PR #5289, opened 2021-09-16
libobs: Add CSPRNG functions

notr1ch:
### Description
Adds three new functions to libobs, backed by the operating system's cryptographically secure pseudo-random number generator (RtlGenRandom on Windows, getrandom(2) on *nix, /dev/urandom on macOS. The output of these functions should be suitable for using in secrets such as encryption keys, tokens, etc.

`os_random_bytes` is added as part of the platform files and covers platform-specific generation of random numbers. A new `libobs/util/rng.{c,h}` file contains helper functions to turn the random bytes into more useful things such as a bounded `uint64_t`.

#### New Functions
`void os_random_bytes(void *buffer, uint32_t length)`
Generate raw random bytes into a buffer.

`uint64_t random_uint64(void)`
Generate a random uint64_t.

`uint64_t random_uint64_bounded(uint64_t min, uint64_t max)` 
Generate a bounded random uint64_t, avoiding modulo bias.

### Motivation and Context
We need a secure random number generator to generate things like state parameters for OAuth flows. Ideally this would be `QRandomGenerator` where needed, but that requires Qt >= 5.10 which is not available for all our supported platforms.

### How Has This Been Tested?
Only tested on Windows. Very basic statistical tests on the output of various ranges to check for bias. Needs testing on other platforms.

### Types of changes

### Checklist:

### DIFF
--- a/libobs/util/platform-nix.c
+++ b/libobs/util/platform-nix.c
@@ -33,6 +33,7 @@
 #if !defined(__APPLE__)
 #include <sys/times.h>
 #include <sys/wait.h>
+#include <sys/random.h>
 #include <libgen.h>
 #if defined(__FreeBSD__) || defined(__OpenBSD__)
 #include <sys/param.h>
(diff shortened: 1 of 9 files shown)


libobs/util/platform-nix.c
```diff
@@ -1033,3 +1034,32 @@ uint64_t os_get_free_disk_space(const char *dir)
 
 	return (uint64_t)info.f_frsize * (uint64_t)info.f_bavail;
 }
+
+#if !defined(__APPLE__)
+void os_random_bytes(void *buffer, uint32_t length)
+{
+	getrandom(buffer, length, 0);
```
kkartaltepe, line 1041: `getrandom` should also be called in a loop like `fread` as its not guaranteed to read `length` bytes in one call unless `length` < 256

notr1ch: Good point. It's unlikely any consumers will need to generate such a large amount of randomness but it should be handled properly. I'll get this fixed.

derrod:
Now that we're at Qt 6 and our oldest supported platform (Ubuntu 20.04) is on Qt 5.12 should this be closed or is this still something we might want?

notr1ch:
It might still be useful for plugins or other code that doesn't use Qt. But for our current use-case, the Qt support should be enough.

PatTheMav:
@notr1ch I guess we can close this for now? If we come across a plugin that will need it, we probably have to revisit or reimplement it anyway as some acknowledged issues haven't been addressed anyway.

PR #5292, opened 2021-09-17
win-capture: Fix UI block caused by game title

walker-WSH:
### Description

get_window_title(...) is called from game_capture_tick(...) which is not in UI thread.
That will make GetWindowTextLengthW blocked sometimes.
GetWindowTextLengthW should be called from UI thread.

### Motivation and Context

When this function is blocked in video-tick, some critial locks can't be released which can cause UI frozen.

### How Has This Been Tested?

I am using another worker thread to reproduce this issue.
In worker thread, enum all windows frequently and call GetWindowTextLengthW for every HWND.
Then, GetWindowTextLengthW  will be blocked soon.

### Types of changes

### Checklist:

### DIFF
--- a/plugins/win-capture/window-helpers.c
+++ b/plugins/win-capture/window-helpers.c
@@ -106,17 +106,13 @@ fail:
 
 void get_window_title(struct dstr *name, HWND hwnd)
 {
-	wchar_t *temp;
-	int len;
-
-	len = GetWindowTextLengthW(hwnd);
-	if (!len)
-		return;
-
-	temp = malloc(sizeof(wchar_t) * (len + 1));
-	if (GetWindowTextW(hwnd, temp, len + 1))
-		dstr_from_wcs(name, temp);
-	free(temp);
+	wchar_t name_temp[MAX_PATH] = {0};
+	ULONG_PTR copy_count = 0;
+	LRESULT res = SendMessageTimeoutW(hwnd, WM_GETTEXT, MAX_PATH, name_temp,
+					  SMTO_ABORTIFHUNG, 2000, &copy_count);
+	if (0 != res && copy_count > 0) {
+		dstr_from_wcs(name, name_temp);
+	}
 }
 
 void get_window_class(struct dstr *class, HWND hwnd)


notr1ch:
Can you elaborate on the situation that leads to blocking? GetWindowTextLengthW should only block if a message is sent to an unresponsive window, it shouldn't matter which thread it is called from. Perhaps you have an unresponsive window on your system?

jpark37:
> GetWindowTextLengthW should be called from UI thread.

The window involved belongs to another process, so I don't think the OBS UI thread matters.

Sending WM_GETTEXT is another process's window seems inappropriate. See: https://devblogs.microsoft.com/oldnewthing/20030821-00/?p=42833
> if you are trying to get text from a window in another process, then GetWindowText() will not send a message; it will just retrieve the string from the “special place”. Since the most common reason for getting text from a window in another process is to get the title of the frame, and since frame windows typically do not do custom window text manipulation, this usually gets the right string.

walker-WSH:
> Can you elaborate on the situation that leads to blocking? GetWindowTextLengthW should only block if a message is sent to an unresponsive window, it shouldn't matter which thread it is called from. Perhaps you have an unresponsive window on your system?

The blocked event on my PC will not always happen with same HWND.
When I invoke that API frequently in non-UI thread, block event will happen sometimes. 

Besides, I find so many users blocked here while finding mathched game window in video-tick thread.

walker-WSH:
> > GetWindowTextLengthW should be called from UI thread.
> 
> The window involved belongs to another process, so I don't think the OBS UI thread matters.
> 
> Sending WM_GETTEXT to another process's window seems inappropriate. See: https://devblogs.microsoft.com/oldnewthing/20030821-00/?p=42833
> 
> > if you are trying to get text from a window in another process, then GetWindowText() will not send a message; it will just retrieve the string from the “special place”. Since the most common reason for getting text from a window in another process is to get the title of the frame, and since frame windows typically do not do custom window text manipulation, this usually gets the right string.

Yes, the block reason should be destinate process.
As the link it is really inappropriate.

But I find so many users blocked here while finding mathched game window in video-tick thread.
Maybe other one has the correct solution.

PR #5294, opened 2021-09-17
linux-capture: Lookup session handle without typechecks

GeorgesStavracas:
### Description

Retrieve the session handle for the PipeWire screencast without performing strict type checks.

### Motivation and Context

g_variant_lookup() obligatorily receives the type of the variant to lookup. This function is used when retrieving the session handle from the portal's response, and the variant type passed is "s" (a string).

However, xdg-desktop-portal had a bug: the documentation explicitly mentions that the session handle is an object path (of variant type "o"), but it passed a string (of variant type "s"). This mismatch [was fixed](https://github.com/flatpak/xdg-desktop-portal/pull/609) in the xdg-desktop-portal release 1.10, but that broke the PipeWire capture code, which was passing specifically the "s" value to the variant lookup.

Fix this by not checking the variant type at all. Object paths ("o") are simply strings with a few extra checks, and we don't actually need to perform these checks.

This change probably broke other apps, and that makes me extremely sad :(

### How Has This Been Tested?

 - Run OBS Studio with xdg-desktop-portal >= 1.10
 - PipeWire screencast still works
 - Run OBS Studio with xdg-desktop-portal < 1.10
 - PipeWire screencast still works too

### Types of changes

### Checklist:

### DIFF
--- a/plugins/linux-capture/pipewire.c
+++ b/plugins/linux-capture/pipewire.c
@@ -958,6 +958,7 @@ static void on_create_session_response_received_cb(
 	UNUSED_PARAMETER(interface_name);
 	UNUSED_PARAMETER(signal_name);
 
+	g_autoptr(GVariant) session_handle_variant = NULL;
 	g_autoptr(GVariant) result = NULL;
 	struct dbus_call_data *call = user_data;
 	obs_pipewire_data *obs_pw = call->obs_pw;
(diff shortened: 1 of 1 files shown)


columbarius:
@GeorgesStavracas: did you check if obs could just use the variant type 'o' in pre 1.10 xdg-desktop-portal? xdpw uses the variant type 'o' for session_handles and it works with xdg-desktop-portal 1.8.

GeorgesStavracas:
The current code works against all versions of xdg-desktop-portal, we don't need to perform any additional checks on OBS side because what is sent must be validated by D-Bus already.

PR #5307, opened 2021-09-20
libobs: move GetRemoteFile to libobs

paulgoulpie:
### Description

Hello,

This patch proposal makes it possible to reuse the following internal OBS functions from a plugin: 
- GetRemoteFile (sending an HTTP request)

### Motivation and Context

These functions being already implemented in OBS and can be considered as utilities, it seems relevant to me to share them for plugin development.
This makes it possible to reduce the weight and the dependency constraints during the development of plugins.
This makes it possible to locate the maintenance efforts on the OBS side.
This contributes to the robustness of plugin development.
This alleviates the porting directives when creating a plugin.

### How Has This Been Tested?

This patch has been tested on the Nudgis plugin from the Ubicast company which is under development.

### Types of changes

The GetRemoteFile function (declaration and implementation) has been moved from internal obs code to libobs.

### Checklist:

### DIFF
--- a/UI/remote-text.hpp
+++ b/UI/remote-text.hpp
@@ -62,11 +62,3 @@ public:
 	{
 	}
 };
-
-bool GetRemoteFile(
-	const char *url, std::string &str, std::string &error,
-	long *responseCode = nullptr, const char *contentType = nullptr,
-	std::string request_type = "", const char *postData = nullptr,
-	std::vector<std::string> extraHeaders = std::vector<std::string>(),
-	std::string *signature = nullptr, int timeoutSec = 0,
-	bool fail_on_error = true, int postDataSize = 0);
--- a/libobs/CMakeLists.txt
+++ b/libobs/CMakeLists.txt
@@ -1,5 +1,9 @@
 project(libobs)
 
+find_package(Libcurl REQUIRED)
+include_directories(${LIBCURL_INCLUDE_DIRS})
+add_definitions(${LIBCURL_DEFINITIONS})
+
 find_package(Threads REQUIRED)
 
 find_package(FFmpeg REQUIRED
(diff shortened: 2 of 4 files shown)


RytoEX:
As I understand it, libobs is strictly C and not C++.  I'll let others weigh in on this, but I don't think this change will be accepted as-is.

tt2468:
As stated by Jim in the initial commit: 
```
      The API was poorly designed because most of it came after I originally
      wrote the application, it was more of an afterthought, and plugin API
      would routinely break for plugin developers due to changing C++
      interfaces (one of the reasons the core is now C).
```

paulgoulpie:
I understand, in this case I can suggest that you rewrite the whole in pure C code. What do you think?

jp9000:
Hi, I understand that @tt2468 didn't respond in the most tactful way (which @tt2468 probably realizes now), but I don't think he meant to be rude. No one asked you to rewrite anything. Please don't respond in that sort of way to anyone again in our community -- getting code merged relies on people cooperating with one another.

That being said, I agree that it is definitely annoying to set up libcurl in plugins. I might consider adding something to help, although I want minimize C++ functions in the backend. But especially in this case, because exporting C++ also makes language interoperability more difficult. C plugins would not have access to this function for example. If you're making a plugin for obs, you're always going to have access to libcurl. My recommendation would just be to temporarily make a copy of this convenience function into your plugin for the time being and just directly link with libcurl in your CMake (which we include with our dependencies).

paulgoulpie:
Hello, I am very sorry for my language error. I absolutely did not interpret @ tt2468's post as rude, and I did not wish to formulate a response that could be considered aggressive.
Far from me this thought and this will.
I made a typo (I don't speak native English) in the sentence 
“I can suggest that you rewrite”, I meant that I was proposing to rewrite the libcurl part myself.
Sorry again for my wording error.
My sincere, friendly greetings to everyone who contributes to OBS.

PR #5308, opened 2021-09-20
UI: Fix YouTube remember settings issues

cg2121:
### Description
-The remember checkbox wouldn't be checked if remember settings
was enabled when loading the YouTube broadcast dialog

-If the user would click remember settings one time and not
the next, the dialog would still load the remembered settings
from the first time, instead of defaults

### Motivation and Context
Fixes bugs that I have noticed.

### How Has This Been Tested?
Created broadcast with saved settings enabled/disabled, to make sure everything worked as expected.

### Types of changes

### Checklist:

### DIFF
--- a/UI/window-youtube-actions.cpp
+++ b/UI/window-youtube-actions.cpp
@@ -704,13 +704,19 @@ void OBSYoutubeActions::UiToBroadcast(BroadcastDescription &broadcast)
 	broadcast.projection = ui->check360Video->isChecked() ? "360"
 							      : "rectangular";
 
-	if (ui->checkRememberSettings->isChecked())
-		SaveSettings(broadcast);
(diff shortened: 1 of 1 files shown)


derrod:
> The remember checkbox wouldn't be checked if remember settings was enabled when loading the YouTube broadcast dialog

This was intentional, so that the user can essentially save new defaults without overwriting them every time (unless desired).

> If the user would click remember settings one time and not the next, the dialog would still load the remembered settings from the first time, instead of defaults

I didn't really know a good way to implement a "reset to defaults" option, but figured that it's probably not a big deal since the defaults aren't great and users probably want to define their own anyway.

cg2121:
Closing, as @derrod explained, these are intentional.

PR #5317, opened 2021-09-21
Fix README.rst

snxx-lppxx:


derrod:
Aside from the unnecessary second commit, I believe this is intentional and does not need to be fixed.

Issue #5314, opened 2021-09-20
"Error uninhibiting", Hangs for 25s when stopping output


Frogging101:
### Operating System Info

Ubuntu 20.04

### OBS Studio Version

Git

### OBS Studio Log URL

https://obsproject.com/logs/CeTN4PzJep6wRMfc

### Expected Behavior

Recording/streaming should stop immediately.

### Current Behavior

When I stop a recording or stream, OBS' UI is unresponsive for 25 seconds. In the log file, I see:

```
Error uninhibiting: GDBus.Error:org.gtk.GDBus.UnmappedGError.Quark._g_2dio_2derror_2dquark.Code24: Timeout was reached
```

### Steps to Reproduce

1. Start a recording or stream
2. Stop the recording or stream

### Anything else we should know?

I compiled OBS myself from today's master branch. The options I used were:
`cmake -DUNIX_STRUCTURE=1 -DENABLE_PIPEWIRE=OFF -DENABLE_WAYLAND=OFF -DBUILD_BROWSER=ON -DCEF_ROOT_DIR=[...]`

I am using Kubuntu 20.04 on an X11 session.

I attached gdb to the obs process while it was stuck. Here is an excerpt of the stack trace:
```
#0  0x00007fd6068e6aff in __GI___poll (fds=0x5587f5742b40, nfds=1, timeout=25000) at ../sysdeps/unix/sysv/linux/poll.c:29
#1  0x00007fd605a5036e in  () at /lib/x86_64-linux-gnu/libglib-2.0.so.0
#2  0x00007fd605a506f3 in g_main_loop_run () at /lib/x86_64-linux-gnu/libglib-2.0.so.0
#3  0x00007fd601d345b5 in g_dbus_connection_send_message_with_reply_sync () at /lib/x86_64-linux-gnu/libgio-2.0.so.0
#4  0x00007fd601d349af in  () at /lib/x86_64-linux-gnu/libgio-2.0.so.0
#5  0x00007fd601d369a7 in g_dbus_connection_call_sync () at /lib/x86_64-linux-gnu/libgio-2.0.so.0
#6  0x00007fd6078b202b in do_uninhibit (info=0x5587f2c562e0)
    at /home/john/gits/obs-studio/libobs/util/platform-nix-portal.c:138
#7  portal_inhibit (info=0x5587f2c562e0, reason=<optimized out>, active=active@entry=false)
    at /home/john/gits/obs-studio/libobs/util/platform-nix-portal.c:204
[...]

PR #5319, opened 2021-09-21
libobs: Make portal inhibitor asynchronous

GeorgesStavracas:
### Description

Make the portal inhibitor asynchronous, by using g_dbus_connection_call() instead of g_dbus_connection_call_sync(). If an uninhibit call is made before the previous inhibit call finishes, cancel the inhibit call instead.

### Motivation and Context

Currently, libobs contains two inhibitors: the portal-based one, and the regular one based on each desktop environment's session manager. The portal inhibitor takes precedence over the session manager one, when available.

Like the session manager inhibitor, the portal inhibitor performs all D-Bus calls synchronously in the calling thread, which can lead to stalls. This is not a good practice.

Fixes obsproject/obs-studio#5314

### How Has This Been Tested?

 - Run `dbus-monitor` in a separate terminal
 - Open OBS and start and stop recording / streaming
 - Notice that the screen is inhibited and uninhibited properly, like before

### Types of changes

### Checklist:

### DIFF
--- a/libobs/util/platform-nix-portal.c
+++ b/libobs/util/platform-nix-portal.c
@@ -24,6 +24,7 @@
 
 struct portal_inhibit_info {
 	GDBusConnection *c;
+	GCancellable *cancellable;
 	unsigned int signal_id;
 	char *sender_name;
 	char *request_path;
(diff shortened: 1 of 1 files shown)


libobs/util/platform-nix-portal.c
```diff
@@ -117,41 +138,58 @@ static void do_inhibit(struct portal_inhibit_info *info, const char *reason)
...
 }
 
+static void do_uninhibit(struct portal_inhibit_info *info)
+{
+	if (info->cancellable) {
+		/* If uninhibit is called before the inhibit call is finished,
+		 * cancel it instead.
+		 */
+		g_cancellable_cancel(info->cancellable);
+		g_clear_object(&info->cancellable);
```
Frogging101, line 172: Is it okay to clear this both here and in portal_inhibit_info_destroy?

Frogging101 (approved):

libobs/util/platform-nix-portal.c
GeorgesStavracas: It is. `g_clear_*()` macros are handy precisely because they perform NULL checks :)

Frogging101:
Excellent!

Sep 21, 2021 13:52:07 Georges Basile Stavracas Neto ***@***.***>:

PR #5322, opened 2021-09-22
UI: Prevent Restream OAuth disconnection

### LINKED ISSUE #5290: Constantly required to "Connect" to Restream.io.  Won't "Keep" Restream.io connection
### Operating System Info

Windows 10

### Other OS

_No response_

### OBS Studio Version

27.1.0-rc2

_No response_

### OBS Studio Log URL

https://obsproject.com/logs/vSMtl383oC4vEiDb

### OBS Studio Crash Log URL

Not Applicable

### Expected Behavior

After I have properly "Settings -> Stream" "Connect Account (recommended)" to Restream.io, then if I go to "Stream -> Settings" page, it should then show that I am "Connected" to Restream.io.  Should not matter whether I use "Server -> Autodetect" or "Server -> Dallas, TX", it should show that I am still connected to Restream.io.

### Current Behavior

[OBS Latest Log](https://obsproject.com/logs/vSMtl383oC4vEiDb)
[[PrintScrn of Settings -> Stream](https://www.flickr.com/photos/garyberth/51484415818/in/datetaken-public/)]

OBS Studio constantly requires "reconnection".  Worth noting: Even though it displays the "Connect Account" selection available, it appears that it is actually connected to Restream.io regardless.

### Steps to Reproduce

1. Settings -> Stream
2. Select "Restream.io" from the "Service" drop down pick item list
3. Select "Autodetect" or "Dallas, Texas" from the "Server" drop down pick list
4. Click on "Connect Account (recommended)"
5. Provide a valid Restream.io account email.
6. Provide a valid password for the Restream.io account email.
7. Click "OK"
8. Click "Settings -> Stream"
9. Now with "Restream.io" selected as "Service", the "Connect Account (recommended)" reappears.

Usually, after Step No. 9 above, the "Settings -> Stream" page will show that there is already a "Connected" to the "Service" Restream.io.  That's how it always worked in previous versions of OBS Studio.  Might be worth noting that, although the "Connect Account (recommended)" appears on Step 9 above, the connection to Restream.io has been saved and it works fine with streaming through Restream.io.  It just appears confusing as the "Connect Account (recommended)" appears to make a person think Restream.io isn't connected.

RytoEX:
### Description

Fix the service check for services where its name in the UI does not match its Auth::Def service string.

In Restream's case, the service name in the UI and rtmp-services is "Restream.io" while the service name in its Auth::Def is "Restream".  This mismatch causes the `service_check` bool to be false, whereas the previous condition (`!!main->auth && service.find(main->auth->service()) != std::string::npos`) would have evaluated to true.

This was broken in commit e6f1daab8c64aa4cd57c7615647ad80362d72d72.

Fixes #5290.

### Motivation and Context

We want service integrations to work.

### How Has This Been Tested?

Compiled and tested on Windows 10.  Connected to accounts for Twitch, YouTube, and Restream and checked against the STR in #5290.

### Types of changes

### Checklist:

### DIFF
--- a/UI/window-basic-settings-stream.cpp
+++ b/UI/window-basic-settings-stream.cpp
@@ -481,7 +481,8 @@ void OBSBasicSettings::on_service_currentIndexChanged(int)
 	}
 
 	auto system_auth_service = main->auth->service();
-	bool service_check = service == system_auth_service;
+	bool service_check = service.find(system_auth_service) !=
+			     std::string::npos;
 #if YOUTUBE_ENABLED
 	service_check = service_check ? service_check
 				      : IsYouTubeService(system_auth_service) &&
--- a/UI/window-basic-auto-config.cpp
+++ b/UI/window-basic-auto-config.cpp
@@ -686,7 +686,8 @@ void AutoConfigStreamPage::ServiceChanged()
 
 	if (main->auth) {
 		auto system_auth_service = main->auth->service();
-		bool service_check = service == system_auth_service;
+		bool service_check = service.find(system_auth_service) !=
+				     std::string::npos;
 #if YOUTUBE_ENABLED
 		service_check =
 			service_check ? service_check


RytoEX:
Alternatively, we can change the `services_check` bool assignment to something like this:
```
	bool service_check = !!main->auth &&
			service.find(main->auth->service()) !=
			std::string::npos;
```

And drop the second assignment.

UI/window-basic-settings-stream.cpp
```diff
@@ -482,6 +482,10 @@ void OBSBasicSettings::on_service_currentIndexChanged(int)
 
 	auto system_auth_service = main->auth->service();
 	bool service_check = service == system_auth_service;
+	if (!service_check)
+		service_check = !!main->auth &&
```
derrod, line 486: The null check here is superfluous, there already is one a few lines above with an early return. You could also replace the simple comparison in the line above with the `.find() != ...npos` check.

This change might also need to be reflected in the Auto-Configuration wizard.

derrod (changes requested):

UI/window-basic-settings-stream.cpp
RytoEX: True.  Fixing it now.  And yes, this does also need changed in auto-config, which I'll also fix.

RytoEX:
Addressed the concerns brought up in review

derrod (approved):

Issue #5325, opened 2021-09-23
libdrm/drm_fourcc.h is missing while building obs-studio on fedora 33


aquisx:
### Operating System Info

Other

### Other OS

Fedora 33

### OBS Studio Version

27.1.0-rc3

### OBS Studio Log URL

-

### Expected Behavior

Building OBS from Source with the following install instructions:
https://obsproject.com/wiki/Install-Instructions#fedora-build-directions 

### Current Behavior

Getting the following error:
```
/home/user/Downloads/obsnew/obs-studio/plugins/linux-capture/pipewire.c:33:10: schwerwiegender Fehler: libdrm/drm_fourcc.h: Datei oder Verzeichnis nicht gefunden
   33 | #include <libdrm/drm_fourcc.h>
      |          ^~~~~~~~~~~~~~~~~~~~~
Kompilierung beendet.
make[2]: *** [plugins/linux-capture/CMakeFiles/linux-capture.dir/build.make:188: plugins/linux-capture/CMakeFiles/linux-capture.dir/pipewire.c.o] Fehler 1

```

### Steps to Reproduce

```
sudo dnf install \
        cmake \
        ffmpeg-devel \
        fontconfig-devel \
        freetype-devel \
        gcc \
        gcc-c++ \
        gcc-objc \
        git \
        glib2-devel \
        libcurl-devel \
        libglvnd-devel \
        libv4l-devel \
        libX11-devel \
        libXcomposite-devel \
        libXdamage \
        libXinerama-devel \
        luajit-devel \
        make \
        mbedtls-devel \
        pipewire-devel \
        pulseaudio-libs-devel \
        python3-devel \
        qt5-qtbase-devel \
        qt5-qtbase-private-devel \
        qt5-qtsvg-devel \
        qt5-qtwayland-devel \
        qt5-qtx11extras-devel \
        speexdsp-devel \
        swig \
        systemd-devel \
        wayland-devel \
        x264-devel
```

```
wget https://cdn-fastly.obsproject.com/downloads/cef_binary_4280_linux64.tar.bz2
  tar -xjf ./cef_binary_4280_linux64.tar.bz2
  git clone --recursive https://github.com/obsproject/obs-studio.git
  cd obs-studio
  mkdir build && cd build
  cmake -DUNIX_STRUCTURE=1 -DBUILD_BROWSER=ON -DCEF_ROOT_DIR="../../cef_binary_4280_linux64" ..
  make -j$(nproc)
  sudo make install
[...]

PR #5326, opened 2021-09-23
linux-capture: Add libdrm dependency check

kkartaltepe:
### Description

We require libdrm for its header so add the cmake module and header path
to the build. We don't need to link libdrm though so we dont add it to
libraries.

### Motivation and Context

Some users might fail to build due to missing deps if they dont have this library installed. See #5325

### How Has This Been Tested?

Compiles and checks the module on my machine.

### Types of changes

### Checklist:

### DIFF
--- a/cmake/Modules/FindLibdrm.cmake
+++ b/cmake/Modules/FindLibdrm.cmake
@@ -1,0 +1,33 @@
+# Once done these will be defined:
+#
+#  LIBDRM_FOUND
+#  LIBDRM_INCLUDE_DIRS
+#  LIBDRM_LIBRARIES
+
+find_package(PkgConfig QUIET)
+if (PKG_CONFIG_FOUND)
+	pkg_check_modules(_LIBDRM QUIET libdrm)
+endif()
+
+find_path(LIBDRM_INCLUDE_DIR
+	NAMES libdrm/drm_fourcc.h
+	HINTS
+		${_LIBDRM_INCLUDE_DIRS}
+	PATHS
+		/usr/include /usr/local/include /opt/local/include)
+
+find_library(LIBDRM_LIB
+	NAMES drm libdrm
+	HINTS
+		${_LIBDRM_LIBRARY_DIRS}
+	PATHS
+		/usr/lib /usr/local/lib /opt/local/lib)
+
+include(FindPackageHandleStandardArgs)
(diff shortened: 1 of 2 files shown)


GeorgesStavracas (approved):
I thought this would be pulled by PipeWire's pkgconf, but I was wrong. Thank you for fixing it!

GeorgesStavracas:
The Flatpak job is also happy with this change :+1:

PatTheMav:
Moving forward, could we please not merge PRs that add/change build-level stuff that uses outdated CMake patterns?

The finder doesn't create proper targets for the library, like so:

```
  if(NOT TARGET Libdrm::Libdrm)
    if(IS_ABSOLUTE "${LIBDRM_LIBRARIES}")
      add_library(Libdrm::Libdrm UNKNOWN IMPORTED)
      set_target_properties(Libdrm::Libdrm PROPERTIES IMPORTED_LOCATION
                                                      "${LIBDRM_LIBRARIES}")
    else()
      add_library(Libdrm::Libdrm INTERFACE IMPORTED)
      set_target_properties(Libdrm::Libdrm PROPERTIES IMPORTED_LIBNAME
                                                      "${LIBDRM_LIBRARIES}")
    endif()

    set_target_properties(
      Libdrm::Libdrm PROPERTIES INTERFACE_INCLUDE_DIRECTORIES
                                "${LIBDRM_INCLUDE_DIRS}")
  endif()
```

The library then needs to be added using `target_link_librares(linux-capture PRIVATE Libdrm::Libdrm)` (though the target can probably be used instead of the variable as-is, linux-capture annoyingly uses _both_ `target_link_libraries` and `add_library`. Both files are also not formatted using `cmake-format`.

CMake scripts are code. Treat it with the same care.

kkartaltepe:
Libdrm the library is not a dependency and very much should not be linked.

PatTheMav:
Linking an interface library just adds the interface include directory to the target. Thus `target_link_library` is the correct way with interface libraries.

kkartaltepe:
Got it, I hope you are not too frustrated but I didnt want to dig in your PR to figure out how to do this and assumed you would know better :)

PatTheMav:
Not at all - just trying to pave the way to good practices early. I already adopted the change. 👍🏻

PR #5330, opened 2021-09-25
linux-capture: Fail when libdrm missing

kkartaltepe:
### Description

I forgot to extend the FATAL_ERROR messages to include libdrm.

### Motivation and Context

The build should fail like it does with the other required components. Also take the time to re-add the quiet that I failed to commit in my last commit.

### How Has This Been Tested?

Ruining my filesystem worked to fail the build.

### Types of changes

### Checklist:


GeorgesStavracas (approved):
:+1: from my side. From Discord:

> PatTheMav: looks fine to me, uses the existing paradigm, makes sense

So seems like it's good to go :slightly_smiling_face:

PR #5218, opened 2021-08-29
win-wasapi: Add support for capturing a process

jpark37:
### Description
Use new Windows API for capturing audio from a process.

### Motivation and Context
Users want it.

### How Has This Been Tested?
Not extensively tested, but seems to work. Tested input/output/process, and fixed various issues.

### Types of changes
- New feature (non-breaking change which adds functionality)

### Checklist:
- [x] My code has been run through [clang-format](https://github.com/obsproject/obs-studio/blob/master/.clang-format).
- [x] I have read the [**contributing** document](https://github.com/obsproject/obs-studio/blob/master/CONTRIBUTING.rst).
- [x] My code is not on the master branch.
- [x] The code has been tested.
- [x] All commit messages are properly formatted and commits squashed where appropriate.
- [x] I have included updates to all appropriate documentation.

PR #5335, opened 2021-09-27
libobs/media-io: Register audio thread with MMCSS

jpark37:
### Description
Ensure audio gets more priority to help prevent glitching.

### Motivation and Context
MS samples old and new seem to be using MMCSS, so we probably should too. Also someone reported glitching with the new process loopback audio capture, so let's try to make sure we meet audio deadlines as best we can.

### How Has This Been Tested?
Audio has been verified to work in recording, thread priority was increased in Process Explorer, and the API calls were inspected in debugger to return success. I don't have any glitch/load tests though.

### Types of changes

### Checklist:

### DIFF
--- a/libobs/CMakeLists.txt
+++ b/libobs/CMakeLists.txt
@@ -120,7 +120,7 @@ if(WIN32)
 	set(libobs_audio_monitoring_HEADERS
 		audio-monitoring/win32/wasapi-output.h
 		)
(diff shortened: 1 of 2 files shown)


RytoEX:
Is this specifically or largely for #5218?

jpark37:
It's probably good to have either way although it's hard to say how much audio glitching OBS experiences of its own fault.

PR #5342, opened 2021-09-29
UI: Adjust minimum size of source toolbar smaller

### LINKED ISSUE #5343: Can't resize to width smaller than 800px
### Operating System Info

Windows 10

### Other OS

_No response_

### OBS Studio Version

27.1.1

_No response_

### OBS Studio Log URL

https://example.com

### OBS Studio Crash Log URL

_No response_

### Expected Behavior

Obs should be able to resize to a width of 636px like it did in the previous version.

### Current Behavior

Doesn't let me resize to smaller than 800px of width.

### Steps to Reproduce

1. Put  Scenes, Controls, Sources, Audio Mixer and Scene transitions in 2 columns under the scene preview.
2. Deselect any source that makes anything appear after Filters under the scene preview.
3. Resize the window hotizontally until you can't anymore.
...

This was working before I upgraded to 27.1.1. I use a monitor of 2560x1080 resolution.

Warchamp7:
### Description
#5187 added a minimum size enforcement on the source toolbar to ensure it remains functional at lower window sizes.

That minimum size was set to 800 pixels, which causes issues for users on 1366x768 monitors in a vertical orientation.

This is still a rather common resolution, so this reduces the minimum size below 768

image
_740px wide plus border width_

### Motivation and Context
Fixes OBS not fitting on 768px wide screens with the source toolbar enabled

### How Has This Been Tested?
Checked all source types to ensure toolbar controls were at least mildly useable at minimum width.

Checked all themes displayed correctly.

### Types of changes

### Checklist:

### DIFF
--- a/UI/forms/OBSBasic.ui
+++ b/UI/forms/OBSBasic.ui
@@ -205,7 +205,7 @@
       </property>
       <property name="minimumSize">
        <size>
-        <width>800</width>
+        <width>740</width>
         <height>30</height>
        </size>
       </property>
--- a/UI/forms/source-toolbar/game-capture-toolbar.ui
+++ b/UI/forms/source-toolbar/game-capture-toolbar.ui
@@ -70,7 +70,7 @@
      </property>
      <property name="minimumSize">
       <size>
-       <width>150</width>
+       <width>120</width>
        <height>22</height>
       </size>
      </property>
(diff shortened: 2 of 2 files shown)


notr1ch:
I'm beginning to think we shouldn't enforce a minimum size for this. Sure users can get it into an ugly state, but losing control over the size of the preview is a worse outcome IMO. Perhaps at lower widths it could hide the source name and only show the icon?

Issue #5364, opened 2021-10-01
Trying to open the auto configuration wizard with a linked YouTube account will crash OBS


Gol-D-Ace:
### Operating System Info

Windows 10

### OBS Studio Version

27.1.1

### OBS Studio Log URL

https://obsproject.com/logs/33n3wo1PrMZr5YjZ

### OBS Studio Crash Log URL

https://obsproject.com/logs/Sz8yC9seiNmbbG69

### Expected Behavior

Have the auto configuration wizard not Crash  OBS when a YouTube account is linked.

### Current Behavior

OBS will crash when opening the auto configuration wizard with a linked YouTube account.

### Steps to Reproduce

1. Link a YouTube account through the stream settings.
2. Try to open the auto configuration wizard from the tools menu
3. OBS will crash

### Anything else we should know?

Unlinking the YouTube account stops the crashing.
Also doesn't happen with a twitch account for example.

Doesn't happen always but often enough that I would call it reproducible on my system.

PR #5365, opened 2021-10-01
UI: Remove thread from YouTube auto config

notr1ch:
### Description
Removes thread from auto configuration code that was accessing UI objects and causing crashing. Given we immediately start and wait on the thread, it seemed largely pointless anyway.

### Motivation and Context
Fixes illegal Qt behavior that caused a crash. Fixes https://github.com/obsproject/obs-studio/issues/5364.

### How Has This Been Tested?
Ran OBS, auto config loaded when it crashed before.

### Types of changes

### Checklist:


Gol-D-Ace:
Looking good here. All crashing I could produce with the auto config is gone with this.

PR #5367, opened 2021-10-01
UI: Do not disable AutoConf bandwidth test for YouTube

derrod:
### Description

This change removes the disabling of auto-config bandiwdth testing for youtube, as the test pattern does no longer get your account flagged (previously it would result in a community guidelines strike).

It also removed the recursive call that would cause YouTube API request to be executed twice once the bandwidth testing was enabled again.

### Motivation and Context

Makes code a little cleaner, avoid unnecessary API requests.

### How Has This Been Tested?

Made sure stuff still works.

### Types of changes

### Checklist:

### DIFF
--- a/UI/window-basic-auto-config.cpp
+++ b/UI/window-basic-auto-config.cpp
@@ -470,31 +470,20 @@ void AutoConfigStreamPage::OnOAuthStreamKeyConnected()
 			ui->connectedAccountText->setText(
 				QTStr("Auth.LoadingChannel.Title"));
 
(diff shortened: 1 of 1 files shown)


notr1ch:
With the removal of this code, we can probably remove the `bool isYoutube = false;` variable too.

https://github.com/obsproject/obs-studio/blob/41ce5a5b2569c70417a5b7995daee4d352d4669b/UI/window-basic-auto-config.cpp#L722

derrod:
Right, of course

PR #5148, opened 2021-08-18
UI: Remove duplicate missing files code

cg2121:
### Description
The same code was in different places, so call single function to
cleanup code.

### Motivation and Context
Code cleanup.

### How Has This Been Tested?
Tested with missing files to make sure dialog still worked properly.

### Types of changes
- Code cleanup (non-breaking change which makes code smaller or more readable)

### Checklist:
- [x] My code has been run through [clang-format](https://github.com/obsproject/obs-studio/blob/master/.clang-format).
- [x] I have read the [**contributing** document](https://github.com/obsproject/obs-studio/blob/master/CONTRIBUTING.rst).
- [x] My code is not on the master branch.
- [x] The code has been tested.
- [x] All commit messages are properly formatted and commits squashed where appropriate.
- [x] I have included updates to all appropriate documentation.

PR #5401, opened 2021-10-11
UI,libobs,obs-filters: Add missing files check to filters

### LINKED ISSUE #5182: Missing files dosen't include missing 'Image Mask/Blend' filter path
### Operating System Info

Windows 10

### Other OS

_No response_

### OBS Studio Version

27.0.1

_No response_

### OBS Studio Log URL

https://obsproject.com/logs/IG-BJQp7FqeT4Jhe

### OBS Studio Crash Log URL

_No response_

### Expected Behavior

As well as the missing media source files, I expected the mask file to be found in the missing file search.

### Current Behavior

Only missing media source files are found and fixed

### Steps to Reproduce

1. Create a scene collection with media sources in it, and have at least one source with a mask on it that links to a mask image file
2. Export the scene collection and then reimport it (make sure to move all the media files so that when you import it OBS cannot find the files)
3. Open the newly imported scene collection and use the dialog box to find the missing files.
...

_No response_

gxalpha:
### Description

- Filters are not respected by the missing files dialog, even though they are sources (which makes this comparitibly easy to implement).
This adds the missing files check for filters. The libobs change is needed since filter names aren't unique on a global context, so having the parent source name is needed so that the user is be able to identify which filter is meant.

- Having a source with a LUT filter and an image mask filter where the files are missing would look like this in the missing files dialog:
image

- From what I've seen, only the "Apply LUT" and "Image Mask/Blend" filters have files, so those have been included. Please comment in case I've missed one.

- Depends on #5148 (Well, doesn't strickly depend on it, but makes that PR makes this one way cleaner). *Draft until that is merged.*

### Motivation and Context

- Fixes #5182

### How Has This Been Tested?

macOS 12.0 beta 6, compiled and run.
For both filter types, the check shows that the file is missing (see above).
Clicking apply replaces it (if a new one has been selected) in the source properties and refreshes the source.

### Types of changes

### Checklist:

### DIFF
--- a/UI/window-basic-main.cpp
+++ b/UI/window-basic-main.cpp
@@ -4861,14 +4861,15 @@ void OBSBasic::on_actionShowMissingFiles_triggered()
 	};
 	obs_enum_sources(cb_sources, files);
 
-	auto cb_transitions = [](void *data, obs_source_t *source) {
-		if (obs_source_get_type(source) != OBS_SOURCE_TYPE_TRANSITION)
-			return true;
+	auto cb_transitions_filters = [](void *data, obs_source_t *source) {
+		enum obs_source_type type = obs_source_get_type(source);
+		if (type == OBS_SOURCE_TYPE_TRANSITION ||
+		    type == OBS_SOURCE_TYPE_FILTER)
+			AddMissingFiles(data, source);
 
-		AddMissingFiles(data, source);
 		return true;
 	};
-	obs_enum_all_sources(cb_transitions, files);
+	obs_enum_all_sources(cb_transitions_filters, files);
 
 	if (obs_missing_files_count(files) > 0) {
 		/* When loading the missing files dialog on launch, the
--- a/libobs/obs-missing-files.c
+++ b/libobs/obs-missing-files.c
@@ -95,7 +95,16 @@ obs_missing_file_t *obs_missing_file_create(const char *path,
 
 	switch (src_type) {
 	case OBS_MISSING_FILE_SOURCE:
-		file->src_name = bstrdup(obs_source_get_name(src));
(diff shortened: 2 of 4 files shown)


gxalpha:
As #5148 got merged, this PR is now ready for review.

norihiro:
> I'm not sure I get what you mean, could you maybe elaborate on your comment?

My point is that this line actually didn't apply the settings right now but set `defer_update_count`.
https://github.com/obsproject/obs-studio/blob/009fc43bee3781347ef188d71e12a12cd0a9ca9f/plugins/obs-filters/color-grade-filter.c#L426
After that, `obs_source_video_tick` will call `color_grade_filter_update` later.
If setting slow frame rate, `color_grade_filter_missingfiles` will be called earlier than `color_grade_filter_update` so that missing-file cannot be detected.

Example step is as below.
1. Set frame rate slower, such as 10 fps.
2. Prepare `Apply LUT` filter and specify a file.
3. Exit OBS
4. Remove the file pointed by the `Apply LUT` filter.
5. Start OBS

I'd like to suggest one more change as below.

```patch
diff --git a/plugins/obs-filters/color-grade-filter.c b/plugins/obs-filters/color-grade-filter.c
index c5d3ce81b..662c43235 100644
--- a/plugins/obs-filters/color-grade-filter.c
+++ b/plugins/obs-filters/color-grade-filter.c
@@ -423,7 +423,7 @@ static void *color_grade_filter_create(obs_data_t *settings,
 		bzalloc(sizeof(struct lut_filter_data));
 	filter->context = context;
 
-	obs_source_update(context, settings);
+	color_grade_filter_update(filter, settings);
 	return filter;
 }
 
```

gxalpha:
This will have to be completely reimplemented.

PR #5412, opened 2021-10-13
UI: Minimize context bar when too small

gxalpha:
### Description

This change removes the minimum total size from the context bar, instead
removing certain elements when the size is too small. This means that
users will still be able to make OBS smaller in width while keeping the
context bar enabled.

First, the properties, filter and interact buttons will get reduced to
just the symbols, if it gets even smaller the source specific menus are
hidden.

This is how it looks:

https://user-images.githubusercontent.com/59806498/137125517-df9f4abc-b267-49ae-bea4-33ab8b782d7b.mp4

This is a draft for two (three/four) reasons:
- ~~It doesn't yet refresh when docks are popped out.~~
- The first threshold (740) is the current minimum size of the toolbar. However, the second threshold (600) was chosen arbitrarily. I'm not a UX expert, feedback or suggestions on that number would be appreciated.

### Motivation and Context

27.1 introduced a minimum size for the context bar / toolbar, which meant that
users with smaller widths had to disable it.
I think this is a more elegant approach to the problem of small toolbars.

### How Has This Been Tested?

macOS 12.0 beta 6, compiled and run.

For the result, see the video above.

### Types of changes

### Checklist:

### DIFF
--- a/UI/forms/OBSBasic.ui
+++ b/UI/forms/OBSBasic.ui
@@ -205,7 +205,7 @@
       </property>
       <property name="minimumSize">
        <size>
-        <width>740</width>
+        <width>0</width>
         <height>30</height>
        </size>
       </property>
--- a/UI/window-basic-main.cpp
+++ b/UI/window-basic-main.cpp
@@ -301,6 +301,8 @@ OBSBasic::OBSBasic(QWidget *parent)
 
 		if (obs_get_video_info(&ovi))
 			ResizePreview(ovi.base_width, ovi.base_height);
+
+		UpdateContextBarVisibility();
 	};
 
 	connect(windowHandle(), &QWindow::screenChanged, displayResize);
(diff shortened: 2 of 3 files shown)


notr1ch:
There seems to be a frame or two where the labels lose their text before they get resized. This feels a bit janky - is there a way to avoid this?
GGqzIO6 1

Warchamp7 (approved):

flaeri:
hmm, wonder if this will "accidentally" fix https://github.com/obsproject/obs-studio/issues/5560
Remdinder to myself to test it

Azeirah:
> hmm, wonder if this will "accidentally" fix #5560 Remdinder to myself to test it

I was investigating #5560, with the most recent build (including this merge patch) the browser docks don't get squished anymore, so this does fix #5560!

PR #5421, opened 2021-10-16
CI: Add copyright information to macOS bundle plist

gxalpha:
### Description

Adds the [NSHumanReadableCopyright](https://developer.apple.com/documentation/bundleresources/information_property_list/nshumanreadablecopyright) key to the macOS app bundle property list, with the value `© 2012-2021 OBS Project` (the year and name are from the [windows installer](https://github.com/obsproject/obs-studio/blob/5fccf86b075cb66fd30c21e6b86d27dbbaac151a/UI/installer/mp-installer.nsi#L503-L504)).

In the macOS info panel, it now looks like this:
image

Whereas before, it was just this: 
image

### Motivation and Context

I noticed that most other apps have this, but OBS doesn't.

### How Has This Been Tested?

macOS 12.0 beta 10
Bundled using CI/full-build-macos.sh, result see above.

### Types of changes

### Checklist:

### DIFF
--- a/CI/scripts/macos/app/Info.plist
+++ b/CI/scripts/macos/app/Info.plist
@@ -10,6 +10,8 @@
 	<string>obs</string>
 	<key>CFBundleIdentifier</key>
 	<string>com.obsproject.obs-studio</string>
+	<key>NSHumanReadableCopyright</key>
+	<string>(c) 2012-2021 Hugh Bailey</string>
 	<key>CFBundlePackageType</key>
 	<string>APPL</string>
 	<key>LSMinimumSystemVersion</key>


WizardCM:
As someone who doesn't use macOS much, is it common to include the © symbol in this string?

gxalpha:
@WizardCM yes, both Apple apps do so universally (except Chess.app :P), as well as nearly all 3rd-party apps on my machine.
The only difference is whether to use `Copyright ©` or just `©`, not even Apple seems to be sure about that one.

Though if your question was more about instead using `(c)`, the only app on my machine doing that seems to be VLC, everyone else uses the symbol.

PR #5424, opened 2021-10-17
mac-vth264: Add GPU name to hardware encoder names

### LINKED ISSUE #4611: OBS does not correctly label additional [Video Toolbox] GPUs in Encoder list (27.0.0 RC2 macOS)
### Operating System Info

macOS 11.2

### Other OS

_No response_

### OBS Studio Version

27.0.0-rc2

_No response_

### OBS Studio Log URL

N/A

### OBS Studio Crash Log URL

_No response_

### Expected Behavior

OBS should differentiate additional AppleVT Hardware encoders, if available **or** default to an external GPU encoder (if connected).

Previous versions of OBS hardcoded the internal GPU (Intel QuickSync) encoder using Apple Video Toolbox. **Release Candidate 2 for version 27.0.0 remedies this issue**. However...

### Current Behavior

Currently, the eGPU & iGPU encoders are listed identically within the encoder list.
OBS Encoder Menu

### Steps to Reproduce

1. Install OBS 27.0.0 RC2 for Mac
2. Connect an external GPU with AppleVT support to the Mac.
3. Launch OBS
4. Open Settings and navigate to the encoding options under **Output**
5. Open the dropdown list — 2 Apple VT H.264 Hardware encoders will be listed

# Additional Information & Suggestions

When using **[VTCopyVideoEncoderList](https://developer.apple.com/documentation/videotoolbox/1522108-vtcopyvideoencoderlist)** in a simple test application, one can see that the eGPU encoder is additionally identified by both an identifer at the end of its **Encoder ID**, as well as a **GPURegistryID**, which is used when specifying external encoders.

Test Application In Playgrounds

Without obtaining the name of each GPU connected, it should at the very least identify each additional hardware encoder as "Apple VT H.264 Hardware Encoder (External)", identifying it as an external encoder by property of a GPU Registry ID (these are only listed with secondary/external GPUs).

Alternatively, using MetalKit, one can query the available GPUs using **[MTLCopyAllDevices](https://developer.apple.com/documentation/metal/1433367-mtlcopyalldevices)**, iterate through the list, and when GPURegistryID is present in a VTEncoder, search for its matching device by registryID in the list, and label it based on the device name.

**Here is an example program**, runnable via Swift Playgrounds, which will enumerate VTEncoder devices, identify any external hardware devices, and label them based on their Metal device name: **[Code (Gist)](https://gist.github.com/MisutaaUrufu/22206996d21182c8f431506bc0ed458b)**

screenshot

gxalpha:
### Description

Adds the name of the GPU to the hardware encoder if its RegistryID is
exposed by VideoToolbox (which it usually is for dedicated GPUs, but not
integrated ones).

This makes it possible to more easily differentiate between the different
"Apple VT H264 Hardware Encoder"s, which wasn't really possible before.

With a dedicated RX580 and an Intel chip with iGPU, it now looks like this:
Bildschirmfoto 2021-10-18 um 00 27 29

### Motivation and Context

Right now, there really isn't a way to tell the GPUs apart:
image
Fixes #4611

### How Has This Been Tested?

macOS 12.0 beta 10 (Intel), compiled and run.
As you can see, encoders with dedicated GPUs will get shown with the name of the GPU.

Apparently, the Apple T2 chip has encoding capabilities as well. As I don't have one on my Mac, it would be good if someone who does could test this and see whether it gets shown.

### Types of changes

### Checklist:

This PR is split into two commits to keep git from thinking I deleted a file and created a new one.


MisutaaAsriel:
>   If my theory is true, I would argue that while being very strange, this is not on OBS to fix. It could be done (checking if an encoder with the GPURegistryID already exists before registering it in OBS - Keep in mind this could cause peoples setups to break), but I'm not sure it's worth the hassle.

I'd argue this is most appropriate. Setups with dGPUs operate differently from those without, as well as those with a dGPU but no iGPU.

It also should only cause minimal breakage; OBS forgets my GPU on the regular anyhow (iGPU + eGPU), and it would be a one-time fix to go in and reset the encoder settings for those whose are lost. OBS has done similar breaking changes as needed before (e.g. changes to the color correction filter)

Alternatively, you could, in theory, "map" one encoder to the other, so it only shows one option in the drop down, masking the fact there are 2 options internally.

PatTheMav:
So what's the action here? It sounds like there is nothing further we could do to differentiate those values as reported on the Mac Pro and the overall feature is good to go?

gxalpha:
That is my understanding as well.
I didn't notice the conflicts, will fix them up.

MisutaaAsriel:
Is there anything this is waiting on? I've been using a custom build with this for some time, and it's actually **highly preferable** *because* of #6077, as otherwise I *literally* have to change my encoder selection, *set all my custom settings up*, **do a test recording**, *then watch OBS in **Activity Monitor** just to see which GPU VTEncoderServiceXPC spawns on.

With this pull I know **for a fact** I'm at least selecting the right GPU.

PatTheMav:
@gxalpha let's review this after HEVC and ProRes are merged because the encoder source will have changed a lot by then.

gxalpha:
I longer think that this is a good PR.

Given that `kVTVideoEncoderList_GPURegistryID` appears to only be added to eGPU's (and in rare cases dedicated GPU's, but crucially never on Apple Silicon), this PR is only relevant for eGPU machines. It's very clear Apple believes that eGPU's are not the way forward (and thinks the same about Intel), so this PR is effectively already legacy code. I do not wish to add maintenance cost (and waste reviewer's time) by adding legacy code with my PR's.

As such, I'm closing this PR. If someone else wants to pick it up, fix the conflicts, drive the review process and maintain the code into the future that'd be fine by me, but personally I don't think it's a good idea.

PR #5449, opened 2021-10-25
mac-avcapture: Capture audio if supported

### LINKED ISSUE #3661: Video Capture Device does not capture iOS device audio
## Platform

Operating system and version: macOS 11.0 (20A5395g)
OBS Studio version: 26.0.2

## Expected Behavior

When adding an iOS device using a Video Capture source, ***audio output*** for the source should be added within the sound mixing panel of OBS.

## Current Behavior

- Audio is not captured nor controlled by OBS
- As OBS has activated A/V capture, system disables _audio-only pass-through_ for iOS device in **Audio MIDI Setup**.
- As *audio-only pass-through* is disabled, by design, OBS cannot capture the iOS device when adding an Audio Capture source.

## Steps to Reproduce

1. Create a scene.
2. Connect an unlocked, trusted iOS device.
3. Add a Video Capture source in OBS.
4. Choose the iOS device as the source device.
5. Add an Audio Capture source in OBS.
6. Fail to locate the iOS device from step 4 in the drop-down list.

## Workaround
1. Connect an unlocked, trusted iOS device.
2. Open QuickTime.
3. Open a New Movie Recording window.
4. Set the iOS device as both the video and audio capture device.
5. Pass system audio through a virtual audio device, such as Blackhole or Soundflower.
6. Add new Window Capture source in OBS.
7. Set QuickTime as the Window Capture window.
8. Set Desktop Audio device in settings to the virtual audio device from step 5.

## Additional information

macOS disables the separate pass-through audio device, for a connected iOS device, when a program begins A/V capture of said device. The capturing program can access both Audio & Video (e.g. QuickTime) capture, however. This suggests that the method of enumerating iOS audio devices when A/V capture is in progress is different from standard system audio devices.

gxalpha:
### Description

Some webcams, or other AVCaptureDevices like connected iOS devices which
are supported since 162450c, have an audio output which so far got
ignored.
Now, if supported, the audio will be captured alongside the video.

### Motivation and Context

- Fixes #3661

### How Has This Been Tested?

macOS 12.0.1 RC 2, compiled and run.
My with USB connected iPad will correctly output audio to OBS.
My camera, which doesn’t have a microphone, has the audio disabled in OBS (not visible in the mixer).
~~Also there still is some significant audio buffering which I need to figure out. [Seems this was caused by a faulty USB cable, doesn't happen with a new one]~~

While that sounds like an unfinished PR that shouldn’t be opened yet, I’m opening it now since I don’t have a real camera with a microphone (so unfortunately can’t test that) and it would be great if someone who does did. Theoretically it should just work, but it’s better to test before.

### Types of changes

### Checklist:


gxalpha:
Uhhmm, so:
My USB-Cable or something broke and now my iPad outputs weird video to OBS/QuickTime/Anything and shortly after, the video stream completely stops, and I have to reboot :D
This makes debugging very difficult and annoying, and I can't rule out the issues that arose before might have been related to the faulty cable, so for now (until I buy a new cable) I can't really work on this :(

.github/workflows/main.yml
```diff
@@ -5,7 +5,7 @@ on:
     paths-ignore:
       - '**.md'
     branches:
-      - master
+      - '**'
```
PatTheMav, line 8: I guess this was needed for testing on different branches in your fork, but that has no place in a feature PR.

plugins/mac-avcapture/av-capture.mm
```diff
@@ -593,6 +597,52 @@ static inline bool update_frame(av_capture *capture, obs_source_frame *frame,
...
+
+	status = CMSampleBufferGetAudioBufferListWithRetainedBlockBuffer(
+		sample_buffer, nullptr, list, requiredSize, nullptr, nullptr,
+		kCMSampleBufferFlag_AudioBufferList_Assure16ByteAlignment,
+		&buffer);
+
+	if (status != noErr || !list) {
+		if (list)
+			bfree(list);
+		blog(LOG_WARNING,
```
RytoEX, line 626: Either separate the `if` contents and the following line with a blank line or wrap the `if` contents in braces.

RytoEX (changes requested):
One style nitpick.

> In terms of users who have added this source previously, this will enable audio for devices that were previously silent, yes?

I think @WizardCM 's question is: will devices in existing scene collections that didn't produce sound before now suddenly, perhaps _unexpectedly_, begin producing sound?

gxalpha:
> will devices in existing scene collections that didn't produce sound before now suddenly, perhaps unexpectedly, begin producing sound?

Oh yeah. Long story short, I just wrote a paragraph about how there is an issue with audio channels I though I couldn't fix, only to notice I probably can.
My plan now is to add a button "Enable audio" to the properties of existing sources, which would then enable it.

gxalpha:
Added a _v2 of the source with audio enabled by default.
Existing v1 sources will have the following button in the properties, pressing which will enable audio (and also tick all the channels which are disabled by default).
Bildschirmfoto 2022-01-21 um 15 15 46
If the user wants to have audio disabled (again), this should be done through the mixer and not the properties (the button disappears once pressed).

MisutaaUrufu:
Wouldn't a checkbox for "Enable Audio (Supported Devices Only") be more adequate? This is a feature even I myself have been wanting, but I will admit, it would become a nuisance in some cases to be _forced_ to capture audio from select devices. — Audio sources themselves can be added or removed in OBS, independently from muting them in the mixer, so following this behavior by allowing the audio portion to be "toggled" would closer fit the existing UX, yes?

It could still be left enabled by default, but having a checkbox (toggle) both alerts users that the capture device contains audio (as they could be unaware), whilst giving advanced users the ability to micromanage their audio sources, including mixed AV sources.

gxalpha:
> Audio sources themselves can be added or removed in OBS, independently from muting them in the mixer, so following this behavior by allowing the audio portion to be "toggled" would closer fit the existing UX, yes?

I'd consider this behaviour closer to video or game capture sources, which can only be muted through the mixer, which is why I've gone for this approach. The button is only there to make sure existing setups don't break (suddenly play sound).

I'm open to other opinions though, what do other people think about the above?

PatTheMav:
I'm really torn on this. It introduces a behaviour that I find is "wrong" on Windows already, namely too much implicit and magic configuration.

If I add a video source, I only want the **video** source (it's in the name). If I _also_ want the associated audio, I add it explicitly as an audio source. This PR muddles the water and makes a video source also add an audio source and that is bad™ in my book.

_But_ that ship has unfortunately sailed as OBS has too many of those implicit "conveniences" on Windows already.

dodgepong:
Personally I would argue that the existing non-magic Window behavior has always been bad, and video capture devices, despite the source name, ought to always include associated audio unless the user explicitly disables it. The problem instead is that audio is not set to monitor by default, so users might not notice that they have a new fader in the audio mixer with audio levels going up and down since they can't hear it.

MisutaaUrufu:
> I'm really torn on this. It introduces a behaviour that I find is "wrong" on Windows already, namely too much implicit and magic configuration.
> 
> If I add a video source, I only want the **video** source (it's in the name). If I _also_ want the associated audio, I add it explicitly as an audio source. This PR muddles the water and makes a video source also add an audio source and that is bad™ in my book.

To be fair, this change is only necessary to begin with due to how macOS handles AVKit sources.

Most "off the shelf" capture devices present their audio and video separately as sources. OBS made incorrect assumptions that this is how *all* devices behave, and so its activation of audio and video sources treat all inputs as such.

But *this breaks capturing both for **joint** sources*. E.G. the iPhone *does* present separate audio & video sources, *but you cannot activate both*. — macOS disables the audio-only source if a video source is being captured (as audio is being presented through the A/V session).

So the way OBS works (on Mac) has always been "wrong", as it assumed an AVKit video capture wouldn't have accompanying audio source directly.

Personally, I think a checkbox is a good middle ground, short of greatly modifying OBS to either add an audio source when present automatically, or breaking mixed A/V capture into its own source type (So you'd have "Audio Capture Device", "Video Capture Device", and "Media Capture Device").

Honestly, though, this could be pushed for now as an update, and future updates could provide changes to the UX for better clarity and uniformity. There are a lot of improvements to be had on the macOS platform, so any improvement to stability or functionality, even if it isn't principally *the best*, would be better than holding something back for reasons like this.

PatTheMav:
@MisutaaUrufu Good input, thanks!

ArtemBernatskyy:
Guys, I'm sorry, any updates when this will be merged?
Thx! ❤️❤️❤️

ArtemBernatskyy:
Thx, just tested and it works well with Macbook 14 M1 + iPad Pro M1 !

ArtemBernatskyy:
@PatTheMav , I'm sorry, can this be merged? Or what's the status?

I mean for MacOS users with M1 it's becoming more and more tedious to keep track that they need to get obs image from this pull request. And if there will be more fixes to the main branch it will require incorporating changes in to this pull request.

TLDR this pull request is very important for M1 macs which work with iPads/iPhones, thx!!!

gxalpha:
@ArtemBernatskyy Thank you very much for testing this!
Please understand that reviewing PRs does take time and none of the Mac people here are working full time on OBS. Feel free to continue using and testing it, but in the future please don't send messages like "When does this PR get merged?"/"Please merge this" as they just clog up conversation. While it might be very important for you (and I understand that, it's important for myself as well, otherwise I wouldn't have created it), right now it's not the top priority in terms of open PRs. Don't worry, we won't forget about it, and it will get merged when we are confident with it.

Also fyi, this PR does not change anything M1-specific, everything changed here applies to both arm and x86_64.

ArtemBernatskyy:
@gxalpha let's agree to disagree )(
Yes, they clog up the conversations but they also help to prioritize PRs that need to be taken into consideration first.
(I've used to commit to open source at some point in time, so know how important it is to prioritize PRs).

Also, there are situations when such PRs (as this one) will be unmerged for a long time because only a specific subset of the community needs it.

Fenrirthviti:
> @gxalpha let's agree to disagree )( Yes, they clog up the conversations but they also help to prioritize PRs that need to be taken into consideration first. (I've used to commit to open source at some point in time, so know how important it is to prioritize PRs).
> 
> Also, there are situations when such PRs (as this one) will be unmerged for a long time because only a specific subset of the community needs it.

No, they don't help prioritize anything when the peanut gallery constantly asks "when is this going to be merged". They actually take away from time people have to work on things as we have to stop what we're doing and write up replies. You claim to have experience with open source, so you should understand that those kinds of comments are rarely, if ever, helpful. One or two loud people asking for something to be merged for their own specific needs doesn't mean it aligns with the wider priorities of the project.

You've been consistently passive aggressive, rude, and dismissive since you've started commenting on our repo. Consider this a warning to curb the attitude and respect the developers who work on the project. Keep the comments to feedback on testing and the features themselves, please.

MisutaaAsriel:
> Also, there are situations when such PRs (as this one) will be unmerged for a long time because only a specific subset of the community needs it.

This PR fixes a long standing bug with OBS that severely limits its functionality on macOS hardware when interfacing with AVKit hardware. I'd hardly call it something "only a specific subset … needs". — Unless you're simplifying the entire macOS user base.

I would greatly appreciate this PR being merged myself. But if one explores this PR conversation, one would realize there's some concerns as to implementation and consistency with the Windows version, as well as lack of testing.

E.G. I made a very simple PR that fixed a bug with how macOS rendered the dock icon, and it was accepted and merged pretty rapidly once I made adequate corrections to both the commit and the code.

PRs like this are more complex, and OBS has some uphill battles in getting more to participate and aid in macOS development (e.g. bugs in GPU selection rendered some machines to default to the integrated GPU for years, to the point some considered Mac hardware unsuitable for streaming).

It's an important PR. It just needs more time to be polished, lest even more bugs and inconsistencies be introduced.

gxalpha:
To elaborate on the testing still required:
I'd really like to see this tested with a physical webcam that has a microphone built-in to make sure that works as well. Theoretically it should, but it would be great to get confirmation on that if someone happens to have such a camera.

WizardCM:
I've tested with a Logitech C930e, a Logitech C270, a cheap $10 capture card.

Both with the v1 and v2 sources and none show an audio entry automatically. It could be that they're not properly reported to AVKit in the first place, but they do show up separately as audio devices so I'm not too worried if they're not added automatically.

image

MisutaaAsriel:
> Both with the v1 and v2 sources and none show an audio entry automatically. It could be that they're not properly reported to AVKit in the first place, but they do show up separately as audio devices so I'm not too worried if they're not added automatically.

Isn't this outside the scope of this commit anyhow then? Not _all_ devices report as a single device to AVKit, just _some_.

In my experience the _common_ behavior is that devices will report as 2 separate devices (to AVKit): an audio device and a video device. And most software will handle this just fine.

The problem is that _not all_ devices will report this way. Some devices will either show up as a _singular_ device; or as two **mutually exclusive** devices (such as iOS devices), one containing both audio & video.

As far as my knowledge goes, iOS devices are the only to behave this way in the wild. It's possible that some Macintosh-specific hardware also exhibits this behavior, but most devices on the market report as two separate devices, I'm assuming for compatibility, in my experience.

Essentially, this is fixing an oversight. — I'm getting the impression that how iOS devices are handled is what Apple might describe as the _intended behavior_, but due to some long standing technical debt or industrial practice, most devices just report separately for some form of compatibility.

******

Also, I'm not able to test _at the moment_, but after messing with the Windows version of OBS more, I can say this in regards to audio options on sources and consistency in UI design (as this was a concern earlier on with this commit):

**[OBS on PC already provides complex audio capture options on PC](https://obsproject.com/hu/kb/video-capture-sources)**; primarily:
- **Audio Output Mode**: On Windows, you can set to direct capture, if audio exists, in which audio is included as part of this source; or to desktop capture, in which audio (if it exists) is routed to the default desktop audio capture device.
- **Use custom audio device**: Specifies that this source is to incorporate an audio device _other_ than the default for this source.
- **Audio Device**: The audio device linked to the source, if custom audio is enabled.

Thus, **I propose the following changes to maintain consistency:**
1. Add the following options to the macOS Video Capture source properties:
    - Audio Capture (dropdown):
        - None: _No Audio will be captured_ (default **for upgraded sources**)
        - Direct: _Audio will be captured if available, as part of this source_ (default **for new sources**)
    - Use Custom Audio Device (Checkbox):
        - Audio Device (dropdown): the list of Audio (input) Devices currently detected by OBS.
2. Add the ability to capture separate audio sources as part of this source.
    - If Custom Audio is specified, the custom Audio Device specified will be captured as part of this source.
    - If Custom Audio **is NOT** specified, and Audio Capture is set to **Direct**, the AVKit device will be treated as a dual purpose, A/V source, and any integrated audio will be captured.
    - If Audio Capture is set to **None**, audio will not be captured as part of this source, and this source shall not appear in the volume mixer.

WizardCM:
@MisutaaAsriel The audio-related options in the Windows version of the Video Capture Device source are largely legacy and we'd really like to phase them out or remove them entirely, and they're generally quite confusing to users so we don't want to introduce the same options on macOS.

If AVKit's merged audio is only for iOS devices, that is a perfectly OK limitation. If a user wants to add a the audio for a video capture device, they can do it as they do now, via an Audio Input Capture source.

MisutaaAsriel:
> If AVKit's merged audio is only for iOS devices, that is a perfectly OK limitation.

@WizardCM That's the thing: **There's really no way of knowing, one way or another.** AVKit, from an API perspective, allows devices to provide both audio and video through a single capture "session", if you will. "Dual-session" may be the _norm_, and iOS devices may be the _largest_ "single-session" device in play, but that does not guarantee that it'll never become a problem again in the future.

Idealistically changes like this either try to "get things right the first time", or "merge as you go". Towing a middle ground feels rather ineffective at both getting a fix out the door, and making sure it's actually fixed. (For example, a number of commits were made in regards to macOS and window/dock icons over time ___and they were all wrong.___ — The dock icon is to be set by the OS using the PLIST, not by the application, under normal circumstance.)

Being said,

> The audio-related options in the Windows version of the Video Capture Device source are largely legacy and we'd really like to phase them out or remove them entirely, and they're generally quite confusing to users so we don't want to introduce the same options on macOS.

Then I propose this alternative: AVFoundation has the [deviceType struct](https://developer.apple.com/documentation/avfoundation/avcapturedevice/devicetype) and the [DiscoverySession function](https://developer.apple.com/documentation/avfoundation/avcapturedevice/discoverysession).

I recommend **when OBS is enumerating AVKit devices for a capture session, it also maintains a separate dictionary of devices which contain builtInMicrophone**, which includes a link back to the primary list of devices for matches between the two using some sort of ID.

The properties window will have one singular audio property: A dropdown. labeled "Capture Audio", and two options:
- "Automatically"
- "Disabled"

(Using a dropdown instead of a checkbox, as this allows for more interactive explanation of what is happening to the end user).

Then, **if and only if the device presents itself as a `video` device with `builtInMicrophone`,** will OBS present an audio device for that source in the mixer.

I will admit **I do not know if `builtInMicrophone` is the correct feature to search for**, or if devices such as iOS devices present some undocumented feature for their audio. **I will be running a test in Swift Playgrounds to confirm and will write back in a follow up comment.**

MisutaaAsriel:
To append to my previous comment, it would seem my initial hunch is _not_ the correct way to go (whilst general AV devices do appear, **iOS Devices do not**, which changes my understanding of this API.). As such my comment has been amended to reflect this, and the recommendation properly separated.

MisutaaAsriel:
**After a little more experimentation on my own, using XCode this time (as it supports Mac Only code in playgrounds)**, here is my suggestion:

For each AVKit device **check [hasMediaType](https://developer.apple.com/documentation/avfoundation/avcapturedevice/1389487-hasmediatype?language=objc)**, for  a media type of **[muxed](https://developer.apple.com/documentation/avfoundation/avmediatype/1388256-muxed)**.

**If this value is `true`**, then present the property "**Capture Device Audio**" as _either_ a dropdown, in which case label the options "**Automatically**" & "**Never**"; or as a checkbox.

Then **If [said] "Capture Device Audio" is enabled,** _and only then_, should OBS try and capture audio from the source, alongside video.

**If "Capture Device Audio" is not enabled, or is subsequently _disabled_**, OBS should remove said audio source from the mixer, and cease audio capture from the multiplexed, or "muxed", source.

_Optionally_, OBS may present the "Capture Device Audio" as **greyed out** and unchecked, if a source is not multiplexed, with a tooltip stating _"The selected video capture device does not support audio capture."_

Default behavior for **upgraded sources** shall be to leave audio **disabled**. **New sources** shall **enable** audio automatically, if it is available.

****

Through this:

1. OBS will only capture audio **when it is present**. — This prevents user confusion for sources without audio.
2. OBS' Audio Options will be presented in a clear and concise manner, avoiding issues relating to its PC counterpart.
3. OBS will functionally be compatible with any `muxed` (multiplexed) A/V device on macOS.
4. OBS will not break older sources, whilst avoiding any "jank" or otherwise friction-inducing UX design.

gxalpha:
> Then I propose this alternative:
> 
> I recommend when OBS is enumerating AVKit devices for a capture session, it also maintains a separate dictionary of devices which contain audio, which includes a link back to the primary list of devices for matches between the two using some sort of ID.
>
>The properties window will have one singular audio property: A dropdown. labeled "Capture Audio", and two options:
>
>"Automatically"
>"Disabled"
>(Using a dropdown instead of a checkbox, as this allows for more interactive explanation of what is happening to the end user).
>
>Then, if and only if the device presents itself as a video device with audio will OBS present an audio device for that source in the mixer.

This implementation would seem like overkill to me, especially since this PR already does most of that (but much simpler). It already only shows devices that report having audio (`AVMediaTypeMuxed`) in the mixer, and I really don't see a benefit of yet another dropdown menu.

MisutaaAsriel:
> This implementation would seem like overkill to me, especially since this PR already does most of that (but much simpler). It already only shows devices that report having audio (`AVMediaTypeMuxed`) in the mixer, and I really don't see a benefit of yet another dropdown menu.

Audio Capture may not be desired in all cases, and whilst Windows' UX is _unruly and set to be renovated_, giving users the _option_ to capture audio instead of forcing it would make for a better user experience, if done right.

The "dropdown" option was merely a UI suggestion. A simple checkbox would do, even. The dropdown would just be more "descriptive". — Behind the scenes it'd be a boolean, either way.

E.G. I use my iPhone for VTubing. I **don't** want to capture audio from it. But if I stream my iPad, I **do** want to capture audio from that.

gxalpha:
> I'm trying to acknowledge and account for concerns around UX here with my suggestions.

Don't worry, I'm just trying to find out advantages and disadvantages of either approach, and your feedback is very welcome!

MisutaaAsriel:
Curious, but I cant get this to actually build and run properly. The logs just spit out being unabled to open any plugins due to "`load commands exceed length of first segmentnot a mach-o file`", and OBS fails to initialize the encoder.

Is there any particular steps I should be taking to build? On a 2017 i3 Intel MacBook Pro 13" Fn model.

MisutaaAsriel:
image

So I went ahead and (with some trial and error) changed your code to my checkbox idea, **and it's actually less code**.

By changing:
```C
	if (!obs_data_get_bool(current_settings, "enable_audio")) {
		auto cb = [](obs_properties_t *, obs_property_t *prop,
			     void *data) {
			struct av_capture *capture =
				static_cast<av_capture *>(data);
			OBSDataAutoRelease settings = obs_data_create();
			obs_data_set_bool(settings, "enable_audio", true);
			obs_source_update(capture->source, settings);

			// Enable all audio tracks
			obs_source_set_audio_mixers(capture->source, 0x3F);
			obs_property_set_visible(prop, false);
			return true;
		};
		obs_properties_add_button2(props, "enable_audio_button",
					   obs_module_text("EnableAudio"), cb,
					   capture);
	}
```
	
To:
```C
	obs_properties_add_bool(props, "enable_audio", obs_module_text("EnableAudio"));
```
and adding:
```C
obs_property_t *ap = obs_properties_get(props, "enable_audio");
	if ([dev hasMediaType:AVMediaTypeMuxed]) {
		obs_property_set_visible(ap, true);
	} else {
		obs_property_set_visible(ap, false);
	}
```

To `properties_device_changed` before the `return`, in `av-capture.mm`, and changing `EnableAudio` to `"Capture Device Audio"` in the language file, it arguably provides a cleaner, more controllable feature, whilst hiding the checkbox when audio is unavailable.

**Granted, I do not know the behavior when you change sources. Does it save the boolean value, or reset? What even is the expected behavior here?** — If I was more confident in this code I'd make a pull to your fork, but as it is, I'll leave the _how_ in this comment, for those more experience than I to update.

gxalpha:
> So I went ahead and (with some trial and error) changed your code to my checkbox idea, **and it's actually less code.**

My point regarding "overkill" was the suggestion regarding separate dictionaries. I know that adding a checkbox would result in less code (especially since you could just get rid of v1 and v2 altogether).
My point is that I think that adding a checkbox is a much worse user experience. I can't help but think that if you want to adjust audio (including turning it off) that's exactly what the mixer is for. When adding a checkbox, all you're doing is adding a second way to do that.

MisutaaAsriel:
> > So I went ahead and (with some trial and error) changed your code to my checkbox idea, **and it's actually less code.**
> 
> My point regarding "overkill" was the suggestion regarding separate dictionaries. I know that adding a checkbox would result in less code (especially since you could just get rid of v1 and v2 altogether). My point is that I think that adding a checkbox is a much worse user experience. I can't help but think that if you want to adjust audio (including turning it off) that's exactly what the mixer is for. When adding a checkbox, all you're doing is adding a second way to do that.

User Experience. It's better to have a checkbox in the actual device properties, than have to mute and hide every single source that uses muxed audio.

Again, not _every_ muxed source will _need_ to have its audio captured. Throwing everything into the mixer is both obnoxious and in worst case , cause unnecessary bloat to the UI.

WizardCM:
> > So I went ahead and (with some trial and error) changed your code to my checkbox idea, **and it's actually less code.**
> 
> My point regarding "overkill" was the suggestion regarding separate dictionaries. I know that adding a checkbox would result in less code (especially since you could just get rid of v1 and v2 altogether). My point is that I think that adding a checkbox is a much worse user experience. I can't help but think that if you want to adjust audio (including turning it off) that's exactly what the mixer is for. When adding a checkbox, all you're doing is adding a second way to do that.

I second this - sources shouldn't need a checkbox for enabling disabling audio if we can avoid it (such as programmatically detecting if the source's content actually supports it) because the Audio Mixer handles that - the user can mute it if they really want, and then right click -> Hide from Mixer. Neither Decklink, AJA or Media Source have such a toggle, and we plan on removing it from the Video Capture Device source on Windows.

MisutaaAsriel:
> I second this - sources shouldn't need a checkbox for enabling disabling audio if we can avoid it (such as programmatically detecting if the source's content actually supports it) because the Audio Mixer handles that - the user can mute it if they really want, and then right click -> Hide from Mixer. Neither Decklink, AJA or Media Source have such a toggle, and we plan on removing it from the Video Capture Device source on Windows.

Fair enough. If this is the intended design going forward for all sources, then it makes sense here, too, I'd agree with that. — Im not going to complain about OBS becoming more consistent.

WizardCM (approved):

plugins/mac-avcapture/data/locale/en-US.ini
```diff
@@ -11,3 +11,4 @@ VideoRange.Partial="Partial"
 VideoRange.Full="Full"
 Auto="Auto"
 Unknown="Unknown ($1)"
+EnableAudio="Enable audio (if supported by device)"
```
RytoEX, line 14: The parenthetical probably doesn't need to be a parenthetical, but is also probably fine either way.

RytoEX (commented):
Just a string nit/comment.

plugins/mac-avcapture/av-capture.mm
```diff
@@ -710,20 +785,40 @@ static bool init_session(av_capture *capture)
...
+	}
+
 	capture->session = session;
 	capture->delegate = delegate;
 	capture->out = out;
+	capture->audioOut = audioOut;
 	capture->queue = queue;
+	capture->audioQueue = audioQueue;
+
+	capture->audio.format = AUDIO_FORMAT_16BIT;
```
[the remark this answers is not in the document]
gxalpha: Usually (like with iOS devices) but we obviously shouldn't rely on it. Fixed.

PatTheMav (approved):

RytoEX (approved):
Changes I've requested have been made. I assume by the other approvals that everything here is good to go, @PatTheMav @WizardCM ?

RytoEX:
Confirmed off-thread that this is good to land, so doing that now.

PR #5155, opened 2021-08-18
CI: Add capability for native Apple Silicon and Universal binary builds

PatTheMav:
### Description
Adds the necessary changes to build OBS on macOS for x86_64, arm64, and universal binary builds. Heavy lifting is done by cmake itself (via the `CMAKE_OSX_ARCHITECTURES` flag) which allows compilation for single or dual architectures.

**Notes:**

* This PR requires all the work done in the cmake overhaul PR, which is why its commits/changes are included here. Once that PR is merged, the rebase will make those additional changes disappear
* ~~For the time being this doesn't work with browser source enabled (which it is by default) as we don't have a native patched CEF build for arm64 available~~
* ~~That said, with browser source functionality disabled, this will compile a fully native Apple Silicon build~~
* This PR uses the recently updated obs-deps version that contains universal libraries
* Dependencies are _not_ expanded to `/tmp` anymore but instead in the `obs-build-dependencies` directory used by all other build dependencies. This avoids the issue of `obs-deps` disappearing at reboot (as macOS empties `/tmp`) and also makes use of the new `obs-deps` not using absolute path IDs in signature anymore

### Motivation and Context
* Future-proof OBS for Apple Silicon machines
* Proper multi-arch handling for macOS
* Dependency cleanup

### How Has This Been Tested?
* Universal, x86_64, and arm64 builds created on x64_host
* Builds running on their native archs
* Universal builds running on both archs with their native runtimes
* Further tests pending

### Types of changes
- New feature (non-breaking change which adds functionality)
- Code cleanup (non-breaking change which makes code smaller or more readable)

### Checklist:
- [x] My code has been run through [clang-format](https://github.com/obsproject/obs-studio/blob/master/.clang-format).
- [x] I have read the [**contributing** document](https://github.com/obsproject/obs-studio/blob/master/CONTRIBUTING.rst).
- [x] My code is not on the master branch.
- [x] The code has been tested.
- [x] All commit messages are properly formatted and commits squashed where appropriate.
- [x] I have included updates to all appropriate documentation.

PR #5453, opened 2021-10-26
Fix compiler warnings in several modules

PatTheMav:
### Description
Fixes current compiler warnings when using `clang`, also introduces a global `OBS_UNUSED` decorator for cross-platform code to explicitly mark variables as unused (MSVC might be able to use the generic C++ decorator but I leave that up to Windows folks to amend, it's inert for MSVC right now).

Also fixes type mismatches and casts from incompatible sizes in some places.

### Motivation and Context
* Clean up compiler output to enable better focus on actual errors or code issues

### How Has This Been Tested?
* Built and run several times on macOS

### Types of changes

### Checklist:

### DIFF
--- a/plugins/vlc-video/vlc-video-source.c
+++ b/plugins/vlc-video/vlc-video-source.c
@@ -547,7 +547,7 @@ static bool valid_extension(const char *ext)
 	if (!ext || !*ext)
 		return false;
 
-	b = EXTENSIONS_MEDIA + 1;
+	b = &EXTENSIONS_MEDIA[1];
 	e = strchr(b, ';');
 
 	for (;;) {


PatTheMav:
The clang/gcc notation seems very common among OSS projects it also works for both languages. If we want to use `[[maybe_unused]]`, then we'd need a separate macro just for C++ code _or_ use it in C++ code directly (leaving the macro for C code or mixed codebases).

That means that I cannot just use it as an alternative for MSVC as it would break C code. If that's ok for now, I'd un-draft this PR.

RytoEX:
It seems like some of this made it into #5155 instead. Could you rebase this?

It also seems like this might overlap a bit with #5653.

PatTheMav:
The change that was most discussed in comments to the PR seems to have been adopted to the main OBS codebase a while ago, so they seem moot.

PR now contains code-level fixes only.

plugins/mac-virtualcam/src/dal-plugin/OBSDALPlugInInterface.mm
```diff
@@ -85,13 +85,15 @@ HRESULT HardwarePlugIn_QueryInterface(CMIOHardwarePlugInRef self, REFIID uuid,
 }
 
 // I think this is deprecated, seems that HardwarePlugIn_InitializeWithObjectID gets called instead
-OSStatus HardwarePlugIn_Initialize(CMIOHardwarePlugInRef self)
+OSStatus HardwarePlugIn_Initialize(CMIOHardwarePlugInRef self
+				   __attribute__((unused)))
```
gxalpha, line 89: Since it's c++, could we just omit the name here (and the other instances)? We do that for browser a lot and it makes things more clean imo

PatTheMav: I don't think that will work in these specific cases, because the parameters are indeed used when the project is compiled in `Debug` config - otherwise the `DLogFunc` macro is empty and the parameters are not used.

If the macros themselves are restored you're back at square one because the macro itself then emits an "unused variable" warning for every invocation as its function block is empty. Might require an entirely different solution.

PR #1144, opened 2018-01-10
[WIP] UI: show scene item spacing helpers in preview editor

Palakis:
Same as the green distance lines in SLOBS. Visibility configurable and enabled by default.

image

Known issues:
- Some lines are still drawn even if outside of the video zone. Can actually be useful to keep it that way.
 ([example](https://user-images.githubusercontent.com/1812130/34773993-0283bc34-f60e-11e7-8312-4e66f6d0b811.png)).
- [ ] Rotation support needs more love. It's fine to keep lines perpendicular, but over 90° it messes up which side is where ([example of correct behaviour](https://user-images.githubusercontent.com/1812130/34774028-253d58e8-f60e-11e7-87f6-bfd9931cb42e.png), [image of example behaviour](https://user-images.githubusercontent.com/1812130/34774047-39d7a57e-f60e-11e7-88be-0128a032753e.png)).
- [ ] Wrong behaviour when several items are selected simultaneously

PR #5471, opened 2021-11-01
UI: Show spacing helpers in preview

cg2121:
### Description
This shows distance between sides of preview and edges of sources.
This will allow users to more easily align sources.

Screenshot from 2021-11-01 02-29-40

### Motivation and Context
Based on original PR by @Palakis https://github.com/obsproject/obs-studio/pull/1144

This version properly supports rotation of sources and refactors the code.

### How Has This Been Tested?
Selected items in preview to if if everything was rendered properly.

### Types of changes

### Checklist:


WizardCM:
Palakis' main concern with the original implementation was the (unrealistically) high performance impact when these were visible. Does this solve those concerns?

I'm also not personally a fan of this being the same colour as the outlines for the item, but I'm not sure what other colour to use. Maybe an orange?

RytoEX:
> 
> 
> Palakis' main concern with the original implementation was the (unrealistically) high performance impact when these were visible. Does this solve those concerns?
> 
> I'm also not personally a fan of this being the same colour as the outlines for the item, but I'm not sure what other colour to use. Maybe an orange?

See also this branch where @Palakis and I tried to reduce the CPU usage:  https://github.com/RytoEX/obs-studio/commits/pr-1144-sceneitem-spacers

I think we had previously used green for that color.  @Dillon would probably ultimately want to make it configurable.

cg2121:
The only thing that causes more CPU usage, from what I can tell, is the rendering of the source labels with obs_source_video_render(). I don't think anything can be done about it in the scope of this PR, as it would be an internal OBS issue.

WizardCM:
This does not render correctly on Windows.

image

Additionally, it is not visible/readable on a white background.

image

To avoid the performance impact, maybe we only show the numbers when the item is being moved or when the user is holding the Alt key?

cg2121:
The Windows text issues are now fixed.

The rendering with multiple items selected is intentional, it would confusing to the user if many more line/labels were drawn, plus not very good on performance.

WizardCM:
Looks much better, but I'd probably go for a slightly larger font size to improve readability.

image

cg2121:
The text should be more readable now.

MatuaDoc:
Possible to have an opaque background for the size labels? Would address potential accessibility/usability issues like WizardCM's screenshot of the white background but also (and especially) against busy backgrounds. The screenshots so far only show completely blank backgrounds.

feetstv:
Would it be possible to have an opaque background for the labels?

This would address potential accessibility/usability issues like WizardCM's screenshot of the white background but also (and especially) against busy backgrounds. The screenshots so far only show completely blank backgrounds which isn't the only scenario where one would resize/move items.

Warchamp7:
I think for multiple item selection, we should go one of two routes:
- Don't draw the lines at all
- Draw the line from the closest item to that edge

I worry the latter could get complicated or weird pretty quickly though

I also think these lines/labels should be disabled for any rotation that is not an exact multiple of 90 degrees

obs64_r6uPXCC2HG

obs64_oLBFzaFTC5

---

On the performance side, I was seeing a 0.4% - 1.0% CPU increase, but only while moving a source.

While that's not insignificant, it's a massive improvement over what Palakis was seeing. Anyone concerned with that impact should try moving a source with the Edit Transform window open :D

WizardCM:
Heads up, there are merge conflicts.

I do agree with Warchamp's thoughts.

cg2121:
Updated to latest master.

Warchamp7:
Are you able to address these points @cg2121 

> I think for multiple item selection, we should go one of two routes:
> 
> * Don't draw the lines at all
> * Draw the line from the closest item to that edge
> 
> I worry the latter could get complicated or weird pretty quickly though
> 
> I also think these lines/labels should be disabled for any rotation that is not an exact multiple of 90 degrees
> 
> obs64_r6uPXCC2HG
> 
> obs64_oLBFzaFTC5
> 
> On the performance side, I was seeing a 0.4% - 1.0% CPU increase, but only while moving a source.
> 
> While that's not insignificant, it's a massive improvement over what Palakis was seeing. Anyone concerned with that impact should try moving a source with the Edit Transform window open :D

cg2121:
I updated to fix bug with rotations, so we don't have to hide the labels when the item is rotated, as they should render correctly now.

cg2121:
I've also disabled the labels when multiple items are selected.

UI/window-basic-preview.cpp
```diff
@@ -2007,3 +2007,292 @@ OBSBasicPreview *OBSBasicPreview::Get()
...
+	if (oti.scale.y < 0.0f) {
+		vec3_set(&top, 0.5f, 1.0f, 1.0f);
+		vec3_set(&bottom, 0.5f, 0.0f, 1.0f);
+	}
+
+	// Decide which side to use with box transform, based on rotation
+	// Seems hacky, probably a better way to do it
+	float rot = oti.rot;
+
+	if (rot >= 90.0f) {
```
Warchamp7, line 2220: Can we adjust this 'breakpoint' to 45 instead of 90? Should avoid the near parallel line effect at rotations near multiples of 90

Warchamp7 (changes requested):
I think this is good but left one request about the edge swapping

UI/window-basic-preview.cpp
cg2121: I don't think this can be avoided, as far as I know, because the parallel lines happen regardless of what angle you put this at.

Warchamp7: Here's what I mean

image

https://user-images.githubusercontent.com/1554753/173735628-ad6c039f-7061-47f0-b746-aa0919d5a272.mp4

cg2121: I see what you mean now.

cg2121: Yep, it works much better now. Thanks for the change suggestion.

Warchamp7 (approved):

UI/data/locale/en-US.ini
```diff
@@ -801,6 +801,7 @@ Basic.Settings.General.ScreenSnapping="Snap Sources to edge of screen"
 Basic.Settings.General.CenterSnapping="Snap Sources to horizontal and vertical center"
 Basic.Settings.General.SourceSnapping="Snap Sources to other sources"
 Basic.Settings.General.SnapDistance="Snap Sensitivity"
+Basic.Settings.General.SpacingHelpers="Show pixel spacing helper lines"
```
WizardCM, line 804: Users might expect the word "alignment" and/or "guides" in here somewhere.

cg2121: Changed to "Show pixel alignment guides"

cg2121:
Updated to fix merge conflict

PR #5480, opened 2021-11-02
win-dshow: Save and restore video device configuration properties

exeldro:
### Description
Save and restore webcam settings/ video device configuration properties

Requires https://github.com/obsproject/libdshowcapture/pull/37

Inspired by [Camooz](https://obsproject.com/forum/resources/camooz-tool-to-manage-and-save-and-restore-camera-settings.1271/)

### Motivation and Context
Some webcams don't save the settings and this allows you to have different webcam settings per scene collection

### How Has This Been Tested?
On windows 64 bit

### Types of changes

### Checklist:

### DIFF
--- a/plugins/win-dshow/win-dshow.cpp
+++ b/plugins/win-dshow/win-dshow.cpp
@@ -177,6 +177,7 @@ enum class Action {
 	ConfigAudio,
 	ConfigCrossbar1,
 	ConfigCrossbar2,
+	SaveSettings,
 };
 
 static DWORD CALLBACK DShowThread(LPVOID ptr);
@@ -204,6 +205,7 @@ struct DShowInput {
 
 	WinHandle semaphore;
 	WinHandle activated_event;
+	WinHandle saved_event;
 	WinHandle thread;
 	CriticalSection mutex;
 	vector<Action> actions;
@@ -242,6 +244,10 @@ struct DShowInput {
 		activated_event = CreateEvent(nullptr, false, false, nullptr);
 		if (!activated_event)
 			throw "Failed to create activated_event";
+
+		saved_event = CreateEvent(nullptr, false, false, nullptr);
+		if (!saved_event)
+			throw "Failed to create saved_event";
 
 		thread =
 			CreateThread(nullptr, 0, DShowThread, this, 0, nullptr);
@@ -287,6 +293,8 @@ struct DShowInput {
 
 	bool UpdateVideoConfig(obs_data_t *settings);
 	bool UpdateAudioConfig(obs_data_t *settings);
+	bool UpdateVideoProperties(obs_data_t *settings);
+	void SaveVideoProperties();
 	void SetActive(bool active);
 	inline enum video_colorspace GetColorSpace(obs_data_t *settings) const;
 	inline enum video_range_type GetColorRange(obs_data_t *settings) const;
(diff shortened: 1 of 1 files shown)


cg2121:
Just my opinion, others can chime in: I think the settings should save automatically, so there shouldn't be a need for the checkbox.

WizardCM:
I know others have tried to implement something similar and ended up either reverting or disabling the capability because it was not saving correctly, or potentially corrupting data. I am OK with providing a checkbox to begin with to allow users to disable the functionality if it's not working as expected for them.

Fenrirthviti:
I agree that this should just be automatic without a checkbox. Plenty of applications accomplish this without corruption or failure saving, I'd prefer we just tested this thoroughly and made it automatic.

exeldro:
I think the seeking testers label needs to be added to this PR to allow users to test the change.

exeldro:
Tested and confirmed working with C910, C920, C922, C295
Reported not working C270

WizardCM:
@exeldro I own a C270. Anything specific you'd like me to try?

exeldro:
@WizardCM set some custom camera settings, check if there are settings stored when OBS is closed in the scene collection json, in arrays called `VideoProcAmp` and `CameraControl`. Restart computer or have an other way to change the camera settings. Start OBS to check if the camera settings are restored.

exeldro:
Removed the extra setting, now it is always enabled

exeldro:
All changed to not make it a setting, but always save and restore

WizardCM:
Heads up, this PR has a merge conflict with a submodule.

mat1jaczyyy:
is it possible for this feature to additionally re-activate the video once? so the full process for restoring would be:

1. first load/activation
2. restore settings
3. deactivate
4. activate again
5. restore settings again

This is due to some webcams (such as C920) having really low FPS on manual exposure -6 or below on the first load, but after the device is deactivated and reactivated once, the FPS works perfectly fine. As far as I understand the current system, this sort of fix wouldn't be able to be fully restored.

Tod181:
Is there an updated version for OBS 27.2.1? When I updated OBS it removed the setting from my menu.

johnebgood:
> The latest test version including OBS 27.2.1 changes can be downloaded here: https://github.com/obsproject/obs-studio/actions/runs/1887279360

I think OBS 27.2.1 had an issue that was crashing my computer that was fixed in 27.2.2, would it be possible to get a build with 27.2.3? I can test with OBSBOT Tiny's, Feiyu Pocket 2S and some other oddball webcams.

exeldro:
The latest test version including OBS 27.2.3 changes can be downloaded here:
https://github.com/obsproject/obs-studio/actions/runs/2017636125

exeldro:
The latest test version can be downloaded here:
https://github.com/obsproject/obs-studio/actions/runs/2403350419

gxalpha (changes requested):
There are quite a few cases here where to me it looks like OBSDataAutoRelease or OBSDataArrayAutoRelease could be used instead of manually releasing, which could improve code readability. Could you check where those can be used (or is there a specific reason why they aren’t)?

(Since this PR is open for quite a bit I think the AutoSelect classes just didn’t exist on the main branch when it was created)

josiasmat:
Hello.
Is there any chance this feature will be reenabled anytime soon? Because I'm sticking to version 28.0.0 and skipping all updates as I cannot live without this anymore. I use 3 cameras, and before 28.0.0 had to configure all of them every time.
Would it not be better if added as an experimental feature, available through a checkbox?
Many thanks.

zacgoodwin:
1 year on and crickets still? Setting settings on every reboot like it's the stone ages.

PR #5503, opened 2021-11-07
XComposite: Fix capturing on software rasterization setups

ishitatsuyuki:
### Description
The GLX_EXT_texture_from_pixmap spec says:

> The contents of the texture after the drawable has been bound are defined
> as the result of all rendering that has completed before the call to
> glXBindTexImageEXT.  In other words, the results of any operation which
> has caused damage on the drawable prior to the glXBindTexImageEXT call
> will be represented in the texture.
> 
> Rendering to the drawable while it is bound to a texture will leave the
> contents of the texture in an undefined state.  However, no
> synchronization between rendering and texturing is done by GLX.  It is
> the application's responsibility to implement any synchronization
> required.

In practice, on most systems with a GPU this kept a directly binding to
the framebuffer, which made it work with the previous code. However,
on software rasterization setups using llvmpipe, a blit was done inside
the call to glXBindTexImageEXT. This was the cause of a notoriously
ignored bug where the captured image would "freeze" until the source
was reconfigured.

### Motivation and Context

Example complaint:

https://obsproject.com/forum/threads/xcomposite-window-capture-is-frozen-does-not-update.136729/

### How Has This Been Tested?

This has been tested on a xpra virtual desktop, which uses llvmpipe unless configured otherwise.

### Types of changes

### Checklist:

- N/A I have included updates to all appropriate documentation.

### DIFF
--- a/plugins/linux-capture/xcompcap-main.cpp
+++ b/plugins/linux-capture/xcompcap-main.cpp
@@ -200,6 +200,7 @@ struct XCompcapMain_private {
 	bool show_cursor = true;
 	bool cursor_outside = false;
 	xcursor_t *cursor = nullptr;
+	bool tick_error_suppressed = false;
 };
 
 XCompcapMain::XCompcapMain(obs_data_t *settings, obs_source_t *source)
@@ -300,14 +301,6 @@ static void xcc_cleanup(XCompcapMain_private *p)
 		GLuint gltex = *(GLuint *)gs_texture_get_obj(p->gltex);
 		glBindTexture(GL_TEXTURE_2D, gltex);
 		if (p->glxpixmap) {
-			glXReleaseTexImageEXT(xdisp, p->glxpixmap,
-					      GLX_FRONT_LEFT_EXT);
-			if (xlock.gotError()) {
-				blog(LOG_ERROR,
-				     "cleanup glXReleaseTexImageEXT failed: %s",
-				     xlock.getErrorText().c_str());
-				xlock.resetError();
-			}
 			glXDestroyPixmap(xdisp, p->glxpixmap);
 			if (xlock.gotError()) {
 				blog(LOG_ERROR,
@@ -397,6 +390,8 @@ void XCompcapMain::updateSettings(obs_data_t *settings)
 	Window prevWin = p->win;
 
 	xcc_cleanup(p);
+
+	p->tick_error_suppressed = false;
 
 	if (settings) {
 		/* Settings initialized or changed */
@@ -554,7 +549,7 @@ void XCompcapMain::updateSettings(obs_data_t *settings)
 				     GS_GL_DUMMYTEX);
 	GLuint gltex = *(GLuint *)gs_texture_get_obj(p->gltex);
 	glBindTexture(GL_TEXTURE_2D, gltex);
-	glXBindTexImageEXT(xdisp, p->glxpixmap, GLX_FRONT_LEFT_EXT, NULL);
+	glXBindTexImageEXT(xdisp, p->glxpixmap, GLX_FRONT_EXT, nullptr);
 	if (xlock.gotError()) {
 		blog(LOG_ERROR, "glXBindTexImageEXT failed: %s",
 		     xlock.getErrorText().c_str());
@@ -568,6 +563,12 @@ void XCompcapMain::updateSettings(obs_data_t *settings)
 	glTexParameteri(GL_TEXTURE_2D, GL_TEXTURE_MAG_FILTER, GL_LINEAR);
 	// glxBindTexImageEXT might modify the textures format.
 	gs_color_format format = gs_format_from_tex();
+	glXReleaseTexImageEXT(xdisp, p->glxpixmap, GLX_FRONT_EXT);
+	if (xlock.gotError()) {
+		blog(LOG_ERROR, "glXReleaseTexImageEXT failed: %s",
+		     xlock.getErrorText().c_str());
+		xlock.resetError();
+	}
 	glBindTexture(GL_TEXTURE_2D, 0);
 	// sync OBS texture format based on any glxBindTexImageEXT changes
 	p->gltex->format = format;
@@ -620,7 +621,7 @@ void XCompcapMain::tick(float seconds)
 		p->win = 0;
 	}
 
-	XDisplayLock xlock;
+	XErrorLock xlock;
 	XWindowAttributes attr;
 
 	if (!p->win || !XGetWindowAttributes(xdisp, p->win, &attr)) {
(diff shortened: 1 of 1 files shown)


plugins/linux-capture/xcompcap-main.cpp
```diff
@@ -135,7 +135,7 @@ void XCompcapMain::defaults(obs_data_t *settings)
 	obs_data_set_default_bool(settings, "exclude_alpha", false);
 }
 
-#define FIND_WINDOW_INTERVAL 0.5
```
kkartaltepe, line 138: Is this change related to the fix?

plugins/linux-capture/xcompcap-main.cpp
```diff
@@ -621,6 +621,7 @@ void XCompcapMain::tick(float seconds)
 	}
 
 	XDisplayLock xlock;
+	XErrorLock xerrlock;
```
kkartaltepe, line 624: we are already holding an xserver lock in XDisplayLock, did you mean to replace this with the XErrorLock?

ezette:
Do not know

plugins/linux-capture/xcompcap-main.cpp
ishitatsuyuki: Oops, I accidentally pushed some debug code. It is now removed.

plugins/linux-capture/xcompcap-main.cpp
ishitatsuyuki: Yes, I didn't realize a XErrorLock supersedes a XDisplayLock. Thanks!

norihiro:
The commit title should start from `linux-capture:` instead of `XComposite:`.

ishitatsuyuki:
Fixed.

kkartaltepe:
I have two questions,

1) now that we bind the texture do we we need to move the release into the rendering loop. The recommended calls are 
```
XGrabServer()
glXWaitX() or XSync()
glXBindTexImageEXT()

<Do rendering/compositing>

glXReleaseTexImageEXT()
XUngrabServer()
```
In the extension. Will we be leaking something by repeatedly binding without releasing?

2) Can we avoid the log spam when binding fails, such as for offscreen windows? This change currently introduces an error in the log per frame in this case.

jp9000:
Hey there, just wanted to check back with you, we just had those few remaining questions there from @kkartaltepe.

ishitatsuyuki:
Sorry, couldn't prioritize this due to other work but I'll try to get back on this soon.

ishitatsuyuki:
I have updated the patch.

The bind and release calls are now matched, in order to avoid causing potentially implementation dependent leaks. It also binds the correct texture before `glXBindTexImageEXT` now.

Only the first error from `tick` will be logged in order to avoid spam. Logging is re-enabled on each reconfiguration.

kkartaltepe:
Thanks this looks good on my machine, if you get a chance and can apply this patch that would be nice otherwise ill probably make a PR for it later. (Just avoiding stereoscopic framebuffer name and aesthetic changes to the tick error check).

```diff
--- a/plugins/linux-capture/xcompcap-main.cpp
+++ b/plugins/linux-capture/xcompcap-main.cpp
@@ -549,7 +549,7 @@ void XCompcapMain::updateSettings(obs_data_t *settings)
                                     GS_GL_DUMMYTEX);
        GLuint gltex = *(GLuint *)gs_texture_get_obj(p->gltex);
        glBindTexture(GL_TEXTURE_2D, gltex);
-       glXBindTexImageEXT(xdisp, p->glxpixmap, GLX_FRONT_LEFT_EXT, nullptr);
+       glXBindTexImageEXT(xdisp, p->glxpixmap, GLX_FRONT_EXT, nullptr);
        if (xlock.gotError()) {
                blog(LOG_ERROR, "glXBindTexImageEXT failed: %s",
                     xlock.getErrorText().c_str());
@@ -563,7 +563,7 @@ void XCompcapMain::updateSettings(obs_data_t *settings)
        glTexParameteri(GL_TEXTURE_2D, GL_TEXTURE_MAG_FILTER, GL_LINEAR);
        // glxBindTexImageEXT might modify the textures format.
        gs_color_format format = gs_format_from_tex();
-       glXReleaseTexImageEXT(xdisp, p->glxpixmap, GLX_FRONT_LEFT_EXT);
+       glXReleaseTexImageEXT(xdisp, p->glxpixmap, GLX_FRONT_EXT);
        if (xlock.gotError()) {
                blog(LOG_ERROR, "glXReleaseTexImageEXT failed: %s",
                     xlock.getErrorText().c_str());
@@ -653,13 +653,11 @@ void XCompcapMain::tick(float seconds)
        }
 
        glBindTexture(GL_TEXTURE_2D, *(GLuint *)gs_texture_get_obj(p->gltex));
-       glXBindTexImageEXT(xdisp, p->glxpixmap, GLX_FRONT_LEFT_EXT, nullptr);
-       if (xlock.gotError()) {
-               if (!p->tick_error_suppressed)
-                       blog(LOG_ERROR, "glXBindTexImageEXT failed: %s",
-                            xlock.getErrorText().c_str());
+       glXBindTexImageEXT(xdisp, p->glxpixmap, GLX_FRONT_EXT, nullptr);
+       if (xlock.gotError() && !p->tick_error_suppressed) {
+               blog(LOG_ERROR, "glXBindTexImageEXT failed: %s",
+                    xlock.getErrorText().c_str());
                p->tick_error_suppressed = true;
-               xlock.resetError();
        }
 
        if (p->include_border) {
@@ -672,13 +670,11 @@ void XCompcapMain::tick(float seconds)
                                       height());
        }
 
-       glXReleaseTexImageEXT(xdisp, p->glxpixmap, GLX_FRONT_LEFT_EXT);
-       if (xlock.gotError()) {
-               if (!p->tick_error_suppressed)
-                       blog(LOG_ERROR, "glXReleaseTexImageEXT failed: %s",
-                            xlock.getErrorText().c_str());
+       glXReleaseTexImageEXT(xdisp, p->glxpixmap, GLX_FRONT_EXT);
+       if (xlock.gotError() && !p->tick_error_suppressed) {
+               blog(LOG_ERROR, "glXReleaseTexImageEXT failed: %s",
+                    xlock.getErrorText().c_str());
                p->tick_error_suppressed = true;
-               xlock.resetError();
        }
        glBindTexture(GL_TEXTURE_2D, 0);
 
```

ishitatsuyuki:
Applied. Just FYI, I had to ignore whitespaces when applying the patch and convert the indents to tab later.

kkartaltepe:
Ah right whitespace gets ruined in github, my bad.

univrsal:
Just a heads up, it seems like this is causing performance issues on proprietary nvidia drivers.

Log for 4623a6b4bc4c89ceb5db684e2e7fbd57d01129aa (no rendering lag):
https://obsproject.com/logs/GMoA70ZoAVngkg9S
Log for 316f858c68d3ef3a73166191b6f89b7bc3f3387c:
https://obsproject.com/logs/AqYbf16E3VT5g6cw

For me it just reduced the fps from 60 to 30 when showing a window capture. When moving a window fps would go down even further to around 15. When I checked out the previous commit things went back to normal.

ishitatsuyuki:
@univrsal Can you create an issue? I think this needs more investigation to resolve, and even should the PR be reverted, an issue is needed to track the investigations and progress before this can reland.

PR #5507, opened 2021-11-10
UI: Link python when obs-scripting python is enabled

### LINKED ISSUE #2222: Python scripting fails to load binary modules
When using a Python script in obs, loading a binary module fails due to non-exported CPython 
symbols.

I am using obs on a Linux Debian Buster.  Tested with obs compiled from current git master (ccb7d897) as well as the debian package provided in testing, aka  `obs-studio 24.0.3+dfsg1-2`.

Given the following example which tries to load the `termios` python package (a cython package):

```python
import termios
```
## Expected Behavior

I would expect this python snippet to load properly (and do nothing).

## Current Behavior

I fails with the following traceback:

```
Traceback (most recent call last):
File "/home/david/perso/elec/obs-tally/obsFirmataTally.py", line 9, in <module>
  import termios
ImportError: /usr/lib/python3.7/lib-dynload/termios.cpython-37m-x86_64-linux-gnu.so: undefined symbol: PyExc_TypeError
```

## Steps to Reproduce
1. create a `test_import.py` script with the content being just `import termios`
2. load the script in the Tools->Scripts dialog 
3. open the Script Log dialog to see the described  traceback

## Additional information 

Using Python 3.7 from the system (3.7.3-1).

It also fails to load numpy for example.

kkartaltepe:
### Description

This fixes #2222 by ensuring python symbols are in the global symbol
table through a direct linkage to the executable. When python is
normally linked via frontend-tools.so its symbols are not global since
we dlopen frontend-tools.so. This causes issues for native python
modules which stopped directly linking python around 3.8 as the symbols
will not resolve.

This is pretty much a hack, but it minimizes the number of extra symbols
we add and hopefully isnt too terrible to maintain.

### Motivation and Context

Making python users happy is a good thing.

### How Has This Been Tested?

Running python 3.9 the url-text.py script no longer works with the same issues as #2222. With this fix it works again. The 2222 script with a plain import also works.

Hopefully it doesnt explode on windows/mac if someone could test there that would be great.

### Types of changes

### Checklist:

### DIFF
--- a/UI/CMakeLists.txt
+++ b/UI/CMakeLists.txt
@@ -148,6 +148,7 @@ elseif(APPLE)
 
 elseif(UNIX)
 	find_package(Qt5Gui REQUIRED)
+	find_package(PythonDeps QUIET)
 
 	include_directories(${Qt5Gui_PRIVATE_INCLUDE_DIRS})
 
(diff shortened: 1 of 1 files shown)


kkartaltepe:
And testing `DISABLE_PYTHON` works as expected for anyone who was previously using it.

PatTheMav:
This looks correct to me - macOS is the only platform where Python shouldn't be linked and the plugin compiled with `dynamic_lookup`.

Vis-a-vis the rework (condensed):

```
if(OS_WINDOWS)
  target_link_libraries(obspython PRIVATE Python::Python)
elseif(OS_MACOS)
  get_target_property(_PYTHON_INCLUDE_DIRECTORY Python::Python
                      INTERFACE_INCLUDE_DIRECTORIES)
  target_include_directories(obspython PRIVATE ${_PYTHON_INCLUDE_DIRECTORY})
  target_link_options(obspython PRIVATE -undefined dynamic_lookup)
elseif(OS_POSIX)
  target_link_libraries(obspython PRIVATE Python::Python)
endif()
```

PR #5509, opened 2021-11-10
win-virtualcam: Make sure virtualcam output thread safe

hellowanda:
### Description

Modify Windows virtualcam output to make sure it's thread safe. 

### Motivation and Context

The virtualcam output for Windows is not thread safe.

### How Has This Been Tested?

Tested on Windows 10.

### Types of changes

### Checklist:

### DIFF
--- a/plugins/win-dshow/virtualcam.c
+++ b/plugins/win-dshow/virtualcam.c
@@ -1,10 +1,13 @@
 #include <obs-module.h>
 #include <util/platform.h>
+#include "util/threading.h"
 #include "shared-memory-queue.h"
 
 struct virtualcam_data {
 	obs_output_t *output;
 	video_queue_t *vq;
+	volatile bool active;
+	volatile bool stopping;
 };
 
 static const char *virtualcam_name(void *unused)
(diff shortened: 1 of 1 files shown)


notr1ch:
Can you explain the failure scenario with the old code? Since you aren't taking a mutex lock when starting or stopping I'm curious what part is not considered thread safe.

hellowanda:
> Can you explain the failure scenario with the old code? Since you aren't taking a mutex lock when starting or stopping I'm curious what part is not considered thread safe.

The thread problem is only related to "_stopping_", because the **_virtualcam_stop_** (called from **_obs_output_stop_**) is called from UI thread, while the **_virtual_video_** is called from video-io thread.


If the **_video_queue_close(vcam->vq);_** in **_virtualcam_stop_** is called after **_'if (!vcam->vq)'_** in **_virtual_video_**, then it may crash in _**video_queue_write**_. I've had such a crash once。

I think we need only move the real _stop_ logic to virtual_video, just like _ffmpeg_muxer_ ,no need for a mutex lock.
```
obs_output_end_data_capture(vcam->output);
video_queue_close(vcam->vq);
vcam->vq = NULL;
```

PR #5518, opened 2021-11-11
libobs/graphics: Add support for NT Handle shared textures

VodBox:
### Description

Adds support for using shared textures that were made with the
D3D11_RESOURCE_MISC_SHARED_NTHANDLE flag.

This increases the minimum required Windows version to Windows 8 or the
Platform Update for Windows 7. As official support is only for Win 8+
this does not change official support.

### Motivation and Context

The new shared texture method required by our CEF branch uses shared textures made with the D3D11_RESOURCE_MISC_SHARED_NTHANDLE and D3D11_RESOURCE_MISC_SHARED_KEYEDMUTEX flags, so we need support for NT Handles.

### How Has This Been Tested?

This has been tested against an implementation of the new shared texture method used in our CEF branch.

### Types of changes

### Checklist:

### DIFF
--- a/libobs-d3d11/d3d11-subsystem.cpp
+++ b/libobs-d3d11/d3d11-subsystem.cpp
@@ -2826,6 +2826,23 @@ extern "C" EXPORT gs_texture_t *device_texture_open_shared(gs_device_t *device,
 	return texture;
 }
 
+extern "C" EXPORT gs_texture_t *
+device_texture_open_nt_shared(gs_device_t *device, uint32_t handle)
+{
+	gs_texture *texture = nullptr;
+	try {
+		texture = new gs_texture_2d(device, handle, true);
+	} catch (const HRError &error) {
+		blog(LOG_ERROR, "gs_texture_open_nt_shared (D3D11): %s (%08lX)",
+		     error.str, error.hr);
+		LogD3D11ErrorDetails(error, device);
+	} catch (const char *error) {
+		blog(LOG_ERROR, "gs_texture_open_nt_shared (D3D11): %s", error);
+	}
+
+	return texture;
+}
+
 extern "C" EXPORT uint32_t device_texture_get_shared_handle(gs_texture_t *tex)
 {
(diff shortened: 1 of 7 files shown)


libobs-d3d11/d3d11-texture2d.cpp
```diff
@@ -313,15 +313,24 @@ gs_texture_2d::gs_texture_2d(gs_device_t *device, ID3D11Texture2D *nv12tex,
 		InitRenderTargets();
 }
 
-gs_texture_2d::gs_texture_2d(gs_device_t *device, uint32_t handle)
+gs_texture_2d::gs_texture_2d(gs_device_t *device, uint32_t handle,
+			     bool ntHandle)
```
notr1ch, line 317: Maybe an enum would be better than a bool, to support future types of handles?

libobs-d3d11/d3d11-subsystem.cpp
```diff
@@ -2826,6 +2826,23 @@ extern "C" EXPORT gs_texture_t *device_texture_open_shared(gs_device_t *device,
 	return texture;
 }
 
+extern "C" EXPORT gs_texture_t *device_texture_open_shared1(gs_device_t *device,
```
notr1ch, line 2829: Not a very explanatory function name, maybe something like `device_texture_open_nt_shared` would be better?

notr1ch (commented):
Should this also be split into separate commits? It affects both `libobs-d3d11` and `libobs`.

libobs-d3d11/d3d11-texture2d.cpp
jp9000: To be fair it probably doesn't matter, as this is an internal structure and not exposed as API.

VodBox: As this is an internal function, the API can change in the future if this needs to be expanded, without introducing any kind of breaking change. I do not suspect there to be any new kinds of handles in the future, so I personally am inclined to keep it as is here.

PR #5523, opened 2021-11-13
UI,libobs: Fix delete scene undo problems

### LINKED ISSUE #5515: Undo crash when restoring deleted scene with sources
### Operating System Info

Windows 10

### Other OS

_No response_

### OBS Studio Version

Other

OBS 27.1.3-144-g0815d5c88 (64-bit, windows)

### OBS Studio Log URL

https://obsproject.com/logs/yfgkBu1bEjjz4F_E

### OBS Studio Crash Log URL

https://obsproject.com/logs/EAhC0vxpZqFjVz8n

### Expected Behavior

It will be restored normally.

### Current Behavior

Version: Nightly builds
Crash.

Version : 27.1.3
It doesn't crash, but it doesn't restore properly.

### Steps to Reproduce

Version : Nightly builds
1. Create a scene / scene 2 / scene 3.
2. Add the source to Scene 3. (Browser / Color Source)
3. Add Scene 3 to Scene and Scene 2.
4. Delete Scene 3.
5. Restore Scene 3. (undo)
6. It will not be restored and OBS will crash.

Version : 27.1.3
1. Create a scene / scene 2 / scene 3.
2. Add a source to Scene 3. (Browser / Color Source)
3. Add Scene 3 to Scene and Scene 2.
4. Delete scene 3.
5. Restore Scene 3. (undo)
6. It doesn't crash, but it doesn't restore properly.

https://user-images.githubusercontent.com/1337354/141227049-c65331bb-fa5a-4443-97db-8034a1020544.mp4

We are using translation. We apologize if the English is not correct.

This is a different problem from the previous report. I apologize for the confusing title.

gxalpha:
### Description

This PR consists of three commits, hopefully fixing all undo shenanigans when deleting sources:

Commit 1 makes scenes listen to the removed signal of their source. When that signal is called, all
other scenes will check if they have removed sources, thus removing nested instances of the
scene that got deleted. This makes them release their references to that scene, letting it get destroyed
properly.
Previously, scenes would check for removed sources when they were rendered (and still do, that doesn't
change). However, this would mean that they had to be selected once in the scene list or get rendered 
otherwise to release their references. Without undo/redo, this wasn't a big problem (it only showed in
for example the name of the source still being in use), however undo/redo expects to have to re-create
the scene, causing issues when it wasn't deleted before.

Commit 2 overhauls what gets stored in the undo data when deleting sources, and fixes the deleted
scene not getting added back to scenes it was nested in when undoing the deletion.
Firstly, the undo data json now differentiates between the deleted scene itself (including the sources in
it) and the scenes/groups it was nested in. #4916 relied on sceneitems of the scene still existing in other
scenes when the source got re-added. This lead to the scene only getting re-added when the the scene
it was nested in wasn't rendered in the meantime (see description of 1), leading to weird and sometimes
unpredictable-seeming behaviour.
The differentiation between the scene itself and scenes/groups it's nested in leads to more controllable
behaviour that should be less prone to bugs.
I added a few comments in my code explaining the individual steps, I hope they are understandable.

### Motivation and Context

"Follow-up" from #5495, after #5515 revealed some more problems.
I also discovered more weirdness along the way, that should all be fixed with this PR.

Fixes #5515

### How Has This Been Tested?

I tried quite a few configurations of scenes being nested in other scenes, including in groups. These
include the configurations of #5515 and #5495.
Most of them would break on undo in 27.1.3, but there wasn't a single one that would be
re-created incorrectly with this PR.

### Types of changes

### Checklist:

### DIFF
--- a/UI/window-basic-main.cpp
+++ b/UI/window-basic-main.cpp
@@ -3777,6 +3777,18 @@ static bool save_undo_source_enum(obs_scene_t *scene, obs_sceneitem_t *item,
 	return true;
 }
 
+static inline void RemoveSceneAndReleaseNested(obs_source_t *source)
+{
+	obs_source_remove(source);
+	auto cb = [](void *unused, obs_source_t *source) {
+		UNUSED_PARAMETER(unused);
+		if (strcmp(obs_source_get_id(source), "scene") == 0)
+			obs_scene_prune_sources(obs_scene_from_source(source));
+		return true;
+	};
+	obs_enum_scenes(cb, NULL);
+}
+
 void OBSBasic::RemoveSelectedScene()
 {
 	OBSScene scene = GetCurrentScene();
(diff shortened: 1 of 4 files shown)


gxalpha:
I think I have a better idea on how to do this without modifying libobs
Haven't tried yet, but will convert to draft until I do and update the PR

gxalpha:
Instead of modifying existing libobs behaviour to prune sources when a scene is removed, this now adds an API to libobs, enabling pruning scenes from where-ever else. Maybe not the optimal solution still, but should be much better

PR #5524, opened 2021-11-13
Updated Info.plist with new bool syntax

stoiandan:
I'm not sure if I missread this, or if it's simply wrong, but I've lookedup the defaults man page (https://ss64.com/osx/defaults.html) and also found this:
https://stackoverflow.com/questions/1832157/setting-a-boolean-property-in-info-plist-from-a-user-defined-setting (maybe just on iOS?)
And some other materials saying the <true/> <false/> tags are no longer valid, and <string></string> with either TRUE/FALSE or YES/NO, should be used.


gxalpha:
Could you please fill out the PR template and adjust the commit title according to the contributing document?

PR #5530, opened 2021-11-15
win-virtualcam: Make placeholder fit to virtualcam resolution

hellowanda:
### Description

This commit makes placeholer for win-virtualcam fit to virtualcam resolution instead of stretching.

### Motivation and Context

Currently, the placeholder is stretched to resolution of virtualcam, looks like this:

I think fitting to the virtualcam resolution might looks better, just like this:

### How Has This Been Tested?

Tested on Windows 10 64bit

### Types of changes

### Checklist:

### DIFF
--- a/plugins/win-dshow/virtualcam-module/placeholder.cpp
+++ b/plugins/win-dshow/virtualcam-module/placeholder.cpp
@@ -81,7 +81,7 @@ static void convert_placeholder(const uint8_t *rgb_in, int width, int height)
 	}
 }
 
-static bool load_placeholder_internal()
+static bool load_placeholder_internal(int dest_cx, int dest_cy)
 {
 	Status s;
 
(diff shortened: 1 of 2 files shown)


mufunyo:
If you're going through the effort of changing this, I would suggest making the placeholder image square, and cutting off the excess depending on the virtualcam aspect ratio, avoiding letter-/pillarboxing.

notr1ch:
I'm not sure this is a good idea. Sure, it looks nicer, but it can mislead the user as to the resolution the camera is set to. They may think everything is configured correctly, only to have the virtualcam output scale to fill when they start the camera. Having the placeholder scale in the same way as the output gives immediate feedback that something might be wrong.

WizardCM:
As mentioned above, the existing behaviour provides good feedback to the user in case their resolution is wrong. Thanks for your PR, but we're going to stick with the current behaviour for now.

PR #5535, opened 2021-11-17
libobs: Add raw audio callback function

cg2121:
### Description
This complements the already existing raw video callback function.

### Motivation and Context
Parity with the video callback.

### How Has This Been Tested?
Created audio callback when testing some custom code.

### Types of changes

### Checklist:

### DIFF
--- a/docs/sphinx/reference-core.rst
+++ b/docs/sphinx/reference-core.rst
@@ -473,6 +473,18 @@ Video, Audio, and Graphics
    :param callback:   The callback that receives raw video frames.
    :param param:      The private data associated with the callback.
 
+---------------------
+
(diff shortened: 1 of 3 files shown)


jp9000:
What was the motivation behind this PR? Other than parity.

Secondly, if possible you probably want to change the `track` parameter to `mix_idx` to keep it consistent with the current. Although I'd like to say that I wish I had used `track` instead because I admit that is better.

cg2121:
Yes, the main reason is for parity with the raw video. I have also renamed track to mix_idx.

jp9000:
I'll allow it for the sake of parity in this case, though in general, I'd prefer to add functionality when we actually use it -- not just because it might be used.

PR #5536, opened 2021-11-17
UI: Import Streamlabs Screen Capture source type

RytoEX:
### Description

Add support for importing the Screen Capture source type from Streamlabs scene collections. Since the Screen Capture source type is essentially an abstraction layer for Display Capture, Game Capture, and Window Capture, this adds some translation logic to translate Screen Capture sources into one of those source types.

Caveat:  since the target of the game capture subtype does not seem to be exposed, this does not specify a target process for that source.

That said, this should still be an improvement over failing to create a suitable source at all.

### Motivation and Context

We should support imports for as many source types as possible.

### How Has This Been Tested?

I've tested this with all three subtypes (Display, Game, Window) on my Windows 10 machine.  Could use additional testing for machines with multiple displays.

### Types of changes

### Checklist:

### DIFF
--- a/UI/importers/sl.cpp
+++ b/UI/importers/sl.cpp
@@ -224,6 +224,32 @@ static void get_scene_items(const Json &root, const Json::array &out_sources,
 		Json::object{{"items", out_items}, {"id_counter", length}};
 }
 
+static void translate_screen_capture(Json::object &out_settings, string &type)
+{
+	string subtype_info =
+		out_settings["capture_source_list"].string_value();
+	size_t pos = subtype_info.find(':');
+	string subtype = subtype_info.substr(0, pos);
+
+	if (subtype == "game") {
+		type = "game_capture";
+	} else if (subtype == "monitor") {
+		type = "monitor_capture";
+		out_settings["monitor"] = subtype_info.substr(pos);
+	} else if (subtype == "window") {
+		type = "window_capture";
+		out_settings["cursor"] = out_settings["capture_cursor"];
+		out_settings["window"] =
+			out_settings["capture_window_line"].string_value();
+		out_settings["capture_cursor"] = nullptr;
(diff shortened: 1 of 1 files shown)


dodgepong:
> since the target of the game capture subtype does not seem to be exposed, this does not specify a target process for that source.

It would appear that the either use their [Auto game capture mode](https://github.com/stream-labs/obs-studio/pull/161), or "capture any fullscreen application" mode: https://github.com/stream-labs/obs-studio/blob/153051426200f79da21c7546a847ddc040672cef/plugins/win-capture/screen-capture.c#L462-L469

RytoEX:
> 
> 
> > since the target of the game capture subtype does not seem to be exposed, this does not specify a target process for that source.
> 
> It would appear that they either use their [Auto game capture mode](https://github.com/stream-labs/obs-studio/pull/161), or "capture any fullscreen application" mode: https://github.com/stream-labs/obs-studio/blob/153051426200f79da21c7546a847ddc040672cef/plugins/win-capture/screen-capture.c#L462-L469

My assumption was the former.  I mostly meant that if they are using their auto game capture mode, then we have no way to know which application was detected.  In this case, it seems that it will import as "Capture any fullscreen application", which should be an acceptable step in the right direction, but let me know if more work needs to be done here.

dodgepong:
I think "Capture any fullscreen application" is probably the least-bad fallback. Obviously it won't work for every case, but it's the closest analogue we have.

RytoEX:
> 
> 
> I think "Capture any fullscreen application" is probably the least-bad fallback. Obviously it won't work for every case, but it's the closest analogue we have.

At the very least, this will create the appropriate source type (game capture) in the canvas location where it should be with our default settings. I'm not sure of anything else we could do here.

PR #5538, opened 2021-11-18
libobs: Fix gs_texture_2d::BackupTexture with GS_TEXTURE_CUBE

qqzz0xx:
### Description

Cubemaps weren't even working properly before this. The gs_texture_2d::data vector did not take cubemaps into account. This code fixes that issue. -Jim

### Motivation and Context

Would be nice to have cubemaps working I suppose. -Jim

### How Has This Been Tested?

Tested by Jim -Jim

### Checklist:

### DIFF
--- a/libobs-d3d11/d3d11-texture2d.cpp
+++ b/libobs-d3d11/d3d11-texture2d.cpp
@@ -50,26 +50,31 @@ void gs_texture_2d::InitSRD(vector<D3D11_SUBRESOURCE_DATA> &srd)
 
 void gs_texture_2d::BackupTexture(const uint8_t *const *data)
 {
-	this->data.resize(levels);
-
-	uint32_t w = width;
-	uint32_t h = height;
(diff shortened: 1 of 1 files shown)


RytoEX:
Please fill out the PR template properly.  What is the purpose of this?  What does it fix?  How have you tested it?

jp9000:
> Please fill out the PR template properly. What is the purpose of this? What does it fix? How have you tested it?

The D3D11 renderer code has to save copies of all the resources in case the driver crashes, because we designed the program to seamlessly work with driver crashes. Basically when the driver crashes, all those resources have to be recreated, thus you have to backup the resources. I'm guessing what happened is that we haven't updated the cubemap texture backup process, so when the driver crashes, it likely would recreate the cubemap texture incorrectly. Just one of those cases where it's very easy to go untested.

jp9000:
Tested, seems fine. Thanks for patching this oversight.

PR #5553, opened 2021-11-20
(libobs,)UI: Add ability to move multiple scene items in the source dock at once

gxalpha:
### Description

When sources are moved up and down in the source dock using the buttons,
previously only one would be moved, even if multiple sources were
selected. Now, all of them will be moved.
In case both elements in a group and outside of a group are selected,
the ones inside the group will be ignored.
In the process, this also fixes a bug where if you tried to move the
top-most element up or the bottom-most element down, it would still
create an undo action.
(Essentially, OBSBasic::MoveSceneItem was mostly rewritten)

Also see this demonstration:

https://user-images.githubusercontent.com/59806498/142740286-5182e254-f31b-4584-9332-d7d1cd0e60a1.mp4

### Motivation and Context

From Warchamp's internal list of things to do

### How Has This Been Tested?

Moved sources up and down with the buttons. This includes when having selected sources where some are grouped and others aren't. Also see the video above ^
Made sure the undo and redo actions work correctly and are named correctly.

### Types of changes

### Checklist:


ogmkp:
OK on OBS 27.2.0-beta2 on Debian 11 no issue found on my side.

PR #5559, opened 2021-11-21
Screencast session restore for Wayland / PipeWire

GeorgesStavracas:
### Description

Implement restoring screencast sessions using portals. Do so by storing the restore token received by the portal in the source data, and using it next time the source is loaded. This feature is only available in the v4 version of the ScreenCast portal [1] - if an older version is running in the host system, the current behavior is preserved.

### Motivation and Context

Currently OBS Studio has to recreate screencast sessions from scratch each time a PipeWire source is loaded. This can be annoying,

Use the new screencast restore options provided by xdg-desktop-portal to restore previous sessions.

### How Has This Been Tested?

 - Have xdg-desktop-portal and xdg-desktop-portal-gnome  installed from the master/main branches [2]
 - Run this branch of OBS Studio on Wayland, or X11 with `OBS_USE_EGL=1`
 - Add a PipeWire source, and select a window or a monitor
 - Close OBS Studio
 - Open it again
 - Notice that the sources were restored automatically

### Types of changes

### Checklist:

---

[1] https://github.com/flatpak/xdg-desktop-portal/pull/638
[2] This is **not** a GNOME-specific feature; it's just the only implementation available at this point. KDE folks reviewed and approved this new feature, and might be working on their implementation.

### DIFF
--- a/plugins/linux-capture/pipewire.c
+++ b/plugins/linux-capture/pipewire.c
@@ -65,6 +65,7 @@ struct _obs_pipewire_data {
 
 	char *sender_name;
 	char *session_handle;
+	char *restore_token;
 
 	uint32_t pipewire_node;
 	int pipewire_fd;
@@ -808,6 +809,20 @@ static void on_start_response_received_cb(GDBusConnection *connection,
 	g_variant_iter_loop(&iter, "(u@a{sv})", &obs_pw->pipewire_node,
 			    &stream_properties);
 
+	if (portal_get_screencast_version() >= 4) {
+		g_autoptr(GVariant) restore_token = NULL;
+
+		g_clear_pointer(&obs_pw->restore_token, bfree);
+
+		restore_token = g_variant_lookup_value(result, "restore_token",
+						       G_VARIANT_TYPE_STRING);
+		if (restore_token)
+			obs_pw->restore_token = bstrdup(
+				g_variant_get_string(restore_token, NULL));
+
+		obs_source_save(obs_pw->source);
+	}
+
 	blog(LOG_INFO, "[pipewire] %s selected, setting up screencast",
 	     capture_type_to_string(obs_pw->capture_type));
 
@@ -940,6 +955,16 @@ static void select_source(obs_pipewire_data *obs_pw)
 	else
 		g_variant_builder_add(&builder, "{sv}", "cursor_mode",
 				      g_variant_new_uint32(1));
+
+	if (portal_get_screencast_version() >= 4) {
+		g_variant_builder_add(&builder, "{sv}", "persist_mode",
+				      g_variant_new_uint32(2));
+		if (obs_pw->restore_token && *obs_pw->restore_token) {
+			g_variant_builder_add(
+				&builder, "{sv}", "restore_token",
+				g_variant_new_string(obs_pw->restore_token));
+		}
+	}
 
 	g_dbus_proxy_call(portal_get_dbus_proxy(), "SelectSources",
 			  g_variant_new("(oa{sv})", obs_pw->session_handle,
@@ -1104,6 +1129,8 @@ static bool reload_session_cb(obs_properties_t *properties,
 
 	obs_pipewire_data *obs_pw = data;
 
+	g_clear_pointer(&obs_pw->restore_token, bfree);
+
 	teardown_pipewire(obs_pw);
 	destroy_session(obs_pw);
 
@@ -1123,6 +1150,8 @@ void *obs_pipewire_create(enum obs_pw_capture_type capture_type,
 	obs_pw->settings = settings;
 	obs_pw->capture_type = capture_type;
 	obs_pw->cursor.visible = obs_data_get_bool(settings, "ShowCursor");
+	obs_pw->restore_token =
+		bstrdup(obs_data_get_string(settings, "RestoreToken"));
 
 	if (!init_obs_pipewire(obs_pw))
 		g_clear_pointer(&obs_pw, bfree);
@@ -1138,12 +1167,20 @@ void obs_pipewire_destroy(obs_pipewire_data *obs_pw)
 	teardown_pipewire(obs_pw);
 	destroy_session(obs_pw);
 
+	g_clear_pointer(&obs_pw->restore_token, bfree);
+
 	bfree(obs_pw);
+}
+
+void obs_pipewire_save(obs_pipewire_data *obs_pw, obs_data_t *settings)
+{
+	obs_data_set_string(settings, "RestoreToken", obs_pw->restore_token);
 }
 
 void obs_pipewire_get_defaults(obs_data_t *settings)
 {
 	obs_data_set_default_bool(settings, "ShowCursor", true);
+	obs_data_set_default_string(settings, "RestoreToken", NULL);
 }
 
 obs_properties_t *obs_pipewire_get_properties(obs_pipewire_data *obs_pw,
--- a/plugins/linux-capture/pipewire-capture.c
+++ b/plugins/linux-capture/pipewire-capture.c
@@ -51,6 +51,11 @@ static void pipewire_capture_destroy(void *data)
 	obs_pipewire_destroy(data);
 }
 
+static void pipewire_capture_save(void *data, obs_data_t *settings)
+{
+	obs_pipewire_save(data, settings);
+}
+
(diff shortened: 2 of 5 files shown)


kkartaltepe:
Are there any issues tracking implementation on other platforms you can link?

GeorgesStavracas:
> Are there any issues tracking implementation on other platforms you can link?

I just opened https://bugs.kde.org/show_bug.cgi?id=445875 and https://github.com/emersion/xdg-desktop-portal-wlr/issues/170 - not entirely sure it's a good idea for them to be in the commit message, do you have any preference?

kkartaltepe:
I agree it doesnt make sense in the commit message. I just wanted to know if other implementations actually were tracking the effort of implementing this functionality or not.

plugins/linux-capture/pipewire.c
```diff
@@ -1138,12 +1167,20 @@ void obs_pipewire_destroy(obs_pipewire_data *obs_pw)
...
 	destroy_session(obs_pw);
 
+	g_clear_pointer(&obs_pw->restore_token, bfree);
+
 	bfree(obs_pw);
 }
 
+void obs_pipewire_save(obs_pipewire_data *obs_pw, obs_data_t *settings)
+{
+	obs_data_set_string(settings, "RestoreToken", obs_pw->restore_token);
```
columbarius, line 1177: Could we use an array/object instead of a single string indexed by the first entry of `XDG_CURRENT_DESKTOP` (can be a :-seperated list). This way we could save a restore token for each DE making it possible to switch them without loosing the state each time. This can be usefull for simple setups with scenes for different outputs without creating a new set of scenes for each DE.

GeorgesStavracas: Clients doesn't have access to that information, and that was a fundamental part of the protocol design. This exchange is strictly between xdg-desktop-portal and the portal implementation.

columbarius: does flatpak hide a `getenv("XDG_CURRENT_DESKTOP")`?

GeorgesStavracas: Sorry, I'm not adding one restore token per desktop environment. It doesn't matter if we have access to environment variables or not. What you're proposing is merely a bad hack, OBS is the wrong layer to add this.

columbarius: Well on wlroots based applications using this envvar is actually a valid solution to eg. load settings wrt. the running compositor. But if there is no easy way to represent dynamic key-value pairs with obs_data_t I'm fine with solving this issue with profiles or other means.

GeorgesStavracas: The point is not how easy or hard it would be to implement this in OBS Studio. What I called "bad hack" is doing this goes directly against the purpose of portals. Portals exist to have DE-independent API that apps can use regardless of where they are being executed. And yet here we are, talking about adding DE-specific code in a DE-independent code. So I'm 100% against adding DE-specific code here because that's the wrong place to add that - the correct place is in portal frontends. That's actually what you suggested when you reviewed the restore-token PR. Did you change your mind about It?

columbarius: >  the correct place is in portal frontends. That's actually what you suggested when you reviewed the restore-token PR. Did you change your mind about It?

If I understood it correctly a token is a 1:1 correspondence with a given entry in the permission-store. This token is only valid for one use. After it was used the portal implementation might issue a new token for an identical entry in the permission store. To support mutliple implementations we either need:
* making the token 1:N in the permission store and updating all entries to the new issued token.
* Let the client manage multiple tokens and use the corresponding one.

From the discussion at the restore-token MR I took the assumption that the 1:1 correspondence was important to keep.

> And yet here we are, talking about adding DE-specific code in a DE-independent code.
I'm not talking about DE-specific code. I'm talking about a DE-specific config entry using the same independent codepath.

GeorgesStavracas: > If I understood it correctly a token is a 1:1 correspondence with a given entry in the permission-store. This token is only valid for one use. After it was used the portal implementation might issue a new token for an identical entry in the permission store.

None of this matters to applications. Applications express intent ("I want to restore session this later"), and get a blob to use next time. That's it. Nothing else matters for applications. It doesn't matter what the blob is. It doesn't matter how portals implement it. It doesn't matter that there are different frontends. This is all an implementation detail of portals, and the API design guarantees that none of this ever leaks into application code.

> I'm talking about a DE-specific config entry using the same independent codepath.

That's exactly the same. If you're writing a portals-based app and you have to check which DE the app is running on, then we failed at the API design level.

My question still stands: why did you request a way for portal frontends to exchange data between each other (the `(suv)` variant)? What you're asking here is the exact opposite of that, and I fail to see what your goals could even be when you request the exact opposite in each PR.

columbarius: > why did you request a way for portal frontends to exchange data between each other (the (suv) variant)? 

I requested the (suv) variant to have a more robust input validation in the portal implementation when receiving the blob from the permission-store. With only (v) collisions between the variants of a portal implementation could happen. With (suv) s has to be the name of the portal implementation. (Input validation has to be in place to prevent malice actors going against the spec).  

Parsing the variant of another portal was sth. undesired iirc so interoperability on the implementation side is out.

> What you're asking here is the exact opposite of that, and I fail to see what your goals could even be when you request the exact opposite in each PR.

My goal is to be able to configure DE1 in a specific way (including DE specific settings on different applications), switch to DE2, configure it in a completely different way, switch back to DE1 and be able to continue from where i left DE1. 

I don't care if that is sth. implemented 
* in the application (storing multiple tokens) or 
* in the portal frontend (saving multiple permission-store entries with the same token). 

We could also tell people to 
* use different obs profiles for different DEs 

and every screencast instance will happily save it's token. But as soon as obs starts with the wrong profile all tokens will probably be invalidated and the state of the scene/profile wrt. screencast will be lost.

I could solve this for me probably by using a wrapper
```
#!/bin/sh
obs --profile="$(echo $XDG_CURRENT_DESKTOP | cut -d: -f1)"
```
but loosing your setup because obs started with the "wrong" profile is not nice UX.

I don't know the obs userbase good enough if that is really an issue. I just wanted to make this issue visible.

columbarius: Note: This should not block this MR in any way. 
This is a huge improvement for users using only a single DE, which is the majority of users!
I'm asking if we want to consider this an issue with a solution now or later.

GeorgesStavracas: > I requested the (suv) variant to have a more robust input validation in the portal implementation when receiving the blob from the permission-store. With only (v) collisions between the variants of a portal implementation could happen. With (suv) s has to be the name of the portal implementation. (Input validation has to be in place to prevent malice actors going against the spec).

There is literally no way any application could use the private data of this protocol to create any attack vector.

> My goal is to be able to configure DE1 in a specific way (including DE specific settings on different applications), switch to DE2, configure it in a completely different way, switch back to DE1 and be able to continue from where i left DE1.

It seems to me that this is a premature optimization to cater for an unrealistic use case. People don't change DEs often, and when they do, OBS is going to be the smallest of their problems. With this PR, selecting a monitor becomes a one-time action.

columbarius: > There is literally no way any application could use the private data of this protocol to create any attack vector.

This is only relevant between portal implementations, not applications, the same way I wanted the (suv) var for easier parsing in the implementation.

> People don't change DEs often

Fine for me. (Discussion resolved, can't see the button to collapse this.)

kkartaltepe: > We could also tell people to
>    use different obs profiles for different DEs

This is not possible, profiles hold non-source/scene configuration. In this case the settings concerned are for the source which are contained in the scene collection.

columbarius: > This is not possible, profiles hold non-source/scene configuration. In this case the settings concerned are for the source which are contained in the scene collection.

Thanks. 
So on the script `s/profile/collection/`

GeorgesStavracas: Let's do it like this: if we get a reasonable number of complaints (either through Discord, or the bug tracker here) that the monitor is queried when changing DEs, I'll implement @columbarius' suggestion. Otherwise, we can deem this as not an issue. What do you think?

PR #5580, opened 2021-11-26
libobs: Add AutoRelease OBSRef wrappers for OBS types

VodBox:
### Description

These AutoRelease versions of the C++ OBSRef types do not add a ref on
construction, which is useful when accepting the result of a function
that turns a raw C pointer type that has had a reference added.

The AutoRelease types and names are taken from obs-websocket.

### Motivation and Context

Not having these types has resulted in multiple awkward anti-patterns.
Such as immediately releasing after construction to account for the
extra addref, or avoiding using the C++ type entirely and manually
releasing it later.

Example:
```
OBSSource source = obs_get_source_by_name(name.c_str());
obs_source_release(source);
```

The whole point of these types is to avoid the need for manual releases,
and rely on the RAII mechanisms inside of C++. Additionally, the
immediate release isn't commented anywhere, resulting in confusion for
other developers looking at the code as to why things are being
immediately released.

### How Has This Been Tested?

Using aspects of the UI and the frontend tools that these changes affect, observing memory leaks not going up, and observing no crashes being introduced.

### Types of changes

### Checklist:

### DIFF
--- a/UI/adv-audio-control.cpp
+++ b/UI/adv-audio-control.cpp
@@ -445,9 +445,9 @@ void OBSAdvAudioCtrl::volumeChanged(double db)
 	obs_source_set_volume(source, val);
 
 	auto undo_redo = [](const std::string &name, float val) {
-		obs_source_t *source = obs_get_source_by_name(name.c_str());
+		OBSSourceAutoRelease source =
+			obs_get_source_by_name(name.c_str());
 		obs_source_set_volume(source, val);
-		obs_source_release(source);
 	};
 
 	const char *name = obs_source_get_name(source);
@@ -468,9 +468,9 @@ void OBSAdvAudioCtrl::percentChanged(int percent)
 	obs_source_set_volume(source, val);
 
 	auto undo_redo = [](const std::string &name, float val) {
-		obs_source_t *source = obs_get_source_by_name(name.c_str());
+		OBSSourceAutoRelease source =
+			obs_get_source_by_name(name.c_str());
 		obs_source_set_volume(source, val);
-		obs_source_release(source);
 	};
 
 	const char *name = obs_source_get_name(source);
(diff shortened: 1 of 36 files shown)


UI/obs-packages/Cargo.lock
```diff
@@ -0,0 +1,243 @@
+# This file is automatically @generated by Cargo.
```
derrod, line 1: I don't think you meant to include these files?

cg2121:
Could these be expanded to scripting? As I don't see any scripts changed with this.

cg2121:
Also, would it be possible to change the existing OBSSource, OBSData, etc. to be auto release, instead of creating new auto release types?

cg2121:
Are there any reasons to use manual release still in the UI? If not, I think OBSSource, OBSData, ect. should be set to auto release, instead creating new auto release types.

VodBox:
> Are there any reasons to use manual release still in the UI? If not, I think OBSSource, OBSData, ect. should be set to auto release, instead creating new auto release types.

The existing types do run release automatically once the references/variables go out of scope. The difference is those original types also increments the reference counter while the reference is still valid. This means that if you call a function that increments the reference counter internally, it'll be assigned to the C++ type, which will increment the counter again, but perform only one release. These AutoRelease types do not perform the ref count increment step, and only handle the release aspect, making them useful for accepting the result of functions that increment the ref count.

Additionally, those types are not used just in the UI, and are available to plugins as well with C++, as it's from libobs. Changing the existing behavior/definition of the types could result in double releases, or cases where something isn't ref counted when it should have been.

VodBox:
> Could these be expanded to be used in scripting?

It is my understanding that neither Lua or Python have guaranteed ways to tell if a reference goes out of scope or is still being held, since whether things are garbage collected, or ref counted and freed immediately is an implementation detail of the runtime/compiler, and not something in the language to be relied on. This means it can be all too easy for a change in the implementation or bad detection to result in resources being held indefinitely.

jp9000:
Big thumbs up to both you and palana for this one. Big code cleanup.

tt2468:
> Are there any reasons to use manual release still in the UI? If not, I think OBSSource, OBSData, ect. should be set to auto release, instead creating new auto release types.

Changing the behavior like this would not be possible because plugins use these guards too.

tt2468 (approved):

PR #5583, opened 2021-11-27
UI: Detect cpp memory leaks

cg2121:
### Description
This overloads the cpp new/delete operators, to add an alloc counter
to detect memory leaks. This number is printed when using OBS in debug
mode when exiting.

### Motivation and Context
Makes it easier to find memory leaks in QT or cpp.

### How Has This Been Tested?
Ran the program and checked if the memory leak number was printed when exiting.

### Types of changes

### Checklist:

### DIFF
--- a/UI/obs-app.cpp
+++ b/UI/obs-app.cpp
@@ -70,6 +70,29 @@
 
 using namespace std;
 
+#ifdef _DEBUG
+void *operator new(size_t size)
+{
+	if (size == 0)
+		size++;
+
+	if (void *ptr = bzalloc(size))
+		return ptr;
+
+	throw bad_alloc{};
+}
+
+void operator delete(void *p) noexcept
+{
+	bfree(p);
+}
+
+void operator delete(void *p, size_t) noexcept
+{
+	bfree(p);
+}
+#endif
+
 static log_handler_t def_log_handler;
 
 static string currentLogFile;


UI/obs-app.cpp
```diff
@@ -69,6 +69,25 @@
 #include "ui-config.h"
 
 using namespace std;
+long alloc_count;
+
+void *operator new(size_t size)
+{
+	alloc_count++;
+	return malloc(size);
+}
```
notr1ch, line 78: This violates `new` semantics, if allocation fails a `std::bad_alloc` must be thrown.

notr1ch (commented):
I'm generally not a fan of big changes to global program state like this for such a small feature. Perhaps it could be enabled in debug only builds, but there are far better tools for finding memory leaks, and such tools may expect non-overloaded `new` / `delete` implementations in order to function.

e00E:
This approach is weird to me a for a number of reasons:

1. Will replacing new and delete like this have side effects like different behavior compared to the default versions? For example it is unclear whether the default implementation calls malloc.
2. This implementation is not thread safe. Do these functions need to be threadsafe?
3. When you log the counter you are at the end of main. But at this point destructors from local objects have not yet been called yet so you can get false positives.

A much more complete approach to this is compiler instrumentation like https://clang.llvm.org/docs/LeakSanitizer.html . I do not see the advantage of this implementation over that.

UI/obs-app.cpp
cg2121: Added the std::bad_alloc exception.

cg2121:
Updated to only overload new/delete in debug mode. Also updated to use bzalloc/bfree to use the internal OBS alloc counter, and it seems the program still works as expected.

cg2121:
Closing, as explained, there are other ways to do this.

PR #5588, opened 2021-11-28
UI/importers: Automatically detect SL Collections on macOS

gxalpha:
### Description

Enables the FindFiles function for the Streamlabs importer on macOS, which previously was
windows-only.
This is not needed for OBS Classic or XSplit, since neither exist(ed) on Mac.

### Motivation and Context

`Search known locations for scene collections when importing` would do nothing on macOS (so this is something in between a tweak and a bug fix).

### How Has This Been Tested?

Installed Streamlabs (a big pain, no idea why they would make the process so damn annoying) and created a collection.
It shows up in the importer window and is correctly identified as Streamlabs.
image

### Types of changes

### Checklist:

### DIFF
--- a/UI/importers/sl.cpp
+++ b/UI/importers/sl.cpp
@@ -503,11 +503,12 @@ bool SLImporter::Check(const string &path)
 OBSImporterFiles SLImporter::FindFiles()
 {
 	OBSImporterFiles res;
-#ifdef _WIN32
+#if defined(_WIN32) || defined(__APPLE__)
 	char dst[512];
 
-	int found = os_get_config_path(dst, 512,
-				       "slobs-client\\SceneCollections\\");
+	int found =
+		os_get_config_path(dst, 512, "slobs-client/SceneCollections/");
+
 	if (found == -1)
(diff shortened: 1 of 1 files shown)


UI/importers/sl.cpp
```diff
@@ -503,11 +503,17 @@ bool SLImporter::Check(const string &path)
...
+#if defined(_WIN32) || defined(__APPLE__)
 	char dst[512];
 
 	int found = os_get_config_path(dst, 512,
-				       "slobs-client\\SceneCollections\\");
+#if defined(_WIN32)
+				       "slobs-client\\SceneCollections\\"
+#elif defined(__APPLE__)
+				       "slobs-client/SceneCollections/"
+#endif
```
WizardCM, line 514: Hmm, do we not normalise paths at any point, removing the need for redefining the same string with different slashes?

gxalpha: Yes, and it appears the original implementation didn't use it.
I fixed that (and confirmed Windows still works).

PR #5600, opened 2021-12-02
libobs: Fix UI block caused by deleting browser source

walker-WSH:
### Description

During render loop, there are so many locks requested. If destructure func of browser source is called, block will appear as https://github.com/obsproject/obs-browser/issues/333

### Motivation and Context

fix UI block https://github.com/obsproject/obs-browser/issues/333

### How Has This Been Tested?

repeat creating/deleting browser source 

### Types of changes

### Checklist:

### DIFF
--- a/libobs/obs-video.c
+++ b/libobs/obs-video.c
@@ -1046,12 +1046,34 @@ void *obs_graphics_thread(void *param)
 	context.was_active = false;
 	context.video_thread_name = video_thread_name;
 
+	bool res = true;
+	while (res) {
+		DARRAY(struct obs_source *) temp_list;
+		da_init(temp_list);
+
+		pthread_mutex_lock(&obs->data.sources_mutex);
+		obs_source_t *source = obs->data.first_source;
+		while (source) {
+			if (!obs_source_removed(source)) {
+				obs_source_addref(source);
+				da_push_back(temp_list, &source);
+			}
+			source = (obs_source_t *)source->context.next;
+		}
+		pthread_mutex_unlock(&obs->data.sources_mutex);
+
 #ifdef __APPLE__
-	while (obs_graphics_thread_loop_autorelease(&context))
+		res = obs_graphics_thread_loop_autorelease(&context);
 #else
-	while (obs_graphics_thread_loop(&context))
+		res = obs_graphics_thread_loop(&context);
 #endif
-		;
+
+		for (size_t i = 0; i < temp_list.num; i++) {
+			struct obs_source *src = *(temp_list.array + i);
+			obs_source_release(src);
+		}
+		da_free(temp_list);
+	}
 
 #ifdef _WIN32
 	uninit_winrt_state(&winrt);


tt2468:
Since this is mainly an issue with Browser sources, it would probably be best to apply a patch to obs-browser itself.

walker-WSH:
> Since this is mainly an issue with Browser sources, it would probably be best to apply a patch to obs-browser itself.

it also fix another crash of libobs.
please refer to issue link for detail.

jp9000:
Fixed this in https://github.com/obsproject/obs-studio/commit/b2c09d35231720d0f6cc1a07f4803e8dd3846939 and https://github.com/obsproject/obs-browser/commit/dd3d7eb2ad99c1aebe7c3c6a37f411c080c7fcdf with a bit of a more of a simple solution.

I think these should be sufficient to fix both of those issues for now. Just to be clear, your solution isn't incorrect or anything, I'm mostly trying to find more optimal ways.

Sources being destroyed while inside of the mutexes could cause similar problems in the future, but I have some other ideas of how this might be handled a bit more optimally; not necessarily pretty ideas though. But it should be more optimal. For example, I was thinking about deferring source destruction either to specific cleanup points in the graphics thread (similar to what you do but only for sources that are actually destroyed), or defer them to a separate thread dedicated to source destruction entirely. I'm seriously contemplating the latter because it would ensure that obs_source_release() is always an asynchronous call rather than a potentially blocking call, which is important for time-critical threads such as the video and audio threads.

walker-WSH:
> Fixed this in [b2c09d3](https://github.com/obsproject/obs-studio/commit/b2c09d35231720d0f6cc1a07f4803e8dd3846939) and [obsproject/obs-browser@dd3d7eb](https://github.com/obsproject/obs-browser/commit/dd3d7eb2ad99c1aebe7c3c6a37f411c080c7fcdf) with a bit of a more of a simple solution.
> 
> I think these should be sufficient to fix both of those issues for now. Just to be clear, your solution isn't incorrect or anything, I'm mostly trying to find more optimal ways.
> 
> Sources being destroyed while inside of the mutexes could cause similar problems in the future, but I have some other ideas of how this might be handled a bit more optimally; not necessarily pretty ideas though. But it should be more optimal. For example, I was thinking about deferring source destruction either to specific cleanup points in the graphics thread (similar to what you do but only for sources that are actually destroyed), or defer them to a separate thread dedicated to source destruction entirely. I'm seriously contemplating the latter because it would ensure that obs_source_release() is always an asynchronous call rather than a potentially blocking call, which is important for time-critical threads such as the graphics and audio threads, and would ensure that releases can't ever directly cause a hard lock.

@jp9000 

I checked your commit. It seems only reduce the chance for block, but not fixed.

destruct function of bs is called in render thread.
isvalid()  is called in cef thread.

lock still happen if destruct func in render thread is called between isvalid() and getting graphics lock in cef thread

jp9000:
Oh I see what you're saying, my apologies; I did indeed misunderstand the situation.

jp9000:
I'll take a look into devising a solution, even if it only ends up being a solution for this specific situation. I apologize if it seems like I'm being a bit stubborn about it, I'm mostly just looking for optimal solutions.

walker-WSH:
> I'll take a look into devising a solution, even if it only ends up being a solution for this specific situation. I apologize if it seems like I'm being a bit stubborn about it, I'm mostly just looking for optimal solutions.

OK， I got it !
I am glad that you can understand my bad English ~

jp9000:
Your english is good, it was my fault for thinking that it happens in the tick call. I think what I'm going to do is defer destruction to a dedicated destroy thread or task queue. It feels a bit extreme, but I really like the idea of making obs_source_release() an asynchronous function, and I can't really think of any downside at the moment.

jp9000:
Should be fixed for real now in https://github.com/obsproject/obs-studio/commit/080090c40e81bea82754ffccbda2f625eed64e06 -- I realized that your solution is pretty much unavoidable, but should be slightly more optimal and more explicit to this specific code block at least. I was going to implement a destroy defer system but it's just a bit too much work to implement right now.

PR #5604, opened 2021-12-03
UI: Simplify OBSRef QDataStreams

cg2121:
### Description
The QDataStream can use the OBSRef pointers themselves, instead
of relying on QStrings. The OBSSignal container data stream
is used this way and works just fine.

### Motivation and Context
Code simplification, and it should also prevent any more potential crashes, like the one fixed with https://github.com/obsproject/obs-studio/commit/642dff054de1f7bfea28a9dfbf3209a83ca0bef8

### How Has This Been Tested?
Ran many operations with scenes/sources in the list widgets (drag, drop, reorder, rename, etc.)
Nothing at all seemed to break.

### Types of changes

### Checklist:

### DIFF
--- a/UI/qt-wrappers.cpp
+++ b/UI/qt-wrappers.cpp
@@ -179,47 +179,23 @@ QDataStream &operator>>(QDataStream &in,
 	return in;
 }
 
-QDataStream &operator<<(QDataStream &out, const OBSScene &scene)
+QDataStream &operator<<(QDataStream &out, const OBSScene &)
 {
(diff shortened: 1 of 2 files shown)


cg2121:
That may be the case, but it is still needed, I believe, to register a custom QVariant.

jp9000:
Are these operators required for QVariant? If they are, then you have to pass something rather than do a noop. It's probably how it stores/loads information from the QVariant.

cg2121:
Closing, as I done some more testing, and it doesn't seem to work as we want it.

I'll reopen if I find another solution.

Issue #5603, opened 2021-12-02
A crash caused by refs of obs_source without thread-safe


walker-WSH:
### Operating System Info

Windows 10

### OBS Studio Version

27.1.3

### OBS Studio Log URL

no

### Expected Behavior

refs is threasafe

### Current Behavior

when source is storaged from _obs_enum_sources_，we can't make sure if its refs has been released to -1.
If refs is -1, that means _obs_source_destory(source)_ is being invoked by another thread.
If run ++1 for source in callback，crash happen when --1 later...  

the  value of refs is changed as this order:
-1 （by another thread，destroy doing）
0（by enum thread）
-1（destroy again！）

### Steps to Reproduce

as above

### Anything else we should know?

while enumerating sources，refs should be checked

RytoEX:
Since we implemented the destruction defer system, we consider this Issue resolved, per discussion on the linked PR (#5605).

PR #5605, opened 2021-12-03
libobs: Fix crash caused by refs of obs_source

walker-WSH:
### Description

Check refs of source before calling callback, inlcuding functions:
_void obs_enum_sources(bool (*enum_proc)(void *, obs_source_t *), void *param)
void obs_enum_scenes(bool (*enum_proc)(void *, obs_source_t *), void *param)_

If refs of source has been -1, source can't be provided to callback. That mean source is being destroied by another thread.

By the way, I find some object is released when refs==-1 while some is released when ref==0. 
It's better to clean up these objects with same default refs （TO BE DONE by anyone）.

### Motivation and Context

fix crash in https://github.com/obsproject/obs-studio/issues/5603

### How Has This Been Tested?

mult-thread issue, can't reproduce easy

### Types of changes

### Checklist:


jp9000:
I'm probably going to fix this via the destruction defer system instead (see the [defer-destroy](https://github.com/obsproject/obs-studio/tree/defer-destroy) branch). Besides, we already have `obs_source_get_ref()` which palana argued is probably more ideal to be using over addref. addref should probably just be deprecated at some point.

walker-WSH:
yes，I also think the weak_source is best solution.
I never notice the interfaces for weak source before.

PR #5606, opened 2021-12-03
aja-output.cpp return false

redknotmiaoyuqiao:
### Description

obs-studio/plugins/aja/aja-output.cpp:918:10: error: cannot initialize return object of type 'void *' with an rvalue of type 'bool'


redknotmiaoyuqiao:
obs-studio/plugins/aja/aja-output.cpp:918:10: error: cannot initialize return object of type 'void *' with an rvalue of type 'bool'

norihiro:
- 1st line of the commit message should start from `aja: ` followed by a capitalized verb. For example `aja: Fix compile error ...`.
- [PR template](https://github.com/obsproject/.github/blob/master/pull_request_template.md) should be filled.

Btw, which platform causes the error?

gxalpha:
This is already part of #5591

DDRBoxman:
Closing since #5591 already covers this.

PR #5608, opened 2021-12-03
libobs: Add obs_data_get_float wrapper around get_double

myo:
The api provides an obs_data_add_float (+ _slider) but no obs_data_get_float, instead it is called obs_data_get_double (and also there is no obs_data_add_double).

### Description

simple wrapper around existing api

not the case

### Motivation and Context

solves developer headache

not the case

not the case

### How Has This Been Tested?

not the case

not the case

not the case

### Types of changes

### Checklist:

### DIFF
--- a/libobs/obs-data.c
+++ b/libobs/obs-data.c
@@ -1287,6 +1287,11 @@ double obs_data_get_double(obs_data_t *data, const char *name)
 	return obs_data_item_get_double(get_item(data, name));
 }
 
+double obs_data_get_float(obs_data_t *data, const char *name)
+{
+	return obs_data_item_get_double(get_item(data, name));
+}
+
 bool obs_data_get_bool(obs_data_t *data, const char *name)
 {
 	return obs_data_item_get_bool(get_item(data, name));


libobs/obs-data.c
```diff
@@ -1287,6 +1287,11 @@ double obs_data_get_double(obs_data_t *data, const char *name)
 	return obs_data_item_get_double(get_item(data, name));
 }
 
+double obs_data_get_float(obs_data_t *data, const char *name)
```
gxalpha, line 1290: Are you sure this is supposed to be `double [...]_get_float(`?

myo: not entirely sure, after all it will still be double precision if it's just a wrapper over _get_double

myo:
I'll leave this open so I don't have to open an issue for this small inconvenience. "Float" properties are currently stored as doubles though, so should I add an obs_properties_add_double(_slider) instead? Renaming it would probably break plugins.

jp9000:
Hi there, this PR doesn't do anything.

Firstly, it isn't even exported. In fact, it isn't even in the header file to expose it to any other source files. It's just a new unused function, which cannot be used outside of this source file because it's not exposed in the headers.

Secondly, the `obs_data_get_double()` function was completely duplicated; even the return type. It doesn't do anything different in comparison to `obs_data_get_double()`. And even if you did expose it, there would be no reason to use it because of that fact.

Thirdly, I used doubles because they can just be cast to floats. Is this PR trying to address an annoyance of having to manually cast, therefor you want to write wrapper functions that cast it for you? Because even if we did that, it would still be double internally. It just doesn't really make sense to go so far as to add additional functions just because someone doesn't like to cast a double to a float.

This PR does not serve a useful purpose. It's incomplete and improperly implemented, and the functionality it's trying to add is a wrapper for casting, which doesn't really make sense either. If everybody implemented their APIs like that, we'd have very messy APIs with a lot of superfluous functions. It'll be casted at some point or another. Just learn to live with casting.

myo:
Hi, sorry for wasting your time, was in a hurry so I didn't think much, to fix my issue I added the wrapper around get_double in my project.

**My issue is with naming:**
**there is** a: obs_data_add_float
**there is no**: obs_data_get_float

Writing code for obs in general is an annoyance for the lack of docs, but inconsistent naming makes the matter even worse..

jp9000:
You'll have to forgive me, but I don't see `obs_data_add_float()` anywhere. Were you meaning `obs_properties_add_float()`? That's for the obs_properties_t object, which is a different object. You're right in that it's inconsistent naming between the two files though, considering the properties are used modify the settings data.

PR #5611, opened 2021-12-03
UI: Add UUID to Twitch panel URLs

derrod:
### Description
Some Twitch dashboard panels now require a UUID, it is not used for
anything but local storage right now, and according to Twitch we can
generate it ourselves (or even use a static string). To be safe we can
just use QUuid to generate a "real" UUID and then store it locally until
the user disconnects the account.

### Motivation and Context

Twitch break, I fix.

### How Has This Been Tested?
Logged into panel on new profile, verified it works and UUID is saved/loaded.

### Types of changes

### Checklist:

### DIFF
--- a/UI/auth-twitch.cpp
+++ b/UI/auth-twitch.cpp
@@ -1,8 +1,10 @@
 #include "auth-twitch.hpp"
 
+#include <QRegularExpression>
 #include <QPushButton>
 #include <QHBoxLayout>
 #include <QVBoxLayout>
+#include <QUuid>
 
 #include <qt-wrappers.hpp>
 #include <obs-app.hpp>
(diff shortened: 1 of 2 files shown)


mihawk90:
For reference, the report on the browser repo:
obsproject/obs-browser#334

WizardCM:
As this UUID is specific to Activity Feed, I think the key should be named as such (UUID_af). From my experimentation, all the dashboard docks now accept a UUID parameter, even if they don't use them at the moment.

WizardCM:
This PR will fail on Linux/Ubuntu because `QUuid::Id128` was introduced in Qt 5.11.

Hello991971:
52262183?token=bf813c04e09eb220be74ea12b558cf7a

derrod:
> This PR will fail on Linux/Ubuntu because `QUuid::Id128` was introduced in Qt 5.11.

*Faint screaming in the distance*

jp9000:
Oh well, we can just use "obs" or whatever

derrod:
> Oh well, we can just use "obs" or whatever

We can still use QUuid, we just can't use the build-in formatting, and need to do a simple string replace to remove the brackets/dashes.

jp9000:
You could do what matt did on https://github.com/obsproject/obs-studio/pull/5612

Issue #5468, opened 2021-10-31
Segfault in pipewire.c (Wayland)


outfrost:
### Operating System Info

Other

### Other OS

Manjaro stable

### OBS Studio Version

27.1.3

### OBS Studio Log URL

https://obsproject.com/logs/6cmcf_nCJdt-Kqoe

### Expected Behavior

Sustained broadcast/recording with no crashes

### Current Behavior

Unpredictable crashes every 5 to 180 minutes

### Steps to Reproduce

1. Run obs on Wayland
2. Use a screen capture via PipeWire
3. Start recording
4. Wait anywhere between 5 to 180 minutes

### Anything else we should know?

I'm using:
* OBS Studio 27.1.3, built from [aur: obs-studio-browser](https://aur.archlinux.org/packages/obs-studio-browser/), with obs-browser, obs-vst, CEF compatibility patch and python binary loading patch
* KDE Plasma 5.22.5
* plasma-wayland-session
* Qt 5.15.2
* PipeWire 0.3.38
* cef-minimal 94.4.9

During recent streaming sessions, I've encountered a lot of random crashes, sometimes minutes apart, sometimes after a couple hours of streaming. Running with debug symbols under gdb has revealed a segfault at [plugins/linux-capture/pipewire.c:478](https://github.com/obsproject/obs-studio/blob/3c14e4ece250a862a5fa62aa7e85f476f7545ca4/plugins/linux-capture/pipewire.c#L478):

```
Thread 73 "PipeWire thread" received signal SIGSEGV, Segmentation fault.
[Switching to Thread 0x7ffe737fe640 (LWP 8002)]
on_process_cb (user_data=0x55555628dd20) at /usr/src/debug/obs-studio-browser/plugins/linux-capture/pipewire.c:478
478     /usr/src/debug/obs-studio-browser/plugins/linux-capture/pipewire.c: No such file or directory.
(gdb) backtrace -full
#0  on_process_cb (user_data=0x55555628dd20) at /usr/src/debug/obs-studio-browser/plugins/linux-capture/pipewire.c:478
        bitmap = 0x7fff0bfdf424
        format = GS_BGRA
        obs_pw = 0x55555628dd20
        cursor = 0x7fff0c00f728
        drm_format = 4294769916
        region = <optimized out>
        buffer = <optimized out>
        b = 0x5555588b4758
        swap_red_blue = <optimized out>
        has_buffer = <optimized out>
[...]

PR #5614, opened 2021-12-04
linux-pipewire: check pipewire buffers for corruption

columbarius:
### Description

This patch checks the buffer received from PipeWire for flags indicating that the buffer is corrupt.

### Motivation and Context

Portals can set a buffer as corrupted via flags. This allows obs to skip those frames and avoid trying to import them.

### How Has This Been Tested?

Modified version of [xdg-desktop-portal-wlr](https://github.com/emersion/xdg-desktop-portal-wlr) to alternate between those flags for a given time duration. With this patch your screencast will freeze until a new frame without a flag arrives. Mouse pointer should still be updated when it is not embedded. 

### Types of changes

### Checklist:

### DIFF
--- a/plugins/linux-pipewire/pipewire.c
+++ b/plugins/linux-pipewire/pipewire.c
@@ -454,11 +454,12 @@ static void on_process_cb(void *user_data)
 	obs_pipewire_data *obs_pw = user_data;
 	struct spa_meta_cursor *cursor;
 	uint32_t drm_format;
+	struct spa_meta_header *header;
 	struct spa_meta_region *region;
 	struct spa_buffer *buffer;
 	struct pw_buffer *b;
 	bool swap_red_blue = false;
-	bool has_buffer;
+	bool has_buffer = true;
 
 	/* Find the most recent buffer */
 	b = NULL;
@@ -478,10 +479,19 @@ static void on_process_cb(void *user_data)
 	}
 
 	buffer = b->buffer;
-	has_buffer = buffer->datas[0].chunk->size != 0;
+	header = spa_buffer_find_meta_data(buffer, SPA_META_Header,
+					   sizeof(*header));
+	if (header && (header->flags & SPA_META_HEADER_FLAG_CORRUPTED) > 0) {
+		blog(LOG_ERROR, "[pipewire] buffer is corrupt");
+		pw_stream_queue_buffer(obs_pw->stream, b);
+		return;
+	} else if (!header) {
+		has_buffer = buffer->datas[0].chunk->size != 0;
+	}
 
 	obs_enter_graphics();
 
+	// Workaround for mutter behaviour pre GNOME 43
 	if (!has_buffer)
 		goto read_metadata;
 
@@ -492,6 +502,7 @@ static void on_process_cb(void *user_data)
 		uint64_t modifiers[planes];
 		int fds[planes];
 		bool use_modifiers;
+		bool corrupt = false;
 
 		blog(LOG_DEBUG,
 		     "[pipewire] DMA-BUF info: fd:%ld, stride:%d, offset:%u, size:%dx%d",
@@ -514,6 +525,14 @@ static void on_process_cb(void *user_data)
 			offsets[plane] = buffer->datas[plane].chunk->offset;
 			strides[plane] = buffer->datas[plane].chunk->stride;
 			modifiers[plane] = obs_pw->format.info.raw.modifier;
+			corrupt |= (buffer->datas[plane].chunk->flags &
+				    SPA_CHUNK_FLAG_CORRUPTED) > 0;
+		}
+
+		if (corrupt) {
+			blog(LOG_ERROR,
+			     "[pipewire] buffer contains corrupted data");
+			goto read_metadata;
 		}
 
 		g_clear_pointer(&obs_pw->texture, gs_texture_destroy);
@@ -544,6 +563,19 @@ static void on_process_cb(void *user_data)
 			blog(LOG_ERROR,
 			     "[pipewire] unsupported DMA buffer format: %d",
 			     obs_pw->format.info.raw.format);
+			goto read_metadata;
+		}
+
+		if ((buffer->datas[0].chunk->flags & SPA_CHUNK_FLAG_CORRUPTED) >
+		    0) {
+			blog(LOG_ERROR,
+			     "[pipewire] buffer contains corrupted data");
+			goto read_metadata;
+		}
+
+		if (buffer->datas[0].chunk->size == 0) {
+			blog(LOG_ERROR,
+			     "[pipewire] buffer contains empty data");
 			goto read_metadata;
 		}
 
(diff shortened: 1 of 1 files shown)


plugins/linux-capture/pipewire.c
```diff
@@ -472,6 +488,11 @@ static void on_process_cb(void *user_data)
 			     obs_pw->format.info.raw.format);
 			goto read_metadata;
 		}
+		if ((buffer->datas[0].chunk->flags & SPA_CHUNK_FLAG_CORRUPTED) >
+		    0) {
+			blog(LOG_ERROR, "[pipewire] buffer plane corrupt");
```
kkartaltepe, line 493: Should this be log_debug like the other reports?

kkartaltepe:
Is this supposed to solve #5468 ?

plugins/linux-capture/pipewire.c
columbarius: Oh thanks, missed to change that back.

columbarius:
> Is this supposed to solve #5468 ?

No

GeorgesStavracas (changes requested):
Commit message typo: `metadate` → `metadata`

Could you expand a bit more the context around these CORRUPTED flags? The documentation I found is painfully small, and I don't think I understand when nor how this flag is used. Is it PipeWire itself that sets it? The compositor? Is skipping the frame the best strategy to handle it?

columbarius:
> Could you expand a bit more the context around these CORRUPTED flags? The documentation I found is painfully small, and I don't think I understand when nor how this flag is used. Is it PipeWire itself that sets it? The compositor? Is skipping the frame the best strategy to handle it?

This flag is set by the producer, in our case the compositor. Afaik this flag can be used to mark "broken" buffers, without specifying what the issue is. xdpw will use that flag if a buffer copy by the compositor failed. Since we manage the buffer the memory won't be invalid but the content might be corrupt (filled with garbage). The v4l2 plugin [1] uses this flag for the same purpose. Since a buffer with this flag won't contain wanted contend I can't think of a better solution than skipping it.  In my testing obs just reused the last imported texture and continued after a non corrupted frame was received.

[1] https://gitlab.freedesktop.org/pipewire/pipewire/-/blob/master/spa/plugins/v4l2/v4l2-utils.c#L1264

columbarius:
Updated this to incorporate both meta and junk flags and added a revertible commit for current GNOME behaviour.

The hinted documentation for ignoring the size in case of DMA-BUFs will be written by me probably this week, but got an ack from mesa devs. Will link the PipeWire MR here.

columbarius:
> The documentation I found is painfully small

Try to extend that at the same time.

columbarius:
https://gitlab.freedesktop.org/pipewire/pipewire/-/merge_requests/1189

GeorgesStavracas (changes requested):
Two comments, but this looks generally good to me.

GeorgesStavracas:
I forgot to mention, but could you please rebase against the main branch, and remove the merge commits?

plugins/linux-pipewire/pipewire.c
```diff
@@ -480,7 +481,15 @@ static void on_process_cb(void *user_data)
...
-	has_buffer = buffer->datas[0].chunk->size != 0;
+	if ((header = spa_buffer_find_meta_data(buffer, SPA_META_Header,
+						sizeof(*header)))) {
+		if ((header->flags & SPA_META_HEADER_FLAG_CORRUPTED) > 0) {
+			blog(LOG_ERROR, "[pipewire] buffer is corrupt");
+			goto queue_buffer;
+		}
+	} else {
+		has_buffer = buffer->datas[0].chunk->size != 0;
+	}
```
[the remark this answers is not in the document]
columbarius: we can't do the `has_buffer` assignment. DMABUFS are allowed to have a `chunk->size = 0` and be valid. The reason is that you don't need the size of a DMABUF to import it and gbm has no api to query it. You could do an fseek on the fd, but that would not be useful in handling the buffer anyways, as such PipeWire just allows to omit this value (set it to 0, so clients which mmap the fd won't map anything bad).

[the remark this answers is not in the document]
columbarius: I'd rather remove the size check at all, but we need it for backwards compatibility for now.

columbarius:
> I forgot to mention, but could you please rebase against the main branch, and remove the merge commits?

It's gone. Probably didn't changed the strategy when using githubs update feature.

plugins/linux-pipewire/pipewire.c
```diff
@@ -616,6 +641,7 @@ static void on_process_cb(void *user_data)
 		obs_pw->cursor.y = cursor->position.y;
 	}
 
+queue_buffer:
```
GeorgesStavracas, line 650: This label is unused now

plugins/linux-pipewire/pipewire.c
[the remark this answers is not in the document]
GeorgesStavracas: Hm, alright then

GeorgesStavracas (changes requested):
I understand why you split this in 2 separate commits - and it actually helped me review it - but I'd prefer if you could squash the second commit into the first one before merge. Please mention the same GNOME pre 43 behavior in the first commit's message. (The reason I prefer squashing is to not have any point in the commit history where the capture won't work with a subset of consumers.)

plugins/linux-pipewire/pipewire.c
columbarius: Nice catch. Thanks!

GeorgesStavracas (approved):

columbarius:
> I understand why you split this in 2 separate commits - and it actually helped me review it - but I'd prefer if you could squash the second commit into the first one before merge. Please mention the same GNOME pre 43 behavior in the first commit's message. (The reason I prefer squashing is to not have any point in the commit history where the capture won't work with a subset of consumers.)

Squashed the commit and added the note to the first commit message. Additionally I marked the workaround in code with a comment, such that it will be recognized as such and can be easily removed later.

I'll continue to use multiple commits to ease development and review, but I'm totally fine to squash them just before merging.

GeorgesStavracas (approved):
One last nit: please use `linux-pipewire:` as the commit prefix; and please start the tile in upper case. In other words, this: `linux-pipewire: Reject invalid buffers`

columbarius:
> One last nit: please use `linux-pipewire:` as the commit prefix; and please start the tile in upper case. In other words, this: `linux-pipewire: Reject invalid buffers`

Done ... `linux-capture`... what a nod from the past ^^`

GeorgesStavracas:
> Done ... linux-capture... what a nod from the past ^^`
Heh the good, old, simpler times :slightly_smiling_face:

jp9000:
This change spams my log with corruption messages even though it seems to be capturing fine. Fedora 37

columbarius:
@jp9000 is it `[pipewire] buffer is corrupt` or `[pipewire] buffer contains corrupted data` ?
The second one should also appear when the buffer only contains an cursor update without new screen data. Maybe we should change the loglevel or message.

columbarius:
@GeorgesStavracas thanks for merging!

jp9000:
@columbarius It's `[pipewire] buffer contains corrupted data`, the one on line 533. I just noticed it happens if I have OBS on a monitor different from the one being captured. If OBS is capturing the same monitor it's fine, but when I move it onto my other monitor, it starts spamming that in the logs in spite of seemingly capturing fine.

PR #5617, opened 2021-12-04
mac-capture: add WindowCapture.ShowCursor

kilinbox:
### Description
Add cursor display function to window capture.

### Motivation and Context
Cursor display in window capture is available in Windows but not in Mac, and even in Mac, cursor display is available in Display Capture but not in window capture. So, we will implement the cursor display in window capture.

### How Has This Been Tested?
macOS 12.0.1 Montere, Mac mini(M1, 2020), Mac Book Pro(2016) 

### Checklist:


PatTheMav:
I'm certainly not happy that this uses a private, undocumented, CoreGraphics API, but it looks like a few other projects use this as well. Might break at any point though, so there's that caveat.

Did you measure the performance impact between capturing with and without cursor, given that you introduce an additional draw call to an already heavy draw call? I understand this feature is desirable, but it looks like it might have the side-effect of making the already bad performance of window capture even worse.

kilinbox:
> I'm certainly not happy that this uses a private, undocumented, CoreGraphics API, but it looks like a few other projects use this as well. Might break at any point though, so there's that caveat.
> 
> Did you measure the performance impact between capturing with and without cursor, given that you introduce an additional draw call to an already heavy draw call? I understand this feature is desirable, but it looks like it might have the side-effect of making the already bad performance of window capture even worse.

I'm not happy with the use of private, undocumented either, but I added this feature because some users feel that they need the same cursor display as in Windows.

Enabling this, will increase the CPU usage compared to the current OBS.

vlcsnap-2021-12-07-14h35m02s781

https://imgur.com/a/jbn83vE

kilinbox:
#5875 closed for better results.

PR #5620, opened 2021-12-05
libobs-opengl: Ensure proper draw buffer

kkartaltepe:
### Description
This commit ensures that we set the appropriate draw buffer when making
a context current. Mesa drivers enforce opengl ES semantics where the
targets passed to eglMakeCurrent are bound, but nvidia instead ignores
these parameters after the 1st eglMakeCurrent. In obs we make current
with EGL_NO_SURFACE so our draw targets end up as EGL_NONE on nvidia
and previews fail to render.

This also removes NATIVE_RENDERABLE requirements. Nvidia driver does not
report support for this attribute on any context and after resolving the
draw target issues previews render correctly on nvidia and intel
drivers.

Id appreciate someone on AMD hardware to test as well.

### Motivation and Context
I guess it's nice for users on nvidia hardware to use OBS.

### How Has This Been Tested?
Tested on sway on nvidia and sway on intel. Qt is still busted on nvidia so OBS ends up hanging after you interact with it a couple times. But that is outside of our control.

### Types of changes

- Bug fix (non-breaking change which fixes an issue) (At least nvidia considers our/mesa's interpretation of eglMakeCurrent a bug :) )

### Checklist:


puyoxyz:
Works here, on a GTX 1060 6GB, GNOME 41.2 on Wayland, Artix Linux

> "Qt is still busted on nvidia so OBS ends up hanging after you interact with it a couple times"

Cannot reproduce this. Although it's laggy, it works perfectly, and I can record with PipeWire. (And it doesn't happen on other Qt apps, so not a Qt problem?)

puyoxyz (approved):

kkartaltepe:
> Cannot reproduce this. Although it's laggy, it works perfectly, and I can record with PipeWire. (And it doesn't happen on other Qt apps, so not a Qt problem?)

Well the lag is the same issue, however the deadlocking occurs after applying pending patches for the "lag". Currently it seems both are caused by the interaction between qt's wayland platform and nvidia's wayland drivers. If both of these are not running in the same process you likely wont find any issue (such as software rendered qt applications or xwayland based qt applications), however OBS uses both.

kkartaltepe:
@GeorgesStavracas can you take a look? Someone mentioned you had some concerns for AMD drivers, I have set up the new attributes as a fallback. However since this configuration value is hardcoded to true on mesa for Intel it seemed that it really didnt matter at all, so we are probably better off removing it.

GeorgesStavracas (approved):
Thanks, I'm more comfortable approving this PR with the fallback path. Seems like this is what NVidia developers have submitted for Firefox too, so I don't see any point in not doing that as well.

PR #5625, opened 2021-12-06
libobs: Ignore loading dll which is not an OBS plugin

walker-WSH:
### Description

I guess here lost "return". @notr1ch 

### Motivation and Context

reduce time on startup

### How Has This Been Tested?

Ignore dll which is not OBS plugin while loading plugins

### Types of changes

### Checklist:


notr1ch:
Thanks, updated the commit title to better reflect this change and merged.

walker-WSH:
@notr1ch 

I begin to doubt whether it is worth adding this code. We find the time for startup has been increased obviously.
In debug mode, sometimes it can take 50~300ms on checking a dll.

In fact, the number of dll which is not an OBS plugin is not so much.

notr1ch:
Are you sure that's not from external interference from virus scanners or similar? I did not see such results in my testing, even on a HDD this requires only a handful of seeks compared to a full DLL load.

walker-WSH:
> Are you sure that's not from external interference from virus scanners or similar? I did not see such results in my testing, even on a HDD this requires only a handful of seeks compared to a full DLL load.

Please forget that. It must be affected by any other software or environment.
I can't reproduce it today. Today it takes only 0~2ms for every DLL.

PR #5644, opened 2021-12-10
UI: Add Rosetta Detection

gxalpha:
### Description

Adds a check that detects if OBS is running on Rosetta on macOS and logs
the result.
~~Also adds a popup window that asks the user wether OBS is intended to
run on Rosetta. This window is currently disabled, and should get
get enabled once we officially support ARM natively.~~

### Motivation and Context

With OBS most likely being compilable natively with the next release, it
will be good to know which version users are using when they face
issues (for example with third party plugins not working).

### How Has This Been Tested?

macOS 12.0.1, Intel: 
- Compiled and Run
- Log correctly confirms that it's not translated.

A friends Mac mini (using a CI build of this branch), macOS 12.0.1, M1:
- Log correctly reports translation via Rosetta.

### Types of changes

### Checklist:

### DIFF
--- a/UI/obs-app.cpp
+++ b/UI/obs-app.cpp
@@ -2116,6 +2116,12 @@ static int run_program(fstream &logFile, int argc, char *argv[])
 		}
 #endif
 
+#ifdef __APPLE__
+		bool rosettaTranslated = ProcessIsRosettaTranslated();
+		blog(LOG_INFO, "Rosetta translation used: %s",
+		     rosettaTranslated ? "true" : "false");
+#endif
+
 		if (!created_log) {
 			create_log_file(logFile);
 			created_log = true;
--- a/UI/platform-osx.mm
+++ b/UI/platform-osx.mm
@@ -23,6 +23,7 @@
 #include "obs-app.hpp"
 
 #include <unistd.h>
+#include <sys/sysctl.h>
 
 #import <AppKit/AppKit.h>
 
@@ -233,6 +234,19 @@ void EnableOSXDockIcon(bool enable)
 	else
 		[NSApp setActivationPolicy:
 				NSApplicationActivationPolicyProhibited];
+}
+
+bool ProcessIsRosettaTranslated()
+{
+#ifdef __aarch64__
+	return false;
+#else
+	int ret = 0;
+	size_t size = sizeof(ret);
+	if (sysctlbyname("sysctl.proc_translated", &ret, &size, NULL, 0) == -1)
+		return false;
+	return ret == 1;
+#endif
 }
 
 /*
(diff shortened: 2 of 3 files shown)


Warchamp7:
The warning text reads a bit awkward to me right now, but I'm not familiar enough with Mac to suggest a correction.

You able to aid here @PatTheMav?

feetstv:
Just a comment: if there were to be a message, it should follow the [macOS HIG](https://developer.apple.com/design/human-interface-guidelines/macos/windows-and-views/alerts/) (since macOS is the only platform where it'll be seen) in terms of the text. 

The message should begin with the question phrase, then get followed by explanatory information. For example:

**Are you sure you want to run the Intel version of OBS?**

For better performance on your Mac, download the Apple Silicon version from the [OBS Project website](https://obsproject.com/)

Close - **Launch anyway**

PatTheMav:
As always the answer is: Follow Apple’s published HIG. 😁

gxalpha:
Updated to look like this:
image
The text is rather small, but this is [intended from QT's side](https://doc.qt.io/qt-5/qmessagebox.html#informativeText-prop) (probably to mimic how it would look with a native alert as shown in Apple's HIG), so I guess it's fine.

jp9000:
So I guess this should be merged after we finish up the M1 stuff then?

feetstv:
The text matches the size of normal NSAlert message text and information text, so it's fine.

Perhaps the message text should say "Are you sure you want to **launch** the Intel version of OBS?" (emphasis added) to match with the button text.

Also, should the link to the OBS website be within the information text, or should it be the default button? Feels like it would make more sense as a default action. Mock-up (using system NSAlert because I can't be bothered to install Qt) attached:

Screen Shot 2021-12-12 at 4 20 09 PM

WizardCM:
I'd say remove the `#if 0` because this won't be merged until it's needed for 28.0 anyway.

gxalpha:
Perhaps I miscommunicated this, but the dialog really wasn't supposed to be the main thing of this PR.

My concern is that once OBS will be compilable on ARM (which it will be with updated CEF), people will inevitably do so, no matter what we say about how unofficial it is; and then potentially come into support because stuff (plugins) breaks. In fact, I remember this happening in past (apparently there are build scripts in the wild), and personally expect this to increase once we mention that OBS can unofficially work natively (which iirc was the plan for 27.2).
This was the motivation behind having the logging, so that such builds can easily be identified.

If it would help the logging get merged faster, I'd be perfectly happy removing the popup stuff from the code entirely and put that into another PR (which then would sit there until official ARM support).
Though of course if my concerns about potentially increased support requests seem like not big of a deal (I don't have data or anything, they are just feelings after all), sure I'll remove the `if 0` and have this PR wait.

tommyvct:
This will be helpful for Windows ARM64 as well.

WizardCM:
In that case yes, please split this into two separate PRs.

gxalpha:
Removed popup part and updated Readme accordingly.

PR #5650, opened 2021-12-12
libobs: Add wayland hotkey infrastructure

kkartaltepe:
### Description

Users on wayland are displeased that they cannot see their hotkey
bindings. This enables keys reported like X11, and has the infrastructure
in place in case wayland ever decides to allow for capturing input.

Also fix numpad subtract missing on X11.

### Motivation and Context

Users cannot even see or set hotkeys which might be usable while OBS is in focus. Hotkeys now work while OBS is in focus and display properly on the UI.

### How Has This Been Tested?

Tested in kde that keymapping tables work correctly on US layouts and the keys on my keyboard that work in X11 work in Wayland.

More testing on non-US layouts would be appreciated.

### Types of changes

### Checklist:

### DIFF
--- a/libobs/obs-nix-wayland.c
+++ b/libobs/obs-nix-wayland.c
@@ -19,7 +19,232 @@
 #include "obs-nix-platform.h"
 #include "obs-nix-wayland.h"
 
+#include <unistd.h>
+#include <sys/mman.h>
+#include <string.h>
+
 #include <wayland-client.h>
+#include <xkbcommon/xkbcommon.h>
+
+// X11 only supports 256 scancodes, most keyboards dont have 256 keys so this should be reasonable.
+#define MAX_KEYCODES 256
+// X11 keymaps only have 4 shift levels, im not sure xkbcommon supports a way to shift the state into a higher level anyway.
+#define MAX_SHIFT_LEVELS 4
+
+struct obs_hotkeys_platform {
+	struct wl_display *display;
+	struct wl_seat *seat;
+	struct wl_keyboard *keyboard;
+	struct xkb_context *xkb_context;
+	struct xkb_keymap *xkb_keymap;
+	struct xkb_state *xkb_state;
+	xkb_keysym_t key_to_sym[MAX_SHIFT_LEVELS][MAX_KEYCODES];
+	xkb_keysym_t obs_to_key[OBS_KEY_LAST_VALUE];
+	uint32_t current_layout;
+};
+
+static obs_key_t obs_nix_wayland_key_from_virtual_key(int sym);
+
+static void load_keymap_data(struct xkb_keymap *keymap, xkb_keycode_t key,
+			     void *data)
+{
+	obs_hotkeys_platform_t *plat = (obs_hotkeys_platform_t *)data;
+	if (key >= MAX_KEYCODES)
+		return;
+
+	const xkb_keysym_t *syms;
+	for (int level = 0; level < MAX_SHIFT_LEVELS; level++) {
+		int nsyms = xkb_keymap_key_get_syms_by_level(
+			keymap, key, plat->current_layout, level, &syms);
+		if (nsyms < 1)
+			continue;
+
+		obs_key_t obs_key =
+			obs_nix_wayland_key_from_virtual_key(syms[0]);
+		// This avoids ambiguity where multiple scancodes produce the same symbols.
+		// e.g. LSGT and Shift+AB08 produce `<` on default US layout.
+		if (!plat->obs_to_key[obs_key])
+			plat->obs_to_key[obs_key] = key;
+		plat->key_to_sym[level][key] = syms[0];
+	}
+}
+
+static void rebuild_keymap_data(obs_hotkeys_platform_t *plat)
+{
+	memset(plat->key_to_sym, 0,
+	       sizeof(xkb_keysym_t) * MAX_SHIFT_LEVELS * MAX_KEYCODES);
+	memset(plat->obs_to_key, 0, sizeof(xkb_keysym_t) * OBS_KEY_LAST_VALUE);
+	xkb_keymap_key_for_each(plat->xkb_keymap, load_keymap_data, plat);
+}
+
+static void platform_keyboard_keymap(void *data, struct wl_keyboard *keyboard,
+				     uint32_t format, int32_t fd, uint32_t size)
+{
+	UNUSED_PARAMETER(keyboard);
+	UNUSED_PARAMETER(format);
+	obs_hotkeys_platform_t *plat = (obs_hotkeys_platform_t *)data;
+
+	char *keymap_shm = mmap(NULL, size, PROT_READ, MAP_SHARED, fd, 0);
+	if (keymap_shm == MAP_FAILED) {
+		close(fd);
+		return;
+	}
+
+	struct xkb_keymap *xkb_keymap = xkb_keymap_new_from_string(
+		plat->xkb_context, keymap_shm, XKB_KEYMAP_FORMAT_TEXT_V1,
+		XKB_KEYMAP_COMPILE_NO_FLAGS);
+	munmap(keymap_shm, size);
+	close(fd);
+
+	// cleanup old keymap and state
+	xkb_keymap_unref(plat->xkb_keymap);
+	xkb_state_unref(plat->xkb_state);
+
+	plat->xkb_keymap = xkb_keymap;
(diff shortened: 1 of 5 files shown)


kkartaltepe:
Also most of the key translation was done programmatically, some keys may be missing the only ones I added manually were numpad keys.

libobs/obs-nix-x11.c
```diff
@@ -1179,6 +1179,8 @@ static void obs_nix_x11_key_to_str(obs_key_t key, struct dstr *dstr)
 		return translate_key(key, "Numpad *");
 	case OBS_KEY_NUMPLUS:
 		return translate_key(key, "Numpad +");
+	case OBS_KEY_NUMMINUS:
+		return translate_key(key, "Numpad -");
```
WizardCM, line 1183: This should likely be a separate commit.

kkartaltepe: Sounds good.

libobs/obs-nix-wayland.c
```diff
@@ -19,7 +19,224 @@
...
+{
+	obs_hotkeys_platform_t *plat = (obs_hotkeys_platform_t *)data;
+	if (key >= MAX_KEYCODES)
+		return;
+
+	const xkb_keysym_t *syms;
+	// My setup doesnt have multiple layouts, do people actually use these?
+	// int num_layouts = xkb_keymap_num_layouts_for_key(keymap, key);
+	// blog(LOG_INFO, "key layouts: %d", num_layouts);
+	for (int l = 0; l < MAX_SHIFT_LEVELS; l++) {
```
GeorgesStavracas, line 58: `int level` perhaps?

libobs/obs-nix-wayland.c
```diff
@@ -19,7 +19,224 @@
...
+#include <sys/mman.h> // mmap
+#include <string.h>   // strcmp
+
 #include <wayland-client.h>
+#include <xkbcommon/xkbcommon.h>
+
+// X11 only supports 256 scancodes, most keyboards dont have 256 keys so this should be reasonable.
+#define MAX_KEYCODES 256
+// X11 keymaps only have 4 shift levels, im not sure xkbcommon supports a way to shift the state into a higher level anyway.
+#define MAX_SHIFT_LEVELS 4
```
GeorgesStavracas, line 32: Can this number be derived from the current keymap?

libobs/obs-nix-wayland.c
```diff
@@ -19,7 +19,224 @@
...
 
+#include <unistd.h>   // close
+#include <sys/mman.h> // mmap
+#include <string.h>   // strcmp
+
 #include <wayland-client.h>
+#include <xkbcommon/xkbcommon.h>
+
+// X11 only supports 256 scancodes, most keyboards dont have 256 keys so this should be reasonable.
+#define MAX_KEYCODES 256
```
GeorgesStavracas, line 30: Same for this one

libobs/obs-nix-wayland.c
```diff
@@ -19,7 +19,224 @@
...
+static void platform_keyboard_keymap(void *data, struct wl_keyboard *keyboard,
+				     uint32_t format, int32_t fd, uint32_t size)
+{
+	UNUSED_PARAMETER(keyboard);
+	UNUSED_PARAMETER(format);
+	obs_hotkeys_platform_t *plat = (obs_hotkeys_platform_t *)data;
+
+	char *keymap_shm = mmap(NULL, size, PROT_READ, MAP_SHARED, fd, 0);
+	if (keymap_shm == MAP_FAILED) {
+		return;
```
GeorgesStavracas, line 83: I'm not sure if it's necessary to `close(fd)` here as well

libobs/obs-nix-wayland.c
```diff
@@ -51,24 +281,1337 @@ obs_nix_wayland_hotkeys_platform_is_pressed(obs_hotkeys_platform_t *context,
...
+	/*
+	obs_hotkeys_platform_t *context = obs->hotkeys.platform_context;
+	struct keycode_list *keycodes = &context->keycodes[key];
+
+	for (size_t i = 0; i < keycodes->list.num; i++) {
+		if (get_key_translation(dstr, keycodes->list.array[i])) {
+			break;
+		}
+	}
+	*/
```
GeorgesStavracas, line 393: I'm not sure what were you intending to test here. Was it multi-layout?

libobs/obs-nix-wayland.c
```diff
@@ -51,24 +281,1337 @@ obs_nix_wayland_hotkeys_platform_is_pressed(obs_hotkeys_platform_t *context,
...
+	obs_hotkeys_platform_t *plat = obs->hotkeys.platform_context;
+	// Translate the obs key back down to shift level 1 and then back to obs key.
+	xkb_keysym_t base_sym = plat->key_to_sym[plat->obs_to_key[key]][0];
```
GeorgesStavracas, line 396: Nitpick: could you please put `plat->obs_to_key[key]` in a separate variable?

GeorgesStavracas (changes requested):
I left some comments and questions, but overall this looks pretty good to me. I didn't think of doing a roundtrip to get the wl_keyboard, what I had in mind is somehow get that from Qt, but that's probably a bad idea. What's here is an interesting approach.

libobs/obs-nix-wayland.c
kkartaltepe: Oh sorry this was copied from the X11 impl, meant to delete.

libobs/obs-nix-wayland.c
kkartaltepe: These could be derived but if we consider non-X11 compatible both keycodes and shift levels have no limit. So such a simple implementation may not be feasible. Only supporting X11 compatible keymaps seems like a nice middle ground of simplicity and supporting all existing (x11 compatible) layouts.

libobs/obs-nix-wayland.c
```diff
@@ -19,7 +19,224 @@
...
+			     void *data)
+{
+	obs_hotkeys_platform_t *plat = (obs_hotkeys_platform_t *)data;
+	if (key >= MAX_KEYCODES)
+		return;
+
+	const xkb_keysym_t *syms;
+	// My setup doesnt have multiple layouts, do people actually use these?
+	// int num_layouts = xkb_keymap_num_layouts_for_key(keymap, key);
+	// blog(LOG_INFO, "key layouts: %d", num_layouts);
```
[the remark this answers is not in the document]
kkartaltepe: based on https://www.x.org/releases/X11R7.6/doc/xorg-docs/input/XKB-Config.html#id2520758 and an overview of the keyboard layouts preloaded on archlinux. There are no multi-layout keyboards defined by default, the only case for this would be users with custom defined xkb keymaps (so like 3 people?) where there appears to be like 4 possible layout levels.

I dont mind also hardcoding in layouts if we assume there can be up to 4 of them, though im not sure we really want to support arbitrary groups too, especially not if IMEs do change keymaps instead.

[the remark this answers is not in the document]
kkartaltepe: I played with some IME's and gnome, it looks like IMEs will use their own path for inputting characters for text input and do not use this mechanism at all.

For users configuring keyboards via gnome settings and not using IMEs gnome will only ever configure 4 layouts/groups at a time to remain X11 compatible. If you switch past the 4th keyboard gnome will reconfigure the wayland keyboard.

In my minimal playing this code seems to work fine for multiple languages once we handle multiple layouts/groups.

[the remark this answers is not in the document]
GeorgesStavracas: Alright, thanks for checking this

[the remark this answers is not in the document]
kkartaltepe: Pushed changes, I build the mapping tables any time users switch layouts. In cases where a user has something like a qwerty and azerty layout, scancode for AD01 might be bound to 'q' and 'a' which isnt quite correct. Switching keymaps takes 0.016ms on my machine so I think there is no problem with computing the lookup tables when users swap layouts to make the code simpler.

GeorgesStavracas (changes requested):
The code looks good to me, I think it's good to go. However, I'd like to ask capitalizing `wayland` → `Wayland`

kkartaltepe:
Updated, added some version checks when blinding `wl_seat`, checking the definition nothing important has changed for the past 8 years since v4 so we will bind whatever the compositor reports up to v7.

I think this is good to merge now.

GeorgesStavracas (approved):

PR #5501, opened 2021-11-06
Fix FFmpeg deprecations up to FFmpeg 4.4 and current FFmpeg git

RytoEX:
### Description

These commits address various FFmpeg changes, and deprecations.  They also address removals of deprecated code post FFmpeg 4.x.

Details about the changes and why they are needed are in each commit message.  As far as I can tell, these changes are compatible for the entire FFmpeg 4.x series, and _should_ be fine for FFmpeg 3.2+ releases.

Big thanks to @pkviet who helped me with this!

### Motivation and Context

These changes allow us to use FFmpeg 4.4+ up to FFmpeg to FFmpeg/FFmpeg@9d4989f2e17c11d4ee9daa489b6e4f5195d80b99.

### How Has This Been Tested?

On Windows 10 with FFmpeg 4.2.4, FFmpeg 4.4.1, and FFmpeg git (ffmpeg-N-104473-g9d4989f2e1, FFmpeg/FFmpeg@9d4989f2e17c11d4ee9daa489b6e4f5195d80b99), I successfully tested:
* streaming to Twitch with NVENC and x264
* recording to MKV and MP4 with NVENC, x264, and QSV
* Color Source with Color Correction filter
* Media Source with local files (MP4; WEBM with transparency)
* Media Source with Chroma Key filter
* Image Source with JPG and animated GIF
* Text Source (both GDI+ and FreeType2)
* Video Capture Device
* VLC Source with multiple files selected

Resulting recordings seems fine in VLC.

Needs additional testing on macOS and Linux, for which assistance would be appreciated.

### Types of changes

- Bug fix (non-breaking change which fixes an issue)

- Tweak (non-breaking change to improve existing functionality)

### Checklist:

- [x] My code has been run through [clang-format](https://github.com/obsproject/obs-studio/blob/master/.clang-format).
- [x] I have read the [**contributing** document](https://github.com/obsproject/obs-studio/blob/master/CONTRIBUTING.rst).
- [x] My code is not on the master branch.
- [x] The code has been tested.
- [x] All commit messages are properly formatted and commits squashed where appropriate.
- [x] I have included updates to all appropriate documentation.

PR #5653, opened 2021-12-12
Fix some compillation warnings that appear on Linux

tytan652:
### Description

This PR is meant to fix some compilation warnings, ignoring FFmpeg deprecation, AJA NT2 and FTL related ones:
 - obs-ffmpeg-vaapi emits a format-overflow warning.

 - obs-outputs emits a warning because of a pointer type mismatch in conditional expression and an unused-parameter warning.
 - v4l2-output emits a format-truncation warning.
 -  media-playback, test/cmocka and obs-ffmpeg emits unused parameter warnings in the CI.
 - obs-scripting: PyEval_InitThreads() and PyEval_ThreadsInitialized() are deprecated since Python 3.9. Fixes based on https://github.com/swig/swig/commit/9c50887daa2b44251e77be8123c63df238034cf5.
 - obs-scripting (python wrap) emits a warning because of an unused parameter. A workaround to put the parameter as in a `UNUSED_PARAMETER()` was made. It really just adds `(void)varargs;` to the wrapper code.

#### Unfixed warning
I just wanted to list some warning that I couldn't fix or that can't be fixed:
 - `libobs/util/base.c` emits a warning because a ‘noreturn’ function does return. Can't be fixed because SWIG seems to not support the `noreturn` attribute.
 - ~A PyEval_InitThreads() deprecation warning is still emitted but will be fixed once SWIG 4.1 is out and used by OBS.~ (fixed for version earlier than 4.1)
 - FFmpeg deprecation, a few are normally fixed by #5501.
 - obs-ffmpeg-av1 emits an absolute-value warning caused by making the absolute value of an unsigned type.
 - And warnings from AJA plugins since I can't fix all of them.

### Motivation and Context

Less there is warning, the better is. No ?
Yes, I don't like warnings so I tried to fix what I can fix.

### How Has This Been Tested?

Just build this branch on Arch Linux (with and without obs-browser) and in the Ubuntu CI and no warning were emitted where it should be fixed.
obs-ffmpeg-vaapi and v4l2-output changes were tested.

### Types of changes

### Checklist:

### DIFF
--- a/plugins/linux-v4l2/v4l2-output.c
+++ b/plugins/linux-v4l2/v4l2-output.c
@@ -197,9 +197,14 @@ static bool virtualcam_start(void *data)
 		return false;
 
 	for (int i = 0; i < n; i++) {
-		char device[32];
+		char device[32] = {0};
 
-		snprintf(device, 32, "/dev/%s", list[i]->d_name);
+		// Use the return value of snprintf to prevent truncation warning.
+		int written = snprintf(device, 32, "/dev/%s", list[i]->d_name);
+		if (written >= 32)
+			blog(LOG_DEBUG,
+			     "v4l2-output: A format truncation may have occurred."
+			     " This can be ignored since it is quite improbable.");
 
 		if (try_connect(vcam, device)) {
 			success = true;
--- a/plugins/obs-ffmpeg/obs-ffmpeg-vaapi.c
+++ b/plugins/obs-ffmpeg/obs-ffmpeg-vaapi.c
@@ -600,8 +600,16 @@ static obs_properties_t *vaapi_properties(void *unused)
 			    strcmp(file_name, "..") == 0)
 				continue;
 
-			char path[64] = "\0";
-			sprintf(path, "/dev/dri/by-path/%s", file_name);
+			char path[64] = {0};
+
+			// Use the return value of snprintf to prevent truncation warning.
+			int written = snprintf(path, 64, "/dev/dri/by-path/%s",
+					       file_name);
+			if (written >= 64)
+				blog(LOG_DEBUG,
+				     "obs-ffmpeg-vaapi: A format truncation may have occurred."
+				     " This can be ignored since it is quite improbable.");
+
 			type = strrchr(file_name, '-');
 			if (type == NULL)
 				continue;
--- a/UI/auth-oauth.cpp
+++ b/UI/auth-oauth.cpp
@@ -64,6 +64,8 @@ OAuthLogin::OAuthLogin(QWidget *parent, const std::string &url, bool token)
 	QVBoxLayout *topLayout = new QVBoxLayout(this);
 	topLayout->addWidget(cefWidget);
 	topLayout->addLayout(bottomLayout);
+#else
+	UNUSED_PARAMETER(url);
 #endif
 }
 
--- a/UI/auth-youtube.cpp
+++ b/UI/auth-youtube.cpp
@@ -181,6 +181,8 @@ void YoutubeAuth::SetChatId(QString &chat_id)
 	if (chat && chat->cefWidget) {
 		chat->cefWidget->setURL(chat_url.toStdString());
 	}
+#else
+	UNUSED_PARAMETER(chat_id);
 #endif
 }
 
--- a/deps/media-playback/media-playback/decode.c
+++ b/deps/media-playback/media-playback/decode.c
@@ -97,6 +97,8 @@ static int mp_open_codec(struct mp_decode *d, bool hw)
 #ifdef USE_NEW_HARDWARE_CODEC_METHOD
 	if (hw)
 		init_hw_decoder(d, c);
+#else
+	UNUSED_PARAMETER(hw);
 #endif
 
 	if (c->thread_count == 1 && c->codec_id != AV_CODEC_ID_PNG &&
--- a/deps/obs-scripting/obs-scripting-python.c
+++ b/deps/obs-scripting/obs-scripting-python.c
@@ -1644,9 +1644,11 @@ bool obs_scripting_load_python(const char *python_path)
 	dstr_free(&old_path);
 #endif
 
+#if PY_VERSION_HEX < 0x03070000
 	PyEval_InitThreads();
 	if (!PyEval_ThreadsInitialized())
 		return false;
+#endif
 
 	/* ---------------------------------------------- */
 	/* Must set arguments for guis to work            */
--- a/deps/obs-scripting/obspython/obspython.i
+++ b/deps/obs-scripting/obspython/obspython.i
@@ -30,6 +30,15 @@
 #include "obs-frontend-api.h"
 #endif
 
+/* Redefine SWIG_PYTHON_INITIALIZE_THREADS if:
+ * - Python version is 3.7 or later because PyEval_InitThreads() became deprecated and unnecessary
+ * - SWIG version is not 4.1 or later because SWIG_PYTHON_INITIALIZE_THREADS will be define correctly
+ *   with Python 3.7 and later */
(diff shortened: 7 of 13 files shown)


plugins/obs-ffmpeg/obs-ffmpeg-vaapi.c
```diff
@@ -593,8 +593,12 @@ static obs_properties_t *vaapi_properties(void *unused)
 			    strcmp(file_name, "..") == 0)
 				continue;
 
-			char path[64] = "\0";
-			sprintf(path, "/dev/dri/by-path/%s", file_name);
+			char path[64];
+			char path_suffix[47];
+
+			strncpy(path_suffix, file_name, 47);
```
kkartaltepe, line 599: I think useless copies make these functions harder to understand. Google suggests branching on thbe result of snprintf will calm gcc, if that works then lets blog that you have a malformed entry and continue to skip this.

plugins/linux-v4l2/v4l2-output.c
```diff
@@ -191,8 +191,10 @@ static bool virtualcam_start(void *data)
 
 	for (int i = 0; i < n; i++) {
 		char device[32];
+		char device_suffix[27];
 
-		snprintf(device, 32, "/dev/%s", list[i]->d_name);
+		strncpy(device_suffix, list[i]->d_name, 27);
+		snprintf(device, 32, "/dev/%s", device_suffix);
```
kkartaltepe, line 197: Same as the other comment, lets log that you have a malformed entry and continue the loop.

plugins/obs-ffmpeg/obs-ffmpeg-vaapi.c
tytan652: I switch to `snprintf` and "used" the result to prevent the overflow warning, since the `file_name` prevent to go beyond a length of 64.

plugins/linux-v4l2/v4l2-output.c
tytan652: Same, I think device name will never go beyond a length of 32.

tytan652:
I'm going to make a test build on Ubuntu 20.04 to see if format truncation still appear or other error.

Once this is done I will squash some commits together.

plugins/aja/aja-ui-props.hpp
```diff
@@ -2,8 +2,8 @@
 
 #include <obs-module.h>
 
-static const char *kProgramOutputID = "aja_output";
-static const char *kPreviewOutputID = "aja_preview_output";
+#define AJA_OUTPUT_ID "aja_output"
+#define AJA_PREVIEW_OUTPUT_ID "aja_preview_output"
```
tytan652, line 6: Should I let them as macros, or should I turn them back to `static const char*` and add `[[maybe_unused]]` attribute to them ?

tytan652: I decided to use [[maybe_unused]].

tytan652:
I dropped AJA related changes since f09137a2e00aecbce634d3e46a8a809ca8cd6907 introduces new warnings some that I can't fix all of them.

plugins/obs-ffmpeg/obs-ffmpeg-av1.c
```diff
@@ -408,7 +408,8 @@ static bool av1_encode(void *data, struct encoder_frame *frame,
 		debug("cur: %lld, packet: %lld, diff: %lld", cur_ts,
 		      recv_ts_nsec, cur_ts - recv_ts_nsec);
 #endif
-		if (llabs(cur_ts - recv_ts_nsec) > TIMEOUT_MAX_NSEC) {
+
+		if (llabs(cur_ts - (int64_t)recv_ts_nsec) > TIMEOUT_MAX_NSEC) {
```
tytan652, line 412: This is the change that I have no idea if it's an good idea to do that or just let the warning.

tytan652: Will be removed on the next push.

plugins/obs-ffmpeg/obs-ffmpeg-av1.c
```diff
@@ -417,6 +417,7 @@ static bool av1_encode(void *data, struct encoder_frame *frame,
 		debug("cur: %lld, packet: %lld, diff: %lld", cur_ts,
 		      recv_ts_nsec, cur_ts - recv_ts_nsec);
 #endif
+
```
RytoEX, line 420: Remove this change.

plugins/obs-ffmpeg/obs-ffmpeg-vaapi.c
```diff
@@ -593,8 +593,16 @@ static obs_properties_t *vaapi_properties(void *unused)
 			    strcmp(file_name, "..") == 0)
 				continue;
 
-			char path[64] = "\0";
-			sprintf(path, "/dev/dri/by-path/%s", file_name);
+			char path[64];
+
+			// Use the return value of snprintf to prevent truncation warning
```
RytoEX, line 598: End comment with punctuation.

plugins/linux-v4l2/v4l2-output.c
```diff
@@ -199,7 +199,12 @@ static bool virtualcam_start(void *data)
 	for (int i = 0; i < n; i++) {
 		char device[32];
 
-		snprintf(device, 32, "/dev/%s", list[i]->d_name);
+		// Use the return value of snprintf to prevent truncation warning
+		int written = snprintf(device, 32, "/dev/%s", list[i]->d_name);
+		if (written > 32)
+			blog(LOG_DEBUG,
+			     "v4l2-output: A format truncation may have happened,"
+			     " this can be ignored since it is quite improbable");
```
RytoEX, line 207: Use:
"A format truncation may have occurred."
" This can be ignored since it is quite improbable."

Use this here and in the other location in obs-ffmpeg-vaapi.c.

plugins/linux-v4l2/v4l2-output.c
```diff
@@ -199,7 +199,12 @@ static bool virtualcam_start(void *data)
 	for (int i = 0; i < n; i++) {
 		char device[32];
 
-		snprintf(device, 32, "/dev/%s", list[i]->d_name);
+		// Use the return value of snprintf to prevent truncation warning
+		int written = snprintf(device, 32, "/dev/%s", list[i]->d_name);
+		if (written > 32)
```
kkartaltepe, line 204: snprintf will never return a value greater than the limit passed in. Which reveals a minor bug. 

`device` should be a 33 character long array in order to include the null terminal (or we should limit snprintf to 31 characters). `device` should also be initialized to ensure that the unwritten bytes are 0 (`char device[32 ] = {0}` or an equivalent memset).

This check should then be >= 32 (or 31 if you choose that).

kkartaltepe (approved):

plugins/linux-v4l2/v4l2-output.c
tytan652: char array for VAAPI and V4L2 are now initialized with `{0}` and `>=32` is used.

plugins/linux-v4l2/v4l2-output.c
tytan652: Thank you, changes done.

plugins/obs-ffmpeg/obs-ffmpeg-vaapi.c
tytan652: Removed.

plugins/obs-ffmpeg/obs-ffmpeg-av1.c
tytan652: Removed.

PR #5661, opened 2021-12-16
UI: Fix a stack overlow caused by using OBSScene

walker-WSH:
### Description

Fix a stack overflow caused by using OBSScene in OBSBasicTransform::OBSSceneItemRemoved, which is callback of signal.
OBSSceneItemRemoved is called as this callstack (dead cycle):

### Motivation and Context

I think it is really quite dangerous to use OBSSource/OBSScene/OBSSceneItem in callback of signal, especilly using these variable type to params of invokeMethod.

Because in callback of signal, we can't ensure source's refs. If refs has been -1, that means any thread is being destroying it. 
If we use OBSSource for param, after running function passed to invokeMehtod, refs will be -1 again. Then obs_source_destroy will be called for this deleted object, crash happen!

I have meet this kind of crash many times. Such as https://github.com/obsproject/obs-studio/pull/5605
@jp9000  It's perfect if this can be considered in "destroy defer system".

### How Has This Been Tested?

open window of OBSBasicTransform, exit OBS or delete current scene.
I can't reproduce this callstack because sometimes signal won't be invoked.

### Checklist:

### DIFF
--- a/UI/window-basic-transform.cpp
+++ b/UI/window-basic-transform.cpp
@@ -23,7 +23,7 @@ static bool find_sel(obs_scene_t *, obs_sceneitem_t *item, void *param)
 	return true;
 };
 
-static OBSSceneItem FindASelectedItem(OBSScene scene)
+static OBSSceneItem FindASelectedItem(obs_scene_t *scene)
 {
 	OBSSceneItem item;
 	obs_scene_enum_items(scene, find_sel, &item);
@@ -184,8 +184,8 @@ void OBSBasicTransform::OBSSceneItemRemoved(void *param, calldata_t *data)
 {
 	OBSBasicTransform *window =
 		reinterpret_cast<OBSBasicTransform *>(param);
-	OBSScene scene = (obs_scene_t *)calldata_ptr(data, "scene");
-	OBSSceneItem item = (obs_sceneitem_t *)calldata_ptr(data, "item");
+	obs_scene_t *scene = (obs_scene_t *)calldata_ptr(data, "scene");
+	obs_sceneitem_t *item = (obs_sceneitem_t *)calldata_ptr(data, "item");
 
 	if (item == window->item)
 		window->SetItem(FindASelectedItem(scene));
@@ -205,8 +205,8 @@ void OBSBasicTransform::OBSSceneItemDeselect(void *param, calldata_t *data)
 {
 	OBSBasicTransform *window =
 		reinterpret_cast<OBSBasicTransform *>(param);
-	OBSScene scene = (obs_scene_t *)calldata_ptr(data, "scene");
-	OBSSceneItem item = (obs_sceneitem_t *)calldata_ptr(data, "item");
+	obs_scene_t *scene = (obs_scene_t *)calldata_ptr(data, "scene");
+	obs_sceneitem_t *item = (obs_sceneitem_t *)calldata_ptr(data, "item");
 
 	if (item == window->item) {
 		window->setWindowTitle(


tt2468:
I understand that this is possible, but the rule of thumb is for anything using calldata is to hold a reference to the resource, use calldata, then release the resource. In those cases, it would not matter if we use reference guards or not.

I don't think I understand what the underlying issue is, because if some component of OBS is going to use a resource (scene, source) in a delayed manner, it should be using weak refs.

walker-WSH:
> calldata

While pushing signal, if refs of source/scene storaged into calldata has been increased before calling callback, there won't exist this issue.
However, I can't find _obs_source_addref_ invoked before pushing signal.

walker-WSH:
> I understand that this is possible, but the rule of thumb is for anything using calldata is to hold a reference to the resource, use calldata, then release the resource. In those cases, it would not matter if we use reference guards or not.
> 
> I don't think I understand what the underlying issue is, because if some component of OBS is going to use a resource (scene, source) in a delayed manner, it should be using weak refs.

While pushing signal, if refs of source/scene storaged into calldata has been increased before calling callback, it can reduce this issue. However, I can't find obs_source_addref invoked before pushing signal.

I said "reduce" instead of "fixed", that is because we can't ensure refs is valid while calling addref  (refs == -1).
So I add "return bool" for add_ref in PR#5605

jp9000:
Oh, I was wondering what was causing weird scene references sometimes.

tt2468:
> However, I can't find obs_source_addref invoked before pushing signal.

In the example shown, the calling thread would already own a reference. In the case of `obs_sceneitem_remove()` (a place where `signal_item_remove()` is called from), it is assumed that the caller has already incremented the reference. So any signal callbacks are automatically guarded from this issue.

---

Related to this case:
- Thread A dec to `-1`, starts destroy
- Thread B inc to `0`
- Thread B dec to `-1`, starts second destroy

This issue would theoretically apply to other reference counted systems like shared_ptr. How do they currently solve/avoid this issue?

A few cases that could theoretically lead to that scenario are:
- `obs_get_source_by_name()`
- `obs_enum_sources()`
- `obs_weak_source_get_source()`

jp9000:
It pretty much comes down to addref being used in cases where ownership is not known/guaranteed, because that function is only supposed to be used when ownership is known/guaranteed. When ownership is not known/guaranteed, the obs_source_get_ref() function is supposed to be used instead, but I don't think it's just a problem of using the wrong function for getting a reference.

This discussion has been interesting, because these cases are precisely what I'm discovering with my destroy defer code. I'm finding that this PR isn't the only case where this occurs. Another such case I found is OBSBasic::SourceRemoved -- albeit that race was so remote that I only discovered it in my destruction defer code. I'll be committing more fixes to this as I discover them.

jp9000:
For this particular signal, walker is correct that this should not take a reference here. It's clear it should be used without taking a reference. This is an annoying "gotcha", because we can't add a reference, but we sort of need to have access to the scene pointer. The scene pointer in this function is safe to use for this particular function, but it's definitely an annoying gotcha. I'm not entirely sure if this fix is avoidable or if there's a better alternative for now, just due to the fact that it's called while a scene is in the process of being destroyed.

walker-WSH:
> shared_ptr

shared_ptr won't include the case about "0 -> -1 ->0". 
Because after storaging "A* xxx" to shared_ptr, all access to ${xxx} is using shared_ptr<A> and it is safe to increase or decrease refs.

However in OBS, we often try to get OBSSource from "obs_source*", and never check current value of refs.

walker-WSH:
> For this particular signal, walker is correct that this should not take a reference here. It's clear it should be used without taking a reference. This is an annoying "gotcha", because we can't add a reference, but we sort of need to have access to the scene pointer. The scene pointer in this function is safe to use for this particular function, but it's definitely an annoying gotcha. I'm not entirely sure if this fix is avoidable or if there's a better alternative for now, just due to the fact that it's called while a scene is in the process of being destroyed.

If only used in callback of signal, the object is surely alive and never need to use OBSSource.


> > However, I can't find obs_source_addref invoked before pushing signal.
> 
> In the example shown, the calling thread would already own a reference. In the case of `obs_sceneitem_remove()` (a place where `signal_item_remove()` is called from), it is assumed that the caller has already incremented the reference. So any signal callbacks are automatically guarded from this issue.
> 
> Related to this case:
> 
> * Thread A dec to `-1`, starts destroy
> * Thread B inc to `0`
> * Thread B dec to `-1`, starts second destroy
> 
> This issue would theoretically apply to other reference counted systems like shared_ptr. How do they currently solve/avoid this issue?
> 
> A few cases that could theoretically lead to that scenario are:
> 
> * `obs_get_source_by_name()`
> * `obs_enum_sources()`
> * `obs_weak_source_get_source()`


Oh~ Please know that stack overflow is caused by scene, not sceneitem instead.
The callstack begins from scene_destroy() where refs of scene has been -1.  
Then push signal "item_remove" in remove_all_item for scene. 
In callback of signal, OBSScene cause refs from -1 to 0, then o to -1, so scene_destroy() is called again and again~~

PR #5668, opened 2021-12-18
UI: Post events in scene collection cleanup to properly destroy removed sources

### LINKED ISSUE #2715: When switching scene collections, the new scene collection often loads before the old one has fully unloaded, causing a variety of issues.
## Platform

Operating system and version:  Windows 10 x64 (1909)
OBS Studio version:  25.0.4 (and all other versions for the last 2 years or so)

## Expected Behavior

When switching from one scene collection to another while streaming live and/or recording, any video capture devices which are used in both collections should continue to work when OBS
switches from one collection to another.

## Current Behavior

When starting in scene collection 1, the webcam works as expected.  After switching to another scene collection (any) which use the same webcam, the webcam will not work by default.  It can sometimes be coaxed into working by bring up the video capture properties for the webcam and disconnecting and reconnecting it, however sometimes this doesn't work and restarting OBS is required.

Please note that with a simple OBS configuration with just a scene with a webcam the problem probably wont happen.  It becomes increasingly more likely to occur the more OBS scenes and sources that there are, which suggests some kind of race condition.

https://obsproject.com/logs/HIwvfa0gw0_Qnr5k

## Steps to Reproduce

1.  Create a new OBS scene collection "Cam test 1"
2.  In the default scene, add 2 or 3 video capture devices of different webcams.  I am using 2 C920 cameras here in my reproduction testing.
3.  Now add a media source and choose a MP4 video to play in a loop, and enable hardware decoding.  (This isn't necessarily relevant to the problem but I had to add some scene items and this seemed to be just enough to trigger the problem)
4.  Resize both webcam sources and the media source so you can see all 3 of them successfully in the preview window.
5. Duplicate the scene collection as "Cam test 2".
6. Switch back and forth between "Cam test 1" and "Cam test 2", and observe how one or both webcams no longer work in one of the collections, but may still work in the other collection.
7. If the bug does not show itself, delete "Cam test 2" scene collection and switch back to the "Cam test 1" collection.  Now add an additional Media source of another MP4 file and resize it so all sources are simultaneously visible.
8. Go to step 5 again.  Repeat steps 5-8 a few times if necessary and the bug should show itself by causing one or both webcams to no longer work in one of the scene collections.

## Additional information

This appears to be a bug in which OBS is loading the new scene collection before it has been able to completely unload the first collection leading to a race condition of some sort in which the webcam device is still opened from the first collection causing trying to open it to fail when the second one is loaded.

Matt WizardCM suggested as a troubleshooting workaround to create a new scene collection named "Blank" which contains nothing, and switch to it first, then switch to the desired collection each time.  I tried this and it does seem to work around the race condition and allow both webcams to work.

In summary, switching:
From Cam test 1 to Cam test 2 -> results in one or both webcams failing to start up.
From Cam test 1 to Blank to Cam test 2 -> results in both cams working properly as it adds additional time for OBS to complete unloading the first collection and disconnect the webcam before trying to load it in the new collection.

tt2468:
### Description
- Adds the `obs_source_get_refs()` function (used for logging purposes)
- Adds logging to ClearSceneData() to log lingering sources
- Waits one frame and clears the Qt event queue to ensure all sources have been destroyed

Fixes #2715 

### Motivation and Context
- Logging lingering sources can be extremely helpful while debugging memory leaks.
- Data loss is generally regarded as a bad thing.

### How Has This Been Tested?
- Added logging before posting events
- Opened OBS and switched from one large scene collection to another
- Before waiting, sources get logged
- After posting events, no sources remain

### Types of changes

### Checklist:

### DIFF
--- a/UI/window-basic-main.cpp
+++ b/UI/window-basic-main.cpp
@@ -4563,6 +4563,13 @@ void OBSBasic::ClearSceneData()
 
 	undo_s.clear();
 
+	// Wait one frame for the graphics thread to remove any remaining references
+	obs_queue_task(
+		OBS_TASK_GRAPHICS, [](void *) {}, nullptr, true);
+
+	// Post all UI events to purge any remaining references
+	QApplication::sendPostedEvents(this);
+
 	auto leakCb = [](void *, obs_source_t *source) {
 		blog(LOG_WARNING,
 		     "Source '%s' not destroyed after remove! Leaking refs: %ld",
--- a/UI/window-basic-main.cpp
+++ b/UI/window-basic-main.cpp
@@ -4563,6 +4563,16 @@ void OBSBasic::ClearSceneData()
 
 	undo_s.clear();
 
+	auto leakCb = [](void *, obs_source_t *source) {
+		blog(LOG_WARNING,
+		     "Source '%s' not destroyed after remove! Leaking refs: %ld",
+		     obs_source_get_name(source),
+		     obs_source_get_refs(source) + 1);
+		return true;
+	};
+
+	obs_enum_all_sources(leakCb, nullptr);
+
 	disableSaving--;
 
 	blog(LOG_INFO, "All scene data cleared");
--- a/libobs/obs-source.c
+++ b/libobs/obs-source.c
@@ -712,6 +712,14 @@ void obs_source_addref(obs_source_t *source)
 	obs_ref_addref(&source->control->ref);
 }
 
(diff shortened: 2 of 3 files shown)


tt2468:
@skeletonbow would you be able to test this and confirm that it does fix the issue?

jp9000:
Should be unnecessary now with https://github.com/obsproject/obs-studio/commit/8b3416c1e70bd4dad6baf1d424357a6af540efe0

PR #5669, opened 2021-12-18
libobs: Hold source ref during `source_remove` signal

tt2468:
### Description
Holds an active reference to a source during signaling of the `source_remove` signal, to prevent receivers from being given an already-destroyed source.

### Motivation and Context
All receivers of the `source_remove` signal should receive a valid source to operate on.

- Call obs_source_remove(source)
- Receiver 1 gets signal, calls `obs_source_release(source)`
- Receiver 2 gets signal, calls `obs_source_release(source)`, refs == -1, source destroyed
- Receiver 3 gets signal, source already destroyed, is forced to ignore signal due to invalid source

### How Has This Been Tested?
This is a theoretical situation which is currently possible in obs-websocket. While I can believe I've seen crashes related to this, I cannot prove it.

### Types of changes

### Checklist:

### DIFF
--- a/libobs/obs-source.c
+++ b/libobs/obs-source.c
@@ -793,8 +793,12 @@ void obs_source_remove(obs_source_t *source)
 		return;
 
 	if (!source->removed) {
-		source->removed = true;
-		obs_source_dosignal(source, "source_remove", "remove");
+		obs_source_t *s = obs_source_get_ref(source);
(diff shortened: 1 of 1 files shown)


jp9000:
I changed it to use obs_source_get_ref instead. We should try to avoid using addref in the future.

Issue #5463, opened 2021-10-29
Image files with uppercase files extensions


TomHenderson4:
### Operating System Info

Ubuntu 18.04

### OBS Studio Version

27.1.3

### OBS Studio Log URL

https://obsproject.com/logs/m5_giJr89g0Z4TT_

### Expected Behavior

Files with upper case file extensions are still valid files. For example picture.jpg and Picture.JPG are both completely valid JPEG files and should be treated as such with any software that uses image files.

This is a defect in what is otherwise an amazing software package.

Currently I work with a collection of tens of thousands of image files, many of which are produced by drones which store them with upper case extensions. Renaming these files would be onerous and would trigger a massive backup due tot he file name changes.

Please consider making this change to correct this defect in OBS Studio. 

### Current Behavior

When attempting to select an image file, none of the valid image files with uppercase extensions appear in the file selection dialog, and there appears to be no workaround.

As a result, OBS Studio does not support valid image files. 

### Steps to Reproduce

1. Click the plus sign to add a new source
2. Select "Image"
3. Browse for image files
4. All valid image files with upper case file extensions are excluded.

fuscati:
Can't reproduce. Image with upper case extension appears as expected.

TomHenderson4:
What I see when I look in the folder...
![image](https://user-images.githubusercontent.com/29453298/139508552-b57c3018-8c2e-4eb3-86c1-2d305cb72b12.png)

What I see when I open that folder with OBS Studio using the "Image"
![image](https://user-images.githubusercontent.com/29453298/139508587-e25143a2-76c4-45aa-bcb1-89d5ea9bd96c.png)

Another directory, note the highlighted files with upper case extensions:
![image](https://user-images.githubusercontent.com/29453298/139508623-2165cfd1-4908-4bb4-9d1c-6b284d46cad1.png)

Same directory when I try to open images with OBS Studio. Note the highlighted files are not available on this list because the extension is upper case:
![image](https://user-images.githubusercontent.com/29453298/139508646-d5fd36a5-301a-437b-b2fd-f61cffbc14bc.png)

The URL of the logs from this session:

https://obsproject.com/logs/jQY48iUAHf_QWqXS

tytan652:
This issue is related to the native dialog that is not case insensitive on file extensions.

This is not an issue with OBS Studio but with how the native dialog manage extensions.

MatthewFallon:
[...]

PR #5672, opened 2021-12-19
plugins: Allow adding image files with uppercase extension

Sayan505:
### Description
While adding an Image source to a scene, allow selection of
supported image files having uppercase extensions.

Also changed the label `All formats` to `All Supported Formats` to convey the functionality correctly

#### After:

 <br>

### Motivation and Context
Valid image files having upercase extensions (JPEG, PNG, GIF etc) did not show up in the file browser when an Image source was being added to a Scene.
I modified the `image_filter` string to allow showing of files with such extensions.

This PR fixes the issue #5463

 <br>

### How Has This Been Tested?
Fedora 35 (kernel: 5.15.6-200.fc35.x86_64), AMD Ryzen 5 3500.
* Code compiled and run correctly.
* Got desired output.

 <br>

### Types of changes

 <br>

### Checklist:

### DIFF
--- a/plugins/image-source/image-source.c
+++ b/plugins/image-source/image-source.c
@@ -248,14 +248,15 @@ static void image_source_tick(void *data, float seconds)
 }
 
 static const char *image_filter =
-	"All formats (*.bmp *.tga *.png *.jpeg *.jpg *.gif *.psd *.webp);;"
-	"BMP Files (*.bmp);;"
-	"Targa Files (*.tga);;"
-	"PNG Files (*.png);;"
-	"JPEG Files (*.jpeg *.jpg);;"
-	"GIF Files (*.gif);;"
-	"PSD Files (*.psd);;"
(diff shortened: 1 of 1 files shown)


tytan652:
This issue is related to the native dialog that is not case insensitive on file extensions.

This is not an issue with OBS Studio but with how the native dialog manage extensions.

So this PR is not fixing the issue where it really is, it just a workaround for maybe only one type of native dialog.

gxalpha:
I agree with tytan here, it's not a great idea to hardcoding uppercase extensions everywhere (there are other instances that would need to be changed as well, like the ffmpeg source).
A better workaround might be to use a QFileDialog instead of the native one, something along these lines (untested):
[qfiledialog.diff.txt](https://github.com/obsproject/obs-studio/files/7740745/qfiledialog.diff.txt)
Though I'm personally not sure it's worth the tradeoff, since I don't know how severe the original issue is in practice.

PR #5678, opened 2021-12-20
UI: Remove unneeded call when removing scene

gxalpha:
### Description

Cleans up an unused statement which introduced an unneeded variable.

### Motivation and Context

Less code in window-basic-main is always a good thing :p
No seriously, I just stumbled across it while working on something else.

### How Has This Been Tested?

Still compiles (macOS 12.0.1).

### Types of changes

### Checklist:

### DIFF
--- a/UI/window-basic-main.cpp
+++ b/UI/window-basic-main.cpp
@@ -3798,8 +3798,6 @@ void OBSBasic::RemoveSelectedScene()
 	OBSScene scene = GetCurrentScene();
 	obs_source_t *source = obs_scene_get_source(scene);
 
-	OBSSource curProgramScene = OBSGetStrongRef(programScene);
-
 	if (!source || !QueryRemoveSource(source)) {
 		return;
 	}


notr1ch:
While the variable may be unused, its presence in the function until the destructor is called holds a reference to the program scene. Are you sure this wasn't intended for example to prevent accidental destruction of the program scene during RemoveSelectedScene somewhere?

gxalpha:
The variable was originally introduced with the undo/redo system, which at the time still held references.
This has since been rewritten, and from what I can tell there is no need to hold that reference.
See also: https://github.com/obsproject/obs-studio/commit/60d95cb#diff-4c0a52244f264feca9c3031c90edcbbbd1087a3040c15f9df4ca9700c5154fa3L3631-R3777

(I also did some more tests, deleting scenes with studio mode enabled and disabled, and didn't get scenes/sources to accidentally get destroyed)

notr1ch:
Sounds good, just wanted to make sure it was something that had been considered. Thanks!

PR #5680, opened 2021-12-21
mac-capture: Improve window capture performance

dossy:
### Description

Reduce CPU utilization on macOS when using Window Capture.

### Motivation and Context

The current implementation of Window Capture on macOS uses Core Graphics to render a bitmap raster which is then handed off to OBS.

This PR replaces that with an implementation that accesses the window's CGImage bitmap directly and blits it and hands that off as a frame to OBS, instead.

### How Has This Been Tested?

Tested on a 2020 iMac 27" w/ 3.6 GHz 10-Core Intel Core i9, AMD Radeon Pro 5700 XT 16 GB, macOS Catalina 10.15.7.

Tested with OBS 26.1.2 and Riot Games League of Legends, with an empty scene with only the Window Capture input source, OBS was consuming 55%-70% CPU.

Tested with the changes in this PR, OBS only consumed 30-35% CPU.

### Types of changes

#### Summary of changes

* I removed any code that was solely used by the previous `CGBitmap`-based implementation, as it's no longer needed.
* I added a small quality-of-life improvement in `capture_thread()` that names the thread using the source name to make debugging easier when you have multiple instances of the input source in a scene.
* I added an `assert()` in `window_capture_create_internal()` to ensure that [the Cocoa Framework is properly multithreaded](https://developer.apple.com/library/archive/documentation/Cocoa/Conceptual/Multithreading/CreatingThreads/CreatingThreads.html#//apple_ref/doc/uid/10000057i-CH15-SW21).
* Renamed the locally scoped variable `cap` to `wc` in `window_capture_destroy()` to be consistent with the rest of the file's name for the `window_capture` structure ptr.
* Reordered the setting of property defaults in `window_capture_defaults()` so that anything defined in `mac-window-capture` would override what was defined by `window_defaults()`.
* Renamed `next_search_time` to `last_search_time` to improve the semantics and, I suppose, indirectly fix a bug that resulted from `next_search_time` never being initialized, resulting in `cw->next_search_time > os_gettime_ns()` always evaluating to false, which effectively made `find_window()` a no-op.
* After fixing the issue with `find_window()`, I had to fix the implementation to properly find the correct window, including applications whose window names were `NULL` (such as the League of Legends client).
* I removed the unnecessary `pthread_{lock, unlock}`ing in `init_window()` as that is called from the main thread, before the `capture_thread()` is started, so the locks are unnecessary there.
* Eliminated some NS-value local variable ptrs to make the code easier to read.
* Changed `destroy_window()` to destroy the mutex after the retained vars are released, not before.  In this particular case the order of operations doesn't really matter, but this is just good hygiene.
* Wrapped the `pthread_{lock, unlock}` and other related calls with `assert()` as basic error checking.  Presumably if any of those syscalls actually fail, your process is likely in a state that is unrecoverable and should be terminated, anyway.
* Changed the implementation of `make_name()` because I thought ARC was `free()`ing the buffer before `obs_property_list_add_int()` could `bstrdup()` it, because I was seeing a hard-to-explain crash.  Turns out, that crash was likely fixed by [2d524580](https://github.com/obsproject/obs-studio/commit/2d524580fe7251336169efc46466277a6af8aeff), so I may be able to revert the change relating to the change of this function's implementation.
* Inlined the contents of `find_window_dict()` into its only caller, as it didn't really aid in comprehension of the code having it in its own function, and wasn't called from anywhere else, anyway.
* Fixed the implementation in `window_changed_internal()` to repopulate the window list combobox while matching the currently selected window and disabling it.

I think this is a fairly complete summary of the changes in this PR.  Please let me know if you feel I missed anything, or would like me to elaborate further on any of these.

### Checklist:

### DIFF
--- a/plugins/mac-capture/mac-window-capture.m
+++ b/plugins/mac-capture/mac-window-capture.m
@@ -13,13 +13,7 @@ struct window_capture {
 
 	struct cocoa_window window;
 
-	//CGRect              bounds;
-	//CGWindowListOption  window_option;
 	CGWindowImageOption image_option;
-
-	CGColorSpaceRef color_space;
-
-	DARRAY(uint8_t) buffer;
 
 	pthread_t capture_thread;
 	os_event_t *capture_event;
@@ -28,55 +22,62 @@ struct window_capture {
 
 static CGImageRef get_image(struct window_capture *wc)
 {
-	NSArray *arr = (NSArray *)CGWindowListCreate(
-		kCGWindowListOptionIncludingWindow, wc->window.window_id);
-	[arr autorelease];
+	CFMutableArrayRef arr =
+		CFArrayCreateMutable(kCFAllocatorDefault, 0, NULL);
+	CFArrayAppendValue(arr, (void *)(uintptr_t)wc->window.window_id);
 
-	if (!arr.count && !find_window(&wc->window, NULL, false))
-		return NULL;
+	CGImageRef image = CGWindowListCreateImageFromArray(
+		CGRectNull, (CFArrayRef)arr, wc->image_option);
+	CFRelease(arr);
 
-	return CGWindowListCreateImage(CGRectNull,
-				       kCGWindowListOptionIncludingWindow,
-				       wc->window.window_id, wc->image_option);
+	if (!image && find_window(&wc->window, NULL))
+		image = get_image(wc);
+
+	return image;
 }
 
 static inline void capture_frame(struct window_capture *wc)
 {
 	uint64_t ts = os_gettime_ns();
-	CGImageRef img = get_image(wc);
-	if (!img)
+	CGImageRef image = get_image(wc);
+
+	if (!image)
 		return;
 
-	size_t width = CGImageGetWidth(img);
-	size_t height = CGImageGetHeight(img);
+	size_t width = CGImageGetWidth(image);
+	size_t height = CGImageGetHeight(image);
+	size_t bytes_per_row = CGImageGetBytesPerRow(image);
 
-	CGRect rect = {{0, 0}, {width, height}};
-	da_resize(wc->buffer, width * height * 4);
-	uint8_t *data = wc->buffer.array;
+	if ((!width && !height) || CGImageGetBitsPerPixel(image) != 32 ||
+	    CGImageGetBitsPerComponent(image) != 8) {
+		CGImageRelease(image);
+		return;
+	}
 
-	CGContextRef cg_context = CGBitmapContextCreate(
-		data, width, height, 8, width * 4, wc->color_space,
-		kCGBitmapByteOrder32Host | kCGImageAlphaPremultipliedFirst);
-	CGContextSetBlendMode(cg_context, kCGBlendModeCopy);
-	CGContextDrawImage(cg_context, rect, img);
-	CGContextRelease(cg_context);
-	CGImageRelease(img);
+	CGDataProviderRef provider = CGImageGetDataProvider(image);
+	CFDataRef data = CGDataProviderCopyData(provider);
+	const uint8_t *buffer = CFDataGetBytePtr(data);
 
 	struct obs_source_frame frame = {
 		.format = VIDEO_FORMAT_BGRA,
 		.width = width,
 		.height = height,
-		.data[0] = data,
-		.linesize[0] = width * 4,
+		.data[0] = (uint8_t *)buffer,
+		.linesize[0] = bytes_per_row,
 		.timestamp = ts,
 	};
 
 	obs_source_output_video(wc->source, &frame);
+
+	CFRelease(data);
+	CGImageRelease(image);
 }
 
 static void *capture_thread(void *data)
 {
 	struct window_capture *wc = data;
+
+	os_set_thread_name(obs_source_get_name(wc->source));
 
 	for (;;) {
 		os_event_wait(wc->capture_event);
@@ -98,10 +99,6 @@ static inline void *window_capture_create_internal(obs_data_t *settings,
 
 	wc->source = source;
 
-	wc->color_space = CGColorSpaceCreateDeviceRGB();
-
-	da_init(wc->buffer);
-
 	init_window(&wc->window, settings);
 
 	wc->image_option = obs_data_get_bool(settings, "show_shadow")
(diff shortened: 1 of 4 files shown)


dossy:
This is a resubmission of #4863, which was mistakenly merged and therefore requires a new PR as the previous one cannot be reopened.

plugins/mac-capture/mac-window-capture.m
```diff
@@ -13,71 +13,75 @@
...
 
-	CGContextRef cg_context = CGBitmapContextCreate(
-		data, width, height, 8, width * 4, wc->color_space,
-		kCGBitmapByteOrder32Host | kCGImageAlphaPremultipliedFirst);
-	CGContextSetBlendMode(cg_context, kCGBlendModeCopy);
-	CGContextDrawImage(cg_context, rect, img);
-	CGContextRelease(cg_context);
-	CGImageRelease(img);
+	// float fps = 1e9 / (ts - wc->frame.timestamp);
+	// NSLog(@"FPS %.4f", fps);
```
RytoEX, line 58: Is this debug code?

plugins/mac-capture/window-utils.m
```diff
@@ -37,186 +36,205 @@
...
 
-	pthread_mutex_lock(&cw->name_lock);
+	if (pthread_mutex_lock(&cw->mutex))
+		abort();
 
 	if (!cw->window_name.length && !cw->owner_name.length)
-		goto invalid_name;
+		goto unlock;
 
 	for (NSDictionary *dict in enumerate_windows()) {
```
RytoEX, line 54: See previous PR's unresolved thread about this line.  Have all concerns been addressed?

@MakersF

https://github.com/obsproject/obs-studio/pull/4863#discussion_r724591985

RytoEX (commented):
Porting some unresolved questions from #4863 to this PR.

RytoEX:
Additionally, please mark any resolved threads in #4863 as resolved if they have been addressed so that we don't need to wonder about them.  Thanks!

plugins/mac-capture/mac-window-capture.m
dossy: It is - I thought I'd removed it in the latest rebase/commit cycle, guess I missed it. Thanks!

dossy:
> Additionally, please mark any resolved threads in #4863 as resolved if they have been addressed so that we don't need to wonder about them. Thanks!

I went ahead and marked threads resolved that I felt should be, but in general it's my understanding that the person who opened the thread needs to decide if the subsequent discussion in the thread has reached the point where the thread has been resolved and can be marked as such?

RytoEX:
> > Additionally, please mark any resolved threads in #4863 as resolved if they have been addressed so that we don't need to wonder about them. Thanks!
> 
> I went ahead and marked threads resolved that I felt should be, but in general it's my understanding that the person who opened the thread needs to decide if the subsequent discussion in the thread has reached the point where the thread has been resolved and can be marked as such?

Technically, anyone can resolve or unresolve/reopen a thread.  We don't have a set process, and GitHub doesn't have a way to enforce one anyway.  If you, as the PR author, mark a thread as resolved because you believe you have addressed the concerns within, the commenter is free to unresolve/reopen the thread if they disagree.  In turn, we make the final decision if a concern has truly been addressed to our satisfaction.

Many of us who review the PRs get emails for every single comment anyway, so we see them regardless of if a review thread is resolved or unresolved.  I just don't want to be left wondering what was or wasn't resolved from the previous PR.

WizardCM (changes requested):
As this PR does a few separate things alongside improving performance, would it be possible to split it to multiple commits? If it's not, please put a little information into the commit description to summarise the bigger individual changes.

jp9000:
Once again terribly sorry for the accidental merge.

One thing I wanted to note about this PR, is it has too many changes in one commit -- someone else might have already mentioned this. We really need all individual changes separated into their own individual commits. Things such as changing the way finding windows works -- that needs its own PR.

Additionally, I think someone already stated this, but it should not capture other windows if the current window it's capturing disappears. It should try to find the same exact window only, and not fall back to other windows when possible. I know we do a similar thing on windows, but on windows we have a drop-down that specifies the fallback method -- I don't think window capture on mac has that, so by default it should only try to capture the window it was capturing before.

Another thing, you do not need to change pthread_mutex_lock and pthread_mutex_unlock to abort if they fail. I just think that's kind of a redundant thing for now. If we improve error handling later then I'd say that's fine, but that change feels unrelated to the PR and should be left out. We don't really have precedent for doing that anywhere else in the codebase at the moment, though we may do that later in some other form. But that is another example of an individual separate change that would have been in its own separate commit otherwise.

I know it might be a bit of a pain to separate these changes into individual commits, but it really does help ensure that we can spot issues and bugs easier to have as many separate (working) commits as possible. It also helps ease the review process.

dossy:
> Additionally, I think someone already stated this, but it should not capture other windows if the current window it's capturing disappears. It should try to find the same exact window only, and not fall back to other windows when possible. I know we do a similar thing on windows, but on windows we have a drop-down that specifies the fallback method -- I don't think window capture on mac has that, so by default it should only try to capture the window it was capturing before.

The code prior to this PR was already capturing other windows on Mac if the current window it's capturing disappears, just like on Windows.  On Windows, the drop-down specifies the fallback method, but on Mac there's only one fallback method, to match on window owner and name.  But, capturing a new window with the same name is the current behavior on Mac, which is the desirable behavior, and changing it will likely break a lot of Mac streamer setups that rely on this behavior.

dossy:
> Another thing, you do not need to change pthread_mutex_lock and pthread_mutex_unlock to abort if they fail.

I suppose if the system and/or process get into a state where those fail, then a buffer has probably been overrun somewhere or the system's resources are in a state where the process will probably be killed by the kernel or scheduler, anyway, so I suppose intentionally aborting isn't that necessary as the process is probably going to die shortly, anyway.

gxalpha:
@dossy what's the state on this one? Really hoping this can make it into the next release.

It would be great to see the commits split up as Jim said earlier, as that makes reviewing the PR much easier.

I'm personally also of the opinion that automatic window switching is a bad idea (even with the current behaviour), but I would leave that up to discussion (and it would be much easier to potentially remove if the commits are split up).

gxalpha:
Hey, we split off the part of this PR that improves the performance into another PR and merged it separately (see #6416).
Feel free to rebase and resolve conflicts here so that the other changes in this PR can be reviewed.

RytoEX:
Heads up, we've cherry-picked some of the changes out of this, so this now has merge conflicts. If there is still something you would like us to review here, please rebase this and resolve the merge conflicts.

RytoEX:
Per above comments, I'm going to close this. We've cherry-picked the parts from this that we felt were most impactful. There are still some changes in here that we are still willing to merge if cleaned up. If there's a desire to revisit the remaining parts of this PR, please feel free to reopen this or open a new PR.

PR #5681, opened 2021-12-21
libobs-d3d11: DuplicateOutput1 for DXGI capture

jpark37:
### Description
Despite DuplicateOutput deing documented to always return BGRA8, it is
seen fluctuating between BGRA8 and RGBA16F when HDR is enabled.
DuplicateOutput1 seems to provide a stable format whether SDR or HDR.

Also update the texture recreation logic to check for format changes.

### Motivation and Context
Want to make valid CopyResource calls.

### How Has This Been Tested?
Format doesn't fluctuate. HDR/SDR transitions are handled fine. HDR always displays image even though we don't tonemap for SDR yet.

### Types of changes

### Checklist:

### DIFF
--- a/libobs-d3d11/d3d11-duplicator.cpp
+++ b/libobs-d3d11/d3d11-duplicator.cpp
@@ -36,6 +36,7 @@ static inline bool get_monitor(gs_device_t *device, int monitor_idx,
 
 void gs_duplicator::Start()
 {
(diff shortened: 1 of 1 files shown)


WizardCM:
I assume this doesn't bump any minimum versions for Windows 10 builds?

jpark37:
The Windows 8 path is still used if QueryInterface fails.

PR #5682, opened 2021-12-22
libobs: Ensure all graphics tasks can be handled before exiting graphics thread

walker-WSH:
### Description

In graphics thread, both tick_sources() and output_frame() will invoke obs_source_release, which may cause source::destroy invoked in defer-thread or any thread.  For source::destroy of window/monitor plugin,  freeing WGC is posted into graphcis thread.

### **That is to say, tick_sources() or output_frame() can cause new graphics task be posted.**
So, It is better to check remaining tasks before stopping graphics thread. 

@jpark37   It should be reviewed by you.

### Motivation and Context

before stopping graphics thread, make sure stored WGC tasks can be handled

### How Has This Been Tested?

add window source and display source based on WGC mode;
exit OBS;
make sure all actual destroy can be handled from graphics thread

### Types of changes

### Checklist:

### DIFF
--- a/libobs/obs-video.c
+++ b/libobs/obs-video.c
@@ -964,8 +964,6 @@ bool obs_graphics_thread_loop(struct obs_graphics_context *context)
 		tick_sources(obs->video.video_time, context->last_time);
 	profile_end(tick_sources_name);
 
-	execute_graphics_tasks();
-
 #ifdef _WIN32
 	MSG msg;
 	while (PeekMessage(&msg, NULL, 0, 0, PM_REMOVE)) {
@@ -981,6 +979,8 @@ bool obs_graphics_thread_loop(struct obs_graphics_context *context)
 	profile_start(render_displays_name);
 	render_displays();
 	profile_end(render_displays_name);
+
+	execute_graphics_tasks();
 
 	frame_time_ns = os_gettime_ns() - frame_start;
 


tt2468:
@jp9000 Is this problem solved by the deferred destroy changes?

jpark37 (changes requested):
The message pump happens before tick, so we can draw changes on the same frame. Is there an actual problem here?

walker-WSH:
> The message pump happens before tick, so we can draw changes on the same frame. Is there an actual problem here?

@jpark37 
no issue, this is the reason of this PR:

jpark37:
We already run the graphics thread one extra loop on thread teardown:
```
	/* defer loop break to clean up sources */
	const bool stop_requested = video_output_stopped(obs->video.video);
```

The only way for this to be a problem is if you somehow modify a setting that requires graphics thread recreation, and delete a source at the same time. If we want to be paranoid, we can add another execute_graphics_tasks() call somewhere, but I don't think reordering the existing function calls is the solution we want.

jpark37:
Actually, do this order:

[message pump]
output_frame
render_displays
execute_graphics_tasks

@jp9000 Does that order look reasonable to you?

jp9000:
Does this PR take into account the new defer destroy code? It seems like in your example, you're referencing obs_source_destroy() occurring in the graphics thread, but it shouldn't be happening in the graphics thread anymore.

But yea, that seems fine to me @jpark37.

walker-WSH:
> Does this PR take into account the new defer destroy code? It seems like in your example, you're referencing obs_source_destroy() occurring in the graphics thread, but it shouldn't be happening in the graphics thread anymore.
> 
> But yea, that seems fine to me @jpark37.



This commit only ensure all graphics tasks can be handled before ending obs_graphics_thread. There is nothing with obs_source_destroy.

Even obs_source_actual_destory is invoked in defer-thread, posted graphics task from window::destroy and monitor::destroy should still be handled in graphic thread. 

If this commit is not easy to understand, I can chang it as Jpark37 said:

jpark37:
We changed our mind. Can you do this ordering instead?

[message pump]
output_frame
render_displays
execute_graphics_tasks

walker-WSH:
@jpark37 

code has been updated as your order

jpark37 (approved):

PR #5692, opened 2021-12-24
Automatic file splitting duplicate audio

norihiro:
### Description

When splitting file (#5371), audio packets are duplicated and saved to both previous and new files.
In the new file, audio timestamps will be offset by video timestamp so that video and audio has the same alignment among the split files in one recording.

Currently, the implementation relays on a video encoder that has DTS < PTS to duplicate audio packet. If DTS = PTS, the audio won't be duplicated. This behavior should be resolved before removing draft state.

### Motivation and Context

Audio glitch is the issue in the new WIP feature Automatic File Splitting (#5371).
Since the audio has different frame size than that for video, it is not easy for the boundary audio packet to be sent into just one file without audio glitch.

This change will resolve audio glitch when editing on NLEs. However, since the audio packets are duplicated, ffmpeg `concat` input might not be able to recover the original video. This function should be an optional.

### How Has This Been Tested?

OBS Studio is built with FFmpeg 4.4.1 on Fedora 34. Encoder setting is libx264.
Output video files are tested with Kdenlive 21.08.1.
Kdenlive sometimes treats the length of the file one frame longer, which results audio glitch. User has to manually shrink the video length to have smooth switch to the new file.

It would be welcome if someone can test with other platforms, encoder settings, and NLEs.

### Types of changes

### Checklist:

### DIFF
--- a/UI/data/locale/en-US.ini
+++ b/UI/data/locale/en-US.ini
@@ -941,6 +941,11 @@ Basic.Settings.Output.Adv.FFmpeg.AEncoderSettings="Audio Encoder Settings (if an
 Basic.Settings.Output.Adv.FFmpeg.MuxerSettings="Muxer Settings (if any)"
 Basic.Settings.Output.Adv.FFmpeg.GOPSize="Keyframe interval (frames)"
 Basic.Settings.Output.Adv.FFmpeg.IgnoreCodecCompat="Show all codecs (even if potentially incompatible)"
+Basic.Settings.Output.EnableSplitFile="Automatic File Splitting"
+Basic.Settings.Output.SplitFile.TypeTime="Split by Time"
+Basic.Settings.Output.SplitFile.TypeSize="Split by Size"
+Basic.Settings.Output.SplitFile.Time="Split Time"
+Basic.Settings.Output.SplitFile.Size="Split Size"
 
 # Screenshot
 Screenshot="Screenshot Output"
--- a/UI/data/locale/en-US.ini
+++ b/UI/data/locale/en-US.ini
@@ -946,6 +946,7 @@ Basic.Settings.Output.SplitFile.TypeTime="Split by Time"
 Basic.Settings.Output.SplitFile.TypeSize="Split by Size"
 Basic.Settings.Output.SplitFile.Time="Split Time"
 Basic.Settings.Output.SplitFile.Size="Split Size"
+Basic.Settings.Output.SplitFile.ResetTimestamps="Reset timestamps at the beginning of each split file"
 
 # Screenshot
 Screenshot="Screenshot Output"
--- a/UI/forms/OBSBasicSettings.ui
+++ b/UI/forms/OBSBasicSettings.ui
@@ -2408,6 +2408,85 @@
                             <item row="6" column="1">
                              <widget class="QLineEdit" name="advOutMuxCustom"/>
                             </item>
+                            <item row="7" column="0">
+                             <widget class="QCheckBox" name="advOutSplitFile">
+                              <property name="sizePolicy">
+                               <sizepolicy hsizetype="Minimum" vsizetype="Expanding">
+                                <horstretch>0</horstretch>
+                                <verstretch>0</verstretch>
+                               </sizepolicy>
+                              </property>
+                              <property name="layoutDirection">
+                               <enum>Qt::RightToLeft</enum>
+                              </property>
+                              <property name="text">
+                               <string>Basic.Settings.Output.EnableSplitFile</string>
+                              </property>
+                             </widget>
+                            </item>
+                            <item row="7" column="1">
+                             <widget class="QComboBox" name="advOutSplitFileType">
+                              <property name="enabled">
+                               <bool>false</bool>
+                              </property>
+                              <item>
+                               <property name="text">
+                                <string>Basic.Settings.Output.SplitFile.TypeTime</string>
+                               </property>
+                              </item>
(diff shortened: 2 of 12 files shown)


flaeri:
Hello again :)
As you said you wanted more testers on this, here is my report.
Checked out this PR, and rebased on latest master at the time (5f91bef8758fc228c2bad42c0e21833fd6b6e62c)

Windows 10 21H2
https://obsproject.com/logs/d1tomaYChKYzxSPa

In Davinci Resolve, I am seeing the same/similar, one frame audio dropout between the splits, as before this change.
I redid this test 3 times, confirming the output was the same.

The top green is the audio generated when the split files are placed back to back in the NLE.
The bottom green is the audio when using ffmpeg concat (perfect, cannot hear or see the split)
image

Example output from NLE (a little bit loud). Listen around 8-9 sec:
https://user-images.githubusercontent.com/50419942/162790061-15825d69-187d-42e8-94a1-4d0748d3359b.mov

Files OBS generated:
https://send.tresorit.com/a#4M_xuHED3qVgaPjyly02KQ

norihiro:
Implemented another idea; split only at the timing when both video and audio has the same timestamp.
The implementation relies on 60 fps video and 48 kHz audio setting. It requires more implementation for the case video-frame > audio-frame.
Another concern is that such condition might not occur frequently especially 59.94 fps and/or 44.1 kHz audio. Also note that some encoder could generate key frames randomly, which results much harder to meet the split condition. Maybe need to force encoder to output keyframe at the certain frame.

Picture below is a test results with 2 stitching points.
Added 432 Hz and 431 Hz tones to OBS Studio and recorded with the automatic split file feature. Then, stitched by Kdenlive.
First and second tracks are 432 Hz and 431 Hz tone, respectively. Third track is a stereo track  with 432 Hz and 431 Hz tones. The last track shows difference of OBS's recorded track and tone tracks. It shows no audio gap or glitch at all.
a

PatTheMav:
@norihiro What's the state of this PR? @flaeri's comment makes it seem as if the original implementation does not fix the issue and you came up with a different idea?

Is there a working implementation that fixes the issue? In that case I'd ask for the branch to be rebased and pushed and the PR undrafted so it can be reviewed.

norihiro:
I'm sorry to this activitiy stalled.

This is the branch I was experimenting.
- https://github.com/norihiro/obs-studio/tree/automatic-file-splitting-at-exact-ts
  (I rebased it today.)

Something I was trying are...
- Duplicating audio packets. [539a86c6]
  - The audio packet consists of 1024 samples. When encoding, it looks there is a window function and adjacent packets are overwrapped. So, just splitting audio packets causes fade-in effect at the beggining of each file when decoding. However, the audio is cut-off immediately at the end of each file. (This might depends on decoder or NLEs). That means the audio is missing when concatinating on an NLE. I guess the end of the audio is somehow missing. To compensate it, I output the audio packets to both current and new files when splitting the file so that there are extra audio packets at the end of the file.
  - This implementation currently relies on DTS and PTS has offset.
- Split only when the audio and video packet timestamps are aligned [1cb51603f]
  - This is a trial implementation that works with 60 fps video and 48 kHz audio.
  - I concern the aligned timing won't happen frequent enough when 44.1 kHz audio for example.
  - Maybe, need to adjust the grace time or give-up if the split is required but the aligned timing did not happen for a while such as 10 seconds or so.

One of the difficulties I have was it relies on NLE's implementation...

PatTheMav:
Is there a known implementation that does it "right", or something else that we can point to and say "this software produces splits that behave the way we want them to"? Because I'm currently not sure the acceptance criteria have been agreed upon.

Fenrirthviti:
Just a nudge on this, if this is still relevant and being worked towards, as Pat's question was not addressed.

norihiro:
> Is there a known implementation that does it "right", or something else that we can point to and say "this software produces splits that behave the way we want them to"?

I looked into two camera models; Sony FDR-AX45 and Panasonic HC-VX985M.

Both did not duplicate the audio packets like this PR but it looks the audio stream encoded by one encoder is split; FFmpeg concat can recover the audio without any gap.
So, in my opinion, current implementation in master branch should be sufficient.

FDR-AX45:
End of the 1st M2TS file:
```txt
packet|codec_type=video|stream_index=0|pts=57342195|pts_time=637.135500|dts=57340693|dts_time=637.118811|duration=1501|duration_time=0.016678|size=11369|pos=2125518720|flags=___
packet|codec_type=audio|stream_index=1|pts=57273902|pts_time=636.376689|dts=57273902|dts_time=636.376689|duration=2880|duration_time=0.032000|size=1792|pos=2125518336|flags=K__
```
Start of the 2nd M2TS file.
```txt
packet|codec_type=audio|stream_index=1|pts=57276782|pts_time=636.408689|dts=57276782|dts_time=636.408689|duration=2880|duration_time=0.032000|size=1792|pos=960|flags=K__
packet|codec_type=audio|stream_index=1|pts=57279662|pts_time=636.440689|dts=57279662|dts_time=636.440689|duration=2880|duration_time=0.032000|size=1792|pos=93120|flags=K__
packet|codec_type=video|stream_index=0|pts=57343697|pts_time=637.152189|dts=57342195|dts_time=637.135500|duration=1501|duration_time=0.016678|size=419062|pos=768|flags=K__
```

I also checked the packets from HC-VX985M but the PTS and DTS is reset after the file-split so that the packet DTS and PTS is useless to study how it is working.

I'd like to close this PR. How do you think?

Issue #5685, opened 2021-12-22
Linux: Window capture causes performance issues with proprietary nvidia drivers


univrsal:
### Operating System Info

Other

### Other OS

Arch Linux

### OBS Studio Version

Git

### OBS Studio Log URL

https://obsproject.com/logs/AqYbf16E3VT5g6cw

### Expected Behavior

Window capture should work as it did in https://github.com/obsproject/obs-studio/commit/4623a6b4bc4c89ceb5db684e2e7fbd57d01129aa. It seems that the issue originates from https://github.com/obsproject/obs-studio/commit/316f858c68d3ef3a73166191b6f89b7bc3f3387c which is linked to #5503.

### Current Behavior

Having a window capture visible will reduce the frame rate from 60 fps to 30. Moving any window around will reduce the frame rate even further.

### Steps to Reproduce

1. Have an nvidia card with proprietary drivers and obs built from master
2. Add a window capture

WizardCM:
I can confirm the experience is not good on the proprietary Nvidia drivers.

![image](https://user-images.githubusercontent.com/941350/147089607-477b8870-30f4-4004-957a-1b2406680a84.png)

ishitatsuyuki:
Looking at the logs, there is one suspicious line:

```
xcompcap: glXReleaseTexImageEXT failed: GLXBadPixmap
```

But a [search](https://github.com/search?q=glXReleaseTexImageEXT+obs&type=code) reveals that this thing has happened on Nvidia long before, so the error might or might not be related to actual issue.

As I don't own a Nvidia card, would you mind capturing a profile using the following instructions?

1. Install and set up `perf`. The exact method depends on what distro you are running.
2. Install [flamegraph (Rust version)](https://github.com/flamegraph-rs/flamegraph). If you have cargo installed and PATH properly set to include `~/.cargo/bin`, then it can be installed with `cargo install flamegraph`.
3. Launch OBS (the master version) and add the problematic capture source.
4. Obtain OBS' PID. You can use `pgrep -a obs` for example.
5. Run `flamegraph -p <PID>`. Switch back to OBS and briefly do anything that lags. Then interrupt the `flamegraph` command.
6. A flamegraph should be written to `flamegraph.svg`. Please upload it.

univrsal:
[...]

PR #5694, opened 2021-12-24
linux-capture: Disable strict binding for NVIDIA drivers

ishitatsuyuki:
NVIDIA drivers appears to have a bug where binding would be excessively
slow. Apply a workaround similar to what [KWin] does to prevent the issue.

Also performs a refactor so that the code paths with and without the
workaround can be shared.

[KWin]: https://github.com/KDE/kwin/blob/4f2c3a00c4a86c08321e80d601fecd09d0871c79/src/libkwineffects/kwinglplatform.cpp

Fixes: 316f858c6 ("linux-capture: Fix capturing on software rasterization setups")
Closes: https://github.com/obsproject/obs-studio/issues/5685

Tested-By: univrsal <uni@vrsal.xyz>

### Types of changes

### Checklist:

- N/A I have included updates to all appropriate documentation.

### DIFF
--- a/plugins/linux-capture/xcompcap-main.cpp
+++ b/plugins/linux-capture/xcompcap-main.cpp
@@ -201,6 +201,11 @@ struct XCompcapMain_private {
 	bool cursor_outside = false;
 	xcursor_t *cursor = nullptr;
 	bool tick_error_suppressed = false;
+	// Whether to rebind the GLX Pixmap on every tick. This is the correct
+	// mode of operation, according to GLX_EXT_texture_from_pixmap. However
+	// certain drivers exhibits poor performance when this is done, so
+	// setting this to false allows working around it.
+	bool strict_binding = true;
 };
 
 XCompcapMain::XCompcapMain(obs_data_t *settings, obs_source_t *source)
@@ -209,6 +214,11 @@ XCompcapMain::XCompcapMain(obs_data_t *settings, obs_source_t *source)
 	p->source = source;
 
 	obs_enter_graphics();
+	if (strcmp(reinterpret_cast<const char *>(glGetString(GL_VENDOR)),
+		   "NVIDIA Corporation") == 0) {
+		// Pixmap binds are extremely slow on NVIDIA cards (https://github.com/obsproject/obs-studio/issues/5685)
+		p->strict_binding = false;
+	}
 	p->cursor = xcursor_init(xdisp);
 	obs_leave_graphics();
 
@@ -301,6 +311,14 @@ static void xcc_cleanup(XCompcapMain_private *p)
 		GLuint gltex = *(GLuint *)gs_texture_get_obj(p->gltex);
 		glBindTexture(GL_TEXTURE_2D, gltex);
 		if (p->glxpixmap) {
+			glXReleaseTexImageEXT(xdisp, p->glxpixmap,
+					      GLX_FRONT_EXT);
+			if (xlock.gotError()) {
+				blog(LOG_ERROR,
+				     "cleanup glXReleaseTexImageEXT failed: %s",
+				     xlock.getErrorText().c_str());
+				xlock.resetError();
+			}
 			glXDestroyPixmap(xdisp, p->glxpixmap);
 			if (xlock.gotError()) {
 				blog(LOG_ERROR,
@@ -563,12 +581,6 @@ void XCompcapMain::updateSettings(obs_data_t *settings)
 	glTexParameteri(GL_TEXTURE_2D, GL_TEXTURE_MAG_FILTER, GL_LINEAR);
 	// glxBindTexImageEXT might modify the textures format.
 	gs_color_format format = gs_format_from_tex();
-	glXReleaseTexImageEXT(xdisp, p->glxpixmap, GLX_FRONT_EXT);
-	if (xlock.gotError()) {
-		blog(LOG_ERROR, "glXReleaseTexImageEXT failed: %s",
-		     xlock.getErrorText().c_str());
-		xlock.resetError();
-	}
 	glBindTexture(GL_TEXTURE_2D, 0);
 	// sync OBS texture format based on any glxBindTexImageEXT changes
 	p->gltex->format = format;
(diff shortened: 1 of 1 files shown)


ishitatsuyuki:
cc @kkartaltepe since you reviewed the previous PR

kkartaltepe:
I think this looks good to me, but I wont be able to test it for another two weeks or so.

kkartaltepe:
I still cant test this but I had a chance to look into some of the kwin code. It seems that most drivers should be using loose binding (at least kwin sets this for nvidia/amd and intel seems to need research). Since this ends up being a potential for issues we should probably modify the code to only use strict binding in the case of a software renderer and otherwise stick with our previous behavior of loose binding.

ishitatsuyuki:
> It seems that most drivers should be using loose binding (at least kwin sets this for nvidia/amd and intel seems to need research).

If you are talking about the code I linked in the commit message, no, it doesn't set it for modern AMD cards. I just checked my logs and I get `Requires strict binding:                yes`.

It has caused trouble with Intel in the past (https://bugs.freedesktop.org/show_bug.cgi?id=80349), which is why there's a line setting it to false for Intel cards (the default is false anyway, the assignment could have been removed in the patch).

Since the default is false, it's using strict binding for any other vendors (qcom etc, they actually use GLES/EGL anyway). The remaining question is legacy AMD drivers, for which two variants exists:
- Proprietary drivers have been mostly retired due to binary incompatibility: they don't run with the latest Xorg.
- Non-gallium drivers have got [booted](https://www.phoronix.com/scan.php?page=news_item&px=Mesa-Classic-Retired) from Mesa. Nobody should be using them at this point.
- I tend to believe the Gallium drivers are OK these days, although since r600 is explicitly marked to use loose binding in KWin I can port that 10-years old check to this patch too.

kkartaltepe:
> If you are talking about the code I linked in the commit message, no, it doesn't set it for modern AMD cards. I just checked my logs and I get Requires strict binding: yes.

Ah I see I guess r600 driver is the previous generation, since it looked like it was being set for those then AMD is split. However since there are no issues reported on AMD it seems that strict binding is not required so again my preference is to retain the old functionality unless there are known issues we can point to.

> It has caused trouble with Intel in the past (https://bugs.freedesktop.org/show_bug.cgi?id=80349), which is why there's a line setting it to false for Intel cards (the default is false anyway, the assignment could have been removed in the patch).

Right this patch is incredibly old so its unclear if this was a performance regression for the sake of this issue or not. Also we don't have this issue reported as far as I know so that is why my hesitant to say we should change our binding strategy on intel without more information.

> Since the default is false, it's using strict binding for any other vendors (qcom etc, they actually use GLES/EGL anyway).

This case (mesa drivers for all EGL users) ends up using loose binding it seems https://github.com/KDE/kwin/blob/4f2c3a00c4a86c08321e80d601fecd09d0871c79/src/libkwineffects/kwinglplatform.cpp#L1092 unless you mean proprietary drivers.

ishitatsuyuki:
> However since there are no issues reported on AMD it seems that strict binding is not required so again my preference is to retain the old functionality unless there are known issues we can point to.

OK, I will add r600 to the loose binding special casing list.

> Right this patch is incredibly old so its unclear if this was a performance regression for the sake of this issue or not. Also we don't have this issue reported as far as I know so that is why my hesitant to say we should change our binding strategy on intel without more information.

We should change the strategy because the old one is not spec compliant. At least some of the Mesa developers are aware of this "correct" interpretation: https://oftc.irclog.whitequark.org/dri-devel/2021-11-07#1636283343-1636283065; so it's reasonable to expect that strict binding is not slow on Mesa drivers. FWIW on my AMD setup it had nearly zero overhead (XLock took more time than the binding operations).

> This case (mesa drivers for all EGL users) ends up using loose binding it seems KDE/kwin@4f2c3a0/src/libkwineffects/kwinglplatform.cpp#L1092 unless you mean proprietary drivers.

KWin is actually wrong in this case, the `OES_EGL_image_external` (which supersedes `OES_EGL_image`) spec says

        Sampling an external texture which has been modified since it was
        bound will return samples which may correspond to image values either
        before, during, or after the modification.  Binding (or re-binding if
        already bound) an external texture by calling BindTexture after all
        modifications are complete guarantees that sampling done in future
        draw calls will return values corresponding to the values in the
        buffer at or after the time that BindTexture is called.  (Note that
        calling BindTexture after calling commands that modify the EGLImage
        may not be sufficient to ensure that the modifications are complete;
        additional synchronization (for example eglWaitAPI, eglWaitNative,
        glFinish, vgFinish, or eglWaitSyncKHR, etc) may be required following
        the call initiating modifications, to ensure they have taken effect on
        the texture, before the call to BindTexture is made.)

Basically the same requirement as GLX. The demo package also got updated https://github.com/freedesktop/mesa-demos/commit/f84d2225208b9ccb0f4f027d72fa75bb25439920#diff-ca608d789dbf11c506f2d27011700674e68def4db4b82bc2af047fae3a5e85a0 at around the time, although the KWin patch precedes it but I can imagine it not being communicated due to a delayed release.

According to the KWin commit it's nouveau causing the headache in this case (https://github.com/KDE/kwin/commit/9e2b6edf02b47dc414dac4ce0dbb48c2d1b949e7), we don't support EGL right now but we'll probably need to dig that grave to decide on a strict binding policy when we deal with it...

Techjar:
I was having serious performance issues with Xcomposite window capture on the latest OBS master. NVIDIA card here, of course. This PR brought it right back to nominal performance.

LGCW:
Tested this on the box is the studio this morning after encountering the bug in the latest beta. Nvidia 2060 running 495.46, xfce4.

This PR fixes the issue with Xcomposite bringing everything to a crawl.

WizardCM (changes requested):
This introduces no regressions on Intel, and fixes the performance regression on Nvidia.

However, it does seem to introduce a memory leak, counted on shutdown. This can be reproduced by having 1 window capture source and closing OBS, logging 2 memory leaks.

ishitatsuyuki:
> However, it does seem to introduce a memory leak, counted on shutdown.

I cannot reproduce this with both radeonsi and llvmpipe (`LIBGL_ALWAYS_SOFTWARE=1`). Can you paste your log? I guess it's probably due to a bad error handling path, and I've seen instances of `glXReleaseTexImageEXT` failing on NV for unknown reasons.

jp9000:
I'm not sure how this would add a memory leak. I don't see any allocations in this code.

WizardCM:
Log with this PR:
[2022-01-02 12-24-58.txt](https://github.com/obsproject/obs-studio/files/7807016/2022-01-02.12-24-58.txt)

Log without this PR (master at the time of test):
[2022-01-02 12-28-32.txt](https://github.com/obsproject/obs-studio/files/7807018/2022-01-02.12-28-32.txt)

I reproduced it a second time relaunching without removing/readding the source (after switching back to this branch):
[2022-01-02 12-31-54.txt](https://github.com/obsproject/obs-studio/files/7807025/2022-01-02.12-31-54.txt)

It's entirely possible the leaks are unrelated & a side effect of the branch being based on an older `master`.

ishitatsuyuki:
I honestly have no idea why there's a leak, I just rebased the branch so @WizardCM you may want to test again.

If it persists, maybe it's good time to compile with ASan to get allocations tracked with stack traces.

kkartaltepe:
Tested the latest PR, i dont have any memory leaks on intel or nvidia. Performance seems fine on both.

kkartaltepe (approved):

jp9000:
Nice branch name -- I sympathize with you as a Linux user. NVIDIA has not given me pleasant experiences on Linux.

PR #5700, opened 2021-12-27
UI: Update context bar when exiting properties dialog

gxalpha:
### Description

On current master, when clicking `OK` in the source properties dialog the context bar wouldn't update.
Fixes this issue.
(Changes update from normal to deferred to prevent the context bar "flashing" (going away and
reappearing) once when the properties dialog is closed)

### Motivation and Context

Not sure when this issue was introduced (and I'm not bisecting it just for the sake of finding out; it was probably me doing the minimized context bar anyways), I just randomly noticed it.

### How Has This Been Tested?

Tested with text source and color source, the context bar correctly shows
changes made in the properties dialog (in my case text/color).
 
### Types of changes

### Checklist:

### DIFF
--- a/UI/window-basic-properties.cpp
+++ b/UI/window-basic-properties.cpp
@@ -212,7 +212,7 @@ OBSBasicProperties::~OBSBasicProperties()
 	}
 	obs_source_dec_showing(source);
 	main->SaveProject();
-	main->UpdateContextBar();
+	main->UpdateContextBarDeferred(true);
 }
 
 void OBSBasicProperties::AddPreviewButton()


WizardCM:
Is this an issue of the function call occurring too early, or that the UpdateContextBar() function itself no longer working? What I mean to ask is - is this fixing an issue, or the side effect of a bigger issue?

gxalpha:
The important part of this PR is the `true`, not the `deferred`. The deferred is only there to stop it from disappearing for ~1 frame when updating.
`UpdateContextBar` only updates if the source has changed since the last time it updated, unless `force` is set to true. Personally I would consider this the correct behaviour.

PR #5702, opened 2021-12-27
Various EGL & PipeWire cleanups

GeorgesStavracas:
### Description

Various unassorted cleanups to the PipeWire code of the `linux-capture` plugin, and to `libobs-opengl`.

### Motivation and Context

As the code grows in scope and complexity, we need to do some conscious effort to keep it tidy.

### How Has This Been Tested?

There shouldn't be any actual functional changes here. All (except one) commits are aesthetical cleanups.

### Types of changes

### Checklist:

### DIFF
--- a/plugins/linux-capture/pipewire.c
+++ b/plugins/linux-capture/pipewire.c
@@ -688,10 +688,10 @@ static void on_process_cb(void *user_data)
 		}
 	} else {
 		blog(LOG_DEBUG, "[pipewire] Buffer has memory texture");
-		enum gs_color_format obs_format;
+		enum gs_color_format gs_format;
 
 		if (!lookup_format_info_from_spa_format(
-			    obs_pw->format.info.raw.format, NULL, &obs_format,
+			    obs_pw->format.info.raw.format, NULL, &gs_format,
 			    &swap_red_blue)) {
 			blog(LOG_ERROR,
 			     "[pipewire] unsupported DMA buffer format: %d",
@@ -702,7 +702,7 @@ static void on_process_cb(void *user_data)
 		g_clear_pointer(&obs_pw->texture, gs_texture_destroy);
 		obs_pw->texture = gs_texture_create(
 			obs_pw->format.info.raw.size.width,
-			obs_pw->format.info.raw.size.height, obs_format, 1,
+			obs_pw->format.info.raw.size.height, gs_format, 1,
 			(const uint8_t **)&buffer->datas[0].data, GS_DYNAMIC);
 	}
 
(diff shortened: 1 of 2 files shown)


GeorgesStavracas:
@columbarius could you please double-check the linux-capture changes? I'm confident my changes don't break anything, but it'd be wonderful if you could take a look at it too.

plugins/linux-capture/pipewire.c
```diff
@@ -405,19 +405,18 @@ static inline struct spa_pod *build_format(struct spa_pod_builder *b,
...
-		for (uint32_t i = 0; i < modifier_count; i++) {
-			uint64_t modifier = modifiers[i];
-			spa_pod_builder_long(b, modifier);
-			if (i == 0)
-				spa_pod_builder_long(b, modifier);
-		}
-		spa_pod_builder_pop(b, &f[1]);
+		for (uint32_t i = 0; i < modifier_count; i++)
+			spa_pod_builder_long(b, modifiers[i]);
+		spa_pod_builder_pop(b, &modifier_frame);
```
columbarius, line 419: PipeWire Choice PODs have the structure: Preference, Item1, Item2, .... , ItemN
See eg:
https://gitlab.freedesktop.org/pipewire/pipewire/-/blob/master/src/examples/sdl.h#L176
https://gitlab.freedesktop.org/pipewire/pipewire/-/blob/master/spa/plugins/bluez5/a2dp-codec-ldac.c#L188

Maybe we should add a comment?

```suggestion
		for (uint32_t i = 0; i < modifier_count; i++)
		        // Note: PipeWire Choice PODs start with a preference followed by the list
			if (i == 0)
			        spa_pod_builder_long(b, modifiers[i]);
			spa_pod_builder_long(b, modifiers[i]);
		spa_pod_builder_pop(b, &modifier_frame);
```

plugins/linux-capture/pipewire.c
```diff
@@ -312,63 +312,63 @@ static inline bool has_effective_crop(obs_pipewire_data *obs_pw)
...
-	default:
-		return false;
-	}
+static const struct {
+	uint32_t spa_format;
+	uint32_t drm_format;
+	enum gs_color_format gs_format;
+	bool swap_red_blue;
+	const char *pretty_name;
+} supported_formats[] = {
```
columbarius, line 321: I like this. 
It probably won't solve the issue of having all formats in one place if we aim at adding webcam support via PipeWire. If we won't want to add YUV to RGBA convertion to the `graphics` API we have to use the `media` API with it's own set of supported formats. But that is sth. we can solve at that time.

columbarius (changes requested):
Everything else LGTM!

plugins/linux-capture/pipewire.c
GeorgesStavracas: Thanks, I didn't know that. I've updated the patch so that the first modifier is still added, but outside the loop.

plugins/linux-capture/pipewire.c
GeorgesStavracas: Yeah, I think YUV will be a different beast, we'll figure it out when we start adding these formats.

columbarius:
Opt nit: Document that the "prefered" value is arbitrary in our case, maybe?
LGTM

PR #5150, opened 2021-08-18
UI: Add undo/redo for "Add existing source"

gxalpha:
### Description

Adds an undo/redo action for the "Add existing source", which previously
was missing.

Overall mimics the behaviour of the undo/redo, including the scene-switching to the scene that is affected when undo/redo-ing).

### Motivation and Context

Fixes #5091 

### How Has This Been Tested?

macOS 12 beta 5, compiled and run

Undo/Redo works, on redo the source reappears on the correct scene.

### Types of changes

- Bug fix (non-breaking change which fixes an issue)

### Checklist:

- [x] My code has been run through [clang-format](https://github.com/obsproject/obs-studio/blob/master/.clang-format).
- [x] I have read the [**contributing** document](https://github.com/obsproject/obs-studio/blob/master/CONTRIBUTING.rst).
- [x] My code is not on the master branch.
- [x] The code has been tested.
- [x] All commit messages are properly formatted and commits squashed where appropriate.
- [x] I have included updates to all appropriate documentation.

PR #5704, opened 2021-12-27
UI: Don't switch scenes on undo/redo; Force-switch away from scene when deleting it

gxalpha:
#### Preamble
While these two changes don't directly require each other, merging `Force-switch away from scene when deleting it` alone makes the unpredictability that `Don't switch scenes on undo/redo` is eliminating even worse, which is why I don't think they should be separate PRs.
So, here are the changes:

#### Commit 1: `UI: Force-switch away from scene when deleting it`
"When triggering an undo immediately after deleting a scene, if the
transition was still going the scene wouldn't be destroyed yet, while
undo would try to recreate it. This would lead to unpredictable
behavior, including crashes.
Instead of transitioning to the other scene, OBS will now force-switch
to it, skipping the transition and letting the removed scene get
destroyed immediately."
I think this shouldn't require further explanation.

#### Commit 2: `UI: Don't switch scenes on undo/redo`
"Previously, undo/redo would often times (but not always) force switch
back to the scene in which the original action was done. This is not
only unpredictable, but in many cases doesn't make any sense (for
example when modifying properties of a source).
Now, undo/redo no longer switches scenes, instead doing the changes
quietly."
Specifically, the following actions would switch scenes, but will no longer
do so with this PR:
- Add/Delete/Rename filters
- Add/Duplicate/Delete scene
- Edit transform of SceneItem (both in the dialog and the preview)
- Change source properties in the dialog
- Change source properties in the context bar
- Add source (new/exisiting)
- Rename/Delete source

### Motivation and Context

- 1: Noticed the problems while working on 2. Scenes disappearing and/or OBS crashing isn't good. The new behaviour (changing without transition) has been agreed to on Discord.
- 2: Discussion in https://github.com/obsproject/obs-studio/pull/5150#issuecomment-901448458

### How Has This Been Tested?

- Confirmed that scenes get destroyed immediately after they are deleted
- Tested that each of the undo/redo actions described above individually work correctly, without leaking memory or crashing. Also tested different scenarios with multiple of these actions directly after each other, as well as with nested scenes (because those often cause trouble). Everything in my tests got undone/redone correctly, without the scenes changing.

### Types of changes

### Checklist:

### DIFF
--- a/UI/context-bar-controls.cpp
+++ b/UI/context-bar-controls.cpp
@@ -64,11 +64,7 @@ void SourceToolbar::SetUndoProperties(obs_source_t *source, bool repeatable)
 			obs_data_get_string(settings, "undo_sname"));
 		obs_source_reset_settings(source, settings);
 
-		OBSSourceAutoRelease scene_source =
-			obs_get_source_by_name(scene_name.c_str());
-		main->SetCurrentScene(scene_source.Get(), true);
-
-		main->UpdateContextBar();
+		main->UpdateContextBar(true);
 	};
 
 	OBSDataAutoRelease new_settings = obs_data_create();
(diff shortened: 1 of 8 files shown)


gxalpha:
Closing as discussed - will make a list of inconsistencies at some point

PR #5705, opened 2021-12-27
text-freetype2: Improve properties dialog UX

### LINKED ISSUE #5638: Freetype text source doesn't toggle text/path fields when toggling "Read from file"
### Operating System Info

macOS 11.6

### Other OS

_No response_

### OBS Studio Version

27.1.3

_No response_

### OBS Studio Log URL

https://obsproject.com/logs/BvqNhdneGR7qcacg

### OBS Studio Crash Log URL

_No response_

### Expected Behavior

Ticking the 'Read from file' box should disable the 'Text' box, and replace it with a path to file box.

### Current Behavior

The 'Text' box never disappears, and when a file location is entered into it, nothing happens

### Steps to Reproduce

1. Open OBS (Using an M1 Mac Mini)
2. Add Text (FreeType 2)
3. Click 'Read from file'
Screenshot 2021-12-09 at 13 25 39
Screenshot 2021-12-09 at 13 26 10
Screenshot 2021-12-09 at 13 26 20
Screenshot 2021-12-09 at 13 26 36

Reading the log, I can see that it fails to open file '' - presumably this is because I can't populate the field that it would normally source the path from.

gxalpha:
### Description

Improves the UX of the Freetype text source's properties dialog.
The properties that only apply to "Read from file" are now in their own section, and the "Text" field will be disabled if "Read from file" is selected.
An alternative would be to hide and show the properties depending on whether "Read from file" is selected, but things just appearing and disappearing feels wrong to me personally. GDI+ does it like this, and I think it's horrible, though if someone has other opinions please voice them.

The "Word wrap" option will get disabled if "Custom text width" is below 100 since it only comes into effect if custom text width is bigger at least 100 (for values lower than that both of these settings are ignored).

Also "Enable Antialiasing" is moved out of the way.

New:
Bildschirmfoto 2021-12-27 um 23 13 16
Bildschirmfoto 2021-12-27 um 23 12 43

Old:
Bildschirmfoto 2021-12-27 um 23 15 37

This PR only modifies the properties dialog, it doesn't effect what happens with those settings.

### Motivation and Context

- Fixes #5638

### How Has This Been Tested?

Enabling "Read from file" correctly enables the section and having "Custom text width" at 100 or higher correctly enables the "Word wrap" button (and vice versa). See screenshots above for results.

### Types of changes

### Checklist:

### DIFF
--- a/plugins/text-freetype2/text-freetype2.c
+++ b/plugins/text-freetype2/text-freetype2.c
@@ -138,6 +138,45 @@ static uint32_t ft2_source_get_height(void *data)
 	return srcdata->cy + srcdata->outline_width;
 }
 
+#define SET_ENABLED(name, enabled)                       \
+	prop = obs_properties_get(props, name);          \
+	if (obs_property_enabled(prop) != enabled) {     \
+		obs_property_set_enabled(prop, enabled); \
+		modified = true;                         \
+	}
+
+static bool properties_from_file_changed(obs_properties_t *props,
+					 obs_property_t *prop,
+					 obs_data_t *settings)
+{
+	bool modified = false;
+	bool from_file = obs_data_get_bool(settings, "from_file");
+	SET_ENABLED("text", !from_file);
+	return modified;
+}
+
+static bool properties_log_mode_changed(obs_properties_t *props,
+					obs_property_t *prop,
+					obs_data_t *settings)
+{
+	bool modified = false;
+	bool log_mode = obs_data_get_bool(settings, "log_mode");
+	SET_ENABLED("log_lines", log_mode);
+	return modified;
+}
+
+static bool properties_custom_width_changed(obs_properties_t *props,
+					    obs_property_t *prop,
+					    obs_data_t *settings)
+{
+	bool modified = false;
+	int width = obs_data_get_int(settings, "custom_width");
+	SET_ENABLED("word_wrap", width >= 100);
+	return modified;
+}
+
+#undef SET_ENABLED
+
 static obs_properties_t *ft2_source_properties(void *unused)
 {
 	UNUSED_PARAMETER(unused);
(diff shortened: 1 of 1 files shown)


feetstv:
I agree that hiding/showing UI elements isn't good design. It violates the principle of least surprise.

Possibly in a separate PR, one could envision the colour wells move up to beneath the font selector. And a button to disable the second colour (because selecting two colours for single-colour text is a pain).

WizardCM:
I would personally prefer to bring the Freetype's properties order (and hide/show flags) to match the GDI+ text source.

Warchamp7:
> An alternative would be to hide and show the properties depending on whether "Read from file" is selected, but things just appearing and disappearing feels wrong to me personally. GDI+ does it like this, and I think it's horrible, though if someone has other opinions please voice them.

I'm preferential to the way GDI+ does it. It only shows fields relevant to what the user is doing, and conveys that they can't mix custom text with text from a file

> I agree that hiding/showing UI elements isn't good design. It violates the principle of least surprise.

I don't think the text entry field disappearing and being replaced with a file selection field when checking off read from file violates that principle. It is a clear change directly tied to the users input here.

feetstv:
> I don't think the text entry field disappearing and being replaced with a file selection field when checking off read from file violates that principle. It is a clear change directly tied to the users input here.

Only at the point of input. If a user comes back later, they need to remember that the text field is hidden because the check box is selected. Potential source of surprise for non-technical users.

From [Nielsen's 10 usability heuristics](https://www.nngroup.com/articles/ten-usability-heuristics/), it violates (1) and (6); a binary state is managed by a control that seems, on the surface, to only affect one control but actually affects two (showing/hiding the text field) which requires recall over recognition. That's the violation I mean.

Further example: a screen reader will not tell the user that there is a text field there so a low vision's mental model of the interface may preclude knowing/remembering that unchecking that box will add a text field there.

However, I agree with what you said:

> It only shows fields relevant to what the user is doing, and conveys that they can't mix custom text with text from a file

In that the design as proposed does not make it clear that "Read from file", which is in a group *below* the text field, will cause the text field *above* to become disabled.

Possible solution: radio buttons instead?

```
(o) Text
    [              ]
( ) Read from file

vs.

( ) Text
(o) Read from file
    Text file: [              ] [Browse]
    [x] Chat log mode
    Chat log lines: [              ] [↕]
```

This way, the state is preserved, and it's easier to build a mental model that the (dis)appearance of the text field relates to the choice of radio button. The binary state is shown by two controls that correspond to each state rather than a single control whose purpose would seem to be enabling one control whilst also manipulating the state of another control.

Then (1) is satisfied, (6) is eased, and no more surprises.

Warchamp7:
> > I don't think the text entry field disappearing and being replaced with a file selection field when checking off read from file violates that principle. It is a clear change directly tied to the users input here.
> 
> Only at the point of input. If a user comes back later, they need to remember that the text field is hidden because the check box is selected. Potential source of surprise for non-technical users.
> 
> From [Nielsen's 10 usability heuristics](https://www.nngroup.com/articles/ten-usability-heuristics/), it violates (1) and (6); a binary state is managed by a control that seems, on the surface, to only affect one control but actually affects two (showing/hiding the text field) which requires recall over recognition. That's the violation I mean.
> 
> Further example: a screen reader will not tell the user that there is a text field there so a low vision user's mental model of the interface may preclude knowing/remembering that unchecking that box will add a text field _above_ it.
> 
> However, I agree with what you said:
> 
> > It only shows fields relevant to what the user is doing, and conveys that they can't mix custom text with text from a file
> 
> In that the design as proposed does not make it clear that "Read from file", which is in a group _below_ the text field, will cause the text field _above_ to become disabled.
> 
> Possible solution: radio buttons instead?
> 
> ```
> (o) Text
>     [              ]
> ( ) Read from file
> 
> vs.
> 
> ( ) Text
> (o) Read from file
>     Text file: [              ] [Browse]
>     [x] Chat log mode
>     Chat log lines: [              ] [↕]
> ```
> 
> This way, the state is preserved, and it's easier to build a mental model that the (dis)appearance of the text field relates to the choice of radio button. The binary state is shown by two controls that correspond to each state rather than a single control whose purpose would seem to be enabling one control whilst also manipulating the state of another control.
> 
> Then (1) is satisfied, (6) is eased, and no more surprises.

I see your point, and I agree now. I'm in favour of a radio button solution as you suggested

gxalpha:
Closing as this soon™️ will be superseeded by a PR with radio buttons instead.

NxOne14:
> Closing as this soon™️ will be superseeded by a PR with radio buttons instead.

I think you should also close #5638 and mention the PR in that case.

gxalpha:
We generally only close issues once the fix is merged into master, so the issue should stay open.

PR #4974, opened 2021-07-04
[WIP] obs-nvenc on Linux

openglfreak:
### Description
Adds an OpenGL version of jim-nvenc to obs-ffmpeg.

#### Why is this change required?
The fallback NVENC encoder that is used on Linux copies rendered frames back to the system RAM, copies them again into an FFMPEG `AVFrame`, and makes FFMPEG upload them back to the GPU. This takes a lot of CPU time especially on older or lower-end CPUs, like my Intel i3-4370. Encoding the textures directly without copying them to system RAM solves this issue.

### How Has This Been Tested?

[WIP] I'm testing on Arch Linux using my Intel i3-4370 + Nvidia GTX 1060 6GB. CPU usage is decreased from around 25-30% to around 11%. The resulting video file looks correct.

### Types of changes

- Performance enhancement (non-breaking change which improves efficiency)

### Checklist:
- [x] My code has been run through [clang-format](https://github.com/obsproject/obs-studio/blob/master/.clang-format).
- [x] I have read the [**contributing** document](https://github.com/obsproject/obs-studio/blob/master/CONTRIBUTING.rst).
- [x] My code is not on the master branch.
- [ ] The code has been tested.
- [ ] All commit messages are properly formatted and commits squashed where appropriate.
- [ ] I have included updates to all appropriate documentation.

PR #4931, opened 2021-06-22
Direct ARGB encoding, texture and non-texture (RGB)

cmlin2:
### Description

Allow for direct ARGB (RGB) non-texture and texture encoding, bypassing unneeded NV12 conversion

### Motivation and Context

QSV from ICL+ allows for direct ARGB HW conversion so it improves CPU performance to use it (as unneeded RGB->NV12 conversion is done on CPU)

### How Has This Been Tested?

Tested on CFL machine with NVIDIA card and TGL machine on Windows, making RGB (non-texture and texture) selections and using Windows->Graphics Settings to change if the renderer and encoder are on the same GPU by selecting Power Saving or Let Windows Decide/High Performance. Results on both CFL and TGL (so machines that are ICL- and ICL+) show CPU improvement and demonstrate functionality with and without discrete graphics. 

### Types of changes

New feature (non-breaking change which adds functionality)

### Checklist:

- [X] My code has been run through [clang-format](https://github.com/obsproject/obs-studio/blob/master/.clang-format).
- [X] I have read the [**contributing** document](https://github.com/obsproject/obs-studio/blob/master/CONTRIBUTING.rst).
- [X] My code is not on the master branch.
- [X] The code has been tested.
- [X] All commit messages are properly formatted and commits squashed where appropriate.
- [X] I have included updates to all appropriate documentation.

PR #5706, opened 2021-12-27
ARGB texture encoding support

openglfreak:
### Description
Makes it possible to directly encode ARGB textures.
So far it was only possible to encode NV12 textures, but NVENC and Quick Sync Video also support ARGB textures as input.
This was split out of #4974 and supersedes #4931.

### Motivation and Context
Previously, switching to the RGB video format made jim-nvenc and obs-qsv11 fall back to the old encoders, with lower performance.
RGB support is also a prerequisite for the Linux support for jim-nvenc, see #4974.

### How Has This Been Tested?
Tested on a laptop with Windows 10 Pro version 21H2 and both an Intel UHD Graphics 630 and a Nvidia GeForce GTX 1650 Ti. Both Nvidia NVENC and Intel Quick Sync Video were tested with video color format NV12 and RGB, with no apparent breakages. The resulting video files were playable and looked normal.

### Types of changes

### Checklist:

### DIFF
--- a/libobs/obs-encoder.c
+++ b/libobs/obs-encoder.c
@@ -221,7 +221,7 @@ static void remove_connection(struct obs_encoder *encoder, bool shutdown)
 		audio_output_disconnect(encoder->media, encoder->mixer_idx,
 					receive_audio, encoder);
 	} else {
-		if (gpu_encode_available(encoder)) {
+		if (os_atomic_load_long(&obs->video.gpu_encoder_active)) {
 			stop_gpu_encode(encoder);
 		} else {
 			stop_raw_video(encoder->media, receive_video, encoder);
--- a/libobs/obs-encoder.c
+++ b/libobs/obs-encoder.c
@@ -199,7 +199,7 @@ static inline bool gpu_encode_available(const struct obs_encoder *encoder,
 						      info);
 }
 
-static void add_connection(struct obs_encoder *encoder)
+static bool add_connection(struct obs_encoder *encoder)
 {
 	if (encoder->info.type == OBS_ENCODER_AUDIO) {
 		struct audio_convert_info audio_info = {0};
@@ -220,6 +220,7 @@ static void add_connection(struct obs_encoder *encoder)
 	}
 
 	set_encoder_active(encoder, true);
+	return true;
 }
 
 static void remove_connection(struct obs_encoder *encoder, bool shutdown)
@@ -559,7 +560,7 @@ void pause_reset(struct pause_data *pause)
 	pthread_mutex_unlock(&pause->mutex);
 }
 
-static inline void obs_encoder_start_internal(
+static inline bool obs_encoder_start_internal(
 	obs_encoder_t *encoder,
 	void (*new_packet)(void *param, struct encoder_packet *packet),
 	void *param)
@@ -568,7 +569,7 @@ static inline void obs_encoder_start_internal(
 	bool first = false;
 
 	if (!encoder->context.data)
-		return;
+		return false;
 
 	pthread_mutex_lock(&encoder->callbacks_mutex);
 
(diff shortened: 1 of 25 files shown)


libobs/obs-encoder.c
```diff
@@ -221,7 +241,7 @@ static void remove_connection(struct obs_encoder *encoder, bool shutdown)
 		audio_output_disconnect(encoder->media, encoder->mixer_idx,
 					receive_audio, encoder);
 	} else {
-		if (gpu_encode_available(encoder)) {
+		if (os_atomic_load_long(&obs->video.gpu_encoder_active)) {
```
tytan652, line 244: The commit that change this may need more explanation about it.

tytan652 (commented):
A little while ago, I tried to rebase your branch and made #6510.

Even if I'm inexperienced with texture and encoding I looked at your PR and also #4931.

Thanks to HDR related commits that adds P010 texture encoding, I could kind of understand "how" adding a texture encode support should be done.

In my [tex_encoding_argb](https://github.com/tytan652/obs-studio/commits/tex_encoding_argb) branch (untested), I tried to add ARGB support "like" it was done for P010. I'm not going to PR it since I can't try this code, I have not Windows machine with Nvidia/Intel GPU for that.

But I will give my feedback about how this PR implement it.


- `encode_texture_available` in `obs_encoder_info` is not needed, the encoder implementation use a function like `obs_p010_tex_active()` and `obs_nv12_tex_active()` to determine if OBS is able to do it and fallback to another encoder if not available.
  - In [jim-nvenc](https://github.com/obsproject/obs-studio/blob/master/plugins/obs-ffmpeg/jim-nvenc.c#L1058)
  - In [obs-qsv11](https://github.com/obsproject/obs-studio/blob/master/plugins/obs-qsv11/obs-qsv11.c#L743)

- Like it's done with NV12 and P010, ARGB support support should be check against the device. Like [here](https://github.com/obsproject/obs-studio/blob/master/libobs-d3d11/d3d11-subsystem.cpp#L683).

- Your commit, about allowing `obs_encoder_start` to fail is not required since we shall fallback to another encoder.
  - If you want to propose this change, it should be done in another PR.

libobs/obs-encoder.c
openglfreak: To be honest I don't even remember why this change was needed, but it's definitely more correct than just relying on gpu encoding being active if it is available.

openglfreak:
@tytan652
Neither `obs_nv12_tex_active` nor `obs_p010_tex_active` work in the `.create` function, at least on Linux.
It also feels like bad design to have one function for each texture format and relying on the encoding plugin to do the right thing.
Asking the encoder directly if a format is supported sounds much better to me.
Any encoder that wants the old behavior of assuming NV12 and P010 support (which looking at the `gpu_encode_available` implementation looks like it might be broken too) can just not implement `.encode_texture_available`.

Also regarding the recent implementation of P010, changing `gpu_encode_available` to assume P010 support in all encoders sounds like a backwards-incompatible change. I'm not sure if that was well though-out.

I will implement ARGB support checking. I worked a little bit on that before but didn't think it was really necessary.

libobs/obs-encoder.c
openglfreak: Oh I remember why now: There might be encoders for which `gpu_encode_available` returns false due to the platform not supporting them, but that also don't have a `.encode` function as a fallback.
In that case starting the encoding fails, and it's better to fail gracefully in that case rather than to potentially do something bad.

tt2468:
> @tytan652 Neither `obs_nv12_tex_active` nor `obs_p010_tex_active` work in the `.create` function, at least on Linux. It also feels like bad design to have one function for each texture format and relying on the encoding plugin to do the right thing. Asking the encoder directly if a format is supported sounds much better to me. Any encoder that wants the old behavior of assuming NV12 and P010 support (which looking at the `gpu_encode_available` implementation looks like it might be broken too) can just not implement `.encode_texture_available`.
> 
> Also regarding the recent implementation of P010, changing `gpu_encode_available` to assume P010 support in all encoders sounds like a backwards-incompatible change. I'm not sure if that was well though-out.
> 
> I will implement ARGB support checking. I worked a little bit on that before but didn't think it was really necessary.

In this case, it may be good to make sure that all built in OBS encoders follow that new convention, and mark the old functions with `OBS_DEPRECATED`

libobs/obs-encoder.c
tari3x: I spoke to @openglfreak and it looks like he isn't likely to finish this PR, but would be happy for me to rebase it. 

If I rebase, all the comment history will be lost, as far as I understand, so first I wanted to check if this can easily be resolved as is - hopefully saving work for everyone involved. @tytan652, what are your thoughts?

PatTheMav: Feel free to open a new PR and link to this one. We can close this PR (so it's off our pile of stuff to review), but the comment history would still be preserved.

PatTheMav:
Closing as it was mentioned that the original author does not intend to finish this PR.

PR #5717, opened 2021-12-30
Add additional buffer when writing to file with obs-ffmpeg-mux

notr1ch:
### Description
This adds a circular buffer to ffmpeg-mux when writing to a file. Output from ffmpeg is buffered so that slow disk I/O does not block ffmpeg writes, as this causes the pipe to become full and OBS stops sending frames with a misleading "Encoding overloaded!" warning. The buffer may grow to 256 MB depending on the rate of data coming in and out, if the buffer is full OBS will start waiting in ffmpeg writes.

A separate I/O thread is responsible for processing the contents of the buffer and writing them to the output file. It tries to process 1 MB at a time ~~and align writes to 4 KB for better performance on drives with caching disabled~~. Small writes are batched up to a minimum of 64 KB.

It also supports seeking in case the muxer has to go back to an earlier part of the file. This works by flushing any pending buffers and then moving the file offset and collecting more writes until enough data is ready or another seek occurs. This is necessary for MOV and MP4 in case the user passes `movflags=faststart`, without the seek callback these formats would not even work at all.

### Motivation and Context
A common problem that users experience is an "Encoding overloaded!" warning when writing to external drives or even internal HDDs with formats such as MKV where writes can be small. External drives often have write caching disabled for quick removal, so small writes become latency bound and block the ffmpeg output, resulting in skipped frames.

The large buffer also allows for other I/O hiccups to be less impactful, for example writing to a network drive might stall due to temporary bandwidth use by other network traffic and with this PR the data would be buffered instead of dropped.

### How Has This Been Tested?
Extensive testing of various rates of data coming in and out of the buffer, with artificial sleep calls to simulate a slow disk. Tested cases of buffer being full or too small. Tested a real-world example on Discord where it solved the problem outlined above.

One potential risk this feature adds is a user recording to a disk that cannot physically keep up with the output bitrate. Instead of getting near-immediate feedback with an "Encoding overloaded!" warning, the user will only notice something is wrong once the buffer is full, which could take a long time depending on the output bitrate. I don't consider this a practical issue though, as even an extremely slow disk or network at 100mbps would only take ~30 seconds before filling the buffer.

As there is the chance of introducing a dangerous race condition between threads (output data would be lost if this code has bugs), I'm marking this as draft until it has had more testing. For testing purposes, you only need to replace the `obs-ffmpeg-mux.exe` in a regular OBS install with [obs-ffmpeg-mux.zip](https://github.com/obsproject/obs-studio/files/7795071/obs-ffmpeg-mux.zip).

### Types of changes

### Checklist:

### DIFF
--- a/plugins/obs-ffmpeg/ffmpeg-mux/ffmpeg-mux.c
+++ b/plugins/obs-ffmpeg/ffmpeg-mux/ffmpeg-mux.c
@@ -26,6 +26,9 @@
 #include <stdlib.h>
 #include "ffmpeg-mux.h"
 
+#include <util/threading.h>
+#include <util/platform.h>
+#include <util/circlebuf.h>
 #include <util/dstr.h>
 #include <libavcodec/avcodec.h>
 #include <libavformat/avformat.h>
@@ -41,6 +44,8 @@
 #else
 #define CODEC_FLAG_GLOBAL_H CODEC_FLAG_GLOBAL_HEADER
 #endif
+
+#define AVIO_BUFFER_SIZE 65536
 
 /* ------------------------------------------------------------------------- */
 
@@ -117,6 +122,24 @@ struct audio_info {
 	AVCodecContext *ctx;
 };
 
+struct io_header {
+	uint64_t seek_offset;
+	uint64_t data_length;
+};
+
+struct io_buffer {
+	bool active;
+	bool shutdown_requested;
+	bool output_error;
+	os_event_t *buffer_space_available_event;
+	os_event_t *new_data_available_event;
+	pthread_t io_thread;
+	pthread_mutex_t data_mutex;
+	FILE *output_file;
+	struct circlebuf data;
+	uint64_t next_pos;
+};
+
 struct ffmpeg_mux {
 	AVFormatContext *output;
 	AVStream *video_stream;
@@ -129,7 +152,7 @@ struct ffmpeg_mux {
 	struct header *audio_header;
 	int num_audio_streams;
 	bool initialized;
-	char error[4096];
+	struct io_buffer io;
 };
 
 static void header_free(struct header *header)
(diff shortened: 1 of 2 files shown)


notr1ch:
Pre-compiled `obs-ffmpeg-mux.exe` for easy testing: [obs-ffmpeg-mux.zip](https://github.com/obsproject/obs-studio/files/7794670/obs-ffmpeg-mux.zip)

kkartaltepe:
I noticed stalling due to the small pipe sizes available on linux, some testing showed that increasing the pipe size worked to prevent MKV encoding overload (https://github.com/kkartaltepe/obs-studio/commit/cfa76d37f9e01cda4c67ce8eb86029c5ce2ee309) and a buffer after the pipe worked (https://github.com/kkartaltepe/obs-studio/commit/79c01ab6f4236c307437819985572b7622a3e284) which seems like what you implemented? If so, im onboard with this.

Did you notice this on a non-linux system?

notr1ch:
This occurs often on Windows especially when writing to removable drives (no write cache). I had debated whether to buffer on the input or output side, I chose output to minimize small I/Os during the mux process and unblock input as fast as possible. After running into issues where ffmpeg wants to seek the output file for certain muxers, I'm certainly wondering whether this was the correct idea...

kkartaltepe:
> I chose output to minimize small I/Os during the mux process and unblock input as fast as possible.

I think on the input side (after reading from the pipe) is the "fastest" as you dont need to wait for the muxer and you can unblock the main OBS process. Also at least from my testing i couldnt identify issues with total throughput on the ffmpeg cli so I assumed the output side worked fast enough even on my extremely high latency disks (but with write cache).

> This occurs often on Windows especially when writing to removable drives (no write cache).

Good to know, im happy to test on linux as well later.

notr1ch:
It might be an idea to add a bit of input buffering then just in case. I was under the impression ffmpeg-mux would be mostly I/O bound, since it's not doing any encoding itself and only muxing the already-encoded packets, so there should be minimal processing overhead and disk I/O was what was causing the bottleneck.

carbn:
I tested this on windows and the patched version seems to resolve all issues I had with frame skipping.

kkartaltepe:
For me this is as effective as the branches I was using before for linux.

NarryG:
Been testing this out in my own fork and it's been working great for me and a friend. Both on Windows, no issues here.

plugins/obs-ffmpeg/ffmpeg-mux/ffmpeg-mux.c
```diff
@@ -567,6 +619,219 @@ static inline bool ffmpeg_mux_get_extra_data(struct ffmpeg_mux *ffm)
...
+	if (os_atomic_load_bool(&ffm->io.output_error))
+		return -1;
+
+	for (;;) {
+		pthread_mutex_lock(&ffm->io.data_mutex);
+
+		// Avoid unbounded growth of the circlebuf, cap to 256 MB
+		if (ffm->io.data.capacity >= 256 * 1048576 &&
+		    ffm->io.data.capacity - ffm->io.data.size <
+			    buf_size + 1 + sizeof(buf_size)) {
```
notr1ch, line 805: Should be `sizeof(io_header)` now.

RandompelaajaMr:
I have raspberry pi 4 NAS with 2 usb hard disk on raid 0. Was testing recording to straight there with mkv. The encoder was dropping frames when using mkv but with mp4 it was somekind fine. Tried this on windows and everything was fixed. No dropped frames from encoder anymore.

flaeri:
The precompiled files (pr itself, and first comment) are not yet updated to e6e25cb, correct?
Is this intentional?

notr1ch:
Yes, no relevant changes, I just rebased and fixed the overflow calculation.

PR #5727, opened 2022-01-03
UI: Do not localize timestamp in log file

norihiro:
### Description

`%T` is specified for `strtime` instead of `%X` so that the time format is not localized.

### Motivation and Context

Time format in the log was fluctuating in Linux. (I think its also applicable for Windows and macOS.) Sometimes `AM` or `PM` was inserted so that the milisecond was not always located just after the second.
Millisecond time in the log was added in 36ad777b8 but the millisecond was not always added just after the second digits depending on locale settings. Some locale settings add `AM` or `PM` at the end so that second and millisecond are separated.

The log is also read by support volunteers among a lot of country so that it should not be translated. Currently, some Asian language is written.
| `LANG` | Time format in log |
|--------|--------------------|
| `en_US.utf8` | `11:14:38 AM.952` |
| `C` | `14:10:19.180` |
| `ja_JP.UTF-8` | `14時11分44秒.347` |
| `ko_KR.UTF-8` | `14시 13분 57초.998` |
| `zh_CN.UTF-8` | `14:29:53.318` |

Previously, the log analyzer failed to parse the log file to retrieve OBS Version because the number of space (` `) changes.
https://discord.com/channels/348973006581923840/636682839382949888/883004441572278333
This bug was fixed in log analyzer but it would be easier to fix in OBS Studio to avoid similar error in future.

### How Has This Been Tested?

Tested on Fedora 34.

Since the `%T` format is same as the `%X` format in some locale, I think it won't break the behavior of the log analyzer.

The function `CurrentTimeString` is exported through `obs.hpp`. It could affect plugins. (I don't think there is a plugin depending on the format of this function.)

For Windows, `%T` is available for `strftime` according to [this document](https://docs.microsoft.com/en-us/cpp/c-runtime-library/reference/strftime-wcsftime-strftime-l-wcsftime-l?view=msvc-170).

### Types of changes

### Checklist:

  - Only tested on Linux.

### DIFF
--- a/UI/obs-app.cpp
+++ b/UI/obs-app.cpp
@@ -245,7 +245,7 @@ string CurrentTimeString()
 	auto now = system_clock::to_time_t(tp);
 	tstruct = *localtime(&now);
 
-	size_t written = strftime(buf, sizeof(buf), "%X", &tstruct);
+	size_t written = strftime(buf, sizeof(buf), "%T", &tstruct);
 	if (ratio_less<system_clock::period, seconds::period>::value &&
 	    written && (sizeof(buf) - written) > 5) {
 		auto tp_secs = time_point_cast<seconds>(tp);


WizardCM (approved):

tytan652 (approved):
I close https://github.com/obsproject/loganalyzer/pull/85 once this merged.

PR #5394, opened 2021-10-09
Migrate pipewire-capture to linux-pipewire

columbarius:
### Description

Migrates the PipeWire based pipewire-capture from the linux-capture plugin to the linux-pipewire plugin.

### Motivation and Context

PipeWire has to be initialized and deinitialized. Doing the deinitialization twice causes obs to segfault. To avoid this move both into the same plugin.

### Types of changes

- Code cleanup (non-breaking change which makes code smaller or more readable)

### Checklist:

- [x] My code has been run through [clang-format](https://github.com/obsproject/obs-studio/blob/master/.clang-format).
- [x] I have read the [**contributing** document](https://github.com/obsproject/obs-studio/blob/master/CONTRIBUTING.rst).
- [x] My code is not on the master branch.
- [x] The code has been tested.
- [x] All commit messages are properly formatted and commits squashed where appropriate.
- [x] I have included updates to all appropriate documentation.

Requires and contains [#5377 ](https://github.com/obsproject/obs-studio/pull/5155)

PR #5728, opened 2022-01-03
PipeWire-based Camera

columbarius:
### Description

This MR allows users to add a webcam as video source via PipeWire.

### Motivation and Context

PipeWire is the more modern multimedia framework for linux and other unix systems. One improvement over v4l2 is that PipeWire allows multiple clients to access the same device.

### How Has This Been Tested?
1. Run OBS
2. add `Webcam (PipeWire)`
3. Enjoy a very laggy stream (See disclaimer)

### Types of changes

### Checklist:

Requires and contains: #5394 

Disclaimer:
This is a somewhat early WIP from @GeorgesStavracas and me. While there are still some very rough spots (including bad performance and probably crashes) I think the changes in the shared code base are worth discussing.
Additionally this should avoid parallel development.

### DIFF
--- a/plugins/linux-pipewire/camera-portal.c
+++ b/plugins/linux-pipewire/camera-portal.c
@@ -1,0 +1,594 @@
+/* camera-portal.c
+ *
+ * Copyright 2021 Georges Basile Stavracas Neto <georges.stavracas@gmail.com>
+ *
+ * This program is free software: you can redistribute it and/or modify
+ * it under the terms of the GNU General Public License as published by
+ * the Free Software Foundation, either version 2 of the License, or
+ * (at your option) any later version.
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU General Public License for more details.
+ *
+ * You should have received a copy of the GNU General Public License
+ * along with this program.  If not, see <http://www.gnu.org/licenses/>.
+ *
+ * SPDX-License-Identifier: GPL-2.0-or-later
+ */
+
+#include "pipewire.h"
+#include "portal.h"
+
+#include <util/dstr.h>
+
+#include <fcntl.h>
+#include <unistd.h>
+#include <gio/gio.h>
+#include <gio/gunixfdlist.h>
+#include <spa/debug/dict.h>
+#include <spa/node/keys.h>
+#include <spa/pod/iter.h>
+#include <spa/utils/defs.h>
+#include <spa/utils/keys.h>
+
+struct camera_portal_source {
+	obs_source_t *source;
+	obs_data_t *settings;
+
+	obs_pipewire_stream *obs_pw_stream;
+	char *device_id;
+};
+
+/* ------------------------------------------------- */
+
+struct pw_portal_connection {
+	obs_pipewire *obs_pw;
+	GHashTable *devices;
+	GCancellable *cancellable;
+	GPtrArray *sources;
+	bool initializing;
+};
+
+struct pw_portal_connection *connection = NULL;
+
+static void pw_portal_connection_free(struct pw_portal_connection *connection)
+{
+	if (!connection)
+		return;
+
+	g_cancellable_cancel(connection->cancellable);
+
+	g_clear_pointer(&connection->devices, g_hash_table_destroy);
+	g_clear_pointer(&connection->obs_pw, obs_pipewire_destroy);
+	g_clear_pointer(&connection->sources, g_ptr_array_unref);
+	g_clear_object(&connection->cancellable);
+	bfree(connection);
+}
+
+static GDBusProxy *camera_proxy = NULL;
+
+static void ensure_camera_portal_proxy(void)
+{
+	g_autoptr(GError) error = NULL;
+	if (!camera_proxy) {
+		camera_proxy = g_dbus_proxy_new_sync(
+			portal_get_dbus_connection(), G_DBUS_PROXY_FLAGS_NONE,
+			NULL, "org.freedesktop.portal.Desktop",
+			"/org/freedesktop/portal/desktop",
+			"org.freedesktop.portal.Camera", NULL, &error);
+
+		if (error) {
+			blog(LOG_WARNING,
+			     "[portals] Error retrieving D-Bus proxy: %s",
+			     error->message);
+			return;
+		}
+	}
+}
+
+static GDBusProxy *get_camera_portal_proxy(void)
+{
+	ensure_camera_portal_proxy();
+	return camera_proxy;
+}
+
+static uint32_t get_camera_version(void)
+{
+	g_autoptr(GVariant) cached_version = NULL;
+	uint32_t version;
+
+	ensure_camera_portal_proxy();
+
+	if (!camera_proxy)
+		return 0;
+
+	cached_version =
+		g_dbus_proxy_get_cached_property(camera_proxy, "version");
+	version = cached_version ? g_variant_get_uint32(cached_version) : 0;
+
+	return version;
+}
+
+/* ------------------------------------------------- */
+
+struct camera_device {
+	uint32_t id;
+	struct pw_properties *properties;
+	struct pw_proxy *proxy;
+	struct spa_hook proxy_listener;
+};
+
+static struct camera_device *
+camera_device_new(uint32_t id, const struct spa_dict *properties)
+{
+	struct camera_device *device = bzalloc(sizeof(struct camera_device));
+	device->id = id;
+	device->properties = properties ? pw_properties_new_dict(properties)
+					: NULL;
+	return device;
+}
+
+static void camera_device_free(struct camera_device *device)
+{
+	if (!device)
+		return;
+
+	g_clear_pointer(&device->proxy, pw_proxy_destroy);
+	g_clear_pointer(&device->properties, pw_properties_free);
+	bfree(device);
+}
+
+/* ------------------------------------------------- */
+
+static bool update_device_id(struct camera_portal_source *camera_source,
+			     const char *new_device_id)
+{
+	if (strcmp(camera_source->device_id, new_device_id) == 0)
+		return false;
+
+	g_clear_pointer(&camera_source->device_id, bfree);
+	camera_source->device_id = bstrdup(new_device_id);
+
+	return true;
+}
+
+static void stream_camera(struct camera_portal_source *camera_source)
+{
+	struct camera_device *device;
+
+	g_clear_pointer(&camera_source->obs_pw_stream,
+			obs_pipewire_stream_destroy);
+
+	device = g_hash_table_lookup(connection->devices,
+				     camera_source->device_id);
+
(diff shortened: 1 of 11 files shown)


dodgepong:
The only localization file you need to edit is the English one -- all the others will be handled by CrowdIn, including the string deletions.

The parlance we currently use is "Video capture device" rather than "Camera" or "Webcam", since the device could be a capture card rather than an actual camera.

(Yes, there's an argument to be made about using the term "camera"/"webcam" for the same of understandability, but that discussion is out of scope for this PR IMO -- instead, for the time being I think we should maintain consistency in terminology.)

columbarius:
Thanks for your note!

> The only localization file you need to edit is the English one -- all the others will be handled by CrowdIn, including the string deletions.

The localization stuff is ... sigh .. an artefact/rebase error of #5394 and will disappear before this will be marked as ready. Since #5394 will probably wait for the big cmake change (Tell me if someone wants to merge it earlier) I would like to change that at that time.

> The parlance we currently use is "Video capture device" rather than "Camera" or "Webcam", since the device could be a capture card rather than an actual camera.

"Video capture device" is fine for me. This uses the official "Camera portal", but I guess that includes other capture devices? This might be changed to use the "device portal" if that becomes the way to go. @GeorgesStavracas: If there are restriction on what devices this portal will announce, please suggest a more fitting name ;)

columbarius:
> The parlance we currently use is "Video capture device" rather than "Camera" or "Webcam", since the device could be a capture card rather than an actual camera.

About this: I'd be interesting how such a capture device is handled by the linux kernel. Is it just treated as another v4l2 device? The reason I'm asking is that we can use the more optimized code path if we don't have to deal with YUY2 et. al.. Sadly I don't have a capture card to test it.

columbarius:
@GeorgesStavracas 
Rebasing this a had some ideas on how to deal with the registry. New proposal:

We separate camera plugin and registry.
* camera-plugin: `init_registry` -> `query_cameras` -> `user selects camera` -> `query_registry for fd and id` -> `create stream like screencast` -> ... -> `unref_registry`
* camera-registry: global refcounted struct, which deals with the camera/device-portal. Is created on first `init_registry` call and destroyed if the refcount is 0. It queries the portal for a filedescriptor and opens the connection and tracks all announced cameras.

An alternative would be to not lazyload the registry, but start it when obs is launched. This way we can only add the camera plugin, if obs is allowed to access the portal. 

If we lazy load I would suggest to work asynchronous with callbacks from the registry to the plugin after it's initialized, but that's just some idea.

Anyways: should I just continue to rebase the current state, or should I continue with my new proposal?

GeorgesStavracas:
> We separate camera plugin and registry.
>   * camera-plugin: `init_registry` -> `query_cameras` -> `user selects camera` -> `query_registry for fd and id` -> `create stream like screencast` -> ... -> `unref_registry`
>   * camera-registry: global refcounted struct, which deals with the camera/device-portal. Is created on first `init_registry` call and destroyed if the refcount is 0. It queries the portal for a filedescriptor and opens the connection and tracks all announced cameras.

In principle, I think separating different moving parts into their own unit is good. My idea with #6645 is that all the code under `pipewire.c` is generic and can be used by all source types - even audio sources.

I don't know if it would be feasible or even possible, but I think you can already make this `camera-registry` thing completely generic, so it can be reused by the audio source at #6207. It'll probably need to be implemented in `pipewire.c` directly, perhaps as follows:

 * Split the code that actually initializes PipeWire from `obs_pipewire_create` into a new `obs_pipewire_start`
 * Add a new vfunc struct `obs_pipewire_device_callbacks` with the regular `device_added`, `device_removed`, `device_changed`, etc callback (please feel free to suggest better naming here)
 * Add `obs_pipewire_set_device_callbacks()`

That way the camera portal code will be able to react to PipeWire adding and removing cameras, without having to deal with PipeWire itself.

What do you think? 

> An alternative would be to not lazyload the registry, but start it when obs is launched. This way we can only add the camera plugin, if obs is allowed to access the portal.
> 
> If we lazy load I would suggest to work asynchronous with callbacks from the registry to the plugin after it's initialized, but that's just some idea.

I think the best approach here is to do what I've done with the screencasting code, and completely separate camera portal code from PipeWire itself. What the plugin needs to do is to expose the pw-camera sources all the time, and only query the Device Access portal when the user actually adds a pw-camera source. If camera access is denied, we show an error message in the properties dialog.

> Anyways: should I just continue to rebase the current state, or should I continue with my new proposal?

Considering what I've suggested above, feel free to continue with your new proposal

columbarius:
> > Anyways: should I just continue to rebase the current state, or should I continue with my new proposal?
> 
> Considering what I've suggested above, feel free to continue with your new proposal

Moved this discussion to #6669

columbarius:
@GeorgesStavracas
I rebased this into https://github.com/columbarius/obs-studio/tree/pipewire-camera-portal-v2

The registry stuff works pretty nice, but creating a new camera source will hang obs if you create a new plugin and obs creates the property and starts connecting to the camera.

GeorgesStavracas:
Me and @columbarius discussed this yesterday and today about this PR, and we agreed on me pushing commits over.

This is my iteration of the work I've been meaning to do for a while. There's quite a large preparation in the first 19 commits, and the last commit actually implements the Camera portal integration.

plugins/linux-pipewire/pipewire.h
```diff
@@ -22,16 +22,31 @@
 
 #include <obs-module.h>
 
-typedef struct _obs_pipewire_data obs_pipewire_data;
+#include <pipewire/pipewire.h>
 
-void *obs_pipewire_create(int pipewire_fd, int pipewire_node);
-void obs_pipewire_destroy(obs_pipewire_data *obs_pw);
+typedef struct _obs_pipewire obs_pipewire;
+typedef struct _obs_pipewire_stream obs_pipewire_stream;
```
RytoEX, line 28: I think that names that begin with underscore are reserved, though I might be misunderstanding the "reserved names" section of the C standard. @kkartaltepe may understand it more than I.

plugins/linux-pipewire/portal.c
```diff
@@ -121,3 +130,78 @@ void portal_create_session_path(char **out_path, char **out_token)
...
+	bfree(call);
+}
+
+static void on_cancelled_cb(GCancellable *cancellable, void *data)
+{
+	UNUSED_PARAMETER(cancellable);
+
+	struct portal_signal_call *call = data;
+
+	blog(LOG_INFO, "[pipewire] Request cancelled");
```
RytoEX, line 155: Is the logging prefix in this module "[pipewire]" or "[pipewire]: " or something else?

plugins/linux-pipewire/camera-portal.c
```diff
@@ -0,0 +1,594 @@
...
+	g_autoptr(GError) error = NULL;
+	int pipewire_fd;
+	int fd_index;
+
+	result = g_dbus_proxy_call_with_unix_fd_list_finish(
+		G_DBUS_PROXY(source), &fd_list, res, &error);
+	if (error) {
+		if (!g_error_matches(error, G_IO_ERROR, G_IO_ERROR_CANCELLED))
+			blog(LOG_ERROR,
+			     "[camera-portal] Error retrieving pipewire fd: %s",
```
RytoEX, line 330: PipeWire ?

plugins/linux-pipewire/camera-portal.c
```diff
@@ -0,0 +1,594 @@
...
+		return;
+	}
+
+	g_variant_get(result, "(h)", &fd_index, &error);
+
+	pipewire_fd = g_unix_fd_list_get(fd_list, fd_index, &error);
+	if (error) {
+		if (!g_error_matches(error, G_IO_ERROR, G_IO_ERROR_CANCELLED))
+			blog(LOG_ERROR,
+			     "[camera-portal] Error retrieving pipewire fd: %s",
```
RytoEX, line 341: PipeWire ?

RytoEX (changes requested):
Nitpicks only.

https://github.com/obsproject/obs-studio/pull/5728/commits/ab3bcbc82820480dfc095659f5fd034e831fccf3
"separate"

https://github.com/obsproject/obs-studio/pull/5728/commits/c8374e6c824c24e58d12fce2fa2cadae93426a99
"linux-pipewire"

plugins/linux-pipewire/pipewire.h
GeorgesStavracas: I'm not aware of any kind of name reservation mechanism in C. The C language has reserved keyword (if, else, while, ...), and the standard types and functions of `libc` must not be overriden by application code, but I don't think there's anything preventing applications from defining structs like here.

RytoEX: See the following regarding identifiers that begin with an underscore:
* https://www.gnu.org/software/libc/manual/html_node/Reserved-Names.html
* https://en.cppreference.com/w/c/language/identifier

I generally would opt to just avoid starting any identifier with an underscore just to avoid having to wonder if it's acceptable according to the standard.

plugins/linux-pipewire/portal.c
GeorgesStavracas: It's actually "portals"

plugins/linux-pipewire/camera-portal.c
```diff
@@ -0,0 +1,594 @@
...
+
+	obs_property_list_clear(device_list);
+
+	device_found = false;
+	g_hash_table_iter_init(&iter, connection->devices);
+	while (g_hash_table_iter_next(&iter, (gpointer *)&device_id,
+				      (gpointer *)&device)) {
+		const char *device_name;
+
+		device_name = pw_properties_get(device->properties,
```
columbarius, line 214: This is a PipeWire issue, but for now, we should use more than `PW_KEY_NODE_DESCRIPTION`. This value is for my webcam "0807" for some reason. I'd suggest maybe adding the object.path Property to that string (Maybe `descr (path)`) which is sth. like `v4l2:/dev/video0`. This would also allow people to notice why a webcam used by PipeWire can't be accessed via v4l2 from another application.

plugins/linux-pipewire/camera-portal.c
```diff
@@ -0,0 +1,594 @@
...
+	const char *device_id;
+
+	if (strcmp(type, PW_TYPE_INTERFACE_Node) != 0)
+		return;
+
+	registry = obs_pipewire_get_registry(connection->obs_pw);
+	device_id = spa_dict_lookup(props, SPA_KEY_NODE_NAME);
+
+	device = camera_device_new(id, props);
+	device->proxy = pw_registry_bind(registry, id, type, version, 0);
```
columbarius, line 276: Is there a reason, why we create a proxy for each PipeWire? I think we don't need that.

GeorgesStavracas: This is what pw-dump does, and apparently tracks the life of each device individually

ChrisLane:
Having no success running obs-studio wrapped with `pw-v4l2` to access my PipeWire camera stream, I wanted to try this.
Could you fix the merge conflicts please?

GeorgesStavracas:
This is obsoleted by https://github.com/obsproject/obs-studio/pull/9771

PR #5730, opened 2022-01-03
UI: Reenable copy/paste transform & Correctly enable edit menu items with multiple sources

gxalpha:
### Description

Fixes copy/paste transform not working anymore.
Also fixes the Edit menu having things enabled that shouldn't be enabled,
like "Edit Transform" while multiple items are selected (or like "Copy Transform"
being enabled when multiple items are selected, this is why these changes
are in one PR, the second commit depends on changes made in the first).

### Motivation and Context

27.2 issue report

### How Has This Been Tested?

Copy/Paste transform works again.
"Edit Transform", "Copy Transform" and "Copy Filters" are no longer available if
multiple source are selected.

### Types of changes

### Checklist:

### DIFF
--- a/UI/window-basic-main.cpp
+++ b/UI/window-basic-main.cpp
@@ -7771,16 +7771,14 @@ config_t *OBSBasic::Config() const
 
 void OBSBasic::UpdateEditMenu()
 {
-	int idx = GetTopSelectedSourceItem();
 	QModelIndexList items = GetAllSelectedSourceItems();
 	int count = items.count();
 	size_t filter_count = 0;
-	OBSSceneItem sceneItem;
-	OBSSource source;
 
-	if (idx != -1) {
-		sceneItem = ui->sources->Get(idx);
-		source = obs_sceneitem_get_source(sceneItem);
+	if (count == 1) {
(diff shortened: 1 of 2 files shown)


UI/window-basic-main.cpp
```diff
@@ -7842,7 +7846,15 @@ void OBSBasic::on_actionEditTransform_triggered()
 
 void OBSBasic::on_actionCopyTransform_triggered()
 {
+	OBSSceneItem item = GetCurrentSceneItem();
+
+	obs_sceneitem_defer_update_begin(item);
+	obs_sceneitem_get_info(item, &copiedTransformInfo);
+	obs_sceneitem_get_crop(item, &copiedCropInfo);
+	obs_sceneitem_defer_update_end(item);
```
notr1ch, line 7854: I don't think we need to be calling `obs_sceneitem_defer_update_begin` / `obs_sceneitem_defer_update_end` here since the operations are read-only.

gxalpha: I thought it might be safer, especially since the original implementation also had it, but on a second thought you're probably correct.

PR #3360, opened 2020-08-30
frontend-tools: Auto detect Python on Windows

cg2121:
### Description
This automatically detects Python on Windows. If Python is
not installed, a message box will tell the users to install
Python and restart OBS, when they open the scripts dialog

### Motivation and Context
I found it annoying to find the Python path manually.

### How Has This Been Tested?
Needs more testing,  as I couldn't get COMPILE_PYTHON flag to work in cmake.

### Types of changes
- New feature (non-breaking change which adds functionality) -->
- Code cleanup (non-breaking change which makes code smaller or more readable)

### Checklist:
- [x] My code has been run through [clang-format](https://github.com/obsproject/obs-studio/blob/master/.clang-format).
- [x] I have read the [**contributing** document](https://github.com/obsproject/obs-studio/blob/master/CONTRIBUTING.rst).
- [x] My code is not on the master branch.
- [x] The code has been tested.
- [x] All commit messages are properly formatted and commits squashed where appropriate.
- [x] I have included updates to all appropriate documentation.

PR #5738, opened 2022-01-05
libobs: Add Windows X64 Emulation on ARM64 and move Rosetta Detection

tommyvct:
### Description

Adds a check that detects if OBS is running on Windows ARM64 emulation and logs the result.

### Motivation and Context

https://github.com/obsproject/obs-studio/pull/5644

### How Has This Been Tested?

Running X64 OBS Studio gives the following log:

#### Windows:

<details>

```
23:10:09.304: CPU Name: Microsoft SQ1 @ 3.0 GHz
23:10:09.305: CPU Speed: 1670MHz
23:10:09.305: Physical Cores: 8, Logical Cores: 8
23:10:09.305: Physical Memory: 7718MB Total, 3202MB Free
23:10:09.305: Windows Version: 10.0 Build 22538 (release: 2009; revision: 1010; ARM 64-bit)
23:10:09.305: Windows ARM64: Running with X64 Emulation
23:10:09.306: Running as administrator: false
23:10:09.307: Aero is Enabled (Aero is always on for windows 8 and above)
23:10:09.307: Windows 10 Gaming Features:
23:10:09.307: 	Game Bar: On
23:10:09.307: 	Game DVR: On
23:10:09.307: 	Game DVR Background Recording: Off
23:10:09.317: Current Date/Time: 2022-01-25, 23:10:09
23:10:09.317: Browser Hardware Acceleration: false
23:10:09.317: Qt Version: 5.15.2 (runtime), 5.15.2 (compiled)
23:10:09.317: Portable mode: false
23:10:10.338: OBS 27.2.0-beta4-23-g026987a2b (64-bit, windows)
```

</details>

#### macOS:

<details>

```
17:58:05.833: CPU Name: Apple M1 Pro
17:58:05.833: CPU Speed: 2400MHz
17:58:05.833: Physical Cores: 10, Logical Cores: 10
17:58:05.833: Physical Memory: 16384MB Total
17:58:05.833: OS Name: macOS
17:58:05.833: OS Version: Version 12.1 (Build 21C52)
17:58:05.833: Rosetta translation used: true
17:58:05.833: Kernel Version: 21.2.0
17:58:05.842: hotkeys-cocoa: Getting keyboard keys failed
17:58:05.842: hotkeys-cocoa: Using layout 'com.apple.keylayout.Dvorak'
17:58:05.843: Current Date/Time: 2022-01-26, 17:58:05
17:58:05.843: Browser Hardware Acceleration: true
17:58:05.843: Qt Version: 5.15.2 (runtime), 5.15.2 (compiled)
17:58:05.843: Portable mode: false
17:58:05.965: OBS 27.2.0-beta4-23-g026987a2b (mac)
```

</details>

### Types of changes

### Checklist:

### DIFF
--- a/UI/window-basic-main-outputs.cpp
+++ b/UI/window-basic-main-outputs.cpp
@@ -2043,8 +2043,7 @@ bool AdvancedOutput::StartReplayBuffer()
 			error_reason = QT_UTF8(error);
 		else
 			error_reason = QTStr("Output.StartFailedGeneric");
-		QMessageBox::critical(main,
-				      QTStr("Output.StartRecordingFailed"),
+		QMessageBox::critical(main, QTStr("Output.StartReplayFailed"),
 				      error_reason);
 		return false;
 	}


WizardCM:
To be clear, this will print if the user is running the x86_64 binary on Windows ARM64, yes?

tommyvct:
> To be clear, this will print if the user is running the x86_64 binary on Windows ARM64, yes?

Yes.

UI/obs-app.cpp
```diff
@@ -2122,6 +2122,11 @@ static int run_program(fstream &logFile, int argc, char *argv[])
 		     rosettaTranslated ? "true" : "false");
 #endif
 
+#if defined(_WIN32)
+		if (IsARM64Emulated())
+			blog(LOG_INFO, "Windows ARM64 Emulation used: true");
```
WizardCM, line 2127: I'm not sold on this wording as it's a bit confusing to someone who doesn't know exactly what it means in advance, but I can't think of something better. Microsoft labels this as "x64 emulation on Arm" and is apparently limited to Windows 11 going forward.

tommyvct: How about `IsEmulatedOnARM64Windows()`?

Name aside, the problem is that line of log doesn't show up. I'll try to fix that

WizardCM: For the function itself, `IsEmulatedOnARM64()` does sound better. The log message I'm unsure about.

UI/obs-app.cpp
```diff
@@ -2122,6 +2122,22 @@ static int run_program(fstream &logFile, int argc, char *argv[])
...
 #endif
 
+#if defined(_WIN32)
+		if (IsEmulatedOnARM64Windows()) {
+#if defined(_WIN64)
+			blog(LOG_INFO,
+			     "Windows ARM64: Running with X64 Emulation");
+#else
+			blog(LOG_INFO,
+			     "Windows ARM64: Running with X86 Emulation");
```
WizardCM, line 2132: I don't believe ARM64 has x86 emulation, and by the time we have a Windows ARM build, we won't support x86 OBS anyway (it exists mostly for legacy reasons, to support 32-bit VSTs and things). I'd be OK with ignoring that use case. Otherwise I like this wording much better.

tommyvct: Believe or not, when Surface Pro X launched, Windows ARM64 only comes with x86 emulation. Until sometime last year in Windows Insider Dev branch, Windows on ARM64 gained X64 emulation. X64 emulation should come with now dead Windows 10 Sun Vally, later the Sun Vally become the Windows 11, and X64 emulation is a Windows 11 exclusive feature now.

And you are right, we don't need x86 anymore if there's X64.

tommyvct:
This should be merged along with https://github.com/obsproject/loganalyzer/pull/91 .

WizardCM (changes requested):
I did some digging yesterday, and I feel this'd make more sense as part of the below function in `libobs`. `log_emulation_status()` or something.

https://github.com/obsproject/obs-studio/blob/ae4797f8cc690c0c94781767b9b1de4a29d3b4aa/libobs/obs-windows.c#L390-L405

gxalpha:
If we ever want to add a dialog like originally proposed in #5644, this needs some kind of return value to the UI. I thought that’s something we still wanted once the native versions are released officially, since we don’t want to do universal packages, right?

tommyvct:
You are right, I forgot the need to do the dialogue, in that case we do need something back from that function.

tommyvct:
How about we take the detection logic out of `log_emulation_status()`, put it into a function like `get_emulation_status()`? This way we can avoid duplicating things. 

But we also need to figure out where to put `get_emulation_status()` and should it be renamed to something else.

gxalpha:
Libobs' util/platform.h header sounds like the right place to me (though I'm not really sure). Perhaps something like `bool os_is_emulated()` or `bool os_get_emulation_status()` (keep in mind this has to be implemented by every OS, but everything besides macOS and Windows could just return false)

libobs/obs-cocoa.m
```diff
@@ -138,6 +138,14 @@ static void log_processor_cores(void)
 	     os_get_physical_cores(), os_get_logical_cores());
 }
 
+static void log_emulation_status(void)
+{
+#ifndef __aarch64__
```
gxalpha, line 143: What’s the reason for this ifdef? I think this should always be logged, no matter the build arch.

libobs/obs-windows.c
```diff
@@ -107,6 +107,18 @@ static void log_processor_cores(void)
 	     os_get_physical_cores(), os_get_logical_cores());
 }
 
+static void log_emulation_status(void)
+{
+#if defined(_M_ARM64EC)
+	blog(LOG_INFO, "Windows ARM64: Compatiable with X64 plugins");
+	return;
+#else
+	if (os_get_emulation_status()) {
```
gxalpha, line 116: See above, I feel it should always be logged wether OBS is being emulated/translated.

libobs/util/platform-windows.c
```diff
@@ -975,6 +975,24 @@ bool is_64_bit_windows(void)
...
+	auto result = IsWow64Process2(GetCurrentProcess(), &_, &nativeMachine);
+	return (result && (nativeMachine == IMAGE_FILE_MACHINE_ARM64));
+}
+
+bool os_get_emulation_status()
+{
+#if defined(_M_ARM64) || defined(_M_ARM64EC)
+	return false;
+#else
+	if (is_arm64_windows()) {
```
gxalpha, line 990: Is there a reason for this being a separate function? Doesn’t *really* matter, but could probably be inlined.

gxalpha (changes requested):
Small nitpicks. Imo the logging should always happen, not just on some systems - it could get rather confusing looking at the log otherwise.
Rest looks good to me.

libobs/util/platform-windows.c
tommyvct: I got your point, I was equally confused when I wrote this. 

If you look a few lines above, you'll see a function `is_64_bit_windows()` which have the similar purpose. 

These 2 functions serve 2 different purpose but can have exactly the same implementation.

libobs/obs-cocoa.m
tommyvct: I don't think it makes sense to log Rosetta or anything related to emulation when emulation is guaranteed not gonna happen. Both Rosetta and Windows X64 emulation works one way and one way only, and both companies are currently having no plan to make it work both ways. If we are ARM64 binaries already, then there is really no need to log it.

If we want to know are we ARM64 binaries or not, then we should do it somewhere else that is not here.

libobs/obs-windows.c
```diff
@@ -137,10 +149,13 @@ static void log_windows_version(void)
 	bool b64 = is_64_bit_windows();
 	const char *windows_bitness = b64 ? "64" : "32";
 
+	bool arm64 = is_arm64_windows();
+	const char *arm64_windows = arm64 ? "ARM " : "";
```
tommyvct, line 153: @gxalpha  This is what I talked about, we should declare whether we are ARM64 or not here.

libobs/util/platform-windows.c
```diff
@@ -975,6 +975,26 @@ bool is_64_bit_windows(void)
...
+}
+
+bool os_get_emulation_status()
+{
+#if defined(_M_ARM64) || defined(_M_ARM64EC)
+	return false;
+#else
+	return is_arm64_windows();
+#endif
+}
```
tommyvct, line 996: @gxalpha Actually there is a difference. 

For `is_arm64_windows(void)`, if we want to know if we are running on ARM64 Windows, and we are ARM64 binary, then return true because only ARM64 Windows can run ARM64 binary. 

For `os_get_emulation_status()`, if we want to know if we are being emulated, and we are ARM64 binary, then return false because emulation only works one way as I said.

libobs/obs-cocoa.m
gxalpha: In my view it very much does make sense to always have that log line - logs about system information should just always be there so that issues (especially on Mac where you can't mix translated and native) can be detected at first glance, and you don't have to check somewhere else on why the log line isn't where you expected (or isn't there at all).

tommyvct: I suggest that we can do it in `libobs/obs-cocoa.m`, either in `log_os()` or `log_kernel_version()`. I think we should detect the host machine in this way, just like what I did with Windows in `libobs/obs-windows.c`.

libobs/util/windows/win-version.h
```diff
@@ -48,6 +48,7 @@ static inline int win_version_compare(const struct win_version_info *dst,
 }
 
 EXPORT bool is_64_bit_windows(void);
+EXPORT bool is_arm64_windows(void);
```
gxalpha, line 51: Is this function used anywhere outside platform-windows.c or intended to be used? Or could the definition here get removed?

gxalpha (commented):
Assuming this works as it should (which I can't test right now, I can only look at the code), this mostly looks good to me. Only this one question left.

libobs/util/windows/win-version.h
WizardCM: It could get used in something like Python scripting #3360

gxalpha: Oh yeah that makes sense.
Now where did GH put the *Mark Resolved* button? 🤔

gxalpha (approved):

libobs/obs-cocoa.m
derrod: Going back to the original review here, I also agree with @gxalpha that emulation status should always be logged regardless of compiled architecture.

derrod:
This may not yet be relevant (although I suspect *somebody* will inevitably try it), but can we also detect https://github.com/ptitSeb/box86 emulation on *nix?

tommyvct:
We can, it's as simple as adding the implementation to `bool os_get_emulation_status(void)`in `libobs/util/platform-nix.c`.

You can propose another PR with the said implementation either with my forked repo or here in the upstream.

RytoEX (changes requested):
@gxalpha I see there have been changes since you last reviewed this. Could you re-review this?

tommyvct:
Thanks for proofreading, but I can't really do much about it right now, sorry for that. I'm extremely busy with schoolwork these days. I'll be back approximately next month.

If I remember correctly, except the wording and typo issue, this PR is logically ready to go.

libobs/obs-cocoa.m
```diff
@@ -156,7 +163,7 @@ static void log_available_memory(void)
 static void log_os(void)
 {
 	NSProcessInfo *pi = [NSProcessInfo processInfo];
-	blog(LOG_INFO, "OS Name: Mac OS X");
+	blog(LOG_INFO, "OS Name: macOS");
```
gxalpha, line 166: this change is unrelated, but I guess it's fine since by now most OS commands also report macOS

libobs/obs-windows.c
gxalpha, line 116: I'm fine with the initial ifdef above, but still think this here should still be logged in any case, like on macOS - not just in an `#else` and `if` (or was there other conversation about this that I didn't see? This code is getting full of comments :D)

docs/sphinx/reference-libobs-util-platform.rst
```diff
@@ -478,3 +478,12 @@ Other Functions
 .. function:: uint64_t os_get_proc_virtual_size(void)
 
    Returns the virtual memory size of the current process.
+
+---------------------
+
+.. function:: bool os_get_emulation_status(void)
+
+   Returns true if the current process is a x64 binary and is being emulated or translated
```
RytoEX, line 486: "is a x64 binary" -> "is an x64 binary"
here and throughout this text

libobs/obs-windows.c
```diff
@@ -107,6 +107,18 @@ static void log_processor_cores(void)
 	     os_get_physical_cores(), os_get_logical_cores());
 }
 
+static void log_emulation_status(void)
+{
+#if defined(_M_ARM64EC)
+	blog(LOG_INFO, "Windows ARM64: Compatible with x64 plugins");
```
RytoEX, line 113: I don't understand this sentence. What is it communicating?

RytoEX (changes requested):

RytoEX:
> @gxalpha did the original Rosetta detection, but it was implemented in the UI layer instead of `libobs`.
> 
> They asked me to do it only for macOS, and that's what I did in #6622. In fact, #6622 is stripped from this PR. Because of this. merge conflicts are normal and expected.
> 
> One of the commit in #6622 reverts the original detection and the other commit re-implemented the Rosetta detection.
> 
> ~#6622 have some minor importance changes (e.g. ""an X64 binary") which is not yet backported to this PR. I'll do that once I got a chance.~ It's already done.

The first commits in each both modify:
* docs/sphinx/reference-libobs-util-platform.rst
* libobs/util/platform-cocoa.m
* libobs/util/platform-nix.c
* libobs/util/platform.h

with nearly identical changes. The main differences are:
* libobs/util/platform-windows.c
* the commit message

That's what I mean.

tommyvct:
Git submodules are nasty. 

I'll fix this tomorrow.

tommyvct:
Okay, now it have 6 commits. The top 5 commits are related to logs, especially I did the log in `libobs`, not UI layer. 

We'll see how it goes.

libobs/obs-windows.c
RytoEX: Why was this marked as resolved? Nothing has been clarified regarding the logged string or the line of code.

RytoEX (changes requested):
Commits 2 and 3 are unrelated to the scope of this PR and could (and _should_) simply be submitted in a separate code cleanup PR. Please keep this PR strictly scoped to logging emulation statuses.

tommyvct:
#6791 seperated the 2 commits suggested by @RytoEX.

RytoEX:
With #6791 merged, this will need to be rebased.

libobs/obs-windows.c
driver1998: ARM64EC is a new ARM64 ABI in Windows that is compatible with the x64 emulator and can load x64 libraries.

In OBS, that would be a special OBS build which is native by itself but can some of its plugins still being x64.

Note ARM64 and ARM64EC are not inter-compatible but ARM64EC and x64 are.

I don't think ARM64 would be that useful in the context of OBS though.

PR #5747, opened 2022-01-06
frontend-tools: Correctly describe ARM Python path for scripting

WizardCM:
### Description

If OBS is built natively for ARM, currently it'll display "64bit" in the "Python Install Path" heading. This is incorrect, and will fail to load due to an incompatible architecture.

`__aarch64__` is used on Linux & macOS, `_M_ARM64` is used on Windows.

### Motivation and Context

As we don't show UI warnings about architecture incompatibility, we need to be clearer about which version of Python OBS needs to be pointed to.

### How Has This Been Tested?

* Built on M1 macOS using #5155
* Opened Scripts dialog
* Select Python tab

### Types of changes
### Checklist:


UI/frontend-plugins/frontend-tools/data/locale/en-US.ini
```diff
@@ -38,6 +38,7 @@ OpenFileLocation="Open file location"
 PythonSettings="Python Settings"
 PythonSettings.PythonInstallPath32bit="Python Install Path (32bit)"
 PythonSettings.PythonInstallPath64bit="Python Install Path (64bit)"
+PythonSettings.PythonInstallPathARM64="Python Install Path (ARM64)"
```
WizardCM, line 41: I'm not 100% sure on displaying these as 64bit and ARM64 (as Apple shows it as "Intel" and "Apple"/"Native" for example). Thoughts @PatTheMav?

PatTheMav: I‘d drop the suffix entirely, because as you mentioned specifying 64-bit doesn’t make sense on macOS (there is no 32-bit anymore) and M1 is either ‘Apple Silicon’ or just ‘Apple’ vs. ‘Intel’.

But with Universal libraries being a thing even that doesn’t make sense as those can be used for both.

Unfortunately due to the way the scripting plug-in is designed, it will be compiled with the Python root directory inside a Python Framework found on the build machine, which requires expert knowledge to customise this path from any macOS user (as the plug-in is very picky about the _exact_ path).

With Homebrew using a different path on M1 Macs vs Intel we should focus on automatic detection at runtime instead.

WizardCM: Yeah I did some research on automatic detection too. I think the only option there would be to provide a set of predefined search paths, then let the user choose one of them (if they exist)? Should be pretty easy for our scripting plugin to know which Python version it was built with & only list compatible entries. On Windows we'd have to check the registry, but it's also doable there as folders are based on version number.

PatTheMav: At least on macOS the following locations of a Python framework need to be considered:

* `/System/Library/Frameworks/Python.framework` (Python 2.7 only)
* `/Applications/Xcode.app/Contents/Developer/Library/Frameworks/Python3.framework`
* `/usr/local/Cellar/python@<MAJOR.MINOR>/<MAJOR.MINOR.PATCH>/Frameworks/Python.framework` (Homebrew Python on Intel-based Macs)
* `/opt/homebrew/Cellar/python@<MAJOR.MINOR>/<MAJOR.MINOR.PATCH>/Frameworks/Python.framework` (Homebrew Python on Apple Silicon-based Macs)

Several Python3 versions can be installed side by side as some Homebrew formulas require the most recent Py3 while others still use Py3.8 or Py3.9 because of API differences.

And on top of this, the scripting plugin needs to be pointed at the non-symlinked(!) root path to the specific version inside the Framework, so the actual "Python Install Path" would be `/usr/local/Cellar/python@3.10/3.10.1/Frameworks/Python.framework/Versions/3.10/` (because it doesn't "understand" that the Python library is a binary called `Python` at the root of the Framework bundle.

All in all the way the scripting plugin loads Python needs to be overhauled for macOS anyway. 😬

WizardCM: Similar idea on Windows, but you would search in the registry:

All users:
```
HKEY_LOCAL_MACHINE\SOFTWARE\Python\PythonCore\<MAJOR.MINOR>\InstallPath
```

Current User:
```
HKEY_CURRENT_USER\SOFTWARE\Python\PythonCore\<MAJOR.MINOR>\InstallPath
```
Which is basically what #3360 does but it provides no way to choose which installation to use, nor a way to specify a custom path. I'll close my PR in favour of getting that one closer to a mergeable state.

PR #5752, opened 2022-01-08
UI: Disable auto-reconnects for srt  when there's no host

pkviet:
### Description
When streaming with srt protocol, if the server is down and if auto-reconnect is enabled
(in Settings > Advanced), the stream hangs forever and obs must be killed.
This PR disables auto-reconnect for srt URLs when libsrt complains there's no host so that obs doesn't hang any more.
If there's a host, the auto-reconnect capability is preserved for srt.
In order to achieve that, librist log message "Failed to resolve hostname"
(https://github.com/FFmpeg/FFmpeg/blob/master/libavformat/libsrt.c#L423) is transmitted to obs-ffmpeg-mux
which then signals OBS_OUTPUT_CONNECT_FAILED instead of OBS_OUTPUT_DISCONNECTED

### Motivation and Context
Bug reported in beta-testing channel of obs discord server by TheBigG.
`Using srt for streaming I noticed that if you get disconnected from your streaming ingest and you want to stop the stream, OBS will be stuck at "Attempting to reconnect" for ever and you have to kill it with taskmanager.`

NB: bug confirmed by @Gol-D-Ace and myself.

### How Has This Been Tested?
Tested against brime.tv Frankfurt srt server which was down at the time of testing.
Obs no longer hangs and disconnects gracefully.

### Types of changes

### Checklist:

### DIFF
--- a/plugins/obs-ffmpeg/ffmpeg-mux/ffmpeg-mux.c
+++ b/plugins/obs-ffmpeg/ffmpeg-mux/ffmpeg-mux.c
@@ -231,7 +231,6 @@ static bool get_audio_params(struct audio_params *audio, int *argc,
 static void ffmpeg_log_callback(void *param, int level, const char *format,
 				va_list args)
 {
-#ifdef DEBUG_FFMPEG
 	char out_buffer[4096];
 	struct dstr out = {0};
 
@@ -240,7 +239,7 @@ static void ffmpeg_log_callback(void *param, int level, const char *format,
 	if (global_stream_key && *global_stream_key) {
 		dstr_replace(&out, global_stream_key, "{stream_key}");
 	}
-
+#ifdef DEBUG_FFMPEG
 	switch (level) {
 	case AV_LOG_INFO:
 		fprintf(stdout, "info: [ffmpeg_muxer] %s", out.array);
@@ -258,13 +257,13 @@ static void ffmpeg_log_callback(void *param, int level, const char *format,
 			out.array, ANSI_COLOR_RESET);
 		fflush(stderr);
 	}
-
-	dstr_free(&out);
 #else
-	UNUSED_PARAMETER(level);
-	UNUSED_PARAMETER(format);
-	UNUSED_PARAMETER(args);
+	if (level == AV_LOG_ERROR) {
+		fprintf(stderr, "%s", out.array);
+		fflush(stderr);
+	}
 #endif
+	dstr_free(&out);
 	UNUSED_PARAMETER(param);
 }
 
--- a/plugins/obs-ffmpeg/obs-ffmpeg-mux.c
+++ b/plugins/obs-ffmpeg/obs-ffmpeg-mux.c
@@ -443,13 +443,19 @@ static void signal_failure(struct ffmpeg_muxer *stream)
 
 	ret = deactivate(stream, 0);
 
+#define SRT_HOST_ERROR "Failed to resolve hostname"
+
 	switch (ret) {
 	case FFM_UNSUPPORTED:
 		code = OBS_OUTPUT_UNSUPPORTED;
 		break;
 	default:
 		if (stream->is_network) {
-			code = OBS_OUTPUT_DISCONNECTED;
+			char *result = strstr(error, SRT_HOST_ERROR);
+			if (result != NULL)
+				code = OBS_OUTPUT_CONNECT_FAILED;
+			else
+				code = OBS_OUTPUT_DISCONNECTED;
 		} else {
 			code = OBS_OUTPUT_ENCODE_ERROR;
 		}


UI/window-basic-main-outputs.cpp
```diff
@@ -870,6 +870,7 @@ void SimpleOutput::SetupVodTrack(obs_service_t *service)
 		clear_archive_encoder(streamOutput, SIMPLE_ARCHIVE_NAME);
 }
 
+#define SRT_PROTOCOL "srt"
```
notr1ch, line 873: Can this be moved to the top of the file along with the other protocol definitions?

https://github.com/obsproject/obs-studio/blob/92d1e5a8ccace361bf5908d4c29062e44bd99600/UI/window-basic-main-outputs.cpp#L19-L20

pkviet: done

eric:
Please don't commit this without further investigation. I have had many experiences with SRT reconnecting successfully.

In my experience if the `timeout=`/`listen_timeout=` parameter is used with the SRT URL it will not hang.

pkviet:
Adding a timeout param doesn't solve the issue.
@eric in the case at hand, there's no server at all due to a default on the service side. So this differs from your experience where there's an existing server.

Also found out that the reconnect thread hangs forever because for some reason the srt reconnection is flagged as successful at first. This resets the reconnection counter to 0; later on, a disconnection is logged and one restarts the reconnect loop ...

So something's definitely rotten.

eric:
Now that your describe this issue, I recall a problem I ran into with auto-reconnect when Twitch reset all of the Stream Keys. When I would try to connect with the old (bad) Stream Key OBS would hang forever and had to be force quit. I never had an opportunity to investigate what was causing that hang, but I wonder if this is all an issue with how auto-reconnection works around initial connection failures.

Needing to tell my remote contributors to have to watch their OBS and manually hit connect if there are any connection failures would be a big step back for me, so I would really appreciate if we could get to the bottom of this connection issue instead of removing the feature.

pkviet:
Update: 
- propagated libsrt error message to obs-ffmpeg-mux so that one knows that there's no host and it's not a plain disconnection issue.
- disabled auto-reconnect if srt log shows no host;
- otherwise, auto-reconnect is preserved for srt streams.

plugins/obs-ffmpeg/ffmpeg-mux/ffmpeg-mux.c
```diff
@@ -258,13 +257,13 @@ static void ffmpeg_log_callback(void *param, int level, const char *format,
...
-
-	dstr_free(&out);
 #else
-	UNUSED_PARAMETER(level);
-	UNUSED_PARAMETER(format);
-	UNUSED_PARAMETER(args);
+	if (level == AV_LOG_ERROR) {
+		fprintf(stderr, "%s", out.array);
+		fflush(stderr);
+	}
```
notr1ch, line 264: If ffmpeg logs `AV_LOG_ERROR` for non-fatal errors then our `stderr` will fill up and block since OBS doesn't read it unless there's a fatal error. Are we sure all paths that his this are fatal errors?

plugins/obs-ffmpeg/obs-ffmpeg-mux.c
```diff
@@ -443,13 +443,19 @@ static void signal_failure(struct ffmpeg_muxer *stream)
...
+#define SRT_HOST_ERROR "Failed to resolve hostname"
+
 	switch (ret) {
 	case FFM_UNSUPPORTED:
 		code = OBS_OUTPUT_UNSUPPORTED;
 		break;
 	default:
 		if (stream->is_network) {
-			code = OBS_OUTPUT_DISCONNECTED;
+			char *result = strstr(error, SRT_HOST_ERROR);
```
notr1ch, line 454: I'm also not sure this is a great solution to the problem, for example if someone's internet goes down briefly then their DNS server will likely be unreachable, causing a "Failed to resolve hostname" error even though it is a situation that could be recovered from by trying again later.

plugins/obs-ffmpeg/ffmpeg-mux/ffmpeg-mux.c
pkviet: right; i'll put a condition to identify that this is the log entry targeted from librist; if it is i'll pass to sdterr; if not, I won't.

plugins/obs-ffmpeg/obs-ffmpeg-mux.c
pkviet: damn; you're our network guru so i'll defer to you.

pkviet:
Closing because this is not a satisfactory solution; see @notr1ch comment about dns.
People encountering this issue will have to disable manually auto-reconnect.

PR #5754, opened 2022-01-08
Fix window flashing when starting minimized to tray

### LINKED ISSUE #2668: [BUG] --minimize-to-tray launch option flashes the main window at startup
## Platform

Operating system and version: Windows 10 64-bit
OBS Studio version: 25.0.4

## Expected Behavior
When launching OBS with the `--minimize-to-tray` launch option, it should run silently without any window popups.

## Current Behavior
OBS seems to flash its main window for a moment before minimizing it, briefly stealing focus from the currently active application.

## Steps to Reproduce

1. Add ``--minimize-to-tray`` to your shortcut's launch options.
2. Run OBS with this shortcut.

## Additional information
After doing this, there seems to be a rare case where running OBS **without** the launch option might not create a visible window until you perform two single clicks on the system tray icon, minimizing and un-minimizing it.

CodyJung:
### Description

Moves the processing to hide the window earlier in the startup process - rather than showing the main window and hiding it in the tray initialization, just _don't_ show the main window. This prevents the window from briefly flashing and taking focus from other apps.

This requires a new check for `sysTrayEnabled` - we need to know whether the tray icon will be visible. (If it's not, and the window is hidden, there would be no way to maximize the window.)

Lastly, this PR removes the Apple-specific workarounds, as they no longer appear to be necessary.

### Motivation and Context

Fixes #2668 and prevents the main window flashing on startup.

### How Has This Been Tested?

_Setting_
1. Enable **System Tray** and **Minimize to system tray when started**.
2. Restart OBS.
2. Verify tray icon appears and window does not flash.
3. Disable **System Tray** and restart OBS.
4. Verify window opens on start.

_Launch Flag_
1. Enable **System Tray**
2. Start OBS with `--minimze-to-tray` option.
3. Verify tray icon appears and window does not flash.
4. Disable **System Tray**
6. Start OBS with `--minimze-to-tray` option.
7. Verify window opens on start (you cannot minimize to tray if no tray icon)

Verified on Windows 10 x64 (19043.1415) and MacOS Big Sur (11.6.2).

### Types of changes

### Checklist:

### DIFF
--- a/UI/window-basic-main.cpp
+++ b/UI/window-basic-main.cpp
@@ -1981,9 +1981,7 @@ void OBSBasic::OBSInit()
 	ui->lockUI->setChecked(docksLocked);
 	ui->lockUI->blockSignals(false);
 
-#ifndef __APPLE__
 	SystemTray(true);
-#endif
 
 #if defined(_WIN32) && QT_VERSION < QT_VERSION_CHECK(6, 0, 0)
 	taskBtn->setWindow(windowHandle());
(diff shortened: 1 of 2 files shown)


WizardCM (approved):
Tested this on Windows 10, looks good to me.

PR #5759, opened 2022-01-08
UI: Stop locking filter mutex while loading properties

exeldro:
### Description
Reordering the filters makes the filter mutex of the source locked while loading a new the properties of a filter which may require other locks. By loading the properties of the filter on a QueuedConnection we make sure the filter mutex is cleared when that happens.

### Motivation and Context
I don't like deadlocks.
The main UI thread gets a lock on the filter mutex followed by a lock on the app sources mutex to fill the properties of the filter
while the graphics thread does a lock on the app sources mutex followed by a lock on the filter mutex

### How Has This Been Tested?
On windows 64 bit by reordering move transition override filters

### Types of changes

### Checklist:


WizardCM:
Is this a new issue in the beta, or slightly more longer-standing?

PR #5760, opened 2022-01-08
libobs,docs: Allow to enforce unique name and value in list property

tytan652:
### Description

Add to the Properties API the possibility to not allow adding duplicate by name and/or by value in lists with `obs_property_list_set_enforce_unique_name` and `obs_property_list_set_enforce_unique_val`.

If duplicates are not allowed, adding/inserting functions will not add data to the `DARRAY` if duplicate.

Duplicates are allowed by default to not break how the Properties API actually behave.

#### My use case
I'm working on my RFCs and I actually get into a situation where I have a list of items (server with their protocol) where I get duplicates (protocol when listed) that I don't want to have inside my list (of protocol in my case).

So rather implementing it in my properties function, I thought that adding it to the Properties API could be a nice addition since my use case could not be the only one to have a use of it.

### Motivation and Context

Rather then forcing developer to check if they are inserting an item that is already inside the property by reinventing the wheel, why not add the check inside the libobs itself.

### How Has This Been Tested?

Adding this to a properties function and each of this list should have only one item.
```c
p = obs_properties_add_list(props, "test_str_dup", "Test name duplicate string", OBS_COMBO_TYPE_LIST, OBS_COMBO_FORMAT_STRING);
obs_property_list_set_enforce_unique_name(p, true);
obs_property_list_add_string(p, "TEST", "test");
obs_property_list_add_string(p, "TEST", "test2");

p = obs_properties_add_list(props, "test_int_dup", "Test name duplicate integer", OBS_COMBO_TYPE_LIST, OBS_COMBO_FORMAT_INT);
obs_property_list_set_enforce_unique_name(p, true);
obs_property_list_add_int(p, "TEST", 1);
obs_property_list_add_int(p, "TEST", 2);

p = obs_properties_add_list(props, "test_val_dup", "Test value duplicate integrer", OBS_COMBO_TYPE_LIST, OBS_COMBO_FORMAT_INT);
obs_property_list_set_enforce_unique_value(p, true);
obs_property_list_add_int(p, "TEST", 1);
obs_property_list_add_int(p, "TEST 2", 1);
```

### Types of changes

### Checklist:

### DIFF
--- a/docs/sphinx/reference-properties.rst
+++ b/docs/sphinx/reference-properties.rst
@@ -186,7 +186,8 @@ Property Object Functions
 .. function:: obs_property_t *obs_properties_add_list(obs_properties_t *props, const char *name, const char *description, enum obs_combo_type type, enum obs_combo_format format)
 
    Adds an integer/string/floating point item list.  This would be
-   implemented as a combo box in user interface.
+   implemented as a combo box in user interface. Duplicated item are
+   allowed by default.
 
    :param    name:        Setting identifier string
    :param    description: Localized name shown to user
@@ -215,6 +216,8 @@ Property Object Functions
    - :c:func:`obs_property_list_insert_float`
    - :c:func:`obs_property_list_item_remove`
    - :c:func:`obs_property_list_clear`
+   - :c:func:`obs_property_list_set_enforce_unique_name`
+   - :c:func:`obs_property_list_set_enforce_unique_value`
 
 ---------------------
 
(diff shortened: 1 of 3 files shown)


tytan652:
Put as draft, I really want to make a `allow_duplicate` and not `show_duplicate`, I will try again tomorrow.

WizardCM:
What is the use case for this?

tytan652:
I finally made it after some rest, my PR now prevent duplicates of being added rather than just not showing them.
I updated my PR description, if you think it's not worth to be merged so be it.

WizardCM:
Maybe rather than `duplicate` we could use the term `unique`?

tytan652:
> Maybe rather than `duplicate` we could use the term `unique`?

You mean rather than `allow_duplicate` something like `enforce_unique` ?

tytan652:
> You mean rather than `allow_duplicate` something like `enforce_unique` ?

WizardCM anwsered me yes off-thread, so I made the changes.

PR #5762, opened 2022-01-08
libobs: Open a separate X11 connection for hotkeys

### LINKED ISSUE #4843: Push-to-talk no longer functions correctly
### Operating System Info

Other

### Other OS

Solus

### OBS Studio Version

27.0.0

_No response_

### OBS Studio Log URL

https://obsproject.com/logs/6tCS6jlZTFHZFYg0

### OBS Studio Crash Log URL

_No response_

### Expected Behavior

In OBS Studio releases prior to 27.0, I could use various hotkeys such as my Right Control or Mouse 6 (DPI typically) for push-to-talk on the default Audio Input Capture (Pulseaudio).

### Current Behavior

In OBS Studio 27.0, I am no longer able to use push-to-talk and have had to temporarily switch to constant transmit with various filters like a limiter and noise gat.

### Steps to Reproduce

1. Open the OBS Studio Settings window
2. Under Audio, choose Enable Push-to-talk on Audio Input Capture (Pulseaudio)
3. Under Hotkeys, type "push" to filter irrelevant items, and press a relevant key such as Right Control or a mouse button
4. Click OK to apply the settings and close the window.
5. Click off the window, press relevant key / button, and notice missing audio input in any scene which has it as a source (you may need to add it if you have a fresh configuration, I am not sure).

This is under X11 and Pulseaudio, as opposed to Wayland and Pipewire. Package build instructions are at https://dev.getsol.us/source/obs-studio/browse/master/package.yml, as you can see we only apply one patch from related to CMake, otherwise it is stock OBS Studio.

kkartaltepe:
### Description

Qt will mask certain input events. In order to avoid them masking things
like mouse button events we open a new display here.

This ended up getting introduced in 506b950 but we didnt notice until later.

### Motivation and Context

Fixes #4843

### How Has This Been Tested?

Mouse 4 works when pressed while another window is active. This is the same as the behavior we had for 26.

### Types of changes

### Checklist:

### DIFF
--- a/libobs/obs-nix-x11.c
+++ b/libobs/obs-nix-x11.c
@@ -828,7 +828,8 @@ static inline void registerMouseEvents(struct obs_core_hotkeys *hotkeys)
 
 static bool obs_nix_x11_hotkeys_platform_init(struct obs_core_hotkeys *hotkeys)
 {
-	Display *display = obs_get_nix_platform_display();
+	// Open a new X11 connection here, this avoids Qt masking events we care about.
+	Display *display = XOpenDisplay(NULL);
 	if (!display)
 		return false;
 


GeorgesStavracas (approved):
Thanks, this makes sense. I tried to look into this issue in the past, but never could have imagined this would be the cause, so thank you a lot for digging into it and figuring this out.

PR #5769, opened 2022-01-10
win-capture: Ignore window that can't be enumerated by EnumWindows

walker-WSH:
### Description

About HWND list of window source, some minimized app is listed. 
However, when user selects this app, no video is rendered and no HWND can be matched. 

**Such as:** 
Calculate.exe, 
Groove Music

### Motivation and Context

minimized app should not be listed.
Is a window is listed, matched HWND should be found at least.

### How Has This Been Tested?

Bug steps:
1. open Groove music and make it be minimized.
2. add window capture, open property window, select Groove music.
3. No video is rendered. After debugging, we can find no HWND is found by _find_window_top_level()_

Reason: 
1. The listed window is gotten by FindWindow+NextWindow.
2. The matched window for WGC is gotten by EnumWindows.
I find that minimized app can be enumerated by FindWindow while can't be found by EnumWindows.

### Types of changes

### Checklist:

### DIFF
--- a/plugins/win-capture/window-helpers.c
+++ b/plugins/win-capture/window-helpers.c
@@ -347,18 +347,74 @@ static HWND first_window(enum window_search_mode mode, HWND *parent,
 	return window;
 }
 
+struct check_window_helper {
+	enum window_search_mode enum_mode;
+	DARRAY(HWND) window_list;
+};
+
+BOOL CALLBACK enum_windows_proc_test(HWND window, LPARAM lParam)
+{
+	struct check_window_helper *helper =
+		(struct check_window_helper *)lParam;
+
+	if (!check_window_valid(window, helper->enum_mode))
+		return TRUE;
+
+	int cloaked;
+	if (SUCCEEDED(DwmGetWindowAttribute(window, DWMWA_CLOAKED, &cloaked,
+					    sizeof(cloaked))) &&
+	    cloaked)
+		return TRUE;
+
+	struct dstr exe = {0};
+	if (!get_window_exe(&exe, window)) /* This is for HWND filter */
+		return TRUE;
+	dstr_free(&exe);
+
+	if (is_uwp_window(window)) {
+		HWND child = get_uwp_actual_window(window);
+		if (child)
+			window = child;
+	}
+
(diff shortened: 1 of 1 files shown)


walker-WSH:
I am not sure if there exists a better solution.
@jpark37

Fenrirthviti:
I'm not sure if I like completely hiding the window from the list. Is there a way we can add some kind of indicator, maybe something like `(Minimized, cannot capture)` to the list and grey it out for windows that are minimized? I understand that they might not render while captured, but should the window be un-minimized it should continue to capture properly unless I am misunderstanding something. Users will typically expect to see all windows they have visible in the task bar, so if some are missing with no indication of why that is, it leads to bad UX.

I'd also welcome @Warchamp7's thoughts from the UX standpoint here. Overall I do think this is a problem we should address, I don't think hiding them from the list is the right solution though.

walker-WSH:
> I'm not sure if I like completely hiding the window from the list. Is there a way we can add some kind of indicator, maybe something like `(Minimized, cannot capture)` to the list and grey it out for windows that are minimized? I understand that they might not render while captured, but should the window be un-minimized it should continue to capture properly unless I am misunderstanding something. Users will typically expect to see all windows they have visible in the task bar, so if some are missing with no indication of why that is, it leads to bad UX.
> 
> I'd also welcome @Warchamp7's thoughts from the UX standpoint here. Overall I do think this is a problem we should address, I don't think hiding them from the list is the right solution though.


At present, about UX of window capture, all minimized window won't be listed. 
Those code has been inlcuded for checking:
IsWindowVisible, IsIconic, RECT.

But these logic does not success to filter the app I said.
I was also searching any attribute for filter.

WizardCM:
@walker-WSH Please let me know if #5810 solves the issues you ran into.

walker-WSH:
> @walker-WSH Please let me know if #5810 solves the issues you ran into.



It seems we are fixing same issue. Your commit is just the very solution，I will test my case with your PR tomorrow

walker-WSH:
@WizardCM  
Test result: Your PR also fix this case.
I will close my PR

PR #5774, opened 2022-01-11
Add some features to NVENC codec settings in Advanced mode targeted t…

Lokomojo:
### Description
Add some advanced features to NVENC codec settings targeted towards recording / archival

- Add CRF support (the most efficient mode for recording. Unsuitable for streaming)
- Configurable lookahead depth
- Configurable AQ Strength
- Configurable B-Frame ref mode
- Configurable BRef Distance
- Removed unused enum RC_MODE

0bd0b6ab5b195b63edd576d299eded19

### Motivation and Context
These options permit improved encoding efficiency for recording purposes.

- CRF produces smaller files than CQP for the same visual quality.
- Lookahead depth - previously defaulted to 32, but for low-end systems a value of 8 is significantly faster yet significantly better than no lookahead at all
- AQ Strength - While it tries to be imperceptible via psychovisual modelling, AQ (Psycho visual tuning) is very lossy from the original image, so having this configurable is important for archival
- B-Frame Ref mode / BRef Distance - This option provides significant efficiency improvements

### How Has This Been Tested?
Tested recording / streaming with all possible combinations of options on Windows 10.

However, I only have an RTX3080 GPU, this needs to be tested on older GPUs.

### Types of changes

### Checklist:

- [ ] The code has been tested. **Needs more testing that I can't do.**


gxalpha:
Hey,
please make sure to name the commit according to the [commit guidelines](https://github.com/obsproject/obs-studio/blob/master/CONTRIBUTING.rst#commit-guidelines).

TomMettam:
Sorry, fixed.

flaeri:
Hello @TomMettam :)

I was a bit curious as to why the CRF value is a double/float, and not an integer?

Log output is missing from jim-nvenc for CRF (line 606)

Log output in ffmpeg gets output as 0, instead of proper value.
Perhaps a problem with CRF stored as float instead of INT I suspect.

Line 289 in obs-ffmpeg-nvenc.c is %d

Besides that, it seems to work okay on my GTX 1070 :)

TomMettam:
Hi @flaeri 

This value is supposed to be fractional

```
config->rcParams.targetQuality
config->rcParams.targetQualityLSB 
```

The LSB is the fractional part

I will fix the format string :)

TomMettam:
Fixed

TomMettam:
Fixed formatting as per CI

WizardCM:
So, how much of this **needs** to be configurable? We generally prefer to keep a "sane defaults" approach, even in Advanced output mode. I think adding CRF is fine, but the rest feel potentially unnecessary. The screenshot you provided feels noisy due to too many settings.

TomMettam:
Well, aside from the the fact that all of this is already tucked away under "Advanced" output settings..

* The "look-ahead" depth is hidden if look ahead is unchecked
* The reference distance is hidden if the b-frame mode is disabled
* AQ strength is hidden if Psycho Visual Tuning is unchecked

So actually there's no additional clutter unless these options are in-use (aside from the b-frame mode, which I would argue is needed)

Fenrirthviti:
Being in advanced mode I think this is *probably* ok, but I would prefer to have tooltips or the ? button next to each option to explain what they do. I can already hear the added support costs of users coming to us and asking what these new options do.

TomMettam:
@Fenrirthviti  I can do that. I've already added the ones i thought were most unclear, i'll add the rest.

Another possibility might be to add an "Expert" output mode if we're really worried about clutter.

Fenrirthviti:
Also, there was mention that we originally used CRF (which is actually just VBR with a target quality) but it was changed because it capped out at 128Mbps which was causing issues. Can you confirm if this is still the case?

TheB1gG:
> * Lookahead depth - previously defaulted to 32, but for low-end systems a value of 8 is significantly faster yet significantly better than no lookahead at all

Wouldn't it be useful to expose p1-p7 then too? That does also greatly affect the performance.

TomMettam:
@Fenrirthviti Can you elaborate on that..

Do you mean that the problem was that the bitrate couldn't go higher than 128Mbps, or that it was permitted to go that high?

TomMettam:
> 
> Wouldn't it be useful to expose p1-p7 then too? That does also greatly affect the performance.

I would like that, but we're using an out of date nvenc api (or rather, ffmpeg is). That needs to be updated before these presets can be used.

TheB1gG:
> > Wouldn't it be useful to expose p1-p7 then too? That does also greatly affect the performance.
> 
> I would like that, but we're using an out of date nvenc api (or rather, ffmpeg is). That needs to be updated before these presets can be used. This will also allow us to use `2Pass-Full`

Oh so you tested with 27.1.3?
27.2.0 betas use ffmpeg 4.4.1 which exposes that presets and also deprecates or breaks some like cbr_hq and vbr_hq.
Using CBR or VBR with -tune hq would be for ffmpeg 4.4.1 the correct way I think.

TomMettam:
I developed and tested against master, but `win64\include\ffnvcodec\nvEncodeAPI.h` in the pre-built dependencies package doesn't have the constants like NV_ENC_PRESET_P7_GUID.

Is there a newer package? Or do I have to build my own?

Fenrirthviti:
> @Fenrirthviti Can you elaborate on that..
> 
> Do you mean that the problem was that the bitrate couldn't go higher than 128Mbps, or that it was permitted to go that high?

My understanding is that it was not permitted to go higher.

flaeri:
> My understanding is that it was not permitted to go higher.

Interesting, I never knew, but it seems you are quite right.
crfnvenc (Medium)

Experienced a strange issue with `jim-nvenc` where it would seemingly wouldnt go past 3mbit bitrate, but after restarting I can no longer reproduce it. Seemed to be related to lookahead, but again, can no longer repro :\

-----

Just wanted to mention that the log now outputs the float, but its missing/ignoring what comes after the `.`
16.5 = crf: 16.000000

`jim-nvenc` should also get the same log output fix as ffmpeg-nvenc did :) Its completely missing at the moment.

I personally am not a big fan of having the users set a float, and floats being output to the log. Like you say, its in the implementation, but I've never set a float before when I've this rate control in ffmpeg for instance. I dont think that granular of control is meaningful.

How do you feel about the naming it CRF over CQ (or VBR-CQ)?  
Both FFMPEG and NVIDIA uses "CQ" for instance.
`-cq Set target quality level.. for constant quality mode in VBR rate control`

I'm a little bit worried people would expect the CQ/CRF scale in nvenc to be the same as it is in x264 (which is quite far apart imo).

flaeri:
> Is there a newer package? Or do I have to build my own?

Yeah, but it landed very recently
https://github.com/obsproject/obs-deps/pull/86

You could build them yourself, yeah. Not sure when the pre-built one will be updated.
https://github.com/obsproject/obs-deps

TomMettam:
Looks like it is bitrate limited but I think it's the vbv buffer size. Investigating

TomMettam:
* Fixed the constrained bitrate issue: https://gyazo.com/9aa4f46e7863150a9127e0e521115141
* Changed Quality to an int (I would prefer to have had it left as a float for fine-grained control, but i'm a mere contributor)
* I can't get the `obs_property_set_long_description` to work on these int fields..  help would be appreciated

flaeri:
crf-jim (Medium)
Testing and it seems like it does indeed solve it for jim-nvenc. 250mbit was my cap.

crf-ffmpeg (Medium)
However with the ffmpeg implementation it seems to still have the unwanted cap. 250mbit was my set cap for this as well.

I'm sorry if I came of a bit strong regarding the INT vs float. It's not for me to demand or decide, I was just of the opinion that some user confusion could be avoided, as OBS does INT for CRF for libx264 as well, even it can accept floats. They are used to specifying all the other quality based rate controls in whole numbers (cq on quicksync, CQP on AMF and nvenc etc). 

I also didnt see much value in having such granular control over the value, as I dont think users are going to feel the need for having a point in-between CRF/CQ 22 and 23 for instance.

Unfortunately I can't be of much help regarding `obs_property_set_long_description`, but hopefully others will :)
If you're not already on the discord, I think it would be a good place for that kind of question, plus discussions.
https://obsproject.com/discord

RytoEX:
> > Is there a newer package? Or do I have to build my own?
> 
> Yeah, but it landed very recently [obsproject/obs-deps#86](https://github.com/obsproject/obs-deps/pull/86)
> 
> You could build them yourself, yeah. Not sure when the pre-built one will be updated. https://github.com/obsproject/obs-deps

It is not necessary to build them.  They are already built and released on obs-deps.  See the [OBS Windows Deps build 2022-01-01 release](https://github.com/obsproject/obs-deps/releases/tag/win-2022-01-01).


> > Wouldn't it be useful to expose p1-p7 then too? That does also greatly affect the performance.
> 
> I would like that, but we're using an out of date nvenc api (or rather, ffmpeg is). That needs to be updated before these presets can be used. This will also allow us to use `2Pass-Full` (We're currently using vbr-hq which is deprecated and is not full resolution multipass)

If you add support for this, you will need to be careful to migrate the old presets to sensible new ones and keep the old ones around in case users downgrade.  You will have to support both upgrades and downgrades gracefully.

notr1ch:
In addition to exposing P1 - P7, I think we should also expose the tuning and two-pass config.

Be aware that moving to the new presets also means the simple output mode recording options need migrating to an equivalent CRF value to the current CQP values and sensible defaults used (eg not too much lookahead and b-frames, maybe 2-pass etc) to ensure concurrent usage of NVENC by other apps isn't impacted - eg from what I can tell, P7 is pretty intense and a user is unlikely to be able to have three concurrent encodes going.

I'd started a bit of work on this myself but hadn't got very far, it's great to see this PR and even though the scope is growing considerably, I think it will be a big upgrade for all NVENC users. I do have some concerns over exactly how much control we should offer (exposing b-ref modes and high lookahead for example is a great way to let users shoot themselves in the foot when using streaming output).

TheB1gG:
> I'd started a bit of work on this myself but hadn't got very far, it's great to see this PR and even though the scope is growing considerably, I think it will be a big upgrade for all NVENC users. I do have some concerns over exactly how much control we should offer (exposing b-ref modes and high lookahead for example is a great way to let users shoot themselves in the foot when using streaming output).

Tough question, but since different GPUs and scenarios would benefit from different settings I would prefer for myself a lot of controll.

For example a nvenc gen 6 benefits from a look ahead depth of up to 12, while gen 7 benefits from a depth of 32. Also gen 6 lacks support for multiple reference frames while gen 7 supports that.

nvenc with settings maxed out for quality (which still runs fine in realtime) vs the defaults used by the old jim-nvenc implementation is like jumping from x264 very fast to x264 medium/slowish. So having atleast the option to set stuff in some way would be important IMHO, so users who know what they are doing can set what they need.

TomMettam:
* Bitrate constraint in ffmpeg module are fixed
* Tooltips are implemented
* I'm going to defer the new preset implementation, it's going to be a breaking change and therefore will require significant testing, it doesn't belong here.

I am very much in favour of giving the user more control. If they've chosen the Advanced settings, we should give them the benefit of the doubt.  It's not going to shoot their granny.

derrod:
With the latest commit you seem to be limiting/setting the maximum bitrate in target quality mode to 288 Mbps, why?

TomMettam:
Because it's actually limited internally https://obsproject.com/forum/threads/nvenc-288000-max-bitrate-bug-or-limitation.50380/

TomMettam:
I've removed the 28800 max-bandwidth cap

However, OBS prevents values higher than 300,000 here, probably due to the same legacy reasons as me.

Does anyone know of any reason why this can't be raised?

derrod:
Just as a note: This really shouldn't be called CRF, because it isn't. CRF is an x264 thing/term and both Nvidia/ffmpeg refer to this as "Target Quality" or "Constant Quality" mode which is a subset of VBR. So really this should be "VBR CQ" if we want to keep using abbreviations.

TomMettam:
> Just as a note: This really shouldn't be called CRF, because it isn't. 

I agree in principal, however...

- Most users will be looking for the NVENC "equivalent" of common x264 settings
- CRF ≈ CQ + AQ
- CQ is confusingly similar to CQP

We already have "Psycho Visual Tuning" which.. i don't know where that name came from, but it's adaptive quantization. If we're going for accuracy, shall we change this too?

Fenrirthviti:
The consistency comes from how NVIDIA documents it, in their API and documentation. We shouldn't be looking to imply that modes are the same when they're not, and using the way other encoders refer to their own methods. NVENC should stay consistent to NVENC. They call it PVT, so we call it PVT.

TheB1gG:
About what documentation do you speak? https://docs.nvidia.com/video-technologies/video-codec-sdk/nvenc-video-encoder-api-prog-guide does not mention PVT ones, while the adaptive quantization modes are documented there.

TomMettam:
Yeah, googling for Psycho Visual Tuning only produces results for OBS. It's called adaptive quantisation everywhere else.

PatTheMav:
What is the state of this PR? From what I could tell there were discussions about:

* Whether this PR adds all the advanced NVENC knobs users should be able to twiddle around with, and
* Whether those knobs' labels should contain the nVidia-specific terms or crudely be translated into terms people might be aware of from other encoders.

Specifically to the second point, I agree with @Fenrirthviti that for advanced settings of NVENC this should also use the NVENC-specific terms. I have faith in users to educate themselves what terms mean if they find them in these labels and don't understand them. If they skip this step, well that's their choice to make.

derrod:
This PR is long-obsoleted by #10536 which also adds these options, and more.

PR #5777, opened 2022-01-12
decklink-output-ui: Stop outputs on OBS_FRONTEND_EVENT_EXIT

DDRBoxman:
### Description

Shuts down the decklink UI outputs when the OBS_FRONTEND_EVENT_EXIT is fired.

### Motivation and Context

When outputting over decklink and exiting OBS there was a crash

### How Has This Been Tested?

Tested locally on my windows machine by starting an obs decklink output and then pressing the exit button on the main ui.

### Types of changes

### Checklist:


notr1ch:
The outputs are already shut down when the module is unloaded at https://github.com/obsproject/obs-studio/blob/ed329c433dde0968239949b2952ec847eb94ffbd/UI/frontend-plugins/decklink-output-ui/decklink-ui-main.cpp#L350-L359

The fact that there is still a crash indicates unsafe shutdown ordering in libobs. This PR does fix the crash by moving the shutdown prior to libobs shutdown, but I believe it masks the real bug in libobs.

PR #5778, opened 2022-01-12
Set desktopFileName for QApplication

nicolasfella:
This needs to matcht the base name of the .desktop file name.

This is needed for propagating the correct app id on Wayland,
which is needed e.g. for proper window icons on Plasma Wayland


notr1ch:
Please read the [commit guidelines](https://github.com/obsproject/obs-studio/blob/master/CONTRIBUTING.rst#commit-guidelines) and fill out the PR template.

WizardCM:
Additionally, as we don't use the name "com.obsproject.Studio" outside of Flatpak, it may be worth scoping this.

PR #5785, opened 2022-01-12
media-playback: Fix bug with rist streams.

pkviet:
### Description
We currently pass a 'buffer_size' option to avformat in bytes.
Unfortunately, avformat/librist.c admits an option with the same name but in milliseconds.
The default values set in the UI (2 MB= 2 10^6) go beyond the range admitted in ms (0-30 000)
so avformat issues errors.
This fix disables the option when rist protocol is detected.

### Motivation and Context

Bug notice given in discord #beta channel : 

- Deagan: `Was RIST support only for outputting from OBS then, not currently for an inbound Media Source? I am attempting to use Larix Streamer to receive a RIST transmission using the "rist://@192.168.x.x:PORT?cname=RECEIVER" formatting, still to no resolve.`
- leb: `while i can't seem to get it working either with the same setup, i can replicate a crash with a rist url in the media source when the network buffer is set to 0mb`

rist streams currently aren't working with Media Source because of the buffer_size param
which is passed in wrong units.
It's better to be able to decode these streams.

Note that instead of disabling the buffer_size option which is set in UI (0-16 MB),
another option would be to translate the MB into milliseconds.
Since we don't know the incoming bitrate, it can't be an exact correspondence
though. 
An option I considered would be to have defaults in correspondence (2 MB <-> 1000 ms).
I leave this open to discussion. It wouldn't be difficult to implement.

### How Has This Been Tested?
Throw a rist stream (ex: w/ Larix Broadcaster) to obs-studio and check it is displayed.
(tested on windows 10 and macos 10.15.4)

### Types of changes

### Checklist:

### DIFF
--- a/deps/media-playback/media-playback/media.c
+++ b/deps/media-playback/media-playback/media.c
@@ -606,6 +606,8 @@ static int interrupt_callback(void *data)
 	return stop;
 }
 
+#define RIST_PROTO "rist"
+
 static bool init_avformat(mp_media_t *m)
 {
 #if LIBAVFORMAT_VERSION_INT < AV_VERSION_INT(59, 0, 100)
@@ -624,7 +626,8 @@ static bool init_avformat(mp_media_t *m)
 	}
 
 	AVDictionary *opts = NULL;
-	if (m->buffering && !m->is_local_file)
+	bool is_rist = strncmp(m->path, RIST_PROTO, strlen(RIST_PROTO)) == 0;
+	if (m->buffering && !m->is_local_file && !is_rist)
 		av_dict_set_int(&opts, "buffer_size", m->buffering, 0);
 
 	m->fmt = avformat_alloc_context();


WizardCM:
I also wouldn't be against doing a silent conversion.

pkviet:
> I also wouldn't be against doing a silent conversion.

ok, which kind ? assume for instance a 2.5 Mbs stream ? or a higher value ? or make the defaults correspond like I suggested.

pkviet:
One of the bug reporters confirm the bug is fixed.

Deagan:
Within the Media Source < Properties menu, I think it'd be nice to implement dynamic settings and/or a "Protocol" dropdown for selecting specific options relevant to each protocol. This may assist in presenting the Buffer Size/Millisecond difference for RIST you mentioned on the Github Issue.

ie.
Protocol Selection dropdown < Local File/SRT/RIST
*selects* ***Local File*** < **Source:** ***//path/to/file***
*selects* ***SRT*** < show ***Buffer Size*** slider < **Source:** *help dialog - "enter using* ***srt://192.168.1.x:PORT*** *formatting. (link to resources)"*
*selects* ***RIST*** < show ***Milliseconds*** slider < **Source:** *help dialog - "enter using* ***rist://@192.168.1.x:PORT?cname=RECEIVER*** *formatting. (link to resources)"*

pkviet:
@Deagan currently only the Main profile can be sent. For receiving , when using VLC source in obs, you must use Simple profile in Larix. When using Media Source, i've tested w/ Main profile but didn't check Simple profile. I'll retest to confirm. You might have to add a profile param in Media Source url.
For the other suggestions, I agree this would be convenient to add specific settings to UI but in practice our support for this protocol is just experimental at the moment. If the protocol gets widely adopted, this would change the discussion. But at this early stage, it's difficult to justify the extra work and maintainance debt.

pkviet:
reconnection issues and the stream showing only on preview not on scene are separate bugs; the first one is a known issue of the ffmpeg librist implementation on windows. The second would need repro steps and an issue should be filed.

PR #5800, opened 2022-01-15
UI: Fix undo causing window captures to disappear in studio mode

### LINKED ISSUE #5516: Window and display captures disappear after undo/redo.
### Operating System Info

Windows 10

### Other OS

_No response_

### OBS Studio Version

27.1.3

_No response_

### OBS Studio Log URL

https://obsproject.com/logs/FJ-gpQ3roLx5boRD

### OBS Studio Crash Log URL

_No response_

### Expected Behavior

When undoing, the window and display captures should still stay fine, even if they were resized.

### Current Behavior

When the window capture was resized, then undo'ed, the feed from the window capture disappeared and its source size became 0 x 0. The only way to bring it back is by selecting a different scene and going back, or bringing it out to program output.

### Steps to Reproduce

1. Have a scene (scene 1) with a window capture, and another (scene 2) with anything else.
2. Turn on Studio Mode.
3. Set up your window capture to capture any window, using the Windows 10 Graphics Capture (doesn't happen with BitBlt). 
4. Transition to Scene 2.
5. On Scene 1, move anything (could be the window capture, or anything else; it happened even if I moved only a color source).
6. Undo with Ctrl+Z. Notice that the window capture just turns black, although its bounding box does indeed get undo'ed (only if Bounding Box Type is not "No Bounds"; otherwise, there is no bounding box).
7. Edit Transform and notice that the source size becomes 0 x 0.
8. Redo with Ctrl+Y and see that the redo did work, but the capture is still black.
9. Select Scene 2 and go back to Scene 1. The capture should work again.
10. Try the steps again with a desktop capture, using Windows 10 Capture (I'm not sure about DXGI Desktop Duplication because it doesn't work on my laptop, I'm assuming due to 2 gpus).

Seems to work fine with BitBlt (tested with notepad, not really sure about other windows since I always use the Windows 10 capture method).
Even if only other sources are moved then undo'ed, window captures using Windows 10 capture break.

CodyJung:
### Description

When group undo/redo behavior was added in 2932cd0fd3, the saved json was modified such that the root object no longer contained the `"scene"` key (instead it is now contained within `"scenes_and_groups"`).

The lines of code removed in this patch were failing to find the scene in the json, so were setting the current scene to nullptr. This caused the window capture width/height to be set to 0 - and so they disappeared from the preview.

### Motivation and Context

This should fix #5516. I'm not positive of the original intent of this code, however - I _suspect_ it was intended to reset the preview to the scene you had edited if you clicked away, but I don't know if that ever worked correctly.

### How Has This Been Tested?

Tested on Windows 10 x64.

1. Be in Studio Mode
2. Move a Window Capture source in one of the following ways:
   - dragging in the preview scene
   - via the Edit Transform dialog
   - via the Reset Transform
3. Hit Ctrl-Z
4. Verified source transform un-dos without disappearing
5. Verified the same worked while not in Studio Mode

### Types of changes

### Checklist:

### DIFF
--- a/UI/window-basic-main.cpp
+++ b/UI/window-basic-main.cpp
@@ -7854,13 +7854,12 @@ void OBSBasic::on_actionCopyTransform_triggered()
 
 void undo_redo(const std::string &data)
 {
-	OBSDataAutoRelease dat = obs_data_create_from_json(data.c_str());
-	OBSSourceAutoRelease source =
-		obs_get_source_by_name(obs_data_get_string(dat, "scene_name"));
-	reinterpret_cast<OBSBasic *>(App()->GetMainWindow())
-		->SetCurrentScene(source.Get(), true);
+	obs_scene_t *scene = obs_scene_load_transform_states(data.c_str());
 
-	obs_scene_load_transform_states(data.c_str());
+	if (scene != nullptr) {
+		reinterpret_cast<OBSBasic *>(App()->GetMainWindow())
+			->SetCurrentScene(scene, true);
+	}
 }
 
 void OBSBasic::on_actionPasteTransform_triggered()
--- a/UI/window-basic-main.hpp
+++ b/UI/window-basic-main.hpp
@@ -474,7 +474,6 @@ private:
 
 	void SetPreviewProgramMode(bool enabled);
 	void ResizeProgram(uint32_t cx, uint32_t cy);
-	void SetCurrentScene(obs_scene_t *scene, bool force = false);
 	static void RenderProgram(void *data, uint32_t cx, uint32_t cy);
 
 	std::vector<QuickTransition> quickTransitions;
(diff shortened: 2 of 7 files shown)


jp9000:
Is there any way to fix this without removing the "switch to scene that the undo action occurred in" code?

CodyJung:
I'm definitely less comfortable with my changes here (the memory management especially), but I _think_ this works like I'd expect, now.

CodyJung:
Sorry about the delay! I've done some more testing, so I'm confident that this does what I want it to:

* Verified that if you've chosen another scene to preview, Ctrl-Z will now switch you back to the edited scene.
  * To accomplish that, I promoted void `SetCurrentScene(obs_scene_t *, bool);`  to public, and modified `obs_scene_load_transform_states()` to return an `obs_scene_t *` representing the modified scene. (If there were somehow multiple scenes in the json, it would return the last scene.)

* Verified that this works on both groups and sources within a group.

Issue #4670, opened 2021-05-08
Apple M1 VT Hardware Encoder (audio) fails Facebook stream with "Limit bitrate" enabled


kellemar:
### Operating System Info

macOS 11.3

### OBS Studio Version

27.0.0-rc3

### OBS Studio Log URL

https://gist.github.com/kellemar/64412d39920a76b059b5d8cb424f5022

### Expected Behavior

On Macbook Pro 13" Apple Silicon:

When VT Hardware Encoder is selected, and "Start Streaming" is clicked, OBS should connect to Facebook Live successfully.

### Current Behavior

No connection is made. However, x264 and VT Software Encoder works.

### Steps to Reproduce

1. In settings -> Output, select VT Hardware Encoder under the Advanced tab.
2.Leave settings as default.
3. Exit settings screen, and click on Start Streaming.

WizardCM:
What do you mean by "no connection is made"? Does an error appear? Did you wait longer than the 10 second test streram in your log?

kellemar:
Actually, this also happened with "Limit Bitrate" unchecked. Anyway I found the problem had to do with DIsabling B-frames.

With B-frames disabled, I wouldn't see anything on the stream, even though the Stream Health shows some data being sent. Enabling B-frames fixed the issue.

This issue doesn't occur on VT Software and x264 encoding.

RytoEX:
> 
> 
> Actually, this also happened with "Limit Bitrate" unchecked. Anyway I found the problem had to do with DIsabling B-frames.
> 
> With B-frames disabled, I wouldn't see anything on the stream, even though the Stream Health shows some data being sent. Enabling B-frames fixed the issue.
> 
> This issue doesn't occur on VT Software and x264 encoding.

If it's related to b-frames, these are the only places that option is used, for anyone's reference:
https://github.com/obsproject/obs-studio/blob/6a9f25c8ea1b244b4a439667be33f4a5cc6cdf4b/plugins/mac-vth264/encoder.c#L319-L321
https://github.com/obsproject/obs-studio/blob/6a9f25c8ea1b244b4a439667be33f4a5cc6cdf4b/plugins/mac-vth264/encoder.c#L671-L673

LPFchan:
Does this issue have to do with why VT Hardware Encoder can't limit the bitrate at all? 

[...]

PR #5809, opened 2022-01-16
mac-vth264: Fix DTS timestamps when blank

AuroraWright:
### Description

This changes fixes DTS timestamps in videotoolbox packets being always zero (at least on ARM macs). Credit for the changes goes to rcombs <rcombs@rcombs.me>, I was just helping her to debug this stuff.

### Motivation and Context

This was seemingly one of the factors causing https://github.com/obsproject/obs-studio/issues/4670 (Inability to save recordings when limiting bitrate, streams crashing). That issue isn't completely fixed however, as OBS still seems to drop frames after ~5 seconds when "Limit bitrate" is turned on (regardless of settings).

### How Has This Been Tested?

I tried streaming and recording before and after the change with "Limit bitrate" turned on, streaming would hang on a black screen and recording would save an empty file before, they both work now (albeit with frame drops).

### Types of changes

### Checklist:

### DIFF
--- a/plugins/mac-vth264/encoder.c
+++ b/plugins/mac-vth264/encoder.c
@@ -669,8 +669,10 @@ static bool parse_sample(struct vt_h264_encoder *enc, CMSampleBufferRef buffer,
 	dts = CMTimeMultiplyByFloat64(dts,
 				      ((Float64)enc->fps_num / enc->fps_den));
 
+	if (CMTIME_IS_INVALID(dts))
+		dts = pts;
 	// imitate x264's negative dts when bframes might have pts < dts
-	if (enc->bframes)
+	else if (enc->bframes)
 		dts = CMTimeSubtract(dts, off);
 
 	bool keyframe = is_sample_keyframe(buffer);


RytoEX:
Does this fix #4670 in its entirety, or are there yet other causes?  Basically, I want to know if it makes sense to continue to leave that Issue open if this gets merged, or can it be closed if this gets merged?

AuroraWright:
@RytoEX I think the issue should be left open but the title should be changed (since it's not related to audio and it's not exclusive to Facebook or even to streaming). This PR fixes the blackscreen (streaming) or hang (recording) with "limit bitrate" checked, but it still heavily drops frames after a couple seconds, regardless of which bitrate you set as the limit.

jp9000:
I suppose this is fine, although it feels like we're fixing something that's not necessarily our bug if that makes sense. I guess we have no choice if this is happening though. I really wish videotoolbox was more reliable.

AuroraWright:
It does fix the blackscreen when "limit bitrate" is checked, yeah (which, from what I can tell, is what the other person experienced). In my experience on a M1 Max machine and a native OBS build the "b-frames" setting didn't really affect this, "limit bitrate" did. There has also been some (related?) unknown issue causing frame drops on top of this timestamp issue, if you think it's best to open a new issue for this (and close the existing one) I can do that

RytoEX:
I don't have any Apple hardware currently, so I can only tell you [what the issue reporter said about enabling/disabling B-frames](https://github.com/obsproject/obs-studio/issues/4670#issuecomment-841804623):
> Actually, this also happened with "Limit Bitrate" unchecked. Anyway I found the problem had to do with DIsabling B-frames.
> 
> With B-frames disabled, I wouldn't see anything on the stream, even though the Stream Health shows some data being sent. Enabling B-frames fixed the issue.
> 
> This issue doesn't occur on VT Software and x264 encoding.

So perhaps they suffered from two distinct issues:
1. timestamps, which [incidentally were adjusted with an offset if B-frames were enabled](https://github.com/obsproject/obs-studio/issues/4670#issuecomment-841823805) and were not adjusted if B-frames were disabled
2. "limit bitrate" doing "something" that is as yet undetermined

If I understand correctly, your PR adjusts timestamps if the DTS is an invalid time ("all of its fields set to 0") by setting DTS to PTS whether or not B-frames are enabled instead of the current/previous behavior of adjusting DTS by an offset only if B-frames are enabled.  Perhaps enabling B-frames alleviated their (first) issue because it was also adjusting DTS?

Of course, it could also have been a red herring and the streaming platform simply couldn't display a stream without B-frames.  I've asked them to retest with a recent CI build.

----

> There has also been some (related?) unknown issue causing frame drops on top of this timestamp issue, if you think it's best to open a new issue for this (and close the existing one) I can do that

Yes, I much prefer narrowly scoped Issues as they tend to be more actionable, and it's easier to avoid confusion with multiple related issues/behaviors tied together. Please include detailed steps-to-reproduce and an OBS log from an affected session. Thanks!

PR #5816, opened 2022-01-16
UI: Use regexp to filter filename formatting

notr1ch:
### Description
Avoids unbounded stack growth when pasting content that has many invalid characters. Fixes https://github.com/obsproject/obs-studio/issues/5815.

### Motivation and Context
Fix a crash.

### How Has This Been Tested?
Pasted content containing many invalid characters. No crash.

### Types of changes

### Checklist:

### DIFF
--- a/UI/window-basic-settings.cpp
+++ b/UI/window-basic-settings.cpp
@@ -4031,17 +4031,18 @@ bool OBSBasicSettings::AskIfCanCloseSettings()
 
 void OBSBasicSettings::on_filenameFormatting_textEdited(const QString &text)
 {
+	QString safeStr = text;
+
 #ifdef __APPLE__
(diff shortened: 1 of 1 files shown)


RytoEX (changes requested):
QRegExp is not Qt6-compatible and is only available in the Qt Core5Compat component, which I'd prefer to avoid utilizing.  I suggest using QRegularExpression instead, which is available in all versions of Qt5 and Qt6.

See:
* https://github.com/obsproject/obs-studio/pull/4299
* https://github.com/obsproject/obs-vst/pull/71

See also Qt docs:
* https://doc.qt.io/qt-6/qtcore-changes-qt6.html#regular-expression-classes
* https://doc.qt.io/qt-6/qregexp.html#porting-to-qregularexpression

PR #5818, opened 2022-01-16
UI: Add ability to toggle projector frame/border

davidmyersdev:
### Description

This feature adds a "Toggle window border" option to the projector window's context (popup) menu.

Screen Shot 2022-01-16 at 3 01 13 PM
Screen Shot 2022-01-16 at 3 01 24 PM

### Motivation and Context

When using the OBS projector as a source for screen sharing in various apps (Discord, Meet, Zoom, etc), the title bar is included in the screen share and distracts from the presentation.

- https://obsproject.com/forum/threads/please-make-title-less-projector-windows-for-better-screen-sharing.151889/
- https://obsproject.com/forum/threads/is-it-possible-to-hide-the-title-bar-on-projector-when-on-windowed-mode.117065/
- https://obsproject.com/forum/threads/remove-menue-bar-in-projector-mode.41633/
- https://obsproject.com/forum/threads/macos-windowed-projector-preview-remove-title-bar.124256/

### How Has This Been Tested?

I have manually tested this on my machine.

```
System Version:	macOS 11.4 (20F71)
Kernel Version:	Darwin 20.5.0
```

```
Model Name:	MacBook Pro
Model Identifier:	MacBookPro16,1
Processor Name:	8-Core Intel Core i9
Processor Speed:	2.4 GHz
Number of Processors:	1
Total Number of Cores:	8
L2 Cache (per Core):	256 KB
L3 Cache:	16 MB
Hyper-Threading Technology:	Enabled
Memory:	64 GB
System Firmware Version:	1554.120.19.0.0 (iBridge: 18.16.14663.0.0,0)
```

### Types of changes

### Checklist:

### DIFF
--- a/UI/data/locale/en-US.ini
+++ b/UI/data/locale/en-US.ini
@@ -101,7 +101,7 @@ LogViewer="Log Viewer"
 ShowOnStartup="Show on startup"
 OpenFile="Open file"
 AddValue="Add %1"
-ToggleWindowBorder="Toggle window border"
+HideProjectorFrame="Hide the window frame"
 
 # warning if program already open
 AlreadyRunning.Title="OBS is already running"
@@ -796,6 +796,7 @@ Basic.Settings.General.WarnBeforeStoppingStream="Show confirmation dialog when s
 Basic.Settings.General.WarnBeforeStoppingRecord="Show confirmation dialog when stopping recording"
 Basic.Settings.General.Projectors="Projectors"
 Basic.Settings.General.HideProjectorCursor="Hide cursor over projectors"
+Basic.Settings.General.HideProjectorFrame="Hide window frames on projectors"
 Basic.Settings.General.ProjectorAlwaysOnTop="Make projectors always on top"
 Basic.Settings.General.Snapping="Source Alignment Snapping"
 Basic.Settings.General.ScreenSnapping="Snap Sources to edge of screen"
--- a/UI/data/locale/en-US.ini
+++ b/UI/data/locale/en-US.ini
@@ -101,6 +101,7 @@ LogViewer="Log Viewer"
 ShowOnStartup="Show on startup"
 OpenFile="Open file"
 AddValue="Add %1"
+ToggleWindowBorder="Toggle window border"
 
 # warning if program already open
 AlreadyRunning.Title="OBS is already running"
--- a/UI/forms/OBSBasicSettings.ui
+++ b/UI/forms/OBSBasicSettings.ui
@@ -473,20 +473,27 @@
                     </widget>
                    </item>
                    <item row="1" column="1">
+                    <widget class="QCheckBox" name="hideProjectorFrame">
+                     <property name="text">
+                      <string>Basic.Settings.General.HideProjectorFrame</string>
+                     </property>
+                    </widget>
+                   </item>
+                   <item row="2" column="1">
                     <widget class="QCheckBox" name="projectorAlwaysOnTop">
                      <property name="text">
                       <string>Basic.Settings.General.ProjectorAlwaysOnTop</string>
                      </property>
                     </widget>
                    </item>
-                   <item row="2" column="1">
+                   <item row="3" column="1">
                     <widget class="QCheckBox" name="saveProjectors">
                      <property name="text">
                       <string>Basic.Settings.General.SaveProjectors</string>
                      </property>
                     </widget>
                    </item>
-                   <item row="3" column="1">
+                   <item row="4" column="1">
                     <widget class="QCheckBox" name="closeProjectors">
                      <property name="text">
                       <string>Basic.Settings.General.CloseExistingProjectors</string>
@@ -5817,6 +5824,7 @@
   <tabstop>sourceSnapping</tabstop>
   <tabstop>centerSnapping</tabstop>
   <tabstop>hideProjectorCursor</tabstop>
+  <tabstop>hideProjectorFrame</tabstop>
   <tabstop>projectorAlwaysOnTop</tabstop>
   <tabstop>saveProjectors</tabstop>
   <tabstop>systemTrayEnabled</tabstop>
--- a/UI/window-basic-main.cpp
+++ b/UI/window-basic-main.cpp
@@ -10071,6 +10071,12 @@ void OBSBasic::UpdateProjectorHideCursor()
 		projectors[i]->SetHideCursor();
 }
 
+void OBSBasic::UpdateProjectorHideFrame(bool hideFrame)
+{
+	for (size_t i = 0; i < projectors.size(); i++)
+		projectors[i]->SetHideFrame(hideFrame);
+}
+
 void OBSBasic::UpdateProjectorAlwaysOnTop(bool top)
 {
 	for (size_t i = 0; i < projectors.size(); i++)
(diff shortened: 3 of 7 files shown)


WizardCM:
While toggling the window decorations on macOS seems to retain its position (according to your screenshots), this is not the case on Windows - each time it's toggled, the window shifts down the screen (based on the height of the top decorations). Ideally, it should retain its position.

Additionally, I believe the toggle should persist. You can set Projectors to save (in Settings -> General), so restarting OBS should remember whether decorations are supposed to be off.

Finally, when decorations are off, the window cannot be moved. Luckily I worked on that last year, so I can merge that commit alongside this PR when it's ready. https://github.com/wizardcm/obs-studio/commit/9ee79acaf8d62ba9c12fb4b2a497fa1ad912cfba

voraciousdev:
@WizardCM thanks for the feedback! I have access to a Windows machine, so I will triage that soon. If you think the setting should be persisted, I'm happy to make an attempt at saving it. I was just worried about the movement/resize issue in frameless mode.

WizardCM:
Yeah resizing in frameless mode might be a concern, but shouldn't be too hard to add - it'd require initialising (or just hiding/showing) a QSizeGrip.

Apparently if we were using a QDialog (which as far as I can tell, we're not) it'd just be a matter of calling `setSizeGripEnabled()`.

notr1ch:
I think this should be a checkable option, not a toggle command. E.g. [ ] Hide window frame

voraciousdev:
@notr1ch I'll attempt that change. I'm in the process of setting up a dev environment on my Windows machine, and I hope to get back to this over the weekend. I've been using my custom build throughout the week for work, and things are looking pretty smooth on Mac.

WizardCM:
I had some free time today so I tested this on Ubuntu (GNOME). It seems toggling the border does the opposite of Windows - hiding the frame shifts the window up, showing the frame doesn't move the window.

Warchamp7:
Agree with R1CH in having this be a checkable option like the "Always on top" entry.

Otherwise excited to see this, I've wanted something like this for a while!

voraciousdev:
@WizardCM what do you think would be the best approach to fixing this? Should we calculate the frame dimensions and set the frameless window's position explicitly?

WizardCM:
> @WizardCM what do you think would be the best approach to fixing this? Should we calculate the frame dimensions and set the frameless window's position explicitly?

I think this is most likely the easiest way to handle this, especially if the behavior differs per platform.

notr1ch:
Personally I think retaining the window position is somewhat out of scope if it's OS-dependent behavior.

WizardCM:
Sure, but we might as well make the UX nice in the first attempt.

voraciousdev:
Hey @WizardCM. I haven't had a chance to come back to this yet. I think I'll be able to tackle it this week though, so I'll keep you posted.

ftisiot:
This would be indeed a nice feature to have!

Warchamp7:
As a new question, when border is turned off is there any way to drag/reposition the window still?

WizardCM:
@Warchamp7 per my earlier comment:

> Finally, when decorations are off, the window cannot be moved. Luckily I worked on that last year, so I can merge that commit alongside this PR when it's ready. https://github.com/wizardcm/obs-studio/commit/9ee79acaf8d62ba9c12fb4b2a497fa1ad912cfba

voraciousdev:
Time sure does slip away. 😅 I know it's been over a month since I last posted an update here, but I wanted to pop back in to say I haven't forgotten about this PR. If things work out on my end, I'll be able to set aside time for this soon. 🤞

voraciousdev:
@WizardCM It's been a while, but I think I fixed the window position problem!

WizardCM:
I can confirm the position of the window now remains correct when the decorations are toggled.

Remaining tasks:

1. r1ch's suggestion: the menu item should behave like the Always on top (checkable) rather than a toggle
2. needs testing on linux

Did you also want to try adding the QSizeGrip yourself, or should we keep the scope of this PR smaller and I can do it when I submit the changes for dragging the window?

voraciousdev:
I'll convert to a toggle, and I can take a stab at the QSizeGrip. If it ends up being a bigger effort, maybe I can do it in a follow-up PR?

voraciousdev:
@WizardCM I converted the menu item to a checkbox and added it to the settings menu. I was able to get QSizeGrip working on Windows, but it doesn't appear to work on macOS, so I haven't committed it yet. I think we could do it with mouse events, but I don't want to go that route if there's a better way. I likely won't get a chance to pick this back up until next weekend, but let me know what you think makes sense!

WizardCM:
Hmm, now that I see the implementation, I think it might need to differ from Always On Top a little, as I personally see value in only setting specific projectors to be borderless. Having a global setting for the default is good, but it might be worth persisting per-projector. Though I'll let @notr1ch chime in on how he expects the behaviour to work. (feel free to leave it as-is until he responds)

------------

In terms of the QSizeGrip, I had a similar experience on macOS. I think it's due to how Qt translates certain widgets to native behaviour. I believe you have to specifically place your cursor on the QSizeGrip itself rather than where you'd normally expect to when resizing a window on macOS. Feel free to push that as a separate commit and I'll happily do some further tests of my own to try and work around the quirk.

Warchamp7:
> Hmm, now that I see the implementation, I think it might need to differ from Always On Top a little, as I personally see value in only setting specific projectors to be borderless. Having a global setting for the default is good, but it might be worth persisting per-projector. Though I'll let @notr1ch chime in on how he expects the behaviour to work. (feel free to leave it as-is until he responds)

Always On Top is a per projector setting. The setting under General is only a default. I would expect the same behaviour here.

chriskorosu:
Any updates on this? I would love to have this functionality

RytoEX:
> Any updates on this? I would love to have this functionality

Per previous messages:
> 2\. needs testing on linux

> I was able to get QSizeGrip working on Windows, but it doesn't appear to work on macOS, so I haven't committed it yet.

Other than that, as ever, if there were updates to provide on our end, we would post those updates here.

fsworld009:
Personally, I prefer the initial approach in the PR (toggle frameless via context menu), so we can avoid dealing with QSizeGrip since users can turn off frameless and adjust sizing.

Besides, with the use case the author describes (screen sharing in other video calling apps), I would expect most users would only need one projector to be frameless at one time, so seems fine to be an independent option for each projector, just like the `Always on top` option.

fsworld009:
I rebased this branch and tested on PopOS: (OS: Pop!_OS 22.04 LTS x86_64, Cosmic desktop, which is based on Gnome), works fine from what I can see
Screenshot from 2022-11-24 16-45-42
Screenshot from 2022-11-24 16-45-53

branch: https://github.com/fsworld009/obs-studio/commits/toggle-projector-frame

Re: `QSizeGrip`, would it be possible to do it in another PR instead, so we can have frameless support first?

Chiaki666:
Hello, I tried compile it on windows 11 but I had so many compile errors in visual studio so guess it doesn't work on windows yet?
Errors like:
error C3861: 'sender': identifier not found
error C2065: 'ui': undeclared identifier 
etc.

davidmyersdev:
Closing this in favor of https://github.com/obsproject/obs-studio/pull/9430.

PR #5819, opened 2022-01-17
libobs: Modify playback_mutex in audio monitor

walker-WSH:
### Description

1. add logic to destroy playback_mutex;
2. lock playback_mutex for audio_monitor_free

### Motivation and Context

as Description

### How Has This Been Tested?

switch on/off for monitor

### Types of changes

### Checklist:

### DIFF
--- a/libobs/audio-monitoring/win32/wasapi-output.c
+++ b/libobs/audio-monitoring/win32/wasapi-output.c
@@ -393,6 +393,8 @@ static inline void audio_monitor_free(struct audio_monitor *monitor)
 			monitor->source, on_audio_playback, monitor);
 	}
 
+	pthread_mutex_lock(&monitor->playback_mutex);
+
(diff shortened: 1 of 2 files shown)


jp9000:
It seems like the code on the OBS Studio master branch has changed to using a `SRWLOCK` for the `playback_mutex` object rather than a `pthread_mutex_t` object. The code in this PR may need to be fixed or reevaluated.

walker-WSH:
> It seems like the code on the OBS Studio master branch has changed to using a `SRWLOCK` for the `playback_mutex` object rather than a `pthread_mutex_t` object. The code in this PR may need to be fixed or reevaluated.

Oh, I did not notice type of playback_mutex has been changed on master.
So there no longer needs any destroy logic for it. 
Besides, since init/uninit for monitor object are both called from UI thread, the lock is not necessary now.

PR #5822, opened 2022-01-17
Enable curl ALPN support

notr1ch:
### Description
ALPN was disabled in 2016 due to poor server support, occasionally resulting in failed negotiations. Six years later, ALPN is now widely supported, and NPN is being removed in the latest version of nginx.

### Motivation and Context
NPN support is being removed from major web servers, leaving no way to negotiate protocol upgrades.

### How Has This Been Tested?
Verified connectivity still works on Windows 10. **Needs testing on Windows 7 / 8**.

### Types of changes

### Checklist:

### DIFF
--- a/UI/remote-text.cpp
+++ b/UI/remote-text.cpp
@@ -77,11 +77,6 @@ void RemoteTextThread::run()
 			curl_easy_setopt(curl.get(), CURLOPT_TIMEOUT,
 					 timeoutSec);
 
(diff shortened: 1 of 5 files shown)


jp9000:
What was the original issue that required setting that anyway?

notr1ch:
I don't fully recall, but I believe it was something like curl with schannel would fail to connect to some servers that uses ALPN at the time.

Issue #5621, opened 2021-12-05
Changing Global Audio Devices settings can produce sources with duplicate names if manually added to scenes


VodBox:
### Operating System Info

Windows 10

### OBS Studio Version

27.1.3

### OBS Studio Log URL

https://obsproject.com/logs/9yxVGx_Ad0J6_SDa

### Expected Behavior

Changing the Global Audio Devices options in Settings shouldn't produce name conflicts.

### Current Behavior

Changing the Global Audio Devices options in Settings can produce name conflicts if the underlying source has been added to a scene using "Add Existing".

### Steps to Reproduce

1. Set a Global Audio Device to some valid device and hit OK/Apply.
2. Add an Audio Input/Output Capture source and use add existing to pick the Global Audio Device.
3. Go back to Global Audio Device settings, change the device to disabled, then hit Apply.
4. Set the Global Audio Device back to some valid device and hit OK/Apply.
5. Repeat.

### Anything else we should know?

![image](https://user-images.githubusercontent.com/16394849/144736456-823947cc-6632-4eac-aa36-e155fbe36912.png)

sora-blue:
The root cause of name conflicts in the scene item list probably lies in the fact that scene item uses name of its source as its name while scene items are allowed to have the same source. It should be noted that sources themselves have checked unique names, but more than one scene items are using the name of one source. 

The simplest approach is to prevent scene items from using the same source, as it was in my last pr. But it is pointed out that users will not expect this, while you cannot duplicate sources directly sometimes, so the only approach is to add independent
name property for scene items. But this involves bunches of parts in, like releasing memory usage, which may cause further problems, and i'm confusing about it. 

In the current version, name of scene item is name of its source. So in most cases, when you get name of scene item, `obs_source_get_name(item->source)` is used. When you rename scene item, you rename the source, then other scene items using the same source. Adding independent name property for scene items means you leave name of sources away, which is nonsense.

sora-blue:
I didn't expect that name conflicts in source list would be intentional, so i totally mistook the took last time. The problem is specific to global audio sources.

Now i suppose the point is, it is expected that if you disable the global audio source on settings windows, it should be removed on spot, while the fact is that, if you added a scene items using global audio source, you added **extra strong reference** to it thus leading to its not being removed. On the other hand, changing from "disabled" to some available source directly causes creation of new source, so there are name conflicts.

[...]

PR #5825, opened 2022-01-19
libobs: check for adding duplicated source in one scene

sora-blue:
### Description

The added code checks if a source is already in the scene when adding existing sources. If so, nothing will be done except for a log warning.
The reason for not renaming is that I have noticed that flag `OBS_SOURCE_DO_NOT_DUPLICATE` may be set for sources, so that 
duplicating and renaming may not be proper.

Also, though no change is done to `/UI/window-basic-source-select.cpp` during this commit, I doubt that line 166 at: /UI/window-basic-source-select.cpp, namely
```cpp
obs_source_release(source);
```
It may should be 
```cpp
obs_source_release(from);
```
as `source` is newly created from `obs_source_duplicate`.

### Motivation and Context

This pull request is to fix [issue #5621](https://github.com/obsproject/obs-studio/issues/5621).

### How Has This Been Tested?

The effect is tested on Windows 10 21H2.

### Types of changes

Bug fix

### Checklist:

### DIFF
--- a/libobs/obs-scene.c
+++ b/libobs/obs-scene.c
@@ -1955,6 +1955,17 @@ static obs_sceneitem_t *obs_scene_add_internal(obs_scene_t *scene,
 		return NULL;
 	}
 
+	{
+		obs_sceneitem_t *item = scene->first_item;
+		while (item) {
+			if (item->source == source) {
+				blog(LOG_WARNING, "Tried to add a duplicated source to a scene");
+				return NULL;
+			}
+			item = item->next;
+		}
+	}
+
 	if (pthread_mutex_init(&mutex, NULL) != 0) {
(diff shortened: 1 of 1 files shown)


derrod:
Adding multiple scene items with the same source is perfectly fine, and something many users do for various reasons (e.g. alternative positions without that can be switched between having to have entirely separate scenes).

The issue linked is related to sources themselves not properly checking for unique names. And this PR probably does not fix it, since switching to a different scene that does not have the duplicated scene item in it and then contiuing the reproduction steps should still work.

sora-blue:
So I suppose the point here is that scene items, namely `obs_scene_item`, should have their independent name properties here in the source list of scenes. Thanks for the kindly detailed description!

PR #5827, opened 2022-01-19
libobs: Reset monitor-self if device of audio source is changed

walker-WSH:
### Description

If device of wasapi plugin is changed, `monitor::ignore` should be reset.

in fact，`ignore` should be re-checked in below situations：

1. monitor device is changed in settings window （already done at master）
2. captured device is changed for source （done by this PR）
3. default monitor device is changed while using default device for audio monitor （ done by #5398）

### Motivation and Context

ignore state should be updated

### How Has This Been Tested?

open monitor, 
then modify captured device for source，
check monitor state

### Types of changes

### Checklist:


walker-WSH:
in fact，ignore variable should be re-checked in below situations：

monitor device is changed in settings window （already done）

captured device is changed for source （by this PR）

default monitor device is changed  （to be done，need detect event）

PatTheMav:
@derrod looks sane to me at first glance? The `libobs` signal is only implemented on Windows which should be fine (macOS with SCK audio capture is device-independent, so the "self-monitoring check" doesn't work there to begin with, I dunno about Linux though).

PR #5514, opened 2021-11-10
obs-qsv11: Remove Intel discrete device ID checking

cmlin2:
### Description
This commit removes the old checking of Intel discrete device IDs and sets the default to integrated when Intel discrete and integrated graphics (i+I) are co-existing on the platform. 

### Motivation and Context
As Intel continues to add discrete graphics cards, maintaining a long list of discrete device IDs to check against isn't wise. When i+I platforms are detected, default to integrated graphics. 

### How Has This Been Tested?
Tested with the following scenarios:
- i+I: RKL+DG2, TGL+DG2, ADL+DG2
- DG2 only: TGL
- Monitor connected to iGPU in i+I cases and user selection of OBS on iGPU/dGPU (Low Power/High Performance Graphics Settings)
- Monitor connected to dGPU in i+I cases and user selection of OBS on iGPU/dGPU (Low Power/High Performance Graphics Settings)
- DG2 A0 128 EU, DG2 A0 512 EU and DG2 B0 512 EU cards

### Types of changes

 Tweak (non-breaking change to improve existing functionality)

### Checklist:

- [X] My code has been run through [clang-format](https://github.com/obsproject/obs-studio/blob/master/.clang-format).
- [X] I have read the [**contributing** document](https://github.com/obsproject/obs-studio/blob/master/CONTRIBUTING.rst).
- [X] My code is not on the master branch.
- [X] The code has been tested.
- [X] All commit messages are properly formatted and commits squashed where appropriate.
- [X] I have included updates to all appropriate documentation.

PR #5829, opened 2022-01-20
obs-qsv11: Memory leak bug fix in QSV plugin module

cmlin2:
### Description
This is a simple bug fix for a memory leak in the QSV plugin module

### Motivation and Context
Memory leaks (exacerbated by 4K and discrete graphics) will cause encoding failures on long recordings. These destructor checks/calls should have been in the original code but were missed and now fixed.

### How Has This Been Tested?
Re-tested on TGL and DG2 systems with QSV recording for 12+ hours to verify fix is working. 

### Types of changes

### Checklist:

- [ X] My code has been run through [clang-format](https://github.com/obsproject/obs-studio/blob/master/.clang-format).
- [ X] I have read the [**contributing** document](https://github.com/obsproject/obs-studio/blob/master/CONTRIBUTING.rst).
- [ X] My code is not on the master branch.
- [ X] The code has been tested.
- [ X] All commit messages are properly formatted and commits squashed where appropriate.
- [ X] I have included updates to all appropriate documentation.

### DIFF
--- a/plugins/obs-qsv11/common_directx11.cpp
+++ b/plugins/obs-qsv11/common_directx11.cpp
@@ -460,6 +460,9 @@ mfxStatus simple_copytex(mfxHDL pthis, mfxMemId mid, mfxU32 tex_handle,
 
 	km->ReleaseSync(*next_key);
 
+	km->Release();
+	input_tex->Release();
+
 	return MFX_ERR_NONE;
 }
 


RytoEX:
Can these be verified on non-TGL or DG2 systems?  If so, what are the steps-to-reproduce?

RytoEX:
The first commit here ("obs-qsv11: Remove Intel discrete device ID checking") was already merged into OBS from #5514 as https://github.com/obsproject/obs-studio/commit/3995326647f7255351a357b9f251659a3bdf75c7.  Was it mistakenly reincluded here?

cmlin2:
Yes, it's most easy to reproduce (< 10 minutes) on a DG2 system only doing 4K recording. But a non-DG2 TGL system was able to reproduce it after 12 hours of 4K recording as well. The steps to reproduce on a non-DG2 or non-TGL system would just be to do 4K recording for a long period of time. But with the memory leak fix, it shouldn't happen anymore.

cmlin2:
Sorry, my local branch was not updated, but I've fixed the issue and force pushed so that just the latest commit is included in the pull request (and not the one already merged into master).

jpark37:
I wouldn't bother setting the variables to nullptr. They're invalid as soon as you exit the function anyway.

jpark37:
Also don't need null checks because they have both been dereferenced by that point in the code.

cmlin2:
Thanks, I've updated the code to just do the needed "->Release()" calls and re-pushed.

hye5 (approved):

RytoEX (changes requested):
Please squash the two commits and use a commit subject that uses a verb after the prefix, such as:
> obs-qsv11: Fix memory leak in QSV plugin

cmlin2:
I've squashed the commits and force pushed with correct verb in commit subject

jpark37:
This is an OCD thing, but I prefer to release in reverse order of acquisition, mimicking RAII. If you could flip the two lines please. Commit looks good to me otherwise.

jpark37:
Doesn't look like the flip was made.

> There isn’t anything to compare.

cmlin2:
Oops, I forgot to do the "git add" before the "git commit --amend" and force push. Should be fixed now. :)

PR #5833, opened 2022-01-20
libobs: Don't destroy mutex before destroying sources is done

exeldro:
### Description

Don't destroy mutex before destroying sources is done

### Motivation and Context
Don't like crashes

### How Has This Been Tested?
on windows 64 bit

### Types of changes

### Checklist:

### DIFF
--- a/libobs/obs.c
+++ b/libobs/obs.c
@@ -699,6 +699,8 @@ static void obs_free_data(void)
 	FREE_OBS_LINKED_LIST(display);
 	FREE_OBS_LINKED_LIST(service);
 
+	os_task_queue_wait(obs->destruction_task_thread);
+
 	pthread_mutex_destroy(&data->sources_mutex);
 	pthread_mutex_destroy(&data->audio_sources_mutex);
 	pthread_mutex_destroy(&data->displays_mutex);


notr1ch:
In the context of OBS, closing the OBS window fires `OBSBasic::closeEvent` which clears all scenes via `ClearSceneData()`, which calls `obs_wait_for_destroy_queue()` which finally results in a call to `os_task_queue_wait(obs->destruction_task_thread)`.

While this code solves the case where non-OBS usage of libobs was missing a wait during libobs shutdown, how is OBS reaching `obs_free_data()` without the previous `os_task_queue_wait` call succeeding?

exeldro:
leaked references are destroyed after `ClearSceneData()`, so `os_task_queue_wait(obs->destruction_task_thread)` has to be called again after that

notr1ch (approved):

notr1ch:
Makes sense, I see libobs also calls `obs_wait_for_destroy_queue` at the start of `obs_shutdown` but that obviously misses the sources freed as a result of `obs_free_data`.

walker-WSH:
> Makes sense, I see libobs also calls `obs_wait_for_destroy_queue` at the start of `obs_shutdown` but that obviously misses the sources freed as a result of `obs_free_data`.

Since `obs_source_addref `and `obs_source_release `are called from `graphic-thread` and `mixer-thread`,
new destroy task for source may be posted before ending the two threads.

walker-WSH (approved):
Since `obs_source_addref and obs_source_release `are called from `graphic-thread and mixer-thread,`
new destroy task for source may be posted before ending the two threads.

So, it's right to add this code after `stop_video() and stop_audio()`

walker-WSH:
By the way, I think adding sleep during video_ticks can help reproduce this crash.

```
while (for each source)
{
    obs_source_add_ref(source);

    source->info.tick();
    // before release refs, add sleep, 
    // make sure the following release can be called when UI thread runs into obs_shutdown
    Sleep(2000); 

    obs_source_release(source)
}
```

PR #5848, opened 2022-01-23
UI: More accurately describe "hide OBS windows" option, show first time warning

WizardCM:
### Description
This makes the name of the setting a little more clear, and adds a tooltip to provide a proper scope to its effect.

image

image

 * Original PR: #5698

### Motivation and Context

I have confirmed this setting hides OBS from the following applications:

1. TeamViewer
2. Discord "Share your screen" (entire display)
3. Jump Desktop
4. Parsec
5. Skype (display & window)
6. Microsoft Teams (display & window)
7. ShareX (all modes, screenshots & videos)
8. Built in Windows screenshot tools
    * Snip & Sketch (including Win+Shift+S)
    * Snipping Tool
    * PrntScr (to clipboard)
    * Win+PrntScr (automatically saves to Pictures\Screenshots)
9. Windows Magnifier
10. [ZoomIt](https://docs.microsoft.com/en-us/sysinternals/downloads/zoomit)
11. Twitch Studio (Entire Screen)
12. Chrome Remote Desktop
13. A second OBS instance, in either Display or Window capture mode, in either DXGI, BitBlt or WGC

Improving the explanation to the user of what the setting does is beneficial.

### How Has This Been Tested?
* Open the Settings window
* Hover over the setting
* Enable the setting

### Types of changes

### Checklist:


notr1ch:
I'm worried this goes a bit too far in trying to explain that it affects other apps. I think the focus should be on hiding OBS from display capture on a single monitor setup so users are likely to find the option, the other apps part should be a secondary aspect.

WizardCM:
Sure, but I also want it to be clear to users why OBS would disappear from other apps. Maybe the setting itself could be "Hide OBS windows from screen capture"? That way if they just want to hide it for OBS it's clear enough?

WizardCM:
New discovery: it seems enabling this option immediately terminates Shadowplay recordings (at least in Desktop mode). If a recording is currently occurring, it ends about half a second after the setting is enabled. If starting a new recording, it terminates immediately and does not create a recording file.

I feel this would be a good warning to include in the tooltip, though I'm unsure how to word it.

notr1ch:
That doesn't sound good. It might be worth reaching out to Nvidia about that, surely that isn't the intended behavior.

Fenrirthviti:
Just logging here, I've reached out to our NVIDIA contacts on the Shadowplay issue.

notr1ch:
Given the variety of use cases affected, I'm beginning to think a one-time dialog when turning it on for the first time is a better option. It's difficult to convey all the information in a single description or tooltip.

WizardCM:
Converting to draft until I've made the changes to add a dialog to show on first toggle of this option.

Wording suggestions are welcome.

Fenrirthviti:
Spitball draft:

Enabling this option hides the main OBS Studio window from being captured by OBS and may affect other applications, such as conferencing, screen sharing, remote desktop, or other capture software. Are you sure you want to enable it?

WizardCM:
I have now made the warning only appear the first time.

I also experimented with disabling the Apply button after the warning is dismissed with "No", however I noticed a pretty bad bug with it and felt better just not touching that aspect for now.

I've also adjusted the wording of the tooltip to match the dialog.

This PR is now ready for review.

notr1ch:
Small text nitpick, since we know for sure it affects almost every app we've tested, I think we can drop the "may affect" in "may affect other applications". And I think we should lowercase "Projector".

Is the tooltip still necessary now given we have the dialog? It feels like it might be a bit too much information overload for one option.

WizardCM:
My reason for keeping the tooltip is that the dialog only shows the first time, and the user may forget after not touching the setting for a long time.

I can make those two text changes, absolutely.

notr1ch:
I also wonder if this should be an informational dialog with just the OK button. "Are you sure" dialogs create an unpleasant user experience as it makes the user thing they are doing something bad or dangerous, when this is a very simple non-destructive operation. If the user decides they don't want to enable it after reading the full details, they can untick the box again.

WizardCM:
That is also an option, I just reused existing paradigms from other Settings dialogs. I'm personally not fussed either way.

WizardCM:
Made the requested changes.

PR #5860, opened 2022-01-24
Deprecate addref for obs objects

jp9000:
### Description
Deprecate most addref calls for OBS objects: sources, scenes, outputs, encoders, services. The get_ref functions should be used instead.

Additionally, almost all cases where addref is used in OBS have also been replaced with their equivalent get_ref for objects that support it.

### Motivation and Context
With 2416dfbd5e5a77, I now have seen that some third-party plugins are calling addref on sources that are somehow already fully released, which is dangerous. With this PR, this will encourage plugin developers not to use the addref functions, and instead hopefully rely on the get_ref functions instead, which will only return a valid incremented object if those objects are valid. In the future perhaps we might end up doing this with other objects, but the primary issue is with sources.

### How Has This Been Tested?
Compiled and tested, but could probably use some wider testing via beta/release candidates.

### Types of changes

### Checklist:


WizardCM:
Probably better to use this in the docs rather than a "Note:"
```
.. deprecated:: 27.2.0
   Use :c:func:`obs_encoder_get_ref()` instead.
```
[Reference](https://www.sphinx-doc.org/en/master/usage/restructuredtext/directives.html#directive-deprecated).

PR #5869, opened 2022-01-25
flatpak: Enable communication with org.a11y.Bus

TheMuso:
### Description
This allows Linux accessibility tools such as the Orca screen reader to be used with OBS.

### Motivation and Context
Without this change, it is not possible to use the Orca screen reader to get speech and braille feedback from the OBS beta flatpak.

### How Has This Been Tested?
1. I added the flathub-beta flatpak repository to my local flatpak installation.
2. I installed com.obsproject.Studio from the flathub-beta repository using the flatpak command-line utuility.
3. I enabled the Orca screen reader using the Super + Alt + S shortcut key.
4. I searched for and ran OBS from the activities overiew in my GNOME desktop session.
5. OBS loaded and ran, but when attempting to navigate the initial configuration wizard, I did not hear anything spoken with text to speech from Orca. I exited the wizard, but still did not hear anything.
6. After making the change in this pull request, I built and installed a local copy of the OBS git master branch as a flatpak using flatpak-builder: "flatpak-builder builddir --force-clean CI/flatpak/com.obsproject.Studio.json --user --install".
7. I re-ran OBS as per the above steps, also making sure Orca was running.
8. This time, I was able to navigate the initial configuration wizard with the keyboard, and get audible feedback from Orca. I also got audible feedback when navigating the main OBS window.

### Types of changes

- Bug fix (non-breaking change which fixes an issue) -->

### Checklist:

### DIFF
--- a/CI/flatpak/com.obsproject.Studio.json
+++ b/CI/flatpak/com.obsproject.Studio.json
@@ -16,6 +16,7 @@
     "--talk-name=org.kde.StatusNotifierWatcher",
     "--talk-name=org.freedesktop.Flatpak",
     "--talk-name=org.freedesktop.Notifications",
+    "--talk-name=org.a11y.Bus",
     "--own-name=org.kde.StatusNotifierItem-2-2",
     "--system-talk-name=org.freedesktop.Avahi",
     "--env=VST_PATH=/app/extensions/Plugins/lxvst"


GeorgesStavracas (approved):
Took the liberty to confirm with KDE / Qt folks if this was a Flatpak problem, or an OBS problem, or a Qt problem, and seems like this PR is correct. Thanks!

GeorgesStavracas:
Quick note: we can revert this when https://invent.kde.org/packaging/flatpak-kde-runtime/-/merge_requests/74 is merged and reaches Flathub. It might take some time, so I think we can keep this change until then.

PR #5874, opened 2022-01-27
win-dshow: Log buffered state

flaeri:
### Description

Adds entry for buffering state of dshow sources to the log.

In order for this to operate successfully, and continue to do so in the future, it moves the `SetupBuffering()` above the `blog` so the buffering is set before the log prints.

### Motivation and Context

Dshow issues show up fairly frequently in support context, and knowing the state of buffering can be useful when troubleshooting. Applicable for issues regarding smoothness/stutter as well as latency.

### How Has This Been Tested?

Tested on windows 10 20H2, verifying correct log output using auto and enabled/disable buffering on the following:
* virtual devices (virtualcam)
* eos Webcam utility
* basic capture card (cantlink)

### Types of changes

### Checklist:

### DIFF
--- a/plugins/win-dshow/win-dshow.cpp
+++ b/plugins/win-dshow/win-dshow.cpp
@@ -972,6 +972,8 @@ bool DShowInput::UpdateVideoConfig(obs_data_t *settings)
 	os_wcs_to_utf8_ptr(videoConfig.path.c_str(), videoConfig.path.size(),
 			   &path_utf8);
 
+	SetupBuffering(settings);
+
 	blog(LOG_INFO, "---------------------------------");
 	blog(LOG_INFO,
 	     "[DShow Device: '%s'] settings updated: \n"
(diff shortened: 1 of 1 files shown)


plugins/win-dshow/win-dshow.cpp
```diff
@@ -980,11 +980,13 @@ bool DShowInput::UpdateVideoConfig(obs_data_t *settings)
...
+	     "\tformat: %s\n"
+	     "\tbuffering: %s",
 	     obs_source_get_name(source), (const char *)name_utf8,
 	     (const char *)path_utf8, videoConfig.cx, videoConfig.cy_abs,
 	     (int)videoConfig.cy_flip, fps, videoConfig.frameInterval,
-	     formatName->array);
+	     formatName->array,
+	     obs_source_async_unbuffered(source) ? "disabled" : "enabled");
 
 	SetupBuffering(settings);
```
notr1ch, line 991: It looks like this is logged before the buffering is actually applied to the source. Perhaps checking the settings would be better, i.e. `obs_data_get_int(settings, BUFFERING_VAL)`

flaeri: The dshow logs always print twice (prior to my pr as well). Before your proposed change it would print the prior value, and then the updated one. Now it prints the new value in the first and second print.

Not sure if its unintentional that its printing twice (before and after?), but yeah, I'll happily change it.

flaeri: hmm, it seems to me that `BUFFERING_VAL` is outputting what the user has selected (auto, enable or disable).
I find that less valuable than whether the buffering is in use or not.

If the user set auto, then you wont actually know based on the log whether its in use or not.
I believe the original implementation would solve `auto` into whether buffering is enabled or not.
It might happen on the 2nd print, but it would do it (I believe).

flaeri: Ok, so I think that this should take care of the concern while still doing what I intended it to do.

Spent some time looking at the double logging (too complicated for me), but seeing as it was always there, I feel like that will have to be for someone else to dig into, at some point, but is unrelated to the PR.

Issue #2760, opened 2020-04-20
Apple VT Encoder (mac-vth264) recording with fractional frame rate does not match OBS frame rate


juniorporciuncula:
## Platform
Operating system and version: macOS Mojave (10.14.6)
OBS Studio version: 25.0.7

## Expected Behavior
Streaming and Recording must record the files with the selected frame rate on OBS

## Current Behavior
Streaming and Recording are recorded with 25 fps (even if I set OBS for 30fps) and 50fps (even if I set OBS for 60fps)
This doesn't happen with x264. On x264 the recording will respect the OBS frame rate.

## Steps to Reproduce
Set OBS for 30 fps and record the video using Apple VT Hardware Encoder.

## Additional information
I've tried a lot of versions of OBS, but this problem occurs with all version, in any macOS I've tested (Yosemite, El Capitan, High Sierra and Mojave)
![Captura de Tela 2020-04-19 às 21 52 56](https://user-images.githubusercontent.com/10005660/79705133-f5410c00-828a-11ea-976d-c52a240ade1e.png)

Fenrirthviti:
Can you please properly fill out the bug report template?

juniorporciuncula:
Sorry. I'll do that!

snicolai-blog:
I just used the VT hardware encoder at 30 fps for a recording over an hour long this morning.  The local recording is at 30 fps as shown by the QuickTime player.  The primary video source was a video camera at 1280x720 30 fps.

MacOS X 10.14.6 (18G4032) MacBook Pro (Retina, 15-inch, Mid 2014)  Running with en-us locale settings.

OBS 25.0.6

juniorporciuncula:
> I just used the VT hardware encoder at 30 fps for a recording over an hour long this morning. The local recording is at 30 fps as shown by the QuickTime player. The primary video source was a video camera at 1280x720 30 fps.
> 
> MacOS X 10.14.6 (18G4032) MacBook Pro (Retina, 15-inch, Mid 2014) Running with en-us locale settings.
> 
> OBS 25.0.6

Please, download the software "Invisor Lite" and check the REAL (Original) fps of your video

snicolai-blog:
That shows 30 fps as well.
<img width="477" alt="Screen Shot 2020-04-19 at 8 49 37 PM" src="https://user-images.githubusercontent.com/59782525/79706895-5ca58e80-8280-11ea-932b-848ec8e1b21c.png">

[...]

PR #5880, opened 2022-01-28
Various fixes for mac-vth264

### LINKED ISSUE #5840: "RealTime" VT property + "limit bitrate" setting cause heavy framedrops on ARM macs with HW VideoToolbox encoder
### Operating System Info

macOS 12.0

### Other OS

_No response_

### OBS Studio Version

Git

_No response_

### OBS Studio Log URL

https://obsproject.com/logs/mCaYXIhd7YJsEBtj

### OBS Studio Crash Log URL

_No response_

### Expected Behavior

Frames should not be dropped when using the hardware VideoToolbox encoder and "limit bitrate" turned on in preferences.

### Current Behavior

Heavy frame drops occur no matter the target bitrate limit when the "limit bitrate" checkbox is toggled. After some tests I found out this is caused by the kVTCompressionPropertyKey_RealTime property, which is currently always set by OBS here: https://github.com/obsproject/obs-studio/blob/2c5b1a837f0912554f82cfb60f278dc06ca514b0/plugins/mac-vth264/encoder.c#L325 . No drops at all occur (even at really high bitrates) if that property is not set. A possible solution could be to make that property user-selectable (if you like the idea I might make a PR as soon as I have some time to check out how Qt UI stuff works).

### Steps to Reproduce

1. Select the VideoToolbox hardware encoder on macOS
2. Check "limit bitrate" and enter any bitrate you like
3. Check output
...

_No response_

RytoEX:
### Description

This PR consists of three commits that provide various fixes for the mac-vth264 encoder plugin.

----

mac-vth264: Remove OSX 10.8 compatibility code
    
OBS no longer supports building on OSX. The minimum version of macOS that we support is macOS 10.13. We can safely remove this 10.8 compatibility code.

The changes in this commit apparently make the minimum version supported macOS 10.10 due to no longer providing a custom enum for [kCMFormatDescriptionBridgeError_InvalidParameter](https://developer.apple.com/documentation/coremedia/1416276-format_description_bridge_error_/kcmformatdescriptionbridgeerror_invalidparameter?language=objc).

----

mac-vth264: Use float for expected frame rate

The encoder property kVTCompressionPropertyKey_ExpectedFrameRate is a hint to the video encoder. Since frame rates can be fractional, let's use a float here instead of an int.

----

mac-vth264: Set RealTime property to False

While the Apple documentation currently seems to indicate that we should set RealTime to True, it appears that this may be causing issues with the encoder not being able to meet the target frame rate. Both FFmpeg and Handbrake have recently explicitly set this value to False, and preliminary tests seem to indicate that setting this to false in OBS has favorable results. 

----

Given that this is a Bug Fix.  I'm pushing for this to be fast-tracked into OBS Studio 27.2.0.

I'm willing to split commits off into separate PRs if needed.

### Motivation and Context

Fixes #5840.

Bad or sub-par behavior is bad, and we should fix it.

### How Has This Been Tested?

Preliminary tests by @Shirk and @AuroraWright seem promising.  Would appreciate more testing on any or all of these commits by macOS devs.

### Types of changes

### Checklist:

### DIFF
--- a/plugins/mac-vth264/encoder.c
+++ b/plugins/mac-vth264/encoder.c
@@ -319,13 +319,13 @@ static bool create_encoder(struct vt_h264_encoder *enc)
 
 	// This can fail depending on hardware configuration
 	code = session_set_prop(s, kVTCompressionPropertyKey_RealTime,
-				kCFBooleanTrue);
+				kCFBooleanFalse);
 	if (code != noErr)
-		log_osstatus(LOG_WARNING, enc,
-			     "setting "
-			     "kVTCompressionPropertyKey_RealTime, "
-			     "frame delay might be increased",
-			     code);
+		log_osstatus(
+			LOG_WARNING, enc,
+			"setting kVTCompressionPropertyKey_RealTime failed, "
+			"frame delay might be increased",
+			code);
 
 	STATUS_CHECK(session_set_prop(s, kVTCompressionPropertyKey_ProfileLevel,
 				      obs_to_vt_profile(enc->profile)));
--- a/plugins/mac-vth264/encoder.c
+++ b/plugins/mac-vth264/encoder.c
@@ -19,20 +19,6 @@
 #define VT_BLOG(level, format, ...) \
 	VT_LOG_ENCODER(enc->encoder, level, format, ##__VA_ARGS__)
 
-// Clipped from NSApplication as it is in a ObjC header
-extern const double NSAppKitVersionNumber;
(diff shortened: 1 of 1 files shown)


DDRBoxman (approved):

energizerfellow:
Would this correct #2760 as well?

RytoEX:
> Would this correct #2760 as well?

I thought that that may be a possibility, but it would need to be tested on an affected system.

PR #5891, opened 2022-01-29
aja-output-ui: Fix crash when stopping AJA Preview output

paulh-aja:
### Description

Stopping the Preview output in the AJA Output UI can lead to a crash.

**Problem**
The function `preview_output_stop` in `UI/frontend-plugins/aja-output-ui/aja-ui-main.cpp` calls `video_output_stop`, which results in a call to `pthread_mutex_destroy`: https://github.com/obsproject/obs-studio/blob/364f102561faf565ae22e5fa0f1f4aabf3e07c48/UI/frontend-plugins/aja-output-ui/aja-ui-main.cpp#L103

This results in the `data_mutex` field of `struct video_output` from being deleted and made `NULL` while still in use by the `render_preview_source` function in aja-ui-main.cpp.

The call to `video_output_close` a few lines after this also results in a call to `video_output_stop` on the same structure. Allowing this call to perform the `video_output` mutex cleanup seems to be the correct action.

https://github.com/obsproject/obs-studio/blob/364f102561faf565ae22e5fa0f1f4aabf3e07c48/libobs/media-io/video-io.c#L267

**Resolution**
Removing the first call to `video_output_stop` appears to fix the crash and not result in any other harmful effects.

### Motivation and Context

AJA QA found a sporadic crash when stopping the Preview output from the AJA Output UI. Some machines exhibit the crash more often than others.

### How Has This Been Tested?

Tested by spamming Preview Start/Stop, both manually and by holding down the Enter key on Windows to engage/disengage it as fast as the keyboard repeat rate would allow.

### Types of changes

### Checklist:

### DIFF
--- a/UI/frontend-plugins/aja-output-ui/aja-ui-main.cpp
+++ b/UI/frontend-plugins/aja-output-ui/aja-ui-main.cpp
@@ -100,7 +100,6 @@ void preview_output_stop()
 {
 	obs_output_stop(context.output);
 	obs_output_release(context.output);
-	video_output_stop(context.video_queue);
 
 	obs_remove_main_render_callback(render_preview_source, &context);
 	obs_frontend_remove_event_callback(on_preview_scene_changed, &context);


paulh-aja:
This code originated from the `decklink-output-ui` code. I don't have a way to check and see if the crash does in fact occur there too. @DDRBoxman please make the fix for that plugin.

PR #5907, opened 2022-02-02
UI: Add Wine detection/warning

derrod:
Screenshot from Ubuntu 20.04 running in Wine stable:
Screenshot_from_2022-02-06_11-26-27

### Description
Adds a warning on startup if OBS is run in Wine and logs Wine version.

Note: this moves the creation of the log file so it's ready once the check happens. This will also fix the Rosetta status not being logged.

### Motivation and Context
While running OBS in Wine works surprisingly well, you shouldn't.
This just warns people who don't realise that (they exist, we had them in #linux-support).

### How Has This Been Tested?
- Ran OBS in Wine
- Ran OBS not in Wine

### Types of changes

### Checklist:

### DIFF
--- a/UI/data/locale/en-US.ini
+++ b/UI/data/locale/en-US.ini
@@ -109,6 +109,9 @@ AlreadyRunning.LaunchAnyway="Launch Anyway"
 ChromeOS.Title="Unsupported Platform"
 ChromeOS.Text="OBS appears to be running inside a ChromeOS container. This platform is unsupported"
 
+Wine.Title="Wine detected"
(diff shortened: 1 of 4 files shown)


gxalpha:
As noticed in #5738, `blog` isn't initialized at this point, meaning the message won't appear in the log file.
Also, I'm wondering if this should also be a libobs function, with the logging in `log_system_info`?

derrod:
Yeah this could be reworked to be similar to #5738 (should that be merged). Not sure what the best way of getting that to the UI is though (should we decide to show a warning).

derrod:
I'm not entirely sure why `Qt::RichText` doesn't seem to work there.

tommyvct:
We should really setup a standard of where should we do the logging and user warning (this dialogue and similar ones).

I suggest that we keep the logging part more or less strictly in `libobs`, then user warning dialogue here in UI.

From a portability standpoint, considering `libobs` can be used elsewhere in another project, moving the log into the `libobs` makes more sense.

There is a drawback though, that is we are doing the same work for twice (namely check this kind of thing once in UI, once in `libobs`). But such cost should be negligible.

PR #5929, opened 2022-02-08
fix virtual studio warning: C6011

seker:
### Description
fix virtual studio warning: C6011

### Motivation and Context
in vs2019, there is a compile warning: C6011
https://docs.microsoft.com/zh-cn/cpp/code-quality/c6011?view=msvc-170

### How Has This Been Tested?
just compile warning, no test case

### Types of changes
Code cleanup 

### Checklist:
fix virtual studio warning: C6011

### DIFF
--- a/libobs/util/platform-windows.c
+++ b/libobs/util/platform-windows.c
@@ -1031,6 +1031,10 @@ static inline void rtl_get_ver(struct win_version_info *ver)
 static inline bool get_reg_sz(HKEY key, const wchar_t *val, wchar_t *buf,
 			      const size_t size)
 {
+	if (!buf) {
+		return false;
+	}
+
 	DWORD dwsize = (DWORD)size;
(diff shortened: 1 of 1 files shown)


jpark37:
I think this would fix the warning without adding a NULL check, and terminates the string in the correct location.
```
static inline bool get_reg_sz(HKEY key, const wchar_t *val, wchar_t *buf,
			      DWORD size)
{
	const LSTATUS status =
		RegGetValueW(key, NULL, val, RRF_RT_REG_SZ, NULL, buf, &size);
	return status == ERROR_SUCCESS;
}
```

Also, please do not ignore the checklist. Your commit message is not formatted correctly, and your changes should not be on the master branch.

seker:
> 

the warning is on this line:

	buf[(size / sizeof(wchar_t)) - 1] = 0;

PR #5930, opened 2022-02-08
libobs/UI: solving name conflicts in source list

sora-blue:
### Description

As is explained in my comment at [issue #5621](https://github.com/obsproject/obs-studio/issues/5621), my resolution is to add an independent property for scene items. And because it involves many parts in, finally I store it in `item->private_settings` to prevent memory allocating & releasing issues. 
(see `obs_sceneitem_set_itemname` and more).

Generally, after modification, when you add an existing source, arg `existing` will be set `true` at `obs_scene_add_internal`, thus leading to searching for a new name. On the other side, when you rename scene items, you will never rename source itself but only "item_name" of scene items. **New name property will only be shown on source list and context bar.** 

Also, there are 2 main bold changes that I'm worrying about. 

One is changes in properties of `struct AddSourceData` and arguments of `obs_scene_add_internal` & `obs_scene_add`. I found it hard to directly set "item_name" in function `AddExisting` at windows-basic-source-select.cpp and above, so I add arg & property instead to ask for a new name.

The other is line 539 at source-tree.cpp. Although renaming sources directly is stopped, you can still rename sources in mixer, which leads to incorrect labels changes in source list of scene. As I cannot think of a better idea to stop it, I stopped it at line 539 directly.

Detailed explanation is attached.

### Motivation and Context

This pr is to fix [issue #5621](https://github.com/obsproject/obs-studio/issues/5621).

### How Has This Been Tested?

The effect is tested on my laptop running Windows 10 21H2 and it works.

### Types of changes

Bug fix

### Checklist:

### DIFF
--- a/UI/source-tree.cpp
+++ b/UI/source-tree.cpp
@@ -39,7 +39,7 @@ SourceTreeItem::SourceTreeItem(SourceTree *tree_, OBSSceneItem sceneitem_)
 	setMouseTracking(true);
 
 	obs_source_t *source = obs_sceneitem_get_source(sceneitem);
-	const char *name = obs_source_get_name(source);
+	const char *name = obs_sceneitem_get_itemname(sceneitem);
 
 	OBSDataAutoRelease privData =
 		obs_sceneitem_get_private_settings(sceneitem);
@@ -434,8 +434,7 @@ void SourceTreeItem::ExitEditModeInternal(bool save)
 	/* ----------------------------------------- */
 	/* Check for same name                       */
 
-	obs_source_t *source = obs_sceneitem_get_source(sceneitem);
-	if (newName == obs_source_get_name(source))
+	if (newName == obs_sceneitem_get_itemname(sceneitem))
 		return;
 
 	/* ----------------------------------------- */
@@ -452,16 +451,28 @@ void SourceTreeItem::ExitEditModeInternal(bool save)
 	}
 
 	/* ----------------------------------------- */
+	/* check for existing scene item             */
+	exists = obs_scene_if_duplicated_itemname(scene, newName.c_str());
+	if (exists) {
+		OBSMessageBox::information(main, QTStr("NameExists.Title"),
+					   QTStr("NameExists.Text"));
+		return;
+	}
+
+	/* ----------------------------------------- */
 	/* rename                                    */
 
 	SignalBlocker sourcesSignalBlocker(this);
-	std::string prevName(obs_source_get_name(source));
+	std::string prevName(obs_sceneitem_get_itemname(sceneitem));
 	std::string scene_name =
 		obs_source_get_name(main->GetCurrentSceneSource());
-	auto undo = [scene_name, prevName, main](const std::string &data) {
-		OBSSourceAutoRelease source =
-			obs_get_source_by_name(data.c_str());
-		obs_source_set_name(source, prevName.c_str());
+	auto itemid = obs_sceneitem_get_id(sceneitem);
+	obs_sceneitem_t *item_ = sceneitem.Get();
+	auto label_ = label;
+	
+	auto undo = [scene_name, prevName, main, item_, label_](const std::string &data) {
+		obs_sceneitem_set_itemname(item_, prevName.c_str());
+		label_->setText(QT_UTF8(prevName.c_str()));
 
 		OBSSourceAutoRelease scene_source =
 			obs_get_source_by_name(scene_name.c_str());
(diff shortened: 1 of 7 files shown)


sora-blue:
Detailed explanation: 

bold changes that may not proper:
1. line 2033 at obs-scene.c on hot keys
2. line 466~486 at source-tree.h on redo/undo
3. line 539 at source-tree.cpp on ignoring signal "rename"
4. line 889 at source-tree.cpp on data


at obs-scene.c

1. four methods added to support a unique name property for:
const char *obs_sceneitem_get_itemname(const obs_sceneitem_t *);
void obs_sceneitem_set_itemname(obs_sceneitem_t *item, const char *newName);
bool obs_scene_if_duplicated_itemname(const obs_scene_t *scene,
				      const char *newName);
char* obs_scene_get_new_itemname(const obs_scene_t *scene, const char* name);

2. transform from #obs_source_get_name(item->source)# to obs_sceneitem_get_itemname(item)
at line 656, 2043

3. arguments change at #obs_scene_add_internal# and #obs_scene_add#
in this file related to changes of: line 890, 1622, 1946, 2011~2016, 2055, 2058, 3170, 3282

at obs.h

1. declaration changes at #obs_scene_add_internal# and #obs_scene_add#
2. new declaration at #obs_sceneitem_get_itemname#, #obs_sceneitem_set_itemname#, #obs_scene_if_duplicated_itemname# and #obs_scene_get_new_itemname#

at source-tree.cpp

1. transform from #obs_source_get_name(item->source)# to obs_sceneitem_get_itemname(item)
at line 42, 437, 454~460, 466~486, 498, 889
*note that line 498 is due to the fact that
renaming source is redirected to renaming "item_name" so ContextBar is not updated in time

2. prevent renaming source itself from renaming label of scene items
at line 539

at window-basic-main-dropfiles.cpp

1. related to arguments change at #obs_scene_add_internal# and #obs_scene_add#

at window-basic-main.cpp

1. transform from #obs_source_get_name(item->source)# to obs_sceneitem_get_itemname(item)

at window-basic-properties.cpp

1. related to arguments change at #obs_scene_add_internal# and #obs_scene_add#

at windows-basic-source-select.cpp

1. property "bool existing" added in AddSourceData at line 27, 180
2. related to arguments change at #obs_scene_add_internal# and #obs_scene_add# at line 121

exeldro:
I think adding an extra name (on the scene item) is not the way to go for this.
In my opinion there should be a check when creating a global audio source, when the name exists pick the same name with the next free number after it or something like that.

sora-blue:
> I think adding an extra name (on the scene item) is not the way to go for this. In my opinion there should be a check when creating a global audio source, when the name exists pick the same name with the next free number after it or something like that.

When it comes to name conflict checking on global audio source, perhaps it is already here in the current version(see screenshot below). Name conflicts for sources is working fine. But the problem is when >1 scene items are using the same source, it's hard to differentiate them without extra names. 

img

exeldro:
In case of multiple references to the same source in a single scene and you want to keep them apart in the sources list I would add the scene item id or something like that to be displayed after the source name, but always keep the source name.

sora-blue:
> In case of multiple references to the same source in a single scene and you want to keep them apart in the sources list I would add the scene item id or something like that to be displayed after the source name, but always keep the source name.

Actually the source name is kept by default after these changes (see line 2011-2017 at obs-scene.c), while leaving room for users to freely edit it to some extent. That's a good idea for simplifying anyway.

WizardCM:
This doesn't really fix the underlying bug, this fixes a side effect of the bug. The behaviour in the bug is specific to the global audio sources - my super uneducated guess is that the scene items aren't properly referencing the global source, **or** we just need to filter out the global sources from Add Existing.

sora-blue:
> This doesn't really fix the underlying bug, this fixes a side effect of the bug. The behaviour in the bug is specific to the global audio sources - my super uneducated guess is that the scene items aren't properly referencing the global source, **or** we just need to filter out the global sources from Add Existing. Either that or the Settings window is doing something incorrectly.

I've tested it with other types of sources like images, and it produces the similar result (See screenshot below, running OBS 27.1.3 64-bit Windows, also through AddExisting). So I guess the problem is not specific to the global audio sources. The multiple scene items are referencing the same global source in fact, thus leading to name conflicts. And if you try to change the source name, all scene items referencing the same source will change in return due to "rename" signal. 

I know the changes are too bold actually, thus expecting a fully discussion to get the issue solved with a better resolution.

img

WizardCM:
Add Existing for normal sources works as it should, they are supposed to share a display name.

sora-blue:
> Add Existing for normal sources works as it should, they are supposed to share a display name. That is intentional.

So I totally mistook the point. Thanks for your kind instructions.

WizardCM:
That's OK! A better example of the issue would be in the Add dialog itself, where "Desktop Audio" or "Mic/Aux" appears multiple times, which (theoretically) should not be possible.

PR #5931, opened 2022-02-08
UI: Correctly style "Chrome OS"

gxalpha:
### Description

The operating system is called "Chrome OS" (sometimes also "chromeOS", but that
is used mostly in pictures), not "ChromeOS". OBS should correctly use that name.

### Motivation and Context

When going through Crowdin to find out why the German translation of
"ChromeOS container" changed from "ChromeOS-Container" to
"Chrome-OS-Container" (the latter actually being more wrong than
the former), I looked at how it actually should be, noticing the English
locale was wrong as well.

Btw, the correct German translation for the new locale is "Chrome OS-Container"

### How Has This Been Tested?

N/A, just a locale change

### Types of changes

### Checklist:

// see above - [ ] The code has been tested.

### DIFF
--- a/UI/data/locale/en-US.ini
+++ b/UI/data/locale/en-US.ini
@@ -107,7 +107,7 @@ AlreadyRunning.Text="OBS is already running! Unless you meant to do this, please
 AlreadyRunning.LaunchAnyway="Launch Anyway"
 
 ChromeOS.Title="Unsupported Platform"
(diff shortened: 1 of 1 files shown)


gxalpha:
This aged like milk.
https://www.theverge.com/tldr/2022/7/14/23219443/google-chromeos-not-chrome-os-branding

PR #5946, opened 2022-02-10
libobs: Fix overflow subtracting unsigned numbers

exeldro:
### Description
Fix overflow subtracting unsigned numbers
When offset_size is greater than size, size would overflow.

### Motivation and Context
Don't like overflowing numbers causing crashes
image
image

### How Has This Been Tested?
On windows 64 bit

### Types of changes

### Checklist:

### DIFF
--- a/libobs/obs-encoder.c
+++ b/libobs/obs-encoder.c
@@ -1086,13 +1086,15 @@ static inline void push_back_audio(struct obs_encoder *encoder,
 				   struct audio_data *data, size_t size,
 				   size_t offset_size)
 {
+	if (offset_size >= size)
+		return;
+
 	size -= offset_size;
 
 	/* push in to the circular buffer */
-	if (size)
-		for (size_t i = 0; i < encoder->planes; i++)
-			circlebuf_push_back(&encoder->audio_input_buffer[i],
-					    data->data[i] + offset_size, size);
+	for (size_t i = 0; i < encoder->planes; i++)
+		circlebuf_push_back(&encoder->audio_input_buffer[i],
+				    data->data[i] + offset_size, size);
 }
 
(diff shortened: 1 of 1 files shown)


exeldro:
I have seen the error 3 times over the past 2 weeks on my debug environment. So rough estimate is 1 in 100 starts I get the error. My debug environment has a lot of plugins that might help causing the the error.

WizardCM:
I assume this is not new to 27.2?

libobs/obs-encoder.c
```diff
@@ -1085,13 +1085,14 @@ static inline void push_back_audio(struct obs_encoder *encoder,
 				   struct audio_data *data, size_t size,
 				   size_t offset_size)
 {
+	if (offset_size >= size)
+		return;
 	size -= offset_size;
```
RytoEX, line 1090: To provide a distinct separation, I'd prefer a blank line between the `if` contents and the following line, or else wrap the `if` contents in curly braces.

RytoEX (changes requested):
Just a simple spacing request.

RytoEX (approved):
Style is good. Seems functionally okay. I'll tentatively approve this based on those points, though further review is welcome.

Could you answer the earlier questions on this PR? Just seeking some clarity to make it easier to test and review.

1. Was this causing problems in practice or in the wild? (seems implied that people have only run into this in Debug builds so far)
2. Is this new to 27.2? Or has it been a problem before 27.2?
3. Is there an easy way to reproduce this?
4. What problems does it cause?

exeldro:
The cases I have seen have all been debug environment and OBS 27.2. I have not found an easy way to reproduce it. The problem is it will try to allocate close to max value of size_t in memory.

PR #5949, opened 2022-02-12
 libobs: Add function to load private sources 

cg2121:
### Description
If a private source is loaded with the obs_load_source function,
it is loaded as a normal source, so add a new function
to load private sources.

### Motivation and Context
Makes it easier to save/load private sources.

### How Has This Been Tested?
Saved/loaded private source.

### Types of changes
- New feature
- Bug fix

### Checklist:


WizardCM:
Does this fix a new bug in 27.2 or something older?

cg2121:
Updated to just add a separate function to load a private source.

Always been a problem, since there was never a way to load a private source from settings, without a bunch a duplicate code.

PR #5951, opened 2022-02-13
obs-vst: Remove submodule in favour of direct merge

WizardCM:
### Description

This deletes the submodule of `plugins/obs-vst` in favour of being included in the main repo.

**IMPORTANT**: Developers will have to delete the `obs-vst` folder **manually** before pulling these commits, otherwise Git will error.

This also removes its custom `.clang-format` file in favour of using the main repo's, alongside the README and License file.

### Motivation and Context

obs-vst is first party, and including it as a submodule introduces a lot of unnecessary headaches.

### How Has This Been Tested?
* Build OBS

### Types of changes

### Checklist:


Conan-Kudo:
Is this okay with the Steinburg header dependency?

gxalpha:
Just as a heads-up when rebasing this against master: The cmake rewrite added a [check for whether the submodule files exist](https://github.com/obsproject/obs-studio/blob/7d07b57993cdd8114afca6df2c4eecc619264447/plugins/CMakeLists.txt#L15-L21) which now can get inlined to just `add_subdirectory(obs-vst)` with this PR.

WizardCM:
Per @gxalpha's comment I've made the relevant additional changes. Would love it if a second person could test this.

I do fear users may potentially have to 
* delete `.git/modules/plugins/obs-vst` 
* run `git config --remove-section submodule.plugins/obs-vst`

Chiitoo:
Without removing the deletion of `plugins/obs-vst` from the patch I'm getting `File plugins/obs-vst is not a regular file -- refusing to patch` (with and without cloning the obs-vst repository).

Other than that, seems to work for me so far.

gxalpha:
Sorry for being unclear, what I meant was just "replace each `check_obs_vst()` call with `add_subdirectory(obs-vst)`". To me at least it looks like obs-vst is disabled on purpose on FreeBSD/OpenBSD, so the differentiation between the operating systems is probably still needed.

RytoEX (changes requested):
Per discussion off-thread, this will need to be two commits: one which deletes the submodule, and a second one which merges the submodule into the tree.

plugins/CMakeLists.txt
```diff
@@ -43,7 +35,6 @@ if(OS_WINDOWS)
   endif()
 
   check_obs_browser()
-  check_obs_vst()
```
tytan652, line 46: Did not get replaced by `add_subdirectory(obs-vst)` under `if(OS_WINDOWS)`.

plugins/CMakeLists.txt
```diff
@@ -79,6 +70,7 @@ elseif(OS_FREEBSD)
   add_subdirectory(vlc-video)
   add_subdirectory(oss-audio)
   add_subdirectory(sndio)
+  add_subdirectory(obs-vst)
 
   obs_status(STATUS "obs-browser plugin not available.")
   obs_status(STATUS "obs-vst plugin not available.")
```
tytan652, line 76: obs-vst is not available on FreeBSD.

plugins/obs-vst/CMakeLists.txt
```diff
@@ -0,0 +1,132 @@
...
+    if(NOT TARGET Qt::${_COMPONENT} AND TARGET
+                                        Qt${FIND_QT_VERSION}::${_COMPONENT})
+
+      add_library(Qt::${_COMPONENT} INTERFACE IMPORTED)
+      set_target_properties(
+        Qt::${_COMPONENT} PROPERTIES INTERFACE_LINK_LIBRARIES
+                                     "Qt${FIND_QT_VERSION}::${_COMPONENT}")
+    endif()
+  endforeach()
+endmacro()
```
tytan652, line 37: The macro was there because the plugin was not directly inside obs-studio repo. I think this can be removed while direct merging this.

@PatTheMav ?

PatTheMav: Yes, it's unnecessary then. You could even argue that plugins that are submodules but only supposed to be built in-tree don't need it anyway.

tytan652: > You could even argue that plugins that are submodules but only supposed to be built in-tree don't need it anyway.

I keep that for #6782.

RytoEX: I have resolved this manually.

RytoEX (approved):

PR #5955, opened 2022-02-13
linux-capture: Fix for pipewire capture leaking texture handles

fzwoch:
Enter the graphics context before calling `gs_texture_destroy()` or else the call will fail causing the texture handles to leak.

### Description

Upon Pipewire session destruction, enter the graphics context when textures are being destroyed.

### Motivation and Context

I was getting leak reports from OBS on shutdown when using Pipewire capture, thought I'd investigate..

### How Has This Been Tested?

Run with and without the change.

Before on OBS Shutdown:
> info: Number of memory leaks: 2

Also in debug build:
> debug: gs_texture_destroy: called while not in a graphics context

With the patch:
> info: Number of memory leaks: 0

### Types of changes

### Checklist:

### DIFF
--- a/plugins/linux-capture/pipewire.c
+++ b/plugins/linux-capture/pipewire.c
@@ -298,8 +298,10 @@ static void destroy_session(obs_pipewire_data *obs_pw)
 	}
 
 	g_clear_pointer(&obs_pw->sender_name, bfree);
+	obs_enter_graphics();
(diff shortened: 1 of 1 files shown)


GeorgesStavracas (approved):
Few words describe the awe and relief that I feel right now. Thank you so much for figuring this one out!

This should be backported to 27.2.

PR #5956, opened 2022-02-13
libobs, docs: Add option to defer source saving

cg2121:
### Description
Currently all sources are saved by the frontend, with no
ability to skip sources, unless they are private. A plugin
might want to add a source to the audio mixer, for example, which
doesn't work with private sources. The defer option is added so
developers can save sources in plugins as they see fit.

### Motivation and Context
Developers might want to save sources with their own code.

### How Has This Been Tested?
Tested with plugin I'm making.

### Types of changes

### Checklist:

### DIFF
--- a/UI/window-basic-main.cpp
+++ b/UI/window-basic-main.cpp
@@ -504,6 +504,9 @@ static obs_data_t *GenerateSaveData(obs_data_array_t *sceneOrder,
 		if (obs_source_is_group(source))
 			return false;
 
+		if (obs_source_saving_deferred(source))
+			return false;
+
 		return find(begin(audioSources), end(audioSources), source) ==
 		       end(audioSources);
 	};
--- a/docs/sphinx/reference-sources.rst
+++ b/docs/sphinx/reference-sources.rst
@@ -1566,6 +1566,12 @@ Functions used by transitions
    with the new transition, then call
    :c:func:`obs_transition_swap_begin()`.
 
+---------------------
+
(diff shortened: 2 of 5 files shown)


tt2468:
Considering the UI is what really is saving the source, perhaps it would be better to set this within the source's private settings?

jp9000:
I don't want to add functions unless someone out there actually, genuinely needs them. We've already added so much stuff that people never really use or have use for. What's the context?

A PR should never be written on the idea that someone might need something. A PR should be written when something is actually needed, and that need should always be specified and elaborated clearly in the pull request description.

cg2121:
My original thoughts when writing this PR:
- In my soundboard plugin, https://github.com/cg2121/obs-soundboard/, I wanted to add a source to the volume mixer, but private sources can't be added because their signals are blocked
- My idea was to add a source save defer function to libobs so I could use a public source while being able to still save/load it myself, as public sources are all saved by the frontend
- The reason why I wanted to save it myself is because I set it to be an output channel, like the global audio devices
- The global audio devices are filtered when saving at https://github.com/obsproject/obs-studio/blob/efce0f41d06062a6f24042d3ba9fcf4eb7ef3219/UI/window-basic-main.cpp#L503, so this would be similar to that
- This could also simplify the saving global audio devices code as well

cg2121:
An alternative idea I thought of was to add a frontend function to add a sources to the mixer.

cg2121:
Closing, as this is not the desired solution.

PR #5962, opened 2022-02-15
obs-ffmpeg: Force mpegts & disable restart on activate for srt/rist

pkviet:
### Description
This fixes crashes in Media Source when :
- the user forgets to set the input format for srt or rist URLs;
- disabling / enabling / disabling etc Media source with a rist URL when the restart_on_activate setting is set to true.
The said setting is therefore set to false for rist URLs.

### Motivation and Context
Crashes are bad.

### How Has This Been Tested?
Tested by playing srt or rist streams in Media Source.
They play fine even if the 'input format field is empty.
Rist streams also play fine when one disables/re-enables the source.

### Types of changes

### Checklist:

### DIFF
--- a/plugins/obs-ffmpeg/obs-ffmpeg-source.c
+++ b/plugins/obs-ffmpeg/obs-ffmpeg-source.c
@@ -386,6 +386,15 @@ static void ffmpeg_source_tick(void *data, float seconds)
 	}
 }
 
+#define SRT_PROTO "srt"
+#define RIST_PROTO "rist"
+
+static bool requires_mpegts(char *path)
+{
+	return !astrcmpi_n(path, SRT_PROTO, sizeof(SRT_PROTO) - 1) ||
+	       !astrcmpi_n(path, RIST_PROTO, sizeof(RIST_PROTO) - 1);
+}
+
 static void ffmpeg_source_update(void *data, obs_data_t *settings)
 {
 	struct ffmpeg_source *s = data;
@@ -406,6 +415,10 @@ static void ffmpeg_source_update(void *data, obs_data_t *settings)
 		input = (char *)obs_data_get_string(settings, "input");
 		input_format =
 			(char *)obs_data_get_string(settings, "input_format");
+		if (requires_mpegts(input)) {
+			input_format = "mpegts";
+			obs_data_set_string(settings, "input_format", "mpegts");
+		}
 		s->reconnect_delay_sec =
 			(int)obs_data_get_int(settings, "reconnect_delay_sec");
 		s->reconnect_delay_sec = s->reconnect_delay_sec == 0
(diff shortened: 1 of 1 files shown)


plugins/obs-ffmpeg/obs-ffmpeg-source.c
```diff
@@ -386,6 +386,15 @@ static void ffmpeg_source_tick(void *data, float seconds)
 	}
 }
 
+#define SRT_PROTO "srt"
+#define RIST_PROTO "rist"
+
+static bool requires_mpegts(char *path)
+{
+	return !strncmp(path, SRT_PROTO, sizeof(SRT_PROTO) - 1) ||
```
RytoEX, line 394: Is this always lowercase at this point, or will uppercase user input still be uppercase when it's checked here?

RytoEX (changes requested):
I think it would be safer for the string comparisons to be case insensitive, unless the strings can only be lowercase at the time of comparison.

plugins/obs-ffmpeg/obs-ffmpeg-source.c
pkviet: The checks don't work for lowercase indeed. But the urls won't be recognized by ffmpeg in uppercase. We have the exact same check in ffmpeg-mux (for srt/rist outputs). I haven't seen any issues as yet with people using uppercase rather than lowercase so i'd say the check is safe.

pkviet: i don't think we do any check at all for strings to be uuper or lower case; it seems complicated to implement (ex: https://stackoverflow.com/questions/5820810/case-insensitive-string-comparison-in-c ) and germane to the problems i'm trying to fix. So i'd say until someone opens an issue because we don't handle correctly his url (be it rtmp, rist srt or else) because it's upper case, I'll leave this as it is

RytoEX: For the record, we have case-insensitive comparison functions.

https://obsproject.com/docs/reference-libobs-util-dstr.html?highlight=astrcmpi#c.astrcmpi

However, if the expectation is that these are always lowercase and would fail in uppercase anyways, then that's another thing entirely, though I don't know if I agree with it.

pkviet: ok swapped the comp function to astrcmp_n.

pkviet:
Updated to use astrcmp_n function to allow comparison of both upper and lowercase.

RytoEX:
astrcmp_n is case sensitive. The case insensitive variant is astrcmpi_n.

pkviet:
thx, it's fixed.

PR #5969, opened 2022-02-16
wayland: Map keymap with MAP_PRIVATE

mahkoh:
### Description

Fixes a bug where keymaps were mapped with MAP_SHARED.

### Motivation and Context

OBS currently segfaults under weston because xkb_state is NULL if mmap fails.

### How Has This Been Tested?

OBS no longer segfaults under weston.

### Types of changes

### Checklist:

### DIFF
--- a/libobs/obs-nix-wayland.c
+++ b/libobs/obs-nix-wayland.c
@@ -84,7 +84,7 @@ static void platform_keyboard_keymap(void *data, struct wl_keyboard *keyboard,
 	UNUSED_PARAMETER(format);
 	obs_hotkeys_platform_t *plat = (obs_hotkeys_platform_t *)data;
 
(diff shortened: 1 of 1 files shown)


kkartaltepe (approved):

mahkoh:
> Right, because this is supposed to be read-only I guess...

The reason is that compositors don't want to create a new file descriptor for each client. Instead they want to create a file descriptor once and then send it to all clients. To make this work, they have to ensure that the clients cannot modify the memory behind the file descriptor. On linux, they do this by creating a file descriptor with `memfd_create` and then sealing it with `F_SEAL_WRITE` to disallow any future write operations. However, the linux kernel does not allow mapping such sealed file descriptors with MAP_SHARED. I do not know why this is but in version 7 of the protocol the MAP_PRIVATE requirement was added for this reason.

PR #5972, opened 2022-02-17
LUT optimizations

jpark37:
### Description
Two optimizations:
- Filter in linear space which is arguably more correct, and simplifies the shader.
- Replace tetrahedral interpolation with trilinear. Just use a bigger LUT.

This has a breaking change in that the Amount multiplier now applies in linear space, but I suspect most users use 1.0, so I don't think it's worth versioning the filter.

### Motivation and Context
Speed.

### How Has This Been Tested?
Linear filter, Intel UHD Graphics 750, 1080p draw: 980 µs -> 860 µs
Trilinear interpolation, Intel UHD Graphics 750, 1080p draw: 860 µs -> 591 µs
Remove domain checks, Intel UHD Graphics 750, 1080p draw: 591 µs -> 532 µs
Alpha passthrough, Intel UHD Graphics 750, 1080p draw: 532 µs -> 486 µs

Tested on D3D11 and OpenGL, and checked image diffs to make sure there weren't major regressions.

### Types of changes

### Checklist:


Nofew:
I've seen at least two people complain about the inverse of this change being done a while back, and both of them requested a toggle to change the behavior. This feels like a wonderful time to mention this, and suggest that such a toggle be added.

Nofew:
I imagine if some people want the old behavior, there may also be some people who want the new behavior as well.

jpark37:
Nah, I haven't heard anyone care either way. I just switched it over on a whim when adding CUBE support because the spec wanted it, but now I just want it to be fast.

Vainock:
Could you describe in more detail what "Passthrough Alpha" does? I would like to provide more context on Crowdin as this is quite technical.

jpark37:
The issue is that we're working with premultiplied colors, so a LUT conversion of the RGB channels can result in an invalid color. The RGB values should not exceed the alpha value. We can avoid this by dividing by alpha, performing the RGB LUT, and then remultiplying the alpha.

This checkbox allows the user to skip these extra calculations, e.g. because they know the incoming colors are opaque, which provides a modest speedup.

SuslikV:
I think, the "Passthrough Alpha" setting should be hidden from the user.
Speed-up optimizations shouldn't depend on user (if they don't affect the quality).
The option shouldn't make incorrect results.

Notes:
1) I don't know if the check for opaque colors will preserve any benefits of the "Passthrough Alpha" optimization.
2) It is possible to transform this setting to "Make the source opaque", if it has any sense here.

PR #5996, opened 2022-02-21
UI: Avoid calling obs_source_update multiple times

kkartaltepe:
### Description

When we added undo/redo in 86eb7aeb a distinction between Updates and
Visual updates was made, presumably for filters which need special
handling of the update case. These functions were also added to regular
source properties, but didnt use any special code instead calling
obs_source_update an additional time after a short time.

Instead lets remove the 2nd delayed call and only call obs_source_update
once per settings change.

### Motivation and Context

some sources perform work during update or log additional information (xcomposite is both) this second
call causes graphical weirdness for the user in this case and extra logging to read through for support.

### How Has This Been Tested?

Undo/redo of xcomposite and v4l2 capture didnt break. Its unclear to me if this change was made to workaround some slow
source which might desync between their visual and actual settings, but if it was I think that should be resolved in the source 
itself.

### Types of changes

### Checklist:

### DIFF
--- a/UI/window-basic-properties.cpp
+++ b/UI/window-basic-properties.cpp
@@ -81,20 +81,10 @@ OBSBasicProperties::OBSBasicProperties(QWidget *parent, OBSSource source_)
 	OBSDataAutoRelease nd_settings = obs_source_get_settings(source);
 	obs_data_apply(oldSettings, nd_settings);
 
-	auto handle_memory = [](void *vp, obs_data_t *old_settings,
-				obs_data_t *new_settings) {
-		obs_source_t *source = reinterpret_cast<obs_source_t *>(vp);
-
-		obs_source_update(source, new_settings);
-
-		UNUSED_PARAMETER(old_settings);
-		UNUSED_PARAMETER(vp);
-	};
-
 	view = new OBSPropertiesView(
 		nd_settings.Get(), source,
 		(PropertiesReloadCallback)obs_source_properties,
-		(PropertiesUpdateCallback)handle_memory,
+		(PropertiesUpdateCallback) nullptr, // No special handling required for undo/redo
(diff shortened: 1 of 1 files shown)


UI/window-basic-properties.cpp
```diff
@@ -81,20 +81,10 @@ OBSBasicProperties::OBSBasicProperties(QWidget *parent, OBSSource source_)
...
-
-		UNUSED_PARAMETER(old_settings);
-		UNUSED_PARAMETER(vp);
-	};
-
 	view = new OBSPropertiesView(
 		nd_settings.Get(), source,
 		(PropertiesReloadCallback)obs_source_properties,
-		(PropertiesUpdateCallback)handle_memory,
+		(PropertiesUpdateCallback) nullptr, // No special handling required for undo/redo
```
WizardCM, line 87: With this change, the 'OK' button now crashes in both the Media Source and Browser Source because there's a `nullptr` check missing on `callback` in `UpdateSettings()` function that calls the callback.

https://github.com/obsproject/obs-studio/blob/45f737fb302810a225b17c9ab7fbec94147ae51e/UI/properties-view.hpp#L191

Unfortunately, adding a `nullptr` check exposes an issue where properties update, but the `Update` function for that source no longer gets called, so the Properties view displays the updated properties but the source doesn't apply the new properties (eg. new URL in Browser is not loaded, even after multiple Properties changes).

jp9000: I'm fixing it at the moment

jp9000: Should be fixed with https://github.com/obsproject/obs-studio/commit/89881147c7f31dd1869f3f877a4b1d342de863cf

PR #5998, opened 2022-02-21
linux-capture: Simplify xcompcap

kkartaltepe:
### Description

This attempts to simplify and remove many helpers from xcompcap capture
methods and rearrange some functionality to fix error handling broken some
time after 4ac77ee (xlocks collided again).

Removes some lock helpers, replaced with a RAII based defer macro to
call the functions directly.

Splits updateSettings into updateTexture to be called from
updateSettings and tick. This also avoids excessive logging when windows
are resized.

Renames xcc_cleanup to cleanupPixmap and moves the non-pixmap related
cleanup directly into XCompCap destructor. This also involves some
changes to texture validity checks to preserve behavior but makes things
more straightforward.

Removes redundant lockX option because XErrorLock already implements
this functionality and is held during the same lifetime.

Removes excessive error checking from cleanup functions. Because of the
earlier xlock collisions our state tracking was incorrect and we saw
errors in these functions. After fixing xlock issues that state tracking works and
those errors go away so lets remove the extra logging as well.

Don't release TexImage on nvidia since we are using loose binding.
Tested via nvidia-smi that resources are released as expected.

### Motivation and Context

Deleting code is good, but primarily this restores the correct errors to our logs again which makes it easier to
tell if/when we introduce regressions. And fixes our error handling that further simplifies the code. Removing
some unused code and codifying our requirements that xlocks not be taken recursively should prevent regression
by crashing new code that breaks this assumption before it reaches a PR, but may be we want this to be a log instead.

### How Has This Been Tested?

Tested on nvidia on i3/gnome and on intel on i3/gnome. Nothing appears to have regressed but by reducing the
log spam it becomes easier to see some underlying errors such as gs_copy_texture failures occurring (due to cross process sync issues?). These copy failures could be replicated on master on intel and nvidia so I don't think this is
a regression (the rate seems more or less the same as master).

### Types of changes

### Checklist:

### DIFF
--- a/plugins/linux-capture/xcompcap-helper.cpp
+++ b/plugins/linux-capture/xcompcap-helper.cpp
@@ -172,7 +172,8 @@ static pthread_mutex_t changeLock = PTHREAD_MUTEX_INITIALIZER;
 
 void registerSource(XCompcapMain *source, Window win)
 {
-	PLock lock(&changeLock);
+	pthread_mutex_lock(&changeLock);
+	DEFER(pthread_mutex_unlock(&changeLock));
 
 	blog(LOG_DEBUG, "registerSource(source=%p, win=%ld)", source, win);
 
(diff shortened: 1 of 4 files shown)


kkartaltepe:
Marked draft, as there is some agreement that a more substantial simplification might be better.

RytoEX:
Should we close this in favor of #6091? Or is there a particular reason to leave this open?

kkartaltepe:
I was just keeping it for reference of a less invasive option, ill close it though as there is no real reason to leave it in draft.

PR #6001, opened 2022-02-21
UI: Restore LC_NUMERIC to C locale on Mac/Linux

### LINKED ISSUE #3770: Muxer/Encoder settings containing decimal points should NOT respect system locale
## Platform

Operating system and version: elementary OS 5.1.4
OBS Studio version: 26.0.0 - installed from PPA

## Expected Behavior
Muxer/encoder settings that take decimals (such as `hls_time`) should always use `.` as a decimal separator

## Current Behavior
On a system with locale `LC_NUMERIC=es_ES.UTF-8`, decimals are expected to use `,` - otherwise, an "Invalid argument" error is shown.

```21:54:34.617: Using muxer settings: 
21:54:34.617: 	hls_time=0.5
21:54:34.617: 	hls_init_time=0.5
21:54:34.617: 	hls_list_size=3
21:54:34.617: Error opening '/home/granpc/hls/wtf-locales.m3u8': Invalid argument
21:54:34.618: ffmpeg_data_init failed
21:54:34.619: ==== Recording Stop ================================================
```

Using the local decimal separator (`,`) causes a corrupted file to be written.

```#EXTM3U
#EXT-X-VERSION:3
#EXT-X-TARGETDURATION:1
#EXT-X-MEDIA-SEQUENCE:2
#EXTINF:0,500000,
wtf-locales2.ts
#EXTINF:0,500000,
wtf-locales3.ts
#EXTINF:0,500000,
wtf-locales4.ts
```

Notice how EXTINF is using comma as a decimal separator. This kills the HLS file.

## Steps to Reproduce

1. Set your system locale to any of the places in the world that don't use dots as a decimal separator, such as Spain
2. Configure OBS to record to an HLS file with `hls_time` set to a value with decimal places:
2.1. Settings > Output > Recording
2.2. Set type to "Custom Output (FFmpeg)" and output to a file
2.3. Set container format to `hls`
2.4. Set Muxer Settings to `hls_time=0.5 hls_init_time=0.5 hls_list_size=3`
2.5. Any other parameters can be left as default
3. Attempt to initiate the recording
4. Watch as it immediately errors out

## Workaround

Move to another country. Or alternatively, run OBS with `LOCALE=C LC_NUMERIC=C LANG=C LANGUAGE=C obs`.

### LINKED ISSUE #5385: .cube LUTs not working in Linux
### Operating System Info

Other

### Other OS

Manjaro Cinnamon

### OBS Studio Version

27.1.3-3 (most recent)

_No response_

### OBS Studio Log URL

`see below`

### OBS Studio Crash Log URL

_No response_

### Expected Behavior

Working LUT

### Current Behavior

Not working, nothing happens.
.png LUTS work fine though.

### Steps to Reproduce

1.Select source
2.Add filter 'LUT'
3.Select .cube LUT file

In my case i also copied the LUT files to the specific OBS folder (obs-filter/LUTs) to see if that was the problem, but to no avail.

_No response_

jpark37:
### Description
Suspicion is that floats were not being parsed correctly when a foreign locale is set.
- CUBE files now load correctly.
- HLS arguments to custom FFmpeg now parse correctly.

The Qt recommendation is to set LC_NUMERIC to "C" in the QCoreApplication doc:
https://doc.qt.io/qt-5/qcoreapplication.html
> On Unix/Linux Qt is configured to use the system locale settings by default. This can cause a conflict when using POSIX functions, for instance, when converting between data types such as floats and strings, since the notation may differ between locales. To get around this problem, call the POSIX function setlocale(LC_NUMERIC,"C") right after initializing [QApplication](https://doc.qt.io/qt-5/qapplication.html), [QGuiApplication](https://doc.qt.io/qt-5/qguiapplication.html) or QCoreApplication to reset the locale that is used for number formatting to "C"-locale.

### Motivation and Context
Fixes #3770
Fixes #5385

### How Has This Been Tested?
- Verified broken -> fixed on Ubuntu with `LANG=es_ES.UTF-8` for both issues.
- Verified broken -> fixed on Mac with `LANG=es_ES.UTF-8` for CUBE load, and no regression for HLS arguments.
- Scene JSON does not seem to care what the locale is.

### Types of changes

### Checklist:

### DIFF
--- a/UI/obs-app.cpp
+++ b/UI/obs-app.cpp
@@ -1156,6 +1156,12 @@ bool OBSApp::InitTheme()
 OBSApp::OBSApp(int &argc, char **argv, profiler_name_store_t *store)
 	: QApplication(argc, argv), profilerNameStore(store)
 {
+	/* fix float handling */
+#if defined(Q_OS_UNIX)
+	if (!setlocale(LC_NUMERIC, "C"))
+		blog(LOG_WARNING, "Failed to set LC_NUMERIC to C locale");
+#endif
+
 	sleepInhibitor = os_inhibit_sleep_create("OBS Video/audio");
 
 #ifdef __APPLE__
--- a/UI/platform-x11.cpp
+++ b/UI/platform-x11.cpp
@@ -213,7 +213,6 @@ string GetDefaultVideoSavePath()
 
 vector<string> GetPreferredLocales()
 {
-	setlocale(LC_ALL, "");
 	vector<string> matched;
 	string messages = setlocale(LC_MESSAGES, NULL);
 	if (!messages.size() || messages == "C" || messages == "POSIX")


plugins/obs-filters/color-grade-filter.c
```diff
@@ -183,6 +189,21 @@ static void *load_cube_file(const char *const path, uint32_t *const width,
...
 {
+#ifndef _WIN32
+	locale_t locale = newlocale(LC_ALL_MASK, "C", 0);
+	locale_t previous = 0;
+	if (locale) {
+		previous = uselocale(locale);
+		if (!previous) {
+			blog(LOG_WARNING, "uselocale: Failed to use C locale");
+			freelocale(locale);
+			locale = 0;
```
kkartaltepe, line 200: freelocale could be done in the last block only conditioned on if(locale) if we remove this early free?

jpark37: Updated.

kkartaltepe: The else around the free in the last block can be removed too, since we dont do the early free.

jpark37: I don't think it can be removed. If we can't dislodge the locale, we would be freeing an active locale.

kkartaltepe: Hmm looks like you are right, I see why you did the early free now.

jpark37: The early free was just some conservative code I wrote early on. It's not necessary.

I don't like writing code that can leak, but it's the lesser of evils, and also very unlikely to happen.

jpark37:
Marked as draft. Planning to redo this change using Qt's recommendation:
> On Unix/Linux Qt is configured to use the system locale settings by default. This can cause a conflict when using POSIX functions, for instance, when converting between data types such as floats and strings, since the notation may differ between locales. To get around this problem, call the POSIX function setlocale(LC_NUMERIC,"C") right after initializing [QApplication](https://doc.qt.io/qt-5/qapplication.html), [QGuiApplication](https://doc.qt.io/qt-5/qguiapplication.html) or QCoreApplication to reset the locale that is used for number formatting to "C"-locale.

That explains why we weren't seeing this on Windows/Mac. Dunno why Qt is treating Unix/Linux special though.

kkartaltepe:
I'm good with using C for LC_NUMERIC everywhere too, if we do that we can probably mark https://github.com/obsproject/obs-studio/issues/3770 as fixed by this as well.

jpark37:
K, I'll try to test 3770 as well.

jpark37:
Could end up breaking user profile/scene setups though, ugh. We'll see.

jpark37:
Both bugs are addressed, and the locale doesn't seem to affect saved JSON settings. Should probably merge this PR before a long beta for test coverage.

majkrzak:
As it is a QT application, shouldn't [QLocale::setDefault](https://doc.qt.io/qt-5/qlocale.html#setDefault) be used instead? Also isn't it just workaround for poor/lack of parsers implementation?

jpark37:
`QLocale::setDefault` doesn't affect the C runtime locale, which is the one we want to control.

I don't think we want to add a custom sscanf implementation if the standard function is good enough.

jpark37:
It doesn't seem like LC_ALL overrides LC_NUMERIC if there's an explicit LC_NUMERIC set. At least, this change wouldn't work if that was the case, but it does work. None of the setlocale man pages seem to guarantee one way or the other though.

majkrzak:
In glibc sources it looks like it will override all entries. https://sourceware.org/git/?p=glibc.git;a=blob;f=locale/setlocale.c;h=38e9bec6bbb61eaa049e36b5ad6e9ffc5c2c6f44;hb=HEAD#l279
Also quick test confirms it:
```c
#include <locale.h>
#include <stdio.h>

int main() {
    setlocale(LC_ALL, "C");
    printf("%f\n",1.0); // .
    setlocale(LC_ALL, "");
    printf("%f\n",2.0); // ,    
    setlocale(LC_ALL, "C");
    printf("%f\n",3.0); // .
    setlocale(LC_NUMERIC,"pl_PL.UTF_8");
    printf("%f\n",4.0); // ,
    setlocale(LC_ALL, "C");
    printf("%f\n",5.0); // .
}
```

jpark37:
Blah. I thought I retested both platforms, but I probably only retested Mac. I'll convert that one case to uselocale.

jpark37:
Actually, I verified this fix works on Ubuntu as-is although it doesn't seem like it should. Still digging.

jpark37:
I had to remove `Language=English` from my global.ini to take the code path, and reproduce the issue. Fixed by removing the extra `setlocale` call. I didn't see any other uses of `setlocale` in OBS code.

PR #3724, opened 2020-11-12
UI: Fix tray behaviour and app icon update for macOS

PatTheMav:
### Description
Adds visual fixes for macOS in general and for macOS Big Sur specifically:

* New AppIcon follows Apple's new interface guidelines for Big Sur: Neumorphism and Squircles
* Made Qt use the AppIcon as well
* New mask-based icons for the tray icon
* Fix tray icon handling in macOS

### Motivation and Context
AppIcon: I whipped up an icon real quick and added some vibrancy to it because the base colour made it to dull and looked too muted especially in dark mode. There were also reports about Qt not using the same icon as macOS does, this has also been fixed with this PR.

Mask-based icons: Tray icons in macOS are supposed to be vectors or masks as macOS will tint them white or black depending on the UI setting (dark mode/bright mode). This PR adds icon variants for this, including the record and pause states. I had to reverse engineer the icon to create shapes from it, maybe someone can build "official" variants using the source icon.

Tray icon handling: Qt seems to get a `trigger` event on a menubar icon even if a `ContextMenu` is set - this meant that on macOS the main window could be hidden/shown by clicking on the tray icon, but the context menu was never usable. This change disables the one-click functionality and restores the menu.

*Note:* The reported issue of the icon only disabling preview is due to plugins attaching modal windows to OBS' main window, which leads to a failed check necessary to trigger the window status.

### How Has This Been Tested?
* OBS compiled and run locally, icons tested in Dock and Task Switcher, as well as Launchpad. 
* Tray icons tested by running OBS, starting and pausing recording, observing the icons.

### Types of changes
- Bug fix (non-breaking change which fixes an issue)
- New feature (non-breaking change which adds functionality)

### Checklist:
- [x] My code has been run through [clang-format](https://github.com/obsproject/obs-studio/blob/master/.clang-format).
- [x] I have read the [**contributing** document](https://github.com/obsproject/obs-studio/blob/master/CONTRIBUTING.rst).
- [x] My code is not on the master branch.
- [x] The code has been tested.
- [x] All commit messages are properly formatted and commits squashed where appropriate.
- [x] I have included updates to all appropriate documentation.

Issue #3093, opened 2020-06-23
Dock icon is overriden on Mac by setWindowIcon


tbodt:
## Platform

Operating system and version: macOS 10.15.6
OBS Studio version: 25.0.8-1-gc57481bcb-modified

## Expected Behavior
The icon used in the dock should be the obs.icns icon.

## Current Behavior

When OBS starts up it calls [setWindowIcon](https://github.com/obsproject/obs-studio/blob/5a6b417fcc9718573f6946c1a972cd02ddd04bf2/UI/obs-app.cpp#L1109) and this [also overrides the dock icon](https://doc.qt.io/qt-5/macos-issues.html#dock), leading to a noticeable "jump" in the dock icon because the two icons are not quite the same.

## Steps to Reproduce

1. Start OBS on Mac

unknowndomain:
Great spot, personally I can't see any need for those calls on macOS due to how the OS works and presume the Windows bias of the development probably resulted in people not noticing as it also requires you to have the application saved in the dock or in recent applications on Catalina.

I've modified the code in my commit to just not change them as I presume the icons are different for visiblity reasons at small sizes on the Windows start bar and title bars.

PatTheMav:
@unknowndomain Did you submit a PR for this? I've noticed the same "jumpy" behaviour but didn't have time to look into it.

unknowndomain:
The issue was that the icon of the app bundle was different to the icon set at start up, the solution I proposed was the #ifdef for macOS all calls to set the icon however people complained this was break the ability for themes to set the icon, why you would do that this way on macOS is beyond me but that was why the PR was closed without merging it.

Someone else was going to replace all the icons to be consistent.

PatTheMav:
I see - yeah in general I agree with you, there should be no themes on macOS IMO (instead there should be a bright and dark mode variant and support for upcoming custom highlights in macOS Big Sur), but that's non-essential.

Fixing the icons for the default skin(s) should be good enough.

unknowndomain:
I don't really mind the themes, but changing the icon in the theme seems unnecessary and not normal of Mac applications, but to be honest I hadly ever use the Dock on my Mac so I don't really care about the issue either.

I mean I can't help but feel the future of OBS on macOS could be very limited anyway, there are a small number of features that don't work or exist, the UI is nothing like any other macOS app and Apple Silicon might be a nail in the coffin, that would be very sad because I love my Mac but to be honest OBS on Mac is not a great experience in my view.

[...]

PR #6009, opened 2022-02-22
Mac os icon patch

MisutaaAsriel:
### Description
macOS **default behavior** is to set the application dock icon _to the one listed in the App's information PLIST_, OR **to a user provided ICNS file,** if added via an app's info pane in finder. This change remedies unexpected behavior regarding the OBS dock icon.

### Motivation and Context
Manually setting the icon by code overrides default macOS behavior, causing the icon to revert to what was set by either the information PLIST, or by the user using Finder. This can cause the icon to not only "revert" to the default OBS icon during launch, when a user provided icon is set, but if the image file and ICNS were to ever differ, this would cause OBS to use the incorrect icon.

Previously #3093 mentioned a similar issue, with `setWindowIcon` causing non-standard dock icon behavior, and #3724 _purported_ to fix this. However, #3724 introduced _yet-another_ non-standard behavior which this pull seeks to remedy.

Solution found is to do **_nothing_** if on Apple platforms when setting the application icon.

#### Testing Environment:
- OS: macOS Monterey
- Hardware: MacBookPro14,1 / 128GB Storage / 16GB RAM

#### Tests:
1. Opening the application
2. Opening various dialogs and windows within OBS.
3. Closing OBS
4. Setting a custom icon via Finder
    - Open finder
    - Highlight `OBS.app`
    - `CMD + I` (or "_Get Info_" from the context menu)
    - Find an `ICNS` file
    - Drag `ICNS` into `Info` pane for OBS, in top-left
    - Drop Icon
5. Opening OBS with custom icon applied.

#### Results
- OBS opens normally.
- Dock bounce behavior unchanged.
- Custom icon remains, when applied.

### Types of changes

### Checklist:

    - I added a comment where the code used to be, explaining the change.


Fenrirthviti:
The commit title needs to have the proper prefix, as well as the commits being squashed.

Please update before this can be merged.

MisutaaAsriel:
> The commit title needs to have the proper prefix, as well as the commits being squashed.
> 
> Please update before this can be merged.

Fixed, AFAIK.

RytoEX (changes requested):
Please wrap your commit message body at 72 characters, per our [Commit Guidelines in our Contributing Guidelines](https://github.com/obsproject/obs-studio/blob/b3870476324e2dc79a9d30371a1199376bcd0ad7/CONTRIBUTING.rst#commit-guidelines).

Per our Commit Guidelines, please use an appropriate commit message prefix.  In this case, the prefix should be "UI: ".  Please note that the commit message is different from the PR title.

MisutaaAsriel:
> Please wrap your commit message body at 72 characters, per our [Commit Guidelines in our Contributing Guidelines](https://github.com/obsproject/obs-studio/blob/b3870476324e2dc79a9d30371a1199376bcd0ad7/CONTRIBUTING.rst#commit-guidelines).
> 
> Per our Commit Guidelines, please use an appropriate commit message subject prefix. In this case, the prefix should be "UI: ". Please note that the commit message is different from the PR title.

Apologies I misunderstood the request earlier.

Granted, I had to remove most of the explanation of the commit itself, due to character count, and was recommended to also remove the code comment explaining it as well. Hope that's ok.

RytoEX:
> > Please wrap your commit message body at 72 characters, per our [Commit Guidelines in our Contributing Guidelines](https://github.com/obsproject/obs-studio/blob/b3870476324e2dc79a9d30371a1199376bcd0ad7/CONTRIBUTING.rst#commit-guidelines).
> > Per our Commit Guidelines, please use an appropriate commit message subject prefix. In this case, the prefix should be "UI: ". Please note that the commit message is different from the PR title.
> 
> Apologies I misunderstood the request earlier.
> 
> Granted, I had to remove most of the explanation of the commit itself, due to character count, and was recommended to also remove the code comment explaining it as well. Hope that's ok.
> 
> Let me know if I'm missing anything else. I tried following guidelines to best of my ability but I clearly missed some things.
> 
> _~This is my first PR, so I apologize for the mistakes.~_

"Wrap at 72 characters" means wrap the text/line so that each line is no more than 72 characters.

Instead of this:
> macOS default behavior is to set the application dock icon to the one listed in the App's info PLIST, OR to a user provided ICNS file, if added via an app's info pane in finder. Manually setting the icon by code overrides this behavior, causing the icon to revert to what was set here.
> 
> Solution is to do nothing if on Apple platforms when setting the application icon.

You would do this:
> macOS default behavior is to set the application dock icon to the one 
> listed in the App's info PLIST, OR to a user provided ICNS file, if 
> added via an app's info pane in finder. Manually setting the icon by 
> code overrides this behavior, causing the icon to revert to what was set 
> here. 
> 
> Solution is to do nothing if on Apple platforms when setting the 
> application icon.

MisutaaAsriel:
> "Wrap at 72 characters" means wrap the text/line so that each line is no more than 72 characters.

Apologies; the guidelines _do not make that clear [to me]_. (To be honest, they are a bit _obtuse_ for first-timers.)

> Instead of this:
> 
> >...
> 
> You would do this:
> 
> > ...

I did something similar just now, only better formatted, to prevent strange line-breaks. Comment is 72 characters or less per line, and the title comes in at 43 characters.

> Fixes overriding of macOS dock icon behavior.
> 
> macOS sets the icon on launch based on:
> - The Information PLIST
> - A user provided icon by Finder
> 
> setWindowIcon overrides this behavior.
> 
> Therefore, this commit resolves this issue,
> by doing nothing if target is macOS.

Is there any other amendments to be made? I did read the contribution guidelines, but as aforementioned, some of it is difficult to parse (apparently), even for the technically inclined.

MisutaaAsriel:
Since I'm a first-timer, I can't run the workflows; should I re-request review for that, or— ?

PatTheMav:
> Previously https://github.com/obsproject/obs-studio/issues/3093 mentioned a similar issue, with setWindowIcon causing non-standard dock icon behavior, and https://github.com/obsproject/obs-studio/pull/3724 purported to fix this. However, https://github.com/obsproject/obs-studio/pull/3724 introduced yet-another non-standard behavior which this pull seeks to remedy.

That is not correct, https://github.com/obsproject/obs-studio/pull/3724 ensured that the ICNS file (used by macOS for launchpad and in file listings) and the icon used by Qt have the same appearance. It was never intended to rely on the AppIcon file alone.

I'm not opposed to removing it, though I'd rather have @Warchamp7 give their feedback on whether they are fine with the icon being entirely separated from Qt's theming.

PatTheMav (approved):
I just checked the Qt docs and it seems that setting the AppIcon using `setWindowIcon` is not the correct way to set the icon, but instead using the AppIcon system as specified by Apple's App Bundle specs.

MisutaaAsriel:
> That is not correct, https://github.com/obsproject/obs-studio/pull/3724 ensured that the ICNS file (used by macOS for launchpad and in file listings) and the icon used by Qt have the same appearance. It was never intended to rely on the AppIcon file alone.

It was a relative statement; #3093 was an issue relating to the dock icon and non-uniform behavior, which at the root was related to the same bug as this pull fixes; setting it to a 256x256 version, as was implemented in #3724, likely only masked this bug, as that is the resolution the dock uses from the ICNS, assumedly, so nothing "changed" as the icon being set was now "identical".

Sorry if I wasn't clear on that. I'm aware it was never _intended_ to let macOS set the icon; the implication was [supposed to be] that while #3724 fixed the _symptoms_ of #3093, it didn't actually fix the core problem.

PatTheMav:
> > That is not correct, #3724 ensured that the ICNS file (used by macOS for launchpad and in file listings) and the icon used by Qt have the same appearance. It was never intended to rely on the AppIcon file alone.
> 
> It was a relative statement; #3093 was an issue relating to the dock icon and non-uniform behavior, which at the root was related to the same bug as this pull fixes; setting it to a 256x256 version, as was implemented in #3724, likely only masked this bug, as that is the resolution the dock uses from the ICNS, assumedly, so nothing "changed" as the icon being set was now "identical".
> 
> Sorry if I wasn't clear on that. I'm aware it was never _intended_ to let macOS set the icon; the implication was [supposed to be] that while #3724 fixed the _symptoms_ of #3093, it didn't actually fix the core problem.

All good, I approved the PR already - it was an oversight on our end that (at least according to Qt's docs) it should've never been implemented that way.

**Do note** that without bundling OBS into an application bundle (which does not happen by default until the Universal Builds PR is merged), there will be no bundle and as such OBS will have no icon in the dock.

PR #6021, opened 2022-02-24
win-dshow: Add hardware decode toggle

jpark37:
### Description
Off by default because it can be buggy, or overburden the GPU.

Removal of logic before calling `ffmpeg_decode_init` looks like a breaking change, but it's not. Hardware decoding was not enabled at all even if you passed true.

### Motivation and Context
Decoding video streams on the CPU is very expensive.

### How Has This Been Tested?
Elgato 4K60 S+ in 4K H264 mode on AMD 3950X: 9% -> 2% CPU
Elgato 4K60 S+ in 4K HDR HEVC mode on AMD 3950X: 20% -> 3% CPU

Verified checkbox toggle works on the fly, and localization string appears in dialog.

### Types of changes

### Checklist:


WizardCM:
"Buggy" how? Would it be worth documenting somewhere?

jpark37:
https://github.com/obsproject/libdshowcapture/pull/41#issuecomment-1047392511
> H265 seems work with nvidia gpu, but decoding h264 will crash inside ffmpeg. Still haven't figure it out why.

Maybe he did something wrong, but maybe it's real.

jpark37:
I also kept CUDA out of the decoder list because of this commit: 86572935afda66442ea2a24c47d1dcca21185d97
> Cuda appears to be a bit buggy and can result in corruption where as
it's not being seen with other hardware decoders.  Remove cuda for now
as a hardware decoder to prevent the possibility of corruption.

PR #6025, opened 2022-02-24
UI: Refresh edit menu on item locked signal

### LINKED ISSUE #6024: Locked out of source resize
### Operating System Info

Windows 10

### Other OS

_No response_

### OBS Studio Version

27.2.1

_No response_

### OBS Studio Log URL

https://obsproject.com/logs/bo9SSBcO0JpwcC-k

### OBS Studio Crash Log URL

_No response_

### Expected Behavior

When I unlock a source, I can Right click -> Transform -> fit to screen, reset transform..

### Current Behavior

When I unlock a source, I Right click -> Transform ->fit to screen, reset transform.. and it's GREYED OUT, like when it's locked. Sometimes it works, most of the time it's greyed out. Sometimes, even when the source is locked, these options are NOT greyed out, but they don't do anything if I click them. This is very weird.
I was running OBS 27.1.3 until I updated to this version, and I had never had this problem before.

I installed and older version, but now I get a message from Twitch saying that I must run this latest version 27.2.1 to be able to connect my account. Why??

### Steps to Reproduce

1.Unlock source.
2.Lock source.
3.Unlock source
...

transform disabled when source is UNlocked
transform enabled when source is locked

gxalpha:
### Description

Makes sure that the edit menu gets refreshed when changing the lock state of a source, enabling/disabling options like
Reset Transform. This includes the right click menu, the Edit menu and Hotkeys.

### Motivation and Context

Fix #6024

### How Has This Been Tested?

Locking and then right clicking a source now has the Reset Transform menus etc correctly disabled (and enabled for unlocking).

### Types of changes

### Checklist:

### DIFF
--- a/UI/source-tree.cpp
+++ b/UI/source-tree.cpp
@@ -520,6 +520,7 @@ void SourceTreeItem::VisibilityChanged(bool visible)
 void SourceTreeItem::LockedChanged(bool locked)
 {
 	lock->setChecked(locked);
+	OBSBasic::Get()->UpdateEditMenu();
 }
 
 void SourceTreeItem::Renamed(const QString &name)


cg2121:
Wouldn't it be better to use the aboutToShow signal for the menu to update it, so it would be in one place?

gxalpha:
No, that is what caused the Hotkey issue in the first place and would just bring it back.

PR #6026, opened 2022-02-24
Fix building OBS Studio on FreeBSD

obiwac:
### Description

Fix building OBS Studio on FreeBSD:
- `libobs/graphics/graphics.h`: The `gs_texture_create_from_dmabuf` & `gs_query_dmabuf_*` functions were marked as Linux-only as of f7a55f45fdacf9c8383cafae8170294ac8997954.
- `plugins/linux-v4l2/v4l2-output.c`: `versionsort` (and by extension, `__strverscmp`) are not available on FreeBSD, as they are GNU extensions (`versionsort` is used by `virtualcam_start` to read `/dev/video*` devices in the right order), so implemented them.
- `UI/platform-x11.cpp`: Was calling `to_string` on FreeBSD when not `using namespace std;`, so calling `std::to_string` instead.

### Motivation and Context

Fixes building on FreeBSD.

### How Has This Been Tested?

Tested on FreeBSD 13.0-RELEASE-p4 and FreeBSD-CURRENT, hardware is non-applicable.
Compiled with clang version 11.0.1 & 13.0.1.

### Types of changes

### Checklist:

Perhaps this is usererror, but `git-clang-format` doesn't seem to work?

```
YAML:35:21: error: unknown key 'AfterExternBlock'
  AfterExternBlock: false
                    ^~~~~
Error reading /usr/home/obiwac/obs-studio/.clang-format: Invalid argument
error: `clang-format -lines=178:178 -lines=187:187 -lines=191:191 -lines=196:196 -lines=198:255 -lines=258:258 -lines=260:260 plugins/linux-v4l2/v4l2-output.c` failed
```

So I attempted to format things myself.
Hope that's alright :)

### DIFF
--- a/plugins/linux-v4l2/v4l2-output.c
+++ b/plugins/linux-v4l2/v4l2-output.c
@@ -185,7 +185,14 @@ static bool virtualcam_start(void *data)
 			return false;
 	}
 
-	n = scandir("/dev", &list, scanfilter, versionsort);
+	n = scandir("/dev", &list, scanfilter,
+#if defined(__linux__)
+		    versionsort
+#else
+		    alphasort
+#endif
+	);
+
 	if (n == -1)
 		return false;
 
--- a/UI/platform-x11.cpp
+++ b/UI/platform-x11.cpp
@@ -136,7 +136,7 @@ const char *RunOnce::thr_name = "OBS runonce";
 void PIDFileCheck(bool &already_running)
 {
 	std::string tmpfile_name =
-		"/tmp/obs-studio.lock." + to_string(geteuid());
+		"/tmp/obs-studio.lock." + std::to_string(geteuid());
 	int fd = open(tmpfile_name.c_str(), O_RDWR | O_CREAT | O_EXLOCK, 0600);
 	if (fd == -1) {
 		already_running = true;
--- a/libobs/graphics/graphics.h
+++ b/libobs/graphics/graphics.h
@@ -936,7 +936,7 @@ EXPORT gs_stagesurf_t *gs_stagesurface_create_nv12(uint32_t width,
 EXPORT void gs_register_loss_callbacks(const struct gs_device_loss *callbacks);
 EXPORT void gs_unregister_loss_callbacks(void *data);
 
-#elif __linux__
+#elif defined(__linux__) || defined(__FreeBSD__)
 
 EXPORT gs_texture_t *gs_texture_create_from_dmabuf(
 	unsigned int width, unsigned int height, uint32_t drm_format,


kkartaltepe:
I don't think we will want to include an implementation of versionsort from glibc just for freebsd. Is there an alternative function that can be used on freebsd?

obiwac:
> I don't think we will want to include an implementation of versionsort from glibc just for freebsd. Is there an alternative function that can be used on freebsd?

Yeah, it's not ideal indeed, but again, to my knowledge and after a good look, I don't think there is.
There does exist `alphasort` though, as can be seen in the [`scandir(3)`](https://www.freebsd.org/cgi/man.cgi?query=scandir&sektion=3&format=html) manual, but that means differing behaviour on different platforms.

Also, `alphasort` is not a good solution as it would order devices as follows:

```
/dev/video0
/dev/video1
/dev/video10
/dev/video2
...
/dev/video9
```

kkartaltepe:
> Also, alphasort is not a good solution as it would order devices as follows:

It seems a decent solution to enable building on freebsd without introducing a large amount of platform specific code. You could consider leaving the results unsorted on freebsd as well perhaps the devices will simply be in the proper order for most users.

WizardCM:
From CI:
```
clang-format-12 is /usr/bin/clang-format-12
=================================
Files were not formatted properly
plugins/linux-v4l2/v4l2-output.c
```

obiwac:
> You could consider leaving the results unsorted on freebsd as well perhaps the devices will simply be in the proper order for most users.

Can confirm this is not the case, unfortunately :(

Something like this could always be used:

```c
static int devsort(const struct dirent **_a, const struct dirent **_b) {
	int a = atoi((*_a)->d_name + 5);
	int b = atoi((*_b)->d_name + 5);

	return a - b;
}
```

This would eliminate the need for any platform specific code whatsoever, as it works both on Linux & FreeBSD. (Strings aren't sorted if not numeric.)

> From CI:
> 
> ```
> clang-format-12 is /usr/bin/clang-format-12
> =================================
> Files were not formatted properly
> plugins/linux-v4l2/v4l2-output.c
> ```

That's my bad, still figuring out how to use this `clang-format` stuff.

obiwac:
Implemented that function I mentioned in my previous comment. Is it better like that or should I just choose between `alphasort` on FreeBSD and `versionsort` on Linux?

obiwac:
Sure, I was operating under that assumption initially, and from that point of view I don't think anything bad could happen.

Regardless, I'll replace it with simple `alphasort`/`versionsort` depending on platform in a bit.

obiwac:
Done

plugins/linux-v4l2/v4l2-output.c
```diff
@@ -1,5 +1,6 @@
 #define _GNU_SOURCE
 
+#include <ctype.h>
```
kkartaltepe, line 3: I dont think this is needed anymore?

obiwac: Indeed, my bad

obiwac:
Yup, of course

obiwac:
There we go!

kkartaltepe (approved):
You squashed the to_strig fix into the v4l2 fix if you could split that back out or make note in the commit desc.

obiwac:
There we go.

PR #6030, opened 2022-02-25
UI: Add profile/scene collection rename events

cg2121:
### Description
Events make it more clear when profiles or scene collection are renamed.

### Motivation and Context
Found issue when making a plugin.

### How Has This Been Tested?
Tested with plugin I've made.

### Types of changes

### Checklist:


WizardCM:
Does the same change need to be made for renaming profiles?

As this was introduced in 27.2, I'd be happy marking this as a bug fix for a 27.2 patch.

tt2468:
I disagree. Since scene collections are only defined by their names, removing this event would make it very difficult to statefully track what's going on.

At the very least, I would suggest adding a new event like `SCENE_COLLECTION_RENAMED` to replace this functionality.

cg2121:
@WizardCM profiles were fixed with https://github.com/obsproject/obs-studio/commit/2720e6d00fb6600f16714700acd24f980157b694

gxalpha:
As far as I can tell, the events in #5945 were not removed, just deduplicated. They are still being emitted.

cg2121:
@tt2468 I think that what's the LIST_CHANGING events are for.

cg2121:
`@WizardCM profiles were fixed with https://github.com/obsproject/obs-studio/commit/2720e6d00fb6600f16714700acd24f980157b694`
Never mind with this.

I've updated to include profiles as well.

tytan652:
> Don't call the changing and changed event when renaming a scene
> collection or profile, as they are not changed.

The path and name of the profile get changed and a way to notify about this to plugins is necessary to avoid issues.

Maybe `_RENAMING` and `_RENAMED` events.

cg2121:
I've now added events for when profiles and scene collections are renamed.

tt2468:
Looks good to me

Issue #4926, opened 2021-06-21
plugins/linux-v4l2/v4l2-input.c issues negative select timeouts


Mikael-Lovqvist:
### Operating System Info

Other

### Other OS

Arch Linux

### OBS Studio Version

27.0.1

### OBS Studio Log URL

Log file is 96 gigabytes and OBS removed it before I realized I had to back it up

### Expected Behavior

I expect the behavior in Current Behavior to not happen.

### Current Behavior

When using the v4l2-input plugin the log file is filled up quickly with the following two lines
```
07:53:56 AM.045: v4l2-input: /dev/video0: failed to log status
07:53:56 AM.045: v4l2-input: /dev/video0: select timed out
```
This happens 131.000 times per second on my system, caused me to get a 96 gigabyte log file.
I traced the problem to https://github.com/obsproject/obs-studio/blob/master/plugins/linux-v4l2/v4l2-input.c#L175-L179 where `fps_num`and `fps_denom`are both zero. This causes `ffps` to be `NaN` which I think causes the `select` call to misbehave.

### Steps to Reproduce

1. Start OBS 
2. Use the v4l2-linux plugin
3.
...

### Anything else we should know?

I did see that someone else had this issue on the [forum](https://obsproject.com/forum/threads/huge-amount-of-error-logs-when-camera-doesnt-return-framerate.145191/).

I am not sure how to properly make a pull request but I did make a solution that works for me here https://github.com/Mikael-Lovqvist/obs-studio/commit/3cd5742be5168f625558c14c1b38fedf23116277

If `framerate` is 0 during `v4l2_init` I  set it to a fallback framerate. This works for me and it does not limit the framerate from the camera as far as I can tell.

If logs are truly needed I can reinstall the version I did not modify on my system to recreate them.

tt2468:
Outside of the logging spam, does the source work as expected when this happens?

Mikael-Lovqvist:
> Outside of the logging spam, does the source work as expected when this happens?

Sort of. Since the select times out immediately I suspect it wastes a lot of CPU power but the issue mainly manifested itself through unmanageable log spam filling up disk til other applications started failing due to it. It is not correct to call select with negative timeouts and it should not be done but the source did seem to function properly from the point of view of a user using the application. I did not perform any measurements to quantify any performance difference with or without negative timeouts.

t0msk:
still not fixed after 2 years :(

PR #6037, opened 2022-02-26
linux-v4l2: Fix fallback framerate for camera

9p4:
### Description

Fixes very large log sizes and STDOUT spam when camera does not report framerate.

Code taken from https://github.com/Mikael-Lovqvist/obs-studio/commit/3cd5742be5168f625558c14c1b38fedf23116277

This adds a fallback framerate if the camera does not report one. It defaults to 10 frames per second, but can be manually changed later in the UI (included).

### Motivation and Context

- Fixes https://github.com/obsproject/obs-studio/issues/4926
- Fixes https://github.com/obsproject/obs-studio/issues/8397

### How Has This Been Tested?

Tested on `Linux hostname 5.16.11-arch1-1 #1 SMP PREEMPT Thu, 24 Feb 2022 02:18:20 +0000 x86_64 GNU/Linux`. Built with `mkdir build && cd build && cmake -DCMAKE_INSTALL_PREFIX="/usr" -DBUILD_BROWSER=OFF -DBUILD_VST=OFF -DDISABLE_VLC=ON .. && make -j$(nproc) && mkdir dist && make install DESTDIR=dist`. Tested by checking log sizes and STDOUT for affected camera. Log was as expected: without the hundreds of thousands of error messages.

### Types of changes

### Checklist:

### DIFF
--- a/plugins/linux-v4l2/v4l2-input.c
+++ b/plugins/linux-v4l2/v4l2-input.c
@@ -39,6 +39,8 @@ along with this program.  If not, see <http://www.gnu.org/licenses/>.
 #include "v4l2-helpers.h"
 #include "v4l2-decoder.h"
 
+#define FALLBACK_FRAMERATE 30
+
 #if HAVE_UDEV
 #include "v4l2-udev.h"
 #endif
@@ -1028,6 +1030,11 @@ static void v4l2_init(struct v4l2_data *data)
 	if (v4l2_set_framerate(data->dev, &data->framerate) < 0) {
 		blog(LOG_ERROR, "Unable to set framerate");
 		goto fail;
+	}
+	if (data->framerate == 0) {
+		blog(LOG_ERROR, "Framerate is not set, falling back to %i",
+		     FALLBACK_FRAMERATE);
+		data->framerate = v4l2_pack_tuple(1, FALLBACK_FRAMERATE);
 	}
 	v4l2_unpack_tuple(&fps_num, &fps_denom, data->framerate);
 	blog(LOG_INFO, "Framerate: %.2f fps", (float)fps_denom / fps_num);


kkartaltepe:
Can you please provide more information about the device where this occurs? What framerates does it support that is resulting in us picking 0/0?

9p4:
The camera seems to not report any supported framerate whatsoever, which is causing OBS to default to a framerate of zero. The culprit in question is (I think, I'm not quite sure) a Logitech QuickCam Express (320x240 @ 30 frame/s,640x480 @ 15 frame/s).

9p4:
Converting to draft since it seems as if it's not particularly reliable at the moment.

9p4:
> What tools have you used to determine the reported framerates? Why does this work in any other software if it reports no supported framerates?

I have not determined the reported framerates. I took those values from the datasheet online. I chose 10 since it seemed like a reasonable default. Keep in mind that this is a stopgap to prevent filling up the logs with a strange webcam, and the user is meant to change the FPS of the camera to the actual value once the application is loaded.

jp9000:
I have to be honest, I care more about fixing the log file issue than I do trying to support an ancient camera

plugins/linux-v4l2/v4l2-input.c
```diff
@@ -39,6 +39,8 @@ along with this program.  If not, see <http://www.gnu.org/licenses/>.
 #include "v4l2-helpers.h"
 #include "v4l2-mjpeg.h"
 
+#define FALLBACK_FRAMERATE 10
```
kkartaltepe, line 42: Use 30 to match chromes implementation

plugins/linux-v4l2/v4l2-input.c
```diff
@@ -169,7 +171,7 @@ static void *v4l2_thread(void *vptr)
 	struct v4l2_buffer buf;
 	struct obs_source_frame out;
 	size_t plane_offsets[MAX_AV_PLANES];
-	int fps_num, fps_denom;
+	uint fps_num, fps_denom;
```
kkartaltepe, line 174: Why did you change the type?

plugins/linux-v4l2/v4l2-input.c
```diff
@@ -1008,11 +1010,18 @@ static void v4l2_init(struct v4l2_data *data)
...
+		     FALLBACK_FRAMERATE);
+		data->framerate = v4l2_pack_tuple(1, FALLBACK_FRAMERATE);
+		if (v4l2_set_framerate(data->dev, &data->framerate) < 0) {
+			blog(LOG_ERROR, "Unable to set framerate");
+			goto fail;
+		}
 	}
 	v4l2_unpack_tuple(&fps_num, &fps_denom, data->framerate);
+
+	blog(LOG_INFO, "fps_num: %d", fps_num);
```
kkartaltepe, line 1024: Remove this added logging.

plugins/linux-v4l2/v4l2-input.c
```diff
@@ -1008,11 +1010,18 @@ static void v4l2_init(struct v4l2_data *data)
...
 
 	/* set framerate */
-	if (v4l2_set_framerate(data->dev, &data->framerate) < 0) {
-		blog(LOG_ERROR, "Unable to set framerate");
-		goto fail;
+	if (data->framerate == 0) {
+		blog(LOG_ERROR, "Framerate is not set, falling back to %i",
+		     FALLBACK_FRAMERATE);
+		data->framerate = v4l2_pack_tuple(1, FALLBACK_FRAMERATE);
+		if (v4l2_set_framerate(data->dev, &data->framerate) < 0) {
```
kkartaltepe, line 1017: This should be outside of the if statement to preserve functionality for non-broken webcams.

kkartaltepe (commented):
I looked at chrome and it seems they do a similar fallback for broken cameras so I guess there is no reason not to do this.

9p4:
I'll test out this commit with the old webcam when I get back home for spring break.

plugins/linux-v4l2/v4l2-input.c
```diff
@@ -1025,6 +1027,11 @@ static void v4l2_init(struct v4l2_data *data)
 	blog(LOG_INFO, "Linesize: %d Bytes", data->linesize);
 
 	/* set framerate */
+	if (data->framerate == 0) {
```
kkartaltepe, line 1030: I think it got mixed when you were applying my suggestions. We should move this after the v4l2_set_framerate (tries to apply the chosen fps and returns the resulting fps, which may be 0) and before the unpack on L1039.

9p4: whoops. should be fixed now

9p4:
Can confirm that this works.

kkartaltepe (approved):

kkartaltepe:
Please rebase instead of including merge commits in your PR.

9p4:
Done.

PR #6047, opened 2022-02-27
UI: Dropfiles Undo/Redo

gxalpha:
### Description

- Adds undo/redo to dropfiles, so that sources created by drag-and-drop can be removed
again via Ctrl-Z.
~~- Adds a dialog telling the user that the file is not supported if a file with an unknown
extension is drag-and-dropped into OBS. So far, this fails silently.~~

### Motivation and Context

- Source actions like this should have undo/redo
~~- Having nothing happen if a user drags an unknown file into OBS can be confusing to the
user who might have no idea what's going on (and is just wondering why no new source
appears)~~

### How Has This Been Tested?

macOS 12.3 Beta 4, compiled and run
- Undo/Redo works, tested with raw text, a media file and a url
~~- Dialog appears if I try to add an unknown file type: 
*<<Removed>>*~~

### Types of changes

- Bug fix (undo/redo)
~~- Tweak (dialog)~~

### Checklist:

### DIFF
--- a/UI/window-basic-main-dropfiles.cpp
+++ b/UI/window-basic-main-dropfiles.cpp
@@ -106,7 +106,6 @@ void OBSBasic::AddDropSource(const char *data, DropType image)
 {
 	OBSBasic *main = reinterpret_cast<OBSBasic *>(App()->GetMainWindow());
 	OBSDataAutoRelease settings = obs_data_create();
-	OBSSourceAutoRelease source = nullptr;
 	const char *type = nullptr;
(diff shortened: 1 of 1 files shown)


WizardCM:
I personally don't think it's necessary to pop a full dialog if the user tries to drag an unsupported file.

dodgepong:
If not a popup, then how should users be notified that the file was not successfully added to OBS due to an unsupported file type?

gxalpha:
Removed the unknown file warning as discussed.

PR #6049, opened 2022-02-28
mac-syphon: Mark source as deprecated

gxalpha:
### Description

Marks the syphon client source as deprecated, moving it to the
deprecated section in the Add Source popup:
image

### Motivation and Context

Syphon Inject was the main usage of the syphon source, mostly as a game
capture replacement, but no longer works since macOS 10.14, causing more
support requests than value.
Additionally, Syphon itself appears to be mostly abandoned for a year
now, which doesn't induce much confidence its future.

### How Has This Been Tested?

macOS 12.3 Beta 4, compiled and run

Syphon Client source has moved, as shown in the screenshot above.

### Types of changes

- Tweak (non-breaking change to ***improve*** existing functionality)

### Checklist:

### DIFF
--- a/plugins/mac-syphon/syphon.m
+++ b/plugins/mac-syphon/syphon.m
@@ -1223,7 +1223,7 @@ struct obs_source_info syphon_info = {
 	.id = "syphon-input",
 	.type = OBS_SOURCE_TYPE_INPUT,
 	.output_flags = OBS_SOURCE_VIDEO | OBS_SOURCE_CUSTOM_DRAW |
-			OBS_SOURCE_DO_NOT_DUPLICATE,
+			OBS_SOURCE_DO_NOT_DUPLICATE | OBS_SOURCE_DEPRECATED,
 	.get_name = syphon_get_name,
 	.create = syphon_create,
 	.destroy = syphon_destroy,


gxalpha:
Which icon would you change it to?

dodgepong:
The icon should definitely be changed, since it hasn't been the equivalent of game capture in years.

I'm on the fence as to whether this should be deprecated, though. It would be nice if more applications supported an open source texture-sharing library like Syphon. I'd be interested to hear if there are any Syphon users out there actually using the Syphon source as-is without SyphonInject.

Is the deprecation proposal motivated more by UX concerns, or by maintenance overhead?

gxalpha:
A bit of both.

The plugin is essentially (with minor exceptions) the same as it was in 2015, including the framework. This gives us about half the warnings of the entire project. All of the warnings are deprecation warnings, so there's a good chance it will just break with a future macOS update. While this could get updated, having no real usages means that no one wants to maintain it.

From the UX side, it's not much better. When you add the source, you get greeted with a properties dialog where you to already need know what you're doing (also, notice the "Launch SyphonInject" button in there? It doesn't work for 3 years now since SyphonInject doesn't either):
image
Again, this could get fixed - but with no usages, it looks like no one is doing that.

Regarding other integrations: There is a list of compatible apps on [their website](http://syphon.v002.info), though like half of them are plugins, not even native integrations. Don't get me wrong, I'd *love* to see Syphon or something similar more broadly adopted, but with it existing in this state for years now, I can't see it happen.

Regarding icons, which one would you suggest? (List of available options [here](https://obsproject.com/docs/reference-sources.html#c.obs_source_info.icon_type))

dodgepong:
> Regarding icons, which one would you suggest? (List of available options [here](https://obsproject.com/docs/reference-sources.html#c.obs_source_info.icon_type))

I suppose the least-bad one would be Unknown, since I feel like it would do the least to confuse users about what it actually does.

If we keep Syphon source, at the very least the Launch SyphonInject button should be deleted (as well as SyphonInject itself, if we're somehow still shipping it).

Obviously we have no telemetry about how many people are actually using this feature, but I would be curious if there are any legitimate use cases out there being used in the wild.

mufunyo:
> I'd be interested to hear if there are any Syphon users out there actually using the Syphon source as-is without SyphonInject.

FWIW, I used Syphon with OBS a couple years back when I was streaming an event that had a VJ present. The VJ software they were using (I forget what it was) supported Syphon output, and I used OBS' Syphon client to stream the VJ's Macbook feed over NDI to my stream PC.

gxalpha:
Internal discussion yielded that we should probably start maintaining this source again instead.

PR #6053, opened 2022-03-01
UI: Simplify multi-instance check

gxalpha:
### Description

Changes the various multi-instance check function names into one, making
calling it much more clean, since it no longer needs to differentiate between
macOS, Linux, FreeBSD and whatever DragonFlyBSD is.
Unfortunately, Windows requires carrying the RunOnceMutex, so it's not *as*
clean as it could be, but imo better than what we currently have.

### Motivation and Context

The ifdef-ladder in obs-app.cpp is ugly and mostly unneeded, since the
check works the same on all operating systems, except Windows where
the function has a return value. Having four different functions doing
the same thing is not very nice to look at.

### How Has This Been Tested?

Compiles on macOS, Windows and Ubuntu (CI).
Tested on macOS 12.3 and Windows 11 that the check still works.

Not tested on other unix-like (since I don't have them), but other than
the function names the code wasn't changed there and it still compiles.

### Types of changes

### Checklist:

### DIFF
--- a/UI/obs-app.cpp
+++ b/UI/obs-app.cpp
@@ -2093,15 +2093,10 @@ static int run_program(fstream &logFile, int argc, char *argv[])
 		bool cancel_launch = false;
 		bool already_running = false;
 
-#if defined(_WIN32)
-		RunOnceMutex rom = GetRunOnceMutex(already_running);
-#elif defined(__APPLE__)
-		CheckAppWithSameBundleID(already_running);
-#elif defined(__linux__)
-		RunningInstanceCheck(already_running);
-#elif defined(__FreeBSD__) || defined(__DragonFly__)
-		PIDFileCheck(already_running);
+#ifdef _WIN32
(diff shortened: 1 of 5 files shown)


UI/obs-app.cpp
```diff
@@ -2093,15 +2093,10 @@ static int run_program(fstream &logFile, int argc, char *argv[])
...
-#elif defined(__APPLE__)
-		CheckAppWithSameBundleID(already_running);
-#elif defined(__linux__)
-		RunningInstanceCheck(already_running);
-#elif defined(__FreeBSD__) || defined(__DragonFly__)
-		PIDFileCheck(already_running);
+#ifdef _WIN32
+		RunOnceMutex rom =
 #endif
+			CheckAlreadyRunning(already_running);
```
WizardCM, line 2099: I'd probably name it `CheckIfAlreadyRunning` for clarity?

PR #6055, opened 2022-03-01
libobs, frontend-tools, docs: Free frontend Decklink modules

cg2121:
### Description
If Decklink was not found, the UI woudn't load, but the modules were
still loaded.

### Motivation and Context
Less memory usage if not using Decklink devices.

### How Has This Been Tested?
Loaded OBS on a computer without Decklink devices.

### Types of changes

### Checklist:

### DIFF
--- a/UI/frontend-plugins/decklink-captions/decklink-captions.cpp
+++ b/UI/frontend-plugins/decklink-captions/decklink-captions.cpp
@@ -158,8 +158,10 @@ bool obs_module_load(void)
 
 void obs_module_post_load(void)
 {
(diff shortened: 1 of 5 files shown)


notr1ch:
I don't think this is a good idea. If the `os_dlclose` in `free_module` is ever uncommented, this will free the currently executing code and cause a crash. And since it is currently commented, this PR doesn't do what it says since the module remains loaded.

A safer implementation would be to return a value from `obs_module_post_load` to indicate unloading desire, but that would also be an API break.

Issue #3418, opened 2020-09-05
[Linux] Preview remain disabled after maximized and difficult to re-enable


Aikatsui:
**Steps to Reproduce**

1. Maximize OBS studio from tray. (Tested sources: screen capture and window capture)
2. [Test with screen capture] when maximize obs shows 'Preview is currently disabled'
3. 'Enable preview' button only helps seldom otherwise needs to click properties on screen capture source then click ok and click enable preview button again then works smooth.

It's good auto disable preview  when minimized (if not recording)

Screen capture:XSHM | Window capture:Xcomposite
PopOS 20.04.01 LTS | Gnome | 25.0.8 64bit (using apt)

Fenrirthviti:
Please fully fill out the issue template.

KizunaH:
@Fenrirthviti Above included all possible details able to provide and cover the issue template However let me know if that details insufficient to recognize the problem or required to add screenshots and other details

WizardCM:
On Ubuntu 20.04 I can only reproduce this if "Always minimise to system tray instead of task bar" is enabled in Settings -> General.
1. After enabling the setting, press the "-" minimise button in the app titlebar
2. Pressing "Show" the first time in the tray icon after minimising will flash the window for a split-second but not display it properly
3. Pressing "Show" a second time will make it appear, with "Preview is currently disabled"
4. Pressing the "Enable Preview" button the first time will do nothing
5. Pressing it a second time will correctly enable the preview

Doesn't happen on Windows 10.

Tested with 0b163018b3f239e41cb5fa493a9904aec7c02404.

PR #6057, opened 2022-03-01
UI: Fix show/hide toggle with minimize to taskbar

cg2121:
### Description
If the user would minimize to tray instead of taskbar, the
window would flash when clicking show in the tray menu, and
the preview would be disabled. This would only happen on Linux.

### Motivation and Context
Fixes https://github.com/obsproject/obs-studio/issues/3418

### How Has This Been Tested?
- Set OBS to minimize to system tray instead of taskbar
- Clicked show in tray menu to make sure everything worked as expected

### Types of changes

### Checklist:

### DIFF
--- a/UI/window-basic-main.cpp
+++ b/UI/window-basic-main.cpp
@@ -4716,6 +4716,7 @@ void OBSBasic::changeEvent(QEvent *event)
 			if (trayIcon && trayIcon->isVisible() &&
 			    sysTrayMinimizeToTray()) {
 				ToggleShowHide();
+				return;
 			}
 
 			if (previewEnabled)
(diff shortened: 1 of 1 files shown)


WizardCM:
Has this change been tested on all 3 platforms, or just Linux? And on Linux, which DE?

cg2121:
Just Linux on Ubuntu 21.10.

cg2121:
Tested now on Windows 10 without problems

WizardCM:
Commit title needs improvement.

cg2121:
Title has been updated.

UI/window-basic-main.cpp
```diff
@@ -8902,9 +8903,10 @@ void OBSBasic::SetShowing(bool showing)
 
 		if (showHide)
 			showHide->setText(QTStr("Basic.SystemTray.Show"));
-		QTimer::singleShot(250, this, SLOT(hide()));
+		QTimer::singleShot(0, this, SLOT(hide()));
```
WizardCM, line 8906: Does this actually still need to be in a QTimer if it's being triggered at a delay of 0, or is this more "rather not break anything unnecessarily" territory?

WizardCM (approved):
Verified this fixes the bug on Linux.

UI/window-basic-main.cpp
cg2121: Yes, as sometimes the window would hang when showing/hiding, so it needed to be toggled on the next event loop.

PR #6058, opened 2022-03-01
libobs: Show plugin search paths in log

norihiro:
### Description

This commit add log lines that shows `bin` and `data` paths to search plugins.

### Motivation and Context

Some users try to install a plugin from zip file and possibly make a mistake to expand the plugin into wrong location.
In other case, some users installed OBS Studio to unusual locations or installed multiple locations on Windows, which causes executable installer to take unexpected install path.
When a plugin is installed by a user but not shown up, there is no way to support the user from the log file. The new log lines should be helpful to find the plugin is installed in the correct location.

### How Has This Been Tested?

OS: Fedora 34
Output of the log with this change is as below.
  https://obsproject.com/logs/195if0CMVjtI6JTa

### Types of changes

### Checklist:

### DIFF
--- a/libobs/obs-module.c
+++ b/libobs/obs-module.c
@@ -466,6 +466,7 @@ void obs_find_modules(obs_find_module_callback_t callback, void *param)
 
 	for (size_t i = 0; i < obs->module_paths.num; i++) {
 		struct obs_module_path *omp = obs->module_paths.array + i;
+		blog(LOG_INFO, "finding modules in %s %s", omp->bin, omp->data);
 		find_modules_in_path(omp, callback, param);
 	}
 }


WizardCM:
I would say this is probably only an issue on Linux where certain containers will limit or customize search paths.

norihiro:
> I would say this is probably only an issue on Linux where certain containers will limit or customize search paths.

Usually deb package is provided for Ubuntu so that it is not an issue for Linux.
I mainly experienced support requests for macOS.

WizardCM:
macOS search paths are always the same, so this logging would provide no insight to someone in support channels. Users should not be putting third party plugins into OBS.app at all.

norihiro:
Thank you, WizardCM. I should reconsider this PR. Maybe documentation should be the better place.

PR #3153, opened 2020-07-11
UI: Match Windows taskbar state to tray icon (v2)

WizardCM:
### Description

This adds a status icon overlay for active/paused outputs. It is shown in the bottom right corner of the icon in the taskbar. This is consistent with other apps like Discord (notification count) and Chrome (current profile, if multiple available).
Note: this status icon isn't visible when the taskbar is in "small icon" mode.

Example from other apps:
image
image

Stripped down version of #2442

### Motivation and Context

While the coloured taskbar may have been considered overkill, I believe it's still worth including a visual on the taskbar when an output is active. It is common enough for users in support channels (especially Twitter) to mention that they either forgot to hit record, or didn't realise that they had stopped recording. The icons are large enough to be noticeable, but small enough to not be distracting.

### How Has This Been Tested?

Windows 10. Start an output, pause it, then stop it.

### Types of changes
- New feature (non-breaking change which adds functionality)

### Checklist:
- [x] My code has been run through [clang-format](https://github.com/obsproject/obs-studio/blob/master/.clang-format).
- [x] I have read the [**contributing** document](https://github.com/obsproject/obs-studio/blob/master/CONTRIBUTING.rst).
- [x] My code is not on the master branch.
- [x] The code has been tested.
- [x] All commit messages are properly formatted and commits squashed where appropriate.
- [x] I have included updates to all appropriate documentation.

PR #6060, opened 2022-03-01
UI: Add status overlay for macOS dock icon

gxalpha:
### Description

macOS equivalent of #3153.
Position of the icon on the bottom left matches the status bar.

Recording/Streaming:
image

Paused:
image

Normal (existing):
image

Doesn't rely on QtMacExtras, which means it will survive the transition to Qt6 (which the existing Windows implementation won't, that needs reworking before Qt6).
Implementation for other Unix-like is not included, the functions are just added so that everything compiles.

### Motivation and Context

This is a nice QOL addition on Windows and should be available on more platforms.

### How Has This Been Tested?

Compiled on macOS 12.3, Windows CI and Linux CI.
Made sure Windows implementation still works (Windows 11).
Tested that macOS works as intended, see the screenshot above.

### Types of changes

### Checklist:


tytan652:
I don't think it's implementable on Linux on all Desktop Environments, most of them just use the SVG in the `.desktop` file.

This shall maybe be macOS and Windows only in the code.

gxalpha:
> I don't think it's implementable on Linux on all Desktop Environments

Does this mean it's possible in *some* environments?
> This shall maybe be macOS and Windows only in the code.

It essentially is. I really hate the idea of ifdefing all the statements again, makes the code hard to read. I personally think having these empty functions in platform-x11 is fine; and they'll likely get optimized away anyways.

tytan652:
> Does this mean it's possible in _some_ environments?

I seriously don't know, I just consider that not all DEs are not made the same way.

But I think we should avoid per DEs features on Linux and try to provide the same amount feature on all DEs.

UI/platform-x11.cpp
```diff
@@ -257,3 +257,8 @@ bool SetDisplayAffinitySupported(void)
 	// Not implemented yet
 	return false;
 }
+
+// Not implemented yet
```
tytan652, line 261: It's this comment that made me write my first comment.

gxalpha:
I disagree. Why should one DE hold others back? We're generally happy to introduce features to specific operating systems only (a prime example is this exact feature), I don't see why this should be different for specific DE's. That is to say, if someone were to come and to implement it for only one DE, straight up saying "all or none" seems wrong to me.

But that's a discussion that's not in scope of this PR.
If it's just about the comment, I can remove that if it would be a blocker for you.

tytan652:
> If it's just about the comment, I can remove that if it would be a blocker for you.

You can leave it, just wanted to tell my opinion that we should try to treat DEs equally. And there is also the Flatpak that ignore in which DE it is run.

For Windows there is only one "DE", macOS same. But on Linux, many of them.

You are right this discussion is out of the scope of this PR, sorry for the inconvenience.

gxalpha:
No problem, thanks for bringing it up!
I guess it can get discussed again should someone be doing a Linux implementation.

UI/platform-osx.mm
```diff
@@ -255,6 +255,51 @@ bool ProcessIsRosettaTranslated()
...
+@implementation DockView
+- (id)initWithIcon:(QIcon)icon
+{
+	self = [super init];
+	self->icon = icon;
+	return self;
+}
+- (void)drawRect:(NSRect)dirtyRect
+{
+	CGSize size = dirtyRect.size;
```
PatTheMav, line 273: @gxalpha How often is this called? If the icon (including the status icon) only changes on actual changes of the app's status, then redrawing the overlay on every draw call of the composer seems wasteful to me.

UI/platform-windows.cpp
```diff
@@ -469,3 +474,26 @@ bool IsRunningOnWine()
 
 	return false;
 }
+
+#if QT_VERSION < QT_VERSION_CHECK(6, 0, 0)
+QWinTaskbarButton *taskBtn;
+void TaskbarOverlayInit()
```
PatTheMav, line 480: If this PR also adds functionality to Windows, it should either be renamed or the changes decoupled into another PR.

Right now Jim (and I) were under the assumption this only adds functionality to macOS, but if it also changes other platforms I cannot fully review+merge this on my own.

gxalpha: This functionality already exists for Windows. It was only moved from window-basic-main to the platform file, functionally was not changed.

UI/platform-osx.mm
gxalpha: It has to be called manually with `[tile display]`, so it's only called when `TaskbarOverlaySetIcon` is.

UI/platform-windows.cpp
PatTheMav: Ah right, you just moved the code around. Did you check the functionality on Windows?

gxalpha: I downloaded CI builds on my fork and tested it still works for recordings

PatTheMav:
Code looks fine to me, but I'm not entirely happy with the composition of the icon and the overlays - the record icon is too similar to the notification badge and detached elements of icons usually are put in the bottom right corner (compare: Photo Booth, Preview, Disk Utility, Chess).

Maybe if both icons are contained in another shape and this shape is the "common" element between both it allows it to be distinct from the notification badge and also make the <pause> icon feel less "floaty".

gxalpha:
> detached elements of icons usually are put in the bottom right corner

I deliberately put it in the bottom left corner since with the Windows implementation, users semi-regularly complained that it doesn't match the taskbar icon. If you want me to, I can put it in the bottom right, sure.

> Maybe if both icons are contained in another shape and this shape is the "common" element between both it allows it to be distinct from the notification badge and also make the icon feel less "floaty".

I'd need a mockup to understand what you mean, I'm pretty bad at visualizing things.

PatTheMav:
I was thinking along the lines of this - just quickly whipped-up examples:

Recording Active

example_obs_icon_record

Recording Paused

example_obs_icon_pause

Colors/stroke widths obviously needs tweaking, but I guess you get the idea.

WizardCM:
Would you be willing to split off the changes that standardise/version-gate the Windows function into a separate commit?

gxalpha:
Split off into #6371. Will be rebase this one once 6371 is merged.

PatTheMav:
Has the champ already spoken on this?

gxalpha:
no

MisutaaAsriel:
Does this take into account PR #6009? macOS sets the dock icon by default via the PLIST, not by code, and users (such as myself) may wish to change the user icon in the dock via Finder (which changes to the icon by code override).

_Personally_, the option to enable status in the dock via settings would be preferable to being forced to use this. **I genuinely do not like dock icons changing themselves, or attempting to behave like a status bar**.

gxalpha:
> Does this take into account PR https://github.com/obsproject/obs-studio/pull/6009?

While I didn’t test it, I wrote it in a way that should respect any custom icons.

gxalpha:
Rebased to fix conflicts introduced in #6371.
Confirmed that custom icons are respected and do not get overridden, with the overlay correctly placed on top.

gxalpha:
As a note, this still requires a comment and/or design from @Warchamp7 regarding https://github.com/obsproject/obs-studio/pull/6060#issuecomment-1062185211 / https://github.com/obsproject/obs-studio/pull/6060#issuecomment-1067134057

Warchamp7:
> As a note, this still requires a comment and/or design from @Warchamp7 regarding [#6060 (comment)](https://github.com/obsproject/obs-studio/pull/6060#issuecomment-1062185211) / [#6060 (comment)](https://github.com/obsproject/obs-studio/pull/6060#issuecomment-1067134057)

Are there any images of how this looks currently? I'd personally like to match Windows having this indicator bottom left, but I defer to @PatTheMav as resident Mac person if he thinks having it bottom right would be more important.

Visually if we want a shared shape for the icons to sit on, I'd propose a white rounded square matching the radius of the whole icon, and removing the colored outline (or at least the drop shadow on it)

PatTheMav:
> > As a note, this still requires a comment and/or design from @Warchamp7 regarding [#6060 (comment)](https://github.com/obsproject/obs-studio/pull/6060#issuecomment-1062185211) / [#6060 (comment)](https://github.com/obsproject/obs-studio/pull/6060#issuecomment-1067134057)
> 
> Are there any images of how this looks currently? I'd personally like to match Windows having this indicator bottom left, but I defer to @PatTheMav as resident Mac person if he thinks having it bottom right would be expected on Mac and more important than having parity with Windows status location.
> 
> Visually if we want a shared shape for the icons to sit on, I'd propose a white rounded square matching the radius of the whole icon, and removing the colored outline (or at least the drop shadow on it)
> 
> Edit: Based off Pat's mockup
> 
> Mac Example
> 
> [Mac Record Top](https://user-images.githubusercontent.com/1554753/175789488-a70707e8-6356-47b3-bcd2-2daf38333a72.png) [Mac Pause Top](https://user-images.githubusercontent.com/1554753/175789489-f908b11e-85ec-454c-96f9-089072caf9b3.png)

In general dock icon overlays are uncommon in macOS (because that's what menu bar icons and the control center are for), which also means there aren't any rules for this.

If there's bigger support for having the icon in the bottom left, I'm fine with putting it there. The square variant I'm also fine with. Personally I'd like to be able to disable this, but I guess @gxalpha has to adapt the code to use macOS-specific overlay icons first anyway.

gxalpha:
Updated the icons to macOS specific ones and moved them to the bottom right.
Also made them a tiny bit bigger because with the ones posted above, they are very small and not really visible.

image
image

MisutaaAsriel:
> In general dock icon overlays are uncommon in macOS (because that's what menu bar icons and the control center are for), which also means there aren't any rules for this.
> 
> If there's bigger support for having the icon in the bottom left, I'm fine with putting it there. The square variant I'm also fine with. Personally I'd like to be able to disable this, but I guess @gxalpha has to adapt the code to use macOS-specific overlay icons first anyway.

Wouldn't the doc overlay be more appropriate as a circle in the top right? This is where macOS currently puts any form of "status" icon, **e.g.** notification icons. This would at least match the look and feel of native system behavior.

Mockup based on system appearance:
obs_live obs_paused

gxalpha:
If you look at https://github.com/obsproject/obs-studio/pull/6060#issuecomment-1062185211, it was specifically made to *not* look like a notification. That being said, I'm personally open to any position, as long as the rest of the team agrees on it.

MisutaaAsriel:
> If you look at [#6060 (comment)](https://github.com/obsproject/obs-studio/pull/6060#issuecomment-1062185211), it was specifically made to _not_ look like a notification. That being said, I'm personally open to any position, as long as the rest of the team agrees on it.

I would actually argue that using a badge similar to the notification badge would be ideal. Some applications throw "generic" notifications that either show just a red circle, or a red circle with a dot, as a way of grabbing user attention.

When OBS is recording, if there's going to be a dock notification, it is, well, **notifying** you that OBS is recording.

But to be fair, **Pat has a point** — [Apple says it's bad practice to mimic badges in apps, because badges can be turned off by the system](https://developer.apple.com/design/human-interface-guidelines/components/system-experiences/notifications). — If a user tried disabling it in Notifications, only for it to still show, they'd probably be confused. So there is an argument for making them *dissimilar* to badges.

**Being said**, the visibility of the current versions *leaves much to be desired*, and can be argued as an accessibility issue. In [#6060 (comment 1189657207)](https://github.com/obsproject/obs-studio/pull/6060#issuecomment-1189657207), it's really hard to make out the status of the pause icon, and in general the visibility to me is really poor. This is why the solid color background with white (for dark colors) or black (for bright colors) glyph seems to make more sense.

As it's not standard practice anyways *to* display statuses on the dock, we're kinda throwing the rulebook out anyways. Likewise, the option to enable, disable these in OBS settings **should** be present, idealistically, in a final implementation, ergo I would argue using badging is not inappropriate in this particular case. The whole point of the icons are to grab user attention, so visibility and accessibility should take priority, imho.

You also have to recognize that a record icon looks exactly like the generic badge icon, so any changes to it to make it distinct are also going to interfere with visibility.

It really could go either way. I'm not fond of how it looks now in these screenshots, but I admit that while badging does increase visibility and recognizability, *it's a minor faux pas against the Human Interface Guidelines.*

...just, not any more than status icons in the dock in general are.

PatTheMav:
FWIW I've been arguing that it is standard practise in macOS to put such status displays into the menu bar icon and _not_ on the dock icon, which avoids any such discussions (especially as that icon has all the desired functionality already).

If it has to be on the dock icon (and I agree that best-case it should be a setting that users can disable), the top right corner is not available because it's already taken by the OS.

MisutaaAsriel:
> If it has to be on the dock icon (and I agree that best-case it should be a setting that users can disable), the top right corner is not available because it's already taken by the OS.

Well, if the OS prevents top-right display, I really cannot argue with that.

*Being said*, as it stands, I still find the icons present an accessibility issue. Ideally, if there are going to be icons, they should meet criteria such as:

- Easy to discern
- High contrast
- Low detail
- Large glyphs

The icons as they are are rather small in nature, and in the pause icon in particular, places a light colored glyph atop a white background. This harms legibility and defeats the purpose of a dock icon, *as is*. I'm not opposed to them, but the current design is rather counterintuitive.

Taking in the feedback previous of this conversation however, I've attached some sample icons which are equal in canvas resolution to the icons in this commit\*, but seek to increase visibility, whilst taking design cues from the previous rendition, and seeking to blend them in with the default OS aesthetic without being too similar to notifications. Also: squircles.

image image

Also attached are icon samples of what it might look like; I'm assuming whatever method is being used would support the full icon "canvas" or image size, and not just to the border of the icon itself.

image image

\**The actual icon is about 96px, similar to other dock icon overlays **like** notifications, with a slight drop-shadow, resulting in about a 10 pixel difference (112 vs 122). This is to be consistent with the OS, not to copy it.*

(Sorry if I come off as annoying btw; I'm passionate about both software design and streaming, and as an OBS user and Mac user, I want to help wherever I am able.)

PatTheMav:
> > If it has to be on the dock icon (and I agree that best-case it should be a setting that users can disable), the top right corner is not available because it's already taken by the OS.
> 
> Well, if the OS prevents top-right display, I really cannot argue with that.
> 

The OS doesn't prevent that, as we're in essence replacing the dock item with an image that has the badge "baked-in". If there'd be a notification, macOS would in essence draw over our status icon, as it doesn't care for icon decorations by apps.

It's for that reason that I deem that corner "unavailable".

As for the visual design I defer to Warchamp's judgement.

Warchamp7:
> > If it has to be on the dock icon (and I agree that best-case it should be a setting that users can disable), the top right corner is not available because it's already taken by the OS.
> 
> Well, if the OS prevents top-right display, I really cannot argue with that.
> 
> _Being said_, as it stands, I still find the icons present an accessibility issue. Ideally, if there are going to be icons, they should meet criteria such as:
> 
> * Easy to discern
> * High contrast
> * Low detail
> * Large glyphs
> 
> The icons as they are are rather small in nature, and in the pause icon in particular, places a light colored glyph atop a white background. This harms legibility and defeats the purpose of a dock icon, _as is_. I'm not opposed to them, but the current design is rather counterintuitive.
> 
> Taking in the feedback previous of this conversation however, I've attached some sample icons (mockups, more or less) which are equal in canvas resolution to the icons in this commit*, but seek to increase visibility, whilst taking design cues from the previous rendition, and seeking to blend them in with the default OS aesthetic without being too similar to notifications. Also: squircles.
> 
> image image (_Updated for graphical error_)
> 
> Also attached are icon samples of what it might look like; I'm assuming whatever method is being used would support the full icon "canvas" or image size, and not just to the border of the icon itself.
> 
> image image
> 
> *_The actual icon is about 96px, similar to other dock icon overlays **like** notifications, with a slight drop-shadow, resulting in about a 10 pixel difference (112 vs 122). This is to be consistent with the OS, not to copy it._
> 
> (Sorry if I come off as annoying btw; I'm passionate about both software design and streaming, and as an OBS user and Mac user, I want to help wherever I am able.)

I'm good with these icons but can you adjust the colors to #ff001e for Recording and #fecb34 for Pause?

MisutaaAsriel:
> @MisutaaAsriel I'm good with these icons but can you adjust the colors to #ff001e for Recording and #fecb34 for Pause?

Those icons were a little rough around the edges; more concept than productionr ready. But if you're serious about using them, here's some somewhat cleaned up versions with the changes requested:

image image

gxalpha:
Newest update (with the icons from https://github.com/obsproject/obs-studio/pull/6060#issuecomment-1193447442):
image
image

mufunyo:
> image image

Compared to Warchamp's version, the outer border is now mismatched with the inner border. This looks extremely off.

MisutaaAsriel:
> Compared to Warchamp's version, the outer border is now mismatched with the inner border. This looks extremely off.

I'll try making changes to it; now that you mention it you are correct, though i didn't see it at the time. Thank you for pointing that out.

MisutaaAsriel:
@Warchamp7 @mufunyo @gxalpha Reworked them in a proper vector graphics tool; This should close up any visual discrepancies caused by my other program and be more consistent. This also allows me to provide SVGs.

Icon Icon

SVG:
Icon Icon2

gxalpha:
In case you're wondering, we merged it as-is because it didn't care too much that the icons were imperfect since they are very small when shown anyways, and we didn't want to hold this PR up any more.
I can just update them to this new version.

MisutaaAsriel:
> In case you're wondering, we merged it as-is because it didn't care too much that the icons were imperfect since they are very small when shown anyways, and we didn't want to hold this PR up any more.
> I can just update them to this new version, thanks for making that.

Noted. And thank you for being gracious enough to update them. No matter how small, I'd be embarrassed if something I did was visually flawed, and this _is_ a Mac feature; Mac users might be more inclined to notice such things, especially at larger dock sizes.

gxalpha:
Pushed the new icons. Again, thanks for making them.

PR #6071, opened 2022-03-03
UI: Change 'Last Log' to 'Previous Log' in order to disambiguate things

Nofew:
### Description
Change "Last Log" to "Previous Log" in the Help menu.

### Motivation and Context
https://discord.com/channels/348973006581923840/376042662580715521/948928695182168084
"Last Log" and "Current Log" are ambiguous in some dialects. Changing "Last Log" to "Previous Log" disambiguates things.

### How Has This Been Tested?
N/A

### Types of changes

### Checklist:
- [X] The code has been tested. (See above)


gxalpha (approved):
I like this change.

As a note, I'm wondering if the locale *keys* should also be changed, though that would definitely invalidate all existing translations on Crowdin. I think like it's done here it doesn't(?), in which case it's probably better to keep it as is.

Nofew:
I thought that too. If it's possible to make the change silent, then it might be a good idea to change those too. I didn't change them here though, since that'd mean I'd have to alter code, and I was trying to keep this PR limited to changing strings only.

WizardCM (approved):

PR #6072, opened 2022-03-03
UI: Make theme change require restart under Wayland and enforce Qt default decoration

tytan652:
#### Restart for theme change on Wayland

With Qt default decoration, the decoration match OBS theme.
But changing theme doesn't change the color of decoration of already created Window until OBS is restarted.
So under Wayland, I made the theme setting will behave like the language setting. It will ask to restart and be unchangeable when if a video feature is active.

Look at the title bar after a theme change.
Screenshot from 2022-03-04 07-00-20

This happen should happen with the Flatpak installed from a non-GNOME DE and on distro that does not install Wayland decoration like QGnome by default.

#### Enforce Qt default decoration

Fedora and Flatpak under GNOME apply/install QGnome Wayland decoration by default. But this decoration gave big title bar to docks.

Screenshot from 2022-03-03 13-08-23

So this PR enforce the use Qt default decoration that is used if no other is installed. Making docks title bar less big and even match OBS theme.

The user still can override this by settings `QT_WAYLAND_DECORATION` is his environments.

Screenshot from 2022-03-03 13-08-07

Screenshot from 2022-03-03 14-27-28

### Motivation and Context

User can see this color mismatch as a bug.

I think those GNOME title bar are too big for the floating docks and doesn't match them.
And also maybe unify OBS look between distros.

### How Has This Been Tested?

- Run under GNOME Wayland with QGnome installed, and check that the decoration are not QGnome's.
- Run under Wayland, and check if the theme is changed only after a restart and the settings locked when the replay buffer is enabled.
- Run under X11/Xorg and check if the theme is changeable without restart and even with replay buffer enabled.

### Types of changes

### Checklist:

### DIFF
--- a/UI/obs-app.cpp
+++ b/UI/obs-app.cpp
@@ -2070,9 +2070,15 @@ static int run_program(fstream &logFile, int argc, char *argv[])
 	 * use other QPA platforms using this env var, or the -platform command
 	 * line option. */
 
+	/* NOTE 2: Some Wayland decoration can gave too big title bar to docks
+	 * like with QGnome decorations. So we force Qt default decoration (bradient).
+	 * It's still possible to use other decoration using this env var. */
+
 	const char *session_type = getenv("XDG_SESSION_TYPE");
-	if (session_type && strcmp(session_type, "wayland") == 0)
+	if (session_type && strcmp(session_type, "wayland") == 0) {
 		setenv("QT_QPA_PLATFORM", "wayland", false);
+		setenv("QT_WAYLAND_DECORATION", "bradient", false);
+	}
 #endif
 
 	OBSApp program(argc, argv, profilerNameStore.get());
--- a/UI/window-basic-settings.cpp
+++ b/UI/window-basic-settings.cpp
@@ -50,6 +50,10 @@
 #include <util/dstr.hpp>
 #include "ui-config.h"
 
+#if !defined(_WIN32) && !defined(__APPLE__) && defined(ENABLE_WAYLAND)
+#include <obs-nix-platform.h>
+#endif
+
 #define ENCODER_HIDE_FLAGS \
 	(OBS_ENCODER_CAP_DEPRECATED | OBS_ENCODER_CAP_INTERNAL)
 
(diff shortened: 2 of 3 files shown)


GeorgesStavracas:
This looks like a massive hack around qt-wayland, and as such, I feel strongly against this

tytan652:
I realized that the restart requirement in fact fix an issue that could happen on user that don't use Flatpak on GNOME or Fedora.

On other distros, Qt's default (bradient) is used leading to a color mismatch when changing theme without restart. I updated my PR description with details and illustration.

WizardCM:
These two commits seem very unrelated, and I feel they should be submitted & reviewed separately.

I don't see the need for making window decorations respect the OBS theme, I don't think anyone has been asking for that - and it might feel out of place.

tytan652:
> I don't see the need for making window decorations respect the OBS theme, I don't think anyone has been asking for that - and it might feel out of place.

The Qt default decoration does it automatically, I just prevent the mismatch that could happen when changing theme.

> These two commits seem very unrelated, and I feel they should be submitted & reviewed separately.

Yeah, I recently split them when I realise that the decoration not getting updated to match the theme was could be seen as a bug for people that do not use GNOME.

I feel that the second commit may be rejected, so I'm thinking of waiting @GeorgesStavracas opinion about making a second PR with only the second commit.

GeorgesStavracas:
I still do not think this should land. Allow me to punctuate my precise concerns with this PR:

 * When QGnomePlatform is available and the user is running GNOME, all Qt apps would have the GNOME titlebar style. With this PR, only OBS Studio would not follow that, making it inconsistent with the rest of the system.
 * We should avoid using and, more importantly, changing environment variable as much as we can. OBS Studio already has to trick Qt to use Wayland on GNOME using env vars, and that makes me super uncomfortable, and I think we consumed all our env vars cards with that. 
 * As much as possible, OBS Studio should not override upstream Qt policies. I think, in this case, we should perhaps raise this stylistic issue to QGnomePlatform developers directly.
 * Requiring a restart on Wayland seems to be a qt-wayland issue, and should be fixed there instead (unless, of course, there is evidence OBS Studio is at fault here)

Based on these points, I'll go ahead and close this PR, because I think none of the commits here should land. Regardless, thanks for raising these issues for discussion, and for your contributions.

tytan652:
Just a last message to tell that I really to put in my head that sometime the issue is upstream and shall be resolved there and not downstream.

PR #6089, opened 2022-03-05
UI: Display dock-relevant context menu on titlebar

WizardCM:
### Description

When the user right clicks a dock titlebar, display a context-aware context menu rather than the generic Docks menu.

**Before**:

image

**After**:

image

### Motivation and Context

**At @Warchamp7's request.**

His belief is it shouldn't matter where you're right clicking on a dock, it should show that dock's context menu.

Additionally, we moved the Docks menu to the main menu bar so accessing it from a context menu should no longer be necessary.

### How Has This Been Tested?

* Right click on a dock titlebar

### Types of changes

### Checklist:

### DIFF
--- a/UI/window-basic-main.cpp
+++ b/UI/window-basic-main.cpp
@@ -9922,11 +9922,25 @@ void OBSBasic::on_customContextMenuRequested(const QPoint &pos)
 {
 	QWidget *widget = childAt(pos);
 	const char *className = nullptr;
-	if (widget != nullptr)
+	QString objName;
+	if (widget != nullptr) {
 		className = widget->metaObject()->className();
+		objName = widget->objectName();
+	}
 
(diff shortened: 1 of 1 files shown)


dodgepong:
Does this include a "Refresh" option for custom browser docks?

WizardCM:
Currently, no. I could do a separate custom menu for docks that can include a refresh button, but it wouldn't have all the other options that dynamically appear when right clicking the contents of a browser dock.

lindenkron:
From a user perspective a few things feels off with this change.

It makes sense that right-clicking 'scenes' and 'sources' would show the relevant context menu for those, however seeing as they're still docks it feels off to not have the quick option to lock/unlock them from the right-click context menu.

This PR also leaves custom docks without any sort of interaction - where still having the current context menu would be intuitive.

PR #6091, opened 2022-03-05
linux-capture: Rewrite xcomposite

kkartaltepe:
### Description

Generally moves all the plugin code into xcomposite-input.cpp with a
mostly C style but keep some heavily c++ helpers xcomposite-helpers.cpp

Migrate as much as possible to xcb from Xlib to enable us to handle
errors and attribute them to the correct callers. This caused many other
knock on issues such as wrongly attributed errors and cleanup code
working incorrectly.

That allows us to use the xcursor-xcb implementation and delete the pure
Xlib implementation. We also add the missing functionality from the Xlib
implementation to the xcb implementation.

Capture glXCreatePixmap errors which occur most commonly on
nvidia+gnome due to nvidia's driver being unable to allocate more than 1
pixmap per window and gnome being the only compositor to read window
data via glx pixmaps.

Fix cleanup after failed glXCreatePixmap that might have leaked pixmaps
and prevented later captures on nvidia drivers for the same reason.

Migrate the x11 window event watcher to process events on its own
thread. This fixes a render pipeline stall in obs and compositors that
caused i3 and obs to reach ~10fps or gnome and obs to reach ~30fps.

### Motivation and Context

This is a much more drastic version of #5998 but removes lots of the
abstraction and deadly sharp edges of Xlib to hopefully avoid us reintroducing
even more bugs. Also fixes bugs identified in earlier PR and some additional
perf improvement.

### How Has This Been Tested?

Tested on i3/gnome/nvidia and i3/gnome/intel. Once taken out of draft there should be
no regressions from master.

### Types of changes

### Checklist:


kkartaltepe:
If anyone on AMD hardware can take this for a spin and let me know if anything behaves differently/worse than master that would be great.

Note if you do: Consider building master without the "window updated" `blog` so you can see existing gs_copy_texture failures on master. They appear more prominently in this PR because the window update message no longer spams the log when the window is resized (which is the same point where the pixmap texture might be invalid and print a gs_copy_texture failure).

Chiitoo:
What does without the "window updated" `blog` mean?

So far haven't noticed anything too weird with the actual capturing using AMD (navi10, LLVM 13.0.1 / Mesa 22.0.0-rc2), but there is something odd happening, that probably isn't specific to AMD, possibly in

https://github.com/obsproject/obs-studio/pull/6091/files#diff-3219ad45ddd4eafa31d9cdeaa31fea7f980a3e7b2f5ec28dba6fafba21aec821R84

which results into some of the window names having extra characters suffixed into their names (for the most part, they don't show up in the terminal output info messages, but for some they do).

One example that seems to easily affect things here is the Falkon browser.

The characters appearing can be parts of other window names, box characters, numbers, or other characters such as the exclamation mark.  Need to play around more with it to find out more.

kkartaltepe:
> The characters appearing can be parts of other window names, box characters, numbers, or other characters such as the exclamation mark. Need to play around more with it to find out more.

Sorry i didnt mention why this was draft but window names have garbage in them right now, xcb doesnt implement proper string parsing so I havnt put that code back in yet, probably will finish that up today.

> What does without the "window updated" blog mean?

Previously if you resized a window the `info: [window-capture: 'Window Capture (Xcomposite) 4'] update settings:` block was pushed to your log every frame. This was instead changed to only be output when you update your settings.

ogmkp:
If I test this one, should I drop #5998?

kkartaltepe:
> If I test this one, should I drop #5998?

Yes, these are two different approaches.

kkartaltepe:
This should be equivalent to master now. There is one annoyance to cleanup which is that windows that fail to capture will repeatedly log the failing glxCreatePixmap error.

Known errors on this branch and master (thus not regressions): gs_copy_texture failures due to invalid pixmap textures being captured, displaying a stale frame sometimes when multiple windows resize at once (such as in tiling window managers).

kkartaltepe:
This should be at least as good as master now. I had to strip out the thread for handling X11 events though as dealing with locking correctly would make things quite a bit more complicated, so we can retry that later if we really want (and as written still didnt perfectly solve all the slowdowns during things like resizes).

plugins/linux-capture/xcursor-xcb.h
```diff
@@ -70,10 +69,16 @@ void xcb_xcursor_update(xcb_xcursor_t *data,
 void xcb_xcursor_render(xcb_xcursor_t *data);
 
 /**
- * Specify offset for the cursor
+ * Specify a manual offset for the cursor
```
RytoEX, line 72: End with punctuation.

plugins/linux-capture/xhelpers.h
```diff
@@ -135,7 +135,6 @@ void xshm_xcb_detach(xcb_shm_t *shm);
  * @return screen info structure
  */
 xcb_screen_t *xcb_get_screen(xcb_connection_t *xcb, int screen);
-
```
RytoEX, line 138: Probably don't _need_ to remove this.

RytoEX (changes requested):
Mostly style/comment nitpicks. I'll have to defer a functional review to someone else who is familiar with this plugin or its subject material.

plugins/linux-capture/xcomposite-helpers.cpp
```diff
@@ -0,0 +1,568 @@
...
+int xcb_get_screen_for_root(xcb_connection_t *conn, xcb_window_t root)
+{
+	xcb_screen_iterator_t iter =
+		xcb_setup_roots_iterator(xcb_get_setup(conn));
+	for (int i = 0; iter.rem > 0; xcb_screen_next(&iter)) {
+		if (iter.data->root == root)
+			return i;
+		i++;
+	}
+	return 0;
```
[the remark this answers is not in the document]
kkartaltepe: It is written to never fail, so it will always return a valid screen. For valid input that will be the screen associated with the root id, for invalid input we just give you screen 0 which always exists. Avoiding callers having to error check and modeling the Xlib implementation of this function that its standing in for.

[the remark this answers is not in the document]
RytoEX: Fair enough.

plugins/linux-capture/CMakeLists.txt
```diff
@@ -1,42 +1,39 @@
 project(linux-capture)
 
 find_package(X11 REQUIRED)
-if(NOT TARGET X11::Xcomposite)
-  obs_status(FATAL_ERROR "linux-capture - Xcomposite library not found.")
+find_package(XCB COMPONENTS XCB XFIXES RANDR SHM XINERAMA COMPOSITE)
+if(NOT TARGET XCB::COMPOSITE)
+  message(
+    FATAL_ERROR
+      "OBS:  DISABLED   linux-capture - xcb composite library not found")
```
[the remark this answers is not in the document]
kkartaltepe: Thanks I didnt see this when I was rebasing with the cmake changes.

plugins/linux-capture/xhelpers.h
kkartaltepe: This is a result of clang-format.

plugins/linux-capture/xcomposite-helpers.cpp
```diff
@@ -0,0 +1,569 @@
...
+	if (s->tex) {
+		gs_texture_destroy(s->tex);
+		s->tex = 0;
+	}
+}
+
+static gs_color_format gs_format_from_tex()
+{
+	GLint iformat = 0;
+	// consider GL_ARB_internalformat_query
```
RytoEX, line 309: Is this a to-do?

plugins/linux-capture/xcomposite-helpers.cpp
```diff
@@ -0,0 +1,569 @@
...
+	free(check2);
+
+	return true;
+}
+
+std::list<xcb_window_t> xcomp_top_level_windows(xcb_connection_t *conn)
+{
+	std::list<xcb_window_t> res;
+
+	// ewmh top level window listing is not supported.
```
RytoEX, line 206: ewmh -> EWMH

plugins/linux-capture/xcomposite-input.hpp
```diff
@@ -0,0 +1,92 @@
...
+	gs_texture_t *gltex;
+
+	pthread_mutex_t lock;
+
+	bool show_cursor;
+	bool cursor_outside;
+	xcb_xcursor_t *cursor;
+	// strict_binding determines whether we rebind the GLX Pixmap on every
+	// tick. "Strict binding" is the correct mode of operation, according to
+	// GLX_EXT_texture_from_pixmap. However certain drivers exhibits poor
```
RytoEX, line 48: exhibits -> exhibit

plugins/linux-capture/xcomposite-input.cpp
```diff
@@ -0,0 +1,382 @@
...
+
+	watcher_register(conn, s);
+	xcomp_cleanup_pixmap(disp, s);
+	xcomp_create_pixmap(conn, disp, s, LOG_ERROR);
+	xcb_xcursor_offset_win(conn, s->cursor, s->win);
+	xcb_xcursor_offset(s->cursor, s->cursor->x_org + s->crop_left,
+			   s->cursor->y_org + s->crop_top);
+
+	obs_leave_graphics();
+	pthread_mutex_unlock(&s->lock);
```
RytoEX, line 309: I think the preference is to have these in reverse order. This is reflected in the other parts of this PR.

RytoEX (changes requested):
Some nits.

plugins/linux-capture/xcomposite-input.cpp
kkartaltepe: Good catch!

plugins/linux-capture/xcomposite-helpers.cpp
kkartaltepe: This was a prexisting comment. That query will let us know the real format to avoid our hacks but it requires a very recent version of opengl so its unlikely to be reasonable.

plugins/linux-capture/xcomposite-input.hpp
```diff
@@ -0,0 +1,92 @@
...
+template<typename F> ScopeExit<F> MakeScopeExit(F f)
+{
+	return ScopeExit<F>(f);
+};
+
+#define XSTRING_JOIN2(arg1, arg2) STRING_JOIN2(arg1, arg2)
+#define STRING_JOIN2(arg1, arg2) arg1##arg2
+#define DEFER(code)                                 \
+	auto XSTRING_JOIN2(scope_exit_, __LINE__) = \
+		MakeScopeExit([=]() { code; })
```
GeorgesStavracas, line 70: Is this the only reason for using C++ here? If so, allow me to suggest `__attribute__((__cleanup__(cleanup_func)))`, which can be similarly wrapped in a macro like this:

```c
#define autocleanup(cleanup_func) __attribute__((__cleanup__(cleanup_func)))
```

You can then use it in the code as follows:

```c
autocleanup(free) char *string_to_be_freed_by_the_end_of_the_scope = NULL;
autocleanup(XFree) GLXFBConfig config = NULL;
```

It's important to initialize these values as NULL otherwise exiting the scope could lead to freeing uninitialized memory.

If this is enough to cover all C++ features used here, it would be good to make these files .c and .h after :slightly_smiling_face:

GeorgesStavracas (changes requested):
I have only one big suggestion - the flow of the code seems to match the expected xcb / Xlib / GLX usage, and mostly matches what the previous code did, so that looks fine to me.

plugins/linux-capture/xcomposite-input.hpp
kkartaltepe: No, I left c++ for xcompcap_props/xcomp_top_level_windows/xcomp_find_window support for some string handling that would grow a fair bit without std::string and stdlib's lexicographical sort which didnt seem to have a libc equivalent so I left them c++.

Since i kept c++ I left some usage of this little defer where it was nice to have, but its not a reason to keep c++.

GeorgesStavracas: Alright, that's reasonable. My suggestion above still stands though - I feel like the autocleanups are easier to reason about compared to the DEFER() approach

plugins/linux-capture/xcomposite-input.cpp
```diff
@@ -0,0 +1,382 @@
...
+		props, "capture_window", obs_module_text("Window"),
+		OBS_COMBO_TYPE_LIST, OBS_COMBO_FORMAT_STRING);
+
+	struct WindowInfo {
+		std::string lex_comparable;
+		std::string name;
+		std::string desc;
+	};
+
+	std::vector<WindowInfo> window_strings;
```
GeorgesStavracas, line 203: This could be replaced by a dynamic array `darray` from `libobs/util/darray.h`

plugins/linux-capture/xcomposite-input.cpp
```diff
@@ -0,0 +1,382 @@
...
+		std::string name = xcomp_window_name(conn, disp, win);
+		std::string cls = xcomp_window_class(conn, win);
+		std::string id = std::to_string((uint32_t)win);
+		std::string desc =
+			(id + WIN_STRING_DIV + name + WIN_STRING_DIV + cls);
+
+		std::string name_lower = name;
+		std::transform(name_lower.begin(), name_lower.end(),
+			       name_lower.begin(),
+			       [](unsigned char c) { return std::tolower(c); });
```
GeorgesStavracas, line 215: You can replace the `std::string` usage by `dstr`, and use `dstr_to_lower()` here

plugins/linux-capture/xcomposite-input.cpp
```diff
@@ -0,0 +1,382 @@
...
+	}
+
+	std::sort(window_strings.begin(), window_strings.end(),
+		  [](const WindowInfo &a, const WindowInfo &b) -> bool {
+			  return std::lexicographical_compare(
+				  a.lex_comparable.begin(),
+				  a.lex_comparable.end(),
+				  b.lex_comparable.begin(),
+				  b.lex_comparable.end());
+		  });
```
GeorgesStavracas, line 229: `strcmp()` and, consequently, `dstr_cmp()`, are also lexicographical

plugins/linux-capture/xcomposite-input.hpp
GeorgesStavracas: This comment still applies, if you want scoped variables - or, alternatively, you could just drop the autocleanup changes entirely (though I personally enjoy scoped variables with the `autocleanup` approach above)

GeorgesStavracas (changes requested):
So, yesterday Jim expressed the wish to drop all C++ code from the linux-capture plugins. I too don't feel particularly great that there's still C++ around. Looking at this file, I think there are enough replacements to cover most of the string operations here, I'll comment each one of them inline below.

kkartaltepe:
It seems to work still after rewriting the cpp parts. Ill probably take another pass at cleaning it up before its worth reviewing but feel free to take it for a spin and see if i missed any `free`s or it explodes on your machine.

plugins/linux-capture/xcomposite-input.c
```diff
@@ -0,0 +1,397 @@
...
+	sinfo.get_width = xcompcap_getwidth;
+	sinfo.get_height = xcompcap_getheight;
+	sinfo.icon_type = OBS_ICON_TYPE_WINDOW_CAPTURE,
+
+	obs_register_source(&sinfo);
+}
+
+void xcomposite_unload(void)
+{
+	conn = NULL;
```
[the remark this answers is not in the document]
kkartaltepe: We are using the Xlib derived xcb conn so I dont think we should close it. We could close the display though.

plugins/linux-capture/xcomposite-input.h
```diff
@@ -0,0 +1,83 @@
...
+struct darray xcomp_top_level_windows(xcb_connection_t *conn);
+struct dstr xcomp_window_class(xcb_connection_t *conn, xcb_window_t win);
+struct dstr xcomp_window_name(xcb_connection_t *conn, Display *disp,
+			      xcb_window_t win);
+xcb_window_t xcomp_find_window(xcb_connection_t *conn, Display *disp,
+			       const char *str);
+bool xcomp_window_exists(xcb_connection_t *conn, xcb_window_t win);
+void xcomp_cleanup_pixmap(Display *disp, struct xcompcap *s);
+void xcomp_create_pixmap(xcb_connection_t *conn, Display *disp,
+			 struct xcompcap *s, int log_level);
```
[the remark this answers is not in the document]
kkartaltepe: They could, personally I prefer the xcomposite-input just holding the functions for the struct and the rest of these in the helpers file. I also dislike managing forward declarations in file versus a header.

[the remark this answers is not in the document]
GeorgesStavracas: I think it's a good practice in C codebases to keep functions internal to a compile unit out of the header file - head files usually only expose functions meant to be used by other compile units.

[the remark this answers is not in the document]
kkartaltepe: I can split them into two compilation units, do you really prefer it all in one file?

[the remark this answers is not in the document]
GeorgesStavracas: If that works for you, sure. I think the only functions exposed in the xcomposite-input.h are ` void xcomposite_load(void)` and `void xcomposite_unload(void)`. I guess you'd be moving the watcher helpers into a different file?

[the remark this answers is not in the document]
kkartaltepe: Not quite, ive pushed what i mean.

plugins/linux-capture/xcomposite-helpers.h
```diff
@@ -0,0 +1,30 @@
...
+			      xcb_window_t win);
+xcb_window_t xcomp_find_window(xcb_connection_t *conn, Display *disp,
+			       const char *str);
+bool xcomp_window_exists(xcb_connection_t *conn, xcb_window_t win);
+void xcomp_cleanup_pixmap(Display *disp, struct xcompcap *s);
+void xcomp_create_pixmap(xcb_connection_t *conn, Display *disp,
+			 struct xcompcap *s, int log_level);
+
+void watcher_register(xcb_connection_t *conn, struct xcompcap *s);
+void watcher_unregister(xcb_connection_t *conn, struct xcompcap *s);
```
GeorgesStavracas, line 28: Could these functions receive `xcb_window_t win` instead of `struct xcompcap`?

plugins/linux-capture/xcomposite-input.h
```diff
@@ -0,0 +1,54 @@
...
+	bool show_cursor;
+	bool cursor_outside;
+	xcb_xcursor_t *cursor;
+	// strict_binding determines whether we rebind the GLX Pixmap on every
+	// tick. "Strict binding" is the correct mode of operation, according to
+	// GLX_EXT_texture_from_pixmap. However certain drivers exhibit poor
+	// performance when this is done, so setting this to false allows working
+	// around it.
+	bool strict_binding;
+};
```
GeorgesStavracas, line 51: With the suggestion above, you can move this entire struct to xcomposite-input.c

plugins/linux-capture/xcomposite-input.h
```diff
@@ -0,0 +1,54 @@
...
+
+#define WIN_STRING_DIV "\r\n"
+#define FIND_WINDOW_INTERVAL 0.5
```
GeorgesStavracas, line 54: These 2 defines can be moved to xcomposite-helpers.c about it

plugins/linux-capture/xcomposite-input.c
```diff
@@ -0,0 +1,401 @@
...
+	obs_properties_add_bool(props, "include_border",
+				obs_module_text("IncludeXBorder"));
+
+	obs_properties_add_bool(props, "exclude_alpha",
+				obs_module_text("ExcludeAlpha"));
+
+	return props;
+}
+
+void xcompcap_defaults(obs_data_t *settings)
```
GeorgesStavracas, line 275: `static void`

plugins/linux-capture/xcomposite-input.c
```diff
@@ -0,0 +1,401 @@
...
+	obs_data_set_default_int(settings, "cut_left", 0);
+	obs_data_set_default_int(settings, "cut_right", 0);
+	obs_data_set_default_int(settings, "cut_bot", 0);
+	obs_data_set_default_bool(settings, "swap_redblue", false);
+	obs_data_set_default_bool(settings, "show_cursor", true);
+	obs_data_set_default_bool(settings, "include_border", false);
+	obs_data_set_default_bool(settings, "exclude_alpha", false);
+}
+
+void xcompcap_update(void *data, obs_data_t *settings)
```
GeorgesStavracas, line 288: `static void`

plugins/linux-capture/xcomposite-input.c
```diff
@@ -0,0 +1,401 @@
...
+	obs_leave_graphics();
+}
+
+static const char *xcompcap_getname(void *data)
+{
+	UNUSED_PARAMETER(data);
+	return obs_module_text("XCCapture");
+}
+
+void xcomposite_load(void)
```
GeorgesStavracas, line 335: This should be in the xcomposite-input.h header (and the `extern void xcomposite_{load,unload}` declarations from linux-capture.c then can be removed)

plugins/linux-capture/xcomposite-input.c
```diff
@@ -0,0 +1,401 @@
...
+		blog(LOG_ERROR,
+		     "window manager does not support Extended Window Manager Hints (EWMH).\nXComposite capture disabled.");
+		return;
+	}
+
+	struct obs_source_info sinfo = {
+		.id = "xcomposite_input",
+		.output_flags = OBS_SOURCE_VIDEO | OBS_SOURCE_CUSTOM_DRAW |
+				OBS_SOURCE_DO_NOT_DUPLICATE,
+
```
GeorgesStavracas, line 378: Nitpick: remove this empty line

GeorgesStavracas (changes requested):
Some more comments

plugins/linux-capture/xcomposite-helpers.h
kkartaltepe: No they map sources to the window they capture and must know this mapping.

plugins/linux-capture/xcomposite-input.h
kkartaltepe: WIN_STRING_DIV is used in both sources.

plugins/linux-capture/xcomposite-helpers.h
```diff
@@ -0,0 +1,30 @@
...
+struct darray xcomp_top_level_windows(xcb_connection_t *conn);
+struct dstr xcomp_window_class(xcb_connection_t *conn, xcb_window_t win);
+struct dstr xcomp_window_name(xcb_connection_t *conn, Display *disp,
+			      xcb_window_t win);
+xcb_window_t xcomp_find_window(xcb_connection_t *conn, Display *disp,
+			       const char *str);
+bool xcomp_window_exists(xcb_connection_t *conn, xcb_window_t win);
+void xcomp_cleanup_pixmap(Display *disp, struct xcompcap *s);
+void xcomp_create_pixmap(xcb_connection_t *conn, Display *disp,
+			 struct xcompcap *s, int log_level);
```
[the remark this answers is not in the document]
kkartaltepe: Perhaps we could move the more general helpers (those that dont depend on knowing the struct) to xhelpers. But these helpers are intended to be tightly coupled and just access and modify the source struct.

plugins/linux-capture/xcomposite-input.h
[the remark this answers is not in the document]
kkartaltepe: Resolving baed on other comments.

plugins/linux-capture/xcomposite-input.c
```diff
@@ -0,0 +1,400 @@
...
+#include "xcursor-xcb.h"
+#include "xcomposite-input.h"
+#include "xcomposite-helpers.h"
+
+static Display *disp = NULL;
+static xcb_connection_t *conn = NULL;
+
+static void xcompcap_update(void *data, obs_data_t *settings);
+
+static uint32_t xcompcap_getwidth(void *data)
```
GeorgesStavracas, line 24: Since you're already touching this code anyway, would you mind adding an underscore to `get_width` (same for `get_height`)? More generally, I think it would be good to make the suffix of these static functions match the vfuncs of `struct obs_source_info`

GeorgesStavracas (changes requested):
Just a few more comments

plugins/linux-capture/xcomposite-helpers.c
```diff
@@ -0,0 +1,637 @@
...
+}
+
+struct reg_item {
+	struct xcompcap *src;
+	xcb_window_t win;
+};
+static DARRAY(struct reg_item) watcher_registry = {0};
+static pthread_mutex_t watcher_lock = PTHREAD_MUTEX_INITIALIZER;
+
+void watcher_register(xcb_connection_t *conn, struct xcompcap *s)
```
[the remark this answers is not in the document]
kkartaltepe: We depend on knowing the struct to signal the render thread via the `window_changed` attribute. This is why we use the struct pointer, so we have access to the shared state to signal.

We could layer up the abstraction and create a watcher_id+*bool and pass those around but it seemed excessive when a simple pointer would do.

kkartaltepe:
Thanks ill try and get to these tomorrow

plugins/linux-capture/xcomposite-helpers.c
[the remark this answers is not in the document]
GeorgesStavracas: I'm trying my best not to suggest moving this back to xcomposite-input.c since you expressed the wish to keep it in a separate file, but it's getting increasingly hard for me to justify to myself it's worth keeping this split. It seems like the watcher routines are tightly locked to the xcompcap structure, and thus should be together. Apologies for the back and forth with the code :slightly_frowning_face:

plugins/linux-capture/xcomposite-input.c
kkartaltepe: Done, good idea

kkartaltepe:
I've updated it once more time and merged most code into xcomposite-input.c now. I think we have very different understandings of .c/.h files and so merging seems like the best way to avoid any future reader whos views may differ from both of ours from being confused by any odd file structures. They will just have to deal with the giant file.

GeorgesStavracas (approved):
Thank you! I think this version looks great, and I'm happy to land it and continue the EGL port, etc, from there

PR #6106, opened 2022-03-08
frontend-tools: Add ability to enable/disable scripts

cg2121:
### Description
This adds the ability to enable or disable scripts by clicking
the visibility icon in the scripts list.

Screenshot from 2022-03-08 02-08-30

### Motivation and Context
If a user doesn't want to use a script, but use it later, they can now just disable it without unloading and loading it.

### How Has This Been Tested?
Tested by enabling/disabling scripts.

### Types of changes

### Checklist:


exeldro:
Does it log the enabled state of scripts for support?

cg2121:
@exeldro the only reason I didn't do that, is that I felt the log would get too spammy when users enable/disable scripts.

WizardCM:
As long as the state on startup is logged, that should be fine.

dodgepong:
I'd like feedback from @Warchamp7 but my gut reaction to using the 👁 icon for enabling/disabling a script seems wrong. Like, it makes sense as an "active" checkbox but the eye usually implies "visibility" somehow, not just being on or off.

WizardCM:
The eye symbol is consistent with our Filters dialog, including audio filters.

dodgepong:
The eye makes sense for visual filters, though I've never really liked it for audio filters. I think it only ends up working in the audio filters dialog because of the parallel use in the video filters dialog.

WizardCM:
Heads up, there are merge conflicts.

mufunyo:
Audio filters should have an ear or speaker icon instead of an eye. Scripts should have a play (active) / pause (inactive) icon.

Warchamp7:
Agreeing with @dodgepong I'd like to see this as a checkbox instead of the eye toggle.

In the future I'd like to add a proper UI toggle button to replace checkboxes in many places (The left/right style dot one)

cg2121:
Updated to latest master.

I agree with @Warchamp7, that we should have the slider toggle. But for now, I think the eye icon is fine because it is used elsewhere in the program.

WizardCM:
UX-wise it does feel weird that the properties section displays "No properties available" when a script is disabled but selected. Outside of that, this seems to do what it says on the tin.

Warchamp7:
> UX-wise it does feel weird that the properties section displays "No properties available" when a script is disabled but selected. Outside of that, this seems to do what it says on the tin.

Could this be changed to show a new string indicating "Script is disabled" instead?

Otherwise fine with this. We use the eye icon for toggling audio filters too.

cg2121:
Updated to have the label show that the script is disabled.
Screenshot from 2022-07-27 21-53-12

ryantheleach:
> Updated to have the label show that the script is disabled. Screenshot from 2022-07-27 21-53-12

As a user, I puzzle why the description of a script is now set to "Script is disabled".

I would have expected, a description of the script...

Reading this PR, I understand how we ended up here, and that it's the execution of the script that probably populates it's description, but as a User I'd very much like to know what a script does, without having to enable it.

Fenrirthviti:
Added seeking testers, but conflicts will need to be resolved. Once CI is re-run, we can drop this PR for people in the scripting channels to test.

cg2121:
The merge conflicts are now fixed.

> As a user, I puzzle why the description of a script is now set to "Script is disabled".

It is not possible to get the description of the scripts, as they are unloaded when disabled.

Warchamp7:
Coming back around to this, how hard would it be to still load the description and possibly even properties (disabled) of scripts but not run their other functions? I imagine it would require some deeper changes, but should be possible, no?

cg2121:
Updated to use a single checkbox in the corner instead of the visibility icons. This will also look nicer when we use the new OBS widgets in #8447.

Enabled:
Screenshot from 2023-10-10 15-22-12

Disabled:
Screenshot from 2023-10-10 15-22-20

Warchamp7:
I don't think it makes sense to have this toggle in the corner. It's visually disconnected from everything else in the window. Can we put the checkbox back on the list items and just an actual checkbox instead of the eye icon?

It looks like in the latest screenshot we are able to load the description when the script is disabled. Are we able to load the properties?

If we CAN, I'd like to still show all the property controls but have them disabled.

If we CAN'T, I'd like to amend the info text to say
```Script properties cannot be viewed or modified when a script is disabled.```

Fenrirthviti:
Reminder that this PR has pending UI changes required, but largely seems to be an acceptable change in concept. If feedback can be addressed and this PR rebased, we can add it back to the normal review cycles.

Warchamp7:

PR #6115, opened 2022-03-09
obs-frontend-api: Add function to get frontend translated string

cg2121:
### Description
Developers now can easily get a translated string from the frontend
instead of using their own translations.

### Motivation and Context
Makes it easier for plugin developers.

### How Has This Been Tested?
Tested with my plugin.

### Types of changes

### Checklist:

### DIFF
--- a/UI/api-interface.cpp
+++ b/UI/api-interface.cpp
@@ -620,6 +620,11 @@ struct OBSStudioAPI : obs_frontend_callbacks {
 		return bstrdup(recordOutputPath);
 	}
 
+	const char *obs_frontend_get_locale_string(const char *string) override
+	{
+		return Str(string);
+	}
+
 	void on_load(obs_data_t *settings) override
 	{
 		for (size_t i = saveCallbacks.size(); i > 0; i--) {
--- a/UI/obs-frontend-api/obs-frontend-api.cpp
+++ b/UI/obs-frontend-api/obs-frontend-api.cpp
@@ -559,3 +559,9 @@ char *obs_frontend_get_current_record_output_path(void)
 		       ? c->obs_frontend_get_current_record_output_path()
 		       : nullptr;
 }
+
+const char *obs_frontend_get_locale_string(const char *string)
+{
+	return !!callbacks_valid() ? c->obs_frontend_get_locale_string(string)
+				   : nullptr;
+}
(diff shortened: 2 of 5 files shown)


gxalpha:
For your info, should the plugin depend on QT you could also just use QObject::tr (Example here:
https://github.com/gxalpha/obs-properties-dock/blob/b7538b08ebcbf1659996aae4720e8a1ade62fb6d/src/obs-classes-helper-functions.hpp#L24)

cg2121:
Non QT plugins could use this as well, with this change.

gxalpha:
I know, that was more meant as „In case this is only needed for a specific plugin which happens to also depend on Qt, you could use this workaround“

UI/api-interface.cpp
```diff
@@ -620,6 +620,12 @@ struct OBSStudioAPI : obs_frontend_callbacks {
 		return bstrdup(recordOutputPath);
 	}
 
+	char *obs_frontend_get_translated_string(const char *string) override
+	{
+		QString out = QTStr(string);
```
exeldro, line 625: Why the conversion to QString? I think its better to return a const char* and use:
`return App()->GetString(string);`
or
`return Str(string);`

UI/api-interface.cpp
```diff
@@ -620,6 +620,12 @@ struct OBSStudioAPI : obs_frontend_callbacks {
 		return bstrdup(recordOutputPath);
 	}
 
+	char *obs_frontend_get_translated_string(const char *string) override
```
dodgepong, line 623: Is `obs_frontend_get_translated_string` the right name here? Should it be `obs_frontend_get_locale_string` like `obs_module_get_locale_string`?

cg2121: Done

UI/api-interface.cpp
cg2121: Done

UI/api-interface.cpp
```diff
@@ -620,6 +620,11 @@ struct OBSStudioAPI : obs_frontend_callbacks {
 		return bstrdup(recordOutputPath);
 	}
 
+	char *obs_frontend_get_locale_string(const char *string) override
+	{
+		return bstrdup(Str(string));
```
jp9000, line 625: There's probably no need to duplicate the string. Pretty sure the original pointer is safe in terms of lifetime. This whole thing should probably not duplicate, and the return should probably be const.

PR #6128, opened 2022-03-11
UI: Fix projectors not closing when source is deleted

cg2121:
### Description
The projectors would still hold on to a reference of a source, even if the source was deleted, so use weak sources

### Motivation and Context
A source projector would still be opened, even when its source is deleted.

### How Has This Been Tested?
- Added source projector
- Deleted the source
- Made sure the projector was closed properly

### Types of changes

### Checklist:

### DIFF
--- a/UI/window-projector.cpp
+++ b/UI/window-projector.cpp
@@ -17,11 +17,12 @@ static bool updatingMultiview = false, mouseSwitching, transitionOnDoubleClick;
 
 OBSProjector::OBSProjector(QWidget *widget, obs_source_t *source_, int monitor,
 			   ProjectorType type_)
-	: OBSQTDisplay(widget, Qt::Window),
-	  source(source_),
-	  removedSignal(obs_source_get_signal_handler(source), "remove",
-			OBSSourceRemoved, this)
+	: OBSQTDisplay(widget, Qt::Window), weakSource(OBSGetWeakRef(source_))
 {
+	OBSSource source = GetSource();
+	destroyedSignal.Connect(obs_source_get_signal_handler(source),
+				"destroy", OBSSourceDestroyed, this);
+
 	isAlwaysOnTop = config_get_bool(GetGlobalConfig(), "BasicWindow",
 					"ProjectorAlwaysOnTop");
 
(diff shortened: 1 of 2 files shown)


jp9000:
I don't think you need to have the remove signal changes. Just check to see if GetSource returns null and close the projector if that ends up being the case.

cg2121:
It would benefit in other instances where the remove signal is used, as it only works for scenes not other sources.

tt2468:
I had previously made a PR to call `source_remove` when a source is destructed, but after some discussion decided against trying to pursue it. My suggestion is to make sure the source is stored as a weak ref, then listen to the `source_destroy` signal to know when to close the projector.

cg2121:
I've updated the PR to remove the source_remove signal change. I've changed the signal in the projector class to use 'destroy' instead of 'remove', per suggestion of @tt2468.

UI/window-projector.cpp
```diff
@@ -17,11 +17,14 @@ static bool updatingMultiview = false, mouseSwitching, transitionOnDoubleClick;
...
-	: OBSQTDisplay(widget, Qt::Window),
-	  source(source_),
-	  removedSignal(obs_source_get_signal_handler(source), "remove",
-			OBSSourceRemoved, this)
+	: OBSQTDisplay(widget, Qt::Window), weakSource(OBSGetWeakRef(source_))
 {
+	OBSSource source = GetSource();
+
+	if (source)
+		removedSignal.Connect(obs_source_get_signal_handler(source),
```
tt2468, line 25: Since we're using `destroy` instead of `remove`, we should probably change the name of this. Other than that, this looks good provided we don't get memory leaks.

cg2121: Updated the name of the function and the signal.

UI/window-projector.cpp
```diff
@@ -371,7 +373,7 @@ void OBSProjector::UpdateProjectorTitle(QString name)
 
 OBSSource OBSProjector::GetSource()
 {
-	return source;
+	return OBSGetStrongRef(weakSource);
```
tt2468, line 376: Wasn't able to find an answer myself. Does this increment the source ref when returning? Seems like it would increment the strong ref then return.

cg2121: No, it does not. No memory leaks occur. It is also done this way in media-controls.cpp. https://github.com/obsproject/obs-studio/blob/2ca0e7cc1af0314d4564f501b956fd58919ac30b/UI/media-controls.cpp#L286

tt2468 (approved):

RytoEX:
Merging based on approval above.

Issue #5791, opened 2022-01-13
SRT: OBS unusable and uncloseable after starting stream to invalid srt server


basisbit:
### Operating System Info

Windows 10

### OBS Studio Version

27.1.3

### OBS Studio Log URL

https://obsproject.com/logs/i0MmL1JbOIWU7u7x

### Expected Behavior

- OBS should not show that it connected successful when the srt stream server URL is invalid, but instead show the usual stream error message
- User should be able to stop streaming and later be able to start streaming again after correcting the srt server url
- User should be able to close OBS

### Current Behavior

- OBS shows that it connected successfully, stream bitrate counter stays at 0b/s. After some time it shows the `OBS Studio: Disconnected. Reconnecting ...` notification.
- Pressing the `Stop streaming` button only makes its text change to `Stopping Stream...` and then OBS is unusable for any further streaming.
- OBS then can not be closed any more. Or rather, if you press the close button or File -> Exit the window disappears but the obs.exe process stays open infinitely until manually killed using Task Manager or restarting the computer
- Dropped Frame counter seems to be not working at all when using SRT, and `Dynimaically change bitrate to manage congestion (Beta)` also does not work when using srt. (However, srt support seems to be rather new in OBS so far and SRT already handles packet loss very well, so those two features are absolutely low priority imho.)

![image](https://user-images.githubusercontent.com/10870269/149400185-40a1e784-e904-4976-9825-6cbe978e6351.png)

### Steps to Reproduce

1. Set Settings -> Stream -> Server to `srt://google.com:443/this/is/not/a/srt/server?latency=500000`
2. Set Settings -> Advanced -> `Automatically Reconnect` to Enabled with Retry Delay `10s`.
2. Close settings. Start streaming. Watch OBS for 30 seconds.
3. Press the `Stop streaming` button, then wait 30 seconds.
4. Close OBS, wait 30 seconds, then open Task Manager and check in the Details/Processes tab to see OBS still being run.

### Anything else we should know?

Streaming using SRT is infinitely more reliable than using rtmp, especially when network congestion happens on some peerings between the streamer and the ingest server, or when short bursts of packet loss happen for example on LTE/5G or Wifi/WLAN or Docsis based shared last mile. OBS adding SRT support some time in 2020 (afaik) meant a huge improvement in reliability when using self-hosted or small video streaming distribution services which don't have ingest servers in every country close to the user. Thank you so much for adding it!

pkviet:
[...]

PR #6134, opened 2022-03-12
obs-ffmpeg: Support native SRT & RIST output

pkviet:
### Description
This PR is a stand in replacement of the obs-ffmpeg-mpegts output
which currently relies on a pipe to obs-ffmpeg-mux binary.
This improves srt/rist handling.
Additionally this fixes several bugs related to auto-reconnect : 
- auto-reconnect is partially broken with srt & rist; the UI will display a successful reconnect even though it is not the case
because in obs-ffmpeg-mux.c the obs_output_can_begin_data_capture function is not called in the thread doing the reconnect. Fixes: https://github.com/obsproject/obs-studio/issues/6749 
- when non existing urls are used, if auto-reconnect is enabled, obs is locked up.  Fixes: https://github.com/obsproject/obs-studio/issues/5791

### Motivation and Context
Currently srt & rist protocols are supported in obs through their implementations
in FFmpeg/avformat.
It is much preferable to support them natively in obs because this allows for better
logging and easier debugging. We also don't depend on upstream for bug fixes
or for functionalities.
This PR achieves this in the following way:
- it creates an obs-ffmpeg-mpegts muxer separate from obs-ffmpeg-mux project;
The use of obs-ffmpeg-mux allowed for easy support of srt & rist but logging was
an issue.
- it brings the protocol handling in-house by using custom AVIO_Contexts.

Note that the mpegts muxing is still delegated to avformat library.
The mpegts spec is quite complex so it doesn't seem reasonable to do the mpegts
muxing ourselves.
However all the protocol handling is now done in obs itself.

Improvements over the previous srt/rist implementations are the following:
- logging of all protocol stages (opening, writing, closing, failures);
- stats collections (the PR collects stats every 60 seconds; this is not configurable, yet).
The stats inform on the RTT and the link bandwidth (effective bandwidth from obs to ingest)
for srt for instance. The figures can be quite useful for optimization of the stream. They could
also be leveraged for dynamic bitrate.

Further improvements are possible but are dependent on whether a better support of these 
protocols is decided upon by the obs dev team. For instance:
- better customize protocol settings (at the plugin level) and create an UI for that.
- add support for multi-track streaming (audio) (this is mostly at UI level).
- add support for opus codec (requires some UI work in obs Settings).
- support rist multi-destinations (rist can stream to up to 10 ingest points).
- support bonding (link aggregation for srt & rist).

### How Has This Been Tested?
This was tested on windows 10 & macos 10.15.4 against : 
- nimble server,
- wowza server,
- vlc 3 & vlc 4,
- ffplay & ffmpeg.

Works fine.

NOT tested on linux.

TEST builds here: https://github.com/obsproject/obs-studio/actions/runs/2656507203

**Important note on deps**
macos obs-deps require libsrt 1.4.4 and break stats collection with 1.4.1.
The first commit will be dropped once it's merged into master; it enables the right deps which work with this PR.

**CI**
CI is currently broken for various reasons on linux, ~~mac & win~~.
On win: ~~the deps need to ship the import libs of libsrt & librist. This will require an update of obs-deps.~~
On macOS: ~~the deps do not ship the srt & rist dylibs. This will be fixed by the obs-deps PR mentioned above.~~
On ubuntu 18.04, the CI script needs to be updated to manually build libsrt & librist. (The ubuntu version is 
too old to have recent enough packages for srt & rist.)

### Types of changes

### Checklist:

### DIFF
--- a/plugins/obs-ffmpeg/CMakeLists.txt
+++ b/plugins/obs-ffmpeg/CMakeLists.txt
@@ -1,6 +1,7 @@
 project(obs-ffmpeg)
 
 option(ENABLE_FFMPEG_LOGGING "Enables obs-ffmpeg logging" OFF)
+option(ENABLE_NEW_MPEGTS_OUTPUT "Use native SRT/RIST mpegts output" ON)
 
 find_package(
   FFmpeg REQUIRED
@@ -15,9 +16,11 @@ find_package(
 add_library(obs-ffmpeg MODULE)
 add_library(OBS::ffmpeg ALIAS obs-ffmpeg)
 
-find_package(Librist QUIET)
-find_package(Libsrt QUIET)
 add_subdirectory(ffmpeg-mux)
+if(ENABLE_NEW_MPEGTS_OUTPUT)
+  find_package(Librist)
+  find_package(Libsrt)
+endif()
 
 configure_file(${CMAKE_CURRENT_SOURCE_DIR}/obs-ffmpeg-config.h.in
                ${CMAKE_BINARY_DIR}/config/obs-ffmpeg-config.h)
@@ -30,17 +33,13 @@ target_sources(
           obs-ffmpeg-av1.c
           obs-ffmpeg-nvenc.c
           obs-ffmpeg-output.c
-          obs-ffmpeg-mpegts.c
           obs-ffmpeg-mux.c
           obs-ffmpeg-mux.h
           obs-ffmpeg-hls-mux.c
           obs-ffmpeg-source.c
           obs-ffmpeg-compat.h
           obs-ffmpeg-formats.h
-          ${CMAKE_BINARY_DIR}/config/obs-ffmpeg-config.h
-          obs-ffmpeg-srt.h
-          obs-ffmpeg-rist.h
-          obs-ffmpeg-url.h)
+          ${CMAKE_BINARY_DIR}/config/obs-ffmpeg-config.h)
 
 target_include_directories(obs-ffmpeg PRIVATE ${CMAKE_BINARY_DIR}/config)
 
@@ -55,9 +54,18 @@ target_link_libraries(
           FFmpeg::avdevice
           FFmpeg::avutil
           FFmpeg::swscale
-          FFmpeg::swresample
-          Librist::Librist
-          Libsrt::Libsrt)
+          FFmpeg::swresample)
+
+if(ENABLE_NEW_MPEGTS_OUTPUT)
+  target_sources(obs-ffmpeg PRIVATE obs-ffmpeg-mpegts.c obs-ffmpeg-srt.h
+                                    obs-ffmpeg-rist.h obs-ffmpeg-url.h)
+
+  target_link_libraries(obs-ffmpeg PRIVATE Librist::Librist Libsrt::Libsrt)
+  if(OS_WINDOWS)
+    target_link_libraries(obs-ffmpeg PRIVATE ws2_32.lib)
+  endif()
+  target_compile_definitions(obs-ffmpeg PRIVATE NEW_MPEGTS_OUTPUT)
+endif()
 
 if(ENABLE_FFMPEG_LOGGING)
   target_sources(obs-ffmpeg PRIVATE obs-ffmpeg-logging.c)
@@ -77,7 +85,6 @@ if(OS_WINDOWS)
   if(MSVC)
     target_link_libraries(obs-ffmpeg PRIVATE OBS::w32-pthreads)
   endif()
-    target_link_libraries(obs-ffmpeg PRIVATE ws2_32.lib)
 
   set(MODULE_DESCRIPTION "OBS FFmpeg module")
   configure_file(${CMAKE_SOURCE_DIR}/cmake/bundle/windows/obs-module.rc.in
--- a/plugins/obs-ffmpeg/CMakeLists.txt
+++ b/plugins/obs-ffmpeg/CMakeLists.txt
@@ -14,6 +14,9 @@ find_package(
 
 add_library(obs-ffmpeg MODULE)
 add_library(OBS::ffmpeg ALIAS obs-ffmpeg)
+
+find_package(Librist QUIET)
+find_package(Libsrt QUIET)
 add_subdirectory(ffmpeg-mux)
 
 configure_file(${CMAKE_CURRENT_SOURCE_DIR}/obs-ffmpeg-config.h.in
@@ -27,13 +30,17 @@ target_sources(
           obs-ffmpeg-av1.c
           obs-ffmpeg-nvenc.c
           obs-ffmpeg-output.c
+          obs-ffmpeg-mpegts.c
           obs-ffmpeg-mux.c
           obs-ffmpeg-mux.h
           obs-ffmpeg-hls-mux.c
           obs-ffmpeg-source.c
           obs-ffmpeg-compat.h
           obs-ffmpeg-formats.h
-          ${CMAKE_BINARY_DIR}/config/obs-ffmpeg-config.h)
+          ${CMAKE_BINARY_DIR}/config/obs-ffmpeg-config.h
+          obs-ffmpeg-srt.h
+          obs-ffmpeg-rist.h
+          obs-ffmpeg-url.h)
 
 target_include_directories(obs-ffmpeg PRIVATE ${CMAKE_BINARY_DIR}/config)
 
@@ -48,7 +55,9 @@ target_link_libraries(
           FFmpeg::avdevice
           FFmpeg::avutil
           FFmpeg::swscale
-          FFmpeg::swresample)
+          FFmpeg::swresample
+          Librist::Librist
+          Libsrt::Libsrt)
 
 if(ENABLE_FFMPEG_LOGGING)
   target_sources(obs-ffmpeg PRIVATE obs-ffmpeg-logging.c)
@@ -68,6 +77,7 @@ if(OS_WINDOWS)
   if(MSVC)
     target_link_libraries(obs-ffmpeg PRIVATE OBS::w32-pthreads)
   endif()
+    target_link_libraries(obs-ffmpeg PRIVATE ws2_32.lib)
 
   set(MODULE_DESCRIPTION "OBS FFmpeg module")
   configure_file(${CMAKE_SOURCE_DIR}/cmake/bundle/windows/obs-module.rc.in
--- a/plugins/obs-ffmpeg/obs-ffmpeg-output.h
+++ b/plugins/obs-ffmpeg/obs-ffmpeg-output.h
@@ -5,12 +5,14 @@
 #include <libavcodec/avcodec.h>
 #include <libavformat/avformat.h>
 #include <libswscale/swscale.h>
+#include "obs-ffmpeg-url.h"
 
 struct ffmpeg_cfg {
 	const char *url;
 	const char *format_name;
 	const char *format_mime_type;
 	const char *muxer_settings;
+	const char *protocol_settings; // not used yet for SRT nor RIST
 	int gop_size;
 	int video_bitrate;
 	int audio_bitrate;
@@ -32,6 +34,7 @@ struct ffmpeg_cfg {
 	int scale_height;
 	int width;
 	int height;
+	int frame_size; // audio frame size
 };
 
 struct ffmpeg_audio_info {
@@ -74,5 +77,38 @@ struct ffmpeg_data {
 	char *last_error;
 };
 
+struct ffmpeg_output {
+	obs_output_t *output;
+	volatile bool active;
+	struct ffmpeg_data ff_data;
+
+	bool connecting;
+	pthread_t start_thread;
+
+	uint64_t total_bytes;
+
+	uint64_t audio_start_ts;
+	uint64_t video_start_ts;
+	uint64_t stop_ts;
+	volatile bool stopping;
+
+	bool write_thread_active;
+	pthread_mutex_t write_mutex;
+	pthread_t write_thread;
+	os_sem_t *write_sem;
+	os_event_t *stop_event;
+
+	DARRAY(AVPacket *) packets;
+	/* used for SRT & RIST */
+	URLContext *h;
+	AVIOContext *s;
+	bool got_headers;
+};
 bool ffmpeg_data_init(struct ffmpeg_data *data, struct ffmpeg_cfg *config);
 void ffmpeg_data_free(struct ffmpeg_data *data);
+
+#define SRT_PROTO "srt"
+#define UDP_PROTO "udp"
+#define TCP_PROTO "tcp"
+#define HTTP_PROTO "http"
+#define RIST_PROTO "rist"
--- a/plugins/obs-ffmpeg/obs-ffmpeg-output.h
+++ b/plugins/obs-ffmpeg/obs-ffmpeg-output.h
@@ -5,7 +5,9 @@
 #include <libavcodec/avcodec.h>
 #include <libavformat/avformat.h>
 #include <libswscale/swscale.h>
+#ifdef NEW_MPEGTS_OUTPUT
 #include "obs-ffmpeg-url.h"
+#endif
 
 struct ffmpeg_cfg {
 	const char *url;
@@ -99,16 +101,12 @@ struct ffmpeg_output {
 	os_event_t *stop_event;
 
 	DARRAY(AVPacket *) packets;
+#ifdef NEW_MPEGTS_OUTPUT
 	/* used for SRT & RIST */
 	URLContext *h;
 	AVIOContext *s;
 	bool got_headers;
+#endif
 };
 bool ffmpeg_data_init(struct ffmpeg_data *data, struct ffmpeg_cfg *config);
 void ffmpeg_data_free(struct ffmpeg_data *data);
-
-#define SRT_PROTO "srt"
-#define UDP_PROTO "udp"
-#define TCP_PROTO "tcp"
-#define HTTP_PROTO "http"
-#define RIST_PROTO "rist"
--- a/plugins/obs-ffmpeg/obs-ffmpeg-srt.h
+++ b/plugins/obs-ffmpeg/obs-ffmpeg-srt.h
@@ -1,0 +1,834 @@
+/*
+ * The following code is a port of FFmpeg/avformat/libsrt.c for obs-studio.
+ * Port by pkv@obsproject.com.
+ *
+ * Permission to use, copy, modify, and distribute this software for any
+ * purpose with or without fee is hereby granted, provided that the above
+ * copyright notice and this permission notice appear in all copies.
+ *
+ * THE SOFTWARE IS PROVIDED "AS IS" AND THE AUTHOR DISCLAIMS ALL WARRANTIES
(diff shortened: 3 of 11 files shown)


pkviet:
PR updated to allow fallback to previous mpegts output.
CI updated to use old mpegts output for linux (ubuntu 18.04).
Note: the CI linux artefacts are not used. We only ship flatpak & manually built deb packages.

plugins/obs-ffmpeg/CMakeLists.txt
```diff
@@ -1,6 +1,7 @@
 project(obs-ffmpeg)
 
 option(ENABLE_FFMPEG_LOGGING "Enables obs-ffmpeg logging" OFF)
+option(USE_NEW_MPEGTS_OUTPUT "Use native srt/rist mpegts output" ON)
```
tytan652, line 4: I think this option should follow the same convention as others (`ENABLE_`) and be named `ENABLE_NEW_MPEGTS`.

plugins/obs-ffmpeg/CMakeLists.txt
```diff
@@ -14,7 +15,11 @@ find_package(
 
 add_library(obs-ffmpeg MODULE)
 add_library(OBS::ffmpeg ALIAS obs-ffmpeg)
-add_subdirectory(ffmpeg-mux)
```
tytan652, line 17: Is this normal ? It prevents CMake to generate a `obs-ffmpeg-mux` binary.

tytan652 (commented):
Aside my comments about `CMakeLists.txt`.

Why `obs-ffmpeg-rist.h` and `obs-ffmpeg-srt.h` are full of complete function definitions rather than having them in their `.c` file counterpart and just declaration in the `.h` ?

plugins/obs-ffmpeg/CMakeLists.txt
pkviet: cmake overhaul had not been merged iirc.
But yeah, it's better to follow the pattern of other options.

plugins/obs-ffmpeg/CMakeLists.txt
pkviet: goof

pkviet:
> Aside my comments about `CMakeLists.txt`.
> 
> Why `obs-ffmpeg-rist.h` and `obs-ffmpeg-srt.h` are full of complete function definitions rather than having them in their `.c` file counterpart and just declarations in the `.h` ?

just lazyness. Works fine as is.

pkviet:
Updated to reflect comments by @tytan652

plugins/obs-ffmpeg/CMakeLists.txt
```diff
@@ -14,7 +15,12 @@ find_package(
 
 add_library(obs-ffmpeg MODULE)
 add_library(OBS::ffmpeg ALIAS obs-ffmpeg)
+
 add_subdirectory(ffmpeg-mux)
+if(USE_NEW_MPEGTS_OUTPUT)
```
tytan652, line 20: Here

plugins/obs-ffmpeg/CMakeLists.txt
```diff
@@ -50,6 +56,15 @@ target_link_libraries(
           FFmpeg::swscale
           FFmpeg::swresample)
 
+if(USE_NEW_MPEGTS_OUTPUT)
```
tytan652, line 59: And here

tytan652 (commented):
You forgot to rename those two.

plugins/obs-ffmpeg/CMakeLists.txt
pkviet: damnit!
thx

plugins/obs-ffmpeg/CMakeLists.txt
pkviet: thx

tt2468:
Hmm, so I'm looking at the way this is implemented, and I'm not quite sure if it's the right way around. What dependencies exist for this to be in obs-ffmpeg, aside from libav itself? If we're using the mpegts muxer that we already use anyway, what benefit is there over relying on ffmpeg's SRT and RIST implementations? Keep in mind, I'm coming from almost no context other than the fact that we're using the ffmpeg mux in a dirty way as a subprocess.

I would even be tempted to suggest a restructure of obs-ffmpeg first to separate components like the ffmpeg source into their own directories, among other code cleanup, before we go stacking more functionality on top of it.

pkviet:
> Hmm, so I'm looking at the way this is implemented, and I'm not quite sure if it's the right way around. What dependencies exist for this to be in obs-ffmpeg, aside from libav itself?
We still use ffmpeg mpegts muxer.
The mpegts spec is very complex so we don't want to write our own mpegts muxer.
There's an open source library written by Kieran Kunhya for mpegts which is better than avformat one but it's way simpler to directly use ffmpeg.

>  If we're using the mpegts muxer that we already use anyway, what benefit is there over relying on ffmpeg's SRT and RIST implementations? 

native means we can track and solve bugs directly without reliance on upstream fixing the bugs ages after they're uncovered.

> Keep in mind, I'm coming from almost no context other than the fact that we're using the ffmpeg mux in a dirty way as a subprocess. I would even be tempted to suggest a restructure of obs-ffmpeg first to separate components like the ffmpeg source into their own directories, among other code cleanup, before we go stacking more functionality on top of it.

this doesn't use ffmpeg-mux subprocess.

basisbit:
> If we're using the mpegts muxer that we already use anyway, what benefit is there over relying on ffmpeg's SRT and RIST implementations?

From my understanding, most noteworthy would be:
- being able to track frame-loss and respond to that for example by showing it in the bottom right in the UI as well as lowering encoding bitrate if that setting is enabled in options,
- being able to do usable error handling and giving the user actually helpful error messages,
- hopefully fix those memory-leaks and process-leaks which currently happen when srt streaming fails or aborts or fails to initialize, see #5791

The current implementation which is based on `relying on ffmpeg's SRT and RIST implementation` is quite a pita in real-world-usage. (But it still is way more reliable than RTMP, as long as the streamer knows those bugs in current OBS, so please DO NOT remove SRT support...)

plugins/obs-ffmpeg/obs-ffmpeg-output.h
```diff
@@ -74,5 +79,34 @@ struct ffmpeg_data {
...
+	pthread_mutex_t write_mutex;
+	pthread_t write_thread;
+	os_sem_t *write_sem;
+	os_event_t *stop_event;
+
+	DARRAY(AVPacket *) packets;
+#ifdef NEW_MPEGTS_OUTPUT
+	/* used for srt & rist */
+	URLContext *h;
+	AVIOContext *s;
```
notr1ch, line 107: Could these perhaps be friendlier named variables? `h` and `s` don't give much context when looking at the code.

plugins/obs-ffmpeg/obs-ffmpeg-srt.h
```diff
@@ -0,0 +1,836 @@
...
+ */
+
+#pragma once
+#include <obs-module.h>
+#include "obs-ffmpeg-url.h"
+#include <srt/srt.h>
+#include <libavformat/avformat.h>
+#ifdef _WIN32
+#pragma comment(lib, "ws2_32.lib")
+#endif
```
notr1ch, line 25: Shouldn't this be part of the CMakefile?

plugins/obs-ffmpeg/obs-ffmpeg-srt.h
```diff
@@ -0,0 +1,836 @@
...
+			       &s->payload_size,
+			       sizeof(s->payload_size)) < 0) ||
+	    (/*(h->flags & AVIO_FLAG_WRITE) &&*/
+	     libsrt_setsockopt(h, fd, SRTO_SENDER, "SRTO_SENDER", &yes,
+			       sizeof(yes)) < 0) ||
+	    (s->tsbpd >= 0 &&
+	     libsrt_setsockopt(h, fd, SRTO_TSBPDMODE, "SRTO_TSBPDMODE",
+			       &s->tsbpd, sizeof(s->tsbpd)) < 0)) {
+		return AVERROR(EIO);
+	}
```
notr1ch, line 406: This feels like it should be a macro or helper function, way too much code for a single statement.

plugins/obs-ffmpeg/obs-ffmpeg-output.h
pkviet: although the plan is eventually to add more and more of our own code, I've kept on purpose the current code very close to avformat/libsrt.c .
This will allow us to port more easily changes/bugfixes done in avformat.
At some point when the divergence will be too large, yeah I agree it'd make sense to have more friendly names.

plugins/obs-ffmpeg/obs-ffmpeg-srt.h
pkviet: i've moved it back to cmake where it was initially; can't remember why i moved it here, was 5 months ago ...

plugins/obs-ffmpeg/obs-ffmpeg-srt.h
pkviet: the conditions for the options are different; some must be >=0 , >=-1, or just exist. Not sure how to go about a macro. The code is strictly that in avformat/libsrt.c . I didn't change anything and maybe for the initial merge it's better to stay close to the source.

basisbit:
@pkviet that build for windows x64 looks good so far. Only edge case I got an issue with is: set a valid srt stream url as server address, change its port number to something that is firewalled (action: drop) by the target server, so that OBS never gets a response from the server. This causes OBS to show `Connecting...` without any error message or any timeout (waited a long time for it) and then only closing and reopening OBS makes it usable again.
image

SRT streaming to OvenMediaEngine as SRT server worked well. Having an invalid dns name as part of the SRT URL instantly shows an error message that it couldn't connect, and that also works as expected when auto-reconnect is enabled. 🥳

pkviet:
> @pkviet that build for windows x64 looks good so far. Only edge case I got an issue with is: set a valid srt stream url as server address, change its port number to something that is firewalled (action: drop) by the target server, so that OBS never gets a response from the server. This causes OBS to show `Connecting...` without any error message or any timeout (waited a long time for it) and then only closing and reopening OBS makes it usable again. image
> 

Yeah that's something I need to look into; implement disconnect upon timeout.

pkviet:
> @pkviet that build for windows x64 looks good so far. Only edge case I got an issue with is: set a valid srt stream url as server address, change its port number to something that is firewalled (action: drop) by the target server, so that OBS never gets a response from the server. This causes OBS to show `Connecting...` without any error message or any timeout (waited a long time for it) and then only closing and reopening OBS makes it usable again. image
> 
as a follow-up, it does disconnect fine unless auto-reconnect is enabled; the time out is implemented correctly. Did you have auto-reconnect enabled ? if you did, then what you get is normal.
The auto-reconnect is set to retry in an exponential way. It retries after 10sec, then 20 later etc ... If you use the current default settings for auto-reconnect , it'll stop after 242 days ...
On master, it's a shorter time now but still a huge value.
So try with saner values. I personally use 2 sec and 4 reconnect attempts ...

pkviet:
PR updated to add hevc support.

erickskrauch:
@pkviet, I apologize for the stupid question, but I don't understand where and how I can select the HEVC for SRT? I downloaded the latest build from CI, but unfortunately, I never found anything new related to adding native SRT support using the HEVC codec.

We use SRT broadcasts a lot for multi-camera broadcasts and are enthusiastic about this PR. We will be glad to take part in testing :)

tytan652:
ENABLE_HEVC build flag needs to be set to ON to build OBS with HEVC hardware encoder support.

1div0:
$700P1D $0F7W4R3 P473N7$.!?

plugins/obs-ffmpeg/CMakeLists.txt
```diff
@@ -1,6 +1,7 @@
 project(obs-ffmpeg)
 
 option(ENABLE_FFMPEG_LOGGING "Enables obs-ffmpeg logging" OFF)
+option(ENABLE_NEW_MPEGTS_OUTPUT "Use native srt/rist mpegts output" ON)
```
RytoEX, line 4: Nit: SRT/RIST

plugins/obs-ffmpeg/obs-ffmpeg-output.h
```diff
@@ -5,12 +5,16 @@
...
+#ifdef NEW_MPEGTS_OUTPUT
+#include "obs-ffmpeg-url.h"
+#endif
 
 struct ffmpeg_cfg {
 	const char *url;
 	const char *format_name;
 	const char *format_mime_type;
 	const char *muxer_settings;
+	const char *protocol_settings; // not used for srt nor rist
```
RytoEX, line 17: Nit: SRT nor RIST

plugins/obs-ffmpeg/obs-ffmpeg-output.h
```diff
@@ -32,6 +36,7 @@ struct ffmpeg_cfg {
 	int scale_height;
 	int width;
 	int height;
+	int frame_size; //audio frame size
```
RytoEX, line 39: Nit: `// audio`

plugins/obs-ffmpeg/obs-ffmpeg-output.h
```diff
@@ -74,5 +79,34 @@ struct ffmpeg_data {
...
+
+	bool write_thread_active;
+	pthread_mutex_t write_mutex;
+	pthread_t write_thread;
+	os_sem_t *write_sem;
+	os_event_t *stop_event;
+
+	DARRAY(AVPacket *) packets;
+#ifdef NEW_MPEGTS_OUTPUT
+	/* used for srt & rist */
```
RytoEX, line 105: Nit: SRT & RIST

plugins/obs-ffmpeg/obs-ffmpeg-url.h
```diff
@@ -0,0 +1,143 @@
...
+	void *priv_data; /* SRTContext or RISTContext */
+	char *url;       /* URL */
+	int max_packet_size;
+	AVIOInterruptCB interrupt_callback;
+	int64_t rw_timeout; /* max time to wait for write completion in mcs */
+} URLContext;
+
+#define UDP_DEFAULT_PAYLOAD_SIZE 1316
+
+/* we need to override libsrt/win/syslog_defs.h due to conflicts w/ some libobs
```
RytoEX, line 29: Nit: We

plugins/obs-ffmpeg/obs-ffmpeg-url.h
```diff
@@ -0,0 +1,143 @@
...
+#define LOG_LOCAL6 (22 << 3)
+#define LOG_LOCAL7 (23 << 3)
+
+#define LOG_NFACILITIES 24
+#define LOG_FACMASK 0x03f8
+#define LOG_FAC(p) (((p)&LOG_FACMASK) >> 3)
+#endif
+#endif
+
+/* we need to override libsrt/logging_api.h due to conflicts with some libobs
```
RytoEX, line 82: Nit: We

plugins/obs-ffmpeg/obs-ffmpeg-url.h
```diff
@@ -0,0 +1,143 @@
...
+#define LOG_DEBUG_TRACE 8
+#endif
+
+// Flags
+#define SRT_LOGF_DISABLE_TIME 1
+#define SRT_LOGF_DISABLE_THREADNAME 2
+#define SRT_LOGF_DISABLE_SEVERITY 4
+#define SRT_LOGF_DISABLE_EOL 8
+
+// Handler type.
```
RytoEX, line 112: Nit: No period needed. This is a label rather than a full sentence.

RytoEX (changes requested):
Mostly nits.

Will try to test this soon, probably OBS-to-OBS unless there's an easy test I can do to some other ingest.

plugins/obs-ffmpeg/CMakeLists.txt
pkviet: done

plugins/obs-ffmpeg/obs-ffmpeg-output.h
pkviet: done

plugins/obs-ffmpeg/obs-ffmpeg-output.h
pkviet: done

plugins/obs-ffmpeg/obs-ffmpeg-url.h
pkviet: done

plugins/obs-ffmpeg/obs-ffmpeg-url.h
pkviet: done

plugins/obs-ffmpeg/obs-ffmpeg-url.h
pkviet: done

plugins/obs-ffmpeg/obs-ffmpeg-output.h
pkviet: done

RytoEX (approved):
Tested and discussed off-thread with @pkviet . Both SRT and RIST output work as expected with this PR. SRT behaves the same as before, with consideration needed for its client parameters, such as latency if you are sending to an ingest far away.

Functionally, this worked fine in my tests, so I am fine merging this unless there are major objections that must be addressed before merge.

basisbit:
> [...] SRT behaves the same as before, with consideration needed for its client parameters, such as latency if you are sending to an ingest far away.

Do you mean that it is still possible to specify the srt parameters as part of the srt url, as before, or has that changed? Not being able to specify the latency parameter (buffer size on the receiving side) would be a breaking change / loss of critical feature imho, especially because SRT tends to be used when packet loss or high latency or bufferbloat makes rtmp unusable...

RytoEX:
> > [...] SRT behaves the same as before, with consideration needed for its client parameters, such as latency if you are sending to an ingest far away.
> 
> Do you mean that it is still possible to specify the srt parameters as part of the srt url, as before, or has that changed? Not being able to specify the latency parameter (buffer size on the receiving side) would be a breaking change / loss of critical feature imho, especially because SRT tends to be used when packet loss or high latency or bufferbloat makes rtmp unusable...

I simply mean that the behavior of the protocol itself is unchanged. If you are streaming to an ingest that is far away (has a large RTT), then you must specify parameters, such as latency, to deal with it. This PR does not change that behavior and does not take away the ability to specify such parameters.

basisbit:
thanks for the explanation. then all good! 👍

eric:
Are there plans to have dropped packets from SRT reflected in the Dropped metric visible in OBS?

pkviet:
> Are there plans to have dropped packets from SRT reflected in the Dropped metric visible in OBS?

Didn't think of it tbh.
Feel free. I don't use srt so my future contributions will probably be limited. i wrote this pr because i was not satisfied with the previous output i had authored (relying on ffmpeg-mux).

PR #6139, opened 2022-03-13
UI: Use weak source refs for dialogs

cg2121:
### Description
When a source is deleted, the dialogs still hold a reference to it.

### Motivation and Context
Sources are actually deleted when deleted in the UI, even when the dialogs are still open. The dialogs will now close now on the remove signal (relies on the second commit in #6128). The signals were always added to the dialogs but never worked, as the dialogs would still remain opened when removing a source. 

### How Has This Been Tested?
Went through all of the dialogs to make sure nothing broke.

### Types of changes

### Checklist:


PatTheMav:
Conceptionally this looks correct to me, the changes are also consistent.

@tt2468 any objections to this?

Warchamp7:
Going through some old UI PRs.

This is a good change but will be superceded soon by the audio mixer refactor, and other work that is in progress around the properties window.

PR #6140, opened 2022-03-14
libobs/graphics: Add support for color spaces

jpark37:
### Description
When we add support for HDR, we need to track the active space somewhere. All else being equal, I would have preferred to keep this state on the libobs level, but it's cleaner to put it inside libobs/graphics, so our swap chain abstraction can interact with the color space without additional changes to the public API.

The reason we need an active space is to handle color space conversions at source boundaries. For example, if we have an HDR source being drawn into an SDR canvas, we are going to want to tonemap the HDR source for sanity. The logic for this will come in a later PR.

TODO: Wait for #6069 to be merged.

### Motivation and Context
HDR support is coming.

### How Has This Been Tested?
Code works in HDR branch.

TODO: Test all modified code paths in this PR branch continue to work as expected for SDR.

### Types of changes

### Checklist:


jpark37:
Makes more sense to test/merge changes as part of #6146.

PR #6146, opened 2022-03-15
libobs, UI: Add support for color spaces

jpark37:
### Description
When we add support for HDR, we need to track the active space somewhere. All else being equal, I would have preferred to keep this state on the libobs level, but it's cleaner to put it inside libobs/graphics, so our swap chain abstraction can interact with the color space without additional changes to the public API.

The reason we need an active space is to handle color space conversions at source boundaries. For example, if we have an HDR source being drawn into an SDR canvas, we are going to want to tonemap the HDR source for sanity. The logic for this will come in a later PR.

Add SDR white luminance UI setting, and wire through to main preview display. This is kind of weird because "SDR White Level" really only affects SDR preview on HDR displays for now, and will otherwise be useful only for compositing SDR with HDR in the future. Maybe it should be called "SDR White Level for HDR" but the name already seems kind of long.

HDR swap chains only on Windows for now.

### Motivation and Context
Getting ready to display HDR by displaying SDR on HDR monitors as we would for an HDR canvas.

### How Has This Been Tested?
Tested SDR preview on SDR and HDR monitors. Verified colors are as expected in RenderDoc. Verified increasing and decreasing SDR setting affects preview brightness accordingly.

- [x] SDR/SDR autoswitch drag, 1/2/1/2
- [x] SDR/SDR autoswitch resize, 1/2/1/2, 2/1/2/1
- [x] SDR/HDR autoswitch drag, 1/2/1/2
- [x] SDR/HDR autoswitch resize, 1/2/1/2, 2/1/2/1
- [x] HDR/HDR autoswitch drag, 1/2/1/2
- [x] HDR/HDR autoswitch resize, 1/2/1/2, 2/1/2/1
- [x] HDR/SDR autoswitch drag, 1/2/1/2
- [x] HDR/SDR autoswitch resize, 1/2/1/2, 2/1/2/1
- [x] display settings toggle, SDR/HDR/SDR/HDR
- [x] display settings toggle, HDR/SDR/HDR/SDR
- [x] Check Mac source with filter

### Types of changes

### Checklist:


jpark37:
Verified SDR white level still works after update. This should be good to go.

PR #6156, opened 2022-03-16
libobs: use rwlock instead of mutex for obs_context_data

exeldro:
### Description
use rwlock instead of mutex for obs_context_data

### Motivation and Context
less time threads waiting on each other is better

### How Has This Been Tested?
On windows 64 bit

### Types of changes

### Checklist:

### DIFF
--- a/libobs/obs-video.c
+++ b/libobs/obs-video.c
@@ -59,7 +59,7 @@ static uint64_t tick_sources(uint64_t cur_time, uint64_t last_time)
 	/* ------------------------------------- */
 	/* call the tick function of each source */
 
-	pthread_mutex_lock(&data->sources_mutex);
+	pthread_rwlock_rdlock(&data->sources_rwlock);
 
 	source = data->first_source;
 	while (source) {
@@ -73,7 +73,7 @@ static uint64_t tick_sources(uint64_t cur_time, uint64_t last_time)
 		source = (struct obs_source *)source->context.next;
 	}
 
-	pthread_mutex_unlock(&data->sources_mutex);
+	pthread_rwlock_unlock(&data->sources_rwlock);
 
 	return cur_time;
 }
--- a/libobs/obs.c
+++ b/libobs/obs.c
@@ -634,17 +634,17 @@ static bool obs_init_data(void)
 	pthread_mutex_init_value(&obs->data.displays_mutex);
 	pthread_mutex_init_value(&obs->data.draw_callbacks_mutex);
 
-	if (pthread_mutex_init_recursive(&data->sources_mutex) != 0)
+	if (pthread_rwlock_init(&data->sources_rwlock, NULL) != 0)
 		goto fail;
 	if (pthread_mutex_init_recursive(&data->audio_sources_mutex) != 0)
 		goto fail;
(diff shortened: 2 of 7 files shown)


libobs/obs.c
```diff
@@ -634,17 +634,17 @@ static bool obs_init_data(void)
 	pthread_mutex_init_value(&obs->data.displays_mutex);
 	pthread_mutex_init_value(&obs->data.draw_callbacks_mutex);
 
-	if (pthread_mutex_init_recursive(&data->sources_mutex) != 0)
+	if (pthread_rwlock_init(&data->sources_rwlock, NULL) != 0)
```
tt2468, line 637: Since we're losing recursive mutex locking functionality here, any calls that risk requiring an upgrade from read-only to rw need to assume rw. Unfortunately, this can be in a lot of places. The most common I can gather is anywhere `obs_source_release()` is called, that might trigger a destroy which requires holding a rw lock.

libobs/obs.c
```diff
@@ -1469,7 +1469,7 @@ void obs_enum_sources(bool (*enum_proc)(void *, obs_source_t *), void *param)
 {
 	obs_source_t *source;
 
-	pthread_mutex_lock(&obs->data.sources_mutex);
+	pthread_rwlock_rdlock(&obs->data.sources_rwlock);
```
tt2468, line 1472: Given my previous statement about recursive mutexes, this would either have to be an rwlock, or it would need to make an incremented copy of the sources array, release the lock, then perform the enumeration on the copy.

libobs/obs-video.c
```diff
@@ -59,7 +59,7 @@ static uint64_t tick_sources(uint64_t cur_time, uint64_t last_time)
 	/* ------------------------------------- */
 	/* call the tick function of each source */
 
-	pthread_mutex_lock(&data->sources_mutex);
+	pthread_rwlock_rdlock(&data->sources_rwlock);
```
tt2468, line 62: Dangerous to hold the lock as read-only due to obs_source_release()

tt2468 (commented):
I was able to verify a deadlock using obs-websocket 5.0.0 request batches, sending 200 `GetInputList` requests with a `CreateInput` in the middle in quick succession causes a destroy during a `obs_enum_sources` callback due to an obs_source_release releasing the final reference of a removed source.

tt2468:
I should note that by applying some strategic delays to my script, I can avoid the deadlock and gather performance data still. Benchmarking average delay of 500 batches of 201 requests results in an average of 15.4ms per batch on `master`, while on this PR, the average goes down to 14.5ms. The difference is there, but arguably miniscule. I think maybe it would be best to play it safe and continue using standard recursive mutexes, as this is a very core component of OBS that needs to be as reliable as possible.

notr1ch:
I like the idea of this, but as @tt2468 has pointed out, there are a lot of places where a seemingly read-only path can become read-write. I do think moving away from mutexes to RW locks should be a goal eventually, but perhaps part of a bigger refactoring.

RytoEX:
Per the above concerns, I'm going to close this PR. Feel free to reopen this PR if the above concerns are addressed.

PR #6168, opened 2022-03-18
CMake: Fix targets not being copied into rundir on Windows and Linux

PatTheMav:
### Description
Windows and Linux do not require a contained application bundle to run
and debug OBS - as such targets can and should be copied independently
from the main OBS application target.

Also silences the output of the `install` step that sets up those files.

### Motivation and Context
Allows updating single plugins inside the build tree without the need to build and package the main OBS project.

### How Has This Been Tested?
Checked on Linux and Windows by changing a single plugin to force a recompile, then comparing the checksums of the files copied into the `rundir` to confirm they've been updated.

### Types of changes

### Checklist:

### DIFF
--- a/cmake/Modules/ObsHelpers.cmake
+++ b/cmake/Modules/ObsHelpers.cmake
@@ -39,9 +39,19 @@ function(setup_binary_target target)
   # Set up installation paths for development rundir
   install(
     TARGETS ${target}
-    RUNTIME DESTINATION ${OBS_EXECUTABLE_DESTINATION} COMPONENT obs_rundir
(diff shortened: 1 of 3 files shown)


jpark37:
Works on my machine.

RytoEX (changes requested):
The correct prefix for this is actually `cmake` rather than `CI`.

I believe people confirmed both on-thread and off-thread that this has fixed the issues they were experiencing, so this can be merged once the prefix is correct.

RytoEX (approved):

PR #6172, opened 2022-03-19
cmake: Fix disabling rtmp-services

Chiitoo:
It looks like this option wasn't actually doing anything.

The old logic was quite different, so it's no surprise it got lost in translation.

### Description

Add a check for the option status, and exit if not enabled.

### Motivation and Context

Without this change, 'rtmp-services' is always enabled.

### How Has This Been Tested?

Running configure with ENABLE_SERVICE_UPDATES disabled or enabled, and comparing the output as well as build directories.

Tested on Gentoo Linux.

### Types of changes

### Checklist:

### DIFF
--- a/plugins/rtmp-services/CMakeLists.txt
+++ b/plugins/rtmp-services/CMakeLists.txt
@@ -1,6 +1,10 @@
 project(rtmp-services)
 
 option(ENABLE_SERVICE_UPDATES "Checks for service updates" OFF)
+if(NOT ENABLE_SERVICE_UPDATES)
+  message(STATUS "OBS:  DISABLED   rtmp-services")
+  return()
+endif()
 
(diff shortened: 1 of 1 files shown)


tytan652:
`ENABLE_SERVICE_UPDATES` is not related to the whole `rtmp-services` plugin but to the part of the code that update the `services.json`.

derrod:
On ~~second~~ third thought, I think @tytan652 is correct that this PR is wrong.

Chiitoo:
Hm, right, so it's not an option that's supposed to be toggled really.

Configure shows `-- OBS:   ENABLED   rtmp-services`, so I went looking and this is where it got me.  :]

Testing with a previous release I do see now that it does seem to be working as it did before, only the enabled message would not be shown.

Thanks!

PR #6173, opened 2022-03-19
obs-output: Add presets to dynamic bitrate (rtmp)

pkviet:
### Description
Also modifies UI.
We add three presets:
- dynamic bitrate disabled (default);
- a recommended preset where bitrate recovery is much faster (bitrate increase tested every 2 seconds).
- a preset which corresponds to earlier settings (but with a faster 15 sec recovery).

The recommended preset targets good internet connections with occasional hiccups.
The slower recovery preset is more adapted to challenging internet connections with instabilities.

Furthermore, the minimum bitrate (floor) is set to :
- 30% of initial bitrate for default preset & drop frames instead of going under the floor.
- 50 kbps for the slower recovery preset.

Note that the basic algorithm for dbr has not been changed. Only some of its hard-coded
settings have been changed.

obs64_2022-03-19_21-10-14

devenv_2022-03-19_21-09-59

The old dynamic bitrate settings have been ported to the recommended setting (fast recovery).

### Motivation and Context
The current dynamic bitrate was written with specifications from a major service,
with tests to increase bitrate every 30 seconds.
This was unfortunately very slow for the majority of users because it would mean recovery can
take several minutes.
The service doesn't require from us such long recovery times, which allows us to make some changes.
There's broadly two populations we target : 
- people who have excellent internet (broadband, fiber), very stable but with very rare occasional network
issues. The default preset targets them. The recovery rate is set to 2 seconds so overall recovery from a hiccup
will be very fast.
- people with unstable network; the recovery rate needs to be much slower in order to ensure stable
transmission.

For the presets we also set different minimum bitrates, below which we drop frames again.
For unstable network, the floor is set to 50 kbps (as it was before) because stability is prioritized over
frame drops.
For stable networks, we picked 30% of the initial bitrate, meaning we can absorb bandwidth contractions
down to that value. We don't want to go too low because this would increase the recovery time.
And we know that on these networks, the hiccups are really short and occasional.

### How Has This Been Tested?
This was tested on windows 10 with a network limiter app (NetLimiter 4).
Works as expected.

### Types of changes

### Checklist:


UI/data/locale/en-US.ini
```diff
@@ -1048,6 +1048,9 @@ Basic.Settings.Advanced.Network.BindToIP="Bind to IP"
 Basic.Settings.Advanced.Network.EnableNewSocketLoop="Enable network optimizations"
 Basic.Settings.Advanced.Network.EnableLowLatencyMode="Enable TCP pacing"
 Basic.Settings.Advanced.Network.TCPPacing.Tooltip="Attempts to make RTMP output friendlier to other latency sensitive applications on the network by regulating the rate of transmission.\nIt may increase the risk of dropped frames on unstable connections."
+Basic.Settings.Advanced.Network.Dbr.Presets="Dynamic Bitrate Presets"
+Basic.Settings.Advanced.Network.Dbr.Presets.Fast="Reliable connection with occasional stability issues"
+Basic.Settings.Advanced.Network.Dbr.Presets.Slow="Unreliable connection (ex: mobile)"
```
WizardCM, line 1053:
```suggestion
Basic.Settings.Advanced.Network.Dbr.Presets.Slow="Unreliable connection (eg. mobile)"
```

WizardCM:
I think it might be better to change from having a checkbox for On/Off plus this dropdown to a single dropdown with Off / Reliable / Unreliable, and then migrate existing users Off -> Off and On -> Unreliable.

I'll let @Warchamp7 weigh in on this first though.

UI/data/locale/en-US.ini
pkviet: done

pkviet:
> I think it might be better to change from having a checkbox for On/Off plus this dropdown to a single dropdown with Off / Reliable / Unreliable, and then migrate existing users Off -> Off and On -> Unreliable.
> 
> I'll let @Warchamp7 weigh in on this first though.

i have no preference; the latter choice requires some rework since the previous setting was a bool and it's changed into an int.

pkviet:
Updated the PR to follow @WizardCM recommendation, following @Warchamp7 advice.
The porting of ancient dbr settings has been done through casting bool to int (the c bool in obs is actually an int anyway).
I've chosen to set the 'recommended' preset instead of the slower preset for the porting, since the previous slow recovery of 30 seconds was oftentimes unsuitable. Anyway, changing the preset is one click away so it's a minor matter.

UI/data/locale/en-US.ini
```diff
@@ -1048,6 +1045,11 @@ Basic.Settings.Advanced.Network.BindToIP="Bind to IP"
 Basic.Settings.Advanced.Network.EnableNewSocketLoop="Enable network optimizations"
 Basic.Settings.Advanced.Network.EnableLowLatencyMode="Enable TCP pacing"
 Basic.Settings.Advanced.Network.TCPPacing.Tooltip="Attempts to make RTMP output friendlier to other latency sensitive applications on the network by regulating the rate of transmission.\nIt may increase the risk of dropped frames on unstable connections."
+Basic.Settings.Advanced.Network.Dbr.Presets="Dynamic Bitrate"
+Basic.Settings.Advanced.Network.Dbr.Presets.TT="Instead of dropping frames to reduce congestion, dynamically changes bitrate on the fly.\n\nNote that this can increase delay to viewers if there is significant sudden congestion.\nWhen the bitrate drops, it can take up to a few minutes to restore. \nOn recommended setting, frames will be dropped below 30% of the original bitrate. \nOn the setting for unstable networks, frames are not dropped at all but bitrate can go down as low as 50 kbps.\n\nCurrently only supported for RTMP."
+Basic.Settings.Advanced.Network.Dbr.Presets.Disabled="Dynamic bitrate disabled (frames dropped in case of network congestion)"
```
WizardCM, line 1050:
```suggestion
Basic.Settings.Advanced.Network.Dbr.Presets.Disabled="Disabled (frames dropped in case of network congestion)"
```

notr1ch:
I feel like having this much text in the listbox is bad UX. The tooltip can explain the details, I would much prefer if we could simplify the list to more concrete options instead of terms which could mean different things to different users. I was thinking something like "Disabled, Normal, Aggressive" and leave the explanations to the tooltips (or even a KB link if we think this feature needs significant explanation).

RytoEX:
> I feel like having this much text in the listbox is bad UX. The tooltip can explain the details, I would much prefer if we could simplify the list to more concrete options instead of terms which could mean different things to different users. I was thinking something like "Disabled, Normal, Aggressive" and leave the explanations to the tooltips (or even a KB link if we think this feature needs significant explanation).

I agree with this. I'd prefer short names for the options, and use explanatory text and/or KB link in a tooltip.

pkviet:
Per discussion w/ @notr1ch & @RytoEX I'm changing to less wordy entries : 
disabled / normal / weak for the three options.

WizardCM:
As a user "Normal / Aggressive" and "Normal / Weak" tell me absolutely nothing about what to expect from each option. You might as well just make it "On1 / On2" at that stage.

SuslikV:
Dynamic Bitrate:

- Disabled
- For wired connection
- For wireless connection

or

Unstable Connection:

- Drop frames
- Drop bitrate
- Dramatically drop bitrate

spiderfaced:
Is it possible for this to get working for SRT streaming as well, or is it too complex to alter? 

I'm happy to show my ignorance, but I do recall hearing that OBS Dynamic bitrate is both slow and may only work for RTMP.

Dynamic bitrate in IRL solutions is a huge game changer, especially with HEVC (albeit using an obviously forked version - if anyone knows of onw I'd be happy to know :)

pkviet:
Srt would require a separate implementation.
The two protocols use completely different code.

spiderfaced:
Do you know how the built in and advertised Adaptive bitrate works in vanilla SRT? Is it at all comparable to the RTMP dynamic bitrate in performance or does it need to be working closely with the encoder?

pkviet:
> Do you know how the built in and advertised Adaptive bitrate works in vanilla SRT? Is it at all comparable to the RTMP dynamic bitrate in performance or does it need to be working closely with the encoder?

Got nothing to do with dynamic bitrate.
It's ARQ, automated repeat request, meaning an auto resend of lost packets. It's faster than rtmp which is tcp because it uses udp but improved with some ack exchanges.
Dynamic bitrate decreases the video bitrate when bandwidth decreases.

spiderfaced:
> > Do you know how the built in and advertised Adaptive bitrate works in vanilla SRT? Is it at all comparable to the RTMP dynamic bitrate in performance or does it need to be working closely with the encoder?
> 
> Got nothing to do with dynamic bitrate. It's ARQ, automated repeat request, meaning an auto resend of lost packets. It's faster than rtmp which is tcp because it uses udp but improved with some ack exchanges. Dynamic bitrate decreases the video bitrate when bandwidth decreases.

Gotcha on resending packets, but I am very interested in getting the bitrate lowered without having to resend packets. I use a high SRT latency as it is (located in Cambodia for IRL streaming and 4g is iffy).

Perhaps I will just testing SRT without forcing my mind that it requires dynamic bitrate or HEVC level low bitrate qualities.

DmitriyFrogo:
Last two months, I have issues with dynamic bitrate - white screen for a long time, and only sound is heard
Without enabled dynamic bitrate, my stream just drops
If possible, can anyone give me latest master build with this patch?
I want to test these changes

pkviet:
I'll rebase it today. @DmitriyFrogo

DmitriyFrogo:
@pkviet
I also interested in testing https://github.com/obsproject/obs-studio/pull/9807
If you can give me one build with both changes, I can test both

pkviet:
> @pkviet
> I also interested in testing https://github.com/obsproject/obs-studio/pull/9807
> If you can give me one build with both changes, I can test both

MS might have solved the bug. We're waiting for confirmation but it's on their side. Not ours. So #9807 is not necessary. It solves another minor bug but not the cracks.

DmitriyFrogo:
> > @pkviet
> > I also interested in testing #9807
> > If you can give me one build with both changes, I can test both
> 
> MS might have solved the bug. We're waiting for confirmation but it's on their side. Not ours. So #9807 is not necessary. It solves another minor bug but not the cracks.

Ah, ok

sunpenghao:
> MS might have solved the bug.

By this do you mean the audio cracks? This have been bothering me for quite some time and I can still (pretty consistently) reproduce the issue on 30.0.2 after ~10h of recording. If MS has done something I'll try updating my OS instead.

pkviet:
Closing this since it's now quite old.

PR #6185, opened 2022-03-22
CI: Add Steam build uploader

derrod:
### Description

Adds a GitHub actions workflow as well as necessary files for Steam build to be uploaded directly from this repo.

This is version 1 of this workflow, which does not yet account for M1/Apple Silicon builds. Once we get closer to starting the beta phase of the next release and have decided on a naming scheme for builds, whether to use universal builds for Steam, etc. the action will be updated to a "v2".

### Motivation and Context

OBS is now on Steam, we don't want to have to deal with manually uploading builds. As originally proposed in https://github.com/obsproject/rfcs/pull/30 this adds the necessary automation via GitHub actions.

Thanks to [OpenTTD](https://github.com/OpenTTD/OpenTTD) whose workflow has served an inspiration for this version.

### How Has This Been Tested?

A number of runs of various iterations on my own repo: https://github.com/derrod/obs-studio/actions/workflows/steam.yml 

A run with the version in this PR at the time of submission is here: https://github.com/derrod/obs-studio/actions/runs/2021121357

I'm sure some people have ideas to optimise it further as I'm not exactly experienced with writing github actions.

### Types of changes

### Checklist:

### DIFF
--- a/.github/workflows/steam.yml
+++ b/.github/workflows/steam.yml
@@ -1,0 +1,215 @@
+name: Steam Upload
+
+on:
+  release:
+    types:
+    - published
+  workflow_dispatch:
+    inputs:
+      tag:
+        description: 'Tag to fetch and upload (nightly if none)'
+        required: false
+      win_url_override:
+        description: 'Windows build to use (.zip only)'
+        required: false
+      mac_url_override:
+        description: 'Mac build to use (.dmg only)'
+        required: false
+  schedule:
+  - cron: 0 0 * * *
+
+env:
+  WORKFLOW_ID: 583765
+  GIT_NIGHTLY_BRANCH: master
+  STEAM_NIGHTLY_BRANCH: nightly
+  STEAM_STABLE_BRANCH: staging
+  STEAM_BETA_BRANCH: beta_staging
+
+jobs:
+  upload:
+    name: Steam upload
+    runs-on: ubuntu-20.04
+
+    steps:
+    - name: Checkout
+      uses: actions/checkout@v3
+      with:
+        path: source
+
+    - name: Setup 7-Zip (PPA)
+      # The 7-Zip version available in the default ubuntu repos (p7zip) is wildly out-of-date and does not properly support DMG files.
+      run: |
+        sudo add-apt-repository -y ppa:spvkgn/sevenzip
+        sudo apt update -y
+        sudo apt install -y 7-zip
+
+    - name: Get build information
+      id: build-info
+      run: |
+        EVENT='${{ github.event_name }}'
+        if [[ ${EVENT} == 'release' || ( ${EVENT} == 'workflow_dispatch' && -n '${{ github.event.inputs.tag }}') ]]; then
+            if [[ ${EVENT} == "release" ]]; then
+                DESC='${{ github.event.release.tag_name }}'
+                if [[ '${{ github.event.release.prerelease }}' == 'true' ]]; then
+                  BRANCH='${{ env.STEAM_BETA_BRANCH }}'
+                else
+                  BRANCH='${{ env.STEAM_STABLE_BRANCH }}'
+                fi
+                ASSETS_URL='${{ github.event.release.assets_url }}'
+            else
+                RELEASE="$(curl -s '${{ github.api_url }}/repos/obsproject/obs-studio/releases/tags/${{ github.event.inputs.tag }}')"
+                
+                DESC="$(jq -r '.tag_name' <<< ${RELEASE})"
+                if [[ "$(jq -r '.prerelease' <<< ${RELEASE})" == 'true' ]]; then
+                  BRANCH='${{ env.STEAM_BETA_BRANCH }}'
+                else
+                  BRANCH='${{ env.STEAM_STABLE_BRANCH }}'
+                fi
+                ASSETS_URL="$(jq -r '.assets_url' <<< ${RELEASE})"
+            fi
+
+            ASSETS="$(curl -s "${ASSETS_URL}")"
+            WIN_ASSET_URL="$(jq -r '.[] | select(.name|test(".*x64.zip")) .browser_download_url' <<< ${ASSETS})"
+            MAC_ASSET_URL="$(jq -r '.[] | select(.name|test(".*.dmg")) .browser_download_url' <<< ${ASSETS})"
+            TYPE='release'
+        else
+            BRANCH='${{ env.STEAM_NIGHTLY_BRANCH }}'
+            BUILDS="$(curl -s '${{ github.api_url }}/repos/obsproject/obs-studio/actions/workflows/${{ env.WORKFLOW_ID }}/runs?per_page=1&event=push&status=success&branch=${{ env.GIT_NIGHTLY_BRANCH }}')"
+            ARTIFACTS_URL="$(jq -r '.workflow_runs[].artifacts_url' <<< ${BUILDS})"
+            DESC="g$(jq -r '.workflow_runs[].head_sha' <<< "${BUILDS}" | cut -c1-9)"
+
+            ARTIFACTS="$(curl -s ${ARTIFACTS_URL})"
+            WIN_ASSET_URL="$(jq -r '.artifacts[] | select(.name|test(".*win-x64.*")) .archive_download_url' <<< ${ARTIFACTS})"
+            MAC_ASSET_URL="$(jq -r '.artifacts[] | select(.name|test(".*macos-x86_64.*")) .archive_download_url' <<< ${ARTIFACTS})"
+            TYPE='nightly'
+        fi
+
+        # Apply overrides from workflow_dispatch
+        if [[ ${EVENT} == 'workflow_dispatch' ]]; then
+            if [[ -n '${{ github.event.inputs.win_url_override }}' ]]; then
+                WIN_ASSET_URL='${{ github.event.inputs.win_url_override }}'
+            fi
+            
+            if [[ -n '${{ github.event.inputs.mac_url_override }}' ]]; then
+                MAC_ASSET_URL='${{ github.event.inputs.mac_url_override }}'
+            fi
+        fi
+
+        if [[ -z ${WIN_ASSET_URL} || -z ${MAC_ASSET_URL} ]]; then
+            echo "Missing at least one asset URL!"
+            exit 1
+        fi
+
+        # set env variables for subsequent steps
+        echo "::set-output name=type::${TYPE}"
+        echo "::set-output name=branch::${BRANCH}"
+        echo "::set-output name=desc::${DESC}"
+        echo "::set-output name=win_url::${WIN_ASSET_URL}"
+        echo "::set-output name=mac_intel_url::${MAC_ASSET_URL}"
+
+    - name: Restore build cache
+      id: cache
+      uses: actions/cache@v3
+      with:
+        path: ${{ github.workspace }}/steam/build
+        key: ${{ steps.build-info.outputs.branch }}-${{ steps.build-info.outputs.desc }}
+        # Using "restore-keys" will restore the most recent cache for the branch, even if the exact cache doesn't exist.
+        # This doesn't set cache-hit to true so it won't skip the upload for nightlies.
+        restore-keys: ${{ steps.build-info.outputs.branch }}
+
+    - name: Determine if Steam upload should run
+      # If the nightly build has already been uploaded and thus a cache exists skip this and the following steps.
+      # Steam does not prevent us from uploading duplicate builds so this would just pollute the dashboard.
+      # This is a bit of a hack and can fail to work if our cache has been evicted or we somehow have no commits for 7 days,
+      # but it's better than nothing!
+      id: should-run
+      run: |
+        if [[ '${{ steps.build-info.outputs.type }}' == 'release' || '${{ steps.cache.outputs.cache-hit }}' != 'true' ]]; then
+            echo "::set-output name=result::true"
+        else
+            echo "::set-output name=result::false"
+        fi
+
+    - name: Download and prepare builds
+      if: steps.should-run.outputs.result == 'true'
+      run: |
+        echo "::group::Download Windows build"
+        if [[ '${{ steps.build-info.outputs.win_url }}' == *'api.github.com'* ]]; then
+            curl -L -H 'Authorization: Bearer ${{ secrets.GITHUB_TOKEN }}' '${{ steps.build-info.outputs.win_url }}' -o windows.zip
+        else
+            curl -L '${{ steps.build-info.outputs.win_url }}' -o windows.zip
+        fi
+        echo "::endgroup::"
+      
+        echo "::group::Download Mac build"
+        if [[ '${{ steps.build-info.outputs.mac_intel_url }}' == *'api.github.com'* ]]; then
+            curl -L -H 'Authorization: Bearer ${{ secrets.GITHUB_TOKEN }}' '${{ steps.build-info.outputs.mac_intel_url }}' -o mac_x86.dmg.zip
+        else
+            curl -L '${{ steps.build-info.outputs.mac_intel_url }}' -o mac_x86.dmg
+        fi
+        echo "::endgroup::"
+
+        mkdir -p steam && cd steam
+  
+        echo "::group::Extract and prepare Win64"
+        mkdir steam-windows
+        (
+            cd steam-windows
+            unzip ../../windows.zip
+            # CI builds can be double-zipped
+            if compgen -G "*.zip" > /dev/null; then
+                unzip *.zip
+                rm *.zip
+            fi
+            # copy install scripts and create sentinel file
+            cp -r ../../source/CI/steam/scripts scripts
+            touch disable_updater
+        )
+        echo "::endgroup::"
+
+        echo "::group::Extract macOS (x86)"
+        mkdir steam-macos
+        # CI builds are zipped
+        if [[ -f ../mac_x86.dmg.zip ]]; then
+            unzip ../mac_x86.dmg.zip
+            # 7-Zip will have an exit code of 2 due to the "unsafe" 'Applications' symlink.
+            # GitHub treats this as a failure so ignore non-zero exit codes here.
+            7zz x *.dmg -otmp || true
+        else
+            7zz x ../mac_x86.dmg -otmp || true
+        fi
+          
+        mv tmp/*/OBS.app steam-macos
+        echo "::endgroup::"
+
+    - name: Setup steamcmd
+      if: steps.should-run.outputs.result == 'true'
+      uses: CyberAndrii/setup-steamcmd@e19cd1516315ce46dbfffa47193f92fe42d1419e
+
+    - name: Generate Steam auth code
+      if: steps.should-run.outputs.result == 'true'
+      id: steam-totp
+      uses: CyberAndrii/steam-totp@0fc9e59dc5bbf4368d23d5a33956f104248da31a
+      with:
+        shared_secret: ${{ secrets.STEAM_SHARED_SECRET }}
+
+    - name: Upload to Steam
+      if: steps.should-run.outputs.result == 'true'
+      run: |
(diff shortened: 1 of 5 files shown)


.github/workflows/steam.yml
```diff
@@ -0,0 +1,214 @@
...
+      uses: actions/checkout@v2
+      with:
+        path: source
+
+    - name: Setup 7zip
+      run: |
+        mkdir 7z
+        (
+            cd 7z
+            curl -s https://www.7-zip.org/a/7z2107-linux-x64.tar.xz | tar -xJ
```
PatTheMav, line 44: Any reason why you don't use `apt` to install 7z? Also might be worthwhile checking for a pre-existing `7z` on the runner to skip this step.

RytoEX: 7z doesn't seem to be listed in the ubuntu-20.04 image.

https://github.com/actions/virtual-environments/blob/main/images/linux/Ubuntu2004-Readme.md

derrod: p7zip is unmaintained and based on an ancient 7z version, so it doesn't support many newer formats like HFS+

derrod: We could use this PPA instead: https://launchpad.net/~spvkgn/+archive/ubuntu/sevenzip
It has a recent enough version that would work.

derrod:
Updated to get a more recent 7zip from a Ubuntu PPA and replaced all the ugly temporary json files with using variables that are piped into `jq`.

.github/workflows/steam.yml
```diff
@@ -0,0 +1,212 @@
...
+      # The 7zip version available in the default ubuntu repos (p7zip) is wildly out-of-date and does not properly support DMG files
+      run: |
+        sudo add-apt-repository -y ppa:spvkgn/sevenzip
+        sudo apt update -y
+        sudo apt install -y 7-zip
+
+    - name: Get build information
+      run: |
+        EVENT=${{ github.event_name }}
+        if [[ $EVENT == "release" || ( $EVENT == "workflow_dispatch" && ! -z "${{ github.event.inputs.tag }}") ]]; then
```
PatTheMav, line 49: I always advise to wrap variable names in curly brackets (`${EVENT}`) just as a good practise.

Not quoting the lparam is fine in bash and zsh when using `[[ ]]`. Also if you don't need substitution, I'd use single quotes here.

.github/workflows/steam.yml
```diff
@@ -0,0 +1,212 @@
...
+            if [[ $EVENT == "release" ]]; then
+                DESC=${{ github.event.release.tag_name }}
+                if [ "${{ github.event.release.prerelease }}" = "true" ]; then
```
PatTheMav, line 52: Prefer `[[ ]]` on Bash, also use `==` to make it fail if used with `test` by accident.

.github/workflows/steam.yml
```diff
@@ -0,0 +1,212 @@
...
+
+            ARTIFACTS=$(curl -s $ARTIFACTS_URL)
+            WIN_ASSET_URL=$(jq -r '.artifacts[] | select(.name|test(".*win64")) .archive_download_url' <<< "$ARTIFACTS")
+            MAC_ASSET_URL=$(jq -r '.artifacts[] | select(.name|test(".*macos-x86_64")) .archive_download_url' <<< "$ARTIFACTS")
+            TYPE="nightly"
+        fi
+        
+        # Apply overrides from workflow_dispatch
+        if [[ $EVENT == "workflow_dispatch" ]]; then
+            if [[ ! -z "${{ github.event.inputs.win_url_override }}" ]]; then
```
PatTheMav, line 88: I'm not sure whether `-n` might suffice here - testing that only returns true if the string is set _and_ is not zero-length.

.github/workflows/steam.yml
```diff
@@ -0,0 +1,212 @@
...
+        restore-keys: ${{ env.BRANCH }}
+
+    - name: Determine if Steam upload should run
+      # If the nightly build has already been uploaded and thus a cache exists skip this and the following steps.
+      # Steam does not prevent us from uploading duplicate builds so this would just pollute the dashboard.
+      # This is a bit of a hack and can fail to work if our cache has been evicted or we somehow have no commits for 7 days,
+      # but it's better than nothing!
+      id: should-run
+      run: |
+        SHOULD_RUN=$(if [[ "${{ env.UPLOAD_TYPE }}" == "release" || "${{ steps.cache.outputs.cache-hit }}" != "true" ]]; then echo "true"; else echo "false"; fi)
```
PatTheMav, line 126: You can `echo` the `set-output` part directly within the branch, no need to store it in an intermediary variable.

.github/workflows/steam.yml
```diff
@@ -0,0 +1,212 @@
...
+        echo "::set-output name=result::$SHOULD_RUN"
+    
+    - name: Download and prepare builds
+      if: steps.should-run.outputs.result == 'true'
+      run: |
+        echo "::group:: Download Windows build"
+        if [[ "${{ env.WIN_ARTIFACT_URL }}" == *"api.github.com"* ]]; then
+            wget --header="Authorization: Bearer ${{ secrets.GITHUB_TOKEN }}" "${{ env.WIN_ARTIFACT_URL }}" -O windows.zip
```
PatTheMav, line 134: We use `curl` above, would prefer to stick with one tool.

.github/workflows/steam.yml
```diff
@@ -0,0 +1,212 @@
...
+        else
+            wget "${{ env.MAC_x86_ARTIFACT_URL }}" -O mac_x86.dmg
+        fi
+        echo "::endgroup::"
+
+        mkdir -p steam && cd steam
+  
+        echo "::group::Extract and prepare Win64"
+        mkdir steam-windows
+        (
```
PatTheMav, line 152: Any particular reason why you use a subshell here? If you want to return to the original directory you came from, I had to kick myself to use `pushd` and `popd` for that purpose.

Otherwise I'd just use a subshell if I want to change environment variables for just a few commands.

.github/workflows/steam.yml
```diff
@@ -0,0 +1,212 @@
...
+        )
+        echo "::endgroup::"
+
+        echo "::group::Extract macOS (x86)"
+        mkdir steam-macos
+        # CI builds are zipped
+        if [[ -f ../mac_x86.dmg.zip ]]; then
+            unzip ../mac_x86.dmg.zip
+            # ignore exit code from 7zip
+            7zz x *.dmg -otmp || true
```
PatTheMav, line 172: That is indeed the correct way to "capture" non-0 return codes and avoid a script failure. (Either the or capturing it in a variable for later handling).

PatTheMav:
Startled the witches of nitpick forest, but looks fine overall.

Do we have to use batch scripts (i.e. prescribed by Steam)? I'd prefer Powershell, but with Windows preventing user-initiated Powershell scripts by default I wouldn't know if it works out of the box.

.github/workflows/steam.yml
derrod: This was mostly taken from OpenTTD's steam deploy script, using the subshell with indentation just seemed like a "pretty" way of separating those things and not having to worry about where you are in the tree for subsequent commands.

derrod:
The batch script were also simply adapted from the existing virtualcam install script. Since powershell is blocked by default that probably wouldn't work, and iirc I tried running a powershell script in an earlier iteration but that didn't work.

PatTheMav:
> The batch script were simply adapted from the existing virtualcam install script. Since powershell is blocked by default that probably wouldn't work, and iirc I tried running a powershell script in an earlier iteration but that didn't work for some reason (maybe steam does a bit more than just telling the OS to run a file and didn't recognise `.ps1`?).

Yeah I'm worried that Windows will just block execution and batch scripts are just the way to go.

CI/steam/scripts/install.bat
```diff
@@ -0,0 +1,85 @@
...
+
+goto checkAdmin
+
+
+:checkAdmin
+	net session >nul 2>&1
+	if %errorLevel% == 0 (
+		echo.
+	) else (
+		echo Administrative rights are required, please re-run this script as Administrator.
```
RytoEX, line 12: "Administrative rights are required. Please re-run this script as Administrator."

CI/steam/scripts/install.bat
```diff
@@ -0,0 +1,85 @@
...
+	reg add "HKLM\SOFTWARE\OBS Studio" /f /t REG_SZ /d %1 /reg:64
+
+:setupProgramData
+	:: Required for UWP applications
+	mkdir "%PROGRAMDATA%\obs-studio-hook"
+	icacls "%PROGRAMDATA%\obs-studio-hook" /grant "ALL APPLICATION PACKAGES":(OI)(CI)(GR,GE)
+
+:checkDLL
+	echo Checking for 32-bit Virtual Cam registration...
+	reg query "HKLM\SOFTWARE\Classes\CLSID\{A3FCE0F5-3493-419F-958A-ABA1250EC20B}" >nul 2>&1 /reg:32
```
RytoEX, line 27: Do we want to hard-code this in this file, or use a .bat.in file and grab this GUID from CI before producing the Steam build (which is what we do for other builds, I think)?

CI/steam/scripts/uninstall.bat
```diff
@@ -0,0 +1,33 @@
+@echo off
+@cd /d "%~dp0"
+goto checkAdmin
+
+:checkAdmin
+	net session >nul 2>&1
+	if %errorLevel% == 0 (
+		echo.
+	) else (
+		echo Administrative rights are required, please re-run this script as Administrator.
```
RytoEX, line 10: "Administrative rights are required. Please re-run this script as Administrator."

.github/workflows/steam.yml
```diff
@@ -0,0 +1,215 @@
...
```
RytoEX, line 175: Is 7zz some CLI executable?

.github/workflows/steam.yml
```diff
@@ -0,0 +1,215 @@
...
+  STEAM_BETA_BRANCH: beta_staging
+
+jobs:
+  upload:
+    name: Steam upload
+    runs-on: ubuntu-20.04
+
+    steps:
+    - name: Checkout
+      uses: actions/checkout@v2
```
RytoEX, line 35: Might as well update the first-party actions to their latest versions unless there's a reason we need the older versions.

.github/workflows/steam.yml
```diff
@@ -0,0 +1,215 @@
...
+      with:
+        path: source
+
+    - name: Setup 7-zip (PPA)
+      # The 7zip version available in the default ubuntu repos (p7zip) is wildly out-of-date and does not properly support DMG files
```
RytoEX, line 40: 7zip -> 7-Zip

End with a period.

.github/workflows/steam.yml
```diff
@@ -0,0 +1,215 @@
...
+
+        echo "::group::Extract macOS (x86)"
+        mkdir steam-macos
+        # CI builds are zipped
+        if [[ -f ../mac_x86.dmg.zip ]]; then
+            unzip ../mac_x86.dmg.zip
+            # ignore exit code from 7zip
+            7zz x *.dmg -otmp || true
+        else
+            # 7zip will have an exit code of 2 due to the 'Applications' symlink
```
RytoEX, line 177: 7zip -> 7-Zip

End with a period.

RytoEX (changes requested):
Did a quick review.

.github/workflows/steam.yml
derrod: Yes, don't ask me why they went with that.

CI/steam/scripts/install.bat
derrod: Could be done, though I don't see us changing it nor anybody else needing this specific script to have their own.

CI/steam/scripts/install.bat
derrod: This is taken straight from https://github.com/obsproject/obs-studio/blob/master/plugins/win-dshow/virtualcam-install.bat.in#L10 so I'd suggest updating it there first.

CI/steam/scripts/install.bat
RytoEX: It would just be one less place to have to consider updating it should we change it, or if we want to register the Steam build and the normal build with different GUIDs to avoid clobbering each other or having one load the other (though that may also require us to differentiate the camera names).

As mentioned, that's how we currently handle this:
* https://github.com/obsproject/obs-studio/blob/ccd1ef2bd0abb7c9d86d25871dfd8744fdd450cb/plugins/win-dshow/virtualcam-install.bat.in
* https://github.com/obsproject/obs-studio/blob/ccd1ef2bd0abb7c9d86d25871dfd8744fdd450cb/plugins/win-dshow/virtualcam-uninstall.bat.in

I'm fine with this as-is for now, but something to consider.

.github/workflows/steam.yml
```diff
@@ -0,0 +1,214 @@
...
+        description: 'Tag to fetch and upload (nightly if none)'
+        required: false
+      win_url_override:
+        description: 'Windows build to use (.zip only)'
+        required: false
+      mac_url_override:
+        description: 'Mac build to use (.dmg only)'
+        required: false
+  schedule:
+  - cron: 0 0 * * *
```
RytoEX, line 19: Do we want this to run every 24 hours or every 12 hours? The latter can be achieved with this:
```
0 */12 * * *
```

I don't have a strong opinion here.

RytoEX (approved):
Seems fine. I left a comment on the frequency of the cron timing. Approved otherwise.

.github/workflows/steam.yml
derrod: I went with 24 hours so it's an actualy "night"-ly :P

PatTheMav:
Where does Steam install the app bundle to on macOS? Did anyone test this already?

derrod:
From memory it should be in `~/Library/Application Support/Steam/steamapps/OBS Studio`

PatTheMav:
> From memory it should be in `~/Library/Application Support/Steam/steamapps/OBS Studio`

Ugly, but should cause no issues.

derrod:
> > From memory it should be in `~/Library/Application Support/Steam/steamapps/OBS Studio`
> 
> Ugly, but should cause no issues.

Well it did cause the issue of browser source not loading, which was fixed in https://github.com/obsproject/obs-studio/commit/269d48f6816d77045724b050b9a6e890a301376c :P

Fun fact: Epic puts stuff into `/Users/Shared/Epic Games/<stuff>`.

RytoEX:
@PatTheMav If this is still good to go, I'll land it.

.github/workflows/steam.yml
```diff
@@ -0,0 +1,214 @@
...
+        sudo apt update -y
+        sudo apt install -y 7-zip
+
+    - name: Get build information
+      run: |
+        EVENT='${{ github.event_name }}'
+        if [[ ${EVENT} == 'release' || ( ${EVENT} == 'workflow_dispatch' && -n '${{ github.event.inputs.tag }}') ]]; then
+            if [[ ${EVENT} == "release" ]]; then
+                DESC='${{ github.event.release.tag_name }}'
+                if [ '${{ github.event.release.prerelease }}' == 'true' ]; then
```
PatTheMav, line 52: Use bash' `[[ ]]` here, because `==` shouldn't work with POSIX `test`.

.github/workflows/steam.yml
```diff
@@ -0,0 +1,214 @@
...
+                  BRANCH='${{ env.STEAM_BETA_BRANCH }}'
+                else
+                  BRANCH='${{ env.STEAM_STABLE_BRANCH }}'
+                fi
+                ASSETS_URL='${{ github.event.release.assets_url }}'
+            else
+                RELEASE=$(curl -s '${{ github.api_url }}/repos/obsproject/obs-studio/releases/tags/${{ github.event.inputs.tag }}')
```
PatTheMav, line 59: In bash it's advised to wrap the result of `$()` in double quotes to avoid tokenisation.

.github/workflows/steam.yml
```diff
@@ -0,0 +1,214 @@
...
+                
+                DESC=$(jq -r '.tag_name' <<< "$RELEASE")
```
PatTheMav, line 61: For consistency all variable accesses should use the `${}` form.

.github/workflows/steam.yml
```diff
@@ -0,0 +1,214 @@
...
+                  BRANCH='${{ env.STEAM_BETA_BRANCH }}'
+                else
+                  BRANCH='${{ env.STEAM_STABLE_BRANCH }}'
+                fi
+                ASSETS_URL=$(jq -r '.assets_url' <<< "$RELEASE")
+            fi
+
+            ASSETS=$(curl -s "$ASSETS_URL")
+            WIN_ASSET_URL=$(jq -r '.[] | select(.name|test(".*x64.zip")) .browser_download_url' <<< "$ASSETS")
+            MAC_ASSET_URL=$(jq -r '.[] | select(.name|test(".*.dmg")) .browser_download_url' <<< "$ASSETS")
```
PatTheMav, line 72: I'm fine with this as-is, but one should be able to do that with a single invocation of `jq`, see this example from `obs-plugintemplate`:

```
read -r deps_version deps_hash <<< \
  "$(jq -r --arg key "obs-deps" --arg target "${target}" \
    '.dependencies[$key][$target] | {version, hash} | join(" ")' \
    ${buildspec_file})"
```

You can have `jq` select the desired fields and output/join them so that `read` can extract them into separate variables.

.github/workflows/steam.yml
```diff
@@ -0,0 +1,214 @@
...
+            fi
+        fi
+
+        if [[ -z $WIN_ASSET_URL || -z $MAC_ASSET_URL ]]; then
+            echo "Missing at least one asset URL!"
+            exit 1
+        fi
+
+        # set env variables for subsequent steps
+        echo "UPLOAD_TYPE=${TYPE}" >> $GITHUB_ENV
```
PatTheMav, line 103: This is also fine as-is, I just started to prefer setting using step output variables instead of polluting the environment (especially as step outputs can also be re-used as job outputs and then also used in later jobs).

.github/workflows/steam.yml
```diff
@@ -0,0 +1,214 @@
...
+        echo "::endgroup::"
+
+    - name: Setup steamcmd
+      if: steps.should-run.outputs.result == 'true'
+      uses: CyberAndrii/setup-steamcmd@v1
+
+    - name: Generate Steam auth code
+      if: steps.should-run.outputs.result == 'true'
+      id: steam-totp
+      uses: CyberAndrii/steam-totp@v1
```
PatTheMav, line 191: This is just my personal choice, but when it comes to Marketplace actions that handle secret tokens, I rather specify the commit hash that I reviewed as tags can be changed after the fact and can point to unreviewed commits without anyone noticing.

.github/workflows/steam.yml
derrod: I believe that was the only instance that I missed in the first round, i changed it now.

PatTheMav:
In the best tradition of our joint indulgence of nitpicking, I added a few picked nits of my own.

.github/workflows/steam.yml
derrod: Good point, I'll use the hashes for the steamcmd/totp action.

.github/workflows/steam.yml
derrod: In this case I would disagree on the basis that the readability suffers once you get too crazy with `jq`, and given that we're talking milliseconds either way here I'm not sure it's worth changing.

.github/workflows/steam.yml
derrod: I converted this to use outputs instead (although there's only one job, so reusability is kind of a moot point).

RytoEX:
@PatTheMav Re-requesting review until we get to an approval.

PatTheMav (approved):

PR #6194, opened 2022-03-22
cmake: Add platform-specific include paths back to FindLibAJANTV2

paulh-aja:
### Description

This change adds the platform-specific include paths for the "ntv2driverinterface" and "ntv2publicinterface" sources and headers back to the `FindAJANTV2.cmake` module script. These include paths were removed from the find script as the paths were thought not to exist. This is due to the NTV2 16.1 CMake Install directives accidentally copying those files into the base `ajantv2/include` and `ajantv2/src` directories. The current obs-deps build uses the NTV2 16.1 release with that bug.

These files normally live in and should be copied by CMake Install to `ajantv2/src/win`, `ajantv2/src/mac` and `ajantv2/src/lin` paths. The NTV2 16.2 release corrects this in its CMake scripts but causes the AJA OBS plugin to throw a build error if attempting to build the plugin with 16.2 build artifacts.

We are discussing internally on the AJA NTV2 the feasibility of simply moving those sources and headers to live directly under the base `ajantv2/includes` and `ajantv2/src` directories but for the time being they must remain in their platform-specific paths.

*IMPORTANT: obs-deps PR*
This pull request coincides with and is dependent on another pull request to the "obs-deps" project, which updates the Windows and macOS OBS deps builds to use NTV2 16.2 code: https://github.com/obsproject/obs-deps/pull/100. Once the obs-deps have been updated to use NTV2 16.2 code, the AJA OBS plugin build will break. This PR should be applied at the same time that the obs-deps PR is applied.

### Motivation and Context

The NTV2 16.1 CMake Install directives placed the platform-specific ntv2driverinterface and ntv2publicinterface .cpp and .h files into the wrong sub-directory.

### How Has This Been Tested?

I've tested with the above-mentioned PR to the "obs-deps" repo on Windows, and then verified with this OBS PR that the AJA plugin will build with the NTV2 16.2 Windows build artifacts produced by the updated obs-deps build. I don't have a way to test on macOS at the moment though. I can get that set up again if requested but it will take a bit of time.

### Types of changes

### Checklist:


paulh-aja:
How does OBS handle PRs that depend upon obs-deps PRs?

paulh-aja:
The non-existent path that the macOS build is erroring about should be resolved by my PR to the obs-deps project, which updates the deps to use AJA NTV2 16.2. I just noticed that the same CMake error, complaining about the missing include path, does occur in the Windows CI build here but doesn't trigger a build error.

Fenrirthviti:
> How does OBS handle PRs that depend upon obs-deps PRs?

Usually, a PR that depends on another PR to be merged first will be marked as draft until the other one is merged, with a mention in the description of the PR that it depends on that needs to be merged first. You've already done the latter, so this can just be marked as draft which shouldn't affect any other review.

cmake/Modules/FindLibAJANTV2.cmake
```diff
@@ -106,6 +106,17 @@ if(LIBAJANTV2_FOUND)
...
 
+  if(WIN32)
+    set(AJA_LIBRARIES_INCLUDE_DIRS ${AJA_LIBRARIES_INCLUDE_DIRS}
+                                   ${AJA_LIBRARIES_INCLUDE_DIR}/ajantv2/src/win)
+  elseif(APPLE)
+    set(AJA_LIBRARIES_INCLUDE_DIRS ${AJA_LIBRARIES_INCLUDE_DIRS}
+                                   ${AJA_LIBRARIES_INCLUDE_DIR}/ajantv2/src/mac)
+  elseif(UNIX AND NOT APPLE)
+    set(AJA_LIBRARIES_INCLUDE_DIRS ${AJA_LIBRARIES_INCLUDE_DIRS}
+                                   ${AJA_LIBRARIES_INCLUDE_DIR}/ajantv2/src/lin)
```
tytan652, line 117: Please replace:
- `WIN32` by `OS_WINDOWS`
- `APPLE` by `OS_MACOS`
- `UNIX AND NOT APPLE` by `OS_LINUX` if it only targets Linux and not FreeBSD and OpenBSD

This is explained in OBS CMake files [here](https://github.com/obsproject/obs-studio/blob/master/plugins/CMakeLists.txt#L23)

PatTheMav (commented):
Apart from the change @tytan652 mentioned, I'm fine with this overall.

Given the amount of include directories necessary, I'm slightly gravitating towards explicitly using the `list` function here:

```
list(APPEND AJA_LIBRARIES_INCLUDE_DIRS <DIR>)
if(OS_WINDOWS)
  list(APPEND AJA_LIBRARIES_INCLUDE_DIRS <DIR>/ajantv2/src/win)
[...]
endif()
```

The list variable itself is quoted in line 153, so it should be passed as-is.

WizardCM (changes requested):
Commit prefix should be `cmake:`

obinnaokechukwu:
Drive by comment: Is it possible to make the aja dependencies optional, so all builds don't have to include it?

RytoEX:
> Drive by comment: Is it possible to make the aja dependency optional, so all builds don't have to include it?

Just disable the plugin when building: `-DENABLE_AJA=OFF`

obinnaokechukwu:
> Just disable the plugin when building: `-DENABLE_AJA=OFF`

Thanks. It may be worth adding this note to the error message so one knows what flag to turn off.

paulh-aja:
> Apart from the change @tytan652 mentioned, I'm fine with this overall.
> 
> Given the amount of include directories necessary, I'm slightly gravitating towards explicitly using the `list` function here:
> 
> ```
> list(APPEND AJA_LIBRARIES_INCLUDE_DIRS <DIR>)
> if(OS_WINDOWS)
>   list(APPEND AJA_LIBRARIES_INCLUDE_DIRS <DIR>/ajantv2/src/win)
> [...]
> endif()
> ```
> 
> The list variable itself is quoted in line 153, so it should be passed as-is.

I've force pushed a change which uses the `list(APPEND ...)` rather than `set`.

RytoEX:
@paulh-aja This should be unblocked now that #6576 has been merged.

paulh-aja:
It looks to me that this PR is no longer required and can be closed.

OBS deps were updated to 2022-05-23 obs-deps as of https://github.com/obsproject/obs-studio/pull/6576.
That PR incorporated an update of the AJA NTV2 SDK to the 16.2 bugfix 5 release (https://github.com/obsproject/obs-deps/pull/100)

I've verified that an obs-studio CI build after that deps change was made is still building the AJA plugin properly, and that the plugin is still operating.

PR #6196, opened 2022-03-23
CI: Move obspython.py to Resources

derrod:
### Description

Moves `obspython.py` to `Resources` on macOS.

### Motivation and Context

Having `obspython.py` in the code directories requires it to be signed, since it's a text file this is done via extended attributes. This breaks on Steam or when using another packaging method that does not retain those.

Additionally per Apple themselves it is not recommended to store scripts there, and they should live in resources instead: https://developer.apple.com/library/archive/technotes/tn2206/_index.html#//apple_ref/doc/uid/DTS40007919-CH1-TNTAG201

### How Has This Been Tested?

- Made sure python scripting still works

### Types of changes

### Checklist:


deps/obs-scripting/obs-scripting-python.c
```diff
@@ -1672,17 +1672,25 @@ bool obs_scripting_load_python(const char *python_path)
...
+	dstr_init_move_array(&bundle_path, os_get_executable_path_ptr(""));
+	dstr_cat(&bundle_path, "../PlugIns");
+	char *absolute_plugin_path = os_get_abs_path_ptr(bundle_path.array);
+
+	if(absolute_plugin_path != NULL) {
+		add_to_python_path(absolute_plugin_path);
+		bfree(absolute_plugin_path);
+	}
+
+	dstr_cat(&bundle_path, "/../Resources");
```
PatTheMav, line 1686: Just a minor thing - either we create both paths with a leading slash, or without.

derrod: Without wouldn't work, because it'd concatenate it wrongly the second time, but I can just add the slash to the first one as well.

PatTheMav: I'd make it explicit:

```
    struct dstr plugin_path;
    struct dstr resource_path;
    
    dstr_init_move_array(&plugin_path, os_get_executable_path_ptr(""));
    dstr_init_copy(&resource_path, plugin_path.array);
    dstr_cat(&plugin_path, "../PlugIns");
    dstr_cat(&resource_path, "../Resources");

    char *absolute_plugin_path = os_get_abs_path_ptr(plugin_path.array);
    char *absolute_resource_path = os_get_abs_path_ptr(resource_path.array);
    
    if(absolute_plugin_path != NULL) {
        add_to_python_path(absolute_plugin_path);
        bfree(absolute_plugin_path);
    }
    dstr_free(&plugin_path);

    if(absolute_resource_path != NULL) {
        add_to_python_path(absolute_resource_path);
        bfree(absolute_resource_path);
    }
    dstr_free(&resource_path);
```

PatTheMav (approved):

PR #6197, opened 2022-03-23
CI/deps: Move obspython.py to Resources (27.2 branch)

derrod:
### Description

Moves `obspython.py` to `Resources` on macOS.

This is a variation of #6196 but for the legacy build system in case we want to create another 27.2.x bugfix release.

### Motivation and Context

Having `obspython.py` in the code directories requires it to be signed, since it's a text file this is done via extended attributes. This breaks on Steam or when using another packaging method that does not retain those.

Additionally per Apple themselves it is not recommended to store scripts there, and they should live in resources instead: https://developer.apple.com/library/archive/technotes/tn2206/_index.html#//apple_ref/doc/uid/DTS40007919-CH1-TNTAG201

### How Has This Been Tested?

- Checked that files are in the correct folders, but was unable to test scripts on my Mac since only arm64 python was installed

### Types of changes

### Checklist:

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -211,7 +211,7 @@ jobs:
           if [ -d ./OBS.app/Contents/Resources/data/obs-scripting ]; then
             mv ./OBS.app/Contents/Resources/data/obs-scripting/obslua.so ./OBS.app/Contents/MacOS/
             mv ./OBS.app/Contents/Resources/data/obs-scripting/_obspython.so ./OBS.app/Contents/MacOS/
-            mv ./OBS.app/Contents/Resources/data/obs-scripting/obspython.py ./OBS.app/Contents/MacOS/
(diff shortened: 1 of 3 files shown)


derrod:
Note: I have no idea why the Windows build fails on CEF, I haven't touched it (obviously), I guess the CEF archive may have been updated for the cmake rewrite and broken the old workflow in the process?

RytoEX:
> Note: I have no idea why the Windows build fails on CEF, I haven't touched it (obviously), I guess the CEF archive may have been updated for the cmake rewrite and broken the old workflow in the process?

I didn't think we changed the CEF archive itself in any meaningful way, unless @PatTheMav repackaged it differently. He can probably answer that question.

WizardCM:
Yes it was repackaged differently and is now incompatible with the 27.2 branch.

derrod:
I guess that's fine then. I'll remove the last commit once the builds have finished and I have an artifact for someone to at least verify that this doesn't break python on Intel Macs.

PatTheMav (approved):

PR #6205, opened 2022-03-25
cmake: Fix compiler flags from FindWayland and FindPipeWire

Arthus:
${WAYLAND_DEFINITIONS} is set to  "-I/usr/include/wayland" on my system

Setting it as INTERFACE_COMPILE_DEFINITIONS leads to CMake emitting "-D-I/usr/include/wayland" in the compiler flags
this breaks the compilation

Instead add WAYLAND_DEFINITIONS as compile option so it just gets appended without adding the superflours "-D"

### Motivation and Context
I was trying to compile OBS for the first time using openSUSE Tumbleweed. I was able to set the dependencies and cmake up, but during compilation GCC throws an error:

> \<command-line>: error: macro names must be identifiers

This is caused by cmake emitting
`cd /home/user/Programming/OBS/obs-studio/build/libobs && /usr/bin/cc -D-I/usr/include/wayland -DENABLE_DARRAY_TYPE_TEST [...]`
-D makes gcc wait for a macro, but instead the include flag follows immediately.

To fix it we just have to tell cmake to not interpret WAYLAND_DEFINITIONS as macro definition, but only append it to the compiler flags. (This is inline with what is done in [FindPipeWire.cmake.](https://github.com/obsproject/obs-studio/blob/ba68eda590b2621b13691a40bb9541ebb353257d/cmake/Modules/FindPipeWire.cmake#L89))

This issue was likely introduced by commit df94b02075636dc0ace7b4d1e9706abeb64dfbbd from PR #5155

P.S.: First day of working with OBS and first patch. I've tried following all contributor guidelines, but please tell me if I could've done better. :)

### How Has This Been Tested?

With this commit I am able to compile OBS using openSUSE Tumbleweed on a Thinkpad P14s.
Snapshot 20220322 with CMake 3.22.2
Snapshot 20220323 with CMake 3.22.3
I'm able to launch the compiled binary on X11 and Wayland using KDE 5.24.3, do some simple screen capture using XSHM and Pipewire, record my desktop and click through menus.

### Types of changes

### Checklist:

### DIFF
--- a/cmake/Modules/FindPipeWire.cmake
+++ b/cmake/Modules/FindPipeWire.cmake
@@ -10,8 +10,8 @@
 # be passed to target_link_libraries() instead of the ``PipeWire::PipeWire``
 # target ``PIPEWIRE_INCLUDE_DIRS`` This should be passed to
 # target_include_directories() if the target is not used for linking
-# ``PIPEWIRE_DEFINITIONS`` This should be passed to target_compile_options() if
-# the target is not used for linking
+# ``PIPEWIRE_COMPILE_FLAGS`` This should be passed to target_compile_options()
+# if the target is not used for linking
 #
 # If ``PIPEWIRE_FOUND`` is TRUE, it will also define the following imported
 # target:
@@ -56,7 +56,7 @@ find_package(PkgConfig QUIET)
 pkg_search_module(PKG_PIPEWIRE QUIET libpipewire-0.3)
 pkg_search_module(PKG_SPA QUIET libspa-0.2)
 
-set(PIPEWIRE_DEFINITIONS "${PKG_PIPEWIRE_CFLAGS}" "${PKG_SPA_CFLAGS}")
+set(PIPEWIRE_COMPILE_FLAGS "${PKG_PIPEWIRE_CFLAGS}" "${PKG_SPA_CFLAGS}")
 set(PIPEWIRE_VERSION "${PKG_PIPEWIRE_VERSION}")
 
 find_path(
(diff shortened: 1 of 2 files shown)


PatTheMav:
The fix is correct insofar as `WAYLAND_DEFINITIONS` contains the `CFLAGS` as reported by `pkg-config`.

I'd prefer if the variable is renamed as well, to avoid the confusion introduced by the fix ("definitions" being set as compiler options).

PatTheMav (changes requested):
Please also rename `WAYLAND_DEFINITIONS` to something like `WAYLAND_COMPILE_FLAGS`.

Arthus:
I've renamed WAYLAND_DEFINITIONS as requested and also added a commit to rename PIPEWIRE_DEFINITIONS as all your comments also apply to that one.

Question: 
Is it ok to have two commits for that, should I squash them or even open another PR?

PatTheMav (approved):

PatTheMav:
Multiple commits is fine, but I'd suggest splitting the changes between one commit for Wayland and another for PipeWire.

Arthus:
Hi, sorry, just force pushed the commits again after some fiddling with git.

It's just as you've proposed: One commit makes the changes for pipewire. The other one fixes the compilation with wayland and also renames the wayland variable. :)

PatTheMav:
Minor nit to pick: One commit explicitly mentions Wayland, the other doesn't. I'd suggest renaming the commit to explicitly call out that it changes something related to PipeWire.

Arthus:
Done :)

RytoEX (changes requested):
Minor nitpicks on the commit messages.

Use a capitalized word immediately after the commit message prefix. Punctuate paragraphs in the commit message body correctly.

The first commit doesn't seem to have anything to do directly with libobs, so mentioning it is confusing. The changes are just in our CMake finders, so I don't think it makes sense to mention libobs, unless you can elaborate?

RytoEX (approved):
I've adjusted the commit messages and the PR title. Merging shortly.

GeorgesStavracas (approved):

PR #6214, opened 2022-03-26
libobs: Fix stale active_copy_surfaces entries

jpark37:
### Description
OBS can crash when recording again after resetting video after a previous record.

### Motivation and Context
Crashes are bad.

### How Has This Been Tested?
Crash repro no longer crashes.

### Types of changes

### Checklist:


RytoEX:
Is this reproducible on current master, or only in your branch/PRs? What are the Steps to Reproduce?

jpark37:
I didn't check master, but it's likely the same. Record a video, switch profiles to reset video, then record a video. This member was added after the last release, so it shouldn't be out in the wild.

PR #6221, opened 2022-03-27
cmake: check empty OBS_MODULE_LIST and OBS_SCRIPTING_MODULE_LIST on macOS 

quink-black:
# Description

Check empty OBS_MODULE_LIST and OBS_SCRIPTING_MODULE_LIST before use them, since empty list leads to add_dependencies() failure.

# Motivation and Context

Don't fail when -DENABLE_SCRIPTING=OFF.

# How Has This Been Tested?

On MacOS, tested with -DENABLE_SCRIPTING=OFF and -DENABLE_SCRIPTING=ON.

# Types of changes

* Bug fix (non-breaking change which fixes an issue) 

# Checklist:
 
- [x] I have read the [contributing document](https://github.com/obsproject/obs-studio/blob/master/CONTRIBUTING.rst).

### DIFF
--- a/cmake/Modules/ObsHelpers_macOS.cmake
+++ b/cmake/Modules/ObsHelpers_macOS.cmake
@@ -281,25 +281,28 @@ function(setup_obs_modules target)
 
   get_property(OBS_SCRIPTING_MODULE_LIST GLOBAL
                PROPERTY OBS_SCRIPTING_MODULE_LIST)
-  add_dependencies(${target} ${OBS_SCRIPTING_MODULE_LIST})
+  if("${OBS_SCRIPTING_MODULE_LIST}")
+    add_dependencies(${target} ${OBS_SCRIPTING_MODULE_LIST})
 
-  install(
-    TARGETS ${OBS_SCRIPTING_MODULE_LIST}
-    LIBRARY DESTINATION "PlugIns"
-            COMPONENT obs_plugin_dev
-            EXCLUDE_FROM_ALL)
-
-  if(TARGET obspython)
     install(
-      FILES "$<TARGET_FILE_DIR:obspython>/obspython.py"
-      DESTINATION "Resources"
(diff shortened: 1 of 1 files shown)


gxalpha:
Please make sure to fill out the PR template instead of deleting it.

RytoEX:
I've converted this to a draft until the PR template is filled out and details are provided. What platform do you observe this behavior on? What is the error message?

PatTheMav (approved):
Looks fine to me, having an empty list evaluate to an empty string (and thus to `false`) is a bit of implicit CMake sorcery, but I'll allow it.

Note that the same check should be added for `OBS_MODULE_LIST` as well, preferably in a separate PR.

RytoEX:
> Looks fine to me, having an empty list evaluate to an empty string (and thus to `false`) is a bit of implicit CMake sorcery, but I'll allow it.
> 
> Note that the same check should be added for `OBS_MODULE_LIST` as well, preferably in a separate PR.

Is there a more correct/modern method? If so, we should probably just use that.

PatTheMav:
> > Looks fine to me, having an empty list evaluate to an empty string (and thus to `false`) is a bit of implicit CMake sorcery, but I'll allow it.
> > Note that the same check should be added for `OBS_MODULE_LIST` as well, preferably in a separate PR.
> 
> Is there a more correct/modern method? If so, we should probably just use that.

Not really, there is a more convoluted method of using the `list` function to return the actual length of the list into a parameter, followed by a numerical `> 0` check on that. But CMake lists are in essence just semicolon-separated strings, so checking for string length is as good.

WizardCM:
Please validate using cmake-format, and once you've pushed the fix I'll merge this.

gxalpha:
This PR breaks macOS builds on both my local build as well as CI (plugins don't get copied over into the app bundle anymore at all).
I think it should get reverted for now.

PR #6222, opened 2022-03-27
obs-transitions: Add HDR support to cut/fade

jpark37:
### Description
Both transitions are capable of passing through HDR sources now.

Also update VS solution to list transition effect files.

### Motivation and Context
Need transitions to support HDR to render their HDR scene children.

### How Has This Been Tested?
HDR scenes were being tonemapped behind these transitions before, and now they aren't.
SDR scenes continue to look SDR.
Tested fade transition for scenes and sources using SDR & HDR.

### Types of changes

### Checklist:

### DIFF
--- a/plugins/obs-transitions/CMakeLists.txt
+++ b/plugins/obs-transitions/CMakeLists.txt
@@ -14,6 +14,23 @@ target_sources(
           transition-luma-wipe.c
           transition-stinger.c)
 
+if(NOT OS_MACOS)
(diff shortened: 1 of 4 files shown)


PatTheMav (approved):
Looks like adding the filter files was copied verbatim from the `libobs` target, so seems fine as-is.

jpark37:
Well, I copied from obs-filters, which had more stuff than the libobs build script.

PR #6228, opened 2022-03-28
UI: Fix manage broadcast crash

### LINKED ISSUE #6227: Youtube manage broadcast dialog crashes in wayland on gnome 41
### Operating System Info

Other

### Other OS

Fedora 35

### OBS Studio Version

27.2.3

_No response_

### OBS Studio Log URL

https://obsproject.com/logs/yXBM6JlqutxUfGQM

### OBS Studio Crash Log URL

_No response_

### Expected Behavior

Obs doesnt close.

### Current Behavior

The dialog causes gnome to issue a fatal wayland protocol exception and closes the connection. This doesnt appear to cause a fatal error on other compositors (sway) but the window behaves weirdly suggesting something is definitely incorrect.

Replicated from a report in discord.

### Steps to Reproduce

1. Launch obs
2. Link with youtube
3. Select "Manage Broadcast"
4. Select "Select Existing Broadcasts"
5. obs closes immediately.

I doubt this is useful for us but the wayland debug log right before the crash.

```
[ 718440.740]  -> xdg_toplevel@48.set_app_id("com.obsproject.Studio")
[ 718441.113]  -> wl_surface@42.set_buffer_scale(1)
[ 718441.484]  -> wl_surface@42.set_buffer_transform(0)
[ 718441.804]  -> wl_surface@42.commit()
[ 718442.070]  -> wl_compositor@4.create_region(new id wl_region@49)
[ 718442.487]  -> wl_region@49.add(3, 30, 616, 712)
[ 718443.011]  -> wl_surface@42.set_opaque_region(wl_region@49)
[ 718443.388]  -> wl_region@49.destroy()
[ 718444.213] wl_buffer@32.release()
[ 718445.252]  -> xdg_toplevel@48.set_min_size(562, 738)
[ 718445.781]  -> xdg_toplevel@48.set_max_size(720, 880)
[ 718446.180]  -> wl_surface@42.commit()
[ 718446.497]  -> wl_compositor@4.create_region(new id wl_region@50)
[ 718446.906]  -> wl_region@50.add(3, 30, 616, 738)
[ 718447.396]  -> wl_surface@42.set_opaque_region(wl_region@50)
[ 718447.752]  -> wl_region@50.destroy()
...
[ 719721.243] wl_pointer@3.frame()
[ 719722.174]  -> xdg_toplevel@48.set_min_size(746, 738)
[ 719722.386]  -> xdg_toplevel@48.set_max_size(720, 880)
[ 719722.567]  -> wl_surface@42.commit()
[ 719722.716]  -> xdg_surface@47.set_window_geometry(0, 0, 752, 771)
[ 719722.962]  -> wl_compositor@4.create_region(new id wl_region@54)
[ 719723.156]  -> wl_region@54.add(3, 30, 746, 738)
[ 719723.414]  -> wl_surface@42.set_opaque_region(wl_region@54)
[ 719723.594]  -> wl_region@54.destroy()
[ 719723.753] wl_buffer@34.release()
[ 719723.980] wl_callback@49.done(3395)
[ 719724.247]  -> wl_shm_pool@52.destroy()
[ 719724.455]  -> wl_buffer@53.destroy()
[ 719724.678]  -> wl_surface@25.attach(wl_buffer@34, 0, 0)
[ 719724.975]  -> wl_surface@25.damage(0, 0, 2147483647, 2147483647)
[ 719725.222]  -> wl_shm_pool@46.destroy()
[ 719725.449]  -> wl_buffer@51.destroy()
[ 719725.690]  -> wl_surface@25.commit()
[ 719725.932]  -> wl_display@1.sync(new id wl_callback@49)
[ 719726.198]  -> wl_shm@5.create_pool(new id wl_shm_pool@50, fd 45, 2319168)
[ 719726.818]  -> wl_shm_pool@50.create_buffer(new id wl_buffer@55, 0, 752, 771, 3008, 0)
[ 719728.263] wl_display@1.error(wl_surface@42, 4, "Invalid min/max size")
```

Invalid surface size seems about right as we see the window size changing unexpectedly on other compositors which maybe is them accommodating this incorrect behavior. Likely the qt ui for this needs to be modified.

kkartaltepe:
### Description

This removes the min and max size properties which lead to crashes
on wayland compositors like gnome which are strict about these
values.

I'm not sure if there is a better way to fix this via the qt gui editor or if these will be reinserted if someone uses that tool. If they will be then likely we need to find a more permanent solution. But i dont have that tool on my system (nor do i want to learn how to use it...). Enforcing sizing constraints on the top level surface seems weird anyway (the dialog gets very wide with non-english languages).

fixes #6227

### Motivation and Context

Crashes are bad.

### How Has This Been Tested?

Tested on fedora 35

### Types of changes

### Checklist:

### DIFF
--- a/UI/forms/OBSYoutubeActions.ui
+++ b/UI/forms/OBSYoutubeActions.ui
@@ -16,18 +16,6 @@
     <verstretch>0</verstretch>
    </sizepolicy>
   </property>
-  <property name="minimumSize">
-   <size>
-    <width>0</width>
-    <height>0</height>
-   </size>
-  </property>
-  <property name="maximumSize">
-   <size>
-    <width>720</width>
-    <height>880</height>
-   </size>
-  </property>
   <property name="windowTitle">
    <string>YouTube.Actions.WindowTitle</string>
   </property>


tt2468:
These constraints would only be re-added if someone were to explicitly set them.

WizardCM:
Does wayland break due to both existing, or only them minimum? We could either just remove that, or set sane minimums.

kkartaltepe:
> Does wayland break due to both existing, or only the minimum? We could either just remove that, or set sane minimums.

I think the issue is due to the maximums. I removed the minimums as well since they seemed to contribute nothing. But i didnt test either in isolation.

RytoEX:
@GeorgesStavracas Is this something we should look to report upstream somewhere?

@Warchamp7 I know you were trying to prevent users from maximizing the Manage Broadcast, but crashing is bad. The only alternative I can think of is to set a higher maximum size, but a user may still run into a scenario where the dialog requests a larger size and this occurs again.

image

GeorgesStavracas:
[QTBUG-102626](https://bugreports.qt.io/browse/QTBUG-102626) now tracks the Wayland protocol violation from qt-wayland. I think in this case, the best course of action is do the following:

 * Remove max sizes (this PR already does this)
 * Let the dialog propagate its minimum sizes to the compositor
 * Make the dialog unresizable

If the contents of the dialog a just too large to be left unbounded, then it should be wrapped into a scrollview.

RytoEX:
> [QTBUG-102626](https://bugreports.qt.io/browse/QTBUG-102626) now tracks the Wayland protocol violation from qt-wayland. I think in this case, the best course of action is do the following:
> 
>     * Remove max sizes (this PR already does this)
> 
>     * Let the dialog propagate its minimum sizes to the compositor
> 
>     * Make the dialog unresizable
> 
> 
> If the contents of the dialog a just too large to be left unbounded, then it should be wrapped into a scrollview.

I'll follow that QTBUG. Thanks for submitting it.

To clarify, the contents of the dialog are too _wide_ because the buttons are in the second tab are wider than the maximum width set. I don't think we want the contents to scroll left and right.

I don't know how we "Let the dialog propagate its minimum sizes to the compositor", unless that is just a side effect of not manually setting minimum sizes.

We can probably set the dialog to be unresizable if @Warchamp7 thinks that's okay.

Right now, I'm mostly looking for whether or not this is okay to land as-is, or if it needs further adjustments.

GeorgesStavracas:
I can massage this dialog from a design)ish) perpective in a separate pull request. For now I think the only change this PR needs is rewording the commit to mention something like "maximum window sizes are a pattern to be avoided" instead of "fix this crash"

RytoEX:
> I can massage this dialog from a design(ish) perpective in a separate pull request. For now I think the only change this PR needs is rewording the commit to mention something like "maximum window sizes are a pattern to be avoided" instead of "fix this crash"

To be fair, the commit message is more detailed than the PR title:
> UI: Remove top level size constraints
> 
> This removes the min and max size properties from the youtube "Manage
> Broadcast" dialog which lead to crashes on wayland compositors like
> gnome which are strict about these values.

I'm not sure if we need the additional message. @kkartaltepe ? 

I'll let you discuss the pattern with @Warchamp7 off-thread, since he has previously expressed a desire to set UI constraints where possible.

kkartaltepe:
I've updated the message.

RytoEX (changes requested):
Commit message nits:
1. YouTube
2. Windows
3. Wayland
4. GNOME

Other than that, seems fine to me. Will ping the others again, but this should be good to go other than these nits, unless there's a significant objection.

kkartaltepe:
Updated the comment, windows here is not the windows platform so I clarified that one.

RytoEX:
> Updated the comment, windows here is not the windows platform so I clarified that one.

Whoops. I might've skimmed too quickly.

RytoEX (approved):
Looks good to me.

GeorgesStavracas (approved):
Looks good to me as well :)

PR #6233, opened 2022-03-28
cmake: Fix obs-ffmpeg-mux missing rpath entries for libobs in build tree

PatTheMav:
### Description
Fixes broken recording functionality on macOS when running OBS from Xcode.

### Motivation and Context
When using `install`, CMake will fix up a binary's rpath entries automatically. As `install` is also used to copy all build artifacts into a common runtime directory, this will lead to independent binaries such as "obs-ffmpeg-mux" to be fixed up to run within a fully portable app bundle.

Yet during development, the app bundle is not fully portable, so rpath entries to e.g. libobs need to be retained. Using CMake's `install PROGRAMS` variant will achieve just that.

### How Has This Been Tested?
OBS built, run, and recording created straight from Xcode.

### Types of changes

### Checklist:

### DIFF
--- a/cmake/Modules/ObsHelpers_macOS.cmake
+++ b/cmake/Modules/ObsHelpers_macOS.cmake
@@ -304,11 +304,15 @@ function(setup_obs_modules target)
   if(TARGET obs-ffmpeg-mux)
     add_dependencies(${target} obs-ffmpeg-mux)
 
+    install(TARGETS obs-ffmpeg-mux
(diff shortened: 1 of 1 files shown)


jpark37:
This fixes my issue of not being able to record.

RytoEX:
> so rpath entries to e.g. libobs need to be retained

This is probably more correctly written as:
> so rpath entries (e.g., to libobs) need to be retained

But I won't hold up merging this over that.

PR #6238, opened 2022-03-28
obs-ffmpeg: Add mxf extension to open file dialog filters

mvji:
### Description

Add .mxf extension to local file dialog extension filters.

### Motivation and Context

Let choose an MXF file in Obs FFmpeg source, using the open file selector

### How Has This Been Tested?

Mac os 11.6.2
Mac intel
Xcode

Tested with trying to load mxf file with and without the patch

### Types of changes

### Checklist:

### DIFF
--- a/plugins/obs-ffmpeg/obs-ffmpeg-source.c
+++ b/plugins/obs-ffmpeg/obs-ffmpeg-source.c
@@ -117,9 +117,9 @@ static void ffmpeg_source_defaults(obs_data_t *settings)
 }
 
 static const char *media_filter =
-	" (*.mp4 *.ts *.mov *.flv *.mkv *.avi *.mp3 *.ogg *.aac *.wav *.gif *.webm);;";
+	" (*.mp4 *.ts *.mov *.mxf *.flv *.mkv *.avi *.mp3 *.ogg *.aac *.wav *.gif *.webm);;";
 static const char *video_filter =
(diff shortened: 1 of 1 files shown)


RytoEX (changes requested):
Per our [Contributing guildelines](https://github.com/obsproject/obs-studio/blob/b3870476324e2dc79a9d30371a1199376bcd0ad7/CONTRIBUTING.rst), please make these changes:
1. Use the correct commit prefix, in this case "obs-ffmpeg: ". Please note that there is no space between the prefix term ("obs-ffmpeg") and the colon after it.
2. Properly capitalize the first word after the commit prefix.

WizardCM:
An example file would be useful, as I'd like to see this tested on all platforms.

Additionally, should this also be supported in drag and drop?

mvji:
> An example file would be useful, as I'd like to see this tested on all platforms.

You can download two MXF test files here (One dnxhd, and one XDcam HD 422) : 
https://www.swisstransfer.com/d/57a12a04-b6cc-42c1-bdba-9b21270dd89b

> Additionally, should this also be supported in drag and drop?
Mxf files are already work (before this patch) if we drag and drop a media in the sources panel.

RytoEX (approved):

RytoEX:
Will merge after CI completes successfully.

PR #6239, opened 2022-03-28
obs-ffmpeg: Add end action UI property for local file source

mvji:
For local file, replace UI properties :
"Loop" and "Show nothing when playback ends"
by a property list call End Action
with three possible values : Hide, Freeze, Loop

### Description

Currently in Obs FFmpeg source, when using local file
the end action of a video file, can be set to three mode, 
using two checkbox (loop and Show nothing when playback ends)
If loop is choose, the checkbox "Show nothing when playback ends", do nothing.
The UI proposal, is to use a property list, with the three mode when local file is checked
Doesn't change the UI when Local file is unchecked.
Internally, doesn't change the way is_looping and clear_media_end works.

UI_EndAction

### Motivation and Context

Having something more easy to understand for user, to define end action of a local video file.

### How Has This Been Tested?

Mac os 11.6.2
Mac intel, Xcode
Tested using local file, and checking the three mode.

### Types of changes

### Checklist:

I'm not sure it's the cleanest way to implement it.
Mainly for UI change for non local file, and using integer or enum for property list.


RytoEX:
When changing properties like this, you need to account for existing setups in the wild. You should cleanly migrate such settings to the equivalent value in the new settings or version the source such that old versions still work unchanged while newly added sources use the new behavior.

mvji:
I update the patch. And let the previous property "looping", but hide it.
And add callback, when parameter change, to update the new property list.
Tested with a scene with 3 local file (in each end action mode). Opening and modify it in previous version and current version. Behaviour stay the same when switching between the two version

WizardCM:
I'm not 100% sold on terminology. I think at the very least, "Freeze on final frame" (if that's what happens) would be better than "Freeze". If the "Hide" action doesn't actually toggle the eye to automatically hide the source, then I'd like to avoid the term "Hide" in favour of the previous "Show nothing".

I'm unsure how better to communicate the term "End Action" without it being very long. "When Playback Ends.." maybe? I don't know.

mvji:
> I'm not 100% sold on terminology. I think at the very least, "Freeze on final frame" (if that's what happens) would be better than "Freeze". If the "Hide" action doesn't actually toggle the eye to automatically hide the source, then I'd like to avoid the term "Hide" in favour of the previous "Show nothing".

Yes we can use Show nothing, who avoid confusion with show/hide property.

Another call for the "Freeze" action (who keep the last frame), can be "Frame hold" (is something close to what Premiere Pro used for this), or "Freeze last frame" who is shorter than "Freeze on final frame".

> 
> I'm unsure how better to communicate the term "End Action" without it being very long. "When Playback Ends.." maybe? I don't know.

I think shorter text are cleaner for UI. But if you prefer another text, will update.

Username69992872:
Will this function finally auto disable Media Sources which finshed playing? Streamdeck is a pain to use with OBS Studio right now as Media Sources stay active, event they finished playing. IF changing scenes, all Medisa Sources will re-start which is a pain.

mvji:
This PR doesn't change the existing behaviour (it change UI in local file mode).
There is already other properties for Media source, for what you want to do.

Username69992872:
> This PR doesn't change the existing behaviour (it change UI in local file mode). There is already other properties for Media source, for what you want to do.

yes I was referring to local video files indeed. So with this new "Hide" option, after the local Video has played, it would hide the local media source in OBS right?

mvji:
There is no new options in this PR, only a proposal for reorganize two checkbox to a property list with 3 parameters.

Currently the behaviour is :

Hide = Show nothing when playback ends (Checked) / Loop (Unchecked)
Freeze = Show nothing when playback ends (Unchecked) / Loop (Unchecked)
Loop = Loop (Checked)

> Is there a way I could test that allready witihout having to compile myself for win 64 bit? I guess it requires only 1 updated dll

Sorry didn't know how to enable test build in a PR.

gxalpha:
@Username69992872 I added the required label and restarted CI, once it's finished you should be able to download test artifacts from the "Checks" section of this PR.

SuslikV:
Show nothing
Freeze last frame
Play media again

tooltip:
Show nothing = make transparent
Freeze last frame = pause at last frame
Play media again = make endless loop

mvji:
Hello,

Thanks for your comments.

Don't see a function to add tooltip to a property list item.
"obs_property_set_long_description", seems to be able to set tooltip only for the entire property list.

Warchamp7 (changes requested):
I like this as a UX improvement.

Only suggestions I have are to phrasing and keeping familiar wording where we can.

End Action -> End of Playback
Freeze -> Freeze on last frame
Hide -> Show Nothing

SuslikV:
Don't forget to explain to translators that "loop" is verb here.

mvji:
> Don't forget to explain to translators that "loop" is verb here.

How to do that ?

SuslikV:
The way you can do this:
* reword
* ask translations maintainer (@Gol-D-Ace or whoever is now)

plugins/obs-ffmpeg/obs-ffmpeg-source.c
```diff
@@ -87,20 +87,73 @@ static bool is_local_file_modified(obs_properties_t *props,
...
+	bool looping = obs_data_get_bool(settings, "looping");
+	int end_playback = 0;
+
+	if (looping == true) {
+		end_playback = 2;
+	} else if (clear_on_media_end == false) {
+		end_playback = 1;
+	}
+
+	obs_data_set_int(settings, "end_playback", end_playback);
```
PatTheMav, line 146: You probably also want to use `obs_data_erase` to remove the old setting when upgrading it to the new variant. Otherwise the old setting will persist in a user's configuration and you will always go and update it.

Probably worthwhile to check if the old setting exists first and only go through the effort to upgrade when it's necessary in the first place.

RytoEX: A "safer" method would be to only upgrade the setting if the old setting exists and the new setting does not. Erasing old setting values completely breaks configs for older versions.

PatTheMav: Isn't that expected? If I were to create the source anew after the change, I could not go back to an older version of OBS and expect the setting to be "downgraded" either.

I'm aware of programs commonly upgrading settings but not that these upgraded settings are still expected to work with older versions (new settings would just be ignored).

RytoEX: I'm not saying that the upgraded settings should work with old versions. I'm saying that if you open a newer OBS one time that deletes settings entirely, and then you open an older OBS because you only wanted to run the newer OBS once to check it out, your settings are gone. I'm aware that we don't want to encourage or support downgrading, but purposefully breaking stuff seems a bit risky for users who just want to test stuff (and don't know that we encourage using portable mode or backing up your settings beforehand).

RytoEX: > Fine with me, there's no issue with keeping it. The property should be removed nevertheless though.

Agree with that.

PR #6242, opened 2022-03-29
obs-filters: Fix scale undistort, attempt two

jpark37:
### Description
Undistort checkbox should be ignored if bilinear lowres effect is used.

Fix regression caused by #6206.

### Motivation and Context
Clear rectangle without fix; proper draw with fix.

Undistort checkbox still works with bicubic and lanczos.

### How Has This Been Tested?
Checked bicubic, lanczos with and without undistort. Also checked point/bilinear/area this time.

### Types of changes

### Checklist:


WizardCM:
Could you provide some before/after screenshots so testers know what to look for?

jpark37:
This isn't testing the distort. Just having distort on in Bicubic/Lanczos is making other draws fail. Before is not seeing the draw, and after is seeing the draw. I don't think you need screenshots for that.

WizardCM:
Ah ok, understood.

PR #6243, opened 2022-03-29
cmake: Fix usage of relative paths for CEF finder

PatTheMav:
### Description
Re-enables usage of relative paths for `CEF_ROOT_DIR`.

### Motivation and Context
On some build systems (and in some usage scenarios) it is necessary to pass a relative path as `CEF_ROOT_DIR`. During testing it occurred that relative paths seem to be supported only for cache variables that have been explicitly marked as `PATH` variables.

### How Has This Been Tested?
Tested locally with relative paths used for `CEF_ROOT_DIR`.

### Types of changes

### Checklist:

### DIFF
--- a/cmake/Modules/FindCEF.cmake
+++ b/cmake/Modules/FindCEF.cmake
@@ -1,7 +1,9 @@
 include(FindPackageHandleStandardArgs)
 
-set_property(CACHE CEF_ROOT_DIR PROPERTY HELPSTRING
-                                         "Path to CEF distributed build")
+set(CEF_ROOT_DIR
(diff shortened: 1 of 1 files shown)


PatTheMav:
It should be noted that `CMAKE_PREFIX_PATH doesn't support relative paths (and I encountered other unexpected errors with relative paths): https://gitlab.kitware.com/cmake/cmake/-/issues/16644

This PR fixes the issue, but it's a porcelain-feature. Using `$(pwd)` or `${pwd}` and composing an absolute path with those should be preferred.

PR #6254, opened 2022-03-30
Unassorted PipeWire cleanups

GeorgesStavracas:
### Description

Some cleanups and small tweaks for the PipeWire capture code.

### Motivation and Context

The information that we currently log is not enough to give proper support, and at best allows us to know that the first few steps of negotiation worked. Improve these logs.

Also do some code shuffling to make it more sensibly organized in different files.

### Types of changes

### Checklist:

### DIFF
--- a/plugins/linux-pipewire/pipewire.c
+++ b/plugins/linux-pipewire/pipewire.c
@@ -582,7 +582,7 @@ static void renegotiate_format(void *data, uint64_t expirations)
 	obs_pipewire_data *obs_pw = (obs_pipewire_data *)data;
 	const struct spa_pod **params = NULL;
 
-	blog(LOG_DEBUG, "[pipewire] Renegotiating stream ...");
+	blog(LOG_INFO, "[pipewire] Renegotiating stream ...");
 
 	pw_thread_loop_lock(obs_pw->thread_loop);
 
@@ -806,18 +806,18 @@ static void on_param_changed_cb(void *user_data, uint32_t id,
 	if (has_modifier || check_pw_version(&obs_pw->server_version, 0, 3, 24))
 		buffer_types |= 1 << SPA_DATA_DmaBuf;
 
-	blog(LOG_DEBUG, "[pipewire] Negotiated format:");
+	blog(LOG_INFO, "[pipewire] Negotiated format:");
 
-	blog(LOG_DEBUG, "[pipewire]     Format: %d (%s)",
+	blog(LOG_INFO, "[pipewire]     Format: %d (%s)",
 	     obs_pw->format.info.raw.format,
 	     spa_debug_type_find_name(spa_type_video_format,
 				      obs_pw->format.info.raw.format));
 
-	blog(LOG_DEBUG, "[pipewire]     Size: %dx%d",
+	blog(LOG_INFO, "[pipewire]     Size: %dx%d",
 	     obs_pw->format.info.raw.size.width,
 	     obs_pw->format.info.raw.size.height);
 
-	blog(LOG_DEBUG, "[pipewire]     Framerate: %d/%d",
+	blog(LOG_INFO, "[pipewire]     Framerate: %d/%d",
 	     obs_pw->format.info.raw.framerate.num,
 	     obs_pw->format.info.raw.framerate.denom);
 
@@ -857,7 +857,7 @@ static void on_state_changed_cb(void *user_data, enum pw_stream_state old,
 
 	obs_pipewire_data *obs_pw = user_data;
 
-	blog(LOG_DEBUG, "[pipewire] stream %p state: \"%s\" (error: %s)",
+	blog(LOG_INFO, "[pipewire] stream %p state: \"%s\" (error: %s)",
 	     obs_pw->stream, pw_stream_state_as_string(state),
 	     error ? error : "none");
 }
@@ -973,7 +973,7 @@ static void play_pipewire_stream(obs_pipewire_data *obs_pw)
 		PW_STREAM_FLAG_AUTOCONNECT | PW_STREAM_FLAG_MAP_BUFFERS, params,
 		n_params);
 
-	blog(LOG_INFO, "[pipewire] playing stream…");
+	blog(LOG_INFO, "[pipewire] playing stream %p…", obs_pw->stream);
 
 	pw_thread_loop_unlock(obs_pw->thread_loop);
 	bfree(params);
--- a/plugins/linux-pipewire/pipewire.c
+++ b/plugins/linux-pipewire/pipewire.c
@@ -1228,15 +1228,19 @@ static void select_source(obs_pipewire_data *obs_pw)
 
 	available_cursor_modes = portal_get_available_cursor_modes();
 
-	if (available_cursor_modes & 4)
-		g_variant_builder_add(&builder, "{sv}", "cursor_mode",
-				      g_variant_new_uint32(4));
-	else if ((available_cursor_modes & 2) && obs_pw->cursor.visible)
-		g_variant_builder_add(&builder, "{sv}", "cursor_mode",
-				      g_variant_new_uint32(2));
+	if (available_cursor_modes & PORTAL_CURSOR_MODE_METADATA)
+		g_variant_builder_add(
+			&builder, "{sv}", "cursor_mode",
+			g_variant_new_uint32(PORTAL_CURSOR_MODE_METADATA));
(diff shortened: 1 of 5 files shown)


RytoEX:
This looks pretty straightforward. Are you waiting for anything in particular before marking it ready to review?

GeorgesStavracas:
I marked it as a draft because I wanted to check if this amount of logging was enough to figure out a situation in the linux-support channel. It was not enough. I'll add some more logging, and also reword some of the messages in there, and only then mark this PR as ready.

GeorgesStavracas:
I'm going to stop here because this branch is about to grow way too much. I'll push the other local changes in a separate PR.

plugins/linux-pipewire/pipewire.c
```diff
@@ -1136,7 +1134,7 @@ static void start(obs_pipewire_data *obs_pw)
 
 	new_request_path(obs_pw, &request_path, &request_token);
 
-	blog(LOG_INFO, "[pipewire] asking for %s…",
+	blog(LOG_INFO, "[pipewire] Asking for %s…",
```
RytoEX, line 1137: Replace the special character `…` with `...` (three separate `.` characters).

RytoEX (changes requested):
Minor nitpicks since we're changing log messages anyway.

plugins/linux-pipewire/pipewire.c
GeorgesStavracas: Ellipses are no more.

plugins/linux-pipewire/pipewire.c
```diff
@@ -1227,10 +1226,12 @@ static void select_source(obs_pipewire_data *obs_pw)
...
 			      g_variant_new_string(request_token));
 
-	if (obs_pw->available_cursor_modes & 4)
+	available_cursor_modes = portal_get_available_cursor_modes();
+
+	if (available_cursor_modes & 4)
 		g_variant_builder_add(&builder, "{sv}", "cursor_mode",
 				      g_variant_new_uint32(4));
-	else if ((obs_pw->available_cursor_modes & 2) && obs_pw->cursor.visible)
+	else if ((available_cursor_modes & 2) && obs_pw->cursor.visible)
```
notr1ch, line 1234: Could these constants be made into a `#define` to make it easier to understand what the code does?

GeorgesStavracas: Added an extra commit enumerating the available values.

RytoEX: @notr1ch Is the enum sufficient, or do you prefer defines for a particular reason?

plugins/linux-pipewire/portal.c
```diff
@@ -73,6 +73,25 @@ uint32_t portal_get_available_capture_types(void)
...
 
+enum portal_cursor_mode portal_get_available_cursor_modes(void)
+{
+	g_autoptr(GVariant) cached_cursor_modes = NULL;
+	uint32_t available_cursor_modes;
+
+	ensure_proxy();
+
+	if (!proxy)
+		return 0;
```
RytoEX, line 84: Should this be `0` or should `0` be represented by an enum entry or define?

RytoEX (commented):
A quick question.

plugins/linux-pipewire/portal.c
GeorgesStavracas: I'd say it should be zero, because this code path should return something to represent "couldn't fetch", which is a state we shouldn't reach when things are fine. This is different from the HIDDEN state, which is set to 1.

plugins/linux-pipewire/pipewire.c
notr1ch: The enum should be fine, I couldn't find anything other than numbers upstream so it seems there's no preferred way to represent these values.

plugins/linux-pipewire/portal.c
RytoEX: I guess my question was more, "Should we explicitly add an entry for `0` in the enum so that it's descriptive?" I'm fine either way.

GeorgesStavracas: Sorry, I was *trying* to answer this question :slightly_smiling_face: 

My usual stance on enums is that there needs to be a very specific reason to enumerate "invalid" - for example, when the "invalid" value is passed as actual, expected input somewhere. This is not the case here. In the case of this code, the invalid case is just something we have to loosely protect against, so I don't see any reason to upcast it into the enum.

plugins/linux-pipewire/portal.h
```diff
@@ -23,7 +23,20 @@
 #include <stdint.h>
 #include <gio/gio.h>
 
-uint32_t portal_get_available_capture_types(void);
+enum portal_capture_type {
```
kkartaltepe, line 26: nit, if these are intended to be bitflags they shouldnt be enums. If they are not intended to be bitflags they shouldnt be using different bits.

for example portal_get_available_capture_types now returns an invalid enum member if multiple capture types are available. It should continue to use a uint32_t, and not defining bitsets as enums avoids this kind of problem.

GeorgesStavracas: This is... tricky. The portal documents these values like this, and they're used both as flags AND enums throughout the stack. In our case, we already have code that uses that in both ways.

kkartaltepe: The portal is defined in dbus syntax, so i dont pretend to know what constitutes a valid enum in that syntax. But in C, we violate the enum definition by setting it to 3, and can see that things like switch statements would fail on such a value. C does not include a type for bitsets so using integers is what we have left.

plugins/linux-pipewire/portal.c
RytoEX: Good enough for me.

plugins/linux-pipewire/portal.h
GeorgesStavracas: For all intents and purposes, this is an enum. It's an enum that, in one specific situation, acts as a bit flag. This is all expressed in the code: this is declared as an enum (satisfying the enum condition), and as per the portal documentation, each value is defined as a shifted bit (satisfying the bit flag condition).

It must be both, simultaneously. The values must be able to be bit-shifted, otherwise there's no way to express support for multiple capture types through the `AvailableCaptureTypes` D-Bus property. But on every other circumstance, the plugin treats it as an enum, because it never handler nor sends more than one value.

On every situation that this is treated as an enum, receiving 3 must be considered an invalid value.

kkartaltepe: I think you misunderstood me, again the abuse is in using an enum as a bitset (which is common if you treat them like bitflags). And the error is in `portal_get_available_capture_types` which is written to return an enum but can (and will) return 3, which is an invalid value for the enum. Which is then casted away (luckily) to a uint32_t and used correctly by luck.

The only reason to not use bitflag enums is to make such a mistake less likely. Not because bitflag style enums are inherently in violation of any C rules by themselves.

GeorgesStavracas: Well, okay. I'll be honest with you, in almost 20 years of reading and writing C code, I have never heard of this "bitflags must not be enums" rule. What would you suggest here then? Drop the last commit?

kkartaltepe: 1) change the return types for `portal_get_available_capture_types` and `portal_get_available_capture_types` to `uint32_t` to fix the latent bug where we cast invalid values into the enum type.

2) consider if this experience has changed your opinion on using bitflags as enums.

GeorgesStavracas: Thanks, changed the return types to `uint32_t`

kkartaltepe (approved):

plugins/linux-pipewire/portal.h
RytoEX: > 1. change the return types for `portal_get_available_capture_types` and `portal_get_available_capture_types` to `uint32_t` to fix the latent bug where we cast invalid values into the enum type.

You wrote `portal_get_available_capture_types` twice. Did you mean something else @kkartaltepe ?

kkartaltepe: Sorry i must have fat fingered copy and paste, it was supposed to be `portal_get_available_cursor_modes` but georges got it.

RytoEX: Understood.

RytoEX (approved):

PR #6256, opened 2022-03-31
CI: Fix Windows/macOS release jobs

cg2121:
### Description
- Hardcode short hash length to 8, as it is not guaranteed that short hashes
are the same length. This caused problems with upload/download action names, as
the hashes sometimes didn't match.

- Fixes Windows installer job, as the download artifact name was incorrect

- Same as above but for the macOS notarize job

### Motivation and Context
Found bugs when testing workflows.

### How Has This Been Tested?
Tested with fork.

### Types of changes

### Checklist:

### DIFF
--- a/.github/workflows/flatpak.yml
+++ b/.github/workflows/flatpak.yml
@@ -36,10 +36,15 @@ jobs:
         with:
           submodules: 'recursive'
 
+      - name: 'Setup build environment'
+        id: setup
+        run: |
+          echo "::set-output name=commitHash::$(git rev-parse --short=9 HEAD)"
+
       - name: Build Flatpak Manifest
         uses: bilelmoussaoui/flatpak-github-actions/flatpak-builder@master
         with:
-          bundle: obs-studio-${{ github.sha }}.flatpak
+          bundle: obs-studio-${{ steps.setup.outputs.commitHash }}.flatpak
           manifest-path: CI/flatpak/com.obsproject.Studio.json
           cache-key: flatpak-builder-${{ github.sha }}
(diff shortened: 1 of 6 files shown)


RytoEX:
> Hardcode short hash length to 8

CI currently appears to be using 9 character short hashes. Do we want to stick with 9? Do we want to use 10?

> Fixes Windows installer job, as the download artifact name was incorrect
> Same as above but for the macOS notarize job

Could you elaborate on what was incorrect? Or do you mean that the aforementioned short hash length was causing the artifact names to be incorrect?

RytoEX:
> GitHub uses 7 even, I guess making this explicit is best to ensure no surprises, but also have no real opinion on what the right length should be.

As I understand it, Git defaults to 7 characters for the short hash, and expands as collisions occur (or become more likely to occur, I'm unclear on this). I may be misunderstanding the literature, but my rudimentary understanding is that we're at roughly 20% chance of a hash collision with our current number of commits.

Using this to get a unique string length given our current commits:
```
git rev-list --abbrev-commit --max-count=1 --format="%h" HEAD
```
I get a 9-character string.

That being the case, it seems that a 9-character short hash right now is safe, though may not always be safe. I do not know what the thresholds are for when a 10-character short hash would be safe/required.

At the moment, we can either hard-code it to 9-characters or use something like the command I used above to use git to determine a safe short hash length for any given commit.

RytoEX:
I've pushed changes to adjust the short hash length from 8 characters to 9 characters. I've also made these changes in the provided FreeBSD package script as well as the Flatpak CI workflow and job.

PR #6257, opened 2022-04-01
libobs,UI: Add support for nonlinear SRGB blending

jpark37:
### Description
Nonlinear blending is useful to emulate blending behavior of other applications.

### Motivation and Context
This is mainly for the browser source, since the web will probably never change to use linear color blending.

### How Has This Been Tested?
- Verified in RenderDoc that alpha blending for browser source behaves as expected when "SRGB Off" method is used, and linearly when "Default" method is used.
- Setting is preserved when source is copy/pasted.
- Setting is preserved across OBS restarts.
- Localized strings appear correctly in English.

### Types of changes

### Checklist:

### DIFF
--- a/UI/data/locale/en-US.ini
+++ b/UI/data/locale/en-US.ini
@@ -480,6 +480,11 @@ ScaleFiltering.Bicubic="Bicubic"
 ScaleFiltering.Lanczos="Lanczos"
 ScaleFiltering.Area="Area"
 
+# blending methods
+BlendingMethod="Blending Method"
+BlendingMethod.Default="Default"
+BlendingMethod.SrgbOff="SRGB Off"
+
 # blending modes
 BlendingMode="Blending Mode"
 BlendingMode.Normal="Normal"
--- a/UI/window-basic-main.cpp
+++ b/UI/window-basic-main.cpp
@@ -5320,6 +5320,36 @@ QMenu *OBSBasic::AddScaleFilteringMenu(QMenu *menu, obs_sceneitem_t *item)
 	return menu;
 }
 
+void OBSBasic::SetBlendingMethod()
+{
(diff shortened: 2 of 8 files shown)


UI/data/locale/en-US.ini
```diff
@@ -480,6 +480,11 @@ ScaleFiltering.Bicubic="Bicubic"
 ScaleFiltering.Lanczos="Lanczos"
 ScaleFiltering.Area="Area"
 
+# blending methods
+BlendingMethod="Blending Method"
+BlendingMethod.Default="Default"
+BlendingMethod.SrgbOff="SRGB Off"
```
WizardCM, line 486:
```suggestion
BlendingMethod.SrgbOff="sRGB Off"
```

For consistency.

WizardCM (commented):
Just for posterity and my own understanding, is the plan for obs-browser to automatically use sRGB Off for all existing and future browser source scene items by default, and then the user can disable it themselves if necessary? Meanwhile all other sources will just be set to Default

UI/data/locale/en-US.ini
jpark37: The capital S is on purpose.

sRGB - The color space.
SRGB - The GPU automatic transfer space conversion. (e.g. D3D: UNORM_SRGB, GL: FRAMEBUFFER_SRGB)

jpark37:
Browser source will not automatically opt in. SRGB Off has an additional side effect of not respecting the linear filtering on texture read, so enabling by default would alter rescaling even without alpha blending.

Anyone who wants to opt into SRGB Off will need to do so manually. This setting is also compatible with image/media source when linear alpha is disabled to get blend behavior comparable to v26.

Issue #5877, opened 2022-01-28
obs crashs on exit in pipewire_capture_unload -> pw_deinit


jirislaby:
### Operating System Info

Other

### Other OS

openSUSE

### OBS Studio Version

27.1.3

### OBS Studio Log URL

https://obsproject.com/logs/HYbYvnhMak0vBV5v

### Expected Behavior

No crash.

### Current Behavior

A crash on exit happened.

### Steps to Reproduce

1. start OBS
2. quit OBS

### Anything else we should know?

Stack trace:
```
#0  pw_deinit () at ../src/pipewire/pipewire.c:687
#1  0x00007f13565f699c in free_module (mod=0x5569be2abe40) at /usr/src/debug/obs-studio-27.1.3-1.26.x86_64/libobs/obs-module.c:493
#2  0x00007f13565d934f in obs_shutdown () at /usr/src/debug/obs-studio-27.1.3-1.26.x86_64/libobs/obs.c:1022
#3  0x00005569bbe9f215 in OBSApp::~OBSApp() (this=0x7ffca46e1bb0, __in_chrg=<optimized out>) at /usr/src/debug/obs-studio-27.1.3-1.26.x86_64/UI/obs-app.cpp:1190
#4  0x00005569bbe7c2d4 in run_program (argv=0x7ffca46e2218, argc=<optimized out>, logFile=...) at /usr/src/debug/obs-studio-27.1.3-1.26.x86_64/UI/obs-app.cpp:2149
#5  main(int, char**) (argc=<optimized out>, argv=0x7ffca46e2218) at /usr/src/debug/obs-studio-27.1.3-1.26.x86_64/UI/obs-app.cpp:2825
```
NULL dereference in more detail:
```
#0  pw_deinit () at ../src/pipewire/pipewire.c:687
687                     p->ref++;
(gdb) p p
$4 = (struct plugin *) 0x0
(gdb) l
682             pthread_mutex_lock(&init_lock);
683             pthread_mutex_lock(&support_lock);
684             pw_log_set(NULL);
685             spa_list_consume(p, &registry->plugins, link) {
686                     struct handle *h;
687                     p->ref++;
688                     spa_list_consume(h, &p->handles, link)
689                             unref_handle(h);
690                     unref_plugin(p);
691             }
```

kkartaltepe:
[...]

PR #6260, opened 2022-04-01
linux-pipewire: remove call to pw_deinit

kkartaltepe:
### Description
Per upstream this function is not supposed to be called by users of
pipewire. This also removes the nearly empty pipewire-common files.

ref: https://gitlab.freedesktop.org/pipewire/pipewire/-/issues/2238#note_1311774

### Motivation and Context

May fix crashes on close for users in case anything else in their process calls pw_deinit. e.g. #5877

### How Has This Been Tested?

It compiled.

### Types of changes

### Checklist:


plugins/linux-pipewire/linux-pipewire.c
```diff
@@ -53,5 +53,5 @@ bool obs_module_load(void)
 
 void obs_module_unload(void)
 {
-	obs_pipewire_unload();
+	// This space intentionally left blank.
```
GeorgesStavracas, line 56: Is this comment really necessary? I think we can go without :slightly_smiling_face:

GeorgesStavracas (changes requested):
Please change the commit message capitalization to `linux-pipewire: Remove call to pw_deinit`

RytoEX (changes requested):
Commit message nitpicks:
1. Write "PipeWire" when possible. The commit prefix need not comply with this.
2. Use sentence case for the first word after the commit prefix (capitalize the first letter).

columbarius:
The PipeWire documenation was updated again:
> Since 0.3.49 this function must be paired with an equal amount of pw_init() calls to deinitialize the PipeWire library.

https://gitlab.freedesktop.org/pipewire/pipewire/-/blob/master/src/pipewire/pipewire.c#L681

Should we still remove `pw_deinit` or just hide it when compiled with PipeWire versions < 0.3.49?

kkartaltepe:
> Should we still remove pw_deinit or just hide it when compiled with PipeWire versions < 0.3.49?

Does pipewire provide an appropriate macro to test this? We can add it back in if we want to have a clean(er) memsan/valgrind

columbarius:
PipeWire has the macro `PW_CHECK_VERSION(major,minor,micro)`
https://gitlab.freedesktop.org/pipewire/pipewire/-/blob/master/src/pipewire/version.h.in#L59

GeorgesStavracas:
Interesting, thanks @columbarius 

I guess we can go with

```c
void obs_pipewire_unload(void)
{
#if PW_CHECK_VERSION(0,3,50)
	pw_deinit();
#endif
}
```

columbarius:
LGTM. I had plans to add additional stuff to pipewire-common when unifying common code between the plugins, but the changes just made things more complicated and they are probably lost for good.

kkartaltepe:
Thanks, i set the version check to 0.3.49 since that should be the first safe version

GeorgesStavracas (changes requested):
There's one extra `pipewire` in the commit message that should be changed to `PipeWire`.

LGTM other than that

GeorgesStavracas (approved):

PR #6261, opened 2022-04-01
linux-pipewire: Provide the parent_window identifier to the portal

Sodivad:
### Description
Send the parent_window specifier as described in https://flatpak.github.io/xdg-desktop-portal/#parent_window
On X for this the window id is used. on Wayland a handle
acquired from the compositor via the xdg-foreign protocol.

### Motivation and Context

This enables the system to associate any windows/dialogs opened
by the portal with the OBS main window. For example positioning
them on/near the OBS window instead randomly somewhere on the
screen.

### How Has This Been Tested?

Tested on Plasma Wayland,  the portal dialog is not placed somewhere randomly but on top
of obs.

### Types of changes

### Checklist:

### DIFF
--- a/plugins/linux-pipewire/CMakeLists.txt
+++ b/plugins/linux-pipewire/CMakeLists.txt
@@ -9,6 +9,7 @@ endif()
 find_package(PipeWire REQUIRED)
 find_package(Gio QUIET)
 find_package(Libdrm QUIET) # we require libdrm/drm_fourcc.h to build
+find_qt(VERSION ${QT_VERSION} COMPONENTS Widgets)
 
 if(NOT TARGET PipeWire::PipeWire)
   obs_status(
@@ -33,6 +34,7 @@ add_library(OBS::pipewire ALIAS linux-pipewire)
 target_sources(
   linux-pipewire
   PRIVATE linux-pipewire.c
+          parent_window.cpp
           pipewire-common.c
           pipewire-common.h
           pipewire.c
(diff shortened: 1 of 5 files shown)


Sodivad:
Can I see why the cmake-format check failed somewhere? 
When I run the script on my machine it doesn't complain

WizardCM:
Forgive me as my Linux knowledge (especially around Pipewire) is limited, but this seems like a lot of work to just grab the parent window. I think it'd be worth attempting a method that doesn't require importing Qt Widgets unnecessarily.

Sodivad:
>Forgive me as my Linux knowledge (especially around Pipewire) is limited, but this seems like a lot of work to just grab the parent window. I think it'd be worth attempting a method that doesn't require importing Qt Widgets unnecessarily.

That's fair. On X the winId is enough (aka `obs_frontend_get_main_window_handle`), but on Wayland need to access somehow the `wl_surface` of the window.

GeorgesStavracas:
The reason I have decided against this in the past is that this is a massive layering violation; qt-wayland itself should implement this protocol, and there should be explicit APIs on Qt side that we can use from `obs_frontend_get_main_window_handle`.

RytoEX:
> The reason I have decided against this in the past is that this is a massive layering violation; qt-wayland itself should implement this protocol, and there should be explicit APIs on Qt side that we can use from `obs_frontend_get_main_window_handle`.

I agree with this, and I would rather work went into Qt on this issue by anyone interested in getting this resolved.

Sodivad:
FWIW I have also a MR to implement for Qt to implement xdg-foreign when it uses portals itself. Once that's in we can  probably figure out a way to access it.
https://codereview.qt-project.org/c/qt/qtwayland/+/400122

GeorgesStavracas:
We've discussed the changes in this PR, and I think the approach taken here is not ideal, in the sense that this is a layering violation. Once the appropriate APIs land in Qt, I'd love to see a PR passing the parent_window handler using that. For now, let's keep an eye on the progress of the Qt contributions you've linked in here.

Thanks for your contribution, looking forward seeing your next ones! :slightly_smiling_face:

PR #6268, opened 2022-04-03
UI: Make OAuth base URL configurable

derrod:
### Description

Makes OAuth URLs consistent and configurable by changing the base url.

### Motivation and Context

- OAuth URLs need ot be adjusted when migrating from legacy PHP scripts to cloudflare workers based setup (https://github.com/obsproject/obs-oauth-cf)
- Third-party distributions (e.g. on Linux) don't have access to our API credentials, so this will make it easier for them to self-host their own OAuth endpoints with their own

### How Has This Been Tested?
Built, didn't fail, more TBD.

### Types of changes

### Checklist:


UI/auth-oauth.cpp
```diff
@@ -91,7 +93,7 @@ void OAuthLogin::urlChanged(const QString &url)
 	if (code_idx == -1)
 		return;
 
-	if (url.left(22) != "https://obsproject.com")
+	if (url.startsWith(OAUTH_BASE_URL))
```
tytan652, line 96: It should be `(!url.startsWith(OAUTH_BASE_URL))` or I'm missing something.

derrod: No you're right, I forgot to change that.

derrod:
Verified to be working on my MacBook with my own credentials for Twitch, haven't been able to test Restream yet.

jp9000:
Did we add the subdomain already? The new paths work?

derrod:
> Did we add the subdomain already? The new paths work?

Not yet. I don't think @notr1ch had the time to complete the move yet.

Fenrirthviti:
Going to mark this as draft so it doesn't get accidentally merged until the supporting server-side stuff is taken care of and tested.

derrod:
Worker is now live at https://auth.obsproject.com/ but we still need to have the new redirect URLs added to Restream/Twitch for it to work properly.

derrod:
Just wanted to make sure that the CMake changes make sense and aren't "legacy" ways of doing things.

UI/CMakeLists.txt
```diff
@@ -26,6 +26,12 @@ if(TARGET obs-browser
     obs-browser-panels INTERFACE ${CMAKE_SOURCE_DIR}/plugins/obs-browser/panel)
 endif()
 
+set(OAUTH_BASE_URL
```
PatTheMav, line 29: Why use a cache variable here? If it's not supposed to be changeable by the user (which it isn't, you `force` the value over whatever a user specifies), make it either a normal project variable (by using `set` without qualifiers) or make it an internal cache variable (which usually only makes sense if you need it globally).

derrod: This was essentially just copied over from https://github.com/obsproject/obs-studio/blob/master/plugins/rtmp-services/CMakeLists.txt#L5-L9

The variable is supposed to be changeable in some cases (e.g. Linux packages or developers running their own OAuth servers since they don't have access to our credentials).

So in order to achieve that would I need to simply remove `FORCE` or also another part of it?

PatTheMav: Yeah remove the `FORCE` part or only set+force it if the value is an empty or illegal string.

PatTheMav (approved):
CMake change looks fine to me.

RytoEX:
> CMake change looks fine to me.

Thanks. Guess we're just waiting on third-parties for this now.

derrod:
> > CMake change looks fine to me.
> 
> Thanks. Guess we're just waiting on third-parties for this now.

That is my understanding, hopefully @jp9000 can get through to someone at Twitch.

derrod:
Just another ping to @jp9000 for adding the new redirect URLs to Restream/Twitch.

derrod:
Alright, that leaves Restream, as of now that still shows an error saying the `redirect_uri` does not match.

PR #6276, opened 2022-04-05
UI: Truncate displayed file paths in the middle

gxalpha:
### Description

In the Remux window, file paths currently truncate at the back, leading to the most important part of the path, the file name itself, often times being cut off.
This changes that behaviour so that it now truncates the path in the middle.

Before:
Bildschirmfoto 2022-04-05 um 01 58 44
After:
Bildschirmfoto 2022-04-05 um 01 57 57

The `setWordWrap` part is needed since wrapping is also considered truncating, independent of what the `setTextElideMode` says. However, as a side effect this also fixes a bug where in rare cases (and from what I can tell, only on macOS) it would actually wrap instead of truncating, which looked very weird:
Bildschirmfoto 2022-04-05 um 02 00 35:

### Motivation and Context

[Suggestion](https://discord.com/channels/348973006581923840/376042662580715521/960565563405525094) on the Discord server.

### How Has This Been Tested?

macOS 12.3, see screenshots above.

Resizing works as expected.

### Types of changes

### Checklist:


energizerfellow:
Would it be possible to have the displayed line items auto-size vertically with line wrap to show the entire path? Maybe make them interactable for sizing where if the user collapses the line item vertically it truncates?

WizardCM (approved):
Outside of additional testing on Windows and Linux, I think this change is perfect.

gxalpha:
> Would it be possible to have the displayed line items auto-size vertically with line wrap to show the entire path? Maybe make them interactable for sizing where if the user collapses the line item vertically it truncates?

That would be *possible*, but much more complex - and honestly, I find the UX of the current window to be suboptimal at best, with the implementation looking rather complicated (and sometimes being buggy in regards to focus as a result). I don't think making it more complicated would be a great idea.

PR #6278, opened 2022-04-05
cmake: Fixes plugins not being copied into application bundle on macOS

PatTheMav:
### Description
Fixes plugins not being copied into the application bundle for macOS builds.

### Motivation and Context
Makes the check for the number of plugins in the global list an explicit length check (not relying on CMake implicit functionality), also moves it in line with how the same was solved in `ObsHelpers.cmake` globally.

### How Has This Been Tested?
Built OBS with a fresh build directory on macOS and checked application bundle contents.

### Types of changes

### Checklist:


quink-black:
Sorry for the break. Now I know empty string is false, but non-empty string doesn't equal to true (most of the case it's false).
Thanks for the fix!

PR #6311, opened 2022-04-10
Add automated setup for Mac desktop audio capture

gxalpha:
### Description

Adds an automated setup process for desktop audio on macOS, transparent to the user.

#### Test Builds
* [M1 (ARM64)](https://github.com/obsproject/obs-studio/suites/6364990662/artifacts/229998535)
* [Intel (x86_64)](https://github.com/obsproject/obs-studio/suites/6364990662/artifacts/229998536)

#### UX
In the preferences, on macOS the user sees a button in the properties:
image
Pressing this brings up a popup, explaining what will happen (as well as a drawback regarding volume controls, explained further down in this PR):
image
Pressing "Yes" will bring up an admin prompt and a dialog showing that it's installing (during which any audio on the system will cut out), and after a few (~5) seconds the process will be done.
In the "Desktop Audio" dropdowns, as well as the device selection dropdowns in the "Audio Output Capture" source, the user will now be able to select "OBS Desktop Audio":
image
Any desktop audio will come through that source.
Reverting the process just takes pressing the "Uninstall" button. Should a user uninstall, but not unselect, this is how the device name looks, similar to `[Device not connected or available]`:
image

#### How it works
The PR is split into two (three) parts:
Firstly, there is mac-capture/HAL. This is a loopback device like BlackHole.
It is basically the sample device from the [official Apple sample code](https://developer.apple.com/documentation/coreaudio/creating_an_audio_server_driver_plug-in) with a ring buffer added to it, adapted from BlackHole (thus being GPLv3). BlackHole has a few more niceties and features, like volume controls for the HAL device, however we don't need that since this device is *completely* invisible to the user (and other apps on the system). To find it, you need to know its UID. This is also the motivation behind the `mac-capture: Simplify coreaudio_get_device_id` commit, OBS wouldn't be able to find it otherwise.
Secondly, there are the mac-capture changes. This adds the logic for the setup process. This includes:
- Installing/Uninstalling the HAL device. This means copying over the driver bundle and kickstarting CoreAudio. This is the part where the user would download BlackHole.
- Installing/Uninstalling the Multi-Output-Device. This includes creating a dictionary with the devices properties (including its sub-devices, which by default is our HAL device, as well as the current system output device), telling CoreAudio to create it and to set it as the default output device. This is the part for which the user would have to go into `Audio-Midi-Setup.app` and set up the Multi-Output-Device.
- Checking the status' of said devices.

This is directly linked with the UI changes. UI and mac-capture communicate via a procedure handler added on plugin initialization.
In the UI, the button is added. What it does depends on the installation status of the devices. It can
- Install (Which runs the installation process)
- Deinstall (Which runs the deletion process)
- Update (Which updates the HAL device) (this will show after OBS updates, where the HAL device could've changed, like the virtualcam)
- Fix (Which runs the deletion process) (this is if somethings gone terribly wrong, like the user manually deleting either device)

The device dropdown must update after something has changed, so that the device gets selectable without having to re-open the settings window, which is why `LoadAudioOutputDevices` was split out of `LoadAudioDevices`.

#### Drawbacks / Potential issues
- While the entire process is transparent, users can no longer adjust system volume because Multi-Output-Devices are dumb and wouldn't know what to do. This is the same when doing the setup manually, due to the nature of Multi-Output-Devices. I added the warning before installing, but some users might forget (or not read).
- After copying the HAL device, CoreAudio needs to be restarted to register it. Apple's recommended way of doing this is restarting the computer. We can't have that since we want the process to be seamless, so this includes a hack: Kickstaring CoreAudio via launchctl. It's clear that this is not something Apple expects you to do, since it doesn't seem to give you any indication of when it's back up, all API calls (like adding the Multi-Output-Device) return without an error while actually failing. This is why this PR includes a hack to the hack, which is waiting five seconds. In that time, it *should* be up again. If it isn't, it will be non-critical, but the user will need to reopen the settings page.
- The Multi-Output-Devices appear to only support two channels (and I think Apple is to blame for that). The HAL device itself is currently configured for 8, but only two of them get used in any scenario. Should we still have all 8, in case this changes on the Multi-Output-Device side in the future, or stick to two for now?

#### Conclusion
I think this makes a great addition towoards the user-friendliness of OBS on macOS.
Actually getting this PR to work nicely turned out to be much more painful than I expected it to be (which is why it took me so long to actually open the PR).

### Motivation and Context

The steps in the current macaudio guide are somewhat complicated and the guide is fairly long in general, and often times users don't even find it or make mistakes during the setup.
This newly introduced process should be user-proof, with nearly no way to screw it up, as a user there's no reason to even look into `Audio-Midi-Setup.app`.

### How Has This Been Tested?

So far, only from local and CI builds on my own machine (Intel), where I've been running a version of this for the last few weeks without noticing any issues.
This definitely needs more testing, especially on M1x machines, to ensure everything works properly.

### Types of changes

### Checklist:


prgmitchell:
I installed the build to give this a test. The install process seemed to go smooth and didn't throw any errors. Upon trying to select the dropdown to add my speakers as my desktop audio source, it still said disabled despite it properly reporting it was installed in the OBS audio settings.

Session log: https://obsproject.com/logs/MUdJqRbqSvqoneZS

Screenshots attached of the sound settings and device it made

https://ibb.co/mbkndQB
https://ibb.co/mv4SHLs

gxalpha:
Thanks for the report, should hopefully be fixed now - could you try again?

asher-the-thrasher:
Overall really sweet addition and happy to see it finally happening!
1) The logging is kinda weird, some of the log entries have `[mac-capture] blah blah blah` while others have `[mac-capture]: blah blah blah` making it kinda weird to see in the logs, would be nice to have just the second option to make it consistent with other log entries

2) something that needs to be discussed is where it makes sense to put the device on the list. Currently, it sits between disabled and default but I think it would make more sense to have it in the list with other options, or right under default and not have the device show up if it isn't installed because that could cause some confusion possibly.

plugins/mac-capture/mac-audio.m
```diff
@@ -637,6 +645,214 @@ static bool coreaudio_init(struct coreaudio_data *ca)
...
+		const char *errorMessage = [[error
+			objectForKey:@"NSAppleScriptErrorMessage"] UTF8String];
+		blog(LOG_WARNING,
+		     "[mac-capture]: Failed to modify HAL plugin: %s",
+		     errorMessage);
+		return false;
+	}
+
+	// Hack to wait until CoreAudio has reloaded. I hate it too.
+	sleep(5);
```
PatTheMav, line 717: Is there any way to make this async? I can imagine that it might the UI code a bit messy because you'd have to add a signal to handle the eventual status of the audio subsystem after waiting for 5 seconds.

plugins/mac-capture/mac-audio.m
```diff
@@ -637,6 +645,214 @@ static bool coreaudio_init(struct coreaudio_data *ca)
...
+			blog(LOG_INFO, "Deleting old device failed. Reason: %d",
+			     status);
+			return false;
+		}
+		return true;
+	}
+	return false;
+}
+
+static inline bool get_default_monitoring_output_device(CFStringRef *cf_uid)
```
PatTheMav, line 749: Is this function misnamed maybe? It suggest that it retrieves the default monitoring output device, but it returns a boolean value instead.

PatTheMav: Upon second reading I realized this is a C API and the default device name is probably stored at the address the `CFStringRef` pointer refers to.

PatTheMav:
I like that there is a warning before the app prompts for elevated privileges (something we should adapt for the virtualcam at some point as well), but it's also on the knife's edge of "TL;DR". The white-on-dark theme is not really great for reading text, so maybe on of our native speakers can help shrink down the wording.

UI/data/locale/en-US.ini
```diff
@@ -1031,6 +1031,15 @@ Basic.Settings.Audio.EnablePushToTalk="Enable Push-to-talk"
...
 Basic.Settings.Audio.UnknownAudioDevice="[Device not connected or not available]"
 Basic.Settings.Audio.Disabled="Disabled"
+Basic.Settings.Audio.MacAudio.Install="Install Desktop Audio Capture"
+Basic.Settings.Audio.MacAudio.Installing="Installing Desktop Audio Capture..."
+Basic.Settings.Audio.MacAudio.Fix="Fix Desktop Audio Capture installation"
+Basic.Settings.Audio.MacAudio.Update="Update Desktop Audio Capture installation"
+Basic.Settings.Audio.MacAudio.Updating="Updating Desktop Audio Capture installation..."
+Basic.Settings.Audio.MacAudio.Uninstall="Uninstall Desktop Audio Capture"
+Basic.Settings.Audio.MacAudio.Uninstalling="Uninstalling Desktop Audio Capture..."
+Basic.Settings.Audio.MacAudio.ConfirmInstall="Installing requires you to enter your adminstrator password and may take a few seconds.\n\nPlease note that system volume controls will no longer work natively.\nTo adjust audio volume after installation, switch the output device to your usual device in System Preferences, change the volume and then switch back to \"OBS Audio Capture\".\n\nDo you want to continue?"
```
RytoEX, line 1041: adminstrator -> administrator

RytoEX, line 1041: "volume and" -> "volume, and"

RytoEX (changes requested):
Minor string nitpicks. I'm not familiar with the recommended/required workflow being described by the text, so I can't well rephrase it to something shorter.

asher-the-thrasher:
another thing is maybe there should be 2 popups instead of one. The first one something like `Please note that system volume controls will no longer work natively.
To adjust audio volume after installation, switch the output device to your usual device in System Preferences, change the volume and then switch back to "OBS Audio Capture".`
and then `Installing requires you to enter your administrator password and may take a few seconds. Would you like to continue?` 

this way it is less of a text block and they are accepting 2 different things, how it will affect their system and then that they need their admin password

prgmitchell:
Has any thought been put into integrated this into the Audio Output Capture dialogue as well?

PatTheMav:
> Minor string nitpicks. I'm not familiar with the recommended/required workflow being described by the text, so I can't well rephrase it to something shorter.

In essence, once a user clicks the button a prompt for elevated privileges will pop up, requiring the user to enter their account password. This is necessary because we effectively run a shell command with `sudo` to copy the required files to their target location (which itself requires elevated privileges to be written to).

This change will also automatically create a composite audio device on macOS (consisting of the current default audio devices as well as the installed OBS loopback device) and attempt to make that one the system's default audio device.

Alas the built-in volume control will seize to work once this has been done, so the user is "locked-in" to the volume of their default output device and unable to change this directly (users have to open the MIDI Control Panel and manually change the volume sliders there).

---

If I'm honest - I am particularly unhappy about that last part. My dislike for "implicit magic" is probably well-known around these parts by now, and I don't think OBS should change the state of the system especially if it negatively impacts user expectation as much as it does here ("Why is my audio gone for 15s?", "Why can't I control my volume anymore?", "How do I get back my original setup?").

Requiring users to explicitly change their default output to the composite device requires them to familiarize themselves with macOS' audio settings, making them aware how to switch back. _Magically_ changing this (as this PR does) keeps users in the dark.

gxalpha:
Removed some of the *magic* as discussed. Also changed the order of the installation steps, which should make it so that in case of the 5 seconds not being enough, it's non-critical. Everything still installs properly, the user will just have to reopen the settings page.

I'm still a bit unhappy with the confirmation text (which I tweaked to compensate for the changes), if a native english speaker could chime in that'd be greatly appreciated!

The links to CI builds in the PR description will be updated once CI is done (should take ~10 minutes).

plugins/mac-capture/mac-audio.m
gxalpha: Yeah, I think it'd be overkill.
While I didn't try, it's likely to be too much code for too little reward (and there's little purpose to making it async imo, since the next thing the user does after clicking the button is selecting the device, which only is available after the 5 seconds).

plugins/mac-capture/mac-audio.m
```diff
@@ -637,6 +645,214 @@ static bool coreaudio_init(struct coreaudio_data *ca)
...
+	removeCmd = [NSString
+		stringWithFormat:@"rm -rf '%@' && ", halPluginFileName];
+	if (reinstall) {
+		NSString *halPluginSourcePath = [[NSBundle mainBundle]
+			pathForResource:@"OBSDesktopAudioCapture.driver"
+				 ofType:nil];
+		NSString *halPluginDestinationPath =
+			@"/Library/Audio/Plug-Ins/HAL/";
+		installCmd =
+			[NSString stringWithFormat:@"cp -R '%@' '%@' && ",
```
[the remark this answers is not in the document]
PatTheMav: @gxalpha have you solved this a different way?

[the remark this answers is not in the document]
gxalpha: Sorry 'bout that, I missed that comment before I pushed two days ago and only noticed afterwards (which is why I left it makes as not resolved), but then forgot about it.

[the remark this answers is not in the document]
gxalpha: Fixed (essentially copied from the virtualcam). Tested that the folder gets created if it doesn't exist.
Also fixed the logging @asher-the-thrasher had concerns about, it should now be `[mac-capture]:` everywhere.

gxalpha:
At the time when this PR was written, it was not foreseeable that Apple would add native audio capture to macOS. 3 months later they did, as part of ScreenCaptureKit on macOS Ventura. As that functionality is now integrated in OBS and Ventura is out of Beta, it would effectively only benefit old versions of macOS.
This would not be an issue in general. However in this case, we're talking about *a lot* of code, with changes not even exclusive to macOS. I fear that this could become a big maintainability cost for relatively little rewards.

It's frustrating since I spent quite a bit of time making this PR, but I've come to the conclusion that we should not merge it for the reasons laid out above. As such, I'm closing this PR. Still though, big thanks to everyone here who tested it.

PR #6325, opened 2022-04-13
libobs,UI: Issue appropriate signals on group / ungroup

Elgato-AStory:
### Description
I identified a deficiency in the API wherein using group and ungroup actions would not issue the expected `item_add` and `item_remove` signals. I've put together a basic implementation that meets my expectations for the API behavior in that regard.

The `item_add` signal is issued after all of the scene items are re-parented to the new group, and the `item_remove` signal is issued right before the group is modified. This allows the API consumer to more easily inspect the group's contents in a meaningful manner.

Unfortunately a side effect I noticed was that the UI's logic for inserting the group row in the sources list had a conflict with the newly issued signal, as the `item_add` signal was introducing a row for the new group automatically. I just went with the least intrusive option, which was to rely on that implicit behavior and get rid of the explicitly added row in the grouping logic. An alternative would be to suppress the refresh triggered by the `item_add` signal, but I feel like that would probably be an even uglier solution, albeit more explicit.

### Motivation and Context
A plugin that I'm working on needs precise and complete knowledge of the scene structure in as close to real time as possible. Without properly working `item_add` / `item_remove` signals, this is a much more expensive task where groups are involved, as presently the only way to get the information reliably would be to poll the entire scene structure continuously. 

### How Has This Been Tested?
Manual testing on Windows 10 with a stub listener plugin that echos signals. I simply just grouped and ungrouped scene items in the UI.

### Types of changes

   - It's likely to break some small number of integrations with OBS unfortunately.

### Checklist:

### DIFF
--- a/UI/source-tree.cpp
+++ b/UI/source-tree.cpp
@@ -967,28 +967,17 @@ void SourceTreeModel::GroupSelectedItems(QModelIndexList &indices)
 	for (obs_sceneitem_t *item : item_order)
 		obs_sceneitem_select(item, false);
 
-	int newIdx = indices[0].row();
-
-	beginInsertRows(QModelIndex(), newIdx, newIdx);
-	items.insert(newIdx, item);
-	endInsertRows();
-
-	for (int i = 0; i < indices.size(); i++) {
-		int fromIdx = indices[i].row() + 1;
-		int toIdx = newIdx + i + 1;
-		if (fromIdx != toIdx) {
-			beginMoveRows(QModelIndex(), fromIdx, fromIdx,
-				      QModelIndex(), toIdx);
-			MoveItem(items, fromIdx, toIdx);
-			endMoveRows();
-		}
-	}
-
 	hasGroups = true;
 	st->UpdateWidgets(true);
 
 	obs_sceneitem_select(item, true);
 
+	/* ----------------------------------------------------------------- */
+	/* obs_scene_insert_group triggers a full refresh of scene items via */
+	/* the item_add signal. No need to insert a row, just edit the one   */
+	/* that's created automatically.                                     */
+
+	int newIdx = indices[0].row();
 	QMetaObject::invokeMethod(st, "NewGroupEdit", Qt::QueuedConnection,
 				  Q_ARG(int, newIdx));
 }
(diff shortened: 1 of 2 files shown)


RytoEX:
> It's likely to break some small number of integrations with OBS unfortunately.

Pinging @tt2468 and @WarmUpTill. Are you aware of anything that would break in your plugins because of this?

tt2468:
While this appears to fix the group/ungroup functionality, there are still some unresolved issues which are related to this one:

(To pretext, my use of "parent scene" refers to the scene which items are currently being displayed in the source tree in UI.)

- This PR seems to miss some functionality specifically relating to the `Group Selected Items` / `Ungroup` features:
  - Grouping With PR:
    - `scene_create` for the group
    - `item_add` for the group being added to the parent scene
  - Expected grouping behavior:
    - `scene_create` for the group
    - `item_add` for the group being added to the parent scene
    - `item_add` for all items in the group
    - `item_remove` for all items in parent scene
- Dragging a scene item between a group and a parent scene does not fire `item_add` or `item_remove` signals on either the group or the scene
  - Expected behavior:
    - `item_add` signal for the item in the group
    - `item_remove` signal for the item in the parent scene (the old location)
- Reordering scene items in a group using the UI fires the `reorder` signal on the parent scene instead of the group itself. While the reason is likely to trigger a full scene tree refresh, this implementation is flawed since it is a libobs-level signal and not internal to the UI.
- (unconfirmed) Dragging a scene item from a group to the currently selected scene does not recalculate the item's ID, leading to the possibility of multiple items with the same ID being in the same scene, causing undefined behavior.

If it wasn't clear, groups are a can of worms. I feel that while this PR is not directly flawed, more needs to be done to properly fix the issues with group signals. If you happen to be in our Discord server, I would greatly appreciate if you have time to have a more extensive discussion about what we can do to fix this functionality.

WarmUpTill:
> > It's likely to break some small number of integrations with OBS unfortunately.
> 
> Pinging @tt2468 and @WarmUpTill. Are you aware of anything that would break in your plugins because of this?

I don't think the advanced scene switcher would be affected by this change.
But thank you very much for the heads up!

PR #6326, opened 2022-04-13
UI: Add shortcut for larger movement steps in preview

gxalpha:
### Description

Adds the ability to move a selected item 10 pixels in the preview when
holding shift while using the arrow keys, instead of going just 1 pixel
per step.
Implementing this was really simple since the system is already made with
distances other than 1 in mind, it just wasn't used anywhere.

https://user-images.githubusercontent.com/59806498/163171601-fffb4924-fe3a-4aa8-97ab-bfdba5905b1f.mov

### Motivation and Context

Moving a sceneitem a medium to large distance in a straight line currently
requires opening the Edit Transform dialog, manually entering a number, closing
the dialog again and then potentially doing fine adjustments in the preview.
While this works, it's an extra step that could be avoided.

Apparently the Shift modifier is used in Photoshop as well, so I assume
it's a rather common shortcut.

### How Has This Been Tested?

macOS 12.4 Beta 1

### Types of changes

### Checklist:

### DIFF
--- a/UI/window-basic-main.cpp
+++ b/UI/window-basic-main.cpp
@@ -391,6 +391,10 @@ OBSBasic::OBSBasic(QWidget *parent)
 	addNudge(Qt::Key_Down, SLOT(NudgeDown()));
 	addNudge(Qt::Key_Left, SLOT(NudgeLeft()));
 	addNudge(Qt::Key_Right, SLOT(NudgeRight()));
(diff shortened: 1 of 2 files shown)


Warchamp7:
Does this only work in the preview, or does it affect things like the Edit Transform/Filters input boxes as well?

Either way I'm good with merging this.

gxalpha:
This only affects the preview

PR #6335, opened 2022-04-14
UI: Unify function definitions in properties-view

gxalpha:
### Description

The AddProperty function was a mess, partly due to the Add# functions
having very different headers. That would lead to redundant code and
having quite a few cases where then in AddProperty things would get
replaced again.
Now, all of them return void, and none of them add the entry to the
QFormLayout themselves. Instead, every function gets the property, and
either a QLayout or QWidget (or both) which it can modify.
This cuts down on code, and makes it more readable.

### Motivation and Context

The `AddProperty` function was a mess, with a lot of special cases for
different property types, like
```
if (!label && type != OBS_PROPERTY_BOOL &&
	    type != OBS_PROPERTY_BUTTON && type != OBS_PROPERTY_GROUP)
		label = new QLabel(QT_UTF8(obs_property_description(property)));
```
(or more examples), where this configuration could be done universally or in
the AddCheckbox/AddButton function.
The only distinction left are groups, since they need the entire width of the
QFormLayout for the widget.

It's so much cleaner this way.

### How Has This Been Tested?

The following widget types have been explicitly tested to work, both normal and with a `obs_property_set_enabled(property, false)` in the beginning to make sure they are
correctly disabled:
- Bool
- Int
- Text
    - Multiline
    - Password
    - Normal
- Path
- List
    - Normal
    - Editable
- Color (w/ alpha)
- Font
- Button
- Editable List (e.g. slideshow)
- Frame Rate
- Group

Ran static analysis to make sure no memory is being leaked.

### Types of changes

### Checklist:

### DIFF
--- a/UI/properties-view.cpp
+++ b/UI/properties-view.cpp
@@ -248,19 +248,46 @@ QWidget *OBSPropertiesView::NewWidget(obs_property_t *prop, QWidget *widget,
 	return widget;
 }
 
-QWidget *OBSPropertiesView::AddCheckbox(obs_property_t *prop)
+void OBSPropertiesView::AddCheckbox(obs_property_t *prop, QLabel *&label,
+				    QWidget *&widget)
 {
+	delete label;
+	label = nullptr;
 	const char *name = obs_property_name(prop);
 	const char *desc = obs_property_description(prop);
 	bool val = obs_data_get_bool(settings, name);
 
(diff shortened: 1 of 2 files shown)


gxalpha:
Rebasing this is impossible. I might decide to completely redo this PR in the future (or to just not care).

Issue #6145, opened 2022-03-15
Delete the scene, the video source is not completely deleted


zengsansui:
### Operating System Info

Windows 11

### OBS Studio Version

27.2.3

### OBS Studio Log URL

https://obsproject.com/logs/Mhe9F38qMmxQDGHm

### Expected Behavior

When deleting a scene, the sources in the scene should all be deleted.

### Current Behavior

1. Set show stinger transition for video/music in scene1.
2.  Click show/hide video/music
3.  Delete scene1
4.  Add video/music in scene2，video/music stay in existing. Add it.
5.  Delete video/music， but video/music stay in audio mixer, and the sound keeps coming out.

### Steps to Reproduce

1.
2.
3.
...

WizardCM:
You have the following third party plugins loaded. Please remove them first and test again:
* StreamFX
* obs-virtualbg

zengsansui:
The issue still exists after removing the third-party plugin.
LOG URL : https://obsproject.com/logs/LY9n4EqkUmeBtfDu

sora-blue:
I reproduced the problem on the lastest version 27.2.4 and the version built from pulled repo on Windows 10. And if you add the source back, you will find its transtition settings gone. So i suppose when the scene is deleted, the transition video source is deleted but the original source is falsely ignored.

jpc0:
I am finding it quite difficult to reproduce this on the current release/27.2 branch. Can you try without any sources loaded or alternatively give more clear instructions to reproduce?

sora-blue:
@jpc0 i reprpduced it on the version built from main branch & 27.4 release by following these steps:
 1. create a media source at current scene
 2. add a stinger transition for it 
 3. double click hide button at sceneitem list to make transition start to play 
 4. create a new scene and delete previous scene 
 Then you should be able to see media source left at Mixer dock.

jpc0:
@sora-blue Just to confirm, should step 4 occur while the stinger transition is still running? The more specific the steps the more likely to get exactly what causes the issue.

jpc0:
Can confirm, I recreates this with any source show transition, if the scene is deleted while a source transition is still running the source is not correctly cleaned up. 

Some more observed behaviour, if the source is readded from the add existing menu the trabsition will no longer fire and the source will not autoplay but will play when using the media controls, it will also lose any information relating to the transition.


Steps:

- Create a source that has an audio element
- Add a show transition with a long duration, I used 10 seconds
[...]

PR #6337, opened 2022-04-15
libobs: Extra callback when removing scene

sora-blue:
### Description

This small patch ensures media source to be removed when being removed with scene.

### Motivation and Context

This PR is to fix [issue #6145 ](https://github.com/obsproject/obs-studio/issues/6145).

### How Has This Been Tested?

The effect is tested on my computer running Windows 10 21H2 and it solves the problem.
But i'm not sure if extra codes are needed to prevent affecting other functions.

### Types of changes

Bug fix

### Checklist:

### DIFF
--- a/libobs/obs-scene.c
+++ b/libobs/obs-scene.c
@@ -277,7 +277,7 @@ static void scene_enum_sources(void *data, obs_source_enum_proc_t enum_callback,
 				 transition_active(item->hide_transition))
 				enum_callback(scene->source,
 					      item->hide_transition, param);
-			else if (os_atomic_load_long(&item->active_refs) > 0)
+			if (os_atomic_load_long(&item->active_refs) > 0)
 				enum_callback(scene->source, item->source,
 					      param);
 		} else {


PatTheMav:
As mentioned in https://github.com/obsproject/obs-studio/issues/6145 I'm not able to reproduce the issue, but maybe I'm mistaken and the issue is still present?

Please confirm or otherwise close the PR as it might not be needed anymore.

sora-blue:
Hi there. I tested it with obs 30.0.0 on win11x64, and it seems still being a problem. But since it had been a year and half ago, the PR may need further debug.
image

PatTheMav:
> Hi there. I tested it with obs 30.0.0 on win11x64, and it seems still being a problem. But since it had been a year and half ago, the PR may need further debug. image

What are the exact reproduction steps?

I tried adding the source with a show transition set to 5 seconds, then deleted the source while the transition was running. Both sources (media source and transition) were correctly removed then.

derrod (approved):
Commit message is maybe not the best, but I believe this change is correct as this callback should also enumerate the underlying source if it is active.

PatTheMav (approved):
The underlying issue is that in the case of a scene destruction some part of the program holds a reference to the media source (could be the UI thread, the source toolbar, or something else) which prevents it from being scheduled for destruction in the helper thread.

Ensuring that the enumeration callback runs even during a show transition seems to also lead to these references being released.

sora-blue:
> What are the exact reproduction steps?
> 
> I tried adding the source with a show transition set to 5 seconds, then deleted the source while the transition was running. Both sources (media source and transition) were correctly removed then.

Here's a recorded reproduction process for it.
[](https://freeimage.host/i/JISE2Ub)

PR #6338, opened 2022-04-15
libobs, UI: Default mixer volume meter to two channels

gxalpha:
### Description

Audio sources that have not yet output (or whatever) any audio currently
do not know which speaker configuration and how many channels the audio
will have, so the audio mixer defaults to 1.
This changes this to 2 since the majority of sources are stereo.
Browser source that hasn't put out any audio yet:
image

This changes how `obs_volmeter_get_nr_channels` works (it will now default
to 0 if the number is unknown instead of 1), technically making it a breaking
change. However, that function is neither documented in the docs (like the
rest of `obs-audio-controls.h`, I'll maybe write that in the future, but don't
really have the time right now) nor used anywhere else in OBS; so I'd personally
assume plugins don't use it and find it safe enough to make this change.

### Motivation and Context

From Warchamps list

### How Has This Been Tested?

macOS 12.3, compiled and run
Verified it now defaults to 2 channels if sources have yet to output audio

### Types of changes

### Checklist:

### DIFF
--- a/UI/volume-control.cpp
+++ b/UI/volume-control.cpp
@@ -750,6 +750,14 @@ inline void VolumeMeter::resetLevels()
 bool VolumeMeter::needLayoutChange()
 {
 	int currentNrAudioChannels = obs_volmeter_get_nr_channels(obs_volmeter);
+
+	if (!currentNrAudioChannels) {
+		struct obs_audio_info oai;
+		obs_get_audio_info(&oai);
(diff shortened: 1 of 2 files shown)


WizardCM:
What happens if the user has OBS itself set to mono output?

gxalpha:
Fixed, channels will still default to 1 now should OBS be set to mono, I somehow didn't realize that was a thing :D.

PR #6351, opened 2022-04-18
libobs: Implement asynchronous sample rate conversion for audio

### LINKED ISSUE #4600: Random short audio dropouts when recording or streaming

norihiro:
### Description

This PR implements asynchronous sample-rate conversion by utilizing a soft compensation feature in libswresample. A digital approach of asynchronous sample-rate conversion was proposed in [1][1], a PDF is available [here](https://www.eecg.utoronto.ca/~johns/ece1371/papers/1994_adams.pdf). Also described in [2][2]. The soft compensation in libswresample uses similar technique as [1][1]. In addition, a traditional lag-lead filter is implemented to eliminate errors at steady state.

As described in the description of the PR #6133, `next_audio_ts_min` should be same as `data->timestamp + conv_frames_to_time(sample_rate, in.frames)`. However, just overwriting `next_audio_ts_min` cannot compensate insufficient or too many number of samples. Instead, the difference between these values will be fed to the soft compensation.

A lag-lead filter is widely used in a conventional PLL design. In the feedback loop, the lag-lead filter is employeed for these purposes.
- to eliminate errors at the steady state
- to suppress jitter induced by fluctuations such as CPU workload and USB transfer.

The unity gain frequency of the feedback loop is set to 1/180 Hz so that the fluctuations faster than 3 minutes will be suppressed
and I believe users won't notice transition at the startup.

This PR also changes built-in plugins to enable the feature by default.
- [x] linux-pulseaudio
- [x] decklink (implemented but not tested)
- [x] aja
- [x] mac-capture (implemented but not tested)
- [x] win-wasapi
- [ ] win-dshow
- [x] linux-alsa
- [ ] linux-jack
- [ ] oss-audio
- [ ] sndio
- ~~obs-ffmpeg-source and vlc-video (for realtime streaming)~~ It's not easy to implement since the flow is controlled in the thread `mp_media_thread`.

Note: Though the implementation looks working, I'd like to clean the code to make it more readable.

[1]: https://ieeexplore.ieee.org/document/280698
[2]: https://www.analog.com/media/en/technical-documentation/application-notes/ee268v01.pdf

### Motivation and Context

Since most of audio sources are asynchronous,
the number of samples provided from the audio sources is sometimes insufficient or too many,
which causes audio glitch as #4600 and audio/video desync.

Fix #4600.

### How Has This Been Tested?

The code is tested on Fedora 34 with MOTU M2 and recording a tone signal.

Also analyzed step response and frequency response of the feedback system with [my SPICE model](https://github.com/norihiro/obs-async-audio-filter/tree/main/model).

obs-async-audio-filter-ac
Frequency response of the open loop transfer function. Gain curve crosses 0 dB at 5.55 mHz (inverse of 3 minutes) and shows suppression for higher frequency components to eliminate fluctuation such as USB and load of the host. At 5.55 mHz, the phase curve shows 60 degree phase margin so that the feedback system is stable.

async-sample-rate-conversion-step-response
Simulated and measured step response of the system. This figure shows step response how the compensation is applied (vc1) and another internal node in the lag-lead filter (vc2), measured with MOTU M2 with Fedora 34. It estimated roughly 4.5 samples are are missing for every second. The implementation gradually started to add samples and reached maximum compensation -5.3 samples per second after 80 seconds from the start, then goes to the steady state compensating 4.5 samples per second. The other node vc2 slowly track vc1 so that it eliminates offset at the steady state. The curves of measured and simulated model are well matched except between 100 seconds to 300 seconds. I think we can rely on the model so that the gain and phase curves are well simulated and conclude the feedback system has enough stability.

### Types of changes

### Checklist:


digitalchou:
Hi, if you need a macos intel tester Ill be glad!! Figthing with this bug till i bought a new Maudio Mtrack DUO 48kHz soundcard, but it sound a little worse than mi UMC404-192kHz one(wich dont exactly match the clock in 48kHz mode and drops a few ms of audio every 2:20 min aprox)
Thanks for your work, your source color minitor is one of the best obs plugins ever!!

norihiro:
> Hi, if you need a macos intel tester Ill be glad!!

I appreciate your comment. You can download a package for macOS by [this instruction](https://github.com/obsproject/obs-studio/wiki/Download-Multiplatform-Build). Since the code is not signed and notarized, you might need to adjust security settings on your macOS.
With this package, you will see an option `Enable Asynchronous Compensation` for `Audio Input Capture`.

tt2468:
Other than for testing purposes, is there any reason to want to have async compensation be a configurable value in sources? To me it seems like it should always be on unless there is a good reason to want it off.

norihiro:
> Other than for testing purposes, is there any reason to want to have async compensation be a configurable value in sources? To me it seems like it should always be on unless there is a good reason to want it off.

It's a good point.
- Iwant to provide the same behavior as old versions just in case it works wrong.
- If async-SRC is enabled, a low-pass filter is conformed to cut off very high frequency. Someone might want to avoid the low-pass filter instead of avoiding glitches.
- Some audio interfaces support genlock. If a user somehow do genlock to OBS, async-SRC is unnecessary.

Anyway, libobs need to have the switch to have async-SRC. Obviously, media source and VLC source have to disable the async-SRC.

Btw, so far I have one thing I need to work on. When I turn off async-SRC, I guess the audio resampler is still running even though it's unnecessary. Other behavior looks fine for me but I don't know the owner of this project is happy to have this PR.

plugins/win-wasapi/win-wasapi.cpp
```diff
@@ -1506,12 +1514,33 @@ static obs_properties_t *GetWASAPIPropertiesInput(void *)
...
-				obs_module_text("UseDeviceTiming"));
+	obs_property_t *device_timing_prop =
+		obs_properties_add_bool(props, OPT_USE_DEVICE_TIMING,
+					obs_module_text("UseDeviceTiming"));
+
+	obs_properties_add_bool(props, OPT_ASYNC_COMPENSATION,
+				obs_module_text("AsyncCompensation"));
+
+	obs_property_set_modified_callback(device_timing_prop,
+					   device_timing_changed);
```
WizardCM, line 1525: 'device_timing_changed': undeclared identifier

norihiro:
I will fix the CI's failure (Windows' failure was solved by above force-pushed commit. Flatpak's failure might be the upstream's issue).
Also try to make `struct lag_lead_filter` a little more general.

plugins/win-wasapi/win-wasapi.cpp
norihiro: The compile error is resolved.

norihiro:
Rebased to the latest master.
So far, the behavior is ok.
I'm afraid the interface of a newly added lag-lead filter in libobs/media-io/lag_lead_filter.h is not so well organized. Magic numbers at [lag_lead_filter_set_parameters](https://github.com/obsproject/obs-studio/blob/7f85ebb81f302526444bd8cd5f65d01c8781467a/libobs/media-io/audio-resampler-ffmpeg.c#L221-L222) were sought experimentally.

gxalpha:
I'd like to see https://github.com/obsproject/obs-studio/pull/6351#issuecomment-1166478853 be discussed in a UI/UX context. This option seems arcane from a users perspective and most won't know what to do with it.
If it's not buggy it can probably be reverted in a hotfix release (instead of having a toggle for the individual user to disable), so besides the latter two points brought up in the linked comment, are there other disadvantages of this feature being enabled?

CC-ing @Warchamp7 @GeorgesStavracas so that it hopefully doesn't get forgotten.

AlexGxtr:
Is that fixing the audio dropouts described here? https://github.com/obsproject/obs-studio/issues/4600

SuslikV:
Thread from the OBS forum (USB device in use) mentioned above : https://obsproject.com/forum/threads/audio-dropouts-each-7-8-minute.168856/

SuslikV:
By some reason, I got crash with the build "obs-studio-win-x64-6dffb6e9a.zip", OBS 29.1.0-474-g6dffb6e9a (64-bit, windows) in **VirtualBox** environment:
```
Unhandled exception: c0000005
Date/Time: 2023-08-18, 09:30:13
Fault address: 7FF7487B0C30 (c:\obs6\bin\64bit\obs64.exe)
libobs version: 29.1.0 (64-bit)
Windows version: 10.0 build 19042 (release: 20H2; revision: 985; 64-bit)
CPU: AMD Athlon(tm) II X2 250 Processor


Thread 10A4: (Crashed)
Stack            EIP              Arg0             Arg1             Arg2             Arg3             Address
0000006B19EF6FC0 00007FF7487B0C30 0000022DC6802EC0 0000000000000000 0000006B19EF9F50 0000000000000000 obs64.exe!SelectFormat+0x130
0000006B19EF70B0 00007FF74879FF29 0000006B19EF9F50 0000006B19EF7200 0000006B19EF9F00 0000022DC0D33AE0 obs64.exe!OBSBasicSettings::LoadAdvOutputFFmpegSettings+0x419
0000006B19EF7180 00007FF7487A73A4 FFFFFFFFFFFFFFFF 0000006B19EF9F50 0000022DC8241510 0000000000000000 obs64.exe!OBSBasicSettings::LoadOutputSettings+0x114
0000006B19EF71B0 00007FF7487A7D65 0000000000000002 0000000000000002 0000022DC8241510 0000006B19EF9A78 obs64.exe!OBSBasicSettings::LoadSettings+0x45
0000006B19EF71E0 00007FF7487951A4 0000006B19EF9F50 0000006B19EF9FB0 00000000000000C2 0000022DBE3D1E20 obs64.exe!OBSBasicSettings::OBSBasicSettings+0x6684
0000006B19EF9EB0 00007FF748752B02 0000000000000000 00000000000000C2 0000006B19EFA3A0 0000006B19EFA5F0 obs64.exe!OBSBasic::on_action_Settings_triggered+0xc2
0000006B19EFA320 00007FF74889A520 00000000000000C2 0000006B19EFA5F0 0000000000000000 0000006B19EFA460 obs64.exe!OBSBasic::qt_static_metacall+0xea0
0000006B19EFA3D0 00007FF7488A05D2 03E88CB3C9484E2B 0000006B19EFA540 0000022DC0D01590 00007FFD260782AF obs64.exe!OBSBasic::qt_metacall+0x42
0000006B19EFA400 00007FFD25B2BAB1 0000022DC0CAA6C0 00007FFD26062491 0000006B19EFA511 00007FFD26073871 qt6core.dll!0x7ffd25b2bab1
... and so on
```
Whole _config_ folder of OBS after the crash (in .zip archive, about 26.8 KB to download, MD5:8F1251D0945561C5B9A1500C64C24413):
[obs-config-folder--obs-studio-win-x64-6dffb6e9a.zip](https://github.com/obsproject/obs-studio/files/12376493/obs-config-folder--obs-studio-win-x64-6dffb6e9a.zip)

The crash happens only on new profile if you attempt to click _Setting_ button of OBS (when the profile is almost empty). If you import existing profile (filled with the settings) all is OK.

**Notes.**
After two sequential crashes OBS shows "Safe mode" dialog, so at least something works:) (not helps, of course)
Haven't tested other builds, yet.

norihiro:
Rebased to the latest master branch. I hope the crash was a common issue in the previous base because this PR does not touch profiles and settings.

SuslikV:
Maybe stupid question, but the _Application Audio Capture (Beta)_ asynchronous source or not?

norihiro:
> _Application Audio Capture (Beta)_ asynchronous source or not

Yes, it's also an asynchronous source. I added the property.

SuslikV:
1. The build "obs-studio-windows-x64-fcf36ed56.zip", OBS 30.0.0-beta1-25-gfcf36ed56 (64-bit, windows): no longer crashed (was mentioned above, in https://github.com/obsproject/obs-studio/pull/6351#issuecomment-1683478193). All OK now.

2. The _Application Audio Capture (Beta)_ inside **VirtualBox** **(!)** environment - I see no difference either checked "Enable Asynchronous Compensation" or not in the properties of the source, probably because of https://github.com/obsproject/obs-studio/issues/8064 . Recorded files are distorted, dropouts and corruption that heard as "crackling" present. My sources were running at different sample rates (if it makes any sense here):
```
09:24:39.248: WASAPI: Device '[VIRTUAL_AUDIO_DEVICE_PROCESS_LOOPBACK]' [48000 Hz] initialized
09:24:39.664: WASAPI: Device 'Speakers (High Definition Audio Device)' [44100 Hz] initialized
```

----------------

Anyway, state of the "Enable Asynchronous Compensation" is not logged unlike "use device timing" in OBS log-file. So, until the async setting is optional, I think, it is better to log state of the function too.

What other disadvantages of usage of "RadioButton of 3 states" or "ComboBox of 3 options" (Timings: Device timings>OBS timings>OBS timings+Asynchronous Compensation) instead of 2 CheckBoxes? Only disadvantage I know - compatibility reasons with previous versions of OBS. But maybe there is more?

For example, for me it is not clear, when I see 2 CheckBoxes:

- [x] "Use device timings" checked;
- [x] ~~"Enable Asynchronous Compensation" checked **but disabled**~~

is the async algorithm enabled or not in this case? (log has no info about it)

norihiro:
You can find a log like below after the shutdown of OBS or disabling an audio source.
> info: pulse-input: Stopped recording from 'alsa_input.usb-MOTU_M2_M2AE0698VI-00.pro-input-0'
> info: pulse-input: Got 2543 packets with 3051600 frames
> info: audio_resampler: input 3026400 samples, output 3027135 samples, estimated input frequency 47988.075921 Hz

> is the async algorithm enabled or not in this case? (log has no info about it)
No, the async-SRc is not enabled when the checkbox is disabled.

AlexGxtr:
> This recent PR fixed my problems with OBS (audio stuttering each 5-8 minutes). However, I will perform more tests just to confirm it works fine on windows 10 64bit. Thanks for the help and for dedicating time to solving OBS itself.

Just to confirm this works on my setup. Performed 60-70 minutes test a few days ago.

AlexGxtr:
I've noticed one single dropout during 2 hrs dj set, not sure what caused it.

norihiro:
> I've noticed one single dropout during 2 hrs dj set, not sure what caused it.

Could you check the log file when the dropout happened?
Not always but sometimes the reason for the dropout will be logged.

AlexGxtr:
> > I've noticed one single dropout during 2 hrs dj set, not sure what caused it.
> 
> Could you check the log file when the dropout happened? Not always but sometimes the reason for the dropout will be logged.

here's the log file: 

The drop happened at the 34:20 min - the second recording.
[2023-08-28 20-37-03.7z.txt](https://github.com/obsproject/obs-studio/files/12504627/2023-08-28.20-37-03.7z.txt)
PS: rename the .txt file to .7z file then uzip to see the exact txt file.

norihiro:
> The drop happened at the 34:20 min - the second recording. [2023-08-28 20-37-03.7z.txt](https://github.com/obsproject/obs-studio/files/12504627/2023-08-28.20-37-03.7z.txt) PS: rename the .txt file to .7z file then uzip to see the exact txt file. and thanks for the help :)

I'm sorry I cannot understand where is "34:20 min". Your log file started at 20:37:03 and ended at 02:01:30, it only contains 324 minutes.
One of your win-dshow capture devices ('LiveCam2GoPro', 'LiveCam2GoPro', or 'LiveCam2GoPro') emit too many error messages, 1 set of 3 lines for every 0.033 ms. Probably, the device sends 1 empty video frame every 2 frames, which means the device is claimed to be capable of 60 FPS but actually only 30 FPS. I recommend setting the capturing frame rate to 30 FPS or buying a correct capture device.
It would be possible that such a huge log could make your system slow. I'd like to suggest fixing the video capture device issue first.

AlexGxtr:
> > The drop happened at the 34:20 min - the second recording. [2023-08-28 20-37-03.7z.txt](https://github.com/obsproject/obs-studio/files/12504627/2023-08-28.20-37-03.7z.txt) PS: rename the .txt file to .7z file then uzip to see the exact txt file. and thanks for the help :)
> 
> I'm sorry I cannot understand where is "34:20 min". Your log file started at 20:37:03 and ended at 02:01:30, it only contains 324 minutes. One of your win-dshow capture devices ('LiveCam2GoPro', 'LiveCam2GoPro', or 'LiveCam2GoPro') emit too many error messages, 1 set of 3 lines for every 0.033 ms. Probably, the device sends 1 empty video frame every 2 frames, which means the device is claimed to be capable of 60 FPS but actually only 30 FPS. I recommend setting the capturing frame rate to 30 FPS or buying a correct capture device. It would be possible that such a huge log could make your system slow. I'd like to suggest fixing the video capture device issue first.

Hi, yes this is my GoPro 7 hero black camera. It's capable of 60fps but the goPro software itself is janky it's probably a bug. I'm using a 3m HDMI cable too.

zbynekdrlik:
When we could test it?

norihiro:
Rebased and the build flow has passed. You can test it from [the summary page](https://github.com/obsproject/obs-studio/actions/runs/11219435426?pr=6351).

moonlightes:
Hi. Is this change missing in the current version of obs 31? If so, is it expected to be added soon?

Fenrirthviti:
This has not been merged yet, so it is not missing from any release.

We prefer not to provide estimations on PR reviews as the project priorities are constantly in flux.

libobs/media-io/audio-resampler-ffmpeg.c
```diff
@@ -169,6 +178,15 @@ audio_resampler_t *audio_resampler_create(const struct resample_info *dst, const
 void audio_resampler_destroy(audio_resampler_t *rs)
 {
 	if (rs) {
+		uint64_t total_output_samples = rs->total_output_samples + swr_get_delay(rs->context, rs->output_freq);
+		if (rs->compensation_filter_configured)
+			blog(LOG_INFO,
+			     "audio_resampler: input %" PRIu64 " samples, "
```
pkviet: I would add some info that compensation is triggered, 
ex: audio_resampler (frequency compensation): input etc ...

libobs/media-io/lag_lead_filter.h
```diff
@@ -0,0 +1,63 @@
+/*
+ *  Lag lead filter for asynchronous sample rate converter
+ *  Copyright (C) 2022 Norihiro Kamae <norihiro@nagater.net>
```
pkviet: update to 2022-2026

libobs/obs-source.c
```diff
@@ -1491,6 +1491,16 @@ static void source_output_audio_data(obs_source_t *source, const struct audio_da
 		}
 	}
 
+	if (source->async_compensation && source->resampler) {
```
pkviet: is the bool async_compensation thread safe ? it is set in UI thread by a source  but consumed in audio thread.

plugins/linux-alsa/alsa-input.c
```diff
@@ -140,6 +141,10 @@ void *alsa_create(obs_data_t *settings, obs_source_t *source)
 #if !SHUTDOWN_ON_DEACTIVATE
 	_alsa_try_open(data);
 #endif
+
+	async_compensation = obs_data_get_bool(settings, "async_compensation");
+	obs_source_set_async_compensation(data->source, async_compensation);
```
pkviet, line 146: in theory this is a no-op, if async_compensation value has not changed. So it's OK.
But in general, we store source settings as members of the base data struct, and update it only if its value has changed, so this saves a call. I will readily agree that here it does not make much difference though.
(same remark for other sources where you followed the same pattern)

libobs/media-io/audio-resampler-ffmpeg.c
```diff
@@ -187,6 +226,17 @@ bool audio_resampler_resample(audio_resampler_t *rs, uint8_t *output[], uint32_t
 	struct SwrContext *context = rs->context;
 	int ret;
 
+	if (rs->compensation_filter_configured) {
+		lag_lead_filter_tick(&rs->compensation_filter, rs->input_freq, in_frames);
+		double drift = lag_lead_filter_get_drift(&rs->compensation_filter);
+
+		ret = swr_set_compensation(rs->context, -(int)(drift * 65536 / rs->input_freq), 65536);
```
pkviet, line 232: where does that figure come from ?

libobs/media-io/audio-resampler-ffmpeg.c
```diff
@@ -178,6 +196,27 @@ void audio_resampler_destroy(audio_resampler_t *rs)
 	}
 }
 
+void audio_resampler_set_compensation_error(audio_resampler_t *rs, int error_ns)
+{
+	if (!rs->compensation_filter_configured) {
+		// Parameters are determined experimentally by SPICE simulation
+		// to have these characteristics.
+		// - Unity gain frequency: 1/180 Hz (inverse of 3 minutes)
```
pkviet, line 203: did you experiment with shorter and longer times ? say 1 min so 1/60 Hz and 5 min so 1/300 Hz ?
Regarding this curve:
https://user-images.githubusercontent.com/780600/180902267-e95694da-7aed-46c2-839b-36c5fe097f15.png

how is it modified ?
Is the sample/sec negative compensation due to input lag ? have you tested the reverse condition (through a simulation for instance)
More generally, can you explain why the filter converges to a plateau ? it seems to mean that the input source is intrinsically lagging if the filter must always compensate ?

pkviet (approved):
The code is very clean and well written.
I appreciated the explanations with reference to the research behind the compensation filter.
As discussed in another PR, FFmpeg has the aresample filter which has both hard and soft compensation. Can you summarize the differences and see if this can suggest improvements to your current implementation ?
Also could this filter help solve the desync issues with all audio outputs (chiefly, decklink and ndi)
(currently there is a low latency audio buffering option which sets oai->max_buffering_ms to 20 ms and triggers : 
https://github.com/obsproject/obs-studio/blob/master/libobs/obs.c#L1612
and following)
Sidenote: please rebase the PR

norihiro:
Rebased to resolve conflicts.

libobs/media-io/audio-resampler-ffmpeg.c
norihiro: Replaced `audio_resampler:` with `audio_resampler (asynchronous compensation):`.

libobs/obs-source.c
norihiro: Thank you. Updated to use `os_atomic_store_bool` and `os_atomic_load_bool`.

Issue #5905, opened 2022-02-02
UI : Audio Mixer show the scrollbar even when not necessary


Trouffman:
### Operating System Info

Windows 11

### OBS Studio Version

27.1.3

### OBS Studio Version (Other)

portable

### OBS Studio Log URL

https://obsproject.com/logs/uhXtMHTXT7nqiE2b

### Expected Behavior

The Audio Mixer Dock show not show the scrollbar when not required.

For example : when only 2 audio input are displayed, the scroll bar is usually not needed (unless you reduce the size of the dock.

![image](https://user-images.githubusercontent.com/2151317/152230974-ea817dcb-ed5d-4ec0-8339-359dbde976aa.png)

### Current Behavior

The scrollbar on the right side of the Audio Mixer dock is visible at all time even whne not required.

![image](https://user-images.githubusercontent.com/2151317/152230850-5c788ac3-7a57-4ab0-9b7a-195dcf280510.png)

Btw this is the same on OBS 27.2 RC1
![image](https://user-images.githubusercontent.com/2151317/152231314-82fc4cac-35dc-4890-a9cc-0524a782acc3.png)

### Steps to Reproduce

1. Open OBS
2. Look at the Audio Mixer

### Anything else we should know?

This is purely cosmetic.

Warchamp7:
This is intentional but I forget if it's to prevent controls jumping around, or if it was a performance thing because of the mixers render update shenanigans. Will chat with some folks on this

WizardCM:
For reference this was last reported in #5176.

The code for this is:

https://github.com/obsproject/obs-studio/blob/f52f34b614804b43486117aa2b06da0fdf3c6260/UI/forms/OBSBasic.ui#L1018-L1019

https://github.com/obsproject/obs-studio/blob/f52f34b614804b43486117aa2b06da0fdf3c6260/UI/forms/OBSBasic.ui#L1074-L1075

Trouffman:
Thanks @WizardCM 

Do you remember the reason behind it? (So this can be referenced if this comment pops-up again)

As far a I could go back I could not find more details.

I went all the way back to this back in the days:
https://github.com/obsproject/obs-studio/commit/52746c25230d4cd724246ea8a56800de675fc341?diff=split

Ricky-Tigg:
Far from cosmetic. In the depicted context its very presence does imply that there is more to be seen. This behaviour is eligible to be interpreted as an issue rather than a intentional.

[...]

PR #6352, opened 2022-04-18
UI: Only use volume scrollbars when needed

cg2121:
### Description
The scrollbars were always set to be on, no matter
how many volume controls that were being used.

### Motivation and Context
It was brought up before that the scrollbars were always on because of rendering problems, but after some testing, I found that everything still works as expected.

Fixes https://github.com/obsproject/obs-studio/issues/5905

### How Has This Been Tested?
Made sure scrollbars only showed up when needed.

### Types of changes

### Checklist:

### DIFF
--- a/UI/forms/OBSBasic.ui
+++ b/UI/forms/OBSBasic.ui
@@ -1016,7 +1016,7 @@
          <enum>QFrame::Sunken</enum>
         </property>
         <property name="verticalScrollBarPolicy">
-         <enum>Qt::ScrollBarAlwaysOn</enum>
+         <enum>Qt::ScrollBarAsNeeded</enum>
         </property>
         <property name="horizontalScrollBarPolicy">
          <enum>Qt::ScrollBarAlwaysOff</enum>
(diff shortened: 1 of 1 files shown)


RytoEX:
We'll need someone on macOS to verify that this does not cause issues. The only note I could find on this was that we _think_ this was done intentionally to prevent some issue on macOS, but there were no details as to what issues we were aware of. It's possible that it was done to avoid unnecessary repaints, which are particularly painful on performance on macOS, but I don't know for sure.

PatTheMav:
I tried to review this just now, but building on master I don't get any scrollbars to begin with. Is the observed behaviour tied to a specific macOS version?

PatTheMav:
@gxalpha Can you replicate the "bad" behaviour on master?

gxalpha:
yes. you need to have the mixer change once though, for example changing the volume of a source in the mixer.

PatTheMav:
Finally was able to replicate the behaviour and confirm the PR fixes it - the bug won't exhibit itself if macOS is set to automatic scrollbars (which is the default when using Magic Mouse or a trackpad).

PatTheMav (approved):

PR #6354, opened 2022-04-19
Add JSONSchema definitions for services & validate using CI

WizardCM:
### Description

This adds full JSONSchema definition files for both services.json and package.json, which can be validated using various tools, including an IDE. This includes user-friendly descriptions pulled from [Service Submission Guidelines](https://github.com/obsproject/obs-studio/wiki/Service-Submission-Guidelines).

image

image

image

image

image

Additionally, this adds a Python-based validation runner when users submit Pull Requests, informing them of any failing validations directly within the diff.

image

image

Unfortunately, I had to write a custom python script to perform complete validation, as none of the existing libraries provide line numbers by default - it's exclusively _paths_ within the JSON structure. Additionally, errors would be poorly formatted and generally hard to read. Instead I have to return a custom JSON result and write that to a file, to be read by an annotation GitHub Action.

By default, CI will only validate the services.json file. If needed, I can add the package.json to the CI run too.

Bonus: json_source_map was built incorrectly and only works on Python 3.9 and above, so I had to manually install that too.

If the services.json format is changed, the schema should be bumped too, which is why the file is v3.

### Motivation and Context

When new services are submitted, there isn't enough feedback to the submitter about what was done incorrectly.

### How Has This Been Tested?

Install Python 3.9 and the required dependencies.

```sh
sudo apt install python3.9-dev
python3.9 -m pip install jsonschema json_source_map
```

Then run the validation script.

```sh
python3.9 CI/check-jsonschema.py plugins/rtmp-services/data/services.json plugins/rtmp-services/data/schema/service-schema-v3.json
```

Note that it silently succeeds. Modify the services.json to be invalid, then run the script again.

Additionally, ran the CI script manually.

### Types of changes

### Checklist:

### DIFF
--- a/.github/workflows/services-json.yml
+++ b/.github/workflows/services-json.yml
@@ -1,0 +1,40 @@
+name: Services Validator
+
+on:
+  push:
+    paths:
+      - "plugins/rtmp-services/data/services.json"
+      - "plugins/rtmp-services/data/package.json"
+  pull_request:
+    paths:
+      - "plugins/rtmp-services/data/services.json"
+      - "plugins/rtmp-services/data/package.json"
+
+jobs:
+  schema:
+    name: Schema
+    runs-on: [ubuntu-latest]
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v3
+
+      - name: Install & Configure Python
+        run: |
+          sudo apt install python3.9-dev
+          python3.9 -m pip install jsonschema json_source_map
+
+      - name: Validate Service JSON Schema
+        run: |
+          JSON_FILES=(
+            plugins/rtmp-services/data/services.json
+            plugins/rtmp-services/data/package.json
+          )
+          python3.9 CI/check-jsonschema.py "${JSON_FILES[@]}"
+
+      - name: Annotate Errors
+        if: failure()
+        uses: yuzutech/annotations-action@v0.4.0
+        with:
+          repo-token: "${{ secrets.GITHUB_TOKEN }}"
+          title: "Service JSON Errors"
+          input: "./validation_errors.json"
--- a/CI/check-jsonschema.py
+++ b/CI/check-jsonschema.py
@@ -1,0 +1,84 @@
+import json
+from jsonschema import Draft7Validator
+from json_source_map import calculate
+from json_source_map.errors import InvalidInputError
+import os
+import sys
+
+
+errors = []
+
+
+def main():
+    if len(sys.argv) < 2:
+        print("JSON path required.")
+        return 1
+
+    for filename in sys.argv[1:]:
+        prep(filename)
+
+    try:
+        with open('validation_errors.json', 'w') as outfile:
+            json.dump(errors, outfile)
+    except OSError as e:
+        print(f'Failed to write validation output to file: {e}')
+        return 1
+
+    if errors:
+        return 1
+
+    return 0
+
+
+def prep(filename):
+    try:
+        with open(filename) as json_file:
+            json_string = json_file.read()
+            json_data = json.loads(json_string)
+    except OSError as e:
+        print(f'Failed to load file "{filename}": {e}')
+        return
+
+    schema_filename = json_data.get('$schema')
+    if not schema_filename:
+        print('File has no schema:', filename)
+        return
+
+    file_path = os.path.split(filename)[0]
+    schema_file = os.path.join(file_path, schema_filename)
+
+    try:
+        with open(schema_file) as json_file:
+            schema = json.load(json_file)
+    except OSError as e:
+        print(f'Failed to load schema file "{schema_file}": {e}')
+        return
+
+    validate(filename, json_data, json_string, schema)
+
+
+def validate(filename, json_data, json_string, schema):
+    try:
+        servicesPaths = calculate(json_string)
+    except InvalidInputError as e:
+        print("Error with file:", e)
+        return
+
+    cls = Draft7Validator(schema)
+
+    for e in sorted(cls.iter_errors(json_data), key=str):
+        print(f'{e}\nIn "{filename}"\n\n')
+        errorPath = '/'.join(str(v) for v in e.absolute_path)
+        errorEntry = servicesPaths['/' + errorPath]
+        errors.append({
+            "file": filename,
+            "start_line": errorEntry.value_start.line + 1,
+            "end_line": errorEntry.value_end.line + 1,
+            "title": "Validation Error",
+            "message": e.message,
+            "annotation_level": "failure"
+        })
+
+
+if __name__ == '__main__':
+    sys.exit(main())
--- a/plugins/rtmp-services/data/schema/service-schema-v3.json
+++ b/plugins/rtmp-services/data/schema/service-schema-v3.json
@@ -1,0 +1,212 @@
+{
+    "$schema": "http://json-schema.org/draft-07/schema#",
+    "type": "object",
+    "properties": {
+        "format_version": {
+            "type": "integer",
+            "description": "Identifier for parsing this file.\n- v3 introduced 'ffmpeg_hls_muxer' to services/recommended/output\n - v2 introduced 'alt_names' to services"
+        },
+        "services": {
+            "type": "array",
+            "items": {
+                "type": "object",
+                "properties": {
+                    "name": {
+                        "type": "string",
+                        "description": "Name of the streaming service. Will be displayed in the Service dropdown.",
+                        "minLength": 1
+                    },
+                    "common": {
+                        "type": "boolean",
+                        "description": "Whether or not the service is shown in the list before it is expanded to all services by the user.",
+                        "default": false
+                    },
+                    "stream_key_link": {
+                        "$ref": "#/definitions/saneUrl",
+                        "description": "Link where a logged-in user can find the 'stream key', presented as a button alongside the stream key field."
+                    },
+                    "servers": {
+                        "type": "array",
+                        "description": "List of servers.",
+                        "items": {
+                            "type": "object",
+                            "properties": {
+                                "name": {
+                                    "type": "string",
+                                    "description": "Name of the server (e.g. location, primary/backup), displayed in the Server dropdown.",
+                                    "minLength": 1
+                                },
+                                "url": {
+                                    "$ref": "#/definitions/serviceUri",
+                                    "description": "RTMP(S) or HLS URL of the ingest server.",
+                                    "minLength": 1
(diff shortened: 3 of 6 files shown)


.github/workflows/services-json.yml
```diff
@@ -0,0 +1,36 @@
+name: Services Validator
+
+on:
+  push:
```
derrod, line 4: The push rule is missing the path filter, I guess this was temporary for testing? Otherwise it should use the same path filter. Also since the annotation doesn't happen for non-PRs how is the failure logged?

.github/workflows/services-json.yml
```diff
@@ -0,0 +1,36 @@
...
+    paths-ignore:
+      - "cmake/**"
+    branches-ignore: [master]
```
derrod, line 7: Ignoring master branch is probably not desired. Since it's the only one we really use for active development.

.github/workflows/services-json.yml
```diff
@@ -0,0 +1,36 @@
...
+  pull_request:
+    paths:
+      - "plugins/rtmp-services/data/services.json"
+      - "plugins/rtmp-services/data/package.json"
+    branches-ignore: [master]
```
derrod, line 12: The ignore filter is evaluated against the target branch, not source branch, so this is not doing what you probably intended.

.github/workflows/services-json.yml
```diff
@@ -0,0 +1,36 @@
...
+jobs:
+  schema:
+    name: Schema
+    runs-on: [ubuntu-latest]
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v3
+        with:
+          submodules: "recursive"
+      - name: Install & Configure Python
```
derrod, line 23: Nitpick: for better readability I personally prefer having a blank line between steps

.github/workflows/services-json.yml
```diff
@@ -0,0 +1,36 @@
...
+
+jobs:
+  schema:
+    name: Schema
+    runs-on: [ubuntu-latest]
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v3
+        with:
+          submodules: "recursive"
```
derrod, line 22: I don't think we need to check out submodules here?

derrod (commented):
Just a first pass with some minor things I noticed without checking the python stuff itself.

.github/workflows/services-json.yml
WizardCM: If any validation errors are found and it's not a PR, it'll be treated properly as an error in the 'Validate Service JSON Schema' step itself, as the script will return a status code of 1 via `sys.exit(1)`.

derrod: Does the action support adding annotations to files in a commit rather than just leaving the errors in this log?

plugins/rtmp-services/data/schema/service-schema-v3.json
```diff
@@ -0,0 +1,206 @@
...
+                                "description": "OBS output module used.",
+                                "enum": [
+                                    "ffmpeg_hls_muxer",
+                                    "ftl_output"
+                                ]
+                            },
+                            "profile": {
+                                "type": "string",
+                                "description": "H.264 Profile.",
+                                "minLength": 1
```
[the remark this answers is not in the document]
WizardCM: I did consider doing that - I'll add it.

.github/workflows/services-json.yml
WizardCM: I am actually unsure. It's likely. I'll test it and get back to you.

RytoEX:
It kind of bothers me that a lot of the description strings use punctuation while not actually being complete sentences. At the moment, I'm not sure how best to address that concern though, since the easiest way to fix that is adding something like "This value is/represents/controls/etc." to the beginning of each sentence, and I understand that that may not be ideal in some cases.

Overall, I am in favor of the concept in this PR though. If I can't come up with any improvements to the language, then I can always defer those to a later time.

.github/workflows/services-json.yml
WizardCM: I have confirmed that annotations do work on straight commits, so we could enable the Annotate Errors step on all runs.

image

PatTheMav: Yeah I also think it might be better to have CI just check for errors but committers should run the validator locally before committing anyway, so a local check should then give the explicit error anyway.

CI/check-jsonschema.py
```diff
@@ -0,0 +1,45 @@
+import json
```
PatTheMav, line 1: As I mentioned in the related PR (https://github.com/obsproject/obs-studio/pull/6360#discussion_r856933794) I'd prefer if this script uses a `main` method and also handles file access exceptions.

WizardCM:
Rebased to master and rewrote the Python script based on feedback from @derrod and @PatTheMav - 

* there's now a `main` function
* script parameters are individual files to be validated
* CI will now validate both services.json and package.json
* CI will annotate regardless of whether it's a PR or a commit.

CI/check-jsonschema.py
```diff
@@ -0,0 +1,90 @@
...
+    if len(errors) > 0:
+        return 1
+
+    return 0
+
+
+def prep(filename):
+    global json_data
+    global json_string
+    global schema
```
derrod, line 36: Rather than using globals, why not just assing them here and pass them to the next function (since they're not used anywhere else)?

CI/check-jsonschema.py
```diff
@@ -0,0 +1,90 @@
...
+            json_string = json_file.read()
+            json_data = json.loads(json_string)
+    except OSError as e:
+        print(f'Failed loading "{filename}" with {e}')
+        return 1
+
+    schema_filename = json_data.get('$schema')
+    if not schema_filename:
+        print('File has no schema:', filename)
+        return 1
```
derrod, line 49: The return value of this function is never used.

CI/check-jsonschema.py
```diff
@@ -0,0 +1,90 @@
...
+            schema = json.load(json_file)
+    except OSError as e:
+        print(f'Could not load schema file: "{schema_file}" with {e}')
+        return 1
+
+    return validate(filename)
+
+
+def validate(filename):
+    global errors
```
derrod, line 65: `errors` doesn't need to be declared global if you're not reassigning it.

CI/check-jsonschema.py
```diff
@@ -0,0 +1,90 @@
...
+        print("JSON path required.")
+        return 1
+
+    for filename in inputArgs[1:]:
+        prep(filename)
+
+    with open('validation_errors.json', 'w') as outfile:
+        json.dump(errors, outfile)
+
+    if len(errors) > 0:
```
derrod, line 27: Personal taste thing, but `if errors:` works fine as well as `errors` is only truthy if it is non-empty.

CI/check-jsonschema.py
```diff
@@ -0,0 +1,90 @@
...
+
+def main():
+    if len(sys.argv) < 2:
+        print("JSON path required.")
+        return 1
+
+    for filename in inputArgs[1:]:
+        prep(filename)
+
+    with open('validation_errors.json', 'w') as outfile:
```
derrod, line 24: Missing exception handling when writing the output file.

CI/check-jsonschema.py
```diff
@@ -0,0 +1,90 @@
...
+inputArgs = sys.argv
+errors = []
+
+
+def main():
+    if len(sys.argv) < 2:
+        print("JSON path required.")
+        return 1
+
+    for filename in inputArgs[1:]:
```
derrod, line 21: Instead of declaring `inputArgs` just use `sys.argv` again and remove the unnecessary assignment at the top

derrod (changes requested):
I've had a couple issues with the python script speciifcally. Just a few oddities probably as a result from refactoring.

CI/check-jsonschema.py
```diff
@@ -0,0 +1,90 @@
...
+        servicesPaths = calculate(json_string)
+    except InvalidInputError as e:
+        print("Error with file:", e)
+        return 1
+
+    cls = Draft7Validator(schema)
+
+    for e in sorted(cls.iter_errors(json_data), key=str):
+        print(e)
+        print (f'    in {filename}')
```
PatTheMav, line 76: Extraneous whitespace between function name and parameters.

WizardCM:
All feedback actioned. Ready for re-review.

CI/check-jsonschema.py
```diff
@@ -0,0 +1,88 @@
...
+    json_string = ''
+    schema = {}
+
+    try:
+        with open(filename) as json_file:
+            json_string = json_file.read()
+            json_data = json.loads(json_string)
+    except OSError as e:
+        print(f'Failed to load file "{filename}": {e}')
+        return 1
```
derrod, line 44: This still returns >:(

CI/check-jsonschema.py
```diff
@@ -0,0 +1,88 @@
...
+    schema_file = os.path.join(file_path, schema_filename)
+
+    try:
+        with open(schema_file) as json_file:
+            schema = json.load(json_file)
+    except OSError as e:
+        print(f'Failed to load schema file "{schema_file}": {e}')
+        return
+
+    return validate(filename, json_data, json_string, schema)
```
derrod, line 61: You can just remove the `return` here.

CI/check-jsonschema.py
```diff
@@ -0,0 +1,88 @@
...
+    if errors:
+        return 1
+
+    return 0
+
+
+def prep(filename):
+    json_data = {}
+    json_string = ''
+    schema = {}
```
derrod, line 36: You don't need to assign them here.

plugins/rtmp-services/data/schema/service-schema-v3.json
```diff
@@ -0,0 +1,211 @@
...
+                                "type": "string",
+                                "description": "Additional x264 encoder options. Space-separated.",
+                                "pattern": "^(\\S+=\\S+\\s*)+$"
+                            },
+                            "output": {
+                                "type": "string",
+                                "description": "OBS output module used.",
+                                "enum": [
+                                    "ffmpeg_hls_muxer",
+                                    "ftl_output"
```
derrod, line 86: Technically `rtmp_output` is also valid.

derrod (changes requested):

WizardCM:
@derrod Done. Also rebased on master.

PR #6357, opened 2022-04-20
UI: Make window-projector its own top level QWidget

### LINKED ISSUE #6283: Preview window - no titlebar (Wayland, Gnome)
### Operating System Info

Other

### Other OS

Linux

### OBS Studio Version

27.2.4

_No response_

### OBS Studio Log URL

https://obsproject.com/logs/xV__iKzOBdld9tor

### OBS Studio Crash Log URL

_No response_

### Expected Behavior

After launching preview window it should have a titlebar with proper window controls.

### Current Behavior

It opens window without decoration / controls of any sorts.

### Steps to Reproduce

1. Create a screen / window capture.
2. Launch preview window

_No response_

kkartaltepe:
### Description
Previously OBSProjector was a subclass of OBSQTDisplay, which is
intended to own the entire surface it draws on. On wayland some
compositors expect the client to draw decorations (window titlebar), but
because OBSQTDisplay draws on the entire surface directly the client
decorations from QT are overwritten.

Instead promote OBSProjector to a simple QWidget with a single layout
containing the OBSQTDisplay. This allows the OBSQTDisplay to own the
entire contents while OBSProjector draws the decorations.

fixes #6283

### Motivation and Context
Broken rendering on wayland isnt that great.

### How Has This Been Tested?
Tested on gnome/i3. I couldnt immediately find any issues with projector (windowed/fullscreen) or with multi-scene previews (windowed/fullscreen). But some more testing that this doesnt adversely affect other platforms would be good, though this brings the display closer to rendering like the preview than it was before.

I'm also not very proficient at Qt so there could be some subtle issues with this minor change.

### Types of changes

### Checklist:

### DIFF
--- a/UI/window-projector.cpp
+++ b/UI/window-projector.cpp
@@ -9,6 +9,16 @@
 #include "qt-wrappers.hpp"
 #include "platform.hpp"
 
+#ifdef ENABLE_WAYLAND
+#include <obs-nix-platform.h>
+#endif
+
+#ifdef _WIN32
+#define NOMINMAX
+#define WIN32_LEAN_AND_MEAN 1
+#include <Windows.h>
+#endif
+
 static QList<OBSProjector *> multiviewProjectors;
 static QList<OBSProjector *> allProjectors;
 
@@ -19,11 +29,16 @@ static size_t maxSrcs, numSrcs;
 
 OBSProjector::OBSProjector(QWidget *widget, obs_source_t *source_, int monitor,
 			   ProjectorType type_)
-	: OBSQTDisplay(widget, Qt::Window),
+	: display(this),
 	  source(source_),
 	  removedSignal(obs_source_get_signal_handler(source), "remove",
 			OBSSourceRemoved, this)
 {
+	QVBoxLayout *layout = new QVBoxLayout;
+	layout->addWidget(&display);
+	layout->setContentsMargins(0, 0, 0, 0);
+	this->setLayout(layout);
+
 	isAlwaysOnTop = config_get_bool(GetGlobalConfig(), "BasicWindow",
 					"ProjectorAlwaysOnTop");
 
@@ -32,7 +47,7 @@ OBSProjector::OBSProjector(QWidget *widget, obs_source_t *source_, int monitor,
 
 	// Mark the window as a projector so SetDisplayAffinity
 	// can skip it
-	windowHandle()->setProperty("isOBSProjectorWindow", true);
+	display.windowHandle()->setProperty("isOBSProjectorWindow", true);
 
 #if defined(__linux__) || defined(__FreeBSD__) || defined(__DragonFly__)
 	// Prevents resizing of projector windows
@@ -69,12 +84,13 @@ OBSProjector::OBSProjector(QWidget *widget, obs_source_t *source_, int monitor,
 	auto addDrawCallback = [this]() {
 		bool isMultiview = type == ProjectorType::Multiview;
 		obs_display_add_draw_callback(
-			GetDisplay(),
+			display.GetDisplay(),
 			isMultiview ? OBSRenderMultiview : OBSRender, this);
-		obs_display_set_background_color(GetDisplay(), 0x000000);
+		obs_display_set_background_color(display.GetDisplay(),
+						 0x000000);
 	};
 
-	connect(this, &OBSQTDisplay::DisplayCreated, addDrawCallback);
+	connect(&display, &OBSQTDisplay::DisplayCreated, addDrawCallback);
 	connect(App(), &QGuiApplication::screenRemoved, this,
 		&OBSProjector::ScreenRemoved);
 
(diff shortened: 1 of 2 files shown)


UI/qt-display.hpp
```diff
@@ -5,12 +5,16 @@
...
 
+class OBSProjector;
+
 class OBSQTDisplay : public QWidget {
 	Q_OBJECT
 	Q_PROPERTY(QColor displayBackgroundColor MEMBER backgroundColor READ
 			   GetDisplayBackgroundColor WRITE
 				   SetDisplayBackgroundColor)
 
+	friend class OBSProjector;
```
jpark37, line 16: Can you just make the necessary members public instead?

UI/window-projector.cpp
```diff
@@ -19,11 +19,16 @@ static size_t maxSrcs, numSrcs;
 
 OBSProjector::OBSProjector(QWidget *widget, obs_source_t *source_, int monitor,
 			   ProjectorType type_)
-	: OBSQTDisplay(widget, Qt::Window),
+	: display(OBSQTDisplay(this)),
```
jpark37, line 22: Is this doing a construct and move? I would expect to see `display(this)` here.

jpark37 (commented):
This might break window projector SDR/HDR auto-update when dragging across monitors or changing OS settings on Windows.

EDIT: I don't know off the top of my head if the necessary events are wired up in this new setup.

UI/qt-display.hpp
kkartaltepe: I couldnt connect the signal without making this a friend class, is there some way to mark signals as public to enable OBSProjector to connect them?

jpark37: Do you mean the two mouse events? I'm not sure those parent calls were even doing anything useful; they might be there out of good habit/practice. If you want to keep them in case the base functionality changes, you can add public methods to reach them, and call those from the projector instead.
```
void OBSQTDisplay::OnMouseDoubleClick(QMouseEvent *event)
{
	mouseDoubleClickEvent(event);
}

void OBSQTDisplay::OnMousePress(QMouseEvent *event)
{
	mousePressEvent(event);
}
```

jpark37: Never mind the new public functions; you probably want this to match original intent.

```
void OBSProjector::mouseDoubleClickEvent(QMouseEvent *event)
{
	QWidget::mouseDoubleClickEvent(event);
```
```
void OBSProjector::mousePressEvent(QMouseEvent *event)
{
	QWidget::mousePressEvent(event);
```

kkartaltepe: No this is being used to connect the DisplayCreated signal, in window-project.cpp:83

I can try also passing the mouse events through to the QWidget instead of the display directly as well.

jpark37: K, I'll take a look at the signal after work.

kkartaltepe:
Marked draft for a bit, there are some more oddities that occur on wayland which were working on the original setup.

UI/qt-display.hpp
jpark37: >No this is being used to connect the DisplayCreated signal, in window-projector.cpp:83

I'm not seeing an issue. addDrawCallback is being called fine as part of DisplayCreated, at least on Windows. DisplayCreated is already public since `signals:` is the same as `public QT_ANNOTATE_ACCESS_SPECIFIER(qt_signal):`, so you shouldn't need to be a friend.

Also connect is a static member of QObject, so you shouldn't need to use `display.connect`, just
`connect(&display, &OBSQTDisplay::DisplayCreated, addDrawCallback);`

kkartaltepe: Yes that worked, I must have had something different when it was complaining initially.

kkartaltepe:
> 	virtual void paintEvent(QPaintEvent *event) override;
>	virtual void moveEvent(QMoveEvent *event) override;
>	virtual void resizeEvent(QResizeEvent *event) override;

Can you describe how these work for the preview? I dont see them being implemented for preview just the `nativeEvent` override is done.

jpark37:
Sorry, copy/pasted from the wrong header. I meant this:
```
	virtual void mousePressEvent(QMouseEvent *event) override;
	virtual void mouseDoubleClickEvent(QMouseEvent *event) override;
	virtual void closeEvent(QCloseEvent *event) override;
#if QT_VERSION >= QT_VERSION_CHECK(6, 0, 0)
	virtual bool nativeEvent(const QByteArray &eventType, void *message,
				 qintptr *result) override;
#else
	virtual bool nativeEvent(const QByteArray &eventType, void *message,
				 long *result) override;
#endif
```
Just add virtual to the existing declarations. I don't like disguising virtual functions as nonvirtual, even with the override keyword as a hint. I prefer them to be nice and obvious.

UI/window-projector.cpp
```diff
@@ -32,7 +47,7 @@ OBSProjector::OBSProjector(QWidget *widget, obs_source_t *source_, int monitor,
 
 	// Mark the window as a projector so SetDisplayAffinity
 	// can skip it
-	windowHandle()->setProperty("isOBSProjectorWindow", true);
+	display.windowHandle()->setProperty("isOBSProjectorWindow", true);
```
jpark37, line 50: Not sure this line needs to change. I suspect it should probably be left alone.

WizardCM: I agree. The property should be set on the window, otherwise the check which uses this property will fail.

kkartaltepe:
Sorry it seems its still premature for this one, i mostly got the projector working but the noticed many other issues with displays caused by this and havnt had time to look into and fix them.

kkartaltepe:
Closing this as i probably wont get to it unless more people complain about not being able to see the title bars on wayland.

PR #6360, opened 2022-04-20
CI: Add automatic service checks

derrod:
### Description

Adds scheduled/manually triggered workflow run to check if service entries are still valid (i.e. if servers are reachable).

Also reformats the currently hand-formatte d`package.json` and `services.json` with python to get a consistent result.

(Waiting for #6354 to be reviewed/merged)

### Motivation and Context

If a service dies but nobody uses it, does it make a sound?

Well, no. So we have to periodically check ourselves and purge defunct servers/services if necessary.

### How Has This Been Tested?

Ran manually on my fork:
- Workflow run: https://github.com/derrod/obs-studio/runs/6098001774?check_suite_focus=true
- Resuling PR: https://github.com/derrod/obs-studio/pull/5

### Types of changes

### Checklist:

### DIFF
--- a/.github/workflows/services-json.yml
+++ b/.github/workflows/services-json.yml
@@ -9,6 +9,9 @@ on:
     paths:
       - "plugins/rtmp-services/data/services.json"
       - "plugins/rtmp-services/data/package.json"
+  schedule:
+  - cron: 0 0 * * *
+  workflow_dispatch:
 
 jobs:
   schema:
@@ -38,3 +41,57 @@ jobs:
           repo-token: "${{ secrets.GITHUB_TOKEN }}"
           title: "Service JSON Errors"
           input: "./validation_errors.json"
+
+  service_check:
+    name: Service Check
+    runs-on: ubuntu-20.04
+    needs: schema
+    if: ${{ github.repository_owner == 'obsproject' && (github.event_name == 'schedule' || github.event_name == 'workflow_dispatch') }}
+
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v3
+        with:
+          fetch-depth: 0
+
+      - name: Restore cache
+        uses: actions/cache@v3
+        with:
+          path: ${{ github.workspace }}/other
+          key: service-check
+
+      - name: Install & Configure Python
+        run: |
+          sudo apt install python3.9
+          python3.9 -m pip install requests
+
+      - name: Check Services
+        id: check
+        run: |
+           python3.9 -u CI/check-services.py
+           # if there are changes, run the PR step
+           if ! git diff --quiet; then
+             echo "::set-output name=make_pr::true"
+           else
+             echo "::set-output name=make_pr::false"
+           fi
+        env:
+          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+          WORKFLOW_RUN_ID: ${{ github.run_id }}
+          REPOSITORY: ${{ github.repository }}
+
+      - uses: actions/upload-artifact@v3
+        with:
+          name: timestamps
+          path: ${{ github.workspace }}/other/*
+
+      - name: Create Pull Request
+        uses: peter-evans/create-pull-request@f094b77505fb89581e68a1163fbd2fffece39da1
+        if: steps.check.outputs.make_pr == 'true'
+        with:
+          author: "Service Checker <commits@obsproject.com>"
+          commit-message: "rtmp-services: Remove defunct servers/services"
+          title: "rtmp-services: Remove defunct servers/services"
+          branch: "automated/clean-services"
+          body: "Automatic PR to remove dead servers\nCreated by workflow run: https://github.com/${{ github.repository }}/actions/runs/${{ github.run_id }}"
+          delete-branch: true
--- a/CI/check-services.py
+++ b/CI/check-services.py
@@ -1,0 +1,248 @@
+import json
+import socket
+import ssl
+import os
+import time
+import requests
+import sys
+import zipfile
+
+from io import BytesIO
+from random import randbytes
+from urllib.parse import urlparse
+
+MINIMUM_PURGE_AGE = 9.75 * 24 * 60 * 60  # slightly less than 10 days
+TIMEOUT = 10
+SKIPPED_SERVICES = {'YouNow', 'SHOWROOM', 'Dacast'}
+SERVICES_FILE = 'plugins/rtmp-services/data/services.json'
+PACKAGE_FILE = 'plugins/rtmp-services/data/package.json'
+CACHE_FILE = 'other/timestamps.json'
+
+context = ssl.create_default_context()
+
+
+def check_ftl_server(hostname) -> bool:
+    """Check if hostname resolves to a valid address - FTL handshake not implemented"""
+    try:
+        socket.getaddrinfo(hostname, 8084, proto=socket.IPPROTO_UDP)
+    except socket.gaierror as e:
+        print(f'⚠️ Could not resolve hostname for server: {hostname} (Exception: {e})')
+        return False
+    else:
+        return True
+
+
+def check_hls_server(uri) -> bool:
+    """Check if URL responds with status code < 500 and not 404, indicating that at least there's *something* there"""
+    try:
+        r = requests.post(uri, timeout=TIMEOUT)
+        if r.status_code >= 500 or r.status_code == 404:
+            raise Exception(f'Server responded with {r.status_code}')
+    except Exception as e:
+        print(f'⚠️ Could not connect to HLS server: {uri} (Exception: {e})')
+        return False
+    else:
+        return True
+
+
+def check_rtmp_server(uri) -> bool:
+    """Try connecting and sending a RTMP handshake (with SSL if necessary)"""
+    parsed = urlparse(uri)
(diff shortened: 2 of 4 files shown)


derrod:
Based on feedback on Discord this has been slightly reworked to use a cache to store negative results, and only remove servers after they have been unreachable for three consecutive days. The cronjob frequency has also been increased to once per day instead of once per month to ensure the cache is not evicted.

derrod:
PR and workflow run with the revised workflow, only containining servers that were unresponsive 3 days in a row:  
https://github.com/derrod/obs-studio/pull/6  
https://github.com/derrod/obs-studio/runs/6133734392?check_suite_focus=true

.github/workflows/services-json.yml
```diff
@@ -0,0 +1,89 @@
...
+      - name: Checkout
+        uses: actions/checkout@v3
+
+      - name: Install & Configure Python
+        run: |
+          sudo apt install python3.9-dev
+          python3.9 -m pip install jsonschema json_source_map
+
+      - name: Validate Service JSON Schema
+        run: |
```
PatTheMav, line 30: If it's a single line, you can omit the pipe - but maybe it's easier for readability to do something like:

```
JSON_FILES=(
  plugins/rtmp-services/data/services.json
  plugins/rtmp-services/data/schema/service-schema-v3.json
)
python3.9 CI/check-jsonschema.py "${JSON_FILES[@]}"
```

In that case you'd just need to add additional JSON files into the array at the top and avoid having to use backslashes to make the single line more readable.

.github/workflows/services-json.yml
```diff
@@ -0,0 +1,89 @@
...
+          repo-token: "${{ secrets.GITHUB_TOKEN }}"
+          title: 'Service JSON Errors'
+          input: './validation_errors.json'
+
+  service_check:
+    name: Service Check
+    runs-on: ubuntu-latest
+    needs: schema
+    # todo remove my repo before un-drafting PR
+    if: (github.repository_owner == 'obsproject' ||  github.repository_owner == 'derrod') && (github.event_name == 'schedule' || github.event_name == 'workflow_dispatch')
```
PatTheMav, line 46: Longer evaluations need to be wrapped in `${{ }}` last I checked, otherwise not the entire statement might be executed. I remember I had peculiar issues when I didn't wrap those.

.github/workflows/services-json.yml
```diff
@@ -0,0 +1,89 @@
...
+        run: |
+          sudo apt install python3.9
+          python3.9 -m pip install requests
+
+      - name: Check Services
+        id: check
+        run: |
+           python3.9 -u CI/check-services.py
+           # if there are unstaged changes, run the PR step
+           if [ -n "$(git status --porcelain=v1 2>/dev/null)" ]; then
```
PatTheMav, line 70: Use `if git diff --exit-code &>/dev/null; then` - no need to do string comparisons here.

CI/check-jsonschema.py
```diff
@@ -0,0 +1,45 @@
+import json
```
PatTheMav, line 1: Maybe that's just me, but I was taught (a lot of moons ago) to wrap code _at the very least_ in a `main` function and run with:

```
if __name__ == '__main__':
    sys.exit(main())
```

Also there are not checks whether the file specified by `sys.argv[1]` exists, so at the very least the `open` calls should be wrapped in a `try..except..finally` block.

CI/check-services.py
```diff
@@ -0,0 +1,168 @@
+import json
```
PatTheMav, line 1: See comment above - wrap main code in a `main` function, also handle file exceptions.

.github/workflows/services-json.yml
derrod: Seems like `git diff --quiet` would be an even better choice here, it implies `--exit-code` but silences all output so the redirect isn't required.

derrod:
Leaving the remaining comments open for @WizardCM's PR. Also squashed everything before the inevitable rebase.

.github/workflows/services-json.yml
derrod: As an amendment, the `package.json` file should probably also be verified. Perhaps the script could be reworked to take just the to-be-verified file as input, and then load the schema from the `$schema` key and then ran in a loop over `JSON_FILES`?

e.g.
```python
try:
    with open(filename) as f:
        raw_file = f.read()
        data = json.loads(raw_file)
except OSError as e:
    print(f'Failed loading "{filename}" with {e}')
    continue

schema_filename = data.get('$schema')
if not schema_filename:
    print('File has no scehma:', filename)
    return 1

file_path = os.path.split(filename)[0]
schema_file = os.path.join(file_path, schema_filename)

try:
    with open(schema_file) as f:
        json.load(f)
except OSError as e:
    print(f'Could not load schema file: {schema_file} (Exception: {e}')
    return 1

# Schema check here
...
```

derrod:
Rebased on #6354

derrod:
Rebased on master and ready for merge.

.github/workflows/services-json.yml
```diff
@@ -38,3 +41,57 @@ jobs:
           repo-token: "${{ secrets.GITHUB_TOKEN }}"
           title: "Service JSON Errors"
           input: "./validation_errors.json"
+
+  service_check:
+    name: Service Check
+    runs-on: ubuntu-latest
```
RytoEX, line 47: If this job relies on any specific version packages, consider pinning the image at ubuntu-2004 or ubuntu-2204.

.github/workflows/services-json.yml
```diff
@@ -38,3 +41,57 @@ jobs:
...
+          WORKFLOW_RUN_ID: ${{ github.run_id }}
+          REPOSITORY: ${{ github.repository }}
+
+      - uses: actions/upload-artifact@v3
+        with:
+          name: timestamps
+          path: ${{ github.workspace }}/other/*
+
+      - name: Create Pull Request
+        uses: peter-evans/create-pull-request@v4
```
RytoEX, line 89: While we usually use `@v#` for first-party actions, for third-party actions, consider pinning them to a specific commit or tag.

.github/workflows/services-json.yml
```diff
@@ -38,3 +41,57 @@ jobs:
...
+        if: steps.check.outputs.make_pr == 'true'
+        with:
+          author: "Automaton <noreply@obsproject.com>"
```
RytoEX, line 92: Did you mean Automation, or was this intentional? Consider using commits@obsproject.com.

plugins/rtmp-services/data/package.json
```diff
@@ -1,11 +1,11 @@
...
+    "$schema": "schema/package-schema.json",
+    "url": "https://obsproject.com/obs2_update/rtmp-services",
+    "version": 197,
+    "files": [
+        {
+            "name": "services.json",
+            "version": 197
+        }
+    ]
+}
```
RytoEX, line 11: End files with a newline. If this was automated, we may need to adjust the formatter.

plugins/rtmp-services/data/services.json
```diff
@@ -2617,4 +2624,4 @@
             }
         }
     ]
-}
+}
```
RytoEX, line 2627: End files with a newline. If this was automated, we may need to adjust the formatter.

RytoEX (changes requested):
Some nits and some questions.

plugins/rtmp-services/data/package.json
derrod: Why? JSON data is generally not terminated with a newline.

.github/workflows/services-json.yml
derrod: It was intentional, I can use commits as well, I think we discussed this briefly but there was no consensus.

.github/workflows/services-json.yml
derrod: Shouldn't cause any issues, the python setup action it uses work with 18.04, 20.04, and 22.04.

plugins/rtmp-services/data/package.json
RytoEX: IIRC, _some_ Linux tools expect all files to end with a newline character, otherwise the last line is not treated like a full line and is never read/parsed correctly. I don't know if this is a historical artifact that is no longer relevant, or if there are still some tools for which this is true.

Also, GitHub's web UI yells at you about it.

.github/workflows/services-json.yml
RytoEX: It's mostly to keep a sudden change in two years from breaking things on us, but I won't block on this.

plugins/rtmp-services/data/package.json
derrod: I believe it's true of a variety of tools like `wc` (word count), `diff` will work but yell at you. I suppose one valid argument is that - sadly - we're hand editing these data files, so while technically JSON is terminated with `}` and a newline is superfluous some editors will automatically add a newline. So I guess for compatibility with those we'll have to have one (if we want cleaner diffs).

derrod:
Ignore the build failure, Flatpak job having a hiccup.

RytoEX (approved):
Looks good to me.

PR #6361, opened 2022-04-20
CI: Respect user-specified build directory

RytoEX:
### Description

The Windows build script would respect the user-specified build directory for the configure step, but not for the build step. Fix that by applying the same logic in Configure-OBS to Build-OBS.

### Motivation and Context

We don't want to clobber the default build32 or build64 directories if the user specifies a build directory.

### How Has This Been Tested?

I ran the build script locally before and after this change to confirm this works as expected.

### Types of changes

### Checklist:


PatTheMav (approved):
Looks fine to me.

PR #6372, opened 2022-04-23
obs-filter: Add new RTX audio fx (dereverb & combined denoiser+ dereverb) 

pkviet:
### Description
The newer Nvidia RTX audio SDK in addition to noise reduction supports : 

- Room echo removal or Dereverberation;
- a combined FX with both noise suppression & dereverberation.

This PR adds these two FX as new options for noise suppression.
This requires the RTX audio FX sdk to be updated.

**Testing**
Test builds are available here: https://github.com/obsproject/obs-studio/actions/runs/2712752768
(requires a github account)

**Note:** a nasty bug with RTX audio creates an interference with RTX video FX.
See https://github.com/obsproject/obs-studio/pull/6365.
The same commits upgrading the RTX audio sdk + the bugfix are included both there and here.
Of course they should not be merged from both PRs. But in order to be able to have
test builds, it's better to have the audio sdk upgrade commit included in both PRs.

**WARNING: RTX Audio bugs**
- high GPU usage ? windows task manager indicates very high GPU loads (for instance 35%) when using the FX; upon investigation, these numbers don't seem representative at all of the real GPU load, incurred by the use of RTX audio. The impact on fps is according to nvidia between 3-5 %.

I'm in discussion w/ nvidia and hopefully they'll fing fixes for both bugs in the next SDK release.

### Motivation and Context
Echoes in large rooms (and sometimes not so large) can be a nuisance.
So allow for their removal.

### How Has This Been Tested?
Tested on windows 10 x64.
Here are demoes : 
(imo in all fairness the FX seems better w/ both denoiser & dereverb rather than just dereverb).
1) original file with echoes / reverberations (from nvidia):
 
https://user-images.githubusercontent.com/9283407/164916237-6110172d-54d6-4ffd-aa5e-d513b2251d56.mp4

 2) recording from obs w/ rtx  dereverb filter :

https://user-images.githubusercontent.com/9283407/164916274-97bcfa0d-9421-4858-b8f4-5a41192cef4a.mp4

3) original with both noise and echoes (from nvidia):

https://user-images.githubusercontent.com/9283407/164916298-809377f4-cbf7-4776-ab63-f08942397920.mp4

4) recording from obs w/ rtx denoiser+ dereverb filter :

https://user-images.githubusercontent.com/9283407/164916337-b9bdd37b-f065-42c7-80e3-9122b773b41f.mp4

### Types of changes

### Checklist:


pkviet:
**update (april 30 2022):**

- fixed a bug at load time where the fx was not set at initialization.
- decreased GPU load for mono sources. If the output is to stereo, we copy the fx processed data for one channel to the second instead of processing twice the same data. (Not doable for more general speaker layouts because we'd need to know the layout conversion matrix).
- did some profiling; the fx processing takes 5 ms for 40 ms of audio data. (480 x 4 samples, 480 samples = 10 ms)
- the high GPU load indicated by windows task manager is difficult to interpret, because it's related to GPU copy and not GPU 3D. When playing a game, the added % of both can go above 100 % (ex 140%). A better metrics is to measure the fps drop : I find that there's about 9% fps decease at 150 fps (150 ==> 135 fps) with Unigine Heaven. Nvidia tells me that their own measurements yield about 3-5% fps drops. The discrepancy is not understood at this time. It could be because we're not benchmarking in the same manner.

plugins/obs-filters/data/locale/en-US.ini
```diff
@@ -75,7 +75,9 @@ NoiseSuppress.Intensity="Suppression Intensity"
 NoiseSuppress.Method="Method"
 NoiseSuppress.Method.Speex="Speex (low CPU usage, low quality)"
 NoiseSuppress.Method.RNNoise="RNNoise (good quality, more CPU usage)"
-NoiseSuppress.Method.nvafx="NVIDIA Noise Removal (good quality, no CPU usage)"
+NoiseSuppress.Method.Nvafx="NVIDIA Noise Removal FX"
+NoiseSuppress.Method.Nvafx.Dereverb="NVIDIA Dereverb FX"
+NoiseSuppress.Method.Nvafx.Dereverb.Denoiser="NVIDIA Noise Removal + Dereverb FX"
```
RytoEX, line 80: If using "Noise Removal" as the user-facing string, I'd use "Room Echo Removal" instead of "Dereverb". That aligns with the phrasing in both of these documents:
* https://docs.nvidia.com/deeplearning/maxine/audio-effects-sdk/index.html#afx-intro
* https://nvidia.frontify.com/d/uAobRitG8H8B/with-nvidia#/nvidia-broadcast-partner-branding-guidelines/messaging/feature-naming

The full string would then be:
> NVIDIA Noise Removal and Room Echo Removal

I don't think we need "FX" appended to each string.

plugins/obs-filters/data/locale/en-US.ini
```diff
@@ -75,7 +75,9 @@ NoiseSuppress.Intensity="Suppression Intensity"
 NoiseSuppress.Method="Method"
 NoiseSuppress.Method.Speex="Speex (low CPU usage, low quality)"
 NoiseSuppress.Method.RNNoise="RNNoise (good quality, more CPU usage)"
-NoiseSuppress.Method.nvafx="NVIDIA Noise Removal (good quality, no CPU usage)"
+NoiseSuppress.Method.Nvafx="NVIDIA Noise Removal FX"
```
RytoEX, line 78: Should use `NoiseSuppress.Method.Nvafx.Denoise` for the string name for specifying "Noise Removal" now that we have more than one NVAFX method.

plugins/obs-filters/data/locale/en-US.ini
RytoEX, line 80: Prefer: `NoiseSuppress.Method.Nvafx.DenoisePlusDereverb`

RytoEX (changes requested):
Some string nits.

Commit message nits:
1. Always write "NVIDIA" in all caps.
2. Correct "finctions" to "functions".
3. Write a comma after "Additionally".
4. Write "NVIDIA Audio Effects SDK" or "NVIDIA AFX SDK". I don't think it's called the "NVIDIA RTX Audio FX SDK" anywhere.
3. Write "CUDA" in call caps.
5. Write "SDK" in all caps.
6. Write "Noise Removal" instead of "Background Noise suppression".
7. Wrap commit message bodies at 72 characters.

plugins/obs-filters/data/locale/en-US.ini
pkviet: done

plugins/obs-filters/data/locale/en-US.ini
pkviet: done

plugins/obs-filters/data/locale/en-US.ini
pkviet: done

plugins/obs-filters/data/locale/en-US.ini
pkviet: The SDK also uses Dereverb but ok, i've changed to Room Echo Removal.

pkviet:
> Some string nits.
> 
> Commit message nits:
> 
> 1. Always write "NVIDIA" in all caps.

ok corrected that.

> 2. Correct "finctions" to "functions".

done
> 3. Write a comma after "Additionally".
done

> 4. Write "NVIDIA Audio Effects SDK" or "NVIDIA AFX SDK". I don't think it's called the "NVIDIA RTX Audio FX SDK" anywhere.

done
> 5. Write "CUDA" in call caps.

done
> 6. Write "SDK" in all caps.
done
> 7. Write "Noise Removal" instead of "Background Noise suppression".
done
> 8. Wrap commit message bodies at 72 characters.
done

plugins/obs-filters/noise-suppress-filter.c
```diff
@@ -414,9 +460,40 @@ static void noise_suppress_update(void *data, obs_data_t *s)
...
+				if (NvAFX_DestroyEffect(ng->handle[i]) !=
+				    NVAFX_STATUS_SUCCESS) {
+					do_log(LOG_ERROR,
+					       "FX failed to be destroyed.");
+					ng->use_nvafx = false;
+				} else
+					ng->handle[i] = NULL;
+				pthread_mutex_unlock(&ng->nvafx_mutex);
+				nvafx_initialize_internal(data);
+				pthread_mutex_unlock(&ng->nvafx_mutex);
```
RytoEX, line 480: Why is the same mutex unlocked twice here?

RytoEX (commented):
A visual pass raised a question about a mutex unlock.

plugins/obs-filters/noise-suppress-filter.c
pkviet: i goofed. 
Fixed in new push.

RytoEX (changes requested):
Single nit on a comment. Looks fine to me from a visual review otherwise.

plugins/obs-filters/noise-suppress-filter.c
```diff
@@ -224,58 +230,74 @@ static void noise_suppress_destroy(void *data)
...
-	pthread_mutex_lock(&ng->nvafx_mutex);
-	pthread_mutex_lock(&nvafx_initializer_mutex);
 	if (!ng->handle[0]) {
 		ng->sample_rate = NVAFX_SAMPLE_RATE;
-
 		for (size_t i = 0; i < ng->channels; i++) {
-			err = NvAFX_CreateEffect(NVAFX_EFFECT_DENOISER,
-						 &ng->handle[i]);
+			// Create FX
+			CUcontext old = {0}, curr = {0};
```
RytoEX, line 243: Style: Split these assignments onto individual lines.

RytoEX (changes requested):
I'd missed this because I'm mixing up this with #6365.

Just want to confirm that _this_ should be merged first and _then_ #6365 gets cleaned up and merged after this. Correct?

plugins/obs-filters/noise-suppress-filter.c
```diff
@@ -414,9 +460,39 @@ static void noise_suppress_update(void *data, obs_data_t *s)
...
+				pthread_mutex_unlock(&ng->nvafx_mutex);
+			}
+		}
+	} else {
+		ng->fx = bstrdup(method);
 	}
+#endif
+	ng->use_nvafx = ng->nvafx_enabled && nvafx_requested;
+
+	/* Process 10 millisecond segments to keep latency low */
```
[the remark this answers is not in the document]
pkviet: done

plugins/obs-filters/noise-suppress-filter.c
pkviet: done

RytoEX:
While retesting this to confirm that a bug had been fixed, I noticed that the volume mixer activity didn't seem to match up with my voice activity. Upon checking a test recording, it seems that I ended up with around 6 seconds of delay between audio and video. Not sure if this was because the laptop was in a low power state.

RytoEX:
Discussed off-thread and the delay I experienced was not reproduced by @pkviet . I retested and confirmed that the delay is a power state issue. As previously stated, the test system I used is a laptop with a mobile RTX 2070. When removed from AC power, the GPU clock speeds drop precipitously, which causes the issue to appear. My guess is that the GPU cannot process the audio fast enough, so it returns the audio data late (6 seconds late, in my case). The issue resolves as soon as you plug back into AC power. I think this is extremely unlikely to occur on desktop systems. This not a blocker, in my opinion. We'll just have to be aware of it in case we see any complaints about audio delay.

RytoEX (changes requested):
Final nits:
1. Remove the period from the commit subject in the second commit.
2. Wrap commit message bodies at 72 characters.

After that, I'll merge this immediately.

pkviet:
> Final nits:
> 
> 1. Remove the period from the commit subject in the second commit.
> 2. Wrap commit message bodies at 72 characters.
> 
> After that, I'll merge this immediately.

done & pushed; i'll rebase greenscreen once it's merged. Thx

RytoEX (approved):

PR #6374, opened 2022-04-24
obs-outputs: Implement send timeout in librtmp (Fix RTMP output deadlock)

tt2468:
### Description
This fixes a bug where the RTMP send thread can deadlock upon the underlying TCP connection being broken. By introducing a send timeout, this allows the thread to unblock and give up, triggering a reconnect as normal. The correct solution to this problem would be to rewrite librtmp with asynchronous IO, but that seems like something unlikely to happen.

### Motivation and Context
Having the TCP connection between OBS and the media server broken has been pretty common in my experience, and OBS currently doesn't handle this very well. If this happens, you usually have to either force close OBS or wait an indefinite amount of time for it to give up on the connection naturally.

### How Has This Been Tested?
**Before**:
- Start stream in OBS
- Use tool (pfSense) to invalidate connection state
- OBS bitrate drops to 0
- Output does not respond to stop signals, and hangs for an undefined
amount of time (usually multiple minutes) before finally giving up

**After**:
- Start stream in OBS
- Use tool (pfSense) to invalidate connection state
- OBS bitrate drops to 0
- Output sits in blocked state for maximum of 8 seconds, then cleans
up and triggers the reconnect logic

### Types of changes

### Checklist:

### DIFF
--- a/plugins/obs-outputs/librtmp/rtmp.c
+++ b/plugins/obs-outputs/librtmp/rtmp.c
@@ -484,7 +484,8 @@ RTMP_Reset(RTMP *r)
     r->m_fVideoCodecs = 252.0;
     r->Link.curStreamIdx = 0;
     r->Link.nStreams = 0;
-    r->Link.timeout = 30;
+    r->Link.receiveTimeout = 30;
+    r->Link.sendTimeout = 6;
     r->Link.swfAge = 30;
 }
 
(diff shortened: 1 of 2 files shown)


plugins/obs-outputs/librtmp/rtmp.c
```diff
@@ -928,11 +929,19 @@ RTMP_Connect0(RTMP *r, struct sockaddr * service, socklen_t addrlen)
...
+            RTMP_Log(RTMP_LOGERROR, "%s, Setting socket receive timeout to %ds failed!",
+                     __FUNCTION__, r->Link.timeout);
+        }
+
+        SET_SNDTIMEO(tvs, r->Link.sendTimeout);
+        if (setsockopt
+                (r->m_sb.sb_socket, SOL_SOCKET, SO_SNDTIMEO, (char *)&tvs, sizeof(tvs)))
+        {
+                RTMP_Log(RTMP_LOGERROR, "%s, Setting socket send timeout to %ds failed!",
                      __FUNCTION__, r->Link.timeout);
```
notr1ch, line 945: The `RTMP_Log` calls need to be changed to use the new timeout fields, this won't compile as-is.

plugins/obs-outputs/librtmp/rtmp.c
```diff
@@ -484,7 +484,8 @@ RTMP_Reset(RTMP *r)
     r->m_fVideoCodecs = 252.0;
     r->Link.curStreamIdx = 0;
     r->Link.nStreams = 0;
-    r->Link.timeout = 30;
+    r->Link.receiveTimeout = 30;
+    r->Link.sendTimeout = 4;
```
notr1ch, line 488: How was this value determined? The PR says 8 seconds but the code uses 4. I'm worried 4 seconds might be a little short to allow for proper TCP retransmission / recovery on higher latency connections such as mobile data.

plugins/obs-outputs/librtmp/rtmp.c
tt2468: On my system (Ubuntu) the actual timeout is double whatever the configured timeout is. It seems to be a common problem which is why I set the number. Just increased it to 6 seconds since Windows probably behaves normally.

plugins/obs-outputs/librtmp/rtmp.c
tt2468: On my system (Ubuntu) the actual timeout is double whatever the configured timeout is. It seems to be a common problem which is why I set the number. Just increased it to 6 seconds since Windows probably behaves normally.

plugins/obs-outputs/librtmp/rtmp.c
tt2468: Fixed

RytoEX:
Per comment from @notr1ch, we're merging this now, but we'll need to keep an eye out for possible side effects in nightlies and any OBS 28 beta releases.

PR #6387, opened 2022-04-28
CI: Update macOS image to macOS 12

RytoEX:
### Description

Update the macOS image used on GitHub Actions from macOS 11 to macOS 12.

### Motivation and Context

Use the latest Xcode and macOS SDK. Unblock #5875 by allowing CI to produce builds using the macOS 12.3 SDK and ScreenCaptureKit.

### How Has This Been Tested?

CI is the test. CI must return all green for macOS builds before this can be merged.

### Types of changes

### Checklist:

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -65,7 +65,7 @@ jobs:
 
   macos_build:
     name: '02 - macOS'
-    runs-on: [macos-11]
+    runs-on: [macos-12]
     strategy:
       matrix:
         arch: ['x86_64', 'arm64']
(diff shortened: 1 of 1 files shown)


gxalpha (approved):
made sure the artifact still works (it does) (honestly though, there wouldn't be any reason why it shouldn't).
lgtm.

PatTheMav:
While we're changing that line, might make sense to remove the array and specify it as a simple string (I think that's what the schema prefers as well).

PatTheMav (approved):
Apart from the comment above, looks fine to me.

Generated artefact ran without issues and compatibility versions were correctly set on the binary:

```
minos 11.0
sdk 12.3
```

(`minos` is 11 because I need to use the M1 build).

RytoEX:
> While we're changing that line, might make sense to remove the array and specify it as a simple string (I think that's what the schema prefers as well).

If you mean changing `[macos-12]` to `macos-12`, then we should probably change those for all other instances other than the macOS jobs. Though, I don't think it makes any difference functionally. Since there are other places where this syntax is used, I'd prefer to just clean that up separately at some point if needed.

PR #6388, opened 2022-04-28
Fix indent on multiline comments

norihiro:
### Description

Fix indent of 2nd line and later for multiline comment blocks.

The most frequent comment indentation style is as below. I followed this style to update comments.
- Indent the first line of a comment with tab(s).
- Indent remaining lines with the same amount of tab(s) + one space so that the first `*` is vertically aligned.

### Motivation and Context

Some comment blocks are ugly indented like below. `clang-format` did not touch the indent of comments.
https://github.com/obsproject/obs-studio/blob/a82bbb416f84461172224ac33d59d281668e17e0/libobs/obs-source.c#L1508-L1513

### How Has This Been Tested?

`git show -w` shows nothing was changed.
Ran `./CI/check-format.sh` and the updated indent is not modified on Fedora 34.

### Types of changes

### Checklist:

### DIFF
--- a/libobs/obs-source.c
+++ b/libobs/obs-source.c
@@ -1505,13 +1505,13 @@ static void source_output_audio_data(obs_source_t *source,
 		if (diff < TS_SMOOTHING_THRESHOLD) {
 			push_back = true;
 
-			/* This typically only happens if used with async video when
-		 * audio/video start transitioning in to a timestamp jump.
-		 * Audio will typically have a timestamp jump, and then video
-		 * will have a timestamp jump.  If that case is encountered,
-		 * just clear the audio data in that small window and force a
-		 * resync.  This handles all cases rather than just looping. */
 		} else if (diff > MAX_TS_VAR) {
+			/* This typically only happens if used with async video when
+			 * audio/video start transitioning in to a timestamp jump.
+			 * Audio will typically have a timestamp jump, and then video
+			 * will have a timestamp jump.  If that case is encountered,
+			 * just clear the audio data in that small window and force a
+			 * resync.  This handles all cases rather than just looping. */
 			reset_audio_timing(source, data->timestamp, os_time);
 			in.timestamp = data->timestamp + source->timing_adjust;
 		}
--- a/plugins/win-capture/get-graphics-offsets/d3d9-offsets.cpp
+++ b/plugins/win-capture/get-graphics-offsets/d3d9-offsets.cpp
@@ -94,32 +94,32 @@ static const uint8_t mask[][MAX_CMP_SIZE] = {
 
 static const uint8_t mask_cmp[][MAX_CMP_SIZE] = {
 	/*
-	* Windows 7
-	* 48 8B 83 B8 3D 00 00                    mov     rax, [rbx+3DB8h]
-	* 44 39 B8 68 50 00 00                    cmp     [rax+5068h], r15d
-	* 75 12                                   jnz     short loc_7FF7AA90530
-	* 41 B8 F9 19 00 00                       mov     r8d, 19F9h
-	*/
+	 * Windows 7
+	 * 48 8B 83 B8 3D 00 00                    mov     rax, [rbx+3DB8h]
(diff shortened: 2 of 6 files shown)


RytoEX:
I would rather that corrections to SIMDE be made upstream, if necessary.

jp9000:
Making corrections to an upstream usually requires that upstream to be responsive. If obs is any indicator that can take a while for some projects, although maybe simde isn't quite as big as obs. I personally think it's okay to make modifications downstream in conjunction to submitting those same changes upstream. If it's further modified upstream later, then we'll just merge the changes after that whenever they happen.

PatTheMav:
@RytoEX looks fine to me, what do you think?

libobs/obs-source.c
```diff
@@ -1506,11 +1506,11 @@ static void source_output_audio_data(obs_source_t *source,
...
-		 * audio/video start transitioning in to a timestamp jump.
-		 * Audio will typically have a timestamp jump, and then video
-		 * will have a timestamp jump.  If that case is encountered,
-		 * just clear the audio data in that small window and force a
-		 * resync.  This handles all cases rather than just looping. */
+			 * audio/video start transitioning in to a timestamp jump.
+			 * Audio will typically have a timestamp jump, and then video
+			 * will have a timestamp jump.  If that case is encountered,
+			 * just clear the audio data in that small window and force a
+			 * resync.  This handles all cases rather than just looping. */
```
RytoEX, line 1513: This actually looks like it's a comment about the below `else if`, in which case the first line should probably be de-indented. That, or the whole comment block gets moved into the `else if` so that the indentation makes sense.

plugins/obs-libfdk/obs-libfdk.c
```diff
@@ -133,8 +133,8 @@ static void *libfdk_create(obs_data_t *settings, obs_encoder_t *encoder)
 		break;
 
 		/* lib_fdk-aac > 1.3 required for 7.1 surround;
-         * uncomment if available on linux build
-         */
+		 * uncomment if available on linux build
+		 */
```
RytoEX, line 137: If this is about `case 8`, then the first line should probably be de-indented, and then the following lines made to match that.

Perhaps this should be placed inside the `ifndef` so that it is not separated from the section it's commenting about.

plugins/win-capture/get-graphics-offsets/d3d9-offsets.cpp
```diff
@@ -229,7 +229,7 @@ void get_d3d9_offsets(struct d3d9_offsets *offsets)
 					break;
 
 					/* check to make sure offsets actually point
-				 * toward expected data */
+					 * toward expected data */
```
RytoEX, line 232: Again, this seems like the first line should be de-indented one level. Perhaps this should be placed inside the `ifdef` so that it is not separated from the section it's commenting about.

RytoEX (changes requested):
See my comments about comment alignment.

WizardCM:
Is there maybe a clang-format configuraiton option we could enable to automatically align comments?

libobs/obs-source.c
norihiro: The comment is moved inside the closure for `if (diff > MAX_TS_VAR)`.

plugins/win-capture/get-graphics-offsets/d3d9-offsets.cpp
norihiro: De-indented one level and moved under the line `#ifdef _MSC_VER`.

plugins/obs-libfdk/obs-libfdk.c
norihiro: I will make another PR #6393 to remove the outdated code. This change is removed.

norihiro:
> Is there maybe a clang-format configuraiton option we could enable to automatically align comments?

`ReflowComments: true` enables clang-format to format comments.
However, it also touch more, such as line-break in the comments.
Setting  `ReflowComments`, amount of changes will be as below.
```
126 files changed, 808 insertions(+), 621 deletions(-)
```

PR #6391, opened 2022-04-29
libobs/util: Add %s string replacement for unix time

tt2468:
### Description
Adds %s which replaces the keyword with the current unix epoch time.

The %s keyword is common in languages like PHP and Python, and even
some C implementations, but it is not standard so this is a fallback.

### Motivation and Context
There is not currently a good way to parse recording filenames with scripts, and also not a timezone-agnostic format available. Unix epoch time is an effective and commonly used method of portraying time.

Requested by a friend who needed it.

### How Has This Been Tested?
Added %s to my recording format and it gets replaced with UNIX time.

### Types of changes

### Checklist:


Vainock:
It would make sense for the translation of this string to convert it into many smaller pieces, but that's probably not possible to implement in a 'clean' way.

norihiro:
Is there any concern the newly added format is conflicting with existing `%ss`?

tt2468:
> Is there any concern the newly added format is conflicting with existing `%ss`?

On linux, it does not appear to. Will test on windows though.

tt2468:
Indeed it is unaffected on Windows 10.

norihiro:
> > Is there any concern the newly added format is conflicting with existing `%ss`?
> 
> On linux, it does not appear to. Will test on windows though.

I just concerned the case a user want to add `s` followed by `%s`. Maybe it’s not so important for most cases.

tt2468:
> > > Is there any concern the newly added format is conflicting with existing `%ss`?
> > 
> > 
> > On linux, it does not appear to. Will test on windows though.
> 
> I just concerned the case a user want to add `s` followed by `%s`. Maybe it’s not so important for most cases.

Well in that case it would qualify as the `%ss` macro. Using an escape `\` should work fine.

PR #6395, opened 2022-04-29
UI: Fix portable mode detection on Linux

PatTheMav:
### Description
Fixes broken portable mode detection for configuration file resolution introduced by the CMake rework.

### Motivation and Context
UI code uses both simple checks for a preprocessor define as well as boolean checks inside an `if`-conditional.

The define will be always set however (albeit using an empty string if the build flag is not enabled), which breaks both kinds of checks (an empty string always resolves as `true` and "defined").

With this change, the variable is not set if the build flag is not enabled and only simple `ifdef` checks are used.

### How Has This Been Tested?
* Tests on Linux VM pending

### Types of changes

### Checklist:


PatTheMav:
Checked the code in my VM - without `LINUX_PORTABLE` the app uses the default config path (in my case at `${HOME}/.config`) regardless of the command line switch.

With `LINUX_PORTABLE` enabled, only passing `--portable` triggers the desired behaviour: OBS uses a config subdirectory two levels above the binary.

RytoEX (approved):
@notr1ch approved off-thread. Looks fine to me too.

RytoEX (changes requested):
Ah, one commit message nit:
maco -> macro

PatTheMav:
> Ah, one commit message nit:
> maco -> macro

Should be fixed now.

RytoEX (approved):

PR #6407, opened 2022-05-01
plugins: Fix decklink gradual audio desync

tt2468:
### Description
Fixes desync with Decklink output by dropping audio frames when the card's audio buffer grows too large as a result of clock drift.

### Motivation and Context
All my homies hate clock drift.

### How Has This Been Tested?
Ubuntu 20.04:
- Load OBS
- Run an AV sync test mp4 and record it using a known-good reference recorder
- Wait as long as you want (hours or days)
- Check it again, AV should have the same amount of desync as the original test

It needs to be noted that this fix is not meant to make decklink perfect, it's just meant to fix *one of many* problems causing desync. The point is to prevent desync caused by clock drift from growing over time.

IF YOU WANT TO TEST, download the latest CI artifact from this PR. You should perform the steps above to validate that desync does not get worse than when you first started. Log messages are posted every time audio frames are dropped to compensate for desync. `Total drift offset` should be the amount of time that audio would lag further behind video without this fix.

I used another computer with a Mini Recorder 4k and Media Express to record the raw frames coming from a duo2, and played the audio slowed down to determine how large the delay got.

### Types of changes

### Checklist:

### DIFF
--- a/UI/window-basic-main.cpp
+++ b/UI/window-basic-main.cpp
@@ -4436,7 +4436,7 @@ bool OBSBasic::ResetAudio()
 {
 	ProfileScope("OBSBasic::ResetAudio");
 
-	struct obs_audio_info ai;
+	struct obs_audio_info2 ai;
 	ai.samples_per_sec =
 		config_get_uint(basicConfig, "Audio", "SampleRate");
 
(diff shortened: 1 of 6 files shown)


thomaslfessler:
I'd like to test this on an Arm64 build of OBS with a Mac Studio.   It looks like the build did not complete due to an error with the Code Format Check. Any progress?   I have a Thunderbolt and PCIe decklink capture card running on a mac studio and while it works with an x86_64 version of OBS, it does not work with a recent pipeline build of the ARM64 version of OBS.

If the test version runs with my hardware config, I'd be happy to test it for you.  Feel free to reach out to me at thomasfesslerchannel@gmail.com

gxalpha:
@thomaslfessler the build did complete, you can find artifacts by scrolling down on [on this page](https://github.com/obsproject/obs-studio/actions/runs/2253173277). In particular, [this](https://github.com/obsproject/obs-studio/suites/6332056709/artifacts/227845871) is the arm64 macOS build.

PR #6409, opened 2022-05-01
UI: Raise replay buffer max size

TheSwerik:
(I don't know which branch to point the PR to, so please tell me in a comment if master is incorrect)

### Description
This Raises the Maximum Size of the Replay Buffer from 8192 MB to 1,000,000 MB in the UI. (arbitrarily large number)
This already works if you change this setting in the basic.ini file but the changes from there do not get reflected in the UI.

### Motivation and Context
I had a full discussion about this topic on [ideas.obsproject.com](https://ideas.obsproject.com/posts/1784/remove-replay-buffer-max-size-limit-inconsistency).
Many people argued in the forum or on ideas.obsproject.com for the maximum size of the Replay Buffer to be raised for various needs.

But I argue that:
1. the maximum size of the Replay Buffer is already higher than 8192 MB, you just have to set it in the basic.ini file but in the UI 8192 MB will still be displayed even though OBS uses the basic.ini file's number
2. if the file size would be capped at 8192 MB then this would be inconsistent because if you use Constant Bitrate, you can have settings like 300,000 kbits 600 seconds which works fine and calculates to 21549 MB which is way more than 8192 MB

### How Has This Been Tested?
I tested this on the latest OBS Version (27.2.4) on Windows 10:
I changed the setting in the basic.ini file to 16,000 MB and then hit "Start Replay Buffer" and then I waited until I saw OBS in the Task manager taking up around 18 GB.
Then I tested the actual change of this Pull Request and it works:
obs64_KJHHoPnjsP

### Types of changes

### Checklist:
- [x] I have included updates to all appropriate documentation. (I think there is no documentation on this)

### DIFF
--- a/UI/forms/OBSBasicSettings.ui
+++ b/UI/forms/OBSBasicSettings.ui
@@ -1866,7 +1866,7 @@
                        <number>20</number>
                       </property>
                       <property name="maximum">
-                       <number>8192</number>
+                       <number>1000000</number>
                       </property>
                       <property name="value">
                        <number>512</number>
@@ -3902,7 +3902,7 @@
                            <number>20</number>
                           </property>
                           <property name="maximum">
-                           <number>8192</number>
+                           <number>1000000</number>
                           </property>
                           <property name="value">
                            <number>512</number>


derrod:
While I agree that the limit *should* be higher, this is probably not the right approach. Ideally the limit should be something like 75% of total available system RAM, just to prevent people from shooting themselves in the foot too easily (they can still edit the config, but then they have no right to complain to us about pulling the trigger).

TheSwerik:
1. I found out how to build it on my machine, I will test my commit shortly

2. @derrod I don't know how I would calculate that dynamically. Or are you suggesting just using 75% of 16GB which would be 12GB? I disagree with the latter. What happens when the Replay Buffer overflows the System RAM, do you know that? like will the recording just crash or will OBS crash or will a BSOD happen?

derrod:
I mean setting it dynamically. We already query how much system RAM is installed (as you can see in the log).

If you use too much RAM your system will just start swapping/paging (unless you disabled that). It should never BSOD, but I'm not sure if it might terminate OBS if you were to overrun.

TheSwerik:
Okay that sounds reasonable I think
I still don't know how I would do that though

TheSwerik:
@derrod But in my opinion, this is still inconsistent with the fact that you can have settings like 300,000 kbits 600 seconds which calculates to 21549 MB even though you only have 16GB installed.
And not even in the basic.ini, you can do this through the UI.
I don't like this inconsistency there...

TheSwerik:
updated the Description because I tested the change now.

derrod:
Closing this because we do not think this is the right approach. At this point you might as well just remove the limit entirely and that's not what we want.

TheSwerik:
> @derrod But in my opinion, this is still inconsistent with the fact that you can have settings like CBR 300,000 kbits 600 seconds which calculates to 21549 MB even though you only have 16GB installed. And not even only through the basic.ini, you can do this through the UI. I don't like this inconsistency there...
> 
> That's why I like either capping it even through the ini or just not capping it in the UI (which is this PR)

@derrod Why is that not wanted? As I said, this is inconsistent with CBR because there the max RAM usage is uncapped. People who don't know what they are doing can have settings that will result in a replay buffer that is bigger than their actual installed RAM. 
So because of this inconsistency, I would say either remove the limit from VBR and CQP as well or limit all 3 because this inconsistency does not make sense.

PR #6416, opened 2022-05-03
mac-capture: Improve window capture performance

gxalpha:
### Description

Replaces the current implementation of Window Capture, which uses Core
Graphics to render a bitmap raster, with one that accesses the window's
CGImage bitmap directly, blits it, and hands that off as a frame to OBS
instead.

Taken out of #5680.

### Motivation and Context

@dossy did great work figuring this out, the entire pull request
is just rather big (also containing other changes and cleanups),
with everything being in one commit which makes it somewhat
difficult to review (It also contains changes to how the window
selection works, which is somewhat controversial and will require
more debate).
The performance enhancement itself though is pretty crucial for
anyone using window capture, it should not get held up in other
debates.

### How Has This Been Tested?

macOS 12.4 Beta
I can no longer count individual frames when using window capture :P
No seriously, while I haven't done any benchmarking (others have
done that on the original PR already) it's *noticeably* less laggy (on
my rather old machine).

### Types of changes

### Checklist:

### DIFF
--- a/plugins/mac-capture/mac-window-capture.m
+++ b/plugins/mac-capture/mac-window-capture.m
@@ -50,28 +50,28 @@ static inline void capture_frame(struct window_capture *wc)
 	size_t width = CGImageGetWidth(img);
 	size_t height = CGImageGetHeight(img);
 
-	CGRect rect = {{0, 0}, {width, height}};
-	da_resize(wc->buffer, width * height * 4);
-	uint8_t *data = wc->buffer.array;
+	if (!width || !height || CGImageGetBitsPerPixel(img) != 32 ||
+	    CGImageGetBitsPerComponent(img) != 8) {
+		CGImageRelease(img);
+		return;
+	}
 
-	CGContextRef cg_context = CGBitmapContextCreate(
-		data, width, height, 8, width * 4, wc->color_space,
-		kCGBitmapByteOrder32Host | kCGImageAlphaPremultipliedFirst);
-	CGContextSetBlendMode(cg_context, kCGBlendModeCopy);
-	CGContextDrawImage(cg_context, rect, img);
-	CGContextRelease(cg_context);
-	CGImageRelease(img);
+	CGDataProviderRef provider = CGImageGetDataProvider(img);
+	CFDataRef data = CGDataProviderCopyData(provider);
 
 	struct obs_source_frame frame = {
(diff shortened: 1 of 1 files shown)


plugins/mac-capture/mac-window-capture.m
```diff
@@ -50,28 +50,28 @@ static inline void capture_frame(struct window_capture *wc)
...
-	uint8_t *data = wc->buffer.array;
-
-	CGContextRef cg_context = CGBitmapContextCreate(
-		data, width, height, 8, width * 4, wc->color_space,
-		kCGBitmapByteOrder32Host | kCGImageAlphaPremultipliedFirst);
-	CGContextSetBlendMode(cg_context, kCGBlendModeCopy);
-	CGContextDrawImage(cg_context, rect, img);
-	CGContextRelease(cg_context);
-	CGImageRelease(img);
+	if ((!width && !height) || CGImageGetBitsPerPixel(img) != 32 ||
```
PatTheMav, line 53: If I read this right, if either width or height is missing (but not both), the function won't return early.

plugins/mac-capture/mac-window-capture.m
```diff
@@ -50,28 +50,28 @@ static inline void capture_frame(struct window_capture *wc)
...
 		.width = width,
 		.height = height,
-		.data[0] = data,
-		.linesize[0] = width * 4,
+		.data[0] = (uint8_t *)CFDataGetBytePtr(data),
+		.linesize[0] = CGImageGetBytesPerRow(img),
 		.timestamp = ts,
 	};
 
 	obs_source_output_video(wc->source, &frame);
```
PatTheMav, line 71: I take it `obs_source_output_video` blocks and uploads the data to the GPU before returning?

gxalpha: I'm not entirely sure about the internal details, but I do know that it does a copy of the memory before returning, so we can release the data immediately afterwards (if that's the question).

plugins/mac-capture/mac-window-capture.m
gxalpha: You are right, that should probably be an "or".

plugins/mac-capture/mac-window-capture.m
PatTheMav: Yeah should be fine, there is a `memcpy` involved.

PR #6420, opened 2022-05-05
UI: Add "H.264" to simple hardware encoders

jpark37:
### Description
Hopefully stops people from assuming other video compression standards
are usable from these settings.

### Motivation and Context
People were assuming NVENC setting could be used for HEVC.

### How Has This Been Tested?
- [x] Verified all GUI string locations.
- [x] All diffs have been breakpoint debugger inspected to behave as expected.

### Types of changes

### Checklist:


jp9000:
The recording part is fine, but for streaming we may have to check to see what services support it and remove it from the streaming encoder list if they don't support HEVC.

We could merge this as-is while HEVC is disabled by default, although having a supported codec list for each service will probably be important when they support AV1 too.

Gol-D-Ace:
Is this respecting that for example Pascal (10xx) cards do not support bframes?

I noticed that while trying advanced output mode that bframes are enabled by default which will result in an error message.

jpark37:
I'm starting to think that adding HEVC to simple mode is anything but.

What if I just update the GUI strings for the existing hardware encoders in the list to point out that they are H.264? I'm just trying to resolve the confusion that a hardware encoder might implicitly choose a codec when it doesn't, which multiple early testers of HDR have incorrectly assumed.

jpark37:
This change just modifies the strings a little.

Fenrirthviti:
Just to throw in my opinion, I think keeping HEVC in the advanced section is going to be fine. I would prefer we didn't expose it to regular users, both from a support standpoint, and a personal standpoint of not wanting to push a morally-bankrupt licensed encoder.

Users who are looking for the option and know what HEVC is are typically going to have a better understanding on what it is and how to use it in the first place. Simple mode users just want to stream, and don't care about the technical minutiae of encoder choice past "works" and "doesn't work"

jpark37:
We may not want users to venture into advanced settings to use HDR, but I think we're looking at a more extensive reshuffling of settings if that's a goal.

Issue #6295, opened 2022-04-07
VLC Media Player Source does not properly support 5.1 video files when streaming to any output sound profile. (Stereo, 2.1, etc)


dencur:
### Operating System Info

Windows 11

### Other OS

Ubuntu 20.04, Windows 10

### OBS Studio Version

27.2.4

### OBS Studio Log URL

https://obsproject.com/logs/wpBbuDdp4GsJhMtF

### Expected Behavior

Audio playback through the VLC Media Player Source that has 5.1 Audio should correctly output audio that balances dialog for L and R channels.

The Media Player Source does not have this problem, however it can not play subtitles.  The only solution as of now is to Window Capture VLC, or Re-encode the video file so that subtitles are embedded and use the Media Player Source.

### Current Behavior

When listening to any videos with 5.1 Sound (3F2M/LFE), the audio channels for dialog favor the Left speaker, unless multiple people are speaking, during which audio fades between Left and Right speakers.  In a movie, this means most dialog comes out of the Left Speaker/Channel.

### Steps to Reproduce

1.  Add VLC Media Source
2.  Add Video with 5.1 Sound. (i.e. 3F2M/LFE)
3.  Configure Audio as any: Stereo, 2.1, 3.1, 5.1, 6.1, 7.1, 8.1, or 9.1
4.  Configure Audio Output for Local Monitoring -OR- Stream and Listen to the output
5.  Audio Without Dialog sounds well balanced.  Any Dialog heavily favors Left Channel unless multiple people are speaking, in which it might output some dialog to the right channel.

### Anything else we should know?

This seems to have been attempted to be fixed here, but there are still issues noted above.

https://github.com/obsproject/obs-studio/pull/5881

I have tested this on Windows 10, Windows 11, and Ubuntu 20.04 which all exhibit the same issue.

pkviet:
is it possible to get a sample of your source ? i'll have a look.
It's quite odd. It 's probably that the channel layout is not correctly retrieved by swresample lib.

dencur:
Thanks!  Here's a public sample that has 5.1, and showcases the issue very well:  https://drive.google.com/file/d/0BwxFVkl63-lETFVpOWRVVGNfeEU/view?usp=sharing&resourcekey=0-e9HrVO61IGkoTc_GBMNbww

Pulled from: https://kodi.wiki/view/Samples

pkviet:
[...]

PR #6424, opened 2022-05-05
vlc-source: Fix surround sound not properly downmixed

pkviet:
### Description
This fixes issue https://github.com/obsproject/obs-studio/issues/6295 .
libvlc does some downmixing/upmixing when the number of channels
requested is less than the number of channels of the source.
We take advantage of that feature to avoid doing it with swresample
because we're missing info that libvlc is not giving (the exact channel
layout of the source).

### Motivation and Context
Fixes a bug.

**Important Note**

- surprisingly libvlc does not provide any speaker layout info about a source it decodes. I checked with ePirat. This is a real bummer because this means we can't let swresample do its usual task of downmixing or upmixing to the output speaker layout.
- the libvlc audio callback has however an interesting feature which I haven't seen documented, which is if we request less channels than the original audio soruce has, then vlc will downmix it.
- That's the feature we take advantage of in this PR. A 5.1 source will be then properly downmixed to stereo.
- However a 5.1(side) source won't be touched at all and since obs default corresponds to 5.1(back) (default speaker layout for aac), there will be some channels which are swapped in the wrong positions. That's not something as easy to solve and is beyond the scope of this PR. A solution would be to have the user select a speaker layout and that the plugin calls swresample. That's not something I want to implement at the moment, but maybe in the future.

### How Has This Been Tested?
Fixes the issue https://github.com/obsproject/obs-studio/issues/6295 

### Types of changes

### Checklist:

### DIFF
--- a/plugins/vlc-video/vlc-video-source.c
+++ b/plugins/vlc-video/vlc-video-source.c
@@ -458,10 +458,13 @@ static int vlcs_audio_setup(void **p_data, char *format, unsigned *rate,
 {
 	struct vlc_source *c = *p_data;
 	enum audio_format new_audio_format;
+	struct obs_audio_info aoi;
+	obs_get_audio_info(&aoi);
+	int out_channels = (int)get_audio_channels(aoi.speakers);
 
 	new_audio_format = convert_vlc_audio_format(format);
-	if (*channels > 8)
-		*channels = 8;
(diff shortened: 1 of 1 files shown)


pkviet:
Bug confirmed as fixed. Removing the Draft status

PR #6425, opened 2022-05-06
CMake headers installation fix with HEVC include fix

tytan652:
### Description

- Stop including `obs-hevc.{h,c}` in sources target if HEVC is disabled with a fix for `jim-nvenc`
- Headers are now installed per default on Linux
- `obs-hevc.h` is not installed if HEVC is disabled
- OS exclusive headers are no longer installed on the wrong OS

### Motivation and Context

This needed to be fixed so I fixed it.

The HEVC change was made at the same time since the HEVC header is not required if disabled.

### How Has This Been Tested?

- Linux OBS Portable installation have headers without Windows or macOS headers
- OBS build with HEVC disabled or enabled

### Types of changes

### Checklist:

### DIFF
--- a/cmake/Modules/ObsHelpers.cmake
+++ b/cmake/Modules/ObsHelpers.cmake
@@ -295,28 +295,6 @@ function(export_target target)
     EXCLUDE_FROM_ALL)
 endfunction()
 
-# Helper function to install header files
-function(install_headers target)
-  install(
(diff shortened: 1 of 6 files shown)


tytan652:
Just remembered that obs-nix headers will be copied on macOS and Windows even if header are not present per default.

PR #6431, opened 2022-05-07
obs-outputs: Improve librtmp timeouts on Linux

tt2468:
### Description
Replaces the usage of SO_SNDTIMEO with TCP_USER_TIMEOUT on Linux.
- Noted as more effective than SO_SNDTIMEO by multiple sources.
- Drops TCP connection on timeout instead of returning send() errors, meaning more predictable reconnect timing and reconnect.
- Timeout period actually reflected in practice. SO_SNDTIMEO takes double the timeout time than actually requested to timeout on unix, whereas WinSock works correctly.

Depends #6651 

### Motivation and Context
Timeouts should function correctly regardless of OS.

### How Has This Been Tested?
Ubuntu 20.04: Works as expected
Windows: Works as expected

### Types of changes

### Checklist:

### DIFF
--- a/plugins/obs-outputs/librtmp/rtmp.c
+++ b/plugins/obs-outputs/librtmp/rtmp.c
@@ -85,7 +85,16 @@ static const char *my_dhm_G = "4";
 #include <openssl/bio.h>
 #include <openssl/buffer.h>
 #endif
+#endif
 
+#if defined(TCP_USER_TIMEOUT)
+#define SOCKET_LEVEL IPPROTO_TCP
+#define SOCKET_TIMEOUT_OPT TCP_USER_TIMEOUT
+#define SOCKET_TIMEOUT_VAR(tv, s) int tv = s*1000
+#else
+#define SOCKET_LEVEL SOL_SOCKET
+#define SOCKET_TIMEOUT_OPT SO_SNDTIMEO
+#define SOCKET_TIMEOUT_VAR(tv, s) SET_RCVTIMEO(tv, s)
 #endif
 
 #define RTMP_SIG_SIZE 1536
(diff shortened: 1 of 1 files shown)


tytan652:
The second commit should prefixed `librtmp:` or it's fine like that ?

tt2468:
> The second commit should prefixed `librtmp:`, no ?

Many previous commits to librtmp use the `obs-outputs` prefix.

notr1ch:
The documentation is quite vague, but I don't think `TCP_MAXRT` is a total timeout, rather it is a maximum retransmission timeout. So this will probably not do what you want it to do, though having a low retransmission timeout is a potentially good thing to experiment with as well since we desire a fast recovery where possible.

Either way this will need some experimenting with on Windows.

tt2468:
I've focused the scope of this PR to address the unix socket timeout discrepancies.

plugins/obs-outputs/librtmp/rtmp.c
```diff
@@ -85,7 +85,16 @@ static const char *my_dhm_G = "4";
...
 
+#if defined(TCP_USER_TIMEOUT)
+#define SOCKET_LEVEL IPPROTO_TCP
+#define SOCKET_TIMEOUT_OPT TCP_USER_TIMEOUT
+#define SOCKET_TIMEOUT_VAR(s) int to = s*1000
+#else
+#define SOCKET_LEVEL SOL_SOCKET
+#define SOCKET_TIMEOUT_OPT SO_SNDTIMEO
+#define SOCKET_TIMEOUT_VAR(s) SET_RCVTIMEO(to, s)
 #endif
```
notr1ch, line 98: Not a fan of the `SOCKET_TIMEOUT_VAR` style of setting the timeout. It makes the
```SOCKET_TIMEOUT_VAR(r->Link.sendTimeout);```
line later on in the code impossible to understand without having to refer back to the define. I understand the desire to avoid code duplication but I think this makes it too unclear what's going on.

File: plugins/obs-outputs/rtmp-stream.c [hunk 2]
```diff
@@ -1050,19 +1049,10 @@ static int try_connect(struct rtmp_stream *stream)
...
+	// free any existing RTMP TLS context
 	RTMP_TLS_Free(&stream->rtmp);
-	RTMP_TLS_Init(&stream->rtmp);
 
-	// since we don't call RTMP_Init above, there's no other good place
-	// to reset this as doing it in RTMP_Close breaks the ugly RTMP
-	// authentication system
-	memset(&stream->rtmp.Link, 0, sizeof(stream->rtmp.Link));
-	stream->rtmp.last_error_code = 0;
+	RTMP_Init(&stream->rtmp);
```
notr1ch, line 1055: I'm still worried this breaks something as there was a reason for the partial reset code we were doing before. I am almost certain it was for RTMP authentication, but as long as that still works we're probably fine.

tt2468: I spent hours looking through commits in librtmp to determine that the iterations of partial reset code were just the result of misunderstandings. The actual RTMP object should not be used at all across connections.

plugins/obs-outputs/librtmp/rtmp.c
tt2468: I've updated the macro here to align with the `SET_RCVTIMEO(tv, s)` macro.

PR #6438, opened 2022-05-08
UI: Add ability to send messages to YouTube chat

gxalpha:
#### Test builds
Click [here](https://github.com/obsproject/obs-studio/actions/runs/2339957170) for CI test builds with YouTube enabled.

### Description

Adds a custom text box to the YouTube chat which allows for sending messages using the YouTube API.

image

#### Drawbacks
- Due to how the chat polls, after sending a message it takes 2-3 seconds for the message to appear in the chat. This would only be fixable by implementing a completely custom chat which has already been decided against in the past for various reasons (which still apply).
- The YouTube API doesn't appear to give you any info on whether the live chat is actually enabled on a stream (until you try to send a message). This means that when it's disabled, the text input field is still visible and users can still try to send messages. In that case, an error message will appear after sending.

### Motivation and Context

Users are used to being able to send messages to Twitch, they expect to be able to do so on YouTube as well.

### How Has This Been Tested?

macOS 12.4 Beta 4
Sending messages works!

Also tested on Windows 11.

### Types of changes

### Checklist:

### DIFF
--- a/UI/data/locale/en-US.ini
+++ b/UI/data/locale/en-US.ini
@@ -1321,3 +1321,8 @@ YouTube.Errors.livePermissionBlocked="Live streaming is unavailable on the selec
 YouTube.Errors.errorExecutingTransition="Transition failed due to a backend error. Please try again in a few seconds."
 YouTube.Errors.errorStreamInactive="YouTube is not receiving data for your stream. Please check your configuration and try again."
 YouTube.Errors.invalidTransition="The attempted transition was invalid. This may be due to the stream not having finished a previous transition. Please wait a few seconds and try again."
+# Chat errors
+YouTube.Errors.liveChatDisabled="Live chat is disabled on this stream."
+YouTube.Errors.liveChatEnded="Live stream has ended."
+YouTube.Errors.messageTextInvalid="The message text is not valid."
+YouTube.Errors.rateLimitExceeded="You are sending messages too quickly."
--- a/UI/data/locale/en-US.ini
+++ b/UI/data/locale/en-US.ini
@@ -1315,6 +1315,12 @@ YouTube.Actions.AutoStartStreamingWarning.Title="Manual start required"
 YouTube.Actions.AutoStartStreamingWarning="Auto-start is disabled for this event, click \"Go Live\" to start your broadcast."
 YouTube.Actions.AutoStopStreamingWarning="You will not be able to reconnect.<br>Your stream will stop and you will no longer be live."
 
+YouTube.Chat.Input.Send="Send"
+YouTube.Chat.Input.Placeholder="Enter message here..."
+YouTube.Chat.Input.Sending="Sending..."
+YouTube.Chat.Error.Title="Error while sending message"
+YouTube.Chat.Error.Text="The message couldn't be sent: %1"
+
 # YouTube API errors in format "YouTube.Errors.<error reason>"
 YouTube.Errors.liveStreamingNotEnabled="Live streaming is not enabled on the selected YouTube channel.<br/><br/>See <a href='https://www.youtube.com/features'>youtube.com/features</a> for more information."
 YouTube.Errors.livePermissionBlocked="Live streaming is unavailable on the selected YouTube Channel.<br/>Please note that it may take up to 24 hours for live streaming to become available after enabling it in your channel settings.<br/><br/>See <a href='https://www.youtube.com/features'>youtube.com/features</a> for details."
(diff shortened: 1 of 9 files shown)


derrod:
Since chat messages can have multiple lines, I think we should probably use a multiline editor, and then either have the button below (e.g. like twich where it's in the right of a vbox with a spacer on the left) or make it taller.

High quality paint-based concept art:
concept

I was also thinking that perhaps we can set a placeholder text in the text box starting with something like "Type here..." and changing it to something like "Message sent, it may take a few seconds for it to show up" (just shorter) after a message is sent so people aren't *as* confused when it takes a bit for it to show up.

gxalpha:
Are you sure that the YouTube chat allows multiline? I went for the single-line since when I tested this, pressing enter even with shift always sent the message, and the API seems to just ignore any `\n` newline.
image

Or do you just mean that the text can get so long that it could extend over multiple lines and it would be a better experience when typing to see everything?

derrod:
Mh I must've remembered wrong, I thought YouTube chat messages could contain newlines, but apparently that is not true.

As far as better visual presentation having a larger multi-line input box is probably still a good idea (not sure if you can disable newlines there and just have it break the text on the rendering side).

gxalpha:
> As far as better visual presentation having a larger multi-line input box is probably still a good idea (not sure if you can disable newlines there and just have it break the text on the rendering side).

Yeah that should be possible, albeit a bit complicated because this is Qt and Qt doesn't have such settings, meaning I might have to implement my own QTextBox.

gxalpha:
Added a three second `Sending...` placeholder text after the message got sent as discussed.

UI/window-youtube-actions.cpp
```diff
@@ -497,15 +497,19 @@ bool OBSYoutubeActions::CreateEventAction(YoutubeApiWrappers *api,
...
+		if (!apiYouTube->BindStream(broadcast.id, stream.id, json)) {
 			blog(LOG_DEBUG, "No stream binded.");
 			return false;
 		}
 
-		if (broadcast.privacy != "private")
-			apiYouTube->SetChatId(broadcast.id);
-		else
+		if (broadcast.privacy != "private") {
+			std::string apiLiveChatId =
```
RytoEX, line 507: Could be const here?

RytoEX (changes requested):
Check opportunities for passing parameters as const, ref, or const-ref. I haven't compiled my suggestions, so please double check these when looking into them (no new warnings/problems).

gxalpha:
> Check opportunities for passing parameters as const, ref, or const-ref. I haven't compiled my suggestions, so please double check these when looking into them (no new warnings/problems).

Think I found them all. No new warnings.

UI/data/locale/en-US.ini
```diff
@@ -1319,9 +1319,20 @@ YouTube.Actions.AutoStartStreamingWarning.Title="Manual start required"
...
+YouTube.Chat.Error.Text="The message couldn't be sent: %1"
+
 # YouTube API errors in format "YouTube.Errors.<error reason>"
 YouTube.Errors.liveStreamingNotEnabled="Live streaming is not enabled on the selected YouTube channel.<br/><br/>See <a href='https://www.youtube.com/features'>youtube.com/features</a> for more information."
 YouTube.Errors.livePermissionBlocked="Live streaming is unavailable on the selected YouTube Channel.<br/>Please note that it may take up to 24 hours for live streaming to become available after enabling it in your channel settings.<br/><br/>See <a href='https://www.youtube.com/features'>youtube.com/features</a> for details."
 YouTube.Errors.errorExecutingTransition="Transition failed due to a backend error. Please try again in a few seconds."
 YouTube.Errors.errorStreamInactive="YouTube is not receiving data for your stream. Please check your configuration and try again."
 YouTube.Errors.invalidTransition="The attempted transition was invalid. This may be due to the stream not having finished a previous transition. Please wait a few seconds and try again."
+# Chat errors
+YouTube.Errors.liveChatDisabled="The live chat has been disabled on this stream."
```
WizardCM, line 1335:
```suggestion
YouTube.Errors.liveChatDisabled="Live chat is disabled on this stream."
```

UI/data/locale/en-US.ini
```diff
@@ -1319,9 +1319,20 @@ YouTube.Actions.AutoStartStreamingWarning.Title="Manual start required"
...
+YouTube.Errors.liveChatEnded="The live chat is no longer live."
```
WizardCM, line 1336:
```suggestion
YouTube.Errors.liveChatEnded="Live stream has ended."
```

WizardCM (approved):

Warchamp7 (approved):

RytoEX (approved):
Finally tested this, and it works as advertised. May double check with Rich on one item before merging, but this is otherwise good to go.

PR #6445, opened 2022-05-09
UI: Add support for backup/redundant streams

obinnaokechukwu:
**Description**
This change adds support for having simultaneous streams to multiple backends of the same streaming service provider.

**Motivation and Context**
To enhance the reliability of a stream, it is sometimes helpful to send multiple (exact) copies of the same video stream to various backends of the same streaming platform. The redundant copy of the video can then be used as a backup when there is a problem with the main video stream.

This change adds the ability to send redundant copies of a given stream to the same streaming service provider.

Redundant streams can be enabled by specifying a `backup_urls` field in the services.json file and adding a url to it.
```json
                {
                    "name": "Server 1 and Server 2",
                    "url": "rtmps://server1.foo.bar",
                    "backup_urls": [
                        "rtmps://server2.foo.bar"
                    ]
                }
```

**How Has This Been Tested?**
Verified changes work in the dual-ingestion-final branch.

**Types of changes**

**Checklist:**
 - [x]  My code has been run through [clang-format](https://github.com/obsproject/obs-studio/blob/master/.clang-format).
- [x]  I have read the [contributing document](https://github.com/obsproject/obs-studio/blob/master/CONTRIBUTING.rst).
- [x]  The code has been tested.
- [x]  All commit messages are properly formatted and commits squashed where appropriate.


RytoEX (changes requested):
Per our [Contributing guildelines](https://github.com/obsproject/obs-studio/blob/b3870476324e2dc79a9d30371a1199376bcd0ad7/CONTRIBUTING.rst), please make these changes:
1. Do not use periods at the end of commit message subjects (the first line of commit messages).
2. Do not exceed 70 characters for total commit message subject length. Try to keep the text after the prefix to 50 characters.
3. Wrap commit message bodies to 72 characters. Do not exceed 72 characters per line in the commit message body.

Also, please rebase this on the latest git HEAD and resolve any merge conflicts.

obinnaokechukwu:
Thanks for taking a look. I have fixed the formatting and rebased to HEAD.

tytan652:
Firstly, I want to specify I have nothing against this feature. But for me backup services should be generated at the last moment (when the stream starts) by the service itself if the UI ask for it:

The actual UI generate a service settings "object" and then create a service "object" (obs_service_t) with it. And then this service is *normally* manipulated only with libobs Service API functions.

In my opinion the feature should be something like:

The backup URLs became a settings of the service like the URL.

When the stream starts, the UI ask with a libobs function to the service "object" (if there is backup URLs) to generate copy of itself with the server URL replace with a backup one (if there is backup URLs).

And the the UI starts streams for each of those services.

The functions that generate "backup services" should maybe be similar to libobs enum functions and his implementation is provided through registered `obs_service_info`.

Here is a poorly made diagram:
```mermaid
sequenceDiagram
    participant Service UI
    participant service
    Note over service: obs_service_t
    participant Output UI
    Service UI->>service: Create service from JSON settings
    Note right of Service UI: which contain backup URLs
    Output UI->>service: Ask for backup url services if backup URLs
    Note right of service: With a libobs function
    alt there is
        service->>Output UI: Send services copy with backup URL
        Output UI->>Output UI: Start streams with the standard and backup services
        Note right of Output UI: The backup URL replace the original URL
    else there is not
        Output UI->>Output UI: Start stream with the standard service
    end
```

obinnaokechukwu:
> My question would be does it care about timestamps after a disconnect? If it does, then what happens in this scenario? If it doesn't, then I assume it just starts using the alternate stream's data where ever it currently is, correct? I feel like the latter scenario has to be the more likely way that the server would have to handle it, right? Just start reading the alternating stream's data where ever it is, right?


The intent of the implementation was to be protocol agnostic and not specific to RTMP, so it spins up several obs_output_t on startstream() and lets them push the same stream to the server. The server then decides how to handle the redundancy. In our case, the system makes no assumptions about timestamps being aligned across multiple streams.

> What interruption scenarios does this specifically handle? I feel like the only scenario this properly covers is when the ingest or nodes to the ingest have to disconnect the user for whatever reason. It would not handle local internet issues because then both streams would likely have issues, correct? 

You are correct that ideally the copies are pushed via distinct ISP connections, however the stream ingestions are targeted to distinct geographic regions,  so it provides protection from any  network/system issues after the two paths split.

obinnaokechukwu:
Thanks for your comments and suggestions! I'll take a look at the potential issues. 

On the first 3 points mentioned (settings, auto config, and storage), I've actually had earlier discussions about these specific points in which it was either directly suggested or okay-ed to take this approach (since I wanted to make sure the design is satisfactory enough before going ahead with its implementation). The point you brought up about removing access to backup services from the frontend-api sounds good, I can move that into the window-basic-main and access the outputs from there if that's okay.

Also, looking at the RFCs you referenced, I'm not actually sure this PR fundamentally makes the RFCs more difficult to implement. And unless I'm missing something here, it might even simplify things a bit. Regarding timelines, since the RFCs mentioned haven't been fully implemented, wouldn't it be fair to say that this PR can go in first? Especially since the actual details of the RFCs are still being ironed out and they seem to have been open for a fair amount of time and we also don't have a landing date at the moment.

Thanks again for your comments and suggestions.

tytan652:
> Also, looking at the RFCs you referenced, I'm not actually sure this PR fundamentally makes the RFCs more difficult to implement. And unless I'm missing something here, it might even simplify things a bit. Regarding timelines, since the RFCs mentioned haven't been fully implemented, wouldn't it be fair to say that this PR can go in first? Especially since the actual details of the RFCs are still being ironed out and they seem to have been open for a fair amount of time and we also don't have a landing date at the moment.

Don't worry about this part of my comment.

obinnaokechukwu:
I've addressed the issues you brought up -- outputs api removal, translatable text, dynamic stats page and output disconnection.

UI/window-basic-main-outputs.cpp
```diff
@@ -761,90 +765,109 @@ bool SimpleOutput::SetupStreaming(obs_service_t *service)
...
+			type = "rtmp_output";
+			const char *url = obs_service_get_url(service);
+			if (url != NULL && strncmp(url, FTL_PROTOCOL,
+						   strlen(FTL_PROTOCOL)) == 0) {
+				type = "ftl_output";
+			} else if (url != NULL &&
+				   strncmp(url, RTMP_PROTOCOL,
+					   strlen(RTMP_PROTOCOL)) != 0) {
+				type = "ffmpeg_mpegts_muxer";
+			}
```
tytan652, line 783: I think primary and backup server share the same type of output, the type has no need to be set multiple times.

UI/window-basic-main-outputs.cpp
```diff
@@ -761,90 +765,109 @@ bool SimpleOutput::SetupStreaming(obs_service_t *service)
...
+		/* XXX: this is messy and disgusting and should be refactored */
+		if (outputType != type) {
+			if (resetOutputs) {
+				streamDelayStarting.Disconnect();
+				streamStopping.Disconnect();
+				startStreaming.Disconnect();
+				stopStreaming.Disconnect();
+				streamOutputs.clear();
+				resetOutputs = false;
+			}
```
tytan652, line 795: Outputs and their signals need to get cleared if the user switch to a service without redundancy.

UI/window-basic-main-outputs.cpp
```diff
@@ -1702,98 +1772,119 @@ bool AdvancedOutput::SetupStreaming(obs_service_t *service)
...
+			type = "rtmp_output";
+			const char *url = obs_service_get_url(service);
+			if (url != NULL && strncmp(url, FTL_PROTOCOL,
+						   strlen(FTL_PROTOCOL)) == 0) {
+				type = "ftl_output";
+			} else if (url != NULL &&
+				   strncmp(url, RTMP_PROTOCOL,
+					   strlen(RTMP_PROTOCOL)) != 0) {
+				type = "ffmpeg_mpegts_muxer";
+			}
```
tytan652, line 1790: Same as simple output.

UI/window-basic-main-outputs.cpp
```diff
@@ -1702,98 +1772,119 @@ bool AdvancedOutput::SetupStreaming(obs_service_t *service)
...
+		if (outputType != type) {
+			if (resetOutputs) {
+				// Only reset output once.
+				streamOutputs.clear();
+				streamDelayStarting.Disconnect();
+				streamStopping.Disconnect();
+				startStreaming.Disconnect();
+				stopStreaming.Disconnect();
+				resetOutputs = false;
+			}
```
tytan652, line 1803: Same as simple output

tytan652 (commented):
I think some things have no need to be set multiple time in the `for` loop.

Also audio encoder creation should be done once, and used by all outputs.

UI/window-basic-main-outputs.cpp
obinnaokechukwu: Moved this out of loop.

UI/window-basic-main-outputs.cpp
obinnaokechukwu: refactored this

UI/window-basic-main-outputs.cpp
obinnaokechukwu: Moved this out of loop.

UI/window-basic-status-bar.cpp
```diff
@@ -461,7 +461,11 @@ void OBSBasicStatusBar::StreamDelayStarting(int sec)
 	if (!main || !main->outputHandler)
 		return;
 
-	streamOutput = main->outputHandler->streamOutput;
+	if (!main->outputHandler->streamOutputs.empty()) {
+		streamOutput = main->outputHandler->streamOutputs[0];
+	} else {
+		streamOutput = nullptr;
+	}
```
norihiro, line 468:
```suggestion
	if (!main->outputHandler->streamOutputs.empty())
		streamOutput = main->outputHandler->streamOutputs[0];
	else
		streamOutput = nullptr;
```

UI/window-basic-stats.cpp
```diff
@@ -485,11 +509,24 @@ void OBSBasicStats::Reset()
...
 
-	outputLabels[0].Reset(strOutput);
-	outputLabels[1].Reset(recOutput);
+	int recordingIndex = (int)outputLabels.size() - 1;
+	int numOutputs = (int)streamingOutputs.size();
+
+	for (int i = 0; i < recordingIndex; i++) {
+		if (i < numOutputs) {
+			outputLabels[i].Reset(streamingOutputs[i]);
+		}
```
norihiro, line 525:
```suggestion
		if (i < numOutputs)
			outputLabels[i].Reset(streamingOutputs[i]);
```

UI/window-basic-stats.cpp
```diff
@@ -429,12 +446,19 @@ void OBSBasicStats::Update()
...
 	/* recording/streaming stats                   */
+	int recordingIndex = (int)outputLabels.size() - 1;
+	int numOutputs = (int)streamingOutputs.size();
 
-	outputLabels[0].Update(strOutput, false);
-	outputLabels[1].Update(recOutput, true);
+	for (int i = 0; i < recordingIndex; i++) {
+		if (i < numOutputs) {
+			outputLabels[i].Update(streamingOutputs[i], false);
+		}
```
norihiro, line 455:
```suggestion
		if (i < numOutputs)
			outputLabels[i].Update(streamingOutputs[i], false);
```

norihiro: or, maybe `i < recordingIndex && i < numOutputs` at the 2nd sentence of the `for` statement.

UI/window-basic-stats.cpp
```diff
@@ -287,11 +303,12 @@ void OBSBasicStats::Update()
...
 	obs_get_video_info(&ovi);
 
-	OBSOutputAutoRelease strOutput = obs_frontend_get_streaming_output();
+	auto streamingOutputs = main->GetStreamingOutputs();
 	OBSOutputAutoRelease recOutput = obs_frontend_get_recording_output();
 
-	if (!strOutput && !recOutput)
+	if (streamingOutputs.empty() && !recOutput) {
 		return;
+	}
```
norihiro, line 311:
```suggestion
	if (streamingOutputs.empty() && !recOutput)
		return;
```

UI/window-basic-main.cpp
```diff
@@ -6936,7 +6998,9 @@ void OBSBasic::StreamingStart()
 	ui->streamButton->setText(QTStr("Basic.Main.StopStreaming"));
 	ui->streamButton->setEnabled(true);
 	ui->streamButton->setChecked(true);
-	ui->statusbar->StreamStarted(outputHandler->streamOutput);
+	if (!outputHandler->streamOutputs.empty()) {
+		ui->statusbar->StreamStarted(outputHandler->streamOutputs[0]);
+	}
```
norihiro, line 7003:
```suggestion
	if (!outputHandler->streamOutputs.empty())
		ui->statusbar->StreamStarted(outputHandler->streamOutputs[0]);
```

UI/window-basic-main.cpp
```diff
@@ -4228,20 +4274,31 @@ void OBSBasic::RenderMain(void *data, uint32_t cx, uint32_t cy)
...
+	return services;
+}
+
+std::vector<OBSOutput> OBSBasic::GetStreamingOutputs()
+{
+	std::vector<OBSOutput> outputs;
+	for (OBSOutputAutoRelease &output : outputHandler->streamOutputs) {
+		outputs.push_back(output.Get());
 	}
```
norihiro, line 4293:
```suggestion
	for (OBSOutputAutoRelease &output : outputHandler->streamOutputs)
		outputs.push_back(output.Get());
```

UI/window-basic-main.cpp
```diff
@@ -1239,11 +1240,54 @@ bool OBSBasic::LoadService()
...
+	// If service contains no backup servers, return original service.
+	OBSDataAutoRelease settings = obs_service_get_settings(service);
+	const char *srv = obs_data_get_string(settings, "server");
+	OBSDataArrayAutoRelease backup_servers =
+		obs_data_get_array(settings, "backup_servers");
+	size_t backup_count = obs_data_array_count(backup_servers);
+
+	if (backup_count == 0) {
+		return services;
+	}
```
norihiro, line 1268:
```suggestion
	if (backup_count == 0)
		return services;
```

UI/window-basic-main.cpp
```diff
@@ -1239,11 +1240,54 @@ bool OBSBasic::LoadService()
...
-	return !!service;
+	return !services.empty();
+}
+
+std::vector<OBSService> OBSBasic::ExpandService(obs_service_t *service)
+{
+	std::vector<OBSService> services;
+	if (!service) {
+		return services;
+	}
```
norihiro, line 1255:
```suggestion
	if (!service)
		return services;
```

UI/window-basic-auto-config.cpp
```diff
@@ -892,14 +912,22 @@ void AutoConfig::SaveStreamSettings()
...
 	OBSDataAutoRelease settings = obs_data_create();
 
 	if (!customServer)
 		obs_data_set_string(settings, "service", serviceName.c_str());
 	obs_data_set_string(settings, "server", server.c_str());
+
+	QJsonArray qBackupServers;
+	for (size_t i = 0; i < backupServers.size(); i++) {
+		qBackupServers.append(QString::fromStdString(backupServers[i]));
+	}
```
norihiro, line 927:
```suggestion
	for (size_t i = 0; i < backupServers.size(); i++)
		qBackupServers.append(QString::fromStdString(backupServers[i]));
```

UI/window-basic-auto-config-test.cpp
```diff
@@ -254,89 +274,135 @@ void AutoConfigTestPage::TestBandwidthThread()
...
 	bool success = false;
-
 	for (size_t i = 0; i < servers.size(); i++) {
 		auto &server = servers[i];
 
-		connected = false;
-		stopped = false;
+		for (auto &wrapper : outputs) {
+			wrapper.state = OutputWrapper::OutputState::Default;
+		}
```
norihiro, line 405:
```suggestion
		for (auto &wrapper : outputs)
			wrapper.state = OutputWrapper::OutputState::Default;
```

UI/window-basic-auto-config-test.cpp
```diff
@@ -345,28 +411,68 @@ void AutoConfigTestPage::TestBandwidthThread()
...
+			obs_data_set_string(service_settings, "server",
+					    urls[j].c_str());
+			obs_service_update(services[j], service_settings);
+		}
+
+		bool start_failed = false;
+
+		for (size_t j = 0; j < urls.size(); j++) {
+			start_failed |= !obs_output_start(outputs[j].output);
+		}
```
norihiro, line 433:
```suggestion
		for (size_t j = 0; j < urls.size(); j++)
			start_failed |= !obs_output_start(outputs[j].output);
```

UI/window-basic-auto-config-test.cpp
```diff
@@ -376,47 +482,90 @@ void AutoConfigTestPage::TestBandwidthThread()
...
+		for (size_t j = 0; j < urls.size(); j++) {
+			if (outputs[j].state ==
+			    OutputWrapper::OutputState::Started) {
+				obs_output_stop(outputs[j].output);
+			}
+		}
+
+		while (not_stopped(urls)) {
+			cv.wait(ul);
+		}
```
norihiro, line 531:
```suggestion
		while (not_stopped(urls))
			cv.wait(ul);
```

UI/streaming-helpers.cpp
```diff
@@ -181,7 +245,27 @@ void StreamSettingsUI::UpdateServerList()
...
+
+		for (auto &url : urls) {
+			qbackup_servers.append(
+				QString::fromStdString(url.string_value()));
+		}
+
+		std::vector<std::string> backup_servers;
+		for (auto &url : urls) {
+			backup_servers.push_back(url.string_value());
+		}
```
norihiro, line 260:
```suggestion
		for (auto &url : urls)
			backup_servers.push_back(url.string_value());
```

UI/api-interface.cpp
```diff
@@ -385,7 +385,10 @@ struct OBSStudioAPI : obs_frontend_callbacks {
 
 	obs_output_t *obs_frontend_get_streaming_output(void) override
 	{
-		OBSOutput output = main->outputHandler->streamOutput.Get();
+		if (main->outputHandler->streamOutputs.empty()) {
+			return nullptr;
+		}
```
norihiro, line 390:
```suggestion
		if (main->outputHandler->streamOutputs.empty())
			return nullptr;
```

UI/streaming-helpers.cpp
```diff
@@ -68,6 +74,64 @@ Json get_service_from_json(const Json &root, const char *name)
...
+
+std::string create_server_id(const std::string &server,
+			     const std::vector<std::string> &backupServers)
+{
+	std::stringstream id;
+
+	id << server;
+	for (const auto &backup_server : backupServers) {
+		id << "," << backup_server;
+	}
```
norihiro, line 114:
```suggestion
	for (const auto &backup_server : backupServers)
		id << "," << backup_server;
```

norihiro (commented):
https://github.com/torvalds/linux/blob/master/Documentation/process/coding-style.rst
> Do not unnecessarily use braces where a single statement will do.

The coding guideline request not to have unnecessary braces.

PatTheMav:
Closing because this PR is now 4 years old and would probably have to be redesigned with the current state of outputs in mind (particularly multi-track encoding).

PR #6446, opened 2022-05-10
Update adv-audio-control.cpp

cloudhzc:
A new line? Why?


RytoEX:
The clang-format rules are why.
image

cloudhzc:
ok

PR #6452, opened 2022-05-11
win-capture: Fix bug sometimes UWP window can't be found

walker-WSH:
### Description

Some UWP window can't be found while using WGC mode.
Such as Microsoft.Photo window.

According to my test result, this situation can't be reproduced on every PC, not related with OS version.
As the issue, Microsoft.Photo can be found in the window list. But when you choose this item, window can't be found in window-plugin.Tick (using Auto/WGC capture mode).

### Motivation and Context

find uwp window correctly

### How Has This Been Tested?

add window source and use WGC capture mode;
choose Microsoft.Photo/Calculate.exe;
check result: no video is captured even photo window is not minimized;

### Types of changes

### Checklist:

### DIFF
--- a/plugins/win-capture/window-helpers.c
+++ b/plugins/win-capture/window-helpers.c
@@ -509,8 +509,10 @@ BOOL CALLBACK enum_windows_proc(HWND window, LPARAM lParam)
 	if (IsWindowCloaked(window))
 		return TRUE;
 
-	const int rating = window_rating(window, data->priority, data->class,
-					 data->title, data->exe,
+	HWND check_rate = is_uwp_window(window) ? get_uwp_actual_window(window)
+						: window;
+	const int rating = window_rating(check_rate, data->priority,
+					 data->class, data->title, data->exe,
 					 data->generic_class);
 	if (rating < data->best_rating) {
 		data->best_rating = rating;


RytoEX:
I can find and capture Photos just fine. Is there a particular scenario in which it cannot be found in the window list?

walker-WSH:
> I can find and capture Photos just fine. Is there a particular scenario in which it cannot be found in the window list?

According to my test result, this situation can't be reproduced on every PC.

As the issue, Microsoft.Photo can be found in the window list. But when you choose this item, window can't be matched in window-plugin.Tick.

RytoEX:
Could you please provide some detailed reproduction steps for reproducing the issue? We're having trouble reproducing the issue, and we'd like to confirm the behavior before and after this PR.

walker-WSH:
> Could you please provide some detailed reproduction steps for reproducing the issue? We're having trouble reproducing the issue, and we'd like to confirm the behavior before and after this PR.

the steps is so simple. add window source and choose photo window & WGC method，then check captured view.

this seems to be related with environment, sorry I have no more tips. I verified it on 5 PC，only 2 computers hold this bug.

even photo window is normally captured on your PC，if you debug the window_rating, you can find only title is matched since exe and class are gotten from host window of UWP

walker-WSH:
I am not sure if you can understand my bad English. Maybe @jpark37 or @jp9000 knows me ？

RytoEX:
Just to confirm: does this actually happen in OBS Studio, or does this only happen in other apps that load this plugin?

walker-WSH:
> Just to confirm: does this actually happen in OBS Studio, or does this only happen in other apps that load this plugin?

In OBS, including master branch

walker-WSH:
hi  @RytoEX    

Now this issue can be reproduced 100% on my WIN11 PC
here is detail for version and recording:

[2023-07-27 15-14-38.zip](https://github.com/obsproject/obs-studio/files/12180391/2023-07-27.15-14-38.zip)

PatTheMav:
The underlying implementation has changed since this PR was authored (the window-helper has moved to `libobs`) and was also changed to accommodate UWP applications as well.

As such this PR attempts to change a non-existing/moved file.

Please check if the issue persists with OBS v30 and open a new issue with reproduction steps or a new PR with mitigating code changes.

PR #6453, opened 2022-05-11
UI: Update show/hide transitions for all selected items

cg2121:
### Description
Instead of just updating the currently selected scene item's visibility
transition, when updating show/hide transition or duration, this updates
all of the selected items.

### Motivation and Context
Idea from @Warchamp7 

### How Has This Been Tested?
Made sure when updating show/hide transition, it affected all selected items.

### Types of changes

### Checklist:


Warchamp7:
The transition choice seems to apply to all selected items, but not the configured options.

Ex. Applying a Slide transition to multiple selected items will put Slide on all of them, but only the one that was clicked will get configuration changes, such as swapping the slide direction in the transition settings.


Also got a crash while testing, trying to narrow it down

cg2121:
Marking as draft, as it needs work with undo/redo. It undo/redos each source at a time instead of all of the selected ones.

cg2121:
This now properly works with undo/redo.

Warchamp7:
Looks to properly copy and paste transitions, but duration is incorrect. It always uses the global transition duration. However, that is not unique to this PR so may be out of scope to fix here.

Warchamp7:
Will see if the above issue is solved naturally by #10511

Warchamp7 (changes requested):
Issues still to be addressed:
- Copy duration as well when copying transition
- Don't display a checkmark in the list when the selected sources have mismatched selections

Warchamp7:
This PR is being closed due to being stale. If you would still like to submit these changes, please open a new PR.

PR #6464, opened 2022-05-13
Accept Qt 5 or Qt 6 in CopyMSVCBins and other improvements

RytoEX:
### Description

Three commits for Qt 5/6 compatibility and some general tidying in cmake/CopyMSVCBins.cmake:
1. Make CopyMSVCBins.cmake work with either Qt 5 or Qt 6 depending on what version the user (builder) has chosen.
2. Resolve the relative path to an absolute path once and then use that for all other Qt file paths.
3. Use correct capitalization for Qt in messages.

### Motivation and Context

Want to be able to build against Qt 5 or Qt 6. Want cleaner paths in CMake log output instead of a triple parent directory path.

### How Has This Been Tested?

I've tested this locally against Qt 6. CI built against Qt 5.

### Types of changes

### Checklist:


cmake/Modules/CopyMSVCBins.cmake
```diff
@@ -179,57 +179,61 @@ file(GLOB RNNOISE_BIN_FILES
      "${RNNOISE_INCLUDE_DIR}/../bin${_bin_suffix}/rnnoise*.dll"
      "${RNNOISE_INCLUDE_DIR}/../bin/rnnoise*.dll")
 
+set(QtCore_DIR "${Qt${QT_VERSION}Core_DIR}")
+cmake_path(SET QtCore_DIR_NORM NORMALIZE "${QtCore_DIR}/../../..")
```
PatTheMav, line 183: Is `cmake_path` ensured to always resolve a path with a final forward slash?

RytoEX: In my testing on Windows, it appeared to always end with a final forward slash, although that isn't reflected in the documentation. If we want to be really sure that there isn't one, we can try to strip slashes off the right side, but it might be annoying to do so.

RytoEX: If the concern about the slashes is blocking, I can split the cosmetic changes into a separate PR for further examination.

PatTheMav: Nah, fine be me. If it breaks, we'll fix it.

PR #2969, opened 2020-05-21
 UI: Add indicators for current preview/program scene

cg2121:
### Description
Add indicator for preview/program scene in studio mode.

Screenshot from 2021-01-16 05-08-56

### Motivation and Context
Makes it easier to know what scene is the preview or program scene.

### How Has This Been Tested?
Switched scenes in studio mode and made sure the icon was put in the right place.

### Types of changes
- New feature (non-breaking change which adds functionality)

### Checklist:
- [x] My code has been run through [clang-format](https://github.com/obsproject/obs-studio/blob/master/.clang-format).
- [x] I have read the [**contributing** document](https://github.com/obsproject/obs-studio/blob/master/CONTRIBUTING.rst).
- [x] My code is not on the master branch.
- [x] The code has been tested.
- [x] All commit messages are properly formatted and commits squashed where appropriate.
- [x] I have included updates to all appropriate documentation.

PR #6472, opened 2022-05-14
UI: Add scene names to preview/program labels

cg2121:
### Description
Makes it easier for users to tell which scene is in
preview and program.

Capture

### Motivation and Context
https://ideas.obsproject.com/posts/1504/please-show-the-scene-name-in-studio-mode

### How Has This Been Tested?
Changed scenes to make sure labels updated properly.

### Types of changes

### Checklist:


WizardCM:
Minor potential concerns with this:
1. as the length of the name will change, this will result in the text jumping around (including the text that doesn't change). However, I'm not a fan of potentially left-aligning it.
2. Does this correctly clip the text and add an ellipsis if the source name is too long to fit? In my opinion, this should only clip the actual scene name.

UI/window-basic-main-transitions.cpp
```diff
@@ -1865,3 +1868,27 @@ int OBSBasic::GetOverrideTransitionDuration(OBSSource source)
 
 	return (int)obs_data_get_int(data, "transition_duration");
 }
+
+void OBSBasic::UpdateStudioModeSceneLabels()
+{
+	if (!IsPreviewProgramMode())
+		return;
+
+	QString program = QTStr("StudioMode.Program");
```
WizardCM, line 1877: Would this function be cleaner with QString `.arg()`?

cg2121: We would need to add a new string to the en-US.ini file, so I feel like this is just fine.

cg2121:
> 2\. Does this correctly clip the text and add an ellipsis if the source name is too long to fit? In my opinion, this should only clip the actual scene name.

The updated code clips the whole label, so it doesn't resize the preview or program.

UI/window-basic-main-transitions.cpp
WizardCM: Not really. `QString("%1 - %2").arg(program, programSceneName)`

cg2121: Updated to use the args.

WizardCM (approved):

lindenkron:
Will there be an option to disable labels? This takes up a non insignificant amount of screen monopoly - something that isn't worth giving up to a lot of people that have no issues knowing which is program and preview - or don't use studio mode at all. 

Fitting docks & widgets it a never ending battle.

WizardCM:
@lindenkron Yes, as it's tied to the existing label, you can hide using Settings -> General -> Studio Mode -> Show preview/program labels.

derrod:
Rather than `Program - <name>` how about `Program\n<Name>` and using a smaller font size for the name?

cg2121:
> Rather than `Program - <name>` how about `Program\n<Name>` and using a smaller font size for the name?

I just feel it would take up too much vertical space.

sakaeshalom:
After switching the program by clicking `Transition`, the name is not updated.
In our settings, `SwapScenes` is disabled.

sakaeshalom:
Instead of calling `UpdateStudioModeSceneLabels` from `UpdateContextBar`, the function should be called when changing program scene or preview scene. How about calling the function `UpdateStudioModeSceneLabels` from `TransitionToScene` and `SetCurrentScene`?

```patch
diff --git a/UI/window-basic-main-transitions.cpp b/UI/window-basic-main-transitions.cpp
index e3200dffe..d225b9060 100644
--- a/UI/window-basic-main-transitions.cpp
+++ b/UI/window-basic-main-transitions.cpp
@@ -436,6 +436,8 @@ void OBSBasic::TransitionToScene(OBSSource source, bool force,
                        TransitionFullyStopped();
        }
 
+       UpdateStudioModeSceneLabels();
+
 cleanup:
        if (usingPreviewProgram && sceneDuplicationMode)
                obs_scene_release(scene);
@@ -745,6 +747,8 @@ void OBSBasic::SetCurrentScene(OBSSource scene, bool force)
                }
        }
 
+       UpdateStudioModeSceneLabels();
+
        UpdateContextBar(true);
 
        if (scene) {
diff --git a/UI/window-basic-main.cpp b/UI/window-basic-main.cpp
index 329ca4fe3..e18a2a541 100644
--- a/UI/window-basic-main.cpp
+++ b/UI/window-basic-main.cpp
@@ -3073,8 +3073,6 @@ void OBSBasic::UpdateContextBarDeferred(bool force)
 
 void OBSBasic::UpdateContextBar(bool force)
 {
-       UpdateStudioModeSceneLabels();
-
        if (!ui->contextContainer->isVisible() && !force)
                return;
 
```
I tested with both cases that `SwapScenes` is disabled and enabled.

UI/window-basic-main-transitions.cpp
```diff
@@ -1865,3 +1867,23 @@ int OBSBasic::GetOverrideTransitionDuration(OBSSource source)
...
 }
+
+void OBSBasic::UpdateStudioModeSceneLabels()
+{
+	if (!IsPreviewProgramMode())
+		return;
+
+	QString preview = QString("%1 - %2").arg(
+		QTStr("StudioMode.Preview"),
+		QT_UTF8(obs_source_get_name(GetCurrentSceneSource())));
```
gxalpha, line 1878: I believe this could cause issues in right-to-left languages.
It'd probably be better to have something along the lines of 
```
QString preview = QTStr("StudioMode.Preview.Scene").arg(obs_source_get_name(GetCurrentSceneSource()));
```
where the string in the ini file would be
```
StudioMode.Preview.Scene="Preview - %1"
```
A separate string is required because the existing one is used in other places as well.

gxalpha (changes requested):

cg2121:
Updated to fix labels not updating when swap scenes is disabled.

UI/data/locale/en-US.ini
```diff
@@ -86,6 +86,8 @@ HideTransition="Hide Transition"
 None="None"
 StudioMode.Preview="Preview"
 StudioMode.Program="Program"
+StudioMode.PreviewSceneName="Preview - %1"
```
OldBaldGeek, line 89: Using a colon instead of a dash "Preview: %1" and "Program: %1" would make it clearer that the scene name is information about the Preview or Program. It also saves a space, if you are worried about long names

OldBaldGeek:
sakaeshalom's proposed change is an improvement on the original, but if you move the manual transition fader partway to the right, it changes the Program name immediately - and leaves it changed (and thus incorrect) if you move the fader back to the left.

An easy solution is to move the call of UpdateStudioModeSceneLabels from line 440 in TransitionToScene to line 399, where OBS_FRONTEND_EVENT_SCENE_CHANGED is generated. The downside is that the Program title won't change until a transition (fader or automatic) completes. I would argue that title changes should correspond to the generation of OBS_FRONTEND_EVENT_SCENE_CHANGED and OBS_FRONTEND_EVENT_PREVIEW_SCENE_CHANGED.

UI/data/locale/en-US.ini
gxalpha: I think the dash look better. We're not worried about long names, but about ones spanning multiple lines.

OldBaldGeek: Just my suggestion. I'll be happy to see the titles regardless of punctuation.
The colon format is used by "Star Wars: the Last Jedi" and over a dozen items in en-US.ini, such as
- QuickTransitions.HotkeyName="Quick Transition: %1"
- Basic.Settings.Stream.Recommended.MaxVideoBitrate="Maximum Video Bitrate: %1 kbps"
- Basic.Settings.Advanced.StreamDelay.MemoryUsage="Estimated Memory Usage: %1 MB"
- YouTube.AuthError.Text="Failed to get channel information: %1."

cg2121:
Updated to use colon instead of hyphen. Also made the labels left aligned.

cg2121:
> sakaeshalom's proposed change is an improvement on the original, but if you move the manual transition Tbar partway to the right, it changes the Program name immediately - and leaves it changed (and thus incorrect) if you move the Tbar back to the left. Actually, that will probably happen with the original version as well.
> 
> This problem is due to a defect in the TBar logic, and affects any code that calls GetProgramSource() or obs_frontend_get_current_scene after such a TBar movement. TBar issue filed as [https://github.com/obsproject/obs-studio/issues/6586](url)
> 
> As a workaround, this PR could move the call of UpdateStudioModeSceneLabels from line 440 in TransitionToScene to line 399, where OBS_FRONTEND_EVENT_SCENE_CHANGED is generated. The downside is that the Program title won't change until a transition (TBar or automatic) _completes_, rather than when it begins. This matches when OBS_FRONTEND_EVENT_SCENE_CHANGED and OBS_FRONTEND_EVENT_PREVIEW_SCENE_CHANGED are generated. Not ideal, but preferable to showing the wrong scene name.

This doesn't happen anymore with the updated code.

Fenrirthviti:
Your suggestions make a lot of assumptions based on your experience that don't match with our observations on how people use OBS.

> * Other titles on the main window (Scene list, Source list etc) are centered, so this isn't consistent

Comparing a title bar to a label is not really a valid comparison.

> * Most scene titles are likely to be short. Left justifying them just looks odd

In practice, we see that people often have very long scene names.
> * The "jumping around" as length changes argument is dubious: with the scene image under the title changing, the change in the title length isn't really noticeable.

It's far more visually consistent to be left aligned, so you're not having to scan the whole label for where the scene name actually starts. It might "feel weird" at first, but it's a far better experience for the intended purpose, which is a quick glance to see what scene is active.

> * Text overflow is unlikely unless you are running OBS in a small window (unlikely for Studio Mode) and have very long scene names.

Again, not really as unlikely as you think.

> * If you are worried about text overflow, the TruncateLabel function added by [UI: Truncate names in advanced audio dialog #6490](https://github.com/obsproject/obs-studio/pull/6490) seems like a better solution

I'm pretty sure the text will already overflow properly here, given previous comments about it.

cg2121:
I've now combined it with #2969, as they are doing similar things.

studio-mode

I've now also enabled showing the preview/indicator in normal mode, to make the UI more consistent.

normal-mode

teeks99:
> I've now combined it with #2969, as they are doing similar things.
> 
> studio-mode
> 
> I've now also enabled showing the preview/indicator in normal mode, to make the UI more consistent.
> 
> normal-mode

I like the red/green indicators, but that can be tricky for people with red/green colorblindness. Can the shape be updated to something unique? Maybe a circle for the active one and a square for the preview?

OldBaldGeek:
When not in Studio Mode, shouldn't the title say "PROGRAM" and use a red indicator, rather then "Preview" with a green?
The view shows what is going out on the stream.

gxalpha:
Preview is the term we commonly use for the area where the user can make scene edits, even in non-studio mode. It behaves exactly like the preview in studio mode (because it is the same). It should therefore be called Preview here as well.

Unrelated, I think adding the indicators to this PR might have feature-creeped it. Not sure I’m a fan of this, especially since this PR was in a pretty good state already, even though I see that the two changes are closely related.

> I've now also enabled showing the preview/indicator in normal mode, to make the UI more consistent

I’m wondering what’s the use case here. I find it much less likely to confuse scenes when in normal mode, so we’re unnecessarily taking away vertical space for folks who don’t know the setting to turn it off exists, and there’s be no way to have it in studio mode only (don’t get me wrong, I’m not suggesting to add such a setting). On the other hand, I might be starting to bikeshed again.

Fenrirthviti:
I also agree the other changes should be split back out in to their own PRs, as it will make individual reviews and discussion easier.

I am, however, personally opposed to the idea of adding the label to non-Studio mode.

Warchamp7:
I still retain my previous feedback of wanting to use an icon rather than a colored square, both for accessibility and clearer communication of what this icon is.

cg2121:
I've reverted it back to just the scene name labels.

gxalpha (changes requested):
The preview label (and for consistency reasons also the program label) needs a few pixels of margin/padding on the left, so that it doesn't directly stick to the window border.

image

Rest seems to work fine.

cg2121:
I've now put a 4px margin on the left and right side of the central widget.

Capture

gxalpha (approved):
just tested, looks good to me

Warchamp7 (changes requested):
Scene name in Preview doesn't update when renaming the scene

We should reflect the rename immediately in Program too even though _functionally_ it might be a different scene if Duplicate Scenes is enabled, because that is mostly an under the hood thing

UI/window-basic-main.cpp
```diff
@@ -2109,6 +2098,8 @@ void OBSBasic::OnFirstLoad()
 
 	if (showLogViewerOnStartup)
 		on_actionViewCurrentLog_triggered();
+
+	ResetUI();
```
SuslikV, line 2102: Is this general bug fix? Or this is part of the new feature?
(in case of bug fix, can you link the issue #)

cg2121: This is part of the new PR.

cg2121: It has been removed now.

PR #6474, opened 2022-05-14
linux-capture: Implement EGL backend for xcomposite capture

kkartaltepe:
### Description
Enables us to provide xcomposite capture when running on X11 on EGL and makes minor improvements to GLX implementation.

Right now its rebased on the xcomposite rewrite but that should be merged soon and ill rebase this.

### Motivation and Context

Some drivers (such as nvidia) have very limited functionality for pixmap capture such as only allowing a single mapping. On popular compositors such as gnome they capture before OBS has started and so users are unable to capture windows. However these drivers dont have such a restriction when using the EGL extensions, which makes the user experience more pleasant.

### How Has This Been Tested?

Tested on nvidia and intel. Everything seems to be working as expected but we have some extra code for software renders like LLVMPipe which I have not tested on the newer GLX setup or EGL.

### Types of changes

### Checklist:


kkartaltepe:
I have this in WIP since its dependent on the rewrite but its fairly well tested and should be basically done.

GeorgesStavracas:
This needs manual rebase

plugins/linux-capture/xcomposite-input.c
```diff
@@ -712,6 +714,11 @@ static void *xcompcap_create(obs_data_t *settings, obs_source_t *source)
 	s->show_cursor = true;
 	s->strict_binding = true;
 	s->source = source;
+	enum obs_nix_platform_type platform = obs_get_nix_platform();
+	s->egl = platform == OBS_NIX_PLATFORM_X11_EGL;
+	if (s->egl) {
+		s->strict_binding = false;
+	}
```
GeorgesStavracas, line 721: Nitpick: can you remove the curly braces? Does clang-formatter complain about that?

kkartaltepe:
Thanks for the merge :) rebased and removed wip.

deps/glad/CMakeLists.txt
```diff
@@ -30,14 +30,12 @@ if(OS_WINDOWS)
 
 elseif(OS_POSIX AND NOT OS_MACOS)
   find_package(OpenGL REQUIRED)
-  find_package(X11 REQUIRED)
-
-  target_link_libraries(glad PRIVATE X11::X11)
```
GeorgesStavracas, line 35: Is this dep removal possible because of this commit; or is it a separate cleanup? Could you please mention this on the commit message?

libobs-opengl/gl-egl-common.c
```diff
@@ -199,24 +224,39 @@ gl_egl_create_dmabuf_image(EGLDisplay egl_display, unsigned int width,
...
+	if (!init_egl_image_target_texture_2d_ext())
 		return NULL;
-	}
-	const GLuint gltex = *(GLuint *)gs_texture_get_obj(texture);
 
-	gl_bind_texture(GL_TEXTURE_2D, gltex);
-	gl_tex_param_i(GL_TEXTURE_2D, GL_TEXTURE_MIN_FILTER, GL_NEAREST);
-	gl_tex_param_i(GL_TEXTURE_2D, GL_TEXTURE_MAG_FILTER, GL_NEAREST);
+	const EGLAttrib pixmap_attrs[] = {EGL_IMAGE_PRESERVED_KHR, EGL_TRUE,
+					  EGL_NONE};
```
GeorgesStavracas, line 246: Nit: does clang complain if this array is declared as:

```c
const EGLAttrib pixmap_attrs[] = {
	EGL_IMAGE_PRESERVED_KHR,
	EGL_TRUE,
	EGL_NONE,
};
```

?

libobs-opengl/gl-wayland-egl.c
```diff
@@ -383,6 +383,17 @@ static bool gl_wayland_egl_device_query_dmabuf_modifiers_for_format(
...
 
+static struct gs_texture *gl_wayland_egl_device_texture_create_from_pixmap(
+	gs_device_t *device, uint32_t width, uint32_t height,
+	enum gs_color_format color_format, uint32_t target, void *pixmap)
+{
+	struct gl_platform *plat = device->plat;
+
+	return gl_egl_create_texture_from_pixmap(plat->display, width, height,
+						 color_format, target,
+						 (EGLClientBuffer)pixmap);
```
GeorgesStavracas, line 394: Does it make sense to implement this vfunc on Wayland / EGL? I'm about 78% sure EGL will crash.

GeorgesStavracas (changes requested):
This looks mostly good to me, just a few nitpicks and questions

deps/glad/CMakeLists.txt
kkartaltepe: Oh thanks, this was actually just me testing something out, I meant to revert this.

libobs-opengl/gl-egl-common.c
tytan652: I think this is how clang-format 12.0.0 will always format those.

libobs-opengl/gl-wayland-egl.c
kkartaltepe: It shouldnt crash, it should return a null texture because eglCreateImage will fail with BADMATCH since EGL_NATIVE_PIXMAP_KHR shouldnt work on wayland and return EGL_NO_IMAGE (but i havnt tested it). null texture is expected when we fail to import and is used for e.g. pipewire fallback to shm so it should be handled by all callers.

I could just return NULL here to make it more clear.

plugins/linux-capture/xcomposite-input.c
tytan652: With or without is will not complain.

libobs-opengl/gl-egl-common.c
kkartaltepe: Clang-format just leaves this ambiguous. if you dont put an element on the line with the curly brace it should preserve the one element per line format.

GeorgesStavracas (approved):
LGTM

PR #6475, opened 2022-05-14
Au revoir, GLX

GeorgesStavracas:
### Description

Drop all GLX-specific code, and rely only on EGL.

### Motivation and Context

GLX is legacy, and has various quirks and complications that make supporting it a chore. Furthermore, GLX doesn't easily support modern texture-sharing mechanisms (DMA-BUF), which complicates it even more.

Since our EGL code paths have matured over the years, it is time to consider dropping GLX.

### How Has This Been Tested?

With this branch, try and perform routine tasks using X11, and the xcomposite plugins. Everything should works at least equally, or better, than before.

### Types of changes

### Checklist:


plugins/linux-capture/xcomposite-input.c
```diff
@@ -399,30 +381,6 @@ static enum gs_color_format gs_format_from_tex()
...
-	bool is_render_target;
-	bool is_dummy;
-	bool gen_mipmaps;
-
-	gs_samplerstate_t *cur_sampler;
-	struct fbo_info *fbo;
-};
-// End shitty hack.
-
 // XErrorHandler for glXCreatePixmap calls.
```
kkartaltepe, line 384: We can drop the X error handler as well, its not being called in the new egl path.

kkartaltepe:
It looks like i missed an include in my earlier rewrite for the X11 headers which was being covered by the glx headers. Woops.

plugins/linux-capture/xcomposite-input.c
GeorgesStavracas: Good catch!

GeorgesStavracas:
This needs to be tested on a variety of hardware, with special attention to NVIDIA and AMD cards. Intel GPUs should handle this transition just fine.

sekalengrengginang:
tested on obs flatpak with Nvidia GTX 1050ti driver 510.68.02
Xcomposite and XSHM work just fine

file:///var/home/shioriko/Pictures/spectacle/Screenshot_20220515_211034.png

denisschultz:
Tested on AMD Radeon RX 5500 XT, Screen capture and Display capture without any problems.

univrsal:
Seems to work fine for me under Nvidia with proprietary drivers. (Arch linux, 5.17.7-zen1-1-zen, Driver version 515.43.04, GTX 970)

tychii93:
Tested with Flatpak.  Both window and screen capture works with Vega 56, Mesa 22.0.3.  Arch Linux with 5.17.8-zen

libobs/obs-nix-platform.h
```diff
@@ -24,7 +24,6 @@ extern "C" {
 #endif
 
 enum obs_nix_platform_type {
-	OBS_NIX_PLATFORM_X11_GLX,
```
norihiro, line 27: Removing this item, we loose binary compatibility with plugins. These code use the enumerator.
https://github.com/exeldro/obs-source-dock/blob/134f1f614a4382e68de265230aaddc06cc66c8b9/qt-display.cpp#L163-L178
https://github.com/norihiro/obs-color-monitor/blob/e8359b1560c1ea8dede019e8e59943325b48be28/src/obsgui-helper.hpp#L21-L36

GeorgesStavracas: This is an excellent point that I didn't consider, thanks! Though now that I'm thinking, OBS 28 will break ABI anyway (Qt5 → Qt6) so I think we're accidentally safe for now.

norihiro: IIRC, since the package for Ubuntu 20.04 (PPA) will be still compiled with Qt5, isn't that still ABI compatible?

kkartaltepe: Qt is not part of the ABI, but even if it were we should seek to reduce the pain not use it to add more. We can just add a _UNUSED to the name to signify its no longer used but just deleting it makes it that much harder to detect the ABI break as   a plugin developer.

GeorgesStavracas: The enum value is still there now, but nothing can set it, and there's a comment instructing to stop using it in code. Is that good now?

libobs-opengl/gl-nix.c
```diff
@@ -31,7 +30,7 @@ static void init_winsys(void)
 
 	switch (obs_get_nix_platform()) {
 	case OBS_NIX_PLATFORM_X11_GLX:
-		gl_vtable = gl_x11_glx_get_winsys_vtable();
+		exit(1);
```
norihiro, line 33:
```suggestion
```
How about just removing the case item `OBS_NIX_PLATFORM_X11_GLX` or put `break` so that it will reach to `assert` at the end?

plugins/linux-pipewire/linux-pipewire.c
```diff
@@ -45,7 +45,7 @@ bool obs_module_load(void)
 		pipewire_capture_load();
 		break;
 	case OBS_NIX_PLATFORM_X11_GLX:
-		break;
+		return false;
```
norihiro, line 48: I think the entire `switch` statement here can be removed and just call `pipewire_capture_load()`. We can reduce 10 lines.

libobs-opengl/gl-nix.c
GeorgesStavracas: Sorry, I had removed this locally but didn't push afterwards... this is gone now.

kkartaltepe (approved):
I'm still a little apprehensive about cutting out GLX before we have had a release with EGL as the default, but dont let that block merging if you are comfortable with it.

trinitronx:
@Gol-D-Ace :pray: 

[Re-posting this][1] here in hopes that someone with PPA push access will see it.

Can we get the old package version prior to this breaking change pushed back up to the [`obsproject/obs-studio` PPA][2]?  Especially considering LTS releases such as Ubuntu 20.04 Focal.

This was a breaking change for systems with old graphics cards.  Only the latest version is now available in the PPA, and there is no way to pin an old version for such systems with old hardware.

[1]: https://obsproject.com/forum/threads/obs-doesnt-start-any-longer-since-latest-update.159270/#post-587553
[2]: https://launchpad.net/~obsproject/+archive/ubuntu/obs-studio

derrod:
PPA's cannot publish multiple versions at the same time, but you can manually download the old package here: https://launchpad.net/~obsproject/+archive/ubuntu/obs-studio/+packages

PR #6487, opened 2022-05-16
UI: close display before native surfaces

### LINKED ISSUE #6466: Hang after closing properties or filters windows on Wayland with the NVIDIA driver
### Operating System Info

Other

### Other OS

Arhc Linux

### OBS Studio Version

27.2.4

_No response_

### OBS Studio Log URL

https://obsproject.com/logs/THEcoe9zOlxpeXsK

### OBS Studio Crash Log URL

_No response_

### Expected Behavior

OBS studio should not hang after closing the properties or filters windows.

### Current Behavior

OBS studio hangs after closing the properties or filters windows.

### Steps to Reproduce

1. Launch obs on a Wayland desktop with an NVIDIA GPU using the proprietary driver.
2. Open either the "Properties" or "Filters" windows from the gear icon in the "Audio Mixer" panel.
3. Close the window .
4. Behold, obs is no longer responsive.

The hang is occurring in obs_display_destroy acquiring the displays_mutex, which is being called from the OBSQTDisplay destructor after closing the pop-up window.

Meanwhile, the graphics thread is stuck in eglSwapBuffers while holding said muted.

eglSwapBuffers is stuck because it is waiting for a wl_surface::frame callback from the previous swap.

This callback will never arrive because Qt has already destroyed the wl_surface by this point, which discards any frame callbacks registered for the surface.

In general, it is illegal to call eglSwapBuffers on an EGLSurface if the backing platform surface (in this case, the wl_surface) has been destroyed.

kkartaltepe:
### Description

This hooks the platform specific events in order to close the obs
display more accurately. Earlier attempts on hooking visiblity in Qt
, but Qt does not ensure that visibility is changed before the surface is
destroyed. So we ended up racing with the egl driver, and on some
drivers if you lose the race they hang.

Also force closure of the main preview window earlier to avoid similar
ordering issues in qt.

Finally only force display creation if the display is actually visible.
When a source type is not video/drawable (or is missing) this would
force the display to be created for the blank window and also hang.

### Motivation and Context

People prefer if obs doesnt hang on their favorite proprietary drivers. Mostly resolves #6466 

### How Has This Been Tested?

Repeatedly opening and closing source and filter properties worked. Opening xcomposite source properties on wayland also didn't hang the render thread.

### Types of changes

### Checklist:

### DIFF
--- a/UI/qt-display.cpp
+++ b/UI/qt-display.cpp
@@ -42,6 +42,9 @@ protected:
 			else
 				mTimerId = startTimer(67); // Arbitrary
 			break;
+		case QPlatformSurfaceEvent::SurfaceAboutToBeDestroyed:
+			display->DestroyDisplay();
+			break;
 		case QEvent::Expose:
 			createOBSDisplay();
 			break;
@@ -52,7 +55,10 @@ protected:
 		return result;
 	}
 
-	void timerEvent(QTimerEvent *) { createOBSDisplay(true); }
+	void timerEvent(QTimerEvent *)
+	{
+		createOBSDisplay(display->isVisible());
+	}
 
 private:
 	void createOBSDisplay(bool force = false)
--- a/UI/qt-display.hpp
+++ b/UI/qt-display.hpp
@@ -43,6 +43,7 @@ public:
 	void SetDisplayBackgroundColor(const QColor &color);
 	void UpdateDisplayBackgroundColor();
 	void CreateDisplay(bool force = false);
+	void DestroyDisplay() { display = nullptr; };
 
 	void OnMove();
 	void OnDisplayChange();
--- a/UI/window-basic-main.cpp
+++ b/UI/window-basic-main.cpp
@@ -4707,6 +4707,10 @@ void OBSBasic::closeEvent(QCloseEvent *event)
 
 	delete extraBrowsers;
 
+	ui->preview->DestroyDisplay();
+	if (program)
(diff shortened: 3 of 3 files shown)


kkartaltepe:
This needs testing on Windows and Mac to make sure the early delete on the preview doesnt affect them

RytoEX (approved):
Tested the CI artifact on Windows with a debugger attached per @kkartaltepe 's instructions and I observed no adverse effects there. May still need macOS testing.

kkartaltepe:
To clarify the testing steps for mac/windows the changes related to that platform are for the main OBS display. So if you could just enable/disable the preview a few times and verify nothing looks different visually and nothing explodes. Similarly starting and closing obs with the preview enabled under a debugger to make sure nothing crashes on close.

RytoEX (changes requested):
Commit message nits:
1. Qt
2. EGL
3. Use "Close" for the first word of the commit message subject.

Otherwise, my previous approval still stands.

kkartaltepe:
Thanks updated, there was one thing I missed which was testing studio mode. Unfortunately even with this change there is still a chance of hang on close for studio mode it seems (maybe 1/10 in my tests).

kkartaltepe:
Changed around the strategy for handling preview/program so this seems a bit more straightforward and works better in both cases i think. Also reworded the commit message a bit.

PR #6488, opened 2022-05-16
deps/media-playback: Fix video looping

jpark37:
### Description
Previous FFmpeg deprecation fix, and optimization attempt causes short
videos to loop badly. Keep deprecation fix, and fix the rest later.

### Motivation and Context
Want video loop to look nice.

### How Has This Been Tested?
Video loop looks nice again.

### Types of changes

### Checklist:

### DIFF
--- a/deps/media-playback/media-playback/decode.c
+++ b/deps/media-playback/media-playback/decode.c
@@ -193,6 +193,10 @@ bool mp_decode_init(mp_media_t *m, enum AVMediaType type, bool hw)
 
 	if (d->codec->capabilities & CODEC_CAP_TRUNC)
 		d->decoder->flags |= CODEC_FLAG_TRUNC;
+
+	d->orig_pkt = av_packet_alloc();
+	d->pkt = av_packet_alloc();
+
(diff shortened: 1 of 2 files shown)


RytoEX (approved):
I can confirm this fixes the issue I described off-thread, so I'm approving based on that. I'm still not sure if there should be a revert commit with the fixes/changes then committed after that (even after sleeping on it, I'm still not sure what's best for diff/blame), but don't let that hold this up if the code is otherwise fine.

RytoEX (changes requested):
Just noticed that the commit message links to a commit on your fork instead of the commit on this repo. Please fix that, and then I'll move this back to approved from me.

jpark37:
I think GitHub is being weird? It looks like the same hash to me.

RytoEX:
> I think GitHub is being weird? It looks like the same hash to me.

You might be right. Maybe it's just linking to your fork because the commit is technically coming from your fork. Never mind then.

RytoEX (approved):

PR #3437, opened 2020-09-09
linux-v4l2: Save device by path

kkartaltepe:
### Description

This adjusts device listing to be by /dev/v4l/by-path which is based
on where the device is plugged in instead of when it is initialized.
This should help people with multiple devices across restarts.

Old plugin data will still work unchanged, when users modify settings
the new options will be transparently backed by the new keys but names
presented remain consistent.

### Motivation and Context

/dev/videoX is not consistent across reboots. This fixes #3003 for users with multiple devices.

### How Has This Been Tested?

Works with my devices. Seems it should also work with the freeBSD assuming they have readlink. Manually editing scene to have the previous config works fine.

### Types of changes

- Bug fix (non-breaking change which fixes an issue)

### Checklist:

- [x] My code has been run through [clang-format](https://github.com/obsproject/obs-studio/blob/master/.clang-format).
- [x] I have read the [**contributing** document](https://github.com/obsproject/obs-studio/blob/master/CONTRIBUTING.rst).
- [x] My code is not on the master branch.
- [x] The code has been tested.
- [x] All commit messages are properly formatted and commits squashed where appropriate.
- [x] I have included updates to all appropriate documentation.

PR #5160, opened 2021-08-19
Add info (text label) property as text sub-type in properties API

tytan652:
### Description

It adds a property which adds a text label to the properties view. This label can be a simple info, a warning or an error.

This new type of property is a sub-type of text property.

On UI side, Rich Text, clickable links and word wrap are enabled.
Depending on if the description, the string value and the long description are set, the property will be shown differently.

Add_text_info_examples
Forgot a "with" in a string value.

My screenshot tool does not let the cursor appear.
Add_text_info_examples_tooltip

- A text info with only a description will only show the the info without the label in the first column.
- A text info with a description and a string value or a long description will fill the first column with the description and the second with the string value or the long description.
  - If an empty description is set, the first column will be empty.
 - If the three strings are provided, the long description become a tooltip.

### Motivation and Context

A while ago, someone asked on Discord for a property in the properties API which just only add a text label.

I had something from my service overhaul draft 😅  but it needed a revision, and I did it today.

@paulh-aja suggested to make my `add_info` a sub-type of `add_text`, so I also made the change.

### How Has This Been Tested?

I added the property in FFmpeg VAAPI properties function and set some defaults value to the default function to test it on Linux any property view would work.
Default:
```C
obs_data_set_default_string(settings, "test_value", "This a string value with an empty description");
obs_data_set_default_string(settings, "test_value2", "This a string value a description");
obs_data_set_default_string( settings, "test_tooltip", "This a string value with a description and a long description");
```
Properties:
```C
p = obs_properties_add_text(props, "test_normal", "Placeholder desc", OBS_TEXT_INFO);
obs_property_set_long_description(p, "This is a simple normal info label created with properties API,<br> we seriously need to really test Qt rich text");
p = obs_properties_add_text(props, "test_warning", NULL, OBS_TEXT_INFO);
obs_property_text_set_info_type(p, OBS_TEXT_INFO_WARNING);
obs_property_set_long_description(p, "This is a simple warning info label created with properties API");
p = obs_properties_add_text(props, "test_error", "ERROR", OBS_TEXT_INFO);
obs_property_text_set_info_type(p, OBS_TEXT_INFO_ERROR);
obs_property_set_long_description(p, "This is a simple error info label created with properties API, we seriously need to really test word wrapping");

p = obs_properties_add_text(props, "test_desc_only", "This is a info label created with only a description", OBS_TEXT_INFO);
//obs_property_text_set_info_word_wrap(p, false);

[...]

PR #6493, opened 2022-05-16
linux-v4l2: Save device by id or path

### LINKED ISSUE #3003: Webcams ordering is not persistent in Linux
It seems that in Linux, the numbering of the /dev/video (V4L) devices is used to determine which source should use which webcam. This should not be the case, as this numbering is not persistent and often changes depending on whether a device is plugged in before or after boot time.

## Platform

Operating system and version: Arch Linux
OBS Studio version: 25.0.8 (official package from Arch repos)

## Expected Behavior

If webcams are connected in a different order during different boots, OBS should still have the correct webcams selected in the scene sources.

## Current Behavior

I have an internal webcam (in a laptop) and an external USB webcam, to provide two different camera angles. First, I booted with the external webcam disconnected, then plugged it in later. Then I set up some scenes in OBS involving both webcams. Then I rebooted, with the USB webcam still connected, and reopened OBS. My scenes that I had created were broken, because it had swapped the two webcams.

I then rebooted with the webcam disconnected, and then reconnected it and opened OBS. My scenes were now back to normal. This leads me to conclude that OBS is using the `/dev/video*` numbering of the webcams (or something like that) to determine which which source is linked to which webcam. This should not be the case, as this numbering changes often and there are already things like `/dev/v4l/by-id/` for persistent naming of webcams e.g.:
```
❯ ll /dev/v4l/by-id/
total 0
drwxr-xr-x 2 root root 120 May 29 00:21 ./
drwxr-xr-x 4 root root  80 May 29 00:21 ../
lrwxrwxrwx 1 root root  12 May 29 00:21 usb-046d_0825_014C8D90-video-index0 -> ../../video0
lrwxrwxrwx 1 root root  12 May 29 00:21 usb-046d_0825_014C8D90-video-index1 -> ../../video1
lrwxrwxrwx 1 root root  12 May 29 00:21 usb-SunplusIT_Inc_Integrated_Camera-video-index0 -> ../../video2
lrwxrwxrwx 1 root root  12 May 29 00:21 usb-SunplusIT_Inc_Integrated_Camera-video-index1 -> ../../video3
```

This is the output of `sudo v4l2-ctl --list devices` if the USB webcam is disconnected during boot:
```
Integrated Camera: Integrated C (usb-0000:00:14.0-6):
	/dev/video0
	/dev/video1
        /dev/media0

UVC Camera (046d:0825) (usb-0000:00:14.0-2):
	/dev/video2
	/dev/video3
        /dev/media1
```
So the internal camera is at `/dev/video0` and `/dev/video1`.

If the USB webcam is plugged in during boot, then the ordering is like so:
```
UVC Camera (046d:0825) (usb-0000:00:14.0-2):
	/dev/video0
	/dev/video1
        /dev/media0

Integrated Camera: Integrated C (usb-0000:00:14.0-6):
	/dev/video2
	/dev/video3
        /dev/media1
```

## Steps to Reproduce

1. Have two webcams
2. Set up some OBS scenes using both webcams
3. Reboot with devices plugged in/unplugged in a way that causes the `/dev/video*` ordering to change
4. Open OBS and observe that the webcams have been swapped, thus ruining the scenes

ggodlewski:
### Description

Use /dev/v4l/by-id or /dev/v4l/by-path when available in order to persist device binding to specific source.
Fixes #3003
Closes #3437

### Motivation and Context

Currently v4l2 uses /dev/videoX which is not consistent across reboots and device reconnections.

This patch uses directories:

* /dev/v4l/by-id - User has 2 devices: C920 and microscope. Sometimes I use them with another computer. I want to keep them assigned to specific sources/scenes without remembering which USB port to use.
* /dev/v4l/by-path - User has 2 devices: front camera connected to first usb and top camera connected to second usb. If one of the devices is upgraded/changed it remains assigned to the same path.
* /dev/videoX - Devices which are not present in above directories (eg. loopback)

The device can exists twice on the list (by-id vs by-path addressing). It's up to user which connect to the source.

In case of by-path device the source name consist of "bus_info (device name)"
Otherwise it uses current naming: "device name (bus_info)"

Fixes #3003
Closes #3437

### How Has This Been Tested?

Tested with 5 different v4l2 devices:

ID 04fc:6353 Sunplus Technology Co., Ltd General USB Device (Lamax Electra)
ID 0c45:6366 Microdia Webcam Vitade AF (Digital microscope)
ID 534d:2109 MacroSilicon USB3.0 HD VIDEO (Video grabber)
ID 046d:082d Logitech, Inc. HD Pro Webcam C920
ID 046d:085e Logitech, Inc. BRIO Ultra HD Webcam

Test Case 1:

- Plug all the devices, assign them by id, put them on the scene
- It should display all the feeds

Test Case 2:

- Having all devices assigned by id switch their USB ports.
- The scene shouldn't change, all feeds in their old places.

Test Case 3:

- Having all devices assigned by id turn off OBS
- Run OBS
- Nothing is on the scene
- Plug devices in another order
- Devices should be assigned to correct places by udev

Test Case 4:

- Plug 2 devices, assign them by path, put them on the scene
- Switch USB ports
- The scene should have switched cameras

Test Case 5:

- Create v4l2loopback: `sudo modprobe v4l2loopback devices=1 video_nr=88 ; ffmpeg -re -f lavfi -i "movie=something.mp4" -f v4l2 /dev/video88`
- The v4l2loopback should be available on the list and working

### Types of changes

### Checklist:


RytoEX:
If this fixes/closes issues or other PRs, use "Fixes" or "Closes" before the Issue/PR reference to automatically link them together.

Does this resolve the lingering concern raised in #3437?
> it seems like it v4l2loopback devices dont end up registered under the /dev/v4l/ tree. So there is probably more work to also include those video devices without paths.

RytoEX:
I don't think GitHub picks up the keywords if you use a colon between them and the reference.

See:

tytan652:
> Does this resolve the lingering concern raised in #3437?

From PR description:
> - /dev/videoX - Devices which are not present in above directories (eg. loopback)

> Tested with two cameras and v4l2loopback devices.

Seems like a yes.

plugins/linux-v4l2/v4l2-input.c
```diff
@@ -381,20 +376,30 @@ static void v4l2_device_list(obs_property_t *prop, obs_data_t *settings)
 		if (dp->d_type == DT_DIR)
 			continue;
 
-		dstr_resize(&device, 5);
-		dstr_cat(&device, dp->d_name);
+		char *dev_path = dp->d_name;
+		char buf[1024];
+		ssize_t len;
+		if ((len = readlink(dp->d_name, buf, sizeof(buf) - 1)) != -1) {
```
kkartaltepe, line 382: I'm not sure I understand why we need to read the link here?

plugins/linux-v4l2/v4l2-input.c
```diff
@@ -406,26 +411,77 @@ static void v4l2_device_list(obs_property_t *prop, obs_data_t *settings)
...
+ */
+static void v4l2_device_list(obs_property_t *prop, obs_data_t *settings)
+{
+	bool cur_device_found;
+	size_t cur_device_index;
+	const char *cur_device_name;
+
+	obs_property_list_clear(prop);
+
+#ifdef __FreeBSD__
```
kkartaltepe, line 466: With this structure we should be able to clear all the freebsd ifdefs inside of the new function it seems? But maybe im missing something.

kkartaltepe:
Im not a big fan of having both id and path options since I don't think there is really a good way to expose this to the user. I dont think that anyone will get the intention behind the two names in the list without reading this PR. I was leaning towards saving by path but checking with windows it looks like we save by ID there so id prefer we just keep it consistent with windows and only expose by-id to users.

I like the code structure though, seems like a good solution to the issue of virtual devices.

plugins/linux-v4l2/v4l2-input.c
ggodlewski: FreeBSD enumerates all files from /dev which contains "video".
In case of linux we're enumarating all files from /sys/class/video4linux to get the file names. However we're opening device files from /dev.
It's how it used to work before my changes.

plugins/linux-v4l2/v4l2-input.c
ggodlewski: I have no idea. I just borrowed this code from #3437.
Since you're the author and also don't know, I'll remove that. :)

ggodlewski:
As a user I'm unable to remember which USB port I used previously for specific camera.
Linux exposes 2 ways to address v4l2 device to give users some flexibility. I think it's a good idea to expose this choice to OBS users as well.
I added prefix: 'Bus/Device' to unique_device_name.

tytan652:
Between `/dev/v4l/by-id` and `/dev/v4l/by-path`, is there one that manage correctly the scenario where the user connect two device from same model without getting them swapped between reboot.

`/dev/v4l/by-path` seems to be the only one that could avoid this.

> Linux exposes 2 ways to address v4l2 device to give users some flexibility. I think it's a good idea to expose this choice to OBS users as well.

When you want to give choice, you need to check if any user that knows nothing about how device works on Linux could understand the choice they had to/can make.

Here you put everything in a list without even telling the difference between them.

plugins/linux-v4l2/v4l2-input.c
kkartaltepe: In my PR this is to find the real path to the device, but in your case you are joining the fake path to the link which normally would point off to no where, but in your case was working.

kkartaltepe:
I pushed some changes to clean up the freebsd defs. I think the code is fine now but I still dont believe the UX is really usable and we must choose one or the other of id or path. Or find a better UX.

tytan652:
Maybe with #5160 we could add information with another property list which makes the user choose between `/dev/v4l/by-path` and `/dev/v4l/by-id`. And also a third option for `/dev/videoX `.

ggodlewski:
Another property list is more clicking. Most users have just one camera.
#5160 is still open. #3437 was created almost 2 years ago. I don't think we should make this PR dependent on it.

Another, simple proposition:
Lets add disabled item between groups saying: "Advanced. Select device by connection:"

Screenshot at 2022-05-29 13-04-56

Normal user won't bother to click this list.
User with 2 cameras will select by id and ignore rest of the list.
User with cameras of the same model will use the "advanced-by-path" list.

ggodlewski:
I've done more testing and resolved one bug. There was a problem because UDEV is always using real path to device instead of by-* scheme.
Now you're able to reconnect camera to different USB port during streaming or even start without any cameras and plug them during work.
I'd appreciate some testing.

ggodlewski:
@tytan652  it this case someone needs to fork this PR and prepare working proposition. I already spent more time than expected.
My goal was to provide smallest changeset to solve unpredictable OBS behavior while selecting cameras.
Maybe this PR can be merged (because it's better than current state) and UI improvements can be made as a separate PR? #3003 / #3437 are open for about 2 years.

christephan1:
[001-v4l2-hotplug.patch.txt](https://github.com/obsproject/obs-studio/files/9414758/001-v4l2-hotplug.patch.txt)

I have a working patch for your reference. Tested on Ubuntu 22.04.
It supports selecting v4l2 camera by ID. Also hot-plugging USB from 1 port to another.
But my work is based on the Ubuntu 22.04 LTS version of obs-studio-27.2.3+dfsg1-1

What I did is below:
1. Apply the draft in #3437
2. We use by-id instead of by-path. So changed to by-id.
3. Actually no need to readlink, those are symlink back to for example /dev/video0. If we readlink, then it doesn't work as expected.
4. Fix the udev. I assume udev will report the /dev/video0 name in "added" or "removed" callback. So, when comparing the hot-plug device with device_id, we need realpath() of device_id, because our device_id is symlink.
5. I don't assume the device_removed() callback is called before the by-id symlink disappear. So, in v4l2_init(), save the realpath() of device_id to v4l2_data for used in device_removed()
6. Fix The CMakeLists.txt udev issue. (Looks this already been fixed in master branch.)

ggodlewski:
@kkartaltepe can you review it again? I've bought more cameras and tested it again (described in `How Has This Been Tested?`).
Some memory leaks were removed. Also solved some udev problem.
The UI creation was extracted into separated function: `v4l2_build_device_list_ui` if anyone wants to improve it in the future, it is trivial.

kkartaltepe:
Sure ill take another look, but it will probably be a week or two.

DarkDefender:
@kkartaltepe do you have time to take a look at this again in a few weeks?
I don't know if @ggodlewski is still around though.
I'm just asking as I'm eager to have this merged upstream (less third party patches to apply to my own builds).

ggodlewski:
@DarkDefender I will not provide any updates to this project. I wasted many hours to prepare my patches and I bought 2 more devices to do a proper testing. Maintainers ignored me so someone else has to work on that. Obviously linux support is not a priority.

DarkDefender:
I don't know if you had some additional communication with the developers, but I wouldn't attribute any of this to malice or unwillingness to make linux support better.

As I am a Blender developer myself, I do see the same thing on our side of things regarding code contributions. We get a lot of pull requests but we don't really have time to properly spent the time needed on all of them. Usually that means that contributors has to jump through hoops to get things merged. It is not a good situation, but I there is simply not enough developers around to properly take care of every contributor.

I kinda see the same thing here.
@RytoEX do you have any time to get back to reviewing this?
If not, is there anyone else we can collaborate with?

RytoEX:
> @RytoEX do you have any time to get back to reviewing this?
> If not, is there anyone else we can collaborate with?

I am not a Linux-based dev. My previous reviews here were only cosmetic. The only Linux-based contributor that I can think of who I would ask to review this is @kkartaltepe , who volunteers their free time as available. I can't force them to have more free time or force them to look at this.

DarkDefender:
@RytoEX would you be fine if I step in as a reviewer and test this so I can vouch for the changes?
We do have a lot of webcams at the Blender Studio, so I can quite easily create a very complex setup to stress test it.

I'm sure kkartaltepe didn't intentionally abandon this patch as he had a patch of his own addressing the same issue. (But his patch was seemingly abandoned in favor of this one). I think it is just that he legitimately doesn't or didn't have time.

So if I am allowed to step up and help this get merged, I will happily do so.
Especially because I'm going to include this patch in the Blender Studio OBS setup either way. However, I would like others to be able to use multiple v4l sources without ordering issues on reboot.

kkartaltepe:
I did look at it again and it again seemed fine from a code perspective. But as I said " I think the code is fine now but I still dont believe the UX is really usable and we must choose one or the other of id or path. Or find a better UX." and as the author said "@tytan652 it this case someone needs to fork this PR and prepare working proposition. I already spent more time than expected."

This is reasonable code and probably will help whoever wants to do that fork and prepare something with better ux, but for me im not really convinced we should merge this as is and then try to change the UI again later. In fact merging this will probably just prevent us from ever making it more user friendly as such changes would probably break users who depend on having all these options available.

Its definitely at least partially my fault if I wasnt clear the main problem here is not the code but the UX, and at least my position is it should just work like windows.

DarkDefender:
Thanks for clarifying your position

> Main problem here is not the code but the UX, and at least my position is it should just work like windows.

But the issue here is that under the hood this doesn't work like on Windows. So to me, trying to force the interface to be like Windows would be a mistake.

To me, it seems like these two options that tytan652 mentioned are very useful and should be exposed:
>- Per connection (e.g. USB port), /dev/v4l/by-path
>- Per device model, /dev/v4l/by-id

This is because in my experience having "Per connection" would be a life saver when running bigger events like the Blender Conference. There we had cameras or capture cards break during the event and being able to simply delegate certain usb ports to a specific video input would make swapping out equipment really easy.
So instead of requiring people to change source settings, they could simply just plug in the replacement device in the same port and it will work without any further intervention. 

For most users though, I think they will most likely just use the "Per device" mode.
(IE, they can plug the device in where ever they want and it will be mapped to the same source no matter what)

To me simply having the option in the UX to select which method you want to use shouldn't be too confusing or hard to understand. So I don't see why we would need to restrict ourselves to one or the other.

That is how I envision it at least. Do you agree that it sounds reasonable?
If not, perhaps we can discuss it some more to see if we can come to an agreement?

> but for me im not really convinced we should merge this as is and then try to change the UI again later
100% agree. The design of how we want things to work should be agreed upon and implemented before merging.
Otherwise it could just lead to things not getting done later.

ggodlewski:
> We do have a lot of webcams and capture cards at the Blender Studio

@DarkDefender do you have any of Logitech BRIO or Kiyo Pro cameras? There is also #7346 PR which gives more controls like FOV and HDR. I wasn't able to find anyone to test it.

> As I am a Blender developer myself, I do see the same thing on our side of things regarding code contributions.

Oh, what a coincidence. I wanted to provide something for Blender and now I'm afraid ;).

DarkDefender:
I'm unsure if we have a Brio model, I have to check. But I do know that we have a Logitech MeetUp camera. It seems like at least might be able to use the tilt and pan controls. So I could try that out if you want to :)

> Oh, what a coincidence. I wanted to provide something for Blender and now I'm afraid ;).

I think it is much less frustrating if you know before hand the situation.
Before writing any code, I can recommend that you try to contact the developer in charge of the area you want to work in and talk through what you want to contribute.
This usually makes the process a lot smoother as then you can come to an agreement about:

1. Will you contribution be something that would be accepted in the first place
2. Is so, how will it be implemented
3. Get a picture of when the developer in question has time to review and merge your code.

BTW, with the updates posted in this pull request, do you want to continue or should I try to take over?
Of course I haven't heard back yet from kkartaltepe, but for me personally I just think we only need to polish the UI presentation a bit.

It should be nice and intuitive if we either do two buttons like the "Vertices/Edges" toggle buttons in the picture to change between "by-id" and "by-path":
screenshot
Then the user can hover over those buttons and we can have tool tips that explain them in detail.
Of course we could also just have it be a drop down menu instead with tool tips.

This is so it is very clear which "mode" the v4l source is in. What do you think?

ggodlewski:
> I do know that we have a Logitech MeetUp camera. It seems like it at least might be able to use the tilt and pan controls. So I could try that out if you want to :) For me it's so useful to finally have proper FOV support under Linux.

That would be cool. You'll probably need to add GUID for this camera. I curious about the result.

> BTW, with the updates posted in this pull request, do you want to continue or should I try to take over?

My job here is done. The last thing I can maybe do is to rebase #7346 after this is merged. Feel free to take it over.

The interface looks cool.

kkartaltepe:
> @kkartaltepe do you think you will have any time to spend on discussing the UI design? As it seems like you are OK with the code, perhaps I could get someone else involved if you don't have time. Because the remainder of the potential changes do not really need any Linux specific knowledge or expertise.

the best place to have a design discussion is probably on the discord where most developers are. There may be other people more interested in supporting multiple card lookup strategies.

DarkDefender:
That sounds like a great idea!
I didn't think about hopping onto discord to do this. But it makes total sense :sweat_smile:

keriszafir:
Dealing with this issue and here's what I noticed:
1. Fresh OBS config, a scene with a single camera: works okay and the video feed comes back after the camera is disconnected and reconnected; v4l2-input disconnect/reconnect entries appearing in the OBS log
2. Add another camera and a scene, restart OBS and optionally your computer: they tend to be jumbled depending on the /dev/videoN number which is random and assigned by kernel/udev
3. To mitigate this and keep cameras persistent, I made custom rules for my cameras, adding symlinks in /dev directory, e.g.
`KERNEL=="video[0-9]*", ATTR{name}=="Razer Kiyo Pro", ENV{ID_V4L_CAPABILITIES}==":capture:", SYMLINK+="video-kiyo"`
The ENV{ID_V4L_CAPABILITIES}==":capture:" ensures that only the device file with video capture rather than metadata capabilities will be symlinked. 
4. Then I edited the `~/.config/obs-studio/basic/scenes/{scene_collection_name}.json` while OBS was not running, changing the video capture device settings manually `"device_id":"/dev/video2"` to `"device_id":"/dev/video-kiyo"` or `"device_id":"/dev/v4l/by-id/usb-Razer_Inc_Razer_Kiyo_Pro-video-index0"`. Both options break the automatic (udev based) disconnection and reconnection detection, i.e. I can no longer see `info: v4l2-input: Device {path} disconnected` after pulling the USB plug and `info: v4l2-input: Device {path} reonnected` after plugging it back in. In case of unstable USB connections like on my Kiyo Pro, this will lead to camera feed freezing and not restarting, forcing me to close and restart OBS.


Multiple cameras of the same model aren't handled gracefully by the udev persistence rule (in Debian bookworm/sid: `/lib/udev/rules.d/60-persistent-v4l.rules`):

`# do not edit this file, it will be overwritten on update

ACTION=="remove", GOTO="persistent_v4l_end"
SUBSYSTEM!="video4linux", GOTO="persistent_v4l_end"
ENV{MAJOR}=="", GOTO="persistent_v4l_end"

IMPORT{program}="v4l_id $devnode"

SUBSYSTEMS=="usb", IMPORT{builtin}="usb_id"
KERNEL=="video*", ENV{ID_SERIAL}=="?*", SYMLINK+="v4l/by-id/$env{ID_BUS}-$env{ID_SERIAL}-video-index$attr{index}"

# check for valid "index" number
TEST!="index", GOTO="persistent_v4l_end"
ATTR{index}!="?*", GOTO="persistent_v4l_end"

IMPORT{builtin}="path_id"
ENV{ID_PATH}=="?*", KERNEL=="video*|vbi*", SYMLINK+="v4l/by-path/$env{ID_PATH}-video-index$attr{index}"
ENV{ID_PATH}=="?*", KERNEL=="audio*", SYMLINK+="v4l/by-path/$env{ID_PATH}-audio-index$attr{index}"

LABEL="persistent_v4l_end"`

but this file could be overriden with a custom rule in `/etc/udev/rules.d` with more fine-grained rules e.g. adding serial number or other unique identifier from the device's attributes.

GeorgesStavracas:
(Copying this response from https://github.com/obsproject/obs-studio/pull/9246)

Thank you very much for the contribution and for the amount of effort put in this PR. I'm afraid I come with bad news though :( 

After deliberating about this feature, the interaction flow is not what we want to go as the intended user experience. What we want going forward is basically how the macOS integration in OBS Studio handles cameras. Each device gets a unique identifier - even devices of the same model - and that's what's used to address the device. That means that whatever device the user selects in OBS Studio is picked, regardless of where it is connected.

Picking the device of a specific bus might incur losing data, or other sorts of incompatibilities, as if the user swaps the device and the new device doesn't support the same properties of the old one.

As it is right now, I don't think the behaviour above is what our v4l2 code does. I think it should be what it does. Contributions that implement this behaviour would be welcomed.

Hope the user experience considerations that led to rejecting this feature were cleared up. I appreciate that a lot of effort was put in this PR. It is unfortunate that it took so long to close this. I hope we (I) can do better and act faster in the future.

amazingfate:
I have a rk3588 board with hdmirx port, but there is only `/dev/v4l/by-path`:
```
$ ls -l /dev/v4l/*/*
lrwxrwxrwx 1 root root 12 10月 12 19:38 /dev/v4l/by-path/platform-fdb50000.video-codec-video-index0 -> ../../video3
lrwxrwxrwx 1 root root 12 10月 12 19:38 /dev/v4l/by-path/platform-fdb50000.video-codec-video-index1 -> ../../video4
lrwxrwxrwx 1 root root 12 10月 12 19:38 /dev/v4l/by-path/platform-fdb80000.rga-video-index0 -> ../../video2
lrwxrwxrwx 1 root root 12 10月 12 19:38 /dev/v4l/by-path/platform-fdba0000.video-codec-video-index0 -> ../../video5
lrwxrwxrwx 1 root root 12 10月 12 19:38 /dev/v4l/by-path/platform-fdc38000.video-decoder-video-index0 -> ../../video0
lrwxrwxrwx 1 root root 12 10月 12 19:38 /dev/v4l/by-path/platform-fdc70000.video-codec-video-index0 -> ../../video6
lrwxrwxrwx 1 root root 12 10月 12 19:38 /dev/v4l/by-path/platform-fdee0000.hdmi_receiver-video-index0 -> ../../video1
```
Debian has added this pr to the obs-studio is packages: https://salsa.debian.org/multimedia-team/obs-studio/-/commit/4a353a071febf315746a9d184bc6506e31421882, and this pr code is also included in ubuntu 24.04.

When running obs-studio installed from debian or ubuntu's apt repo, if I add a v4l2 capture device, obs will crash. After rebuilding obs-studio without this patch, issue is gone.

kkartaltepe:
There are no debian maintainers here. If you are trying to report a bug about debian you are in the wrong place.

amazingfate:
I'm just posting a FYI message, not asking someone to fix the bug in debian.

ggodlewski:
@amazingfate thanks for the info. I didn't expect that to be included into Debian :).
It's hard to say what is wrong without the logs.
I tried changing the code to use not-existing path `/dev/v4l/by-id2/` and it doesn't crash.

DarkDefender:
@ggodlewski @amazingfate I created a updated version of this PR here: https://github.com/obsproject/obs-studio/pull/9246

I'm working at the Blender Studio and we have been using that patch without issues since then.
We are host the patch ourselves here: https://projects.blender.org/ZedDB/studio_portage_conf/src/branch/master/patches/media-video/obs-studio/fix_v4l_device_order.patch

If you have any issues with the patch you could reported it in the https://projects.blender.orgZedDB/studio_portage_conf repo.

There is also a design task about it here in the OBS repo: https://github.com/obsproject/design/issues/21
Hopefully we can get the ball rolling again as this solves a lot of issues with camera setups.

DarkDefender:
If you guys have any feedback on the design task please drop by.
I really want to know what people think about my proposed UI.

ggodlewski:
@DarkDefender I don't care much about UI as long as it supports all my use cases (id, path, other).

I also another PR which changes slightly UI. It might be less controversial since it only adds missing controls.
https://github.com/obsproject/obs-studio/pull/7346 / https://github.com/ggodlewski/obs-studio/pull/1
It depends on this PR.

christephan1:
I also don't care UI. What I want is just to fix the current OBS bug.
Glad to hear that Ubuntu 24.04 includes this, my original plan is to integrate it by myself when we need to update version.
I don't know why obs team reject this. As the reason they mentioned is exactly why we need this. People still reporting issues in #3003 , they should not suffer that if this PR got merged.

amazingfate:
> @ggodlewski @amazingfate I created a updated version of this PR here: #9246
> 
> I'm working at the Blender Studio and we have been using that patch without issues since then. We are hosting the patch ourselves here: https://projects.blender.org/ZedDB/studio_portage_conf/src/branch/master/patches/media-video/obs-studio/fix_v4l_device_order.patch
> 
> If you have any issues with the patch you could report it in the https://projects.blender.org/ZedDB/studio_portage_conf repo.
> 
> There is also a design task about it here in the OBS repo: [obsproject/design#21](https://github.com/obsproject/design/issues/21) Hopefully we can get the ball rolling again as this solves a lot of issues with various camera setups.

Your patch works fine. After clicking `add video capture device`, I get default `By Device ID`, but no device to choose, then I select `By Bus Path`, and I can see video from my hdmirx device.

DarkDefender:
@amazingfate Awesome! Perhaps you could poke the Debian maintainers to switch the my patch? I'm not using Ubuntu/Debian myself, so I think it would be best for an actual end user to do it. (Especially since you had issues with the current included patch).

Also useful to know that you have a device that only has a "by path"  and not "by id".

amazingfate:
> @amazingfate thanks for the info. I didn't expect that to be included into Debian :). It's hard to say what is wrong without the logs. I tried changing the code to use not-existing path `/dev/v4l/by-id2/` and it doesn't crash.

Here is the log when using obs with your patch:
```
info: v4l2-input: Start capture from 
error: v4l2-input: Unable to open device
error: v4l2-input: Initialization failed, errno: No such file or directory
info: User added source 'Video Capture Device (V4L2)' (v4l2_input) to scene '场景'
free(): invalid pointer
```
It seems that no capture device is deteted.

PR #6494, opened 2022-05-16
UI: Fix memory leak with Manage Broadcast dialog

RytoEX:
### Description

UI: Fix memory leak with Manage Broadcast dialog

When OBS has been connected to a YouTube account, the Manage Broadcast dialog becomes available. Opening and closing this dialog leaks about 2-3 MB of memory each time. This happens because the dialog is never deleted, and a new object is created every time the dialog opens.

If we set Qt::WA_DeleteOnClose on the dialog, then the dialog->exec() call would call delete this. However, if the dialog->Valid() call returned false (this should currently be impossible), then exec is never called, so it would not clean up.

To fix this, make the dialog stack-allocated instead of using new and
delete.

Detected by clang-analyzer.

----

The Qt docs recommend using `dialog->open()` over `dialog->exec()`, but that is a broader change to investigate another day.
* https://doc.qt.io/qt-5/qdialog.html#exec

### Motivation and Context

Memory leaks are bad.

### How Has This Been Tested?

Built and tested locally. Observed memory usage in Task Manager before and after.

### Types of changes

### Checklist:

### DIFF
--- a/UI/window-basic-main.cpp
+++ b/UI/window-basic-main.cpp
@@ -6668,12 +6668,10 @@ void OBSBasic::SetupBroadcast()
 #if YOUTUBE_ENABLED
 	Auth *const auth = GetAuth();
 	if (IsYouTubeService(auth->service())) {
-		OBSYoutubeActions *dialog;
-		dialog = new OBSYoutubeActions(this, auth, broadcastReady);
-		connect(dialog, &OBSYoutubeActions::ok, this,
+		OBSYoutubeActions dialog(this, auth, broadcastReady);
+		connect(&dialog, &OBSYoutubeActions::ok, this,
 			&OBSBasic::YouTubeActionDialogOk);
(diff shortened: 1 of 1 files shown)


notr1ch:
If the lifetime is limited to this scope, perhaps a stack allocation would makes more sense than new / delete?

RytoEX:
> If the lifetime is limited to this scope, perhaps a stack variable would make more sense than new / delete?

As far as I could tell, this is the only place this dialog is initialized. The exec call is blocking/synchronous, so if it is called, it must finish before this function completes.

Otherwise, to your question, I don't have a strong opinion on stack vs. new/delete.

notr1ch:
Looking through the code it seems almost every other dialog is stack allocated, so for consistency I think I'd prefer this instead of new / free. Less memory fragmentation is always good :).

notr1ch (approved):

PR #6499, opened 2022-05-17
libobs: Default 10-bit video to sRGB instead of PQ

jpark37:
### Description
Matches how VLC operates. HDR is only in effect with proper metadata.

### Motivation and Context
Want to behave like other popular programs when no default is specified.

### How Has This Been Tested?
Tested 10-bit video file without video metadata now appears like VLC. PQ and HLG videos still render fine.

### Types of changes

### Checklist:

### DIFF
--- a/libobs/obs-internal.h
+++ b/libobs/obs-internal.h
@@ -912,8 +912,15 @@ static inline enum gs_color_space convert_video_space(enum video_format format,
 {
 	enum gs_color_space space = GS_CS_SRGB;
 	if (convert_video_format(format, trc) == GS_RGBA16F) {
-		space = (trc == VIDEO_TRC_SRGB) ? GS_CS_SRGB_16F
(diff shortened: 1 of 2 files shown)


jp9000:
I wonder if the motivation behind that is because most people generally use 10bit in video files for increased precision video playback rather than HDR specifically.

jpark37:
I don't know about "most," but let's just match VLC, and call it a day.

mufunyo:
> I wonder if the motivation behind that is because most people generally use 10bit in video files for increased precision video playback rather than HDR specifically.

10 bit video files have been around for a lot longer than HDR has, so yes.

ValZapod:
> 10 bit video files have been around for a lot longer than HDR has, so yes.

Not that it works in many places. That is like only openGL in mpv... maybe Vulkan too in mpv. And Photoshop of course.

PR #6501, opened 2022-05-17
decklink-output-ui: Fix crash when stopping preview

### LINKED ISSUE #5832: [Crash] Closing OBS while having an active Decklink Output
### Operating System Info

Windows 10

### Other OS

_No response_

### OBS Studio Version

27.1.3

_No response_

### OBS Studio Log URL

https://obsproject.com/logs/EyGorWuXlARMNGUm

### OBS Studio Crash Log URL

https://obsproject.com/logs/Ar7YSJa5sPwyIDfE

### Expected Behavior

Not to crash.

### Current Behavior

Crashes.

### Steps to Reproduce

1. Tools > Decklink Output
2. Select Device (DeckLink SDI 4k)
3. Press 'Start' under the 'Output' setting
4. Close OBS.

Also got this crash on encoder PC when simply pressing the "Start" button on Decklink output. May be unrelated.
https://obsproject.com/logs/oFw3TDouwRZKGC1Q

RytoEX:
### Description

Removing the first call to `video_output_stop` prevents the `data_mutex` field in `struct video_output` from being destroyed while still in use. The `render_preview_source` function will call `video_output_lock_frame` upon that mutex and encounter a NULL pointer exception.

Co-authored-by: @paulh-aja 

Fixes #5832.

----

This change is from #5891, but for Decklink Output UI. In my experience, the problem results in a hang rather than a crash, but either is bad.

Text from #5891:
Stopping the Preview output in the AJA Output UI can lead to a crash.

**Problem**
The function `preview_output_stop` in `UI/frontend-plugins/aja-output-ui/aja-ui-main.cpp` calls `video_output_stop`, which results in a call to `pthread_mutex_destroy`: https://github.com/obsproject/obs-studio/blob/364f102561faf565ae22e5fa0f1f4aabf3e07c48/UI/frontend-plugins/aja-output-ui/aja-ui-main.cpp#L103

This results in the `data_mutex` field of `struct video_output` from being deleted and made `NULL` while still in use by the `render_preview_source` function in aja-ui-main.cpp.

The call to `video_output_close` a few lines after this also results in a call to `video_output_stop` on the same structure. Allowing this call to perform the `video_output` mutex cleanup seems to be the correct action.

https://github.com/obsproject/obs-studio/blob/364f102561faf565ae22e5fa0f1f4aabf3e07c48/libobs/media-io/video-io.c#L267

**Resolution**
Removing the first call to `video_output_stop` appears to fix the crash and not result in any other harmful effects.

### Motivation and Context

Crashes and hangs are bad.

### How Has This Been Tested?

Tested by spamming Preview Start/Stop, both manually and by holding down the Enter key on Windows to engage/disengage it as fast as the keyboard repeat rate would allow.

### Types of changes

### Checklist:

### DIFF
--- a/UI/frontend-plugins/decklink-output-ui/decklink-ui-main.cpp
+++ b/UI/frontend-plugins/decklink-output-ui/decklink-ui-main.cpp
@@ -124,7 +124,6 @@ void preview_output_stop()
 {
 	obs_output_stop(context.output);
 	obs_output_release(context.output);
-	video_output_stop(context.video_queue);
 
 	obs_remove_main_render_callback(render_preview_source, &context);
 	obs_frontend_remove_event_callback(on_preview_scene_changed, &context);


RytoEX:
@DDRBoxman approved off-thread. Merging.

PR #6507, opened 2022-05-18
UI/installer: Install Media Foundation if missing

flaeri:
Checks for mfplat.dll, and offers user to install Media Foundation. If accepted, Powershell installs it, and installer informs user of  required reboot.

### Description

If the installer detects mfplat.dll is missing, you get the following message and offer.
vmconnect_9F90Kfuk6m
Yes triggers the powershell installer. No exits the installer.

Upon yes, a powershell window is spawned with a progress bar. Window closes automatically when done
vmconnect_1gcUtCsI5I

Once the powershell script is done, the user is informed a reboot is required (it really is, as far as I can tell).
vmconnect_8DJ1TDcbRI
A bit unfortunate that the mfplat.dll does not exist as far as the check can tell until the reboot is done.

### Motivation and Context

A few versions back we bumped FFMPEG to a version that requires Media Foundation. For the time being, I'm not sure this requirement will be removed.

For Windows 10/11 "N" versions, which ships without Media Foundatino installed, it will completely prevent the user from running OBS Studio.

Manually installing this can be quite confusing and difficult for a lot of users.
This installation method *should* work for most supported Win10/11 builds.
I believe this installation method is valid for Windows 10, version 1903, and later (no manual download required)
Have not personally tested on the older builds (only on 21H2), and it's probably worthwhile to test on at least one more build, like 20H2 (still enterprise support).

A fairly large number of users run Windows N are coming by the support channels asking for help. A majority of people are not able to successfully get Media Foundation installed correctly on the first try. I would prefer preReqs were handled for them in an easier way (equivalent to vcredist).

### How Has This Been Tested?

Built the installer successfully. Ran the installer on a clean Win10 21H2 VM, multiple times (uninstall mfplat, retry, re-install OBS). I can provide the installer if that makes it easier for others to test.

### Types of changes

### Checklist:

### DIFF
--- a/UI/installer/mp-installer.nsi
+++ b/UI/installer/mp-installer.nsi
@@ -187,6 +187,23 @@ Function PreReqCheck
 	dxOK:
 	ClearErrors
 
+	; Media Foundation (win N) Check
+	ClearErrors
+	GetDLLVersion "mfplat.dll" $R0 $R1
+	IfErrors mfMissing mfOK
+	mfMissing:
+	IfSilent +1 +3
+		SetErrorLevel 4
+		Quit
+	MessageBox MB_YESNO|MB_ICONEXCLAMATION "Your system is missing Media Foundation components that ${APPNAME} requires. Would you like to install it?" IDYES mftrue IDNO mffalse
+	mftrue:
+	ExecWait 'powershell -ExecutionPolicy bypass -Command ""add-windowsCapability -Online -Name Media.MediaFeaturePack~~~~0.0.1.0"" '
+	MessageBox MB_OK|MB_ICONEXCLAMATION "Media Foundation installation requires you to reboot the computer. Please do so, and run the installer again"
+	mffalse:
+	Quit
+	mfOK:
+	ClearErrors
+
 	; Check previous instance
 	check32BitRunning:
 	OBSInstallerUtils::IsProcessRunning "obs32.exe"


notr1ch:
Personally I get a bad impression whenever an installer opens a command prompt or other console window to do something. How long does the installation typically take - do we actually need to show a progress bar? What happens if the user panics and closes the PowerShell window half way through?

The example dialog shown in this PR does not seem to match the script - "missing components / install it" vs "missing components / install them". I also think the reboot prompt should be in the form of a "Reboot now yes/no" dialog rather than asking the user to do it themselves.

I also noticed the installer quits if Media Foundation is missing when running a silent install. Perhaps we could install MF silently instead? Otherwise we would probably need a separate exit code to differentiate it from a missing DirectX condition.

Fenrirthviti:
Leaving a few thoughts we had discussed on Discord last night:

1. This installed the MF Pack, which includes some of the missing media player components. Users who are using Windows N (which is really the only userbase this affects) may have chosen that version explicitly to not have the media player components, as that's why the version exists in the first place. If we do any kind of installation, it needs to be absolute bare minimum, or be very explicit with the user what it's actually installing. It's a little ambiguous right now.
2. This, supposedly, comes from a requirement in ffmpeg, not OBS currently. If this requirement only exists in ffmpeg, we might consider compiling it without the MF components.

Overall, I'm not opposed to this change, and providing a check of some kind for users who don't have the MF components installed, but I think we need to do a little more digging in to exactly *what* inside OBS is using them, and if there is a better solution than offering to install a nebulous pack from Microsoft.

flaeri:
> Personally I get a bad impression whenever an installer opens a command prompt or other console window to do something. How long does the installation typically take - do we actually need to show a progress bar? What happens if the user panics and closes the PowerShell window half way through?

Yeah, I get that. Unfortunately, it can take a while, even on an SSD. 30sec-a few min. I could run it hidden, but then I really think the experience would feel like the installer is hanging. I don't know if its possible to implement some sort of "please wait" prompt in NSIS, but yeah, this would be nice to improve.

I've forced closed the installer early and late, it all seemed to end up the same. It won't be installed, and it will just re-prompt you to install it, which works fine.

> The example dialog shown in this PR does not seem to match the script - "missing components / install it" vs "missing components / install them". I also think the reboot prompt should be in the form of a "Reboot now yes/no" dialog rather than asking the user to do it themselves.

Thats fair, I'll see if I can't get that working.

> I also noticed the installer quits if Media Foundation is missing when running a silent install. Perhaps we could install MF silently instead? Otherwise we would probably need a separate exit code to differentiate it from a missing DirectX condition.

Yeah... You're right.

flaeri:
> 2\. This, supposedly, comes from a requirement in ffmpeg, not OBS currently. If this requirement only exists in ffmpeg, we might consider compiling it without the MF components.

I've checked. It is from avcodec-58.dll (ffmpeg). I compiled our deps with `--disable-mediafoundation` here
https://github.com/flaeri/obs-deps/commit/64d8bc5b58c72e3ae3d4284643be6688aafd7fe7

This removed the mfplat.dll dependency from ffmpeg (and OBS), and allows me to run OBS without having Media Foundation installed. Just did some basic tests, and media source + custom ffmpeg output was still working for basic actions.
My tests were quite narrow, and if there is anything we would like to confirm, I can do so.

flaeri:
After some more discussions with the team, I think we should move towards excluding Media Foundation as a dependency for the time being.

See PR in OBS-Deps: https://github.com/obsproject/obs-deps/pull/105

Unsure if I should just close this PR for the time being, or if we want to keep/improve it incase of future use.

Fenrirthviti:
This can probably be closed, we can reference the PR in the future if we decide to revisit.

flaeri:
closing in favor of https://github.com/obsproject/obs-deps/pull/105

PR #6514, opened 2022-05-19
UI: Clean up source right-click menu

### LINKED ISSUE #4968: Resize output(source size) on source doesn't support ctrl+z undo correctly
### Operating System Info

Windows 10

### Other OS

_No response_

### OBS Studio Version

27.0.1

_No response_

### OBS Studio Log URL

https://obsproject.com/logs/v42NLzmjD1cl2hCb

### OBS Studio Crash Log URL

_No response_

### Expected Behavior

I expected ctrl+z to undo both the "crop" to the selected source and the resizing of the canvas/output size

### Current Behavior

it uncrops/restores the source setup as it was before, but does not undo the rescaling of the canvas/output resolution

### Steps to Reproduce

1. Resize output(source size) on a source that's smaller than canvas, preferablly not in top left corner too
2. hit ctrl+z
3. be amazed! (at the lack of re-resizing the canvas/output size)
...

_No response_

### LINKED ISSUE #4969: Resize output(source size) on cropped source resizes to source size, not cropped area size
### Operating System Info

Windows 10

### Other OS

_No response_

### OBS Studio Version

27.0.1

_No response_

### OBS Studio Log URL

https://obsproject.com/logs/v42NLzmjD1cl2hCb

### OBS Studio Crash Log URL

_No response_

### Expected Behavior

I believe it should resize output to the cropped area of the selected source, since the cropped area is just empty/black/not filled

### Current Behavior

It resizes output to the source size, and then scales the cropped area with in that according to the transform settings, usually resulting in black bars/empty space if it's not a crop to the same aspect ratio

### Steps to Reproduce

1. Create window capture of smaller-than-canvas window.
2. Crop this source a bit, to show it clearly crop to just the center 1/3
3. Resize output (source size)
4. you will now see the canvas beign resized to the source size(hit ctrl+e, check size), while the actual cropped contents are fitted per the Bounding Box Type selected in the transform

I'm not sure if i'm just not used to the cropping tool, or if cropping afterwards doing these steps also makes the cropping act weird(can't uncrop with alt-drag, doesn't show cropped/green lines?)

gxalpha:
### Description

 
Cleans up the source right click menu mostly according to a mockup and requests from @Warchamp7.

Old menu (source has async video / no audio):
image

New menu (source has async video / no audio):
image

Mockup: 

New menu (no video / with audio):
image

Detailed overview:
- Removed "Resize output (source size)" menu
- Only shows "Interact" menu when source can be interacted with, instead of always showing the menu, but disabling it with most sources
- Grouped Scale Filtering, Blending Mode & Method, and Deinterlacing together
- Moved Projector menus down
- "Hide in mixer" has its own group. This is the only change that doesn't comply with the mockup (and the notes), since just up removing it would remove functionality and a better place for it has yet to be found. This can come at a later point.

### Motivation and Context

List is very long. List can be shorter.

Changes to it were requested many times by multiple people, this follows Warchamps mockup and notes.

### How Has This Been Tested?

macOS 12.4
Made sure the menus still work and appear when they are supposed to.
See screenshots above.

### Types of changes

### Checklist:

Closes #4968 
Closes #4969

### DIFF
--- a/UI/data/locale/en-US.ini
+++ b/UI/data/locale/en-US.ini
@@ -1196,11 +1196,6 @@ AddUrl.Title="Add Source via URL"
 AddUrl.Text="You have dragged a URL into OBS. This will automatically add the link as a source. Continue?"
 AddUrl.Text.Url="URL: %1"
 
-# Dynamic output size
-ResizeOutputSizeOfSource="Resize output (source size)"
-ResizeOutputSizeOfSource.Text="The base and output resolutions will be resized to the size of the current source."
-ResizeOutputSizeOfSource.Continue="Do you want to continue?"
-
 PreviewTransition="Preview Transition"
 
 # Import Dialog
--- a/UI/window-basic-main.cpp
+++ b/UI/window-basic-main.cpp
@@ -5601,18 +5601,6 @@ void OBSBasic::CreateSourcePopupMenu(int idx, bool preview)
 			popup.addSeparator();
 		}
 
-		QAction *resizeOutput =
-			popup.addAction(QTStr("ResizeOutputSizeOfSource"), this,
-					SLOT(ResizeOutputSizeOfSource()));
-
-		int width = obs_source_get_width(source);
-		int height = obs_source_get_height(source);
-
-		resizeOutput->setEnabled(!obs_video_active());
-
-		if (width < 8 || height < 8)
-			resizeOutput->setEnabled(false);
-
 		scaleFilteringMenu = new QMenu(QTStr("ScaleFiltering"));
 		popup.addMenu(
 			AddScaleFilteringMenu(scaleFilteringMenu, sceneItem));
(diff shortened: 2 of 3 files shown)


RytoEX:
> * Removed "Resize output (source size)" menu

It's unusual that we completely remove a feature rather than relocate/replace it. Is this being removed because of a particular motivation, or was it simply missing from @Warchamp7 's mockups and thus you removed it?

> * Only shows "Interact" menu when source can be interacted with, instead of always showing the menu, but disabling it with most sources

I'm not usually fond of the context menu changing on different kinds of sources because it changes the arrangement of menu items based on the type of source you selected. This may cause issues with muscle-memory for mouse movement or keyboard navigation. Is there a guideline somewhere that says this is a best practice?

gxalpha:
> I'm not usually fond of the context menu changing on different kinds of sources because it changes the arrangement of menu items based on the type of source you selected. This may cause issues with muscle-memory for mouse movement or keyboard navigation. Is there a guideline somewhere that says this is a best practice?

These are the Apple Human Interface Guidelines on [Context menus](https://developer.apple.com/design/human-interface-guidelines/macos/menus/contextual-menus/):
> **Include only the most commonly used commands that are appropriate in the current context.** For example, in the contextual menu for selected text, it makes sense to include editing commands but it doesn’t make sense to include a Save or Print command.

RytoEX:
> > I'm not usually fond of the context menu changing on different kinds of sources because it changes the arrangement of menu items based on the type of source you selected. This may cause issues with muscle-memory for mouse movement or keyboard navigation. Is there a guideline somewhere that says this is a best practice?
> 
> These are the Apple Human Interface Guidelines on [Context menus](https://developer.apple.com/design/human-interface-guidelines/macos/menus/contextual-menus/):
> 
> > **Include only the most commonly used commands that are appropriate in the current context.** For example, in the contextual menu for selected text, it makes sense to include editing commands but it doesn’t make sense to include a Save or Print command.
> 
> In the context of most sources, _Interact_ isn't applicable, so it doesn't make sense to include it.

I've read that Apple HIG page. The guideline right above the one you quoted says that the menus should be consistently arranged, which to me, implies that they should not change if the type of item selected is the same (in OBS, a source; in the HIG example, text):
> In general, all menus and menu items should be consistently arranged and titled.

gxalpha:
> I've read that Apple HIG page. The guideline right above the one you quoted says that the menus should be consistently arranged, which to me, implies that they should not change if the type of item selected is the same (in OBS, a source; in the HIG example, text):
>> In general, all menus and menu items should be consistently arranged and titled.

When I read that part, I felt that it was just a "Don't mess with the order of the items for no reason", but your interpretation definitely also makes sense.

I guess @Warchamp7 needs to look at this as well.

Warchamp7:
While they are both sources, they are still different _types_ of sources with some different options available to them. Another issue is that we don't really expose many of these things in any other place, so we need to take a bit of liberty in presenting these menus at least for now.

I agree that common functionality should be consistent between all of them in location (Ex. Higher up the list) and more specific controls should be further down, or in an area that is consistent across all sources for "source type specific" entries.

Example from Google Docs of menus for Text and an Image
chrome_dCXZY1Q0xm

chrome_J3tn01vPcu

Common functionality for all contents of a document are at the top like Cut/Copy/Paste/etc. Other content specific stuff, as well as some light reorganization based on frequency of use. I do not think things like "Hide in Mixer" should always be visible and only enabled for sources with audio, or Interact visible in the menu but disabled for everything but browser sources.

> It's unusual that we completely remove a feature rather than relocate/replace it. Is this being removed because of a particular motivation, or was it simply missing from @Warchamp7 's mockups and thus you removed it?

It can stay for now since as mentioned it doesn't exist anywhere else, and we have no semblance of how commonly it's used. 

However I will say: I think from a design standpoint, it's providing functionality OBS theoretically isn't built for currently and shouldn't be there. We're accelerating a workflow we don't do a very good job of supporting. 

If a user wants to regularly record individual windows or games that vary in dimensions, I think they're better off with a program like ShareX until such a time we can provide a proper and convenient workflow for it.

Warchamp7 (approved):

SuslikV:
When referencing user interfaces (as examples for the OBS) is better to look at the image editors rather than the word processors, IMHO.

PR #6516, opened 2022-05-19
UI: Use correct terminology for Program in Studio Mode

gxalpha:
### Description

Changes transition settings from being called "Swap Preview/Output Scenes After Transitioning" to "Swap Preview/Program Scenes After Transitioning", as well as tooltip strings.

In the tooltips, there are instances of "output" left which I *think* are appropriate, but it would be great to have a native English speaker to chime in on this.

### Motivation and Context

It's called "Preview/Program" everywhere else, so that's the terminology we should use.

### How Has This Been Tested?

N/A, string change.

### Types of changes

### Checklist:

### DIFF
--- a/UI/data/locale/en-US.ini
+++ b/UI/data/locale/en-US.ini
@@ -255,12 +255,12 @@ Updater.GameCaptureActive.Title="Game capture active"
 Updater.GameCaptureActive.Text="Game capture hook library is currently in use. Please close any games/programs being captured (or restart Windows) and try again."
 
 # quick transitions
-QuickTransitions.SwapScenes="Swap Preview/Output Scenes After Transitioning"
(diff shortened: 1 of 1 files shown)


RytoEX:
> When editing the same scene, allows editing transform/visibility of sources without modifying the output.\nTo edit properties of sources without modifying the output, enable 'Duplicate Sources'.\nChanging this value will reset the current program scene (if it still exists).

> When editing the same scene, allows editing properties of sources without modifying the output.

The use of "the output" here is _probably_ okay. If you wanted to go for full consistency and correctness, you could use "the program" instead. It might also be acceptable to say "the program scene", "the program output", "the live program", "the live output", or something similar.

gxalpha:
Changed to "program output"

RytoEX (approved):
Seems fine to me.

Issue #4885, opened 2021-06-12
obs-ffmpeg: AAC encoder fails with Invalid argument (NaN error)


SuslikV:
### Operating System Info

Other

### Other OS

any OS

### OBS Studio Version

27.0.0

### OBS Studio Log URL

about:blank

### Expected Behavior

Common AAC Encoder shouldn't fail on broken audio samples.

### Current Behavior

When decoded audio has NaN as value the OBS Studio fails to encode it into AAC and stops recording/streaming immediately.

### Steps to Reproduce

1. Run OBS Studio
2. Enable Replay Buffer (simple mode)
3. Start Replay Buffer (encoding should start)
4. Add Media source and point it to the attached "issue FFmpeg NaN aac encode error - Short - 50ms.wav" file.
5. Click OK button to close the source properties.
6. If encoder didn't yet fail for you - try to show/hide the Media source (p.4) few times.
7. The OBS Studio stops the Replay Buffer and rises error message: "Recording error - An unsupported error occurred while recording". The log file has next entry: 
```
20:02:03.547: error:   Input contains (near) NaN/+-Inf
20:02:03.547: [FFmpeg aac encoder: 'simple_aac'] avcodec_encode_audio2 failed: Invalid argument
20:02:03.547: Error encoding with encoder 'simple_aac'
```
**TEST FILE:**
[issue FFmpeg NaN aac encode error - Short - 50ms.zip](https://github.com/obsproject/obs-studio/files/6642593/issue.FFmpeg.NaN.aac.encode.error.-.Short.-.50ms.zip)
359 bytes to download, CRC of the .zip archive MD5:1C0440AB08860A39A02822AFA4C369C7

### Anything else we should know?

- The FFmpeg AAC encoder fails on input of the  floating point number that represents NaN (usually it keeps NaN in all math calculations, thus it is possible). This happens inside internal FFmpeg's _aac_encode_frame_ function.
- Audio in OBS can be modified by third-party plugins that can be written improperly (or without awareness of this limitation of the FFmpeg and OBS): https://obsproject.com/forum/threads/obs-studio-27-release-candidate.141857/page-2#post-522499
[...]

PR #6518, opened 2022-05-19
libobs: Clamp audio NaN to 0.0f

claudeha:
### Description

Rearrange comparison order so that NaN is caught.
Comparisons with NaN always result in false.
NaNs cause problems later in audio encoder.

### Motivation and Context

When a NaN (not a number) audio sample is fed to OBS, it immediately stops both streaming and recording because ffmpeg audio encoder doesn't like it.

See #4885, which this patch should theoretically fix, if compiler obeys IEEE semantics (no `-ffast-math` or so).

### How Has This Been Tested?

I have not tested yet.

### Types of changes

### Checklist:

### DIFF
--- a/libobs/media-io/audio-io.c
+++ b/libobs/media-io/audio-io.c
@@ -145,6 +145,7 @@ static inline void clamp_audio_output(struct audio_output *audio, size_t bytes)
 
 			while (mix_data < mix_end) {
 				float val = *mix_data;
+				val = (val == val) ? val : 0.0f;
 				val = (val > 1.0f) ? 1.0f : val;
 				val = (val < -1.0f) ? -1.0f : val;
 				*(mix_data++) = val;


jp9000:
The greater than/less than comparison changes seem unnecessary

Nofew:
Media Autobuild Suite builds ffmpeg with -ffast-math by default, so there may be precedent for libffmpeg inside of obs to also not conform to IEEE standards. Can we confirm if this is the situation somehow, since that compiler flag would break this PR?

Testing this patch seems especially prudent to me.

notr1ch:
Nice catch! I have confirmed we do potentially have NaN in our `clamp_audio_output` inputs with malformed media files (and likely some glitchy audio devices). This PR does fix one test case with an audio file with out of range signals. It's not known at this time whether the FFmpeg errors users experience are mostly occurring with NaN input or due to FFT overshoot, but this PR definitely seems to fix the NaN input case.

This does mean we need to take care when building to make sure `-ffast-math` is disabled, and provide suitable warning for 3rd party maintainers not to do so.

RytoEX:
> The greater than/less than comparison changes seem unnecessary

Ah, I think I see what you mean. The first changed line seems to make the latter two unnecessary. @claudeha can you explain why the second two lines would be needed? Can a NaN get past the first line?

I don't know if we want NaN to clamp to 1.0f as previously proposed or 0.0f as done here. @notr1ch ?

notr1ch:
NaN will always fail an equality check: NaN is both greater than 1 and less than 1.

> In comparison operations, positive infinity is larger than all values except itself and NaN, and negative infinity is smaller than all values except itself and NaN. NaN is unordered: it is not equal to, greater than, or less than anything, including itself. x == x is false if the value of x is NaN.

Setting NaN to 0 probably makes the most sense, since NaN should ideally never be present in the first place and likely represents some kind of corrupt / invalid data.

claudeha:
On 19/05/2022 23:20, Ryan Foster wrote:
>
>     The greater than/less than comparison changes seem unnecessary
>
> Ah, I think I see what you mean. The first changed line seems to make 
> the latter two unnecessary. @claudeha <https://github.com/claudeha> 
> can you explain why the second two lines would be needed? Can a NaN 
> get past the first line?
>
I guess with the first line (explicitly clamp NaN to 0) the changes in 
the other two would be unnecessary (I changed them first, before 
realizing that clamping NaN to 1.0f might not be such a good idea, and 
didn't change them back after adding the first line).

Possibly it's better to use `isnan()`, which might be less performant 
(though I haven't benchmarked) but could maybe be more robust (though I 
haven't checked -ffast-math behaviour) as well as being more 
self-documenting.

Maybe instead of changing +3-2 lines have a +1-0 with this line added 
before the two unchanged clamping lines:

```
val = isnan(val) ? 0.0f : val;
```

Will need `#include <math.h>` if it's not present already, and also 
glibc needs feature test macros to be defined before include which 
complicates things:

 >       isnan():
 >           _ISOC99_SOURCE || _POSIX_C_SOURCE >= 200112L
 >               || _XOPEN_SOURCE
 >               || /* Since glibc 2.19: */ _DEFAULT_SOURCE
 >               || /* Glibc versions <= 2.19: */ _BSD_SOURCE || 
_SVID_SOURCE


A choice between `isnan(val)` (and associated complexity of build system 
macros) and `! (val == val)` (and associated greater risk that 
`-ffast-math` might break it); then should I make a new PR or force push 
adjusted commit or something else?

notr1ch:
`-ffast-math` also breaks `isnan` so I think the current comparison is fine. Generally we prefer force pushes to update PRs unless there's a significant implementation difference that causes it to no longer align with the PR description.

notr1ch (approved):

notr1ch:
I discovered that `-ffast-math` sets a macro in both GCC and Clang, so we can prevent its use, e.g.

```
#ifdef __FAST_MATH__
#error libobs cannot be built with -ffast-math
#endif
```

RytoEX (approved):
Looks good to me. Confirmed NaN values are caught and clamped to 0 by this change.

notr1ch:
Gonna go ahead and merge this; refusing to build with `-ffast-math` is probably a larger discussion.

PR #6508, opened 2022-05-18
obs-ffmpeg: Add texture-based hardware AMD encoder

jp9000:
### Description
Add support for texture-based AMD encoding, with both H264, HEVC, and HDR support. Fall back to FFmpeg when texture-based encoding cannot be used for whatever reason.

### Motivation and Context
The original implementation of the AMD hardware encoder is a submodule that's old, poorly maintained, a bit complicated and difficult to maintain, and should probably be considered deprecated at this point.

This PR instead implements the AMD hardware encoder in a similar approach to jim-nvenc: a very minimalistic texture-based implementation that falls back to FFmpeg if texture-based encoding is not available for whatever reason. With this sort of implementation, it makes the AMD encoder easier to maintain and focuses specifically on texture-based encoding, which should be OBS' strength.

I based this code upon the work done by AMD/Luxoft themselves in obsproject/obs-studio#4538, which already was a simplified texture-based implementation, but I further simplified their implementation according to our preferences/tastes. In addition, I also added support for P010 textures for HEVC, updated the API to the latest release, included support for HDR, added a subprocess to check for AMD encoder support called on startup, and implemented an FFmpeg fallback for when texture-based encoding is unavailable.

I wanted people to be able to pass custom options, but I didn't want to implement a complicated properties system, so instead I opted for the ability to pass custom options via text, similar to how the x264 encoder works; except in this case, we use the FFmpeg option names so the options can safely be used with the FFmpeg fallback. (We should probably do something similar for jim-nvenc at some point).

It'd be preferable to deprecate and eventually fully remove the old AMD hardware implementation in favor of this.

Please leave comments and testing results if anyone has a chance to test this. I have only tested this on two machines.

### How Has This Been Tested?
Tested H264, HEVC, and HDR. Everything appears to be functioning, although I have probably not tested every option or command line param option. This probably needs more testing on various machines as well. To be able to test HEVC, you must build it yourself and turn on the ENABLE_HEVC CMake option when configuring.

Again, if anyone else could test this with their AMD video cards, the help would be greatly appreciated.

### Types of changes
- New feature (non-breaking change which adds functionality)
- Performance enhancement (non-breaking change which improves efficiency)
- Code cleanup (non-breaking change which makes code smaller or more readable)

### Checklist:

- [x] My code has been run through [clang-format](https://github.com/obsproject/obs-studio/blob/master/.clang-format).
[...]

PR #6519, opened 2022-05-19
obs-ffmpeg: Remove duplicate "FFmpeg Options" locale

gxalpha:
### Description

There are two separate locales for "FFmpeg Options", now there is only one.
Also moves the tooltip up to where the other source options are.

### Motivation and Context

Duplicate locales only cause extra code and work for translators.

### How Has This Been Tested?

macOS 12.4
Made sure for the source both the name and tooltip of the property are shown.

### Types of changes

### Checklist:

### DIFF
--- a/plugins/obs-ffmpeg/data/locale/en-US.ini
+++ b/plugins/obs-ffmpeg/data/locale/en-US.ini
@@ -2,6 +2,7 @@ FFmpegOutput="FFmpeg Output"
 FFmpegAAC="FFmpeg Default AAC Encoder"
 FFmpegOpus="FFmpeg Opus Encoder"
 FFmpegOpts="FFmpeg Options"
+FFmpegOpts.ToolTip.Source="Allows you to set FFmpeg options. This only accepts options in the option=value format.\nMultiple options can be set by separating them with a space.\nExample: rtsp_transport=tcp rtsp_flags=prefer_tcp"
 Bitrate="Bitrate"
 MaxBitrate="Max Bitrate"
 Preset="Preset"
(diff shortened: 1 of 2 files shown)


Vainock:
I prefer the version Jim already used in his PR: "Use to specify custom FFmpeg options. For example, \"level=5.2 profile=main\"."

Explains the formatting with a generic example and reduces the length of the sentence to a minimum. Same information much more compact and easier to read.

RytoEX:
> I prefer the version Jim already used in his PR: "Use to specify custom FFmpeg options. For example, \"level=5.2 profile=main\"."
> 
> Explains the formatting with a generic example and reduces the length of the sentence to a minimum. Same information much more compact and easier to read.

The longer string has a specific use case to explain a specific usage in the media source. Options like level and profile may or may not work in that use case. I would not want non-working options to be reflected in that flow. The options described in the longer string are very specific for the flow in which they are used. I'm not sure they are actually interchangeable as suggested.

gxalpha:
I agree with Ryan.
The options here are specifically for the ffmpeg source, while the ones in Jim's PR are for an encoder. The arguments used for an encoder ("level=5.2 profile=main") don't make sense for the source.

What I did do is move the string up to the other ffmpegopts like jim did as it probably makes more sense to have them together.
I'll also recommend in #6508 to add a `.Encoder` to the string to indicate that it isn't universal, just as this has a `.Source`

RytoEX:
Was just typing this, but apparently moments too late. If I want to be pedantic, the key name should probably be: FFmpeg.Source.Options.Tooltip to indicate that it is specifically a tooltip for the FFmpeg Source's Options. I know there's a desire to match it to the existing FFmpegOpts key, but we don't need to have those match. It's not a big deal either way, so long as we don't get confused about its purpose in the future.

PR #6529, opened 2022-05-22
Use EETF when HDR media exceeds HDR scene

jpark37:
### Description
Rendering HDR media without EETF may cause brights to get clipped. Use metadata, and reduce overall brightness with EETF where source is brighter than scene.

Only per-stream metadata is checked. Frame-level metadata is ignored.

### Motivation and Context
Don't want to lose bright detail by clipping..

### How Has This Been Tested?
- [x] Tested 1000-nit PQ video at 400/1000/2000 nits, and saw darker/same/same as expected. Also stepped through shader logic in RenderDoc.
- [x] Tested HLG video at 400/1000/2000 nits with no regressions.

### Types of changes

### Checklist:

### DIFF
--- a/deps/media-playback/media-playback/decode.c
+++ b/deps/media-playback/media-playback/decode.c
@@ -15,7 +15,9 @@
  */
 
 #include "decode.h"
+
 #include "media.h"
+#include <libavutil/mastering_display_metadata.h>
 
 #if LIBAVCODEC_VERSION_INT > AV_VERSION_INT(58, 4, 100)
 #define USE_NEW_HARDWARE_CODEC_METHOD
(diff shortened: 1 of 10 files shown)


jp9000:
There's code in here that changes obs_source_frame, which breaks ABI, but I realized I already merged an ABI break in 0ed0f2cdb4 -- so what I'm going to do is I'm going to merge this, and then I'm going to separate the ABI-breaking parts into an obs_source_frame3. (I know, this is probably annoying, but apparently there's like 20 plugins that rely on this ABI so I feel like this is probably unavoidable).

I wanted to break ABI/API this patch but I'm not sure if we should yet. I feel like if we're going to break ABI/API, we should do it very infrequently and we should try to improve as many things as possible in the API/ABI. I'm undecided on it for now, so right now I'm just going to merge this, then I'm going to fix the ABI for now, then we'll decide what to do later.

jp9000:
Nevermind, ABI not broken, I was wrong; jpark is just incredibly sneaky with his structure modifications. I'll just sit down now.

PR #6530, opened 2022-05-22
UI: Disable network features if a non-RTMP service is set

tytan652:
### Description

Hide UI element related to the advanced network group which if the user set a non-RTMP service.

It actually check if it is a custom service that is set and if the server starts with "rtmp" or if the service is meant to use `rtmp_output`, if not network options will be hidden.

A message is shown explaining why the elements are hidden.
Screenshot from 2022-05-27 08-16-41

### Motivation and Context

Only `rtmp_output` has those network related features implemented.

~~I choose to hide it rather than grey it out, because hiding indicate that the feature is not present.
If greyed out, people might think that they missed a checkbox somewhere.~~

~Network feature UI elements are now greyed out (disabled).~

They now hidden and a message is shown to explain why.

### How Has This Been Tested?

- Set a non-RTMP service, elements are disabled
- Set a RTMP(S) one, they are enabled
- Set a custom server that start with "rtmp", they are enabled otherwise they are not.

### Types of changes

### Checklist:


WizardCM:
I do worry that by hiding it, users will be confused whether we removed the feature entirely, rather than disabled it due to the chosen service.

tytan652:
So greyed it out would be better ?

I was doubting while making this PR, and switching to grey out rather than to hide is not difficult.

tytan652:
Now UI element are greyed out (disabled) and I change the wording (bind IP -> network features)

Fenrirthviti:
I think adding a label that has a simple description, such as "Advanced networking features disabled, output is not RTMP" would be worthwhile here, to avoid any confusion on why it's disabled. I'm in favor of the change overall.

tytan652:
Now a message replace options.

Screenshot from 2022-05-27 08-16-41

gxalpha:
As people mentioned off-thread, bind to IP would still work when someone used a third party plugin to also stream to an RTMP server besides the main non-rtmp stream.
With that in mind, does this PR still make sense?

cc @notr1ch @pkviet

notr1ch:
The multi output plugins still uses the OBS output system, which relies on librtmp. The bind to IP UI being disabled but still applying to RTMP outputs would be a weird edge case. This UI assumes a single stream output which is true with plain OBS (for now), but falls apart when plugins come into the picture. I'm beginning to think that this should be left as-is.

tytan652:
How to  explain this obs-multi-rtmp does not seems to rely on `SimpleOuput` or `AdvancedOutput` which apply those properties on the output, so it does not use it.

tt2468:
Just verified that the multi RTMP plugin does NOT implement any RTMP bind interface logic. Bind interface is passed via the output's settings object, and the multi RTMP plugin does not apply any settings to the outputs it creates.

Warchamp7 (approved):
Minor phrasing nit:
The currently selected streaming protocol does not support changing network settings

LGTM

tytan652:
Changes done.

PR #6533, opened 2022-05-23
linux-capture: Check da_erase target is valid

### LINKED ISSUE #6532: linux-capture: Xcomposite - darray_erase: Assertion `idx < dst->num' failed
### Operating System Info

Other

### Other OS

Gentoo Linux

### OBS Studio Version

Git

_No response_

### OBS Studio Log URL

https://obsproject.com/logs/vnjcnBE1fibN3ZxK

### OBS Studio Crash Log URL

_No response_

### Expected Behavior

No crashy crashy.

### Current Behavior

obs: /[build-path]/libobs/util/darray.h:312: darray_erase: Assertion `idx < dst->num' failed.

Thread 34 "libobs: graphic" received signal SIGABRT, Aborted.

### Steps to Reproduce

1. Starting with a clean profile (after deleting '$HOME/.config/obs-studio', although it seems to be enough if the 'Untitled.json' is deleted from under 'scenes'), run OBS Studio.
2. Add a Window Capture (Xcomposite) source.
3. Select a window to capture, and close said window.

- It may be enough to simply close OBS Studio instead of closing the source window.
- I can also trigger the failure by mousing over the LXQt panel to make it come out of hiding (it is set to automatically hide).

This starts with commit 1604400e488 [1], although I did not have it happening when initially testing the PR for it.

Not that I'm familiar with the code, but it doesn't look too normal that 'idx' is suddenly such a big number (18446744073709551615), when it seems to normally be 0 to 3.  Or maybe 4 was the highest I saw while testing with added logging.

Log with backtrace: https://gist.github.com/Chiitoo/3ac706dc0d134a85d467970489ee9018

1. https://github.com/obsproject/obs-studio/commit/1604400e488f66c885dc43e9ad8fb2a56beaa0bb

kkartaltepe:
### Description

This was leftover from the cpp conversion where the delete would not
fail, but for darray we need to skip da_erase if its unneeded.

### Motivation and Context

fixes #6532

### Types of changes

### Checklist:

### DIFF
--- a/plugins/linux-capture/xcomposite-input.c
+++ b/plugins/linux-capture/xcomposite-input.c
@@ -610,10 +610,13 @@ void watcher_unregister(xcb_connection_t *conn, struct xcompcap *s)
 			sizeof(struct reg_item), &watcher_registry.da, i);
 
 		if (item->src == s) {
+			idx = i;
 			win = item->win;
 			break;
 		}
 	}
+	if (idx == DARRAY_INVALID)
+		goto done;
 
 	da_erase(watcher_registry, idx);
 
@@ -636,6 +639,7 @@ void watcher_unregister(xcb_connection_t *conn, struct xcompcap *s)
 					     vals);
 	}
 
+done:
 	pthread_mutex_unlock(&watcher_lock);
 }
 


plugins/linux-capture/xcomposite-input.c
```diff
@@ -615,7 +615,8 @@ void watcher_unregister(xcb_connection_t *conn, struct xcompcap *s)
 		}
 	}
 
-	da_erase(watcher_registry, idx);
+	if (idx != DARRAY_INVALID)
+		da_erase(watcher_registry, idx);
```
GeorgesStavracas, line 619: Hm, do you not need to early return here even? If the window is not in `watcher_registry`, there's no need to iterate over it yet another time. Early returning would need to unlock `watcher_lock` though.

GeorgesStavracas:
Adding "Seeking Testers" so the OP of the issue can test it after you push again. @kkartaltepe if you decide not to apply the suggestion above, please feel free to rerun CI manually!

plugins/linux-capture/xcomposite-input.c
adsr: Is the intent to set `idx = i` before the `break` in the loop above? As I'm reading it, `idx` will always eq `DARRAY_INVALID` here.

Chiitoo:
Yep, works for me so far.

Thanks!

plugins/linux-capture/xcomposite-input.c
kkartaltepe: Yes that was the original issue, it appears I completely forgot to add in the assignments to `idx` in both places. Likely the size of the watcher table just ended up growing and growing, but double checking with the assignments I forgot in place things look good.

kkartaltepe:
Should be good, confirmed I was in a debug build to replicate. Also confirming the table length looks about right, but it seems there might be an entry leak left in the code.

GeorgesStavracas (approved):
\o/

Gustavokruger:
Hey dad

Em ter, 24 de mai de 2022 08:17, Georges Basile Stavracas Neto <
***@***.***> escreveu:

PR #6535, opened 2022-05-23
cmake: Fix frontend-api built when ENABLE_UI=OFF

jpc0:
### Description

Moved the add_subdirectory call in the UI CMakeLists.txt to after the check
for it ENABLE_UI is set to off

### Motivation and Context
Currently the obs-frontend-api is still being built even when ENABLE_UI
is off and there is no frontend being built

### How Has This Been Tested?
I built it and confirmed the frontent-api isn't being built

### Types of changes

### Checklist:

### DIFF
--- a/UI/CMakeLists.txt
+++ b/UI/CMakeLists.txt
@@ -1,5 +1,3 @@
-add_subdirectory(obs-frontend-api)
-
 option(ENABLE_UI "Enable building with UI (requires Qt)" ON)
 if(NOT ENABLE_UI)
   obs_status(DISABLED "OBS UI")
@@ -12,6 +10,8 @@ if(NOT QT_VERSION)
       CACHE STRING "OBS Qt version [5, 6]" FORCE)
   set_property(CACHE QT_VERSION PROPERTY STRINGS 5 6)
 endif()
+
+add_subdirectory(obs-frontend-api)
 
 project(obs)
 


RytoEX:
This might be intentional for plugin development and DX reasons. @PatTheMav ?

PatTheMav:
> This might be intentional for plugin development and DX reasons. @PatTheMav ?

Exactly, obs-frontend-api is needed for plugin development. This PR would break the plugintemplate.

jpc0:
> > This might be intentional for plugin development and DX reasons. @PatTheMav ?
> 
> 
> 
> Exactly, obs-frontend-api is needed for plugin development. This PR would break the plugintemplate.

Just so I understand this a little better, the [documentation on plugins](https://obsproject.com/docs/plugins.html) do not reference the frontend API at all. Is there an issue with the documentation or is there an issue with the plugin template which puts a hard requirement on the UI even when the UI isn't being built at all or am I missing something else entirely here.

I don't see anything in the public API that would be usable in the code should the UI not be built.

Given the above, could I look at adding a cmake flag to disable this from being built should it not be needed.

Final option is I will just live with it/patch manually on pull should I not need this built.

PatTheMav:
Our currently supported usecases out of the box are:

1. Building the OBS application with the common feature set expected by most users (which already diverge per-OS which isn't optimal, but it is what it is).
2. Building just the libraries necessary for development (which include both libobs and obs-frontend-api).

Not every plugin needs the frontend-api library, but it's a common enough use case to have it be built for development out of the box. As-is this PR would not allow these developers to build just the frontend-api without also building the entire application.

What would be good to understand is why having the frontend-api being built is an issue to begin with?

jpc0:
> 

Likely the same reason -DENABLE_UI=OFF is used in the template, reducing built time for a feature that is not needed. The reason for the pull request is because it makes sense that the ENABLE_UI option should be disabling the entire module as is the case with all the other flags.

I'm not inside one of your supported use cases which is to just use the libobs library and it's related modules with the default plugins. I guessed this was in the supported use case considering it has been kept separated quite well and the only thing that cannot be disable in the build system is building the frontend-api which seemed like a bug.

I'm willing to add an ENABLE_FRONTEND_API option that is defaulted on or as I said I'm happy to just patch before building.

PatTheMav:
You should be to able to run `cmake --build <YOUR_BUILD_DIR> -t libobs` to _just_ build the main library without anything else. CMake will be smart enough to build all dependencies of libobs but no other parts.

Granted, this applies to all other usecases as well, but if it's a single target you need, that should get you what you need.

PR #6537, opened 2022-05-24
mac-vth264: Rename mac-vth264 plugin to mac-videotoolbox

gxalpha:
### Description

Renames the mac-vth264 plugin to mac-videotoolbox to better reflect that it
contains a videotoolbox encoder that mostly isn't h264-specific (except for
very few places), and could easily be extended to support other codecs.

### Motivation and Context

This crossed my mind when I decided to test how hard it is to convert this
to an hevc encoder and noticed it only requires changing 6 lines for an POC.
While hevc isn't decided on at all, I think it's a good idea in general to rename
the plugin to something that better reflects what it's theoretically capable of.

### How Has This Been Tested?

macOS 12.5 Beta
Tested that recording and streaming both still work (at least as far as "working"
can be applied to the encoder in general).
Made sure that existing configurations get picked up and continue working
as expected (the ID stored didn't change, so this is a non-breaking change 
for users).

### Types of changes

### Checklist:


gxalpha:
Verified that our updater does indeed delete old files.
Should be good to merge.

PR #6542, opened 2022-05-26
libobs: Add shader enhancements for better colors

Anastasy888:
### Description

Improve accuracy of color space linear conversion coefficients (709 to 2020 and 2020 to 709).

### Motivation and Context

Achieving better color representation to improve user experience of content creators without significant performance changes.

### How Has This Been Tested?

Cmake variable ENABLE_HEVC should be enabled (-DENABLE_HEVC=ON).
1. Create a scene, add HDR media source
2. Record to file
3. Compare recorded file with original file played on HDR monitor

### Types of changes

### Checklist:


WizardCM:
Hi and welcome! Please split your changes into 3 separate commits. It's also possible one of the other reviewers might ask you to split this into separate PRs, but I'll let them decide.

jpark37:
> Improve accuracy of color space linear conversion coefficients

This is fine, but do you have sources for these numbers, so I can verify them? I should probably have added a comment that I pulled our values from a D3D sample.

> Add separate white level parameter for preview

This seems related to the recent email thread. I think users can manage without this additional setting, and I'm of the opinion that simpler is better, but maybe I don't understand the use-case for adding this parameter.

> Change HDR media source tonemapping from rough Reinhard to more accurate HDR to SDR tonemapping

We don't want to tonemap HDR sources early even for SDR output because users may want to filter in HDR space before tonemapping occurs. This also gives them the opportunity to implement their own tonemapper if they dislike Reinhard, either by using Apply LUT, which currently would not work well since we don't allow shapers, or providing their own custom filter plugin.

As far as switching to BT.2446, it doesn't seem like it has universal approval: https://github.com/mpv-player/mpv/issues/7984#issuecomment-1019553601

I suspect any choice we make for HDR to SDR tonemapping is going to have tradeoffs. We chose Reinhard because it's cheap, and it's good enough for Xbox GameDVR. (Their algorithm is a bit different, but it's mostly Reinhard.) We could make the choice of tonemapper a global setting, or even per-source although I'm not sure we want to get into that right now. Adding that configurability would probably make a mess.

jpark37:
For what it's worth, my own math matches up with yours:
0.62740389593469914, 0.32928303837788397, 0.043313065687417190,
0.069097289358232047, 0.91954039507545904, 0.011362315566309173,
0.016391438875150235, 0.088013307877225860, 0.89559525324762468,

float aliased as uint32_t:
Current: 3f209d8c 3ea897a6 3d31699a 3d8d82ba 3f6b66f9 3c3a2454 3c8647ad 3db44044 3f6545b7
Yours: 3f209d8b 3ea897c9 3d31690b 3d8d82e1 3f6b6700 3c3a2901 3c864757 3db44052 3f6545bb
Mine: 3f209d8b 3ea897c9 3d31690b 3d8d82e1 3f6b6700 3c3a2901 3c864757 3db44052 3f6545bb

Opposite matrix:
1.6604910021084343, -0.58764113878854973, -0.072849863319884745,
-0.12455047452159063, 1.1328998971259603, -0.0083494226043695080,
-0.018150763354905199, -0.10057889800800746, 1.1187296613629123,

jpark37:
Our rec2020_to_rec709 values seem far apart though:
Current: 3fd48af8 bf166fa6 bd953254 bdff1455 3f9102dd bc08cbec bc94b0fd bdcdfc4f 3f8f3289
Yours: 3fd48ba2 bf167433 bd95187e bdff1a0d 3f9104da bc099c69 bc94d466 bdcdf2b7 3f8f327d
Mine: 3fd48af8 bf166fa6 bd95324f bdff1452 3f9102dd bc08cc04 bc94b0e9 bdcdfc4f 3f8f3289

jpark37:
This is what I see in Visual Studio debugger when I multiply your matrices together, should be as close to identity as possible:
1.00000834, -2.28071585e-05, 1.44802034e-05,
-8.84501787e-06, 1.00005114, -4.23016027e-05,
-1.57821923e-05, 2.05859542e-05, 0.999995172,

New multiplied:
1.00000000, -2.60770321e-08, 3.72529030e-09,
-2.61934474e-09, 1.00000000, 0.00000000,
0.00000000, 0.00000000, 1.00000012,

Anastasy888:
> This is fine, but do you have sources for these numbers, so I can verify them?

I was watching to https://www.itu.int/dms_pub/itu-r/opb/rep/R-REP-BT.2407-2017-PDF-E.pdf and https://www.itu.int/dms_pubrec/itu-r/rec/bt/R-REC-BT.2087-0-201510-I!!PDF-E.pdf. Also using already computed coefficients with higher precision should be better, so I'll change it. Leaving coefficients with low precision may lead to color conversion artefacts.

Anastasy888:
> > Add separate white level parameter for preview
> 
> This seems related to the recent email thread. I think users can manage without this additional setting, and I'm of the opinion that simpler is better, but maybe I don't understand the use-case for adding this parameter.

Current use-case is managing HDR scene with HDR sources, which differs from what I asked in the email about. For example, we add HDR media source and HDR window capture. The problem appears when the user wants to change White Level, but the multiplier for capturing and preview is computed with the same parameter, so the scene will be affected, but it will not be relevant with the preview and the result output may be confusing. For example, user will not see the result for capturing, but he will see the result for video, which is not changed. So Preview White Level is an advanced parameter that user can use to calibrate his preview brightness on HDR monitor and see relevant changes on preview when he changes White Level.

Anastasy888:
As for using tonemapping from BT.2446, I agree that it is too complex and the result can look weird. I have chosen it because of better brightness match to SDR version of the same video. But since this scenario is not common, I will omit this changes and leave the simplier implementation.

jpark37:
For example, I don't see 82250100.0f/49532999.0f in BT.2407, and the values in that matrix seem to be inaccurate.

I sent a list of SDR/HDR scenarios in the separate email thread. Can you point out which one the preview SDR white level setting is trying to solve?

Anastasy888:
> For what it's worth, my own math matches up with yours for rec709_to_rec2020: 0.62740389593469914, 0.32928303837788397, 0.043313065687417190, 0.069097289358232047, 0.91954039507545904, 0.011362315566309173, 0.016391438875150235, 0.088013307877225860, 0.89559525324762468,
> 
> float aliased as uint32_t: 
Current: 3f209d8c 3ea897a6 3d31699a 3d8d82ba 3f6b66f9 3c3a2454 3c8647ad 3db44044 3f6545b7 
Yours: 3f209d8b 3ea897c9 3d31690b 3d8d82e1 3f6b6700 3c3a2901 3c864757 3db44052 3f6545bb 
Mine: 3f209d8b 3ea897c9 3d31690b 3d8d82e1 3f6b6700 3c3a2901 3c864757 3db44052 3f6545bb

Current coefficients are less precise (7 digits against 10 digits after point) and they differ from high precise numbers (for the 1st coefficient: 3f209d8c and 3f209d8b), so we suggested more accurate coefficients than the current.
These numbers were obtained by my colleagues in the previous projects. Their goal was not the only high precision of the numbers but also to fix artefacts that appeared through the color conversions.
I should admit that visual difference is not significant.

Anastasy888:
> Our rec2020_to_rec709 values seem far apart though. My computed matrix: 1.6604910021084343, -0.58764113878854973, -0.072849863319884745, -0.12455047452159063, 1.1328998971259603, -0.0083494226043695080, -0.018150763354905199, -0.10057889800800746, 1.1187296613629123,
> 
> float aliased as uint32_t: Current: 3fd48af8 bf166fa6 bd953254 bdff1455 3f9102dd bc08cbec bc94b0fd bdcdfc4f 3f8f3289 Yours: 3fd48ba2 bf167433 bd95187e bdff1a0d 3f9104da bc099c69 bc94d466 bdcdf2b7 3f8f327d Mine: 3fd48af8 bf166fa6 bd95324f bdff1452 3f9102dd bc08cc04 bc94b0e9 bdcdfc4f 3f8f3289

Same as for rec709_to_rec2020, these numbers were obtained to increase precision and to fix the artifacts.

jpark37:
I'm fine with using more precise numbers, but I'm going to need more information how your 2020 -> 709 matrix is computed since they don't seem close to the actual numbers.

i would use these numbers, which I computed using doubles:

709 -> 2020:
0.62740389593469914., 0.32928303837788397., 0.043313065687417190.,
0.069097289358232047., 0.91954039507545904., 0.011362315566309173.,
0.016391438875150235., 0.088013307877225860., 0.89559525324762468.,

2020 -> 709:
1.6604910021084343., -0.58764113878854973., -0.072849863319884745.,
-0.12455047452159063., 1.1328998971259603., -0.0083494226043695080.,
-0.018150763354905199., -0.10057889800800746., 1.1187296613629123.,

jpark37:
Looking back at your comment, You have an:

> HDR scene with HDR sources

And you want to:

> preview brightness on HDR monitor

This doesn't involve the SDR white level at all, so I'm still not sure why you need a new setting.

Anastasy888:
> I sent a list of SDR/HDR scenarios in the separate email thread. Can you point out which one the preview SDR white level setting is trying to solve?

I have checked the use case that had a problem and there was no problem. The only issue I should mention is changing SDR white level while the video from HDR media source is paused. It appears because its frames are asynchronously rendered with the previous value of SDR white level.

libobs/data/color.effect
```diff
@@ -20,17 +20,17 @@ float3 srgb_nonlinear_to_linear(float3 v)
 
 float3 rec709_to_rec2020(float3 v)
 {
-	float r = dot(v, float3(0.6274040f, 0.3292820f, 0.0433136f));
-	float g = dot(v, float3(0.0690970f, 0.9195400f, 0.0113612f));
-	float b = dot(v, float3(0.0163916f, 0.0880132f, 0.8955950f));
+	float r = dot(v, float3(0.6274038959f, 0.3292830384f, 0.0433130657f));
+	float g = dot(v, float3(0.0690972894f, 0.9195403951f, 0.0113623156f));
+	float b = dot(v, float3(0.0163914389f, 0.0880133079f, 0.8955952532f));
```
jpark37, line 25: Could you use the whole numbers provided? Just let the shader compiler truncate to 32 bits.

Also remove the f suffix to match this file's style. Needed in C++ to denote 32-bit float instead of double, but unnecessary for shader.

libobs/data/color.effect
```diff
@@ -20,17 +20,17 @@ float3 srgb_nonlinear_to_linear(float3 v)
...
 	return float3(r, g, b);
 }
 
 float3 rec2020_to_rec709(float3 v)
-{
-	float r = dot(v, float3(1.6604910, -0.5876411, -0.0728499));
-	float g = dot(v, float3(-0.1245505, 1.1328999, -0.0083494));
-	float b = dot(v, float3(-0.0181508, -0.1005789, 1.1187297));
+{
```
jpark37, line 30: Delete tab.

jpark37:
> The only issue I should mention is changing SDR white level while the video from HDR media source is paused. It appears because its frames are asynchronously rendered with the previous value of SDR white level.

HDR media source paused on an SDR monitor or scene, right? I can reproduce the issue, but it would need more plumbing to address, and I don't expect people to be changing the value often if at all. You also can't change the value while streaming and/or recording is active.

jpark37 (approved):

jpark37:
Thanks.

PR #6546, opened 2022-05-26
(Draft) DRM / GBM renderer for Linux

GeorgesStavracas:
(Draft since it depends on #6475)
 
### Description

Introduce a GBM-based GL renderer.

### Motivation and Context

This PR is the foundation for future changes introducing a texture sharing mechanism to libobs, which should allow for a more efficient rendering pipeline on Linux based on DMA-BUF passing. The new GBM renderer introduced here is effectively unreachable, there's no way to set it without patching the codebase.

I don't know how feasible it is to use a texture sharing mechanism on Mac and Windows. I believe it's both possible and efficient.

### How Has This Been Tested?

Currently, this code is unreachable; but it is possible to manually call `obs_set_nix_platform(OBS_NIX_PLATFORM_DRM)` on startup, and observe that the GL renderer does not complain.

(I have further proof that this code works, but it's too early to share, which is another reason for this PR to be a draft)

### Types of changes

### Checklist:


libobs/obs-nix-platform.h
```diff
@@ -29,7 +29,7 @@ enum obs_nix_platform_type {
 #ifdef ENABLE_WAYLAND
 	OBS_NIX_PLATFORM_WAYLAND,
 #endif
-
+	OBS_NIX_PLATFORM_DRM,
 };
 
 /**
```
tytan652, line 35: To avoid API/ABI breakage `OBS_NIX_PLATFORM_WAYLAND` should be added even if Wayland is disabled.

jp9000:
That line added/removed ratio is quite wonderful.

tt2468:
Thought I'd comment on what it took to get this running on my Ubuntu 20.04 system using NVIDIA 510, in case anyone else happens to have issues:

- Install `libgdm-dev`
- I had a weird issue where it tried to load wayland hotkeys, causing a crash. If you have issues like that, build explicitly with `-DENABLE_WAYLAND=OFF`
- Made sure the `nvidia_drm` kernel module was loaded (see `nvidia-modprobe`)
- Added myself to the `video` and `render` groups, to give my account access to `/dev/dri/*`

libobs/obs-nix-platform.h
kkartaltepe: +1 we could leave wayland optional when it was at the end of the list, but now we just need it to be there always in the enum.

libobs/obs-nix-drm.c
```diff
@@ -0,0 +1,79 @@
...
+#include <unistd.h>
+#include <sys/mman.h>
+#include <string.h>
+
+struct obs_hotkeys_platform {
+};
+
+void obs_nix_drm_log_info(void) {}
+
+static bool obs_nix_drm_hotkeys_platform_init(struct obs_core_hotkeys *hotkeys)
```
kkartaltepe, line 31: What is the plan for supporting hotkeys?

GeorgesStavracas: Global shortcuts portal when running under a compositor, libinput otherwise. This is a very far fetched idea though, I haven't spend any time thinking about this. The use case for this renderer is to export the resulting frames as a shared texture, potentially for offscreen rendering or integrating with other exotic cases. The texture sharing part is missing from this PR - I have some working code in a branch, but it's not ready to be published yet.

libobs/obs-nix-platform.h
GeorgesStavracas: Sure, it's just that at this point I spend precisely 0 seconds thinking about API, ABI, or any compatibility whatsoever. There's a plethora of reasons this PR is a draft :slightly_smiling_face:

kkartaltepe:
I'm not totally sure that this how we should approach this I think. DRM as a platform for libobs only makes sense if we want to support a headless mode. But I think the end goal you had discussed is to migrate obs swapchains off of window system dependent surfaces. 

The platform abstraction covers more than just swapchains but also the input handling etc. Since DRM doesnt provide any of that it being a platform doesnt make much sense by itself to me.

My thoughts are that we would either have both window systems share some drm egl context backends for most rendering and then only have specialized swapchain handling code to deal with those platform specific swapchains. And I think later you wanted to push that code out to the UI where it would render from the drm swapchain into the platform window? But even in that case we need window information for input handling.

kkartaltepe:
> Thought I'd comment on what it took to get this running on my Ubuntu 20.04 system using NVIDIA 510, in case anyone else happens to have issues:

Are you sure this is working? On my system at least, after adding in the lines to force DRM platform in obs-app.cpp, the first surface fails to create.

kkartaltepe:
I had some more time to play with this, and at least on nvidia 515 there doesnt seem to be any possible way to create gbm_surface's. Replacing the surfaces with gbm_bo's seems to work and at least looking at the other tools that nvidia's drivers were written to support (xorg and wayland compositors) no one seems to use gbm_surfaces so id be a bit concerned about us trying to use them if we were the only program exercising the driver code.

GeorgesStavracas:
@kkartaltepe does the gbm_surface approach work if you replace `GBM_BO_USE_RENDERING` by `GBM_BO_USE_RENDERING | GBM_BO_USE_LINEAR`?

GeorgesStavracas:
Or perhaps could you please try the following:

```diff
-       wi->gbm_surface = gbm_surface_create(plat->gbm, wi->width, wi->height,
-                                            GBM_FORMAT_ARGB8888,
-                                            GBM_BO_USE_RENDERING);
+       wi->gbm_surface = gbm_surface_create_with_modifiers2(plat->gbm,
+                                                            wi->width, wi->height,
+                                                            GBM_FORMAT_ARGB8888,
+                                                            NULL, 0,
+                                                            GBM_BO_USE_RENDERING | GBM_BO_USE_LINEAR);

```

kkartaltepe:
No neither of those work on my system.

PR #5095, opened 2021-08-07
Don't trigger hotkeys without modifiers if user is holding down modifiers (strict)

OldBaldGeek:
### Description
(This description has been revised from the original, since a proposed UI component has been removed)
Trigger hotkeys only when the key and modifiers pressed by the user exactly match the hotkey definition.

### Motivation and Context

To fix #4408.

Suppose a user defines hotkey actions for "1", "Ctrl-1", "Alt-1", and "Ctrl-Alt-1." With the code in OBS 27 or earlier, if they then press "Ctrl-Alt-1", ALL FOUR actions will be triggered, in an order that may change between runs of OBS (presumably due to how the json is loaded).

While there may be some use case for this, requiring exact matches greatly expands the number of actions that can be invoked independently by hotkeys.

Bug reports
[https://github.com/obsproject/obs-studio/issues/4408](url)
[https://obsproject.com/mantis/view.php?id=925](url)
[https://obsproject.com/mantis/view.php?id=1468](url)
[https://obsproject.com/mantis/view.php?id=1571](url) (duplicate of 1468)

Forum posts complaining about the current behavior
[https://ideas.obsproject.com/posts/...at-ctrl-alt-x-doesnt-trigger-ctrl-x-and-alt-x](url)
[https://obsproject.com/forum/thread...iably-for-scene-selection.146242/#post-535412](url)
[https://obsproject.com/forum/threads/hotkey-modifier-bug.79197/#post-333409](url)
[https://obsproject.com/forum/threads/please-fix-the-hotkeys.79150/#post-333256](url)
[https://obsproject.com/forum/threads/hotkey-modifier-issue.78002/#post-329251](url)

There is a pull request from 2020 that attempts to fix this 
[https://github.com/obsproject/obs-studio/pull/2785](url)
This pull has a number of problems, both in functionality and in coding), as noted in my comment on the pull request.

### How Has This Been Tested?

Tested using
- A scene collection with scenes showing white text at different positions
- A Lua script that logs changes to the Program scene, and with handlers for four hotkeys that simply log the hotkey.

After assigning hotkeys with various modifiers to change scenes and trigger the Lua script, you can observe the script log window and see the behavior of v27 and modified code. In many cases you can see flashes caused by transient scene switching when in v28, or with the fix in legacy mode.

### Types of changes

- Bug fix (non-breaking change which fixes an issue)

### Checklist:

- [X] My code has been run through [clang-format](https://github.com/obsproject/obs-studio/blob/master/.clang-format).
[...]

PR #6550, opened 2022-05-28
libobs,UI: Replace `obs_hotkey_inject_event` with "external" hotkeys

gxalpha:
### Description

Replaces the existing method of injecting hotkeys from the UI (and other
places) with one that doesn't rely on the platform implementation at all.
The new method programatically more closely resembles the platform
implementation and shouldn't cause bugs like the one described under
motivation and context.
A UI for example can set an "external" key to be pressed or released. Libobs
remembers the state of these external keys and looks them up like it would
look up native platform keys when querying hotkeys.

The existing `inject_event` keeps working for now to not break the API,
but should this PR get merged it would only make sense to remove it in the
future, especially since it still has the issues below. I marked it as deprecated
as part of this PR.

This PR should not conflict with #5095 and #3914.

### Motivation and Context

The existing `obs_hotkey_inject_event` method is a bit problematic since
it partly relies on the injected key combination, but also partly on the
native platform implementation. This is because it not only checks for
the modifiers passed in the key combination, but through an `is_pressed`
call in `query_hotkeys` it also check for modifiers on the platform
implementation. If the platform implementation now either doesn't work
at all (like currently on macOS) or is completely disabled when not in
focus (like with the upcoming macOS implementation), hotkeys injected
with modifiers will fail.

You can reproduce the macOS behaviour on Windows by forcing the
native Windows implementation to always return false on
`obs_hotkeys_platform_is_pressed`.
I made [these](https://github.com/gxalpha/obs-studio/actions/runs/2401073563) CI builds for that.

### How Has This Been Tested?

macOS 12.5 Beta:
Triggering hotkeys with modifiers from the UI now works.
Tested:
- Modifierless hotkeys
- Hotkeys with a single modifier, making sure to both work non-strictly as well as strictly with #5095 applied
- Hotkeys with multiple modifiers -> made sure they don't work with modifiers missing
- Modifier-only hotkeys

Windows 11, tested the same as above with CI builds:
- from the UI
- outside the UI (focus on another window)

I tested as well as I could, but it would be great to see other people test
as well.

### Types of changes

### Checklist:

### DIFF
--- a/libobs/obs-hotkey.c
+++ b/libobs/obs-hotkey.c
@@ -1149,6 +1149,7 @@ void obs_hotkeys_free(void)
 	da_free(obs->hotkeys.bindings);
 	da_free(obs->hotkeys.hotkeys);
 	da_free(obs->hotkeys.hotkey_pairs);
+	da_free(obs->hotkeys.external_keys);
 
 	for (size_t i = 0; i < OBS_KEY_LAST_VALUE; i++) {
 		if (obs->hotkeys.translations[i]) {
@@ -1202,7 +1203,8 @@ static inline bool modifiers_match(obs_hotkey_binding_t *binding,
 static inline bool is_pressed(obs_key_t key)
 {
 	return obs_hotkeys_platform_is_pressed(obs->hotkeys.platform_context,
-					       key);
+					       key) ||
+	       obs_hotkeys_external_is_pressed(key);
 }
 
 static inline void press_released_binding(obs_hotkey_binding_t *binding)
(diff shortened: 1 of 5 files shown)


libobs/obs-hotkey.c
```diff
@@ -1365,6 +1391,23 @@ static inline void query_hotkeys()
 		obs->hotkeys.strict_modifiers,
 	};
 	enum_bindings(query_hotkey, &param);
+#ifdef __APPLE__
+	/* Stupid hack to reset the capslock key after keys have been queried
```
PatTheMav, line 1395: Well in some way it still is - if you press caps lock, the green LED will come on, suggesting it is in an "arrested" state just like the old hardware key was.

Pressing it again will "unlock" it, and as such only then is it a released state. It's the only key with a "hold" state, but given Apple's keyboard layout hasn't changed in decades, it's to be expected. 😏

libobs/obs-hotkey.c
```diff
@@ -1328,6 +1330,30 @@ void obs_hotkey_enable_strict_modifiers(bool enable)
 	unlock();
 }
 
+void obs_hotkey_external_set_pressed(obs_key_t key, bool pressed)
+{
+	if (key == OBS_KEY_NONE)
+		return;
+
+	if (pressed)
```
PatTheMav, line 1338: To ensure thread safety (between the hotkey thread and the UI thread/Qt), I'd suggest wrapping write access in locks, like so:

```
    if (!lock())
        return;

    if (pressed)
		da_push_back(obs->hotkeys.external_keys, &key);
	else
		da_erase_item(obs->hotkeys.external_keys, &key);
    unlock();
```

libobs/obs-hotkey.c
```diff
@@ -1328,6 +1330,30 @@ void obs_hotkey_enable_strict_modifiers(bool enable)
...
+		da_push_back(obs->hotkeys.external_keys, &key);
+	else
+		da_erase_item(obs->hotkeys.external_keys, &key);
+}
+
+void obs_hotkey_external_release_all()
+{
+	if (obs->hotkeys.external_keys.num)
```
PatTheMav, line 1346: See above:

```
    if (obs->hotkeys.external_keys.num) {
        if (!lock())
            return;

        da_erase_range(obs->hotkeys.external_keys, 0,
			       obs->hotkeys.external_keys.num);
        unlock();
    }
```

gxalpha:
Honestly, the more I look at this PR the more I dislike it.
If someone really wants it feel free to reopen, but I'm unhappy with it.

PR #6555, opened 2022-05-29
libobs: Deprecate obs_hotkey_enable_strict_modifiers

gxalpha:
### Description

As discussed, deprecates `obs_hotkey_enable_strict_modifiers` so that we can get rid of it later.
I would probably push to master but
a) I'm too scared and
b) I don't actually know how

- Follow-up to #5095

### Motivation and Context

This got changed in the UI and we don't need it for developers either.

### How Has This Been Tested?

Calling it will now warn that it's deprecated.

### Types of changes

- Code cleanup (non-breaking change which makes code smaller or more readable) I guess? Or at least in the future it will

### Checklist:

### DIFF
--- a/libobs/obs-hotkey.h
+++ b/libobs/obs-hotkey.h
@@ -277,7 +277,7 @@ EXPORT void obs_hotkey_inject_event(obs_key_combination_t hotkey, bool pressed);
 
 EXPORT void obs_hotkey_enable_background_press(bool enable);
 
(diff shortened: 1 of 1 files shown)


tt2468:
I would suggest maybe providing some more context to the discussion within your commit description.

gxalpha:
> I would suggest maybe providing some more context to the discussion within your commit description.

Yup, fixed

PR #6561, opened 2022-05-30
flatpak: Install CMake config files

### LINKED ISSUE #6556: Flatpak builds don't ship CMake files
### Operating System Info

Other

### Other OS

Flatpak runtime

### OBS Studio Version

Git

Latest Flatpak build from master

### OBS Studio Log URL

--

### OBS Studio Crash Log URL

_No response_

### Expected Behavior

The Flatpak builds should ship CMake files, to allow building plugins with it as the runtime.

Most notably:
- `LibObsConfig.cmake`
- `obs-frontend-apiConfig.cmake`

### Current Behavior

The CMake files are not shipped. OBS 27.2.4 shipped `LibObsConfig.cmake`, but now it's also absent in the `master` builds.

### Steps to Reproduce

1. Download and install the latest Flatpak build from `master`
2. Run `flatpak run --command=find com.obsproject.Studio//master /app/lib/cmake -name "*.cmake"`
3. It errors out while it should print the CMake files paths

_No response_

rmnvgr:
### Description

Add a post-install command to the OBS module that installs the `obs_libraries` component, which provides CMake config files.

### Motivation and Context

CMake config files are required for building OBS plugins with the Flatpak as a runtime.

Builds from the `master` branch stopped shipping the `LibObsConfig.cmake` file, and never shipped the `obs-frontend-apiConfig.cmake` file.

Fixes #6556 

### How Has This Been Tested?

Built locally, the resulting Flatpak ships the following CMake config files:

- `/app/lib/cmake/libobs/libobsConfig.cmake`
- `/app/lib/cmake/libobs/libobsConfigVersion.cmake`
- `/app/lib/cmake/libobs/libobsTargets.cmake`
- `/app/lib/cmake/libobs/libobsTargets-release.cmake`
- `/app/lib/cmake/obs-frontend-api/obs-frontend-apiConfig.cmake`
- `/app/lib/cmake/obs-frontend-api/obs-frontend-apiConfigVersion.cmake`
- `/app/lib/cmake/obs-frontend-api/obs-frontend-apiTargets.cmake`
- `/app/lib/cmake/obs-frontend-api/obs-frontend-apiTargets-release.cmake`

### Types of changes

### Checklist:

### DIFF
--- a/CI/flatpak/com.obsproject.Studio.json
+++ b/CI/flatpak/com.obsproject.Studio.json
@@ -444,6 +444,7 @@
         "-DYOUTUBE_SECRET_HASH=$YOUTUBE_SECRET_HASH"
       ],
       "post-install": [
+        "cmake --install . --component obs_libraries",
         "install -d /app/plugins",
         "install -d /app/extensions/Plugins"
       ],


tytan652:
Using a cmake command like this is not a good idea because we can't predict if the flatpak-builder change how cmake is used.
Today it is `.` it can be changed in the flatpak-builder code.

@PatTheMav is trying to automate the installation of `obs_libraries` on Linux builds, which could be a better solution.

PatTheMav:
FWIW that's how I would've solved this originally as well, but then I don't know as much about Flatpak internals. 😄

tytan652:
> FWIW that's how I would've solved this originally as well, but then I don't know as much about Flatpak internals.

Same but since we use the `cmake-ninja` buildsystem, we can end up with a broken manifest because flatpak-builder change where the cmake commands are executed.

rmnvgr:
> FWIW that's how I would've solved this originally as well, but then I don't know as much about Flatpak internals. smile

Well I have another problem: I would have changed this in CMake, but I'm not familiar with it :sweat_smile: 

> Same but since we have Flatpak that use the `cmake-ninja` buildsystem, we can end up with a broken manifest because flatpak-builder could change where the cmake commands are executed.

Changing where the post-install commands are run would likely break not only OBS, but most Flatpak out there, so it'd be safe to assume this behavior won't change.

Although I do agree, this is not the cleanest solution, having the build system install them itself would be better, but again I'm not familiar enough in CMake to propose something else.

tytan652:
> Changing where the post-install commands are run would likely break not only OBS, but most Flatpak out there, so it'd be safe to assume this behavior won't change.
> 
> Although I do agree, this is not the cleanest solution, having the build system install them itself would be better, but again I'm not familiar enough in CMake to propose something else.

I want to precise that this PR provides a clean solution, if we assume that flatpak-builder behavior won't change. 

@GeorgesStavracas, what do you think of this ?

rmnvgr:
> if we assume that flatpak-builder behavior won't change.

I asked the Flatpak team on their Matrix channel, and their answer has been unequivocally "[yes](https://matrix.to/#/!RfXaBjokqHAbzZrgHz:matrix.org/$03UGzGQF4-loG2DiDg-BysG_fzCEW2-q-j4ZjGZjM-0?via=matrix.org&via=gnome.org&via=kde.org)".

GeorgesStavracas (approved):
I'm fine with this. At some point I want to experiment with Meson, but this PR should cover our backs as long as we're using CMake.

PR #6565, opened 2022-05-31
libobs: Should not free audio_output before clearing view->channels

walker-WSH:
### Description

When `stop_audio()` is called,  `view->channels` has not been cleared yet, where maybe storage global output/input audio source.
So variables related with audio should not be deleted in stop_audio().

### Motivation and Context

fix possible crash

### How Has This Been Tested?

enable global desktop audio, exit OBS

### Types of changes

### Checklist:

### DIFF
--- a/libobs/media-io/audio-io.c
+++ b/libobs/media-io/audio-io.c
@@ -400,19 +400,22 @@ fail0:
 	return AUDIO_OUTPUT_FAIL;
 }
 
+void stop_audio_thread(audio_t *audio)
+{
+	if (audio->initialized) {
+		audio->initialized = false;
+
+		os_event_signal(audio->stop_event);
+		pthread_join(audio->thread, NULL);
+	}
+}
+
 void audio_output_close(audio_t *audio)
 {
-	void *thread_ret;
-
 	if (!audio)
 		return;
 
-	if (audio->initialized) {
-		os_event_signal(audio->stop_event);
-		pthread_join(audio->thread, &thread_ret);
-		os_event_destroy(audio->stop_event);
-		pthread_mutex_destroy(&audio->input_mutex);
-	}
+	stop_audio_thread(audio);
 
 	for (size_t mix_idx = 0; mix_idx < MAX_AUDIO_MIXES; mix_idx++) {
 		struct audio_mix *mix = &audio->mixes[mix_idx];
@@ -422,6 +425,9 @@ void audio_output_close(audio_t *audio)
 
 		da_free(mix->inputs);
 	}
+
+	os_event_destroy(audio->stop_event);
+	pthread_mutex_destroy(&audio->input_mutex);
 	bfree(audio);
 }
 
--- a/libobs/media-io/audio-io.h
+++ b/libobs/media-io/audio-io.h
@@ -203,6 +203,8 @@ static inline uint64_t ns_to_audio_frames(size_t sample_rate, uint64_t frames)
 	return util_mul_div64(frames, sample_rate, 1000000000ULL);
 }
 
+void stop_audio_thread(audio_t *audio);
+
 #define AUDIO_OUTPUT_SUCCESS 0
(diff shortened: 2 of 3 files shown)


libobs/media-io/audio-io.c
```diff
@@ -400,19 +400,26 @@ int audio_output_open(audio_t **audio, struct audio_output_info *info)
...
 	if (!audio)
 		return;
 
-	if (audio->initialized) {
-		os_event_signal(audio->stop_event);
-		pthread_join(audio->thread, &thread_ret);
-		os_event_destroy(audio->stop_event);
-		pthread_mutex_destroy(&audio->input_mutex);
-	}
+	stop_audio_thread(audio);
```
norihiro, line 422: If calling like this, the function should not be suffixed `_thread`.

libobs/media-io/audio-io.c
```diff
@@ -400,19 +400,26 @@ int audio_output_open(audio_t **audio, struct audio_output_info *info)
...
+		os_event_signal(audio->stop_event);
+
+		void *thread_ret;
+		pthread_join(audio->thread, &thread_ret);
+	}
+}
+
 void audio_output_close(audio_t *audio)
 {
 	void *thread_ret;
```
norihiro, line 417: `void *thread_ret` is an unused variable now.

tt2468:
Are there reproduction steps for the fixed crash or is it just *possible*?

walker-WSH:
> Are there reproduction steps for the fixed crash or is it just _possible_?

Just possible and I meet this crash once.
Difficult to reproduce

norihiro:
Even if it's difficult to reproduce, could you describe how ` view->channels` holds the global audio sources when entering `obs_shutdown()`? If there is a possibility to cause a crash, I think it is important to fix.

I've added a log message `entering stop_audio...` at the beginning of `stop_audio()` and `entering obs_source_destroy...` at the beginning of `obs_source_destroy()`. Then checked which is called earlier. On my run, `obs_source_destroy()` is called earlier.

My understanding is that `obs_shutdown()` is called after all sources are destroyed. And also `obs_wait_for_destroy_queue()` at the beginning of `obs_shutdown()` waits audio thread has exited.

Just in case a reference-count of a source is not correctly decremented, it is a memory leak rather than an invalid order of the shutdown.

walker-WSH:
> view->channels

`view->channels` holds the reference and it is cleared in `obs_main_view_free`.
However, `obs_main_view_free` is called after `stop_audio`().

norihiro:
> `view->channels` holds the reference of root source, and it is released in `obs_main_view_free`. However, `obs_main_view_free` is called after `stop_audio`().

Before `stop_audio`, `OBSBasic::ClearSceneData()` should have released the main-views. The main-views should not be a concern.

https://github.com/obsproject/obs-studio/blob/b4f7499b33b53351e56a02fc6194f11d8af78da5/UI/window-basic-main.cpp#L4640-L4641

If script or third party plugins have not released a reference-count of an audio source, `stop_audio()` might fail. However I feel such a scenario is not a fault of the shutdown process but a memory leak.

walker-WSH:
> > `view->channels` holds the reference of root source, and it is released in `obs_main_view_free`. However, `obs_main_view_free` is called after `stop_audio`().
> 
> Before `stop_audio`, `OBSBasic::ClearSceneData()` should have released the main-views. The main-views should not be a concern.
> 
> https://github.com/obsproject/obs-studio/blob/b4f7499b33b53351e56a02fc6194f11d8af78da5/UI/window-basic-main.cpp#L4640-L4641
> 
> AFAIK, if script or third party plugins have not released a reference-count of an audio source, `stop_audio()` might fail. However I feel such a scenario is not a fault of the shutdown process but a memory leak.

sorry to response late.

yes, this crash is because of another scenario that any source has not been released completely until stop_audio().

Now I agree with you to ignore this crash. I will close it.
Thannks ~

PR #6568, opened 2022-05-31
CI: Use WindowsQtVersion to install Qt

RytoEX:
### Description

Use WindowsQtVersion instead of WindowsDepsVersion when installing Qt on Windows so that we can specify a different version string for Qt than for the main deps package. This unblocks updating obs-deps when we don't have a new Qt update to ship.

### Motivation and Context

Let's us update the obs-deps version for Windows builds without having a corresponding Qt build available. For now, this is useful if there were no Qt updates for Windows between obs-deps releases. We want to be able to update obs-deps versions.

### How Has This Been Tested?

The PR is the test.

### Types of changes

### Checklist:


RytoEX:
While this will work for a single fresh run, it can get into a failure scenario where if you tried to switch between deps versions, it could break if you end up with deps+Qt versions where the build scripts detect a deps folder that only has half the deps.

I'm not really happy with this, and I'd rather spend time getting Qt builds on CI working so that this isn't necessary.

PR #6569, opened 2022-06-01
UI: Cleanup advanced audio window

cg2121:
### Description
This converts the advanced audio window to use
a ui form, so it is easier to modify in the future.

This also fixes sizing issues with the control widgets,
as before the audio tracks would be clipped, because the
widgets in the window were too wide.

Before:
2022-05-31 08_28_27-Advanced Audio Properties

After:
2022-05-31 20_10_43-Advanced Audio Properties

### Motivation and Context
Makes dialog easier to edit and makes it look better.

### How Has This Been Tested?
Opened dialog to make sure everything was working as expected.

### Types of changes

### Checklist:

### DIFF
--- a/UI/window-basic-adv-audio.cpp
+++ b/UI/window-basic-adv-audio.cpp
@@ -1,9 +1,3 @@
-#include <QVBoxLayout>
-#include <QHBoxLayout>
-#include <QGridLayout>
-#include <QScrollArea>
-#include <QPushButton>
-#include <QLabel>
 #include "window-basic-adv-audio.hpp"
 #include "window-basic-main.hpp"
 #include "item-widget-helpers.hpp"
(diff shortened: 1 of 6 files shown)


WizardCM:
I had started working on similar changes to this dialog, but ran into an issue because I attempted to entirely rebuild the way the rows themselves are structured, which end up being more complicated than expected.

Thanks for getting the ball properly rolling on this one. :)

UI/window-basic-adv-audio.cpp
```diff
@@ -224,9 +128,13 @@ void OBSBasicAdvAudio::SetVolumeType(bool percent)
 
 	config_set_int(GetGlobalConfig(), "BasicWindow", "AdvAudioVolumeType",
 		       (int)type);
+
+	ui->usePercent->blockSignals(true);
+	ui->usePercent->setChecked(checked);
+	ui->usePercent->blockSignals(false);
```
WizardCM, line 134:
```suggestion
```

This likely shouldn't be within the on_usePercent_toggled function, as the intent is that this function is called when the checkbox is toggled.

Instead, `blockSignals` should be used in the init function (surrounding a `ui->usePercent->setChecked()` call), and this bit of code

UI/window-basic-adv-audio.cpp
```diff
@@ -1,134 +1,38 @@
...
-	vlayout->setContentsMargins(11, 11, 11, 11);
-	vlayout->addWidget(scrollArea);
-	vlayout->addLayout(buttonLayout);
-	setLayout(vlayout);
-
-	connect(activeOnly, SIGNAL(clicked(bool)), this,
-		SLOT(ActiveOnlyChanged(bool)));
-
-	connect(closeButton, &QPushButton::clicked, [this]() { close(); });
+		on_usePercent_toggled(true);
```
WizardCM, line 27:
```suggestion
		ui->usePercent->blockSignals(true);
		ui->usePercent->setChecked(true);
		ui->usePercent->blockSignals(false);
	}
```

UI/window-basic-adv-audio.cpp
```diff
@@ -1,134 +1,38 @@
...
-	usePercent->setStyleSheet("font-weight: bold;");
-	usePercent->setText("%");
-	connect(usePercent, SIGNAL(toggled(bool)), this,
-		SLOT(SetVolumeType(bool)));
+	ui->setupUi(this);
 
 	VolumeType volType = (VolumeType)config_get_int(
 		GetGlobalConfig(), "BasicWindow", "AdvAudioVolumeType");
 
 	if (volType == VolumeType::Percent)
```
WizardCM, line 26:
```suggestion
	if (volType == VolumeType::Percent) {
```

UI/window-basic-adv-audio.cpp
```diff
@@ -237,7 +145,11 @@ void OBSBasicAdvAudio::SetShowInactive(bool show)
 		return;
 
 	showInactive = show;
-	activeOnly->setChecked(!showInactive);
+
+	ui->activeOnly->blockSignals(true);
+	ui->activeOnly->setChecked(!showInactive);
```
WizardCM, line 150: Same here. We want to respect the Qt event flow. `on_activeOnly_toggled` is called when activeOnly has been toggled, so we shouldn't toggle it again..

WizardCM (changes requested):
Overall the code looks good, and the UI functions as expected. Just a minor change and this'll be good to merge.

UI/window-basic-adv-audio.cpp
cg2121: Fixed. I didn't add the block signals here because we still need to have to toggle function called.

UI/window-basic-adv-audio.cpp
cg2121: These have now been removed.

WizardCM (approved):

PR #6574, opened 2022-06-01
rtmp-services: Add Eventials

evandroduarte:
[Eventials](https://www.eventials.com/) is a platform to broadcast webinars and events.

Eventials is a Brazilian company and was founded in 2010.

### DIFF
--- a/plugins/rtmp-services/data/package.json
+++ b/plugins/rtmp-services/data/package.json
@@ -1,10 +1,10 @@
 {
 	"url": "https://obsproject.com/obs2_update/rtmp-services",
-	"version": 73,
+	"version": 74,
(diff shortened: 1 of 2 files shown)


Fenrirthviti:
Please properly fill out the PR template instead of deleting it.

derrod (changes requested):
Eventials already exists, please update the existing service rather than adding a new one. It also looks like you're basing this on an ancient branch that I'm genuienly baffled it's not causing a merge conflict.

derrod:
Actually let's just stop this here. You should re-submit this PR updating the existing service from a non-master branch that isn't 5 years out of date.

PR #6577, opened 2022-06-01
libobs: Add support for multiple video mixes

chippydip:
### Description
Add the ability to render multiple "video mixes" where a "video mix" is ultimately a `render_texture` created by compositing scenes and sources that can be fed into one or more video encoders/outputs. 

This implementation splits `obs_core_video` keeping global resources in place and moving per-mix resources to a new `obs_core_video_mix` struct. It then updates the main graphics thread to render each channel of the `main_view` onto a separate `render_texture` / `obs_core_video_mix` rather than rendering them one on top of another on the same `render_texture` (I don't believe this old behavior was actual used since scenes and transitions achieve the same effect with a single `main_view` channel). GPU resource allocation has been deferred until the first render tick when a channel is actually in use for video to avoid re-allocating GPU resources for all 64 available channels.

More details on the design can be found in the [Multiple Video Mixes RFC](https://github.com/chippydip/obs-rfcs/blob/multiple-video-mixes/text/0049-multiple-video-mixes.md) and associated [PR discussion](https://github.com/obsproject/rfcs/pull/49).

### Virtual Camera
While this refactor was designed to support "ISO Recording" as an eventual use-case, the primary motivation was to enable source selection for the Virtual Camera feature. This enables users to export just their webcam (including filters) to be used in a video conference while also streaming/recording a fully composited scene.

Settings have been added to select a source override in both SImple:
image

and Advanced output modes (in a new Video tab):
image

These settings are kept in sync and are used when starting the virtual camera. If an override is specified, then channel 7 is configured with the selected source and used for the virtual camera. Otherwise the default channel 0 is used as normal.

### Reviewing Changes
To make reviewing these changes easier I've separated the PR into a couple of individual diffs:
- [libobs: Add support for multiple video mixes](https://github.com/obsproject/obs-studio/pull/6577/commits/16d65849f33b2df4b3a2bca5a0ac43f4d8d9128d)
- [docs: Update render pipeline descriptions for multiple video mixes](https://github.com/obsproject/obs-studio/pull/6577/commits/2d19894b6e034080902aaad05d67ace92c28ea86)
- [UI: Add Virtual Camera source selector to settings](https://github.com/obsproject/obs-studio/pull/6577/commits/3e6766663bfcdfc74478f579428c953603d2c87d)
- [libobs: Format changes for multiple video mixes](https://github.com/obsproject/obs-studio/pull/6577/commits/6f8583ac499eefda934b34ae90458f4e8a547d68) (clang-format only to remove noise from main commit above)

### Motivation and Context
OBS supports multiple video outputs allowing simultaneous streaming, recording, virtual camera, etc. These outputs, however, must all be derived from a single `render_texture`. This makes it impossible to support [Multiple Video Outputs (Selective Recording, ISO recording)](https://ideas.obsproject.com/posts/41/multiple-video-outputs-selective-recording-iso-recording) or to share just a webcam via virtual camera (to participate in a video conference while streaming) with the current architecture.

While plugins do exist to implement these features, they do so in a roundabout way. They provide a no-op filter which intercepts video from a source for the plugin to process. This completely bypasses the normal output pipeline of OBS.

### How Has This Been Tested?
- Tested that existing recording functionality continues to work as expected
- Added test code to set a webcam source as an additional view and use that for virtual camera output while also recording
- Tested on Win10

### Types of changes

  - This is technically a breaking change since it modifies how the renderer behaves when multiple `main_view` channels are used for video, but I don't believe this functionality was being used in practice so it should be more of a non-breaking feature addition in practice.

### Checklist:

### DIFF
--- a/libobs/obs-view.c
+++ b/libobs/obs-view.c
@@ -139,3 +139,50 @@ void obs_view_render(obs_view_t *view)
 
 	pthread_mutex_unlock(&view->channels_mutex);
 }
+
+static inline size_t find_mix_for_view(obs_view_t *view)
+{
+	for (size_t i = 0, num = obs->video.mixes.num; i < num; i++) {
+		if (obs->video.mixes.array[i].view == view)
+			return i;
+	}
+
+	return DARRAY_INVALID;
+}
+
+static inline void set_main_mix()
+{
+	size_t idx = find_mix_for_view(&obs->data.main_view);
+
+	struct obs_core_video_mix *mix = NULL;
+	if (idx != DARRAY_INVALID)
+		mix = obs->video.mixes.array + idx;
+	obs->video.main_mix = mix;
+}
+
+video_t *obs_view_add(obs_view_t *view)
+{
+	struct obs_core_video_mix mix = {0};
+	mix.view = view;
+	if (obs_init_video_mix(&obs->video.ovi, &mix) != 0) {
+		obs_free_video_mix(&mix);
+		return NULL;
+	}
+
+	pthread_mutex_lock(&obs->video.mixes_mutex);
+	da_push_back(obs->video.mixes, &mix);
+	set_main_mix();
+	pthread_mutex_unlock(&obs->video.mixes_mutex);
+
+	return mix.video;
+}
+
+void obs_view_remove(obs_view_t *view)
+{
+	pthread_mutex_lock(&obs->video.mixes_mutex);
+	size_t idx = find_mix_for_view(view);
+	if (idx != DARRAY_INVALID)
+		obs->video.mixes.array[idx].view = NULL;
+	set_main_mix();
+	pthread_mutex_unlock(&obs->video.mixes_mutex);
+}
--- a/UI/CMakeLists.txt
+++ b/UI/CMakeLists.txt
@@ -103,6 +103,7 @@ target_sources(
           forms/OBSBasicSettings.ui
           forms/OBSBasicSourceSelect.ui
           forms/OBSBasicTransform.ui
+          forms/OBSBasicVCamConfig.ui
           forms/OBSExtraBrowsers.ui
           forms/OBSImporter.ui
           forms/OBSLogReply.ui
@@ -257,6 +258,8 @@ target_sources(
           window-basic-transform.cpp
           window-basic-transform.hpp
           window-basic-preview.hpp
+          window-basic-vcam-config.cpp
+          window-basic-vcam-config.hpp
           window-dock.cpp
           window-dock.hpp
           window-importer.cpp
--- a/UI/data/locale/en-US.ini
+++ b/UI/data/locale/en-US.ini
@@ -700,6 +700,17 @@ Basic.Main.Ungroup="Ungroup"
 Basic.Main.GridMode="Grid Mode"
 Basic.Main.ListMode="List Mode"
 
+# virtual camera configuration
+Basic.Main.VirtualCamConfig="Configure Virtual Camera"
+Basic.VCam.VirtualCamera="Virtual Camera"
+Basic.VCam.OutputType="Output Type"
+Basic.VCam.OutputSelection="Output Selection"
+Basic.VCam.Internal="Internal"
+Basic.VCam.InternalDefault="Program Output (Default)"
+Basic.VCam.InternalPreview="Preview Output"
+Basic.VCam.Start="Start"
+Basic.VCam.Update="Update"
+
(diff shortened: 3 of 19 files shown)


jp9000:
I am impressed by the amount of lines of code you did this with.

libobs/obs-view.c
```diff
@@ -139,3 +139,29 @@ void obs_view_render(obs_view_t *view)
...
+	if (channel >= MAX_CHANNELS)
+		return NULL;
+
+	pthread_mutex_lock(&view->channels_mutex);
+	source = view->channels[channel];
+	if (source) {
+		if (source->removed) {
+			obs_source_release(source);
+			view->channels[channel] = NULL;
+			source = NULL;
```
chippydip, line 159: `obs_view_render` was the only place this removal logic was happening so to preserve that behavior I added this new function instead of just using `obs_view_get_source`.

tt2468:
Just curious if you have any examples where this new functionality could be used to say, route an `obs_source_t` as an aux output to an `obs_output_t` object?

chippydip:
This is the test code I was using to set my "Webcam" source an an aux output when starting the virtual camera. Obviously needs some proper settings UI, but the API is very straightforward to use:
image

chippydip:
Updated configuration UX based on https://github.com/obsproject/rfcs/pull/49

Added settings cog next to Start Virtual Camera button:
image

Added support for both default and preview outputs:
image

Added options to select a specific scene (in default order) as output:
image

Added options to select an individual source (in alphabetical order) as output:
image

Closing the dialog will cancel any changes. Pressing the start button with save them and start the virtual cam output. Output can be changed without needing to restart the virtual cam (in which case the "Start" button becomes an "Update" button).

Settings are now saved as part of the scene collection.

RytoEX:
This PR introduces memory leaks and may introduce a crash on exit. Tests were done on Windows 10 and Windows 11. Simply opening the Windows 64-bit build and closing it results in 15 memory leaks in our leak counter (visible at the end of the OBS log file after closing OBS).

I experienced a crash on exit after having opened OBS, adding a single source (color source is sufficient), and then closing OBS. You need not start the virtual camera. Perhaps of note, while I can reliably reproduce this crash on a system with an AMD CPU+GPU, I have been unable to reproduce this on a system with an Intel CPU and NVIDIA GPU. Something to perhaps consider while trying to reproduce.

chippydip:
Thanks for testing and for pointing out the memory leak info in the log!

I found and fixed 2 issues and am now seeing 0 memory leaks locally. I'll get the fixes squashed in and pushed up tomorrow:
- There was a missing call to `obs_free_render_textures()` from `obs_free_video_mix()` in obs.c
- There was an early return in `render_main_texture()` in obs-video.c which was skipping a `obs_source_release()` call

I wasn't able to repro the crash on exit, but I only have access to NVIDIA GPUs right now. It's possible the crash was an additional side effect of not freeing the render textures before shutdown? Several surfaces were not being unmapped and textures were not being destroyed as a result, If not, would it be possible to grab a stack trace of the crash on the AMD machine to help debug?

RytoEX (approved):
I've retested this, and the previously observed memory leaks and crashes appear to be fixed. Functionality all works as described. Great work!

I'm tentatively fine with merging this as-is, though I'm aware that @jp9000 would like to take a look at this first.

jp9000:
Okay, I finally had a chance to do a bit more of an in-depth review of this.

A big issue with this PR is that as currently designed, it creates 64 video mixes. Which makes sense at first glance, but the problem with that is for each video mix, there's a GPU encode thread, and a video thread (via the video_t object which each one has). so it's creating 128 threads where as we only had 2 threads before. In addition to the threading issue, it's also now renders 64 times per main render call as well.

I think linking this up to MAX_CHANNELS might not be ideal. There has to be a more optimal way to deal with this with less overhead. Maybe lazy loading of mixes (with the exception to the main mix)?

chippydip:
@jp9000 Thanks for taking a look. I did some refactoring and was able to revert the MAX_CHANNELS stuff and instead added `obs_view_add()` and `obs_view_remove()` functions to `libobs` API to allow additional views to be added to the main render loop. Once added, a new video mix is created for them to handle the rendering. Let me know what you think.

SuslikV:
Names for the options:

Feeder Type
---Final Composition
---Source
---Scene

Feeder*
---Program Output (Default)
---Preview Output

---Scene
---Scene 2

---Text (GDI+)
---Game Capture

*Tip:  Source of the video

PR #6583, opened 2022-06-02
UI: Always use "OBS Studio" in main window title bar

gxalpha:
### Description

Changes the window title to always say "OBS Studio", whereas before it said "OBS" normally and "OBS Studio" in Studio Mode

Before, when toggling Studio Mode on and off:

https://user-images.githubusercontent.com/59806498/171715969-ada4957b-3a9b-4677-9fa0-871e26f9f7e3.mov

After, always
image

### Motivation and Context

This differentiation made sense when Studio Mode was introduced (and the app still known as OBS Multiplatform) to differentiate it from what was "normal" OBS behaviour, but these days "OBS Studio" is known as the app name so that is what it should probably say all the time, not just when you're in Studio Mode.

### How Has This Been Tested?

Compiled and opened.
See screenshot above.
No difference whether Studio Mode is on or off.

### Types of changes

### Checklist:

### DIFF
--- a/UI/window-basic-main-transitions.cpp
+++ b/UI/window-basic-main-transitions.cpp
@@ -1739,7 +1739,6 @@ void OBSBasic::SetPreviewProgramMode(bool enabled)
 	}
 
 	ResetUI();
-	UpdateTitleBar();
 }
 
 void OBSBasic::RenderProgram(void *data, uint32_t cx, uint32_t cy)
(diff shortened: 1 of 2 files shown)


gxalpha:
> tbh I think this was mostly a cute easter egg, and I don't really see a reason to remove it.

i didn't see it as an easter egg and more, probably because i hate fun :P I'd remove it mostly for the reasons I stated above, but if people think it should be kept i guess we could always close the PR

jp9000:
Author hates fun confirmed.

Fenrirthviti:
I also agree that this was just a fun little easter egg, and since there is no practical benefit to changing it, I am opposed to the change. I understand the reason behind submitting it, but I'll go ahead and make the decision to close for now as discussion here and off-thread seems to indicate the majority would like to keep it the way it is.

PR #6585, opened 2022-06-03
UI, libobs: Add lock aspect ratio to transform dialog

cg2121:
### Description
This adds the ability to lock the aspect ratio for the size
and bounding box, in the transform dialog.

2022-06-03 05_35_32-Edit Transform for 'Image Slide Show'

### Motivation and Context
https://ideas.obsproject.com/posts/1338/button-to-lock-aspect-ratio-under-transform

### How Has This Been Tested?
Locked the aspect ratio and made sure the width/height changed accordingly.

### Types of changes

### Checklist:

### DIFF
--- a/UI/data/locale/en-US.ini
+++ b/UI/data/locale/en-US.ini
@@ -626,6 +626,7 @@ Basic.TransformWindow.CropLeft="Crop Left"
 Basic.TransformWindow.CropRight="Crop Right"
 Basic.TransformWindow.CropTop="Crop Top"
 Basic.TransformWindow.CropBottom="Crop Bottom"
+Basic.TransformWindow.LockAspectRatio="Lock Aspect Ratio"
 
 Basic.TransformWindow.Alignment.TopLeft="Top Left"
 Basic.TransformWindow.Alignment.TopCenter="Top Center"
--- a/UI/forms/OBSBasicTransform.ui
+++ b/UI/forms/OBSBasicTransform.ui
@@ -6,8 +6,8 @@
    <rect>
     <x>0</x>
     <y>0</y>
-    <width>564</width>
-    <height>313</height>
+    <width>778</width>
+    <height>378</height>
    </rect>
   </property>
   <property name="windowTitle">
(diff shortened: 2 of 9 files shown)


Nofew:
The lock/unlock button does not visually appear to be clickable. Can some sort of border be added to suggest that it can be interacted with?

cg2121:
IMO, I don't think that's necessary, as it is the same as every other lock icon in the program.

Nofew:
There's other lock icons? And they're clickable but don't appear to be? This sounds like something that should probably be changed across OBS as a whole. There's precedent for it in Gimp.

gxalpha:
I agree with @nofew, it’s not really obvious that the lock can be clicked.

Unrelated, since this appears to get stored in the sceneitem properties and isn’t just a temporary dialog thing, how would this interact with scene editing in the preview? As in, is there a difference whether one of the locks is clicked and if not, should there be?

cg2121:
This only affects changing the values in the transform dialog.

dodgepong:
Thirding the request to make the lock button more obviously clickable/toggleable. That was my immediate thought upon seeing the screenshot.

That said, I'm definitely thankful for this change, it's something I've heard people ask for and it makes a lot of sense.

Fenrirthviti:
I would agree that while this is a really nice QoL feature to have for the transform dialog, as a user I would also expect locking the aspect ratio to propagate down with any future changes/resizing I do in the preview itself. It's not immediately obvious that locking the ratio is ONLY for manually entering values, as other changes are passed down to the preview (such as the bounding box). This creates inconsistent UX. Is it possible to lock the ratio in the preview as well?

cg2121:
The aspect ratio is already stays the same when dragging to resize in the preview, except when also holding the shift key down. Since the shift key is an extra step, I think that it is fine not to lock the aspect ratio for this.

cg2121:
Updated to also lock the aspect ratio when dragging in the preview.

Warchamp7:
> Updated to also lock the aspect ratio when dragging in the preview.

I'm not sure I like the idea of being able to apply a lock in the transform dialog that is then only visible there.

Is it difficult to modify the canvas selection outlines? A modification to indicate locked ratio would be helpful here, and also handy for making Shift + Drag resizing more obvious to users by appearing when Shift is held

image

SuslikV:
The lock is not the link.

Guys, have you ever edit an images in some editors?
The transform values usually vertically organized (or narrow, single line widgets) and has to be placed on sides of the screen just not to cover the main edit area. This is obvious for the user of any editor.

And this PR change sets even more limits on the _Transform_ window minimum width while preserving its height. It's no good.

The lock(the bind) aspect ratio tool button can be placed **between the values** similar to other editors. To not confuse the users.

More over, there is bug with the _Transform_ window update - the size of grouped sources after the whole group was resized doesn't updates the values in the _Transform_ window until the  _Transform_ window re-opened by the user:
1) select group of few items
2) open Transform window (title says Edit Transform for Group)
3) resize the whole group in preview by mouse (values changes, no issues)
4) select the source inside the group (title says Edit Transform for Source's Name)
5) resize the selected source by mouse in preview (no values for the fields changes in the _Transform_ window !!!)
6) try to select the group again (title says No Item Selected)
7) close  _Transform_ window and re-open it again (updated values are finally shown)

This bug should be addressed first, I think. Then new features can be added to the  _Transform_ window. Or you may step onto unknown lands of the selected items, locked aspect ratios and not selected sources.

Warchamp7:
As currently implemented, this is not going to be merged due to the above concerns. This dialog needs a larger overhaul with features like this in consideration.

PR #6606, opened 2022-06-08
UI: handle theme file names with "." characters

OldBaldGeek:
### Description

The Settings dialog derives theme names from names of the theme files, nominally by removing the ".qss" extension. However, the current code parses the name as "everything left of the first '.'" rather than "everything left of the LAST '.'"

Thus, if there is a file with a name such as "Dark-V27.2.4.qss" in the themes directory, the Settings dialog will show it as "Dark-V27", and an attempt to select the theme will cause an error. This PR corrects the parsing.

files
Theme directory including theme file with "." characters

Ver27 2 4-composite
Setting dialog in 27.2.4 truncates at first "." Selecting the truncated name shows an error, since file can't be loaded.

Fix
Fix correctly shows the name up to the .qss extension.

### Motivation and Context

Fixes a bug in parsing of theme filenames which prevented some valid names from being used.

### How Has This Been Tested?

Tested on Windows 10.

- Rename a theme file to include one or several "." characters before the extension. For example, make a copy of Dark.qss and call it "Dark27.2.4.qss"
- Open the OBS Settings dialog and drop the "Themes" list. Verify correct or incorrect parsing of the theme filename
- Select the theme in question and verify that it is loaded correctly
- Close the Settings dialog and exit OBS
- Run OBS and verify that it correctly loads the selected theme

### Types of changes

### Checklist:

### DIFF
--- a/UI/window-basic-settings.cpp
+++ b/UI/window-basic-settings.cpp
@@ -1215,7 +1215,7 @@ void OBSBasicSettings::LoadThemeList()
 				QDir::Files);
 		while (it.hasNext()) {
 			it.next();
-			QString name = it.fileName().section(".", 0, 0);
+			QString name = it.fileInfo().completeBaseName();
 			ui->theme->addItem(name);
 			uniqueSet.insert(name);
 		}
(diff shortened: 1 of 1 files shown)


jp9000:
Wouldn't QFileInfo be the better option rather than trying to parse strings manually?

PR #6617, opened 2022-06-15
cmake: Enable ENABLE_HEVC by default

jpark37:
### Description
Enable HEVC code by default.

### Motivation and Context
Widespread hardware encode/decode support.

### How Has This Been Tested?
Verified HEVC was enabled by default in fresh CMake build directory. Was able to make HEVC recording with NVENC.

### Types of changes

### Checklist:

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -32,7 +32,7 @@ if(NOT CMAKE_BUILD_TYPE)
 endif()
 
 # Global project options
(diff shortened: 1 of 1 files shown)


tytan652:
As it is, OBS allow users to use HEVC on any services because the following line without checking for compatiblity:
https://github.com/obsproject/obs-studio/blob/master/UI/window-basic-settings.cpp#L1015

https://github.com/obsproject/rfcs/pull/45 is also meant to avoid this scenario.

PR #6621, opened 2022-06-17
aja: Implement buffering property

norihiro:
### Description

Add a property `Use Buffering` to AJA source.
Also set the default to unbuffering, which is a change of the behavior.

### Motivation and Context

Buffering inside libobs is enabled by default. The latency of the buffering changes every startup, which is a problem to adjust delay of multi-camera setup or lip sync with other sources.

The `timestamp` for the video frame is taken from `os_gettime_ns()`, not taken from the timestamp of the received frames. I don't think buffering is not necessary for this source.
https://github.com/obsproject/obs-studio/blob/c5015d0e6cb180d138e1a9e2258afd903254b9ea/plugins/aja/aja-source.cpp#L381

### How Has This Been Tested?

Added a `blog` call in `obs_source_set_async_unbuffered` and checked the setting is sent to libobs. This PR does not contain the added `blog`.

### Types of changes

### Checklist:

### DIFF
--- a/plugins/aja/aja-source.cpp
+++ b/plugins/aja/aja-source.cpp
@@ -913,6 +913,10 @@ static void aja_source_update(void *data, obs_data_t *settings)
 	const std::string &wantCardID =
 		obs_data_get_string(settings, kUIPropDevice.id);
 
+	obs_source_set_async_unbuffered(
+		ajaSource->GetOBSSource(),
+		!obs_data_get_bool(settings, kUIPropBuffering.id));
+
 	const std::string &currentCardID = ajaSource->CardID();
 	if (wantCardID != currentCardID) {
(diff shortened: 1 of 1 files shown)


WizardCM:
Depending on how we feel about this change generally, it may be worth versioning the source and keeping the old default the same for v1 users and defaulting buffering to off for v2 users (basically: users who create the source after this PR is merged).

paulh-aja:
I'm inclined to agree that the correct approach is to discover the existing setting, otherwise default to buffering off.

> Depending on how we feel about this change generally, it may be worth versioning the source and keeping the old default the same for v1 users and defaulting buffering to off for v2 users (basically: users who create the source after this PR is merged).

PR #6630, opened 2022-06-20
cmake: Fix build architectures and deployment target not set to defaults

PatTheMav:
### Description
CMake seems to set either `CMAKE_OSX_ARCHITECTURES` and `CMAKE_OSX_DEPLOYMENT_TARGET` implicitly to empty cache variables when not defined explicitly on the commandline or via a GUI.

Setting a default thus requires the `FORCE` flag to overwrite these empty defaults with our desired results. The branches ensure that user-provided non-empty values are not overwritten as well.

### Motivation and Context
Ensure OBS is built with correct architecture-appropriate build flags on macOS.

### How Has This Been Tested?
* Tested with several CMake configuration runs locally

### Types of changes

### Checklist:

### DIFF
--- a/cmake/Modules/ObsDefaults_macOS.cmake
+++ b/cmake/Modules/ObsDefaults_macOS.cmake
@@ -14,17 +14,30 @@ if(POLICY CMP0025)
 endif()
 
 # Build options
-set(CMAKE_OSX_ARCHITECTURES
-    "x86_64"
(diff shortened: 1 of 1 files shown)


PatTheMav:
@WizardCM Do you remember what we decided as next steps for this? I guess I could raise the limit to 10.15, though that should be in sync with the workflow files and build scripts, so I'd rather scope that into a second PR.

PR #6631, opened 2022-06-20
mac-avcapture: Fix deprecation warning for AVCaptureDevice list

PatTheMav:
### Description
Direct access to `[AVCaptureDevice devices]` is deprecated since macOS 10.15. An `AVCaptureDeviceDiscoverySession` is to be used instead, which this PR implements.

When running on older macOS versions or compiling against older platform SDKs, the deprecated method is used.

PS: The formatting is not proper ObjC style, but that's clang-format's doing with our current settings.

### Motivation and Context
Prepare for lowest supported macOS version increase to 10.15 and be more up-to-date with Apple's internal APIs.

### How Has This Been Tested?
* Compiled on M1 Mac with macOS 11.0 target to check that the new API is used and expected video capture devices are visible.

### Types of changes

### Checklist:

### DIFF
--- a/plugins/mac-avcapture/av-capture.mm
+++ b/plugins/mac-avcapture/av-capture.mm
@@ -3,6 +3,7 @@
 #import <CoreMedia/CoreMedia.h>
 #import <CoreVideo/CoreVideo.h>
 #import <CoreMediaIO/CMIOHardware.h>
+#include <AvailabilityMacros.h>
 
 #include <obs-module.h>
 #include <obs.hpp>
@@ -2241,7 +2242,41 @@ static obs_properties_t *av_capture_properties(void *data)
 		OBS_COMBO_FORMAT_STRING);
 	obs_property_list_add_string(dev_list, "", "");
 
-	for (AVCaptureDevice *dev in [AVCaptureDevice devices]) {
+	NSArray *devices = nil;
+#if (__MAC_OS_X_VERSION_MAX_ALLOWED >= MAC_OS_X_VERSION_10_15)
+	if (@available(macOS 10.15, *)) {
+		AVCaptureDeviceDiscoverySession *mediaDeviceDiscoverySession =
+			[AVCaptureDeviceDiscoverySession
+				discoverySessionWithDeviceTypes:@[
+					AVCaptureDeviceTypeBuiltInWideAngleCamera,
+					AVCaptureDeviceTypeExternalUnknown
+				]
+						      mediaType:AVMediaTypeVideo
+						       position:AVCaptureDevicePositionUnspecified];
+		NSArray *mediaDevices = [mediaDeviceDiscoverySession devices];
+
+		AVCaptureDeviceDiscoverySession *muxedDeviceDiscoverySession =
+			[AVCaptureDeviceDiscoverySession
+				discoverySessionWithDeviceTypes:@[
+					AVCaptureDeviceTypeExternalUnknown
+				]
+						      mediaType:AVMediaTypeMuxed
+						       position:AVCaptureDevicePositionUnspecified];
+		NSArray *muxedDevices = [muxedDeviceDiscoverySession devices];
+
+		devices = [mediaDevices
+			arrayByAddingObjectsFromArray:muxedDevices];
+	} else {
+#pragma clang diagnostic push
+#pragma clang diagnostic ignored "-Wdeprecated-declarations"
+		devices = [AVCaptureDevice devices];
+#pragma clang diagnostic pop
+	}
+#else
+	devices = [AVCaptureDevice devices];
+#endif
+
+	for (AVCaptureDevice *dev in devices) {
 		if ([dev hasMediaType:AVMediaTypeVideo] ||
 		    [dev hasMediaType:AVMediaTypeMuxed]) {
 			obs_property_list_add_string(


plugins/mac-avcapture/av-capture.mm
```diff
@@ -2236,7 +2237,28 @@ static void add_manual_properties(obs_properties_t *props)
...
-	for (AVCaptureDevice *dev in [AVCaptureDevice devices]) {
+	NSArray *devices = nil;
+#if (__MAC_OS_X_VERSION_MAX_ALLOWED >= MAC_OS_X_VERSION_10_15)
+	if (@available(macOS 10.15, *)) {
+		AVCaptureDeviceDiscoverySession *deviceDiscoverySession = [AVCaptureDeviceDiscoverySession
+			discoverySessionWithDeviceTypes:@[
+				AVCaptureDeviceTypeBuiltInWideAngleCamera,
+				AVCaptureDeviceTypeExternalUnknown
+			]
+					      mediaType:AVMediaTypeVideo
```
gxalpha, line 2248: This will need `AVMediaTypeMuxed` as well so that we still pick up the screen of connected i(Pad)OS devices (and other cameras that could potentially have sound)

plugins/mac-avcapture/av-capture.mm
```diff
@@ -2236,7 +2237,28 @@ static void add_manual_properties(obs_properties_t *props)
...
 
-	for (AVCaptureDevice *dev in [AVCaptureDevice devices]) {
+	NSArray *devices = nil;
+#if (__MAC_OS_X_VERSION_MAX_ALLOWED >= MAC_OS_X_VERSION_10_15)
+	if (@available(macOS 10.15, *)) {
+		AVCaptureDeviceDiscoverySession *deviceDiscoverySession = [AVCaptureDeviceDiscoverySession
+			discoverySessionWithDeviceTypes:@[
+				AVCaptureDeviceTypeBuiltInWideAngleCamera,
+				AVCaptureDeviceTypeExternalUnknown
+			]
```
gxalpha, line 2247: Could we include DeskViewCamera here?
Documentation implies we might not (`You can use this device type with AVCaptureMultiCamSession`), but in other cases of special cameras the docs explicitly say `You can only use this...`, so I'm unsure (and I don't have an iOS device to test).

gxalpha (changes requested):
See comment regarding muxed

plugins/mac-avcapture/av-capture.mm
```diff
@@ -2236,7 +2237,28 @@ static void add_manual_properties(obs_properties_t *props)
 		OBS_COMBO_FORMAT_STRING);
 	obs_property_list_add_string(dev_list, "", "");
 
-	for (AVCaptureDevice *dev in [AVCaptureDevice devices]) {
+	NSArray *devices = nil;
+#if (__MAC_OS_X_VERSION_MAX_ALLOWED >= MAC_OS_X_VERSION_10_15)
```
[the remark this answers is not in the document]
PatTheMav: Did we decide on that already? Lowest _supported_ macOS version should be 10.15 for x86_64-based systems, but we want to make the lowest supported _development macOS version_ (that is, the version used to compile OBS) be 11.0?

plugins/mac-avcapture/av-capture.mm
PatTheMav: That would require a second discovery session, merging the results of both arrays, as `AVCaptureDeviceDiscoverySession` only allows a single `mediaType` value.

plugins/mac-avcapture/av-capture.mm
PatTheMav: According to the docs, only valid `AVCaptureDeviceType` values are allowed, can be one of:

* AVCaptureDeviceTypeBuiltInDualCamera
* AVCaptureDeviceTypeBuiltInDualWideCamera
* AVCaptureDeviceTypeBuiltInTripleCamera
* AVCaptureDeviceTypeBuiltInWideAngleCamera
* AVCaptureDeviceTypeBuiltInUltraWideCamera
* AVCaptureDeviceTypeBuiltInTelephotoCamera
* AVCaptureDeviceTypeBuiltInLiDARDepthCamera
* AVCaptureDeviceTypeBuiltInTrueDepthCamera
* AVCaptureDeviceTypeBuiltInMicrophone
* AVCaptureDeviceTypeExternalUnknown

(See also: https://developer.apple.com/documentation/avfoundation/avcapturedevicetype)

plugins/mac-avcapture/av-capture.mm
PatTheMav: FWIW, implemented just that - got my iPhone camera as well as the built-in Studio Display camera.

Will push this once we cleared up the other comments.

plugins/mac-avcapture/av-capture.mm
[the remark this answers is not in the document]
gxalpha: I thought we decided on that (or rather generally on the current version and one lower) - though we didn't actively break anything since, so that decision could be reverted if we wanted.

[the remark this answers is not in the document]
RytoEX: > Did we decide on that already? Lowest _supported_ macOS version should be 10.15 for x86_64-based systems, but we want to make the lowest supported _development macOS version_ (that is, the version used to compile OBS) be 11.0?

I also thought we had decided on this split (minimum 10.15 for runtime, minimum 11.0 for development/building). I'm fine either way and have no strong opinion on this particular bit of code.

plugins/mac-avcapture/av-capture.mm
gxalpha: Discussed off-thread that this can be implemented later.

plugins/mac-avcapture/av-capture.mm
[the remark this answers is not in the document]
PatTheMav: Alright, so here's my suggestion: Given that we don't have a PR merged that "officially" increases the lower limit to 10.15 across the code base (meaning raising that in CMake files, build scripts, workflow files), etc. I'd rather have this PR assume this to haven't happened yet and thus work on master branch "as-is".

I'm fine with removing this in a later pass (which we'd have to do anyway) when we remove this and other workarounds for older macOS versions.

Thoughts?

plugins/mac-avcapture/av-capture.mm
PatTheMav: @gxalpha pushed an update that should now include muxed devices in the list again.

plugins/mac-avcapture/av-capture.mm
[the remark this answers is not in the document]
PatTheMav: Clarified off thread.

PR #6632, opened 2022-06-20
mac-syphon: Fix usage of methods deprecated since macOS 11.0

PatTheMav:
### Description
`absolutePathForAppBundleWithIdentifier` and `openFile` have both been deprecated since macOS 11.0. 

This implements the correct modern APIs when OBS is compiled for macOS 11.0 or more recent and falls back to the legacy APIs when either compiled against or run on older macOS versions.

### Motivation and Context
Ensure compatibility with future versions of macOS.

### How Has This Been Tested?
* Still needs to be tested with a proper Syphon setup

### Types of changes

### Checklist:

### DIFF
--- a/plugins/mac-syphon/syphon.m
+++ b/plugins/mac-syphon/syphon.m
@@ -2,6 +2,7 @@
 #import <ScriptingBridge/ScriptingBridge.h>
 #import "syphon-framework/Syphon.h"
 #include <obs-module.h>
+#include <AvailabilityMacros.h>
 
 #define LOG(level, message, ...)                                    \
 	blog(level, "%s: " message, obs_source_get_name(s->source), \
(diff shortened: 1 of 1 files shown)


PatTheMav:
@gxalpha, @RytoEX please evaluate if this is still necessary - from what I could tell the affected code-parts concern injecting apps with SyphonInject, which _should_ not be possible since macOS 10.15 anyway.

Once we increase the lowest supported macOS version to 10.15, we might instead need to remove the entire functionality instead.

gxalpha:
I don't think this is needed since as you stated, SyphonInject doesn't work with anything newer than 10.13. I don't think there's any way to get it to work on 10.14+ since it relied on a loophole Apple patched.

RytoEX:
> I don't think this is needed since as you stated, SyphonInject doesn't work with anything newer than 10.13. I don't think there's any way to get it to work on 10.14+ since it relied on a loophole Apple patched.

Is any of this code related to SyphonInject, or just the Syphon Source? These are separate functionalities. The last function looks like it's just about displaying the Syphon license, so it shouldn't be strictly related to SyphonInject. I'm not familiar with this plugin, so I'm not sure about the first code block.

gxalpha:
Ah no you're right, I only looked at Pat's comment and glimpsed at the first part of the code - we have a button in the source properties that makes the license open, which isn't tied to Syphon inject.

PatTheMav:
Yeah the button to open the license is independent of any injection functionality. Alas the same applies to the function returning the path of an application.

We don't really "shield" the functionality away from the user on more recent macOS versions and we probably need to clean up the entire plugin once we added the macOS 10.15 change.

As-is, we don't, so this code will be compiled and throw warnings otherwise. As it doesn't impact functionality, I'd be fine with merging it and removing it again later (see similar comment on the AVCaptureDevice PR).

PR #6639, opened 2022-06-22
mac-virtualcam: Fix IOSurface memory leak

fabianishere:
This change fixes a memory leak in the mac-virtualcam plugin that causes OBS to not release the CVPixelBuffers (and underlying IOSurfaces) it emits to the virtual camera consumers.

### Description
I have found that the plugin is leaking Mach ports associated with IOSurfaces, preventing them from being re-used. The previous approach using `NSMachPort` does not seem to properly release the Mach port allocated via `CVPixelBufferGetIOSurface` and `IOSurfaceLookupFromMachPort`. Instead, we must explicitly deallocate the port using `mach_port_deallocate`.

### Motivation and Context
Pull request https://github.com/obsproject/obs-studio/pull/6573 (Avoid transcoding where possible) updated the mac-virtualcam plugin to share the virtual camera feed with other processes via IOSurfaces.

Although the changes work correctly, users have observed that OBS memory usage keeps increasing when the virtual camera is active until OBS runs out of memory or the consuming application is closed.
See the report by @SciTechNick for more information: https://github.com/obsproject/obs-studio/pull/6573#issuecomment-1161979765

### How Has This Been Tested?
I have tested the changes on a Macbook Pro (M1) running macOS Monterey with Google Chrome, Zoom, and Cameo. OBS shows no signs of memory leakage after multiple minutes.

### Types of changes
Bug fix

### Checklist:

### DIFF
--- a/plugins/mac-virtualcam/src/dal-plugin/OBSDALMachClient.mm
+++ b/plugins/mac-virtualcam/src/dal-plugin/OBSDALMachClient.mm
@@ -101,14 +101,30 @@
 		break;
 	case MachMsgIdFrame:
 		VLog(@"Received frame message");
-		if (components.count >= 4) {
+
+		if (components.count < 4)
+			return;
+
+		@autoreleasepool {
 			NSMachPort *framePort = (NSMachPort *)components[0];
+
+			if (!framePort)
+				return;
+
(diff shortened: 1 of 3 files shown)


gxalpha (approved):
Can confirm that this fixes the issue.

PR #6651, opened 2022-06-25
obs-outputs: Rework RTMP context init/deinit

tt2468:
### Description
This commit fixes what is arguably a long-winded series of previous commits that have possibly caused just as many problems as they have fixed.

This provides a slight enhancement by removing the `RTMP_Init` call in `rtmp_stream_create()`, since it effectively just initializes TLS just for `try_connect` to deinitialize it before it is even used.

This also fixes the current `SO_RCVTIMEO` timeout functionality by making sure that `RTMP_Reset` is called last.

### Motivation and Context
I'll spare the details, but basically, there's no reason that any of the RTMP object should ever be used across socket sessions. It's best just to zero out the RTMP context and initialize TLS each time we connect, than to do a bunch of things to only reset parts of it.

### How Has This Been Tested?
Ubuntu 20.04:
- RTMPS streaming - ok
- Standard RTMP streaming - ok
- Cutting network for RTMP stream and allowing reconnect - ok
- RTMP streaming with user/pass authentication - ok

### Types of changes

### Checklist:

### DIFF
--- a/plugins/obs-outputs/librtmp/rtmp.c
+++ b/plugins/obs-outputs/librtmp/rtmp.c
@@ -420,7 +420,7 @@ RTMP_TLS_Init(RTMP *r)
 
 void
 RTMP_TLS_Free(RTMP *r) {
-#ifdef USE_MBEDTLS
+#if defined(CRYPTO) && defined(USE_MBEDTLS)
 
     if (!r->RTMP_TLS_ctx)
         return;
@@ -451,9 +451,7 @@ RTMP_Alloc()
 void
 RTMP_Free(RTMP *r)
 {
-#if defined(CRYPTO) && defined(USE_MBEDTLS)
     RTMP_TLS_Free(r);
-#endif
     free(r);
 }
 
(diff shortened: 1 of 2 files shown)


notr1ch:
RTMP authentication should be tested as that was one of the reasons the code was so awful in the first place - some state has to be maintained between the initial connection and the authorized connection for whatever reason.

tt2468:
> RTMP authentication should be tested as that was one of the reasons the code was so awful in the first place - some state has to be maintained between the initial connection and the authorized connection for whatever reason.

I've now tested to the best of my ability using a Wowza trial and all works well.

tt2468:
This PR also happens to fix the permanent deadlock associated with the TCP connection state being invalidated by an upstream router.

jp9000:
Looks pretty good to me and seems to work fine. It's been tested with authentication so the concern about authentication shouldn't be an issue at this point. @notr1ch this okay with you?

notr1ch:
Yeah, hopefully if there are any issues they will appear during the beta / RC process.

PR #6652, opened 2022-06-25
UI: Improve exponential backoff reconnect on disconnect

tt2468:
### Description
- Changes the default base exponent value to 1.5 from 2.0
- Applies a random skew of +-0.05 to the exponent to lessen the "water hammer" effect caused by predictable backoff techniques
- Fixes the logging associated with exponential backoff to log the true reconnect delay value
- Changes the default reconnect delay to 2 seconds from 10
- Changes the default max retries to 25 from 20

### Motivation and Context
The default delay of 10 seconds is arguably quite excessive given that it already implements exponential backoff.

Depending on how a given stream is disconnected, it's likely that the RTMP server is already ready to accept the publisher again by the time the reconnect process even starts. If it's not ready, then that's why exponential backoff is there in the first place.

Total average reconnect time before abandon is now 2.75 hours, from an original 3.4 hours.

### How Has This Been Tested?
Ubuntu 20.04: Works

### Types of changes

### Checklist:

### DIFF
--- a/libobs/obs-output.c
+++ b/libobs/obs-output.c
@@ -18,6 +18,7 @@
 #include <inttypes.h>
 #include "util/platform.h"
 #include "util/util_uint64.h"
+#include "graphics/math-extra.h"
 #include "obs.h"
 #include "obs-internal.h"
 
@@ -25,6 +26,9 @@
 #include <caption/mpeg.h>
 
 #define get_weak(output) ((obs_weak_output_t *)output->context.control)
+
+#define RECONNECT_RETRY_MAX_MSEC (15 * 60 * 1000)
+#define RECONNECT_RETRY_BASE_EXP 1.5
 
 static inline bool active(const struct obs_output *output)
 {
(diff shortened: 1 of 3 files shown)


Warchamp7 (approved):

RytoEX:
> Changes the default retry count from 20 to 22 to make the effective total retry process similar to before.

Could you elaborate? What would the retry count being 20 or 22 look like with the other change here, in terms of total retry time and final retry interval?

tt2468:
> > Changes the default retry count from 20 to 22 to make the effective total retry process similar to before.
> 
> Could you elaborate? What would the retry count being 20 or 22 look like with the other change here, in terms of total retry time and final retry interval?

libobs/obs-output.c
```diff
@@ -145,6 +149,8 @@ obs_output_t *obs_output_create(const char *id, const char *name,
 
 	output->reconnect_retry_sec = 2;
 	output->reconnect_retry_max = 20;
+	output->reconnect_retry_exp =
+		RECONNECT_RETRY_BASE_EXP + (rand_float(0) / 20);
```
norihiro, line 153:
```suggestion
		RECONNECT_RETRY_BASE_EXP + (rand_float(0) * 0.05);
```
How about writing `* 0.05` so that easily know it's 0.05 second (50 ms) random time.

PR #6653, opened 2022-06-25
CI: Use static 7zip build instead of PPA

derrod:
### Description

Use official static 7zip linux binaries instead of removed PPA.

### Motivation and Context

The PPA has been deleted so the Steam workflow started failing.

### How Has This Been Tested?

Has not been tested, but since this is just reverting the switch to the PPA I don't expect any issues.

### Types of changes

### Checklist:

### DIFF
--- a/.github/workflows/steam.yml
+++ b/.github/workflows/steam.yml
@@ -24,6 +24,7 @@ env:
   STEAM_NIGHTLY_BRANCH: nightly
   STEAM_STABLE_BRANCH: staging
   STEAM_BETA_BRANCH: beta_staging
+  SEVENZIP_HASH: 5290409e7bbe2f133d0bd7e7482548678157ea2be276b0f9cb440600f4be9a2d
 
 jobs:
   upload:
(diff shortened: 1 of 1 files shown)


.github/workflows/steam.yml
```diff
@@ -37,12 +37,15 @@ jobs:
       with:
         path: source
 
-    - name: Setup 7-Zip (PPA)
       # The 7-Zip version available in the default ubuntu repos (p7zip) is wildly out-of-date and does not properly support DMG files.
+    - name: Setup 7zip
```
RytoEX, line 41: Nit: The software name is "7-Zip".

RytoEX (changes requested):
Aside from the nitpick I've highlighted, this looks fine to me as long as Rich doesn't have any concerns.

derrod:
Yeah perhaps we could mirror a known-good version on the OBS server @notr1ch?

notr1ch:
I'd feel safer if we had a hash check as well.

RytoEX:
> I'd feel safer if we had a hash check as well.

Do you want a hash check and mirror the archive, or just the hash check?

derrod:
I added a hash check for the SHA256 of the downloaded file.

RytoEX (approved):
Seems fine to me. @notr1ch ?

jp9000:
I'll merge it for now, if there's any issues we can always fix it

PR #6654, opened 2022-06-25
UI: Move Simple Mode encoder preset out of Advanced (and more)

gxalpha:
### Description

Moves the "Encoder Preset" option in Simple Mode out of the "Advanced" checkbox, and only shows ultrafast-fast. Any users who currently are on slower presets will stay on those, but won't be able to newly select them anymore.
Also moves the audio bitrate below the video bitrate so that it isn't so awkwardly between the encoder settings anymore.

Before:

https://user-images.githubusercontent.com/59806498/175784382-35516e2d-c31c-4a47-8e9e-65fb5f68c31b.mov

Now:

https://user-images.githubusercontent.com/59806498/175784334-24a39eb5-1352-48b9-b9f7-19a2d1eec32e.mov

### Motivation and Context

Simple mode should be simple. If we consider a feature to be simple, it should just be available, if we consider it to be advanced, it shouldn't. An "Advanced" checkbox in Simple Mode doesn't make a lot of sense, if a feature is considered to be simple it should just be available. However, to keep users from foot-shooting, anything slower than fast got removed from simple mode, and short descriptions were added to describe how the presets effect the stream (making them simpler to understand).

The custom encoder arguments (and probably also the custom muxer settings for recordings) will be another story. While those are arguably even worse offenders than encoder presets (they shouldn't be available in a simple mode, users at that level should just go into Advanced Mode), just removing them could/would break existing setups. We would probably have to transparently migrate them to Advanced Mode, which could be a larger undertaking.

### How Has This Been Tested?

macOS 13 Beta 2

Tested that only ultrafast-fast are shown, except when the user has a slower one already selected (see video above). Tested that the presets no longer depend on whether the "Advanced" is enable. Tested that setting gets applied to encoder correctly.

### Types of changes

### Checklist:


Warchamp7:
This was a change suggested/requested by myself, so it has my stamp on the UX side

Warchamp7 (approved):

notr1ch:
This looks good. My only concern is the wording - "average" feels like it would be tempt users to change it immediately, after all who wants only average quality? Perhaps veryfast could be labeled "good quality" or "standard quality" or something less likely to encourage tinkering.

gxalpha:
changed to "medium CPU usage, standard quality"

PR #6657, opened 2022-06-26
libobs, win-wasapi: Handle changes to the default audio monitoring device

### LINKED ISSUE #4463: When switching audio monitor output device - monitor audio stops
### Operating System Info

Windows 10

### Other OS

_No response_

### OBS Studio Version

26.1.1

_No response_

### OBS Studio Log URL

https://obsproject.com/logs/xCZi22mtkq1wYOpF

### OBS Studio Crash Log URL

_No response_

### Expected Behavior

Audio monitoring output continues on new device

### Current Behavior

Audio monitoring output stops

### Steps to Reproduce

1. open Settings - Audio
2. change Monitoring device

After a restart of OBS it will work again

tt2468:
### Description
Changes win-wasapi to reset audio monitoring when the default output device changes.

Fixes #4463
Closes #5398 

### Motivation and Context
When the default Windows output changes, OBS should switch to the new one if it is configured to use default.

### How Has This Been Tested?
- Enable audio monitoring with default device
- Change default output device: OBS responds [ok]
- Change default input device: OBS does not respond [ok]
- Set audio monitoring to non-default device
- Change default input/output devices: OBS does not respond [ok]

### Types of changes

### Checklist:

### DIFF
--- a/docs/sphinx/reference-core.rst
+++ b/docs/sphinx/reference-core.rst
@@ -540,6 +540,12 @@ Video, Audio, and Graphics
 
 ---------------------
 
+.. function:: void obs_reset_audio_monitoring(void)
+
+   Resets all audio monitoring devices.
+
+---------------------
+
 .. function:: void obs_enum_audio_monitoring_devices(obs_enum_audio_device_cb cb, void *data)
 
    Enumerates audio devices which can be used for audio monitoring.
--- a/libobs/obs.c
+++ b/libobs/obs.c
@@ -3018,6 +3018,21 @@ bool obs_obj_is_private(void *obj)
 	return context->private;
 }
 
+void obs_reset_audio_monitoring(void)
+{
+	if (!obs_audio_monitoring_available())
+		return;
+
+	pthread_mutex_lock(&obs->audio.monitoring_mutex);
+
+	for (size_t i = 0; i < obs->audio.monitors.num; i++) {
+		struct audio_monitor *monitor = obs->audio.monitors.array[i];
+		audio_monitor_reset(monitor);
+	}
+
+	pthread_mutex_unlock(&obs->audio.monitoring_mutex);
+}
+
 bool obs_set_audio_monitoring_device(const char *name, const char *id)
 {
 	if (!name || !id || !*name || !*id)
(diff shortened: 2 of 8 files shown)


jp9000:
Are you intending to share this code over multiple modules? An initial take I have of this is I'm not sure if it's really all that necessary to do so. Also, the wasapi code in libobs is not C++, so it will not work with STL callbacks. Plus I'm kind of wondering if the STL usage feels a bit excessive and possibly unnecessary, but maybe that's just my humble opinion on the matter.

Personally, I feel like this particular class (IMMNotificationClient I mean) is so trivial and never really used for any other purposes that I don't really mind if there's a little bit of code duplication just to separate the libobs code from the win-wasapi plugin code. That was kind of one of the issues I had with the implementation walker made as well. Sometimes there are times where trying to solve code duplication can *possibly* make things a little less pleasant than they would be if you just did the same thing twice. In those sort of situations it's not a big deal if you have two sets of code that basically do the same thing as long as it makes things cleaner and is highly unlikely to be modified/customized again. But again, that's just my personal philosophy with that sort of thing.

Sometimes certain approaches can be a bit subjective. For example I probably could have just written this plugin entirely in C, and sometimes I kind of wish I did, but I suppose it doesn't matter that much.

docs/sphinx/reference-core.rst
```diff
@@ -492,6 +492,12 @@ Video, Audio, and Graphics
 
 ---------------------
 
+.. function:: boid obs_reset_audio_monitoring(void)
```
RytoEX, line 495: `boid` -> `void`

RytoEX (changes requested):
Nit. Will try to test this at some later time.

RytoEX:
With this PR, I can confirm that when the OBS audio monitoring device is set to Default, it now follows changes to Windows' default output device.

I understand that there is ongoing discussion about the structure of the code, so I will defer to others on when this is good to merge.

Lain-B (approved):

Lain-B:
Rebased and fixed up issues. The only issue I have with this is that it makes it so that the audio monitoring on windows is sort of dependent upon the win-wasapi to update it when the device has changed, but I suppose it's sufficient for now and keeps all the code in one place for it.

tt2468:
I think it could be moved in the future easily to another place. To me, having the function to reset audio monitoring is helpful regardless of this specific usage of it here.

PR #2778, opened 2020-04-22
UI: Add new theme Yami

Warchamp7:
### Description
This is a new theme designed with the intent of replacing the current default dark theme.

Feedback and issues with the theme itself are welcome below. Discussion of this theme actually replacing Dark as the default should take place on the RFC

obs64_2020-04-22_11-13-26

obs64_2020-04-22_11-13-33

obs64_2020-04-22_11-19-52

obs64_2020-04-22_11-13-42

obs64_2020-04-22_11-19-14

obs64_2020-04-22_11-14-29

obs64_2020-04-22_11-18-13

obs64_2020-04-22_11-18-44

### Motivation and Context
Compared to the default Dark theme this new theme has a larger font size, more space around elements to let them breathe, and a better color heirarchy to improve both readability and usability.

The Dark theme, while very compact, is also incredibly cramped and overwhelming visually, especially for newer users.

### How Has This Been Tested?
Checked all windows and dialogs of OBS

### Types of changes
- New feature (non-breaking change which adds functionality)

### Checklist:
- [ ] My code has been run through [clang-format](https://github.com/obsproject/obs-studio/blob/master/.clang-format).
- [x] I have read the [**contributing** document](https://github.com/obsproject/obs-studio/blob/master/CONTRIBUTING.rst).
- [x] My code is not on the master branch.
- [x] The code has been tested.
- [ ] All commit messages are properly formatted and commits squashed where appropriate.

PR #6664, opened 2022-06-27
UI: Enforce Fusion Qt style on Linux

tytan652:
### Description

Found the issue on Linux, the Plasma default theme Breeze also provide a Qt style plugin.
This plugin add a frame to QDockWidget.
Those frames are not stylable with QSS because those are drawn with QPainter.
- https://github.com/KDE/breeze/blob/4222e8e83ae870ffb1ccc3c02775681322f62b88/kstyle/breezestyle.cpp#L1555
- https://github.com/KDE/breeze/blob/4222e8e83ae870ffb1ccc3c02775681322f62b88/kstyle/breezehelper.cpp#L457

Flatpak install is always affected because the KDE Runtime provide Breeze, and Linux systems with Breeze installed are also affected for non-Flatpak install

So to avoid this the base style is enforce to the Qt default style on Linux: Fusion.

Screenshot made with #2778 

Without the PR:
Thanks_Qt_and_Breeze

With the PR:
Thanks_Fusion

### Motivation and Context

I want Yami to be nice on Linux and Qt doesn't want me to be happy.

### How Has This Been Tested?

- Run without the commit: frames around docks
- Run with the commit: no frames around docks

### Types of changes

### Checklist:

### DIFF
--- a/UI/obs-app.cpp
+++ b/UI/obs-app.cpp
@@ -2078,8 +2078,14 @@ static int run_program(fstream &logFile, int argc, char *argv[])
 	}
 #endif
 
-#if !defined(_WIN32) && !defined(__APPLE__) && defined(USE_XDG) && \
-	defined(ENABLE_WAYLAND)
+#if !defined(_WIN32) && !defined(__APPLE__)
+	/* NOTE: The Breeze Qt style plugin adds frame arround QDockWidget with
+	 * QPainter which can not be modifed. To avoid this the base style is
+	 * enforce to the Qt default style on Linux: Fusion. */
+
+	setenv("QT_STYLE_OVERRIDE", "Fusion", false);
+
+#if defined(ENABLE_WAYLAND) && defined(USE_XDG)
 	/* NOTE: Qt doesn't use the Wayland platform on GNOME, so we have to
 	 * force it using the QT_QPA_PLATFORM env var. It's still possible to
 	 * use other QPA platforms using this env var, or the -platform command
@@ -2088,6 +2094,7 @@ static int run_program(fstream &logFile, int argc, char *argv[])
 	const char *session_type = getenv("XDG_SESSION_TYPE");
 	if (session_type && strcmp(session_type, "wayland") == 0)
 		setenv("QT_QPA_PLATFORM", "wayland", false);
+#endif
 #endif
 
 	OBSApp program(argc, argv, profilerNameStore.get());


SuslikV:
Maybe stupid question... Will work the QApplication setStyle QStyleFactory create("Fusion") sequence for OBS only, but not for every Qt application installed in the system? I don't like the global env vars.

tytan652:
> Maybe stupid question... Will work the QApplication setStyle QStyleFactory create("Fusion") sequence for OBS only, but not for every Qt application installed in the system? I don't like the global env vars.

This will try to apply Fusion over Breeze by "merging" them because the base style is already Breeze. To avoid Breeze to be used as base style this env var is unfortunately necessary.

WizardCM:
While I overall agree with this change due on limitations in the Qt API, the only downside of this change is that for users who use the System theme, it may no longer look like their system if a custom style is applied automatically.

tytan652:
The `System` theme on Linux:
- Will be Breeze or Fusion in Flatpak, so this change is minor on Flatpak
- Is expected to generate bizarre style and to not work really well with dark system theme (like OBS icons still black).

We could:
- Keep `System` and consider it as a Fusion style
- Remove the `System` theme on Linux build
- Rename it to Fusion on Linux

jp9000:
If this affects KDE only, is there a way to only turn this on when KDE is active? Or do we need to have this on gnome as well?

tytan652:
This issue is related to Breeze Qt style plugin which is DE-agnostic, it will affect:
  - every Flatpak installation because of the KDE Runtime.
  - every user that have Breeze installed and use it on their system as Qt default Style.

So the only way is to enforce Qt's default: Fusion. User can still override it if they want to on non-Flatpak install.

jp9000:
Ah okay, I think I misunderstood originally then. Linux is definitely not my expertise.

dimtpap:
This is how Yami looks for me without this PR
image
and then with this PR
image

GloriousEggroll:
This commit causes OBS to not follow Gnome's theming if breeze is installed.

Here is my setup:
> I'm using GNOME
> I have the breeze packages installed because Kdenlive needs them to display properly
> I launch OBS with this commit -- my theme is in dark 
Screenshot from 2022-07-25 22-13-59
mode but OBS's bar is bright white:

If I revert the commit, OBS behaves:
Screenshot from 2022-07-25 22-25-13

tytan652:
Breeze have nothing to do with that, you are actually defaulting to Qgnome (`QT_STYLE_OVERRIDE=gnome`) another Qt Style made by Fedora IIRC which seems to manage the decoration bar theme.

Qt doesn't provide a way to avoid breeze if this is installed and will be used. The only solution to always avoid Breeze is to enforce Fusion.
I don't even know if other Qt Style "break" OBS themes.

The user still can set `QT_STYLE_OVERRIDE` and OBS will not override it. Feel free to used that.

GloriousEggroll:
thanks for the tip -- i resolved this by removing:
qgnomeplatform-qt5
qgnomeplatform-qt6 
from my system after your tip

tytan652:
The Breeze Qt style is DE-agnostic, son it's not a question of Boeing on Plasma or GNOME.

I'm on GNOME and I have this issue.

PR #6207, opened 2022-03-25
linux-pipewire: Add PipeWire audio captures

dimtpap:
### Description

Adds three audio capture sources that utilize PipeWire to capture audio inputs, outputs, as well as audio of specific applications.

### Motivation and Context

PipeWire is increasingly getting adopted by users as well as major distros (like Fedora), and is on track to be the de facto standard for audio on Linux and others. The PipeWire project has created a PulseAudio-PipeWire bridge, so the current PulseAudio sources work as intended. However, pipewire-pulse has had its issues in the past, although they were minor. Furthermore, PipeWire allows us to capture specific application audio, something that currently needs external tools and tinkering to get working.

### How Has This Been Tested?

This has been tested on Arch Linux, on KDE Plasma running X.Org. I am not able to test Wayland captures and PW Audio captures working together since I'm on an Nvidia GPU, and the Nvidia driver is missing features for Wayland to work on it.

### Types of changes

- New feature (non-breaking change which adds functionality)

### Checklist:

- [x] My code has been run through [clang-format](https://github.com/obsproject/obs-studio/blob/master/.clang-format).
- [x] I have read the [**contributing** document](https://github.com/obsproject/obs-studio/blob/master/CONTRIBUTING.rst).
- [x] My code is not on the master branch.
- [x] The code has been tested.
- [x] All commit messages are properly formatted and commits squashed where appropriate.
- [x] I have included updates to all appropriate documentation.

PR #5377, opened 2021-10-05
PipeWire Virtual Camera

columbarius:
### Description

This MR creates a new plugin exposing the obs video feed as a virtual camera via PipeWire. Additionally it should create a framework to share code between different plugins connecting obs to PipeWire.

### Motivation and Context

The current virtual camera plugin for linux via v4l2 requires root access to load a kernel driver. A plugin via PipeWire doesn't need that kind of privileges and additionally works nice in sandboxed environments like flatpak.

### How Has This Been Tested?

* Remove the v4l2loobbak kernel module.
* Start obs
* Start the virtual camera
* Record the stream with cheese or get the pipewire-id and use the video-play example from PipeWire

### Types of changes

- New feature (non-breaking change which adds functionality)

### Checklist:

- [x] My code has been run through [clang-format](https://github.com/obsproject/obs-studio/blob/master/.clang-format).
- [x] I have read the [**contributing** document](https://github.com/obsproject/obs-studio/blob/master/CONTRIBUTING.rst).
- [x] My code is not on the master branch.
- [x] The code has been tested.
- [x] All commit messages are properly formatted and commits squashed where appropriate.
- [x] I have included updates to all appropriate documentation.

Requires and contains: #5383 

Maintainer edit: Fixes #4499

PR #6669, opened 2022-06-28
Draft: Pipewire unified wrappers

columbarius:
### Description
Unify pipewire.c api in the linux-pipewire plugin

### Motivation and Context

We want to create a basic pipewire wrapper to share code between multiple obs plugins. This MR should create the basis for discussions to create sth. useful for various video and audio plugins.

### Types of changes

### Checklist:

### DIFF
--- a/plugins/linux-pipewire/pipewire.c
+++ b/plugins/linux-pipewire/pipewire.c
@@ -65,7 +65,6 @@ struct _obs_pipewire {
 	struct obs_pw_version server_version;
 
 	GPtrArray *streams;
-	DARRAY(struct format_info) format_info;
 };
 
 struct _obs_pipewire_stream {
@@ -97,6 +96,8 @@ struct _obs_pipewire_stream {
 
 	struct obs_video_info video_info;
 	bool negotiated;
+
+	DARRAY(struct format_info) format_info;
 };
 
 /* auxiliary methods */
@@ -293,8 +294,8 @@ static bool build_format_params(obs_pipewire_stream *obs_pw_stream,
 	uint32_t params_count = 0;
 
 	const struct spa_pod **params;
-	params =
-		bzalloc(2 * obs_pw->format_info.num * sizeof(struct spa_pod *));
+	params = bzalloc(2 * obs_pw_stream->format_info.num *
+			 sizeof(struct spa_pod *));
 
 	if (!params) {
 		blog(LOG_ERROR,
@@ -305,22 +306,23 @@ static bool build_format_params(obs_pipewire_stream *obs_pw_stream,
 	if (!check_pw_version(&obs_pw->server_version, 0, 3, 33))
 		goto build_shm;
 
-	for (size_t i = 0; i < obs_pw->format_info.num; i++) {
-		if (obs_pw->format_info.array[i].modifiers.num == 0) {
+	for (size_t i = 0; i < obs_pw_stream->format_info.num; i++) {
+		if (obs_pw_stream->format_info.array[i].modifiers.num == 0) {
 			continue;
 		}
 		params[params_count++] = build_format(
 			pod_builder, &obs_pw_stream->video_info,
-			obs_pw->format_info.array[i].spa_format,
-			obs_pw->format_info.array[i].modifiers.array,
-			obs_pw->format_info.array[i].modifiers.num);
+			obs_pw_stream->format_info.array[i].spa_format,
+			obs_pw_stream->format_info.array[i].modifiers.array,
+			obs_pw_stream->format_info.array[i].modifiers.num);
 	}
 
 build_shm:
-	for (size_t i = 0; i < obs_pw->format_info.num; i++) {
+	for (size_t i = 0; i < obs_pw_stream->format_info.num; i++) {
 		params[params_count++] = build_format(
 			pod_builder, &obs_pw_stream->video_info,
-			obs_pw->format_info.array[i].spa_format, NULL, 0);
+			obs_pw_stream->format_info.array[i].spa_format, NULL,
+			0);
 	}
 	*param_list = params;
 	*n_params = params_count;
@@ -338,9 +340,9 @@ static bool drm_format_available(uint32_t drm_format, uint32_t *drm_formats,
 	return false;
 }
 
-static void init_format_info(obs_pipewire *obs_pw)
+static void init_format_info(obs_pipewire_stream *obs_pw_stream)
 {
-	da_init(obs_pw->format_info);
+	da_init(obs_pw_stream->format_info);
 
 	obs_enter_graphics();
 
@@ -358,7 +360,7 @@ static void init_format_info(obs_pipewire *obs_pw)
 					  drm_formats, n_drm_formats))
 			continue;
 
-		info = da_push_back_new(obs_pw->format_info);
+		info = da_push_back_new(obs_pw_stream->format_info);
 		da_init(info->modifiers);
 		info->spa_format = supported_formats[i].spa_format;
 		info->drm_format = supported_formats[i].drm_format;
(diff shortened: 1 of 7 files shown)


columbarius:
Relevant MRs:
#5377 
#5728
#6207

columbarius:
CC: @GeorgesStavracas @dimtpap

plugins/linux-pipewire/pipewire.h
```diff
@@ -22,9 +22,16 @@
...
 
+#include <pipewire/keys.h>
+#include <pipewire/properties.h>
+
 typedef struct _obs_pipewire_data obs_pipewire_data;
 
-void *obs_pipewire_create(int pipewire_fd, int pipewire_node);
+extern const struct pw_stream_events stream_events;
+
+void *obs_pipewire_create(int pipewire_fd, int pipewire_node, const char *name,
```
kkartaltepe, line 32: Why is this void*, the only assignment from this function i see is directly into the correct type.

columbarius:
Moved comment from https://github.com/obsproject/obs-studio/pull/5728#issuecomment-1168993335
> > We separate camera plugin and registry.
> > 
> > * camera-plugin: `init_registry` -> `query_cameras` -> `user selects camera` -> `query_registry for fd and id` -> `create stream like screencast` -> ... -> `unref_registry`
> > * camera-registry: global refcounted struct, which deals with the camera/device-portal. Is created on first `init_registry` call and destroyed if the refcount is 0. It queries the portal for a filedescriptor and opens the connection and tracks all announced cameras.
> 
> In principle, I think separating different moving parts into their own unit is good. My idea with #6645 is that all the code under `pipewire.c` is generic and can be used by all source types - even audio sources.
> 
> I don't know if it would be feasible or even possible, but I think you can already make this `camera-registry` thing completely generic, so it can be reused by the audio source at #6207. It'll probably need to be implemented in `pipewire.c` directly, perhaps as follows:
> 
>     * Split the code that actually initializes PipeWire from `obs_pipewire_create` into a new `obs_pipewire_start`
> 
>     * Add a new vfunc struct `obs_pipewire_device_callbacks` with the regular `device_added`, `device_removed`, `device_changed`, etc callback (please feel free to suggest better naming here)
> 
>     * Add `obs_pipewire_set_device_callbacks()`
> 
> 
> That way the camera portal code will be able to react to PipeWire adding and removing cameras, without having to deal with PipeWire itself.
> 
> What do you think?

I like the idea to build a registry wrapper with these callbacks. I'm not sure a `obs_pipewire_set_device_callbacks()` is feasable but an `obs_pipewire_registry_create()` with these callbacks as an argument (see stream_events here) will be.

> > An alternative would be to not lazyload the registry, but start it when obs is launched. This way we can only add the camera plugin, if obs is allowed to access the portal.
> > If we lazy load I would suggest to work asynchronous with callbacks from the registry to the plugin after it's initialized, but that's just some idea.
> 
> I think the best approach here is to do what I've done with the screencasting code, and completely separate camera portal code from PipeWire itself. What the plugin needs to do is to expose the pw-camera sources all the time, and only query the Device Access portal when the user actually adds a pw-camera source. If camera access is denied, we show an error message in the properties dialog.

I think we might talk about different things here: The screencast portal creates for each stream a new pipewire_fd together with the pipewire_id which the client can use to access only this single stream. Iirc the camera portal only grants access to a single pipewire_fd, which can be used to access pipewire on the host with permissions on the connection only to show the client cameras, etc. In this case the client is responsible to set up a registry to get the pipewire_id of each camera. Can't we just use the provided pipewire_fd together with the pipewire_id, obtained from the registry, and just create a new stream (including it's own pw_core, pw_thread_loop, etc.) with those?    

> > Anyways: should I just continue to rebase the current state, or should I continue with my new proposal?
> 
> Considering what I've suggested above, feel free to continue with your new proposal

This way we could create one wrapper each for pipewire_stream, pipewire_registry (with the suggested callbacks) and pipewire_node (@dimtpap could you extract such a wrapper from your audio work?)

plugins/linux-pipewire/pipewire.h
columbarius: The reason for this decision,  i can think of, is to hide all (for the plugin) "unnecessary" stuff behind a opaque handle.

kkartaltepe: Can we use an opaque type instead of void*, unless there is a real reason we need to use void* (but since this always returns the same type it doesnt seem so?).

columbarius: Why not? But I'd like implement this after we finished designing the whole wrapper api.

columbarius:
Btw. creating `pw_core` and `pw_thread_loop` and such things should be done by static functions inside pipewire.c and shared between the wrapper APIs.

columbarius:
Added registry stuff, nothing tested and just a sketch.

columbarius:
Does anyone know how to do fa935cc73f1205daf6bd907607b0d35e725a15e7 in nice?

dimtpap:
> Does anyone know how to do [fa935cc](https://github.com/obsproject/obs-studio/commit/fa935cc73f1205daf6bd907607b0d35e725a15e7) in nice?

Why not make it a regular struct and pass it to `pw_core_add_listener`

plugins/linux-pipewire/pipewire.h
```diff
@@ -27,8 +27,11 @@
 
 typedef struct _obs_pipewire_data obs_pipewire_data;
 
+extern const struct pw_stream_events stream_events;
+
```
GeorgesStavracas, line 33: Please don't. **If** allowing plugins to override stream events is a good idea - something I'm not convinced of yet - then it's better to write it as follows:

```c
struct pw_stream_events obs_pipewire_get_default_stream_events (void);
```

then in the C file it returns a copy of the structure:

```c
struct pw_stream_events obs_pipewire_get_default_stream_events (void)
{
	return stream_events;
}
```

Doing this will at least prevent plugins from accidentally screwing up other plugins.

plugins/linux-pipewire/pipewire.c
```diff
@@ -46,6 +46,26 @@ struct obs_pw_version {
 	int micro;
 };
 
+typedef struct _obs_pipewire_data obs_pipewire_data;
+
+// This struct contains the shared fields between all obs_pipewire wrappers
+// The C Standard guarantees that the memory layout off these parts of the
+// struct are identical and as such we can cast any extended struct to this
+// and use the fields without issues. Any change here has to be added to the
+// extended structs and the order of fields has to be preserved.
```
GeorgesStavracas, line 55: I don't think this is a good approach for sharing struct fields. Traditionally, if you want `obs_pipewire_stream_data` to extend `obs_pipewire_data`, you'd do something like:

```c
struct _obs_pipewire_data {
	... fields ...
};

struct _obs_pipewire_stream_data {
	obs_pipewire_data base;

	... other fields ...
};
```

Then, in the places where you want to treat `obs_pipewire_stream_data`  as a `obs_pipewire_data`, you'd do:

```c
obs_pipewire_data *data = &stream_data->base;
```

GeorgesStavracas (changes requested):
I think this PR needs some work, but ultimately what it tries to achieve is desired for the maintainability of the plugin. Technical details aside, I think the internal API here is a bit odd in some places. What I had in mind with device registry callbacks and whatnot was something like this:

```c
struct obs_pipewire_device_callbacks {
	void (*device_added)(void *user_data, obs_pipewire_device *device);
	void (*device_removed)(void *user_data, obs_pipewire_device *device);
};

void obs_pipewire_add_device_callbacks(
	obs_pipewire_data *obs_pw, 
	const struct obs_pipewire_device_callbacks *device_callbacks);

void obs_pipewire_remove_device_callbacks(
	obs_pipewire_data *obs_pw, 
	const struct obs_pipewire_device_callbacks *device_callbacks);
```

To properly do this, we would need to split the initialization of `obs_pipewire` in 2 steps:

```c
obs_pipewire_data * obs_pipewire_create(int pipewire_fd);

obs_pipewire_stream *
obs_pipewire_connect_stream(
	obs_pipewire_data *obs_pw,
	uint32_t stream_node,
	const char *stream_name,
	struct pw_properties *stream_properties);
```

The screencast portal code would look more or less like this:

```c
	capture->obs_pw = obs_pipewire_create(pipewire_fd);
	capture->pipewire_stream = obs_pipewire_connect_stream(capture->obs_pw,
		capture->pipewire_node, "OBS Studio",
		pw_properties_new(PW_KEY_MEDIA_TYPE, "Video",
				  PW_KEY_MEDIA_CATEGORY, "Capture",
				  PW_KEY_MEDIA_ROLE, "Screen", NULL)
	obs_pipewire_stream_set_cursor_visible(capture->pipewire_stream,
					capture->cursor_visible);
```


The camera portal code would look more or less like this:

```c

static const struct obs_pipewire_device_callbacks camera_device_callbacks = {
	camera_portal_device_added,
	camera_portal_device_remove,
};

...

	camera->obs_pw = obs_pipewire_create(pipewire_fd);
	obs_pipewire_add_device_callbacks(camera->obs_pw, &camera_device_callbacks);

... (every time someone selects a camera) ...

	capture->pipewire_stream = obs_pipewire_connect_stream(camera->obs_pw,
		device->pipewire_node, "OBS Studio",
		pw_properties_new(PW_KEY_MEDIA_TYPE, ...,
				  PW_KEY_MEDIA_CATEGORY, ...,
				  PW_KEY_MEDIA_ROLE, "Screen", NULL));
```

columbarius:
Added the case of an output stream to pipewire.c in planing of the virtualcamera pluging.

columbarius:
@GeorgesStavracas I'm pretty happy with your ideas! From my side it's ready for review.

columbarius:
@GeorgesStavracas I'm not sure about the api design. On one hand I'd like to have a single function for `obs_pipewire_stream_pause` on the other hand audio won't need a `_get_width` function. 

Would you prefere one api for video_in, video_out, audio_in or one unified one, created with a 
```
struct {
    type,
    union {
        video_in,
        ...,
    }
}
```
Thing and internally we do a lot of switch case?

columbarius:
I went this weekend on a totally different direction using a struct of function pointers holding the implementation of certain features for each backend which wants to implement the `obs_pipewire_stream` thing/interface. The results are quite nice.
[Here](https://github.com/columbarius/obs-studio/tree/pipewire-unified-dev) is a branch containing rebased versions of #5377 and #5728. 

@GeorgesStavracas Please take a glance at it, if we want to go with this kind of api (`pipewire-internal.h`)

@dimtpap Please look if extending the `impl` struct with audio functions would help you.

wtay:
I rebased the branch and added some small fixes: https://github.com/wtay/obs-studio/tree/pipewire-unified-dev

GeorgesStavracas:
This pull request has become too chaotic and unmanageable. I'll start breaking it in smaller chunks, and will handle it.

I think something went wrong with my rebase, half of the commits don't even build. I'll fix that too.

columbarius:
@GeorgesStavracas please ping me when I should start rebasing this including fixes from @wtay  onto #8153 
Nit: you linked #5614 instead of this MR there.

Same is valid if a review from me is helpful.

PatTheMav:
@GeorgesStavracas What's the intended plan for this PR? Is there a roadmap of changes that need to be merged first and does this PR still make sense in its current form (I guess the change itself is desired, just not the way it's implemented in the PR)?

PatTheMav:
There seems to be a set of PipeWire-related PRs that are stuck in Draft-limbo for one reason or another (and I don't have enough insight into PipeWire to understand for what reason) and this PR in particular was considered unfeasible for merging.

Can someone explain to me what the desired outcome is and what work is required to achieve that outcome? It would be preferable if the explanation is high-level and conceptual and doesn't get too bogged down in the details of PipeWire specifically (I'll leave that to domain experts to figure out).

At best I could figure out that https://github.com/obsproject/obs-studio/pull/5377 is somehow blocked by this?

PR #6677, opened 2022-07-01
aja: Fix memory overrun on aja-source

norihiro:
### Description

This commit moves the check of the size before the DMA transfer to avoid calling DMA transfer with too large size.

### Motivation and Context

When too many audio frames are captured at the same time the ring buffer on the card is wrapped, DMA transfer overrun to write audio data, which resulted in corruption of memory contents, sometimes OBS crashes.

### How Has This Been Tested?

OS: Fedora 34
OBS version: release/27.2 (d9865bdffb) + cherry-pick 1ba03d3f0a2613cf30de317a0381f683a5e946c7 to find memory corruption easily.
Hardware: AJA KONA 4

Configuration:
- Add multiple sources of `AJA I/O Device Capture` with these settings
  - Video format: 1080p59.94a
  - Pixel format: YUV-8
  - SDI transport: 3G Level-A (3Ga)
- Connect a camera to the SDI input on AJA KONA 4

To test, do the steps below several times.
1. Start OBS Studio with option `--verbose`.
2. Turn the power of the camera on and off around 10 times. During this, ensure a log message below appears.
```
AJASource::CaptureThread: Audio overrun (2)! Buffer Size: 410624, Bytes Captured: 1579008
```
3. Exit OBS and check no memory corruption reported by 1ba03d3f0 and ensure OBS did not crash.

Note that I didn't test audio is arriving to OBS since my camera does not output audio at all.

### Types of changes

### Checklist:

### DIFF
--- a/plugins/aja/aja-source.cpp
+++ b/plugins/aja/aja-source.cpp
@@ -40,8 +40,8 @@ AJASource::~AJASource()
 	mTestPattern.clear();
 	mVideoBuffer.Deallocate();
 	mAudioBuffer.Deallocate();
-	mVideoBuffer = NULL;
-	mAudioBuffer = NULL;
+	mVideoBuffer = 0;
+	mAudioBuffer = 0;
 }
 
 void AJASource::SetCard(CNTV2Card *card)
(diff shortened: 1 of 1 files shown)


plugins/aja/aja-source.cpp
```diff
@@ -293,6 +293,20 @@ void AJASource::CaptureThread(AJAThread *thread, void *data)
 				audioOverrun = true;
 			}
 
+			if (offsets.currentAddress - offsets.readOffset >
```
paulh-aja, line 296: Just to make it easier to understand the flow, perhaps move this check just in front of the `DMAReadAudio` call that it is checking against (Line 315 below in your code).

norihiro: Ok, I will update the code later.

norihiro:
The second commit (081c62975) changes structure of the capture thread to improve readability of the code. This should not change the behavior.
When merging, it might be better to squash (or let me know if I need to squash).

plugins/aja/aja-source.cpp
```diff
@@ -184,6 +184,63 @@ static inline uint64_t get_sample_time(size_t frames, uint_fast32_t rate)
 	return os_gettime_ns() - samples_to_ns(frames, rate);
 }
 
+static bool aja_source_check_transfer_audio_dma(
```
paulh-aja, line 187: I'm using the `aja_source_` function prefix for functions which are callbacks to the OBS plugin API. See here:
https://github.com/obsproject/obs-studio/blob/0ed34427e111e9d899b3a6ad708bc98f3139b0f3/plugins/aja/aja-source.cpp#L1164
Could you just rename the new functions without that prefix?

PR #6685, opened 2022-07-02
deps/obs-scripting: Log script load/unload

gxalpha:
### Description

Logs when a script gets loaded and unloaded

### Motivation and Context

Requested for support, since scripts can mess up things real bad and currently there's no way to know if a script is active.

### How Has This Been Tested?

macOS 13 Beta 2

When starting OBS, both python and lua scripts get logged like this:
image
When refreshing, first "loaded" and then "unloaded" shows up.
When adding or removing, it shows as loaded or unloaded.

### Types of changes

### Checklist:


flaeri:
Tested on Win 11 21H2, both with Py and LUA, scripts that we ship as well as oen 3rd party.
* Switching scene collections (with/without scripts)
* Adding and removing
* Reloading scripts
* Launch/Exit

Correctly outputs information to the log. No memory leaks or crashes during 10 launches/trials.

Godsend for support. Thank you very much!

PR #6686, opened 2022-07-03
UI: Yami font adjustments

Warchamp7:
### Description
Removes font-family definition for now, and changes
other font sizes to pt units instead of px

Also adds a margin to the SceneTree/SourceTree line
edits to fix qss padding weirdness

### Motivation and Context
Until we can ship fonts as part the resource file, we should stick to using the existing system fonts

### How Has This Been Tested?
Compared font sizes before and after this change.

### Types of changes

### Checklist:

### DIFF
--- a/UI/data/themes/Yami.qss
+++ b/UI/data/themes/Yami.qss
@@ -77,7 +77,6 @@ QWidget {
     selection-background-color: rgb(40,76,184);
     selection-color: palette(text);
     font-size: 10pt;
-    font-family: 'Open Sans', Helvetica, Arial, sans-serif;
 }
 
 QWidget:disabled {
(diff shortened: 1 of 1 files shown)


GeorgesStavracas:
Tested with the Flatpak bundle from CI, and it seems to work well enough here with Cantarell

SuslikV:
I added comment to https://github.com/obsproject/obs-studio/pull/2778
it's about new theme Yami implementation. The system theme (and other themes) missing some icons - just restart application with the System selected theme.

SuslikV:
Why pt?

pt is good when you have 300dpi/ppi or more.
Is it common for modern PC to have, let's say, the 180ppi display device?

The most software editors I know designed to print the text, so there are "pt"s. The printing resolutions are usually huge. I don't think that the themes designed to be printed.

Not a big deal, of course. Just wondering what was behind the decision to use "pt"s instead of "px"s?

PR #6687, opened 2022-07-03
obs-ffmpeg: Add support for audio track selection

mvji:
This commit also modifies deps/media-playback

### Description

Add support for audio track selection in media source.
The default behaviour is unchanged (auto select).
If we uncheck the autoselect checkbox, user can select another audio track

Default behaviour : Auto select audio track UI :
AutoselectOn

Auto select Audio track unchecked UI :
AutoselectOff

### Motivation and Context

Make Media source features close to VLC source.
Implement Ideas : https://ideas.obsproject.com/posts/737/select-audio-track-on-media-source

### How Has This Been Tested?

Mac os 11.6.2
Mac Intel / Xcode
Tested with basic sample with 1 audio track.
Tested with a 6 mono tracks test file, to check the correct read of audio track
(file can be found here : https://www.swisstransfer.com/d/39ef1a76-df1a-40f7-9dfe-ad6029b673bf)
Tested with invalid audio track number, to check no audio is output in that case.

### Types of changes

### Checklist:


Fenrirthviti:
I'm not thrilled with the UI/UX of this, where a checkbox is controlling the visibility of an element. Would this make more sense to add to the ffmpeg opts field that already exists? `audiotrack=4` or similar to avoid cluttering the UI with a feature most won't need to bother with.

mvji:
> I'm not thrilled with the UI/UX of this, where a checkbox is controlling the visibility of an element. Would this make more sense to add to the ffmpeg opts field that already exists? `audiotrack=4` or similar to avoid cluttering the UI with a feature most won't need to bother with.

Selecting audio track using ffmpeg options, seems not very easy to use. (using ffmpeg opts, is for me for very advanced options).
I try to be similar in term of UI, than the audio track selection of VLC source.
There is already advanced options in Media source, available (mostly put at the end of the properties panel).

Joefis-x20s:
Unless I'm doing it incorrectly but I can't use the FFmpeg options to do anything except state stuff that I could already put in the URL like "mode=listener" or "streamid=142390" for streaming protocals.

GeorgesStavracas:
This PR was discussed recently and ideally it should expose the tracks available in the files, similar to how VLC (the application) operates. Would you please remove the checkbox, and replace the spinner by a dropdown populated with the media file tracks?

GeorgesStavracas:
(By default, the dropdown should pre-select the best available audio track provided by FFmpeg)

mvji:
Last update change the audio track selection UI and restrict it for local file.
Also add "Disable Audio" feature

6 tracks file : 
AudioTrack

Local file without Audio :
NoAudio
 
When a local file is choose (or when propertie panel is open), 
The audio file information is read, to detect if there is audio stream or not.
Adapt a combobox to let user choose only track who exists in the file.
By default, use "Auto select" option (for consistency between existing behaviour, and more convenient to use when select another media in the local file path selector
Add a disable audio mode, to avoid decoding audio when you don't need it at all.
If media file didn't have audio, audio track selection is hidden.

GeorgesStavracas:
@mvji haven't reviewed this again, but I just wanted to thank you for being so receptive and quick in responding to the requests we shared this week. It's always a pleasure to work with contributors like this.

Warchamp7:
@mvji Echoing Georges' comment about the swift response. Thanks a lot for the quick changes!

I want to avoid having a setting that appears/disappears based on whether an audio track exists (and we are pushing to be better about this throughout the rest of the program as well). Instead of hiding the property when there is no audio, can you just disable the dropdown with a single entry that says "No audio tracks available"?

Can we also remove the autoselect option and simply update the dropdown with the choice that autoselect would do? Or would this create problems when swapping between files that have different audio track selections?

deps/media-playback/media-playback/decode.c
```diff
@@ -151,7 +151,29 @@ static uint16_t get_max_luminance(const AVStream *stream)
 	return max_luminance;
 }
 
-bool mp_decode_init(mp_media_t *m, enum AVMediaType type, bool hw)
+static int get_stream(mp_media_t *m, struct mp_decode *d, enum AVMediaType type,
+		      int track)
+{
+	if (d->audio) {
```
tt2468, line 157: Nitpick: Make this function media-agnostic to allow it to be used for selecting the current video track too

plugins/obs-ffmpeg/obs-ffmpeg-source.c
```diff
@@ -448,6 +538,8 @@ static void ffmpeg_source_update(void *data, obs_data_t *settings)
 	s->close_when_inactive =
 		obs_data_get_bool(settings, "close_when_inactive");
 
+	s->audio_track = obs_data_get_int(settings, "audio_track");
```
tt2468, line 541: Related to the above nit, `video_track` could be added here too, allowing very advanced users to select the video track using obs-websocket.

mvji:
@tt2468 : Not entirely sure about the best way to enable videotrack selection using websockets.
I rebased this PR, and made your requested changes : 
- get_stream now use the track type only to select the track (so can be use for video track for example)
- Add default video_track value set to zero (for auto select)

Audio track selection can be test for example with this file : 
(Prores file, with 6 mono audio tracks in this order : L, R, C, Lfe, Ls, Rs)
https://www.swisstransfer.com/d/3c4a3d3d-2a9b-4602-bdf8-c81d523b0674

tt2468:
> Not entirely sure about the best way to enable videotrack selection using websockets.

Simply exposing the track selection values in the source settings is enough for it to be used in obs-websocket via `SetInputSettings`.

mvji:
> > Not entirely sure about the best way to enable videotrack selection using websockets.
> 
> Simply exposing the track selection values in the source settings is enough for it to be used in obs-websocket via `SetInputSettings`.

Last update, should have the feature you want for video track
Also fix build error on linux and windows who was due to warning on sprintf buffer size and int64 to int conversion.

Joefis-x20s:
Is this still being thought about to be merged?

GeorgesStavracas (approved):
I do not have authority over the code, but from the screenshots, this looks fine for a UI / interaction perspective.

gxalpha (changes requested):
You only modify things with audio track, why do you store and access `video_track` variables? Please only change the things that the user also accesses.

mvji:
> You only modify things with audio track, why do you store and access `video_track` variables? Please only change the things that the user also accesses While you're at it, please rebase this PR on the current master branch, there are small conflicts.

The videotrack selection (without UI) was requested before (see previous comments of @tt2468 )

mvji:
> If i'm not mistaken the feature is limited to multi-track files ? I'd enable it for streams too.

I only tested using local file, with multiple audio tracks.
Don't have test case for testing with streams.

pkviet:
> > If i'm not mistaken the feature is limited to multi-track files ? I'd enable it for streams too.
> 
> I only tested using local file, with multiple audio tracks. Don't have test case for testing with streams.

it's fairly common in broadcast; ex: for multiple languages, for multiple surround options (stereo vs 5.1), director track with comments ... It can be tested with mpegts + udp/srt/rist for instance; such a a stream can be created within obs through the custom FFmpeg recording output. Or one can use FFmpeg.
That'd simplify a little bit the code since one doesn't have to check against 'file' or 'stream'.
But it's not a blocker so  ignore for now.

gxalpha:
> The videotrack selection (without UI) was requested before (see previous comments of @tt2468 )

I see. I heavily disagree with the idea of this only being settable via plugins/scripts/websockets, imo *if* we make this customizable it should be a property as well. Ultimately that's not my decision though.

PatTheMav:
I agree with @gxalpha that we should not try to sneak in a hidden feature only exposed via one specific way to interact with the program.

My suggestion would be to limit the scope of _this_ PR to the audio track selection and add video track selection as a separate PR.

@mvji can you give me a high-level rundown of the features this PR adds in its current form? Given that it had UI/UX approval already, I'd like to get this merged soon-ish (no promises on which version this will end up though).

I'm fine if this _just_ allows selecting the audio stream of a local file with e.g. different audio tracks for different languages. Anything else can come later.

Joefis-x20s:
bump

PR #6690, opened 2022-07-06
Use EGL_IMG_context_priority on X11 and Wayland 

HannesMann:
### Description
This change makes OBS use `EGL_IMG_context_priority` on X11 and Wayland to request a high-priority context. 

### Motivation and Context
Using a high-priority context ensures OBS is prioritized over the game being captured when GPU load is high. This extension has been supported on Intel since Mesa 17.3.0 and should be supported on AMD in the next release (since [this commit](https://gitlab.freedesktop.org/mesa/mesa/-/commit/e20718e8fa69bace9d5acf0ea484df6ca9421b98)). It should also be supported on NVIDIA, though I don't have any hardware to confirm it. 

### How Has This Been Tested?
On a Radeon RX 6700 XT running Quake II RTX:
* Before - OBS shows ~29 FPS
* After - OBS shows ~47 FPS 

Both configurations were using VAAPI H.264 at 1920x1080 and showed an "Encoding overloaded" warning, so the change doesn't make encoding perfectly smooth, but it still improves performance. 

Requesting a high-priority context on Mesa requires root or `CAP_SYS_NICE` which can be set with `sudo setcap CAP_SYS_NICE=+ep path-to-obs`. This should be done at install time by the package manager. OBS will drop the capability after the EGL context has been created (this is required for the PipeWire plugin to work). 

### Types of changes

### Checklist:


kkartaltepe:
Maybe other people have some concerns but mine are:

1) We dont want to have any additional capabilities required which cause people to recommend or otherwise configure obs to be run with elevated privileges.
2) obs is intended to be secondary to whatever users are already doing. On systems with constrained resources we prefer for the user's games or applications to receive preferential scheduling so we dont impact game/application performance. With this instead obs will be competing for resources.

HannesMann:
Forgot to mention that this requires libcap, so if you're on Debian or Ubuntu `libcap-dev` is needed to build OBS. Should these dependencies be added somewhere so GitHub Actions succeeds?

kkartaltepe:
> It also seems like the D3D11 renderer already does this? https://github.com/obsproject/obs-studio/blob/master/libobs-d3d11/d3d11-subsystem.cpp#L550

You can refer to the commit for that, but this is a workaround for a quirk of the windows scheduler and not setting OBS to a higher priority than games.

> Well, that's why CAP_SYS_NICE is dropped as soon as possible. You definitely don't need to run it as root for the feature to work, and if CAP_SYS_NICE isn't set the request for high priority will fail silently. I was inspired by a similar feature in KWin.

This commit makes people suggest running obs with higher privileges, which they will probably do with `sudo obs` instead of fiddling with capabilities like you describe in order to get the benefit this supposedly gives. Then since they launched obs with root and you only drop one cap obs is running with a bunch of dangerous capabilities because the user thought they might be benefited by this minor change. This is the scenario we want to avoid and that this kind of a change promotes.

Fenrirthviti:
On Windows, setting OBS' priority higher definitely helped both OBS and the game when the GPU utilization was at or nearing 100%, as on Windows the GPU scheduling was better handled with the higher priority. I'd like a lot of testing and feedback on this, but I certainly wouldn't be surprised if the same results are similar with OpenGL.

How was the in-game FPS affected in the before/after tests?

HannesMann:
Did some more testing:
* Quake II RTX on X11 - From 6 to 7 FPS in OBS
* Quake II RTX on Wayland - From 26 to 35 FPS in OBS
* Unigine Superposition on X11 - From 55 to 57 FPS in OBS
* Unigine Heaven on X11 - From 60 to 60 FPS (no change)

So far, Quake II RTX is the only game I can find that really makes OBS suffer. X11 is terrible no matter what. I had a lot more problems on my old RX 590. 

Considering the minuscule benefit when capturing most games (at least on high-end hardware) and the concerns with the additional capability maybe this isn't worth it, at least not by default. 

> This commit makes people suggest running obs with higher privileges, which they will probably do with sudo obs instead of fiddling with capabilities like you describe in order to get the benefit this supposedly gives

No, users shouldn't have to touch anything or run OBS as root. The package manager or an install script should set `CAP_SYS_NICE` on the OBS binary and nothing else. 

For reference, this is the script Arch runs when installing KWin to grant it the `CAP_SYS_NICE` capability:
https://raw.githubusercontent.com/archlinux/svntogit-packages/packages/kwin/trunk/kwin.install

> How was the in-game FPS affected in the before/after tests?

In-game FPS isn't affected at all, though Quake II RTX runs at ~30 FPS to begin with (I'm using RADV, not AMD's proprietary driver).

kkartaltepe:
> No, users shouldn't have to touch anything or run OBS as root. The package manager or an install script should set CAP_SYS_NICE on the OBS binary and nothing else.

And when the package maintainer has chosen not to do this, then the user is enticed to run obs as root to take advantage of this code. Which is precisely what we dont want to do. Given that you have not talked to every package manager this is more likely how this will be used by users and thus forever recorded in history as the thing to do if you want more fps for obs.

HannesMann:
Doesn't seem to benefit that much on my laptop either. Considering the security issues this clearly isn't worth it.

PR #6694, opened 2022-07-07
obs-scripting: Fix swig runtime header generation for macOS

PatTheMav:
### Description
Fixes swig runtime header generation failing on macOS with updated obs-deps.

### Motivation and Context
Due to swig using a hardcoded swig library path in self-compiled variants, we need to pass our custom swig library path with every invocation of swig on macOS.

CMake's `add_custom_command` function does _not_ inherit the environment variables used during configuration, hence why it needs to be passed explicitly on invocation.

Usage of CMake's "-E env" command module should allow for cross-platform compatible invocation.

### How Has This Been Tested?
* Tested with full OBS-Studio build on macOS
* Tested with full OBS-Studio build on Windows

### Types of changes

### Checklist:


RytoEX (approved):
Tested against the latest obs-deps and the macOS jobs now successfully find SWIG, Lua, and Python components.

Issue #4653, opened 2021-05-05
Can't stream Vulkan games using game capture (specifically Rainbow Six Siege and Detroit: Become Human)


AkalaGit:
### Operating System Info

Windows 10

### OBS Studio Version

26.1.1

### OBS Studio Log URL

https://obsproject.com/logs/xLo52huy_4xWcYwk

### Expected Behavior

I expect Vulkan games to be detected by OBS.

### Current Behavior

Black screen. This has only been happening for a week so I'm not sure what changed.

### Steps to Reproduce

1. Open Rainbow Six Siege or Detroit: Become Human
2. Use Game Capture to capture either game
3.
...

### Anything else we should know?

Haven't had issues until last week. I'm pretty much always on the latest Windows version.

RytoEX:
Since 26.1.1 has been working previously and we haven't shipped any out-of-band changes to Vulkan capture, we think this is related to a recent Windows update rather than anything we've changed.

RytoEX:
Following up, as this is either an issue with a Windows Update or a support issue, I'm going to close this Issue.  Please use the forums or Discord for support.  Thank you for understanding.

https://obsproject.com/help

AkalaGit:
Okay, thanks. Is there a specific Windows version that I should rollback to or would that not help?

PR #6698, opened 2022-07-08
graphics-hook: Avoid conflict between Vulkan and DXGI Present

elaberge:
### Description
Skip DXGI interception when called as a result of a Vulkan "Present" call.

### Motivation and Context
Some Vulkan apps for Windows use a DXGI backend, which is called by the Vk "Present" API call. That makes OBS double-hooked during the same call, first by the Vulkan hook, which then call its native implementation, which then calls the DXGI hook in return.

That makes the Vulkan call to capture_should_init true, then the D3D11 call to capture_should_init false, causing the d3d11_capture method to call d3d11_shmem_capture (as data.using_shtex still having its default "false" value), which in turn uses an unititialized mutex as the d3d11_init method has not been called, and leads to a crash.

### How Has This Been Tested?
This was tested on Windows 10 20H2, on a RTX 2070 Super with driver version 516.40, with the project at the last tagged revision (27.2.4).
Tests were done with the stock version of the graphics-hook64.dll for comparison, and with Debug versions for development. Final tests were done back with a RelWithDebug version to confirm the fix.
Verified using the Vulkan build of Rainbow Six Siege.

### Types of changes

### Checklist:

### DIFF
--- a/plugins/win-capture/graphics-hook/dxgi-capture.cpp
+++ b/plugins/win-capture/graphics-hook/dxgi-capture.cpp
@@ -198,6 +198,13 @@ static void update_mismatch_count(bool match)
 static HRESULT STDMETHODCALLTYPE hook_present(IDXGISwapChain *swap,
 					      UINT sync_interval, UINT flags)
 {
+	if (should_passthrough()) {
+		dxgi_presenting = true;
+		const HRESULT hr = RealPresent(swap, sync_interval, flags);
+		dxgi_presenting = false;
+		return hr;
+	}
+
 	const bool capture_overlay = global_hook_info->capture_overlay;
 	const bool test_draw = (flags & DXGI_PRESENT_TEST) != 0;
 
@@ -255,6 +262,14 @@ static HRESULT STDMETHODCALLTYPE
 hook_present1(IDXGISwapChain1 *swap, UINT sync_interval, UINT flags,
 	      const DXGI_PRESENT_PARAMETERS *params)
 {
+	if (should_passthrough()) {
+		dxgi_presenting = true;
+		const HRESULT hr =
+			RealPresent1(swap, sync_interval, flags, params);
+		dxgi_presenting = false;
+		return hr;
+	}
+
 	const bool capture_overlay = global_hook_info->capture_overlay;
 	const bool test_draw = (flags & DXGI_PRESENT_TEST) != 0;
 
--- a/plugins/win-capture/graphics-hook/vulkan-capture.c
+++ b/plugins/win-capture/graphics-hook/vulkan-capture.c
@@ -115,6 +115,8 @@ struct vk_data {
 	ID3D11DeviceContext *d3d11_context;
 };
 
+__declspec(thread) int vk_presenting = 0;
+
 /* ------------------------------------------------------------------------- */
 
 static void *vk_alloc(const VkAllocationCallbacks *ac, size_t size,
@@ -1216,7 +1218,14 @@ static VkResult VKAPI_CALL OBS_QueuePresentKHR(VkQueue queue,
 		vk_capture(data, queue, info);
 	}
 
-	return funcs->QueuePresentKHR(queue, info);
+	if (vk_presenting != 0) {
+		flog("non-zero vk_presenting: %d", vk_presenting);
+	}
+
+	vk_presenting++;
+	VkResult res = funcs->QueuePresentKHR(queue, info);
+	vk_presenting--;
+	return res;
 }
 
 /* ======================================================================== */
--- a/plugins/win-capture/graphics-hook/graphics-hook.h
+++ b/plugins/win-capture/graphics-hook/graphics-hook.h
@@ -248,6 +248,19 @@ static inline bool capture_should_init(void)
 	return should_init;
 }
 
+#if COMPILE_VULKAN_HOOK
+extern __declspec(thread) int vk_presenting;
+#endif
+
+static inline bool should_passthrough()
+{
+#if COMPILE_VULKAN_HOOK
+	return vk_presenting > 0;
+#else
+	return false;
+#endif
+}
+
 #ifdef __cplusplus
 }
 #endif


jp9000:
> Some Vulkan apps for Windows use a DXGI backend

Which ones?

jpark37 (changes requested):
Not sure we want to detach the hook functions. Might be safer to signal to the DXGI hook functions that they should become passthroughs.

I'd also like to know if there are any popular games/applications to test this against because there is regression risk any time the handshaking is altered, and I'd rather not do this if this is just for a theoretical sample.

elaberge:
This happens with Rainbow Six Siege, when launched in Vulkan. Possibly related to issue #4653.
Launching R6, then OBS and starting game capture might or might not work, and just stopping and restarting capture while the game is running, even on the title screen, causes the game to crash easily, or at best, get a black screen.
With the fix (tested by turning off BattleEye with R6 option "/belaunch -be", required due to the custom build DLL being discarded by the anti-cheat system), I can toggle capture as long as I wish without the game ever crashing.

elaberge:
I agree it might be better to use a passthrough implementation, will do.

jpark37:
Sorry, I missed that you said DXGI present was literally happening inside Vulkan present. Are you using DXVK on Windows?

elaberge:
> Sorry, I missed that you said DXGI present was literally happening inside Vulkan present. Are you using DXVK on Windows?

This is just plain nvidia drivers, with any extra layer disabled to make sure.

elaberge:
That last change is only to keep the history I force-pushed over the first implementation.

jpark37:
I enabled HOOK_VERBOSE_LOGGING, and I'm not seeing any DXGI present logging. I'm launching the Vulkan build through Steam with BattlEye off. Vulkan capture seems to be working fine.

This also seems annoying to debug because the process seems to terminate when I attach VS debugger.

elaberge:
Yeah, debugging without a debugger is a PITA. I found that issue with game code I can actually compile, it's just that R6 seems to exhibit the same problem, and it actually helps me validate the issue.

For the record, here's a run with HOOK_VERBOSE_LOGGING without this fix:
[log-default-2022-07-10 16-45-52.txt](https://github.com/obsproject/obs-studio/files/9079886/log-default-2022-07-10.16-45-52.txt)

And here's with this fix:
[log-fixed-2022-07-10 16-53-20.txt](https://github.com/obsproject/obs-studio/files/9079885/log-fixed-2022-07-10.16-53-20.txt)

elaberge:
Here you go:
```
>	graphics-hook64.dll!hook_present(IDXGISwapChain * swap, unsigned int sync_interval, unsigned int flags) Line 206	C++
 	nvoglv64.dll!00007fff9c1022b3()	Unknown
 	nvoglv64.dll!00007fff9c1351e4()	Unknown
 	nvoglv64.dll!00007fff9c1131c2()	Unknown
 	nvoglv64.dll!00007fff9c136869()	Unknown
 	nvoglv64.dll!00007fff9c23128f()	Unknown
 	nvoglv64.dll!00007fff9c23da5a()	Unknown
 	nvoglv64.dll!00007fff9c2548b5()	Unknown
 	nvoglv64.dll!00007fff9c0d13c6()	Unknown
 	nvoglv64.dll!00007fff9c219f13()	Unknown
 	nvoglv64.dll!00007fff9c1f69a4()	Unknown
 	nvoglv64.dll!00007fff9c1f5f8f()	Unknown
 	nvoglv64.dll!00007fff9c47f169()	Unknown
 	nvoglv64.dll!00007fff9c4810ad()	Unknown
 	nvoglv64.dll!00007fff9c47bc93()	Unknown
 	nvoglv64.dll!00007fff9c563c5a()	Unknown
 	graphics-hook64.dll!OBS_QueuePresentKHR(VkQueue_T * queue, const VkPresentInfoKHR * info) Line 1226	C
```
Followed by game code calling vkQueuePresentKHR

plugins/win-capture/graphics-hook/dxgi-capture.cpp
```diff
@@ -201,6 +201,13 @@ static HRESULT STDMETHODCALLTYPE hook_present(IDXGISwapChain *swap,
 	const bool capture_overlay = global_hook_info->capture_overlay;
 	const bool test_draw = (flags & DXGI_PRESENT_TEST) != 0;
 
+	if (should_passthrough()) {
+		dxgi_presenting = true;
+		const HRESULT hr = RealPresent(swap, sync_interval, flags);
+		dxgi_presenting = false;
+		return hr;
+	}
+
```
jpark37, line 210: Move this block to the top of the function.

plugins/win-capture/graphics-hook/dxgi-capture.cpp
```diff
@@ -258,6 +265,14 @@ hook_present1(IDXGISwapChain1 *swap, UINT sync_interval, UINT flags,
...
 	const bool test_draw = (flags & DXGI_PRESENT_TEST) != 0;
 
+	if (should_passthrough()) {
+		dxgi_presenting = true;
+		const HRESULT hr =
+			RealPresent1(swap, sync_interval, flags, params);
+		dxgi_presenting = false;
+		return hr;
+	}
+
```
jpark37, line 275: Move this block to the top of the function.

jpark37:
What an unexpected call stack... I guess we have to trust that this is legitimate. Just make or at least respond to the requested changes. As long as this doesn't cause a regression, I'll approve it.

jpark37:
Should also update the PR description

jpark37:
Should also squash your commits because we don't want to merge the history of your changes. This should just be one commit.

jpark37:
There's still an item remaining.

jpark37:
Apparently review comments are invisible to everyone except the reviewer until a review is started. Hopefully you can see the feedback now.

plugins/win-capture/graphics-hook/vulkan-capture.c
```diff
@@ -115,6 +115,8 @@ struct vk_data {
 	ID3D11DeviceContext *d3d11_context;
 };
 
+volatile int vk_presenting = 0;
```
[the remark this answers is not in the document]
elaberge: That was my first idea, but the shared graphics-hook.h used by both plain C and C++ code disallowed using the straight thread_local specifier. The project configures pre-C11 compilation, which doesn't have that concept. I went with the __declspec(thread) extension which works with both, as that is Windows-code only anyway.

jpark37 (approved):

PR #6700, opened 2022-07-08
UI: Improve simple mode replay buffer settings

gxalpha:
### Description

Instead of a checkbox in the recording group that reveals the replay
buffer group when pressed, the replay buffer group now has its own
toggle that enables or disables the replay buffer.

Before:
image
image

After:
image
image

### Motivation and Context

There currently is an element of surprise which should be avoided. When clicking the "Enable replay buffer" checkbox, as a visually unrelated group pops up below the recording group, or gets hidden again.
With this change, it's clear on first sight what the checkbox does and what the user can expect when clicking it.

### How Has This Been Tested?

macOS 13
Enabled/Disabled replay buffer, and enabled/disabled lossless quality (and opened the settings window in those states).

### Types of changes

### Checklist:


gxalpha:
It was discussed off-thread that this is generally fine, but buries the warnings which is undesirable. The warnings should rather be floating on the bottom, like so (Red = ScrollView, Blue = Field for warnings):
image
Converting to draft until that is a thing.

gxalpha:
Waits for #7709

gxalpha:
With https://github.com/obsproject/obs-studio/pull/9338 merged, I rebased this on master so it should be ready for review.
Needs @Warchamp7 approval.

Warchamp7 (approved):

UI/forms/OBSBasicSettings.ui
```diff
@@ -7855,7 +7851,6 @@
   <tabstop>simpleOutRecFormat</tabstop>
   <tabstop>simpleOutRecEncoder</tabstop>
   <tabstop>simpleOutMuxCustom</tabstop>
-  <tabstop>simpleReplayBuf</tabstop>
```
WizardCM, line 7858: Missing tabstop, so users might not be able to keyboard-navigate to the settings?

gxalpha: You're right, initially this was removed when the group box kept its name (when I then only changed later).
Undone.

PR #6702, opened 2022-07-09
UI: Add support for theme meta, parent theme palette

WizardCM:
### Description

This enhances the existing theme loading system with Theme Meta, which allows individual themes to provide additional information to OBS.

The primarily goal is for a theme to define a "parent", allowing to extend existing themes with additional attributes, including extending/replacing an existing theme's palette.

### Motivation and Context

Warchamp expressed he would like the ability to provide "palatte swaps" of a single theme - such as dark/light variants, without having to clone the entire QSS file.

While I've updated existing themes to include metadata, I'm leaving creation of palette swaps to Warchamp.

A future version of this will allow for additional QSS, but that requires other changes, so I'm submitting this separately first.

### How Has This Been Tested?

Create a new theme file `YamiChild.qss` with the following contents and load it. It should contain all Yami styling but with yellow text.

```css

OBSThemeMeta {
    parent: 'Yami';
    dark: 'true';
    author: 'WizardCM';
}

OBSTheme {
    text: rgb(234, 255, 0);
}
```

image

### Types of changes

### Checklist:

### DIFF
--- a/UI/data/themes/Acri.qss
+++ b/UI/data/themes/Acri.qss
@@ -1,3 +1,8 @@
+OBSThemeMeta {
+    dark: 'true';
+    author: 'Warchamp7';
+}
+
 /* OBSTheme, main QApplication palette and QML values */
 OBSTheme {
     window: rgb(24,24,25);
--- a/UI/obs-app.cpp
+++ b/UI/obs-app.cpp
@@ -1095,10 +1095,72 @@ void OBSApp::ParseExtraThemeData(const char *path)
 	setPalette(pal);
 }
 
-bool OBSApp::SetTheme(std::string name, std::string path)
+OBSThemeMeta *OBSApp::ParseThemeMeta(const char *path)
 {
-	theme = name;
+	BPtr<char> data = os_quick_read_utf8_file(path);
+	CFParser cfp;
+	int ret;
 
+	if (!cf_parser_parse(cfp, data, path))
+		return nullptr;
+
+	if (cf_token_is(cfp, "OBSThemeMeta") ||
+	    cf_go_to_token(cfp, "OBSThemeMeta", nullptr)) {
+		OBSThemeMeta *meta = new OBSThemeMeta();
+		if (!cf_next_token(cfp))
+			return nullptr;
+
+		if (!cf_token_is(cfp, "{"))
+			return nullptr;
+
+		for (;;) {
+			if (!cf_next_token(cfp))
+				return nullptr;
+
+			ret = cf_token_is_type(cfp, CFTOKEN_NAME, "name",
+					       nullptr);
+			if (ret != PARSE_SUCCESS)
+				break;
+
+			DStr name;
+			dstr_copy_strref(name, &cfp->cur_token->str);
+
+			ret = cf_next_token_should_be(cfp, ":", ";", nullptr);
+			if (ret != PARSE_SUCCESS)
+				continue;
+
+			if (!cf_next_token(cfp))
+				return nullptr;
+
+			ret = cf_token_is_type(cfp, CFTOKEN_STRING, "value",
+					       ";");
+
+			if (ret != PARSE_SUCCESS)
+				continue;
+
+			char *str;
+			str = cf_literal_to_str(cfp->cur_token->str.array,
+						cfp->cur_token->str.len);
+
+			if (strcmp(name->array, "dark") == 0 && str) {
+				meta->dark = strcmp(str, "true") == 0;
+			} else if (strcmp(name->array, "parent") == 0 && str) {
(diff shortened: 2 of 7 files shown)


WizardCM:
Known issues:
* bfree() causes things to go wonky
* If OBSThemeMeta is the very first line in the file, it'll get skipped entirely

UI/obs-app.cpp
```diff
@@ -1098,10 +1098,75 @@ void OBSApp::ParseExtraThemeData(const char *path)
...
+bool OBSApp::ParseThemeMeta(const char *path)
 {
-	theme = name;
+	BPtr<char> data = os_quick_read_utf8_file(path);
+	CFParser cfp;
+	int ret;
+
+	cf_parser_parse(cfp, data, path);
+
+	themeMeta = nullptr;
```
jpark37, line 1109: Not really a fan of this side effect. Would be cleaner to return the OBSThemeMeta object instead of a bool, and assign externally, i.e.: `themeMeta = ParseThemeMeta(path);`

UI/obs-app.cpp
```diff
@@ -1098,10 +1098,75 @@ void OBSApp::ParseExtraThemeData(const char *path)
...
 
-bool OBSApp::SetTheme(std::string name, std::string path)
+bool OBSApp::ParseThemeMeta(const char *path)
 {
-	theme = name;
+	BPtr<char> data = os_quick_read_utf8_file(path);
+	CFParser cfp;
+	int ret;
+
+	cf_parser_parse(cfp, data, path);
```
jpark37, line 1107: Check return value to ensure we don't crash on empty file.

UI/obs-app.cpp
```diff
@@ -1098,10 +1098,75 @@ void OBSApp::ParseExtraThemeData(const char *path)
...
+
+	themeMeta = nullptr;
+
+	if (cf_go_to_token(cfp, "OBSThemeMeta", nullptr)) {
```
jpark37, line 1111: Add `cf_token_is` test to avoid `cf_go_to_token` skipping first token.

UI/obs-app.cpp
WizardCM: Done.

UI/obs-app.cpp
WizardCM: Done.

UI/obs-app.cpp
WizardCM: Done. Thanks for the suggestion.

WizardCM:
Alongside the above suggestions, I've also switched away from char * to std::string to be consistent with OBSApp.

Ready for re-review.

UI/data/themes/Acri.qss
```diff
@@ -1,3 +1,9 @@
+
```
jpark37, line 1: Starting a file with a new line is kind of weird.

UI/obs-app.cpp
```diff
@@ -1100,10 +1100,75 @@ void OBSApp::ParseExtraThemeData(const char *path)
...
+				meta->author = std::string(str);
+			}
+			bfree(str);
+			meta->parent = "Yami";
+
+			if (!cf_go_to_token(cfp, ";", nullptr))
+				return nullptr;
+		}
+
+		themeDarkMode = meta->dark;
```
jpark37, line 1164: Didn't notice this other side effect. Can assign outside function, maybe near where you're already setting themeDarkMode in the null case.

UI/obs-app.hpp
```diff
@@ -103,6 +110,7 @@ class OBSApp : public QApplication {
 	QPalette defaultPalette;
 
 	void ParseExtraThemeData(const char *path);
+	OBSThemeMeta *ParseThemeMeta(const char *path);
```
jpark37, line 113: Mark static if possible.

UI/data/themes/Acri.qss
WizardCM: This must be left over from when the first token couldn't be `OBSThemeMeta`. I'll remove this.

WizardCM: Done.

UI/obs-app.cpp
WizardCM: Done.

UI/obs-app.hpp
WizardCM: Done.

UI/obs-app.cpp
jpark37: Now that there are two sides, test for `themeMeta` instead of `!themeMeta`

WizardCM: Good point. Done.

PR #6705, opened 2022-07-09
UI: Add bitdepth and chroma subsampling to output pixel format name

mvji:
### Description

Change output format name in Advanced Settings to have something more easy to understand for non developpers.
Add Bitdepth, chroma subsampling and number of planes information

Before : 
BEFORE

After:
UPDATE2

### Motivation and Context

Using fourCC to describe pixel format, is not very easy to understand.
Even people with video technical skills, doesn't know all the fourCC.
Add some information to more easy understand the output pixel format selected.

### How Has This Been Tested?

Read Xcode debug log, to check that the color format is correctly change after the update
Check that the basic.ini file have correct values after this patch.

### Types of changes

### Checklist:


Nofew:
I like this a lot! Isn't it rgb, not rgba, though? I thought OBS doesn't output alpha.

RytoEX:
I don't like that the string comparisons have to be much more precise as a result of this change.

UI/window-basic-main.cpp
```diff
@@ -4282,24 +4283,16 @@ static inline enum obs_scale_type GetScaleType(ConfigFile &basicConfig)
 
 static inline enum video_format GetVideoFormatFromName(const char *name)
 {
-	if (astrcmpi(name, "I420") == 0)
+	if (astrcmpi(name, "4:2:0 8 bits (I420)") == 0)
```
derrod, line 4286: There must be a better way to do this. You can probably have a separate internal and displayed string in Qt.

derrod (changes requested):

UI/window-basic-main.cpp
RytoEX: You can, but it requires either abusing translations or using a data model. The former is hacky, and the latter cannot be done from within a .ui file.

jpark37:
The number of planes might be useful to know too:
NV12 - 2
I420 - 3
I444 - 3
P010 - 2
I010 - 3

mvji:
> I like this a lot! Isn't it rgb, not rgba, though? I thought OBS doesn't output alpha.

Change in last update to don't change ui name from RGB to RGBA (internally Obs use RGBA pixel format), but quickly tested in recording mode and seems like Obs can't record with alpha, so maybe better to let it named RGB

The last update, change the way ColorFormat is internally managed in the code (in the commit 2)
By adding functions from/to ui_index, from_to internal text (use in the .ini setting file)
We can't use the video format to const char existing function because RGBA video format is called RGB in the ui.

For now, i put all functions connected to colorformat ui conversion at the end of "window-basic-main.hpp", because don't really know the best place to put it. Probably need to be move to a better place

The benefits of this, is that UI name is not anymore used in the code for colorformat.

UI/window-basic-main.cpp
jpark37: Doesn't this also break compatibility with existing config files? We had a similar issue with making the Color Space entries more descriptive, so we use translations.

jpark37:
I feel like the text is reversed? To me, the format is the main info, and the rest is just support text. I think we could also use some commas.

For example:
NV12 (8 bits, 4:2:0, 2 planes)
...
RGB (8 bits)

Not that I have any authority over the UI strings.

UI/window-basic-main.cpp
mvji: Yes, in the previous version, existing config files will break (and i made a mistake in the previous version (changing the text order without updating this func). In the last version, only "internal" name is used.

jpark37: You won't need as many code changes if you use the translation system.

mvji:
For me, the fourcc of the format is not known by lot of people, but other part is more common (bitdepth, chroma subsampling). But don't have strong opinion on this. I will wait a little bit if other people want other text, before updating.

UI/window-basic-main.cpp
mvji: > You won't need as many code changes if you use the translation system.

Thanks for pointing this. Much cleaner way to change text, and enable translation

mvji:
Update using translation to "rename" the color format. Values save in settings will not be change by this patch (only text display)

UI/window-basic-main.cpp
```diff
@@ -4292,14 +4292,6 @@ static inline enum video_format GetVideoFormatFromName(const char *name)
...
 	else if (astrcmpi(name, "P010") == 0)
 		return VIDEO_FORMAT_P010;
-#if 0 //currently unsupported
-	else if (astrcmpi(name, "YVYU") == 0)
-		return VIDEO_FORMAT_YVYU;
-	else if (astrcmpi(name, "YUY2") == 0)
-		return VIDEO_FORMAT_YUY2;
-	else if (astrcmpi(name, "UYVY") == 0)
-		return VIDEO_FORMAT_UYVY;
-#endif
```
RytoEX, line 4302: I'm not sure if it's worth removing these in this PR as it adds to the scope. These can be removed in a separate follow-up PR.

RytoEX (changes requested):
Nit about scope.

UI/data/locale/en-US.ini
```diff
@@ -1049,6 +1049,12 @@ Basic.Settings.Advanced.FormatWarning10BitSdr="Warning: 10-bit formats are more
 Basic.Settings.Advanced.FormatWarning2100="Warning: Rec. 2100 should use a format with more precision."
 Basic.Settings.Advanced.Audio.BufferingTime="Audio Buffering Time"
 Basic.Settings.Advanced.Video.ColorFormat="Color Format"
+Basic.Settings.Advanced.Video.ColorFormat.NV12="NV12 (8 bits, 4:2:0, 2 planes)"
```
WizardCM, line 1052: After some internal discussion, please change "8 bits" and "10 bits" to "8-bit" and "10-bit" respectively.

UI/window-basic-main.cpp
mvji: I include this patch in the first version, because the dead code was in the function modified in this PR.
Will remove in the next update

UI/data/locale/en-US.ini
mvji: Ok will change in the next update

Joefis-x20s:
> I like this a lot! Isn't it rgb, not rgba, though? I thought OBS doesn't output alpha.

OBS does output alpha for NDI when set to RGB

PR #6696, opened 2022-07-07
CI: Update deps to obs-deps 2022-07-08 release

RytoEX:
### Description

CI: Update deps to obs-deps 2022-07-08 release

Notable changes:
 * Windows: Add Qt 5 and 6 builds
 * Windows: Remove nasm native build
 * Windows: Remove pthread-win32 cross-compile build
 * Windows: Update srt from 1.4.2 to 1.4.4
 * Windows: Update SWIG from 3.0.12 to 4.1.0-git
 * macOS: Update libvpx from 1.10.0 to 1.11.0
 * macOS: Update srt from 1.4.1 to 1.4.4
 * macOS: Update SWIG from 3.0.12 to 4.1.0-git
 * macOS: Update x264 from r3059 to r3060

Caveats:
1. Qt5 is hard-coded for now when fetching Qt deps in the build scripts.
2. The Windows deps URL changed format because the Windows obs-deps releases no longer have the "win-" prefix. Attempts to switch back to older Windows obs-deps releases after this will require reverting this change.

### Motivation and Context

Update obs-deps to the new release assets. Unblock PRs:
* #6134 
* #6706 

### How Has This Been Tested?

Previously ran CI on my fork successfully. I've checked that the Windows builds launch, and I've done some basic testing on the x64 build. Ingesting either an SRT or RIST stream to Media Source works.

Additional testing and sanity checking appreciated.

### Types of changes

- New feature (non-breaking change which adds functionality)

### Checklist:

- [x] My code has been run through [clang-format](https://github.com/obsproject/obs-studio/blob/master/.clang-format).
- [x] I have read the [**contributing** document](https://github.com/obsproject/obs-studio/blob/master/CONTRIBUTING.rst).
- [x] My code is not on the master branch.
- [x] The code has been tested.
- [x] All commit messages are properly formatted and commits squashed where appropriate.
- [x] I have included updates to all appropriate documentation.

PR #6553, opened 2022-05-28
obs-scripting: Fix missing obs-frontend-api bindings for scripting plugins

PatTheMav:
### Description
Fixes `obs-frontend-api` symbols not being available to scripting environments since merging the rework.

Fixes https://github.com/obsproject/obs-studio/issues/6539

### Motivation and Context
The required preprocessor definition activating support for `obs-frontend-api` within the scripting plugin needs to be set for the actual C/C++ compiler as well as the swig pre-compiler.

These commits ensure that the definition exists if UI is enabled in the project.

### How Has This Been Tested?
* Compiled locally and checked for existing symbols using the Lua file provided in [#6539](https://github.com/obsproject/obs-studio/issues/6539).

### Types of changes
- Bug fix (non-breaking change which fixes an issue)

### Checklist:
- [x] My code has been run through [clang-format](https://github.com/obsproject/obs-studio/blob/master/.clang-format).
- [x] I have read the [**contributing** document](https://github.com/obsproject/obs-studio/blob/master/CONTRIBUTING.rst).
- [x] My code is not on the master branch.
- [x] The code has been tested.
- [x] All commit messages are properly formatted and commits squashed where appropriate.
- [x] I have included updates to all appropriate documentation.

Issue #6539, opened 2022-05-25
Lua: Many API functions and constants available in 27.2.4 aren't available in a build of recent master


OldBaldGeek:
### Operating System Info

Windows 10

### OBS Studio Version

Other

### OBS Studio Version (Other)

27.2.0-431-gfb639e303 (64-bit) 5/24/2022 11:29:46 AM from obsbot.rodney.io, and in my local builds

### OBS Studio Log URL

https://obsproject.com/logs/mQ7E5p4qKMgVD4M8

### Expected Behavior

Lua should be able to call API functions such as obs.obs_frontend_get_current_scene and obs.obs_frontend_get_current_preview_scene.

Lua should be able to use API constants such as obs.OBS_FRONTEND_EVENT_SCENE_CHANGED

These items work in the 27.2.4 (64 bit) release, and worked in earlier versions at least back to 27.1.3.

Examples of use may be found in scripts such as refresh-browser-specific.lua [https://obsproject.com/forum/resources/reload-browser-in-preview-scene.1030/](url)

obs-schedulecountdown.lua [https://obsproject.com/forum/resources/advanced-countdown-with-schedule.1306/](url)

obs-visca-control-1.6 [https://obsproject.com/forum/resources/control-visca-over-ip-based-cameras.1173/](url)

### Current Behavior

Lua can call obs.obs_frontend_get_current_scene, but obs.obs_frontend_get_current_preview_scene fails, claiming nil value. 
Various other functions are also missing.

Similarly, many constants such as obs.OBS_FRONTEND_EVENT_SCENE_CHANGED are nil.

The attached script explicitly tests several functions and constants for undefined/nil. Results are shown in the log and in the script UI
[wtf.lua.txt](https://github.com/obsproject/obs-studio/files/8772507/wtf.lua.txt)

### Steps to Reproduce

1. Build master, or get a build from obsbot.rodney.io
2. Run OBS, and install wtf.lua, or one of the scripts listed in "Expected Behavior"
3. Script UI will show results for wtf.lua
4. For refresh-browser-specific.lua, pressing the "Refresh Browsers" button will log an error
5. For the other mentioned scripts, some configuration is required. But inspection of the code shows usage of the missing items.
...

RytoEX:
Testing out each of these items...

[...]

PR #6706, opened 2022-07-09
obs-scripting: Switch Python support to use the Py3 stable ABI

PatTheMav:
### Description
Changes the Python module of obs-scripting to use the Python 3 stable ABI and also allows OBS to load several different recent Python 3 versions (any possible version between 3.6 and 3.10 will be loaded, if available).

Also changes the macOS UI to just require the base path a `Python.framework` is placed in (instead of requiring users to dive into the Framework to find the library path).

**Note:** This PR implements the changes introduced by https://github.com/obsproject/obs-studio/pull/6696 to pull the swig binaries compiled on CI with stable Abu support.

### Motivation and Context
In theory these changes should allow users to run Python scripts in OBS with a wider range of available Python versions installed on the host system, independent from the version used to compile OBS.

Caveats: While the obspython module uses the stable ABI, obs-scripting itself cannot do the same, as OBS uses C-API methods not available in the stable ABI. As the use of these methods is overseen by OBS development, additional care can be put into their implementation, ensuring that only methods available/unchanged in respective Python versions are used.

The stable ABI itself does not contain many C-API methods that _should_ be contained, for the simple reason of not being as actively maintained/updated as Python itself. Hence why we chose a "better is the enemy of good" approach.

### How Has This Been Tested?
Compiled and ran OBS with Python scripts on:
* Windows (Python 3.6 and 3.9)
* Linux (Python 3.10)
* macOS (Python 3.9 and Python 3.10)

### Types of changes

### Checklist:

### DIFF
--- a/deps/obs-scripting/obspython/CMakeLists.txt
+++ b/deps/obs-scripting/obspython/CMakeLists.txt
@@ -21,7 +21,9 @@ endif()
 include(UseSWIG)
 
 set_source_files_properties(
-  obspython.i PROPERTIES USE_TARGET_INCLUDE_DIRECTORIES TRUE SWIG_FLAGS "-py3")
+  obspython.i
+  PROPERTIES USE_TARGET_INCLUDE_DIRECTORIES TRUE
+             SWIG_FLAGS "$<IF:$<BOOL:${OS_LINUX}>,-py3,-py3-stable-abi>")
 
 swig_add_library(
   obspython
--- a/UI/frontend-plugins/frontend-tools/scripts.cpp
+++ b/UI/frontend-plugins/frontend-tools/scripts.cpp
@@ -652,6 +652,19 @@ extern "C" void InitScripts()
 	const char *python_path =
 		config_get_string(config, "Python", "Path" ARCH_NAME);
 
+#ifdef __APPLE__
+	std::string _python_path(python_path);
+	std::size_t pos = _python_path.find("/Python.framework/Versions");
+
+	if (pos != std::string::npos) {
+		std::string _temp = _python_path.substr(0, pos);
+		config_set_string(config, "Python", "Path" ARCH_NAME,
+				  _temp.c_str());
+		config_save(config);
+		python_path = _temp.c_str();
+	}
+#endif
+
 	if (!obs_scripting_python_loaded() && python_path && *python_path)
 		obs_scripting_load_python(python_path);
 #endif
--- a/cmake/Modules/FindPythonWindows.cmake
+++ b/cmake/Modules/FindPythonWindows.cmake
@@ -20,7 +20,7 @@ find_path(
 
 find_library(
   PYTHON_LIB
-  NAMES ${_PYTHON_LIBRARIES} python36
+  NAMES ${_PYTHON_LIBRARIES} python3
   HINTS ${_PYTHON_LIBRARY_DIRS}
   PATH_SUFFIXES
     lib${_lib_suffix}
--- a/cmake/Modules/ObsDefaults_Linux.cmake
+++ b/cmake/Modules/ObsDefaults_Linux.cmake
@@ -93,8 +93,8 @@ macro(setup_obs_project)
     set(OBS_DATA_PATH "../../${OBS_DATA_DESTINATION}")
 
     set(OBS_SCRIPT_PLUGIN_PATH "../../${OBS_SCRIPT_PLUGIN_DESTINATION}")
-    set(CMAKE_INSTALL_RPATH
-        "$ORIGIN/" "${CMAKE_INSTALL_PREFIX}/${OBS_LIBRARY_DESTINATION}")
+    set(CMAKE_INSTALL_RPATH "$ORIGIN/"
+                            "${ORIGIN}/../../${OBS_LIBRARY_DESTINATION}")
   endif()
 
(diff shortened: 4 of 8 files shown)


RytoEX (approved):
Seems fine to me at a glance. Will need to merge #6696 first.

WizardCM:
During my own testing, the CI artifact of this build only worked with Python 3.10 on Windows. Both 3.6 and 3.9 fail to load.

RytoEX:
# Python 3.6

Using Python 3.6.2 installed via pyenv-win doesn't seem to load anything at all. Scripts report "no properties available" and nothing appears in the Script Log. The OBS log reports:
```
22:52:47.032: [Python] Could not load library: C:/pyenv-win/versions/3.6.2/
22:54:01.254: [Python] Could not load library: C:/pyenv-win/versions/3.6.2/Lib/
22:54:18.071: [Python] Python failure in get_defaults:168:
```


# Python 3.9

Using Python 3.9 installed with the default Python installer into Program Files, OBS loads Python, but fails to execute my test script, claiming that "obs_frontend_preview_program_mode_active" does not exist. This reminds me of #6539 where Lua scripts became unable to access Frontend APi functions that were not _explicitly_ made visible to scripts, where previously we relied on some automagic to make them available. See #6553 for how that was addressed at the time.

My test script also says that OBS_FRONTEND_EVENT_STREAMING_STARTING is undefined, which is similar to the linked Issue.

# Python 3.10

With Python 3.10 installed with the default Python installer, I get the same results described in the Python 3.9 section above.

PatTheMav:
> During my own testing, the CI artifact of this build only worked with Python 3.10 on Windows. Both 3.6 and 3.9 fail to load.

Yeah that's an oversight, I fixed this locally already. I can confirm the issues reported by @RytoEX on Windows (macOS works fine).

PatTheMav:
Just pushed an update that should address both issues.

deps/obs-scripting/obspython/CMakeLists.txt
```diff
@@ -21,8 +21,10 @@ endif()
 include(UseSWIG)
 
 set_source_files_properties(
-  obspython.i PROPERTIES USE_TARGET_INCLUDE_DIRECTORIES TRUE SWIG_FLAGS
-                                                             "-builtin;-py3")
+  obspython.i
+  PROPERTIES USE_TARGET_INCLUDE_DIRECTORIES TRUE
+             SWIG_FLAGS
+             "$<IF:$<BOOL:${OS_LINUX}>,-builtin;-py3,-py3-stable-abi>")
```
RytoEX, line 27: The `-py3` flag is deprecated and no longer supported.

RytoEX (changes requested):
It looks like the use of `-py3` is emitting a deprecation warning due to SWIG dropping the flag.

https://github.com/swig/swig/commit/a343b7e254567a64761bc1be7dc55b7b7424ec52

RytoEX (approved):
Tests on Windows look good to me.

WizardCM:
To better clarify the above from Ryan:

I was running into two separate issues. In my opinion, **neither is a blocker for this PR** to be merged but at the very least both should be independently fixed later.

### 1. Python 3.8.0 and 3.8.1 load successfully, but fail to import the obspython.py/pyd

In this situation, the following is logged on startup or on first attempt of loading 3.8.0 or 3.8.1.

```
21:13:25.988: [Python] Python failure in obs_scripting_load_python:1724:
21:13:25.988: [Python] Error importing obspython.py', unloading obs-python
21:13:25.988: [Python] Failed to load python plugin
```

At this point in the code, the Python interpreter is fully loaded and can run Python code, but when it attempts to load our `obspython.py` it spits out the following into stderr, which is not exposed in any way and provides no additional insight. The error itself comes from the Python interpreter.

```
ImportError: DLL load failed while importing obspython: The specified module could not be found.
```

* For users, the fix is to update to 3.8.2 or above.

* For us, we should (at the very least) display some kind of error in the UI that our Python implementation failed to load, and we **should not save** the selected path.

### 2. If Python loads successfully but obspython doesn't, attempting to load a broken _or_ functional Python+obspython a second time causes OBS to crash

I have a feeling this is caused by two separate internal issues, one is related to `python_loaded_at_all` (which is false when it should be true) and `mutexes_loaded` being true when `tick_mutex` and `timer_mutex` are null, as issue (1) skips half the loading steps. There may be one more related to how other "python_is_loaded" variables function because of this.

Either way, without bypassing the `mutexes_loaded` conditional, I get the following crash with the CI build. https://obsproject.com/logs/6Z5oUwFtdYUkYgrL

Steps:

1. Load Python 3.8.0 or 3.8.1
2. Load *any* version fo Python, including 3.6, 3.8 or 3.10
3. Crash

PatTheMav:
From what I could tell the issue in Python 3.8.0 and Python 3.8.1 is related to Py 3.8 changing the way DLLs are loaded (or more precisely, which paths Python supplies as search paths for the library loader):

Before Py 3.8, Python used the `PATH` as possible locations for finding DLLs (and as such DLLs linked to from precompiled python modules), which allowed DLL hijacking. For security reasons this was removed, now the loading application needs to supply DLL search paths using Windows APIs.

I haven't figured out yet what changed in that regard with Python 3.8.2 (which doesn't seem to exhibit the same issue).

PR #6717, opened 2022-07-12
mac-capture: Add ScreenCaptureKit support for improved capture abilities on macOS 12.5+

PatTheMav:
### Description
Adds support for capturing content on macOS using macOS' ScreenCaptureKit framework, enabling:

* Full-Desktop capture
* Window capture
* Application capture (capturing all windows of a specific application in relation to the desktop)
* Audio capture of any selected screen/window/application

Hidden windows and windows positioned on separate spaces (i.e. full screen windows) can be selected and captured by enabling the option to show hidden windows. This allows capturing games and other applications using the built-in full screen API to run in full screen or use custom windows.

**Note:** The feature requires at least macOS 12.5. Additionally macOS 13.0 is required for audio capture support. Until GitHub supports Xcode 14 on GitHub runners, artefacts built on CI **will not(!)** contain audio capture.

### Motivation and Context
ScreenCaptureKit allows the capture of visual and audible contents of applications without the need of 3rd party software and is also guarded behind macOS' screen capture permissions (ensuring user privacy controls). 

Application capture also allows a new capture type for screen casts, ensuring that a whole desktop can be captured but only the contents of the selected application are visible.

### How Has This Been Tested?
* Tested on macOS 12.4 (Apple Silicon)
* Tested on macOS 13 Beta (Apple Silicon)

### Types of changes

### Checklist:

Closes https://github.com/obsproject/obs-studio/pull/5875
Closes https://github.com/obsproject/obs-studio/pull/6600


podycust:
macOS 12.6? that doesn't exist (well not yet).
I thought the screen capture feature was introduced in 12.3?

gxalpha:
@podycust there are a few bugs in ScreenCaptureKit on older 12.x versions (like 12.3) that are fixed in newer ones, with behavior changing between them. While it would work on 12.3, behavior changing between minor versions is not something we want to support, and there's no reason for users on 12.3/12.4/12.5 to not update to 12.6 once that is released.

PatTheMav:
> @gxalpha I get that but there isn't even a 12.6 beta yet (maybe apple had started on internally) the RC for 12.5 was only released the other day! And I see it been tested on 12.4 which is lower than the non existent 12.6

Correct, it was tested on those versions, which is how we came across these issues. To avoid a situation where we're inundated by support requests to whom our only possible answer would be "update your macOS once the next version is available" (which frankly would be disappointing), we're rather gating this feature behind version 12.6 which according to our available information should _not_ have those issues.

plugins/mac-capture/mac-screen-capture.m
```diff
@@ -0,0 +1,1142 @@
...
+
+	obs_properties_t *props = obs_properties_create();
+	obs_property_t *capture_type = obs_properties_add_list(
+		props, "type", obs_module_text("Method"), OBS_COMBO_TYPE_LIST,
+		OBS_COMBO_FORMAT_INT);
+	obs_property_list_add_int(capture_type,
+				  obs_module_text("DisplayCapture"), 0);
+	obs_property_list_add_int(capture_type,
+				  obs_module_text("WindowCapture"), 1);
+	obs_property_list_add_int(capture_type, "Application Capture", 2);
```
gxalpha, line 1021: This needs to be localized

PatTheMav: Fixed.

libobs-opengl/gl-cocoa.m
```diff
@@ -419,9 +419,8 @@ bool gs_texture_rebind_iosurface(gs_texture_t *texture, void *iosurf)
 		blog(LOG_ERROR, "Unexpected pixel format: %d (%c%c%c%c)", pf,
 		     pf >> 24, pf >> 16, pf >> 8, pf);
 
-	if (tex->width != IOSurfaceGetWidth(ref) ||
-	    tex->height != IOSurfaceGetHeight(ref))
-		return false;
+	tex->width = IOSurfaceGetWidth(ref);
+	tex->height = IOSurfaceGetHeight(ref);
```
gxalpha, line 423: Is this directly related to SCK or does it fix an issue only exposed by SCK?
I don't know the direct effects of this, but to me it feels like it should be split into a separate commit with a small explanation of what it does.

plugins/mac-capture/mac-screen-capture.m
```diff
@@ -0,0 +1,1143 @@
...
+	.video_tick = screen_capture_video_tick,
+	.video_render = screen_capture_video_render,
+
+	.get_width = screen_capture_getwidth,
+	.get_height = screen_capture_getheight,
+
+	.get_defaults = screen_capture_defaults,
+	.get_properties = screen_capture_properties,
+	.update = screen_capture_update,
+	.icon_type = OBS_ICON_TYPE_GAME_CAPTURE,
```
gxalpha, line 1115: I find the usage of the game capture icon a bit weird, feels like it should be window capture or maybe display capture instead - or was there some other motivation behind using the game capture icon?

gxalpha (commented):
Testing suggests this generally works fine. Got a few small points:

libobs-opengl/gl-cocoa.m
PatTheMav: I think this allows the source to scale with the size of the captured window - that is, if the window increases in size, the source automagically matches it.

plugins/mac-capture/mac-screen-capture.m
PatTheMav: My hunch is that SCK was mainly used for capturing games at Apple because of OBS historic main usage as a game capture tool. As it's combining game capture, desktop capture, window capture, and a never before available (not even on Windows) app capture, I dunno what the right icon would be.

PatTheMav: FWIW Syphon source also has the game capture icon, so I think it works within the icon constraints that we have.

plugins/mac-capture/mac-screen-capture.m
```diff
@@ -0,0 +1,1143 @@
...
+
+	gs_technique_end_pass(tech);
+	gs_technique_end(tech);
+
+	gs_enable_framebuffer_srgb(previous);
+}
+
+static const char *screen_capture_getname(void *unused OBS_UNUSED)
+{
+	return "macOS ScreenCapture";
```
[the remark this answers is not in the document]
PatTheMav: We could call it "ScreenCapture (macOS)"?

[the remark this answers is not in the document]
gxalpha: I wouldn’t stress too much about the capitalization, this is a bug somewhere else in the program, I just wanted to have it noted down somewhere; if we get it localized that’s fine by me.
Though I do think the current name might proof to be not very localizable, as ScreenCapture doesn’t make much sense as a word. Maybe *macOS Screen Capture* or something along those lines would be better.

plugins/mac-capture/mac-screen-capture.m
gxalpha: Syphon having the game capture icon is a relict of its past as a game capture source in combination with syphon inject, and I think it was discussed before that the icon isn’t really optimal there anymore.

As for here, my feeling is that either window or display capture would be better (I’d personally probably go for window capture): If someone does indeed use it for game capturing, looking at the sources list they will likely understand that this can capture their screen and as such their game. On the opposite way, not so much. I imagine someone just looking to record their Safari will be confused if they have to use the source with the game controller icon.

PatTheMav: I've opted for the desktop capture icon for now, because it's based on ScreenCaptureKit and includes desktop capture.

gxalpha: Sounds good, either should be fine

plugins/mac-capture/mac-screen-capture.m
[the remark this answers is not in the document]
gxalpha: This appears to not use the localized string.

gxalpha (approved):
Besides the one open comment, this seems to work well now.
I'm pretty hyped about SCK + CBR, 28 will be a nice update.

PatTheMav:
Lower compatibility level reduced to macOS 12.5, which has been released as of 2022-07-24.

PR #6720, opened 2022-07-12
libobs: Fix crash handler noreturn compiler warning

gxalpha:
### Description

Fixes a `Function declared 'noreturn' should not return` warning in `bcrash()`.
Compiler: Apple clang version 14.0.0 (clang-1400.0.28.1)
Compiler didn't know that `crash_handler()` doesn't return, so it assumed it did, in which case the entire function would return.

### Motivation and Context

Don't like warnings.

### How Has This Been Tested?

Compiled (macOS 13).
One less warning (186->185).

### Types of changes

### Checklist:

### DIFF
--- a/deps/obs-scripting/obslua/obslua.i
+++ b/deps/obs-scripting/obslua/obslua.i
@@ -55,6 +55,7 @@ static inline void wrap_blog(int log_level, const char *message)
 %ignore blog;
 %ignore blogva;
 %ignore bcrash;
+%ignore base_set_crash_handler;
 %ignore obs_source_info;
 %ignore obs_register_source_s(const struct obs_source_info *info, size_t size);
 %ignore obs_output_set_video(obs_output_t *output, video_t *video);
--- a/deps/obs-scripting/obspython/obspython.i
+++ b/deps/obs-scripting/obspython/obspython.i
@@ -64,6 +64,7 @@ static inline void wrap_blog(int log_level, const char *message)
 %ignore blog;
 %ignore blogva;
 %ignore bcrash;
+%ignore base_set_crash_handler;
 %ignore obs_source_info;
 %ignore obs_register_source_s(const struct obs_source_info *info, size_t size);
 %ignore obs_output_set_video(obs_output_t *output, video_t *video);
--- a/libobs/util/base.c
+++ b/libobs/util/base.c
@@ -103,6 +103,7 @@ OBS_NORETURN void bcrash(const char *format, ...)
 	va_start(args, format);
 	crash_handler(format, args, crash_param);
 	va_end(args);
+	exit(0);
 }
 
 void blogva(int log_level, const char *format, va_list args)


tytan652:
Making SWIG avoid functions that use `no_return` and avoid declaring function with `no_return` in the header with MSVC, should work:

```diff
diff --git a/libobs/util/base.c b/libobs/util/base.c
index 0aabf12c1..ea8b609f3 100644
--- a/libobs/util/base.c
+++ b/libobs/util/base.c
@@ -64,7 +64,8 @@ OBS_NORETURN static void def_crash_handler(const char *format, va_list args,
 }
 
 static log_handler_t log_handler = def_log_handler;
-static void (*crash_handler)(const char *, va_list, void *) = def_crash_handler;
+OBS_NORETURN static void (*crash_handler)(const char *, va_list,
+					  void *) = def_crash_handler;
 
 void base_get_log_handler(log_handler_t *handler, void **param)
 {
@@ -83,7 +84,8 @@ void base_set_log_handler(log_handler_t handler, void *param)
 	log_handler = handler;
 }
 
-void base_set_crash_handler(void (*handler)(const char *, va_list, void *),
+void base_set_crash_handler(OBS_NORETURN void (*handler)(const char *, va_list,
+							 void *),
 			    void *param)
 {
 	crash_param = param;
diff --git a/libobs/util/base.h b/libobs/util/base.h
index b6436045e..7ab20357f 100644
--- a/libobs/util/base.h
+++ b/libobs/util/base.h
@@ -72,9 +72,12 @@ typedef void (*log_handler_t)(int lvl, const char *msg, va_list args, void *p);
 EXPORT void base_get_log_handler(log_handler_t *handler, void **param);
 EXPORT void base_set_log_handler(log_handler_t handler, void *param);
 
-EXPORT void base_set_crash_handler(void (*handler)(const char *, va_list,
-						   void *),
+#ifndef SWIG
+EXPORT void base_set_crash_handler(OBS_NORETURN void (*handler)(const char *,
+								va_list,
+								void *),
 				   void *param);
+#endif
 
 EXPORT void blogva(int log_level, const char *format, va_list args);
 
@@ -88,9 +91,11 @@ PRINTFATTR(2, 3)
 EXPORT void blog(int log_level, const char *format, ...);
 PRINTFATTR(1, 2)
 #ifndef SWIG
+#ifndef _MSC_VER
 OBS_NORETURN
 #endif
 EXPORT void bcrash(const char *format, ...);
+#endif
 
 #undef PRINTFATTR
```

tytan652:
Please exclude functions that use `no_return` form SWIG with `#ifndef SWIG`.

gxalpha:
I know, this is the reason why I converted this PR into draft. However, I'm not a fan of just removing the function from swig.

notr1ch:
I like the idea of hiding it from SWIG. If we still need to make the compiler avoid emitting a warning, perhaps an `exit` call could be added to `bcrash` instead.

gxalpha:
@notr1ch does any compiler still complain? I can't see any warnings in the CI log and it seems to pass #6759

norihiro:
I'd like to suggest adding `exit(0);` to the end of `bcrash` to ensure bcrash won't return rather than changing the attribute of the crash handler.

tytan652:
The new fix cause no issue on #6759.

PatTheMav (approved):
Should be fine now, if there are no further objections this should be ready to merge.

norihiro:
It looks ok to me.

RytoEX:
Per @notr1ch, this is good to merge as-is. We should revisit this later to clean up exit status codes and methods.

PR #6723, opened 2022-07-13
Treat warnings as errors on MSVC

jpark37:
### Description
Address the remaining warnings, and enable /WX on both MSVC compiler & linker to prevent more from being added.

### Motivation and Context
I hate warnings.

### How Has This Been Tested?
- Verified the warnings were treated as errors when not fixed.
- Verified the warnings are gone after fixing.
- Verified Debug and RelWithDebInfo compiled on my machine.

### Types of changes

### Checklist:

### DIFF
--- a/UI/CMakeLists.txt
+++ b/UI/CMakeLists.txt
@@ -367,6 +367,7 @@ if(OS_WINDOWS)
                 _CRT_NONSTDC_NO_WARNINGS)
 
   if(MSVC)
+    target_link_options(obs PRIVATE "LINKER:/IGNORE:4099")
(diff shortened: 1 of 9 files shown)


jpark37:
@paulh-aja The DrawTestPattern deprecation isn't trivial to work around, so I'm suppressing in CMake. The inline function in the SDK header itself seems to trigger the warning even if all OBS usages are clean, which they aren't. Just letting you know in case you want to address this for real.

jpark37:
Marking as draft until the obs-browser submodule PR is merged.

PR #6737, opened 2022-07-15
Fix unused-parameter warnings, unnecessary unused attributes

norihiro:
### Description

- Add `OBS_UNUSED` attribute to unused function parameters.
- Remove `UNUSED_PARAMETER` from used parameters
- Remove `OBS_UNUSED` from a used parameter

### Motivation and Context

Don't want to see warnings.
Don't want to see unnecessary `UNUSED_PARAMETER`.

### How Has This Been Tested?

- Reviewed code to ensure the changes are just touching `OBS_UNUSED` and `UNUSED_PARAMETER`.
- Checked warnings built on Linux, macOS, and Windows.

### Types of changes

### Checklist:

### DIFF
--- a/plugins/obs-filters/noise-suppress-filter.c
+++ b/plugins/obs-filters/noise-suppress-filter.c
@@ -334,8 +334,8 @@ failure:
 }
 
 static inline void alloc_channel(struct noise_suppress_data *ng,
-				 uint32_t sample_rate OBS_UNUSED,
-				 size_t channel, size_t frames)
+				 uint32_t sample_rate, size_t channel,
+				 size_t frames)
 {
 #ifdef LIBSPEEXDSP_ENABLED
 	ng->spx_states[channel] =
--- a/UI/window-basic-main-outputs.cpp
+++ b/UI/window-basic-main-outputs.cpp
@@ -97,8 +97,6 @@ static void OBSStopRecording(void *data, calldata_t *params)
 	QMetaObject::invokeMethod(output->main, "RecordingStop",
 				  Q_ARG(int, code),
 				  Q_ARG(QString, arg_last_error));
-
-	UNUSED_PARAMETER(params);
 }
 
 static void OBSRecordStopping(void *data, calldata_t *params)
@@ -143,8 +141,6 @@ static void OBSStopReplayBuffer(void *data, calldata_t *params)
 	os_atomic_set_bool(&replaybuf_active, false);
 	QMetaObject::invokeMethod(output->main, "ReplayBufferStop",
 				  Q_ARG(int, code));
-
-	UNUSED_PARAMETER(params);
 }
 
 static void OBSReplayBufferStopping(void *data, calldata_t *params)
@@ -182,8 +178,6 @@ static void OBSStopVirtualCam(void *data, calldata_t *params)
 	os_atomic_set_bool(&virtualcam_active, false);
 	QMetaObject::invokeMethod(output->main, "OnVirtualCamStop",
 				  Q_ARG(int, code));
-
-	UNUSED_PARAMETER(params);
 }
 
 /* ------------------------------------------------------------------------ */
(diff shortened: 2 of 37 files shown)


plugins/obs-filters/noise-suppress-filter.c
```diff
@@ -340,8 +340,6 @@ static inline void alloc_channel(struct noise_suppress_data *ng,
 #ifdef LIBSPEEXDSP_ENABLED
 	ng->spx_states[channel] =
 		speex_preprocess_state_init((int)frames, sample_rate);
-#else
-	UNUSED_PARAMETER(sample_rate);
```
tytan652, line 344: The related function definition should not have `OBS_UNUSED` since it is used if speexdsp is enabled. So it would be better to remove the `OBS_UNUSED` and keep the `UNUSED_PARAMETER.`

tytan652: Don't forget to mark this "review" as resolved.

jp9000:
You don't need to mark unused parameters in function definitions in C++. You can just remove the parameter's name. (Can't do that in C though, yet)

RytoEX (changes requested):
The change to remove variables in the first commit I would expect to be on its own.

The fourth commit has a change/revert to obs-filters which should be applied to the third commit to revert the change there.

Otherwise, seems fine.

norihiro:
I have updated as Jim and RytoEX commented.
- Unused parameters are removed from `OBSBasic::LoadTransitions`.
- Separate the commit to remove `ms = ticks * AUDIO_OUTPUT_FRAMES * 1000 / sample_rate;` into standalone commit.
- The commits to `plugins/obs-filters/noise-suppress-filter.c` are reordered so that there is no change-then-revert in the series of commits.

UI/window-basic-main-transitions.cpp
```diff
@@ -1770,8 +1770,7 @@ obs_data_array_t *OBSBasic::SaveTransitions()
 	return transitions;
 }
 
-void OBSBasic::LoadTransitions(obs_data_array_t *transitions,
-			       obs_load_source_cb cb, void *private_data)
```
RytoEX, line 1774: I feel like this change is out of scope for this PR, as we're changing a function signature, even though it's private. I'd be fine with this and its related changes as its own PR.

RytoEX (changes requested):
I think I'd prefer the function signature change with `OBSBasic::LoadTransition` to be its own PR.

norihiro:
After rebasing, I saw some more unused parameter warnings. I will update this PR.

norihiro:
Added these changes after Ryan's review comment.
- File `libobs/util/bmem.c` function `base_set_allocator`: Added `OBS_UNUSED`.
- File `plugins/linux-capture/xcomposite-input.c` function `silence_x11_errors`: Added `OBS_UNUSED`.
- File `plugins/vlc-video/vlc-video-source.c` function `init_screencast_capture`: Removed a local variable `aux`, which is not set, not read at all.
- File `plugins/vlc-video/vlc-video-source.c` function `vlcs_video_format`: Removed two local variables `new_width` and `new_height`, which are initialized to 0 but not set, not read.

RytoEX (approved):
CI built on all jobs. Visually, this seems fine to me. This is probably fine to merge. Will circle back around to this to merge if no one raises any concerns.

tytan652:
On macOS (or with Clang), `LoadTransitions` emits unused-parameter warning.

```diff
diff --git a/UI/window-basic-main-transitions.cpp b/UI/window-basic-main-transitions.cpp
index 629506eb557d..de2492f68bad 100644
--- a/UI/window-basic-main-transitions.cpp
+++ b/UI/window-basic-main-transitions.cpp
@@ -1771,7 +1771,8 @@ obs_data_array_t *OBSBasic::SaveTransitions()
 }
 
 void OBSBasic::LoadTransitions(obs_data_array_t *transitions,
-			       obs_load_source_cb cb, void *private_data)
+			       OBS_UNUSED obs_load_source_cb cb,
+			       OBS_UNUSED void *private_data)
 {
 	size_t count = obs_data_array_count(transitions);
 

```

norihiro:
> On macOS (or with Clang), `LoadTransitions` seems to emit unused-parameter warning.

Thank you for reviewing this PR. The other PR (#5148) was just now merged and it fixed `LoadTransitions`.

PR #6112, opened 2022-03-09
Add pause on deactivate and resume on activate for Media Source

dcmouser:
Added option for ffmpeg (video) sources to say they want to be paused on deactivate and resumed on activate, instead of default to stop and restart.

### Description
This patch adds a checkbox option to video sources (after "close file when inactive"), labeled "pause and resume"; when enabled the video source will be PAUSED when the source is deactivated, and then RESUMED when it is later reactivated.  This is in contrast to the other options which allow EITHER that video sources are stopped/unloaded and restarted, or are left playing in the background.

### Motivation and Context
This is based on this request:
https://obsproject.com/forum/threads/request-for-video-media-source-pause-and-unpause-when-showing-hiding.154270/#post-565354
The two motivations are: 1) to allow video sources to be paused when switched away from and then resumed and 2) to reduce cpu use when not showing a scene with a video that you will return to.

### How Has This Been Tested?
Tested on windows OBS version; it's a super simply modification which makes no assumptions about the video sources.

### Types of changes

New feature (non-breaking change which adds functionality)

### Checklist:
- [x] My code has been run through [clang-format](https://github.com/obsproject/obs-studio/blob/master/.clang-format).
- [x] I have read the [**contributing** document](https://github.com/obsproject/obs-studio/blob/master/CONTRIBUTING.rst).
- [ ] My code is not on the master branch.
- [x] The code has been tested.
- [ ] All commit messages are properly formatted and commits squashed where appropriate.
- [ ] I have included updates to all appropriate documentation.

NOTE: This is my first contribution to OBS -- please feel free to modify it as you see fit.

PR #6739, opened 2022-07-15
obs-ffmpeg: Add pause visibility behaviour

mvji:
### Description

Based on https://github.com/obsproject/obs-studio/pull/6112

Add ability to pause local media src when source is not visible, and unpause when not visible
Use combobox like slideshow/vlc source visibility behaviour

Visibility

Use a similar UI code than PR https://github.com/obsproject/obs-studio/pull/6239

### Motivation and Context

Let user choose to pause a media when not visible
Another step to align media source features with other source.

### How Has This Been Tested?

Mac os Intel 11.6.7
- Check that the two mode of Restart playback..., is keep from a previous version of Obs
- Check that modify checkbox "Restart playback...) (in not local file mode), 
or new combobox "Visibility Behaviour" (in local file mode), correctly update UI
- Check that the three behaviour works as expected with a video file (even in loop mode)

### Types of changes

### Checklist:


plugins/obs-ffmpeg/obs-ffmpeg-source.c
```diff
@@ -105,12 +112,70 @@ static bool is_local_file_modified(obs_properties_t *props,
...
+}
+
+static void update_visibility_behaviour_property_list(obs_data_t *settings)
+{
+
+	bool restart_on_activate =
+		obs_data_get_bool(settings, "restart_on_activate");
+	bool pause_on_deactivate =
+		obs_data_get_bool(settings, "pause_on_deactivate");
+	int visibility_behaviour = 2;
```
cg2121, line 148: You should be using enums, instead of numbers, so it is easier to understand when reading the code.

mvji: Last update use enum instead of numbers for the three modes.

dcmouser:
I'm happy to support this modification over my older pull request with pause functionality.

arno01:
cherry-picked against the master and resolved the conflicts, compiled in Ubuntu 22.04 LTS, this feature is working well.

any reason this PR isn't merged yet?

the code I've tested is [here](https://github.com/obsproject/obs-studio/compare/master...arno01:obs-studio:master+pr6739?expand=1)

Fenrirthviti:
This needs testing against audio, video, and audio/video files, as well as remote content to ensure everything works appropriately. It hasn't been merged because the testing isn't complete, and it hasn't been a high priority for us.

PatTheMav:
There is a major issue with this PR (which might not be this PR's fault though):

* The playback of the source does not happen synchronously with UI interaction but on a different thread
* As such when interacting with the media source via the visibility switch, the playback is not immediately affected
* When hiding a source, the media file continues to play for over a second
* During that time a user can switch the source to be visible again, so playback continues
* However this is not always correctly propagated to UI, so e.g. the source toolbar does not show any playback progress anymore (because it's under the impression that the source has stopped)
* Same with pause/unpause: Because the pausing happens after some lag, it's not consistent (playback restarts from a different point in time than when it was paused)

By itself it all works as intended to a degree, but because of the way the media source is designed (and playback is handled) it introduces race conditions and unintuitive program states because of that.

cab4444:
@dcmouser & @mvji, is this something you're still able to work on/resolve issues for? Unless there is another way to do this as of 2026, I just want to have media sources stop(pause) playback when not visible and resume from the same spot without restarting from the beginning when visible again.

PR #6747, opened 2022-07-17
linux-v4l2: Add socket to signal select on shutdown

kkartaltepe:
### Description

Currently we use signals based on os_event's but the udev code blocks in
a 1s select that cannot be signalled without an FD. This adds an extra
unbound socket which will never recieve events but which can be closed
to signal the select.

### Motivation and Context

Whenever udev is enabled it seems like obs hangs on shutdown which isnt a good feeling.

### How Has This Been Tested?

Starting and stopping obs with a udev enabled build and a v4l2 capture in the source list.

### Types of changes

### Checklist:

### DIFF
--- a/plugins/linux-v4l2/v4l2-udev.c
+++ b/plugins/linux-v4l2/v4l2-udev.c
@@ -15,6 +15,9 @@ You should have received a copy of the GNU General Public License
 along with this program.  If not, see <http://www.gnu.org/licenses/>.
 */
 
+#include <sys/eventfd.h>
+#include <poll.h>
+#include <unistd.h>
 #include <libudev.h>
 
 #include <util/threading.h>
@@ -40,6 +43,7 @@ static pthread_mutex_t udev_mutex = PTHREAD_MUTEX_INITIALIZER;
 
 static pthread_t udev_thread;
 static os_event_t *udev_event;
+static int udev_event_fd;
 
 static signal_handler_t *udev_signalhandler = NULL;
 
@@ -108,8 +112,6 @@ static void *udev_event_thread(void *vptr)
 	UNUSED_PARAMETER(vptr);
 
 	int fd;
-	fd_set fds;
-	struct timeval tv;
 	struct udev *udev;
 	struct udev_monitor *mon;
 	struct udev_device *dev;
@@ -127,12 +129,14 @@ static void *udev_event_thread(void *vptr)
 	fd = udev_monitor_get_fd(mon);
 
 	while (os_event_try(udev_event) == EAGAIN) {
-		FD_ZERO(&fds);
-		FD_SET(fd, &fds);
-		tv.tv_sec = 1;
-		tv.tv_usec = 0;
+		struct pollfd fds[2];
 
-		if (select(fd + 1, &fds, NULL, NULL, &tv) <= 0)
+		fds[0].fd = fd;
+		fds[0].events = POLLIN;
+		fds[1].fd = udev_event_fd;
+		fds[1].events = POLLIN;
+
+		if (poll(fds, 2, 1000) <= 0)
 			continue;
 
 		dev = udev_monitor_receive_device(mon);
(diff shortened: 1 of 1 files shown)


kkartaltepe:
Actually ill probably change this, i have not played around with threading at this level much but it may be more appropriate the create an eventfd instead of the unbound socket instead.

kkartaltepe:
Updated with eventfd instead of socket fds.

plugins/linux-v4l2/v4l2-udev.c
```diff
@@ -23,6 +26,10 @@ along with this program.  If not, see <http://www.gnu.org/licenses/>.
 
 #include "v4l2-udev.h"
 
+#ifndef max
+#define max(a, b) ((a) > (b) ? (a) : (b))
+#endif
+
```
GeorgesStavracas, line 32: Seems like this is not used in this file?

plugins/linux-v4l2/v4l2-udev.c
```diff
@@ -126,13 +132,14 @@ static void *udev_event_thread(void *vptr)
...
-
-		if (select(fd + 1, &fds, NULL, NULL, &tv) <= 0)
+		struct pollfd fds[2];
+		fds[0].fd = fd;
+		fds[0].events = POLLIN;
+		fds[1].fd = udev_event_fd;
+		fds[1].events = POLLIN;
+		if (poll(fds, 2, 1000) <= 0) {
 			continue;
+		}
```
GeorgesStavracas, line 142: Are these extra curly braces added by clang-format? If not, they don't seem necessary

plugins/linux-v4l2/v4l2-udev.c
```diff
@@ -157,6 +164,10 @@ void v4l2_init_udev(void)
 	if (udev_refs == 0) {
 		if (os_event_init(&udev_event, OS_EVENT_TYPE_MANUAL) != 0)
 			goto fail;
+		if ((udev_event_fd = eventfd(0, EFD_CLOEXEC)) < 0) {
+			close(udev_event_fd);
```
GeorgesStavracas, line 168: On error, `udev_event_fd` will be -1 and this `close()` call will error out with `EBADF` I think. Seems like you don't need to close this here.

plugins/linux-v4l2/v4l2-udev.c
```diff
@@ -157,6 +164,10 @@ void v4l2_init_udev(void)
 	if (udev_refs == 0) {
 		if (os_event_init(&udev_event, OS_EVENT_TYPE_MANUAL) != 0)
 			goto fail;
+		if ((udev_event_fd = eventfd(0, EFD_CLOEXEC)) < 0) {
```
GeorgesStavracas, line 167: I wonder if `EFD_NONBLOCK` shouldn't be used here too?

plugins/linux-v4l2/v4l2-udev.c
```diff
@@ -126,13 +132,14 @@ static void *udev_event_thread(void *vptr)
...
-		tv.tv_sec = 1;
-		tv.tv_usec = 0;
-
-		if (select(fd + 1, &fds, NULL, NULL, &tv) <= 0)
+		struct pollfd fds[2];
+		fds[0].fd = fd;
+		fds[0].events = POLLIN;
+		fds[1].fd = udev_event_fd;
+		fds[1].events = POLLIN;
+		if (poll(fds, 2, 1000) <= 0) {
```
GeorgesStavracas, line 140: Wouldn't it be better to pass `-1` as the timeout to `poll`? Something like this:

```c
struct pollfd fds[2];

fds[0].fd = fd;
fds[0].events = POLLIN;
fds[1].fd = udev_event_fd;
fds[1].events = POLLIN;

if (poll(&pollfd, 1, -1) <= 0)
	continue;
```

Passing `-1` there signals that we don't care how long it takes, it can block potentially forever.

GeorgesStavracas (changes requested):

plugins/linux-v4l2/v4l2-udev.c
kkartaltepe: Left the close in the wrong branches, that should be fixed up.

plugins/linux-v4l2/v4l2-udev.c
kkartaltepe: in case we screw this up later it  seems prudent to leave the timeout since return of this thread blocks shutdown. The overhead is negligible imho.

plugins/linux-v4l2/v4l2-udev.c
kkartaltepe: I'm not sure it makes sense, we will never be in a blocking case since we only write to the fd once and we want that write to succeed. If it were nonblocking it could fail instead of blocking which would be unfortunate. Maybe its ok to make it nonblocking if we retain the timeout on poll.

GeorgesStavracas (approved):

PR #6750, opened 2022-07-17
UI: Fix undo stack uninitialized ui warning

tytan652:
### Description

Fix this warning by ~~turning~~ moving `undo_s` after `ui` in the class definition. ~~into a pointer and initialized it inside `OBSBasic` constructor~~.

```
obs-studio/UI/window-basic-main.cpp: In constructor ‘OBSBasic::OBSBasic(QWidget*)’:
OBS/obs-studio/UI/window-basic-main.cpp:231:41: warning: member ‘OBSBasic::ui’ is used uninitialized [-Wuninitialized]
  231 |         : OBSMainWindow(parent), undo_s(ui), ui(new Ui::OBSBasic)
      |                                         ^~

```

### Motivation and Context

Just remembered that this warning was there and wanted to get rid of it.

### How Has This Been Tested?

Tried to use Undo/Redo feature and no issue.

### Types of changes

- Code cleanup (non-breaking change which makes code smaller or more readable)

### Checklist:

- [x] My code has been run through [clang-format](https://github.com/obsproject/obs-studio/blob/master/.clang-format).
- [x] I have read the [**contributing** document](https://github.com/obsproject/obs-studio/blob/master/CONTRIBUTING.rst).
- [x] My code is not on the master branch.
- [x] The code has been tested.
- [x] All commit messages are properly formatted and commits squashed where appropriate.
- [x] I have included updates to all appropriate documentation.

PR #6759, opened 2022-07-18
Treat warnings as errors on Clang and GNU GCC

tytan652:
### Description

Depends on:
- #6720
- #6750
- https://github.com/obsproject/obs-browser/pull/368

Once those are merged some commits will be removed.

Turn every warning into errors except:
- deprecations
- jansson (warning are silenced)
- some of obs-outputs because of FTL

All the errors made by this changed are fixed in this PR.

### Motivation and Context

Do the same as #6723.

### How Has This Been Tested?

- Flatpak local build.
- Build on Arch Linux (need some change related to FFmpeg 5).
- And CI from a PR on my fork.

### Types of changes

### Checklist:

### DIFF
--- a/cmake/Modules/CompilerConfig.cmake
+++ b/cmake/Modules/CompilerConfig.cmake
@@ -82,10 +82,14 @@ else()
   endif()
 
   add_compile_options(
+    -Werror
     -Wextra
     -Wvla
+    -Wformat
+    -Wformat-security
     -Wno-unused-function
     -Wno-missing-field-initializers
+    -Wno-error=deprecated-declarations
     -fno-strict-aliasing
     "$<$<COMPILE_LANGUAGE:C>:-Werror-implicit-function-declaration;-Wno-missing-braces>"
     "$<$<BOOL:${USE_LIBCXX}>:-stdlib=libc++>"
--- a/deps/jansson/CMakeLists.txt
+++ b/deps/jansson/CMakeLists.txt
@@ -701,5 +701,11 @@ set(JANSSON_LIBRARIES
     jansson
     CACHE STRING "Jansson libraries")
 
+if (UNIX)
+  target_compile_options(jansson PUBLIC
+    -Wno-format-truncation
+    -Wno-implicit-fallthrough)
+endif()
+
 target_include_directories(jansson PUBLIC src
                                           "${CMAKE_CURRENT_BINARY_DIR}/include")
--- a/libobs/util/base.c
+++ b/libobs/util/base.c
@@ -64,6 +64,9 @@ OBS_NORETURN static void def_crash_handler(const char *format, va_list args,
 }
 
 static log_handler_t log_handler = def_log_handler;
+#if !defined(_MSC_VER) && !defined(SWIG)
+OBS_NORETURN
+#endif
 static void (*crash_handler)(const char *, va_list, void *) = def_crash_handler;
 
 void base_get_log_handler(log_handler_t *handler, void **param)
@@ -83,8 +86,12 @@ void base_set_log_handler(log_handler_t handler, void *param)
 	log_handler = handler;
 }
 
-void base_set_crash_handler(void (*handler)(const char *, va_list, void *),
-			    void *param)
+void base_set_crash_handler(
+#if !defined(_MSC_VER) && !defined(SWIG)
+	OBS_NORETURN
+#endif
+	void (*handler)(const char *, va_list, void *),
+	void *param)
 {
 	crash_param = param;
 	crash_handler = handler;
--- a/plugins/obs-outputs/CMakeLists.txt
+++ b/plugins/obs-outputs/CMakeLists.txt
@@ -178,6 +178,22 @@ elseif(EXISTS "${CMAKE_CURRENT_SOURCE_DIR}/ftl-sdk/CMakeLists.txt")
       obs-outputs PRIVATE ${CMAKE_CURRENT_SOURCE_DIR}/ftl-sdk/libftl/posix)
   endif()
 
+  if(NOT OS_WINDOWS)
+    target_compile_options(
+      obs-outputs
+      PRIVATE
+        -Wno-error=extra
+        -Wno-error=sign-compare
+        -Wno-error=incompatible-pointer-types
+        -Wno-error=int-conversion
+        "$<$<COMPILE_LANG_AND_ID:C,GNU>:-Wno-error=maybe-uninitialized>"
+        "$<$<COMPILE_LANG_AND_ID:C,AppleClang,Clang>:-Wno-error=pointer-sign>")
+    if((NOT CMAKE_C_COMPILER_ID STREQUAL "GNU") OR CMAKE_C_COMPILER_VERSION
+                                                   VERSION_GREATER_EQUAL "10")
+      target_compile_options(obs-outputs PRIVATE -Wno-error=enum-conversion)
+    endif()
+  endif()
+
   target_compile_definitions(obs-outputs PRIVATE FTL_FOUND)
 endif()
 
--- a/.gitmodules
+++ b/.gitmodules
@@ -9,7 +9,7 @@
 	url = https://github.com/obsproject/obs-amd-encoder.git
 [submodule "plugins/obs-browser"]
 	path = plugins/obs-browser
-	url = https://github.com/obsproject/obs-browser.git
+	url = https://github.com/tytan652/obs-browser.git
 [submodule "plugins/obs-vst"]
 	path = plugins/obs-vst
 	url = https://github.com/obsproject/obs-vst.git
--- a/UI/window-basic-main.hpp
+++ b/UI/window-basic-main.hpp
@@ -818,7 +818,6 @@ private:
 	OBSSource prevFTBSource = nullptr;
 
 public:
-	undo_stack undo_s;
 	OBSSource GetProgramSource();
 	OBSScene GetCurrentScene();
 
@@ -1150,7 +1149,12 @@ public slots:
 	void UpdateContextBarDeferred(bool force = false);
 	void UpdateContextBarVisibility();
 
+private:
+	std::unique_ptr<Ui::OBSBasic> ui;
+
 public:
+	undo_stack undo_s;
+
 	explicit OBSBasic(QWidget *parent = 0);
 	virtual ~OBSBasic();
 
@@ -1162,9 +1166,6 @@ public:
 				   const char *file) const override;
 
 	static void InitBrowserPanelSafeBlock();
-
-private:
-	std::unique_ptr<Ui::OBSBasic> ui;
 };
 
 class SceneRenameDelegate : public QStyledItemDelegate {
--- a/deps/obs-scripting/obslua/CMakeLists.txt
+++ b/deps/obs-scripting/obslua/CMakeLists.txt
@@ -61,4 +61,10 @@ elseif(OS_MACOS)
                                           MACHO_COMPATIBILITY_VERSION 0)
 endif()
 
+if(NOT OS_WINDOWS
+   AND CMAKE_C_COMPILER_ID STREQUAL "GNU"
+   AND SWIG_VERSION VERSION_LESS "4.1")
+  target_compile_options(obslua PRIVATE -Wno-maybe-uninitialized)
+endif()
+
 setup_script_plugin_target(obslua)
--- a/deps/obs-scripting/obslua/obslua.i
+++ b/deps/obs-scripting/obslua/obslua.i
@@ -55,6 +55,7 @@ static inline void wrap_blog(int log_level, const char *message)
 %ignore blog;
 %ignore blogva;
 %ignore bcrash;
+%ignore base_set_crash_handler;
 %ignore obs_source_info;
 %ignore obs_register_source_s(const struct obs_source_info *info, size_t size);
 %ignore obs_output_set_video(obs_output_t *output, video_t *video);
--- a/deps/obs-scripting/obspython/obspython.i
+++ b/deps/obs-scripting/obspython/obspython.i
@@ -64,6 +64,7 @@ static inline void wrap_blog(int log_level, const char *message)
 %ignore blog;
 %ignore blogva;
 %ignore bcrash;
+%ignore base_set_crash_handler;
 %ignore obs_source_info;
 %ignore obs_register_source_s(const struct obs_source_info *info, size_t size);
 %ignore obs_output_set_video(obs_output_t *output, video_t *video);
--- a/libobs/media-io/media-remux.c
+++ b/libobs/media-io/media-remux.c
@@ -34,6 +34,10 @@
 #define CODEC_FLAG_GLOBAL_H CODEC_FLAG_GLOBAL_HEADER
 #endif
 
+#ifndef FF_API_BUFFER_SIZE_T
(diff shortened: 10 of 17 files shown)


gxalpha (changes requested):
[-Wshorten-64-to-32](https://releases.llvm.org/13.0.0/tools/clang/docs/DiagnosticsReference.html#id721) needs to be treated as a warning, or maybe even suppressed - it's enabled by default on Apple Clang 14 and throws tons of warnings, which are now errors and prevent building.

Also a bit of nitpicking, I feel like commit messages could be improved. The `Fix compilation errors` are only errors after warnings are turned into errors - it seems more correct to me to have the "Treat warnings as errors" at the end and then have other fixes as `Fix compilation warnings`.
First (or last, depending on where you put it in the end) commit needs `cmake:` prefix.

cmake/Modules/CompilerConfig.cmake
```diff
@@ -84,16 +84,21 @@ else()
...
 
   add_compile_options(
+    -Werror
     -Wextra
     -Wvla
+    -Wformat
+    -Wformat-security
     -Wno-unused-function
     -Wno-missing-field-initializers
+    -Wno-error=deprecated-declarations
```
PatTheMav, line 94: I'd rather add this to projects that we _know_ use deprecated technology (e.g. libobs-opengl) and still keep this as an error otherwise.

We need to ensure that new code doesn't use deprecated APIs without anyone noticing.

libobs/util/base.c
```diff
@@ -64,6 +64,9 @@ OBS_NORETURN static void def_crash_handler(const char *format, va_list args,
 }
 
 static log_handler_t log_handler = def_log_handler;
+#if !defined(_MSC_VER) && !defined(SWIG)
+OBS_NORETURN
```
PatTheMav, line 68: @notr1ch Correct me if I'm wrong, but I guess this is not good, because this would disable the NORETURN attribute when compiling with MSVC.

plugins/obs-outputs/CMakeLists.txt
```diff
@@ -179,6 +179,22 @@ elseif(EXISTS "${CMAKE_CURRENT_SOURCE_DIR}/ftl-sdk/CMakeLists.txt")
       obs-outputs PRIVATE ${CMAKE_CURRENT_SOURCE_DIR}/ftl-sdk/libftl/posix)
   endif()
 
+  if(NOT OS_WINDOWS)
```
PatTheMav, line 182: This sounds like a mouthful that should be fixed in code and not just ignored.

tytan652: I don't think someones want to fork ftl-sdk and fix all this.

PatTheMav: Ah, I didn't notice the indent, this is for FTL.. Bummer.

PatTheMav:
Apart from open inline comments I'm fine with this.

plugins/mac-syphon/CMakeLists.txt
```diff
@@ -67,7 +67,7 @@ target_compile_options(
   mac-syphon
   PRIVATE -include
           ${CMAKE_CURRENT_SOURCE_DIR}/syphon-framework/Syphon_Prefix.pch
-          -fobjc-arc)
+          -fobjc-arc -Wno-deprecated-declarations)
```
gxalpha, line 70: I don't think we should just ignore all deprecations here but rather fix them or ignore them individually.
I addressed the one I got in #6767, do you have any others?

tytan652: I don't, CI will tell.

cmake/Modules/CompilerConfig.cmake
tytan652: I remembered why I put it there because of Qt6 deprecation warnings, actually testing to apply the flag to each "library/binary" that use Qt.

tytan652:
Changed some CMake conditions to apply Clang/GCC flags with `NOT MSVC` rather than `NOT OS_WINDOWS` or `OS_POSIX` or even `UNIX` for jansson.

tytan652:
Since some between generators there is different enablement (e.g [here between Ninja and Xcode](https://github.com/obsproject/obs-studio/pull/6761#discussion_r924326213)) I ensured that null conversion warning on Clang and GCC is enabled.

tytan652:
#6060 added a new shadow-ivar warning.

PatTheMav:
> local declaration of 'icon' hides instance variable

Yeah needs to be fixed in code, ObjC convention is to prefix private variables with an underscore:

```
@private
	QIcon _icon;
}
@end

[...]
	self->_icon = icon;
```

tytan652:
> Yeah needs to be fixed in code, ObjC convention is to prefix private variables with an underscore

Updated the change with this convention in mind.

tytan652:
#6802 added incompatible-pointer-types warnings.

```
/obs-studio/plugins/rtmp-services/rtmp-common.c:863:31: error: returning ‘char **’ from a function with incompatible return type ‘const char **’ [-Werror=incompatible-pointer-types]
  863 |                 return service->video_codecs;
      |                        ~~~~~~~^~~~~~~~~~~~~~
/obs-studio/plugins/rtmp-services/rtmp-common.c:898:23: error: returning ‘char **’ from a function with incompatible return type ‘const char **’ [-Werror=incompatible-pointer-types]
  898 |         return service->video_codecs;
      |                ~~~~~~~^~~~~~~~~~~~~~
```

tytan652:
Discovered that Xcode 14 deprecates sprintf(), so I placed pragmas to turn related errors back to warnings.

tytan652:
Waiting for a PR to replace sprintf() which is deprecated on macOS with Xcode.

tytan652:
Try to build on Linux with Clang and found that -Wswitch was not enabled on GCC by default, so I added it and fix the resulting warnings.

tytan652:
#6134 added warnings.

Note: ftime() is deprecated with newer glibc, this will may need to be fixed for the next Ubuntu big release.

cmake/Modules/CompilerConfig.cmake
```diff
@@ -84,16 +84,27 @@ else()
   endif()
 
   add_compile_options(
+    -Werror
```
theHamsta, line 87: `-Werror` is a setting for CI where known compilers are used. Enabling it in user space just make just the source code incompatible with future or unknown compilers (very annoying bisecting code and returning to old revision with already fixed warnings) and you always have to disable the flag when experimenting (unfinished features will likely generate warnings). Any reasons why not just enable this on CI to forbid code that generates warning to be submitted? I get the point that some annoyed users might be more likely to submit PRs for a warning when they use a untested compiler, but wouldn't it still be better to just try to cover most compilers in CI?

tytan652:
#6828 add unused parameter warnings and #6838 add switch warnings.

cmake/Modules/CompilerConfig.cmake
jp9000: We could add an cmake option to optionally disable it locally, although after using it recently, I've found that I've managed to get used to it while programming something. It's mostly only been a very minor annoyance if anything, and we disable the warnings that we don't feel make sense for us. I know that future compiler versions may be unpredictable with it, but I generally don't think it's a big deal either. But this is just my opinion on the matter.

theHamsta: I would appreciate a option to disable them, but temporarily disabling them locally wouldn't be a big deal for me. I normally use nightly compilers and prefer to have the warnings on a maximum which I might ignore for a test run but when making a commit.

tytan652:
Changes related to switches  were split to #6960.

tytan652:
#6577 added a format warning because of a wrong format specifier

tytan652:
Worked on fixing issues by compiling OBS and aarch64 Linux.

tytan652:
Rebase and updated how deprecation warning in libobs-opengl on macOS are managed since #7376.

tytan652:
Modified the PR to no longer required #6960.

tytan652:
Rebase, will fail because of https://github.com/obsproject/obs-studio/blob/fcb6df1f636daa9552a8f61ffc78bbf1bc26881d/plugins/obs-ffmpeg/obs-ffmpeg-srt.h#L90

cmake/Modules/CompilerConfig.cmake
```diff
@@ -90,16 +90,36 @@ else()
...
+    "$<$<COMPILE_LANG_AND_ID:CXX,AppleClang,Clang>:-Wnull-conversion;-fcolor-diagnostics;-Wno-error=shorten-64-to-32>"
+    "$<$<COMPILE_LANG_AND_ID:C,AppleClang,Clang>:-Wnull-conversion;-fcolor-diagnostics;-Wno-error=shorten-64-to-32>"
+    "$<$<COMPILE_LANG_AND_ID:CXX,GNU>:-Wconversion-null>")
+
+  if(CMAKE_CXX_COMPILER_ID STREQUAL "GNU")
+    # GCC 12.1.0 has a regression bug which trigger maybe-uninitialized warnings
+    # where there is not. (https://gcc.gnu.org/bugzilla/show_bug.cgi?id=105562)
+    if(CMAKE_CXX_COMPILER_VERSION VERSION_EQUAL "12.1.0")
+      add_compile_options(-Wno-error=maybe-uninitialized)
+    endif()
```
gxalpha, line 116: Is this still needed with 12.2.0 being out for a while now?

libobs/util/c99defs.h
```diff
@@ -51,6 +51,31 @@
...
+#define PRAGMA_WARN_DEPRECATION \
+	_Pragma("GCC diagnostic warning \"-Wdeprecated-declarations\"")
+#define PRAGMA_WARN_STRINGOP_OVERFLOW \
+	_Pragma("GCC diagnostic warning \"-Wstringop-overflow\"")
+#else
+#define PRAGMA_WARN_PUSH
+#define PRAGMA_WARN_POP
+#define PRAGMA_WARN_DEPRECATION
+#define PRAGMA_WARN_STRINGOP_OVERFLOW
+#endif
```
gxalpha, line 77: Should these maybe have an `OBS_` prefix?

gxalpha (commented):
```
libobs,libobs-opengl,obs-ffmpeg-mux: Keep some deprecation as warning…
…s on Clang and GCC
```
needs a shorter commit title (<=72 characters).

Also, do you think it would be possible to squash the commits a bit more? Going through over 20 commits individually (some of which modify changes of others, making GitHub place inline comments on wrong locations or just disappear in the full view which happened to me multiple times now, thanks GitHub!) is a bit tedious :D For example, there are multiple commits changing scripting, filters, c99defs, etc.

cmake/Modules/CompilerConfig.cmake
tytan652: Removed.

tytan652:
Did my best for making the commit name shorter.

cg2121:
Reminder that there are merge conflicts that need to be resolved.

cmake/Modules/CompilerConfig.cmake
tytan652: Will be re-added require for Flatpak.

tytan652: They appear to use 12.1.0 on 22.08.

deps/obs-scripting/obs-scripting-python.c
```diff
@@ -1686,15 +1686,15 @@ bool obs_scripting_load_python(const char *python_path)
...
-#elif PY_VERSION_HEX < 0x030700b0
-	if (true) {
-#else
-	if (false) {
-#endif
 		PyEval_InitThreads();
 		if (!PyEval_ThreadsInitialized())
 			return false;
 	}
+#elif PY_VERSION_HEX < 0x03070000
```
RytoEX, line 1693: Is there a particular reason why this version number changed from `0x030700b0` to `0x03070000`?

RytoEX (commented):
A single question on this.

deps/obs-scripting/obs-scripting-python.c
tytan652: #5653 used `0x03070000`: https://github.com/obsproject/obs-studio/pull/5653/files#diff-07e241c573bec44c5491a58f0e65cf92d43af09a1306a864cb29240980f55759

But #6915 changed it for no given reason: https://github.com/obsproject/obs-studio/pull/6915/files#diff-07e241c573bec44c5491a58f0e65cf92d43af09a1306a864cb29240980f55759 maybe a typo 

And Python docs says (https://docs.python.org/3/c-api/init.html#c.PyEval_InitThreads):
> Changed in version 3.7: This function is now called by Py_Initialize(), so you don’t have to call it yourself anymore.

> Changed in version 3.7: The GIL is now initialized by Py_Initialize().

I will ask Pat off-thread.

tytan652: I look and this `b` just mean beta, and I see no reason to be so specific.

deps/jansson/CMakeLists.txt
```diff
@@ -701,5 +701,11 @@ set(JANSSON_LIBRARIES
     jansson
     CACHE STRING "Jansson libraries")
 
+if (NOT MSVC)
```
PatTheMav, line 704: @RytoEX aren't we using pre-compiled Jansson on Windows already?

deps/obs-scripting/obs-scripting-python.c
PatTheMav: In theory this needs to encompass all versions that contain the breaking change (and that includes the beta version). The real-life impact will probably small (only users of Python 3.7.0 Beta are affected), but semantically speaking _including_ the beta version is more correct.

deps/jansson/CMakeLists.txt
tytan652: Yes, but I keep the change except if jansson get removed form the repo.

tytan652: Also maybe PPA build use it, I didn't check.

deps/obs-scripting/obs-scripting-python.c
tytan652: `0x03070000` is less than `0x030700b0`, so the beta is included.

Note: `f` is final release.

PatTheMav:
Overall seems fine to me - will we merge the separate PRs instead of this "meta" PR? Because as-is the changes lack context (no code comments referencing the warning that is silenced), while separate commits would give some needed context.

libobs/util/c99defs.h
PatTheMav: I think it's better this way as they're closer to the standard `#pragma` syntax.

tytan652:
> Overall seems fine to me - will we merge the separate PRs instead of this "meta" PR?

No, this is one PR, you should have asked that really sooner.

> Because as-is the changes lack context (no code comments referencing the warning that is silenced), while separate commits would give some needed context.

21 commits with descriptions when needed. And you say there is no context.

deps/obs-scripting/obs-scripting-python.c
PatTheMav: The only requirement here is that the comparison needs to hit any version up to the one that still required manual initialisation of threads. If this is true for all 3.7 versions (including alphas) then `0x03070000` is fine.

tytan652: Alpha use `a` so less than beta.

PR #6769, opened 2022-07-19
UI: Add low latency audio buffering mode

jp9000:
### Description

This feature adds a checkbox option in the audio section of settings which is meant to reduce maximum audio buffering and turn off dynamic buffering mode. This allows the lowest possible consistent latency for audio buffering, which is useful for the Decklink and NDI outputs which cannot rely on audio timestamps for synchronization.

This can have a negative effect of making audio segments (partial or in full) cut out. So audio glitching or audio loss can occur if this is enabled.

The downside of this feature is that audio segments from sources that require more buffering can prematurely get cut off or discarded, so this feature should be used with caution. Warning messages have been put in place to ensure that users do not necessarily enable this without knowing the potential consequences.

### Motivation and Context

When either the Decklink output or NDI plugin output is used, the audio buffering which OBS does can introduce audio/video desynchronization on the other end due to the fact that they do not necessarily rely on timestamps. To work around this, this low latency audio buffering feature prevents audio buffering from getting higher than around 20 milliseconds, which helps minimize desynchronization due to buffering.

### How Has This Been Tested?

Fairly simple to test, just adds a setting to the audio settings. Seems to function correctly. Now as for having low latency buffering enabled, that's another story which will probably need more testing by users.

### Types of changes

### Checklist:


UI/data/locale/en-US.ini
```diff
@@ -1044,6 +1044,11 @@ Basic.Settings.Audio.EnablePushToTalk="Enable Push-to-talk"
 Basic.Settings.Audio.PushToTalkDelay="Push-to-talk delay"
 Basic.Settings.Audio.UnknownAudioDevice="[Device not connected or not available]"
 Basic.Settings.Audio.Disabled="Disabled"
+Basic.Settings.Audio.LowLatencyBufferingMode="Low Latency Audio Buffering Mode (For Decklink/NDI outputs)"
+Basic.Settings.Audio.LowLatencyBufferingWarning.Enabled="WARNING: Low latency audio buffering is enabled.""
```
Vainock, line 1048: This string ends with 2 quotation marks.

Vainock (commented):
Found a small issue with a string.

LGCW:
On Debian 11 this causes a segmentation fault when opening Settings if OBS is compiled with -DENABLE_PULSEAUDIO=OFF. 

When PulseAudio is enabled the settings menu opens and the option for low latency audio is displayed.

PR #6778, opened 2022-07-21
Allow HEVC on streaming only if the service is compatible

tytan652:
### Description

While the user change his streaming service, if an incompatible encoder was set the user will got this warning explaining that the encoder was changed because of the incompatibility.
HEVC_Warning

The message differ if the encoder was a simple preset or does not have a H264 alternative.

### Motivation and Context

When HEVC is enabled, HEVC encoder can be used with any service, this PR try to avoid this.
It also add HEVC as supported codec for mpeg-ts based protocols (HLS, RIST, SRT)

Also clear the issue I have with #6617.

This might be refactored afterward because of https://github.com/obsproject/rfcs/pull/45 and https://github.com/obsproject/rfcs/pull/39.

### How Has This Been Tested?

- Set a service compatible with HEVC
- Set an HEVC encoder (they are not shown if the service is not compatible)
- Change the service for one that does not support HEVC and a warning should show.

### Types of changes

### Checklist:

### DIFF
--- a/UI/data/locale/en-US.ini
+++ b/UI/data/locale/en-US.ini
@@ -867,6 +867,10 @@ Basic.Settings.Stream.Recommended.MaxVideoBitrate="Maximum Video Bitrate: %1 kbp
 Basic.Settings.Stream.Recommended.MaxAudioBitrate="Maximum Audio Bitrate: %1 kbps"
 Basic.Settings.Stream.Recommended.MaxResolution="Maximum Resolution: %1"
 Basic.Settings.Stream.Recommended.MaxFPS="Maximum FPS: %1"
+Basic.Settings.Stream.ChangeIncompatibleEncoder="The selected encoder is not compatible with the service.\nThe encoder in the output page was changed.\n\nCheck your encoder settings."
+Basic.Settings.Stream.ChangeIncompatibleEncoder.Title="Incompatible encoder changed"
+Basic.Settings.Stream.ChangeIncompatibleEncoder.Simple="The selected encoder is not compatible with the service.\nThe encoder in the output page was changed to its H.264 version."
+Basic.Settings.Stream.ChangeIncompatibleEncoder.ReplaceWithH264="The selected encoder is not compatible with the service.\nThe encoder in the output page was changed to its H.264 version.\n\nCheck your encoder settings."
 
 # basic mode 'output' settings
 Basic.Settings.Output="Output"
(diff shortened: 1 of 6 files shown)


tytan652:
Stupid me has forgot to stage some changes.

SuslikV:
In my opinion, if this kind of warning/forcing/restriction will be implemented. it should use already available control: _Ignore streaming service setting recommendations_

tytan652:
> In my opinion, if this kind of warning/forcing/restriction will be implemented. it should use already available control: _Ignore streaming service setting recommendations_

Not for codecs, codecs are not a recommendation.

tytan652:
Sidenote: I could improve this by adding codecs in the Service API, but since I made this https://github.com/obsproject/rfcs/pull/45, I don't this PR to be taken for a part of the implementation of this RFC.

lextra2:
IMHO

```
The selected video encoder is not compatible with the streaming service.
Check your output settings.
```

sounds better.

UI/data/locale/en-US.ini
```diff
@@ -867,6 +867,10 @@ Basic.Settings.Stream.Recommended.MaxVideoBitrate="Maximum Video Bitrate: %1 kbp
 Basic.Settings.Stream.Recommended.MaxAudioBitrate="Maximum Audio Bitrate: %1 kbps"
 Basic.Settings.Stream.Recommended.MaxResolution="Maximum Resolution: %1"
 Basic.Settings.Stream.Recommended.MaxFPS="Maximum FPS: %1"
+Basic.Settings.Stream.ChangeIncompatibleEncoder.Title="Incompatitle encoder changed"
```
RytoEX, line 870: Incompatitle -> Incompatible

UI/data/locale/en-US.ini
```diff
@@ -867,6 +867,10 @@ Basic.Settings.Stream.Recommended.MaxVideoBitrate="Maximum Video Bitrate: %1 kbp
...
+Basic.Settings.Stream.ChangeIncompatibleEncoder.Simple="The actually selected encoder is not compatible with the actual service.\nThe encoder in the output page was changed by its H264 version."
```
RytoEX, line 871: Not sure about "The actually selected encoder". Is this distinct somehow from "The selected encoder"? Same with "the actual service" vs. "the service".

RytoEX, line 871: "changed by its H264 version." -> "changed to its H264 version." ?

RytoEX, line 871: "H264" -> "H.264"

RytoEX (changes requested):
Just a quick visual review. Found some string nits.

RytoEX (changes requested):
String nit

SuslikV:
Closed in favor of
https://github.com/obsproject/obs-studio/commit/48819def6df056f5aee58bebb05a64ca308aa378

PR #6782, opened 2022-07-21
Refactor find_qt CMake macro to select Qt version automatically if not specified

tytan652:
### Description

Depends on:
- https://github.com/obsproject/obs-vst/pull/109
- https://github.com/obsproject/obs-browser/pull/372

This refactor allow to use automatically the available Qt lib (5 or 6). If both are present, 5 is the default.

Also remove usage of versioned Qt libs name usually with GuiPrivate, the macro add the unversioned target.

If QT_VERSION is not defined, it gets define with the AUTO value. And definition is moved to the helper file.

find_qt now:
- Check quietly for Qt5 and Qt6
- If QT_VERSION is set to AUTO, it checks firstly if Qt5 was found and it use it. If not it do the same for Qt6
- If QT_VERSION is set to 5 or 6, it checks if the choice was found and use it. And if not, it falls back to the other if found.
- If neither Qt5 or Qt6 are found, a fatal error is emitted.
- The macro saved the _QT_VERSION in the cache and used when QT_VERSION was used, so the process is not repeated each time that find_qt is used.

### Motivation and Context

Allow to automatically use whether Qt5 or Qt6 is available and if both are present prefer Qt5.

### How Has This Been Tested?

- I uninstalled Qt6 from my system and set `QT_VERSION` to 6 and falls back to Qt5. If both installed Qt5 is selected.

- By building the Flatpak with the KDE 6.3 Runtime, Qt6 is selected automatically. Before this PR, `-DQT_VERSION=6` was required to make it build.

### Types of changes

### Checklist:

### DIFF
--- a/cmake/Modules/ObsHelpers.cmake
+++ b/cmake/Modules/ObsHelpers.cmake
@@ -337,38 +337,98 @@ function(define_graphic_modules target)
   endforeach()
 endfunction()
 
+if(NOT QT_VERSION)
+  set(QT_VERSION
+      AUTO
+      CACHE STRING "OBS Qt version [AUTO, 5, 6]" FORCE)
+  set_property(CACHE QT_VERSION PROPERTY STRINGS AUTO 5 6)
+endif()
+
 macro(find_qt)
-  set(oneValueArgs VERSION)
   set(multiValueArgs COMPONENTS COMPONENTS_WIN COMPONENTS_MAC COMPONENTS_LINUX)
   cmake_parse_arguments(FIND_QT "" "${oneValueArgs}" "${multiValueArgs}"
                         ${ARGN})
+  set(QT_NO_CREATE_VERSIONLESS_TARGETS ON)
+  find_package(
+    Qt5
+    COMPONENTS Core
+    QUIET)
+  find_package(
+    Qt6
+    COMPONENTS Core
+    QUIET)
+
+  if(NOT _QT_VERSION AND QT_VERSION STREQUAL AUTO)
+    if(TARGET Qt5::Core)
+      set(_QT_VERSION
+          5
+          CACHE INTERNAL "")
+    elseif(TARGET Qt6::Core)
+      set(_QT_VERSION
+          6
+          CACHE INTERNAL "")
+    endif()
+  elseif(NOT _QT_VERSION)
+    if(TARGET Qt${QT_VERSION}::Core)
+      set(_QT_VERSION
+          ${QT_VERSION}
+          CACHE INTERNAL "")
+    else()
+      if(QT_VERSION EQUAL 6)
+        set(FALLBACK_QT_VERSION 5)
+      else()
+        set(FALLBACK_QT_VERSION 6)
+      endif()
+      message(
+        WARNING
+          "Qt${QT_VERSION} was not found, falling back to Qt${FALLBACK_QT_VERSION}"
+      )
+
+      if(TARGET Qt${FALLBACK_QT_VERSION}::Core)
+        set(_QT_VERSION
+            ${FALLBACK_QT_VERSION}
+            CACHE INTERNAL "")
+      endif()
+    endif()
+  endif()
+
+  set(QT_NO_CREATE_VERSIONLESS_TARGETS OFF)
+
+  if(NOT _QT_VERSION)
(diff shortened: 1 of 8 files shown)


cmake/Modules/ObsHelpers.cmake
```diff
@@ -337,38 +337,96 @@ function(define_graphic_modules target)
   endforeach()
 endfunction()
 
+if(NOT QT_VERSION)
```
PatTheMav, line 340: I'm not sure if this check is correct (given CMake's stringly typed nature).

In _theory_ `if(NOT ..)` checks for the boolean value of a variable. _If_ Qt considers a non-defined variable `false`, this would work. Otherwise (and to be specific) this would need to say:

`if(NOT DEFINED QT_VERSION OR QT_VERSION STREQUAL "")` which ensures that you only `FORCE` the cache variable if it is not defined at all or contains an empty string.

PatTheMav: > `if(<constant>)`
> True if the constant is 1, ON, YES, TRUE, Y, or a non-zero number (including floating point numbers). False if the constant is 0, OFF, NO, FALSE, N, IGNORE, NOTFOUND, the empty string, or ends in the suffix -NOTFOUND. 

Script languages, you gotta love 'em. So I guess the code _works_ fine.

cmake/Modules/ObsHelpers.cmake
```diff
@@ -337,38 +337,96 @@ function(define_graphic_modules target)
...
+    Qt5
+    COMPONENTS Core
+    QUIET)
+  find_package(
+    Qt6
+    COMPONENTS Core
+    QUIET)
+
+  if(NOT _QT_VERSION AND QT_VERSION STREQUAL AUTO)
+    if(Qt5_FOUND)
```
PatTheMav, line 362: I'd rather explicitly check for the target: `if(TARGET Qt5::Core)`.

cmake/Modules/ObsHelpers.cmake
tytan652: The next push will have `NOT DEFINED` for each `NOT QT_VERSION` or `NOT _QT_VERSION`. And also `QT_VERSION STREQUAL ""` will  be added.

cmake/Modules/ObsHelpers.cmake
tytan652: Changed on the next push.

cmake/Modules/ObsHelpers.cmake
tytan652: So I will keep it like that.

tytan652:
Found an issue with where Qt6 Svg was unable to find `Qt::Core`. I fixed by adding Core to the components list to create the missing target.

cmake/Modules/ObsHelpers.cmake
```diff
@@ -337,38 +337,98 @@ function(define_graphic_modules target)
...
+      set(_QT_VERSION
+          5
+          CACHE INTERNAL "")
+    elseif(TARGET Qt6::Core)
+      set(_QT_VERSION
+          6
+          CACHE INTERNAL "")
+    endif()
+  elseif(NOT _QT_VERSION)
+    if(Qt${QT_VERSION}_FOUND)
```
PatTheMav, line 372: See above, you'd need to use `if(TARGET Qt${QT_VERSION}::Core)` here and below.

tytan652: Forgot that one.

PatTheMav:
> Found an issue with where Qt6 Svg was unable to find `Qt::Core`. I fixed by adding Core to the components list to create the missing target.

Interesting effect, I guess because there is no versionless Core target set anymore, but we enable versionless targets to import all the other components, it will not be able to find core.

Another way to fix this would be to add the `Qt::Core` target as an imported lib after you found the correct version, but before finding the other components. Either way should be fine.

tytan652:
> Another way to fix this would be to add the `Qt::Core` target as an imported lib after you found the correct version, but before finding the other components. Either way should be fine.

My fix seems to work even with all plugins using `find_qt` disabled. So I keep it like that.

PatTheMav (approved):
Went over the recent changes again, I think all my suggestions were taken into account.

PR #6785, opened 2022-07-22
obs-outputs: Fix TLS_client init for mbedTLS 3.1.0+

RytoEX:
### Description

When trying to update to mbedTLS 3.2.1, I ran into failures with RTMPS output. After consulting an mbedTLS implementation example, I determined that we were not setting up the SSL/TLS context config in the correct order, causing the connection to fail. Performing the setup in the recommended order fixes that.

### Motivation and Context

Want RTMPS to work with mbedTLS 3.1.0+.

### How Has This Been Tested?

I tested an RTMPS stream to YouTube before and after this change. Before it would fail to connect. After, the stream worked fine. Confirmed that this works with both mbedTLS 2.24.0 and 3.2.1 on Windows.

### Types of changes

### Checklist:


notr1ch (approved):
Looks good as far as I can tell given the awful state of mbedTLS docs.

PR #6786, opened 2022-07-22
UI: Disable default path setting when file path is selected

### LINKED ISSUE #6777: OBS not saving last path when browsing for LUT filter
### Operating System Info

Windows 10

### Other OS

_No response_

### OBS Studio Version

27.2.4

_No response_

### OBS Studio Log URL

https://obsproject.com/logs/qYRCAtNpCPG4fZOM

### OBS Studio Crash Log URL

_No response_

### Expected Behavior

I expect that OBS saves the last folder I was viewing when I selected the filter I want to use, so I can try out multiple filters

### Current Behavior

When selecting a LUT filter for a source, it does not save the last path you were browsing at that moment, but it saves when you click "Close" and open the window up again.

### Steps to Reproduce

1. Add source like a webcam or something
2. Open Filters
3. Apply LUT Filter
4. Browse for a filter, and go to a folder that is not the default obs-filters folder
5. Select one
6. Browse for a filter again
7. The last path you were in did not save.

Now do this again but instead of browsing for a filter again, close the filters window, open it again then browse for a filter, it will then have your last path saved, but this is not the wanted and convenient behavior.
...

_No response_

tuduweb:
### Description

Add path value check (whether empty) when open path/file selecting dialog, to achive only using `default_path` when path value is empty.

### Motivation and Context

Fixes #6777
In most cases, the `default_path` setting should only work when the path is empty, but in current situation, it looks not work well when the path is changed to other directory or is set from empty. Beacuse the `properties view` (default_path argument) is not updated after changing the path.

https://github.com/obsproject/obs-studio/blob/e3f416f3fc72db49d1ba71b8423454d053a9d6bf/plugins/obs-filters/color-grade-filter.c#L392-L399

### How Has This Been Tested?

in Ubuntu20
reproduce the steps mentioned in issue, it works.
and other behaviors link to this `WidgetInfo::PathChanged` function works fine as usual.

### Types of changes

### Checklist:

### DIFF
--- a/UI/properties-view.cpp
+++ b/UI/properties-view.cpp
@@ -1743,7 +1743,11 @@ bool WidgetInfo::PathChanged(const char *setting)
 	const char *desc = obs_property_description(property);
 	obs_path_type type = obs_property_path_type(property);
 	const char *filter = obs_property_path_filter(property);
-	const char *default_path = obs_property_path_default_path(property);
+	const char *last_path = obs_data_get_string(view->settings, setting);
+	const char *default_path =
+		(last_path && *last_path)
+			? NULL
+			: obs_property_path_default_path(property);
 	QString path;
 
 	if (type == OBS_PATH_DIRECTORY)


UI/properties-view.cpp
```diff
@@ -1743,7 +1743,11 @@ bool WidgetInfo::PathChanged(const char *setting)
 	const char *desc = obs_property_description(property);
 	obs_path_type type = obs_property_path_type(property);
 	const char *filter = obs_property_path_filter(property);
-	const char *default_path = obs_property_path_default_path(property);
+	const char *last_path = obs_data_get_string(view->settings, setting);
+	const char *default_path =
+		(last_path && *last_path)
+			? NULL
+			: obs_property_path_default_path(property);
```
PatTheMav, line 1750:
```suggestion
    const char *last_path = obs_data_get_string(view->settings, setting);
	const char *default_path = obs_property_path_default_path(property);
    QString pathCandidate = (last_path && *last_path) ? QT_UTF8(last_path) : QT_UTF8(default_path);
```

Then use `pathCandidate` below instead of the `QT_UTF8` calls.

PatTheMav (commented):
Approach looks fine to me, it also makes sense for filters/plugins that provide their own files to use their location as their default path.

PatTheMav:
Just tested this on macOS and Windows - on Windows it seems to work as expected, on macOS it does not change the overall behaviour because native file pickers store their global state, so it will always open back up in the last directory that was opened.

@Warchamp7 any concerns about this UX-wise?

Warchamp7:
This fixes the mentioned issue but doesn't address things like the Defaults button also changing the value of this field. The underlying problem is that we're using default_path here which is only set when the path property is added to the view.

I've taken a more thorough approach to solving this in #10485

PR #6788, opened 2022-07-23
UI: Search combo property item with QVariant type

norihiro:
### Description

When finding current item for a combobox on properties view, use `QVariant` type to hold current values instead of converting all types into string.

### Motivation and Context

A combobox created with `OBS_COMBO_FORMAT_FLOAT` cannot show the current item. When a modified callback for another item on the properties view returns true, the combobox is initialized again and the selection goes to the top item even though another item is selected.

When adding items into the combobox, `userData` is generated in the code below.
https://github.com/obsproject/obs-studio/blob/e3f416f3fc72db49d1ba71b8423454d053a9d6bf/UI/properties-view.cpp#L505-L523

| OBS's type | `userData` generation function |
| ---------- | --------- |
| `OBS_COMBO_FORMAT_INT`    | `QVariant::fromValue<long long>(long long)` |
| `OBS_COMBO_FORMAT_FLOAT`  | `QVariant::fromValue<double>(double)` |
| `OBS_COMBO_FORMAT_STRING` | `QByteArray(const char *)` |

I'm implementing a plugin on norihiro/obs-color-monitor#54 and found this bug.

### How Has This Been Tested?

OS: Fedora 35
Qt version: 5.15.2-31.fc35 and 6.2.3-2.fc35 (I checked with both versions.)

Checked these steps for `xshm_input` (using `OBS_COMBO_FORMAT_INT`) and `slideshow` (using `OBS_COMBO_FORMAT_STRING`).
- Create a source.
- Open it's properties view.
- Choose another item in combobox.
- Close the properties view and open it again.
- Check the combobox shows the same item.
  - Before this bugfix, the top item is always selected at this step for the combobox with `OBS_COMBO_FORMAT_FLOAT`.

Also checked another source under development using `OBS_COMBO_FORMAT_FLOAT`.

Need to check autoselect functionality but only available on two sources `av_capture_input` on macOS and `dshow_input` on Windows.

I've checked the signature and description of `QComboBox::findData` are same in between Qt5 and Qt6.
Qt5: https://doc.qt.io/qt-5/qcombobox.html#findData
Qt6: https://doc.qt.io/qt-6/qcombobox.html#findData

### Types of changes

### Checklist:

  - I didn't test autoselect functionality.
  - I didn't test on Windows though especially the behavior of `QT_UTF8` is different on Windows.

### DIFF
--- a/UI/properties-view.cpp
+++ b/UI/properties-view.cpp
@@ -541,30 +541,30 @@ static void AddComboItem(QComboBox *combo, obs_property_t *prop,
 template<long long get_int(obs_data_t *, const char *),
 	 double get_double(obs_data_t *, const char *),
 	 const char *get_string(obs_data_t *, const char *)>
-static string from_obs_data(obs_data_t *data, const char *name,
-			    obs_combo_format format)
+static QVariant from_obs_data(obs_data_t *data, const char *name,
+			      obs_combo_format format)
 {
 	switch (format) {
 	case OBS_COMBO_FORMAT_INT:
-		return to_string(get_int(data, name));
+		return QVariant::fromValue(get_int(data, name));
 	case OBS_COMBO_FORMAT_FLOAT:
-		return to_string(get_double(data, name));
+		return QVariant::fromValue(get_double(data, name));
 	case OBS_COMBO_FORMAT_STRING:
-		return get_string(data, name);
+		return QByteArray(get_string(data, name));
 	default:
-		return "";
+		return QVariant();
 	}
 }
 
-static string from_obs_data(obs_data_t *data, const char *name,
-			    obs_combo_format format)
+static QVariant from_obs_data(obs_data_t *data, const char *name,
+			      obs_combo_format format)
 {
 	return from_obs_data<obs_data_get_int, obs_data_get_double,
 			     obs_data_get_string>(data, name, format);
 }
 
-static string from_obs_data_autoselect(obs_data_t *data, const char *name,
-				       obs_combo_format format)
+static QVariant from_obs_data_autoselect(obs_data_t *data, const char *name,
+					 obs_combo_format format)
 {
 	return from_obs_data<obs_data_get_autoselect_int,
 			     obs_data_get_autoselect_double,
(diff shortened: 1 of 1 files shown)


RytoEX:
> I've checked the signature and description of `QComboBox::findData` are same in between Qt5 and Qt6.
> Qt5: https://doc.qt.io/qt-5/qcombobox.html#findData
> Qt6: https://doc.qt.io/qt-6/qcombobox.html#findData

See [QTBUG-104990](https://bugreports.qt.io/browse/QTBUG-104990).
cc: @gxalpha

norihiro:
However [QTBUG-104990](https://bugreports.qt.io/browse/QTBUG-104990) claims the bug was introduced in Qt6, the issue was found in Qt5. Maybe `OBS_COMBO_FORMAT_INT` is broken by Qt6 but `OBS_COMBO_FORMAT_FLOAT` is not working in both Qt5 and Qt6.

RytoEX:
> However [QTBUG-104990](https://bugreports.qt.io/browse/QTBUG-104990) claims the bug was introduced in Qt6, the issue was found in Qt5. Maybe `OBS_COMBO_FORMAT_INT` is broken by Qt6 but `OBS_COMBO_FORMAT_FLOAT` is not working in both Qt5 and Qt6.

I'm saying that the behavior of `QComboBox::findData` changes between Qt5 and Qt6, which is why I highlighted the text saying that the function is the same between Qt5 and Qt6.

UI/properties-view.cpp
```diff
@@ -541,30 +541,30 @@ static void AddComboItem(QComboBox *combo, obs_property_t *prop,
...
 	 const char *get_string(obs_data_t *, const char *)>
-static string from_obs_data(obs_data_t *data, const char *name,
-			    obs_combo_format format)
+static QVariant from_obs_data(obs_data_t *data, const char *name,
+			      obs_combo_format format)
 {
 	switch (format) {
 	case OBS_COMBO_FORMAT_INT:
-		return to_string(get_int(data, name));
+		return QVariant::fromValue<long long>(get_int(data, name));
```
jpark37, line 549: Template type should be inferred from get_int.

UI/properties-view.cpp
```diff
@@ -541,30 +541,30 @@ static void AddComboItem(QComboBox *combo, obs_property_t *prop,
...
 	case OBS_COMBO_FORMAT_FLOAT:
-		return to_string(get_double(data, name));
+		return QVariant::fromValue<double>(get_double(data, name));
```
jpark37, line 551: Template type should be inferred from get_double.

jpark37:
I know I just added two (optional) defects, but please merge this sooner than later. It's quite a nasty bug. The code looks good otherwise.

jpark37 (changes requested):
Update the commit message. This is no longer just about OBS_COMBO_FORMAT_FLOAT.

jpark37 (approved):

jpark37:
Looks good to me. Someone merge this ASAP as possible.

PR #6789, opened 2022-07-23
UI: Remove unused functions for audio mixers

norihiro:
### Description

Remove two functions `save_audio_source` and `load_audio_source`.

### Motivation and Context

These functions are defined in C++ source code in `UI` but never called from any code. Also not defined in any header file.

I found `double` type is passed to `obs_source_set_audio_mixers`, where it requires `uint32_t` type, each bit of which defines enable/disable of sending the source to the output. Even though these functions are not in use, I don't want to see such a type conversion.

### How Has This Been Tested?

Built OBS Studio on Fedora 35 and confirmed it works.
Build OBS Studio on CI and confirmed no link errors.

### Types of changes

### Checklist:

### DIFF
--- a/UI/window-basic-main.cpp
+++ b/UI/window-basic-main.cpp
@@ -4899,42 +4899,6 @@ void OBSBasic::on_actionShowMissingFiles_triggered()
 	}
 }
 
-void save_audio_source(int channel, obs_data_t *save)
-{
-	OBSSourceAutoRelease source = obs_get_output_source(channel);
-	if (!source)
-		return;
-
-	OBSDataAutoRelease obj = obs_data_create();
-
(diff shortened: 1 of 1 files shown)


WizardCM:
Looks like these were introduced as part of https://github.com/obsproject/obs-studio/commit/3c5e2ddc37f40fe3f44788ca8000ebe258abcd00

Unsure if their lack of use is due to no longer being necessary, or caused by a bug from someone deleting those lines unintentionally.

norihiro:
These functions are once added by a commit 3c5e2ddc3 to implement the undo/redo feature but later another commit cec8061d8 implemented the feature in different way. At that time, these two functions are not removed. So I think they are safe to remove.
The CI is failing at `Install dependencies` step, which is not related to this PR.

PR #6792, opened 2022-07-23
obs-filters: Add HDR bypass for color correction

jpark37:
### Description
Not sure how to implement controls for HDR, so skip unless SDR for now.

### Motivation and Context
Don't want to apply SDR color processing on HDR video.

### How Has This Been Tested?
Verify scenarios in RenderDoc:
- [x] SRGB on SRGB -> controls work
- [x] SRGB16 on SRGB -> controls work
- [x] EDR on SRGB -> bypass
- [x] SRGB on SRGB16 -> controls work
- [x] SRGB16 on SRGB16 -> controls work
- [x] EDR on SRGB16 -> bypass
- [x] SRGB on EDR -> controls work
- [x] SRGB16 on EDR -> controls work
- [x] EDR on EDR -> bypass
- [x] SRGB on SCRGB -> controls work (multiply sdr_white_level/80.)
- [x] SRGB16 on SCRGB -> controls work (multiply sdr_white_level/80.)
- [x] EDR on SCRGB -> bypass (multiply sdr_white_level/80.)

### Types of changes

### Checklist:


WizardCM:
I think it may be worth using the new INFO property to inform the user that the filter won't do anything, otherwise they'll be confused why the sliders don't do anything.

jpark37:
Whether or not the source is SDR/HDR can change from frame to frame, and I only know from the graphics thread. Do you want an unconditional label that says the filter only works in SDR? I'm also planning to copy this PR pattern to the other color filters.

jpark37:
I added an info line. If we add real HDR settings later, we should probably replace with groups.

Issue #6787, opened 2022-07-22
OSX ARM version crashing right at launch starting with Build 8975


cmierzwa:
### Operating System Info

macOS 12

### OBS Studio Version

Git

### OBS Studio Version (Other)

Issue starting with Master Build 8975

### OBS Studio Log URL

https://imgur.com/a/whiI5aF

### OBS Studio Crash Log URL

https://imgur.com/a/whiI5aF

### Expected Behavior

Should run as the usual "Master" ARM builds have been running.

### Current Behavior

Starting with Master Build 8975 of the OSX ARM version is crashing right after launching. The IMGUR link I have put above is for Build 9014 but it occurs with every build I've tried starting with build 8975. OSX ARM Build 8947 was working fine like previous builds.

### Steps to Reproduce

1. Simply launch OBS OSX ARM build 8975 through 9022 (latest available as of opening this case)
2.
3.
...

### Anything else we should know?

Mac Studio 
128GB RAM / 4TB SSD / 20 core CPU / 48 core GPU / 32 core NE
OSX v12.5

RytoEX:
Could you please put the contents of the problem reports into text files and upload them to this Issue? Reading the text in screenshots is difficult.

That said, this is most likely related to changes we made for loading Python scripts. What version of Python do you have installed?

cmierzwa:
Hi @RytoEX,

You bet...full report is below. I have brew installed if that makes a difference. That said:

"python3" yields

Python 3.9.13 (main, May 24 2022, 21:13:51) 
[Clang 13.1.6 (clang-1316.0.21.2)] on darwin
Type "help", "copyright", "credits" or "license" for more information.

"brew list" yields

==> Formulae
ca-certificates	gdbm		htop		mpdecimal	ncurses		openssl@1.1	python@3.9	readline	sqlite		xz

Here's the full text of the crash log when launching OSX ARM Build 9014

-------------------------------------
Translated Report (Full Report Below)
-------------------------------------

Process:               OBS [33182]
Path:                  /Applications/OBS-ARM-b9014-FAIL.app/Contents/MacOS/OBS
Identifier:            com.obsproject.obs-studio
Version:               27.2.0 (2718946331)
Code Type:             ARM-64 (Native)
Parent Process:        launchd [1]
User ID:               501

Date/Time:             2022-07-22 16:44:18.2531 -0700
OS Version:            macOS 12.5 (21G72)
Report Version:        12
[...]

PR #6796, opened 2022-07-23
UI: Fix crash on macOS if no python path is set in configuration

PatTheMav:
### Description
Obvious fix guarding against calling the std::string constructor with a NULL pointer.

### Motivation and Context
Fix OBS crashing on macOS without preconfigured Python path.

### How Has This Been Tested?
Tested on macOS with clean config and config without available Python configuration.

### Types of changes

### Checklist:


RytoEX:
Does this fix #6787?

PatTheMav:
> Does this fix #6787?

Probably, given from what I was able to decipher from the crash log screenshots.

RytoEX:
Reporter of #6787 confirmed that this fixes that Issue.

PR #6802, opened 2022-07-24
UI: Omit stream codecs the service doesn't support

jp9000:
### Description
Makes it so that only encoders that are of the supported codecs of the currently selected service are shown to the user in simple/advanced output mode. That way a user can't stream with a codec that a service doesn't support for example.

If the user selects a service that doesn't support that currently configured stream encoder, then it will display a message warning the user that the program will have to automatically select the closest encoder to the current encoder, but of a codec that the service does support (i.e. if you're using NVENC/AMF it'll try to fallback to a compatible NVENC/AMF codec). If they click yes, it'll change the encoders. If they click no, it'll revert the currently selected service to the previously selected service.

Also can I say just how incredibly annoying it was to write this god forsaken PR? I feel like I am suffering the technical debt I created for myself. User interface programming really sucks. Especially when there's a lot of features. UI programming is insufferable. At least if you want good UX. (Sorry for complaining)

Also ignore the fact that the service validator is failing. It's because I added new json entries for each service.

### Motivation and Context
Now that we support HEVC (if only via hardware encoders), certain services (YouTube) need to be able to stream with HEVC. So the user needs to be able to select HEVC as an encoder to stream with. But they can't use that codec for other services which don't support that, so we have to make sure the user can't select them. So basically we have to update the combo boxes every time the user changes services.

It's really dumb and required a lot of code, but it's required for good UX I guess.

### How Has This Been Tested?
I tested switching back and forth between the services and made sure that it properly updated the encoder lists, and ensured it displayed the warning when that warning was supposed to occur.

Basically, I did basic testing and everything... seems to be in order.

### Types of changes

### Checklist:


plugins/rtmp-services/data/services.json
```diff
@@ -1863,6 +2242,12 @@
         {
             "name": "Glimesh",
             "stream_key_link": "https://glimesh.tv/users/settings/stream",
+            "supported video codecs": [
+                "h264"
+            ],
+            "supported audio codecs": [
+                "aac"
+            ],
```
tytan652, line 2250: FTL based services (Glimesh, Younow), should not have opus as audio codec ?

tytan652:
Also IIRC `ffmpeg_hls_muxer` does not have HEVC in its supported codec list.

libobs/obs-encoder.h
```diff
@@ -240,7 +240,7 @@ struct obs_encoder_info {
 
 	/**
 	 * Gets the default settings for this encoder
-	 * 
+	 *
```
tytan652, line 243: Nipick: This whitespace change should maybe not be there.

plugins/rtmp-services/rtmp-common.c
```diff
@@ -850,6 +853,49 @@ static void rtmp_common_get_max_bitrate(void *data, int *video_bitrate,
 	json_decref(root);
 }
 
+const char **rtmp_common_get_supported_video_codecs(void *data)
```
norihiro, line 856: I think this function should be static.
```suggestion
static const char **rtmp_common_get_supported_video_codecs(void *data)
```

libobs/obs-service.c
```diff
@@ -467,3 +467,12 @@ void obs_service_get_max_bitrate(const obs_service_t *service,
...
 }
+
+const char **
+obs_service_get_supported_video_codecs(const obs_service_t *service)
+{
+	if (service->info.get_supported_video_codecs)
+		return service->info.get_supported_video_codecs(
+			service->context.data);
+	return NULL;
+}
```
norihiro, line 478: This function should be documented on `docs/sphinx/reference-services.rst`.
I'd like to suggest that the returned value should not be freed by `bfree`.

tytan652:
I confirm that this PR works (just add to remove the services.json in my config).

Once my comments are resolved, for me this PR is okay.

Note: custom service is considered compatible with all the codecs even if we put a non-SRT/RIST URL, but it's custom so users should maybe know what they actually doing.

UI/window-basic-main-outputs.cpp
```diff
@@ -349,6 +349,36 @@ void SimpleOutput::LoadStreamingPreset_Lossy(const char *encoderId)
 	obs_encoder_release(videoStreaming);
 }
 
+/* mistakes have been made to lead us to this. */
```
norihiro, line 352:
```suggestion
/* Mistakes have been made to lead us to this. */
```

libobs/obs-service.h
```diff
@@ -86,6 +86,8 @@ struct obs_service_info {
 
 	void (*get_max_bitrate)(void *data, int *video_bitrate,
 				int *audio_bitrate);
+
+	const char **(*get_supported_video_codecs)(void *data);
```
norihiro, line 90: The corresponding document `docs/sphinx/reference-services.rst` should be updated.

norihiro (commented):
I'm sorry to leave nitpick comments.

UI/window-basic-main-outputs.cpp
jp9000: It's not a big deal if some comments start with lowercase. They're just comments.

libobs/obs-service.c
jp9000: The return value doesn't need to be freed.

UI/window-basic-settings-stream.cpp
```diff
@@ -1177,3 +1182,241 @@ bool OBSBasicSettings::IsServiceOutputHasNetworkFeatures()
...
 	return false;
 }
+
+static bool service_supports_codec(const char **codecs, const char *codec)
+{
+	if (!codecs)
+		return true;
+
+	while (*codecs) {
+		if (astrcmpi(*codecs, codec) == 0)
```
notr1ch, line 1192: Is there a reason this is a case-insensitive compare? Feels like we should `strcmp` here since this data shouldn't be coming from a user and should be all-lowercase.

plugins/rtmp-services/data/schema/service-schema-v3.json
```diff
@@ -25,6 +25,24 @@
                         "$ref": "#/definitions/saneUrl",
                         "description": "Link where a logged-in user can find the 'stream key', presented as a button alongside the stream key field."
                     },
+                    "supported video codecs": {
+                        "type": "array",
+                        "description": "Video codecs that are supported by the service.",
+                        "items": {
+                            "type": "string",
+                            "description": "Short-form codec names.",
+                            "minLength": 1
```
WizardCM, line 34: Additionally we can specify an enum here of possible strings, that way if someone submits a service with an invalid codec CI/their editor will inform them. Something like

```suggestion
                            "minLength": 1,
                            "enum": [
                                "h264",
                                "hevc"
                            ]
```

notr1ch:
How difficult would it be to make the dialog more specific? For example, instead of :

> This streaming service does not support your current codec. Your current codec will be changed to the closest compatible codec.

I would prefer something like:

> Twitch does not support HEVC. Your encoder will be changed to H264.

Aside from that, I think we should also replace "codec" with "encoder" as we don't use "codec" in the settings anywhere which could lead to confusion over what's being changed.

jp9000:
But H264 isn't an encoder, it's a codec

jp9000:
I'll make a modification to the string that should satisfy your asks.

notr1ch:
We've never made the distinction between a codec and an encoder in the settings - we don't ask users to pick a codec and then an implementation, they just pick an encoder and the codec is implicit. Saying we're changing the codec, while technically correct, may be confusing to users who want to revert the change and can't find a "codec" setting anywhere.

jp9000:
How about that r1ch?

jp9000:
This is what the dialog looks like when it appears: 
obs64_2022-07-26_08-42-07

jp9000:
Basically, it'll use the actual combo box item name which the user selected so the user has a better idea of which setting it was referring to.

notr1ch:
Yeah that's a good improvement IMO, perhaps the : could be dropped but best to let someone else nitpick the UX here :).

jp9000:
Alrighty then, I'll drop the ":", then I'll rebase and merge.

PR #6805, opened 2022-07-24
UI: Add functions to check for and request macOS permissions

PatTheMav:
### Description
Adds functions to check and/or request specific macOS permissions (audio device access, video device access, accessibility access, and screen capture access).

By default only audio capture, video capture, and accessibility are requested on launch - the first two have straight-forward "Yes/No" prompts, the latter requires people to enable OBS in the settings application (but is required for hotkey functionality, independent
of scene setups).

### Motivation and Context
Reduce friction to gain necessary permissions on macOS, especially for fixed hotkey functionality.

### How Has This Been Tested?
Yes.

### Types of changes

### Checklist:

### DIFF
--- a/UI/obs-app.cpp
+++ b/UI/obs-app.cpp
@@ -2185,6 +2185,15 @@ static int run_program(fstream &logFile, int argc, char *argv[])
 		}
 
 #ifdef __APPLE__
+		MacPermissionStatus audio_permission =
+			CheckPermission(kAudioDeviceAccess);
+		MacPermissionStatus video_permission =
+			CheckPermission(kVideoDeviceAccess);
+		MacPermissionStatus accessibility_permission =
+			CheckPermission(kAccessibility);
+		MacPermissionStatus screen_permission =
+			CheckPermission(kScreenCapture);
+
 		bool rosettaTranslated = os_get_emulation_status();
 		blog(LOG_INFO, "Rosetta translation used: %s",
 		     rosettaTranslated ? "true" : "false");
--- a/UI/platform-osx.mm
+++ b/UI/platform-osx.mm
@@ -18,6 +18,7 @@
 #include <sstream>
 #include <dlfcn.h>
 #include <util/base.h>
+#include <util/threading.h>
 #include <obs-config.h>
 #include "platform.hpp"
 #include "obs-app.hpp"
@@ -26,6 +27,9 @@
 #include <sys/sysctl.h>
 
 #import <AppKit/AppKit.h>
+#import <CoreFoundation/CoreFoundation.h>
+#import <AVFoundation/AVFoundation.h>
+#import <ApplicationServices/ApplicationServices.h>
 
 using namespace std;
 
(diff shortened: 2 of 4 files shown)


UI/obs-app.cpp
```diff
@@ -2070,6 +2070,13 @@ static int run_program(fstream &logFile, int argc, char *argv[])
 
 #if __APPLE__
 	InstallNSApplicationSubclass();
+	bool audio_permission =
+		CheckPermissionWithPrompt(kAudioDeviceAccess, true);
+	bool video_permission =
+		CheckPermissionWithPrompt(kVideoDeviceAccess, true);
+	bool accessibility_permission =
+		CheckPermissionWithPrompt(kAccessibility, true);
```
gxalpha, line 2078: Should these use the RequestPermission macro?

gxalpha:
In general, I'm not sure if we should really just request these permissions on startup - especially the prompts for accessibility and screen capture have the tendency to just disappear behind the main OBS window (meaning users don't see them). Also users might wonder why OBS wants all of this even if they haven't done anything yet.
I wonder if we should have some kind of window saying "OBS needs permission x for this, y for this - which of these do you want to grant?". For reference, this is the style I'd think of (but for permissions instead, and this one is a bit too wordy):
image

gxalpha (commented):
Noted this off-thread already but just so that this doesn't get lost: Now that we use an enum the blog calls aren't correct anymore (they're giving wrong information). Also I think it'd be great if each check always calls blog, not just when a prompt is requested.

PatTheMav:
> Noted this off-thread already but just so that this doesn't get lost: Now that we use an enum the blog calls aren't correct anymore (they're giving wrong information). Also I think it'd be great if each check always calls blog, not just when a prompt is requested.

Right, it's only fixed for `kAccessibility` - waiting for your force push before making any other changes.

gxalpha:
I didn't plan to do any more force pushes, did everything I wanted to in the one. Sorry for the confusion.

gxalpha (approved):
Works fine with the dialog (which currently is WIP). No issues on my side.

UI/obs-app.cpp
```diff
@@ -2110,6 +2110,22 @@ static int run_program(fstream &logFile, int argc, char *argv[])
 		OBSTranslator translator;
 		program.installTranslator(&translator);
 
+		if (!created_log) {
+			create_log_file(logFile);
+			created_log = true;
+		}
+
+#ifdef __APPLE__
```
WizardCM, line 2118: Does this actually need to occur before `if (!already_running)`?

gxalpha: We could probably move it further down if that's desirable

PatTheMav: Good point, moved it accordingly.

UI/platform-osx.mm
```diff
@@ -238,6 +242,138 @@ - (void)drawRect:(NSRect)dirtyRect
...
+						      (CFDictionaryRef)options)
+						      ? kPermissionAuthorized
+						      : kPermissionDenied);
+		}
+
+		blog(LOG_INFO, "[macOS] Permission for accessibility %s.",
+		     permissionResponse == kPermissionAuthorized ? "granted"
+								 : "denied");
+		return permissionResponse;
+		break;
```
WizardCM, line 372: Is this standard in Objective C and/or consistent with our other uses of switch statements?

PatTheMav: Removed the early returns, as every block assigns a correct value to `permissionresponse`.

PR #1832, opened 2019-04-16
[WIP] UI: Add notification widget

cg2121:
TODO:
- [ ] Add more functionality to the url button
- [ ] Don't resize window/preview when notification appears

This is a refactoring of #1543

The warning and info icons are from: https://github.com/Keyamoon/IcoMoon-Free
The error icon is from: https://github.com/feathericons/feather
The close icon is from: https://octicons.github.com

Screenshot from 2019-04-16 02-31-58

PR #6815, opened 2022-07-25
libobs: Add api to for notifications

cg2121:
### Description
This adds an api to libobs to show a notification in the
main UI. This is only for the libobs changes, as the UI
will be added later.

### Motivation and Context
Split from #1832 

### How Has This Been Tested?
Tested with #1832 

### Types of changes

### Checklist:

### DIFF
--- a/libobs/obs.h
+++ b/libobs/obs.h
@@ -2617,6 +2617,19 @@ EXPORT void obs_source_frame_copy(struct obs_source_frame *dst,
 /* Get source icon type */
 EXPORT enum obs_icon_type obs_source_get_icon_type(const char *id);
 
+/* ------------------------------------------------------------------------- */
+/* Functions to show notification in main UI */
+enum obs_notify_type {
+	OBS_NOTIFY_TYPE_INFO,
+	OBS_NOTIFY_TYPE_WARNING,
+	OBS_NOTIFY_TYPE_ERROR,
+};
+
+EXPORT uint32_t obs_show_notification(enum obs_notify_type type,
+				      const char *message, bool persist,
+				      void *data);
+EXPORT void obs_close_notification(uint32_t id);
+
 #ifdef __cplusplus
 }
 #endif
--- a/libobs/obs-internal.h
+++ b/libobs/obs-internal.h
@@ -488,6 +488,8 @@ struct obs_core {
 	os_task_queue_t *destruction_task_thread;
 
 	obs_task_handler_t ui_task_handler;
+
+	volatile long notification_id;
 };
 
 extern struct obs_core *obs;
--- a/libobs/obs.c
+++ b/libobs/obs.c
@@ -1113,6 +1113,9 @@ static const char *obs_signals[] = {
 	"void hotkey_unregister(ptr hotkey)",
 	"void hotkey_bindings_changed(ptr hotkey)",
 
+	"void show_notification(int id, int type, string message, bool persist, ptr data)",
+	"void close_notification(int id)",
+
 	NULL,
 };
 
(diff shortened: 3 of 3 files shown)


libobs/obs.h
```diff
@@ -2518,6 +2518,19 @@ EXPORT void obs_source_frame_copy(struct obs_source_frame *dst,
...
 
+/* ------------------------------------------------------------------------- */
+/* Functions to show notification in main UI */
+enum obs_notify_type {
+	OBS_NOTIFY_TYPE_INFO,
+	OBS_NOTIFY_TYPE_WARNING,
+	OBS_NOTIFY_TYPE_ERROR,
+};
+
+EXPORT uint32_t obs_show_notification(enum obs_notify_type type,
```
GeorgesStavracas, line 2529: I'm on the fence about the API returning `uint32_t` for the id. Usually, the id is passed as input, and it's in string format. This is helpful for consumers of this API, so that they are able to recreate the notification id at any point in time.

libobs/obs.h
```diff
@@ -2518,6 +2518,19 @@ EXPORT void obs_source_frame_copy(struct obs_source_frame *dst,
...
+				      const char *message, bool persist,
```
GeorgesStavracas, line 2530: Hm, I feel like this API is missing a few important fields... usually, when it comes to notifications, the following information is passed by API consumers:

 * id → string
 * title → string
 * message → string (alternative names: body, description)
 * priority → enum obs_notification_priority
 * actions → array of ( action name → string )
 * default action → string
 * user data → raw pointer

Actions in notifications are useful to make them actionable. This way, we avoid notifying users about something without giving them a means to act on what is being notified. What do you think?

GeorgesStavracas (changes requested):

cg2121:
@GeorgesStavracas my idea is to have the message as html, so in the future, more features could be added without changing the api. Just the UI needs to be changed.

Example message:
```
// Basic summary of message, required
<summary>The settings need to be changed</summary>

// Use multiple hyperlinks for different actions, optional
<a href="https://obsproject.com/wiki">Learn more</a>
<a href="settings://">Go to settings</a>

// The message in more detail, optional
<description>A more detailed description about the setting</description>
```

libobs/obs.h
cg2121: See https://github.com/obsproject/obs-studio/pull/6815#issuecomment-1195702976

GeorgesStavracas:
Not sure using HTML for notifications is a good idea. A constrained markup language is probably fine, and I'm sure Qt supports that, but using full blown HTML unnecessarily binds notifications to some sort of browser renderer. But actually, my concern is that if the API doesn't have well defined boundaries, nor provides a rich set of details for the UI, we won't be able to properly implement policies on how to present these notifications.

cg2121:
Not html, my bad. xml would be a better descriptor of what it is.

libobs/obs.h
cg2121: My only concern with that is, how would you make sure every id is unique, if the caller is generating the id themselves?

gxalpha:
It would be helpful if the functions and enum values defined were added to the documentation, with some explanation as to how the notifications are expected to be shown in a UI implementation.

\
Also, to throw my two cents in here:
I think that using just one string in some type of XML format seems like unneeded complexity to me if it could just be separate arguments instead:
- title (string)
- description (string, would only allow plaintext, maybe bold and italic?)
- button 1 (probably some custom type)
- button 2

I think it would definitely be beneficial if buttons could a) have different types (normal, danger, etc) and b) execute arbitrary code on click, similar how they do with the properties system. That said, I'm not 100% sure on how that'd be implemented in the API.

tt2468:
Perhaps I'm missing some context, but what was the motivation behind making this a libobs API versus something in the frontend API? It honestly really feels like this doesn't belong here. If plugins want to show notifications, they should honestly just link against the frontend API and use functions there.

cg2121:
@jp9000 has said many times in the past, that he doesn't like non-frontend plugins to be linked with the frontend api.

energizerfellow:
It would also be useful if a button type in the notifications had a URL attached that could be opened by the OS's default browser outside OBS. Typical example would be a "More Info" button linking to a wiki/KB page with more info on said error.

cg2121:
Closing, as UI stuff shouldn't be in libobs, ideally. I will keep the UI PR open, though, as we can just show notifications from the UI side without needing libobs. For plugins to use notifications, the frontend api would be a better choice. The frontend api functions will be added to the UI PR.

PR #6820, opened 2022-07-26
UI: Add macOS permissions window

gxalpha:
### Description

Adds a window for simpler permissions management to macOS.
Pops up the first time when opening OBS, and can be invoked manually via a menu item.

The buttons should be self-explanatory.
Microphone and Camera permissions can be requested and granted directly (if not already rejected), Screen Capture and Accessibility always require a trip into System Preferences. 

Yami (with missing font issue):
image
Dark:
image

Menu entry:
image

### Motivation and Context

Permissions are a pain, and this might make things easier for users.

### How Has This Been Tested?

macOS 13
Tested functionality of buttons and that they correctly show the status.
Tested that the window only pops up by default once. Tested that it is invokable via the menu.

### Types of changes

### Checklist:

### DIFF
--- a/UI/data/locale/en-US.ini
+++ b/UI/data/locale/en-US.ini
@@ -466,6 +466,23 @@ MissingFiles.AutoSearchText="OBS has found additional matches for missing files 
 MissingFiles.NoMissing.Title="Missing Files Check"
 MissingFiles.NoMissing.Text="No files appear to be missing."
 
+# macOS permissions dialog
+MacPermissions.Title="Review App Permissions"
+MacPermissions.Description="OBS Studio requires your permission to be able to provide certain features. It is recommended to enable these permissions, but they are not required to use the app. You can always enable them later."
+MacPermissions.Description.OpenDialog="You can re-open this dialog via the OBS Studio menu."
+MacPermissions.AccessGranted="Access Granted"
+MacPermissions.RequestAccess="Request Access"
+MacPermissions.OpenPreferences="Open %1 Preferences"
+MacPermissions.Item.ScreenRecording="Screen Recording"
+MacPermissions.Item.ScreenRecording.Details="OBS requires this permission to be able to capture your screen."
+MacPermissions.Item.Camera="Camera"
+MacPermissions.Item.Camera.Details="This permission is needed in order to capture content from a webcam or capture card."
+MacPermissions.Item.Microphone="Microphone"
+MacPermissions.Item.Microphone.Details="OBS requires this permission if you want to capture your microphone."
+MacPermissions.Item.Accessibility="Accessibility"
+MacPermissions.Item.Accessibility.Details="For keyboard shortcuts (hotkeys) to work while other apps are focused, please enable this permission."
+MacPermissions.Continue="Continue"
+
 # update dialog
 UpdateAvailable="New Update Available"
 UpdateAvailable.Text="Version %1.%2.%3 is now available. <a href='%4'>Click here to download</a>"
--- a/UI/window-permissions.cpp
+++ b/UI/window-permissions.cpp
@@ -1,0 +1,102 @@
+/******************************************************************************
+    Copyright (C) 2014 by Hugh Bailey <obs.jim@gmail.com>
+
+    This program is free software: you can redistribute it and/or modify
+    it under the terms of the GNU General Public License as published by
+    the Free Software Foundation, either version 2 of the License, or
+    (at your option) any later version.
+
+    This program is distributed in the hope that it will be useful,
+    but WITHOUT ANY WARRANTY; without even the implied warranty of
+    MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+    GNU General Public License for more details.
+
+    You should have received a copy of the GNU General Public License
+    along with this program.  If not, see <http://www.gnu.org/licenses/>.
+******************************************************************************/
+
+#include <QLabel>
+#include "window-permissions.hpp"
+#include "obs-app.hpp"
+
+OBSPermissions::OBSPermissions(QWidget *parent, MacPermissionStatus capture,
+			       MacPermissionStatus video,
+			       MacPermissionStatus audio,
+			       MacPermissionStatus accessibility)
+	: QDialog(parent), ui(new Ui::OBSPermissions)
+{
+	ui->setupUi(this);
+	SetStatus(ui->capturePermissionButton, capture,
+		  QTStr("MacPermissions.Item.ScreenRecording"));
+	SetStatus(ui->videoPermissionButton, video,
+		  QTStr("MacPermissions.Item.Camera"));
+	SetStatus(ui->audioPermissionButton, audio,
+		  QTStr("MacPermissions.Item.Microphone"));
+	SetStatus(ui->accessibilityPermissionButton, accessibility,
+		  QTStr("MacPermissions.Item.Accessibility"));
+}
+
+void OBSPermissions::SetStatus(QPushButton *btn, MacPermissionStatus status,
+			       const QString &preference)
+{
+	if (status == kPermissionAuthorized) {
+		btn->setText(QTStr("MacPermissions.AccessGranted"));
+	} else if (status == kPermissionNotDetermined) {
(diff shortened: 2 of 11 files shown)


UI/data/locale/en-US.ini
```diff
@@ -462,6 +462,22 @@ MissingFiles.AutoSearchText="OBS has found additional matches for missing files
 MissingFiles.NoMissing.Title="Missing Files Check"
 MissingFiles.NoMissing.Text="No files appear to be missing."
 
+# macOS permissions dialog
+MacPermissions.Title="Review App Permissions"
+MacPermissions.Description="OBS Studio requires your permission to be able to provide certain features. It is recommended to enable these permissions, but they are not required to use the app. You can always enable them later via the OBS Studio menu."
+MacPermissions.AccessGranted="Access Granted"
+MacPermissions.RequestAccess="Request Access"
+MacPermissions.OpenPreferences="Open %1 Preferences"
```
WizardCM, line 470: This won't be a blocker for this PR, but I fear this translation key may need to be split up later as I am unsure if it'll work well in every language. Alternatively we could go the lazy way and just wrap the %1 with single quotes - though it might look a bit weird.

UI/data/locale/en-US.ini
```diff
@@ -462,6 +462,22 @@ MissingFiles.AutoSearchText="OBS has found additional matches for missing files
...
+MacPermissions.Item.ScreenRecording="Screen Recording"
+MacPermissions.Item.ScreenRecording.Details="OBS requires this permission to be able to capture your screen."
+MacPermissions.Item.Camera="Camera"
+MacPermissions.Item.Camera.Details="OBS requires this permission to be able to capture content from a webcam or capture card."
+MacPermissions.Item.Microphone="Microphone"
+MacPermissions.Item.Microphone.Details="OBS requires this permission to be able to capture your microphone."
+MacPermissions.Item.Accessibility="Accessibility"
+MacPermissions.Item.Accessibility.Details="OBS requires this permission so that you can use keyboard shortcuts (hotkeys) while other apps are focused."
```
WizardCM, line 478: The other Details strings use "to be able to". As that'd look strange here, we may have to tweak all the Details strings slightly to flow together better.

I think starting each with "OBS requires this permission" is a bit repetitive and may dissuade users from reading what's in the dialog.

UI/window-permissions.cpp
```diff
@@ -0,0 +1,101 @@
...
+void OBSPermissions::on_accessibilityPermissionButton_clicked()
+{
+	OpenMacOSPrivacyPreferences("Accessibility");
+	RequestPermission(kAccessibility);
+}
+
+void OBSPermissions::on_continueButton_clicked()
+{
+	config_set_bool(GetGlobalConfig(), "General",
+			"MacOSPermissionDialogShown", true);
```
WizardCM, line 99: Would it be worth versioning the value in case Apple/we add new permissions in future so that we can forcefully reopen the dialog when requirements change? Although it may be better to just use the notification PR and display a message that way when things change in future.

I just don't want to potentially limit ourselves this time.

UI/data/locale/en-US.ini
```diff
@@ -462,6 +462,22 @@ MissingFiles.AutoSearchText="OBS has found additional matches for missing files
...
+MacPermissions.Continue="Continue"
```
WizardCM, line 479: It may also be worth having a message at the bottom instructing the user on how they can manually reopen this dialog later.

UI/window-permissions.cpp
```diff
@@ -0,0 +1,101 @@
...
+			QTStr("MacPermissions.OpenPreferences").arg(preference));
+	}
+	btn->setEnabled(status != kPermissionAuthorized);
+	btn->setProperty("status", status);
+}
+
+void OBSPermissions::on_capturePermissionButton_clicked()
+{
+	OpenMacOSPrivacyPreferences("ScreenCapture");
+	RequestPermission(kScreenCapture);
```
WizardCM, line 57: This goes for all the buttons: After the user opens the relevant System dialog and enables permissions, will the status in the dialog dynamically change?

I think the only way to do that properly would be to start a timer loop after they click a button, and keep checking until they either close the dialog, close OBS, or successfully change the permission.

WizardCM, line 57: I know that with the main OBS window in our current implementation, the dialog for screen capture often/always appears behind OBS, making it very unclear to the user.

Is that still the case with this implementation? Do we need to force it to the front, or force ourselves to the back?

WizardCM (changes requested):
Thanks so much for doing this on such short notice & so quickly. Most of my feedback is questions rather than actionable, but I wanted to put them down anyway.

I do think as a further enhacement I'd like to see us add icons to the dialog as it's very bland and a bit hard to look at, but we can definitely skip that for this PR so we can get this merged ASAP.

UI/window-permissions.cpp
gxalpha: It updates dynamically only when they request the Microphone and Camera with the prompt.
However this is fine because permissions changed in the settings require an app restart to apply anyways. For example when I enable Screen Recording, I get the following popup:
image
"OBS.app cannot record the contents of your screen until it quit. You can decide whether you want to restart now or quit yourself later." (and then the buttons are "Quit & Reopen" / "Later").

What I did just notice is that the "Quit & Reopen" button doesn't actually quit OBS, which is interesting. Might need to look at what's going on there.

gxalpha: The dialog appears in front of the window. From what I can tell, the issue with current OBS is that the dialog opens before the main window does, so the main window puts itself in focus (in front of the dialog).
Here however the window already exists when the user presses the button, so the dialog appears in front.

UI/window-permissions.cpp
gxalpha: Good idea.

UI/data/locale/en-US.ini
WizardCM: I think moving that specific bit to the bottom would be recommended.

UI/window-permissions.cpp
WizardCM: In tests I've done before, that dialog isn't entirely true - I think as long as you don't attempt to access the thing before the user has allowed it, a restart is not strictly necessary. This could potentially depend on which permission it is and maybe even the macOS version.

UI/window-permissions.cpp
PatTheMav: Why not just pop the dialog if any permission is not given? Maybe guard if with the `DEBUG` macro, because it would pop open on _any_ code change during development otherwise.

gxalpha: I personally wouldn't want the dialog to pop up each time in my dev environment, there are many things that don't require any permissions and having to click "Continue" each time you open OBS would be an annoyance.

UI/data/locale/en-US.ini
gxalpha: Done
image

UI/data/locale/en-US.ini
gxalpha: > we may have to tweak all the Details strings slightly to flow together better

Changed them a bit, but don't think they're great. I'm always open to suggestions (especially from native speakers).

UI/window-permissions.cpp
gxalpha: Added versioning.

UI/window-permissions.cpp
WizardCM: Excellent, thanks for confirming.

PatTheMav:
A few things that came to my mind testing this:

* Menu bar items that trigger UI windows need to have ellipses at the end
* Clicking the button to open screen capture permissions should trigger the availability of OBS in the preference pane to ensure the item is "there" (if you remove the OBS item after launch, you don't get it back) - accessibility seems to work that way some times, but not always
* For those two permissions it'd be great to have an OBS.app proxy icon that people could drag and drop into the permissions dialog (like some other macOS apps do that require such permissions). That's a nice-to-have though.
* I checked some other apps and indeed the "onboarding" dialog for permissions usually appears only once, but doesn't show on consecutive launches even when permissions have been manually removed

PR #6824, opened 2022-07-26
UI: Include OpenSans font with OBS

gxalpha:
### Description

Ships the OpenSans font with OBS and reinstates the font-family statement in Yami.

### Motivation and Context

We want to be able to use the font with Yami without it needing to be installed on the users system.

### How Has This Been Tested?

On my Mac, ran without it installed. Still has the font.

### Types of changes

### Checklist:


notr1ch:
It looks like the TTF contains all variations (light, bold, semibold, extrabold, condensed, etc). Is Qt able to use all of those or would it make more sense to only include the regular typeface?

gxalpha:
@notr1ch ah I am stupid and didn't test that part, I just expected it to work. Which it doesn't.
It looks like it only uses regular right now, even though some parts should be bold. Most likely will need to add regular/bold/italic.

gxalpha:
This should be better (and actually work this time!).

Warchamp7 (approved):
Not tested, but approved from UX side

UI/data/themes/Yami.qss
```diff
@@ -77,6 +77,7 @@ QWidget {
     selection-background-color: rgb(40,76,184);
     selection-color: palette(text);
     font-size: 10pt;
+    font-family: 'Open Sans', Helvetica, Arial, sans-serif
```
WizardCM, line 80: I know Qt uses MS Shell Dlg as a sort of alias for "System font", though their docs don't seem to document it well. I would like to try placing it between Helvetica & Arial just in case.

PR #6832, opened 2022-07-27
UI: Add hotkey for resize output (source size)

cg2121:
### Description
Adds hotkey for resizing the output based on current source size.

### Motivation and Context
https://github.com/obsproject/obs-studio/commit/829e906ac2b15d822173009cde5c90b5ec6c04c1 removed the feature all together, this just removes it from the context menu and changes it to a hotkey

### How Has This Been Tested?
Resized output with the hotkey.

### Types of changes

### Checklist:

### DIFF
--- a/UI/data/locale/en-US.ini
+++ b/UI/data/locale/en-US.ini
@@ -1231,6 +1231,11 @@ AddUrl.Title="Add Source via URL"
 AddUrl.Text="You have dragged a URL into OBS. This will automatically add the link as a source. Continue?"
 AddUrl.Text.Url="URL: %1"
 
+# Dynamic output size
+ResizeOutputSizeOfSource="Resize output (source size)"
+ResizeOutputSizeOfSource.Text="The base and output resolutions will be resized to the size of the current source."
+ResizeOutputSizeOfSource.Continue="Do you want to continue?"
+
 PreviewTransition="Preview Transition"
 
 # Import Dialog
--- a/UI/window-basic-main.cpp
+++ b/UI/window-basic-main.cpp
@@ -5555,6 +5555,18 @@ void OBSBasic::CreateSourcePopupMenu(int idx, bool preview)
 			popup.addSeparator();
 		}
 
+		QAction *resizeOutput =
+			popup.addAction(QTStr("ResizeOutputSizeOfSource"), this,
+					SLOT(ResizeOutputSizeOfSource()));
+
+		int width = obs_source_get_width(source);
+		int height = obs_source_get_height(source);
+
+		resizeOutput->setEnabled(!obs_video_active());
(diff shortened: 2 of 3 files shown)


derrod:
I don't understand why this was removed in the first place, and hotkey discoverability kinda sucks, but it's better than nothing.

notr1ch:
A hotkey feels like a really odd place to put this option as it's unlikely to be used multiple times. Perhaps the option could be a pop-up dialog on the Video page in settings ("Pick a source to set the canvas size to")? There's plenty of free space in the UI there for more esoteric options like this.

gxalpha:
I agree with R1ch, it was removed from the right-click menu since it wasn't something you'd use regularly (or even just semi-regularly). In that sense, a hotkey might be worse than where it was before.

@derrod it was removed since it didn't really fit in the right-click menu, which is supposed to be for actions you'd use semi-regularly, and there wasn't an obvious place in the program where it would fit better. Additionally, it's used for a workflow that OBS in general isn't really made for or good at (dynamic-ish output size), and one of the bigger reasons for it being in Studio was just that it also was in Classic (at least so I was told).

> <details>
> <summary>For more context, a comment on the PR where the menu was removed:</summary>
>
> "However I will say: I think from a design standpoint, it's providing functionality OBS theoretically isn't built for currently and shouldn't be there. We're accelerating a workflow we don't do a very good job of supporting.
> 
> If a user wants to regularly record individual windows or games that vary in dimensions, I think they're better off with a program like ShareX until such a time we can provide a proper and convenient workflow for it."
> </details>

Because of the reasons above, I'm not sure if it's a good idea to reintroduce this at all.

cg2121:
A hotkey is a stop-gap solution until a better UI can be made. Many people have been using this feature, including myself, so removing it entirely is not a good idea.

cg2121:
> Additionally, it's used for a workflow that OBS in general isn't really made for or good at (dynamic-ish output size)

I don't think this should be a concern because it cannot be used when outputs are active.

Chiitoo:
> it was removed since it didn't really fit in the right-click menu, which is supposed to be for actions you'd use semi-regularly, and there wasn't an obvious place in the program where it would fit better.

I guess it is rare, then, but for me this is (or was) pretty much the only thing I ever used from that right-click menu.  I'd almost always use it before starting to record a window, so I'd say it was pretty regular use for me.  : )

Maybe I do things weird-like, but it certainly was useful to me (and evidently to others), and it took me a while to believe it really is gone.

Thanks!

cg2121:
Closing, as this was not the desired solution. This was meant for a quick thing for v28, but obviously it was decided not to be merged. For future versions, a better UI would be desired.

PR #6837, opened 2022-07-28
UI: Replace fake spaces with real spaces

jpark37:
### Description
Somehow NULL characters were typed into the string literals.

### Motivation and Context
Don't like fake spaces.

### How Has This Been Tested?
Binary diff of code file. I don't think any plugins use this control type though.

### Types of changes

### Checklist:

### DIFF
--- a/UI/properties-view.cpp
+++ b/UI/properties-view.cpp
@@ -1216,7 +1216,7 @@ CreateFrameRateWidget(obs_property_t *prop, bool &warning, const char *option,
 
 	auto fps_label = widget->currentFPS = new QLabel{"FPS: 22"};
 	auto time_label = widget->timePerFrame =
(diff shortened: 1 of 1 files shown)


kkartaltepe:
Iooks like its a nbsp, possibly intentionally given it's position

```
$ sed -n '1219p' UI/properties-view.cpp | xxd
00000000: 0909 6e65 7720 514c 6162 656c 7b22 4672  ..new QLabel{"Fr
00000010: 616d 6520 496e 7465 7276 616c 3a20 302e  ame Interval: 0.
00000020: 3132 33c2 a06d 7322 7d3b 0a              123..ms"};
```

At the very least the null character reference in commit message should be corrected.

jpark37:
You're right; the tool I used reported a null char. Maybe these should have a u8 prefix, but I'll just leave this alone since it probably was intentional.

PR #6842, opened 2022-07-28
Remove discouraged and deprecated usage of sprintf across the codebase

PatTheMav:
### Description
Replaces all usage of `sprintf` with the safer variant `snprintf` across the codebase.

### Motivation and Context
Usage of `sprintf` has been discouraged for many years now, because the size of the target buffer is not taken into account. `snprintf` should be used instead, which this PR switches to.

`sprintf` is also officially deprecated in macOS 13/Xcode 14.

### How Has This Been Tested?
Compiled and ran OBS on Windows and macOS.

### Types of changes

### Checklist:


tytan652:
Is the usage of `sizeof(char)` a good idea rather than the size of the char array ?

deps/libcaption/src/caption.c
```diff
@@ -408,10 +408,11 @@ size_t caption_frame_dump_buffer(caption_frame_t* frame, utf8_char_t* buf)
 {
```
notr1ch, line 408: `snprintf` can't be safely used in this function without the size of the buffer being passed as a function.

deps/obs-scripting/obs-scripting-python-import.c
```diff
@@ -65,9 +65,9 @@ bool import_python(const char *python_path)
 
 	char temp[PATH_MAX];
 
-	sprintf(cur_version, VERSION_PATTERN, PY_MAJOR_VERSION_MAX,
-		PY_MINOR_VERSION_MAX);
-	sprintf(temp, FILE_PATTERN, cur_version);
+	snprintf(cur_version, 5, VERSION_PATTERN, PY_MAJOR_VERSION_MAX,
```
notr1ch, line 68: Should be `sizeof(cur_version)`, using literals risks errors when one is changed but not the other.

notr1ch (changes requested):
There's a lot of issues with this, almost all of the changes look incorrect. Using the size of individual characters instead of the whole buffer, using literals instead of sizeof and using `snprintf` without knowing the buffer size seem common throughout. My inline comments have only highlighted a few examples but the whole PR seems to have these issues.

tytan652:
You also need to use the return value of `snprintf()` on Linux because otherwise it will/could emit `format-truncation` warnings.

https://github.com/obsproject/obs-studio/pull/5653#discussion_r767311212

deps/libcaption/src/caption.c
PatTheMav: The buffer is initialised with `CAPTION_FRAME_DUMP_BUF_SIZE`, so it should suffice to carry on with that value?

UI/window-basic-main-outputs.cpp
```diff
@@ -1305,7 +1305,7 @@ AdvancedOutput::AdvancedOutput(OBSBasic *main_) : BasicOutputHandler(main_)
 
 	for (int i = 0; i < MAX_AUDIO_MIXES; i++) {
 		char name[9];
-		sprintf(name, "adv_aac%d", i);
+		snprintf(name, sizeof(char), "adv_aac%d", i);
```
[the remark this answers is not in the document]
PatTheMav: Yep stupid mistake throughout.

PatTheMav:
> You also need to use the return value of `snprintf()` on Linux because otherwise it could emit `format-truncation` warnings with GCC.
> 
> * [Fix some compillation warnings that appear on Linux #5653 (comment)](https://github.com/obsproject/obs-studio/pull/5653#discussion_r767311212)
> * [Fix some compillation warnings that appear on Linux #5653 (comment)](https://github.com/obsproject/obs-studio/pull/5653#discussion_r848955534)
> * [096ec55](https://github.com/obsproject/obs-studio/commit/096ec55c8370052dd9d0547ad1a670b1bdd08f3f)
> * [ce62240](https://github.com/obsproject/obs-studio/commit/ce62240530b0fc102b799be8f0fa824f7a746300)

`snprintf` is already used without checking for the return value throughout the codebase, don't those throw warnings as well?

tytan652:
> `snprintf` is already used without checking for the return value throughout the codebase, don't those throw warnings as well?

Check Ubuntu CIs, some of them throw a format-truncation warning

PatTheMav:
Fixed the most obvious mistakes - when the char array is created on the stack, `sizeof` is used, otherwise the functions have been amended to pass the size of the char array together with the pointer.

If the char array is initialised with a static preprocessor value, that one is used on both ends.

I'll look at the warnings on Linux next.

deps/libcaption/src/caption.c
notr1ch: That should be OK, I would probably also mark the function as `static` to ensure only callers in caption.c (that presumably are aware of the buffer size) can use it.

tytan652:
No more `-Wformat-truncation`, but three `-Wsign-compare` appeared (3 others are FTL). Some casts and they should be gone.

deps/libcaption/src/caption.c
PatTheMav: Removed both functions as discussed.

PatTheMav:
Added new commits fixing existing implementations that used literals as buffer length.

deps/libcaption/src/caption.c
RytoEX: Has this been resolved? Would like to move this forward.

PatTheMav: Yeah should be resolved as the corresponding functions have been removed entirely.

libobs/util/platform.c
```diff
@@ -559,7 +559,7 @@ static inline void from_locale(char *buffer)
 double os_strtod(const char *str)
 {
 	char buf[64];
-	snprintf(buf, 64, "%s", str);
+	snprintf(buf, sizeof(buf), "%s", str);
```
notr1ch, line 562: Seems like weird use of `snprintf`, `strncpy` / `strncpy_s` (if available) seem more appropriate.

plugins/obs-ffmpeg/obs-ffmpeg-vaapi.c
```diff
@@ -631,10 +631,15 @@ static obs_properties_t *vaapi_properties(void *unused)
 	if (obs_property_list_item_count(list) == 0) {
 		char path[32] = "/dev/dri/renderD1";
 		for (int i = 28;; i++) {
-			sprintf(path, "/dev/dri/renderD1%d", i);
+			snprintf(path, sizeof(path), "/dev/dri/renderD1%d", i);
 			if (access(path, F_OK) == 0) {
 				char card[128] = "Card: ";
```
notr1ch, line 636: Initializing this seems weird when we immediately overwrite it.

PatTheMav: Changed both `snprintf` invocations to use a `const char` template string and also removed the initial string assignment (as it's immediately overwritten anyway as you mentioned).

libobs/util/platform.c
notr1ch: Unfortunately `strncpy` is unsafe as-is as it doesn't null terminate on overflow. The preferred pattern is e.g. 
```
strncpy (buf, sizeof(buf)-1, str);
buf[sizeof(buf)-1] = 0;
```

plugins/obs-ffmpeg/obs-ffmpeg-vaapi.c
notr1ch: I'm personally not a fan of moving the format strings away from the `snprintf` invocation as it adds additional programmer overhead to parse what it's doing, but this solves my main concern.

PatTheMav: I wouldn't be opposed to it, I tried to surmise why it was written that way to begin with and thought the idea was to have a single "template" string that needs to be adjusted, but that train of thought also doesn't make sense.

libobs/util/platform.c
PatTheMav: Right, will change.

RytoEX (approved):
Got @notr1ch 's approval off-thread.

Will make a mental note that this modifies libcaption and librtmp, which are technically third-party dependencies even though we have them in-tree. Merging shortly.

PR #6843, opened 2022-07-28
Increase macOS deployment target to 10.15

PatTheMav:
### Description

Increases the deployment target for macOS builds to 10.15 for x86_64 and removes all guards for versions before 10.15.

Also removes now-unnecessary guards for older Mac OS X versions and removes the inject functionality from `mac-syphon` which is defunct on 10.15 anyway.

### Motivation and Context
With the impending release of macOS 13 it is important that not too much time and effort is spent on outdated macOS versions, especially if devices older than 7-8 years are able to install and run macOS 12.5 (and _should_ update).

Hardware older than that is considered "legacy" by Apple and many libraries and Frameworks have moved on to much more recent deployment targets anyway.

Builds for arm64 need to be compiled with a minimum deployment target 11.0 anyway, so our codebase in general needs to target 11.0 feature level anyway, but we'll make an effort to support 10.15 until macOS 14 comes around.

### How Has This Been Tested?
macOS compiled and run on Mac OS X 10.15 VM.

### Types of changes

### Checklist:
- [x] I have read the [**contributing** document](https://github.com/obsproject/obs-stuxio/blob/master/CONTRIBUTING.rst).


tytan652:
I don't know if mac-videotoolbox needs a change like this since https://github.com/obsproject/obs-studio/commit/5ff07c63b6b53274359d8ac54718c0206c903d08.

RytoEX (approved):
I'm fine with this in concept, but I cannot speak to the code itself. Someone more familiar with these areas will have to look over that.

PR #6848, opened 2022-07-28
Add plugin load failure warning on startup

jp9000:
### Description
If plugins fail to load on startup, a warning will now be shown on startup.

Also, if we're building with Qt6, we need to block Qt5 plugins on Windows because sadly all Windows programs are modular, thus Qt5 files from older versions may still exist if someone manually installs the Qt6 version on top of a Qt5 version. So any plugins that are detected as using Qt5 when OBS is building with Qt6 will purposely be blocked to prevent crashing.

### Motivation and Context
Because certain plugins might not be able to load with newer OBS versions (or may intentionally be blocked from loading due to compatibility in this case), we need to display a warning message telling users what plugins failed to load so they can either update or remove them. This is also helpful for plugin authors because it lets users know which plugins specifically are failing to load, so it'll at the very least prevent support issues in that regard.

### How Has This Been Tested?
Tested it by installing the advanced scene switcher (dependent on Qt5) on a Qt6 build of OBS.

### Types of changes

### Checklist:


WizardCM:
A while ago we added an exception on macOS for obs-browser to disable attempting to load it from Application Support, where we used to install it to (otherwise we would load 2 copies of CEF, one of which was sorely outdated).

Any users who have had OBS installed for a while will be given this error dialog, as we never forcefully removed the old obs-browser plugin directory from Application Support.

image

PR #6849, opened 2022-07-28
CI: Restore output of cmake configuration on Windows with Qt version

PatTheMav:
### Description
Restores the configuration output of CMake on Windows and adds the selected Qt version.

### Motivation and Context
Allows better insight into how the Windows build is configured on CI and also enables quicker check for the Qt version selected for a build.

### How Has This Been Tested?
Needs to be tested on CI.

### Types of changes

### Checklist:

### DIFF
--- a/CI/windows/02_build_obs.ps1
+++ b/CI/windows/02_build_obs.ps1
@@ -43,7 +43,7 @@ function Build-OBS {
 
     $BuildDirectoryActual = "${BuildDirectory}$(if (${BuildArch} -eq "x64") { "64" } else { "32" })"
 
-    Invoke-Expression "cmake --build ${BuildDirectoryActual} --config ${BuildConfiguration}"
+    Invoke-External cmake --build "${BuildDirectoryActual}" --config ${BuildConfiguration}
 }
 
 function Configure-OBS {
@@ -64,8 +64,7 @@ function Configure-OBS {
     $GeneratorPlatform = "$(if (${BuildArch} -eq "x64") { "x64" } else { "Win32" })"
 
     $CmakeCommand = @(
-        "-S . -B `"${BuildDirectoryActual}`"",
-        "-G `"${CmakeGenerator}`"",
+        "-G", ${CmakeGenerator}
         "-DCMAKE_GENERATOR_PLATFORM=`"${GeneratorPlatform}`"",
         "-DCMAKE_SYSTEM_VERSION=`"${CmakeSystemVersion}`"",
         "-DCMAKE_PREFIX_PATH:PATH=`"${CmakePrefixPath}`"",
@@ -90,7 +89,7 @@ function Configure-OBS {
         "$(if (Test-Path Variable:$Quiet) { "-Wno-deprecated -Wno-dev --log-level=ERROR" })"
     )
 
-    Invoke-Expression "cmake ${CmakeCommand}"
+    Invoke-External cmake -S . -B  "${BuildDirectoryActual}" @CmakeCommand
 
     Ensure-Directory ${CheckoutDir}
 }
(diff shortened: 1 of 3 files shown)


tytan652:
Why not put it after those three lines ?
```
if(NOT _QT_VERSION)
    message(FATAL_ERROR "Neither Qt5 or Qt6 were found")
  endif()
```

PatTheMav:
> Why not put it after those three lines ?
> 
> ```
>   if(NOT _QT_VERSION)
>     message(FATAL_ERROR "Neither Qt5 or Qt6 were found")
>   endif()
> ```

Because then it would be printed every time `find_qt` is being called, which ended up being spammy and drowning out other information.

CI/windows/02_build_obs.ps1
```diff
@@ -64,8 +64,7 @@ function Configure-OBS {
     $GeneratorPlatform = "$(if (${BuildArch} -eq "x64") { "x64" } else { "Win32" })"
 
     $CmakeCommand = @(
-        "-S . -B `"${BuildDirectoryActual}`"",
-        "-G `"${CmakeGenerator}`"",
+        "-G", ${CmakeGenerator}
```
RytoEX, line 67: Should this be quoted to handle generator names with spaces?

CI/windows/02_build_obs.ps1
```diff
@@ -90,7 +89,7 @@ function Configure-OBS {
         "$(if (Test-Path Variable:$Quiet) { "-Wno-deprecated -Wno-dev --log-level=ERROR" })"
     )
 
-    Invoke-Expression "cmake ${CmakeCommand}"
+    Invoke-External cmake -S . -B  ${BuildDirectoryActual} @CmakeCommand
```
RytoEX, line 92: Should `${BuildDirectoryActual}` be quoted to handle paths with spaces?

RytoEX (changes requested):
Looks okay to me other than two places where quotes might be helpful. Haven't tested locally yet.

CI/windows/02_build_obs.ps1
PatTheMav: Interestingly, no. It beaks when you wrap that Variable in quotes.

The issue is that Powershell applies its own "magic" quoting rules for command invocations. I won't pretend I entirely understood them, just know that it works when it's specified this way _exactly_ even though the string contains spaces.

CI/windows/02_build_obs.ps1
PatTheMav: I'll check this separately, I _think_ Powershell will apply magic quoting here as well, but will amend appropriately if checks prove me otherwise.

CI/windows/02_build_obs.ps1
PatTheMav: That's also why it's a separate token in the array, so Powershell considers the var (and its contents) in isolation.

CI/windows/02_build_obs.ps1
RytoEX: Tested locally with "build test" as the "BuildDirectory" parameter. While it initially succeeds at creating the directory, CMake breaks once invoked.

RytoEX:
Ah, I guess this restores _all_ the CMake output to the log? I did see the Qt version get logged, so I suppose this achieves what it set out to do. Can merge once concerns about quoted strings are addressed.

CI/windows/02_build_obs.ps1
PatTheMav: Build directory string is now quoted.

RytoEX (approved):
Pulled the latest changes and tested with a BuildDirectory path that contains spaces and everything worked as it should.

PR #6851, opened 2022-07-29
obs-websocket: CMake improvements

tytan652:
### Description

- Adds checks if submodules related to obs-websocket are initialized like obs-browser.
- Add `ENABLE_WEBSOCKET` CMake var

### Motivation and Context

Parity between in-tree submodules plugins

### How Has This Been Tested?

Avoid initialize obs-websocket and CMake should fail.
Avoid initialize obs-websocket deps submodules and CMake should fail.
Set `ENABLE_WEBSOCKET` to OFF make CMake not build obs-websocket.

### Types of changes

### Checklist:


PatTheMav:
> * Add `ENABLE_WEBSOCKET` CMake var

I don't see that in the changed files overview?

tytan652:
PR Description:
> Waiting for CMake improvements obs-websocket#969 being merged to add a submodule update commit.

plugins/CMakeLists.txt
```diff
@@ -92,4 +92,9 @@ add_subdirectory(obs-transitions)
 add_subdirectory(rtmp-services)
 add_subdirectory(text-freetype2)
 add_subdirectory(aja)
-add_subdirectory(obs-websocket)
+
+if(EXISTS ${CMAKE_CURRENT_SOURCE_DIR}/obs-websocket/CMakeLists.txt)
```
tt2468, line 96: We should have a `check_obs_websocket()` function for this just like `check_obs_browser()` since they are now our only two plugin submodules

tytan652: This function was made because this plugin is not available on FreeBSD and so we needed to the check for Windows, macOS and Linux.

tytan652:
Update submodule commit added.

plugins/CMakeLists.txt
tt2468: I think then at least moving these lines towards the top of the file would help flow. Right now it's logic followed by includes then followed again by more logic; just doesn't feel right.

tytan652: Moved the code just after the Windows/macOS/Linux/FreeBSD ifs.

PR #6856, opened 2022-07-29
UI: Fix Qt UI call from graphics thread

PatTheMav:
### Description
Removes calls to `devicePixelRatioF` from functions run on the graphics thread and use a property of passed object instead.

### Motivation and Context
Calling `devicePixelRatioF` from any thread but the main UI thread triggers thread-safety warnings at runtime on macOS, because Qt uses NSView calls to determine the value.

NSView calls are only allowed to be made from the main thread on macOS, so instead the value is stored as a property of the OBSQTDisplay at initialization, to be retrieved from the preview object later.

Static functions that do not have access to the preview object have the pixel ratio passed in their call signature.

### How Has This Been Tested?
Tested on macOS, tests on Windows pending.

### Types of changes

### Checklist:

### DIFF
--- a/UI/window-basic-main.cpp
+++ b/UI/window-basic-main.cpp
@@ -316,7 +316,9 @@ OBSBasic::OBSBasic(QWidget *parent)
 			ResizePreview(ovi.base_width, ovi.base_height);
 
 		UpdateContextBarVisibility();
+		dpi = devicePixelRatioF();
 	};
+	dpi = devicePixelRatioF();
 
 	connect(windowHandle(), &QWindow::screenChanged, displayResize);
 	connect(ui->preview, &OBSQTDisplay::DisplayResized, displayResize);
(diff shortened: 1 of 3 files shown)


PatTheMav:
Alas `DrawSelectionBox` also makes use of the pixel ratio, a call which has no access to the preview object, only the main window, so the DPI property needs to be moved into the main window class.

PatTheMav:
Good news, I found the appropriate event and tested it locally - when my screen switched from Retina mode (scale factor 2) to non-Retina (scale factor 1) the signal was handled and the scale factor was updated accordingly.

PatTheMav:
Updated the code to use the `screenChanged` event on the main window (which was already connected). This made sense insofar as the original code used the device pixel ratio of the main window anyway (and it's the only one for which we need to know the DPI of the screen it resides on).

Also the value would only need to be updated if the main window is moved to another screen with a different DPI value.

PR #6862, opened 2022-07-29
mac-capture: Add transparent background option to SCK App Capture

gxalpha:
### Description

Adds an option to SCK Application Capture to make the background transparent.
Only available on macOS 13, because on 12.5 the desktop icons (like drives, files and folders) show. Should this get fixed with a macOS 12 update (like 12.6) we'll make it available there as well.

### Motivation and Context

Users might want transparency instead of just a black background.

### How Has This Been Tested?

macOS 13.
Tested that it doesn't work on 12 by changing the guards to check for macOS 14.

### Types of changes

### Checklist:

### DIFF
--- a/plugins/mac-capture/mac-screen-capture.m
+++ b/plugins/mac-capture/mac-screen-capture.m
@@ -445,6 +445,11 @@ static bool init_screen_stream(struct screen_capture *sc)
 			includingApplications:target_application_array
 			     exceptingWindows:[[NSArray alloc] init]];
 
+		if (@available(macOS 13.0, *))
+			[sc->stream_properties
+				setBackgroundColor:CGColorGetConstantColor(
+							   kCGColorClear)];
+
 		set_display_mode(sc, target_display);
 	} break;
 	}


PatTheMav (changes requested):
Just one thing/idea:

Maybe instead of storing a boolean in the `sc` struct, let's store the color ref (in this case the result of `CGColorGetConstantColor`) and pass that value to `setBackgroundColor`.

That way we could later add a colour picker and (though it probably should be a native colour picker or we'd have to transfer the colour codes given by Qt to create CGColorRefs) and not have to change the internal structure again.

gxalpha:
Would we want to allow custom background colors in the future? I feel users could just be adding a color source behind the SCK source if they need that

PatTheMav:
> Would we want to allow custom background colors in the future? I feel users could just be adding a color source behind the SCK source if they need that

Fair point, no further objections from my side then.

PatTheMav (approved):

WizardCM (approved):

PR #6863, opened 2022-07-29
mac-capture: Don't exclude desktop windows in SCK window capture

gxalpha:
### Description

When looking for content, we exclude the desktop windows (mainly wallpaper) to keep the list in window capture smaller.
However, due to a bug with SCK on macOS 12 we have to use this same list to populate the display.
This means that we need to exclude them when the user has Window Capture selected, but include them for Display Capture so that they appears on screen there.

This PR assumes that the only reason that they are excluded is the window capture list length. @PatTheMav will still need to confirm this. Should there be other reasons for the exclusion this PR might not be a good idea.

### Motivation and Context

Fixes the desktop wallpaper missing on Display Capture on macOS 12

### How Has This Been Tested?

macOS 13
Changed the guard to macOS 14 so that I used the same method to capture the display as macOS 12 would. Without this change, the wallpaper is missing, with the change it exists.
Confirmed window list for window capture still doesn't show them.

### Types of changes

### Checklist:

### DIFF
--- a/plugins/mac-capture/mac-screen-capture.m
+++ b/plugins/mac-capture/mac-screen-capture.m
@@ -542,7 +542,8 @@ bool init_vertbuf_screen_capture(struct screen_capture *sc)
 	return sc->vertbuf != NULL;
 }
 
-static void screen_capture_build_content_list(struct screen_capture *sc)
+static void screen_capture_build_content_list(struct screen_capture *sc,
+					      bool excludingDesktopWindows)
 {
 	typedef void (^shareable_content_callback)(SCShareableContent *,
 						   NSError *);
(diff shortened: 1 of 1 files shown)


PatTheMav (approved):
That's a nice find, should've looked at the SCSharableContent class a bit closer.

It made sense to exclude desktop windows for window capture and keep it that way, because neither Application Capture nor Desktop capture required that list. With our workaround for Desktop capture in 12.5, this rears its ugly head.

Code is good without additional guards, as the usage of a window inclusion list (which has a slight negative performance impact) is guarded below macOS 13 already.

gxalpha:
Merging as per the above comment

PR #6871, opened 2022-07-30
libobs: Fix memory leak when load_module_exports failed

norihiro:
### Description

Release `mod.module` when an error is detected after the module was successfully opened.

Also add a warning message that the module is failed to load exports.

### Motivation and Context

When reading plugin files and `load_module_exports` failed, the object should be released before returning an error code but was not released. This is a memory leak.

### How Has This Been Tested?

Added a random shared object into the plugin directory and checked the module is released.

### Types of changes

### Checklist:

### DIFF
--- a/libobs/obs-module.c
+++ b/libobs/obs-module.c
@@ -126,8 +126,11 @@ int obs_open_module(obs_module_t **module, const char *path,
 	}
 
 	errorcode = load_module_exports(&mod, path);
-	if (errorcode != MODULE_SUCCESS)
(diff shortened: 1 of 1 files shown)


notr1ch:
This is intentional behavior - see the comment in `free_module`:

norihiro:
According to above comment, I close this issue.
The amount of the allocated memory would be not so large and not increasing over the time. And, ideally the failed plugin should be removed by users instead of cleanup by the code.

PR #6872, opened 2022-07-30
plugins/obs-ffmpeg: Mark rist/srt required when flag enabled

kkartaltepe:
### Description

When compiling with the new flag these are required, if we dont fail
here we instead get a more cryptic failure that the targets are
unassigned later in the file. This should make it easier to find the
flag to turn off or packages being searched for.

### Motivation and Context

Compiling obs-ffmpeg failed for an odd reason, this makes it a less odd reason.

### How Has This Been Tested?

Compiled obs-ffmpeg.

### Types of changes

### Checklist:


notr1ch:
The commit prefix should just be "obs-ffmpeg".

PR #6873, opened 2022-07-30
obs-ffmpeg: Disable audio if no audio in source

mvji:
### Description

Disable media source audio output if source doesn't have audio.

### Motivation and Context

Avoid to display audio mixer entry for media source without audio

Partially implement this idea : https://ideas.obsproject.com/posts/1125
And : 
https://ideas.obsproject.com/posts/797

### How Has This Been Tested?

Mac os 11.6 Intel
Test with media file with and without audio
- File with audio still play audio
- File without audio didn't display audio mixer entry anymore

### Types of changes

### Checklist:

### DIFF
--- a/deps/media-playback/media-playback/media.c
+++ b/deps/media-playback/media-playback/media.c
@@ -728,6 +728,9 @@ static bool init_avformat(mp_media_t *m)
 	m->has_video = mp_decode_init(m, AVMEDIA_TYPE_VIDEO, m->hw);
 	m->has_audio = mp_decode_init(m, AVMEDIA_TYPE_AUDIO, m->hw);
 
+	if (m->set_a_cb)
+		m->set_a_cb(m->opaque, m->has_audio);
+
 	if (!m->has_video && !m->has_audio) {
 		blog(LOG_WARNING,
 		     "MP: Could not initialize audio or video: "
@@ -880,6 +883,7 @@ bool mp_media_init(mp_media_t *media, const struct mp_media_info *info)
 	media->v_cb = info->v_cb;
 	media->a_cb = info->a_cb;
 	media->stop_cb = info->stop_cb;
+	media->set_a_cb = info->set_a_cb;
 	media->ffmpeg_options = info->ffmpeg_options;
 	media->v_seek_cb = info->v_seek_cb;
 	media->v_preload_cb = info->v_preload_cb;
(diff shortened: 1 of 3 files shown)


tt2468:
Similar to https://github.com/obsproject/obs-studio/commit/07b0786892c0acfdc83ab3cd6d111e651dfcf173, could we use the `s->media.has_audio` boolean to determine whether to enable audio? Disable audio mixer on source create, then let the `_start` function set the real value.

mvji:
> Similar to [07b0786](https://github.com/obsproject/obs-studio/commit/07b0786892c0acfdc83ab3cd6d111e651dfcf173), could we use the `s->media.has_audio` boolean to determine whether to enable audio? Disable audio mixer on source create, then let the `_start` function set the real value.

It's not possible to use has_audio in this function. Because in the _start function, video/audio decoder can be not start (it's started on another thread).
If we call "obs_source_set_audio_active" in ffmpeg_source_start it's doesn't work properly.
You can test by printing has video and has audio in this function, and the result will be inconsistent.

I think the commit https://github.com/obsproject/obs-studio/commit/07b0786892c0acfdc83ab3cd6d111e651dfcf173
is wrong, and should be revert. It's make strange behaviour in the loop point (probably link that has_video can be zero for file with video at this step).

PatTheMav:
IMO whether a source appears in the mixer should depend on the source capabilities, not on the internal situation of the source. The media source is by default audio-capable, so I should be able to set its volume regardless of whether a file is currently selected for playback or if such a file has no audio track.

This is especially important in cases where a media source doesn't exist in isolation (e.g. when scripting is involved or maybe even other plugins interacting with source types) and I'd rather have the user be able to set the volume to 0 (or mute the source) explicitly than relying on the state of an arbitrary media file.

@Warchamp7 WDYT

tt2468:
While I agree with Pat's opinion, the `obs_source_set_audio_active()` method is already in use by other (mainly capture) sources to enable/disable audio when audio is available. As such, this isn't exactly bringing any new UX concepts to the table.

PatTheMav:
> While I agree with Pat's opinion, the `obs_source_set_audio_active()` method is already in use by other (mainly capture) sources to enable/disable audio when audio is available. As such, this isn't exactly bringing any new UX concepts to the table.

Well as Fenrir mentioned in another PR:

> I don't want sins of the past informing future decisions

Just because we didn't pay enough attention in the past should not mean we cannot re-evaluate that for any new change.

Warchamp7:

PR #6879, opened 2022-07-31
UI: Remove spaces from translation keys

WizardCM:
### Description

Qt translation strings use the full English string as an identifier, rather than a standardised key. Meanwhile, .ini files aren't designed
for keys with spaces, so a translation string for "Restore Defaults" won't match unless the ini line is
`"Restore Defaults"="Defaults"`
which is inconsistent, unpleasant and harder to read.

This fixes the "Restore Defaults" button in Filters and Properties to actually be translated using the right key.

### Motivation and Context

In https://github.com/obsproject/obs-studio/commit/dbb063eae6a4b111ef6dabef5aeab5514d4a7b73#diff-fa7b36911cb29093d60306caac7b90c5e9d2d489b8f0ad563c4061be8d18e483R581 I changed the "Defaults" button from a Reset button with a modified label to use the built-in Qt button "Restore Defaults", not realising that the translation key wasn't being applied.

**Alternate solution**: We could simply restore the below code to the dialogs one-by-one, rather than modifying the translation function, which is called every time Qt attempts to render text.

```cpp
	ui->buttonBox->button(QDialogButtonBox::RestoreDefaults)
		->setText(QTStr("Defaults"));
```

Open to feedback on this.

### How Has This Been Tested?

Open the Properties or Filters dialog for a source.

### Types of changes

### Checklist:

### DIFF
--- a/UI/obs-app.cpp
+++ b/UI/obs-app.cpp
@@ -1720,7 +1720,9 @@ QString OBSTranslator::translate(const char *context, const char *sourceText,
 				 const char *disambiguation, int n) const
 {
 	const char *out = nullptr;
-	if (!App()->TranslateString(sourceText, &out))
+	QString str(sourceText);
+	str.replace(" ", "");
+	if (!App()->TranslateString(QT_TO_UTF8(str), &out))
 		return QString(sourceText);
 
 	UNUSED_PARAMETER(context);


UI/obs-app.cpp
```diff
@@ -1717,7 +1717,9 @@ QString OBSTranslator::translate(const char *context, const char *sourceText,
 				 const char *disambiguation, int n) const
 {
 	const char *out = nullptr;
-	if (!App()->TranslateString(sourceText, &out))
+	std::string str = string(sourceText);
+	str.erase(remove_if(str.begin(), str.end(), isspace), str.end());
+	if (!App()->TranslateString(str.c_str(), &out))
 		return QString(sourceText);
```
tytan652, line 1723: std::remove_if seems to behave differently on Linux/macOS, do we need to remove all type of whitespaces ?

Source text usually only contains spaces ` `, so only replacing spaces should be fine with whether std::string, dstr or QString.

Example with dstr:
```suggestion
	struct dstr str;
	dstr_init_copy(&str, sourceText);
	dstr_replace(&str, " ", "");
	if (!App()->TranslateString(str.array, &out))
	{
		dstr_free(&str);
		return QString(sourceText);
	}
	dstr_free(&str);
```

Add `dstr_depad(&str);` if you really want to remove other type of whitespace.

WizardCM: Just a standard space should be fine in my opinion. I'll test this out. Thanks!

UI/obs-app.cpp
```diff
@@ -1720,7 +1720,9 @@ QString OBSTranslator::translate(const char *context, const char *sourceText,
 				 const char *disambiguation, int n) const
 {
 	const char *out = nullptr;
-	if (!App()->TranslateString(sourceText, &out))
+	QString str(sourceText);
+	str.replace(" ", "");
```
tytan652, line 1724: You could use `.remove(" ")` but replace works too.

WizardCM: I'm OK with either solution. Unsure which is preferred.

tytan652: Me neither.

PR #6883, opened 2022-07-31
UI: Add WhatsNew Feature for *nix systems

derrod:
### Description

Add support for "What's New" dialogue on macOS and Linux.

ToDo:
- [x] Test on Linux
- [x] Remove temporary test URL
- [x] mbedTLS based implementation for Linux

### Motivation and Context

This feature is used to notify users about changes or important updates, and should be available on other platforms as well.

### How Has This Been Tested?

Tested on a Mac.

### Types of changes

### Checklist:


UI/nix-update/nix-update.cpp
```diff
@@ -0,0 +1,275 @@
...
+
+#ifdef BROWSER_AVAILABLE
+#include <browser-panel.hpp>
+#endif
+
+struct QCef;
+extern QCef *cef;
+
+#ifndef MAC_WHATSNEW_URL
+#define MAC_WHATSNEW_URL "https://r2.rodney.io/whatsnew.json"
```
derrod, line 21: The temporary URL https://r2.rodney.io/whatsnew.json is used since it contains the signature in the `X-Signature` header as the real server would.

UI/nix-update/crypto-helpers.hpp
```diff
@@ -0,0 +1,77 @@
...
+	0x4a, 0x77, 0x61, 0x63, 0x46, 0x36, 0x36, 0x68, 0x33, 0x59, 0x55, 0x57,
+	0x36, 0x74, 0x7a, 0x55, 0x5a, 0x68, 0x7a, 0x74, 0x63, 0x6d, 0x51, 0x65,
+	0x70, 0x50, 0x2f, 0x75, 0x37, 0x42, 0x67, 0x47, 0x72, 0x6b, 0x4f, 0x50,
+	0x50, 0x70, 0x59, 0x41, 0x30, 0x4e, 0x45, 0x4a, 0x38, 0x30, 0x53, 0x65,
+	0x41, 0x78, 0x37, 0x68, 0x69, 0x4e, 0x34, 0x76, 0x61, 0x0a, 0x65, 0x45,
+	0x51, 0x4b, 0x6e, 0x52, 0x6e, 0x2b, 0x45, 0x70, 0x42, 0x4e, 0x36, 0x55,
+	0x42, 0x61, 0x35, 0x66, 0x37, 0x4c, 0x6f, 0x4b, 0x38, 0x43, 0x41, 0x77,
+	0x45, 0x41, 0x41, 0x51, 0x3d, 0x3d, 0x0a, 0x2d, 0x2d, 0x2d, 0x2d, 0x2d,
+	0x45, 0x4e, 0x44, 0x20, 0x50, 0x55, 0x42, 0x4c, 0x49, 0x43, 0x20, 0x4b,
+	0x45, 0x59, 0x2d, 0x2d, 0x2d, 0x2d, 0x2d, 0x0a};
```
derrod, line 73: The certificate here should probably be null-terminated for mbedtls, and it would also make it easier to load on macOS since we can just read it as a C-string.

UI/nix-update/crypto-helpers-cocoa.mm
```diff
@@ -0,0 +1,58 @@
+#include "crypto-helpers.hpp"
+
+#import <Foundation/Foundation.h>
+#import <Security/Security.h>
+#import <Security/SecKey.h>
+
+void get_uuid(std::string &guid)
+{
+	NSString *str = [[NSUUID UUID] UUIDString];
```
derrod, line 9: Using UUID here instead of just a random string, does that matter for the server @notr1ch?

derrod: Actually I should probably just do what we did for YouTube OAuth and use `QRandom` as a cross-platform way of doing this, possibly with a C++ std fallback if we still want to support Qt versions < 5.10.

derrod:
Added a mbedtls based version and CMake adjustments for Linux. The mbedtls path was only checked on macOS so far however.

Now all that remains is some cleanup and adding some error logging to the crypto helpers.

derrod:
Verified working on Linux now (at least with Flatpak):
2022-08-01_18-39-46_nTRYJc

I believe this is now ready for review, the CMake changes are what I'd like to get a second opinion on the most.

CMakeLists.txt
```diff
@@ -63,6 +63,7 @@ elseif(OS_POSIX)
   if(OS_LINUX)
     option(ENABLE_WAYLAND "Enable building with support for Wayland (Linux)" ON)
     option(BUILD_FOR_PPA "Build for PPA distribution" OFF)
+    option(ENABLE_WHATSNEW_LINUX "Enable WhatsNew dialog on Linux" OFF)
```
tytan652, line 66: If it on by default and not disable-able on Windows and macOS, I think defaulting it to on is fine and even remove the CMake option.

UI/nix-update/nix-update.cpp
```diff
@@ -0,0 +1,299 @@
...
+#endif
+
+#ifndef LINUX_WHATSNEW_URL
+#define LINUX_WHATSNEW_URL "https://r2.rodney.io/whatsnew.json"
```
tytan652, line 27: Just a `don't forget to change this`.

UI/obs-app.cpp
```diff
@@ -551,7 +551,9 @@ static bool MakeUserDirs()
 		return false;
 	if (!do_mkdir(path))
 		return false;
+#endif
 
+#if defined(_WIN32) || WHATSNEW_ENABLED
```
RytoEX, line 556: Can you link me to another place with this syntax? I believe @PatTheMav prefers `defined(WHATSNEW_ENABLED)` over just dropping the raw definition in place.

CMakeLists.txt
derrod: I kept it off becuase enabling it causes a hard-fail if OBS is compiled without MbedTLS. I could perhaps do something like we do for RTMPS which defaults to an "AUTO" option and then only hard fail if it has been manually set to "ON".

RytoEX (changes requested):
@PatTheMav is probably better suited to review the CMake stuff, but I took a quick look anyway.

UI/CMakeLists.txt
```diff
@@ -348,7 +348,7 @@ if(OS_WINDOWS)
 
   target_compile_definitions(
     obs PRIVATE UNICODE _UNICODE _CRT_SECURE_NO_WARNINGS
-                _CRT_NONSTDC_NO_WARNINGS)
+                _CRT_NONSTDC_NO_WARNINGS WHATSNEW_ENABLED)
```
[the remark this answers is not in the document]
derrod: You cannot disable it. Disabling the Windows updater has been removed from CMake after the infamous OBS 21.1 incident.

UI/CMakeLists.txt
```diff
@@ -421,9 +423,16 @@ elseif(OS_MACOS)
...
   target_sources(obs PRIVATE platform-osx.mm)
+  target_sources(
+    obs
+    PRIVATE nix-update/crypto-helpers.hpp nix-update/crypto-helpers-mac.mm
+            nix-update/nix-update.cpp nix-update/nix-update.hpp
+            nix-update/nix-update-helpers.cpp nix-update/nix-update-helpers.hpp)
   target_sources(obs PRIVATE forms/OBSPermissions.ui window-permissions.cpp
                              window-permissions.hpp)
 
+  target_compile_definitions(obs PRIVATE WHATSNEW_ENABLED)
```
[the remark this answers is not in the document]
derrod: Yes, as this code is always compiled. It's only conditional on Linux where mbedtls may not be present.

UI/obs-app.cpp
derrod: This was a mistake, it was just using `#ifdef ..` before and I forgot the second `defined()`

UI/nix-update/nix-update.cpp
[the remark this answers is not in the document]
derrod: Changed now. Can be overridden via `LINUX_WHATSNEW_URL` or `MAC_WHATSNEW_URL` for testing. `https://r2.rodney.io/whatsnew.json` provides the last 27.2.4/27.2.3 file and a valid signature for testing, so make sure your OBS build is set to one of these versions.

CMakeLists.txt
tytan652: Forgot about this. Let's make it simple, option with default as off as it is is simple.

UI/nix-update/crypto-helpers-mac.mm
```diff
@@ -0,0 +1,50 @@
+#include "crypto-helpers.hpp"
+
+#import <Foundation/Foundation.h>
+#import <Security/Security.h>
+#import <Security/SecKey.h>
+
+bool VerifySignature(const uint8_t *buf, const size_t len, const uint8_t *sig,
+		     const size_t sigLen)
+{
+	NSString *pubKeyStr =
```
PatTheMav, line 10: Have you looked at https://developer.apple.com/documentation/security/1395728-secitemimport?language=objc which supports `kSecFormatPEMSequence`?

PatTheMav:
That's a whole lot of code just to show a "What's new" page - what's the reason for the dependency on mbedTLS and shipping a public key in code(!) for this?

UI/nix-update/crypto-helpers-mac.mm
derrod: I have not, I probably disregarded it while reading documentation because of the "Import" part, but it looks like you can use it to load a key without actually importing it. I'll give it a shot tomorrow.

derrod:
> That's a whole lot of code just to show a "What's new" page - what's the reason for the dependency on mbedTLS and shipping a public key in code(!) for this?

This is based on the Windows implementation, which contains the same signature verification: https://github.com/obsproject/obs-studio/blob/master/UI/win-update/win-update.cpp

I think @notr1ch said that this was basically to prevent someone compromising the OBS server from publishing a malicious entry that would get pushed to users.

On Linux it was a matter of either adding a dependency on something like `libkcapi` (Linux Kernel Crypto API wrapper), or using MbedTLS which we already use for RTMPS, so that seemed like the least worst choice.

> We'd also need to differentiate between "news" and actual update changelogs, the latter of which should appear in the Sparkle update notification instead.

"What's New" is only shown after an update to a new version, whereas the changelog in Sparkle or the Windows updater is shown before. It is generally used to highlight specific larger changes or notify users about important updates (e.g. Twitch API deprecation requiring an OBS update). We've also used it to advertise the Survey to users for example.

tytan652:
> what's the reason for the dependency on mbedTLS

On the Linux side there is multiple libs for that (gnutls, OpenSSL) but nothing "by default", we already use mbedTLS for RTMPS.

PatTheMav:
FWIW we already ship OBSPublicDSAKey.pem on macOS for Sparkle (_if_ it's enabled in CMake), so we could also ship another PEM as well (if it's not the same key).

derrod:
Hardcoding the key does seems somewhat archaic, but I don't know if there's a good way to read a file at compile time (`std::embed` seems to be stuck in limbo, and `#embed` will only come with C23),

That may be a question for @jp9000 as to why this is the way it is and how we should address it.

derrod:
Updated now to use `SecItemImport` for Mac, as well as some switches to Qt utilities to replace some old functions.

RytoEX:
Looks fine at a glance to me. If @notr1ch and @PatTheMav think the implementation is good enough for now and further improvements/changes can be done later, then we just need to decide when we actually want to merge this.

PatTheMav:
> Updated now to use `SecItemImport` for Mac, as well as some switches to Qt utilities to replace some old functions.

I checked "OBSPublicDSAKey.pem" that we ship in the Resources subdirectory, and it's the same key. I'd rather use that key then a hardcoded one. Also the feature requires obs-browser to be present, so configuration should fail if `ENABLE_BROWSER` is disabled for conflicting configuration parameters.

derrod:
I don't feel particularly strongly about hardcoding the key either way, so unless there are any objections from R1CH or Jim I can look into loading it from the file. Although that may require additional modifications to CMake/build scripts to include the file in Linux distributions as well.

What's New is - weirdly - not guarded behind the browser being available on Windows, but it's true that it doesn't really make sense for it to be enabled if it isn't, so I'll update that.

derrod:
Okay implemented the following changes:

* WhatsNew is now only enabled if browser panels are enabled
  + On Windows/macOS it is implicity enabled
  + On Linux is has to be explicitly enabled as it requires MbedTLS and we probably don't want third-party distributions to have it on by default
* Added updater public key to `UI/data/OBSPublicRSAKey.pem`
* Public key is now loaded from disk (Windows implementation remains unchanged)

> I checked "OBSPublicDSAKey.pem" that we ship in the Resources subdirectory, and it's the same key. 

How did you check? Because they're clearly not the same key. The existing pem file also is a DSA key rather than RSA.

CMakeLists.txt
PatTheMav, line 66: I don't think we need an additional flag for this - make `ENABLE_WHATSNEW` an `INTERNAL` cache variable (set to "ON") on macOS and Windows, make it an `option` on `OS_LINUX`, also defaulting to "ON".

That would make the changes above (explicitly enabling it in the build script and for FlatPak) unnecessary and forces other package maintainers to explicitly turn it off (as desired for Linux).

UI/CMakeLists.txt
```diff
@@ -363,6 +363,10 @@ if(OS_WINDOWS)
     target_link_options(obs PRIVATE /LARGEADDRESSAWARE)
   endif()
 
+  if(TARGET OBS::browser-panels)
```
PatTheMav, line 366: Make this global for all OSes and guard behind `TARGET OBS::browser-panels` and `ENABLE_WHATSNEW`.

UI/CMakeLists.txt
```diff
@@ -424,6 +428,22 @@ elseif(OS_MACOS)
   target_sources(obs PRIVATE forms/OBSPermissions.ui window-permissions.cpp
                              window-permissions.hpp)
 
+  if(TARGET OBS::browser-panels)
```
PatTheMav, line 431: See above.

UI/CMakeLists.txt
```diff
@@ -444,6 +464,33 @@ elseif(OS_POSIX)
   if(OS_FREEBSD)
     target_link_libraries(obs PRIVATE procstat)
   endif()
+
+  if(OS_LINUX AND ENABLE_WHATSNEW_LINUX)
```
PatTheMav, line 468: Accordingly just check for `ENABLE_WHATSNEW` here.

UI/obs-app.cpp
```diff
@@ -551,7 +551,9 @@ static bool MakeUserDirs()
 		return false;
 	if (!do_mkdir(path))
 		return false;
+#endif
 
+#if defined(_WIN32) || defined(WHATSNEW_ENABLED)
```
PatTheMav, line 556: With `WHATSNEW_ENABLED` treated as a global define (see above), the specific check for `_WIN32` could be removed.

deps/CMakeLists.txt
```diff
@@ -6,6 +6,8 @@ if(OS_WINDOWS)
 
   add_subdirectory(blake2)
   add_subdirectory(lzma)
+elseif(OS_MACOS OR ENABLE_WHATSNEW_LINUX)
```
PatTheMav, line 9: Could be simplified to just check for `ENABLE_WHATSNEW`.

UI/CMakeLists.txt
derrod: I would keep `OS_LINUX` here to avoid issues with FreeBSD at least.

CMakeLists.txt
derrod: Where would be the most appropiate place to set the internal cache variable? `ObsDefaults_{Windows,macOS}` or just at the top of this file?

derrod:
@tytan652 @GeorgesStavracas Opinions on making this on-by-default for Linux? For RTMPS we use an "auto" setting in case MbedTLS isn't available, so I'm thinking we could mirror this here unless we want package maintainers to have to opt-out explicitly.

GeorgesStavracas:
Didn't give a deep review, but from a quick gland, I don't think there should be any build flags controlling this. I think it should always be enabled if a browser source and mbedtls are available.

tytan652:
> Opinions on making this on-by-default for Linux? For RTMPS we use an "auto" setting in case MbedTLS isn't available, so I'm thinking we could mirror this here unless we want package maintainers to have to opt-out explicitly.

I do want package maintainers to have to opt-out explicitly so `ON` by default. RTMPS may need to have a Linux only default change but it's out of scope.

If the maintainer already disabled the browser plugin, this feature should be disabled with it.
Or if the packager don't want to link to mbedTLS at all, a flag is needed to disable the feature.

derrod:
Alright, we'll go with enabled by default then. Thanks for your input!

derrod:
Updated based on the latest review:
- `ENABLE_WHATSNEW` is now set as an option on Linux/internal cache on macOS/Windows when browser panels are available
- `WHATSNEW_ENABLED` definition is now set when it is enabled
- Linux builds have it enabled by default, will fail if MbedTLS is unavailable
- Fixed some ifdefs that were still using `_WIN32`

UI/CMakeLists.txt
```diff
@@ -444,6 +471,25 @@ elseif(OS_POSIX)
...
     target_link_libraries(obs PRIVATE procstat)
   endif()
+
+  if(OS_LINUX AND ENABLE_WHATSNEW)
+    find_package(MbedTLS)
+    if(NOT MBEDTLS_FOUND)
+      obs_status(
+        FATAL_ERROR
+        "mbedTLS not found, but required for WhatsNew support on Linux")
+    else()
```
PatTheMav, line 481: Triggering a message with `FATAL_ERROR` acts like a `throw` in CMake - the `else` is not strictly necessary.

derrod: This one must've fallen through the cracks during some refactoring, I removed the else now.

PatTheMav:
Finally was able to test this (took a bit of futzing around with the config for it to actually show anything) and got a crash when closing the browser window.

Looks like it requires obs-browser to be updated to include https://github.com/obsproject/obs-browser/pull/374 before this can be merged.

derrod:
> Finally was able to test this (took a bit of futzing around with the config for it to actually show anything) and got a crash when closing the browser window.
> 
> Looks like it requires obs-browser to be updated to include [obsproject/obs-browser#374](https://github.com/obsproject/obs-browser/pull/374) before this can be merged.
> 
> EDIT: It's also unfortunate that users are greeted with an empty window at first (CEF showing a blank, white page) instead of waiting for CEF to having finished the page and _then_ opening the Qt window, but I'm afraid that would require rewriting the feature itself.
> 
> FWIW, it takes so long to load for me that I can see some people closing it before actually reading the contents.

Yup since your last version is presumably already set to 27.2.4 and the "WhatsNew" dialog is for 27.2.4 it won't show unless you remove the `LastVersion*` entries from the config first.

Weirdly I never had any crashes on my system, and it also didn't take particularly long to load, but I also already had a browser source active in my scenes, so perhaps my CEF was already "warmed up"?

Looks like the submodule has already been updated to include https://github.com/obsproject/obs-browser/pull/374 in master.

RytoEX:
Successfully built and tested on Ubuntu 22.04 running Xorg. What's New dialog pops up successfully. There seems to be a _short_ UI hang for 1-2 seconds.

RytoEX (approved):
At a glance, seems fine. I could go either way on "WhatsNew dialog" vs. "What's New dialog" in the CMake strings, so I won't nit or block there. Tested earlier on a native build on Ubuntu 22.04 using Xorg, per my comment above. I think this is good enough as-is and we can iterate and make improvements later if needed.

PR #6894, opened 2022-08-01
mac-capture: Fix macOS 12 SCK Display Capture workaround

### LINKED ISSUE #6893: NEW macOS Screen Capture (one using ScreenCaptureKit) doesn't capture Powerpoint presentation correctly.
### Operating System Info

macOS 12

### Other OS

_No response_

### OBS Studio Version

Git

27.2.0-862-g7ec6f0a66

### OBS Studio Log URL

https://obsproject.com/logs/b-dQqxqq2UbLS0eD

### OBS Studio Crash Log URL

_No response_

### Expected Behavior

Fullscreen Powerpoint presentation should be recorded like this. (This works on deprecated "Display Capture" mode)

Greenshot 2022-08-01 19 00 15

> note: this "expected behavior" is captured using deprecated "Display Capture" source.

### Current Behavior

https://youtu.be/dZ_rbE3XtGM

Only desktop image is recorded.

Greenshot 2022-08-01 19 00 21

### Steps to Reproduce

1. Attach Second Display. (I'm using [BetterDummy](https://github.com/waydabber/BetterDisplay) for that purpose)
2. Run any powerpoint presentation
3. Control surface should appear on the first display, and **the presentation itself should appear on the second display**.
4. Record the second display via macOS Screen Capture - Display Capture.

image

- macOS 12.5
- Powerpoint version 16.63.1 (22071401, Microsoft Office 365 subscription)

gxalpha:
### Description

Adding all windows manually means that any windows opened after the
session was started won't appear. This instead excludes the WindowServer
application from the list, which isn't needed anyways.

### Motivation and Context

Display Capture would not show windows opened after the session (OBS) was started, which is bad.
Possibly fixes #6893, though I'm not 100% sure about that.

### How Has This Been Tested?

Only tested on macOS 13 that using this method, windows show up even if they're opened *after* the capture is started.
Needs testing on macOS 12 to confirm that this doesn't spawn the original bug we're trying to work around in the first place.

### Types of changes

### Checklist:


plugins/mac-capture/mac-screen-capture.m
```diff
@@ -377,9 +377,30 @@ static bool init_screen_stream(struct screen_capture *sc)
 				 initWithDisplay:target_display
 				excludingWindows:[[NSArray alloc] init]];
 		} else {
+			__block SCRunningApplication *excluded;
+			[sc->shareable_content.applications
```
PatTheMav, line 381: Isn't `applications` already an NSArray? Instead of enumerating, extracting a single item and then creating a new NSArray with that single item, `NSArray` should have a block-based filter that returns an NSArray with just the item you want.

Then you can pass this filtered copy of the NSArray directly into the next call.

plugins/mac-capture/mac-screen-capture.m
```diff
@@ -372,14 +372,34 @@ static bool init_screen_stream(struct screen_capture *sc)
 	case ScreenCaptureDisplayStream: {
 		SCDisplay *target_display = get_target_display();
 
-		if (@available(macOS 13.0, *)) {
+		if (@available(macOS 14.0, *)) {
```
derrod, line 375: Is this supposed to say 14 here?

CarrotCodes:
Display Capture doesn't appear to capture anything on macOS 12.5 with 8e9f62ee:

Screenshot 2022-08-01 at 19 48 51

Other SCK methods still work as expected though.

gxalpha:
Yeah the fix unfortunately didn’t work as expected (because of differences between macOS 12 and 13). I’ll push another attempt later. Thanks for testing.

gxalpha:
This changes the PR to exclude Control Center instead. While it's unfortunate that it won't appear in display capture on macOS 12 like this, it seems to be pretty much the only app (besides dock/wallpaper) that's always active (even `SystemUIServer` can be missing from the list when Siri and Time Machine are disabled).

CarrotCodes:
Tried 04de7bdb and it worked for me on macOS 12.5 👍 

I opened OBS, added the Display Capture, made some new Terminal windows, and dragged them in to frame. Previously, the new windows didn't show up in the capture, which was odd.

plugins/mac-capture/mac-screen-capture.m
```diff
@@ -575,9 +593,12 @@ typedef void (^shareable_content_callback)(SCShareableContent *,
 	os_sem_wait(sc->shareable_content_available);
 	[sc->shareable_content release];
 	[SCShareableContent
-		getShareableContentExcludingDesktopWindows:excludingDesktopWindows
-				       onScreenWindowsOnly:!sc->show_hidden_windows
-					 completionHandler:new_content_received];
+		getShareableContentExcludingDesktopWindows:TRUE
```
PatTheMav, line 596: This will remove Finder, Dock, and Desktop from the list of shareable content - is that intended? Finder at the very least should be captureable.

gxalpha: As discussed off-thread, this reinstates the code as it was before #6863, both on 12.5 and 13 Finder and Dock are capturable. The only thing missing is the wallpaper (and that's intentional).

PatTheMav (approved):

PR #6898, opened 2022-08-01
CI: Update Steam workflow for Apple Silicon builds

derrod:
### Description

Updates workflow to process both ARM and x86 builds for macOS.

ToDo:
- [x] Adjust to final DMG names
- [x] Push updates to legacy branches to include `launch.sh` so they keep with the default launch configuration
- [ ] (optional) command line option to override auto-detection?

### Motivation and Context

Users will probably want to use native builds.

### How Has This Been Tested?

Previous version has been tested on playtest app on my own repo.

### Types of changes

### Checklist:


.github/workflows/steam.yml
```diff
@@ -79,7 +82,8 @@ jobs:
 
             ASSETS="$(curl -s "${ASSETS_URL}")"
             WIN_ASSET_URL="$(jq -r '.[] | select(.name|test(".*x64.zip")) .browser_download_url' <<< ${ASSETS})"
-            MAC_ASSET_URL="$(jq -r '.[] | select(.name|test(".*.dmg")) .browser_download_url' <<< ${ASSETS})"
+            MAC_ASSET_URL="$(jq -r '.[] | select(.name|test(".*x86_64.dmg")) .browser_download_url' <<< ${ASSETS})"
+            MAC_ARM_ASSET_URL="$(jq -r '.[] | select(.name|test(".*arm64.dmg")) .browser_download_url' <<< ${ASSETS})"
```
derrod, line 86: Needs adjusted for final names that are going to be used in releases giong forward.

CI/steam/scripts_macos/launch.sh
```diff
@@ -0,0 +1,16 @@
...
+ 
+arch_name="$(uname -m)"
+
+# legacy app installation
+if [ -d OBS.app ]; then
+    exec open OBS.app -W --args "$@"
+fi
+
+if [ "${arch_name}" = "x86_64" ]; then
+    exec open x86/OBS.app -W --args "$@"
```
PatTheMav, line 11: `open` usually exits when `dyld_loader` has finished starting the application - this would usually mean that the shell script exits immediately. Dunno if Steam then still considers the app to be "running".

derrod: That's what `-W` is for ("Wait until the applications exit").

Steam doesn't track process trees on Mac it would otherwise consider the app to be closed unfortunately.

PR #6907, opened 2022-08-02
UI/updater: Add update channels

derrod:
### Description

This PR adds the option to select and update channel in the OBS Studio settings.

This implementation is based on https://github.com/obsproject/rfcs/pull/48

2022-08-02_14-20-40_8XF7Ww

ToDo:
- [ ] UI/UX feedback
- [ ] Update server-side components
- [x] Cache list of branches after first load

### Motivation and Context

We want users to be able to opt-into branches for easy switching between stable/beta and possibly test or nightly builds in the future.

### How Has This Been Tested?

UI parts have been tested with a placeholder `branches.json` in `%APPDATA%\obs-studio\update`:

<details>
<summar>Example JSON</summary>

```json
[
  {
    "description": "Semi-stable test builds",
    "display_name": "Beta / Release Candidates",
    "enabled": false,
    "macos": false,
    "name": "beta",
    "visible": true,
    "windows": true
  },
  {
    "description": "Unstable/Unsigned nightly builds",
    "display_name": "Nightly",
    "enabled": true,
    "macos": false,
    "name": "nightly",
    "visible": true,
    "windows": true
  },
  {
    "description": "Test Build to Fix issue #1234",
    "display_name": "Test, Issue #1234",
    "enabled": true,
    "macos": false,
    "name": "test_1234",
    "visible": false,
    "windows": true
  }
]
```

</details>

### Types of changes

### Checklist:


UI/win-update/updater/updater.cpp
```diff
@@ -1188,6 +1189,8 @@ static bool Update(wchar_t *cmdLine)
 			for (int i = 0; i < argc; i++) {
 				if (wcscmp(argv[i], L"Portable") == 0) {
 					bIsPortable = true;
+				} else {
+					branch = argv[i];
```
derrod, line 1193: I'm not sure if this would work as intended (not familiar with low level argument parsing). Perhaps the branch name could just be prefixed with something to be sure (e.g. `-B<branch>`)?

notr1ch: I think in general there needs to be better command line parsing in the updater, it's also needed for another feature I want where OBS passes the manifest path in the command line to allow updating by non-admin users. Perhaps requiring key=value is the way to go, though for compatibility we have to consider the updater can be ran by any older OBS version :(.

derrod: Perhaps a bit jank, but for simple `--<key>=<value>` matching couldn't we do something like this:
```cpp
wstring branch;
wstraing manifest_path;

for (int i = 0; i < argc; i++) {
    // Legacy OBS
    if (wcscmp(argv[i], L"Portable") == 0) {
        bIsPortable = true;
        break;
    }
    
    if (wcsncmp(argv[i], L"--branch=", 9) == 0) {
        branch = argv[i] + 9;
    } else if (wcsncmp(argv[i], L"--manifest=", 11) == 0) {
        manifest_path = argv[i] + 11;
    }
}
```

notr1ch: That should be OK, assuming CommandLineToArgvW handles quoted paths with spaces in the name.

derrod: Based on https://docs.microsoft.com/en-us/windows/win32/api/shellapi/nf-shellapi-commandlinetoargvw it should:

> CommandLineToArgvW treats whitespace outside of quotation marks as argument delimiters.

With the special note that a quotation mark immediately preceeded by a backslash is considered to be part of a file path, and not an escape sequence.

derrod: I've implemented this in https://github.com/obsproject/obs-studio/pull/6907/commits/c9400cc5fa7e0d512777aa8686a7894e95fd1d92 now, I'm not too big a fan of what's needed to convert from a `std::string` to `std::wstring`, unfortunately I may have to go the slightly uglier route of dealing with the raw string and OBS's utilities for this.

derrod: > warning STL4017: std::wbuffer_convert, std::wstring_convert, and the <codecvt> header (containing std::codecvt_mode, std::codecvt_utf8, std::codecvt_utf16, and std::codecvt_utf8_utf16) are deprecated in C++17. (The std::codecvt class template is NOT deprecated.) The C++ Standard doesn't provide equivalent non-deprecated functionality; consider using MultiByteToWideChar() and WideCharToMultiByte() from <Windows.h> instead. 

Well fuck me then.

derrod: Alright, I think I have fixed it now.

Not sure how I would debug the updater itself quite yet, but I'm positive that at least will work.

derrod:
I've made a few changes once again to clean up some things. If anybody has a better idea than keeping track of available branches in `OBSApp` I'm happy to hear them.

Right now I believe there's also the slim chance of a race condition if the settings window is opened at the same time as the fetch thread finishes. Not sure if it's worth introducing a mutex there or if the risk associated with this is ultimately neglibible.

derrod:
After some refactoring this PR is now down to a net gain of 300 lines. I have consolidated all signature-verified HTTP requests in the update code into one function and added a notification if a user's selected branch is no longer available.

I have not yet verified that the UI is not broken on Mac/Linux when the Windows-specific elements get `delete`d.

derrod:
I'm planning to do a full test setup this week, there's also still something missing for checking the beta/RC revision. Also not sure what to do for nightlies, perhaps commit hash?

derrod:
Added a WIP commit for relase/beta builds so they can be differentiated when checking for updates. This still doesn't address what to do for nightlies if we want to provide them. I'm considering that we could use a 64-bit unix timestamp that gets set when it is built instead of the regular OBS version for the internal check when dealing with nightlies.

derrod:
Added `OBS_COMMIT` as a new define that can be used to check if there's a new update instead. This only applies if the commit hash is set within the manifest, which I expect will only be done for non-release builds (e.g. nightlies, test builds, etc.)

derrod:
I've done a full set of tests today with the following:
- Updating from "stable" to "beta" (with deltas)
- Updating from "beta" to "nightly" (without deltas)
- Downgrading from "beta" to "stable" (without deltas)

In doing so I hopefully fixed most of the outstanding bugs. There are some things we may have to consider related to downgrading as newer versions may contain additional files that would not get deleted unless the manifests were updated to include those files in the delete list.

Potential options here:
- Update the stable manifest when new things are added to beta builds (e.g. new plugins)
- Make it possible for the updater to read both the new and installed manifest, and decide deletions based on that

derrod:
Made the following changes today so this PR can be considered "finished" and ready for review:

- Rebased on latest master
- Squashed all related commits together
- Fixed up getting the commit hash from git describe
- Added `windows`/`macos` properties to branches
- Removed manifest filename from branches list (now created programmatically)

The latter are smaller changes in preparation for macOS support, which will be a separate PR based on this branch since it requires larger changes such as updating to Sparkle 2. That way it can be reviewed and worked on separately and won't block the Windows side of things.

UI/data/locale/en-US.ini
```diff
@@ -266,6 +266,8 @@ Updater.Running.Title="Program currently active"
 Updater.Running.Text="Outputs are currently active, please shut down any active outputs before attempting to update"
 Updater.NoUpdatesAvailable.Title="No updates available"
 Updater.NoUpdatesAvailable.Text="No updates are currently available"
+Updater.BranchNotFound.Title="Update channel removed"
```
notr1ch, line 269: Dialog titles should be Title Case.

derrod: I changed it, although it should be noted that none of the others are Title Case (except the integrity check ones I added earlier this year).

notr1ch: Yeah it's something I hate about our dialogs. One day I'm gonna go through them all...

derrod:
In preparation for macOS support I've added a commit to merge `win-update` and `nix-update` into simply `update`. This includes a CMake change to use MbedTLS instead of Windows's crypto APIs as well as some general changes.

This has not been tested on Linux/macOS yet.

WizardCM:
TODO: Test with disabled updater.

derrod:
> TODO: Test with disabled updater.

Currently this doesn't do anything, I'll see if I can update it to hide/remove the new Update section when the flag is set.

derrod:
Rebased on master plus a small behavioural change to disallow downgrades on the stable branch and fix some whitespace formatting.

This change is mainly so that people installing an RC manually don't immediately see a prompt to update (downgrade) to stable. A better solution to this problem would be to automatically set the selected channel to `beta` for beta/RC builds, but that is a change I'd prefer to add to #7723 since the large amount of refactoring and code cleanup it contains would essentially require me to do that twice otherwise.

cmake/Modules/VersionConfig.cmake
```diff
@@ -32,6 +32,13 @@ if(NOT DEFINED OBS_VERSION_OVERRIDE)
         string(REPLACE "." ";" _CANONICAL_SPLIT ${_CANONICAL_SPLIT})
         list(GET _CANONICAL_SPLIT 0 1 2 _OBS_VERSION_CANONICAL)
         string(REPLACE "." ";" _OBS_VERSION ${_OBS_VERSION})
+        # Get 8-character commit hash without "g" prefix
+        foreach(VERSION_PART ${_CANONICAL_SPLIT})
+          if(VERSION_PART MATCHES "^g")
+            string(SUBSTRING ${VERSION_PART}, 1, 8, OBS_COMMIT)
+            break()
+          endif()
+        endforeach()
```
RytoEX, line 41: We use 9-character hashes in CI. Do we want to use 9-characters here?

RytoEX (changes requested):
Some items from a review I'd forgotten to submit.

cmake/Modules/VersionConfig.cmake
derrod: I cannot remember if this had a specific reason behind it. In theory it shouldn't matter.

UI/win-update/updater/updater.cpp
```diff
@@ -1292,10 +1295,17 @@ static bool Update(wchar_t *cmdLine)
 		if (argv) {
 			for (int i = 0; i < argc; i++) {
 				if (wcscmp(argv[i], L"Portable") == 0) {
+					// Legacy OBS
+					bIsPortable = true;
+					break;
+				} else if (wcsncmp(argv[i], L"--branch=", 9) ==
+					   0) {
+					branch = argv[i] + 9;
+				} else if (wcscmp(argv[i], L"--portable") ==
```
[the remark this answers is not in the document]
derrod: `argv[i]` is a null terminated string, for `--branch=` this contains both the name of the parameter as well as its value, so we only want to compare the first 9 characters. But since `--portable` is a flag we can simply compare the whole string and do not need to specify a maximum length.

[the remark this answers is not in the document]
RytoEX: Fair enough.

cmake/Modules/VersionConfig.cmake
RytoEX: Can probably be adjusted later anyway if need be.

RytoEX (approved):
As expressed off-thread, I think we should merge this and iterate upon it as needed. If it proves problematic, we can revert it, make adjustments, and reland later.

derrod:
> As expressed off-thread, I think we should merge this and iterate upon it as needed. If it proves problematic, we can revert it, make adjustments, and reland later.

Should I wait to hear back from @notr1ch regarding server-side path changes?

Also here is the `branches.json` file that should be signed and uploaded before the first release with this feaeture is published (inside a ZIP because github doesn't allow me to upload `.json` directly???):
[branches.zip](https://github.com/obsproject/obs-studio/files/10051259/branches.zip)

PR #6908, opened 2022-08-02
rtmps-services: Add fallback to H264 and some tweaks and fixes

tytan652:
### Description

Make optional `supported video codecs` and `supported audio codecs` by adding a fallback to H264 in `rtmp-services`.
`supported audio codecs` is not used for now so no fallback to AAC.

Remove field in the JSON that have only H264 or only AAC.

Opus ones are kept but are not used at all.

Also fixes incompatible-ponter-types warnings.

### Motivation and Context

Some services don't have those fields which cause issue and those should be optional.

### How Has This Been Tested?

Service without those field are now considered compatible only with H264.

### Types of changes

### Checklist:

### DIFF
--- a/plugins/rtmp-services/data/services.json
+++ b/plugins/rtmp-services/data/services.json
@@ -6,12 +6,6 @@
             "name": "Twitch",
             "common": true,
             "stream_key_link": "https://dashboard.twitch.tv/settings/stream",
-            "supported video codecs": [
-                "h264"
-            ],
-            "supported audio codecs": [
-                "aac"
-            ],
             "servers": [
                 {
                     "name": "Asia: Hong Kong",
@@ -214,9 +208,6 @@
                 "h264",
                 "hevc"
             ],
-            "supported audio codecs": [
-                "aac"
-            ],
             "servers": [
                 {
                     "name": "Primary YouTube ingest server",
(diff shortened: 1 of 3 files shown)


plugins/rtmp-services/data/services.json
```diff
@@ -6,12 +6,6 @@
             "name": "Twitch",
             "common": true,
             "stream_key_link": "https://dashboard.twitch.tv/settings/stream",
-            "supported video codecs": [
-                "h264"
-            ],
-            "supported audio codecs": [
-                "aac"
-            ],
```
gxalpha, line 14: We just marked this field as required in our service submission guidelines, I’m pretty sure it’s intentionally explicit.

tytan652: This PR is meant to make them optional, because RTMP only use AAC. RTMP and FTL only use H264.

Those should maybe be required if the protocol is not RTMP or FTL, but RFC 45 is not reviewed so I'm stuck for PRing better.

And also I think everyone is not on the same wavelength about those that were added recently. Those are not required if we care to "add fallback"/"consider default" to H264 and AAC.

jp9000: I added the entries, it's not a big deal to make it optional imho. I'm fine with it for now. This is a relatively harmless file to change it later if we end up feeling different later. Plus I really don't want to block the beta over this.

PR #6915, opened 2022-08-02
obs-scripting: Fix issues between runtime and compile-time versions

### LINKED ISSUE #6897: Python: On Windows, Python 3.6 crashes on launch
### Operating System Info

Windows 11

### Other OS

_No response_

### OBS Studio Version

Git

27.2.0-813-g7bfc7e65f

### OBS Studio Log URL

https://gist.github.com/flaeri/bf75a3804ccef2fbe9142497b70bab8d

### OBS Studio Crash Log URL

https://gist.github.com/flaeri/b5a0d6bf5cdfc163952940fe5fb4440e

### Expected Behavior

I expect on OBS Windows will not crash on startup with Python 3.6 path set.

### Current Behavior

It crashes on `7bfc7e65f` and later commits, if python 3.6 is set in the global.ini file.

### Steps to Reproduce

1. Download Python 3.6
<https://www.python.org/ftp/python/3.6.8/python-3.6.8-embed-amd64.zip>
2. Unzip to C:\python36
3. Add it to your path via tools > scripts > python path
4. Observe silent crash

If you upgrade from a version prior, IE the py36 path is already set, you will crash on launch (see attached crashlog).

This happens on python 3.6.0 and 3.6.8. I've tried the installer as well as the zip, no change.

Happens with and without any python scripts loaded.

Earliest commit that causes this: 7bfc7e65f4
You can download it here: <https://github.com/obsproject/obs-studio/actions/runs/2763999685>

Also happens on latest master/nightly for the past few days.

Building these older commits is now quite tricky due to recent changes (C2220, linker errors to py), so I've attached the stack trace in VS via a more recent commit c2122b0

image

PatTheMav:
### Description
Fixes crashing bug when OBS is configured to load Python 3.6.

### Motivation and Context
Calling `PyEval_InitThreads` has been deprecated in Python 3.7 and the function itself will be removed in Python 3.11. The current check guards this function behind a version check that only happens at compile time.

This in turn leads to crashes when run on Python 3.6, as the necessary initialization for `PyEval_ReleaseThread` did not take place.

This commit ensures the manual initialization takes place based on the runtime version of Python and avoids loading the associated symbols on Python 3.9 or later.

**Note:** Once Python 3.11 is out officially, we probably have to drop support for Python 3.6, as the corresponding symbols are not available anymore.

### How Has This Been Tested?
Compiled and ran OBS with Python 3.6.2, 3.8.10, and 3.9.6 on Windows and 3.9.13 on macOS.

### Types of changes

### Checklist:

### DIFF
--- a/deps/obs-scripting/obs-scripting-python-import.c
+++ b/deps/obs-scripting/obs-scripting-python-import.c
@@ -42,7 +42,7 @@
 #define PY_MAJOR_VERSION_MAX 3
 #define PY_MINOR_VERSION_MAX 10
 
-bool import_python(const char *python_path)
+bool import_python(const char *python_path, python_version_t *python_version)
 {
 	struct dstr lib_path;
 	bool success = false;
@@ -50,6 +50,12 @@ bool import_python(const char *python_path)
 
 	if (!python_path)
 		python_path = "";
+
+	if (!python_version) {
+		blog(LOG_DEBUG,
+		     "[Python] Invalid python_version pointer provided.");
+		goto fail;
+	}
 
 	dstr_init_copy(&lib_path, python_path);
 	dstr_replace(&lib_path, "\\", "/");
@@ -96,6 +102,9 @@ bool import_python(const char *python_path)
 		goto fail;
 	}
 
+	python_version->major = PY_MAJOR_VERSION_MAX;
+	python_version->minor = minor_version;
+
 #define IMPORT_FUNC(x)                                                     \
 	do {                                                               \
 		Import_##x = os_dlsym(lib, #x);                            \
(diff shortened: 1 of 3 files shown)


flaeri:
Can indeed confirm this solves https://github.com/obsproject/obs-studio/issues/6897 both on launch and when adding py 3.6 on win. Thank you very much!

RytoEX:
Appears to fix #6897 in my testing. @flaeri ?

RytoEX:
A key to note while trying to reproduce the crash in #6897 is that this crash only happens if one of these is true:
1. The _first_ version of Python you select is 3.6 (crash after select).
2. You already have Python configured to select 3.6, such as users who are already using Python scripts (crash on OBS start).
3. You first select a working Python version, such as Python 3.8.2, and then select Python 3.6.x, then restart OBS (crash on OBS start).

alinsavix:
> **Note:** Once Python 3.11 is out officially, we probably have to drop support for Python 3.6, as the corresponding symbols are not available anymore.

It might be worth considering dropping Python 3.6 *now*, given that it's been completely unsupported (not even security patches) for more than six months at this point.

PatTheMav:
> > **Note:** Once Python 3.11 is out officially, we probably have to drop support for Python 3.6, as the corresponding symbols are not available anymore.
> 
> It might be worth considering dropping Python 3.6 _now_, given that it's been completely unsupported (not even security patches) for more than six months at this point.

In theory that's our plan, but for the immediate future we still debate whether we want to force people to update to Python 3.9 or 3.10 _now_ or whether we want to allow people to keep their existing setups after the update (there's gonna be enough people only updating after the beta period) and gracefully update their Python version then.

RytoEX (approved):
Tested Windows x64. Looks good to me.

PR #6916, opened 2022-08-02
libobs: Reject plugins compiled with newer libobs

norihiro:
### Description

Check `obs_module_ver` and reject if it returned a larger value so that plugins build with newer libobs won't be loaded.

### Motivation and Context

When getting a plugins that is compiled with a newer libobs version in the future, such plugins should not be loaded to avoid potential crash.

### How Has This Been Tested?

Overwrite `LIBOBS_API_VER` on a 3rd paty plugin and compile it, then checked the plugin is not loaded.
```
#define LIBOBS_API_VER MAKE_SEMANTIC_VERSION(28, 0, 0)
OBS_DECLARE_MODULE()
```

### Types of changes

### Checklist:


derrod:
There are some potential edge cases here like a plugin being compiled against the latest patch release which was only released for one operating system (for example OBS 27.2.3 being Windows-only). So I'd suggest that this should only look at the major and minor version and ignore patch.

jpark37:
I think there's a couple issues with this:

- Our recent discussions only involve plugins that rely on Qt5 because of the coming Qt6 upgrade, but all plugins are caught in the crossfire now.
- Plugins that would normally work are now gated by this version check, which means plugin authors are now required to continually bump the version and make new releases that they wouldn't have to before.

@Fenrirthviti mentioned something about the planned plugin manager addressing this in some fashion although I don't know the details.

Fenrirthviti:
Partly because of the above, the draft for the plugin manager would have handled the compatibility checks as part of the plugin manifest for the plugin manager (i.e. the plugin metadata that the plugin manager uses), but that obviously wouldn't work for self-installed plugins or during a migration, so this probably needs some further discussion on the best way to handle. Maybe a combination of both approaches.

derrod:
> * Plugins that would normally work are now gated by this version check, which means plugin authors are now required to continually bump the version and make new releases that they wouldn't have to before. EDIT: Which also means users would lose access to those plugins if they upgrade OBS until those plugin updates are released.

I believe you got that the wrong way around. This PR blocks plugins compiled against a *newer* libobs version from loading, not older. So existing plugins are unaffected, but going forward, should there be another big breaking change, a plugin built for a newer obs will no longer load in an older one.

Due to minor changes in patch versions, I believe this check should be restricted to Major/Minor however.

jpark37:
Oh, I did have it backwards. Never mind everything I said then. I also think patch should be removed from the check though.

norihiro:
> So I'd suggest that this should only look at the major and minor version and ignore patch.

Should only patch-version masked or maybe minor version as well?

> Plugins that would normally work are now gated by this version check, which means plugin authors are now required to continually bump the version and make new releases that they wouldn't have to before. EDIT: Which also means users would lose access to those plugins if they upgrade OBS until those plugin updates are released.

This is opposite. Old plugins are not gated by this PR. The case the plugin does not work is if plugin developers bump the version, and if user still uses old OBS.

Even without this PR, plugin-load sometimes partially fails because of `sizeof(struct obs_source_info)`, for example, when the plugin developer chooses newer libobs version than users' OBS version. This PR will strongly reject loading such plugins entirely.

This PR does not solve the issue on Qt5 vs. Qt6 unless back-ported to the `release/27.2` branch. This PR will avoid _future_ issues similar to the Qt5 vs. Qt6.

If plugin author want to define OBS version more complexly, they can call `obs_get_version()` or `obs_get_version_string()` from `obs_module_load` and return failure, for example.

jpark37:
I would keep minor.

norihiro:
Updated to ignore patch version of the plugin.

norihiro:
> Need to increment LIBOBS_API_VER in the new major beta version (such as 28.0.0-beta, 29.0.0-beta) so that the plugins built with the new major beta version is not loaded by the old OBS.

The versioning in `obs.h` is necessary to be considered so that I put this PR draft.

PatTheMav:
> > Need to increment LIBOBS_API_VER in the new major beta version (such as 28.0.0-beta, 29.0.0-beta) so that the plugins built with the new major beta version is not loaded by the old OBS.
> 
> The versioning in `obs.h` is necessary to be considered so that I put this PR draft, though the code itself is ready.

Is this still necessary? What needs to be done to get this PR ready to be merged?

norihiro:
The current plugin template specifies the version of libobs to compile against. I think this change is ok to be merged to obs-studio.

PatTheMav:
> The current plugin template specifies the version of libobs to compile against. I think this change is ok to be merged to obs-studio.

What would be the impact for plugins who are not based on the template?

norihiro:
> What would be the impact for plugins who are not based on the template?

With this PR, if a plugin is build against a newer libobs major-minor version than the running OBS, the plugin will be rejected and a warning message will be left on the log.
Without this PR, such a plugin is usually loaded but this would be dangerous as the plugin can expect the new feature. Sometimes `obs_register_source` will ignore to register if the size of `struct obs_source_info` is bigger than that on the running OBS. Though, `obs_register_source` returns nothing so the plugin cannot know the source-type was registered or not.

PatTheMav:
> With this PR, if a plugin is build against a newer libobs major-minor version than the running OBS, the plugin will be rejected and a warning message will be left on the log.

That does sound right to me.

@RytoEX WDYT?

RytoEX:
Will discuss, but I think I'll target this for the next major release.

PatTheMav:
@RytoEX So I guess this is the time for us to decide whether we want to pull the trigger on this one? Code looks fine to me, so it's mostly a decision about whether "now" is the right time for the plugin support effort this potentially creates.

RytoEX:
> @RytoEX So I guess this is the time for us to decide whether we want to pull the trigger on this one? Code looks fine to me, so it's mostly a decision about whether "now" is the right time for the plugin support effort this potentially creates.

Yes, I was effectively waiting for a major/breaking release to merge this.

libobs/obs-module.c
```diff
@@ -129,6 +129,16 @@ int obs_open_module(obs_module_t **module, const char *path,
...
 
+	/* Reject plugins compiled with a newer libobs.
+	 * Patch version (lower 16-bit) is ignored. */
+	uint32_t ver = mod.ver ? mod.ver() & 0xFFFF0000 : 0;
+	if (ver > LIBOBS_API_VER) {
+		blog(LOG_WARNING,
+		     "Module '%s' compiled with newer libobs %d.%d", path,
+		     (ver >> 24) & 0xFF, (ver >> 16) & 0xFF);
+		return MODULE_INCOMPATIBLE_VER;
+	}
```
RytoEX: Can unwrap/rewrap to 120 columns.

RytoEX (changes requested):
Nit.

libobs/obs-module.c
norihiro: Thank you. I rebased and fixed the format.

RytoEX: Nothing seems to have changed in this file. You'll likely have to manually unwrap and rewrap the lines, or unwrap them and let clang-format wrap them for you.

norihiro: I'm sorry it was my bad. I did git-add but forgot to run git-commit.

RytoEX (approved):

PR #6929, opened 2022-08-03
UI/installer: Require Windows 10 64bit minimum

WizardCM:
### Description

This updates the Windows Vista check to Windows 8.1 and lower.

image

This also improves the wording of the 32bit error.

Submitting this as a PR to get feedback on wording, and to determine if more 32bit/64bit code should be removed from the installer script.

### Motivation and Context

With Qt 6, Windows 8.1 and below are no longer supported.

Additionally, 32bit is no longer supported.

### How Has This Been Tested?

* Launch the installer on Windows 8.1.
* Click "Next" to go past the first screen
* See error message

### Types of changes

### Checklist:

### DIFF
--- a/UI/installer/mp-installer.nsi
+++ b/UI/installer/mp-installer.nsi
@@ -110,16 +110,17 @@ Function PreReqCheck
 		IfSilent +1 +3
 			SetErrorLevel 3
 			Quit
-		MessageBox MB_OK|MB_ICONSTOP "This version of ${APPNAME} is not compatible with your system. Please use the 32bit (x86) installer."
+		MessageBox MB_OK|MB_ICONSTOP "${APPNAME} is not compatible with your operating system's architecture."
 	${EndIf}
-	; Abort on XP or lower
+	; Abort on 8.1 or lower
 !endif
 
-	${If} ${AtMostWinVista}
+	${If} ${AtLeastWin10}
+	${Else}
 		IfSilent +1 +3
 			SetErrorLevel 3
 			Quit
-		MessageBox MB_OK|MB_ICONSTOP "Due to extensive use of DirectX 10 features, ${APPNAME} requires Windows 7 or higher and cannot be installed on this version of Windows."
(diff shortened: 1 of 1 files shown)


UI/installer/mp-installer.nsi
```diff
@@ -110,16 +110,17 @@ Function PreReqCheck
 		IfSilent +1 +3
 			SetErrorLevel 3
 			Quit
-		MessageBox MB_OK|MB_ICONSTOP "This version of ${APPNAME} is not compatible with your system. Please use the 32bit (x86) installer."
+		MessageBox MB_OK|MB_ICONSTOP "${APPNAME} is not compatible with your system's architecture."
```
RytoEX, line 113: Technically, it's the operating system's architecture and not necessarily the system's (hardware's) architecture. A person could, for _some_ reason, be running a 32-bit OS on 64-bit capable hardware.

RytoEX (changes requested):
Other than my one review comment, this looks fine. I don't think we need to roll a bigger cleanup into this. We should just keep it scoped as-is and deal with bigger cleanups later. In fact, I would defer huge cleanups until _after_ 28.0 is officially released, so that we don't run the risk of having a lot of stuff to roll back.

dodgepong:
Have we confirmed that Qt 6 doesn't work on 8.1, or are we just taking Qt's word for it that it doesn't work on 8.1?

WizardCM:
> Have we confirmed that Qt 6 doesn't work on 8.1, or are we just taking Qt's word for it that it doesn't work on 8.1?

Yep, confirmed it myself just in case. This is the error users see:

image

UI/installer/mp-installer.nsi
RytoEX: If it wasn't clear, I prefer this string:
```
"${APPNAME} is not compatible with your operating system's architecture."
```

RytoEX (approved):
Looks good to me.

PR #6938, opened 2022-08-03
Update README.rst

AugustinMauroy:
fix tag


notr1ch:
Thank you, but this is unnecessary.

PR #6960, opened 2022-08-05
Switch clean up (remove one-case switch and switch warning workaround)

tytan652:
### Description

Replace one-case switches, remove workaround casts and refactor a little the Windows taskbar code to fix its switch.

### Motivation and Context

Clean up switches and switch warnings that can be simply cleaned.

### How Has This Been Tested?

Builds without issue.

### Types of changes

### Checklist:

### DIFF
--- a/UI/platform-windows.cpp
+++ b/UI/platform-windows.cpp
@@ -465,33 +465,32 @@ void TaskbarOverlaySetStatus(TaskbarOverlayStatus status)
 		return;
 	}
 
-	if (status != TaskbarOverlayStatusInactive) {
-		QIcon qicon;
-		switch (status) {
-		case TaskbarOverlayStatusActive:
-			qicon = QIcon::fromTheme(
-				"obs-active", QIcon(":/res/images/active.png"));
-			break;
-		case TaskbarOverlayStatusPaused:
-			qicon = QIcon::fromTheme(
-				"obs-paused", QIcon(":/res/images/paused.png"));
-			break;
-		}
-
-		HICON hicon = nullptr;
-		if (!qicon.isNull()) {
-			Q_GUI_EXPORT HICON qt_pixmapToWinHICON(
-				const QPixmap &p);
-			hicon = qt_pixmapToWinHICON(
-				qicon.pixmap(GetSystemMetrics(SM_CXSMICON)));
-			if (!hicon)
-				return;
-		}
-
-		taskbarIcon->SetOverlayIcon(hwnd, hicon, nullptr);
-		DestroyIcon(hicon);
-	} else {
+	QIcon qicon;
+	switch (status) {
+	case TaskbarOverlayStatusActive:
+		qicon = QIcon::fromTheme("obs-active",
+					 QIcon(":/res/images/active.png"));
+		break;
+	case TaskbarOverlayStatusPaused:
+		qicon = QIcon::fromTheme("obs-paused",
+					 QIcon(":/res/images/paused.png"));
+		break;
+	case TaskbarOverlayStatusInactive:
 		taskbarIcon->SetOverlayIcon(hwnd, nullptr, nullptr);
+		taskbarIcon->Release();
+		return;
 	}
+
+	HICON hicon = nullptr;
+	if (!qicon.isNull()) {
+		Q_GUI_EXPORT HICON qt_pixmapToWinHICON(const QPixmap &p);
+		hicon = qt_pixmapToWinHICON(
+			qicon.pixmap(GetSystemMetrics(SM_CXSMICON)));
+		if (!hicon)
+			return;
+	}
+
+	taskbarIcon->SetOverlayIcon(hwnd, hicon, nullptr);
+	DestroyIcon(hicon);
 	taskbarIcon->Release();
 }
--- a/libobs/obs-scene.c
+++ b/libobs/obs-scene.c
@@ -614,6 +614,9 @@ static void render_item_texture(struct obs_scene_item *item,
 		case GS_CS_SRGB_16F:
 		case GS_CS_709_EXTENDED:
 			multiplier = obs_get_video_sdr_white_level() / 80.f;
+			break;
+		case GS_CS_709_SCRGB:
+			break;
 		}
 	}
 
@@ -623,6 +626,9 @@ static void render_item_texture(struct obs_scene_item *item,
 		case GS_CS_SRGB_16F:
 		case GS_CS_709_EXTENDED:
 			multiplier = 80.f / obs_get_video_sdr_white_level();
+			break;
+		case GS_CS_709_SCRGB:
+			break;
 		}
 	}
 
--- a/libobs/obs-scene.c
+++ b/libobs/obs-scene.c
@@ -608,8 +608,7 @@ static void render_item_texture(struct obs_scene_item *item,
 
 	float multiplier = 1.f;
 
-	switch (current_space) {
-	case GS_CS_709_SCRGB:
+	if (current_space == GS_CS_709_SCRGB) {
 		switch (source_space) {
 		case GS_CS_SRGB:
 		case GS_CS_SRGB_16F:
@@ -618,8 +617,7 @@ static void render_item_texture(struct obs_scene_item *item,
 		}
 	}
 
-	switch (source_space) {
-	case GS_CS_709_SCRGB:
+	if (source_space == GS_CS_709_SCRGB) {
 		switch (current_space) {
 		case GS_CS_SRGB:
 		case GS_CS_SRGB_16F:
@@ -634,10 +632,8 @@ static void render_item_texture(struct obs_scene_item *item,
 		switch (source_space) {
 		case GS_CS_SRGB:
 		case GS_CS_SRGB_16F:
-			switch (current_space) {
-			case GS_CS_709_SCRGB:
+			if (current_space == GS_CS_709_SCRGB)
 				tech_name = "DrawUpscaleMultiply";
-			}
 			break;
 		case GS_CS_709_EXTENDED:
 			switch (current_space) {
@@ -663,10 +659,8 @@ static void render_item_texture(struct obs_scene_item *item,
 		switch (source_space) {
 		case GS_CS_SRGB:
 		case GS_CS_SRGB_16F:
-			switch (current_space) {
-			case GS_CS_709_SCRGB:
+			if (current_space == GS_CS_709_SCRGB)
 				tech_name = "DrawMultiply";
-			}
 			break;
 		case GS_CS_709_EXTENDED:
 			switch (current_space) {
--- a/libobs/obs-source-deinterlace.c
+++ b/libobs/obs-source-deinterlace.c
@@ -386,8 +386,7 @@ void deinterlace_render(obs_source_t *s)
 	switch (source_space) {
 	case GS_CS_SRGB:
 	case GS_CS_SRGB_16F:
-		switch (current_space) {
-		case GS_CS_709_SCRGB:
+		if (current_space == GS_CS_709_SCRGB) {
 			tech_name = "DrawMultiply";
 			multiplier = obs_get_video_sdr_white_level() / 80.0f;
 		}
--- a/UI/source-tree.cpp
+++ b/UI/source-tree.cpp
@@ -636,13 +636,15 @@ void SourceTreeModel::OBSFrontendEvent(enum obs_frontend_event event, void *ptr)
 {
 	SourceTreeModel *stm = reinterpret_cast<SourceTreeModel *>(ptr);
 
-	switch ((int)event) {
+	switch (event) {
 	case OBS_FRONTEND_EVENT_PREVIEW_SCENE_CHANGED:
 		stm->SceneChanged();
 		break;
 	case OBS_FRONTEND_EVENT_EXIT:
 	case OBS_FRONTEND_EVENT_SCENE_COLLECTION_CLEANUP:
 		stm->Clear();
+		break;
+	default:
 		break;
 	}
 }
--- a/UI/window-basic-stats.cpp
+++ b/UI/window-basic-stats.cpp
@@ -22,12 +22,14 @@ void OBSBasicStats::OBSFrontendEvent(enum obs_frontend_event event, void *ptr)
 {
 	OBSBasicStats *stats = reinterpret_cast<OBSBasicStats *>(ptr);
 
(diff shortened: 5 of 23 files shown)


tytan652:
Forgot to clean some switch with cast as a warning workaround. Draft until I remove those.

libobs/media-io/video-matrices.c
```diff
@@ -247,6 +247,8 @@ bool video_format_get_parameters(enum video_colorspace color_space,
 	case VIDEO_CS_2100_PQ:
 	case VIDEO_CS_2100_HLG:
 		bpc = 10;
+	default:
+		break;
```
norihiro, line 251: - If adding some code in future, we should avoid fall-through.
- If adding new enum elements for `video_colorspace`, we should list all cases so that the warning can catch newly added elements.
```suggestion
		break;
	case VIDEO_CS_DEFAULT:
	case VIDEO_CS_601:
	case VIDEO_CS_709:
	case VIDEO_CS_SRGB:
		bpc = 8;
```

tytan652: The warning was literally removed in libobs. So it means that we don't care about managing every member of the enumeration.

norihiro (commented):
Other than a recommendation to avoid fall-through, I've added two comments on `plugins/linux-v4l2/v4l2-decoder.c` and `plugins/obs-ffmpeg/obs-ffmpeg-vaapi.c`.

plugins/obs-ffmpeg/obs-ffmpeg-vaapi.c
```diff
@@ -272,6 +272,8 @@ static bool vaapi_update(void *data, obs_data_t *settings)
 		enc->context->color_primaries = AVCOL_PRI_BT709;
 		enc->context->colorspace = AVCOL_SPC_BT709;
 		break;
+	default:
+		break;
```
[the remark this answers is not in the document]
tytan652: We were not doing it before, the warning was just remove with `−Wno-switch`.

tytan652:
I think I added all the missing breaks.

plugins/linux-v4l2/v4l2-decoder.c
```diff
@@ -116,6 +116,8 @@ int v4l2_decode_frame(struct obs_source_frame *out, uint8_t *data,
 	case AV_PIX_FMT_YUV444P:
 		out->format = VIDEO_FORMAT_I444;
 		break;
+	default:
+		break;
```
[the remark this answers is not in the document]
tytan652: I would like to say it's out of scope of the PR.

plugins/obs-ffmpeg/obs-ffmpeg-vaapi.c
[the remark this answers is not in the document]
notr1ch: Is this really the correct fix though instead of handling the missing enums?

notr1ch:
I'm a little concerned in general that this may mask other bugs such as not all enum values being handled. Sometimes it's appropriate that there's no default.

tytan652:
> I'm a little concerned in general that this may mask other bugs such as not all enum values being handled. Sometimes it's appropriate that there's no default.

Most of them were suppressed (made invisible).

plugins/obs-ffmpeg/obs-ffmpeg-vaapi.c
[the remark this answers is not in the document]
norihiro: I know this modification in the PR does not change the existing behavior. However, I doubt this might be an existing bug.

gxalpha:
> I'm a little concerned in general that this may mask other bugs such as not all enum values being handled. Sometimes it's appropriate that there's no default.

> Most of them were suppressed (made invisible).

I agree that just adding a `default` statement everywhere isn't "fixing" the warnings. At that point you might as well not enable the warning in the first place. I'd even argue it's worse than the status quo, as any developer who looks at a switch statement will think "looks like everything is fine, that default statement looks intentional", while in reality the default statement is just there to suppress the warning.

> When warnings will be treated as errors, only enum switches that have a case per constant or a default will pass.

If that's the issue, just don't turn in into an error, but leave it as a warning. Might encourage developers to go over them one by one and fix them over time instead of just suppressing them all.

tytan652:
I'm going to remove every default that I added in #6759 those switch will be guarded with pragma to not emit errors.

gxalpha:
I agree that removing the defaults here is correct, I still disagree that they should be suppressed in #6759 (assuming that's what you mean by "guarded with pragma to not emit errors "). Suppressing them again leads to them getting ignored, even though they still could be masking bugs.
If all we want is to ensure that *new* switches always respect every case, then this is the correct approach. But I don't agree that that should be the goal. If we want switch warnings to be fixed, that should include the existing ones as well. I don't see a point in making sure that only new switches respect every case, while old ones are fine to not have every case. If switch case warnings are to be checked, it should include each of them.
I completely understand that that's out-of-scope for this PR or (6759), and it really isn't doable by one developer alone (and I'm not asking for that, to be clear). However I'm sure that once yellow warnings pop up all over the code, developers (including me) will start actually fixing them if they feel .
So the TL;DR of the above is that I'd do `-Wno-error=switch` in #6759 instead of silencing via pragma.

And to be absolutely clear, I'm not in a position to make demands and hope that the above doesn't sound like one. Take it as a somewhat opinionated suggestion.

tytan652:
@gxalpha you didn't understood we can use pragma to keep a warning as a warning and not an error.

tytan652:
The purpose is to avoid new "incomplete" switch to be PRed, so we can use pragma to keep actual "incomplete" switch warnings as warnings and not errors.

gxalpha:
> @gxalpha you didn't understood the pragma part, we can use pragma to keep a warning as a warning and not an error.

Oh I see what you mean. That sounds like a lot of code, but I guess that's fine if it does the job.

UI/platform-windows.cpp
```diff
@@ -524,19 +524,22 @@ void TaskbarOverlaySetStatus(TaskbarOverlayStatus status)
...
+		qicon = QIcon::fromTheme("obs-active",
+					 QIcon(":/res/images/active.png"));
+		break;
+	case TaskbarOverlayStatusPaused:
+		qicon = QIcon::fromTheme("obs-paused",
+					 QIcon(":/res/images/paused.png"));
+		break;
+	case TaskbarOverlayStatusInactive:
+		taskbarIcon->SetOverlayIcon(hwnd, nullptr, nullptr);
+		break;
```
norihiro, line 539: How about return here so that it's not necessary to check `status` again later.
```suggestion
		taskbarIcon->Release();
		return;
```

tytan652: I finish my work about MSVC warning and then I will apply your suggestion.

tytan652:
Added switch warning (C4062) on MSVC for parity with Clang and GCC and add pragmas to incomplete switch to emits warning rather than errors.

tytan652:
Waiting for enc-amf submodule to be updated to undraft.

libobs/media-io/video-io.c
```diff
@@ -310,6 +312,7 @@ static enum video_colorspace collapse_space(enum video_colorspace cs)
 	case VIDEO_CS_2100_HLG:
 		cs = VIDEO_CS_2100_PQ;
 	}
+	PRAGMA_WARN_POP
```
norihiro, line 315: Instead of assigning to `cs`, I'd like to suggest just returning from each case condition.
```c
static enum video_colorspace collapse_space(enum video_colorspace cs)
{
	switch (cs) {
	case VIDEO_CS_DEFAULT:
	case VIDEO_CS_SRGB:
		return VIDEO_CS_709;
	case VIDEO_CS_2100_HLG:
		return VIDEO_CS_2100_PQ;
	default:
		return cs;
	}
}
```

libobs/media-io/video-matrices.c
```diff
@@ -243,11 +243,14 @@ bool video_format_get_parameters(enum video_colorspace color_space,
...
 {
 	uint32_t bpc = 8;
+	PRAGMA_WARN_PUSH
+	PRAGMA_WARN_SWITCH
 	switch (color_space) {
 	case VIDEO_CS_2100_PQ:
 	case VIDEO_CS_2100_HLG:
 		bpc = 10;
 	}
+	PRAGMA_WARN_POP
```
norihiro, line 253: Instead of disabling the warning, I'd like to suggest using `default` condition and assigning `8`.
```c
        uint32_t bpc;
        switch (color_space) {
        case VIDEO_CS_2100_PQ:
        case VIDEO_CS_2100_HLG:
                bpc = 10;
                break;
        default:
                bpc = 8;
        }
```

libobs/obs-scene.c
```diff
@@ -608,24 +608,28 @@ static void render_item_texture(struct obs_scene_item *item,
...
+		PRAGMA_WARN_PUSH
+		PRAGMA_WARN_SWITCH
 		switch (source_space) {
 		case GS_CS_SRGB:
 		case GS_CS_SRGB_16F:
 		case GS_CS_709_EXTENDED:
 			multiplier = obs_get_video_sdr_white_level() / 80.f;
 		}
+		PRAGMA_WARN_POP
 	}
```
norihiro, line 621:
```suggestion
	if (current_space == GS_CS_709_SCRGB &&
	    source_space != GS_CS_709_SCRGB) {
		multiplier = obs_get_video_sdr_white_level() / 80.f;
	}
```

libobs/obs-scene.c
```diff
@@ -608,24 +608,28 @@ static void render_item_texture(struct obs_scene_item *item,
...
+		PRAGMA_WARN_PUSH
+		PRAGMA_WARN_SWITCH
 		switch (current_space) {
 		case GS_CS_SRGB:
 		case GS_CS_SRGB_16F:
 		case GS_CS_709_EXTENDED:
 			multiplier = 80.f / obs_get_video_sdr_white_level();
 		}
+		PRAGMA_WARN_POP
 	}
```
norihiro, line 633:
```suggestion
	if (source_space == GS_CS_709_SCRGB &&
	    current_space != GS_CS_709_SCRGB) {
		multiplier = 80.f / obs_get_video_sdr_white_level();
	}
```

libobs/media-io/video-matrices.c
```diff
@@ -272,6 +277,7 @@ bool video_format_get_parameters_for_format(enum video_colorspace color_space,
 		bpc = 12;
 		break;
 	}
+	PRAGMA_WARN_POP
```
norihiro, line 280:
```c
        uint32_t bpc;
        switch (format) {
        case VIDEO_FORMAT_I010:
        case VIDEO_FORMAT_P010:
        case VIDEO_FORMAT_I210:
                bpc = 10;
                break;
        case VIDEO_FORMAT_I412:
        case VIDEO_FORMAT_YA2L:
                bpc = 12;
                break;
        default:
                bpc = 8;
        }
```

libobs/obs-source-deinterlace.c
```diff
@@ -418,8 +419,11 @@ void deinterlace_render(obs_source_t *s)
 			tech_name = "DrawMultiply";
 			multiplier = obs_get_video_sdr_white_level() / 80.0f;
 		}
+		PRAGMA_WARN_POP
```
norihiro, line 422: I think this pragma can put outside of the upper `switch` statement instead of having for each.

norihiro (commented):
I feel using `default` is better than suppressing warnings by pragma. Especially the macro `PRAGMA_WARN_PUSH`, `PRAGMA_WARN_SWITCH`, `PRAGMA_WARN_POP` is too confusing for editors (`vim` for example) to auto-indent.

tytan652:
I put what I wrote in one of my commit:

> We can't fix them all in one PR, so those will need to be fixed in the
future.

I will not modify them in this PR.

norihiro:
> I will not modify them in this PR.

In that case, I don't think it's a good idea to touch the source file to disable the warnings but I suggest leaving the warnings.

libobs/media-io/video-matrices.c
tytan652: `default` is a no go there, it needs to be all cases in the future. But for now, in this PR it's pragma.

libobs/media-io/video-matrices.c
tytan652: `default` is a no go there, it needs to be all cases in the future. But for now, in this PR it's pragma.

libobs/obs-source-deinterlace.c
tytan652: MSVC pragma emits a message because MSVC doesn't know "warning without erroring" and so I want/need one per switch.

tytan652:
I don't make them desperate, I make them not emit errors.

libobs/media-io/video-io.c
tytan652: `default` is a no go there, it seems to need to be all cases in the future. But for now, in this PR it's pragma.

tytan652:
Testing something.

tytan652:
Updated the PR to no longer add compiler flag, only switch fixes are kept.

libobs/obs-scene.c
```diff
@@ -608,24 +608,14 @@ static void render_item_texture(struct obs_scene_item *item,
...
-	case GS_CS_709_SCRGB:
-		switch (source_space) {
-		case GS_CS_SRGB:
-		case GS_CS_SRGB_16F:
-		case GS_CS_709_EXTENDED:
-			multiplier = obs_get_video_sdr_white_level() / 80.f;
-		}
+	if (current_space == GS_CS_709_SCRGB &&
+	    source_space != GS_CS_709_SCRGB) {
+		multiplier = obs_get_video_sdr_white_level() / 80.f;
```
tt2468, line 613: I see we're effectively switching from a whitelist to a blacklist. Could this produce unintended behavior if more color spaces are added in the future?

tt2468: Is it not possible that another color space with its own multiplier could be added, that would require fraction multiplication instead? Say, a hypothetical `GS_CS_420_BLZIT` where it requires a denominator of 75.0f?

tt2468: I'm just worried that in a future where this happens, we are placing a bet on a future developer being able to catch this behavior, and going from a switch here to a single if statement would likely make troubleshooting errors more difficult.

A comment perhaps pointing out that this logic relies on there only being one color space with a denominator would be good IMO.

PR #6962, opened 2022-08-05
UI: Fix small font size on macOS and fallback font

PatTheMav:
### Description
Fixes the too-small font-size used by Yami for macOS and adds the macOS ui default font as a fallback for when OpenSans cannot be found.

This is achieved by extending the event handler of the main window, waiting for the StyleChange event to happen. As patching the font-size itself will emit that event again, we have to guard this second event against patching the font again, resulting in an endless loop.

For that reason a boolean is stored on the object which is reset when another theme but Yami is selected, insuring that the patch will be re-applied if users switch back to Yami.

### Motivation and Context
Qt uses a default font size of 13pt for QWidgets on macOS. By setting the default font size to 10pt, text becomes too small for comfortable use on macOS. 

As the original style is applied via stylesheet, all further changes are suggested by Qt to be done via stylesheets as well, hence why this is implemented via appending the original stylesheet string.

This in turn requires setting the stylesheet on a class of the respective style (so it has to be applied to a QWidget object) or else it won't stick or be applied correctly.

Using a modern fallback font also ensures OBS fits into the overall OS look and feel if for some reason OpenSans cannot be found.

### How Has This Been Tested?
Tested on macOS 12.5 with Yami active.

### Types of changes

### Checklist:


UI/window-basic-main.cpp
```diff
@@ -4847,6 +4847,20 @@ void OBSBasic::changeEvent(QEvent *event)
 			if (previewEnabled)
 				EnablePreviewDisplay(true);
 		}
+#ifdef __APPLE__
+	} else if (event->type() == QEvent::StyleChange) {
+		const char *theme_name = config_get_string(
+			App()->GlobalConfig(), "General", "CurrentTheme3");
+		if (theme_name && strcmp(theme_name, "Yami") == 0) {
```
gxalpha, line 4854: This could get messy when we add other color palettes of Yami (gray, light, etc).

UI/window-basic-main.cpp
```diff
@@ -4847,6 +4847,20 @@ void OBSBasic::changeEvent(QEvent *event)
...
+			if (!fontsizePatched) {
+				fontsizePatched = true;
+				setStyleSheet(styleSheet().append(
+					QString("font-size: 13pt;")));
```
gxalpha, line 4858: This overrides every font-size in the app, which is an issue as not all font sizes are originally 10 (see the Yami.qss file for examples, like `10.5` in QDockWidget or `26` in the about dialog).

gxalpha (commented):
See the comments below.
I'm thinking what we could do is have a `font-size-default` and a `font-size-mac` in the qss file and then do a string replacement of the appropriate one (depending on the OS) with `font-size`. That wouldn't be much less hacky, but work for different font sizes and even across themes.

UI/window-basic-main.cpp
WizardCM: I could add a `font-size-[platform]` in `OBSThemeMeta {}` to do this programmatically rather than hardcoding themes in some way. It's the kind of thing I made `OBSThemeMeta` for.

gxalpha: That could probably work as well, although you could run into the same issue with differing font sizes throughout the app.
(See the second comment:)
> This overrides every font-size in the app, which is an issue as not all font sizes are originally 10 (see the Yami.qss file for examples, like 10.5 in QDockWidget or 26 in the about dialog).

UI/window-basic-main.cpp
PatTheMav: That doesn't seem to be the case though - other font-sizes retain their size:

Bildschirmfoto 2022-08-10 um 13 52 29

Also this only sets the font-size on `OBSBasic`, more specific style sheet properties should still be applied regardless.

gxalpha: This is how it currently looks, the "OBS Studio" and the version are supposed to be larger:
image

PatTheMav:
Updated the PR to patch stylesheet files after building the binary and the files have been copied into the app bundle.

UI/data/themes/Yami.qss
```diff
@@ -82,7 +82,7 @@ QWidget {
     selection-background-color: rgb(40,76,184);
     selection-color: palette(text);
     font-size: 10pt;
-    font-family: 'Open Sans', Helvetica, Arial, 'MS Shell Dlg', sans-serif
+    font-family: 'Open Sans', '.AppleSystemUIFont', Helvetica, Arial, 'MS Shell Dlg', sans-serif;
```
tytan652, line 85: Should not this applied to every Yami variants (Grey, Light, Acri, Rachni) ?

PatTheMav: Correct, fixed.

WizardCM (approved):

PR #6966, opened 2022-08-06
UI: Fix Yami list widget hover color

cg2121:
### Description
With Yami, when hovering a list widget (ie. scenes/settings), the
background wouldn't change like the sources list.

### Motivation and Context
Fix Yami bugs

### How Has This Been Tested?
Hovered over list widgets

### Types of changes

### Checklist:

### DIFF
--- a/UI/data/themes/Yami.qss
+++ b/UI/data/themes/Yami.qss
@@ -201,7 +201,11 @@ SourceTree::item:selected:hover {
 }
 
 QListWidget::item:disabled,
-QListWidget::item:hover {
+QListWidget::item:disabled:hover,
+SourceTree::item:disabled,
+SourceTree::item:disabled:hover,
+SceneTree::item:disabled,
+SceneTree::item:disabled:hover {
     background: transparent;
     color: rgb(153,153,153);
 }


UI/data/themes/Yami.qss
```diff
@@ -196,8 +196,7 @@ SourceTree::item:selected:hover {
     color: palette(text);
 }
 
-QListWidget::item:disabled,
-QListWidget::item:hover {
+QListWidget::item:disabled {
```
WizardCM, line 199: I wonder if the original intent was something closer to 

```
QListWidget::item:disabled,
QListWidget::item:disabled:hover {
```

RytoEX: I have a vague recollection that @Warchamp7 ran into some issues on Windows where QSS behaved strangely without certain combinations or selectors that otherwise did not make sense. As with any UI changes, some before/after images would be helpful to reviewing this.

cg2121: Updated to include all possible combinations of the :disabled and :hover

Warchamp7: Background should only be transparent for disabled items. I think this was originally a fix for the Add Existing source list, where the hover color was appearing even when the list was disabled.

WizardCM (approved):

PR #6969, opened 2022-08-06
UI: Force fixed font in plain text edits

cg2121:
### Description
Since the Yami QSS changes the default font for widgets, we need
to force the system fixed font for plain text edits.

### Motivation and Context
Fix Yami bugs

### How Has This Been Tested?
Opened log viewer to make sure it was the right font

### Types of changes

### Checklist:

### DIFF
--- a/UI/crash-report.cpp
+++ b/UI/crash-report.cpp
@@ -1,12 +1,12 @@
 #include "crash-report.hpp"
 #include <QApplication>
 #include <QFontDatabase>
-#include <QPlainTextEdit>
 #include <QPushButton>
 #include <QVBoxLayout>
 #include <QHBoxLayout>
 #include <QClipboard>
 #include "qt-wrappers.hpp"
+#include "plain-text-edit.hpp"
 
 OBSCrashReport::OBSCrashReport(QWidget *parent, const char *text)
 	: QDialog(parent)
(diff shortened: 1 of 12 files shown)


WizardCM:
I considered this solution, but it'd also be required for *any* monospaced element, such as the CSS input field for browser sources.

cg2121:
Updated to include other QPlainTextEdits.

The only way I could get this working is to set the stylesheet for each text edit. I couldn't get it working at all when setting a global stylesheet in obs-app.cpp

UI/crash-report.cpp
```diff
@@ -20,7 +20,15 @@ OBSCrashReport::OBSCrashReport(QWidget *parent, const char *text)
 	textBox = new QPlainTextEdit;
 	textBox->setPlainText(QT_UTF8(text));
 	textBox->setLineWrapMode(QPlainTextEdit::NoWrap);
-	textBox->setFont(QFontDatabase::systemFont(QFontDatabase::FixedFont));
+
+	const QFont fixedFont =
+		QFontDatabase::systemFont(QFontDatabase::FixedFont);
+	textBox->setFont(fixedFont);
```
PatTheMav, line 26: Is this line actually doing anything? IIRC Qt vaguely tells people not to use `setFont` on Widgets otherwise styled via stylesheets.

cg2121:
Updated to just sub-class QPlainTextEdit. I feel like this is the least horrible way of accomplishing this.

UI/crash-report.cpp
cg2121: Removed.

notr1ch (approved):
Tested and no issues found.

Hellscaped:
6969

PR #6970, opened 2022-08-06
frontend-tools: Display Python version in UI, suggest restart after version switch

WizardCM:
### Description

This adds a visual Python version in the UI to inform the user which version is currently, successfully loaded.

Additionally, if the user attempts to switch versions, a popup is shown to inform them that a restart is required.

~~**I very much need help** writing a better implementation of `obs_scripting_python_version`. I got tired of fighting it.~~

image

image

image

### Motivation and Context

It can currently be very unclear to a user that a) Python loaded successfully and b) changing versions requires an OBS restart.

### How Has This Been Tested?

* Set path to Python 3.8.10
* Change path to Python 3.10

### Types of changes

### Checklist:


UI/frontend-plugins/frontend-tools/scripts.cpp
```diff
@@ -194,6 +194,7 @@ ScriptsTool::ScriptsTool() : QWidget(nullptr), ui(new Ui_ScriptsTool)
 		config_get_string(config, "Python", "Path" ARCH_NAME);
 	ui->pythonPath->setText(path);
 	ui->pythonPathLabel->setText(obs_module_text(PYTHONPATH_LABEL_TEXT));
+	showPythonVersion();
```
PatTheMav, line 197: Given that the label is always "shown" (albeit with an empty text"), `updatePythonVersion` (or even `updatePythonVersionLabel`) would be more descriptive of what the function does.

UI/frontend-plugins/frontend-tools/scripts.cpp
```diff
@@ -219,6 +220,20 @@ ScriptsTool::~ScriptsTool()
 		       ui->scripts->currentRow());
 }
 
+void ScriptsTool::showPythonVersion()
+{
+	bool loaded = obs_scripting_python_loaded();
+	QString label;
+	if (loaded) {
```
PatTheMav, line 227: Minor nitpick - `loaded` isn't used anywhere outside of this branch, so could possibly suffice to use `if(obs_scripting_python_loaded())`.

UI/frontend-plugins/frontend-tools/scripts.cpp
```diff
@@ -455,11 +470,28 @@ void ScriptsTool::on_pythonPathBrowse_clicked()
 
 	ui->pythonPath->setText(newPath);
 
-	if (obs_scripting_python_loaded())
+	bool loaded = obs_scripting_python_loaded();
+	bool pathChanged = !newPath.isEmpty() && curPath.compare(newPath) != 0;
```
PatTheMav, line 474: Minor nitpick, if this were moved into the `if` statement, then the comparisons and checks would only be evaluated when `loaded` is true. As-is, they are evaluated even if python is not loaded at all (when `loaded` is false).

deps/obs-scripting/obs-scripting-python.c
```diff
@@ -1590,6 +1590,25 @@ bool obs_scripting_python_runtime_linked(void)
 	return (bool)RUNTIME_LINK;
 }
 
+char *obs_scripting_python_version(void)
+{
+	struct dstr ver = {0};
```
PatTheMav, line 1595: What you maybe want to do here is initialise `char ver_str[8];` in the calling context (`showPythonVersion`) and pass the address and size of that char array into this function, then use `snprintf` to fill it with the appropriate values (similar to how we "fill" the `python_version_t` with the appropriate values in the first place).

Because you write directly into the char array passed by reference, you don't need to return anything, so the function signature would be `void obs_scripting_python_version(char * version, size_t version_length)`.

I'm also a bit confused that this code works, because you free the `dstr` before returning its internal char pointer, which you'd usually have to do _after_ you passed it via `args` to `QString`.

UI/frontend-plugins/frontend-tools/scripts.cpp
WizardCM: Good point. Fixed.

UI/frontend-plugins/frontend-tools/scripts.cpp
WizardCM: Yeah I did this in prep for a future capability, but better to keep it simple for now. Fixed.

UI/frontend-plugins/frontend-tools/scripts.cpp
WizardCM: True. Fixed.

deps/obs-scripting/obs-scripting-python.c
WizardCM: Yeah I wanted to avoid doing that, but it seems to be the most common way we do it, and the cleanest way to do it overall. Changes made.

WizardCM:
Performed all changes requested in review. Ready for re-review.

UI/frontend-plugins/frontend-tools/scripts.cpp
```diff
@@ -215,6 +216,20 @@ ScriptsTool::~ScriptsTool()
 		       ui->scripts->currentRow());
 }
 
+void ScriptsTool::updatePythonVersionLabel()
+{
+	QString label;
+	if (obs_scripting_python_loaded()) {
+		char version[8];
+		obs_scripting_python_version(version, sizeof(version));
```
PatTheMav, line 224: I wasn't sure if this was right at first, but afaik in C passing the name of the array implicitly passes a pointer to the first element of the array, so it should be correct.

PatTheMav (approved):

productdevbook:
linux dont see, how to change ubuntu ?

PR #6978, opened 2022-08-06
mac-videotoolbox: Reject color formats other than NV12 and I420

gxalpha:
### Description

Changes the Apple VT encoder to fail when an unsupported color format (any other than NV12 or I420) is selected, instead of siliently-ish falling back to NV12.
This code will need tweaking when HEVC is added, as the HEVC encoder also supports P010 (and BRGA?), but that can be done at a later point.

### Motivation and Context

Until recently, the fallback wasn't *that* much of an issue. With the introduction of 10-bit formats however, it means precision loss. And even if a user sets a different 8-bit format, they probably have a reason for doing so, and should know that it doesn't work.

### How Has This Been Tested?

macOS 13
Selecting anything besides NV12 or I420 results in the following error when starting the encoder:
image

### Types of changes

### Checklist:


plugins/mac-videotoolbox/encoder.c
```diff
@@ -497,6 +494,14 @@ static void update_params(struct vt_encoder *enc, obs_data_t *settings)
 	// also sets the enc->vt_pix_fmt
 	vt_video_info(enc, &info);
 
+	if (info.format != voi->format) {
```
PatTheMav, line 497: In the end, this does what it is supposed to, but it seems unfortunate that we pass an "invalid" setting into `vt_video_info`, which then also sets the fallback value of NV12, only to then compare the fallback value with the value in OBS settings.

Ideally, `vt_video_info` (which is also unfortunately named as it doesn't indicate a manipulation of data) should return a boolean when it was given an invalid colour format.

Fenrirthviti:
Just to voice this again, I'm not a fan of silent fallbacks that choose alternate settings than you have set. I feel this should just fail, and the error message be more descriptive on why it's failing, and what the user can do to correct it. Silently failing and changing options for them leads to misinformation and misunderstanding of what works and doesn't work.

plugins/mac-videotoolbox/encoder.c
gxalpha: > Ideally, vt_video_info (which is also unfortunately named as it doesn't indicate a manipulation of data) should return a boolean when it was given an invalid colour format.

I was thinking that as well orgininally, unfortunately it's used directly in the encoder info struct `.get_video_info`. I could refactor the code a bit though.

gxalpha: Cleaned up the existing code.
Removed `.get_video_info` from the struct, since that "can be used to tell the back-end that the video data needs to be automatically converted to a different video format or specific size before being sent to the encoder" (docs), which is exactly the kind of fallback we don't want.

plugins/mac-videotoolbox/encoder.c
```diff
@@ -459,46 +457,40 @@ static void dump_encoder_info(struct vt_encoder *enc)
...
 
-static void vt_video_info(void *data, struct video_scale_info *info)
+static bool set_video_format(struct vt_encoder *enc, enum video_format format,
+			     enum video_range_type range)
 {
-	struct vt_encoder *enc = data;
-
-	if (info->format == VIDEO_FORMAT_I420) {
-		enc->obs_pix_fmt = info->format;
+	bool fullrange = range == VIDEO_RANGE_FULL;
```
PatTheMav, line 463: Minor nitpick, but I'm not sure if the proper snake-case variant would be `full_range` here.

gxalpha: It's definitely not less correct

PatTheMav (approved):

PR #6985, opened 2022-08-07
UI: Fix source tree item color height

cg2121:
### Description
Fixes widget inside of source tree item being incorrect height.

Before:
Screenshot from 2022-08-07 01-47-41

After:
Screenshot from 2022-08-07 02-11-05

### Motivation and Context
Fix theme bugs

### How Has This Been Tested?
Set scene item colors and made sure everything looked good.

### Types of changes

### Checklist:

### DIFF
--- a/UI/source-tree.hpp
+++ b/UI/source-tree.hpp
@@ -8,6 +8,7 @@
 #include <QStaticText>
 #include <QSvgRenderer>
 #include <QAbstractListModel>
+#include <QStyledItemDelegate>
 #include <obs.hpp>
 #include <obs-frontend-api.h>
 
@@ -25,7 +26,7 @@ class SourceTreeSubItemCheckBox : public QCheckBox {
 	Q_OBJECT
 };
 
-class SourceTreeItem : public QWidget {
+class SourceTreeItem : public QFrame {
 	Q_OBJECT
 
 	friend class SourceTree;
@@ -217,4 +218,13 @@ protected:
 	virtual void
 	selectionChanged(const QItemSelection &selected,
 			 const QItemSelection &deselected) override;
+};
+
+class SourceTreeDelegate : public QStyledItemDelegate {
+	Q_OBJECT
+
+public:
+	SourceTreeDelegate(QObject *parent);
+	virtual QSize sizeHint(const QStyleOptionViewItem &option,
+			       const QModelIndex &index) const override;
 };
--- a/UI/data/themes/Acri.qss
+++ b/UI/data/themes/Acri.qss
@@ -189,11 +189,17 @@ SourceTree {
     padding: 3px;
 }
 
+SourceTreeItem,
+QMenu::item,
+SceneTree::item {
+    padding: 6px;
+}
+
+SourceTreeItem,
 QMenu::item,
 SceneTree::item,
 SourceTree::item {
     border-radius: 4px;
-    padding: 6px;
     color: palette(text);
 	border: 0px solid transparent;
 }
--- a/UI/data/themes/Grey.qss
+++ b/UI/data/themes/Grey.qss
@@ -189,11 +189,17 @@ SourceTree {
     padding: 3px;
 }
 
+SourceTreeItem,
+QMenu::item,
+SceneTree::item {
+    padding: 6px;
+}
+
+SourceTreeItem,
 QMenu::item,
 SceneTree::item,
 SourceTree::item {
     border-radius: 4px;
-    padding: 6px;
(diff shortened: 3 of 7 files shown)


WizardCM:
As we're partially breaking existing themes as part of 28 anyway, what if we just restructure this widget so that hacks like this aren't necessary any more? We've had to fight this particular element in various ways for ages.

cg2121:
We can't restructure the widget, as it is a QT problem that it works this way, afaik.

cg2121:
According to a bunch of forum posts, the only way to fix this without changing the qss would to create a delegate and override the paint event in the source tree, which I don't feel like doing. Just some dumb QT things.

WizardCM:
Ah, that's a bummer. Thanks for looking into it.

SuslikV:
The margins should be set in the .qss, not in the code.
Example, 3px from all sides:
```
SourceTree {
	margin: 3px;
}
```
At some point, you should decide either you using .qss or customizing the whole UI manually, in the code.

The padding works OK. Probably, you misunderstanding the box model of the widget. It is not known what the theme should look like (the author should say if the selection border wasn't intentional).

Also, please fix the "group"ed sign width (obviously, it is cut in Yami theme):
At least, like this:
```
SourceTreeSubItemCheckBox {
...
	min-width: 12px;
}
```

SuslikV:
Also, the string:
https://github.com/obsproject/obs-studio/blob/61e0d69779882fdbca2f38856237c7cd0bc4ca86/UI/data/themes/Yami.qss#L184
do nothing. It is seems to be leftover from the attempt to round the background color from the _Set Color_, isn't?

If so, then the:
```
*[bgColor="1"]
{
	border-radius: 4px;
}
```
may help (should be set for all 8 colors)

If not - the string can be removed (only Acri theme uses border here, and the style isn't preserved on theme change, for me at least).

Warchamp7 (approved):
Seems like the lesser of all evils for fixing this.

Pending code review I am fine with this being fixed.

cg2121:
Fixed merge conflicts and restored the padding in the scene tree.

GeorgesStavracas:
Can source rows not lose their rounded corners? Can the scene list not lose its current padding?

GeorgesStavracas:
Apparently the hardcoded heights do not scale on DPI changes.

Warchamp7 (changes requested):
We'll need to find a different approach that doesn't involve hardcoding height, so as not to break due to DPI scaling

cg2121:
@GeorgesStavracas could you post a screenshot of what they look like under high dpi?

cg2121:
I finally figured out a way of doing this without hardcoding the heights.

@GeorgesStavracas could you confirm it works with different dpis?

cg2121:
Fixed merge commits.

RytoEX:
Here's how this PR currently looks on my laptop screen, which uses a 125% custom scaling factor.

image

GeorgesStavracas:
Thanks @RytoEX, this looks good to me now - I wonder if the QSS changes here affect the menu style though, @cg2121 could you confirm menus still have the same style as of lists?

Warchamp7 (approved):

cg2121:
@GeorgesStavracas the menu is styled correctly.

Screenshot from 2022-08-24 21-38-18

RytoEX:
What was the purpose of this change after an approval was given? It looks like some stylistic preference rather than a functional change, so it _shouldn't affect anything, unless there's some importance to the order of identifiers here. At minimum, please point out that you pushed a change and what it was for so we don't miss the fact that a change was pushed after approval, so that we can ascertain if we need to retest or re-review.

https://github.com/obsproject/obs-studio/compare/2ae773a39ecf328c1db79f873450e81aeec7a430..76e8c3da14468c480afcfd9326b9bf85642183ef

cg2121:
It was because of a bad rebase, as I didn't realize a line was removed.

RytoEX:
> It was because of a bad rebase, as I didn't realize a line was removed.

Ah, I see. Some lines were reordering, some were adding in the accidentally removed lines.

UI/source-tree.hpp
```diff
@@ -25,7 +26,7 @@ class SourceTreeSubItemCheckBox : public QCheckBox {
 	Q_OBJECT
 };
 
-class SourceTreeItem : public QWidget {
+class SourceTreeItem : public QFrame {
```
SuslikV, line 29: Changing the UI code after the release candidate issued is bad practice.

PR #6992, opened 2022-08-07
obs-filters: Support HDR AI greenscreen

jpark37:
### Description
Add HDR support to AI greenscreen.

### Motivation and Context
More HDR support.

### How Has This Been Tested?
Checked SDR image/camera, and HDR image/video.

Admittedly, I didn't check every color space combination, but the others should be rare, and follow the same pattern as other filters. We can fix bugs if they come up.

### Types of changes

### Checklist:

### DIFF
--- a/plugins/obs-filters/data/color.effect
+++ b/plugins/obs-filters/data/color.effect
@@ -1,3 +1,13 @@
+float srgb_linear_to_nonlinear_channel(float u)
+{
+	return (u <= 0.0031308) ? (12.92 * u) : ((1.055 * pow(u, 1. / 2.4)) - 0.055);
+}
+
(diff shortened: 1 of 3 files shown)


pkviet (approved):
Couldn't test HDR sources since I don't have the hardware. I checked that the current code still works fine with SDR sources (browser, webcam, media sources, images).
Although technically it's an enhancement, it could be seen also as a bug fix since  HDR is officially supported in obs 28, but there wasn't before this PR any code either supporting it or warning of non support of HDR as with other filters.
So I'd advocate for this being included in the next beta.
Pending further tests from people with HDR sources.
Codewise, I can't comment on the shader part since I know almost nothing about HDR.

PR #7006, opened 2022-08-09
UI: Fix color select buttons with Yami

cg2121:
### Description
The color select buttons would not be rendered properly with the Yami based themes.

Before:
Screenshot from 2022-08-09 07-33-15

After:
Screenshot from 2022-08-09 07-26-50

### Motivation and Context
Found bug when using color select

### How Has This Been Tested?
Set source color to make sure the buttons looked good

### Types of changes

### Checklist:

### DIFF
--- a/UI/forms/ColorSelect.ui
+++ b/UI/forms/ColorSelect.ui
@@ -6,8 +6,8 @@
    <rect>
     <x>0</x>
     <y>0</y>
-    <width>98</width>
-    <height>61</height>
+    <width>126</width>
+    <height>70</height>
    </rect>
   </property>
   <property name="sizePolicy">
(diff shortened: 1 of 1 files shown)


RytoEX:
I cannot find an entry in the Qt docs for a "toolButton" property for QPushButtons. Do you have any references available? What is it for?

cg2121:
@Warchamp7 added it with https://github.com/obsproject/obs-studio/commit/8a3abf1de78e7c52074155a5c7d94afa72872bdf

Warchamp7:
The goal of this fix is fine, but I'm not sure if the toolButton styling property is the correct way to go about it. Feels like a hack

Can this not be accomplished with UI file changes alone?

cg2121:
Since we need to set padding on the buttons, toolButton was set to true, so we don't have to add to the qss files, since it is already there. The toolButton property works good for buttons without text, like the color buttons are.

Warchamp7 (approved):

PR #5174, opened 2021-08-23
UI: Clean up Settings Output form

Warchamp7:
### Description
This is an overhaul of the UI markup for the Output pane of the Settings window

- Adds new frames around certain elements for a better element hierarchy
- Places the categories of settings into group boxes
- Moves the scroll area from on the entire page into the individual Advanced tabs themselves
- Adjusts margins on the various container frames so that all sections line up

I believe this PR also fixes #4858 

## Before
image

## After
image

## Before
image

## After
image

image
_The recording tab scrolled down doesn't hide the output mode dropdown, navigation tabs or recording type dropdown_

### Motivation and Context
The scrollbar being on the entire output page container meant that the six entries in the Audio tab could cause the other tabs to be scrollable for no reason. This also meant that scrolling down would scroll the top level dropdown and tab controls out of view.

### How Has This Been Tested?
Checked that all tabs still functioned as expected.
Checked that all toggles that add/remove warnings, or enable/disable certain controls still functioned correctly.

### Types of changes
- New feature (non-breaking change which adds functionality)

### Checklist:
- [x] My code has been run through [clang-format](https://github.com/obsproject/obs-studio/blob/master/.clang-format).
- [x] I have read the [**contributing** document](https://github.com/obsproject/obs-studio/blob/master/CONTRIBUTING.rst).
- [x] My code is not on the master branch.
- [x] The code has been tested.
- [x] All commit messages are properly formatted and commits squashed where appropriate.
- [x] I have included updates to all appropriate documentation.

PR #7007, opened 2022-08-09
UI: Fix double scrollbar in settings

cg2121:
### Description
The settings would sometimes show a scrollbar with the
properties view (ie. encoder section).

### Motivation and Context
Reported here: https://github.com/obsproject/obs-studio/pull/5174#issuecomment-1200128285

### How Has This Been Tested?
It doesn't break anything. I have never been able to reproduce this, but at least in theory it should fix the problem.

### Types of changes

### Checklist:

### DIFF
--- a/UI/window-basic-settings.cpp
+++ b/UI/window-basic-settings.cpp
@@ -938,6 +938,9 @@ OBSBasicSettings::OBSBasicSettings(QWidget *parent)
 
 	UpdateAudioWarnings();
 	UpdateAdvNetworkGroup();
+
+	connect(App(), &OBSApp::StyleChanged, this,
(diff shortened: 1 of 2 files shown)


cg2121:
@SuslikV does this by chance fix https://github.com/obsproject/obs-studio/pull/5174#issuecomment-1200128285 ?

cg2121:
I've now been able to reproduce this. It happens when switching themes, as they have different widget sizes, scrollbars get added because the widgets no longer fit the original parent size.

I've now added code to recalculate the properties view's height when themes are changed, so scrollbars aren't added.

GeorgesStavracas:
We've discussed this today and the general feeling towards this is that we'd rather not land any workarounds, and eventually report issues like this upstream. Thanks for the contribution though!

PR #7008, opened 2022-08-09
libobs-opengl: Check window creation for errors

kkartaltepe:
### Description

Without this we get somewhat misleading errors from later in the
swapchain creation when this fails. This is currently happening for
users attempting to use multi-gpu with prime offloading on nvidia.

### Motivation and Context

Makes the error more clear, in this case the user identified it was a bad config and not something later in X11 which the old error suggested. This is a new regression as of the beta release, so we should report the correct error before things get released.

### How Has This Been Tested?

It works on my machine and changing the create call to fail it reports the error correctly.

### Types of changes

### Checklist:


GeorgesStavracas (approved):
Looks good, thanks for the patch

PR #7009, opened 2022-08-09
mac-videotoolbox: Don't manually override color space

gxalpha:
### Description

This was done to work around an issue in 2015 which does not appear to
exist anymore in 2022. Thus, remove it to remove complexity, especially
if we want to explicitly add HDR support (even though that already
mostly works).

### Motivation and Context

Less unnecessary code is better.

### How Has This Been Tested?

macOS 13
Confirmed with `CVBufferCopyAttachments` that all of the attributes are already correctly
set on the buffer.
image

Tested with H.264 Software, H.264 Hardware, and HEVC Hardware encoders.

### Types of changes

### Checklist:

### DIFF
--- a/plugins/mac-videotoolbox/encoder.c
+++ b/plugins/mac-videotoolbox/encoder.c
@@ -789,22 +789,6 @@ bool get_cached_pixel_buffer(struct vt_encoder *enc, CVPixelBufferRef *buf)
 
 	CVPixelBufferRef pixbuf;
 	STATUS_CHECK(CVPixelBufferPoolCreatePixelBuffer(NULL, pool, &pixbuf));
-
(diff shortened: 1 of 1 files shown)


PatTheMav (approved):
Confirmed on macOS 12.5 that the values are contained by default. Will still need to check on macOS 10.15.

gxalpha:
Closing.
While these are indeed set *usually*, it appears that when some color *formats* are selected they might not be (For example it's set for NV12, but not when switching to P010 it isn't). Unsure why the color format would play a role here but that's how it is. Sucks that we have to carry this code but looks like it might be unavoidable.
Sorry for wasting your time @PatTheMav, this is something I only noticed now.

PR #7019, opened 2022-08-10
libobs-opengl,UI: Use Metal swap chain for non-blocking vsync

jpark37:
### Description
Using the Metal swap chain allows us check when the swap chain is
ready, so the graphics thread won't get stuck waiting on vsync.

There's also dormant HDR code that isn't turned on right now because it causes Qt to flicker like crazy.

### Motivation and Context
Want preview/projectors to not tear, and also not block the graphics thread from making progress.

### How Has This Been Tested?
Preview and projectors seem to work. Resize is fine. Ran temporary instrumentation to test present is skipped alternating for 120 FPS video on 60 Hz refresh.

Vsync settings are gone from Settings dialog, and tested that that a setting change still works.

### Types of changes

### Checklist:

### DIFF
--- a/UI/forms/OBSBasicSettings.ui
+++ b/UI/forms/OBSBasicSettings.ui
@@ -6643,7 +6643,7 @@
                     <widget class="QLabel" name="label_30">
                      <property name="minimumSize">
                       <size>
-                       <width>0</width>
+                       <width>170</width>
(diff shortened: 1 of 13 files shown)


jpark37:
Retracting. Too many issues. Waiting for real Metal renderer.

PR #7029, opened 2022-08-11
CI: Switch to included Xcode 14 Beta

derrod:
### Description

Set Xcode 14 Beta as the active version on CI.

### Motivation and Context

GitHub updated the runner image to include the 14 beta, so we no longer need to manually install it, saving us about 30 minutes of build time.

### How Has This Been Tested?

CI ran successfully.

### Types of changes

### Checklist:

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -80,7 +80,6 @@ jobs:
       BLOCKED_FORMULAS: 'speexdsp curl php composer'
       CODESIGN_IDENT: '-'
       HAVE_CODESIGN_IDENTITY: ${{ secrets.MACOS_SIGNING_IDENTITY != '' && secrets.MACOS_SIGNING_CERT != '' }}
-      HAVE_XCODE_URL: ${{ secrets.XCODE_DOWNLOAD_URL != '' }}
     defaults:
       run:
         shell: bash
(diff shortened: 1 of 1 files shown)


.github/workflows/main.yml
```diff
@@ -156,9 +155,8 @@ jobs:
 
           echo "::set-output name=commitHash::$(git rev-parse --short=9 HEAD)"
 
-      - name: 'Install Xcode 14 Beta'
-        if: ${{ startsWith(github.ref, 'refs/tags/') && github.event_name != 'pull_request' && env.HAVE_CODESIGN_IDENTITY == 'true' && env.HAVE_XCODE_URL == 'true' }}
-        run: xcversion install "14 beta 4" --url=${{ secrets.XCODE_DOWNLOAD_URL }}
+      - name: 'Swwitch to Xcode 14 Beta'
```
gxalpha, line 158: Swwitch

RytoEX: Do we need the word "Beta" here? Eventually, Xcode 14 will not be beta, and I don't think the xcode-select invocation will change.

derrod: One it is out of beta it will probably become the default and the entire command can be removed.

RytoEX: They don't make it the default right away. May 17, they added Xcode 13.4, but 13.3.1 was default. June 9, they added Xcode 13.4.1, but 13.3.1 remained the default. June 28, they finally announce that Xcode 13.4.1 will become the default on July 11.

Not blocking, but wanted to clarify expected update timetables.

derrod: We may still have to update this manually whenever Xcode 14 stable comes out due to point releases. So we can remove the "Beta" whenever that happens, and then fully remove it once 14 becomes the default.

PR #7032, opened 2022-08-11
obs-filters: Add HDR Tonemap filter

jpark37:
### Description
Allow per-source tonemapping to override default tonemapping.

### Motivation and Context
Multiple HDR sources may need to be tonemapped differently to canvas SDR/HDR space, e.g. HDR camera & HDR game with different ranges. Inserting the tonemap filter manually also allows users to apply SDR filters after conversion.

### How Has This Been Tested?
Verify in RenderDoc:
- [x] EDR source, SRGB canvas, EDR 1000 -> EDR 1000
- [x] EDR source, SRGB canvas, EDR 1000 -> EDR 100
- [x] EDR source, SRGB canvas, EDR 1000 -> EDR 10
- [x] EDR source, SRGB canvas, EDR 1000 -> SRGB 80
- [x] EDR source, SRGB canvas, EDR 1000 -> SRGB 300
- [x] EDR source, SRGB canvas, EDR 1000 -> SRGB 480
- [x] EDR source, SRGB16 canvas, EDR 1000 -> EDR 1000
- [x] EDR source, SRGB16 canvas, EDR 1000 -> EDR 100
- [x] EDR source, SRGB16 canvas, EDR 1000 -> EDR 10
- [x] EDR source, SRGB16 canvas, EDR 1000 -> SRGB16 80
- [x] EDR source, SRGB16 canvas, EDR 1000 -> SRGB16 300
- [x] EDR source, SRGB16 canvas, EDR 1000 -> SRGB16 480
- [x] EDR source, EDR canvas, EDR 1000 -> EDR 1000
- [x] EDR source, EDR canvas, EDR 1000 -> EDR 100
- [x] EDR source, EDR canvas, EDR 1000 -> EDR 10
- [x] EDR source, EDR canvas, EDR 1000 -> SRGB16 80
- [x] EDR source, EDR canvas, EDR 1000 -> SRGB16 300
- [x] EDR source, EDR canvas, EDR 1000 -> SRGB16 480
- [x] SRGB16 source, HDR 1000 -> HDR 1000 (ignore, passthrough SRGB16)
- [x] SRGB16 source, HDR 1000 -> HDR 100 (ignore, passthrough SRGB16)
- [x] SRGB16 source, HDR 1000 -> HDR 10 (ignore, passthrough SRGB16)
- [x] SRGB16 source, HDR 1000 -> SDR 80 (ignore, passthrough SRGB16)
- [x] SRGB16 source, HDR 1000 -> SDR 300 (ignore, passthrough SRGB16)
- [x] SRGB16 source, HDR 1000 -> SDR 480 (ignore, passthrough SRGB16)
- [x] SRGB source, HDR 1000 -> HDR 1000 (ignore, passthrough SRGB)
- [x] SRGB source, HDR 1000 -> HDR 100 (ignore, passthrough SRGB)
- [x] SRGB source, HDR 1000 -> HDR 10 (ignore, passthrough SRGB)
- [x] SRGB source, HDR 1000 -> SDR 80 (ignore, passthrough SRGB)
- [x] SRGB source, HDR 1000 -> SDR 300 (ignore, passthrough SRGB)
- [x] SRGB source, HDR 1000 -> SDR 480 (ignore, passthrough SRGB)

### Types of changes

### Checklist:


plugins/obs-filters/data/locale/en-US.ini
```diff
@@ -3,6 +3,7 @@ ColorGradeFilter="Apply LUT"
 MaskFilter="Image Mask/Blend"
 AsyncDelayFilter="Video Delay (Async)"
 CropFilter="Crop/Pad"
+HdrTonemapFilter="HDR Tonemap"
```
gxalpha, line 6: Just on the wording here, could it be made more obvious that using this filter is optional (perhaps "HDR Tonemap Override")?
I’m a bit worried that otherwise users could be under the impression (perhaps influenced by bad YouTube tutorials) that they need this for every HDR source if their OBS is set to SDR.

WizardCM: I was thinking Override too.

jpark37: I went with "HDR Tonemap (Override)"

gxalpha: wonderful, thanks

jpark37:
Another benefit of this is that you can use SDR filters in the chain after tonemapping (when PR #7037 goes through).

SuslikV:
Maybe "Output" can be replaced by "Convert to"?

jpark37:
Does OBS ever use a preposition to lead into a UI control?

I've considered "Tonemapper" and "Tone Mapper," but I'm not sure if it needs a space or not. I'm looking at Real-TIme Rendering 4, and they seem to consistently use "Tone Transform," so I'll probably update the PR with that when I get the chance.

SuslikV:
I'm seeing the tone mapping as one of the possible algorithms to reduce the amount of data. I have seen it named _Tone Mapping Operator_ (here https://64.github.io/tonemapping/ ). I was thinking that, maybe it not sounds good for the end users (who are not programmers) and the "Output" had no relation to the algorithm itself (but fortunately points to the destination -> the output of this filter is will be converted according the selected algorithm).

mufunyo:
"Apply LUT", "Crop/Pad", "Scroll" and "Sharpen" all use the infinite verb, so "Tone Map" sounds most correct to me.

jpark37:
Wouldn't the pattern be "Map Tones"? :D

mufunyo:
Haha, that's a good point! :D

jpark37:
If we really want to get down to it, this is why I settled on "HDR Tone Mapping":

- HDR is important enough to be at the front because the operations will only work on an HDR input.
- I want the syllables "tone" and "map" to be adjacent, which is why "Map HDR Tones" would be quite weird despite being in pattern. This isn't a phrase anyone has ever spoken in real life.
- The header of the relevant section in RTR4, the definitive aggregator of all things rendering, is "Tone Mapping"
- The compound words tonemapping, tonemapper, and tonemap feel too colloquial and "made up" to be used for UI strings although I do use tonemap in the code.

Even though this may be out of pattern, I feel like the alternatives are worse.

SuslikV:
"Tone Mapping" easier to find and thus easier to translate.

The Limited to Full range in SDR technically tone mapping too, isn't it? (General question not related to the PR)

mufunyo:
> The Limited to Full range in SDR technically tone mapping too, isn't it? (General question not related to the PR)

Tone mapping usually refers to a spatial transform that modifies global contrast while trying to maintain local contrast. By this definition, HDR conversion LUTs are not tone mapping. By your definition any levels adjustment would be tone mapping.

I don't know exactly which definition is technically correct, but I think it helps to maintain the distinction to prevent confusion.

SuslikV:
@mufunyo The HDR signal that goes into 8 bit filter is already fitted into 8 bits (16 bits clipped to 8 bits; clipping - is your tone mapping operator). What your filter will do with these 8 bits - doesn't matter (you can adjust levels or what ever you wish). The Limited vs Full range in SDR has ~6M colors difference. You mapping brightest values to the brightest possible and darkest to the darkest possible. You cannot increase number of color "bands" on the Limited display (it can do only ~10M). Of course, you can ignore the mapping and get dimmed image because you throw away the values that were above the 235/240 and below 16.

PR #7038, opened 2022-08-11
Sorry Video

TheTabKey:
adding Sorry Video FTL


RytoEX:
We are no longer accepting new services that use FTL. Please see [our post on Deprecating FTL](https://github.com/obsproject/obs-studio/discussions/4021).

PR #7043, opened 2022-08-12
libobs: Fix gpu thread termination when additional video mixes are added

chippydip:
### Description
Fixes a bug where enabling the new virtual camera when GPU encoding had already started (for streaming or recording) would cause the GPU thread to exit early. This early exit would result in the output getting stuck in the stopping state later.

### Motivation and Context
Fixes bug reported in v28 beta 2

### How Has This Been Tested?
Repro steps:
- Start recording or streaming with a GPU encoder
- Start he virtual camera (this silently stops the GPU encoder thread)
- Stop the virtual camera (optional)
- Stop the recording or stream -> causes the UI to get stuck in a "Stopping..." state

Re-tested after the fix and was no longer able to reproduce.

The source of the bug was a DARRAY of video mixes containing structs rather than struct pointers. The GPU encoder thread holds a pointer to one of these structs which is invalidated when the DARRAY is reallocated to append a new video mix (which happens when the virtual camera is enabled). This appears to be the only place a pointer to this array is held outside of the mutex so the bug can only be observed when the GPU encoder is running and a new video mix is added causing a realloc.

### Types of changes

### Checklist:

### DIFF
--- a/libobs/obs-internal.h
+++ b/libobs/obs-internal.h
@@ -294,8 +294,8 @@ struct obs_core_video_mix {
 	enum obs_scale_type scale_type;
 };
 
-extern int obs_init_video_mix(struct obs_video_info *ovi,
-			      struct obs_core_video_mix *video);
+extern struct obs_core_video_mix *
+obs_create_video_mix(struct obs_video_info *ovi);
 extern void obs_free_video_mix(struct obs_core_video_mix *video);
 
 struct obs_core_video {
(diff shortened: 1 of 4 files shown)


flaeri:
I can confirm this fixed the following:
* texture-amf
* fallback-amf
* x264

Cleanly stops the recording, does not hang on shutdown, and the recording has content after initiating vcam.

PR #7045, opened 2022-08-12
UI: Fix theme leak

pkviet:
### Description
This fixes leaks when loading and switching themes.

### Motivation and Context
Leaks are bad.
This leak was detected with Visual Leak Detector.

### How Has This Been Tested?
Run obs with VLD; check that the leak is gone after changing a theme.

### Types of changes

### Checklist:

### DIFF
--- a/UI/obs-app.cpp
+++ b/UI/obs-app.cpp
@@ -1203,8 +1203,8 @@ bool OBSApp::SetTheme(std::string name, std::string path)
 	path = GetTheme(name, path);
 	if (path.empty())
 		return false;
-
-	themeMeta = ParseThemeMeta(path.c_str());
+	unique_ptr<OBSThemeMeta> themeMeta;
+	themeMeta.reset(ParseThemeMeta(path.c_str()));
 	string parentPath;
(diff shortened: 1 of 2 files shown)


notr1ch:
Is there a reason `themeMeta` is part of `OBSApp`? As far as I can tell, the only place it is used is within `OBSApp::SetTheme`, so there doesn't appear to be a need to keep it around after the function exits.

pkviet:
Good point, i'll move the delete then

notr1ch:
I was thinking more about removing it from the class and having it as a local stack variable.

pkviet:
updated to follow @notr1ch advice; this is much cleaner, thx !

pkviet:
> I was thinking more about removing it from the class and having it as a local stack variable.

Ah i see. I'll let Matt fix that since he authored that code. 
Dunno what's the reason indeed to have the var in the whole class.

pkviet:
Changed to a unique_ptr; we'll see if Matt was thinking of using the var elsewhere in future code.

PR #7048, opened 2022-08-13
Remove unnecessary unicode characters and BOMs

norihiro:
### Description

Removes unnecessary UTF characters and replaces them with closest ASCII character(s).
- `°` in comments are replaced with `degree`.
- `’`, `“`, and `”` are replaced with `'`, `"`, and `"`, respectively.
- `–` is replaced with `-`.

Also removes BOMs in source codes.

These files are not modified to avoid changing existing behavior or another reason.
- UI/properties-view.cpp
  - NBSP is not modified.
- libobs/util/profiler.c
  - `µ` and `±`
- plugins/linux-v4l2/v4l2-decoder.c
- plugins/linux-v4l2/v4l2-decoder.h
  - It's a name of the author.
- UI/forms/OBSBasicTransform.ui
  - Only `°`, which should not be modified.
- cmake/bundle/windows/ALL_BUILD.vcxproj.user.in
  - Has BOM but I don't know it is ok to remove the BOM.

### Motivation and Context

Want to have ASCII source code as possible.

### How Has This Been Tested?

Eyeball checked modifications other than BOMs are only in the comments.

All source files are checked with a bash command below on Fedora 35.
```
git ls-files | grep -E '\.(cc|hh|in|ui|sh|h|c|cpp|hpp)$' | while read f; do test "$(nkf -g $f)" == ASCII || echo $f; done
```

### Types of changes

### Checklist:

### DIFF
--- a/UI/properties-view.cpp
+++ b/UI/properties-view.cpp
@@ -899,7 +899,7 @@ static bool matches_ranges(media_frames_per_second &best_match,
 	for (auto &pair : fps_ranges) {
 		auto max_ = convert_fn(pair.first);
 		auto min_ = convert_fn(pair.second);
-		/*blog(LOG_INFO, "%lg ≤ %lg ≤ %lg? %s %s %s",
+		/*blog(LOG_INFO, "%lg <= %lg <= %lg? %s %s %s",
 				min_, val, max_,
 				fabsl(min_ - val) < epsilon ? "true" : "false",
 				min_ <= val && val <= max_  ? "true" : "false",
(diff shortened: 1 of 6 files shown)


notr1ch:
If all of our supported platforms have compilers that can process UTF-8 source files, I would prefer adding UTF-8 signatures to the files or specifying appropriate compiler flags (e.g. [/utf8 on MSVC](https://docs.microsoft.com/en-us/cpp/build/reference/utf-8-set-source-and-executable-character-sets-to-utf-8?view=msvc-170)) to avoid character set warnings (which are now treated as errors).

Some more thoughts and notes: https://www.open-std.org/jtc1/sc22/wg21/docs/papers/2021/p2295r4.pdf

norihiro:
Removed my modification in `plugins/obs-ffmpeg/external/nvEncodeAPI.h` since this is a file just copied from an SDK.

PR #7052, opened 2022-08-13
obs-qsv11: Do not apply limits if CPU generation is unknown

derrod:
### Description

Do not restrict QSV resolution if CPU ID lookup returns "Unknown".

This should be removed and replaced with a proper check once the encoder implementation gets reworked for OneAPI-VPL and AV1 support.

### Motivation and Context

Users with Intel dGPU (Arc) and AMD CPU would find themselves restricted to 1200p for no reason.

### How Has This Been Tested?

Hasn't.

### Types of changes

### Checklist:


derrod:
Note: this does not fix a number of other restrictions that are placed upon Arc GPUs without Intel CPU, but those I'm not entirely comfortable changing blindly without Arc hardware on hand.

PR #7077, opened 2022-08-15
libobs: Fix the crash when Async Delay Filter on

### LINKED ISSUE #6188: Crash when using with Video Delay (Async)
### Operating System Info

Windows 10

### Other OS

_No response_

### OBS Studio Version

27.2.3

_No response_

### OBS Studio Log URL

https://obsproject.com/logs/s0_fUTqQxiW9Q9EA

### OBS Studio Crash Log URL

_No response_

### Expected Behavior

-

### Current Behavior

When I change track from VLC playlist (prev, next) after adding Video Delay (Async) to VLC source, I get a Crash error. 

### Steps to Reproduce

1. Add a VLC source
2. Add more than one file to the playlist
3. Add a Video Delay filter to the VLC source
4. Switch tracks until a crash occurs (usually only a few times necessary)
...

_No response_

tuduweb:
### Description

~~1. Add function to limit texture filling size to avoid memcpy overflow in memcpy~~

Correctly set the async texture size,
based on the frame to be rendered on this loop,
rather than the newest frame from encoder when async delay filter on.

### Motivation and Context

1. Fixes #6188
2. And fixes crash when preload frame remain but the source texture changed
(by media file changed or other situation, found in debugging other issues)

When using Video Delay filter, after the source media changed(especially when the media dimension size get bigger),
the frame pop from the delay filter with the old size, but the texture is changed by the newest frame.

https://github.com/obsproject/obs-studio/blob/bad54448ceb65b015a3b2c812a40920f6f78d8b1/libobs/obs-source.c#L2377-L2381

When occurs the mismatch between frame and texture, the crash happens.

The `set_async_texture_size` should be called after the old deleyed frame popped.

### How Has This Been Tested?

Win10
Reproduce those steps mentioned on issue, no crash happen.

### Types of changes

### Checklist:

### DIFF
--- a/libobs/obs-source.c
+++ b/libobs/obs-source.c
@@ -1206,10 +1206,6 @@ static void async_tick(obs_source_t *source)
 
 	source->last_sys_timestamp = sys_time;
 	pthread_mutex_unlock(&source->async_mutex);
-
-	if (source->cur_async_frame)
-		source->async_update_texture =
-			set_async_texture_size(source, source->cur_async_frame);
 }
 
 void obs_source_video_tick(obs_source_t *source, float seconds)
@@ -1975,9 +1971,6 @@ static inline bool init_gpu_conversion(struct obs_source *source,
 bool set_async_texture_size(struct obs_source *source,
 			    const struct obs_source_frame *frame)
 {
-	enum convert_type cur =
-		get_convert_type(frame->format, frame->full_range, frame->trc);
-
 	if (source->async_width == frame->width &&
 	    source->async_height == frame->height &&
 	    source->async_format == frame->format &&
@@ -2007,7 +2000,9 @@ bool set_async_texture_size(struct obs_source *source,
 
 	const enum gs_color_format format =
 		convert_video_format(frame->format, frame->trc);
-	const bool async_gpu_conversion = (cur != CONVERT_NONE) &&
+	const enum convert_type type =
+		get_convert_type(frame->format, frame->full_range, frame->trc);
+	const bool async_gpu_conversion = (type != CONVERT_NONE) &&
 					  init_gpu_conversion(source, frame);
 	source->async_gpu_conversion = async_gpu_conversion;
 	if (async_gpu_conversion) {
@@ -2382,6 +2377,9 @@ static void obs_source_update_async_video(obs_source_t *source)
 
 		source->async_rendered = true;
 		if (frame) {
+			const bool async_update_texture =
+				set_async_texture_size(source, frame);
+
 			check_to_swap_bgrx_bgra(source, frame);
 
 			if (!source->async_decoupled ||
@@ -2391,11 +2389,10 @@ static void obs_source_update_async_video(obs_source_t *source)
 				source->timing_set = true;
 			}
 
-			if (source->async_update_texture) {
+			if (async_update_texture) {
 				update_async_textures(source, frame,
 						      source->async_textures,
 						      source->async_texrender);
-				source->async_update_texture = false;
 			}
 
 			obs_source_release_frame(source, frame);
--- a/libobs/obs-internal.h
+++ b/libobs/obs-internal.h
@@ -744,7 +744,6 @@ struct obs_source {
 	bool async_flip;
 	bool async_linear_alpha;
 	bool async_active;
-	bool async_update_texture;
 	bool async_unbuffered;
 	bool async_decoupled;
 	struct obs_source_frame *async_preload_frame;


jpark37:
I would just fix the mismatch where it's happening instead of adding a new function.

tuduweb:
> I would just fix the mismatch where it's happening instead of adding a new function.

Thanks for your advice.

Yea, it's better to just fix the mismatch, but the texture size has already been changed when filling the frame image to texture.

And the old function can't access the frame size, just pass frame plane data width(linesize), it's hard to fix in the old function.

If just avoid the texture size update when still have old frame to avoid mismatch, it's link to so many procedure.
I think this is the appropriate way.

I'm really happy to receive your opinion, and it is very helpful to me.
Hope to your reply and advice, thx~

jpark37:
The frame dimensions are part of obs_source_frame, and you can get the texture size with gs_texture_get_width/gs_texture_get_height. Can you test, and maybe resize the texture before calling gs_texture_set_image? Which function call has the mismatch?

tuduweb:
> The frame dimensions are part of obs_source_frame, and you can get the texture size with gs_texture_get_width/gs_texture_get_height. Can you test, and maybe resize the texture before calling gs_texture_set_image? Which function call has the mismatch?

it seems not work as expected.(*resize the texture before calling gs_texture_set_image*)
After the texture size changed, the renderer will reset the preview to the newest frame with new size, and also draw `SceneEditing` and `SpacingHelpers`. It just reset(and refresh the preview) again and again within the video delay time set by user.

*Which function call has the mismatch?*
The `async-delay-filter`, receive and push the newest frame to a queue, then pop and output the old frame.
But the source only focus on the newest frame, not care about the delay filter.
This cause the mismatch.

I spent the past half day and tried to fix it, but i didnt find a good solution.  :(

https://user-images.githubusercontent.com/30559085/184940879-9e9a82a7-3dac-499c-af4b-a026e5f51c8b.mp4

jpark37:
I'll look at this later today.

tuduweb:
Thx. And there may be some UX(maybe?) issues here.
Do the source size should be updated when the old frames not gone(all pop out from queue)?

In my opinion, the size shouldn't be updated.
Maybe it needs more other changes after fixing the overflow crash.

jpark37:
Try this: https://github.com/jpark37/obs-studio/commit/c2cf93a53d2bcff173ea7992138471fdeb020da0

tuduweb:
It seems to work fine after patching.

jpark37:
Sounds good. Just update your branch with my changes.

jpark37:
Or I can make a new PR if you'd rather not.

tuduweb:
The changes seems to intro a new issue. I'm debuging now.
I'll give details later.

tuduweb:
The SelectBox and HelperLines seems to not keep up with the renderer changed about 1 frame after patching:
image

And this has not happened before
(*without delay filter*, *with delay filter 0ms/1000ms*).

jpark37:
It's only wrong for one frame? I think that's fine.

tuduweb:
Yep, only one frame.
They are not updated together in one graphic loop after this.

tuduweb:
I updated the branch and PR,
and made some other tiny changes. (I think it maybe improve some performance?)

libobs/obs-source.c
```diff
@@ -2391,12 +2389,10 @@ static void obs_source_update_async_video(obs_source_t *source)
 				source->timing_set = true;
 			}
 
-			if (source->async_update_texture) {
+			if (async_update_texture)
```
jpark37, line 2392: Add the braces back. Even though it's a single statement, it's still multiple lines. The braces make it easier to understand.

jpark37:
The one-frame issue might be fixable with a new filter callback that happens at tick time instead of render time, but I'm fine with this for now. Not sure what the OBS devs will think though.

libobs/obs-source.c
tuduweb: I got it.

tuduweb:
> The one-frame issue might be fixable with a new filter callback that happens at tick time instead of render time, but I'm fine with this for now. Not sure what the OBS devs will think though.

This introduces one-frame issue,(just a moment render on user preview, not on output and record)
but fixes overflow bug now.

jp9000:
I'm going to revert this for now -- I shouldn't have merged this one. You'll have to remake this PR -- if you do, please make sure to reference this PR from the new one.

jpark37:
It also might work to move the filter callback into async_tick to avoid the one-frame issue. I might just make the PR myself.

tuduweb:
got it.
Hope to be fixed one day.~

PR #7080, opened 2022-08-15
obs-qsv11: Prevent 8-bit HDR, and 10-bit anything

jpark37:
### Description
Don't want to silently generate lossy video.

### Motivation and Context
Don't want users generating unintentionally low-quality videos.

### How Has This Been Tested?
Verified QSV H.264 can no longer record 8-bit HDR, and the error message looks as expected.

Verified QSV H.264 can no longer record 10-bit anything, and the error message looks as expected.

### Types of changes

### Checklist:


mufunyo:
Is compositing in 10-bit SDR and encoding in 8-bit still allowed? There are some advantages over 8 bit compositing still, and SDR is fine to be encoded in 8-bit.

jpark37:
No, 10-bit SDR (actually 16-bit) and encoding in 8-bit is not still allowed. Probably need some UI work If we want to bring it back proper.

PR #7082, opened 2022-08-15
CI: remove UNIX_STRUCTURE from FreeBSD Cirrus-CI config

emaste:
### Description
Update FreeBSD .cirrus.yml to stop setting UNIX_STRUCTURE.

### Motivation and Context
UNIX_STRUCTURE was deprecated in df94b0207563.

### How Has This Been Tested?
Cirrus-CI run: https://cirrus-ci.com/task/4900659664257024

### Types of changes

### Checklist:

### DIFF
--- a/.cirrus.yml
+++ b/.cirrus.yml
@@ -20,7 +20,6 @@ task:
   - mkdir build
   - cd build
   - cmake
-      -DUNIX_STRUCTURE=1
       -DENABLE_AJA=OFF
       -DENABLE_ALSA=OFF
       -GNinja ..


.cirrus.yml
```diff
@@ -20,7 +20,6 @@ task:
   - mkdir build
   - cd build
   - cmake
-      -DUNIX_STRUCTURE=1
```
RytoEX, line 23: UNIX_STRUCTURE was deprecated and replaced by LINUX_PORTABLE. To maintain equivalent functionality, this should probably replace UNIX_STRUCTURE with LINUX_PORTABLE. Will defer to @PatTheMav or someone else more familiar with Linux/*nix/FreeBSD though if they have a different conclusion.

RytoEX (changes requested):
Please use sentence case for the first word after the commit message subject prefix. For example:
```
CI: Remove UNIX_STRUCTURE from FreeBSD Cirrus-CI config
```

.cirrus.yml
emaste: I think it's fine for `LINUX_PORTABLE` to be left at the default here -- IIRC I added `-DUNIX_STRUCTURE=1` originally to get the Cirrus-CI build to complete successfully, but Cirrus-CI is green with this PR. I wanted Cirrus-CI basically to make sure that cmake & build succeeds on FreeBSD with settings as close as possible to default.

Of course, if setting `LINUX_PORTABLE` is the desired config I'm happy to make that change.

RytoEX: Ah, my mistake. The params have swapped meanings/values. So UNIX_STRUCTURE=1 (ON/TRUE) is equivalent to LINUX_PORTABLE=0 (OFF/FALSE). This can be left as-is then.

https://github.com/obsproject/obs-studio/blob/28.0.0-beta2/cmake/Modules/DeprecationHelpers.cmake#L128-L132

PatTheMav: `UNIX_STRUCTURE` is the default with the new flags, so passing it should not have made much of an effective difference, but nevertheless we'd rather merge a change that does not rely on deprecated flags.

emaste: Yep - this change removes the use of the deprecated flag.

emaste:
> Please use sentence case for the first word after the commit message subject prefix. For example:

Pull request updated with 1d99b2f45160c6c2e1337e6a6c4f93b6fab812bd

PR #7089, opened 2022-08-16
UI: Enforce macOS plugin bundle on M1; don’t load plugin bundles globally

gxalpha:
### Description

macOS OBS plugins in following locations/formats will no longer get loaded with this PR:
- The plugin is a `.plugin` bundle, but located in the global application support directory (`/Library/Application Support/obs-studio/plugins`)
- OBS and the plugin have the arm64 architecture and the plugin has the `/bin` and `/data` folder structure (which effectively means it's an `.so` plugin)

This means that plugins are now only loaded if:
- They are a `.plugin` bundle in the user application support directory
- They are an x86_64 `.so` plugin

Important note: None of what got removed got released in a stable OBS release yet.
This does not break any setups just upgrading from 27.2 to 28.

### Motivation and Context

Modern plugins on macOS are `.plugin` bundles.
They live in the user application support directory, so that they don't require admin
permissions to be installed and don't effect other users. (@PatTheMav will still need to
confirm whether this assessment is correct.)

Now is our best shot at preventing any other behaviour, since by introducing it we could
potentially force ourselves into supporting it for a long time.
Any plugin author following the plugintemplate (which most of them need to do anyways
due to Qt 6) will already have the plugin in the correct format and installer will install it into
the correct location.

### How Has This Been Tested?

macOS 13
- Tested that a plugin bundle (used [obs-properties-dock](https://github.com/gxalpha/obs-properties-dock) to test) from the global applications support directory gets loaded without my changes, and no longer get loaded with them.
- Tested with CI build that an existing, x86_64-only old-style plugin (used [obs-main-view-source](https://github.com/norihiro/obs-main-view-source) to test) gets loaded successfully from both global and user application support (via running Rosetta). Tested that when running arm64 natively, without this PR (plugin in either location) I get an error (`os_dlopen[...] mach-o file, but is an incompatible architecture (have 'x86_64', need 'arm64'`) message in the log (which means that it tries to load the plugin), but with this PR that message doesn't appear (which means that it correctly does not try to load this old-style plugin).

### Types of changes

### Checklist:

### DIFF
--- a/UI/window-basic-main.cpp
+++ b/UI/window-basic-main.cpp
@@ -165,9 +165,12 @@ static void AddExtraModulePaths()
 	}
 
 	char base_module_dir[512];
-#if defined(_WIN32) || defined(__APPLE__)
+#if defined(_WIN32)
 	int ret = GetProgramDataPath(base_module_dir, sizeof(base_module_dir),
 				     "obs-studio/plugins/%module%");
+#elif defined(__APPLE__)
+	int ret = GetConfigPath(base_module_dir, sizeof(base_module_dir),
+				"obs-studio/plugins/%module%.plugin");
 #else
 	int ret = GetConfigPath(base_module_dir, sizeof(base_module_dir),
 				"obs-studio/plugins/%module%");
@@ -179,22 +182,26 @@ static void AddExtraModulePaths()
 	string path = base_module_dir;
 #if defined(__APPLE__)
 	/* User Application Support Search Path */
-	BPtr<char> config_bin = os_get_config_path_ptr(
-		"obs-studio/plugins/%module%.plugin/Contents/MacOS");
-	BPtr<char> config_data = os_get_config_path_ptr(
-		"obs-studio/plugins/%module%.plugin/Contents/Resources");
-	obs_add_module_path(config_bin, config_data);
+	obs_add_module_path((path + "/Contents/MacOS").c_str(),
+			    (path + "/Contents/Resources").c_str());
 
 #ifndef __aarch64__
 	/* Legacy System Library Search Path */
-	obs_add_module_path((path + "/bin").c_str(), (path + "/data").c_str());
+	char system_legacy_module_dir[PATH_MAX];
+	GetProgramDataPath(system_legacy_module_dir,
+			   sizeof(system_legacy_module_dir),
+			   "obs-studio/plugins/%module%");
+	std::string path_system_legacy = system_legacy_module_dir;
+	obs_add_module_path((path_system_legacy + "/bin").c_str(),
+			    (path_system_legacy + "/data").c_str());
 
 	/* Legacy User Application Support Search Path */
-	BPtr<char> config_bin_legacy =
-		os_get_config_path_ptr("obs-studio/plugins/%module%/bin");
-	BPtr<char> config_data_legacy =
-		os_get_config_path_ptr("obs-studio/plugins/%module%/data");
-	obs_add_module_path(config_bin_legacy, config_data_legacy);
+	char user_legacy_module_dir[PATH_MAX];
+	GetConfigPath(user_legacy_module_dir, sizeof(user_legacy_module_dir),
+		      "obs-studio/plugins/%module%");
+	std::string path_user_legacy = user_legacy_module_dir;
+	obs_add_module_path((path_user_legacy + "/bin").c_str(),
+			    (path_user_legacy + "/data").c_str());
 #endif
 #else
 #if ARCH_BITS == 64
--- a/UI/window-basic-main.cpp
+++ b/UI/window-basic-main.cpp
@@ -178,10 +178,6 @@ static void AddExtraModulePaths()
 
 	string path = base_module_dir;
 #if defined(__APPLE__)
-	/* System Library Search Path */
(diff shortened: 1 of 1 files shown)


PatTheMav (commented):
The change itself looks fine, but I'd maybe go a step further:

* Make the user library path the `base_module_dir` and construct the `bin` and `data` path configs accordingly
* Only on x86_64, add the ability to still load `.so` variants
* Only on x86_64, add the system library path as an optional directory

That way we make the arm64 location and format the "default" and the legacy format plus locations are handled in a separate piece of code.

WizardCM:
If we're making such heavy changes to plugin loading on macOS, should we have a section in obsproject.com/docs outlining how it all works?

PatTheMav:
> If we're making such heavy changes to plugin loading on macOS, should we have a section in obsproject.com/docs outlining how it all works?

I'm not opposed to that, but at first glance I couldn't find any mention of how plugins should be packaged in our docs to begin with, so it'd make sense to document this for all existing platforms anyway.

gxalpha:
@PatTheMav 
> Make the user library path the base_module_dir and construct the bin and data path configs accordingly

Done.

> Only on x86_64, add the ability to still load .so variants
Only on x86_64, add the system library path as an optional directory

If I understand you correctly, this is already in this PR.

gxalpha:
> If we're making such heavy changes to plugin loading on macOS, should we have a section in obsproject.com/docs outlining how it all works?

I missed this message in my reply a few minutes ago, but yeah I don't think there has ever been format documentation to how a plugin look like exactly. The only official guidance was the plugintemplate, other than that developers usually just looked at existing plugins and copied their cmake. The plugintemplate at least is updated to follow the new convention.

Such a section in the docs, or perhaps even the wiki of the plugintemplate would definitely make sense (not just for macOS, afaik in theory we also want Windows plugins to be in a different location to where most of them live now(?), and could even include a Flatpak section for Linux), but I think that shouldn't block this PR, as this PR just enforces what we said since the cmake rewrite.

UI/window-basic-main.cpp
```diff
@@ -165,9 +165,12 @@ static void AddExtraModulePaths()
 	}
 
 	char base_module_dir[512];
```
PatTheMav, line 167: This should probably use `PATH_MAX` (and aliased from `MAX_PATH` on Windows).

gxalpha: I'd consider that to be out-of-scope of this PR, as this is something that would also effect other operating systems (and isn't even modified by this PR).

UI/window-basic-main.cpp
```diff
@@ -178,26 +181,25 @@ static void AddExtraModulePaths()
...
-		"obs-studio/plugins/%module%.plugin/Contents/Resources");
-	obs_add_module_path(config_bin, config_data);
+	obs_add_module_path((path + "/Contents/MacOS").c_str(),
+			    (path + "/Contents/Resources").c_str());
 
+#ifndef __aarch64__
 	/* Legacy System Library Search Path */
-	obs_add_module_path((path + "/bin").c_str(), (path + "/data").c_str());
+	BPtr<char> system_bin_legacy =
+		os_get_program_data_path_ptr("obs-studio/plugins/%module%/bin");
```
PatTheMav, line 191: I wonder if it'd be more consistent to use a `char` array and `GetProgramDataPath/`GetConfigPath` like above instead of `BPtr`s.

gxalpha: Done

aramg:
This seems too abrupt and too heavy of a change. With the "legacy" .so approach it was possible to ship universal arm64/x64 builds instead of 2 separate installers. I don't see why you wouldn't load arm64 .so plugins while the legacy x64 code is still in there anyway. Eventually that's going to get moved (I assume), and all legacy .so plugin support will come to end (making universal builds unnecessary) -- this seems it would be much a smoother transition. 

Just my 2 cents; sorry I didn't see/comment here before pr was merged.

PatTheMav:
The goal is to have all updated plugins for v28 (including Apple Silicon and - if necessary - Qt6 support) use the new plugin bundle by default and not allow a situation where plugins are in an in-between limbo state of supporting some things, but not others.

We will load existing legacy plugins in the Linux library format on Intel, but Apple Silicon is a clean slate situation. As such we'd rather rip off the band-aid once, allowing us to remove legacy code paths earlier rather than later.

gxalpha:
I also think there might be some confusion here.
Universal builds are still possible, in fact they are very much preferred over builds for individual architectures (as you only need to provide one installer). They also need to be bundles in the user application support directory, both x86_64 and arm64 OBS loads plugin bundles from there.

The only thing that does get prevented by this PR is a unified installer location for both OBS 27 on x86_64 *and* OBS 28 on arm64 (as OBS 27 doesn't know how to load bundles). However, given that plugins usually don't support many versions of OBS into the past, this is an issue that will resolve itself over time (and I consider it to be an okay tradeoff to prevent plugins still using the old format in years to come).

aramg:
@gxalpha - thanks, yes, was just looking into that as we speak. 

My original argument then is not as strong, simply moving the (universal) builds into a bundle and into the user appsupport directory is totally fine 🙏

PR #7106, opened 2022-08-16
cmake: Specify utf-8 for MSVC builds

notr1ch:
### Description
If there is no BOM, MSVC defaults to the user's code page which can result in errors if UTF-8 characters are present in the source files.

https://docs.microsoft.com/en-us/cpp/build/reference/utf-8-set-source-and-executable-character-sets-to-utf-8?view=msvc-170

### Motivation and Context
Trying to build OBS with a non-standard codepage would fail.

### How Has This Been Tested?
Hasn't.

### Types of changes

### Checklist:


RytoEX:
I think with https://github.com/obsproject/obs-amd-encoder/pull/431 merged and the submodule updated a while ago, this can be merged now.

PR #7114, opened 2022-08-17
UI: Don't set theme if it didn't change

gxalpha:
### Description

This PR makes sure that the theme doesn't get set again if it's already set, since that'd be a waste and can cause small hiccups sometimes.

### Motivation and Context

Noticed originally that theme switching is very bad on macOS. In the process, I noticed that the theme gets applied each time something in the combobox gets selected, even if it already was selected before.

### How Has This Been Tested?

macOS 13
Selecting the already selected theme no longer calls SetTheme

### Types of changes

### Checklist:


WizardCM:
This was intentionally supported, as it allowed for theme developers to reload the current theme without switching away and back, which would be even slower.

PR #7116, opened 2022-08-17
UI: Only set QStyle on app start

gxalpha:
### Description

Setting a QStyle resets the color palette. This is unhelpful when a few
lines before that we set a new palette. In theory the setStyle call
could just be moved up, but we don't need to since we always set the
same style so there's no reason to call it multiple times.

Massive thanks to @Fenrirthviti for digging into the documentation.

### Motivation and Context

Colors broke when theme switching (especially on macOS and Linux)

### How Has This Been Tested?

macOS 13
Switching themes now no longer appears to break colors.

Tested that the styleHint in OBSIgnoreWheelProxyStyle still gets respected after
switching themes (by setting it to 1 for the test(default on Mac is 0) and testing wheel
scrolling before and after, and it worked in both situations).

### Types of changes

### Checklist:

### DIFF
--- a/UI/obs-app.cpp
+++ b/UI/obs-app.cpp
@@ -1222,7 +1222,6 @@ bool OBSApp::SetTheme(std::string name, std::string path)
 
 	QString mpath = QString("file:///") + lpath.c_str();
 	ParseExtraThemeData(path.c_str());
-	setStyle(new OBSIgnoreWheelProxyStyle);
 	setStyleSheet(mpath);
 	if (themeMeta) {
 		themeDarkMode = themeMeta->dark;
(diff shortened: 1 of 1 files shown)


UI/window-basic-main.cpp
```diff
@@ -249,6 +249,7 @@ OBSBasic::OBSBasic(QWidget *parent)
 #endif
 
 	setAttribute(Qt::WA_NativeWindow);
+	setStyle(new OBSIgnoreWheelProxyStyle());
```
cg2121, line 252: Out of curiosity, why is it set here instead of in OBSApp?

gxalpha: No real reason, mostly because the stylesheet is also set there.
Clarified that a bit by moving it directly above the stylesheet call.

cg2121: Maybe a better idea would to have it in the InitTheme() function in obs-app.cpp, so theming stuff can be in one place. I don't think we even use the setStyleSheet() in window-basic-main.cpp, as it seems to be old code before custom stylesheets were introduced.

gxalpha: moved it to OBSApp::InitTheme

cg2121 (approved):
Everything looks good to me now.

WizardCM (approved):
Tested without issues on Windows and macOS.

PR #7117, opened 2022-08-17
UI: Free virtual cam memory on shutdown

chippydip:
### Description
Fix a memory leak on shutdown if the virtual camera is still active. Also fixes an untracked memory leak (C++ new/delete) of the `vCamConfig` by making it statically allocated.

### How Has This Been Tested?
- Reproduced leak on shutdown, implemented fix and tested that leak was gone.
- Added `new`/`delete` overrides to `VCamConfig` so that leak was tracked, fixed and then tested again that leak was gone before removing the overrides.

### Types of changes

### Checklist:

### DIFF
--- a/libobs/obs-view.c
+++ b/libobs/obs-view.c
@@ -162,6 +162,9 @@ static inline void set_main_mix()
 
 video_t *obs_view_add(obs_view_t *view)
 {
+	if (!view)
+		return NULL;
+
 	struct obs_core_video_mix *mix = obs_create_video_mix(&obs->video.ovi);
 	if (!mix) {
 		return NULL;
@@ -178,6 +181,9 @@ video_t *obs_view_add(obs_view_t *view)
 
 void obs_view_remove(obs_view_t *view)
 {
+	if (!view)
+		return;
+
 	pthread_mutex_lock(&obs->video.mixes_mutex);
 	size_t idx = find_mix_for_view(view);
 	if (idx != DARRAY_INVALID)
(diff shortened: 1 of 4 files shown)


libobs/obs-view.c
```diff
@@ -162,6 +162,9 @@ static inline void set_main_mix()
 
 video_t *obs_view_add(obs_view_t *view)
 {
+	if (!view)
+		return NULL;
+
```
notr1ch, line 167: Is there a reason these null checks were added? Calling add/remove with NULL is clearly wrong behavior in the first place. Either way I feel like it should be a separate commit as it's a libobs API change unrelated to virtual cam shutdown.

chippydip: I noticed when updating the `StopVideo()` method that a precautionary null check wasn't actually needed before calling either `obs_view_remove()` or `obs_view_set_source()`. If called with a null view the remove method ends up doing a little extra work (locking a mutex to search a list of video mixes for one with the matching view) but ultimately is a noop since the view shouldn't be found. I figured it would be good to add the null check there as a minor optimization and do the add method as well. The other pre-existing `obs_view_*()` methods have similar null checks, I just didn't add them when I added the new `obs_view_add/remove()` methods.

In any case, happy to revert that change or split into a separate commit since it's not technically needed.

flaeri:
I can no longer reproduce any of the memory leaks I initially reported, so consider that a thumbs up from me.

libobs/obs-view.c
notr1ch: I'm personally not a fan of "painting over" bad behavior by API clients (none of these functions should be called with NULL) but libobs already does this in many places much to my annoyance, so I don't mind either way. Probably fine to keep it with this commit, I didn't realize view add / remove were also introduced by this feature.

PR #7129, opened 2022-08-18
libobs: Initialize main_view video mix before video thread

notr1ch:
### Description
Fixes a race condition where the video thread would exit because it sees no active mixes, causing OBS to freeze on startup. Introduced in https://github.com/obsproject/obs-studio/pull/6577.

### Motivation and Context
Bug fix.

### How Has This Been Tested?
@tytan652 tested on Discord and confirmed it fixed the freezing he experienced.

### Types of changes

### Checklist:


chippydip (commented):
Ah, good catch! Perhaps at a deeper level it's wrong for `stop_requested()` on `obs-video.c` to return true if the list of mixes is empty? I'm not 100% sure either way, but this tweak to startup order should certainly fix that race condition. 👍

PR #7132, opened 2022-08-19
Update VST path

hfiguiere:
Old path will be deprecated in 22.08

### Description

In Flatpak vst plugins are moving. All current plugin in 21.08 series can use both. In 22.08, this will change.

### Motivation and Context

When you move to a 22.08 based runtime, things will break.

### How Has This Been Tested?

Hasn't been test per see, I know the changes. The only question is how does OBS remember of plugins? By their ID or by their PATH?

### Checklist:

### DIFF
--- a/CI/flatpak/com.obsproject.Studio.json
+++ b/CI/flatpak/com.obsproject.Studio.json
@@ -19,7 +19,7 @@
     "--talk-name=org.a11y.Bus",
     "--own-name=org.kde.StatusNotifierItem-2-2",
     "--system-talk-name=org.freedesktop.Avahi",
-    "--env=VST_PATH=/app/extensions/Plugins/lxvst"
+    "--env=VST_PATH=/app/extensions/Plugins/vst"
   ],
   "add-extensions": {
     "com.obsproject.Studio.Plugin": {
@@ -34,7 +34,7 @@
       "directory": "extensions/Plugins",
       "version": "21.08",
       "add-ld-path": "lib",
-      "merge-dirs": "lxvst",
+      "merge-dirs": "vst",
       "subdirectories": true,
       "no-autodownload": true
     }


CI/flatpak/com.obsproject.Studio.json
```diff
@@ -19,7 +19,7 @@
     "--talk-name=org.a11y.Bus",
     "--own-name=org.kde.StatusNotifierItem-2-2",
     "--system-talk-name=org.freedesktop.Avahi",
-    "--env=VST_PATH=/app/extensions/Plugins/lxvst"
+    "--env=VST_PATH=/app/extensions/Plugins/vst"
```
tytan652, line 22:
```suggestion
    "--env=VST_PATH=/app/extensions/Plugins/lxvst;/app/extensions/Plugins/vst"
```

CI/flatpak/com.obsproject.Studio.json
```diff
@@ -34,7 +34,7 @@
       "directory": "extensions/Plugins",
       "version": "21.08",
       "add-ld-path": "lib",
-      "merge-dirs": "lxvst",
+      "merge-dirs": "vst",
```
tytan652, line 37:
```suggestion
      "merge-dirs": "lxvst;vst",
```

tytan652 (commented):
The commit is not named correctly. Use forced push, don't create a new PR.
And please verify the checkboxes in the PR description.

About the changes, since we can have multiple path in `VST_PATH`, just adding it and not replacing would be better.

hfiguiere:
> About the changes, since we can have multiple path in `VST_PATH`, just adding it and not replacing would be better.

No it would NOT.

hfiguiere:
> You said that both were used in 21.08, we are not going to suddenly switch to 22.08 when available.

I said the were changing and both are available - for a transition. But not that they should be used simultaneously.

Now if you want to not change that know and have it break because you forget when moving to a 22.08 based runtime, it's up to you. Given that OBS flatpak manifest isn't hosted in the flathub organisation I don't have the possibility to fix this myself.

tytan652:
> Now if you want to not change that know and have it break because you forget when moving to a 22.08 based runtime, it's up to you.

I meant by that, that we have time before we switch to 22.08.

This change is needed to avoid a breakage in the future, but this PR can't be merged if you don't comply to the contributing document.

PR #7136, opened 2022-08-19
obs-ffmpeg: Direct setting of  encryption & auth for SRT & RIST

pkviet:
### Description
Also modifies rtmp-services (slight change in locale file).
This allows the direct use of:
- encryption passphrase (SRT & RIST),
- username + password (RIST),
- as well as streamid (SRT).
Previously, these parameters had to be set in the URL in the form:
`URL?option1=value1&option2=value2.`

They can now be set directly in the Settings/Stream menu.
The encryption passphrase for SRT & RIST can be set in the stream key
field of custom rtmp service.
The username field of custom rtmp service is used as the SRT streamid.
Indeed, per libsrt guidelines [1], the streamid is used to identify a
publisher; so we do likewise.
For RIST, the srp_username & srp_password are provided by the username
& password fields [2].
Additionally, some error logging has been added when there's a
disconnect caused by a wrong password.

[1] https://github.com/Haivision/srt/blob/master/docs/features/access-control.md
[2] https://code.videolan.org/rist/librist/-/wikis/Authentication-and-the-ristsrppasswd-Utility

### Motivation and Context
Allow better support for srt services.
Currently, srt can be used in services but streamid & passphrase can't be set
by the service because the info is not pulled by the output.
So I've modified the ffmpeg-mpegts-output to be able to use those settings.
This stemmed from a request on dev channel on discord from a service inquiring
about srt support.

### How Has This Been Tested?

- encryption passphrase tested against wowza server (SRT) + ffplay (SRT, RIST).
- tested that services.json work with an srt service and can use passphrase.

### Types of changes

### Checklist:


UI/data/locale/en-US.ini
```diff
@@ -887,7 +887,7 @@ Basic.Settings.General.MultiviewLayout.25Scene="Scenes only (25 Scenes)"
 Basic.Settings.Stream="Stream"
 Basic.Settings.Stream.StreamType="Stream Type"
 Basic.Settings.Stream.Custom.UseAuthentication="Use authentication"
-Basic.Settings.Stream.Custom.Username="Username"
+Basic.Settings.Stream.Custom.Username="Username /SRT Streamid"
```
notr1ch, line 890: Should be one more space after `/`, though I think I would prefer keeping this as "Username" and changing the entire string to "Streamid" if the selected protocol is SRT.

pkviet: I can remove streamid.
In terms of UI/UX, it's actually not ideal to have srt/rist in the same service as custom rtmp.
It'd be better to have a separate service.
There's the idea that more specific options could then be offered. I'm thinking of the main ones, latency, bandwidth and profile.
There's also the services/outputs/protocols overhaul which is in RFC (from tytan). So there's a timing issue which makes it unclear for me about the best path forward.
For now, i'll remove the streamid string.
The information will be relegated to the wiki.
For upcoming srt services (this is what prompted this PR), that's not ideal though. They'll have to walkthrough their users as to what is what.

pkviet: removed the streamid string

pkviet:
Bug fixed for the current srt output:

when auto-reconnect is set and a wrong
passphrase is provided for srt, the output would keep trying to
reconnect. With this commit, an OBS_OUTPUT_INVALID_STREAM signal is
emitted and the stream is immediately stopped.

plugins/obs-ffmpeg/obs-ffmpeg-mpegts.c
```diff
@@ -327,11 +327,50 @@ static inline int connect_mpegts_url(struct ffmpeg_output *stream, bool is_rist)
 					"Can not allocate memory.");
 		goto fail;
 	}
+	/* For SRT, pass streamid & passphrase; for RIST, pass passphrase, username
+	 * & password.
```
softorangetech200, line 331: You should use `*` if the `/*` is `/**`.

derrod: Please read our contribution and code-style guidelines first: https://github.com/obsproject/obs-studio/blob/master/CONTRIBUTING.rst

niklaskorz:
Wouldn't it make more sense to use the streaming key field for the SRT streamid parameter, and the password field for the encryption passphrase?

pkviet:
> Wouldn't it make more sense to use the streaming key field for the SRT streamid parameter, and the password field for the encryption passphrase?
I did consider the alternative.
1) Why I picked the current option is answered in the PR description. Per libsrt, the streamid identifies the publisher. 
2) You're also just thinking of SRT and not RIST. Both have passwords but in the case of RIST there's an associated user-id. So RIST makes more sense to use the id/password fields for the srp_user/srp_password pair.
Since SRT has a similar passphrase param in addition to stream-id, it wouldn't have made sense to use a different field.

niklaskorz:
The problem I see with using the streaming key as encryption phrase instead of stream id is that the predefined "rtmp services" can only make use of the stream id field and are not able to provide a username. Predefined services that want to make use of SRT would thus be unable to identify the stream.

pkviet:
> The problem I see with using the streaming key as encryption phrase instead of stream id is that the predefined "rtmp services" can only make use of the stream id field and are not able to provide a username. Predefined services that want to make use of SRT would thus be unable to identify the stream.

The streamkey is NOT used at all in this PR.
You misunderstand.

niklaskorz:
It is not used?

> The encryption passphrase for SRT & RIST can be set in the stream key
field of custom rtmp service.

```cpp
		context->passphrase = NULL;
		if (stream->ff_data.config.key != NULL) {
			if (strlen(stream->ff_data.config.key))
				context->passphrase =
					av_strdup(stream->ff_data.config.key);
		}
```

pkviet:
> It is not used?
> 
> > The encryption passphrase for SRT & RIST can be set in the stream key
> > field of custom rtmp service.
> 
> ```c++
> 		context->passphrase = NULL;
> 		if (stream->ff_data.config.key != NULL) {
> 			if (strlen(stream->ff_data.config.key))
> 				context->passphrase =
> 					av_strdup(stream->ff_data.config.key);
> 		}
> ```

damnit, sorry.
I've forgotten how i coded that.
So : 
- RIST has: encryption passphrase, srp_user, srp_password
- SRT has: encryption passphrase, stream_id

I get your concern. I would have to either (1) modify the services to support username/password setting; that's do-able, (2) or follow your suggestion, which is not ideal because srt and rist have the passphrase dealt differently then.
I'll think about it a bit before picking up an option.

pkviet:
**Update**
- SRT stream_id is now set as the stream key.
- SRT encryption passphrase is set in the password auth field.
This will facilitate integration by services.
Tooltips have been added to help users as to what needs to be entered where, depending on the protocol.
It would have been better to change the labels from Stream key to Stream key / SRT stream_id / RIST passphrase.
But R1ch felt the labels were turning too long.
Ideally, there should be a separate MPEGTS service for srt/rist where these fields are explicitly displayed.
This will wait for the service/output/protocol overhaul planned in later obs versions.

UI/forms/OBSBasicSettings.ui
```diff
@@ -1246,7 +1249,11 @@
               </widget>
              </item>
              <item row="9" column="1">
-              <widget class="QLineEdit" name="authUsername"/>
+              <widget class="QLineEdit" name="authUsername">
+               <property name="toolTip">
+                <string>Basic.Settings.Stream.Custom.Username.ToolTip</string>
+               </property>
+		          </widget>
```
RytoEX, line 1256: nit: Fix indenttion

RytoEX (changes requested):
Found a nit on a quick review.

plugins/rtmp-services/data/schema/service-schema-v3.json
```diff
@@ -110,7 +110,8 @@
                                 "enum": [
                                     "rtmp_output",
                                     "ffmpeg_hls_muxer",
-                                    "ftl_output"
+                                    "ftl_output",
+                                    "ffmpeg_mpegts_muxer"
```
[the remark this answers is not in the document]
derrod: Depends on how older version would handle this. In general it shouldn't be too big of an issue since the number if users on old versions or using a service other than the main big ones isn't too large. But if it turns out to be too big of a pain point we can introduce a new version (v4) of the services file.

GeorgesStavracas:
For the tooltip, it was suggested to use the question mark pattern here:

pkviet:
> For the tooltip, it was suggested to use the question mark pattern here:
> 
> 

how ?
your screen cap leverages the Properties API

UI/forms/OBSBasicSettings.ui
pkviet: done

plugins/rtmp-services/data/schema/service-schema-v3.json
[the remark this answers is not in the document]
tytan652: > Depends on how older version would handle this. 

None, handle this.

> But if it turns out to be too big of a pain point we can introduce a new version (v4) of the services file.

I have always https://github.com/obsproject/rfcs/pull/39 in mind, so I don't know if bumping is worth it, if the overhaul will happen.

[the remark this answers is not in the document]
pkviet: ok i'll turn it into v4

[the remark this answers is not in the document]
derrod: Bumping doesn't really cost us anything, and this is exactly what it exists for.

derrod:
Note: you will also need to bump the version in the services json files itself and ensure the validator script still works.

pkviet:
> Note: you will also need to bump the version in the services json files itself and ensure the validator script still works.

ah thx, didn't think of that. Will do

pkviet:
> For the tooltip, it was suggested to use the question mark pattern here:
> 
> 

@GeorgesStavracas 
i implemented your suggestion; it was less a hassle than I expected, I just had to copy a few lines.
The screenshots have been updated.

tytan652:
This one should be updated too:
https://github.com/obsproject/obs-studio/blob/master/plugins/rtmp-services/rtmp-format-ver.h

pkviet:
> This one should be updated too: https://github.com/obsproject/obs-studio/blob/master/plugins/rtmp-services/rtmp-format-ver.h

done thx

pkviet:
**Update**
Removed one extra \n in the stream key tooltip.
For the UI/UX part, got the approval of Warchamp.

notr1ch:
My only nitpick here would be the tooltip texts "type the username, "type the password" etc. No one is typing those, they are probably pasting. "Enter the username" or "Provide the username" or just "Username" etc. might be better.

pkviet:
> My only nitpick here would be the tooltip texts "type the username, "type the password" etc. No one is typing those, they are probably pasting. "Enter the username" or "Provide the username" or just "Username" etc. might be better.

done

derrod (approved):

RytoEX (approved):

PR #7148, opened 2022-08-20
obs-ffmpeg: Change AMF bitrate logging to kbps

flaeri:
### Description
Move multiplication to when its passed to the encoder, so that bitrate
is kept in kbps. Changed for both for H264 and HEVC. Other encoders (x264 and NVENC) already display kbps in the log,
so it makes sense to mimic this with AMF. It's difficult to tell the exact bitrate with bps.

```
01:10:56.878: 	rate_control: VBR
01:10:56.878: 	bitrate:      25000
```

### Motivation and Context
It's difficult to tell the exact bitrate with bps. Helps support volunteers and users who are used to seeing kbps in other encoders. I believe it's important to have units (kbps) as it reduces confusion, especially with AMF which has been displaying bps since its inception.

### How Has This Been Tested?
Tested on Win11, both texture-amf and fallback-amf, using H.264 and HEVC. Records fine, log output is as expected for all 4 variations.

### Types of changes

### Checklist:

### DIFF
--- a/plugins/obs-ffmpeg/texture-amf.cpp
+++ b/plugins/obs-ffmpeg/texture-amf.cpp
@@ -1095,13 +1095,13 @@ try {
 		return true;
 	}
 
-	int64_t bitrate = obs_data_get_int(settings, "bitrate") * 1000;
+	int64_t bitrate = obs_data_get_int(settings, "bitrate");
 	int64_t qp = obs_data_get_int(settings, "cqp");
 	const char *rc_str = obs_data_get_string(settings, "rate_control");
 	int rc = get_avc_rate_control(rc_str);
 	AMF_RESULT res;
 
-	amf_avc_update_data(enc, rc, bitrate, qp);
+	amf_avc_update_data(enc, rc, bitrate * 1000, qp);
 
 	res = enc->amf_encoder->ReInit(enc->cx, enc->cy);
 	if (res != AMF_OK)
@@ -1120,7 +1120,7 @@ static bool amf_avc_init(void *data, obs_data_t *settings)
 {
 	amf_base *enc = (amf_base *)data;
 
-	int64_t bitrate = obs_data_get_int(settings, "bitrate") * 1000;
+	int64_t bitrate = obs_data_get_int(settings, "bitrate");
 	int64_t qp = obs_data_get_int(settings, "cqp");
 	const char *preset = obs_data_get_string(settings, "preset");
 	const char *profile = obs_data_get_string(settings, "profile");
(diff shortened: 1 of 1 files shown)


RytoEX:
Checking other encoders, they seem to keep bitrate as-is until the moment that it's needed to be converted from kbps to bps, and then multiply in-place instead of converting back and forth. I would also be fine with that, though this could also be merged as-is and that could be changed later, if desired.

flaeri:
Other encoders like nvenc and x264 use kbps already, so there is no need to convert. AMF operates it all on bps, hence the need to do anything.

RytoEX:
> Other encoders like nvenc and x264 use kbps already, so there is no need to convert. AMF operates it all on bps, hence the need to do anything.

I beg to differ. While x264 accepts kbps for its bitrate parameter, NVENC does not.

https://github.com/obsproject/obs-studio/blob/f5be6f5fdd2fe34b18518ecd38030f0768845688/plugins/obs-ffmpeg/jim-nvenc.c#L259-L262
https://github.com/obsproject/obs-studio/blob/f5be6f5fdd2fe34b18518ecd38030f0768845688/plugins/obs-ffmpeg/jim-nvenc.c#L363-L364
https://github.com/obsproject/obs-studio/blob/f5be6f5fdd2fe34b18518ecd38030f0768845688/plugins/obs-ffmpeg/jim-nvenc.c#L588-L590
https://github.com/obsproject/obs-studio/blob/f5be6f5fdd2fe34b18518ecd38030f0768845688/plugins/obs-ffmpeg/jim-nvenc.c#L612-L623

https://github.com/obsproject/obs-studio/blob/f5be6f5fdd2fe34b18518ecd38030f0768845688/plugins/obs-ffmpeg/obs-ffmpeg-nvenc.c#L99
https://github.com/obsproject/obs-studio/blob/f5be6f5fdd2fe34b18518ecd38030f0768845688/plugins/obs-ffmpeg/obs-ffmpeg-nvenc.c#L154

derrod:
All of the code you're quoting there is multiplying the bitrate by 1000 to get from kbps in the config to bps for the internal encoder settings.

flaeri:
I've modified it in the way RytoEX suggested, as it matches the other encoders.

Restested H264 and HEVC with CBR, VBR and CQP. Work fine, and logs are still neat and nice.

plugins/obs-ffmpeg/texture-amf.cpp
```diff
@@ -1171,7 +1171,7 @@ static bool amf_avc_init(void *data, obs_data_t *settings)
 
 	info("settings:\n"
 	     "\trate_control: %s\n"
-	     "\tbitrate:      %d\n"
+	     "\tbitrate:      %d kbps\n"
```
derrod, line 1174: Since none of the other items have units, and neither do any of the other encoders, I'd omit that here for now. Eventually we should perhaps revise the logging for this entirely and add units where appropiate.

flaeri: I was actually hoping to PR a change to all of the other encoders to add kbps.

Apple VT already has it, and I would posit that knowing the unit can only be helpful.

Do you think it should be in a separate PR (per encoder, or all together)?

flaeri: Seeing as this breaks the logAnalyzer, I'll just remove it :)

PR #7149, opened 2022-08-20
mac-capture: Clip gamut to sRGB

### LINKED ISSUE #7139: Washed out colors in output when converting from P3 color space on macOS
### Operating System Info

macOS 12

### Other OS

_No response_

### OBS Studio Version

28.0.0-beta2

_No response_

### OBS Studio Log URL

https://obsproject.com/logs/WCP4lAdgetZg-nPR

### OBS Studio Crash Log URL

_No response_

### Expected Behavior

Colors should be accurately reproduced in the output.

### Current Behavior

Colors are increasingly washed out with each iteration of conversion from P3 to the output color space.

### Steps to Reproduce

1. Have monitor or other input using a P3 color space (i.e. most Macs' default color space).
2. Add that input as source (seems to occur on macOS Screen Capture, Display Capture, Window Capture, etc.)

Screen Shot 2022-08-19 at 6 28 18 AM

Screenshot showing the issue.

Can be worked around by selecting a non-P3 color space for the monitor/input. Present in versions before 28, most likely has always been there?

Already discussed to some extent on the discord by Sebastian / gxalpha and rcdrone.

jpark37:
### Description
Mac likes P3, but OBS only supports sRGB for SDR.

Fixes #7139

### Motivation and Context
Aliasing P3 as sRGB is worse than clipping out of range colors, which shouldn't happen too often given that most SDR content fits in sRGB.

### How Has This Been Tested?
Recursive display capture colors were measured to be stable by Digital Color Meter with changes, and unstable without.

### Types of changes

### Checklist:

### DIFF
--- a/plugins/mac-capture/mac-screen-capture.m
+++ b/plugins/mac-capture/mac-screen-capture.m
@@ -474,6 +474,7 @@ static bool init_screen_stream(struct screen_capture *sc)
 	os_sem_post(sc->shareable_content_available);
 	[sc->stream_properties setQueueDepth:8];
 	[sc->stream_properties setShowsCursor:!sc->hide_cursor];
+	[sc->stream_properties setColorSpaceName:kCGColorSpaceSRGB];
 	[sc->stream_properties setPixelFormat:'BGRA'];
 #if __MAC_OS_X_VERSION_MAX_ALLOWED >= 130000
 	if (@available(macOS 13.0, *)) {


jcm93:
Thanks for this! Solves it. Quite huge for Mac users to be honest.

ValZapod:
>but OBS only supports sRGB for SDR.

But why? Just tag it as P3-D65 primaries (12 per H.273) and sRGB transfer (11 per H.273), that is what Apple uses. mpv does even support that.

>which shouldn't happen too often given that most SDR content fits in sRGB.

Tell that to all iPhone camera photos that are all display-p3 (P3-D65 +sRGB above) since [iPhone 7](https://cielab.xyz/forum/download/file.php?id=1229&mode=view) to iPhone 13 (https://www.dpreview.com/sample-galleries/4660266261/apple-iphone-13-pro-sample-gallery/4167123167) and all galaxy s21, s22 screenshots that are also display-p3 or to CSS v4 that supports display-p3 in safari. https://p3colorpicker.cool/

jpark37:
We needed an expedient least of evils solution. A metadata hack for video output would be easy to add technically, but comes with a different set of challenges.

Speaking more generally, we have a wide-gamut color space option; it's HDR. Adding wide-gamut SDR to the mix would increase the test/support matrix for limited benefit.

Both mobile ecosystems offer HDR photography don't they? Wide-gamut SDR feels outmoded in that space.

ValZapod:
> Both mobile ecosystems offer HDR photography don't they?

No one offers HDR photography yet, except Panasonic that can do HLG photos. What are you talking about? It is only planned, though Chrome supports it already in AVIF.

>Wide-gamut SDR feels outmoded in that space.

This is what BT.2020 spec defined, it uses BT.709 transfer with BT.2020 primaries, so SDR WCG.

jpark37:
Sorry, I just googled for iOS/Android HDR and found hits, but forgot HDR is overloaded to mean HDR compressed into SDR. That's why I asked the question. Fwiw, my Pixel 4 saved an sRGB JPEG photo, and I didn't see any options to increase color gamut.

Your suggestion to set P3 metadata when all of OBS SDR internals assumes 8-bit sRGB would be a hack. A proper fix would involve color space data conversions, more bits (i.e. more GPU bandwidth, worse perf), and/or render pipeline refactoring.

I think a higher priority step would be to load P3 undamaged when an HDR color space is selected, which we could do for Mac screen capture with ColorSpaceExtendedLinearSRGB potentially, or a custom GPU shader if that doesn't work. We currently have very limited HDR support for image source (just JXR, Windows only), and that probably needs to be addressed soon.

ValZapod:
>didn't see any options to increase color gamut.

Even my Nikon D5100 has the ability to do Adobe RGB photos, jpeg and (obviously fake) RAW. Here (open in Photoshop) 
https://github.com/Beep6581/RawTherapee/issues/1805#issuecomment-1249696139

>Your suggestion to set P3 metadata when all of OBS SDR internals assumes 8-bit sRGB would be a hack. 

Nope. display p3 dies use sRGB transfer, only primaries are different. ;) it is not a hack just cause apple said so, in fact Android uses the same thing now too.

>when an HDR color space is selected, 

But that is not how it works, HDR must be BT.2020 primaroes, always... The whole point is that apple uses SDR WCG. Just like Android nowadays. My gakaxy s22 Ultra, e.g. see, this is a screenshot of my home screen tagged as ICC P3-D65, cause that is what wallpaper uses.  

Screenshot_20220929-070202_One UI Home

ValZapod:
>Sorry, I just googled for iOS/Android HDR and found hits

That is only HDR cause most displays are not 100 nits. My galaxy s22 is 1750 nits... so SDR images do actually render as HDR. It is what it is.

jpark37:
> only primaries are different

Which requires a matrix multiply that is not in our render pipeline. If you avoid that by treating the whole canvas as P3, then what happens when I add an actual sRGB source next to the P3 source? It will wrongly get reinterpreted as P3.

>> when an HDR color space is selected

> But that is not how it works, HDR must be BT.2020 primaroes

My point was that P3 will fit in Rec. 2020 (except a tiny corner of red apparently), so we can fix Mac screen capture for HDR recordings at least without a lot of churn.

ValZapod:
> then what happens when I add an actual sRGB source next to the P3 source?

Oh. Yes, OBS supports all kinds of overlays, right. Of course then BT.709 primaries will have to be color managed into P3-D65.

>except a tiny corner of red apparently

Yeah, but that should be just one codevalue. It was a shock for me also that P3 is not fully inside BT.2020 :)

ValZapod:
BTW, here is mpv commit that added support for this H.273 code points. https://github.com/mpv-player/mpv/commit/f02ab2aabb630d6618dd99f63af8014c10d79562

ValZapod:
If yiu want more crazy CSS display-p3 use this. Open in Safari. https://blog.joewoods.dev/technology/colors-on-the-web-are-getting-intense-with-display-p3/

mufunyo:
> Oh. Yes, OBS supports all kinds of overlays, right. Of course then BT.709 primaries will have to be color managed into P3-D65.

This is essentially a feature request. A valid one, perhaps, but still a feature request.

Right now, OBS can record/stream either of two things:
- SDR sRGB tagged as Rec709 composited in 8 bits with minimal/no colour management to keep the render code fast
- HDR Rec2100 composited in 16 bits with some colour management to make SDR sources look normal

Supporting anything other than the two scenarios above is much much more than a "bug fix" or a "code tweak" and should be treated as a big ask. Ideally, OBS would support high bit depth SDR, wide gamut SDR, embedded ICC profiles in media files, display ICC profiles, etc, but all of that has to be weighed against the development effort required to support those things, the performance hit of doing all those colour space conversions, and the maintenance burden of having different render code paths. HDR alone was a herculean effort in that respect.

ValZapod:
>HDR alone was a herculean effort in that respect.

And of course it is much more widespread and has WCG too.

PR #7157, opened 2022-08-21
cmake: Update default compiler flags and preprocessor defines

PatTheMav:
### Description
Updates the default compiler flags and preprocessor defines used by obs-studio:

* Removes `mmx`, `sse`, and `sse2` flags for x86_64 hosts on macOS (those flags are enabled by default on 64-bit Intel-based CPUs). This removes warnings for macOS-based targets.
* Remove `USE_LIBCXX` option, as libc++ is used by default since Mac OS X Mavericks
* Set preprocessor defines explicitly via `add_compile_definitions`
* Set DEBUG defines for all platforms
* Store CCache detection result in CMake cache and also allow disabling of CCache in an existing project
* Use new `CMAKE_COLOR_DIAGNOSTICS` variable on CMake 3.24+
* Update the pluginhelper script to use the same updated compiler flags

### Motivation and Context
Clean up compiler flags, some of which had been added 9 years ago, and also fix warnings on macOS when unused/unnecessary feature flags were added.

### How Has This Been Tested?
Compiled obs-studio with new compiler flags on Windows 11, Ubuntu 22, and macOS 12.5.

### Types of changes

### Checklist:


PatTheMav:
Superseded by https://github.com/obsproject/obs-studio/pull/8247.

PR #7164, opened 2022-08-21
UI: Hide reset-timestamp property for file splitting

norihiro:
### Description

This PR hide `Reset timestamps at the beginning of each split file` from the automatic split file when the format is other than MPEG-TS (`ts`).
When setting other than MPEG-TS, the setting will be overwritten to reset the timestamp.

Also invert the checkbox by changing the text.
- Old text: `Reset timestamps at the beginning of each split file`
- Revised text: `Preserve timestamps at the beginning of each split file`

The property marked by an orange box won't show unless user select MPEG-TS.
Recording Settings

### Motivation and Context

The option was originally provided for ease of debugging and also for some containers that support timestamp not starting from zero. However, some users could turn the option off even though the current container requires the timestamp starting from zero. In such case, user experiences the recording contents do not start from the beginning of the video file for the 2nd and later files.

For example, a user reported an issue.
- https://obsproject.com/forum/threads/obs-studio-28-0-release-candidate.158248/post-580393
> if you have Reset Time-Stamp off, the recording continues onto a new file while also producing a black screen for 30m untill the 2nd actually starts.

In this case, the user set the container to MOV so that this is as expected behavior.

To avoid confusion like this, the option should be removed from the settings dialog so that the user cannot change it easily.

### How Has This Been Tested?

1. Start OBS with clean configuration.
2. Set automatic file splitting.
3. Start recording and wait until the 1st split happens.
4. Check the PTS of the 2nd file is starting from zero.

OS: Fedora 35.

### Types of changes

Even though it is a breaking change, I believe the feature is not required for the majority of the users.

### Checklist:

### DIFF
--- a/UI/data/locale/en-US.ini
+++ b/UI/data/locale/en-US.ini
@@ -1022,7 +1022,6 @@ Basic.Settings.Output.SplitFile.TypeSize="Split by Size"
 Basic.Settings.Output.SplitFile.TypeManual="Only split manually"
 Basic.Settings.Output.SplitFile.Time="Split Time"
 Basic.Settings.Output.SplitFile.Size="Split Size"
-Basic.Settings.Output.SplitFile.ResetTimestamps="Reset timestamps at the beginning of each split file"
 
 # Screenshot
 Screenshot="Screenshot Output"
--- a/UI/forms/OBSBasicSettings.ui
+++ b/UI/forms/OBSBasicSettings.ui
@@ -3018,16 +3018,6 @@
                                  </property>
                                 </widget>
                                </item>
-                               <item row="10" column="1">
-                                <widget class="QCheckBox" name="advOutSplitFileRstTS">
-                                 <property name="text">
-                                  <string>Basic.Settings.Output.SplitFile.ResetTimestamps</string>
-                                 </property>
-                                 <property name="checked">
-                                  <bool>true</bool>
-                                 </property>
-                                </widget>
-                               </item>
                               </layout>
                              </widget>
                             </item>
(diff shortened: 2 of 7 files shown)


gxalpha (commented):
Is there any reason to keep it at all?
I think that if we remove it from the UI (and the only way to set it would be to manually edit the config file), we should remove it altogether instead of leaving code that is essentially dead.

norihiro:
Um, I will reconsider hiding the option just for mp4, mov, mkv, flv, and m3u8.
Starting non-zero timestamp is ok for mpegts format and it might be useful for debugging just in case there is a missing frame or user want to have timestamp without resetting.

norihiro:
The PR is re-implemented to just hide the option when the format is other than MPEG-TS. The description of the PR is also revised.

gxalpha:
I think this could get confusing UX wise. If we want to keep the option for `.ts` files but reset the timestamp for everything else, I think the option should get inverted ("Don't reset timestamps at the beginning of each split file"). That way, the option being unchecked has the same effect as when it's hidden, which I think makes more sense (options being hidden is an issue in general though).

norihiro:
If inverting the option, the text will be `Preserve timestamp`. This terminology is not commonly used, where `Reset timestmap` is same as FFmpeg. Maybe need to add tooltip.

norihiro:
The checkbox and the property are inverted.

gxalpha (approved):

Warchamp7:
If this is only useful for debugging or extremely niche cases I'm not sure we should have the option at all. How necessary do you think this is to have available @norihiro?

norihiro:
At this moment, I don't have clear understandings how necessary this option is. What I know is that enabling this option is the default behavior of FFmpeg's segment format.

RytoEX:
> What I know is that enabling this option is the default behavior of FFmpeg's segment format.

Is that specifically just .ts files, or other file types?

norihiro:
The FFmpeg’s default does not depend the container format. The default is not-reset time stamp regardless of the format.

RytoEX:
We've updated this PR to remove the option to reset timestamps, as _not_ resetting timestamps only seems to be applicable in very specific use cases, and we don't want users ending up with unexpected results when splitting recordings into multiple files. We may reintroduce the ability to have continuous timestamps over multiple segments at some point in the future.

PR #7173, opened 2022-08-23
libobs: Fix Windows 10/11 Gamemode detection

derrod:
### Description

It seems that the older key still exists on Win 11 and 10 but is always
set to 0, so use the newer to start with and fall back to the old one if
necessary.

Also changes "10" to "10/11" just for correctness.

### Motivation and Context

Didn't work.

### How Has This Been Tested?

Ran OBS, worked.

### Types of changes

### Checklist:

### DIFF
--- a/libobs/obs-windows.c
+++ b/libobs/obs-windows.c
@@ -27,6 +27,7 @@
 #include <iwscapi.h>
 
 static uint32_t win_ver = 0;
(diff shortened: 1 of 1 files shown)


derrod:
Added logging for the case in newer Windows 10 and 11 builds where game mode/HAGS is on by default with no registry key set.

PR #7177, opened 2022-08-24
libobs: Avoid display clear workaround if possible

jpark37:
### Description
Allow fast clears in cases where we know it's safe.

### Motivation and Context
Performance. Refinement of #7175

### How Has This Been Tested?
- [x] Verified fast clear works on my desktop.
- [x] Verified slow clear path still taken by NVIDIA laptop without flicking.
- [x] Verified slow clear path still taken by M1 MacBook Pro with SRGB correctness.
- [x] Verified fast clear path still taken on Linux.

### Types of changes

### Checklist:

### DIFF
--- a/libobs/obs-display.c
+++ b/libobs/obs-display.c
@@ -19,11 +19,31 @@
 #include "obs.h"
 #include "obs-internal.h"
 
+#if defined(_WIN32)
+#ifndef WIN32_LEAN_AND_MEAN
+#define WIN32_LEAN_AND_MEAN 1
+#endif
+#include <Windows.h>
+#endif
+
 bool obs_display_init(struct obs_display *display,
 		      const struct gs_init_data *graphics_data)
 {
 	pthread_mutex_init_value(&display->draw_callbacks_mutex);
 	pthread_mutex_init_value(&display->draw_info_mutex);
+
+#if defined(_WIN32)
+	/* Conservative test for NVIDIA flickering on Intel display */
+	SYSTEM_POWER_STATUS status;
+	display->use_clear_workaround =
+		!GetSystemPowerStatus(&status) ||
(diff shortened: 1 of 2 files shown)


libobs/obs-display.c
```diff
@@ -19,12 +19,32 @@
...
 {
 	pthread_mutex_init_value(&display->draw_callbacks_mutex);
 	pthread_mutex_init_value(&display->draw_info_mutex);
 
+#if defined(_WIN32)
+	/* Conservative test for NVIDIA flickering on Intel display */
+	SYSTEM_POWER_STATUS status;
+	display->use_clear_workaround =
+		!GetSystemPowerStatus(&status) ||
+		(status.BatteryFlag != 128 && gs_get_adapter_count() != 1);
```
RytoEX, line 40: This would also affect Intel+AMD or AMD+AMD laptops or any multi-adapter system with a battery, wouldn't it? Are we able to check the Vendor ID of the adapter at this point? From my testing, this only occurs when OBS is running on the NVIDIA GPU/adapter. Just wondering if we can narrow the field a bit.

jpark37: I don't want to add to the public interface, nor add a d3d12 lib dependency to libobs for a temp hack. The perf hit isn't horrendous; I just don't want it to affect me. :)

jpark37: And this is still tighter than what we have now, which hits everyone.

jpark37: Actually, maybe you should try your laptop on Linux to see if their Linux driver is also affected.

RytoEX: The flicker does not seem to occur on Linux with proprietary NVIDIA drivers.

PR #7179, opened 2022-08-24
cmake: add obs_status output for ENABLE_{RTMPS,SERVICE_UPDATES} options

tgurr:
### Description

Trying to figure out a build issue locally our CI runs into: https://gitlab.exherbo.org/tgurr/media/-/jobs/48624 I discovered that the CMake `obs_status` summary always reads 

```
-- OBS:  ENABLED    obs-outputs
-- OBS:  ENABLED    rtmp-services
```
no matter if `-DENABLE_RTMPS:BOOL=FALSE` and `-DENABLE_SERVICE_UPDATES:BOOL=FALSE` was passed.

I didn't check if any other plugins might are also be affected in the same way, some do have that checks though.

### Motivation and Context

Build issue.

### How Has This Been Tested?

Local build.

### Types of changes

### Checklist:

### DIFF
--- a/plugins/obs-outputs/CMakeLists.txt
+++ b/plugins/obs-outputs/CMakeLists.txt
@@ -6,6 +6,10 @@ if(NOT DEFINED ENABLE_RTMPS)
       CACHE STRING "Enable RTMPS support with mbedTLS" FORCE)
   set_property(CACHE ENABLE_RTMPS PROPERTY STRINGS AUTO ON OFF)
 endif()
+if(NOT ENABLE_RTMPS)
+  obs_status(DISABLED "RTMPS support with mbedTLS")
+  return()
+endif()
 
 option(ENABLE_STATIC_MBEDTLS "Enable statically linking mbedTLS into binary"
        OFF)
--- a/plugins/rtmp-services/CMakeLists.txt
+++ b/plugins/rtmp-services/CMakeLists.txt
@@ -1,6 +1,10 @@
 project(rtmp-services)
 
 option(ENABLE_SERVICE_UPDATES "Checks for service updates" OFF)
+if(NOT ENABLE_SERVICE_UPDATES)
+  obs_status(DISABLED "Checks for service updates")
+  return()
+endif()
 
 set(RTMP_SERVICES_URL
     "https://obsproject.com/obs2_update/rtmp-services"


RytoEX:
RTMPS is not required for all RTMP services or for all obs-outputs, so the ENABLE_RTMPS change proposed here seems undesirable. ENABLE_SERVICE_UPDATES has nothing to do with enabling/disabling the entire rtmp-services module, it has to do with whether to build the module with the ability to check for updates to the services.json file. The module itself can function just fine without ENABLE_SERVICE_UPDATES.

tgurr:
> RTMPS is not required for all RTMP services or for all obs-outputs, so the ENABLE_RTMPS change proposed here seems undesirable.
> 
> ENABLE_SERVICE_UPDATES has nothing to do with enabling/disabling the entire rtmp-services module, it has to do with whether to build the module with the ability to check for updates to the services.json file. The module itself can function just fine without ENABLE_SERVICE_UPDATES. Furthermore, the rtmp-services updater is always enabled on non-Windows platforms.

If I build with  -DENABLE_RTMPS:STRING=OFF obs-outputs is not built at all hence my initial assumption that the options reflect the plugin as a whole. I've updated the description to read what the actual option description reads, if that would be helpful, if not feel free to close this pull request as both my assumption was not clearly correct as well as the initial benefit I saw adding these hints.

plugins/obs-outputs/CMakeLists.txt
```diff
@@ -6,6 +6,10 @@ if(NOT DEFINED ENABLE_RTMPS)
       CACHE STRING "Enable RTMPS support with mbedTLS" FORCE)
   set_property(CACHE ENABLE_RTMPS PROPERTY STRINGS AUTO ON OFF)
 endif()
+if(NOT ENABLE_RTMPS)
+  obs_status(DISABLED "RTMPS support with mbedTLS")
+  return()
```
tytan652, line 11: This `return()` shall be removed.

plugins/rtmp-services/CMakeLists.txt
```diff
@@ -1,6 +1,10 @@
 project(rtmp-services)
 
 option(ENABLE_SERVICE_UPDATES "Checks for service updates" OFF)
+if(NOT ENABLE_SERVICE_UPDATES)
+  obs_status(DISABLED "Checks for service updates")
+  return()
```
tytan652, line 6: This `return()` shall be removed.

plugins/rtmp-services/CMakeLists.txt
```diff
@@ -1,6 +1,10 @@
 project(rtmp-services)
 
 option(ENABLE_SERVICE_UPDATES "Checks for service updates" OFF)
+if(NOT ENABLE_SERVICE_UPDATES)
```
tytan652, line 4: Change the condition to apply to Windows only.

tytan652 (commented):
I didn't review the wording, but RTMPS and service update are options of their own modules.

There is no point to return if those are disabled.

plugins/obs-outputs/CMakeLists.txt
PatTheMav: It would also make more sense to add this output after line 109 where crypto is being disabled.

plugins/rtmp-services/CMakeLists.txt
PatTheMav: So far there is no status output in the affirmative, so I don't see why there needs to be an output _just_ for the negative case.

PatTheMav:
Conversations have been unresolved for two weeks - the PR will be closed if the open issues aren't addressed within 2 days.

PR #7198, opened 2022-08-27
obs-filters: Clarify that HDR Tone Mapping filter is optional

gxalpha:
### Description

Adds a short description to the "HDR Tone Mapping (Override)" filter clarifying that it is optional.
Also reinforces that it doesn't do anything on SDR sources.
image

### Motivation and Context

Users on both the forum and Discord server have asked whether they have to put this filter on every HDR source they have, indicating that it wasn't clear that the filter is optional.
It was suggested off-thread that it could also be made more clear that the filter is a no-op on SDK sources.

### How Has This Been Tested?

macOS 13
See screenshot.

### Types of changes

### Checklist:


jpark37 (changes requested):
This is how I would phrase it. I feel like mentioning Settings -> Advanced will only lead to more questions.
> OBS can perform HDR to SDR tone mapping automatically. Only use this filter if you need additional control for a particular source.
This filter has no effect on SDR sources.

gxalpha:
Sounds good to me

jpark37 (approved):

PR #7203, opened 2022-08-28
UI: Fix Grid Mode operations

### LINKED ISSUE #5005: Grid Scene Layout not reordering scenes
### Operating System Info

Windows 10

### Other OS

_No response_

### OBS Studio Version

27.0.1

_No response_

### OBS Studio Log URL

https://obsproject.com/logs/UythwM_4H2eGauAG

### OBS Studio Crash Log URL

_No response_

### Expected Behavior

I expected the ability to move scenes around in grid view

### Current Behavior

When I move scenes, it doesn't stay in place. Instead it goes back to the orginal spot it was in. 

### Steps to Reproduce

1. Make sure scenes is in grid view
2. Drag any scene to any spot order
3. Then try dragging it back to its spot
...

List mode scene reordering works fine. Grid mode does not.

RytoEX:
### Description

This PR is a collection of three commits to fix three distinct issues with the Scene Grid Mode:
1. Dropped items do not appear on the grid location where they were dropped
2. Scrollbar disappears when resizing or dragging while scrolled to the bottom and reappears when resizing or dragging after scrolling back up
3. While scrolled down, grid items repositioning while dragging do not reflect the grabbed object correctly

Fixes #5005.

##### Before
https://user-images.githubusercontent.com/624931/187057609-cd0e071d-505d-4adf-9a94-6d82fbcccb3b.mp4
##### After
https://user-images.githubusercontent.com/624931/187057688-73ff8af5-672b-4163-b38e-ef0df938e93b.mp4

##### Before
https://user-images.githubusercontent.com/624931/187057751-f0a7acbe-525a-4a01-915a-0171e1e7ea4e.mp4
##### After
https://user-images.githubusercontent.com/624931/187057766-6d9679b3-d90f-4c35-b162-b5af8810eb34.mp4

##### Before
https://user-images.githubusercontent.com/624931/187057822-4db5d333-1bc3-4da6-b28d-fcd6311c9a8a.mp4
##### After
https://user-images.githubusercontent.com/624931/187057838-44bbbf9c-2164-4c60-aae0-c515da1ab965.mp4

I had originally written a lot of comments while tracking this down. However, as multiple areas of the code required the same changes, I did not find a suitable place to put them. They were also long, so I will instead include them here in a collapsible section.
<details>
<summary>Notes about Qt coordinates</summary>
height() is height of the SceneTree widget excluding any
window frame.

QRect QListWidget::visualItemRect(const QListWidgetItem *item)
Returns the rectangle on THE VIEWPORT occupied by the item at
item.
The viewport's origin changes as you scroll.

int QRect::bottom()
Returns the y-coordinate of the rectangle's bottom edge.
Note that for historical reasons this function returns
top() + height() - 1;
use y() + height() to retrieve the true y-coordinate.

visualItemRect returns the rectangle relative to current
scroll position of the currently visible view of the
QListWidget.
The further "out of view" the last QListWidgetItem is, the
larger the value of h.
The closer the last QListWidgetItem is to the top of the view
of the QListWidget, the smaller the value of h.

h is intended to be the y coordinate of the bottom of the
last item in the QListWidget. This is used to check if the
bottom of the last widget is below the height of the
enclosing frame/widget. However, if scrolled to the bottom,
this incorrectly asserts that the bottom of the last widget
is above the height of the enclosing frame/widget, so
scrollbars are disabled. This can be resolved by using
y() + height() instead of bottom().

If h = visualItemRect(item(count())).bottom(), when scrolled
all the way to the bottom, then h = height() - 1, so
h < height() is always true.
Thus, y() + height(), which should be 1 pixel larger than
bottom(), should resolve this.

The x() and y() values of point are relative to viewport's
scrolled origin, not the widget's true origin.
Since we do not allow horizontal scrolling, the value of x()
is okay.
However, the value of y() needs to be adjusted by an offset
of the top()/y() value for the first widget in the SceneTree.
When not scrolled, this offset will be 0.
When scrolled down, this offset will be a negative value.
</details>

### Motivation and Context

Want less bugs in Grid Mode.

### How Has This Been Tested?

I've tested this locally on Windows 11. Would appreciate tests on macOS and Linux.

### Types of changes

### Checklist:

### DIFF
--- a/UI/scene-tree.cpp
+++ b/UI/scene-tree.cpp
@@ -143,6 +143,11 @@ void SceneTree::dropEvent(QDropEvent *event)
 
 	QListWidget::dropEvent(event);
 
+	// We must call resizeEvent to correctly place all grid items.
+	// We also do this in rowsInserted.
+	QResizeEvent resEvent(size(), size());
+	SceneTree::resizeEvent(&resEvent);
+
 	QTimer::singleShot(100, [this]() { emit scenesReordered(); });
 }
 
--- a/UI/scene-tree.cpp
+++ b/UI/scene-tree.cpp
@@ -70,7 +70,8 @@ void SceneTree::resizeEvent(QResizeEvent *event)
 {
 	if (gridMode) {
 		int scrollWid = verticalScrollBar()->sizeHint().width();
-		int h = visualItemRect(item(count() - 1)).bottom();
+		const QRect lastItem = visualItemRect(item(count() - 1));
+		const int h = lastItem.y() + lastItem.height();
 
 		if (h < height()) {
 			setVerticalScrollBarPolicy(Qt::ScrollBarAlwaysOff);
@@ -113,7 +114,8 @@ void SceneTree::dropEvent(QDropEvent *event)
 
 	if (gridMode) {
 		int scrollWid = verticalScrollBar()->sizeHint().width();
-		int h = visualItemRect(item(count() - 1)).bottom();
+		const QRect lastItem = visualItemRect(item(count() - 1));
+		const int h = lastItem.y() + lastItem.height();
 
 		if (h < height()) {
 			setVerticalScrollBarPolicy(Qt::ScrollBarAlwaysOff);
@@ -154,7 +156,8 @@ void SceneTree::dropEvent(QDropEvent *event)
 void SceneTree::RepositionGrid(QDragMoveEvent *event)
 {
 	int scrollWid = verticalScrollBar()->sizeHint().width();
-	int h = visualItemRect(item(count() - 1)).bottom();
+	const QRect lastItem = visualItemRect(item(count() - 1));
+	const int h = lastItem.y() + lastItem.height();
 
 	if (h < height()) {
 		setVerticalScrollBarPolicy(Qt::ScrollBarAlwaysOff);
--- a/UI/scene-tree.cpp
+++ b/UI/scene-tree.cpp
@@ -114,8 +114,10 @@ void SceneTree::dropEvent(QDropEvent *event)
 
 	if (gridMode) {
 		int scrollWid = verticalScrollBar()->sizeHint().width();
+		const QRect firstItem = visualItemRect(item(0));
 		const QRect lastItem = visualItemRect(item(count() - 1));
 		const int h = lastItem.y() + lastItem.height();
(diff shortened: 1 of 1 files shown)


WizardCM:
In Qt5 (which I believe we're still supporting on Ubuntu?) this does not always behave as expected if placing a scene further down the grid than it was. Its ocurrence is random, but it very clearly places the scene **last** rather than the intended position. 

In the video demo below, I try to move **Scene 8** and **Display** to second-last and third-last respectively, but both end up last.

https://user-images.githubusercontent.com/941350/187212547-2895334a-280c-4327-9f7f-e134d2bdef7b.mp4


It does seem better than before overall. Significantly more predictable, even if not 100%.

RytoEX:
> In Qt5 (which I believe we're still supporting on Ubuntu?) this does not always behave as expected if placing a scene further down the grid than it was. Its ocurrence is random, but it very clearly places the scene **last** rather than the intended position.
> 
> In the video demo below, I try to move **Scene 8** and **Display** to second-last and third-last respectively, but both end up last.
> 2022-08-29_23-29-14.mp4
> 
> It does seem better than before overall. Significantly more predictable, even if not 100%.

What is the current behavior before this PR? Is the new behavior better, worse, or not comparable to the current behavior? Is the problem described blocking for landing?

GeorgesStavracas (approved):

GeorgesStavracas:
Thanks! I've done an in-depth review of the math used here, and it's correct.

WizardCM:
> > In Qt5 (which I believe we're still supporting on Ubuntu?) this does not always behave as expected if placing a scene further down the grid than it was. Its ocurrence is random, but it very clearly places the scene **last** rather than the intended position.
> > In the video demo below, I try to move **Scene 8** and **Display** to second-last and third-last respectively, but both end up last.
> > 2022-08-29_23-29-14.mp4
> > It does seem better than before overall. Significantly more predictable, even if not 100%.
> 
> What is the current behavior before this PR? Is the new behavior better, worse, or not comparable to the current behavior (sounds like better overall)? Is the problem described blocking for landing?
> 
> From off-thread discussion, it sounds like we're inclined to take this as-is as long as we document and track the described issue and try to address it later.

Definitely better, I'm 100% ok with this merged as-is and making further adjustments later.

PR #7245, opened 2022-09-01
UI: Fix sizing of virtual camera button

cg2121:
### Description
This makes the virtual camera button the same
size as the stream, record and replay buttons.
This also makes the secondary buttons line up
with each other.

Before:
Screenshot from 2022-09-01 07-29-37

After:
Screenshot from 2022-09-01 07-34-05

### Motivation and Context
Fix UI bugs.

### How Has This Been Tested?
Looked at buttons and made sure they looked good.

### Types of changes

### Checklist:

### DIFF
--- a/UI/data/themes/Acri.qss
+++ b/UI/data/themes/Acri.qss
@@ -754,6 +754,7 @@ QSpinBox::down-arrow, QDoubleSpinBox::down-arrow {
     margin: 1px;
 }
 
+QPushButton[themeID="vcamButton"],
 #streamButton,
 #recordButton,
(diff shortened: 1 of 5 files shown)


TianQiBuTian:
Maybe it would be better to add it in the middle of the "Replay Buffer" and the "Studio Mode" according to the order of the buttons.

GeorgesStavracas:
Quickly discussed this, and it was suggested to only change the horizontal padding of these pair of buttons specifically (not all of them). Is it too hard to implement, or unreasonable to you?

cg2121:
The button is intended to be the smaller size.

PR #7261, opened 2022-09-02
UI/updater: Only run updater on Windows 10+

RytoEX:
### Description

OBS Studio 28 no longer supports Windows 7, 8, or 8.1, and Qt 6 will fail to run on those versions. Users on those systems being offered the update will end up in a state where OBS will not run if they install the update. Let's prevent the updater executable from running on those versions to avoid that scenario.

### Motivation and Context

Want to prevent broken installations.

### How Has This Been Tested?

Compiled locally, though I don't have a test Win7/8 installation to test the updater.

### Types of changes

### Checklist:

### DIFF
--- a/UI/win-update/updater/updater.cpp
+++ b/UI/win-update/updater/updater.cpp
@@ -1873,6 +1873,18 @@ int WINAPI wWinMain(HINSTANCE hInstance, HINSTANCE, LPWSTR lpCmdLine, int)
 
 	is32bit = wcsstr(cwd, L"bin\\32bit") != nullptr;
 
+	if (!IsWindows10OrGreater()) {
+		MessageBox(
+			nullptr,
+			L"OBS Studio 28 and newer no longer support Windows 7,"
+			L" Windows 8, or Windows 8.1. You can disable the"
+			L" following setting to opt out of future updates:"
+			L" Settings → General → General → Automatically check"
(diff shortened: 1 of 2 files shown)


UI/win-update/updater/updater.cpp
```diff
@@ -1873,6 +1873,18 @@ int WINAPI wWinMain(HINSTANCE hInstance, HINSTANCE, LPWSTR lpCmdLine, int)
 
 	is32bit = wcsstr(cwd, L"bin\\32bit") != nullptr;
 
+	if (!IsWindows10OrGreater()) {
+		MessageBox(
+			nullptr,
+			L"OBS Studio 28 and newer no longer support Windows 7,"
+			L" Windows 8, or Windows 8.1. You can disable the"
+			L" following setting to opt out of future updates:"
+			L" Settings -> General -> General -> Automatically"
```
notr1ch, line 1882: I would recommend using Unicode arrows here to match how OBS itself presents these guides.

RytoEX: Could you recommend a specific character?

notr1ch: We use → in the existing texts.

RytoEX: Pushed changes.

PR #7273, opened 2022-09-02
Revert "UI: Remove "Resize output (source size)" menu"

derrod:
### Description

This reverts commit 829e906ac2b15d822173009cde5c90b5ec6c04c1.

### Motivation and Context

Honestly this change annoyed me because it's a useful feature that got removed without replacement. And clearly I'm not the only one judging by some forum posts/comments about it (like https://obsproject.com/forum/threads/resize-output-missing-in-28-0-0.159123/ and on the PR/commit itself).

**Edit:**
To elaborate; one of the annoying design decisions in OBS is that canvas size is tied to profile, and not scene collection. Meaning that switching between profiles or scene collections can require you to change your canvas/output sizes. Sometimes this gets extra annoying because the scaled dropdown will default to something weird rather than what you select for the canvas. Being able to just right-click a fullscreen source and automatically switch was convenient and quick.

The probably even bigger advantage applies when switching between games that run in different resolutions for local recording with 1:1 scaling, just CTRL+R the source and then resize output to it.

### How Has This Been Tested?

Hasn't yet.

### Types of changes

### Checklist:

### DIFF
--- a/UI/data/locale/en-US.ini
+++ b/UI/data/locale/en-US.ini
@@ -1285,6 +1285,11 @@ AddUrl.Title="Add Source via URL"
 AddUrl.Text="You have dragged a URL into OBS. This will automatically add the link as a source. Continue?"
 AddUrl.Text.Url="URL: %1"
 
+# Dynamic output size
+ResizeOutputSizeOfSource="Resize output (source size)"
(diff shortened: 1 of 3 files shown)


derrod:
Per @Warchamp7 this wasn't supposed to be removed yet, revert for now until we find a better place for it.

PR #7276, opened 2022-09-03
UI: Add fragmented MP4/MOV formats

derrod:
### Description

Partial implementation for RFC [obsproject/rfcs#51](https://github.com/obsproject/rfcs/pull/51).

- Changes format dropdown to include fragmented MOV/MP4 options
- Adds tooltip to format dropdown with some explanations
- Only shows MP4/MOV warning if fragmentation is not enabled
- Adds mp4 as a valid input file extension in the remuxing window
- Does **not** change the default

Screenshot:
2023-01-29_12-52-06_YteuKE

### Motivation and Context

See https://github.com/derrod/rfcs/blob/fragmented-recording/text/0000-fragmented-recording.md#motivation

tl;dr MKV support isn't great in many places, fragmented MP4 offers the same resilience against crashes, but is more compatible with editors and video players.

### How Has This Been Tested?

Recorded a video, verified it was fragmented.

Additionally several users have verified that fMP4 works in various NLEs such as Resolve, Premiere, Kdenlive, etc.

### Types of changes

### Checklist:


SuslikV:
1) Reason for the '+empty_moov+delay_moov' flags?

2) Also, the only `frag_keyframe` will make fragment on each keyframe literally - this makes fragments to small (in size) for any CRF encoding. The file may grow in size for about 20% (depending on content) because of excessive fragmentation.

**Solution (to p2):**
The usage of the parameter `min_frag_duration=16000000` together with the `movflags` will make each fragment at least 16 sec long. They still may be short in size, but in my opinion it is way better, than fragment created on each keyframe that is usually is 2 sec only (or less, depending on the content).

derrod:
> 1. Reason for the '+empty_moov+delay_moov' flags?

´empty_moov` will cause the entire output to be fragmented, otherwise everything up until the second fragment would be a self-contained "movie".Technically it is redundant because `delay_moov` sets it as well.  
The reason `delay_moov` there is because in some situations (e.g. when using file splitting) the timestamp in the written file may not start with 0, so otherwise the muxer would error out.

> 2. Also, the only `frag_keyframe` will make fragment on each keyframe literally - this makes fragments to small (in size) for any CRF encoding. The file may grow in size for about 20% (depending on content) because of excessive fragmentation.

That's nonsense and I have no idea where you got that number from. Here's a comparison (CRF 5, default keyint of 250):
```
5936124776 crf5_fragmented.mp4
5936164678 crf5.mp4
```
As you can see, the fragmented file is actually *smaller*. Not by much, just about 40 KiB, but certainly not 20% larger.

Writing as a fragmented file would also actually decrease memory use as the muxer does not have to store information about each packet in memory (see https://www.ffmpeg.org/ffmpeg-formats.html#mov_002c-mp4_002c-ismv).

SuslikV:
> That's nonsense and I have no idea where you got that number from...

I was wrong. Assumption was based on encoding small static images with zeroed sound and has no real use case.

GeorgesStavracas:
Marked as a Draft so that nobody accidentally merges this

PatTheMav:
Is this fine to merge in tandem with the ProRes/HEVC PR (given that @GeorgesStavracas only draft-ed it to avoid accidental merging)?

We maybe want to create additional PRs to fix remuxing (e.g. shouldn't automatically remux to MP4 with ProRes, maybe remux to MOV by default on macOS), see also the other MKV issues raised by users.

derrod:
Based on discussion regarding ProRes it might make sense to make MOV the default on macOS and use MP4 on Windows.  
As ProRes is incompatible with MP4, and AV1 is incompatible with MOV it so happens that hardware encoding support for those codecs maps perfectly onto those two operating systems.

gxalpha:
> As ProRes is incompatible with MP4, and AV1 is incompatible with MOV it so happens that hardware encoding support for those codecs maps perfectly onto those two operating systems.

This assumes that AV1 hardware encoding won't come to Mac. According to rumors this might change, perhaps even not that far in the future.

derrod:
> This assumes that AV1 hardware encoding won't come to Mac. According to rumors this might change, perhaps even not that far in the future.

Hardware decoding seems likely, encoding I doubt (But since Apple are the arbiters of MOV, perhaps if they actually do add encode support they'd be nice enough to make AV1 a valid codec for MOV as well).

derrod:
Rebased and conflicts resolved. Also changed mac default to MOV.

derrod:
Updated with some UI changes to remove the checkbox and also excluded 0785e67ecc925be96993125d82ca7d629e7e6626 to not switch defaults yet.

Changes:
- Removed checkbox
- Format dropdown is now translateable and includes extra entries for fragmented MP4/MOV
- Tooltip now applies to format dropdown instead

PatTheMav:
> Based on discussion regarding ProRes it might make sense to make MOV the default on macOS and use MP4 on Windows.
> As ProRes is incompatible with MP4, and AV1 is incompatible with MOV it so happens that hardware encoding support for those codecs maps perfectly onto those two operating systems.

Following up on this - did we make this change already? Is this still blocking this PR?

derrod:
> > Based on discussion regarding ProRes it might make sense to make MOV the default on macOS and use MP4 on Windows.
> > As ProRes is incompatible with MP4, and AV1 is incompatible with MOV it so happens that hardware encoding support for those codecs maps perfectly onto those two operating systems.
> 
> Following up on this - did we make this change already? Is this still blocking this PR?

We have not changed any defaults yet. This PR originally did change it to fragmented MOV on macOS and fMP4 everywhere else but I have opted not to change defaults for now to unblock on that discussion and save that for a later time.

PatTheMav:
Yeah I'm also fine with making this available to users without changing defaults.

UI/data/locale/en-US.ini
```diff
@@ -927,6 +927,15 @@ Basic.Settings.Stream.Recommended.MaxFPS="Maximum FPS: %1"
...
 Basic.Settings.Output.Format="Recording Format"
+Basic.Settings.Output.Format.MKV="Matroska Video (.mkv)"
+Basic.Settings.Output.Format.FLV="Flash Video (.flv)"
+Basic.Settings.Output.Format.MP4="MPEG-4 (.mp4)"
+Basic.Settings.Output.Format.MOV="QuickTime (.mov)"
+Basic.Settings.Output.Format.TS="MPEG-TS (.ts)"
+Basic.Settings.Output.Format.HLS="HLS (.m3u8 + .ts)"
+Basic.Settings.Output.Format.fMP4="Fragmented MP4 (.mp4)"
+Basic.Settings.Output.Format.fMOV="Fragmented MOV (.mov)"
+Basic.Settings.Output.Format.TT="Fragmented MP4/MOV writes the recording in chunks and does not require the same finalisation as traditional MP4/MOV files.\nThis ensures the file remains playable even if writing to disk is interrupted e.g. as a result of a BSOD or power loss.\n\nThis may not be compatible with all players and editors. Use File → Remux Recordings to convert the file into a more compatible format if necessary."
```
RytoEX, line 938:
```suggestion
Basic.Settings.Output.Format.TT="Fragmented MP4/MOV writes the recording in chunks and does not require the same finalization as traditional MP4/MOV files.\nThis ensures the file remains playable even if writing to disk is interrupted e.g. as a result of a BSOD or power loss.\n\nThis may not be compatible with all players and editors. Use File → Remux Recordings to convert the file into a more compatible format if necessary."
```

RytoEX (changes requested):
Locale nit.

UI/data/locale/en-US.ini
derrod: Good old en-US/en-GB mixup. I'll fix it in a bit.

UI/data/locale/en-US.ini
```diff
@@ -927,6 +927,15 @@ Basic.Settings.Stream.Recommended.MaxFPS="Maximum FPS: %1"
...
 Basic.Settings.Output.Format="Recording Format"
+Basic.Settings.Output.Format.MKV="Matroska Video (.mkv)"
+Basic.Settings.Output.Format.FLV="Flash Video (.flv)"
+Basic.Settings.Output.Format.MP4="MPEG-4 (.mp4)"
+Basic.Settings.Output.Format.MOV="QuickTime (.mov)"
+Basic.Settings.Output.Format.TS="MPEG-TS (.ts)"
+Basic.Settings.Output.Format.HLS="HLS (.m3u8 + .ts)"
+Basic.Settings.Output.Format.fMP4="Fragmented MP4 (.mp4)"
+Basic.Settings.Output.Format.fMOV="Fragmented MOV (.mov)"
+Basic.Settings.Output.Format.TT="Fragmented MP4/MOV writes the recording in chunks and does not require the same finalization as traditional MP4/MOV files.\nThis ensures the file remains playable even if writing to disk is interrupted e.g. as a result of a BSOD or power loss.\n\nThis may not be compatible with all players and editors. Use File → Remux Recordings to convert the file into a more compatible format if necessary."
```
RytoEX, line 938: "e.g." should always be followed by a comma ("e.g.,").
Parenthesis are not required, but may be clearer:
> This ensures the file remains playable even if writing to disk is interrupted (e.g., as a result of a BSOD or power loss.)

Otherwise, a comma may be encourage:
> This ensures the file remains playable even if writing to disk is interrupted, e.g., as a result of a BSOD or power loss.

RytoEX (changes requested):
Probably last language nit.

derrod:
Are there any concerns/objections as far as merging this for the next release goes?

Currently this does not alter any defaults, and only adds the option for using fragmented MOV/MP4 as a recording format, and I would like to encourage users to test this as part of the 30 Beta to make decisions regarding the change of defaults proposed in https://github.com/obsproject/rfcs/pull/51 for a future release.

PatTheMav:
I think @RytoEX has not finished the review yet? Otherwise I was fine with it.

RytoEX:
Let me take another look next week, but it seemed largely fine to me.

RytoEX (approved):
Seems fine at a glance and from a brief test.

PR #7279, opened 2022-09-03
UI: Reduce settings dialog default height

### LINKED ISSUE #5168: Settings Window height is more than my screen height
### Operating System Info

Windows 10

### Other OS

_No response_

### OBS Studio Version

27.0.1

_No response_

### OBS Studio Log URL

https://obsproject.com/logs/MlsqgVTdI6HoGoT9

### OBS Studio Crash Log URL

_No response_

### Expected Behavior

When I click on Settings button on the Controls panel, settings window should open with its height <= screen/display height. So that "OK", "Cancel" and & "Apply" buttons are visible on the screen/display.

### Current Behavior

When I click on Settings button on the Controls panel, settings window is opening with its height > screen/display height. Due to this, "OK", "Cancel" and & "Apply" buttons are going outside of screen/display area.

https://user-images.githubusercontent.com/43258549/130329102-f83f54b2-7ce3-44c1-90cd-fd84ab894aa3.mp4

### Steps to Reproduce

1. Open OBS
2. Click on Settings button on Controls panel
...

Please find the video attached in the Current Behavior section.

tytan652:
### Description

Reduce the size of the settings window if greater than the current display's available geometry.

### Motivation and Context

Fixes #5168

This change should avoid this to happen again on 720p:

[Example]()
OK, Cancel and Apply are not accessible without having to resize the window.
### How Has This Been Tested?

- Set manually the main window size at something 720p-like.
- Open the settings window.
- The settings window is smaller in height than the main window.

### Types of changes

### Checklist:

### DIFF
--- a/UI/window-basic-settings.cpp
+++ b/UI/window-basic-settings.cpp
@@ -4279,6 +4279,20 @@ void OBSBasicSettings::closeEvent(QCloseEvent *event)
 		event->ignore();
 }
 
+void OBSBasicSettings::showEvent(QShowEvent *event)
+{
+	QDialog::showEvent(event);
+
+	/* Reduce the height of the widget area if too tall compared to the screen
+	 * size (e.g., 720p) with potential window decoration (e.g., titlebar). */
+	const int titleBarHeight =
+		QApplication::style()->pixelMetric(QStyle::PM_TitleBarHeight);
+	const int maxHeight =
+		round(screen()->availableGeometry().height() - titleBarHeight);
+	if (size().height() >= maxHeight)
+		resize(size().width(), maxHeight);
+}
+
 void OBSBasicSettings::reject()
 {
 	if (AskIfCanCloseSettings())
--- a/UI/window-basic-settings.hpp
+++ b/UI/window-basic-settings.hpp
@@ -480,6 +480,7 @@ private slots:
 
 protected:
 	virtual void closeEvent(QCloseEvent *event) override;
+	virtual void showEvent(QShowEvent *event) override;
 	void reject() override;
 
 public:


WizardCM:
It annoys me that Qt in this case is ignoring the monitor's height and treating the default height more like a minimum height. This should be automatic - "if default window size doesn't fit, resize to be smaller, otherwise stay default". Maybe there's a way for us to force it? Maybe we're setting some default heights of internal widgets somewhere causing it to stretch to its default regardless?

tytan652:
Made the fix in code rather than in UI file.

WizardCM:
I wonder if we could use QScreen::availableGeometry to determine the safest maximum size for a window on the current display. Might provide a slightly better experience than basing it on the height of the main window. However, will need to be tested on Wayland.

Warchamp7 (approved):
On one hand, basing this off the size of the main window feels janky.

On the other hand, if someone has made the main window that small despite having more screen real estate, it's not the worst thing to respect that decision in the settings window too.

RytoEX:
> On one hand, basing this off the size of the main window feels janky.
> 
> On the other hand, if someone has made the main window that small despite having more screen real estate, it's not the worst thing to respect that decision in the settings window too.

I think it would be better to use the screen dimensions instead of the main window.

tytan652:
> On the other hand, if someone has made the main window that small despite having more screen real estate, it's not the worst thing to respect that decision in the settings window too.

If the asked height is too small for the windows, it will normally not respect the asked size change.

> I think it would be better to use the screen dimensions instead of the main window.

We can't because of the Windows 10/11 taskbar which is the main cause of this PR.

RytoEX:
> > On the other hand, if someone has made the main window that small despite having more screen real estate, it's not the worst thing to respect that decision in the settings window too.
> 
> If the asked height is too small for the windows, it will normally not respect the asked size change.
> 
> > I think it would be better to use the screen dimensions instead of the main window.
> 
> We can't because of the Windows 10/11 taskbar which is the main cause of this PR.

You can just use 85/90% of the screen dimensions to account for the taskbar's minimum size, which should be sufficient in the vast majority of cases, if not all default cases.

PatTheMav:
@RytoEX While not perfect this seems reasonable to me? Got any objections to having this merged?

UI/window-basic-settings.cpp
```diff
@@ -921,6 +921,13 @@ OBSBasicSettings::OBSBasicSettings(QWidget *parent)
 
 	UpdateAudioWarnings();
 	UpdateAdvNetworkGroup();
+
+	/* Reduce the height of the window if too tall compared to the screen
+	 * size (e.g. 720p) with potential desktop decoration (e.g. taskbar) */
+	int max_height = round(
+		QGuiApplication::primaryScreen()->size().height() * 95 / 100);
```
Warchamp7, line 928: Does primaryScreen in this case get the primary display or the display that OBS is actually running on?

RytoEX: The primary display.

Warchamp7: @tytan652 We should base this off the dimensions of the screen that the settings dialog is actually opening on

tytan652: Made the change but it needs to be tested on Windows.

RytoEX:
Off-thread review yielded that the constructor is too early for this to occur, as the widget does not yet have its screen updated. Overloading showEvent and handling this there is late enough for the widget's screen to be correct. @Warchamp7 is driving this.

RytoEX:
Rebased. Will clean up commits and implement some changes from off-thread review.

RytoEX (approved):
Tested the adjusted PR changes locally on Windows 11 with a 1280x720 display. The Settings window is resized to _exactly_ fit in the usable screen space.

PR #7298, opened 2022-09-05
UI: Modernize about dialog

youk-ai:
### Description

The basic layout of the about dialog got changed. A Tab bar replaces the previous bottom navigation panel as the main tool to navigate the about panel.

The OBS logo got moved into a smaller header above the tab bar containing only the logo itself, project name and version number. In turn general project information and links got consolidated into their own "About" tab and the previous "About" section got moved into the "Patrons" tab.

After:
after

Before:
before

### Motivation and Context

The custom widget at the bottom of the about panel already imitated the functionality of a tab bar, replacing it with a QTabWidget makes the function more recognizable and adjusts well to pre-existing theming.

While the logo is important it simply took up too much space, the intention for shrinking and moving it was to shift the main focus on to the actual tab content.

I do recognize that changing that the UI is more of an art than science. I'd appreciate other thoughts on if this is even any better than the previous dialog, or how it could be improved by e.g: changes to layout, dimension, tab order, tab content...

### How Has This Been Tested?

Tested on Windows 11 22H1

- Resize the dialog to make sure elements resize and re-position accordingly
- Click through each tab to see if the content is being shown
- Check if all links lead to their respective destination

### Types of changes

### Checklist:

### DIFF
--- a/UI/data/locale/en-US.ini
+++ b/UI/data/locale/en-US.ini
@@ -1277,8 +1277,10 @@ About.Info="OBS Studio is a free and open source video recording and live stream
 About.Donate="Make a Contribution"
 About.GetInvolved="Get Involved"
 About.Authors="Authors"
+About.Patrons="Patrons"
 About.License="License"
 About.Contribute="Support the OBS Project"
+About.Copyright="© 2012 - 2022 OBS Project"
 
 # Drag-drop URL
 AddUrl.Title="Add Source via URL"
--- a/UI/forms/OBSAbout.ui
+++ b/UI/forms/OBSAbout.ui
@@ -6,36 +6,39 @@
    <rect>
     <x>0</x>
     <y>0</y>
-    <width>840</width>
-    <height>519</height>
+    <width>650</width>
+    <height>400</height>
    </rect>
   </property>
   <property name="windowTitle">
    <string>About</string>
   </property>
   <layout class="QVBoxLayout" name="verticalLayout_2">
+   <property name="spacing">
+    <number>6</number>
+   </property>
    <property name="leftMargin">
-    <number>0</number>
+    <number>9</number>
    </property>
    <property name="topMargin">
-    <number>0</number>
+    <number>9</number>
    </property>
    <property name="rightMargin">
-    <number>0</number>
+    <number>9</number>
    </property>
    <property name="bottomMargin">
-    <number>0</number>
+    <number>9</number>
    </property>
    <item>
-    <layout class="QHBoxLayout" name="horizontalLayout_2">
+    <layout class="QHBoxLayout" name="horizontalLayout">
      <property name="spacing">
-      <number>30</number>
+      <number>20</number>
      </property>
      <property name="leftMargin">
-      <number>30</number>
+      <number>6</number>
      </property>
      <property name="topMargin">
-      <number>10</number>
+      <number>6</number>
      </property>
      <property name="rightMargin">
       <number>6</number>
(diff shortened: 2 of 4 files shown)


UI/data/locale/en-US.ini
```diff
@@ -1277,8 +1277,10 @@ About.Info="OBS Studio is a free and open source video recording and live stream
 About.Donate="Make a Contribution"
 About.GetInvolved="Get Involved"
 About.Authors="Authors"
+About.Patrons="Patrons"
 About.License="License"
 About.Contribute="Support the OBS Project"
+About.Copyright="© 2012 - 2022 OBS Project"
```
WizardCM, line 1283: Might be worth using a %1 variable and using Qt to dynamically insert the current year, rather than depending on translators.

UI/forms/OBSAbout.ui
```diff
@@ -6,36 +6,39 @@
    <rect>
     <x>0</x>
     <y>0</y>
-    <width>840</width>
-    <height>519</height>
+    <width>650</width>
+    <height>400</height>
```
WizardCM, line 10: I would probably vote to make it taller, to make the content easier to read. At least 500.

WizardCM (changes requested):
I believe displaying Patrons by default is intentional, and should be kept.

I've personally been experimenting with ideas myself, and think the following would be worth discussing.

1. Update the dialog to use the new logo - likely a copy of https://github.com/obsproject/obs-studio/blob/master/UI/xdg-data/icons/obs-logo-128.png
2. Add an icon before the version number for the platform (Windows/macOS/Linux)
3. The OBS logo should be vertically centered to match the text next to it

Thank you for submitting this PR, I'm excited to see this evolve.

UI/data/locale/en-US.ini
youk-ai: I actually did experiment with this and wasn't sure if having the date be dependent on the user environment was desirable. Though in hindsight my concerns were probably misplaced, and **QDate::currentDate().year()** should make this trivially easy to implement.

youk-ai:
> I believe displaying Patrons by default is intentional, and should be kept.
> 
> I've personally been experimenting with ideas myself, and think the following would be worth discussing.
> 
>     1. Update the dialog to use the new logo - likely a copy of https://github.com/obsproject/obs-studio/blob/master/UI/xdg-data/icons/obs-logo-128.png
> 
>     2. Add an icon before the version number for the platform (Windows/macOS/Linux)
> 
>     3. The OBS logo should be vertically centered to match the text next to it
> 
> 
> Thank you for submitting this PR, I'm excited to see this evolve.

In regards to your different points:

1.  You're right!
2.  That's very cute and I love the idea! Any recommendations for a suitable source for platform icons?
3.  This one still eludes me, initially the logo actually is vertically centered with the text but once theming applies and text size is changed the alignment gets thrown out the window. One solution would be to hardcore text size but that wouldn't really be ideal.
4.  About the patrons tab being kept as default, this I didn't think of but it makes sense. It also occurs to me that that's probably the reason why the information I put into the "About" tab was previously displayed at all times so as not to stow away information about the patrons. Ideally the patron tab should be displayed first but the information in the about tab should also be displayed. I'm gonna experiment with different variations one could lay this out.

CodeYan01:
I like the modern design. If the patreons should be shown right away, how about just having it as a separate frame on the right? so the only tabs will be About, Authors, and License? You could maybe even have autoscroll, if that's possible? Just throwing suggestions.

RytoEX:
Please hold off on further changes until @Warchamp7 can provide feedback.

GeorgesStavracas:
Two suggested changes:

 * Make sure that the top vertical divider is in the middle of the dialog
 * Please make the left padding of the content match the right padding

Warchamp7:

If the above feedback from georges is also addressed, this is likely to be merged.

PR #7342, opened 2022-09-11
libobs/util: Reject plugins linking Qt5 library for Linux

kkartaltepe:
### Description

Since obs-studio switched to Qt6, old plugins linking Qt5 should be rejected. On Linux system, OBS and it's plugins link the system libraries so that obs could be crashed by such plugins that links Qt5 when Qt5 functions are called.

Co-authored-by: Norihiro Kamae <norihiro@nagater.net>

### Motivation and Context

Preferred alternative to #7282

### How Has This Been Tested?

Compiled with qt6 and loaded some of the common qt5 plugins.

### Types of changes

### Checklist:


libobs/util/platform-nix.c
```diff
@@ -96,10 +99,56 @@ void os_dlclose(void *module)
...
+}
+
 void get_plugin_info(const char *path, bool *is_obs_plugin, bool *can_load)
 {
 	*is_obs_plugin = true;
+#if OBS_QT_VERSION == 6
 	*can_load = true;
+	fork_and_qt5_check(path, can_load);
+#else
+	*can_load = true;
```
tytan652, line 150: No need to have `*can_load = true;` two times.

```suggestion
	*is_obs_plugin = true;
	*can_load = true;
#if OBS_QT_VERSION == 6
	fork_and_qt5_check(path, can_load);
```

kkartaltepe:
sorry i made a bunch of changes deciding on if we should have gnusource on apple. I ended up just slapping it in ifdefs to avoid affecting apple with any gnusource behavior changes.

tytan652:
Maybe just guard all the additions to be not on Apple.

kkartaltepe:
Yea, just wrapped everything in ifdefs for apple now.

kkartaltepe:
Fixed up the ifdefs so they work now.

libobs/util/platform-nix.c
```diff
@@ -96,10 +102,56 @@ void os_dlclose(void *module)
...
+	if (pid < 0) {
+		return;
+	}
+	int status;
+	if (waitpid(pid, &status, 0) < 0) {
+		return;
+	}
+	if (WIFEXITED(status) && WEXITSTATUS(status) == 0) {
+		*can_load = false;
+	}
```
GeorgesStavracas, line 144: Minor nit, allow me to suggest the following style:

```c
void fork_and_check_qt5(const char *path, bool *can_load)
{
	pid_t pid = fork();


	if (pid == -1) {
		blog(LOG_WARNING, "Error with fork() call");
		*can_load = false;
		return;
	}

	if (pid == 0) {
		// This is the child process
		_exit(module_has_qt5_check(path));
	} else {
		int status;
		if (waitpid(pid, &status, 0) >= 0) {
			*can_load = WIFEXITED(status) && WEXITSTATUS(status) != 0;
		} else {
			blog(LOG_WARNING, "Error waiting for child process");
			*can_load = false;
		}
	}
}
```

libobs/util/platform-nix.c
```diff
@@ -96,10 +102,56 @@ void os_dlclose(void *module)
...
+	struct link_map *list = NULL;
+	if (dlinfo(mod, RTLD_DI_LINKMAP, &list) == 0) {
+		for (struct link_map *ptr = list; ptr; ptr = ptr->l_next) {
+			char *slash = strrchr(ptr->l_name, '/');
+			if (!slash)
+				continue;
+			char *base = slash + 1;
+			if (strncmp(base, "libQt5", 6) == 0) {
+				return 0;
+			}
```
GeorgesStavracas, line 122: I wonder if it wouldn't be simpler to

```c
if (strstr(ptr->l_name, "libQt5") != NULL)
	return 0;
```

here.

(There might be a concern that a well crafted LD_PRELOAD with a maliciously crafted library path would trick strstr() into an out-of-bounds read and potentially arbitrary code execution, but given this is a subprocess and this is purely hypothetical, I don't think it matters in practice.)

GeorgesStavracas (approved):
Left a couple of suggestions, but generally this looks good to me and could land as is. Please consider whether or not to apply the suggestions.

libobs/util/platform-nix.c
kkartaltepe: this introduces more possible failures in loading plugins, which I am specifically trying to avoid. We should only avoid loading these plugins, which causes the incompatible prompt to be shown, if we are quite sure they are incompatible.  Otherwise let OBS behave exactly as it did before.

libobs/util/platform-nix.c
kkartaltepe: We could add a strlen to avoid reading past the end of the string. I don't really feel stongly since I assume no one is insane enough to put their libraries in a folder called `libQt5`. (if you are the person filling the bug report because your install in the libQt5 folder is broken, sorry).

libobs/util/platform-nix.c
GeorgesStavracas: Then `s/can_load = false/can_load = true/g` :slightly_smiling_face: 

I just enjoy the quality of being explicit in what it returns

kkartaltepe: can_load is already prepared, we are only setting it to false if we check and find qt5. Its not being returned so to speak. We could return a value if you want.

Personally i find this branching construction more confusing. Previously we have straightline code for the parent with branches for the child and failures, instead this keeps a growing chain of else's.

kkartaltepe: Updated with return, should appear simpler now.

GeorgesStavracas (approved):

GeorgesStavracas:
Just fix CI and we can land this

kkartaltepe:
Thats what I get for not compiling... Semicolon added.

PR #7347, opened 2022-09-11
UI: Use valueChanged() signal for T-Bar everywhere

gxalpha:
### Description

The `sliderMoved` signal doesn't trigger when clicking on the slider, only the `valueChanged` signal. According to QTBUG-72995 this is intentional, since the documentation states that `sliderDown` also needs to be true for the sliderMoved signal to be triggered. As for why `sliderDown` isn't true when clicking, or why that would even be necessary for a signal that should trigger when the slider is moved (and it very clearly moves) remains a mystery.

Also gets rid of the wrong usage of the `on_foo_bar` slot declaration, which should only be used for signals where the sender is part of a UI file and it gets connected automatically. While it of course is possible to connect it manually, this is against convention and can cause confusion. In this case it was particularly wrong since even if tBar was part of a UI file (which it isn't), the method should have been called `on_tBar_valueChanged` instead of `on_tbar_position_valueChanged`.

Note: The diff on this might look a bit confusing. I kept the old `TBarChanged` method but connected it to the `valueChanged` signal, and then moved the code that used to be connected to the valueChanged into that same `TBarChanged` method.

### Motivation and Context

Noticed that when clicking on the tBar instead of dragging, the slider position would change but the transition wouldn't start / continue. On Windows, this is most notable when middle-clicking (since left-click only jumps a small ), a PR to change this will come in the future.

### How Has This Been Tested?

Clicking on the tBar now triggers the transition, which is at least less incorrect than moving the slider but not doing anything.

### Types of changes

### Checklist:

### DIFF
--- a/UI/window-basic-main-transitions.cpp
+++ b/UI/window-basic-main-transitions.cpp
@@ -793,9 +793,7 @@ void OBSBasic::CreateProgramOptions()
 
 	tBar->setProperty("themeID", "tBarSlider");
 
-	connect(tBar, SIGNAL(sliderMoved(int)), this, SLOT(TBarChanged(int)));
-	connect(tBar, SIGNAL(valueChanged(int)), this,
-		SLOT(on_tbar_position_valueChanged(int)));
+	connect(tBar, &QSlider::valueChanged, this, &OBSBasic::TBarChanged);
 	connect(tBar, SIGNAL(sliderReleased()), this, SLOT(TBarReleased()));
 
 	layout->addStretch(0);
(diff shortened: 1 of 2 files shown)


RytoEX (approved):
Seems fine at a glance.

RytoEX:
If possible, could you explain the before/after behavior here? Optionally, a visual aid might help.

Warchamp7 (approved):
Transition state should actually match the TBar. I approve this change from a UX perspective.

Whether or not we should actually allow left-clicking/middle-clicking the slider to jump the TBar position is an entirely different can of worms I don't want to think about right now.

Issue #7259, opened 2022-09-01
Hotkeys possiblities using onscreen keyboard (accessibility keyboard) downgraded from OBS27 to OBS28


pkej:
### Operating System Info

macOS 12

### OBS Studio Version

28.0.0

### OBS Studio Log URL

https://obsproject.com/logs/P1xh0KkQqIKjy-9J

### Expected Behavior

In OBS27 it is possible to use any combination of the four modifier keys alt, shift, meta and ctrl and selecting any key from the onscreen accessibility keyboard in macOS.

In OBS28 it is only possible to select a single modifier and any other key, as well as the combination alt+shift and any other key. 

Trying to use any other combination of modifier keys and pressing a key onscreen with the mouse will not register. 

### Current Behavior

It is not possible to assign more than two specific modifier keys to a  hotkey when using the onscreen accessibility keyboard on macOS. Only alt+shift will register any accompanying keypress from non-modifier keys. The modifier keys can be used by themselves with any other key.

### Steps to Reproduce

1. Open "System Preferences" -> "Accessibility" -> select tab "Viewer" 
2. Open "Panel Editing" from the above dialog
3. Press "Add Panel" and select "Keyboard - ISO (Large)
4. Go to the "File"-menu and select "Save"
5. Quit the "Panel Editor"
6. Return to the "Accessibility" preferences panel
7. Hook off "Enable Accessibility Keyboard
8. In the "Accessibility Keyboard" shown on screen press the custom button on the first row to the far right
9. In the "Custom" window select "Keyboard - ISO (Large)"
10. Go into OBS 28
11. Open "Preferences"
12. Go to Hotkeys
13. Choose any empty hotkey
14. Press down different modifier keys on your physical keyboard 
15. Try to combine with any key on the onscreen keyboard
16. Only shift+alt + any other non-modifier key will be registered in the hotkey field in OBS
17. Try different combinations to verify that alt+command, alt+meta, shift+meta, shift+command, command+meta, command+alt + any other onscreen keyboard key isn't registered in OBS
18. Try any three modifier combination and note that it can't work with any non-modifier keys on the onscreen keyboard
19. Also note that any of the above will work with physical keys
20. Note that a four modifier key chord (alt+shift+meta+command) will not combine neither with physical keys, nor onscreen keyboard keys.
21. All of the above is working in OBS27
...

pkej:
[...]

PR #7349, opened 2022-09-11
UI: Fix hotkey settings screen not accepting all input on macOS

PatTheMav:
### Description
Re-enables full availability of hotkeys on macOS in OBS 28.

### Motivation and Context
By default Qt will hand off key events to platform-native input methods on macOS because the OS or additional software can intercept key input and transform it (e.g. the international variations popping up when holding down the key for certain letters).

This functionality can lead to certain key combinations being ignored because they don't call the delegate methods implemented by Qt to handle the input after it's been interpreted by the Input Method.

Disabling Input Methods for hotkey input fields fixes this issue, as the native input methods are bypassed entirely and the key events are handled by the widget and its keyboard events directly.

Fixes https://github.com/obsproject/obs-studio/issues/7259.

### How Has This Been Tested?
* Tested on macOS 12.5
* Tested on Windows 11
* Tested on Ubuntu 22

For testing several different combinations of keys and modifiers were used to bind hotkeys in the settings window.

### Types of changes

### Checklist:

### DIFF
--- a/UI/hotkey-edit.cpp
+++ b/UI/hotkey-edit.cpp
@@ -59,6 +59,15 @@ void OBSHotkeyEdit::keyPressEvent(QKeyEvent *event)
 	HandleNewKey(new_key);
 }
 
+QVariant OBSHotkeyEdit::inputMethodQuery(Qt::InputMethodQuery query) const
+{
+	if (query == Qt::ImEnabled) {
+		return false;
+	} else {
+		return QLineEdit::inputMethodQuery(query);
+	}
+}
+
 #ifdef __APPLE__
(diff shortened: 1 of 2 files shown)


PatTheMav:
Per https://bugreports.qt.io/browse/QTBUG-106516?focusedCommentId=679012&page=com.atlassian.jira.plugin.system.issuetabpanels%3Acomment-tabpanel#comment-679012 this change is not OS-specific as input methods are supposed to be disabled for all input that are not "real" text inputs.

RytoEX (approved):
Discussed off-thread with @PatTheMav and this seems fine. The documentation suggests that this is the same for Qt 5 and Qt 6, so there should be no ifdef concerns here.

PR #7352, opened 2022-09-11
linux-capture: Ensure name pixmap is checked

kkartaltepe:
### Description

The void cookie will never return an error unless we use the explicitly checked version of this function.

### Motivation and Context

Avoids getting annoying x11 errors later in the EGL calls.

### How Has This Been Tested?

Tested with some unnamable pixmaps and it works as intended.

### Types of changes

### Checklist:


GeorgesStavracas (approved):
Makes sense

PR #7364, opened 2022-09-12
UI: Allow left-clicking T-Bar on all operating systems

gxalpha:
### Description

The T-Bar currently acts like the a slider in the OS would. This means that when clicking it, on some operating systems the slider won't jump to the mouse. Now it does on every OS.

### Motivation and Context

If you click on a slider like the T-Bar, you'd expect it to jump there.

### How Has This Been Tested?

macOS 13, confirmed that left-click still works. Tested that the styleHint works in general by removing one of the buttons and seeing that it no longer works.
Other than that, hasn't. Needs testing on Windows. Draft until someone has tested.

### Types of changes

### Checklist:


gxalpha:
Discussed that this should instead change the slider to never be clickable, but instead only draggable. Will implement that soon™.

gxalpha:
Updated to not jump to the mouse on any operating system.
When left-clicking, it will now jump only a small step on all operating systems, since there appears to be no simple way to make it truly drag-only (outside of maybe handling it when receiving the event which could end up jank).

PR #7347 is still required to fix the inconsistency between where the slider appears to be after left-clicking (which it looks like we can't eliminate entirely, see above) and what the transition does.

gxalpha:
This probably isn't the right solution.

PR #7369, opened 2022-09-13
UI: Fix Chinese font on Windows

Aira-Sakuranomiya:
### Description

Add Microsoft YaHei (Microsoft YaHei UI) to font-family
Before
After

### Motivation and Context

Default Chinese font on Windows Vista+ should not be the old SimSun, it should be Microsoft YaHei.

### How Has This Been Tested?

Replace the qss and verified it was fixed.

### Types of changes

### Checklist:

### DIFF
--- a/UI/data/themes/Acri.qss
+++ b/UI/data/themes/Acri.qss
@@ -82,7 +82,7 @@ QWidget {
     selection-background-color: rgb(22,31,65);
     selection-color: palette(text);
     font-size: 10pt;
-    font-family: 'Open Sans', '.AppleSystemUIFont', Helvetica, Arial, 'MS Shell Dlg', sans-serif;
+    font-family: 'Open Sans', '.AppleSystemUIFont', Helvetica, Arial, 'Microsoft YaHei UI', 'MS Shell Dlg', sans-serif;
 }
 
 QWidget:disabled {
--- a/UI/data/themes/Grey.qss
+++ b/UI/data/themes/Grey.qss
@@ -82,7 +82,7 @@ QWidget {
     selection-background-color: rgb(40,76,184);
     selection-color: palette(text);
     font-size: 10pt;
-    font-family: 'Open Sans', '.AppleSystemUIFont', Helvetica, Arial, 'MS Shell Dlg', sans-serif;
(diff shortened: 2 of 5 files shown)


PatTheMav:
Does Microsoft YaHei UI contain latin glyphs? Because the way CSS works, on a system without Arial installed (as obscure as that might be), people would get YaHei instead of the default shell dialog font, which would definitely be seen as a "bug" by those users.

Aira-Sakuranomiya:
> Does Microsoft YaHei UI contain latin glyphs? Because the way CSS works, on a system without Arial installed (as obscure as that might be), people would get YaHei instead of the default shell dialog font, which would definitely be seen as a "bug" by those users.



> Does Microsoft YaHei UI contain latin glyphs? Because the way CSS works, on a system without Arial installed (as obscure as that might be), people would get YaHei instead of the default shell dialog font, which would definitely be seen as a "bug" by those users.

yes it does
https://docs.microsoft.com/en-ca/typography/font-list/microsoft-yahei

Aira-Sakuranomiya:
> Does Microsoft YaHei UI contain latin glyphs? Because the way CSS works, on a system without Arial installed (as obscure as that might be), people would get YaHei instead of the default shell dialog font, which would definitely be seen as a "bug" by those users.

yes it does
https://docs.microsoft.com/en-ca/typography/font-list/microsoft-yahei

TianQiBuTian:
Can also add to other themes that have not been added.

Aira-Sakuranomiya:
> Can also add to other themes that have not been added.

added👌

Fenrirthviti:
Is this a locale issue that should really be upstreamed as a bug to Qt? Why is the default not being selected properly? Surely there must be a reason.

I'm not entirely opposed to this change if it is a real issue, but I feel this might have unintended consequences as a manually defined font for non-Chinese locales, and would like to understand why it's choosing a legacy font in the first place. Do we know if this currently affects all users with a Chinese locale setting, or if it might something environment specific?

Aira-Sakuranomiya:
new pull request that put Microsoft YaHei after sans-serif
#7390

PR #7371, opened 2022-09-13
Drop win-ivcam

GeorgesStavracas:
### Description

Drop the win-ivcam plugin.

### Motivation and Context

Apparently this was a prototype from Intel that never really properly launched. Remove the plugin.

### How Has This Been Tested?

 * Open OBS Studio
 * Inhale, exhale
 * Try to integrate with win-ivcam
 * Fail

### Types of changes

### Checklist:


UI/win-update/updater/updater.cpp
```diff
@@ -606,20 +606,13 @@ static inline bool FileExists(const wchar_t *path)
 
 static bool NonCorePackageInstalled(const char *name)
 {
-	if (is32bit) {
-		if (strcmp(name, "obs-browser") == 0) {
+	if (strcmp(name, "obs-browser") == 0) {
+		if (is32bit)
```
derrod, line 610: tbh we can drop all 32 bit checks in the updater now that we've discontinued 32 bit OBS.

PatTheMav (approved):

UI/win-update/updater/updater.cpp
PatTheMav: Had the same thought, though I think it's better to have that in a separate PR and commit.

derrod:
Same deal as #7372, was marked as "merge for next major release" and then was never merged, so let's do it today and reduce that PR count!

(Even if that means I'll have to rebase some of my updater PRs D:)

PR #7372, opened 2022-09-13
Drop win-mf

GeorgesStavracas:
### Description

The win-mf plugin is deprecated, and disabled by default. Let's drop it entirely.

### Motivation and Context

Less code is better code. No code is perfect code.

### How Has This Been Tested?

 * Build OBS Studio
 * Ha! It's disabled by default already, nevermind

### Types of changes

### Checklist:


PatTheMav (approved):

WizardCM:
If we ever want to add support for the Qualcomm encoder on Windows on ARM, I expect it'll either have to be via MediaFoundation or via FFmpeg.

Additionally, at some stage we want to add a MediaFoundation video capture device sourcce (as some devices/virtual drivers only work with MF and not DShow). I wonder if a MF virtual camera would be worth doing at some stage too.

Fenrirthviti:
As far as I am aware, this plugin was entirely for the MF encoders, which we probably won't really need to support at any point. Using FFmpeg for the ARM edge case seems just fine to me, personally. Probably not any overlap in code for MF capture devices, both on input and output, and since this code is so old and unmaintained, likely better to start from scratch on those anyway.

derrod:
I think if we ever add support for MF capture or Qualcomm encoders we'll rewrite this from scratch anyway. If and when that happens we can add a new module.

derrod:
If you don't mind rebasing it, I think it's time we merge this (Next action in PR triage says "Merge for next major release", so much for that 😛).

derrod:
Windows builds passed, that's good enough for me.

Issue #7240, opened 2022-09-01
Pressing enter when renaming a source does not save the name.


CodeYan01:
### Operating System Info

macOS 12

### OBS Studio Version

28.0.0

### OBS Studio Log URL

https://obsproject.com/logs/OSkpKz12GDKGAWoa

### Expected Behavior

I expected it to have similar behavior to when renaming with the Enter key on Finder. Pressing Enter will bring up the text field to rename, then pressing Enter again will save it and remove the text field.

### Current Behavior

Pressing Enter after selecting a source in OBS will bring up the text field, pressing it again doesn't save the name nor remove the text field. I would have to click somewhere else or press Tab to save it.

### Steps to Reproduce

1. Select a source in the sources panel.
2. Press Enter. You should be able to type a new name.
3. Press Enter again. It won't save the new name and the text field will stay there.
4. Press Tab or click somewhere else to save it.
...

### Anything else we should know?

It works correctly in the Scenes panel, but not in the Sources panel. I have also noticed this since 27.2.4 (did not try older versions). This issue exists even on a 2011 iMac that I have that I used OCLP to update to macOS 12 (although you probably won't trust such old hardware).

PR #7373, opened 2022-09-13
UI: Fix source name edit textbox not accepting input on enter

PatTheMav:
### Description
The Enter key is connected to the `Edit` function and is handled by Qt with the highest priority in its key event handler. Alas the boolean return value is not propagated to the shortcut handling system, so the key event will always be consumed and as such the user will be stuck in editing state.

To fix this, `Edit` needs to behave like a toggle, saving the current state of the input at a repeat press of Enter while in editing state.

### Motivation and Context
Fixes https://github.com/obsproject/obs-studio/issues/7240

### How Has This Been Tested?
Tested on macOS, also tested that functionality is not impacted on Windows 11 and Ubuntu 22.

### Types of changes

### Checklist:

### DIFF
--- a/UI/source-tree.cpp
+++ b/UI/source-tree.cpp
@@ -1555,8 +1555,12 @@ bool SourceTree::Edit(int row)
 	QModelIndex index = stm->createIndex(row, 0);
 	QWidget *widget = indexWidget(index);
 	SourceTreeItem *itemWidget = reinterpret_cast<SourceTreeItem *>(widget);
-	if (itemWidget->IsEditing())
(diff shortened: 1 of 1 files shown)


cg2121:
Should do this for other list widgets such as scenes and filters, so everything is consistent.

PatTheMav:
> Should do this for other list widgets such as scenes and filters, so everything is consistent.

Are other rename-able list fields affected by this as well? The root cause of this issue is setting the ‘Enter’ key as the hotkey triggering the edit (as is common on macOS) and not removing that action binding when in the signal handler.

PR #7376, opened 2022-09-14
libobs-opengl: Disable deprecation warnings on macOS

PatTheMav:
### Description
We are aware of OpenGL having been deprecated on macOS since Mac OS X 10.14, so silence the deprecation warnings the "official" way.

### Motivation and Context
Remove unnecessary warnings from build output or Xcode project reports.

### How Has This Been Tested?
Tested with macOS 12.6.

### Types of changes

### Checklist:

### DIFF
--- a/libobs-opengl/gl-cocoa.m
+++ b/libobs-opengl/gl-cocoa.m
@@ -16,6 +16,7 @@
 ******************************************************************************/
 
 #include "gl-subsystem.h"
+#define GL_SILENCE_DEPRECATION
 #include <OpenGL/OpenGL.h>
 
 #import <Cocoa/Cocoa.h>


gxalpha (approved):
Works on my machine. Only leaves four warnings that look like we could/should actually address.

Fenrirthviti:
Just to play devil's advocate here, we know about this, but is it fair to leave in for anyone else who might be compiling and unaware? Does this warning cause any issues or annoyances that makes it worth removing?

gxalpha:
That's a good point, though I believe leaving them isn't much better.
I feel that warnings should act as a call to action, and the OpenGL deprecation warnings silenced by this PR aren't actionable. The only fix is to remove OpenGL altogether and replace it with Metal. Furthermore, they basically hid four warnings that were already present before the 10.15 deprecation wave and look like we can actually fix (just in case those things get removed before the rest of OpenGL, though to be honest that's not likely).

So no, it wouldn't cause much harm to leave them in, but I at least also see no benefit to it.

PatTheMav:
> That's a good point, though I believe leaving them isn't much better. I feel that warnings should act as a call to action, and the OpenGL deprecation warnings silenced by this PR aren't actionable. The only fix is to remove OpenGL altogether and replace it with Metal. Furthermore, they basically hid four warnings that were already present before the 10.15 deprecation wave and look like we can actually fix (just in case those things get removed before the rest of OpenGL, though to be honest that's not likely).
> 
> So no, it wouldn't cause much harm to leave them in, but I at least also see no benefit to it.

Pretty much this: The warnings that we should have switched to Metal 4 years ago are known and the amount of warnings in that single file has buried the _actual_ warnings we might have wanted to pay attention to.

I also don't expect many people to interact with our renderer code and as far as OpenGL is concerned there's a single source file for the macOS-specific implementation, which is simple enough to spot/review.

Fenrirthviti:
Thanks for the explanation, I'm satisfied with the conversation here living as documentation in case anyone complains, which is admittedly unlikely.

LGTM otherwise.

PR #3083, opened 2020-06-20
UI: Enable sps/pps (video headers) repetition (for srt/mpegts)

pkviet:
### Description
Also changes obs-x264, obs-ffmpeg, rtmp-services.
Repetition of sps/pps and AUDs are mandatory for mpegts live streams per spec.
This PR enables that for srt streams with mpegts container. In addition it enables AUDs (x264/nvenc) and annex B option (x264).

For obs-x264, the repeat-headers option had previously been hard-coded to false; we set it to false on default to keep previous behaviour. But we now allow it to be settable in x264 options.
For obs-ffmpeg/jim-nvenc, we add a setting for video headers repetition.
Since srt is dealt with rtmp-custom service, when x264 or jim-nvenc is used with mpegts, we automatically enable the repeat-headers option.

**This fixes some connection issues for srt streams.** (i should have opened an issue to document the bug but I confess I've been a bit lazy)
For instance, the following workflow is currently broken: obs ==> srt-live-transmit ==> ffplay , unless ffplay catches the initial video headers. If obs is started earlier than ffplay, no video will be displayed since sps/pps headers are not delivered.
Another broken case is : obs ==> srt server (wowza, nimble, srt-live-server) ==> client ; in case of a network disconnection, reconnection fails to produce a playable stream because the video headers are missing.

### Motivation and Context
This supersedes https://github.com/obsproject/obs-studio/pull/2725 
For live broadcasts (with mpegts container), repetition of sps/pps is necessary because there's no reason the broadcast catches headers at start of transmission; or in case of interruptions.
Mpeg-ts spec requires such a repetition (couldn't find exact spec refs, which are behind paywalls; this was told to me on slack #video-dev).

### How Has This Been Tested?
Tested against wowza, nimble server, makito X decoder, vlc, ffmpeg/ffplay, srt live server.

### Types of changes
 - New feature (non-breaking change which adds functionality)
 - Performance enhancement (non-breaking change which improves efficiency) 
 - Bugfix

### Checklist:
- [x] My code has been run through [clang-format](https://github.com/obsproject/obs-studio/blob/master/.clang-format).
- [x] I have read the [**contributing** document](https://github.com/obsproject/obs-studio/blob/master/CONTRIBUTING.rst).
- [x] My code is not on the master branch.
- [x] The code has been tested.
- [x] All commit messages are properly formatted and commits squashed where appropriate.
- [x] I have included updates to all appropriate documentation.

PR #7379, opened 2022-09-15
obs-ffmpeg: Fix m3u8 recording in AMF

shinji3:
### Description

SPS/PPS header insert

### Motivation and Context

When recording in m3u8 format using AMD HW H.264, the second and subsequent TS files are broken.
Fixed second and subsequent TS files are broken.

### How Has This Been Tested?

Tested with start recording in OBS Studio

### Types of changes

### Checklist:

### DIFF
--- a/plugins/obs-ffmpeg/obs-ffmpeg-mux.c
+++ b/plugins/obs-ffmpeg/obs-ffmpeg-mux.c
@@ -375,18 +375,33 @@ inline static void ts_offset_update(struct ffmpeg_muxer *stream,
 	stream->found_audio[packet->track_idx] = true;
 }
 
-static bool ffmpeg_mux_start(void *data)
+static inline void update_encoder_settings(struct ffmpeg_muxer *stream,
+					   const char *path)
 {
-	struct ffmpeg_muxer *stream = data;
-	obs_data_t *settings;
-	const char *path;
+	obs_encoder_t *vencoder = obs_output_get_video_encoder(stream->output);
+	const char *ext = strrchr(path, '.');
+
+	/* if using m3u8, repeat headers */
+	if (ext && strcmp(ext, ".m3u8") == 0) {
+		obs_data_t *settings = obs_encoder_get_settings(vencoder);
+		obs_data_set_bool(settings, "repeat_headers", true);
+		obs_encoder_update(vencoder, settings);
+		obs_data_release(settings);
+	}
+}
+
+static inline bool ffmpeg_mux_start_internal(struct ffmpeg_muxer *stream,
+					     obs_data_t *settings)
+{
+	const char *path = obs_data_get_string(settings, "path");
+
+	update_encoder_settings(stream, path);
 
 	if (!obs_output_can_begin_data_capture(stream->output, 0))
 		return false;
 	if (!obs_output_initialize_encoders(stream->output, 0))
 		return false;
 
-	settings = obs_output_get_settings(stream->output);
 	if (stream->is_network) {
 		obs_service_t *service;
 		service = obs_output_get_service(stream->output);
@@ -395,7 +410,6 @@ static bool ffmpeg_mux_start(void *data)
 		path = obs_service_get_url(service);
 		stream->split_file = false;
 	} else {
-		path = obs_data_get_string(settings, "path");
 
 		stream->max_time =
 			obs_data_get_int(settings, "max_time_sec") * 1000000LL;
@@ -427,7 +441,6 @@ static bool ffmpeg_mux_start(void *data)
 	}
 
 	start_pipe(stream, path);
-	obs_data_release(settings);
 
 	if (!stream->pipe) {
 		obs_output_set_last_error(
@@ -444,6 +457,17 @@ static bool ffmpeg_mux_start(void *data)
 
 	info("Writing file '%s'...", stream->path.array);
 	return true;
+}
+
+static bool ffmpeg_mux_start(void *data)
+{
+	struct ffmpeg_muxer *stream = data;
+
+	obs_data_t *settings = obs_output_get_settings(stream->output);
+	bool success = ffmpeg_mux_start_internal(stream, settings);
+	obs_data_release(settings);
+
+	return success;
 }
 
 int deactivate(struct ffmpeg_muxer *stream, int code)
--- a/plugins/obs-ffmpeg/texture-amf.cpp
+++ b/plugins/obs-ffmpeg/texture-amf.cpp
@@ -1158,6 +1158,11 @@ static bool amf_avc_init(void *data, obs_data_t *settings)
 				    : 250;
 
 	set_avc_property(enc, IDR_PERIOD, gop_size);
+
+	bool repeat_headers = obs_data_get_bool(settings, "repeat_headers");
+	if (repeat_headers)
+		set_avc_property(enc, HEADER_INSERTION_SPACING, gop_size);
+
 	set_avc_property(enc, DE_BLOCKING_FILTER, true);
 
 	const char *ffmpeg_opts = obs_data_get_string(settings, "ffmpeg_opts");


WizardCM:
Could you define "broken" for the subsequent files?

shinji3:
Can't play TS files in VLC player

I uploaded a bug test video
https://github.com/shinji3/BugTestVideoAMF

plugins/obs-ffmpeg/texture-amf.cpp
```diff
@@ -1158,6 +1158,7 @@ static bool amf_avc_init(void *data, obs_data_t *settings)
 				    : 250;
 
 	set_avc_property(enc, IDR_PERIOD, gop_size);
+	set_avc_property(enc, HEADER_INSERTION_SPACING, gop_size);
```
derrod, line 1161: This should probably be only conditionally enabled like the implementation for NVENC/x264 is, see PR #3083 for reference.

derrod (dismissed):

plugins/obs-ffmpeg/texture-amf.cpp
```diff
@@ -1008,6 +1009,9 @@ static obs_properties_t *amf_properties_internal(bool hevc)
 	obs_property_set_long_description(p,
 					  obs_module_text("AMFOpts.ToolTip"));
 
+	p = obs_properties_add_bool(props, "repeat_headers", "repeat_headers");
+	obs_property_set_visible(p, false);
+
```
tytan652, line 1014: The property is not required, `obs_data` is enough. The PR should work without it.

pkviet:
your code breaks recording into m3u8 for nvenc and x264: the first segment can't be played.
For file recording, header repetition for these encoders is not required. It breaks hls because it records in annex B bitstream, which the "repeat_headers" setting adds for nvenc and x264.
So two solutions:
- either use another name for the setting than 'repeat_headers', like 'amf_insert_headers'
- add a check against the id of the encoder ("h264_texture_amf")
For extra safety, you could do both.
I haven't been able to confirm the issue yet on an amd gpu. Will report later.

pkviet:
Also, you didn't properly rebase to master. You shouldn't merge the new commits from master but do a proper rebase .
git fetch origin
git rebase -i origin/master 
So remove this ugly merge commit.

plugins/obs-ffmpeg/obs-ffmpeg-mux.c
```diff
@@ -380,6 +380,20 @@ static bool ffmpeg_mux_start(void *data)
...
+	settings = obs_output_get_settings(stream->output);
+	path = obs_data_get_string(settings, "path");
+	ext = strrchr(path, '.');
+	obs_data_release(settings);
+
+	if (strcmp(ext, ".m3u8") == 0) {
+		obs_encoder_t *vencoder;
+		vencoder = obs_output_get_video_encoder(stream->output);
+		settings = obs_encoder_get_settings(vencoder);
+		obs_data_set_bool(settings, "repeat_headers", true);
```
pkviet, line 394: change the name of this setting (ex: amf_repeat_headers) and put everything behind a check of the encoder id (h264_texture_amf or h265_texture_amf) (don't bother with the fallback encoders).

plugins/obs-ffmpeg/texture-amf.cpp
```diff
@@ -933,6 +933,7 @@ static void amf_defaults(obs_data_t *settings)
 	obs_data_set_default_string(settings, "rate_control", "CBR");
 	obs_data_set_default_string(settings, "preset", "quality");
 	obs_data_set_default_string(settings, "profile", "high");
+	obs_data_set_default_bool(settings, "repeat_headers", false);
```
pkviet, line 936: change the setting name, say : amf_repeat_headers

plugins/obs-ffmpeg/texture-amf.cpp
```diff
@@ -1158,6 +1162,11 @@ static bool amf_avc_init(void *data, obs_data_t *settings)
 				    : 250;
 
 	set_avc_property(enc, IDR_PERIOD, gop_size);
+
+	bool repeat_headers = obs_data_get_bool(settings, "repeat_headers");
```
pkviet, line 1166: change the setting name here too, say : amf_repeat_headers

pkviet (changes requested):

pkviet:
also, for future reference, @RytoEX unearthed this discussion :https://github.com/GPUOpen-LibrariesAndSDKs/AMF/issues/293
Without these infos, it would have been impossible to guess that HEADER_INSERTION_SPACING would insert sps/pps ...

pkviet (changes requested):
please make the changes I suggested.
Also in order to help us to reproduce the issue, please provide us with info regarding your GPU (which model).

shinji3:
I'm using a Radeon RX550

pkviet:
a single commit is enough, no need to make 3 commits

flaeri:
Tested the PR (37c8da56d) on texture and fallback, h264 and HEVC. AMD r5 4500U APU, and I am able to playback the local ts files as standalone.

I was not able to prior (on release).

pkviet:
LGTM; resolve the conversations and we should be good

plugins/obs-ffmpeg/texture-amf.cpp
RytoEX: Resolved per off-thread request.

RytoEX:
Waiting on blocking conversations to be resolved before merging.

RytoEX:
> also, for future reference, @RytoEX unearthed this discussion :[GPUOpen-LibrariesAndSDKs/AMF#293](https://github.com/GPUOpen-LibrariesAndSDKs/AMF/issues/293) Without these infos, it would have been impossible to guess that HEADER_INSERTION_SPACING would insert sps/pps ...

Follow-up that the [v1.4.24 AMF_Video_Encode_API.pdf](https://raw.githubusercontent.com/GPUOpen-LibrariesAndSDKs/AMF/v1.4.24/amf/doc/AMF_Video_Encode_API.pdf) mentions:
> ** HeaderInsertionSpacing: Every IDR frame has SPS and PPS regardless of default value of HeaderInsertionSpacing
per VCE logic.

But [AMD says](https://github.com/GPUOpen-LibrariesAndSDKs/AMF/issues/293#issuecomment-1051301399):
> if you want that SPS/PPS inserted with every IDR controlled by AMF_VIDEO_ENCODER_IDR_PERIOD please set AMF_VIDEO_ENCODER_HEADER_INSERTION_SPACING to the same value.

Those seem to say the opposite thing to me, but perhaps I've misunderstood.

Anyway, I _think_ this is okay to land. Will look over it again. Other approvals also welcome.

RytoEX (approved):
Looks fine at a glance.

jp9000:
Do the other encoders have this behavior by default?

pkviet:
> Do the other encoders have this behavior by default?

i checked that it is not the case with nvenc h264/hevc & x264. The ts recording works fine. It's only with amf oddly enough.

jp9000:
> i checked that it is not the case with nvenc h264/hevc & x264. The ts recording works fine. It's only with amf oddly enough.

Respectfully, "oddly enough" is an uncomfortable answer. It would be nice to know why specifically this only affects AMF and not nvenc/x264.

flaeri:
I cannot test nvenc atm, due to not having it accessible for the next few wekks.

I can however confirm that x264 TS segments (from m3u8) does contain PPS and SPS by default.

If someone would be so kind as to drop me said ts segment from nvenc, I'll be happy to check.

AMF as establish does not contain them in the current release build.
image

jp9000:
I wonder if I should just turn on header repetition by default for amf.

derrod:
> I wonder if I should just turn on header repetition by default for amf.

I believe that would result in issues with other container formats such as MP4 or MKV.

shinji3:
>I wonder if I should just turn on header repetition by default for amf.

flv, mp4, mov, mkv, ts, m3u8 could be recorded without issues.

PR #7384, opened 2022-09-15
CI: Use official Flatpak actions

GeorgesStavracas:
### Description

The actions from Bilal were moved [to the Flatpak organization in GitHub](https://github.com/flatpak/flatpak-github-actions/), so let's just use that.

### Motivation and Context

Routine cleanup.

### How Has This Been Tested?

 - Let CI pass

### Types of changes

### Checklist:

### DIFF
--- a/.github/workflows/flatpak.yml
+++ b/.github/workflows/flatpak.yml
@@ -50,7 +50,7 @@ jobs:
     env:
       FLATPAK_BUILD_PATH: flatpak_app/files/share
     container:
-      image: bilelmoussaoui/flatpak-github-actions:kde-6.3
+      image: flatpak/flatpak-github-actions:kde-6.3
       options: --privileged
     strategy:
       matrix:
@@ -68,7 +68,7 @@ jobs:
           echo "::set-output name=commitHash::$(git rev-parse --short=9 HEAD)"
 
       - name: Build Flatpak Manifest
-        uses: bilelmoussaoui/flatpak-github-actions/flatpak-builder@master
+        uses: flatpak/flatpak-github-actions/flatpak-builder@master
         with:
           bundle: obs-studio-${{ steps.setup.outputs.commitHash }}.flatpak
           manifest-path: CI/flatpak/com.obsproject.Studio.json
@@ -94,7 +94,7 @@ jobs:
           ostree commit --repo=repo --canonical-permissions --branch=screenshots/x86_64 flatpak_app/screenshots
 
       - name: Publish to Flathub Beta
-        uses: bilelmoussaoui/flatpak-github-actions/flat-manager@v4
+        uses: flatpak/flatpak-github-actions/flat-manager@master
         if: matrix.branch == 'beta'
         with:
           flat-manager-url: https://hub.flathub.org/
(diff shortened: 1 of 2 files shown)


GeorgesStavracas:
Draft because it doesn't work right now

.github/workflows/flatpak.yml
```diff
@@ -50,7 +50,7 @@ jobs:
     env:
       FLATPAK_BUILD_PATH: flatpak_app/files/share
     container:
-      image: bilelmoussaoui/flatpak-github-actions:kde-6.3
+      image: flatpak/flatpak-github-actions:kde-6.3
```
tytan652, line 53: They still didn't switch to github registry for container so it's still bilelmoussaoui for containers.

.github/workflows/main.yml
```diff
@@ -399,7 +399,7 @@ jobs:
           echo "OBS_GIT_TAG=$(git describe --tags --abbrev=0)" >> $GITHUB_ENV
 
       - name: Build Flatpak Manifest
-        uses: bilelmoussaoui/flatpak-github-actions/flatpak-builder@v4
+        uses: flatpak/flatpak-github-actions/flatpak-builder@master
```
tytan652, line 402: Switching to master is a bad idea, we should stick with v4.

.github/workflows/flatpak.yml
```diff
@@ -68,7 +68,7 @@ jobs:
           echo "::set-output name=commitHash::$(git rev-parse --short=9 HEAD)"
 
       - name: Build Flatpak Manifest
-        uses: bilelmoussaoui/flatpak-github-actions/flatpak-builder@master
+        uses: flatpak/flatpak-github-actions/flatpak-builder@master
```
tytan652, line 71: This one should maybe be reverted to v4.

tytan652 (commented):
bilelmoussaoui's is the official Flatpak actions because it now's redirect to the flatpak repo, it's more a avoid repo redirection PR than a use official repo.

But container side no change should be done, because they still didn't switch registry.

.github/workflows/flatpak.yml
pheiduck: They didn’t update there README https://github.com/flatpak/flatpak-github-actions so change is correct

.github/workflows/main.yml
pheiduck: And yes, for stability version tags are better

.github/workflows/flatpak.yml
pheiduck: Same here

.github/workflows/flatpak.yml
derrod: No the change is incorrect. The docker image is still not published on the flatpak docker hub or github packages (https://hub.docker.com/u/flatpak), only on https://hub.docker.com/u/bilelmoussaoui

tytan652: > They didn’t update there README https://github.com/flatpak/flatpak-github-actions so change is correct

The pull request to switch to Github registry as `flatpak` is not merge, for now it's in docker registry as `bilelmoussaoui`.

GeorgesStavracas:
I'm aware of all these issues folks, and I'm working on fixing them in https://github.com/flatpak/flatpak-github-actions - this is why this PR is a draft in the first place

PR #7393, opened 2022-09-17
libobs: Add "source_update" signal

gxalpha:
### Description

Adds a source signal that gets triggered when a source's settings are updated.

~~There is `update_properties`, but it acts in a very different way.
Confirmation that it's not *supposed* to do what the new signal does would be great.~~

### Motivation and Context

Writing a feature / plugin (probably plugin for the time being) that requires constantly watching a source's settings.

### How Has This Been Tested?

macOS 13
Connected to the signal via `signal_handler_connect(obs_source_get_signal_handler(source), "update", source_update_signal, data)` and saw the signal being triggered each time the source was updated.

### Types of changes

### Checklist:

### DIFF
--- a/docs/sphinx/reference-core.rst
+++ b/docs/sphinx/reference-core.rst
@@ -582,6 +582,10 @@ Core OBS Signals
    Called when a source has been removed (:c:func:`obs_source_remove()`
    has been called on the source).
 
+**source_update** (ptr source)
+
+   Called when a source's settings have been updated.
+
 **source_save** (ptr source)
 
    Called when a source is being saved.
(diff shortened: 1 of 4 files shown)


gxalpha:
I'm now certain that the `update_properties` and the new signal I added are not the same.
> Tells any open source properties windows to perform a UI refresh

Websockets to the rescue again!

PatTheMav:
Seems fine, however I'm not sure if this event shouldn't be named "source_properties_update" because I'd expect a "source_update" to signal an actual change triggered by the source itself.

gxalpha:
`source_properties_update` would indicate that the signal is something about the properties, which it isn't ("properties" is exclusively used with the the properties API in every other OBS terminology).
Other signals like `source_add`, `source_show`, `source_activate` etc aren't triggered by the source either, so this PR is just following convention.

PR #7395, opened 2022-09-18
linux-v4l2: Improve error handling in interacting with V4L2 loopback module

shoffmeister:
### Description

Improve error detection and general (debug, warning, error) diagnostics for interacting with the V4L2 loopback kernel module via the OBS V4L2 output component.

### Motivation and Context

Fix API error checking; enable diagnosis of runtime challenges in log output.

### How Has This Been Tested?

Manual testing of a portable OBS build on Fedora Linux 36.

With (only) this PR applied, diagnostic information now clearly indicates that some functional challenges exist in combination with the version of the V4L2 loopback module shipped with Fedora Linux 36 (which is, at the time of this writing, the latest released version of this module).

This (improved diagnostic, bad functional) situation will remain for the upcoming Fedora Linux 37, and any other Linux distribution using clean release, unpatched versions of the V4L2 loopback Linux kernel module.

Note that these improvements go beyond a specific Linux distribution or a specific version of the V4L2 loopback kernel module; they simply make sure that API contracts are followed and that appropriate logging is emitted in case problems are detected.

### Types of changes

### Checklist:

### DIFF
--- a/plugins/linux-v4l2/v4l2-output.c
+++ b/plugins/linux-v4l2/v4l2-output.c
@@ -26,7 +26,11 @@ static const char *virtualcam_name(void *unused)
 static void virtualcam_destroy(void *data)
 {
 	struct virtualcam_data *vcam = (struct virtualcam_data *)data;
-	close(vcam->device);
+	if (!close(vcam->device)) {
+		blog(LOG_WARNING,
+		     "Failed to close virtual camera file descriptor %d (%d)",
+		     vcam->device, errno);
+	}
 	bfree(data);
 }
 
@@ -55,6 +59,10 @@ static int run_command(const char *command)
 
 	dstr_cat(&str, command);
 	result = system(str.array);
+	if (result == -1) {
+		blog(LOG_ERROR, "Failed to run command '%s' (%d)", str.array,
+		     errno);
+	}
 	dstr_free(&str);
 	return result;
 }
@@ -67,8 +75,10 @@ static bool loopback_module_loaded()
 
 	FILE *fp = fopen("/proc/modules", "r");
 
-	if (!fp)
+	if (!fp) {
+		blog(LOG_WARNING, "Failed to open /proc/modules (%d)", errno);
 		return false;
+	}
 
 	while (fgets(temp, sizeof(temp), fp)) {
 		if (strstr(temp, "v4l2loopback")) {
@@ -77,8 +87,11 @@ static bool loopback_module_loaded()
 		}
 	}
 
-	fclose(fp);
-
+	if (!fclose(fp)) {
+		blog(LOG_WARNING,
+		     "Failed to close loopback module check file descriptor (%d)",
+		     errno);
+	}
 	return loaded;
 }
 
@@ -125,19 +138,57 @@ static bool try_connect(void *data, const char *device)
 
 	vcam->device = open(device, O_RDWR);
 
-	if (vcam->device == -1)
+	if (vcam->device == -1) {
+		blog(LOG_ERROR, "Failed to open device '%s' (%d)", device,
+		     errno);
 		return false;
+	}
 
-	if (ioctl(vcam->device, VIDIOC_QUERYCAP, &capability) == -1)
+	if (ioctl(vcam->device, VIDIOC_QUERYCAP, &capability) == -1) {
+		blog(LOG_WARNING,
+		     "Failed to query device capabilities on '%s' (%d)", device,
+		     errno);
 		goto fail_close_device;
+	}
 
+	blog(LOG_DEBUG,
+	     "Virtual video device: driver='%s', card='%s', capabilities=0x%x, device_caps=0x%x",
+	     capability.driver, capability.card, capability.capabilities,
+	     capability.device_caps);
+
+	/* The v42loopback driver always supports reporting device node capabilities.
+		Here we just make sure to properly follow the API contract which requires
+		that before accessing the content of capability.device_caps, we first
+		check if the driver does in fact support reporting device capabilities */
+	if (!(capability.capabilities & V4L2_CAP_DEVICE_CAPS)) {
+		blog(LOG_ERROR,
+		     "Driver on device '%s' is unable to report on device node capabilities",
+		     device);
+		goto fail_close_device;
+	}
+
+	if (!(capability.device_caps & V4L2_CAP_VIDEO_OUTPUT)) {
+		blog(LOG_ERROR, "Device '%s' does not support video output",
+		     device);
+		goto fail_close_device;
+	}
+
+	memset(&format, 0, sizeof(format));
 	format.type = V4L2_BUF_TYPE_VIDEO_OUTPUT;
 
-	if (ioctl(vcam->device, VIDIOC_G_FMT, &format) == -1)
+	if (ioctl(vcam->device, VIDIOC_G_FMT, &format) == -1) {
+		blog(LOG_ERROR,
+		     "Failed to get video output format from device '%s' (%d)",
+		     device, errno);
 		goto fail_close_device;
+	}
 
 	struct obs_video_info ovi;
-	obs_get_video_info(&ovi);
+	if (!obs_get_video_info(&ovi)) {
+		blog(LOG_WARNING, "Failed to get video info for device '%s'",
+		     device);
+		// .. but do continue anyway
+	}
 
 	memset(&parm, 0, sizeof(parm));
 	parm.type = V4L2_BUF_TYPE_VIDEO_OUTPUT;
@@ -146,16 +197,24 @@ static bool try_connect(void *data, const char *device)
 	parm.parm.output.timeperframe.numerator = ovi.fps_den;
 	parm.parm.output.timeperframe.denominator = ovi.fps_num;
 
-	if (ioctl(vcam->device, VIDIOC_S_PARM, &parm) == -1)
+	if (ioctl(vcam->device, VIDIOC_S_PARM, &parm) == -1) {
+		blog(LOG_WARNING,
+		     "Failed to set video output format parameter for time per frame on device '%s' (%d)",
+		     device, errno);
 		goto fail_close_device;
+	}
 
 	format.fmt.pix.width = width;
 	format.fmt.pix.height = height;
 	format.fmt.pix.pixelformat = V4L2_PIX_FMT_YUYV;
 	format.fmt.pix.sizeimage = vcam->frame_size;
 
-	if (ioctl(vcam->device, VIDIOC_S_FMT, &format) == -1)
+	if (ioctl(vcam->device, VIDIOC_S_FMT, &format) == -1) {
+		blog(LOG_ERROR,
+		     "Failed to set video output format to YUYV on device '%s' (%d)",
+		     device, errno);
 		goto fail_close_device;
+	}
 
 	struct video_scale_info vsi = {0};
 	vsi.format = VIDEO_FORMAT_YUY2;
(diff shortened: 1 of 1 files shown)


shoffmeister:
Note that in the logging I did not add prefixes containing either the source file or the function. There does not seem to be logging API support for that and the existing usage of `blog` is all over the place.

I'll be glad to amend the affected log message to comply with log policy.

shoffmeister:
Ping @kkartaltepe and @GeorgesStavracas - I wonder whether you could review these (IMHO) uncontroversial improvements to error handling?

kkartaltepe:
provide some justification for checking closes and reporting these to the user. Such a failure is almost never going to cause the user any issues or be helpful for any user issues.

`try_connect` is used to scan devices and should not output spam messages about all the devices its checking so all of that logging seems like it would make the log viewing experience worse.

Adding logging to run_command and then duplicating the logging in other places, for example in calls to `loopback_module_load`. This particular example is already reported to the user via the gui and logged that vc failed to start, so doesnt seem particularly helpful to assist users or people assisting users.

Then there is lot of logging of FDs and assorted things that are totally useless without a debugger, but if you have a debugger you can already set watchpoints and introspect the memory to associate FDs so im not sure why these are added.

Maybe someone elses sees more value in this, but it seems like you are trying to make due without a debugger by adding lots of user facing logging. The v4l2 plugin generally logs far too much usually useless information so personally id prefer we not make it worse for our users because the debugger experience in linux is poor.

shoffmeister:
None of that added logging will ever be seen on the happy path, when everything is/was working as expected.

The added logging is meant to, and will help, e.g., if a user starts `obs` from the command-line, notices misbehaviour, and then does copy&paste of the output found there. Users don't tend to have debuggers installed, and do not tend to be proficient in running debuggers.

Note that I use LOG_DEBUG in places - where the default log level is LOG_INFO, so none of that information will ever be shown until the support case comes up; then, a user could be instructed to enable logging at DEBUG level, and more information will be printed.

I do agree that logging failure on `close` is taking a rather highly detailed approach - yet, it will only happen on unhappy paths and may actually contribute to identifying unhappy paths. Or, in short, doing all that is part of return value checking hygiene, typically results in more stable software, in my experience (if only `__attribute__ ((warn_unused_result))` was a thing, and strong API contracts, and "_oh, no, the plain C language_" ;)

There could be further objections that the added logging introduces non-functional detriment. Specifically, it could be argued that
* performing the actual test on the return code adds two or three CPU instructions and may pollute OO execution caches - but this is never on the hot path, so improving maintainability and diagnosability seems to be well-worth it
* even at log level DEBUG the code will make a call to the log _function_ and check against the log level, resulting in CPU consumption - response see above
* the added TEXT and CONST DATA bytes increase software size - response see above

I'll be happy to submit another, separate PR which _reduces_ the currently generated log output in `v4l2-output.c` - but with this PR here I am only focusing on making possible to diagnose things _off the happy path_, in the fault scenario, without a debugger.

kkartaltepe:
> None of that added logging will ever be seen on the happy path, when everything is/was working as expected.
> 
> The added logging is meant to, and will help, e.g., if a user starts `obs` from the command-line, notices misbehaviour, and then does copy&paste of the output found there. Users don't tend to have debuggers installed, and do not tend to be proficient in running debuggers.

Sorry my opinion is that without a debugger in the session, i do not see these logs providing enough information to change anyone's behavior in trying to diagnose the problem (for example what problems do you foresee these lines diagnosing). And adding enough logging would just make it a horrible spamy mess. If you have a debugger it might provide some code tracing and let you avoid some breakpoints but that doesnt seem like its worth maintaining.

> Note that I use LOG_DEBUG in places - where the default log level is LOG_INFO, so none of that information will ever be shown until the support case comes up; then, a user could be instructed to enable logging at DEBUG level, and more information will be printed.
> 
> I do agree that logging failure on `close` is taking a rather highly detailed approach - yet, it will only happen on unhappy paths and may actually contribute to identifying unhappy paths. Or, in short, doing all that is part of return value checking hygiene, typically results in more stable software, in my experience (if only `__attribute__ ((warn_unused_result))` was a thing, and strong API contracts, and "_oh, no, the plain C language_" ;)

Code that exists solely for ideological purity is generally bad code. As above I would want additional logging to change someones behavior, that of the user or that of person helping them remotely.

> There could be further objections that the added logging introduces non-functional detriment. Specifically, it could be argued that
> 
>     * performing the actual test on the return code adds two or three CPU instructions and may pollute OO execution caches - but this is never on the hot path, so improving maintainability and diagnosability seems to be well-worth it
> 
>     * even at log level DEBUG the code will make a call to the log _function_ and check against the log level, resulting in CPU consumption - response see above
> 
>     * the added TEXT and CONST DATA bytes increase software size - response see above
> 

right none of this is on a hot path I have no qualms about adding a few bytes here and a few branches there.

plugins/linux-v4l2/v4l2-output.c
```diff
@@ -77,8 +87,11 @@ static bool loopback_module_loaded()
 		}
 	}
 
-	fclose(fp);
-
+	if (!fclose(fp)) {
+		blog(LOG_WARNING,
```
PatTheMav, line 91: See above, the level should depend on whether it it possible for `fclose` to fail under normal circumstances.

E.g. if the OS itself can close the file descriptor at any point and there is no way for us to notice and this is a "normal" occurrence on Linux.

plugins/linux-v4l2/v4l2-output.c
```diff
@@ -125,19 +138,57 @@ static bool try_connect(void *data, const char *device)
...
+
+	if (ioctl(vcam->device, VIDIOC_QUERYCAP, &capability) == -1) {
+		blog(LOG_WARNING,
+		     "Failed to query device capabilities on '%s' (%d)", device,
+		     errno);
+		goto fail_close_device;
+	}
 
-	if (ioctl(vcam->device, VIDIOC_QUERYCAP, &capability) < 0)
+	blog(LOG_DEBUG,
```
PatTheMav, line 154: Personally I rather guard debug output with a check for the `DEBUG` preprocessor definition so the code doesn't even make it into optimised builds.

plugins/linux-v4l2/v4l2-output.c
```diff
@@ -125,19 +138,57 @@ static bool try_connect(void *data, const char *device)
...
 
-	if (vcam->device < 0)
+	if (vcam->device == -1) {
+		blog(LOG_ERROR, "Failed to open device '%s' (%d)", device,
+		     errno);
 		return false;
+	}
+
+	if (ioctl(vcam->device, VIDIOC_QUERYCAP, &capability) == -1) {
+		blog(LOG_WARNING,
```
PatTheMav, line 148: Would it be possible to store an error message in string here and then add an output call after the `fail_close_device` label instead? That label seems to serve as the bail-out point for all errors after this point.

PatTheMav:
I agree with @kkartaltepe that if `try_connect` is _expected_ to fail in some circumstances (hence the `try` part in its name) then it might make more sense to limit the messages to `LOG_DEBUG` level and also guard the call to `blog` with a preprocessor check for the `DEBUG` define.

Debug output in code has the benefit of not negatively impacting runtime behaviour compared to breakpoints with gdb/lldb outputs attached.

plugins/linux-v4l2/v4l2-output.c
stephematician: _v4l2loopback_ module has a debug parameter which will provide output in the system journal, e.g. `v4l2loopback debug=2 ...` will add output summarising the result of each IOCTL (except for the per-frame spammy ones like e.g. queue/dequeue).

plugins/linux-v4l2/v4l2-output.c
```diff
@@ -125,19 +138,57 @@ static bool try_connect(void *data, const char *device)
...
+		check if the driver does in fact support reporting device capabilities */
+	if (!(capability.capabilities & V4L2_CAP_DEVICE_CAPS)) {
+		blog(LOG_ERROR,
+		     "Driver on device '%s' is unable to report on device node capabilities",
+		     device);
 		goto fail_close_device;
+	}
 
+	if (!(capability.device_caps & V4L2_CAP_VIDEO_OUTPUT)) {
+		blog(LOG_ERROR, "Device '%s' does not support video output",
```
stephematician, line 171: This check has already been added to PR #11906 - but failing should not be considered an 'error'; OBS tests devices in order to find one with output capabilities. It expects that some of them won't have output.

Also - see comment at line 148 about the debug output that is available with v4l2loopback.

plugins/linux-v4l2/v4l2-output.c
```diff
@@ -125,19 +138,57 @@ static bool try_connect(void *data, const char *device)
...
+		goto fail_close_device;
+	}
+
+	memset(&format, 0, sizeof(format));
 	format.type = V4L2_BUF_TYPE_VIDEO_OUTPUT;
 
-	if (ioctl(vcam->device, VIDIOC_G_FMT, &format) < 0)
+	if (ioctl(vcam->device, VIDIOC_G_FMT, &format) == -1) {
+		blog(LOG_ERROR,
+		     "Failed to get video output format from device '%s' (%d)",
```
stephematician, line 181: Also - see comment at line 148 about the debug output that is available with v4l2loopback.

plugins/linux-v4l2/v4l2-output.c
```diff
@@ -146,16 +197,24 @@ static bool try_connect(void *data, const char *device)
 	parm.parm.output.timeperframe.numerator = ovi.fps_den;
 	parm.parm.output.timeperframe.denominator = ovi.fps_num;
 
-	if (ioctl(vcam->device, VIDIOC_S_PARM, &parm) < 0)
+	if (ioctl(vcam->device, VIDIOC_S_PARM, &parm) == -1) {
+		blog(LOG_WARNING,
```
stephematician, line 201: Also - see comment at line 148 about the debug output that is available with v4l2loopback.

plugins/linux-v4l2/v4l2-output.c
```diff
@@ -166,14 +225,20 @@ static bool try_connect(void *data, const char *device)
 	memset(&parm, 0, sizeof(parm));
 	parm.type = V4L2_BUF_TYPE_VIDEO_OUTPUT;
 
-	if (ioctl(vcam->device, VIDIOC_STREAMON, &parm) < 0) {
+	if (ioctl(vcam->device, VIDIOC_STREAMON, &parm) == -1) {
```
stephematician, line 228: Note that a lot of this PR is now (quite) behind - see e.g. PR #11906

plugins/linux-v4l2/v4l2-output.c
```diff
@@ -146,16 +197,24 @@ static bool try_connect(void *data, const char *device)
...
 		goto fail_close_device;
+	}
 
 	format.fmt.pix.width = width;
 	format.fmt.pix.height = height;
 	format.fmt.pix.pixelformat = V4L2_PIX_FMT_YUYV;
 	format.fmt.pix.sizeimage = vcam->frame_size;
 
-	if (ioctl(vcam->device, VIDIOC_S_FMT, &format) < 0)
+	if (ioctl(vcam->device, VIDIOC_S_FMT, &format) == -1) {
```
stephematician, line 212: Also - see comment at line 148 about the debug output that is available with v4l2loopback.

stephematician (commented):
Many of these can be removed by using _v4l2loopback_'s logging capability when `debug=2` is set in the module.

Warchamp7:

PR #7396, opened 2022-09-18
CI: Enable running code format check locally

shoffmeister:
### Description

Supply a script and a Dockerfile enabling containerized local running of the exact rules applied by OBS (remote) CI check format.

### Motivation and Context

Part of OBS CI is a dedicated check that all code conforms to defined policy through _exactly_ clang-format 13.

On some Linux distributions - like Fedora Linux 36 - the _exact_ version of the clang-format dependency is not conveniently available, while other versions of clang-format - e.g. clang-format 14 - apply materially different, incompatible rules in their formatting.

Improve the developer experience around this policy such that these checks can be run locally, through a docker container, before CI kicks in. This will spare developers the effort to rewrite commits after they have been pushed; it also shortens the feedback cycle on PRs.

This contribution might also apply to other host operating systems; any benefits will depend on the respective local developer setup.

### How Has This Been Tested?

My last two pull requests were created supported by the content of this PR.

### Types of changes

### Checklist:

### DIFF
--- a/CI/linux/run-code-format-check.dockerfile
+++ b/CI/linux/run-code-format-check.dockerfile
@@ -1,0 +1,17 @@
+FROM ubuntu:22.04
+
+RUN apt update
+
+# Baseline infrastructure
+RUN apt install -y \
+    apt-transport-https \
+    ca-certificates \
+    curl \
+    git
+
+# Tooling required to run the checks
+RUN apt install -y \
+    clang-format-13 \
+    pip
+
+RUN pip install cmakelang
--- a/CI/linux/run-code-format-check.sh
+++ b/CI/linux/run-code-format-check.sh
@@ -1,0 +1,48 @@
+#!/bin/bash
+
+#
+# This script enables running code format checks in a control
+# fashion independent of the CI environment, but with the
+# results of the CI environment.
+#
+# Execution takes place inside a docker container to be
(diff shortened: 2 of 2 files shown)


tytan652:
There is static binaries of clang-format (check releases assets): https://github.com/muttleyxd/clang-tools-static-binaries

Which is simpler than using docker just to format code.

derrod:
Yeah using docker seems *massively* overkill. We know the clang-format situation sucks, but thankfully 13 and 14 are *mostly* compatible and for those edge cases where they aren't you can use the staic binaries linked by tytan (we should probably add that to the wiki though).

shoffmeister:
FWIW, these two (tiny, very lightweight) files allow running locally what would otherwise fail in CI.

The next step beyond that (and, IMHO, the right step) would be enabling matching build steps when run locally and on CI (which basically translates to a fully dockerized build environment). I do not have the time for implementing this for OBS, I am afraid.

PatTheMav:
Using Docker to achieve this seems very wasteful to me when the alternative is to simply check out llvm and build/install clang-format-13 locally (I use Linuxbrew to install it via a formula).

shoffmeister:
Given the above feedback, I am closing this PR. Thanks for the input!

Issue #7333, opened 2022-09-09
Virtual Cam shows a red screen instead of video


markflorkowski:
### Operating System Info

macOS 12

### OBS Studio Version

28.0.1

### OBS Studio Log URL

https://obsproject.com/logs/i5aP5GOrwhY0Sic1

### Expected Behavior

Virtual camera should show the currently selected output scene

### Current Behavior

Red screen in application attempting to use virtual camera

<img width="1785" alt="image" src="https://user-images.githubusercontent.com/11494384/189426467-c7edf219-e782-4284-9b96-129dfa10c54f.png">

### Steps to Reproduce

1. Open OBS, start virutal camera
2. attempt to use virtual camera in google chrome (eg. meet.google.com)
3. note that camera just shows red
...

### Anything else we should know?

On a 2019 MacBook Pro with the following specs:
- macOS version: 12.3.1
- Processor: 2.4 GHz 8-Core Intel Core i9
- Memory: 64 GB 2667 MHz DDR4
- Graphics: AMD Radeon Pro 5500M 8 GB, Intel UHD Graphics 630 1536 MB

BenDMyers:
I was getting this issue today as well.

**OBS version:** 28.0.1

On a 2019 MacBook Pro with the following specs:
- **macOS version:** 12.5.1
- **Processor:** 2.3 GHz -Core Intel Core i9
- **Memory:** 16 GB
- **Graphics:** AMD Radeon Pro 5500M 8 GB, Intel UHD Graphics 630 1536 MB

Encoded with the Apple VT H264 Hardware Encoder.

markflorkowski:
The placeholder shows correctly if I disable the virtual camera in OBS and use it in an application

<img width="1709" alt="image" src="https://user-images.githubusercontent.com/11494384/189436049-80c076ba-db62-4994-ba5c-9c8cf92356b0.png">

TheoBr:
I have not been able to replicate on either my M1 Macbook Air or M1 Pro, but can reliably repro on my intel machine. I am relatively certain this is intel specific

markflorkowski:
Also was unable to replicate on my Mac Studio, so that is one more point towards it being an Intel-specific issue

prgmitchell:
Can confirm I am getting a blank green screen when trying to use virtual cam on 28.0.1. Also on an intel macbook pro.

PatTheMav:
[...]

Issue #7171, opened 2022-08-23
Vitrual Cam's Picture Black


buduan:
### Operating System Info

macOS 12

### Other OS

macOS

### OBS Studio Version

27.2.4

### OBS Studio Log URL

https://obsproject.com/logs/1gKXEuFQvGuGhOa3

### Expected Behavior

Output the OBS's Scene to Meeting App.
<img width="1497" alt="image" src="https://user-images.githubusercontent.com/39254250/186082051-a8f772db-ee84-4eef-9f2f-5881d44592c5.png">

### Current Behavior

When I chose _OBS Virtual Cam_ as my camera, the output picture was black.

### Steps to Reproduce

1.Edit Scene
2.Start Virtual cam
3.choose OBS Virtual Cam in Dingtalk
4.Black!
...

### Anything else we should know?

I've tried it on Dingtalk and Telegram.All black

Voring:
我今天手贱更新了28版本的obs也是无法正常输出画面，原来好像是25版本的，之前一直正常使用，现在我试试旧版本的吧。

PatTheMav:
Possible duplicate of https://github.com/obsproject/obs-studio/issues/7333.

PR #7403, opened 2022-09-18
mac-virtualcam: Fix frame sharing and colourspace issues

PatTheMav:
### Description
Fixes an issue with empty IOSurface references on Intel-based Macs and also broken output when OBS is set to deliver full colour range output.

### Motivation and Context
The plugin was missing necessary IOSurface locks to ensure that modifications to the underlying memory reference made by the OS do not occur when client code accesses the image data.

On machines without unified memory this can lead to IOSurfaces pointing to GPU memory even though the data has been moved to CPU memory by the OS (to avoid unnecessarily large copy operations) and a PixelBuffer created from the IOSurface won't contain data.

In addition to the issue above, FFmpeg's `swscale` seems to have issues with converting images with full colour ranges to limited colour ranges, resulting in mangled frame data inside OBS.

As macOS natively supports partial range as well as as full ranges for pixel buffers, use these directly, avoiding costly colour space conversions for even more use cases than before - as this skips swscale entirely, the issue doesn't occur anymore.

Fixes https://github.com/obsproject/obs-studio/issues/7333
Possibly fixes https://github.com/obsproject/obs-studio/issues/7171

### How Has This Been Tested?
Tested with macOS 12.6 on Intel-based and Apple-Silicon-based Macs.

### Types of changes

### Checklist:

### DIFF
--- a/plugins/mac-virtualcam/src/dal-plugin/OBSDALMachClient.mm
+++ b/plugins/mac-virtualcam/src/dal-plugin/OBSDALMachClient.mm
@@ -10,6 +10,7 @@
 #import "Logging.h"
 
 @interface OBSDALMachClient () <NSPortDelegate> {
+	uint32_t _seed;
 	NSPort *_receivePort;
 }
 @end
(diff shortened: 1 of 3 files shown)


PatTheMav:
Kudos to @ccgus for dropping the important hint on his blog: https://shapeof.com/archives/2017/12/moving_to_metal_episode_iii.html.

RytoEX:
Per off-thread conversation with @PatTheMav and @gxalpha , this is good to merge as-is, so I'm merging it.

PR #7406, opened 2022-09-19
UI: Fix in Light theme

shiina424:
### Description

Preview/Program Labels and T-Bar in Studio mode are not visible in the Light theme.
If this is incorrect, we would appreciate it if you could correct it.

Before
Before

After
After

### Motivation and Context

I want displayed correctly.

### How Has This Been Tested?

I changed the qss file and tested it.

### Types of changes

### Checklist:

### DIFF
--- a/UI/data/themes/Light.qss
+++ b/UI/data/themes/Light.qss
@@ -1346,6 +1346,12 @@ SceneTree {
     qproperty-icon: url(./Light/save.svg);
 }
 
+/* Studio Mode Labels */
+
+* [themeID="previewProgramLabels"] {
+    color: palette(text);
+}
+
 /* Studio Mode T-Bar */
 
 QSlider[themeID="tBarSlider"] {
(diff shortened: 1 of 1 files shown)


Warchamp7 (changes requested):
Please use the palette values where it makes sense, or when it aligns with the existing theme definitions.

In this case, the `text` palette color should be used for both of these as the labels are normal text, and the existing slider handles also use it

https://github.com/obsproject/obs-studio/blob/a90dc8ef2596e8f425841dcb2bb6fabf7b1423b1/UI/data/themes/Light.qss#L831-L832

RytoEX (approved):
Confirmed that the current changes fix the Light theme as described.

RytoEX:
Though, ideally I would like the commit message to be a bit more descriptive about what is being fixed.

Warchamp7 (approved):

Issue #7409, opened 2022-09-19
"Decay Rate" setting isn't reflected on the audio meters when switching profiles


Miltos95:
### Operating System Info

Windows 10

### OBS Studio Version

28.0.1

### OBS Studio Log URL

https://obsproject.com/logs/P64eMgPlaz0ZMU90

### Expected Behavior

When switching from one profile to another, the "Decay Rate" setting should be reflected in the audio meters, without needing a restart of the application.

### Current Behavior

The setting is correctly loaded in the Settings menu, but the audio meters use the Decay Rate that was being used before the new profile was loaded.

### Steps to Reproduce

1. Set the Decay Rate to Slow
2. Create a new Profile
3. Set the Decay Rate to anything else other than "Slow" (or whatever you chose in step 1)
4. Switch between the Profiles with the different Decay Rates and observe the behaviour of the audio meters

### Anything else we should know?

The setting is correctly reflected on the audio meters if the application is restarted.

PR #7411, opened 2022-09-19
UI: Update volume controls decay rate on profile switch

PatTheMav:
### Description
Ensures the volume controls decay rate is correctly updated on a profile switch, as there is no hook onto that UI event yet.

### Motivation and Context
Without this change the decay rate would only be updated when an audio source is added or when the volume meter itself is changed (e.g. from horizontal to vertical layout).

Fixes https://github.com/obsproject/obs-studio/issues/7409

### How Has This Been Tested?
Tested on macOS 12.6 with 2 profiles and different decay rates (fast vs slow).

### Types of changes

### Checklist:

### DIFF
--- a/UI/window-basic-main-profiles.cpp
+++ b/UI/window-basic-main-profiles.cpp
@@ -338,6 +338,7 @@ bool OBSBasic::CreateProfile(const std::string &newName, bool create_new,
 
 	config_save_safe(App()->GlobalConfig(), "tmp", nullptr);
 	UpdateTitleBar();
+	UpdateVolumeControlsDecayRate();
 
 	Auth::Load();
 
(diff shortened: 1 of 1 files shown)


RytoEX:
As-is, this fixes the issue for switching between profiles. It does not, apparently, fix a similar issue where if your current profile is set to a non-default value, creating a new profile will not change to the default value.

PatTheMav:
> As-is, this fixes the issue for switching between profiles. It does not, apparently, fix a similar issue where if your current profile is set to a non-default value, creating a new profile will not change to the default value.

Correct, amended the commit to also handle profile creation and profile removal.

RytoEX (approved):
Looks good to me.

derrod (approved):

Issue #7325, opened 2022-09-09
OBS Studio crash when deleting video filter


snicolai:
### Operating System Info

macOS 12

### OBS Studio Version

28.0.1

### OBS Studio Log URL

https://obsproject.com/logs/5wenb2HhQXV5RWsS

### Expected Behavior

OBS Studio should not crash.
[OBS Remove Filter Crash.txt](https://github.com/obsproject/obs-studio/files/9531458/OBS.Remove.Filter.Crash.txt)

### Current Behavior

See crash log.

### Steps to Reproduce

1.  I'm using the mac arm NDI plugin from https://github.com/Palakis/obs-ndi/actions/runs/2977380826
2. Add a color correction filter to the NDI source
3.  Mess with the sliders a bit4. 
4.  Click the delete button, watch OBS crash

gxalpha:
Does this only happen with NDI sources? If so, please report this to the plugin instead.

snicolai:
The same crash also happens with a El Gato CamLink source, so NDI isn't needed.

It also happens with a sharpen filter.  You have to change the settings without saving the filter for the crash to happen.  If you've closed the filter dialog (saved the filter) then the crash doesn't happen.

WizardCM:
I cannot reproduce this on Windows 28.0.1 with the Color Correction filter on a browser source and a video capture device source.

PatTheMav:
Confirmed the crash, this is yet another bug probably introduced by Qt6 and the automatic focus switch trying to set input focus on a destroyed widget (see also https://github.com/obsproject/obs-studio/issues/7259).

Exact reproduction steps:
* Add video input source
* Add colour correction filter
* Change a value of the colour correction filter
* Do not close the property window or accept the change
* Instead, delete the colour correction filter
* Observe crash

The root cause seems to be the following order of events:

* Deleting the filter triggers `QListWidget::currentRowChanged` (the selected row was the colour correction filter which has now been deleted)
* This triggers `OBSBasicFilters::on_effectFilters_currentRowChanged`, which calls `UpdatePropertiesView`
* This triggers `ui->propertiesFrame->setVisible(false);`
* Now this `setVisible` call triggers `QWidget::focusNextPrevChild`, which attempts to find a new widget to give the keyboard focus to
* This recursive search yields `FocusList::FocusList` as the Widget to set the focus to (triggering `FocusList::focusInEvent`)
[...]

PR #7412, opened 2022-09-19
UI: Fix crash when removing filter after changing a value.

PatTheMav:
### Description
Deleting a filter will trigger a visibility change, which will also trigger a focus change if the focus has not been on the list itself (e.g. after interacting with the property view).

When an async filter list is available in the view, it will be the first candidate to receive focus. If this list is empty, we hide the property view by default and set the view to a `nullptr`.

When the call for the visibility change returns, we need to check for this possibility, as another event might have hidden (and deleted) the view already.

### Motivation and Context
Fixes https://github.com/obsproject/obs-studio/issues/7325

### How Has This Been Tested?
Tested with Colour Correction filter on macOS 12.6.

### Types of changes

### Checklist:

### DIFF
--- a/UI/window-basic-filters.cpp
+++ b/UI/window-basic-filters.cpp
@@ -275,9 +275,25 @@ void OBSBasicFilters::UpdatePropertiesView(int row, bool async)
 	if (view) {
 		updatePropertiesSignal.Disconnect();
 		ui->propertiesFrame->setVisible(false);
-		view->hide();
-		view->deleteLater();
-		view = nullptr;
+		/* Deleting a filter will trigger a visibility change, which will also
+		 * trigger a focus change if the focus has not been on the list itself
+		 * (e.g. after interacting with the property view).
+		 *
+		 * When an async filter list is available in the view, it will be the first
+		 * candidate to receive focus. If this list is empty, we hide the property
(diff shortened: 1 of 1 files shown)


RytoEX (approved):
Discussed with @PatTheMav off-thread and the explanation of the fix makes sense to me.

derrod:
Kicked GitHub CI, seems to build successfully now.

derrod (approved):

UI/window-basic-filters.cpp
```diff
@@ -275,9 +275,11 @@ void OBSBasicFilters::UpdatePropertiesView(int row, bool async)
 	if (view) {
 		updatePropertiesSignal.Disconnect();
 		ui->propertiesFrame->setVisible(false);
-		view->hide();
-		view->deleteLater();
-		view = nullptr;
+		if (view) {
```
notr1ch, line 278: I think this could use an explanatory comment before it gets removed in a few years because "view was already null checked".

PatTheMav: Funnily enough the commit message started its life as a code comment and I thought to myself "well I explained it in the commit message", but doesn't hurt to add it here as well.

PR #7415, opened 2022-09-20
UI: Add system dark theme

cg2121:
### Description
This adds a system dark theme, so system themes that are dark can use the correct icons. It is just a copy of System.qss, except replacing the icons with the dark versions.

It doesn't show up in the theme list because it is programmatically selected if the System theme is selected and OBS detects it is dark.

**Using a KDE dark theme**
Before:
Screenshot_20220920_032420

After:
Screenshot_20220920_035716

### Motivation and Context
Make dark system themes work.

### How Has This Been Tested?
Installed a KDE dark theme.

### Types of changes
- Bug fix

### Checklist:

### DIFF
--- a/UI/data/themes/System-Dark.qss
+++ b/UI/data/themes/System-Dark.qss
@@ -11,47 +11,47 @@
 /* We need to set back the icons, or the preview wont stick. */
 
 * [themeID="addIconSmall"] {
-    qproperty-icon: url(:/res/images/plus.svg);
+    qproperty-icon: url(./Dark/plus.svg);
 }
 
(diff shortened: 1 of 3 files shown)


RytoEX:
I would rather we implement the upcoming System Dark Mode Aware support in Qt in the System Theme and let System Theme obey whatever the user has set in their OS Settings.

Fenrirthviti:
Thank you for this, we will likely use this as a reference, but let's hold off on any further changes to system themes until we can use the proper OS-aware switching. It will keep maintenance and having to clean up additional themes and potentially breaking user setups to a minimum.

PR #7417, opened 2022-09-20
obs-vst: Fix crashes and unusable property button on M1 Macs

PatTheMav:
### Description
Fixes crashes on M1 Macs introduced by the fact that most (if not all) VSTs available do not support the Apple Silicon (arm64) architecture.

### Motivation and Context
VST bundles might be present on user systems, but symbols will not be loaded as most VSTs do not contain arm64 binaries. This can lead to non-null bundle pointers to invalid references, which crash the plugin once it tries to free the references.

This PR ensures that no valid bundle pointers exist if the bundle wasn't loaded.

Also hides the VST UI button in the properties if no valid VST is currently loaded.

### How Has This Been Tested?
Tested on macOS 12.6 with M1 Max.

### Types of changes

### Checklist:

### DIFF
--- a/plugins/obs-vst/headers/VSTPlugin.h
+++ b/plugins/obs-vst/headers/VSTPlugin.h
@@ -98,6 +98,7 @@ public:
 	void getSourceNames();
 	obs_audio_data *process(struct obs_audio_data *audio);
 	bool openInterfaceWhenActive = false;
+	bool vstLoaded();
 
 	bool isEditorOpen();
 	void onEditorClosed();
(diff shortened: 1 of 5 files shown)


WizardCM (approved):
The check for `isEditorOpen()` does slightly concern me as this particular check has been a bit finnicky in the past (either from our own code or VSTs themselves misreporting).

Otherwise, this code looks fine to me. Haven't tested it myself though.

RytoEX:
Per off-thread conversation with @DDRBoxman , this is good to merge as-is.

PR #7420, opened 2022-09-20
UI: Alt-drag, alt-scroll to change cropped area

taxilian:
### Description

Currently it is very inconvenient to change what part of a video source you are zoomed into or to adjust to zoom in or out slightly if your needs change. Just as you can currently alt-resize to change the cropped area, this makes it possible to alt-drag to move the crop allowing you to "pan" around. It also adds a scroll option when holding alt which will "zoom" in or out of the cropped area, making that easy to adjust as well. This is the single biggest feature I've wished I had since I started using OBS, so I finally implemented it myself =]

There is no visible change to the UI except when in use -- when hovering over any video source you can hold alt and scroll in or out to change the zoom, preserving aspect ratio of the clipped area. You can also select and then click and drag a video source while holding alt to pan the image around inside the visible bounds.

### Motivation and Context
I use a high resolution (4k) camera and then use OBS to "zoom" in and out, allowing me to have nice zooming effects when broadcasting church meetings. This has been great, but if the camera get bumped or if I find I set it wrong I have to carefully drag the resize handles until I get everything back where it should be; that can be a huge hassle. Similarly sometimes I want to zoom in more if it's e.g a child speaking or out if it's a taller or larger person. With this change those are really really easy to do, whereas without it is quite a pain. I've seen multiple ways to work around this, but none are ideal. This feels to me like how it should have worked already =]

### How Has This Been Tested?
I tested this using the OBS scenes I use for church; there may be impacts in other parts of the application that I'm unfamiliar with, since I don't use e.g. full screen previewing, etc, but since alt / drag and alt / scroll are not AFAIK used for anything else I don't anticipate issues.

### Types of changes

### Checklist:


taxilian:
I'm having trouble running clang-format, so I'll circle back to that and push a fix for whatever style things I violated -- sorry about that.

Please let me know what concerns you have and anything else I can do to help this get merged -- every person I have talked to who was streaming meetings like I do has expressed the desire to have it.

gxalpha (changes requested):
Please make your code compatible with Qt 5 (via ifdefs), that will fix the CI.

Also, does this have undo/redo? I haven’t tested it, but from looking at the code it doesn’t look like it’s implemented.

Warchamp7:
The way many users accomplish this goal currently is by adding their camera to a scene, and then adding that scene as a source where they want the camera, cropping it to the size/position they want, and then using studio mode to resize/reposition the camera inside that "container" scene.

Workflow-wise, this is certainly an improvement!

I am good with the alt-drag functionality as is. I think it's intuitive enough and super useful for it's usecase.

The alt-scroll however is not currently working correctly for me, nor did it behave the way I expected. They could be the same bug, or they could not.

Currently, alt-scroll is always increasing the crop values regardless of whether I scroll up or down. I believe this is a bug.

Functionality-wise, my expectation for alt-scroll was that it would increase or decrease the size of the source within the bounds, centered around the cursor position. Currently it only does this sort of behaviour when bounding box is set to a value other than no bounds. When set to no bounds, it changes the size of the source due to the cropping changes.

Additionally, it also seems to perform this pseudo-scaling based around the positional alignment property. This makes sense from a design standpoint, but I feel like it's unintuitive.

---

In summary: I think the alt-drag is an excellent change. I would be fine with splitting that out to it's own PR and giving that my UX stamp of approval, pending a code review.

For alt-scroll, I'm not currently happy with it's behaviour given the above problems. I am also not sure it is a simple solution to make it work the way I "expect" given the current way source sizing and positioning works in OBS. It may simply not be possible.

taxilian:
It's possible that alt-scroll is only working correctly when you use it with a source which has a transformation set such that changing the crop doesn't change the bounds; I forget what that is called. There are a couple of challenges that I think are still there and could use some help from someone who knows the code better than I do.

The core idea is that when you scroll it should adjust all sides of the crop box in or out relative to the aspect ratio such that the center position remains the same but the image inside gets larger or smaller. The main challenges I hit were:

* First, you have to save the aspect ratio when you start scrolling rather than calculating it each time otherwise it will gradually change aspect ratio due to rounding issues
* Second, it's grabbing the item under the mouse pointer and not the *selected* item under the mouse pointer

It seems that it is in some cases resizing the bounds of the item as well, so I'll need to figure out why.

I will try to find some time to circle back and update this within a few days; I did use this on Sunday and found it extremely helpful, much more convenient than the other options. The workflow you describe certainly works, but only if you are using the whole canvas -- I like to use a picture-in-picture e.g. for the chorister conducting the music and it's common for me to need to adjust that quickly because ours likes to move around between songs for some reason. Being able to quickly zoom in / out and pan around saves a lot of time.

taxilian:
> Please make your code compatible with Qt 5 (via ifdefs), that will fix the CI.
> 
> Also, does this have undo/redo? I haven’t tested it, but from looking at the code it doesn’t look like it’s implemented.

I will have to figure out how to do that; I'll try to circle back in the next week to look at this and fix that and a few other issues.

Undo / redo should work fine AFAIK -- I used the same APIs that everything else in the file uses and when I tested it undo / redo worked fine. Granted it isn't convenient with alt-scroll because that's actually a sequence of rapid actions rather than a single action. I assume that the reason it works in alt-drag is because it's just using the same lifecycle as other drag events and the mousepress / mouseup events already capture things?

UI/window-basic-preview.cpp
```diff
@@ -49,6 +49,18 @@ vec2 OBSBasicPreview::GetMouseEventPos(QMouseEvent *event)
 
 	return pos;
 }
+vec2 OBSBasicPreview::GetWheelEventPos(QWheelEvent *event)
+{
+	OBSBasic *main = reinterpret_cast<OBSBasic *>(App()->GetMainWindow());
+	float pixelRatio = main->GetDevicePixelRatio();
```
PatTheMav, line 55: Please check if this function is only called in the main thread - if it is called from the GPU thread, you need to use `OBSBasic::GetDevicePixelRatio()` instead. Calling UI functions off the UI thread is no bueno.

taxilian: The only time I call this function is in response to a mousewheel event, so I would expect that it should only be called on the UI thread -- this is a direct copy of a similar function that just uses a different event type. I would assume if the other one is safe then this one should be safe as well. Do you have any reason to expect otherwise?

gxalpha:
> If there is an IRC room, discord channel, or similar that I could use to ask questions to speed things up that would help

Feel free to join https://obsproject.com/discord. The #development channel is what you're looking for, feel free to ask anything :)

taxilian:
... 

> > If there is an IRC room, discord channel, or similar that I could use to ask questions to speed things up that would help
> 
> There is https://obsproject.com/discord. The #development channel is what you're looking for, feel free to ask anything :)

... and this is the moment where I realize that I'm actually already *on* that server and simply forgot about it. *smh*.  Some days....

Warchamp7:

I'll also just note I have an open PR making other adjustments to cropping that includes similar alt drag behaviour, but I do not have a timeline for it.

PR #7267, opened 2022-09-02
UI: Refine YouTube dialog

cg2121:
### Description
- Make the settings a scroll area. This makes it work on smaller screens and if more items are added in the future, the dialog doesn't get bigger.
- Made the additional settings label bold to distinguish better. Also removed the colon, as I felt it wasn't needed.
- Made the margins consistent between the settings and events pages
- Made the scrollbar show only when needed in the events page
- Moved the remember settings checkbox to the top of the settings because it just seems better to be there.

Before:
Screenshot from 2022-09-02 01-36-33

After:
Screenshot from 2022-09-02 01-32-54

### Motivation and Context
Make UI look better and work on smaller screens.

### How Has This Been Tested?
Opened manage broadcast dialog to see how it looked.

### Types of changes
- Tweak (non-breaking change to improve existing functionality)

### Checklist:
- [x] My code has been run through [clang-format](https://github.com/obsproject/obs-studio/blob/master/.clang-format).
- [x] I have read the [**contributing** document](https://github.com/obsproject/obs-studio/blob/master/CONTRIBUTING.rst).
- [x] My code is not on the master branch.
- [x] The code has been tested.
- [x] All commit messages are properly formatted and commits squashed where appropriate.
- [x] I have included updates to all appropriate documentation.

PR #2751, opened 2020-04-18
UI: Remove exit button from controls dock

cg2121:
### Description
Remove exit button from controls dock.

Screenshot from 2020-04-23 22-24-36

### Motivation and Context

- This button is redundant because these are also in the file menu
- This button made the UI feel cluttered
- Users might accidentally click the exit button

### How Has This Been Tested?
Ran OBS as normal.

### Types of changes
- UI tweak

### Checklist:
- [x] My code has been run through [clang-format](https://github.com/obsproject/obs-studio/blob/master/.clang-format).
- [x] I have read the [**contributing** document](https://github.com/obsproject/obs-studio/blob/master/CONTRIBUTING.rst).
- [x] My code is not on the master branch.
- [x] The code has been tested.
- [x] All commit messages are properly formatted and commits squashed where appropriate.

PR #7421, opened 2022-09-20
UI: Remove exit button from control dock

skrueger-ftc:
### Description
Removes the exit button from the control dock

Current:
wihtExitButton

Without exit button:
withoutExitButton

### Motivation and Context
With the release of the Yami theme, the padding around the UI increased significantly causing the preview to be squashed.
Next to the YouTube broadcast button issue (fixed in #7267), the most common complaint is the preview is too small. Look at the screenshots posted by these users and the control dock is almost always the culprit. By removing the unnecessary exit button, the minimum control dock size shrinks and allows users to have a bigger preview. 

Based on the feedback around #2751, many see the button as unnecessary. Users can still exit from within the app via file > exit, which is still accessible via keyboard. 

### How Has This Been Tested?
Tested on Windows 10 21H2, launched obs and used the file > exit button to close

### Types of changes

### Checklist:


Vainock:
Please also remove the string 'Exit' in UI/data/locale/en-US.ini as it is hereby no longer used.

tytan652:
Enable the virtual camera and the replay buffer, the "issue" still there.

SuslikV:
Just in case, if some users like to make custom themes that inherits the Yami and want to hide/remove some Control buttons, they may try suggestions discussed on the forum:
https://obsproject.com/forum/threads/recording-button.83929/post-572914
(only modification of the .qss theme file required in this case, so you can experiment right now without waiting when the changes will get its final shape)

lextra2:
Hide the scene transition tab while you're at it. The tab name doesn't even fit into the window in your screenshot, plus I'd argue it is the least used tab from the bottom.

gxalpha:
> Hide the scene transition tab while you're at it. The tab name doesn't even fit into the window in your screenshot, plus I'd argue it is the least used tab from the bottom.

No, please don't. Please keep PR scope limited.
@lextra2 you can move the dock somewhere else or close it via the "Docks" menu.

lextra2:
> > Hide the scene transition tab while you're at it. The tab name doesn't even fit into the window in your screenshot, plus I'd argue it is the least used tab from the bottom.
> 
> No, please don't. Please keep PR scope limited. @lextra2 you can resize the dock or move it somewhere else or close it via the "Docks" menu.

I know I can hide it myself. But the point stands.

skrueger-ftc:
> Enable the virtual camera and the replay buffer and the "issue" still there.

That was worded poorly on my part. I didn't intend or mean to present this as a solution to the "issue". Personally, I actually prefer the larger padding of Yami. For many users on smaller displays though, any extra reduction in the minimum height of the bottom docks would be potentially beneficial. 
This was fix was more fueled by my dislike of the button with the potential added benefit for some users. It also helps as a stepping stone towards a redesigned output panel like that of obsproject/rfcs#12 and #6421

skrueger-ftc:
> Please also remove the string 'Exit' in UI/data/locale/en-US.ini as it is hereby no longer used.

Will do, I was unsure if it was used elsewhere and figured it would be safer not to.

RytoEX:
Thank you for this PR, but we'd rather handle this in a separate effort.

PR #7427, opened 2022-09-21
docs: Add missing sceneitem functions

CodeYan01:
### Description

Adds functions related to selected scene items and show/hide transitions of sceneitems.
I did see that obs.h says that the `...selected` and `...locked` functions are supposed
to be replaced, but `...locked` is documented.

I had not added docs for other related functions yet, since I have yet to
figure out what they do (like `obs_sceneitem_transition_load`, `obs_sceneitem_save`,
and `obs_sceneitem_force_update_transform`).

### Motivation and Context

Missing documentation for some functions.

### How Has This Been Tested?

Built the docs and viewed the html file in the browser (with a webserver).

### Types of changes

### Checklist:

### DIFF
--- a/docs/sphinx/reference-scenes.rst
+++ b/docs/sphinx/reference-scenes.rst
@@ -482,6 +482,15 @@ Scene Item Functions
 
 ---------------------
 
+.. function:: void obs_sceneitem_select(obs_sceneitem_t *item, bool select)
+              bool obs_sceneitem_selected(const obs_sceneitem_t *item)
+
+   Sets/gets the selection state of the scene item. Note that toggling
+   the selected state will not affect the selected state of other scene items,
+   as multiple scene items can be selected.
+
+---------------------
+
 .. function:: bool obs_sceneitem_set_visible(obs_sceneitem_t *item, bool visible)
               bool obs_sceneitem_visible(const obs_sceneitem_t *item)
 
@@ -562,10 +571,33 @@ Scene Item Functions
 
 ---------------------
 
+.. function:: void obs_sceneitem_set_transition(obs_sceneitem_t *item, bool show, obs_source_t *transition)
+
+   Sets a transition for showing or hiding a scene item.
+
+   :param item:       The target scene item
+   :param show:       If *true*, this will set the show transition.
+                      If *false*, this will set the hide transition.
+   :param transition: The transition to set. Pass *NULL* to remove the transition.
+
+---------------------
+
 .. function:: void obs_sceneitem_set_show_transition(obs_sceneitem_t *item, obs_source_t *transition)
               void obs_sceneitem_set_hide_transition(obs_sceneitem_t *item, obs_source_t *transition)
 
-   Set a transition for showing or hiding a scene item. Set *NULL* to remove the transition.
+   Sets a transition for showing or hiding a scene item. Set *NULL* to remove the transition.
+
+.. deprecated:: 27.2.4
+   Use :c:func:`obs_sceneitem_set_transition()` instead.
+
+---------------------
+
+.. function:: obs_source_t *obs_sceneitem_get_transition(obs_sceneitem_t *item, bool show)
+
+   :param item: The target scene item
+   :param show: If *true*, this will return the show transition.
+                If *false*, this will return the hide transition.
+   :return:     The transition for showing or hiding a scene item. *NULL* if no transition is set.
 
 ---------------------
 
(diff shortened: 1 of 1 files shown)


CodeYan01:
Do i need to run clang-format with this kind of PR?

RytoEX:
> Do i need to run clang-format with this kind of PR?

No. If nothing covered by clang-format has changed (e.g., C/C++ files), then clang-format does not need to be run.

CodeYan01:
thanks

CodeYan01:
In the header, it seems like `obs_sceneitem_set_show_transition` and `obs_sceneitem_set_show_transition_duration` are not marked deprecated, but the others are, which seems like an oversight. Do I still consider them all the transition functions deprecated?

As for showing they are deprecated, all the existing deprecation in the docs are done only with one function, unlike this where I combined the new function with the old functions (since I didn't know beforehand). What do you suggest I do? Separate the new functions from the old functions, then put a deprecation warning on the old ones?

docs/sphinx/reference-scenes.rst
```diff
@@ -472,6 +472,15 @@ Scene Item Functions
 
 ---------------------
 
+.. function:: void obs_sceneitem_select(obs_sceneitem_t *item, bool select)
+              bool obs_sceneitem_selected(const obs_sceneitem_t *item)
+
+   Sets/gets the selected/deselected state of the scene item. Note that toggling
+   the selected state will not affect the selected state of other scene items,
+   so multiple scene items can be selected.
```
WizardCM, line 480:
```suggestion
   as multiple scene items can be selected.
```

docs/sphinx/reference-scenes.rst
```diff
@@ -472,6 +472,15 @@ Scene Item Functions
 
 ---------------------
 
+.. function:: void obs_sceneitem_select(obs_sceneitem_t *item, bool select)
+              bool obs_sceneitem_selected(const obs_sceneitem_t *item)
+
+   Sets/gets the selected/deselected state of the scene item. Note that toggling
```
WizardCM, line 478:
```suggestion
   Sets/gets the selection state of the scene item. Note that toggling
```

WizardCM (changes requested):

gxalpha (approved):
I'm not completely happy with the `; hide transition otherwise` in a few places but can't think of something better, and at some point we should just merge these.

CodeYan01:
The changes requested by @WizardCM are already applied.

WizardCM (approved):

docs/sphinx/reference-scenes.rst
```diff
@@ -554,10 +563,31 @@ Scene Item Functions
 
 ---------------------
 
+.. function:: void obs_sceneitem_set_transition(obs_sceneitem_t *item, bool show, obs_source_t *transition)
+
+   Sets a transition for showing or hiding a scene item.
+
+   :param item:       The target scene item
+   :param show:       If *true*, this will set the show transition; hide transition otherwise
```
RytoEX, line 571:
```suggestion
   :param show:       If *true*, this will set the show transition.
                      If *false*, this will set the hide transition.
```

docs/sphinx/reference-scenes.rst
```diff
@@ -554,10 +563,31 @@ Scene Item Functions
...
+   :param transition: The transition to set; *NULL* to remove the transition
```
RytoEX, line 572:
```suggestion
   :param transition: The transition to set. Pass *NULL* to remove the transition.
```

docs/sphinx/reference-scenes.rst
```diff
@@ -554,10 +563,31 @@ Scene Item Functions
...
+
+.. deprecated:: 27.2.4
+   Use :c:func:`obs_sceneitem_set_transition()` instead.
+
+---------------------
+
+.. function:: obs_source_t *obs_sceneitem_get_transition(obs_sceneitem_t *item, bool show)
+
+   :param item: The target scene item
+   :param show: If *true*, this returns the show transition; hide transition otherwise
```
RytoEX, line 589:
```suggestion
   :param show: If *true*, this will set the show transition.
                If *false*, this will set the hide transition.
```

docs/sphinx/reference-scenes.rst
```diff
@@ -566,12 +596,40 @@ Scene Item Functions
...
+
+---------------------
+
+.. function:: void obs_sceneitem_set_transition_duration(obs_sceneitem_t *item, bool show, uint32_t duration_ms)
+
+   Sets the transition duration for showing or hiding a scene item.
+
+   :param item:        The target scene item
+   :param show:        If *true*, this will set the duration of the show transition;
+                       hide transition otherwise
```
RytoEX, line 610:
```suggestion
   :param show:        If *true*, this will set the duration of the show transition.
                       If *false*, this will set the duration of the hide transition.
```

docs/sphinx/reference-scenes.rst
```diff
@@ -566,12 +596,40 @@ Scene Item Functions
...
+
+---------------------
+
+.. function:: uint32_t obs_sceneitem_get_transition_duration(obs_sceneitem_t *item, bool show)
+
+   Gets the transition duration for showing or hiding a scene item.
+
+   :param item: The target scene item
+   :param show: If *true*, this will return the duration of the show transition;
+                hide transition otherwise
```
RytoEX, line 631:
```suggestion
   :param show: If *true*, this will return the duration of the show transition.
                If *false*, this will return the duration of the hide transition.
```

RytoEX (changes requested):
Circling back around to this and trying to address @gxalpha 's earlier concerns about the wording. Let me know if you disagree with these suggestions.

CodeYan01:
Currently looks like:
image

Suggested change looks like:
image

My opinion is that the `; ... otherwise` is fine, especially considering that it's just a bool, it's either show or hide. I feel like the additional words do not contribute much, other than for readability reasons (which is arguable, since there are more words for saying the same thing that can't be misunderstood imo). That said, I have no strong feelings on the matter, and I'm fine either way, so just say the word if the suggested changes should be applied. 

For some of them, like 
image
I actually like the slight change.

I also noticed an oversight of mine that you pointed out
image
where `returns` is wrong, and that will definitely be added.

Again, just say the word if the suggested changes on the `; ... otherwise` parts are better, I'm fine with it.

RytoEX:
> since there are more words for saying the same thing that can't be misunderstood imo)

I do not think that semi-colons are universally understood, and I do not think the implied words hidden by the use of "otherwise" are universally understood. In technical documentation, I prefer explicit text over implicit text. Would welcome a second opinion from @gxalpha or @WizardCM since they are usually involved on documentation PRs.

gxalpha:
I agree the suggested wording looks nicer than the current one

CodeYan01:
Applied all suggested changes, but changed one to this:
image
the suggested was "will set", i changed it to "will return"

CodeYan01:
Are there still any changes necessary?

CodeYan01:
Any chance we could get this merged before public release of 30?

RytoEX (approved):
> @RytoEX any remaining change requests for this?

Seems fine.

PR #7430, opened 2022-09-22
Add NVENC AV1 support

jp9000:
### Description
Adds NVENC AV1 encoding support

### Motivation and Context
AV1 is super sweet and we need to encourage the industry to move forward on newer, better codecs

### How Has This Been Tested?
On a test machine with applicable hardware

### Types of changes

### Checklist:


UI/window-basic-main-profiles.cpp
```diff
@@ -862,6 +862,12 @@ void OBSBasic::CheckForSimpleModeX264Fallback()
 				name = SIMPLE_ENCODER_X264;
 				return false;
 			}
+		} else if (strcmp(name, SIMPLE_ENCODER_NVENC_AV1) == 0) {
+			if (!nve_supported) {
+				changed = true;
+				name = SIMPLE_ENCODER_X264;
```
Conan-Kudo, line 868: This should fallback to rav1e, svt, or aom here, shouldn't it? Otherwise this makes no sense.

jp9000: You know, I think I'd rather make it just explicitly fail than fall back to a software AV1 encoder. Software AV1 encoders aren't really usable for most people due to their core requirements

Conan-Kudo: Are we sure about that? Realtime AOM modes seemed to work reasonably okay on my PCs from 2014 (4th generation Intel) and 2016 (6th generation Intel).

jp9000: If it did, it probably didn't compress very well.

Conan-Kudo: But does that matter when you're streaming to the other end? Realtime encoding trades off compression for quality and speed.

jp9000: I suppose I'm fine with falling back to AOM or something.

Conan-Kudo: That would be great, I know I would certainly appreciate it!

RytoEX: I don't think falling back from HW encoder to SW encoder is a good idea. Just select the SW encoder if you want to use that. Error out with a helpful message if we can't use the HW encoder.

Fenrirthviti: Also confirming falling back from a hardware to a software encoder is a bad idea. We don't fall back to x264 for any of the other hardware h264 encoders, and I don't see a reason why we should start doing that for those or AV1.

UI/window-basic-main-outputs.cpp
```diff
@@ -290,9 +290,7 @@ struct SimpleOutput : BasicOutputHandler {
 	void UpdateRecordingSettings_x264_crf(int crf);
 	void UpdateRecordingSettings_qsv11(int crf);
 	void UpdateRecordingSettings_nvenc(int cqp);
-#ifdef ENABLE_HEVC
```
Conan-Kudo, line 293: If you're going to remove these guards, can you do it in a separate commit?

UI/window-basic-main-profiles.cpp
Conan-Kudo: I'm fine with no fallback at all too. Though I'd love to eventually see an AV1 alternative to the x264 plugin.

jp9000: We already have AOM and SVT.

UI/window-basic-main-outputs.cpp
tytan652: Those guards are removed because `UpdateRecordingSettings_nvenc_hevc` is used for AV1.

Maybe `UpdateRecordingSettings_nvenc_hevc` needs to be renamed since it is no longer HEVC only.

Conan-Kudo: Yeah, that would be a good idea.

UI/window-basic-main-profiles.cpp
jp9000: This code doesn't have anything to do with silently falling back. This code is for when you start up obs and the encoder no longer exists with the current config anymore (i.e. switched parts, copied config from another computer).

UI/window-basic-main-outputs.cpp
jp9000: I changed the name of the function to be more clear.

PR #7433, opened 2022-09-22
docs: Add property functions

CodeYan01:
### Description

Add undocumented property functions.

`obs_property_text_monospace` only having an effect on multiline
text properties were something I noticed by testing myself, so
correct me if I'm wrong.

### Motivation and Context

Missing documentation for functions

### How Has This Been Tested?

Built and viewed locally. 

### Types of changes

### Checklist:

### DIFF
--- a/docs/sphinx/reference-properties.rst
+++ b/docs/sphinx/reference-properties.rst
@@ -75,6 +75,18 @@ General Functions
 
 ---------------------
 
+.. function:: void obs_properties_remove_by_name(obs_properties_t *props, const char *property)
+
+   Removes a property from a properties list. Only valid in ``get_properties``,
+   ``script_properties`` for scripts, ``modified_callback``, and ``modified_callback2``.
+   ``modified_callback`` and ``modified_callback2`` *must* return true so that
+   all UI properties are rebuilt. Returning false is undefined behavior.
+
+   :param props:    Properties to remove from.
+   :param property: Name of the property to remove.
+
+---------------------
+
 
 Property Object Functions
 -------------------------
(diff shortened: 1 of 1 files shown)


RytoEX (approved):
Some odd line breaks, but otherwise seems fine. @gxalpha ? @WizardCM ?

CodeYan01:
> Some odd line breaks
like which ones? If you are referring to line breaks like this:
image
it should not have an effect on the actual output, but I'm open to moving the line breaks to where they should be.

RytoEX:
> > Some odd line breaks
> 
> like which ones? If you are referring to line breaks like this: image it should not have an effect on the actual output, but I'm open to moving the line breaks to where they should be.

Yes. I was just wondering why they are line wrapped at something lower then 80 or 100 characters (whatever is commonly used in the documentation). It seemed like an odd choice when there should already be a specified column limit.

CodeYan01:
I sorta fixed the line breaks. There isn't really any established column limit in the docs that I'm aware of, and different documentation files have different limits, especially considering the function declarations that aren't even wrapped.

PatTheMav:
@RytoEX probably good to merge, haven't seen any objections from others yet.

RytoEX:
> Add undocumented property functions. In the `obs_properties_add_button`, I replaced `description` with `text` because there's no such param.

This change doesn't seem to be present here anymore.

> Please check the description I put for `priv`, since I can't test it myself since it's not accessible in python scripting.

This change doesn't seem to be here either.

> `obs_property_text_monospace` only having an effect on multiline text properties were something I noticed by testing myself, so correct me if I'm wrong.

The commit adding `obs_property_text_monospace` says that it was added for multiline text and the `if` block where it is checked is within an `if` for `type == OBS_TEXT_MULTILINE`, so this seems correct.
https://github.com/obsproject/obs-studio/commit/62504dc651c75996890cdb1c3a43bc99cafa5768

CodeYan01:
PR description updated, commit description is just the title so there's nothing to update. 

Those changes are indeed gone now, as another PR already fixed those.

Thanks for confirming the info about multiline.

RytoEX (approved):
Looks good to me.

PR #7436, opened 2022-09-22
UI: Apply OBSDock themes to all QDockWidgets

glikely:
### Description
The OBSDock theme settings are equally applicable to other QDockWidgets in the application. For example, plugins implementing a QDockWidget should have the same theme. This patch changes OBSDock specific qss stanzas to match QDockWidget

### How Has This Been Tested?
Tested on macOS 12 and Ubuntu by visual inspection that existing docks look correct, and that it fixes the theme problems observed with the obs-ptz plugin.

Attached are before and after screenshots showing the change. As can be seen, the stock dock widgets are unchanged, but the obs-ptz window is better matched to the OBS docks.

Screenshot 2022-09-22 at 14 44 40
Screenshot 2022-09-22 at 14 45 49
Screenshot 2022-09-22 at 14 45 15
Screenshot 2022-09-22 at 14 46 07

### Types of changes

### Checklist:

### DIFF
--- a/UI/data/themes/Acri.qss
+++ b/UI/data/themes/Acri.qss
@@ -270,13 +270,13 @@ OBSBasicSettings #PropertiesContainer {
 }
 
 /* Dock Widget */
-OBSDock > QWidget {
+QDockWidget > QWidget {
     background: palette(dark);
     border-bottom-left-radius: 4px;
     border-bottom-right-radius: 4px;
 }
 
-OBSDock QFrame {
+QDockWidget QFrame {
     background: palette(dark);
     border-bottom-left-radius: 4px;
     border-bottom-right-radius: 4px;
@@ -287,12 +287,12 @@ OBSDock QFrame {
     padding: 4px 6px;
 }
 
-OBSDock QLabel {
+QDockWidget QLabel {
     background: transparent;
 }
 
-OBSDock QComboBox,
-OBSDock QPushButton {
+QDockWidget QComboBox,
+QDockWidget QPushButton {
     margin: 1px 2px;
 }
 
--- a/UI/data/themes/Grey.qss
+++ b/UI/data/themes/Grey.qss
@@ -270,13 +270,13 @@ OBSBasicSettings #PropertiesContainer {
 }
 
 /* Dock Widget */
-OBSDock > QWidget {
+QDockWidget > QWidget {
     background: palette(dark);
     border-bottom-left-radius: 4px;
     border-bottom-right-radius: 4px;
 }
 
-OBSDock QFrame {
+QDockWidget QFrame {
     background: palette(dark);
     border-bottom-left-radius: 4px;
     border-bottom-right-radius: 4px;
@@ -287,12 +287,12 @@ OBSDock QFrame {
     padding: 4px 6px;
 }
 
-OBSDock QLabel {
+QDockWidget QLabel {
     background: transparent;
 }
 
-OBSDock QComboBox,
-OBSDock QPushButton {
+QDockWidget QComboBox,
+QDockWidget QPushButton {
     margin: 1px 2px;
 }
 
--- a/UI/data/themes/Light.qss
+++ b/UI/data/themes/Light.qss
@@ -270,13 +270,13 @@ OBSBasicSettings #PropertiesContainer {
 }
 
 /* Dock Widget */
-OBSDock > QWidget {
+QDockWidget > QWidget {
     background: palette(dark);
     border-bottom-left-radius: 4px;
     border-bottom-right-radius: 4px;
 }
 
-OBSDock QFrame {
+QDockWidget QFrame {
     background: palette(dark);
     border-bottom-left-radius: 4px;
     border-bottom-right-radius: 4px;
@@ -287,12 +287,12 @@ OBSDock QFrame {
     padding: 4px 6px;
 }
 
-OBSDock QLabel {
+QDockWidget QLabel {
     background: transparent;
 }
 
-OBSDock QComboBox,
-OBSDock QPushButton {
+QDockWidget QComboBox,
+QDockWidget QPushButton {
     margin: 1px 2px;
 }
 
--- a/UI/data/themes/Rachni.qss
+++ b/UI/data/themes/Rachni.qss
@@ -272,13 +272,13 @@ OBSBasicSettings #PropertiesContainer {
 }
 
 /* Dock Widget */
-OBSDock > QWidget {
+QDockWidget > QWidget {
     background: palette(dark);
     border-bottom-left-radius: 4px;
     border-bottom-right-radius: 4px;
 }
 
-OBSDock QFrame {
+QDockWidget QFrame {
     background: palette(dark);
     border-bottom-left-radius: 4px;
     border-bottom-right-radius: 4px;
@@ -289,12 +289,12 @@ OBSDock QFrame {
     padding: 4px 6px;
 }
 
-OBSDock QLabel {
+QDockWidget QLabel {
     background: transparent;
 }
 
-OBSDock QComboBox,
-OBSDock QPushButton {
+QDockWidget QComboBox,
+QDockWidget QPushButton {
     margin: 1px 2px;
 }
 
--- a/UI/data/themes/Yami.qss
+++ b/UI/data/themes/Yami.qss
@@ -274,13 +274,13 @@ OBSBasicSettings #PropertiesContainer {
 }
 
 /* Dock Widget */
-OBSDock > QWidget {
+QDockWidget > QWidget {
     background: palette(dark);
     border-bottom-left-radius: 4px;
     border-bottom-right-radius: 4px;
 }
 
-OBSDock QFrame {
+QDockWidget QFrame {
     background: palette(dark);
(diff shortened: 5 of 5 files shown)


tytan652:
Please name the commit correctly by following the [guidelines](https://github.com/obsproject/obs-studio/blob/master/CONTRIBUTING.rst#commit-guidelines).
TLDR: Add the `UI:` prefix to the commit name.

glikely:
> Please name the commit correctly by following the [guidelines](https://github.com/obsproject/obs-studio/blob/master/CONTRIBUTING.rst#commit-guidelines). TLDR: Add the `UI:` prefix to the commit name.

Done, and thanks for the feedback.

GeorgesStavracas:
Does the stats dock look okay with this PR? Could you please share a screenshot of it?

glikely:
Here are screenshots before and after of the stats window using Yami theme.
Screenshot 2022-09-28 at 15 26 42
Screenshot 2022-09-28 at 15 27 51

Here is the same using the Light theme
Screenshot 2022-09-28 at 15 29 50
Screenshot 2022-09-28 at 15 30 20

PatTheMav:
I am reading it right that the top screenshot is the "before" state?

Because looking at the PTZ controls it seems to me that from a UX/UI point of view the "before" state is actually correct, with selection widgets being visually distinct from their labels.

In the "fixed" state the entire control becomes visually indistinct and harder to understand (as there is no visual element differentiating a section/control title from the actual control).

@GeorgesStavracas what do you think?

glikely:
> I am reading it right that the top screenshot is the "before" state?
> 
> Because looking at the PTZ controls it seems to me that from a UX/UI point of view the "before" state is actually correct, with selection widgets being visually distinct from their labels.
> 
> In the "fixed" state the entire control becomes visually indistinct and harder to understand (as there is no visual element differentiating a section/control title from the actual control).

Look along the bottom of the PTZ panel. In the before shot, the PTZ panel doesn't get the rounded corners, and it has a different background colour. In the 'after' shot it looks correct.

I wouldn't worry about the look of the PTZ list controls themselves. The UI design on them isn't very good and I'll be changing them in the next release of the plugin.

glikely:
Any more feedback on this change? Will this be acceptable, or do I need to come up with a way to apply themes within the plugin itself?

Warchamp7:
Hi @glikely, apologies for the delay in reviewing this.

As Pat mentioned, the qss was written to target `OBSDock` specifically because OBS docks follow a particular widget hierarchy. The intent was that plugins would need to opt-in to that specialized styling and then structure their widgets accordingly. As demonstrated in the Before screenshots, this means other docks by default retain a visual hierarchy for things like lists properly.

However, I mistakenly believed that plugins could use the OBSDock class to opt-in to that. This is not the case as it is a private type. We may look into fixing that so there is a method they can get the appropriate styling, such as via a custom property.


With all that said: if your primary goal is for the PTZ QPushButtons to have the squarer, more even style of toolbar buttons, that IS available via the custom properties. Our usage of custom properties in this way needs more fleshing out and documenting, but it is available.

Example in [code](https://github.com/obsproject/obs-studio/blob/e438037960b65d491a5ae4f0bd687b8ffb203e95/UI/properties-view.cpp#L640) and [in a UI form](https://github.com/obsproject/obs-studio/blob/14bd880822ef4162f950eada92bbe8afa967dec2/UI/forms/OBSBasic.ui#L1353-L1355) and [how the QSS gets applied](https://github.com/obsproject/obs-studio/blob/0df774cd01c08d1e813103faf6fe6fae20a68597/UI/data/themes/Yami.qss#L460-L466)

glikely:
> Hi @glikely, apologies for the delay in reviewing this.
> 
> As Pat mentioned, that particular qss was written to target `OBSDock` specifically because OBS docks follow a particular widget hierarchy. The intent was that plugins would need to opt-in to that specialized styling and then structure their widgets accordingly. As demonstrated in the Before screenshots, this means other docks by default retain a visual hierarchy for things like lists properly.
> 
> However, I mistakenly believed that plugins could use the OBSDock class to opt-in to that. This is not the case as it is a private type. We may look into fixing that so there is a way they can get the OBSDock specific styling, such as via a custom property. I'll need to add consideration for QListView as well.

Thanks for replying.

I did a quick experiment with using `* [themeID="obsDock"]` in place of `OBSDock` in the theme files, and it did the job for me. The main thing I want is to follow the theme color scheme and margins so that the dock looks consistent, and so that the list items are visibly highlighted. On some of the themes, if the theme colours aren't applied then the currently selected list item isn't highlighted.

Would adding themeIDs for the OBSDock and QList{Widget,View} work for you?

> With all that said: if your primary goal is for the PTZ QPushButtons to have the squarer, more even style of toolbar buttons, that IS available via the custom properties. Our usage of custom properties in this way needs more fleshing out and documenting, but it is available.

That is helpful, thanks! I'll make use of that.

> Example in [code](https://github.com/obsproject/obs-studio/blob/e438037960b65d491a5ae4f0bd687b8ffb203e95/UI/properties-view.cpp#L640) and [in a UI form](https://github.com/obsproject/obs-studio/blob/14bd880822ef4162f950eada92bbe8afa967dec2/UI/forms/OBSBasic.ui#L1353-L1355) and [how the QSS gets applied](https://github.com/obsproject/obs-studio/blob/0df774cd01c08d1e813103faf6fe6fae20a68597/UI/data/themes/Yami.qss#L460-L466)

glikely:
> > Hi @glikely, apologies for the delay in reviewing this.
> > With all that said: if your primary goal is for the PTZ QPushButtons to have the squarer, more even style of toolbar buttons, that IS available via the custom properties. Our usage of custom properties in this way needs more fleshing out and documenting, but it is available.
> 
> That is helpful, thanks! I'll make use of that.

After a bit of experimentation, the toolButton property may not do what I need because unlike on a toolbar which has 0 margin above and below (on some themes), those buttons need to be evenly spaced on a grid. So, having a bit finer grained control of which bits of the theme are picked up on would be useful for me. (e.g., being able to pick up the OBSDock theme for the dock part of the window, but not necessarily having to pick up the QPushButton margin styling https://github.com/obsproject/obs-studio/blob/0df774cd01c08d1e813103faf6fe6fae20a68597/UI/data/themes/Yami.qss#L298-L301)

> 
> > Example in [code](https://github.com/obsproject/obs-studio/blob/e438037960b65d491a5ae4f0bd687b8ffb203e95/UI/properties-view.cpp#L640) and [in a UI form](https://github.com/obsproject/obs-studio/blob/14bd880822ef4162f950eada92bbe8afa967dec2/UI/forms/OBSBasic.ui#L1353-L1355) and [how the QSS gets applied](https://github.com/obsproject/obs-studio/blob/0df774cd01c08d1e813103faf6fe6fae20a68597/UI/data/themes/Yami.qss#L460-L466)

Warchamp7:
> > Hi @glikely, apologies for the delay in reviewing this.
> > As Pat mentioned, that particular qss was written to target `OBSDock` specifically because OBS docks follow a particular widget hierarchy. The intent was that plugins would need to opt-in to that specialized styling and then structure their widgets accordingly. As demonstrated in the Before screenshots, this means other docks by default retain a visual hierarchy for things like lists properly.
> > However, I mistakenly believed that plugins could use the OBSDock class to opt-in to that. This is not the case as it is a private type. We may look into fixing that so there is a way they can get the OBSDock specific styling, such as via a custom property. I'll need to add consideration for QListView as well.
> 
> Thanks for replying.
> 
> I did a quick experiment with using `* [themeID="obsDock"]` in place of `OBSDock` in the theme files, and it did the job for me. The main thing I want is to follow the theme color scheme and margins so that the dock looks consistent, and so that the list items are visibly highlighted. On some of the themes, if the theme colours aren't applied then the currently selected list item isn't highlighted.
> 
> Would adding themeIDs for the OBSDock and QList{Widget,View} work for you?
> 
> > With all that said: if your primary goal is for the PTZ QPushButtons to have the squarer, more even style of toolbar buttons, that IS available via the custom properties. Our usage of custom properties in this way needs more fleshing out and documenting, but it is available.
> 
> That is helpful, thanks! I'll make use of that.
> 
> > Example in [code](https://github.com/obsproject/obs-studio/blob/e438037960b65d491a5ae4f0bd687b8ffb203e95/UI/properties-view.cpp#L640) and [in a UI form](https://github.com/obsproject/obs-studio/blob/14bd880822ef4162f950eada92bbe8afa967dec2/UI/forms/OBSBasic.ui#L1353-L1355) and [how the QSS gets applied](https://github.com/obsproject/obs-studio/blob/0df774cd01c08d1e813103faf6fe6fae20a68597/UI/data/themes/Yami.qss#L460-L466)

I would like to go the route we did with toolButtons rather than themeID (themeID is the older, original way we tackled this, and restricts widgets to only a single style 'hint')

i.e. 
```OBSDock,
* [OBSDock="true"] {}

OBSDock QPushButton,
* [OBSDock="true"] QPushButton```
etc.

I may have to think harder about how to make the QLists look. The new themes palette was not designed for more than 3 levels of visual heirarchy. A QList surrounded by other widgets within a dock would create a 4th level. (This is something I need to address for #8305 as well)

You can use palette(button) as the background color in the interim and I will adjust it in the future if needed.

PR #7446, opened 2022-09-23
linux-capture: Fixup window name/class checking

### LINKED ISSUE #7404: Window Capture(Xcomposite) resets its target every time OBS launches
### Operating System Info

Other

### Other OS

Fedora 36 XFCE

### OBS Studio Version

28.0.1

28.0.1 64bit

### OBS Studio Log URL

https://obsproject.com/logs/bVsiDlwzGq8M0Eoh

### OBS Studio Crash Log URL

_No response_

### Expected Behavior

When I use Window Capture.   If I do not have the application open then it should be black.   If I do have the application open then it should show the application.   This is how it worked in the previous flatpak version of OBS.

### Current Behavior

Every time I launch OBS, it changes the window capture target in my scene to whatever options is at the to for new window captures.   It remembers the cropping but not the target application.  It did not use to do this.   I have to set the correct target every time I launch the application.

### Steps to Reproduce

1.  Launch OBS
2.  See that  a scene with Window capture is on the wrong target
3.  Launch the application
4.   See that  a scene with window capture is still on the wrong target.

Most of the time it defaults to my desktop.   Sometimes it defaults to Discord if I have Discord open.

kkartaltepe:
### Description

There were a couple mistakes here that caused the fallback checks other than window ID to fail to ever return valid results. This restores this functionality that was broken since the c++/c transition.

### Motivation and Context

fixes #7404

### How Has This Been Tested?

Tested that capturing obs works, closing obs, openign and closing some windows and then starting obs again regains capture. Sometimes X11 will reuse window IDs so opening and closing some windows in the interim helps avoid this.

### Types of changes

### Checklist:

### DIFF
--- a/plugins/linux-capture/xcomposite-input.c
+++ b/plugins/linux-capture/xcomposite-input.c
@@ -310,11 +310,10 @@ xcb_window_t xcomp_find_window(xcb_connection_t *conn, Display *disp,
 	char *wcls = bzalloc(strEnd - thirdStr + 1);
 	memcpy(wname, secondStr, secondMark - secondStr);
 	memcpy(wcls, thirdStr, strEnd - thirdStr);
-	xcb_window_t wid = (xcb_window_t)strtol(str, NULL, 10);
+	ret = (xcb_window_t)strtol(str, NULL, 10);
 
 	// first try to find a match by the window-id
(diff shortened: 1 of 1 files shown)


GeorgesStavracas (approved):

RytoEX:
This should be picked to the release/28.0 branch.

PR #7448, opened 2022-09-24
CI: Run PVS Studio analysis

derrod:
### Description

Adds PVS Studio analysis to CI on manual or scheduled runs.
(Pending confirmation from PVS-Studio that it's okay for us to run this on CI)

Currently this is Windows and Linux only.

This is using the PVS CMake module for direct integration and running the analysis when `ENABLE_PVS` is set.

### Motivation and Context

Static analysis good, bugs bad.

### How Has This Been Tested?

My fork with a trial license: https://github.com/derrod/obs-studio/actions/runs/3121413583

(Note: The results in the security tab require the necessary permission on the repo and are not visible to non-members.)

### Types of changes

### Checklist:


derrod:
Based on discussion on Discord I've split the analysis off into its own workflow. The Windows version will still require some work and figuring out why the Sarif produced is invalid.

Depending on how things go we may also consider using a different static analyser (CodeQL, clang, etc.) in addition or as a replacement for PVS should it continue to fail us on Windows.

derrod:
So, I ended up trying to get it up and running with the CMake module they provide. It wasn't fun (~~if you can't tell by the commit titles~~ rebased now, so you can't relive my pain).

The Windows build however still does not actually work entirely. I'd appreciate some help in debugging this if anybody has some time: https://github.com/derrod/obs-studio/actions/runs/3120773820/jobs/5061614867

derrod:
Seems like in order for PVS to work properly we manually have to add the `/Zc:__cplusplus /std:c++17` compiler flags. `pvs_studio_add_target` supports adding compiler flags so this could potentially be done via a variable there rather than setting it for OBS entirely.

Here's a successful run with the flags added globally: https://github.com/derrod/obs-studio/actions/runs/3120908618/jobs/5061856858

However, since it's only set to the `obs` target it currently doesn't analyse any plugins, so gotta figure out what the best approach there is to add all (current) targets to that list.

Also not sure if Windows and Linux should use different category names to avoid one overwriting the other. The documentation suggests yes, but it also says a run should fail if you try to upload it twice, and that evidently didn't happen...

derrod:
It only took several hours and my sanity but I finally did it: https://github.com/derrod/obs-studio/actions/runs/3121413583

The exclusion rules are of course not working *for some reason* but that's not a huge issue, we can just dismiss those reports and ignore them.

The deduplication in `PVS-Studio.cmake` also isn't perfect since we use a bunch of relative paths. Perhaps normalising them will work. So far it doesn't seem like an issue though (GitHub might deduplicate those anyway).

cmake/Modules/ObsHelpers.cmake
```diff
@@ -546,3 +546,28 @@ function(_install_obs_datatarget target destination)
     COMMENT "Installing ${target} to OBS rundir"
     VERBATIM)
 endfunction()
+
+# Helper function to get all added build targets
```
PatTheMav, line 550: That's not necessary - the `obs` target has all sub-targets added as dependencies (otherwise you couldn't just build that target and get a full OBS installation), so you just need to get the targets recursively from that one, without the need for file-system shenanigans.

So in essence you'd just need to add the `obs` target for a full check, alas we need to filter out the 3rd party ones.

PVS has this already implemented as `pvs_studio_get_recursive_targets`. Only requirement is that you'd need to set up PVS _after_ the OBS target has been fully created.

.github/workflows/analysis.yml
```diff
@@ -0,0 +1,157 @@
...
+        uses: actions/cache@v3
+        env:
+          CACHE_NAME: 'cef-cache'
+        with:
+          path: ${{ github.workspace }}/obs-build-dependencies/cef_binary_${{ env.CEF_BUILD_VERSION_WIN }}_windows_x64
+          key: ${{ runner.os }}-pr-${{ env.CACHE_NAME }}-${{ env.CEF_BUILD_VERSION_WIN }}-x64-${{ env.CACHE_REVISION }}
+
+      - name: 'Setup PVS Studio'
+        run: |
+          New-Item -Path ..\PVSTestResults -ItemType Directory
```
PatTheMav, line 132: Good enough for now - for the next revision of the build scripts I'd add this to the dependency resolver based on build configuration.

CI/analysis/pvs_config.cfg
```diff
@@ -0,0 +1,26 @@
...
+exclude-path=*/lzma/*
+exclude-path=*/w32-pthreads/*
+exclude-path=*/plugins/enc-amf/*
+exclude-path=*/plugins/obs-browser/*
+exclude-path=*/plugins/obs-websocket/*
+exclude-path=*/plugins/win-dshow/libdshowcapture/*
+exclude-path=*/plugins/obs-outputs/ftl-sdk/*
+exclude-path=*/obs-build-dependencies/*
+exclude-path=*/simde/*
+exclude-path = *\blake2\*
```
PatTheMav, line 14: Duplication necessary because Windows is still basically DOS? ;)

CMakeLists.txt
```diff
@@ -95,3 +95,63 @@ endif()
...
+    "obs-websocket"
+    "obs-x264-test"
+    "test_bitstream"
+    "test_darray"
+    "test_serializer"
+    "w32-pthreads")
+
+  # The formatting on this is terrible and cmake-format should feel bad about
+  # it.
+  pvs_studio_add_target(
```
PatTheMav, line 138: Let's add this to the cmake config so it formats correctly - I'll see if I can send you the change off-thread later.

cmake/Modules/PVS-Studio.cmake
```diff
@@ -0,0 +1,719 @@
+# 2006-2008 (c) Viva64.com Team 2008-2018 (c) OOO "Program Verification Systems"
```
PatTheMav, line 1: I accept this file as-is, we should probably exclude it from cmake-format somehow to make diffs easier to work with upon updates.

CI/analysis/pvs_config.cfg
derrod: Not entirely sure yet, the PVS documentation is not great and things like possible config file options are split across multiple pages and none of them ever spell out how exactly it needs to be formatted. And since a single run takes ~80 minutes I didn't have time to test this yet. Will do later though.

CMakeLists.txt
derrod: Added it to `.cmake-format.json` in f5f4c928d484f62620955b9d18778335a981a7c0

cmake/Modules/PVS-Studio.cmake
derrod: I moved it to `cmake/external` and excluded that folder from formatting in `CI/check-cmake.sh`. I also added a note on the modifications made from upstream for it to work correctly here.

cmake/Modules/ObsHelpers.cmake
derrod: This may be my ignorance in how CMake processes things, but when using `pvs_studio_get_recursive_targets` instead (at the very end of the main `CMakeLists.txt`) I only get the following (on Windows): `OOBS::libobs;Jansson::Jansson;OBS::caption;OBS::w32-pthreads;OBS::frontend-api;OBS::blake2` so plugins themselves appear to be not found that way.

derrod: Okay so by adding the following to `pvs_studio_get_recursive_targets`:
```diff
    get_target_property(deps "${TARGET}" MANUALLY_ADDED_DEPENDENCIES)
    list(APPEND libs ${deps})
```
I can get *almost* there. The following targets are still missing (when compared to the current method):
```
file-updater
libobs-winrt
media-playback
obs-amf-test
obs-virtualcam-module
opts-parser
```

CI/analysis/pvs_config.cfg
derrod: Seems to work with just unix-style paths as well. It's just odd that it isn't logged.

cmake/Modules/ObsHelpers.cmake
PatTheMav: Wouldn't that require us to modify their CMake script whenever they update it, possibly breaking our changes?

Also some targets are header-only libraries, whose source code will be added to a linked target, so there's nothing to "compile" for these targets (and hence check).

derrod: Right now the script is already slightly modified since by default it would try to add a custom command multiple times for shared dependencies, which causes a fatal cmake error. So that is already problem (perhaps Tracking the changes in a .patch file would be a good idea).

I'm not too worried about header only targets, since it should just don't do anything for those, but the ones currently missing aren't header-only and I'm not entirely sure why are they being omitted.

RytoEX (changes requested):
Nit.

derrod:
This PR needs to be basically rebuilt from the ground up due to CI and CMake changes, so I'll close it and then resubmit whenever I get around to it.

PR #7450, opened 2022-09-24
obs-scripting: Fix compile when python is not found

exeldro:
### Description
Fix compile when python is not found

### Motivation and Context
I want to be able to compile without python

### How Has This Been Tested?
On windows 64 bits compiling without python

### Types of changes

### Checklist:

### DIFF
--- a/deps/obs-scripting/obs-scripting.c
+++ b/deps/obs-scripting/obs-scripting.c
@@ -458,4 +458,9 @@ bool obs_scripting_python_runtime_linked(void)
 {
 	return (bool)true;
 }
+
+void obs_scripting_python_version(char *version, size_t version_length)
+{
+	version[0] = 0;
+}
 #endif


tytan652:
I think a better fix would be to guard some code in `frontend-tools` with a python related macros.

exeldro:
I think the solution should be be same as the functions `obs_scripting_load_python`, `obs_scripting_python_loaded` and `obs_scripting_python_runtime_linked`. All are exported so have an empty implementation when not `Python_FOUND`

PatTheMav:
This should be fine in theory, however this requires that the char array used is zero-initialised, because otherwise Qt will receive random data to build the final output string.

So `void ScriptsTool::updatePythonVersionLabel()` should initialise via `char version[8] = {0};`. If you could add this to `obs-frontend-api` this should work without side-effects.

PR #7455, opened 2022-09-24
obs-ffmpeg: Prevent media source restart

CodeYan01:
Some options do not require a media source restart. Saving source unchanged source settings should also not trigger a restart.

I tried to also make it so speed changes do not require a restart by manipulating the elapsed timestamp  for the source, but it sometimes causes the first few milliseconds after saving settings to be rough or slightly distorted. I am also unsure whether a/v frame buffers should be cleared on a speed change (and how to) so I made it so restarts are required for speed changes. Chiyo also mentioned that successive quick speed changes will cause maximum audio buffering (which I was able to reproduce once).

Nonlocal file playback is unaffected, and will cause the media source to restart because I am not aware of the possible consequences and because there's no restart button for it.

Changing "Apply alpha in linear space" is only visible after a few milliseconds (when mp_media_next_video is called for new frames).

### Motivation and Context

Idea: https://ideas.obsproject.com/posts/1622/do-not-restart-media-source-when-not-necessary
To be honest, that's mine. I know no one else voted on it, but I know people who wouldn't
want to restart the media source either.

The media source currently restarts every time the source is saved, even if it is unnecessary.
It restarts even when no setting is changed. It should not restart if I just want to toggle _Loop_,
_Restart playback when source becomes active_, _Close file when inactive_, and _Show nothing when playback ends_.
If a video only has transparency in a certain part of the video, you would have to toggle the _linear alpha_
option, then skip to that part again. In addition, workflows could have adding media sources directly
by dragging and dropping, and while it's playing out, I realize that I forgot to turn on _Show Nothing when playback ends_. 

This was my attempt at making speed changes not need media restarts. I tried it many times, and it was mostly fine, other than the fact the first few milliseconds after changing it makes audio a bit distorted. There's also probably some buffered data/frames that I would need to free or flush, but I have no idea about that, so I left it alone and decided that speed changes should require a restart.
image

Nonlocal file playback is left with original behavior of restarting all the time because of the lack of a restart button and my lack of knowledge on which properties can be updated without restarting the source.

### How Has This Been Tested?

Windows 10 Build 19043 64-bit on a Lenovo Thinkpad T570
Ran this in portable mode.
Used monitor and output for the media source.

Tests:
- Repeatedly saving settings without changes, ensuring that playback continues.
- Toggling the settings that should not cause a restart and ensuring playback continues.
- Tried to change input file path, speed, color range, hardware decoding, and ffmpeg options, along with combinations I can think of, and ensuring that the media source is restarted.
- Toggling "Local File" restarts the source (if it is actually changed upon saving).
- If local file is off, saving properties, even if unchanged, will restart the source.
- Tested that linear alpha change works using the test file Flaeri kindly provided 
[test.webm](https://user-images.githubusercontent.com/65320293/192115161-677fd8c0-800d-4f39-be3a-dfa30b9b105f.webm)
Note that linear alpha change does not get applied instantly, but it is only a matter of a few milliseconds, which is better than having it restart.
- Verified that toggling Restart Playback when source becomes active works without a restart by switching to a blank scene.

### Types of changes

### Checklist:

### DIFF
--- a/deps/media-playback/media-playback/media-playback.c
+++ b/deps/media-playback/media-playback/media-playback.c
@@ -85,6 +85,23 @@ void media_playback_stop(media_playback_t *mp)
 		mp_media_stop(&mp->media);
 }
 
+void media_playback_set_looping(media_playback_t *mp, bool looping)
+{
+	if (mp->is_cached)
+		mp->cache.looping = looping;
+	else
+		mp->media.looping = looping;
+}
+
+void media_playback_set_is_linear_alpha(media_playback_t *mp,
+					bool is_linear_alpha)
+{
+	if (mp->is_cached)
+		mp->cache.m.is_linear_alpha = is_linear_alpha;
+	else
+		mp->media.is_linear_alpha = is_linear_alpha;
+}
+
 void media_playback_preload_frame(media_playback_t *mp)
 {
 	if (!mp)
--- a/deps/media-playback/media-playback/media.c
+++ b/deps/media-playback/media-playback/media.c
@@ -483,7 +483,7 @@ void mp_media_next_video(mp_media_t *m, bool preload)
 	frame->height = f->height;
 	frame->max_luminance = d->max_luminance;
 	frame->flip = flip;
-	frame->flags |= m->is_linear_alpha ? OBS_SOURCE_FRAME_LINEAR_ALPHA : 0;
+	frame->flags = m->is_linear_alpha ? OBS_SOURCE_FRAME_LINEAR_ALPHA : 0;
 	switch (f->color_trc) {
 	case AVCOL_TRC_BT709:
 	case AVCOL_TRC_GAMMA22:
--- a/plugins/obs-ffmpeg/obs-ffmpeg-source.c
+++ b/plugins/obs-ffmpeg/obs-ffmpeg-source.c
@@ -435,22 +435,33 @@ static void ffmpeg_source_update(void *data, obs_data_t *settings)
 {
 	struct ffmpeg_source *s = data;
 
+	bool active = obs_source_active(s->source);
 	bool is_local_file = obs_data_get_bool(settings, "is_local_file");
 	bool is_stinger = obs_data_get_bool(settings, "is_stinger");
+	bool should_restart_media = (is_local_file != s->is_local_file) ||
+				    (is_stinger != s->is_stinger);
 
 	const char *input;
 	const char *input_format;
 	const char *ffmpeg_options;
 
-	bfree(s->input);
+	bool is_hw_decoding;
+	enum video_range_type range;
+	bool is_linear_alpha;
+	int speed_percent;
+	bool is_looping;
+
 	bfree(s->input_format);
-	bfree(s->ffmpeg_options);
 
 	if (is_local_file) {
 		input = obs_data_get_string(settings, "local_file");
 		input_format = NULL;
-		s->is_looping = obs_data_get_bool(settings, "looping");
+		is_looping = obs_data_get_bool(settings, "looping");
+
+		if (s->input && !should_restart_media)
+			should_restart_media |= strcmp(s->input, input) != 0;
 	} else {
+		should_restart_media = true;
 		input = obs_data_get_string(settings, "input");
 		input_format = obs_data_get_string(settings, "input_format");
 		if (requires_mpegts(input)) {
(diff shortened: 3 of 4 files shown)


plugins/obs-ffmpeg/obs-ffmpeg-source.c
```diff
@@ -463,26 +472,40 @@ static void ffmpeg_source_update(void *data, obs_data_t *settings)
...
+	range = obs_data_get_int(settings, "color_range");
+	should_restart_media |= (s->range != range);
+	speed_percent = (int)obs_data_get_int(settings, "speed_percent");
+	if (speed_percent < 1 || speed_percent > 200)
+		speed_percent = 100;
+	should_restart_media |= (s->speed_percent != speed_percent);
 	ffmpeg_options = obs_data_get_string(settings, "ffmpeg_options");
+	if (s->ffmpeg_options && !should_restart_media)
+		should_restart_media |=
+			strcmp(s->ffmpeg_options, ffmpeg_options) != 0;
```
Lain-B, line 487: I'd personally prefer something like:

```cpp
	is_hw_decoding = obs_data_get_bool(settings, "hw_decode");
	range = obs_data_get_int(settings, "color_range");
	speed_percent = (int)obs_data_get_int(settings, "speed_percent");
	if (speed_percent < 1 || speed_percent > 200)
		speed_percent = 100;
	ffmpeg_options = obs_data_get_string(settings, "ffmpeg_options");

	/* Restart media source if these properties are changed */
	if (s->is_hw_decoding != is_hw_decoding || s->range != range ||
	    s->speed_percent != speed_percent ||
	    (s->ffmpeg_options &&
	     strcmp(s->ffmpeg_options, ffmpeg_options) != 0))
		should_restart_media = true;
```
It's not that much cleaner, but it explicitly separates variable assignment and the check to restart media into their own chunks of code. Also would prefer `/* */` style comments (more info on that in the next comment)

plugins/obs-ffmpeg/obs-ffmpeg-source.c
```diff
@@ -491,17 +514,22 @@ static void ffmpeg_source_update(void *data, obs_data_t *settings)
...
-	if (s->media) {
+	if (s->media && should_restart_media) {
 		media_playback_destroy(s->media);
 		s->media = NULL;
 	}
 
 	bool active = obs_source_active(s->source);
-	if (!s->close_when_inactive || active)
+
+	if (s->media) { // directly set options if media is playing
```
Lain-B, line 524: I'd prefer the comment on the line above, and I'd prefer `/* */` style comments for C code. I know we've been inconsistent on this in the past though. I've just kind of starting thinking about this just recently after seeing it and I'd really like to have consistency with kernel normal form styling.

deps/media-playback/media-playback/media-playback.c
```diff
@@ -69,6 +69,23 @@ void media_playback_stop(media_playback_t *mp)
 		mp_media_stop(&mp->media);
 }
 
+void media_playback_set_looping(media_playback_t *mp, bool looping)
+{
+	if (mp->is_cached)
+		mp->cache.looping = looping;
+	else
+		mp->media.looping = looping;
+}
```
Lain-B, line 78: If this function called after media has already stopped playback, it will not start up playback again, which is what a user would expect if they re-tick the `Loop` option in properties. Instead it will just continue to not play back.

I'm not entirely sure how to deal with this elegantly right now considering playback is in another thread. I think the best thing is just to remove this looping function and restart the media source for the time being until we can come up with a more elegant solution.

Lain-B (changes requested):
All in all this PR is mostly good, just the looping idea doesn't really work, and needs a bit of code cleanup.

The looping change will need to either be dropped or changed somehow. I'd just recommend dropping the looping stuff and have it restart if looping has changed as well -- at least for the time being

Lain-B:
Sorry it took me so long to get back around to this. I'll be monitoring this closely from now on, so if you change it I'll know and get it re-reviewed in short order.

deps/media-playback/media-playback/media.c
```diff
@@ -474,6 +474,7 @@ void mp_media_next_video(mp_media_t *m, bool preload)
 	frame->height = f->height;
 	frame->max_luminance = d->max_luminance;
 	frame->flip = flip;
+	frame->flags = 0;
 	frame->flags |= m->is_linear_alpha ? OBS_SOURCE_FRAME_LINEAR_ALPHA : 0;
```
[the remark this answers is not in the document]
CodeYan01: I'm fine with this change, but I think I decided on this in case other flags are added. Changing `|=` to `=` would be the same here only because there is currently only one flag?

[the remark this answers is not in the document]
Lain-B: Correct. And I doubt other flags will be added.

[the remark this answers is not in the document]
CodeYan01: Okay

deps/media-playback/media-playback/media-playback.c
CodeYan01: I applied all other suggestions, but for this one, i didn't drop the looping. Instead I did this https://github.com/obsproject/obs-studio/pull/7455/files#diff-09b0325afe3e08bb7529d6918f3e970513784c791bc496e473c78b2e2519f9a9R499 which I think solves the issue. Please see if it is what you intended. As much as possible, I really don't want to restart the media if only the looping is changed, hence this solution.

Lain-B: Sounds good.

Lain-B (approved):

PR #7458, opened 2022-09-25
CI: Add cherry picker

derrod:
### Description

Adds automated cherry picks to branches specified in `cherry-pick-to-<branch>` tags.

### Motivation and Context

Want to make branching and picking easier.

### How Has This Been Tested?

PR on my repo: https://github.com/derrod/obs-studio/pull/21

Edit: Might require another test with a PR from an actual fork, it could be that the cherry-pick in that case since the commit hashes might not be available in the `pull_request_target` context.

### Types of changes

### Checklist:

### DIFF
--- a/.github/workflows/cherry-picker.yml
+++ b/.github/workflows/cherry-picker.yml
@@ -1,0 +1,36 @@
+name: Cherry Picker
+
+on:
+  pull_request_target:
+    types: [closed]
+
+jobs:
+  cherry_picker:
+    name: Cherry Picker
+    runs-on: ubuntu-22.04
+    if: ${{ github.event.pull_request.merged == true }}
+
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v3
+        with:
+          fetch-depth: 0
+
+      - name: Install & Configure Python
+        run: |
+          sudo apt install python3.10
+          python3.10 -m pip install requests
+
+      - name: Setup git config
+        run: |
+          git config user.name "Service Checker"
+          git config user.email "commits@obsproject.com"
+
+      - name: Run picker
+        id: picker
+        run: python3.10 -u CI/cherry-picker.py
+        env:
+          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+          WORKFLOW_RUN_ID: ${{ github.run_id }}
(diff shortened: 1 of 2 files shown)


PatTheMav:
Are we doing this often enough to warrant the maintenance burden?

derrod:
Marking as draft until I can properly test and - if need be - fix it.

gxalpha:
Was this already discussed to be per PR or is it just the current implementation?

With this being on PRs instead of commit, I see two issues:
- Some PR's contain multiple commits of which only some should get cherry-pick (and although you could argue that in such a case you're doing it wrong and it should instead be split into multiple PR's, this situation is something that does happen)
- People pushing directly to master (especially Jim) would have no way of automatically cherry-picking

derrod:
> Some PR's contain multiple commits of which only some should get cherry-pick (and although you could argue that in such a case you're doing it wrong and it should instead be split into multiple PR's, this situation is something that does happen)

Yes, in those cases it would still have to be done manually.

> People pushing directly to master (especially Jim) would have no way of automatically cherry-picking

Also true, but if you have write access you can do the cherry-pick locally and then push both branches.

I know Qt uses a commit-message based approach, but getting PR authors to put the cherry pick message into their commit descriptions would probably be more trouble than its worth.

.github/workflows/cherry-picker.yml
```diff
@@ -0,0 +1,36 @@
...
+
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v3
+        with:
+          fetch-depth: 0
+
+      - name: Install & Configure Python
+        run: |
+          sudo apt install python3.10
```
PatTheMav, line 21: How stable is this? Might it make sense to fetch the result of `python --version` (or something similar) to check the Python version present on the system?

derrod: This could probably be removed, `python3.10` is the default on GitHub runners and in theory this should be forwards-compatible. I don't know why I specifically put the version in there rather than just using `python3`.

derrod:
Closing for now as we probably don't need it quite yet and I'm not sure when I'll get around to finish this. I'll probably rewrite and resubmit whenever it's actually necessary.

PR #7467, opened 2022-09-26
CI: fix build on non-x86 Linux platforms

glikely:
### Description
The CI build script is useful to build OBS Studio even outside the CI environment. However, the current script hard codes adding the 'amd64' architecture to dpkg, the Debian/Ubuntu package management database. This breaks the system when run on non-x86 machines. e.g., a linux-aarch64 VM running on an Apple silicon macbook. Fixing the system requires telling dpkg to remove 'amd64' again with 'sudo dpkg --remove-architecture amd64'

As the CI build always runs on an amd64 machine, adding amd64 is not necessary and can be safely removed.

### How Has This Been Tested?
Tested locally on an Ubuntu linux-aarch64 virtual machine. Also tested by pushing out to GitHub and passing all Linux build actions.

### Types of changes

### Checklist:

### DIFF
--- a/CI/linux/01_install_dependencies.sh
+++ b/CI/linux/01_install_dependencies.sh
@@ -98,7 +98,6 @@ install_dependencies() {
          libpulse-dev libsndio-dev libspeexdsp-dev libudev-dev libv4l-dev libva-dev libvlc-dev libdrm-dev"
     )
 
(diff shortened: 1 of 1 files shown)


PatTheMav (approved):
Correct fix, had to do the same to build in my arm64-based Linux VM.

Issue #7407, opened 2022-09-19
release tarballs miss required sources


umlaeute:
### Operating System Info

Other

### Other OS

Debian

### OBS Studio Version

28.0.1

### OBS Studio Log URL

https://github.com/obsproject/obs-studio/releases/tag/28.0.1

### Expected Behavior

downloading the release tarballs from https://github.com/obsproject/obs-studio/releases/tag/28.0.1 should include the entire source required to build obs-studio

### Current Behavior

the release tarballs do not contain submodules, e.g. plugins/obs-websocket is missing.

this means that obs-studio cannot be built from the source tarballs.

### Steps to Reproduce

1. download https://github.com/obsproject/obs-studio/archive/refs/tags/28.0.1.tar.gz and extract it
2. install all the 3rd party libraries required
3. attempt to build
the build fails with:
```
CMake Error at cmake/Modules/ObsHelpers.cmake:446 (message):
  OBS: obs-websocket submodule not available.
Call Stack (most recent call first):
  plugins/CMakeLists.txt:86 (obs_status)
```

### Anything else we should know?

This is for creating the Debian-official packages of `obs-studio` (which will also be part of Ubuntu and other derivatives).
On Debian, we use release tarballs for creating our packages. 

Since github does not include any gitmodules in their generated "Source Downloads" (which is just a wrapper around `git archive`), crucial parts of the source code are missing.
The typical way to solve this (on the upstream (that is *your*) side) is to manually create a release tarball that contains all the necessary sources.

tytan652:
Maybe just download tarballs with the right commit hashes for each sub-modules.

kkartaltepe:
There is no way to disable github's broken tarballs or we would have. You can assume OBS does not distribute tarballs and it is incumbent on packagers to use git.

kkartaltepe:
We have added error messages to the missing submodules so that package managers know that they are disabling functionality if they refuse to use git. But you can continue to disable the functionality as it was silently disabled if you previously used github tarballs.

umlaeute:
@tytan652 wrone:
> Maybe just download tarballs with the right commit hashes for each sub-modules.

ah well, we have infrastructure to download tarballs for us, but unfortunately that doesn't work well if we have to assemble upstream sources from multiple tarballs.

@kkartaltepe wrote
> But you can continue to disable the functionality as it was silently disabled if you previously used github tarballs.

[...]

PR #7475, opened 2022-09-27
CI: add source_build job

umlaeute:
### Description

This adds a new job to the BUILD workflow, that generates a tarball with all the submodules (as GitHub's "Source Download" will happily exclude all the submodules, resulting in an incomplete tarball that fails to build).
    
The generated tarball excludes all repository configuration files (everything starting with .git*),
in order to not interfere with downstream consumers of the tarball, who might have their own ideas how to configure a repository.

@PatTheMav suggested in https://github.com/obsproject/obs-studio/issues/7407#issuecomment-1258023718 that i should extend the `build_linux` job, which I did **not** do for the following reasons:
- the build-artifacts are platform-independent (there's nothing "linux"-specific to them, except that i picked `tar.gz` as the output format)
- there's little point in creating the same tarball on multiple Ubuntu versions
- the tarballs might be of interest, even if the Linux builds fail (for whatever reasons)

### Motivation and Context
This is motivated by creating the Debian-official packages of obs-studio ("Debian-official" being the packages that are shipped by Debian itself (and thus by derivatives like Ubuntu), as opposed to .deb-packages shipped by obs-studio itself).

On Debian, we use release tarballs for creating our packages, and as such it is important that these packages contain all the necessary bits and bolts.

Closes: https://github.com/obsproject/obs-studio/issues/7407

### How Has This Been Tested?

This has been tested by running the workflow in my fork of OBS-studio.
https://github.com/umlaeute/obs-studio/actions/runs/3135647019

For whatever reasons, all the other build-jobs fail with some Cmake specific error. Since I did not touch these jobs, I think the problem is somewhere else (e.g. hidden in the project settings).

### Types of changes
?

### Checklist:
- [x] My code has been run through [clang-format](https://github.com/obsproject/obs-studio/blob/master/.clang-format). (*does not apply*)
- [x] I have included updates to all appropriate documentation. (*does not apply*)

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -369,6 +369,54 @@ jobs:
           name: 'obs-studio-windows-${{ matrix.arch }}-${{ steps.setup.outputs.commitHash }}'
           path: '${{ env.FILE_NAME }}'
 
+  source_build:
+    name: '02 - Source'
+    runs-on: [ubuntu-22.04]
+    if: always()
+    needs: [clang_check]
+    defaults:
+      run:
+        shell: bash
+        working-directory: 'obs-studio'
+    steps:
+      - name: 'Checkout'
+        uses: actions/checkout@v3
+        with:
+          submodules: 'recursive'
+          path: 'obs-studio'
+          fetch-depth: 0
+
+      - name: 'Setup environment'
+        id: setup
+        run: |
+          echo "commitHash=$(git rev-parse --short=9 HEAD)" >> $GITHUB_OUTPUT
+
+      - name: 'Create source artifact'
+        id: tarball
+        if: ${{ success() && github.event_name != 'pull_request' }}
+        run: |
+          prefix=obs-studio-${{ steps.setup.outputs.commitHash }}
+          commit_date=$(git show -s --format=%cI ${{ steps.setup.outputs.commitHash }})
+          outfile=${prefix}.tar.gz
+          ARTIFACT_NAME=${outfile}
+          outfile=${{ github.workspace }}/${outfile}
+          echo "outfile: $outfile"
+          echo "artifact: $ARTIFACT_NAME"
+          # create a tarball with *all* the sources
+          # - prefix paths with 'obs-studio/' (since we are creating the tarball from within the source dir)
+          # - exclude all repository files
+          # - make the tarball reproducible
+          tar -cvzf ${outfile}  --transform='s|^\.|obs-studio|'  --exclude-vcs --exclude-vcs-ignores --exclude '.git*'  --sort=name  --owner=0 --group=0 --numeric-owner --pax-option="exthdr.name=%d/PaxHeaders/%f,delete=atime,delete=ctime" --mtime="${commit_date}"  .
+          # announce artifact for upload
+          echo "fileName=${ARTIFACT_NAME}" >> $GITHUB_OUTPUT
+
+      - name: 'Upload source artifact'
+        if: ${{ success() && (github.event_name != 'pull_request') }}
+        uses: actions/upload-artifact@v3
+        with:
+          name: 'obs-studio-source-${{ steps.setup.outputs.commitHash }}'
+          path: '${{ github.workspace }}/${{ steps.tarball.outputs.fileName }}'
+
   linux_package:
     name: '02 - Flatpak'
     runs-on: [ubuntu-latest]


umlaeute:
sidenote: the action i linked above as a "proof of test" is obviously on my `master` branch and contains multiple commits.
this is because the given job is limited to only run on `master` and `release/*` branches (so it wouldn't run on my `release-tarballs` branch).
the commits have been squashed into the final branch for this PR.

.github/workflows/main.yml
```diff
@@ -361,6 +361,56 @@ jobs:
...
+          prefix=obs-studio-${{ steps.setup.outputs.commitHash }}
+          outfile=${prefix}.tar
+          rm -f ${outfile}
+          # exclude repository configuration files
+          find .git -type d -name info -exec touch {}/attributes ";"
+          echo ".git* export-ignore" | find .git -type f -path "*/info/attributes" -exec tee -a {} +
+          # create a tarball with all the submodules
+          # this uses uncompressed 'tar', so we can append the archives
+          git archive --format tar -o ${outdir}/${outfile}  --prefix=$prefix/ HEAD
+          PREFIX=${prefix} OUTFILE=${outdir}/${outfile} git submodule --quiet foreach --recursive 'git archive --format tar -o submodule.tar --prefix=$PREFIX/$displaypath/ $sha1; tar -Af ${OUTFILE} submodule.tar; rm -f submodule.tar'
```
PatTheMav, line 400: This doesn't take a possible error in each previous step into account and will lead to broken archives. `tar` should only be invoked if the submodule fetch was successful.

Also `rm` will throw an error, if the tar file is not created. One could argue that the entire command will abort the workflow because of that, but you should abort early, especially as consecutive commands implicitly depend on prior commands to be successful.

.github/workflows/main.yml
```diff
@@ -361,6 +361,56 @@ jobs:
...
+          echo "::set-output name=commitHash::$(git rev-parse --short=9 HEAD)"
+
+      - name: 'Create source artifact'
+        if: ${{ success() && github.event_name != 'pull_request' }}
+        run: |
+          outdir=$(pwd)/build
+          mkdir -p ${outdir}
+          prefix=obs-studio-${{ steps.setup.outputs.commitHash }}
+          outfile=${prefix}.tar
+          rm -f ${outfile}
```
PatTheMav, line 393: Check if the file exists before attempting to remove it, don't rely on `-f` - it's an unnecessary call to an external program.

Given that this script only runs on CI, you could even remove this entirely as you're always working on a fresh checkout.

.github/workflows/main.yml
```diff
@@ -361,6 +361,56 @@ jobs:
...
+          # exclude repository configuration files
+          find .git -type d -name info -exec touch {}/attributes ";"
+          echo ".git* export-ignore" | find .git -type f -path "*/info/attributes" -exec tee -a {} +
+          # create a tarball with all the submodules
+          # this uses uncompressed 'tar', so we can append the archives
+          git archive --format tar -o ${outdir}/${outfile}  --prefix=$prefix/ HEAD
+          PREFIX=${prefix} OUTFILE=${outdir}/${outfile} git submodule --quiet foreach --recursive 'git archive --format tar -o submodule.tar --prefix=$PREFIX/$displaypath/ $sha1; tar -Af ${OUTFILE} submodule.tar; rm -f submodule.tar'
+          # zip and upload
+          gzip ${outdir}/${outfile}
+          ARTIFACT_NAME=$(/usr/bin/find ${outdir} -maxdepth 1 -type f -name "${outfile}.*" -exec basename {} ";" -quit)
```
PatTheMav, line 403: Given that you create a single file with a pre-defined filename, you should not need to use `find` here. `-exec` is also.. not great (see the requirement to add `";"` or `\;` after it to be able to use `-quit`), but not a blocker.

PatTheMav (changes requested):
This seems overly complicated to me - I'm fine with running this as a separate job for the reasons you mentioned, but the whole business of manipulating files and appending submodules to the tar archive seems unnecessary to me.

Just run `git ls-files --recurse-submodules | tar caf ../obs-studio.tar.gz -T-` in the checkout directory and don't mind the `.gitignore` files that might be left over - IMO they belong to our source tree.

From what I could tell that single command gives us all the files from all submodules, except if I missed something?

You also forgot to add the artefact to the release creation job - we'd also want to add the tarball to our releases as well then.

umlaeute:
> don't mind the .gitignore files that might be left over - IMO they belong to our source tree.

how so?
having repository configuration in a source tarball is an endless source of troubles for us downstream packagers.

1. when you download the tarball, there is no repository anywhere. what's the benefit of having repository configuration in it then?
2. if you then import the tarball into your own repository, the configuration for a different repository (upstream) keeps interfering with the downstream needs.

the thing is: our needs are different from yours. we **do** want to see *any* build artifacts that is left around.

you might of course argue that we could simply remove the repository configuration that we don't like on our side, but in general we want to change the upstream sources as little as possible, and as i don't see any reason to do include the repo-config in an explicitly non-repo artifact, i ask you to just remove it. if the source tarballs had already been useable as such, i would have submitted a patch to drop any repository configuration anyway.

.github/workflows/main.yml
umlaeute: that would rely on `gzip` adding a `.gz` prefix (which of course it does).

in any case, i mimicked the behaviour of the existing jobs, that already uses `find` for what seems to be pre-defined filenames as well, see https://github.com/obsproject/obs-studio/blob/dc30cf0843c54f3f02ad412daa10038784ae3be1/.github/workflows/main.yml#L192

i personally find the use of `-exec` with `";"` and `-quit` more clean and resilient than the existing use of piping the output of `find` into `head`.

.github/workflows/main.yml
umlaeute: although, i don't full y understand the problem with the "unecessary call" (in case the file does exist).

in any case, i'd go for the entire line...

.github/workflows/main.yml
umlaeute: i'm not sure i can follow.
the code is written on the assumption that the sCript is executed with `set -e` (as is the case on CI-systems).

so the only problem i see is that within the `foreach` loop, `tar -Af` is executed even if `tar -o submodule.tar` failed...
i'll fix that.

in the CI context, i'll remove the deletion of `submodule.tar` (as it happens in a throwaway container anyhow)

PatTheMav: You execute a command and rely on its success for a consecutive command. Even without `set -e` this should use `&&` to ensure that consecutive commands only run on success to avoid relying on functionality you cannot control. Also the shell grammar allows you to be explicit about it.

Also have you checked if git aborts if an iteration of a `foreach` command has a non-zero return?

umlaeute: i did check `git foreach` (admittedly, after your critique), and it indeed aborts if a sub-command returns non-0

umlaeute:
> Fine by me, still `git ls-files --recurse-submodules -- . ':!:.git*' | tar caf ../obs-studio.tar.gz -T-` should get you a working tarball in a single command - or are there files missing from that archive?

your command is of course much simper and more elegant than mine.
it doesn't exclude the.git*` files in subdirectories, though. (i'll try to come up with a solution for this, which might take over the weekend - there are other pressing matters to attent to :-))

PatTheMav:
> > Fine by me, still `git ls-files --recurse-submodules -- . ':!:.git*' | tar caf ../obs-studio.tar.gz -T-` should get you a working tarball in a single command - or are there files missing from that archive?
> 
> your command is of course much simpler and more elegant than mine. it doesn't exclude the.git*` files in subdirectories, though. (i'll try to come up with a solution for this, which might take over the weekend - there are other pressing matters to attent to :-))

That’s weird - the Git glob expression should take care of that, might need some tweaking.

Polynomial-C:
If you're not using an ancient tar version, you could also use the `--exclude-vcs` option.

umlaeute:
so i force pushed a change using the suggestion by @Polynomial-C 

the code has become significantly simpler, basically running

- `tar -cvzf ${outfile}` 
- `--transform='s|^\.|obs-studio|'` - as we are running this from checked out source tree
- `--exclude-vcs --exclude-vcs-ignores --exclude '.git*'` - removing all repository data and configuration
- `--sort=name  --owner=0 --group=0 --numeric-owner --pax-option="exthdr.name=%d/PaxHeaders/%f,delete=atime,delete=ctime" --mtime="${commit_date}"` - making the tarball reproducible (so multiple runs should create exactly the same tarball)
-  `.` - the source tree

PatTheMav:
Nicely done, looks neat now. I will take a closer look tomorrow but looks fine at a first glance.

umlaeute:
ping @PatTheMav

umlaeute:
ping...

PatTheMav (approved):
Seems fine to me - @tytan652 any objections to merge this?

.github/workflows/main.yml
```diff
@@ -361,6 +361,53 @@ jobs:
...
+          outfile=${{ github.workspace }}/${outfile}
+          echo "outfile: $outfile"
+          echo "artifact: $ARTIFACT_NAME"
+          # create a tarball with *all* the sources
+          # - prefix paths with 'obs-studio/' (since we are creating the tarball from within the source dir)
+          # - exclude all repository files
+          # - make the tarball reproducible
+          tar -cvzf ${outfile}  --transform='s|^\.|obs-studio|'  --exclude-vcs --exclude-vcs-ignores --exclude '.git*'  --sort=name  --owner=0 --group=0 --numeric-owner --pax-option="exthdr.name=%d/PaxHeaders/%f,delete=atime,delete=ctime" --mtime="${commit_date}"  .
+          # announce artifact for upload
+          echo "FILE_NAME=${ARTIFACT_NAME}" >> $GITHUB_ENV
```
tytan652, line 402: This has no need to be an env var, turn it into an output.

tytan652 (commented):
Running this all the time is my biggest complain, creating useless artifacts except when it's a release.

.github/workflows/main.yml
umlaeute: i've (rebased on current master and) force pushed a version that now uses outputs rather than envvars.

#### run only on tag/release

afaict, the current pipelines will generate the following artifacts on each push:
|artifact|size|
|--------|----|
|obs-studio-flatpak-5b370d4e3-x86_64|139 MB|
|obs-studio-macos-arm64-5b370d4e3|146 MB|
|obs-studio-macos-x86_64-5b370d4e3|156 MB|
|obs-studio-ubuntu-20.04-5b370d4e3|108 MB|
|obs-studio-ubuntu-20.04-5b370d4e3-dbgsym|80.8 MB|
|obs-studio-ubuntu-22.04-5b370d4e3|108 MB|
|obs-studio-ubuntu-22.04-5b370d4e3-dbgsym|72.9 MB|
|obs-studio-windows-x64-5b370d4e3|186 MB|
|obs-studio-windows-x86-5b370d4e3|167 MB|
|**TOTAL**|**1160MB**|

the source-job adds another artifact with a size of approximately 32MB.
The worry that this job imposes a problem, seems a bit ... exaggerated to me.

But anyhow: I see that e.g. the `Windows Installer` resp `macOS notarized image` jobs seem to only run for releases by using

```yaml
if: ${{ startsWith(github.ref, 'refs/tags/') && github.event_name != 'pull_request' }}
```

if you want I can add this condition (but that means that the pipeline won't be run before you merge, which imho is even less desirable; but ymmv, so just tell me if you require me to do so).

umlaeute:
> Running this all the time is my biggest complain, creating useless artifacts except when we need one for a release/tag.
> 
> Edit: But there is no alternative, if I understood correctly.


afaict, the current pipelines will generate the following artifacts on each push:
|artifact|size|
|--------|----|
|obs-studio-flatpak-5b370d4e3-x86_64|139 MB|
|obs-studio-macos-arm64-5b370d4e3|146 MB|
|obs-studio-macos-x86_64-5b370d4e3|156 MB|
|obs-studio-ubuntu-20.04-5b370d4e3|108 MB|
|obs-studio-ubuntu-20.04-5b370d4e3-dbgsym|80.8 MB|
|obs-studio-ubuntu-22.04-5b370d4e3|108 MB|
|obs-studio-ubuntu-22.04-5b370d4e3-dbgsym|72.9 MB|
|obs-studio-windows-x64-5b370d4e3|186 MB|
|obs-studio-windows-x86-5b370d4e3|167 MB|
|**TOTAL**|**1160MB**|

the source-job adds another artifact with a size of approximately 32MB.
The worry that this job imposes a problem, seems a bit ... exaggerated to me.

But anyhow: I see that e.g. the `Windows Installer` resp `macOS notarized image` jobs seem to only run for releases by using

```yaml
if: ${{ startsWith(github.ref, 'refs/tags/') && github.event_name != 'pull_request' }}
```

if you want I can add this condition (but that means that the pipeline won't be run before you merge, which imho is even less desirable; but ymmv, so just tell me if you require me to do so).

umlaeute:
@tytan652 how does this sound?

tytan652:
> @tytan652 how does this sound?

I let @PatTheMav having the final say.

> The worry that this job imposes a problem, seems a bit ... exaggerated to me.

What I talked about was a piece of the problem. The main problem was discussed off-thread.

umlaeute:
> What I talked about was a piece of the problem. The main problem was discussed off-thread.

ah i see. sorry for being snarky then.
is the off-thread discussion available somewhere for external people like me?

tytan652:
> is the off-thread discussion available somewhere for external people like me?

It was between OBS developers, so no.
But long-story short it was a quick discussion about:
- how we do releases and this PR
- the WIP future CI rewrite

So do not worry about any of that.

umlaeute:
> > @tytan652 how does this sound?
> 
> I let @PatTheMav having the final say.
> 

any news on this?

umlaeute:
@tytan652 @PatTheMav ping

umlaeute:
ping @tytan652 @PatTheMav

umlaeute:
ping @tytan652 @PatTheMav

PatTheMav:
This is already partially implemented in the updated workflows, but will be enabled once Ubuntu-based builds use the updated CMake path sometime after the next release (similar to how it's already implemented in the plugintemplate).

As this was based on the old workflows and is not compatible with the updated CI system, I'll close this PR. But the feature will definitely be implemented.

PR #7476, opened 2022-09-27
[Feature Request] | 'Facebook VOD Track' ?

azgoodaz:
### Description

More and more creators are jumping ship from Twitch and it would be great to get a Facebook VOD Track like how Twitch has this feature. Getting the ability to select a setting like 'Twitch VOD Track' which records a different audio for the VOD so stream doesn't get any copyrighted notices.

### Motivation and Context

This will allow Facebook creators to livestream content without having streams being taken down due to copyright notices. If you stream copyrighted music on YouTube by mistake from a video game the stream has a chance to be taken down then to have to press 'Livestream' again in order to livestream again. It's a bit redundant.

### Types of changes

### DIFF
--- a/libobs/obs-config.h
+++ b/libobs/obs-config.h
@@ -41,7 +41,7 @@
  *
  * Reset to zero each major or minor version
  */
-#define LIBOBS_API_PATCH_VER 1
+#define LIBOBS_API_PATCH_VER 2
 
 #define MAKE_SEMANTIC_VERSION(major, minor, patch) \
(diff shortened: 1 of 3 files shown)


tytan652:
Pull request is not a place for feature request, issue neither.

Post your ideas and suggestions: https://ideas.obsproject.com/

Fenrirthviti:
As an additional note, VOD tracks would need to be supported by the platform, so this is not something we could implement.

PR #7480, opened 2022-09-28
docs: Add links to python functions

CodeYan01:
### Description

Adds hyperlinks to the scripting functions that replace the C functions.

Example screenshots:
image
image

### Motivation and Context

When looking through the documentation, I don't memorize all the overriding scripting functions, so I want to know immediately that the scripting equivalent has a different function signature.

### How Has This Been Tested?

Built and viewed locally, and tested the hyperlinks.

### Types of changes

### Checklist:

### DIFF
--- a/docs/sphinx/reference-core.rst
+++ b/docs/sphinx/reference-core.rst
@@ -278,6 +278,8 @@ Libobs Objects
    :c:func:`obs_source_get_weak_source()` if you want to retain a
    reference after obs_enum_sources finishes.
 
+   For scripting, use :py:func:`obs_enum_sources`.
+
 ---------------------
 
 .. function:: void obs_enum_scenes(bool (*enum_proc)(void*, obs_source_t*), void *param)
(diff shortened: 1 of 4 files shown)


docs/sphinx/reference-core.rst
```diff
@@ -278,6 +278,8 @@ Libobs Objects
    :c:func:`obs_source_get_weak_source()` if you want to retain a
    reference after obs_enum_sources finishes.
 
+   For scripting, use :py:func:`obs_enum_sources`.
```
WizardCM, line 281: Any reason you went with `:py:` instead of `:c:`?

CodeYan01: That's under .. function:: void obs_enum_sources(bool (*enum_proc)(void*, obs_source_t*), void *param), which is a C function. If I used :c:, that would link to the same function it's under. I used :py: so it links to the python documentation for the function of the same name. Remember that many of them have different function signatures from their C function equivalents.

WizardCM: Right, good point!

PR #7482, opened 2022-09-28
obs-scripting: Re-enable Python annotations; Enable Python autodoc

WizardCM:
### Description

Re-enables Python annotations which were automatically disabled by the upgrade to SWIG 4. swig/swig@2072ae19c
This was previously available in OBS 27.2.

Additionally enables autodoc to provide the same information to apps such as VS Code.

### Python Annotations

**Before:**

```py
def obs_property_set_long_description(p, long_description):
    return _obspython.obs_property_set_long_description(p, long_description)
```

**After:**

```py
def obs_property_set_long_description(p: "obs_property_t *", long_description: "char const *") -> "void":
    return _obspython.obs_property_set_long_description(p, long_description)

```

### Python autodoc

**Before (same with/without annotations):**

image

```py
def obs_property_set_long_description(p: "obs_property_t *", long_description: "char const *") -> "void":
    return _obspython.obs_property_set_long_description(p, long_description)

```

**After:**

image

```py
def obs_property_set_long_description(p: "obs_property_t *", long_description: "char const *") -> "void":
    r"""
    obs_property_set_long_description(p, long_description)

    Parameters
    ----------
    p: obs_property_t *
    long_description: char const *

    """
    return _obspython.obs_property_set_long_description(p, long_description)
```

### Motivation and Context

Providing more context to Python developers about what parameters they need to provide makes writing scripts faster.

### How Has This Been Tested?

Compile OBS. Check `obspython.py` and VS Code autocomplete.

### Types of changes

### Checklist:

### DIFF
--- a/deps/obs-scripting/obspython/obspython.i
+++ b/deps/obs-scripting/obspython/obspython.i
@@ -42,6 +42,7 @@
 %}
 
 %feature("python:annotations", "c");
+%feature("autodoc", "2");
 
 #define DEPRECATED_START
 #define DEPRECATED_END
--- a/deps/obs-scripting/obspython/obspython.i
+++ b/deps/obs-scripting/obspython/obspython.i
@@ -41,6 +41,8 @@
 
 %}
 
+%feature("python:annotations", "c");
+
 #define DEPRECATED_START
 #define DEPRECATED_END
 #define OBS_DEPRECATED


PatTheMav:
I guess this is not automatically enabled given our "stable ABI" variant.

What's the current state of type hints in SWIG? We should use valid PEP 484 hints if available, as this would enable type checking via LSP and other toolchains. Using the C-types has only visual value, proper types have code safety value.

deps/obs-scripting/obspython/obspython.i
```diff
@@ -41,6 +41,9 @@
 
 %}
 
+%feature("python:annotations", "c");
```
PatTheMav, line 44: Would prefer using Python annotations or PEP 484 annotations if available.

PatTheMav: So only C annotations are available right now? Fair enough, we take what we can get.

WizardCM: Unfortunately, yeah. @CodeYan01 and I are experimenting with potential alternatives, but honestly none of them are looking fantastic.

PR #7497, opened 2022-09-30
Retain P3 for Mac screen capture

jpark37:
### Description
Added support for retaining P3 colors when capturing Mac screens. Gets clipped for SDR recordings, but kept for HDR recordings.

### Motivation and Context
Want to keep wide colors when possible.

### How Has This Been Tested?
Verified SDR looks the same, and AOM AV1 HDR recording retains P3 colors.

### Types of changes

### Checklist:

### DIFF
--- a/plugins/mac-capture/mac-screen-capture.m
+++ b/plugins/mac-capture/mac-screen-capture.m
@@ -462,8 +462,8 @@ static bool init_screen_stream(struct screen_capture *sc)
 	os_sem_post(sc->shareable_content_available);
 	[sc->stream_properties setQueueDepth:8];
 	[sc->stream_properties setShowsCursor:!sc->hide_cursor];
-	[sc->stream_properties setColorSpaceName:kCGColorSpaceSRGB];
-	[sc->stream_properties setPixelFormat:'BGRA'];
+	[sc->stream_properties setColorSpaceName:kCGColorSpaceDisplayP3];
+	[sc->stream_properties setPixelFormat:'l10r'];
 #if __MAC_OS_X_VERSION_MAX_ALLOWED >= 130000
 	if (@available(macOS 13.0, *)) {
 		[sc->stream_properties setCapturesAudio:TRUE];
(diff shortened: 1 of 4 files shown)


plugins/mac-capture/mac-screen-capture.m
```diff
@@ -462,8 +462,8 @@ static bool init_screen_stream(struct screen_capture *sc)
 	os_sem_post(sc->shareable_content_available);
 	[sc->stream_properties setQueueDepth:8];
 	[sc->stream_properties setShowsCursor:!sc->hide_cursor];
-	[sc->stream_properties setColorSpaceName:kCGColorSpaceSRGB];
-	[sc->stream_properties setPixelFormat:'BGRA'];
+	[sc->stream_properties setColorSpaceName:kCGColorSpaceDisplayP3];
+	[sc->stream_properties setPixelFormat:'l10r'];
```
PatTheMav, line 466: Is it correct to set this to `l10r` in all cases?

jpark37: Yeah. I don't think 8 bits is sufficient for P3. Shouldn't be much of a perf hit since it's still 32 bits per pixel, just need a little extra shader math.

PatTheMav: Right, I checked the docs, it looks like P3 is effectively RGB with 16-bit per channel:

> use the Display P3 color profile at 16 bits per pixel (per channel)

https://developer.apple.com/design/human-interface-guidelines/foundations/color/#color-management)

jpark37: l10r is the only format in the list greater than 8 bit per channel, so it seems like the best fit.
> BGRA: packed little endian ARGB8888
l10r: packed little endian ARGB2101010
420v: two-plane “video” range YCbCr 4:2:0
420f: two-plane “full” range YCbCr 4:2:0

https://developer.apple.com/documentation/screencapturekit/scstreamconfiguration/3919825-pixelformat

PR #7513, opened 2022-10-02
mac-videotoolbox: Remove "None" profile

gxalpha:
### Description

Selecting this would actually fall back to "main", so we should just use
main as the default.

### Motivation and Context

Noticed this when looking at VT again the other day. 
I guess this is what happens when half of the source file is copied from x264.

### How Has This Been Tested?

macOS 13
Options now looks like this:
image

### Types of changes

- Bug fix. The currently displayed setting is wrong.

### Checklist:


PatTheMav:
Where does this fallback take place? If we do this when the config value is an empty string, couldn't we just remove this property value and move the "main" value to the top (making it the default)?

gxalpha:
> Where does this fallback take place?

https://github.com/obsproject/obs-studio/blob/2a76ffaf02334527c1434be140c680c8886f4969/plugins/mac-videotoolbox/encoder.c#L86-L87

> couldn't we just remove this property value and make "main" the default?

I don't think we can, because of existing setups.
While someone who never touched that setting (which means the value doesn't exist) would correctly get moved to main, someone who explicitly set it to None (meaning the value now is an empty string, which in this context isn't the same as the value missing) will still have said empty string (which falls back to `main`) in the config, while the settings will say `baseline` (the first value in the dropdown) until they edit it.

> and move the "main" value to the top

That would fix the problem described above - however, it would also be very confusing to users as the profiles would now be out-of-order. Not sure I'd be a big fan of that.

gxalpha:
I think that's fair, especially considering that the only annoyance is that when opening the properties, it shows something different, but if the user doesn't look at the settings nothing will break.

I updated the PR accordingly.

PR #7514, opened 2022-10-02
libobs: Bind to only one wayland seat

markbolhuis:
### Description
Bind to only one wl_seat.

### Motivation and Context
Compositors like Sway are mulit-seat aware. This means the registry will send multiple global events for each wl_seat.
This cause an issue where obs will leak every seat other than the last one sent, and will try to get the keyboard for a seat that doesn't support it, causing a protocol error.

Relevant Wayland log:
```
[2453323.467] wl_seat@30.name("seat0")
[2453323.508] wl_seat@30.capabilities(3)
[2453323.516]  -> wl_seat@31.get_keyboard(new id wl_keyboard@34)
[2453323.526] wl_seat@31.name("seat1")
[2453323.533] wl_seat@31.capabilities(0)
[2453323.539]  -> wl_keyboard@34.release()
[2453323.719] wl_display@1.error(wl_seat@31, 0, "wl_seat.get_keyboard called when no keyboard capability has existed")
wl_seat@31: error 0: wl_seat.get_keyboard called when no keyboard capability has existed
The Wayland connection experienced a fatal error: Protocol error
```

### How Has This Been Tested?
- ArchLinux: 5.19.12.arch1-1
- Sway: 1.7 (configured with 2 seats)

### Types of changes

### Checklist:

### DIFF
--- a/libobs/obs-nix-wayland.c
+++ b/libobs/obs-nix-wayland.c
@@ -219,6 +219,11 @@ static void platform_registry_handler(void *data, struct wl_registry *registry,
 	obs_hotkeys_platform_t *plat = (obs_hotkeys_platform_t *)data;
 
 	if (strcmp(interface, wl_seat_interface.name) == 0) {
+		if (plat->seat != NULL) {
+			blog(LOG_WARNING,
+			     "[wayland] multiple seats detected, only using the first one");
+			return;
+		}
 		if (version < 4) {
 			blog(LOG_WARNING,
 			     "[wayland] hotkeys disabled, compositor is too old");


kkartaltepe:
It looks like this will break if the first seat provided is not the one with the keyboard.

Sorry im not familiar enough with precisely what a seal is intended to represent, but probably we should either listen to every keyboard or pick one. Since we need to bind the seat to check its capabilities we still need to bind to all of them.

markbolhuis:
A seat is an event aggregator. Basically it represents a collection of input devices that all logically belong to one person sitting at the computer. 

‐--------------------

If the first seat doesn't have a keyboard capability then the keyboard object isn't created:

https://github.com/obsproject/obs-studio/blob/6f115d60143547b37600aac806bd978ced4932a0/libobs/obs-nix-wayland.c#L192

The issue wasn't that there was no keyboard at all, but that the capabilities event triggered after binding to the first seat has the keyboard bit set, but by then any other seat had replaced that one, so attempting to get the keyboard terminated the connection if that new seat didn't support it.

-------------

We should not try to be clever and pick different seats just because they have a keyboard capability. As I said a seat represents a person, so picking a different one would mean a completely different collection of physical keyboards intended for a different person.

Ideally OBS should be multiseat aware, but for my first commit I decided to not be too clever and just fix it crashing.
Proper multiseat support is much more involved. For single seat apps picking the first one is fine.

kkartaltepe:
> We should not try to be clever and pick different seats just because they have a keyboard capability. As I said a seat represents a person, so picking a different one would mean a completely different collection of physical keyboards intended for a different person.

We already (try) pick the seat with a keyboard. Please provide protocol & compositor reference that we shouldn't bind to the most appropriate seat given to us if you think it should be changed.

markbolhuis:
> Please provide protocol & compositor reference that we shouldn't bind to the most appropriate seat given to us if you think it should be changed.

Unfortunately we don't know which seat is the most appropriate. A seat may gain or lose the keyboard capability at any time. So just because one of the seats has a keyboard on the first capabilities event doesn't mean it will persist, or another won't gain it later. 
[wayland.xml:1817](https://gitlab.freedesktop.org/wayland/wayland/-/blob/main/protocol/wayland.xml#L1817)

We also cannot use the name of the seat, since there is no naming convention. 
[wayland.xml:1891](https://gitlab.freedesktop.org/wayland/wayland/-/blob/main/protocol/wayland.xml#L1891)

------------------------------------

The only compositors that I'm aware of that handle multiple seats are Sway and Weston, and each handle it differently.

Weston's first seat is named "default" but more can be added with udev rules.
[Tizen Docs](https://wiki.tizen.org/Multi-user_DisplayManagement) 
[Weston Docs](https://wayland.pages.freedesktop.org/weston/toc/running-weston.html#running-weston-on-a-different-seat-on-a-stand-alone-back-end)

Sway's first seat is named "seat0", and more are added by attaching devices in `.config/sway/config`.
[Sway seat configuration](https://github.com/swaywm/sway/blob/master/sway/sway-input.5.scd#seat-configuration)

---------------------------------

Ideally we need to know which seats Qt is using.
From a brief loot at their [source code](https://code.qt.io/cgit/qt/qtwayland.git/tree/src/client/qwaylanddisplay.cpp#n490) it seems Qt binds to all seats. Although I'm not sure how it handles multiple focus.

kkartaltepe:
sway developers suggest we listen to every keyboard, so that can be an enhancement to support multi-seat setups.

markbolhuis:
> sway developers suggest we listen to every keyboard, so that can be an enhancement to support multi-seat setups.

Ok, I'll close this MR and move the discussion to #7517

PR #2198, opened 2019-11-26
UI: Implement UI collections

Programatic:
### Description
This adds ui collections commonly referred to as "workspaces". It allows you to create mutliple different formats and easily switch between them.

### Motivation and Context
It simply allows for more flexibility when using OBS for different purposes. It implements the suggestion found here: https://ideas.obsproject.com/posts/615/ui-collections

### How Has This Been Tested?
I tested it on my Linux machine but nothing else yet.

### Todo
- Needs testing
- Implementing Import/Export (foundation is there from scene collection but I haven't touched it yet)
- Need to add more configurables being saved
- Need to work on how the configs are saved and loaded
- Rename to workspaces
- Add API
- Possibly will implement hotkeys

### Known Bugs
- [x] Misbehaves when working with custom browser docks
- [x] Startup gives unnecessary reset ui notification
- [x] Stores window position as well
- [x] completely ignores old configuration so user would have to start from scratch
- [x] Memory Leaks

### Types of changes

- New feature (non-breaking change which adds functionality)

### Checklist:

- [ ] My code has been run through [clang-format](https://github.com/obsproject/obs-studio/blob/master/.clang-format).
- [x] I have read the [**contributing** document](https://github.com/obsproject/obs-studio/blob/master/CONTRIBUTING.rst).
- [x] My code is not on the master branch.
- [ ] The code has been tested.
- [ ] All commit messages are properly formatted and commits squashed where appropriate.

PR #7542, opened 2022-10-06
Added toolbar menu to export and import dock sets with recent dockset…

dcmouser:
Added toolbar submenu for exporting and importing docksets (configuration files containing layout of docks), as well as a recent dockset list.

### Description
Added new submenu under Docks to allow user to export and import docksets (saved configurations of docks), along with a recent dockset list.  This makes it easy to save configurations of docks (where they are, which ones are visible, etc.).  Very lightweight, uses the built-in dock saving and loading code in OBS.

### Motivation and Context
Lots of requests for this over the years.

### How Has This Been Tested?

### Types of changes

### Checklist:

### DIFF
--- a/UI/data/locale/en-US.ini
+++ b/UI/data/locale/en-US.ini
@@ -793,6 +793,13 @@ Basic.MainMenu.Docks.ResetDocks="&Reset Docks"
 Basic.MainMenu.Docks.LockDocks="&Lock Docks"
 Basic.MainMenu.Docks.CustomBrowserDocks="&Custom Browser Docks..."
 
+#ATTN: importing and exporting docksets
+Basic.DocksMenu.menuDocksets="&Dock sets"
+Basic.DocksMenu.menuDocksets.importDockset="&Import dockset..."
+Basic.DocksMenu.menuDocksets.exportDockset="&Export dockset..."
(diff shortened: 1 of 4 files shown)


Fenrirthviti:
As a note, we have an open RFC which is being strongly considered to move to a new docking system (ADS):

https://github.com/obsproject/rfcs/pull/47

If we decide to accept this RFC, it would mean that this functionality would need to be replaced, and won't be compatible. However, switching to ADS will break existing dock setups anyway even without saving of layouts, so this may or may not be a blocker depending on how far away the ADS implementation is likely to be, or if it's accepted in the first place.

For the time being, let's evaluate this PR with the assumption that ADS is not on the table since merging this would more or less be unrelated, and would just be something implemented as part of ADS (so this would be removed at this time).

dcmouser:
@Fenrirthviti That seems very reasonable.  This code is very un-intrusive to the base code, and the minor menu additions could be adapted to a future dock system and the menu code for import/export and showing recent dock set files will all still be useful.

I feel like the ability to save and load dock configurations could be useful while waiting for any proposed new dock system.  And I hope it will help the new dock system coders see that loading and saving dock configurations is an important feature to implemented in any new system.

Warchamp7:
There was some relevant discussion on this idea previously in #2198

PR #7543, opened 2022-10-06
Added feature to image source slideshow to rescan for changes in directory on activation

dcmouser:
Super simple change to image source slideshow functionality to rescan the directory for changes in available files when activating the source.

### Description
Super simple change to image source slideshow functionality to rescan the directory for changes in available files when activating the source.

### Motivation and Context
  This has been requested on the OBS forum a few times.

### How Has This Been Tested?

### Types of changes

### Checklist:

### DIFF
--- a/plugins/image-source/obs-slideshow.c
+++ b/plugins/image-source/obs-slideshow.c
@@ -59,6 +59,12 @@
 
 /* clang-format on */
 
+
+// ATTN: jr
+// to rescan for changes in files
+static void jrRescanFiles(struct slideshow *ssdata);
+
+
 /* ------------------------------------------------------------------------- */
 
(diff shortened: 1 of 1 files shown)


TeaseTest:
Is it possible to make this change to OBS myself? If so, what file(s) would I need to update? Also, can it be coded to refresh the image directory when the slideshow source is hidden and unhidden? I only use one scene so if it only updates when switching scenes then it doesn't help. Thanks!

JamesOsborn-SE:
@dcmouser you may want to merge with master as it fails clang-format but it's not your added code.Also whats with the "jr" prefixes and ATTN: stuff?
I don't' think you're following the naming convention with: `static void jrRescanFiles(`

PR #7544, opened 2022-10-06
Added new events broadcast_starting, broadcast_started, broadcase_start_failed

dcmouser:
Added new events triggered by Go Live functionality in OBS for starting broadcasts on YouTube. 

### Description
Just a couple of new events to match the new GoLive button in OBS that transitions stream between streaming and actually going live.

### Motivation and Context
These events can be very useful for script to know when the stream has actually gone live (as opposed to when streaming was started, which may have happened many minutes before).  These events are particularly useful for calculating timestamps on YouTube.

### How Has This Been Tested?

### Types of changes

### Checklist:


UI/window-basic-main.cpp
```diff
@@ -6648,8 +6651,14 @@ void OBSBasic::BroadcastButtonClicked()
 					QTStr("Output.BroadcastStartFailed"),
 					last_error, true);
 				ui->broadcastButton->setChecked(false);
+				// broadcast start failed event
```
norihiro, line 6654:
```suggestion
```
This comment is not meaningful because the event name describes the same thing.

UI/window-basic-main.cpp
```diff
@@ -6648,8 +6651,14 @@ void OBSBasic::BroadcastButtonClicked()
...
+				api->on_event(OBS_FRONTEND_EVENT_BROADCAST_START_FAILED);
+
 				return;
-			}
+			} else {
+				// broadcast started successfully event
+				api->on_event(OBS_FRONTEND_EVENT_BROADCAST_STARTED);
+				}
```
norihiro, line 6661:
```suggestion
			}
```

UI/window-basic-main.cpp
```diff
@@ -6648,8 +6651,14 @@ void OBSBasic::BroadcastButtonClicked()
 					QTStr("Output.BroadcastStartFailed"),
 					last_error, true);
 				ui->broadcastButton->setChecked(false);
+				// broadcast start failed event
+				api->on_event(OBS_FRONTEND_EVENT_BROADCAST_START_FAILED);
+
 				return;
-			}
+			} else {
+				// broadcast started successfully event
```
norihiro, line 6659:
```suggestion
```

UI/obs-frontend-api/obs-frontend-api.h
```diff
@@ -60,6 +60,9 @@ enum obs_frontend_event {
 	OBS_FRONTEND_EVENT_SCRIPTING_SHUTDOWN,
 	OBS_FRONTEND_EVENT_PROFILE_RENAMED,
 	OBS_FRONTEND_EVENT_SCENE_COLLECTION_RENAMED,
+	OBS_FRONTEND_EVENT_BROADCAST_STARTED,
+	OBS_FRONTEND_EVENT_BROADCAST_STARTING,
```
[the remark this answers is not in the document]
dcmouser: Good idea; I will swap them.

UI/window-basic-main.cpp
dcmouser: changed and submitted.

UI/window-basic-main.cpp
dcmouser: changed and submitted.

UI/window-basic-main.cpp
dcmouser: changed and submitted.

dcmouser:
Added documentation to docs/sphinx/reference-frontend-api.rst as suggested.

cg2121:
There should probably be a stopping and stopped event as well, as the streaming and recording events have these, so we are consistent.

UI/window-basic-main.cpp
```diff
@@ -6648,7 +6652,10 @@ void OBSBasic::BroadcastButtonClicked()
 					QTStr("Output.BroadcastStartFailed"),
 					last_error, true);
 				ui->broadcastButton->setChecked(false);
+				api->on_event(OBS_FRONTEND_EVENT_BROADCAST_START_FAILED);
```
notr1ch, line 6655: This will wait until the user confirms the OBSMessageBox dialog before firing the event which is probably not good for an API use case. It might be better to fire this earlier.

dcmouser: very good point; i have pushed the change.

dcmouser:
"There should probably be a stopping and stopped event as well, as the streaming and recording events have these, so we are consistent." <-- This would make intuitive sense, but it felt to me that there wasn't any real difference between stopping a broadcast and stopping streaming.  Unlike the starting case, where broadcast can start much later than streaming, in the broadcasting case, when you stop broadcasting you stop streaming (as far as i can tell).  However I would be happy to add the BroadcastStopping and BroadcastStopped events if that was the only thing holding up this code.

PatTheMav:
I agree that if events specific to a broadcast lifecycle are added that we should add all of them (and not just the ones related to starting a broadcast).

Warchamp7:

PR #7553, opened 2022-10-07
mac-capture: Log CoreAudio device sample rate

gxalpha:
### Description

Adds the sample rate of an audio input/output capture device to the log file.
image

### Motivation and Context

Requested for support purposes.
Useful to easily see sample rate mismatches.

### How Has This Been Tested?

macOS 13 Beta 10
Initialized different devices, saw the sample rate being logged (see above).

### Types of changes

### Checklist:

### DIFF
--- a/plugins/mac-capture/mac-audio.c
+++ b/plugins/mac-capture/mac-audio.c
@@ -629,7 +629,8 @@ static bool coreaudio_init(struct coreaudio_data *ca)
 	if (!coreaudio_start(ca))
 		goto fail;
 
-	blog(LOG_INFO, "coreaudio: device '%s' initialized", ca->device_name);
+	blog(LOG_INFO, "coreaudio: Device '%s' [%" PRIu32 " Hz] initialized",
+	     ca->device_name, ca->sample_rate);
 	return ca->au_initialized;
 
 fail:


plugins/mac-capture/mac-audio.c
```diff
@@ -629,7 +629,8 @@ static bool coreaudio_init(struct coreaudio_data *ca)
 	if (!coreaudio_start(ca))
 		goto fail;
 
-	blog(LOG_INFO, "coreaudio: device '%s' initialized", ca->device_name);
+	blog(LOG_INFO, "coreaudio: Device '%s' [%d Hz] initialized",
+	     ca->device_name, ca->sample_rate);
```
jp9000, line 633: Include <inttypes.h>, then change string to:
`"coreaudio: Device '%s' [%" PRIu32 " Hz] initialized"`
instead

PatTheMav: `PRIu32` resolves to `u` on macOS - given that mac-capture is macOS only, I don't think we necessarily need a portable macro?

gxalpha: Fixed

jp9000: Sorry about that, already changed though so either way is fine I suppose. I'll mark as resolved for now

PR #7554, opened 2022-10-07
rtmp-services: Add Vindral service

mattiaslandin:
Adding Vindral rtmp endpoints to streaming services dropdown, wrapped with the following changes:

- Adding active Vindral rtmp endpoints to **services.json** file.
- Incrementing rtmp-services version in **package.json** file.

### Description
Updating the rtmp-services ingestion list to introduce Vindral as a service

### Motivation and Context
Vindral Live CDN delivers next-generation live streaming at sub-second latency, synchronized playout, and 4K video quality at a global scale.

[https://www.vindral.com/](https://www.vindral.com/)
[https://docs.vindral.com/docs/vindral-cdn/](https://docs.vindral.com/docs/vindral-cdn/)

### How Has This Been Tested?
Took the updated JSON and had it validated at jsonlint.com.
Build testing not required.

### Types of changes

### Checklist:

### DIFF
--- a/plugins/rtmp-services/data/services.json
+++ b/plugins/rtmp-services/data/services.json
@@ -2525,6 +2525,27 @@
                 "keyint": 2,
                 "max video bitrate": 16000
             }
+        },
+        {
+            "name": "Vindral",
+            "more_info_link": "https://docs.vindral.com/docs/vindral-cdn/",
+            "servers": [
+                {
+                    "name": "eu-north",
+                    "url": "rtmps://rtmp.eu-north.cdn.vindral.com/publish"
+                },
+                {
+                    "name": "eu-west",
+                    "url": "rtmps://rtmp.eu-west.cdn.vindral.com/publish"
+                }
+            ],
+            "recommended": {
+                "keyint": 1,
+                "profile": "high",
+                "bframes": 0,
+                "max video bitrate": 20000,
+                "max audio bitrate": 192
+            }
         }
     ]
 }
--- a/plugins/rtmp-services/data/package.json
+++ b/plugins/rtmp-services/data/package.json
@@ -1,11 +1,11 @@
 {
     "$schema": "schema/package-schema.json",
     "url": "https://obsproject.com/obs2_update/rtmp-services",
-    "version": 210,
+    "version": 211,
     "files": [
         {
             "name": "services.json",
-            "version": 210
+            "version": 211
         }
     ]
 }


plugins/rtmp-services/data/services.json
```diff
@@ -2484,6 +2484,32 @@
...
+                "h264"
+            ],
+            "supported audio codecs": [
+                "aac"
+            ],
+            "recommended": {
+                "keyint": 1,
+                "profile": "baseline",
+                "bframes": 0,
+                "output": "rtmp_output"
```
tytan652, line 2511: Supported codecs and output are not needed for RTMP services.

```suggestion
            "recommended": {
                "keyint": 1,
                "profile": "baseline",
                "bframes": 0,
```

tytan652 (changes requested):

plugins/rtmp-services/data/services.json
```diff
@@ -2484,6 +2484,26 @@
...
+                    "url": "rtmps://rtmp.eu-north.cdn.vindral.com/publish"
+                },
+                {
+                    "name": "eu-west",
+                    "url": "rtmps://rtmp.eu-west.cdn.vindral.com/publish"
+                }
+            ],
+            "recommended": {
+                "keyint": 1,
+                "profile": "baseline",
```
derrod, line 2503: Are you sure about this? The quality hit from limiting to baseline is fairly considerable.

Additionally, you may want to consider setting a maximum audio and video bitrate.

mattiaslandin: Many of our customers still target devices that only support baseline. As long as the option to select a different profile exists we believe this is the best setting for us.

We don't have limits for max video and audio bitrate but added high numbers for sanity check

plugins/rtmp-services/data/services.json
```diff
@@ -2525,6 +2525,28 @@
...
+                {
+                    "name": "eu-west",
+                    "url": "rtmps://rtmp.eu-west.cdn.vindral.com/publish"
+                }
+            ],
+            "recommended": {
+                "keyint": 1,
+                "profile": "high",
+                "bframes": 0,
+                "output": "rtmp_output",
```
tytan652, line 2546: I think you missed it, so I reiterate that "output" is not needed if the protocol is RTMP(S).

```suggestion
```

plugins/rtmp-services/data/services.json
mattiaslandin: I understand, in our case it's not really misguided. We have cases where `baseline` is preferred. But after some more consideration we've decided to go with `high` instead, for the same reasons that you've already stated

plugins/rtmp-services/data/services.json
mattiaslandin: Yes I did, removed

derrod:
Other than squashing the commits I do not believe there's anything left to do.

mattiaslandin:
> Other than squashing the commits I do not believe there's anything left to do.

Great! Squashing complete

PR #7557, opened 2022-10-07
UI: Shortcut filter refactor to fix numerous memory leaks

notr1ch:
### Description
This PR refactors the shortcut filter implementation to create a single, public `OBSEventFilter` instance for the entire application instead of relying on every widget to create and free its own shortcut filter. An event filter can be shared provided it's created in the same thread, and since these are all UI operations they should all be executing in the Qt UI thread.

### Motivation and Context
Many `CreateShortcutFilter` calls were not tracked and would leak memory.
```
Indirect leak of 96 byte(s) in 2 object(s) allocated from:
    #0 0x55fe148a765d in operator new(unsigned long) (/home/r1ch/obs-studio/build/rundir/RelWithDebInfo/bin/64bit/obs+0x21365d) (BuildId: 361227000132ec995ad483aef4b23909e4e88dc4)
    #1 0x55fe148e4f0a in CreateShortcutFilter() /home/r1ch/obs-studio/UI/obs-app.cpp:115:9
    #2 0x55fe14c90c82 in OBSBasicProperties::OBSBasicProperties(QWidget*, OBSSafeRef<obs_source*, &(obs_source_get_ref), &(obs_source_release)>) /home/r1ch/obs-studio/UI/window-basic-properties.cpp:91:21
    #3 0x55fe14b13f4e in OBSBasic::CreatePropertiesWindow(obs_source*) /home/r1ch/obs-studio/UI/window-basic-main.cpp:2849:19
    #4 0x7f03bd1a9d7c in QWidget::event(QEvent*) (/lib/x86_64-linux-gnu/libQt6Widgets.so.6+0x1d9d7c) (BuildId: 1b8a809c3e427a6488ddc759bd04d21b53094660)
```

### How Has This Been Tested?
Compiled and ran OBS, verified filter behaved as intended.

### Types of changes

### Checklist:


UI/obs-app.cpp
```diff
@@ -1294,6 +1296,8 @@ OBSApp::~OBSApp()
 	os_inhibit_sleep_set_active(sleepInhibitor, false);
 	os_inhibit_sleep_destroy(sleepInhibitor);
 
+	delete shortcutFilter;
```
cg2121, line 1299: Probably a good idea to use QScopedPointer, so we don't have to manually delete the shortcut filter.

notr1ch:
My reasons for keeping this draft at the moment:
- Is having the `shortcutFilter` on `OBSApp` OK?
- Is there a better way to do this, e.g. with the global event hook we use for display affinity, so it also applies to windows and dialogs created by plugins?

UI/obs-app.cpp
notr1ch: Not really sure how to do this, do you have an example? I tried `QScopedPointer<QObject> shortcutFilter;` which seemed correct from the documentation but it does not compile. It works if I explicitly call `shortcutFilter.data());` but according to the docs that shouldn't be necessary?
```
error C2664: 'void QObject::installEventFilter(QObject *)': cannot convert argument 1 from 'QScopedPointer<QObject,QScopedPointerDeleter<T>>' to 'QObject *'
```

notr1ch:
Undrafting to  encourage feedback :).

gxalpha:
> Is there a better way to do this, e.g. with the global event hook we use for display affinity, so it also applies to windows and dialogs created by plugins?

I think it might be possible to just install an event filter on OBSApp (e.g., in the OBSApp constructor)?
Question is if/how that would interfere with widget-level filters that we still use in some places.

UI/obs-app.cpp
PatTheMav: Afaik that's necessary - you either have to use `data()`, `take()`, or you pass a reference to the pointer (which would be `installEventFilter(&App()->shortcutFilter);`) when using a `QScopedPointer`.

PatTheMav:
IMO a single global event filter is desirable over a multitude of event filters all doing their own thing without any oversight (also makes debugging events a chore).

I think this change is a good enough step in the right direction and if this doesn't break in major ways, we could extend it to other areas after that.

Warchamp7:

Issue #7287, opened 2022-09-04
OBS Virtual Camera crashes when switching breakout rooms in Zoom.


attaboyBrad:
### Operating System Info

macOS 12

### OBS Studio Version

28.0.0

### OBS Studio Version (Other)

28.0.1

### OBS Studio Log URL

https://obsproject.com/logs/Chwt_hwR_9cbLuE8

### Expected Behavior

I expected to switch from one breakout room to another without issue.

### Current Behavior

The OBS virtual camera froze in Zoom, and in attempting to restart it the virtual camera ceased to show up as an option in Zoom and Zoom fell back to a raw webcam feed.  This has been an issue since the first OBS 28 beta and was mentioned a couple of times in the beta announcement forum thread.

### Steps to Reproduce

1. Open OBS 28
2. Open Zoom (fully updated as of 22-09-04) and start a meeting.
3. Start virtual camera in OBS 28. 
4. Enter a breakout room or try to switch breakout rooms in Zoom.

### Anything else we should know?

This has been happening basically every time I try to enter, or especially to switch, breakout rooms in Zoom since the first OBS 28 beta (still an issue with 28.0.1).

It's not an issue on 27.2.4, and I've reverted back to that for the time being.

attaboyBrad:
I have since retested using a completely blank profile and scene collection and the issue still happens--in fact Zoom crashed outright.  Retested while using cameras other than OBS 28's Virtual Camera and Zoom worked fine so it appears to be very specific to OBS 28 and it's Virtual Camera implementation.

fischernick:
I've been running into the same problem on an M1 MacBook Pro. While using Virtual Camera in Zoom, going into a breakout room will cause zoom to just lock and display the spinner on the mouse icon. The other participant video and audio still works.  

I've also run into related issue where I am in one meeting, leave, and start another meeting, and Zoom disables my video and I'm unable to switch to another video or turn on the video.  If I close / force quit Zoom and rejoin it works.

fabiosantossilva:
Same thing here. My zoom has been crashing randomly since I started using OBS 28.0.1 on my Mac Studio. The virtual camera stops working and zoom becomes unresponsive as soon as I change meetings. Closing zoom and reopening seems to be the only solution for now.

stevevrporter:
Exactly the same issues here.

OBS 28.0.1
Zoom Version: 5.11.11 (10514)
OSX Version 12.6
2021 Macbook pro, Apple M1 Max

[...]

PR #7581, opened 2022-10-10
mac-virtualcam: Fix crashes in apps loading virtual camera and performance fixes

PatTheMav:
### Description
Fixes crashes that could happen in applications that load the virtual camera multiple times consecutively or in parallel and improves performance on M1-based Macs.

### Motivation and Context
Without invalidating the mach port used for sharing the IOSurface between OBS and the application displaying the virtual camera output, IOKit seems to run into the issue of receiving "shared" mach ports, possibly because of port exhaustion. IOKit requires a "new" port however and crashes upon that error otherwise.

Fixes https://github.com/obsproject/obs-studio/issues/7287

Additionally, because Apple Silicon-based Macs have a unified memory architecture, an IOSurface will always be available in memory accessible to the CPU and GPU (and an off-load of the IOSurface will not take place).

eGPUs are not supported on Apple Silicon-based Macs either, so an IOSurface lock to ensure data is copied back to CPU memory is not necessary.

### How Has This Been Tested?
Tested on macOS 12.6 with OBS itself and using Cameo to load and initialise the virtual camera plugin multiple times.

### Types of changes

### Checklist:

### DIFF
--- a/plugins/mac-virtualcam/src/dal-plugin/OBSDALMachClient.mm
+++ b/plugins/mac-virtualcam/src/dal-plugin/OBSDALMachClient.mm
@@ -10,7 +10,6 @@
 #import "Logging.h"
 
 @interface OBSDALMachClient () <NSPortDelegate> {
-	uint32_t _seed;
 	NSPort *_receivePort;
 }
 @end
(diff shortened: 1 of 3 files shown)


sa1:
In my limited testing, I wasn't able to crash it with breakout rooms or video configuration screen in Zoom. Seems like it works.

steven-michaud:
When is OBS Studio 29.0 expected to be released?

If (as it seems) that's not going to be for a while, would you consider landing the crash fix (https://github.com/obsproject/obs-studio/pull/7581/commits/9eb1b7c5bbc381b814ab2f387f27b7df69adb32f), at least, in an earlier release?

It's very simple, and unlikely to cause trouble.

steven-michaud:
Thanks, @gxalpha, for changing the target milestone to 28.1! When can we expect that to be released? General timeframe only.

gxalpha:
Very soon, in fact a first beta (without this fix) is already out. That’s the reason why originally it didn’t look like it could make it.
However, it was decided that *if* the fix works reliably enough and doesn’t cause any new issues (and this has yet to undergo thorough testing, as far as I know) it could be included.

attaboyBrad:
> However, it was decided that _if_ the fix works reliably enough and doesn’t cause any new issues (and this has yet to undergo thorough testing, as far as I know) it could be included.

Is there testing I can assist with as a user? I've been trying to break this version and haven't had a hiccup.  Not sure if there's a 'proper' way to report those results, though I see the "seeking testers" tag.

I've also dropped the fixed mac-virtualcam plugin from this version into 28.0.3 and the regular 28.1 Beta 1 versions (installed via Steam) and it seemed to work a treat.

RytoEX:
To clarify, only the crash fix commit from this will be shipped with OBS Studio 28.1.0. The other optimizations for Apple Silicon will be targeted for the release after that (likely OBS Studio 29).

PR #7584, opened 2022-10-11
obs-frontend-api: Add audio mixer functions

cg2121:
### Description
Adds functions to the frontend api to add and remove sources from the audio mixer

### Motivation and Context
In one of my plugins, the source is private, so these functions allow me to add it to the audio mixer.

### How Has This Been Tested?
Tested with my plugin.

### Types of changes

### Checklist:

### DIFF
--- a/docs/sphinx/reference-frontend-api.rst
+++ b/docs/sphinx/reference-frontend-api.rst
@@ -913,3 +913,19 @@ Functions
                       This uses the undo action from the first and the redo action from the last action.
 
    .. versionadded:: 29.1
+
+---------------------------------------
+
+.. function:: void obs_frontend_audio_mixer_add_source(obs_source_t *source)
+
+   Adds a source to the audio mixer.
+
+   :param source: The source to add to the mixer
+
+---------------------------------------
+
+.. function:: void obs_frontend_audio_mixer_remove_source(obs_source_t *source)
+
+   Removes a source from the audio mixer.
+
+   :param source: The source to remove from the mixer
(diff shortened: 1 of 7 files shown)


tytan652:
I think switching the view style of the mixer may remove added sources through this method.

cg2121:
> I think switching the view style (vertical to horizontal and vice versa) of the mixer may remove added sources through this method.

That works as expected. The only problem was that it wouldn't work when unhiding all audio sources. The latest update to this PR fixes that.

Fenrirthviti:
> That works as expected. The only problem was that it wouldn't work when unhiding all audio sources. The latest update to this PR fixes that.

Can you confirm if this means the sources are persisted, or removed? This comment is not clear on what the actual result is. I would hope that changing modes does not remove sources added using this API.

cg2121:
The items added in the mixer are persistent.

CodeYan01:
General question: What happens when you name your private source the same as an existing user-added source? And add it to the mixer?

docs/sphinx/reference-frontend-api.rst
```diff
@@ -818,3 +818,19 @@ Functions
...
+
+   Adds a source to the audio mixer.
+
+   :param source: The source to add to the mixer
+
+---------------------------------------
+
+.. function:: void obs_frontend_audio_mixer_remove_source(obs_source_t *source)
+
+   Removes a source to the audio mixer.
```
CodeYan01, line 834:
```suggestion
   Removes a source from the audio mixer.
```

cg2121:
I've found out that there is a function called obs_source_set_hidden, that allows me to disable saving the source in the main ui, so I can save it in my plugin instead. This means I no longer need a private source and my source can show up in the audio mixer, so this PR is no longer needed.

Issue #7592, opened 2022-10-12
Enhancement: option to Hide Transition widgets in Studio Mode


kamalmostafa:
### Operating System Info

Other

### Other OS

all

### OBS Studio Version

28.0.3

### OBS Studio Log URL

n/a

### Expected Behavior

New Settings option: General -> Studio Mode -> **Hide Transition widgets**

When checked, the Transition button and associated widgets are hidden, allowing the side-by-side Preview and Program video windows to display as large as possible.

Note: When the Transition widgets are hidden, transitions can still be invoked by the "Transition to scene when double-clicked" method, websocket, or scripting methods.

![obs-HideTransitionWidgets](https://user-images.githubusercontent.com/1181534/195223128-56f96dd8-12b7-4a96-beaa-4807dbf18a36.png)

### Current Behavior

By default, the placement of the Transition widgets consumes quite a bit of screen real estate.  For some workflows, the Transition widgets might be unnecessary, and the screen space better used for bigger video windows.

The default layout (without enabling the new feature) looks like this for comparison:

![obs-default](https://user-images.githubusercontent.com/1181534/195223667-be1b3ebf-4f4c-4910-830e-248b7b203911.png)

### Steps to Reproduce

1.
2.
3.
...

RytoEX:
We do not track feature requests on GitHub Issues. Please submit feature requests on [our Ideas site](https://ideas.obsproject.com/) after confirming that it has not already been submitted there. Thank you for your understanding.

PR #7593, opened 2022-10-12
UI: Add option to "Hide Transitions widgets" in Studio Mode

kamalmostafa:
### Description

New Settings option: General -> Studio Mode -> Hide Transition widgets

When checked, the Transition button and associated widgets are hidden, allowing the side-by-side Preview and Program video windows to display as large as possible.

Note: When the Transition widgets are hidden, transitions can still be invoked by the "Transition to scene when double-clicked" method, websocket, or scripting methods.

obs-HideTransitionWidgets

### Motivation and Context

https://ideas.obsproject.com/posts/1129/studio-mode-ability-to-hide-transition-panel

Closes obsproject/obs-studio#7592

### How Has This Been Tested?

Functional testing of the option switch behavior on Ubuntu 22.04.

### Types of changes

### Checklist:

### DIFF
--- a/UI/window-basic-main.cpp
+++ b/UI/window-basic-main.cpp
@@ -4373,6 +4373,11 @@ void OBSBasic::ResetUI()
 		ui->previewLayout->setDirection(QBoxLayout::LeftToRight);
 
 	UpdatePreviewProgramIndicators();
+
+	bool hideTransitionWidget = config_get_bool(
+		App()->GlobalConfig(), "BasicWindow", "HideTransitionWidget");
+	if (programOptions)
+		programOptions->setVisible(!hideTransitionWidget);
 }
 
 int OBSBasic::ResetVideo()
--- a/UI/window-basic-main.cpp
+++ b/UI/window-basic-main.cpp
@@ -311,6 +311,11 @@ OBSBasic::OBSBasic(QWidget *parent)
 
 	ui->scenes->setItemDelegate(new SceneRenameDelegate(ui->scenes));
 
+	bool hideTransitionWidget = config_get_bool(
+		App()->GlobalConfig(), "BasicWindow", "HideTransitionWidget");
+	ui->hiddenTransitionsLabel->setVisible(previewProgramMode &&
+					       hideTransitionWidget);
+
 	auto displayResize = [this]() {
 		struct obs_video_info ovi;
 
@@ -4378,6 +4383,8 @@ void OBSBasic::ResetUI()
 		App()->GlobalConfig(), "BasicWindow", "HideTransitionWidget");
 	if (programOptions)
 		programOptions->setVisible(!hideTransitionWidget);
+	ui->hiddenTransitionsLabel->setVisible(previewProgramMode &&
+					       hideTransitionWidget);
 }
 
 int OBSBasic::ResetVideo()
--- a/UI/data/locale/en-US.ini
+++ b/UI/data/locale/en-US.ini
@@ -866,6 +866,7 @@ Basic.Settings.General.OverflowSelectionHidden="Show overflow even when source i
 Basic.Settings.General.Importers="Importers"
 Basic.Settings.General.AutomaticCollectionSearch="Search known locations for scene collections when importing"
 Basic.Settings.General.SwitchOnDoubleClick="Transition to scene when double-clicked"
+Basic.Settings.General.HideTransitionWidget="Hide Transition widgets"
 Basic.Settings.General.StudioPortraitLayout="Enable portrait/vertical layout"
 Basic.Settings.General.TogglePreviewProgramLabels="Show preview/program labels"
 Basic.Settings.General.Multiview="Multiview"
--- a/UI/forms/OBSBasic.ui
+++ b/UI/forms/OBSBasic.ui
@@ -440,6 +440,22 @@
             </property>
            </widget>
           </item>
+	  <item>
+	   <widget class="QLabel" name="hiddenTransitionsLabel">
+	    <property name="sizePolicy">
+	     <sizepolicy hsizetype="Expanding" vsizetype="Fixed">
+	      <horstretch>0</horstretch>
+	      <verstretch>0</verstretch>
+	     </sizepolicy>
+	    </property>
+	    <property name="text">
+	     <string>Note that "Hide Transitions widget" is enabled in Settings</string>
+	    </property>
+	    <property name="alignment">
+	     <set>Qt::AlignLeading|Qt::AlignRight</set>
+	    </property>
+	   </widget>
+	  </item>
          </layout>
         </widget>
        </item>
--- a/UI/forms/OBSBasicSettings.ui
+++ b/UI/forms/OBSBasicSettings.ui
@@ -723,6 +723,13 @@
                      </property>
                     </widget>
                    </item>
+                   <item row="1" column="1">
+                    <widget class="QCheckBox" name="hideTransitionWidget">
+                     <property name="text">
+                      <string>Basic.Settings.General.HideTransitionWidget</string>
+                     </property>
+                    </widget>
+                   </item>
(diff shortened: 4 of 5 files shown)


Fenrirthviti:
Please correct the PR description to properly fill out the template.

kamalmostafa:
- properly fill out the template
 - fixed clang-format
 - rebased

cg2121:
This should be in the view menu not the settings.

kamalmostafa:
> This should be in the view menu not the settings.

I think it's better in Settings -> General -> Studio Mode, because:

- This is presumably a quite rare thing to want to enable, only suitable for a small number of (advanced) users.  Why clutter the View menu with it?

- It is *only* relevant for Studio Mode, and could easily confuse a (new) user who might select the "hide" option if it were in the View menu before ever switching to Studio Mode for the first time -- there would be no visible effect to it.  But later when they first explore Studio Mode, they'll never even see the Transition widgets at all (and have no idea that they're "missing").

- It makes sense to keep this hide switch *right* next to "Transition to scene when double-clicked" (which is in Settings), since many users will want to ensure that the latter is enabled, if that's how they intend to invoke their transitions.  The Studio Mode block in Settings contains other switches which affect both layout and functionality.

Are there stronger reasons why it should be in the View menu?  (Really, whichever 50 lines of framework invoke the single `setVisibility()` call that this all boils down to is fine with me :smile:).

UI/window-basic-main.cpp
```diff
@@ -4373,6 +4373,11 @@ void OBSBasic::ResetUI()
 		ui->previewLayout->setDirection(QBoxLayout::LeftToRight);
 
 	UpdatePreviewProgramIndicators();
+
+	bool hideTransitionWidget = config_get_bool(
+		App()->GlobalConfig(), "BasicWindow", "hideTransitionWidget");
```
tytan652, line 4378: Mismatch with others.
```suggestion
		App()->GlobalConfig(), "BasicWindow", "HideTransitionWidget");
```

tytan652 (changes requested):

kamalmostafa:
Huh.  I clicked the "Request re-review" button for @tytan652 ... and that seems to have removed the review request for @Warchamp7 (that was not my intention).

tytan652 (commented):
What I asked to fix on my first review was fixed.

Fenrirthviti:
Can we have a clear problem statement on what exact problem this is solving? Is this simply a space issue, or is this allowing a workflow that doesn't otherwise exist? Have you considered using a multi-view projector for a cleaner view, and popping out the required docks for a better/more compact setup with a similar end result?

As it stands, I have one fundamental issue with this solution, which is that when you hide the transition controls (as mentioned) you lose functionality in the program unless you know hidden information, which is a really bad situation to be in. I agree that putting it in the Settings menu means it's less likely that someone will accidentally turn the transition controls off, but this is still a concern.

I just want to make sure we fully understand what the problem is, and that this is the correct solution to that problem as it's a little ambiguous right now.

As an aside, is this something that has always been an issue, or is it only now an issue due to the new default theme?

kamalmostafa:
Thanks for your review @Fenrirthviti, 

> Can we have a clear problem statement on what exact problem this is solving? Is this simply a space issue, or is this allowing a workflow that doesn't otherwise exist?

This is purely a "space issue"; no new workflow is enabled.  The purpose is to make better use of the screen real-estate "wasted" by the Transition widget panel, for setups where those buttons are never used (like mine, where all transition control is by websocket).  I'd much rather have bigger Preview/Program video windows than a bunch of buttons that I never touch.  Presumably, the other commenters in the ideas.obsproject.com topic have similar reasons for wanting to hide them.
 
> Have you considered using a multi-view projector for a cleaner view, and popping out the required docks for a better/more compact setup with a similar end result?

I experimented with this, but it did not provide a better layout for me.  I note that there isn't actually a 2x1 "just-Preview/Program" multi-view panel.  And I don't particularly want the video windows detached from the main window anyway.

On the topic of pop-out docks though ... Another idea mentioned in the ideas.obsproject.com topic was to make the Transition widgets panel a detachable dock, and that would also allow for my desired layout, of course.  (The "Scene Transitions" detachable dock even looks awfully similar to the widget panel.  Hmmm.).

> As it stands, I have one fundamental issue with this solution, which is that when you hide the transition controls (as mentioned) you lose functionality in the program unless you know hidden information, which is a really bad situation to be in. I agree that putting it in the Settings menu means it's less likely that someone will accidentally turn the transition controls off, but this is still a concern.

A fair concern, though I'm not sure this switch is any "more dangerous" than myriad other settings checkboxes.

That said, I've added another commit [8cee669](https://github.com/obsproject/obs-studio/pull/7593/commits/8cee669ac931acb5835af480bf35aecd77e5fb27) which adds an indicator label "Note" whenever the hidden Transisition widgets are being hidden.  I'm not in love with it, but it does address the "hidden information" concern.  Meh?

obs-HiddenTransitionWidgetsLabel

 
> I just want to make sure we fully understand what the problem is, and that this is the correct solution to that problem as it's a little ambiguous right now.
> 
> As an aside, is this something that has always been an issue, or is it only now an issue due to the new default theme?

This is not related to the new theme; I carried a hack version of this in v27 too and the ideas.obsproject.com topic is years old.

Fenrirthviti:
> This is purely a "space issue"; no new workflow is enabled. The purpose is to make better use of the screen real-estate "wasted" by the Transition widget panel, for setups where those buttons are never used (like mine, where all transition control is by websocket). I'd much rather have bigger Preview/Program video windows than a bunch of buttons that I never touch. Presumably, the other commenters in the ideas.obsproject.com topic have similar reasons for wanting to hide them.

Thanks for the confirmation here.

> A fair concern, though I'm not sure this switch is any "more dangerous" than myriad other settings checkboxes.

I don't want sins of the past informing future decisions, though I'd be curious what you're referring to as we're making a concerted effort to reduce the amount of dangerous or hidden options. This is probably better suited off-thread here though, I'd definitely love to hear your thoughts on our Discord, if you ping me (Fenrir) in the #development channel.
 
> That said, I've added another commit [8cee669](https://github.com/obsproject/obs-studio/pull/7593/commits/8cee669ac931acb5835af480bf35aecd77e5fb27) which adds an indicator label "Note" whenever the hidden Transisition widgets are being hidden. I'm not in love with it, but it does address the "hidden information" concern. Meh?
> 
> image snipped

This uses the space reserved for the source control bar, so not a viable solution unfortunately.

In general, I'm unhappy with this as a solution as I feel it's not exactly the correct way to solve the problem. Internally (and in some public places I don't have handy) we've discussed changing the transition settings to a separate dialog where you can configure and set up transitions, set a matrix for to/from scenes for individual transitions, manage source show/hide transitions, etc., and while that hasn't been a high priority for the project as a whole, it's something that might alleviate some of the issues here.

However, the transition controls themselves are something that are fundamental to how Studio Mode works, and information that shouldn't ever be hidden, IMO. This is just my opinion, so don't take this as a rejection or that this won't be merged, I just wanted to voice my concerns. I understand that for advanced users who are absolutely sure of what they are doing, it might make sense as they're moving the functionality somewhere else, but we as maintainers need to ask if this helping or beneficial to a majority of the userbase, or if this cause more harm to the average user who doesn't understand what they're doing. This is basically an impossible question to answer as we have no user telemetry, and it's hard to set aside bias and preconceptions about the program as a whole.

@Warchamp7 will also need to weigh in here as the current UI/UX decision maker before a final decision can be made.

Warchamp7:
Sorry for the long delay on responding to this.

I appreciate the desire to remove unused space in certain workflows, but unfortunately this makes Studio Mode non-functional on a fundamental level without having explicitly enabled other settings (Ex. Double click to switch scenes) or other workflow methods (Hotkeys, third-party devices/plugin API calls) and I am not comfortable with that on a UX level.

PR #7594, opened 2022-10-12
added support for custom css in browser docks as in browser sources

dcmouser:
added support for custom css in browser docks as in browser sources

### Description
Browser sources allow the user to specify custom css overrides; this is super useful and has been missing for the browser DOCK (UI dock) elements.  This pull request addes a custom css column to the front end browser dock system.
The browser dock code in OBS is not the most elegant thing, but my changes parallel the original code and do not introduce anything unusual.
The only imperfect aspect of the code is that changing the custom css for an already visible browser dock cannot have any effect until the dock is closed and recreated; refreshing the page does not solve this.  If someone can figure out a clean way to force the browser to reapply the script startup code it would be welcome.

### Motivation and Context
OBS forum post here: https://obsproject.com/forum/threads/custom-css-for-browser-docks-like-browser-sources.155735
As an example of how and why one might use this, is to create a youtube studio dock showing the users youtube studio page/account.  But such pages are filled with extraneous information that makes them poorly suited to show in a small dock.  With this mod it is easy to customize the css to show a compact version of these pages which have stats on the current stream, etc.  I will post some sample custom css for these usecases.

### How Has This Been Tested?
Tested in OBS 28, tested by setting complicated custom css for youtube pages in docks.

### Types of changes

### Checklist:

- [X ] My code is not on the master branch.
- [X ] The code has been tested.

### DIFF
--- a/UI/window-basic-main.hpp
+++ b/UI/window-basic-main.hpp
@@ -546,13 +546,14 @@ private:
 	QList<QSharedPointer<QDockWidget>> extraBrowserDocks;
 	QList<QSharedPointer<QAction>> extraBrowserDockActions;
 	QStringList extraBrowserDockTargets;
+	QStringList extraBrowserDockCustomCss;
 
 	void ClearExtraBrowserDocks();
 	void LoadExtraBrowserDocks();
 	void SaveExtraBrowserDocks();
 	void ManageExtraBrowserDocks();
 	void AddExtraBrowserDock(const QString &title, const QString &url,
-				 const QString &uuid, bool firstCreate);
+				 const QString &uuid, const QString &customCss, bool firstCreate);
 #endif
 
 	QIcon imageIcon;
--- a/UI/window-extra-browsers.cpp
+++ b/UI/window-extra-browsers.cpp
@@ -18,6 +18,7 @@ using namespace json11;
 enum class Column : int {
 	Title,
 	Url,
+	CustomCss,
 	Delete,
 
 	Count,
@@ -39,6 +40,7 @@ void ExtraBrowsersModel::Reset()
 		item.prevIdx = i;
 		item.title = dock->windowTitle();
 		item.url = main->extraBrowserDockTargets[i];
+		item.customCss = main->extraBrowserDockCustomCss[i];
 		items.push_back(item);
 	}
 }
@@ -71,6 +73,8 @@ QVariant ExtraBrowsersModel::data(const QModelIndex &index, int role) const
 			return items[idx].title;
 		case (int)Column::Url:
 			return items[idx].url;
+		case (int)Column::CustomCss:
+			return items[idx].customCss;
 		}
 	} else if (idx == count) {
 		switch (column) {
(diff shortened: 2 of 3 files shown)


dcmouser:
Screenshot:
browseDockCustomCss

dcmouser:
Video demonstration:
https://www.youtube.com/watch?v=tQUEz2gGCfo

WizardCM:
This implementation has the following flaws:

* It makes the table of Custom URLs harder to read/browse
* It provides an unpleasant single line to squeeze all CSS into
* It does not provide a way to set the CSS for a built in browser dock, such as Twitch chat

Personally, I would want the implementation be done via a "Configure" option that opens a dialog from the right click menu of a dock using `OnBeforeContextMenu`.

https://github.com/obsproject/obs-browser/blob/d1fa35dcbc384136e9e883f2d6071b14fd8f9a65/panel/browser-panel-client.cpp#L231

dcmouser:
Well that's unfortunate.  I get that you would like a different solution - I'd be happy to use your proposed hypothetical solution over my working one.  But rejecting a functionally working useful feature because the custom browser configuration list looks a little cramped is a bit demoralizing...

WizardCM:
It's definitely a useful feature and one that has been often requested (and that we'd like to see) but I expect most of the code would go into obs-browser instead of obs-studio. I can reopen this particular PR if I am incorrect and changes need to be applied to both repositories.

Fenrirthviti:
Just to comment on this not being accepted, and provide some context on to why this isn't being accepted even though it's "working".

When we accept a PR, we are accepting responsibility for that code. Both the code quality, and the resulting user experience of that code. This PR, while presumably functional (at the time of closing, it wasn't even building properly on Windows), introduces a lot of negative and undesirable patterns for UI expectations and the UX of our userbase. We have to accept the cost of both maintaining this code in all future releases, and supporting users who want to utilize this feature. I understand that your perspective is "working" is better than "not working" in this case, but from our perspective, there's a significant cost to accepting an unideal solution, that we have to bear the consequences of essentially forever. This goes a lot further past being "a little cramped" as the reason why we aren't accepting.

At the end of the day, "this feature works" is not sufficient enough of a justification for us to merge *anything*. It is also not personal, so please don't take it as such.

dcmouser:
I appreciate the responsibilities and complications of maintaining such a large codebase with so many contributors.  And you're doing something right because OBS is a wonderful, incredible piece of software.  And I won't take it personally.  Having said all that, trying to contribute has been a sufficiently unpleasant and unfruitful exercise, and it's clear I'm going to have to do my own private builds for the forseeable future.  So for now I'll worry about streamlining that workflow, and not mess around with trying to submit code.

kurodz:
I really need this. Please make it happen 🥲

REFLEXGAMING007:
Someone please implement this!!!

FerLuisxd:
I really need this too!!

Weltraumeule:
Please can you tell my, how I can use this? I need custom css :( Every time I start OBS, I open the inspector and paste my css code, it is so annoying!

phillypro:
lol here we are a year later

laurentlasalle:
This feature would be awesome, especially for my multistreaming needs.

flleeppyy:
almost 2 years later...

PR #7600, opened 2022-10-13
UI: Swap mq and hq NVENC Preset mappings

RytoEX:
### Description

The old preset "mq" was "Max Quality", which used the old HQ preset with two-pass enabled. The old preset "hq" was "Quality", which used the old HQ preset without two-pass. Of the two, "mq" would have been considered "slower" or "higher quality" than "hq", so let's swap these entries.

Note: There is perhaps some argument to be made that "mq" and "hq" should simply both map to P4/P5 with different multipass settings, but let's get the simple fix out of the way before getting stuck on other items.

Reference:
https://docs.nvidia.com/video-technologies/video-codec-sdk/nvenc-preset-migration-guide/index.html

### Motivation and Context

Want more accurate migrations.

### How Has This Been Tested?

Checked locally that recording encoder settings are migrated as follows on application start:
* hq -> p4
* mq -> p6

### Types of changes

### Checklist:

### DIFF
--- a/UI/obs-app.cpp
+++ b/UI/obs-app.cpp
@@ -2821,12 +2821,12 @@ static void convert_28_1_encoder_setting(const char *encoder, const char *file)
 			const char *preset =
 				obs_data_get_string(data, "preset");
 
-			if (astrcmpi(preset, "hq") == 0) {
+			if (astrcmpi(preset, "mq") == 0) {
 				obs_data_set_string(data, "preset2", "p6");
 				obs_data_set_string(data, "tune", "hq");
 				obs_data_set_string(data, "multipass", "qres");
 
-			} else if (astrcmpi(preset, "mq") == 0) {
+			} else if (astrcmpi(preset, "hq") == 0) {
 				obs_data_set_string(data, "preset2", "p4");
 				obs_data_set_string(data, "tune", "hq");
 				obs_data_set_string(data, "multipass", "qres");


Nofew:
For what it's worth, the only reason I used MQ was for smarter b-frame placement, which it's self requires enabling 2-pass, which is what MQ does, to my knowledge. There may be some surprise if p6 with 2-pass enabled takes more resources than MQ did previously.

jp9000:
I thought mq was medium quality

RytoEX:
> For what it's worth, the only reason I used MQ was for smarter b-frame placement, which it's self requires enabling 2-pass, which is what MQ does, to my knowledge. There may be some surprise if p6 with 2-pass enabled takes more resources than MQ did previously.

Again, I don't want to get lost in "what other fixes could we make here", I simply want to first correct what is obviously backwards.

> I thought mq was medium quality

This is a common misconception, but no, it is/was "Max Quality".

Here you see that "mq" is simply the old "hq" with twopass:
* https://github.com/obsproject/obs-studio/blob/28.0.3/plugins/obs-ffmpeg/jim-nvenc.c#L383-L385
* https://github.com/obsproject/obs-studio/blob/28.0.3/plugins/obs-ffmpeg/jim-nvenc.c#L642-L644
* https://github.com/obsproject/obs-studio/blob/28.0.3/plugins/obs-ffmpeg/obs-ffmpeg-nvenc.c#L128-L131

It is above "hq" in this list here:
* https://github.com/obsproject/obs-studio/blob/28.0.3/plugins/obs-ffmpeg/obs-ffmpeg-nvenc.c#L500-L502

PR #7601, opened 2022-10-13
libobs: Enable fast clear on Windows always

jpark37:
### Description
NVIDIA driver 527.37 fixes flickering when multiple GPUs are installed.

### Motivation and Context
Gotta go fast.

### How Has This Been Tested?
Clear color is still visible on Windows and Mac.

Flicker verified gone in 528.49.

### Types of changes

### Checklist:

### DIFF
--- a/libobs/obs-display.c
+++ b/libobs/obs-display.c
@@ -19,21 +19,18 @@
 #include "obs.h"
 #include "obs-internal.h"
 
+#if defined(__APPLE__)
+/* Apple Silicon GL driver doesn't seem to track SRGB clears correctly */
+#define OBS_USE_CLEAR_WORKAROUND 1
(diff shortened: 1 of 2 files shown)


RytoEX:
This really should be gated on the user having the correct driver version. We don't want to intentionally introduce flickering for users who do not have super-recent drivers.

jp9000:
It's annoying but I can probably write a driver gate or something if we need to for the time being.

RytoEX:
Discussed with @Fenrirthviti and @jp9000 . I'm going to back out this change and work on putting a driver version guard in place. Once that work is done, I'll adjust the `display->use_clear_workaround` branches for that guard, so we'll always use fast clear when it's safe.

PR #7603, opened 2022-10-15
CI: Update GitHub Actions for set-output deprecation

RytoEX:
### Description

GitHub Actions has deprecated set-output. Replace usages of set-output in stdout with the new syntax to save the output to the new environment variable.

See:
https://github.blog/changelog/2022-10-11-github-actions-deprecating-save-state-and-set-output-commands/

### Motivation and Context

Don't want deprecation warnings hiding real issues.

### How Has This Been Tested?

It hasn't. This was written by following documentation. A CI run will be the test.

### Types of changes

### Checklist:

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -153,7 +153,7 @@ jobs:
             brew uninstall ${REMOVE_FORMULAS}
           fi
 
-          echo "::set-output name=commitHash::$(git rev-parse --short=9 HEAD)"
+          echo "commitHash=$(git rev-parse --short=9 HEAD)" >> $GITHUB_OUTPUT
 
       - name: 'Switch to Xcode 14.1'
         run: sudo xcode-select -switch /Applications/Xcode_14.1.app
@@ -253,7 +253,7 @@ jobs:
       - name: 'Setup build environment'
         id: setup
         run: |
-          echo "::set-output name=commitHash::$(git rev-parse --short=9 HEAD)"
+          echo "commitHash=$(git rev-parse --short=9 HEAD)" >> $GITHUB_OUTPUT
 
       - name: 'Install dependencies'
         env:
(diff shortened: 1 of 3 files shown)


CI/check-services.py
```diff
@@ -323,7 +323,7 @@ def main():
             service_authors = dict()
 
         # set GitHub outputs
-        print(f'::set-output name=make_pr::true')
+        print(f'make_pr=true') >> $GITHUB_OUTPUT
```
derrod, line 326: This is not valid python, annoyingly this change makes it more difficult here. You'd have to write everything to a file.

Perhaps adding a helper for that would work, e.g.
```
def set_output(name, value):
    try:
        with open(os.environ['GITHUB_OUTPUT'], 'a', newline='\n') as f:
            f.write(f'{name}={value}')
    except Exception as e:
        print(f'Writing to github output files failed: {e!r}')
```
And then replace these calls with `set_output('make_pr', 'true')` etc.

RytoEX: Removed the file from this PR to make it less troublesome. Rodney will revisit this point sometime later.

derrod (approved):

PatTheMav (approved):
Looks like a straight-forward change to me.

RytoEX:
Per GitHub documentation, adjusted PowerShell (pwsh / PowerShell Core) syntax to:
```
"{name}=value" >> $env:GITHUB_OUTPUT
```

Running CI without and with "Seeking Testers" to ensure all is well, then will merge.

.github/workflows/main.yml
```diff
@@ -449,7 +449,7 @@ jobs:
           CI/windows/03_package_obs.ps1 -CombinedArchs -Package
 
           $ArtifactName = (Get-ChildItem -filter "obs-studio-*-windows-x86+x64.zip" -File).Name
-          Write-Output "::set-output name=filename::${ArtifactName}"
+          "filename=${ArtifactName}" >> $env:GITHUB_OUTPUT
```
derrod, line 452: I believe this does not add a newline character by default, you should probably end each string with `\n`

RytoEX: The documentation doesn't specify.
See: https://docs.github.com/en/actions/using-workflows/workflow-commands-for-github-actions?tool=powershell#setting-an-output-parameter

PR #7560, opened 2022-10-08
CI: Upload beta builds as Steam Playtest

derrod:
### Description

Uploads pre-release builds to Steam Playtest as well as beta branch.

### Motivation and Context

Want to potentially use playtests in the future to allow side-by-side installation and better/easier discoverability of betas on Steam: https://partner.steamgames.com/doc/features/playtest

### How Has This Been Tested?

Tested on my fork.

### Types of changes

- New feature (non-breaking change which adds functionality)

### Checklist:
- [x] My code has been run through [clang-format](https://github.com/obsproject/obs-studio/blob/master/.clang-format).
- [x] I have read the [**contributing** document](https://github.com/obsproject/obs-studio/blob/master/CONTRIBUTING.rst).
- [x] My code is not on the master branch.
- [x] The code has been tested.
- [x] All commit messages are properly formatted and commits squashed where appropriate.
- [x] I have included updates to all appropriate documentation.

PR #7604, opened 2022-10-16
UI: Hide Donation CTA when launched from Steam

derrod:
### Description

Adds new internal `steam` flag and hides donate link when running under Steam.

### Motivation and Context

Playtest app failed Valve review due to donation link in-app, to make use of it and make #7560 useful just hide the link when launched from Steam.

### How Has This Been Tested?

Hasn't yet, I'm too tired and my PC is too slow to compile before my face will hit the keyboard

### Types of changes

### Checklist:

### DIFF
--- a/UI/obs-app.hpp
+++ b/UI/obs-app.hpp
@@ -240,6 +240,7 @@ static inline int GetProfilePath(char *path, size_t size, const char *file)
 }
 
 extern bool portable_mode;
+extern bool steam;
 
 extern bool opt_start_streaming;
 extern bool opt_start_recording;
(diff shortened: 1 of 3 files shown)


UI/obs-app.hpp
```diff
@@ -240,6 +240,7 @@ static inline int GetProfilePath(char *path, size_t size, const char *file)
 }
 
 extern bool portable_mode;
+extern bool steam;
```
WizardCM, line 243: Would we be better off using something like `hide_contribute`?

WizardCM (commented):
I generally don't love the idea of a launch flag that can hide UI elements almost abitrarily. I understand this is mostly outside of our control though.

PR #7614, opened 2022-10-18
Update Yami.qss and Rachni.qss to remove dotted border in OBS UI

RoadToDream:
### Description

Before:
image

After: 
image

### Motivation and Context

Dotted box on ListWidget doesn't feels consistent with the overall UI design. 
### How Has This Been Tested?

Tested after build. 

### Types of changes

Minor UI modification.

### Checklist:

### DIFF
--- a/UI/data/themes/Rachni.qss
+++ b/UI/data/themes/Rachni.qss
@@ -183,6 +183,7 @@ QMenuBar::item:selected {
 /* Item Lists */
 QListWidget {
     border-radius: 4px;
+    outline: 0px;
 }
 
 QListWidget::item {
@@ -194,6 +195,7 @@ QMenu,
 SceneTree,
 SourceTree {
     padding: 3px;
+    outline: 0px;
 }
 
 QListWidget::item,
--- a/UI/data/themes/Yami.qss
+++ b/UI/data/themes/Yami.qss
@@ -181,7 +181,7 @@ QMenuBar::item:selected {
 /* Item Lists */
 QListWidget {
     border-radius: 4px;
-    outline: 0;
+    outline: 0px;
 }
 
 QListWidget::item {
@@ -193,6 +193,7 @@ QMenu,
 SceneTree,
 SourceTree {
     padding: 3px;
+    outline: 0px;
 }
 
 QListWidget::item,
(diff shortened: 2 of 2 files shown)


Fenrirthviti:
This is actually an accessibility feature, that shows the current keyboard focus. We probably shouldn't remove it.

RoadToDream:
Sorry I was not aware of this consideration. I will close it.

RoadToDream:
Upon re-checking OBS code, this UI choice should not be related to accessibility feature as buttons already highlighted while selected by arrow keys. No need to say, dotted line around buttons is not unified across the app, dotted line around buttons in Scene dock is not behaving the same as the Sources dock. I would say removal of the dotted line will give OBS a cleaner look and not sacrificing functionality from accessibility point of view. 
Correct me if I am wrong. 
image
image

Fenrirthviti:
While the visual inconsistency should definitely be corrected (I agree that the sources dock line looks better here, and not all elements are showing focus), you are misunderstanding the point. It is not showing a selected item, it is showing a focused item. If you click (or use tab to cycle) away from a scene or source, the focus is no longer on that, and now on something else. The dotted line shows current focus. Without it, there is no other way to determine what focus is.

https://pub.rachni.com/vid/obs64_2022-10-28_12-28-56.webm

RoadToDream:
Thank you for the detailed explanation! I now understand the dotted line is useful when there are multiple docks. I didn't get it before as the tab behavior doesn't cover many ui components so I was thinking using tab to navigate is not a designed behavior. 
Like in settings panel. Tab or arrow key will not only indicate selection by a bounding dotted box but also make the selection. Also some object is highlightable via tab at UI element level(Buttons, dropdown), others can only be selected at dock level(Audio Mixer for example.) I guess these inconsistencies make me overlook the intention of this design.

Fenrirthviti:
I think it would probably be better to update parts of the UI where the selection box is inconsistent, and the way that the sources list displays it is far less intrusive/off-looking than the settings window or the scenes list, but if I recall there were issues with getting it to display properly like that in some cases.

I defer to @Warchamp7 on this, but probably better suited for an issue post. I'll create that myself next week, unless you'd like to, as I think it's something that does need to be tracked for accessibility.

RoadToDream:
Thank you for putting efforts in it. Would be great to have a discussion.

Issue #7617, opened 2022-10-19
'Open [VST] Plugin Interface' Button doesn't show up when OBS is installed under 'Applications' but does when installed via Steam


attaboyBrad:
### Operating System Info

macOS 12

### OBS Studio Version

28.1.0-beta1

### OBS Studio Version (Other)

Issue occurs with 28.0.3 as well

### OBS Studio Log URL

https://obsproject.com/logs/XJ9aujHkkrsBAX9_

### Expected Behavior

When I add a 'VST 2.x Plug-In' Audio Filter to an audio source and select the plugin, a button labeled 'Open Plug-In Interface' like that pictured below usually appears.
![Screen Shot 2022-10-18 at 18 12 49](https://user-images.githubusercontent.com/4070613/196582592-3f53364b-ea52-4770-8532-59753f907c9f.png)

### Current Behavior

That button now only appears when I install OBS via Steam, and not when I install OBS via a downloaded installer.  The image from 'Expected Behavior' is from a Steam based install.  The image below is from a regular install--note the absence of said button.
![Screen Shot 2022-10-18 at 18 10 59](https://user-images.githubusercontent.com/4070613/196582562-44845ad0-e2ac-4b2e-9f7a-8cca33100baf.png)

### Steps to Reproduce

1. Launch OBS from an install in '/Applications'
2. Add a 'VST 2.x Plug-In' and select a plugin--I use the ones from Izotope Elements Suite. They're installed in '/Library/Audio/Plug-Ins/VST'
3. Look for the 'Open Plug-In Interface' button.
...

### Anything else we should know?

I've reproduced this issue with both 28.0.3 and 28.1 Beta 1.  When installed via Steam, the button shows up just fine for both versions.  When installed via downloaded dmg, the button fails to appear.  When 27.4.2 is installed via dmg the button shows up.

From the logs, I've noticed the following lines (or similar) appear when installed from dmg and don't appear when running a Steam-based version:

18:55:08.948: User selected new VST plugin: '/Library/Audio/Plug-Ins/VST/Neutron 3 Elements.vst'
18:55:08.949: Couldn't get a pointer to plug-in's main()

I've tried resetting all of the permissions settings I could find.  I tried copy/pasting the .vst files to '~/Library/Audio/Plug-Ins/VST'.  I tried uninstalling and re-installing the VST Plugins.  I tried using an empty profile as well as replacing the whole settings folder.  No joy.

ArmchairGuru:
[...]

PR #7620, opened 2022-10-19
CI: Fix Steam launching x86 version under Rosetta

derrod:
### Description

Apparently Steam running the OBS launch script makes it run inside Rosetta as well, so it ended up launching the x86 version even on arm64 systems.

Explicitly detect Rosetta translation and set arm64 in those cases. Still allow users to override this by specifying `--intel` as a launch parameter.

This wasn't caught until #7617 due to the script working fine when just run from a terminal (and our macOS user numbers on Steam being rather small).

### Motivation and Context

Want to run native version where possible.

### How Has This Been Tested?

Modified script locally before launching through Steam to verify.

### Types of changes

### Checklist:

### DIFF
--- a/CI/steam/scripts_macos/launch.sh
+++ b/CI/steam/scripts_macos/launch.sh
@@ -2,6 +2,17 @@
  
 arch_name="$(uname -m)"
 
+# When the script is launched from Steam, it'll be run through Rosetta.
+# Manually override arch to arm64 in that case.
+if [ "$(sysctl -in sysctl.proc_translated)" = "1" ]; then
+    arch_name="arm64"
+fi
+
+# Allow users to force Rosetta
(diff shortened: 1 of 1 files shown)


derrod:
Note: Since this change is kind of breaking I do not necessarily want to push it to the 28.0 release already on Steam, but 28.1 rc1 and later should have this fix. I will make a note of this in the rc1/release announcement post.

derrod:
Arlgith, tested and verified this works both with zsh and bash and behaves as intended. Merging once CI passes.

RytoEX (approved):
Seems fine to me, and is probably no worse than what is happening now.

RytoEX:
Added to the cherry-pick PR as well.

PR #4741, opened 2021-05-21
UI: Add audio track mixer

cg2121:
### Description
This adds a dock, hidden by default, which shows volume controls for each audio track. Each track can be hidden.

Screenshot from 2022-11-17 19-48-55

### Motivation and Context
IMO, this feature is sorely needed in OBS.

I took the code from #1469, and tried to simplify it as much as possible, so it would be more likely to be merged.

### How Has This Been Tested?
Controlled different audio tracks and used the audio monitor with them to make sure it functions as expected.

### Types of changes
- New feature (non-breaking change which adds functionality)

### Checklist:
- [x] My code has been run through [clang-format](https://github.com/obsproject/obs-studio/blob/master/.clang-format).
- [x] I have read the [**contributing** document](https://github.com/obsproject/obs-studio/blob/master/CONTRIBUTING.rst).
- [x] My code is not on the master branch.
- [x] The code has been tested.
- [x] All commit messages are properly formatted and commits squashed where appropriate.
- [x] I have included updates to all appropriate documentation.

PR #7626, opened 2022-10-20
libobs: Deprecate obs_get/set_master_volume

### LINKED ISSUE #3786: LIBOBS Set Master Volume not working
## Platform

Operating system and version:  Window 10
OBS Studio version: 26.02

## Expected Behavior

A call to **obs_set_master_volume** must set the overall volume level

## Current Behavior

Nothing happens, no change in volume

## Steps to Reproduce

1.  call **obs_set_master_volume** passing any value between 0 and 1
2.
3.
4.

gxalpha:
### Description

Marks the `obs_set_master_volume` and `obs_get_master_volume` as deprecated
as they aren't used by OBS and don't actually do anything.
Removes the associated event (that never gets called, because the function doesn't get called).
Removes the associated variable. The struct is in `obs-internal.h` so this shouldn't break the ABI.

### Motivation and Context

Reported in #3786. We shouldn't have dead functions.
Closes #3786.

### How Has This Been Tested?

macOS 13 RC 1

Code still compiles.
Check that nothing in OBS uses these functions. Checked with git grep that the `master_volume` isn't used anywhere (tbf, it would have been odd if it was used).
Checked that the functions aren't used in #4741 (that PR is per channel)

### Types of changes

### Checklist:

### DIFF
--- a/docs/sphinx/reference-core.rst
+++ b/docs/sphinx/reference-core.rst
@@ -479,13 +479,19 @@ Video, Audio, and Graphics
 
 .. function:: void obs_set_master_volume(float volume)
 
-   Sets the master user volume.
+   No-op, only exists to keep ABI compatibility.
+
+   .. deprecated:: 29.0
 
 ---------------------
 
 .. function:: float obs_get_master_volume(void)
 
-   :return: The master user volume
+   No-op, only exists to keep ABI compatibility.
+
+   :return: Always returns 1
+
+   .. deprecated:: 29.0
 
 ---------------------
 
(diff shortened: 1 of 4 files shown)


PatTheMav:
I'd advocate to not only deprecate it, but also explicitly put in a removal notice (e.g. OBS 30), otherwise we're dragging unused code around with us and any plugin that _might_ actually have it implemented will never remove it (because it never breaks compilation).

cg2121:
I agree with removing these, as the master mixer PR doesn't even use these functions. https://github.com/obsproject/obs-studio/pull/4741

docs/sphinx/reference-core.rst
```diff
@@ -478,13 +478,19 @@ Video, Audio, and Graphics
 
 .. function:: void obs_set_master_volume(float volume)
 
-   Sets the master user volume.
+   No-op, only exists to keep ABI compatibility.
+
```
WizardCM, line 482:
```suggestion
.. deprecated:: 29.0.0
```

Issue #7595, opened 2022-10-12
'Text (FreeType 2)' element only reading first line of file (when 'Read from file' is checked)


raleighlittles:
### Operating System Info

Ubuntu 22.04

### OBS Studio Version

28.0.3

### OBS Studio Log URL

https://obsproject.com/logs/VJ5xkLXGk9ICAlaP

### Expected Behavior

Expected behavior: Each line of the text file would be read, and a multi-line string would be shown.

### Current Behavior

Only the first line of the text file is read, and other lines aren't shown. See screenshot: 

![screenshot](https://i.imgur.com/oFbrTHL.png)

To summarize:

* I have 'Text(FreeType 2)' set to read a text file. The text file has 4 lines. Only the first line of the file is being rendered.

### Steps to Reproduce

1. Create a text file with multiple lines of text.
2. Create a 'Text (FreeType 2)' source element
3. Check "Read from file", and in the 'Text File' field, select your multi-line file.
4. Observe that only one line of output is shown, regardless of how many lines your actual file has.
...

kkartaltepe:
I cannot replicate, provide a sample file.

raleighlittles:
Ah okay I figured out what the actual problem is. When a file is read, OBS is incorrectly treating the null character in a file as the end-of-file indicator.

I ran `cat` with the extended option and saw this:

```
➜ cat -A test.txt 
Manufacturer: EVGA^@^@^@^@^@^@^@^@^@^@^@^@^@^@^@^@^@^@^@^@^@^@^@^@^@^@^@^@^@^@^@^@^@^@^@^@^@^@^@^@^@^@^@^@^@^@^@^@^@^@^@^@^@^@^@^@^@^@^@^@^@^@^@^@^@^@^@^@^@^@^@^@^@^@^@^@^@^@^@^@^@^@^@^@^@^@^@^@^@^@^@^@^@^@^@^@^@^@^@^@^@^@^@^@^@^@^@^@^@^@^@^@^@^@^@^@^@^@^@^@^@^@$
[...]

PR #7622, opened 2022-10-20
text-freetype2: Permit null-chars

adsr:
### Description
Permit null-chars in text plugin.

### Motivation and Context
https://github.com/obsproject/obs-studio/issues/7595 for Hacktoberfest (Requesting `hacktoberfest-accepted` label.)

### How Has This Been Tested?
Tested with and without text containing null chars.

### Types of changes
- Bug fix (non-breaking change which fixes an issue)
- Tweak (non-breaking change to improve existing functionality)

### Checklist:

- [x] My code has been run through [clang-format](https://github.com/obsproject/obs-studio/blob/master/.clang-format).
- [x] I have read the [**contributing** document](https://github.com/obsproject/obs-studio/blob/master/CONTRIBUTING.rst).
- [x] My code is not on the master branch.
- [x] The code has been tested.
- [x] All commit messages are properly formatted and commits squashed where appropriate.
- [x] I have included updates to all appropriate documentation.

PR #7628, opened 2022-10-21
Reading nulls frm txtfiles

SBaori:
### Description
#7595 
Reading text containing NULL chars by skipping over them.

### Motivation and Context
Improve user experience and also a good learning opportunity.

### How Has This Been Tested?
Tested on POP OS 22.04
Text File causing issue taken from https://github.com/obsproject/obs-studio/files/9793734/test.txt

### Types of changes

### Checklist:

### DIFF
--- a/libobs/util/utf8.c
+++ b/libobs/util/utf8.c
@@ -146,8 +146,13 @@ size_t utf8_to_wchar(const char *in, size_t insize, wchar_t *out,
 	wlim = out == NULL ? NULL : out + outsize;
 
 	for (; p < lim; p += n) {
-		if (!*p)
+		if (!*p && !insize) {
 			break;
+		} else if (!*p) {
+			//skip intermediate null chars
+			n = 1;
+			continue;
+		}
 
 		if (utf8_forbidden(*p) != 0 && (flags & UTF8_IGNORE_ERROR) == 0)
 			return 0;
--- a/plugins/text-freetype2/text-functionality.c
+++ b/plugins/text-freetype2/text-functionality.c
@@ -421,7 +421,7 @@ static void remove_cr(wchar_t *source)
 void load_text_from_file(struct ft2_source *srcdata, const char *filename)
 {
 	FILE *tmp_file = NULL;
-	uint32_t filesize = 0;
+	uint32_t filesize = 0, textsize;
 	char *tmp_read = NULL;
 	uint16_t header = 0;
 	size_t bytes_read;
@@ -464,9 +464,10 @@ void load_text_from_file(struct ft2_source *srcdata, const char *filename)
 		bfree(srcdata->text);
 		srcdata->text = NULL;
 	}
-	srcdata->text = bzalloc((strlen(tmp_read) + 1) * sizeof(wchar_t));
-	os_utf8_to_wcs(tmp_read, strlen(tmp_read), srcdata->text,
-		       (strlen(tmp_read) + 1));
+
+	textsize = (uint32_t)os_utf8_to_wcs(tmp_read, filesize, NULL, 0) - 1;
+	srcdata->text = bzalloc((textsize + 1) * sizeof(wchar_t));
+	os_utf8_to_wcs(tmp_read, filesize, srcdata->text, (textsize + 1));
 
 	remove_cr(srcdata->text);
 	bfree(tmp_read);


WizardCM:
Note: this is an alternative fix to #7622

SBaori:
Should I create a new pull request? Or commit the existing formatted files with a new title on the same branch?

WizardCM:
Pushing / force-pushing to the same branch is preferred. Merge commits should be avoided.

plugins/text-freetype2/text-functionality.c
```diff
@@ -464,9 +464,8 @@ void load_text_from_file(struct ft2_source *srcdata, const char *filename)
 		bfree(srcdata->text);
 		srcdata->text = NULL;
 	}
-	srcdata->text = bzalloc((strlen(tmp_read) + 1) * sizeof(wchar_t));
-	os_utf8_to_wcs(tmp_read, strlen(tmp_read), srcdata->text,
-		       (strlen(tmp_read) + 1));
+	srcdata->text = bzalloc((filesize + 1) * sizeof(wchar_t));
```
kkartaltepe, line 467: Please measure the size of the decoded wcs data by calling `os_utf8_to_wcs` without a destination pointer.

kkartaltepe: And use that as the text size instead of filesize.

plugins/text-freetype2/text-functionality.c
```diff
@@ -464,9 +464,8 @@ void load_text_from_file(struct ft2_source *srcdata, const char *filename)
...
+	os_utf8_to_wcs(tmp_read, filesize, srcdata->text, (filesize + 1));
 
 	remove_cr(srcdata->text);
```
kkartaltepe, line 470: Please also fix remove_cr to handle nulls within the string

libobs/util/utf8.c
```diff
@@ -146,8 +146,12 @@ size_t utf8_to_wchar(const char *in, size_t insize, wchar_t *out,
 	wlim = out == NULL ? NULL : out + outsize;
 
 	for (; p < lim; p += n) {
-		if (!*p)
+		if (!*p && !n)
```
kkartaltepe, line 149: We should actually run through the rest of this function so it can correctly measure the length of the string. This function already handles nulls correctly so you can likely just remove the null check. If some other errors come up after removing the null check those should probably be fixed.

kkartaltepe (commented):
Please address the comments in the files. But also rebase your changes and remove the duplicate commits and merge commits. Let us know if you need some help with that.

libobs/util/utf8.c
SBaori: I actually tried doing that before but it won't work. The previous function code handled the NULL case by breaking the loop thinking the string has ended once it sees one. So the rest of the code in loop does not handle the NULL case (logically). But we want to continue reading till we hit the file size by skipping over the NULLs and not including it in the file size. 
correct me if I am missing something here.

plugins/text-freetype2/text-functionality.c
SBaori:
```suggestion
	textsize = os_utf8_wcs(tmp_read, filesize, NULL, 0);
	srcdata->text = bzalloc((textsize+1) * sizeof(wchar_t));
	os_utf8_to_wcs(tmp_read,filesize,srcdata->text,(textsize+1));
```

All good?

SBaori:
> Please address the comments in the files. But also rebase your changes and remove the duplicate commits and merge commits. Let us know if you need some help with that.

The duplicate commits are the ones after running clang format on them. Should I change the commit message to something like "text-freetype2: clang formated"? Or just revert them?

kkartaltepe:
> > Please address the comments in the files. But also rebase your changes and remove the duplicate commits and merge commits. Let us know if you need some help with that.
> 
> The duplicate commits are the ones after running clang format on them. Should I change the commit message to something like "text-freetype2: clang formated"? Or just revert them?

Squash the format fixes into their respective commits, and ensure there are no merge commits in the PR.

libobs/util/utf8.c
kkartaltepe: Which part of the function does not handle nulls correctly? If its simple to fix maybe we can fix that as well. Otherwise skipping sounds fine.

plugins/text-freetype2/text-functionality.c
kkartaltepe: Yes something like that

libobs/util/utf8.c
SBaori: I think when we encounter a null, it moves inside the else block at line 180 and since `flag=0` it breaks and returns 0.

kkartaltepe: Sorry it seems i misunderstood this fix, and we are instead just not copying out the null values in this function. Null (0x00) is treated like the other 7bit ascii characters in line 163 so it shouldnt error out. But it seems we were just skipping copying out the nulls with this fix, im less sure if this is what we would really want to do with this function but its probably workable as adding length handling everywhere ends up requiring many more changes.

SBaori: Oh yes line 163 handles the NULL case, I wonder where it breaks then cause the text does not get display correctly.

SBaori:
@kkartaltepe please take a look and let me know about the workaround and if any more changes required.

plugins/text-freetype2/text-functionality.c
kkartaltepe: Not needed with if we just dont copy internal nulls.

plugins/text-freetype2/text-functionality.c
SBaori: Should I create the textsize variable at the top along with others?

SBaori:
Anything else to add/modify? @kkartaltepe

kkartaltepe:
I think this is probably fine as is, thanks.

SBaori:
Greetings, any updates on this one?

kkartaltepe:
Hey, sorry there was some discussion of whether we want to allow this quirk in utf8 decoding to fix the problem or do a larger rewrite of freetype and we ended up just putting this off. Probably we will end up preferring a larger change to freetype to correctly handle string lengths and interior nulls.

PR #7633, opened 2022-10-22
[WIP] obs-ffmpeg: Add av1 vaapi implementation

kkartaltepe:
### Description

Work in progress. Add a vaapi encoder for AV1, tested on non-upstreamed patchsets for ffmpeg from intel so this wont actually work if you using official ffmpeg sources until those patches land. Once those patches do land this may or may not need some changes.

This is also kind of a mess thrown together, we might be able to use the new ffmpeg-video-encoders tooling that was added for software AV1 but I havnt looked too hard at the code.

### Motivation and Context

I wanted to test this intel gpu encoder, it works and the encoder is fast ~250fps but since only CQP is supported its not really suitable for streaming. Maybe this will motivate vendors to try again to land their patches. Also its nice to have for anyone wanting to play around with it.

### How Has This Been Tested?

Encoding and playback on an a750 worked fine.

### Types of changes

### Checklist:

### DIFF
--- a/plugins/obs-ffmpeg/obs-ffmpeg-vaapi.c
+++ b/plugins/obs-ffmpeg/obs-ffmpeg-vaapi.c
@@ -48,6 +48,11 @@
 #define info(format, ...) do_log(LOG_INFO, format, ##__VA_ARGS__)
 #define debug(format, ...) do_log(LOG_DEBUG, format, ##__VA_ARGS__)
 
+enum obs_vaapi_codec {
+	OBS_VAAPI_CODEC_H264,
+	OBS_VAAPI_CODEC_AV1,
+};
+
 struct vaapi_encoder {
 	obs_encoder_t *encoder;
 
@@ -56,6 +61,7 @@ struct vaapi_encoder {
 
 	const AVCodec *vaapi;
 	AVCodecContext *context;
+	enum obs_vaapi_codec type;
 
 	AVPacket *packet;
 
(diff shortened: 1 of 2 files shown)


kkartaltepe:
Someone requested how to actually run this, you must use intel's dg2 enablement ubuntu. Instructions https://dgpu-docs.intel.com/installation-guides/ubuntu/ubuntu-jammy-arc.html

Conan-Kudo:
This should also work with Linux systems with AMD Radeon RX 7000 series GPUs too, right? AV1 through Mesa VA-API would be available on such systems.

kkartaltepe:
As mentioned, it depends on ffmpeg patches from Intel. So this will not work in any platform unless those are merged. Beyond the ffmpeg changes I have no clue about the state of AV1 on AMD and/or the compatibility with the Intel patches.

eclairevoyant:
> Which patches are needed? Couldn't find them via google. `vainfo` on the 7900 XTX reports AV1 as supported seemingly, so I'd potentially give it a shot on that platform too.

You'll most likely need to compile ffmpeg (git version) on your own using [this patch](https://github.com/intel/cartwheel-ffmpeg/blob/master/patches/0071-lavc-vaapi-support-av1-encode.patch), but you may need to apply all the patches in https://github.com/intel/cartwheel-ffmpeg/tree/master/patches, in case there's dependencies.

cromefire:
Got it compiling with only patch 69 and 71 on latest upstream ffmpeg.

kkartaltepe:
Closing this one out as bringing qsv to linux seems more likely than ffmpeg av1 support.

PR #7636, opened 2022-10-22
Refactor main docks toggle and integration and browser docks

tytan652:
### Description

**Dependency of:**
- https://github.com/obsproject/obs-studio/pull/7637
- https://github.com/obsproject/obs-studio/pull/7638

Groundwork for other PRs and also avoid creating QActions by using the one included in the dock by modifying its connections making it usable even while the docks are locked.

### Motivation and Context

Preparation for PRs that depends on this one.

### How Has This Been Tested?

- Test docks toggle with docks locked or not
- Connect to services integration
- Modify custom browser docks

### Types of changes

- Tweak (non-breaking change to improve existing functionality)

### Checklist:

- [x] My code has been run through [clang-format](https://github.com/obsproject/obs-studio/blob/master/.clang-format).
- [x] I have read the [**contributing** document](https://github.com/obsproject/obs-studio/blob/master/CONTRIBUTING.rst).
- [x] My code is not on the master branch.
- [x] The code has been tested.
- [x] All commit messages are properly formatted and commits squashed where appropriate.
- [x] I have included updates to all appropriate documentation.

PR #7637, opened 2022-10-22
Avoid having two docks with the same object name

tytan652:
### Description

**Depends on:**
- https://github.com/obsproject/obs-studio/pull/7636

**Dependency of:**
- https://github.com/obsproject/obs-studio/pull/7638

https://doc.qt.io/qt-6/qmainwindow.html#saveState
> You should make sure that this property is unique for each QToolBar and QDockWidget you add to the QMainWindow.

We should avoid letting plugins add docks with the same object name. So to implement that:
- `obs_frontend_add_dock()` is deprecated in favor of `obs_frontend_add_dock_by_id()` and `obs_frontend_add_custom_qdock()` which ask for an unique id (object name). The dock is not added if the id is already used and return false in this case. 
- Also the new method does not return a QAction since internally we use the one included in QDockWidget.
- `obs_frontend_add_dock()` implements a temporary workaround to avoid this.
- `obs_frontend_add_custom_qdock()` replace the usage of `obs_frontend_add_dock()` where the returned QAction is deleted (e.g. Sources Dock plugin). The dock is not added if the id is already used and return false in this case. 
- If the object name of docks added with `obs_frontend_add_dock()` and  `obs_frontend_add_custom_qdock()` happen to be changed, the object name is changed back to its original.
- Docks added with new methods can be removed from the UI with `obs_frontend_remove_dock()`

**Note: If `obs_frontend_add_custom_qdock()` usecase is something we want to avoid, review the code about removing it and deprecate the usecase through `obs_frontend_add_dock()` deprecation.**

### Motivation and Context

Fix a potential issue and preparation for PRs that depends on this one.

Also make the API more neutral toward the used dock system for non-custom methods.

### How Has This Been Tested?

- Modified Downstream Keyer plugin to use `obs_frontend_add_dock_by_id()` or `obs_frontend_add_custom_qdock()`

### Types of changes

### Checklist:

### DIFF
--- a/UI/api-interface.cpp
+++ b/UI/api-interface.cpp
@@ -398,6 +398,35 @@ struct OBSStudioAPI : obs_frontend_callbacks {
 		return (void *)main->AddDockWidget(d);
 	}
 
+	bool obs_frontend_add_dock_by_id(const char *id, const char *title,
+					 void *widget) override
+	{
+		if (main->IsDockObjectNameUsed(QT_UTF8(id))) {
+			blog(LOG_WARNING,
+			     "Dock id '%s' already used!  "
+			     "Duplicate library?",
+			     id);
+			return false;
+		}
+
+		OBSDock *dock = new OBSDock(main);
+		dock->setWidget((QWidget *)widget);
+		dock->setWindowTitle(QT_UTF8(title));
+		dock->setObjectName(QT_UTF8(id));
+
+		main->AddDockWidget(dock, Qt::RightDockWidgetArea);
+
+		dock->setFloating(true);
+		dock->setVisible(false);
+
+		return true;
+	}
+
+	void obs_frontend_remove_dock(const char *id) override
+	{
+		main->RemoveDockWidget(QT_UTF8(id));
+	}
+
 	void obs_frontend_add_event_callback(obs_frontend_event_cb callback,
 					     void *private_data) override
 	{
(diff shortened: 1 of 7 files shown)


PatTheMav:
Personally I'd prefer to avoid the whole "function_call[insert integer here]" shenanigans, it always seems half-arsed to me. If code uses the old signature it should break and throw a compile-time error, forcing any authors relying on it to update to the new API, especially as this probably won't end up in a minor version update anyway.

Otherwise we're dragging a whole ecosystem of mixed implementations with us until we end up with `obs_frontend_add_dock5`.

PatTheMav:
Is there other code that uses this function? Because if it is only us, then we should fix all invocations of the code, but I'm afraid you might tell me that plugins are allowed to use this function?

tytan652:
If you are talking about `obs_frontend_add_dock()`, it's a Frontend API function used by many plugins.

PatTheMav:
I was afraid you'd say that - internal UI methods shouldn't be exposed as-is to the API, but that's for later to fix.

I'd suggest naming the function `obs_frontend_add_dock_by_id` then, because it's a) different and b) clearly communicates what it does.

tytan652:
Renamed `obs_frontend_add_custom_dock` to `obs_frontend_add_custom_qdock`.

tytan652:
Replace QUuid usage in `obs_frontend_add_dock` by `os_generate_uuid`.

tytan652:
Added this note to the PR description:
> Note: If `obs_frontend_add_custom_qdock()` usecase is something we want to avoid, review the code about removing it and deprecate the usecase through `obs_frontend_add_dock()` deprecation.

PR #7640, opened 2022-10-22
Improve Premultiplication Performance

CryZe:
### Description
There's no need to go through floating point numbers in the non sRGB case as you can easily multiply color channels as integers. 

### Motivation and Context
It's faster by around 30%.

### How Has This Been Tested?
I've verified that this results in the same output with all possible value combinations.

### Types of changes

### Checklist:


CryZe:
I didn't do the same optimization for the sRGB case, but arguably sRGB should just be a table lookup anyway, and then it could be optimized the same way.

norihiro:
> It's about two times as fast. 3.687 seconds for 4 GiB of RGBA data compared to the previous 7.677 seconds.

How this was tested? Do you have a code so that I (or someone else) can test the same condition? I'd like to test whether it improves on my host.
What was your setup to measure that, such as CPU and compiler?

Something I'm thinking of is adding `if (a == 255) return;` before touching the RGB values. It might perform further better in some conditions.

The commit message should be added a prefix `libobs/graphics: `. Only first letter should be capitalized, so `P` should be `p`.

CryZe:
I did the performance testing in Rust, but here's the code: [Playground](https://play.rust-lang.org/?version=stable&mode=release&edition=2021&gist=49add8ad475e1c979131ae034cc8904e)

I also threw both versions into Godbolt and used llvm-mca to check the cycle count on various CPUs, it was always somewhere between 1.5x and 2x faster.

It turns out that an early return is indeed a bit faster too. So I included it in all the four variants.

I also updated the commit message.

norihiro:
I compared another proposal from rcdrone on Discord.
```c
static inline void gs_premultiply_xyza_rcdrone(uint8_t *data)
{
        uint32_t a = data[3] * 2;
        data[0] = (data[0] * a + 255) / 510;
        data[1] = (data[1] * a + 255) / 510;
        data[2] = (data[2] * a + 255) / 510;
}
```

Comparing the outputs from a code below with (a) original `gs_premultiply_xyza`, (b) modified code by this PR, and (c) above code commented by rcdrone. These 3 cases show no difference. (I use GCC on Fedora 35 but the interger calculation should be as same in other supported platforms.)
```c
int main()
{
        for (int a = 0; a < 256; a++) {
                for (int b = 0; b < 256; b++) {
                        uint8_t data[4] = {b, 0, 0, a};
                        gs_premultiply_xyza(data);
                        printf("%d\t%d\t%d\n", b, a, data[0]);
                }
        }
        return 0;
}
```

CryZe:
> I compared another proposal from rcdrone on Discord.

What's the advantage of this implementation? At a first glance it looks slower as there's now a longer data dependency on alpha as it first needs to do a multiplication with 2 now. I did a similar optimization for the float case where I moved a common multiplication / division with 255 before hand into the alpha calculation, reducing the amount of operations by two. However it ended up being slower as RGB needed to wait for alpha to be calculated for longer, stalling the pipeline. I'm fairly confident the same is true here.

norihiro:
> What's the advantage of this implementation?

The point is that `127/255` is smaller than 0.5 but `255 / 510` equals 0.5.

CryZe:
Ah, I see. My approach was different in that I simply thought: What do I need to add to ensure 128 gets rounded up. So 127 + 128 = 255 which after the division with 255 rounds it up. I didn't try to make it equal to 0.5. And as you see it produces the correct result too and as elaborated, is faster, so I'd keep it.

CryZe:
Considering Aarch64 machines compile with NEON by default, I did push a version now with the early return conditionally compiled out in when it would hurt performance.

nyabinary:
shouldn't this be merged by now?

libobs/graphics/srgb.h
```diff
@@ -117,28 +124,38 @@ static inline void gs_premultiply_xyza_srgb(uint8_t *data)
...
+	} else {
+		uint8_t u[4];
+		float f[4];
+		memcpy(&u, src, sizeof(u));
+		gs_u8x4_to_float4(f, u);
+		gs_float3_srgb_nonlinear_to_linear(f);
+		gs_premultiply_float4(f);
+		gs_float3_srgb_linear_to_nonlinear(f);
+		gs_float3_to_u8x3(u, f);
+		memcpy(dst, &u, sizeof(u));
```
PatTheMav: Isn't this identical to what `gs_premultiply_xyza_srgb` does? Is there a drawback to calling that function instead?

libobs/graphics/srgb.h
```diff
@@ -92,17 +92,24 @@ static inline void gs_float3_srgb_linear_to_nonlinear(float *f)
...
+// the vectorization in both cases.
+#if !(defined(__ARM_NEON) || defined(__AVX2__))
+	if (data[3] == 0xFF)
+		return;
+#endif
+
+	uint16_t a = data[3];
+	data[0] = (data[0] * a + 127) / 255;
+	data[1] = (data[1] * a + 127) / 255;
+	data[2] = (data[2] * a + 127) / 255;
```
PatTheMav: While reading this here I wondered how this passed CI, but when building locally I got what I expected:

```
Implicit conversion loses integer precision: 'int' to 'uint8_t' (aka 'unsigned char')
```
The multiplication result will automatically be elevated to `uint16_t` and the compiler cannot assume that the final value will always fit within 8 bytes again. 

Casting each result to `uint8_t` will fix that warning and make it explicit to the reader that "yes, we are actively truncating the result, but we expect this value never to exceed the range, so it's fine".

PatTheMav (changes requested):
How was the vectorisation tested? Because at least naively testing with `clang` on ARM64 in compiler explorer, I can see the code getting simplified with `-O3 -march=armv8-a+simd+fp` but I see no vectorisation taking place.

PR #7643, opened 2022-10-23
libobs: Make wcs<->utf8 conversion consistent

derrod:
### Description

On Windows NULL characters would be included in the output if insize is non-zero, but on *nix it would abort on the first NULL.
This just changes the condition to only `break` if insize is 0.

(This is arguably also more consistent with the function description in the comments.)

### Motivation and Context

First step towards fixing #7595, will need changes to gdi/freetype source to handle `NULL`s in output, for instance by replacing it with a placeholder character such as �

### How Has This Been Tested?

Hasn't as of yet.

### Types of changes

### Checklist:

### DIFF
--- a/libobs/util/utf8.c
+++ b/libobs/util/utf8.c
@@ -146,7 +146,7 @@ size_t utf8_to_wchar(const char *in, size_t insize, wchar_t *out,
 	wlim = out == NULL ? NULL : out + outsize;
 
 	for (; p < lim; p += n) {
-		if (!*p)
+		if (!*p && insize == 0)
 			break;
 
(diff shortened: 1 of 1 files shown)


kkartaltepe:
we had an external contributor also working on this in #7628 , are you aware of any issues with removing internal nulls from wcs strings? If not we can go with #7628 and then re-introduce nulls once we have freetype2 support for sized strings, as that will require a bit of work in the plugin.

derrod:
I don't believe there is any issue with removing nulls, but this makes the Windows and *nix implementations consistent.

kkartaltepe (approved):
Probably fine to merge this and work on the fixes for internal nulls where we have to (just freetype I guess)

Issue #7623, opened 2022-10-20
White image gets alpha blended as grey image


CryZe:
### Operating System Info

Windows 10

### OBS Studio Version

28.0.3

### OBS Studio Log URL

doesn't apply

### Expected Behavior

In https://github.com/obsproject/obs-studio/pull/6257 blending modes have been added. And it seems to indeed "mostly" fix my issue with LiveSplit not properly being captured transparently. However I feel like there's still "something wrong".  If you look at this image:

![https://i.imgur.com/vqiJFUh.png](https://i.imgur.com/vqiJFUh.png)

you'll notice that it's a white square that only contains white pixels. However the opaqueness changes from left being fully opaque to right being transparent. As far as I understand sRGB really is just a curve applied to the colors to such that darker colors get more of the 8-bit available per channel as human vision is not actually linear and we can see changes in brightness in dark colors really well. So in the end sRGB often is approximated as pow(color, 2.2). However if we think about it, the RGB values in my image are fully white and thus it's pow(1.0, 2.2), i.e. RGB stays at (1, 1, 1). The alpha however is somewhere between 0 and 1 as it's a gradient, so with 0.5 or so you get pow(0.5, 2.2) = 0.21... but more importantly, it basically just moves the alpha somewhere else between 0 and 1. So in the end what we should see is an image that no matter whether you consider it to be sRGB or not, should look white, but the gradient may look a little different.

### Current Behavior

 But that's not at all what the blending mode in OBS seems to do. Instead I'm getting a grey image:
 
 ![https://i.imgur.com/mqTRpAk.png](https://i.imgur.com/mqTRpAk.png)

### Steps to Reproduce

1. Add an image source.
2. Select the linked image.
3. Make sure "Apply alpha in linear space is deactivated"
4. Put it above a white background.

### Anything else we should know?

This is a clean reproducible continuation of: https://github.com/obsproject/obs-studio/issues/5674

Also: People in Discord are suggesting that `SRGB` and `sRGB` are two separate concepts and that this might be related. However no one was able to link a source nor was I able to find any.

CryZe:
[...]

PR #7647, opened 2022-10-23
UI: Remove the blending method

CryZe:
### Description
This removes the blending method from the User Interface.

### Motivation and Context
As concluded in #7623 this is not something that the user can or should be setting.

### How Has This Been Tested?
This has been tested on Windows 10.

### Types of changes

### Checklist:


Vainock:
You only need to update en-US.ini, the rest will be updated with the next Crowdin update.

CryZe:
Ah alright, I'll change it back.

UI/window-basic-main.hpp
```diff
@@ -869,7 +867,10 @@ private slots:
 		return os_atomic_load_bool(&previewProgramMode);
 	}
 
-	inline bool VCamEnabled() const { return vcamEnabled; }
+	inline bool VCamEnabled() const
+	{
+		return vcamEnabled;
+	}
```
gxalpha, line 873: Please run your code through `clang-format` to make sure that these lines don't get reformatted

gxalpha (changes requested):

UI/window-basic-main.hpp
CryZe: I did, weirdly it got formatted differently on my machine.

gxalpha: This could depend on the version of clang-format you're running. We're using 13, which iirc should be in the current VS Code.
Otherwise, in this case it would probably work to manually undo the changes.

Fenrirthviti:
As a note for anyone reading this PR, the conclusion was made solely by the author of this PR, and has not been evaluated by anyone on the core OBS team from a technical standpoint at the time of writing this.

This removes an option, but does not provide a workaround or alternate solution to the issue. Can you elaborate why removing this option is better for users than providing a workaround that doesn't require rebuilding assets (which, in many cases by many users will be provided by someone else, and they would not have access to the source files)? Please try to provide practical examples on why this is better, not simply as a philosophical debate.

From my perspective as someone who doesn't fully understand the nuance of alpha blending, which would include a majority of OBS Studio users, removing this option means that half of my assets will no longer work how I expect them and I have no way to correct it.

To stress again, I would like to focus this conversation on how this is *better for our users* not start another discussion on what is "right" or "wrong" here.

CryZe:
> not simply as a philosophical debate.

If you check the issue linked the bottom comment is a full technical explanation of why this setting does not work from a technical perspective.

> To stress again, I would like to focus this conversation on how this is better for our users

The best thing for the user is that the setting is automatically chosen by the source. This would be a followup PR, or could be included as part of this PR.

Fenrirthviti:
> If you check the issue linked the bottom comment is a full technical explanation of why this setting does not work from a technical perspective.

I have checked the issue, but from practical experience, despite the claims that things are "wrong" the end result is working just fine, and as expected. However, let's just leave it at that for the purpose of this PR, any further discussion can happen on that issue itself.

> The best thing for the user is that the setting is automatically chosen by the source. This would be a followup PR, or could be included as part of this PR.

Thank you, I would very much prefer that these two options are PRed together so that we are not left in a state with an at worst imperfect, but viable workaround. A more reliable automatic way to select the method would be better, I agree.

UI/window-basic-main.hpp
PatTheMav: Alternatively you could just cherry pick the lines/hunks you want to have changed for the commit instead of the entire file if clang-format still formats parts that you actually don't intend to change.

Warchamp7:

PR #7652, opened 2022-10-24
UI: Change adv audio background color

cg2121:
### Description
The color of the disabled sliders would be the same as the background in the advanced audio dialog.

Before:
Screenshot from 2022-10-23 22-33-46

After:
Screenshot from 2022-10-23 22-27-55

### Motivation and Context
Fix Yami bugs.

### How Has This Been Tested?
Opened advanced audio dialog to make sure the background color was correct.

### Types of changes

### Checklist:

### DIFF
--- a/UI/data/themes/Acri.qss
+++ b/UI/data/themes/Acri.qss
@@ -1519,3 +1519,8 @@ QDialogButtonBox {
 OBSBasicStats {
     background: palette(dark);
 }
+
+/* Advanced audio dialog */
+OBSBasicAdvAudio #scrollAreaWidgetContents {
+    background: palette(dark);
+}
--- a/UI/data/themes/Grey.qss
+++ b/UI/data/themes/Grey.qss
@@ -1507,3 +1507,8 @@ QDialogButtonBox {
 OBSBasicStats {
     background: palette(dark);
 }
+
+/* Advanced audio dialog */
+OBSBasicAdvAudio #scrollAreaWidgetContents {
+    background: palette(dark);
+}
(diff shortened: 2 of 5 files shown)


notr1ch:
I'm not sure this is the correct way to resolve the issue. The widget is supposed to indicate that the panning is permanently locked at 50% and the controls are disabled. Why then is the 50% in the left solid black and not a greyed out blue as is used in the enabled sliders?

Thinking about it some more, the whole idea of it being a slider with a visual indication of the current value doesn't really make sense. This isn't a volume control - setting it to 100% means the left channel is completely disabled, not fully enabled, counter-intuitive to how the widget looks.

Is it possible to remove the blue part entirely instead so it's just a slider?

PatTheMav:
> I'm not sure this is the correct way to resolve the issue. The widget is supposed to indicate that the panning is permanently locked at 50% and the controls are disabled. Why then is the 50% in the left solid black and not a greyed out blue as is used in the enabled sliders?
> 
> Thinking about it some more, the whole idea of it being a slider with a visual indication of the current value doesn't really make sense. This isn't a volume control - setting it to 100% means the left channel is completely disabled, not fully enabled, counter-intuitive to how the widget looks.
> 
> Is it possible to remove the blue part entirely instead so it's just a slider?

I agree, if there should be a highlighted area in the slider at all, it should grow from the center to whichever position the slider was moved - if that turns out to be too complicated setting the "highlight" colour to the background colour should suffice.

I'd be fine with merging this PR as it just fixes the panel's background colour and have the fixes for the slider be in a separate PR, as it affects the enabled variant as well as the disabled one (which creates this visual confusion).

GeorgesStavracas (approved):
Warchamp seemed happy with this change

Warchamp7 (approved):

PR #7664, opened 2022-10-26
obs-ffmpeg: Fix FFmpeg NVENC presets on Ubuntu 20.04

jp9000:
### Description
Fix Ubuntu 20.04 NVENC requiring old NVENC presets

### Motivation and Context
Ubuntu 20.04 and 22.04 both have the same FFmpeg versions, but FFmpeg on 20.04 was built with version 9 of NVENC headers, whereas 22.04 was built with 11.

Unfortunately, that means we have to revert back to the old presets when Ubuntu 20.04 is detected. The way this detection is done is a bit hacky, but it's the only way to preserve startup performance. Otherwise we'd have to write yet another startup subprocess test program to detect whether the FFmpeg being used was built with older or newer NVENC, which is both slow to start up and annoying to write. So instead, just get the distro name/version strings and detect Ubuntu 20.04 that way.

### How Has This Been Tested?
Hasn't really been tested, could probably use a tester, even if it just means forcing the ubuntu 20.04 variable

### Types of changes

### Checklist:

### DIFF
--- a/plugins/obs-ffmpeg/obs-ffmpeg-nvenc.c
+++ b/plugins/obs-ffmpeg/obs-ffmpeg-nvenc.c
@@ -39,6 +39,10 @@ struct nvenc_encoder {
 	DARRAY(uint8_t) sei;
 };
 
+#ifdef __linux__
+extern bool ubuntu_20_04_nvenc_fallback;
+#endif
+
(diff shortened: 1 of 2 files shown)


jp9000:
Going to merge for now, if anyone has any issues please let me know

PR #7667, opened 2022-10-26
vlc-video: Fix videos larger than 1080p being squished

### LINKED ISSUE #7666: Overflow of `display_width` in vlc-video plugin
### Operating System Info

Other

### Other OS

Arch Linux

### OBS Studio Version

Git

_No response_

### OBS Studio Log URL

https://obsproject.com/logs/-9kLz92IqEfzryb8

### OBS Studio Crash Log URL

_No response_

### Expected Behavior

Valid width and height of VLC video source.

### Current Behavior

Playing a video with very high resolution (tested with 3860x2160) multiplexed via MKVToolNix causes overflow of `display_width` variable in `calculate_display_size` function (`plugins/vlc-video/vlc-video-source.c`), because `track->video->i_sar_num` and `track->video->i_sar_den` contain a large number (in my scenario total number of pixels): 3840*8294400 > 2^31-1.

### Steps to Reproduce

1. Make example videos
    - `ffmpeg -f lavfi -i testsrc=duration=15:size=3840x2160:rate=30 -c:v libx264 -preset fast -crf 22 4k_test_ffmpeg.mkv`
    - `mkvmerge -o 4k_test_toolnix.mkv 4k_test_ffmpeg.mkv`
2. In OBS, create new VLC source and select:
    - `4k_test_ffmpeg.mkv`, the video plays correctly
    - `4k_test_toolnix.mkv`, the video is horizontally squished due overflow

_No response_

CoalZombik:
### Description

Current behavior leads to overflow in computing source width with Sample Aspect Ratio. Changing to `uint64_t` fixed this problem.

### Motivation and Context

Resolve #7666.

### How Has This Been Tested?

Playing testing videos described in #7666.

### Types of changes

Changed precision of computing `display_width` with SAR.

### Checklist:

### DIFF
--- a/plugins/vlc-video/vlc-video-source.c
+++ b/plugins/vlc-video/vlc-video-source.c
@@ -401,8 +401,8 @@ static void calculate_display_size(struct vlc_source *c, unsigned *width,
 			if (track->i_type != libvlc_track_video)
 				continue;
 
-			int display_width = track->video->i_width;
-			int display_height = track->video->i_height;
+			unsigned display_width = track->video->i_width;
+			unsigned display_height = track->video->i_height;
 
 			if (display_width == 0 || display_height == 0)
 				continue;
@@ -410,9 +410,9 @@ static void calculate_display_size(struct vlc_source *c, unsigned *width,
 			/* Adjust for Sample Aspect Ratio (SAR) */
 			if (track->video->i_sar_num > 0 &&
 			    track->video->i_sar_den > 0) {
-				display_width = display_width *
-						track->video->i_sar_num /
-						track->video->i_sar_den;
+				display_width = (unsigned)util_mul_div64(
+					display_width, track->video->i_sar_num,
+					track->video->i_sar_den);
 			}
 
 			switch (track->video->i_orientation) {


PatTheMav:
Wouldn't it make more sense to put the devision into brackets to yield the ratio of those two big values (which usually should yield "1" for digital displays anyway) and then multiply with that comparatively small ratio?

Doing any kind of division with arbitrary values without using a fractional type is also not great to begin with, but not a fault of the PR per se.

CoalZombik:
If we put division into brackets, we must compute it as floating-point value, because ratio can be fraction (for example 4/3 = 1 in integer division). With integer division, first multiplication then division must be done and because we multiplying two 32bit values (in scenario with normal compiler), it cannot overflow with 64bit number type. 

I try change as little code as possible in this fix, so I did it this way. Computing as floating-point values will be probably more elegant way.

tt2468:
We have the `util_mul_div64()` utility function, which may be of use here.

CoalZombik:
Sorry for late response. `util_mul_div64` is definitely better option. I created new commit with this change.

RytoEX:
Please fix the following warning:
```
D:\a\obs-studio\obs-studio\obs-studio\plugins\vlc-video\vlc-video-source.c(413,21): error C2220: the following warning is treated as an error [D:\a\obs-studio\obs-studio\obs-studio\build32\plugins\vlc-video\vlc-video.vcxproj]
D:\a\obs-studio\obs-studio\obs-studio\plugins\vlc-video\vlc-video-source.c(413,21): warning C4244: '=': conversion from 'uint64_t' to 'int', possible loss of data [D:\a\obs-studio\obs-studio\obs-studio\build32\plugins\vlc-video\vlc-video.vcxproj]
```

tuduweb:
I guess you should simplify the PR and only keep the latest commit.

just `git reset --soft f065f20` to the basic version, commit the changed again and then `git push -f` to your own repository.

CoalZombik:
@tuduweb Thank you for the recommendation.

It is squashed to one commit now, but I don't understand why Flatpak build failed.

gxalpha:
@CoalZombik looks like the Flatpak failure was due to something outside this Pull Request. I restarted it and would expect it to complete.

PR #7673, opened 2022-10-27
obs-ffmpeg: Fix Ubuntu 20.04 detection

derrod:
### Description

The current Ubuntu 20.04 check doesn't work since both name and version id are quoted, but quotes are not stripped, resulting in the `dstrcmp`s here failing:

https://github.com/obsproject/obs-studio/blob/dd26fe4f8a5fc603a405841bfb02069bf59ee5c8/plugins/obs-ffmpeg/obs-ffmpeg.c#L274-L275

`VERSION_CODENAME` on the other hand is not quoted, and also easier since it requires just a single string.

### Motivation and Context

Fixes.

### How Has This Been Tested?

Hasn't.

### Types of changes

### Checklist:


tt2468 (approved):
Looks good, separate fix needs to be made for the old preset translation strings. Looks like they may have been removed and need to be added back.

PR #7679, opened 2022-10-28
Allow building with wayland alone (no X11)

lanodan:
- CMake: Add ENABLE_X11 option
- UI: Building with ENABLE_X11 OFF
- deps/glad: Building with ENABLE_X11 OFF
- libobs-opengl: Building with ENABLE_X11 OFF
- libobs: Building with ENABLE_X11 OFF

### Motivation and Context
This allows to build OBS on machines with Wayland present but X11 being absent (ie. embedded devices).

### How Has This Been Tested?
Using Gentoo Linux, `USE=-X`

### Checklist:


tytan652 (changes requested):
Quick review:
- Please in CMake, forbid `ENABLE_X11` and `ENABLE_WAYLAND` being both OFF.
- Changes in `deps/glad` seems unrelated to the PR.

tytan652:
> This allows to build OBS on machines with Wayland present but X11 being absent (ie. embedded devices).

Can you please add more information ?
Because for now almost every distro have usually both because Wayland only desktop systems happen to have XWayland.
And for the embedded devices part, I think more information is required.

libobs/obs-nix-platform.h
```diff
@@ -18,14 +18,17 @@
...
 #endif
 
 enum obs_nix_platform_type {
+#ifdef ENABLE_X11
 	OBS_NIX_PLATFORM_X11_GLX OBS_DEPRECATED,
 	OBS_NIX_PLATFORM_X11_EGL,
+#endif
 #ifdef ENABLE_WAYLAND
 	OBS_NIX_PLATFORM_WAYLAND,
 #endif
```
tytan652, line 34: This change cause an API/ABI breakage, please remove this change.

I will make a bigger review once CI is all green.

tytan652 (changes requested):

lanodan:
[2022-10-28 06:23:25-0700] tytan652:
>> This allows to build OBS on machines with Wayland present but X11 being absent (ie. embedded devices).
>
>Can you please add more information ?
>Because for now almost every distro have usually both because Wayland only desktop systems happen to have XWayland.
>And for the embedded devices part, I think more information is required.

XWayland should be irrelevant here as OBS can use the wayland protocol, but without changes like this PR it's still stuck to requiring X11 libraries.

SailfishOS is one example of an X11-less mobile distro, which has ports to tablets.  
That said my use case is more on making sure that software will still work as Xorg goes less and less maintained.

lanodan:
***@***.*** requested changes on this pull request.
>- Changes in `deps/glad` seems unrelated to the PR.

They are relevant, *glx* is an X11 extension, `libGL.so` is also OpenGL+GLX, meaning it's gone when you build mesa without X11.

I plan to make a PR to `glad` btw, otherwise it could end up being reverted in an update, but that will likely be for glad2.

tytan652:
> XWayland should be irrelevant here as OBS can use the wayland protocol, but without changes like this PR it's still stuck to requiring X11 libraries.

Qt Wayland support is not perfect, CEF support neither.

> SailfishOS is one example of an X11-less mobile distro, which has ports to tablets.

Sorry to say that, but it is a bad example. If I remember correctly, this OS has no "desktop mode" and OBS Studio is only designed for desktop.

> That said my use case is more on making sure that software will still work as Xorg goes less and less maintained.

Not against comment but there is still a long way before Xorg can be completely replaced by Wayland.

> They are relevant, *glx* is an X11 extension, `libGL.so` is also OpenGL+GLX, meaning it's gone when you build mesa without X11.
>
> I plan to make a PR to `glad` btw, otherwise it could end up being reverted in an update, but that will likely be for glad2.

You should have made it before this one if glad can be generated without GLX, the bump to glad2 is not plan for now so going this way is not recommended.

Note for those who don't know about glad: Most of glad files are generated so manually modifying will may be a no-go.

PR #7680, opened 2022-10-29
obs-frontend-api: Add functions to get last saved files

cg2121:
### Description
Adds functions to the frontend api to get the last recording, screenshot and replay buffer saved.

### Motivation and Context
Makes it easier for developers to find these files.

### How Has This Been Tested?
Tested with script.

### Types of changes

### Checklist:

### DIFF
--- a/docs/sphinx/reference-frontend-api.rst
+++ b/docs/sphinx/reference-frontend-api.rst
@@ -798,3 +798,21 @@ Functions
 .. function:: bool obs_frontend_is_theme_dark(void)
 
    :return: Checks if the current theme is dark or light.
+
+---------------------------------------
+
+.. function:: const char *obs_frontend_get_last_recording(void)
+
+   :return: The file path of the last recording.
+
+---------------------------------------
+
+.. function:: const char *obs_frontend_get_last_screenshot(void)
+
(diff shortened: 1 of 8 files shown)


docs/sphinx/reference-frontend-api.rst
```diff
@@ -798,3 +798,21 @@ Functions
 .. function:: bool obs_frontend_is_theme_dark(void)
 
    :return: Checks if the current theme is dark or light.
+
+---------------------------------------
+
+.. function:: const char *obs_frontend_get_last_recording(void)
+
+   :return: Gets the last recording.
```
CodeYan01, line 806: Just a nit, I think this should be "The file path of the last recording", and similarly for the other two functions. The :returns: should not need to say "Gets". Also, explicitly saying "file path" is clearer, especially for the screenshot, which someone might interpret as the raw image data. Of course, the return type is a const char *, but imo it's better to spend a few more characters for clarity.

tt2468:
I believe to match with other existing frontend API functions, returning a `char*` would be better.

docs/sphinx/reference-frontend-api.rst
cg2121: Fixed.

PR #7687, opened 2022-10-30
UI: Check item whether selected before select

### LINKED ISSUE #7685: Pasting multiple sources puts the bottom source at the top of the pasted sources
### Operating System Info

Windows 10

### Other OS

_No response_

### OBS Studio Version

28.0.3

_No response_

### OBS Studio Log URL

N/A

### OBS Studio Crash Log URL

_No response_

### Expected Behavior

The copied sources are pasted in the same order they were in when copied

### Current Behavior

The bottom most source in the list of copied sources is placed at the top of the pasted sources

### Steps to Reproduce

1. Create three sources named 1, 2 and 3 top to bottom
2. Copy all three sources

image

3. Paste all three sources

Result: 3 is at the top of the list, followed by 1 and then 2 in proper order.
image

Happens with any number of sources. All sources in the correct order except the bottom one, which is placed at the top

image
image

_No response_

tuduweb:
### Description

Check item whether selected before select
To prevent item from being selected again,
then mess up the qt internal list order.

When Click SourceTree Item, then the App will run like following steps:

image

and the last step is not as expected, it **mess up** the qt internal list order. (found after debugging qtcore `QItemSelectionModel::select`)

So Add a judgement to Check item whether selected before select.

### Motivation and Context

Fixes #7685

### How Has This Been Tested?

reproduce steps mentioned in issues, and work fine.

Paste items order as expected.

Click item 0, then Press Shift and Click item n.
the order is `0->n` not `n, 0 -> n-1`

### Types of changes

### Checklist:

### DIFF
--- a/UI/source-tree.cpp
+++ b/UI/source-tree.cpp
@@ -1152,7 +1152,7 @@ void SourceTree::SelectItem(obs_sceneitem_t *sceneitem, bool select)
 		return;
 
 	QModelIndex index = stm->createIndex(i, 0);
-	if (index.isValid())
+	if (index.isValid() && select != selectionModel()->isSelected(index))
 		selectionModel()->select(
 			index, select ? QItemSelectionModel::Select
 				      : QItemSelectionModel::Deselect);


PatTheMav:
I don’t think this fixes the issue - the order in the list mostly resolved to the order in which items were selected, but user expectation is to retain the order in which the items existed in the scene collection, which can be arbitrary and needs to be retained as metadata.

I’m happy to be corrected on this, but that’s the impression I got from looking at the code.

tuduweb:
> I don’t think this fixes the issue - the order in the list mostly resolved to the order in which items were selected, but user expectation is to retain the order in which the items existed in the scene collection, which can be arbitrary and needs to be retained as metadata.
> 
> I’m happy to be corrected on this, but that’s the impression I got from looking at the code.

I review my code, and find that I misunderstood the issues.
I just found the implement of AI, and other Adobe product.It just pasted in the same order as the previous one(in layers).

And return to this mentioned issues, there is indeed some bugs so as to disturb the order(found when I debugging). I will try to fix this.

PatTheMav:
Even with the changes my main issue is not addressed: The _actual_ order of items (as they are present in a scene) is not copied or retained.

To achieve that, you'd need to copy the scene item _and_ associated scene item metadata (namely the position/index of the scene item within its _current_ scene) and then upon pasting the scene items ensure that the order of items inserted into the scene takes this positional data into account.

As long as you don't add this positional data to the copy and paste operations, the issue is not actually addressed because we need to ensure the order at which items are inserted is not based on the order items were selected.

Even with your fix, I could select the 3rd, 5th, then 2nd and 1st item and that'd be the exact order they'd be pasted (3,5,2,1 instead of 1,2,3,5).

PatTheMav:
I've revisited this today and while it doesn't fix the issue of truly retaining order of copied sources it fixes an important architectural issue:

When selecting source items in the source tree, selection is handled by `QtListView` and its `QItemSelectionModel`. When selecting source items in the preview, this step is omitted, so we need to "manually" add the selected items to the `QItemSelection`.

Checking if an item is already contained in the selection (avoiding to add it twice) is one way to fix it (another would be to fix the root cause of the issue and only manually add items to the selection if it happened in the preview, alas that's not how the selection handling is implemented).

I guess this happens because of the deduplication the `QItemSelection` does, dropping earlier entries.

@RytoEX any objections to merge this?

RytoEX:
> @RytoEX any objections to merge this?

None that I can think of.

tuduweb:
Thx for your in-depth issue analysis @PatTheMav.
I get a lot from it and understand the root cause of this issue.

it seems caused by the deduplication the `QItemSelection::merge`: dropping earlier entries when *split the old ranges*.
https://github.com/qt/qtbase/blob/df47d7d1ff2ac45ab5c6cdade89e5e1857f12479/src/corelib/itemmodels/qitemselectionmodel.cpp#L475-L494

PR #7690, opened 2022-10-31
mac-capture: Sort application and window list alphabetically

Colerar:
### Description

Order application and window list in mac screen capture panel.

image

image

### Motivation and Context

The list is now unordered, making it difficult to find the app or window you want to capture.

### How Has This Been Tested?

It is a small tweak, so I just run and see if it works.

### Types of changes

### Checklist:


PatTheMav (commented):
It's unfortunate this requires two loop iterations, especially as we throw away lots of entries in the default case. Usually you want the filtering happen before the sorting to minimise the amount of comparisons required to achieve the output.

Colerar:
Thanks for your advice! But I'm a newbie for objective-c. Actually, it's my first commit with obj-c. I'll try to do this, although maybe not as clear or idiomatic.

PatTheMav (approved):
Conceptually this seems fine given that the sorted arrays contain only pointers (so no unnecessary copies of data take place). Will check functionality tomorrow.

jp9000:
Seems to be a case-sensitive sort. Should probably be case-insensitive.

plugins/mac-capture/mac-screen-capture.m
```diff
@@ -835,24 +835,52 @@ static bool build_window_list(struct screen_capture *sc,
...
+				}
+				return true;
+			}]];
+
+	NSArray<SCWindow *> *sortedWindows;
+	sortedWindows =
+		[filteredWindows sortedArrayUsingComparator:^NSComparisonResult(
+					 SCWindow *window, SCWindow *other) {
+			NSComparisonResult appNameCmp =
+				[window.owningApplication.applicationName
```
PatTheMav, line 865: Per Jim's comment, please add `options: NSCaseInsensitiveSearch` to the `compare` message.

gxalpha (changes requested):
In addition to the comment above regarding case sensitivity, please fix the merge conflicts and squash the commits into one. Ideally, capitalize the first letter in the commit title after the prefix (`mac-capture: Sort apps and windows alphabetically`).

PR #6410, opened 2022-05-02
Adaptive Replay Buffer Memory Limit 

derrod:
### Description

Adds libobs utility function to get total available memory and changes UI code to adjust the maximum replay buffer accordingly. Also adds a warning for CBR/VBR modes when estimated usage exceeds suggested maximum.

### Motivation and Context

8 GiB is a little small considering most users have 16, and 32 is getting increasingly popular.

### How Has This Been Tested?

Tested on Windows 10 with advanced output replay buffer.

### ToDo

- Test on Linux

### Types of changes
- New feature (non-breaking change which adds functionality)
- Tweak (non-breaking change to improve existing functionality)

### Checklist:

- [x] My code has been run through [clang-format](https://github.com/obsproject/obs-studio/blob/master/.clang-format).
- [x] I have read the [**contributing** document](https://github.com/obsproject/obs-studio/blob/master/CONTRIBUTING.rst).
- [x] My code is not on the master branch (ignore the typo).
- [x] The code has been tested.
- [x] All commit messages are properly formatted and commits squashed where appropriate.
- [x] I have included updates to all appropriate documentation.

PR #7709, opened 2022-11-02
UI: Add warning box in settings

tytan652:
### Description

All settings warnings and errors are shown in this box except too verbose one.

Unlike warnings, errors prevent the user to save changes. The user shall fix them to be able to save changes.

Errors are to use as a last resort, checks in the settings UI are preferred.

Settings with many warnings and the verbose "Channels" warning.
Screenshot from 2022-11-02 19-41-27

Dialog when quitting the setting with errors.
Screenshot from 2022-11-08 20-28-04

Dialog when clicking apply or OK with errors.
Screenshot from 2022-11-08 20-27-56

### Motivation and Context

@Warchamp7 requested a change like this one.

### How Has This Been Tested?

- Do what is required to trigger those warnings and error
- Try to save with errors

### Types of changes

### Checklist:


feetstv:
Feedback

- The dialog should summarise the errors. A generic "There are errors" is very confusing.
- The dialog should direct users to the settings page and tab where the error is located. If you're not on the page/tab where the error is located, or haven't scrolled down to see the error, the dialog will be seen as just an annoyance.

Although ideally, instead of these dialogues, it would be better to prevent the errors from being settable in the first place.

tytan652:
> Although ideally, instead of these dialogues, it would be better to prevent the errors from being settable in the first place.

Like I said:
> Errors are to use as a last resort, checks in the settings UI are preferred.

Actually there is only one error, but replacing this one by a better solution is out of scope of this PR.

> - The dialog should summarise the errors. A generic "There are errors" is very confusing.
>
> - The dialog should direct users to the settings page and tab where the error is located. If you're not on the page/tab where the error is located, or haven't scrolled down to see the error, the dialog will be seen as just an annoyance.

The dialog are error neutral, they already appear in red in the warning box. And like I said earlier, this is not meant to be commonly use.

tytan652:
Last pushes were rebase because of the recently merged #7710.

derrod:
I agree with Feets here that the dialog should probably contain some info about the error. Users are bad at reading, an itemised list of all currently set errors here could help. I'd also change the title so the dialog looks something like:

> **Unable To Save**
>  The following errors have been found in your configuration:
>
> * Thing A is wrong
> * Thing B is *very* wrong
>
> You must resolve these errors or discard your changes via "Cancel"

Both dialogs should imo be the same and not offer any destructive option to force the user to interact with the main window and make a concious choice to hit "Cancel" should they want to discard their changes.

tytan652:
Update with #6410 and requests about dialogs, still two because the settings window size issue is still a thing.

Replay buffer warning:
Screenshot from 2022-11-08 20-28-50

tytan652:
Just forgot to remove old sentences in the INI file.

gxalpha:
Are we sure we want *every* warning to be visible on *every* page? This can get crowded really quickly:
This is a perfectly okay setup, yet 1/4th of the hotkey section is blocked despite none of the warnings having anything to do with hotkeys:

While I

feetstv:
> Are we sure we want _every_ warning to be visible on _every_ page? This can get crowded really quickly: This is a perfectly okay setup, yet 1/4th of the hotkey section is blocked despite none of the warnings having anything to do with hotkeys

This was partly my concern with the out-of-context messages. They're more confusing than just putting the issues into the dialog itself in addition to their original, individual locations.

PR #7710, opened 2022-11-02
Add support for native HEVC and ProRes encoders on macOS

PatTheMav:
### Description
Implements https://github.com/obsproject/obs-studio/pull/7010 and adds additional convenience functionality to the app:

* An error is displayed if ProRes is selected as encoder, but neither mov nor mkv are selected as output formats.
* If ProRes is selected as encoder, automatic remuxing uses mov as output container instead of mp4
* If HEVC is selected as encoder and an output container without proper HEVC tagging support is used (e.g. mkv), automatic remuxing correctly sets `hvc1` tagging to the resulting file, enabling default playback functionality on Apple devices
* HEVC is added as possible encoder to the simple setup

Closes https://github.com/obsproject/obs-studio/pull/7010

### Motivation and Context
Enable full breadth of available encoders in macOS devices including ProRes as high-end output format.

### How Has This Been Tested?
* Tested on macOS 13 with ProRes and HEVC recordings.

### Types of changes

### Checklist:


gxalpha (approved):

jcm93:
Some small feedback– I see the code changes for ProRes to default to the .mov container in the auto-remuxer, but in the 'File->Remux Recordings' menu, it still defaults to mp4, and the error text output there on remux is "Recording remuxed, but the file may be incomplete" instead of any error about mp4 being invalid. The user can manually change the output to .mov there, but defaulting to the .mov extension would be nice if the current encoder is ProRes.

PatTheMav:
That is on purpose because we cannot detect the codec used by files added to the remux dialog without doing a file inspection first.

And we cannot assume that just because ProRes is the _current_ encoder that the files added here were encoded using ProRes as well.

If someone is willing to add codec detection of source files to this window, I'm happy to review a respective PR for it.

PR #7711, opened 2022-11-02
mac-videotoolbox: Support P010 and HDR color spaces

gxalpha:
### Description

Adds HDR support to the VT HEVC and ProRes encoders.

### Motivation and Context

OBS on macOS currently can't do any HDR hardware encoding. The only way to properly encode HDR content is with the AOM AV1 software encoder, which isn't optimal.

### How Has This Been Tested?

With HEVC and ProRes, recorded 8-bit SDR (still works the same), 10-bit SDR (pointless), 8-bit HDR (looks like you'd expect it to) and 10-bit HDR all using PQ, by 1) looking at them and 2) throwing them into MediaInfo. Tested 10-bit HLG as well.
Streamed to YouTube in glorious HDR! (with HEVC, ProRes isn't a supported streaming protocol).

Also limited testing by yam on Discord.

### Types of changes

### Checklist:

### DIFF
--- a/plugins/mac-videotoolbox/encoder.c
+++ b/plugins/mac-videotoolbox/encoder.c
@@ -107,7 +107,8 @@ static void log_osstatus(int log_level, struct vt_encoder *enc,
 }
 
 static CFStringRef obs_to_vt_profile(CMVideoCodecType codec_type,
-				     const char *profile)
+				     const char *profile,
+				     enum video_format format)
 {
 	if (codec_type == kCMVideoCodecType_H264) {
 		if (strcmp(profile, "baseline") == 0)
@@ -120,8 +121,14 @@ static CFStringRef obs_to_vt_profile(CMVideoCodecType codec_type,
 			return kVTProfileLevel_H264_Main_AutoLevel;
 #ifdef ENABLE_HEVC
 	} else if (codec_type == kCMVideoCodecType_HEVC) {
-		if (strcmp(profile, "main") == 0)
-			return kVTProfileLevel_HEVC_Main_AutoLevel;
+		if (strcmp(profile, "main") == 0) {
+			if (format == VIDEO_FORMAT_P010) {
+				VT_LOG(LOG_WARNING, "Forcing main10 for P010");
+				return kVTProfileLevel_HEVC_Main10_AutoLevel;
+			} else {
+				return kVTProfileLevel_HEVC_Main_AutoLevel;
+			}
+		}
 		if (strcmp(profile, "main10") == 0)
 			return kVTProfileLevel_HEVC_Main10_AutoLevel;
 #if __MAC_OS_X_VERSION_MAX_ALLOWED >= 120300 // macOS 12.3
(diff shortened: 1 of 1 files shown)


PatTheMav:
> Can you rework your color space handling to match mine? [09aaa8c#diff-fe5ca9c79f5125fb7dc1d00f6292489433b9fd62dc3c88010135990ccd59190a](https://github.com/obsproject/obs-studio/commit/09aaa8cd9d93e983e061abf47bb5bdb5fcb320bc#diff-fe5ca9c79f5125fb7dc1d00f6292489433b9fd62dc3c88010135990ccd59190a)
> 
> EDIT: Ignoring the P216 part.

Yeah that'd be great if those could be incorporated.

PatTheMav (approved):
Looks good to me - if @jpark37 is fine with the changes I guess this would be good to merge.

jpark37:
I just don't know if it's good idea to allow 10-bit formats since Apple doesn't recommend them for ProRes, but I feel like a larger overhaul will be needed in the future if we want to disallow suboptimal combinations (and choose optimal combinations more easily). I'm not opposed to allowing 10-bit for now.

PatTheMav:
> I just don't know if it's good idea to allow 10-bit formats since Apple doesn't recommend them for ProRes, but I feel like a larger overhaul will be needed in the future if we want to disallow suboptimal combinations (and choose optimal combinations more easily). I'm not opposed to allowing 10-bit for now.

Sounds fine to me - merge this and refine formats for ProRes in a separate PR, except if @gxalpha wants to amend it with a check for ProRes and usage of the 16-bit formats, though we’d also need to decide what to do with full range setups then given that 16-bit formats only support partial range.

gxalpha:
I'd be happy to make 10-bit exclusive to HEVC if that'd be desired. Amending it by 16-bit will be difficult given that it's not yet supported by OBS.

gxalpha:
Updated the PR to make `P010` exclusive to HEVC.

PatTheMav:
> Updated the PR to make `P010` exclusive to HEVC.

I'm happy to be corrected on this, but I don't think this was the intended outcome - ProRes 422 is a 10-bit codec, so feeding it 10-bit data is better in any case than feeding 8-bit data (quality wise), especially if it's available.

The most beneficial pixel buffer format would indeed be a 16-bit variant, but that shouldn't exclude ProRes from receiving 10-bit inputs.

@jpark37 or am I mistaken?

jpark37:
I'm planning to make a 4:2:2 16-bit formats PR after the PR goes in. I'm fine with blocking P010 ProRes if that would lead to wasteful processing.

The drawback is that a user won't be able to stream with P010, but also record with ProRes. In an ideal world, OBS would ask the encoders what pixel formats they want based on color space and bit depth (replacing color format), and generate the ideal set of output textures in the best formats, multiple if necessary, all on the GPU. The expedient, somewhat evil, solution to this would be to rename the existing "Color Format" setting to "Stream Color Format" and add another "Recording Color Format" setting that has the same options with an extra default option to "(Use stream color format)," which I would be fine with implementing for 29 although we're running short on time I think.

PatTheMav:
Sounds fair to me, I just find it weird to exclude 10-bit support from an encoder that clearly supports it (even if it's visually not "best") and leave users with just HEVC for 10-bit recordings.

PatTheMav:
Ok, I defer to your suggestions in this case. I haven't checked the recordings myself, otherwise this seems good to merge.

gxalpha:
This was re-approved off-thread by jpark37. Merging.

PR #7723, opened 2022-11-05
Sparkle 2 Update / Mac Update Channels

derrod:
### Description

This PR merges nix-update and win-update, upgrades OBS to Sparkle 2.3.2, and adds support for update channels on macOS.

Closes #7869 if merged first.

The new Sparkle feed URLs will be:
- `https://obsproject.com/osx_update/updates_arm64_v2.xml`
- `https://obsproject.com/osx_update/updates_x86_64_v2.xml`

This is due to the incompatibilites between Sparkle 1 and 2 and the addition of channels.

### Motivation and Context

Want update channels on mac.

### How Has This Been Tested?

Compiles and runs, channels work as expected on macOS (tested with local webserver and some appcasts generated using Sparkle's `generate_appcast` tool).

### Types of changes

### Checklist:


gxalpha:
Marking as draft to prevent accidental merges

UI/update/sparkle-updater.mm
```diff
@@ -0,0 +1,100 @@
...
+					       @selector(canCheckForUpdates))];
+	}
+}
+
+@end
+
+// Creates and starts the updater. There's nothing else required.
+OBSSparkle::OBSSparkle(const char *branch, QAction *checkForUpdatesAction)
+{
+	@autoreleasepool {
```
derrod, line 76: This file is basically mostly just copied from Sparkle's Qt example (it works!), not sure if we really need this here. https://sparkle-project.org/documentation/programmatic-setup/#:~:text=Create%20an%20Updater%20in%20Qt

derrod:
Rebased on #6907's latest changes.

Also have a macOS screenshot (I don't yet know why this is misaligned, it's fine on Windows...):
Screenshot 2022-11-06 at 18 31 25

derrod:
Still not sure why, but adding that `fieldGrowthPolicy` fixed it:
Screenshot 2022-11-06 at 18 38 53

derrod:
Rebased on master now that #6907 is merged. Also made sure that update check runs in background after switching channels.

derrod:
I've been able to confirm that updating from one build to another via Sparkle 2 works. Downloading the following build should update to the next one: https://github.com/obsproject/obs-studio/actions/runs/3521880459

cmake/Modules/ObsHelpers_macOS.cmake
```diff
@@ -360,27 +360,17 @@ function(setup_obs_bundle target)
...
+        "plutil -replace SUFeedURL -string https://obsproject.com/osx_update/stable/updates_v2.xml \"$<TARGET_BUNDLE_CONTENT_DIR:${target}>/Info.plist\""
       VERBATIM)
 
     add_custom_command(
       TARGET ${target}
       POST_BUILD
       COMMAND
         /bin/sh -c
-        "plutil -replace SUPublicDSAKeyFile -string OBSPublicDSAKey.pem \"$<TARGET_BUNDLE_CONTENT_DIR:${target}>/Info.plist\""
+        "plutil -replace SUPublicEDKey -string \"<INSERT ED KEY HERE>\" \"$<TARGET_BUNDLE_CONTENT_DIR:${target}>/Info.plist\""
```
derrod, line 371: This either needs the public key to be added, or we can make it configurable and then add it as an env var/secret to CI.

CI/macos/02_build_obs.sh
```diff
@@ -69,6 +69,14 @@ _configure_obs() {
         UNITTEST_OPTIONS="-DENABLE_UNIT_TESTS=ON"
     fi
 
+    if [ "${CI}" ]; then
+        if [ "${ARCH}" = "x86_64" ]; then
+            BUILD_NUMBER=${GITHUB_RUN_ID}
+        else
+            BUILD_NUMBER=$((${GITHUB_RUN_ID} + 1))
```
derrod, line 76: This exists as a workaround for Sparkle only allowing one item per "version" in the appcast, even if they are on different channels.

Sadly the only real solution to this would be to continue having separate arm64/x86 appcasts, which I may go back to tomorrow for the sake of fixing this.

cmake/bundle/macOS/bundleutils.cmake
```diff
@@ -117,13 +117,13 @@ if(EXISTS
   execute_process(
     COMMAND
       /usr/bin/codesign --remove-signature
-      "${CMAKE_INSTALL_PREFIX}/${_BUNDLENAME}/Contents/Frameworks/Sparkle.framework/Versions/A/Resources/Autoupdate.app"
+      "${CMAKE_INSTALL_PREFIX}/${_BUNDLENAME}/Contents/Frameworks/Sparkle.framework/Versions/B/Autoupdate"
```
derrod, line 120: I assume this could be removed entirely since there's no app bundle to (re-)sign anymore?

CI/macos/02_build_obs.sh
PatTheMav: That's annoying, maybe report this issue upstream, as having arm64 and x86_64 items for the same version+build seems like a very obvious use case for any application that has to ship an entire Google browser with itself..

UI/update/sparkle-updater.mm
```diff
@@ -0,0 +1,105 @@
+#include "mac-update.hpp"
+
+#include <qaction.h>
+
+#import <Cocoa/Cocoa.h>
+#import <Sparkle/Sparkle.h>
+
+#ifdef __aarch64__
+#define BRANCH_SUFFIX "_arm64"
```
PatTheMav, line 9: Don't use macros for constants, use _actual_ constants for this:

```
NSString *const OBSBranchSuffix = @"_arm64";
```

We really need to stop using macros and macro constants as much as possible in our codebase moving forward.

cmake/Modules/ObsHelpers_macOS.cmake
PatTheMav: Add it as cache variables, like `-DSPARKLE_APPCAST_URL` and `-DSPARKLE_KEY_FILE`.

The _actual_ fix would be to have a separate `Info.plist.in` just for the app bundle, add the keys there, with both of these cache variables as their values, and just have CMake generate the file automatically.

If Sparkle is not enabled or those cache keys are not set, the plist will just contain empty values, which is fine by me.

CI/macos/02_build_obs.sh
derrod: I opened a discussion for now to see what they think: https://github.com/sparkle-project/Sparkle/discussions/2283

cmake/Modules/ObsHelpers_macOS.cmake
derrod: For ED keys we actually can't use a key file, but have to put the key itself into the property. Same CMake changes apply though.

UI/CMakeLists.txt
```diff
@@ -448,12 +443,20 @@ elseif(OS_MACOS)
...
-              nix-update/crypto-helpers-mac.mm
-              nix-update/nix-update.cpp
-              nix-update/nix-update.hpp
-              nix-update/nix-update-helpers.cpp
-              nix-update/nix-update-helpers.hpp)
+      PRIVATE update/crypto-helpers.hpp update/crypto-helpers-mac.mm
+              update/shared-update.cpp update/shared-update.hpp
+              update/update-helpers.cpp update/update-helpers.hpp)
+
+    if(ENABLE_SPARKLE_UPDATER)
```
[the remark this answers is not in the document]
derrod: I'd say that's something for later, although I'm not sure what benefit there would be to only optionally loading Sparkle?

derrod:
Reminder to self, if/when #7855 is merged update the ifdef to include macOS.

derrod:
Sparkle 2.3.1 has been released since this PR was made: https://github.com/sparkle-project/Sparkle/releases/tag/2.3.1

The main difference is that `generate_appcast` now no longer supports adding DSA signatures at all. This shouldn't be an issue for us as the migration allows changing key types as long as the developer signature is the same, but I would like to see #7869 perhaps being shipped with 29.0.x just to be on the safe side before updating to Sparkle 2.

derrod:
Updated to Sparkle 2.3.1 now. I think all that's left is getting the EdDSA public key from @jp9000, though arguably we could merge it before we have that and then add it later.

derrod:
The Sparkle 2 public EdDSA key has now been added. This PR is therefore "finished" any may be reviewed/merged at your earliest convenience.

PatTheMav:
During my testing it occurred to me that you're not catching the edge case of someone manually enabling Sparkle, but not provide the public key and new appcast URL, so you end up with the source files added but cannot successfully compile.

derrod:
> During my testing it occurred to me that you're not catching the edge case of someone manually enabling Sparkle, but not provide the public key and new appcast URL, so you end up with the source files added but cannot successfully compile.

Should that be a CMake Error or should the `ENABLE_SPARKLE` option be retired in favour of automatically building it when appcast/key are set?

PatTheMav:
> > During my testing it occurred to me that you're not catching the edge case of someone manually enabling Sparkle, but not provide the public key and new appcast URL, so you end up with the source files added but cannot successfully compile.
> 
> Should that be a CMake Error or should the `ENABLE_SPARKLE_UPDATER` option be retired in favour of automatically building it when appcast/key are set?

I'm fine with enabling it only when the values are set as cache variables (similar to our oauth modules).

derrod:
> I'm fine with enabling it only when the values are set as cache variables (similar to our oauth modules).

Done.

PatTheMav (approved):

PR #7730, opened 2022-11-07
UI: Remove number from multiview labels

Warchamp7:
### Description
Removes the number and dash that prefix the scene name in the multiview

### Motivation and Context
The primary use of a multiview is in larger scale, professional productions. In these environments, it is very common for a director to need to be calling the shots as they are labeled in the multiview.

Scenes themselves have arbitrary names and as such - in the absence of the ability to customize multiview labels - it makes the most sense to display labels with scene names exactly as they exist. A scenario where a label displays "3 - CAM 1" or "2 - 4 BOX" is undesirable. If numbers before name are desired by someone, they can name the scene as such with the added bonus of their scene list having parity with the multiview display.

### How Has This Been Tested?
Viewed multiview with my eyes, used a build of 28 with this change cherrypicked in for a 5 day broadcast comprising 20+ staff viewing it with their eyes.

### Types of changes

### Checklist:

### DIFF
--- a/UI/multiview.cpp
+++ b/UI/multiview.cpp
@@ -174,9 +174,8 @@ void Multiview::Update(MultiviewLayout multiviewLayout, bool drawLabel,
 		multiviewScenes.emplace_back(OBSGetWeakRef(src));
 		obs_source_inc_showing(src);
 
-		std::string name = std::to_string(numSrcs) + " - " +
-				   obs_source_get_name(src);
-		multiviewLabels.emplace_back(CreateLabel(name.c_str(), h / 3));
+		multiviewLabels.emplace_back(
+			CreateLabel(obs_source_get_name(src), h / 3));
 	}
 
 	obs_frontend_source_list_free(&scenes);


UI/multiview.cpp
```diff
@@ -174,8 +174,7 @@ void Multiview::Update(MultiviewLayout multiviewLayout, bool drawLabel,
 		multiviewScenes.emplace_back(OBSGetWeakRef(src));
 		obs_source_inc_showing(src);
 
-		std::string name = std::to_string(numSrcs) + " - " +
-				   obs_source_get_name(src);
+		std::string name = obs_source_get_name(src);
 		multiviewLabels.emplace_back(CreateLabel(name.c_str(), h / 3));
```
gxalpha, line 178: This could just use `CreateLabel(obs_source_get_name(src), ...)` without the detour via the std string now

gxalpha:
Didn't notice the first time, the commit title also needs the `UI: ` prefix

UI/multiview.cpp
Warchamp7: Done

sebastian-s-beckmann (approved):
lgtm code wise

PR #7731, opened 2022-11-07
CI: Ignore libdshowcapture deps when formatting

derrod:
### Description

Ignores the `external/` folder in libdshowcapture when running `check-format.sh`

### Motivation and Context

Have you ever gotten annoyed that running the formatting script results in a submodule change on your Mac?
Have you ever had to manually go in and `git restore` that folder just to get a clean `git status` output?

Well I have so that's why I added this (btw no clue why this doesn't cause issues on CI)

### How Has This Been Tested?

Ran locally, no longer formats that folder.

### Types of changes

### Checklist:

### DIFF
--- a/CI/check-format.sh
+++ b/CI/check-format.sh
@@ -48,7 +48,8 @@ find . -type d \( \
     -path ./plugins/enc-amf -o \
     -path ./plugins/mac-syphon/syphon-framework -o \
     -path ./plugins/obs-outputs/ftl-sdk -o \
-    -path ./plugins/obs-websocket/deps \
+    -path ./plugins/obs-websocket/deps -o \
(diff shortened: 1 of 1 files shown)


RytoEX:
Would it not be better to add a .clang-format file to the `external` directory and have it set to disable clang-format?

derrod:
Maybe? We should do that with websockets as well then I think.

RytoEX:
> Maybe? We should do that with websockets as well then I think.

I don't recall if the obs-websocket repo has its own clang-format rules, but yes, generally we should avoid running against any submodule unless we want to enforce our clang-format rules on them for some reason.

derrod:
Closing in favour of https://github.com/obsproject/libdshowcapture/pull/48

PR #7736, opened 2022-11-08
mac-capture: Disable all SCK modes besides WindowCapture on macOS 12

PatTheMav:
### Description
SCK has too many open issues on macOS 12 to enable full functionality on that version. Window Capture has the biggest performance uplift and so far the least amount of quirks, so leave this variant (with the "Beta" qualifier) for macOS versions before Ventura.

### Motivation and Context
Reduce support effort on macOS 12 as we are aware of SCK capture methods having unfixed issues on that version of the OS and also that there will not be fixes for them.

Legacy Display Capture works fine as-is, SCK Window-Capture has a big performance benefit compared to the legacy variant, but caveats might persist.

### How Has This Been Tested?
Tested on macOS 12 runner.

### Types of changes

### Checklist:


gxalpha:
I’m a bit worried about the impact on existing setups - what happens to someone who already has selected Display Capture updating from 28.1 to 29?

Migration to macOS 13 with this is pretty unclean - changing defaults between macOS versions means changing setups, so anyone updating to macOS 13 will have to re-select window capture.

Also remind me, what was the issue with Application Capture on macOS 12 (I only remember all the Display Capture problems).

PatTheMav:
Application Capture suffers from the same issues as all the other SCK features - in theory the entire feature should be disabled for macOS 12, but I'm willing to throw such users a bone and give them a beta-quality window capture instead.

Limiting the functionality to Window Capture reduces the amount of problematic use cases, hence why the other modes are disabled.

As for upgrading to 29 on macOS 12, breaking setups is intended - there is no way to "recover" that functionality, the source should stay empty. We should've never launched 28 with that, alas that's what we decided on back then.

Upgrading from 12 to 13 should be doable though, just requires explicitly storing the capture type even on 12.

PatTheMav:
Implemented the following changes:

* No actual `SCStream` instance is created when the pre-existing capture type is not Window Capture
* An error message is displayed in place of the capture type dropdown informing about the availability on macOS 13 or newer
* Error message will disappear after configuring a window to capture and re-opening the properties window
* The capture type is explicitly written to the configuration on macOS 12, when updating to macOS 13 this should ensure the capture type doesn't suddenly change
* The warning about audio capture only being available on macOS 13 is also displayed with the warning text decorations now

gxalpha:
Hmm, when I Cmd-F replace all `macOS 13` instances to `macOS 14` to "simulate" being on an old version, create an SCK source, select a window, then undo my version changes and re-open, I still get switched to Display Capture. To be honest, I'm not completely sure how we could properly fix this.
As an idea, maybe we leave the dropdown and default, and just show the warning message and no frames when Display/Application Capture is selected? Not the greatest UX, but would prevent users from getting things messed up completely when updating OBS or macOS.

PatTheMav:
> Hmm, when I Cmd-F replace all `macOS 13` instances to `macOS 14` to "simulate" being on an old version, create an SCK source, select a window, then undo my version changes and re-open, I still get switched to Display Capture. To be honest, I'm not completely sure how we could properly fix this.
> As an idea, maybe we leave the dropdown and default, and just show the warning message and no frames when Display/Application Capture is selected? Not the greatest UX, but would prevent users from getting things messed up completely when updating OBS or macOS.

Yeah let me try that, effectively just requires removing some fancy workarounds.

PatTheMav:
@gxalpha Capture type selection is restored, but no actual capture is initialised on Desktop or App capture, instead an error is displayed below the audio warning.

gxalpha (approved):
Personally I like this solution a lot. The string could maybe be feature something along the lines of "Use Display Capture or Window capture instead" but at this point I'm nitpicking.

Tested that it works fine, both how it will behave when upgrading to OBS 29 on macOS 12 and then when upgrading to macOS 13, using the methods outlined earlier.

With https://github.com/obsproject/obs-studio/pull/7497 merged it looks this will need to be rebased though.

gxalpha (approved):

PR #7749, opened 2022-11-10
libobs: Fix mixed wide and narrow characters

GkvJwa:
### Description

Fix mixed wide and narrow characters on windows
vsnprintf cannot format wide and narrow at the same time
### Motivation and Context

The log is found in the windows environment as follows:
12:37:44.371: Sec. Software Status:
12:37:44.372: 
12:37:44.372: 
12:37:44.373: Current Date/Time: 2022-11-10, 12:37:44
Then I read the code and found that vsnprintf goes to format wide characters and utf8 strings at the same time
### How Has This Been Tested?

Checked that the log analyser continues to correctly read the Windows version
### Types of changes

### Checklist:


libobs/obs-windows.c
```diff
@@ -316,8 +316,23 @@ static void log_security_products_by_type(IWSCProductList *prod_list, int type)
...
+		size_t utf8_name_length =
+			os_wcs_to_utf8(name, name_length, NULL, 0);
+
+		if (utf8_name_length) {
+			utf8_name = bmalloc(utf8_name_length);
+			os_wcs_to_utf8(name, name_length, utf8_name,
+				       utf8_name_length);
+		}
+
+		if (utf8_name) {
```
notr1ch, line 330: This null check seems unnecessary since the code above already writes to the pointer without checking.

GkvJwa: review again

notr1ch: The check can be removed entirely as `bmalloc` will never return NULL.

GkvJwa: > The check can be removed entirely as `bmalloc` will never return NULL.
review again

GkvJwa: > The check can be removed entirely as `bmalloc` will never return NULL.

review?

gxalpha (changes requested):
Please make sure to always squish your commits back together into one commit after making changes.

Also please show some patience, most people here are volunteers doing this in their free time. You may need to wait a few days (in some cases even weeks) for someone to re-review.

PR #7759, opened 2022-11-11
UI, obs-ffmpeg: Various logging enhancements

tt2468:
### Description
Fixes a few logging-related issues I've noticed while playing with OBS.

### Motivation and Context
I like logs that are easy to read and cohesive.

### How Has This Been Tested?
- OBS starts
- OBS streams
- OBS happy :)

### Types of changes

### Checklist:

### DIFF
--- a/plugins/obs-ffmpeg/obs-ffmpeg-srt.h
+++ b/plugins/obs-ffmpeg/obs-ffmpeg-srt.h
@@ -83,7 +83,7 @@ static int libsrt_neterrno(URLContext *h)
 	SRTContext *s = (SRTContext *)h->priv_data;
 	int os_errno;
 	int err = srt_getlasterror(&os_errno);
-	blog(LOG_ERROR, "[obs-ffmpeg mpegts muxer / libsrt] : %s\n",
+	blog(LOG_ERROR, "[obs-ffmpeg mpegts muxer / libsrt]: %s",
 	     srt_getlasterror_str());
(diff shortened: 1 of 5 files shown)


plugins/obs-ffmpeg/obs-ffmpeg-srt.h
```diff
@@ -801,8 +801,8 @@ static int libsrt_write(URLContext *h, const uint8_t *buf, int size)
 #endif
 		if (time > (s->time + 60.0)) {
 			srt_bistats(s->fd, &perf, 0, 1);
-			blog(LOG_INFO,
-			     "[obs-ffmpeg mpegts muxer / libsrt] : rtt [%.2f ms], link bandwidth [%.1f Mbps]\n",
+			blog(LOG_DEBUG,
```
pkviet, line 804: i'd prefer to keep this one to info rather than debug, that's useful health info during the stream. It logs currently every minute. You find this too spammy ?

tt2468: I changed it to debug, as users can be directed to launch OBS with --verbose in order to log this message. If it were to only log when some network issue detection logic triggers, I would say it's okay for it to be INFO. But there are people who run OBS streaming 24/7, and any message logged on an interval will eventually create an incredibly large log file.

pkviet: ok works for me then

PR #7767, opened 2022-11-13
mac-videotoolbox: Default to High profile

gxalpha:
### Description

Changes the default profile of the VT H.264 encoder to High.
HEVC stays at main.

### Motivation and Context

[R1ch's stream analyzer](https://r-1.ch/analyzer/) got mad at me for using the main profile.
We really should be using High in Simple Mode, I consider current behavior to be a bug.

### How Has This Been Tested?

Tested that recordings using the simple Apple H.264 streaming encoder use the High profile now (whereas before it was Main) by throwing the file into MediaInfo.
Tested that x264 recordings using the stream encoder in simple mode use the High profile both before and after.
Checked manually (by looking at the code) that both NVENC and AMF already use High.

Tested that HEVC recordings using the simple Apple HEVC streaming encoder still use the Main profile (which is correct for HEVC).

### Types of changes

### Checklist:

### DIFF
--- a/plugins/mac-videotoolbox/encoder.c
+++ b/plugins/mac-videotoolbox/encoder.c
@@ -1328,7 +1328,10 @@ static void vt_defaults(obs_data_t *settings, void *data)
 	obs_data_set_default_int(settings, "max_bitrate", 2500);
 	obs_data_set_default_double(settings, "max_bitrate_window", 1.5f);
 	obs_data_set_default_int(settings, "keyint_sec", 0);
-	obs_data_set_default_string(settings, "profile", "main");
+	obs_data_set_default_string(
+		settings, "profile",
+		type_data->codec_type == kCMVideoCodecType_H264 ? "high"
+								: "main");
 	obs_data_set_default_int(settings, "codec_type",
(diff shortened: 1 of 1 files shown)


derrod (commented):
The "proper" solution is probably to change the default at in https://github.com/obsproject/obs-studio/blob/master/plugins/mac-videotoolbox/encoder.c#L1331

See NVENC for example which conditionally uses "high" for AVC and "main" for everything else:

https://github.com/obsproject/obs-studio/blob/master/plugins/obs-ffmpeg/obs-ffmpeg-nvenc.c#L434-L435

gxalpha:
> The "proper" solution is probably to change the default

You think so? This would change every existing setup (although it probably wouldn't matter too much).

derrod:
I'm not sure it would affect existing setups unless this is one of those cases where OBS doesn't save your settings unless you change them from the default. Shouldn't break anything either though even if it does change.

gxalpha:
It is indeed one of those cases where defaults aren't saved.
I guess you're right though, there's virtually no reason for anyone to use main for H.264 these days.

PR #7770, opened 2022-11-13
CI,obs-vst: Update Flatpak KDE Runtime to version 6.4

tytan652:
### Description

Closes #7132

The 6.4 runtime rely on Freedesktop 22.08 which provide a more recent version of Mesa and PipeWire.

The switch to 22.08 also requires to migrate paths of VST 2 plugins.

### Motivation and Context

We need to bump the runtime version to provide a more recent version of Mesa and other stuff (Qt 6.4, remove custom PipeWire module).

### How Has This Been Tested?

- I installed `org.freedesktop.LinuxAudio.Plugins.DISTRHO-Ports//21.08` and `org.freedesktop.LinuxAudio.Plugins.DISTRHO-Ports//22.08`.
- Add a VST filter (from 21.08) with OBS Flatpak from Flathub (so 28.1.2 with 6.3)
- Run OBS Flatpak with this change, the filter stilsl work because the VST path has successfully been updated (and now use the 22.08 one)

Note: for unknown reason I had to build the flatpak with "beta" or "stable" has branch to test it, "master" seems to not allow to see VST filters (6.3 or 6.4).

To build and run the flatpak with `beta` branch
```
flatpak-builder _flatpak_build CI/flatpak/com.obsproject.Studio.json --force-clean --verbose --default-branch=beta
flatpak-builder --run _flatpak_build CI/flatpak/com.obsproject.Studio.json obs
```

### Types of changes

### Checklist:

### DIFF
--- a/.github/workflows/flatpak.yml
+++ b/.github/workflows/flatpak.yml
@@ -50,7 +50,7 @@ jobs:
     env:
       FLATPAK_BUILD_PATH: flatpak_app/files/share
     container:
-      image: bilelmoussaoui/flatpak-github-actions:kde-6.3
+      image: bilelmoussaoui/flatpak-github-actions:kde-6.4
       options: --privileged
     strategy:
       matrix:
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -378,7 +378,7 @@ jobs:
       run:
         shell: bash
     container:
-      image: bilelmoussaoui/flatpak-github-actions:kde-6.3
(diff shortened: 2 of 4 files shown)


GeorgesStavracas (approved):
Generally LGTM. I can't comment on the VST change, it looks correct but I cannot prove it is. Someone should also check that.

WizardCM (approved):
Verified 6.4 undocking works better than 6.3 - docks will stick to the cursor rather than being offset strange amounts.

hfiguiere:
you should have removed jack2 as well. The headers are in the runtime now.

PR #7778, opened 2022-11-14
obs-frontend-api: Add function to get the virtual camera view

exeldro:
### Description
Add function to get the virtual camera view

### Motivation and Context
Since in OBS 28 a view is used for the virtual camera instead of the main output, other channels that are set on the output in the [downstream keyer](https://obsproject.com/forum/resources/downstream-keyer.1254/) plugin are not visible anymore in the output of the virtual camera. This change will allow plugins like the downstream keyer plugin to set the other channels of the view for the virtual camera.

### How Has This Been Tested?
On windows 64 bit with changes to the downstream keyer plugin.

### Types of changes

### Checklist:

### DIFF
--- a/UI/api-interface.cpp
+++ b/UI/api-interface.cpp
@@ -3,6 +3,7 @@
 #include "qt-wrappers.hpp"
 #include "window-basic-main.hpp"
 #include "window-basic-main-outputs.hpp"
+#include "window-basic-vcam-config.hpp"
 
 #include <functional>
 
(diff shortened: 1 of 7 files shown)


tytan652:
Please note that:
- There is a refactor of the feature #7169.
- And that the view can be destroyed at any moment.

tytan652:
After some reflection, I think that the fact that the Virtual Camera does not show what is really put in the Program View can be seen as a bug.

tytan652:
Since #7169 got merged, this PR is no longer needed since the former fix what this PR was meant to fix an another.

So I close this PR, @exeldro feel free to comment back if you have an issue with this decision.

PR #7787, opened 2022-11-16
CI: Fix building in PowerShell 7.3.x

RytoEX:
### Description

The backtick double-quote pattern does not work in PowerShell 7.3.x. There are still some other possible PowerShell 7.3.x issues in the packaging steps, but let's fix this first to get regular builds working in PowerShell 7.3.x again.

### Motivation and Context

Updates are coming to the Windows CI images. Don't want CI to immediately fail all builds.

Some devs are getting PowerShell 7.3. Want them to be able to build locally.

### How Has This Been Tested?

Tested locally on Windows 11 with PowerShell 7.3.

### Types of changes

### Checklist:

### DIFF
--- a/CI/windows/02_build_obs.ps1
+++ b/CI/windows/02_build_obs.ps1
@@ -63,31 +63,61 @@ function Configure-OBS {
     $BuildDirectoryActual = "${BuildDirectory}$(if (${BuildArch} -eq "x64") { "64" } else { "32" })"
     $GeneratorPlatform = "$(if (${BuildArch} -eq "x64") { "x64" } else { "Win32" })"
 
-    $CmakeCommand = @(
-        "-G", ${CmakeGenerator}
-        "-DCMAKE_GENERATOR_PLATFORM=`"${GeneratorPlatform}`"",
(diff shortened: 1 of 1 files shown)


tommyvct:
This is temporary, right?

Does the new syntax without the backticks works on pwsh 7.2?

Are you sure this change in behaviour is not at fault on the pwsh side?

RytoEX:
> This is temporary, right?
> 
> Does the new syntax without the backticks works on pwsh 7.2?
> 
> Are you sure this change in behaviour is not at fault on the pwsh side?

We've previously tested the syntax without the backtick+double-quote syntax in PowerShell 7.2 and it does not work unless we used that syntax. I've tested the old backtick+double-quote syntax in PowerShell 7.3, and it does not work if used. Checking the error messages led me to conclude that this syntax is producing bad input to CMake. My guess is that array handling (or passing args via command exection) does some implicit quoting of strings. Feel free to debug or dig in if you like, but this is appropriately version gated to continue using the existing syntax that _has worked_ on PowerShell 7.1.x and 7.2.x, while using the new syntax only in PowerShell 7.3.x and later.

PR #7797, opened 2022-11-17
libobs: Allow overriding video resolution per view

aet:
### Description

- Add new variant of obs_view_add to customize view specific canvas/encoder output resolution

### Motivation and Context

I recently got familiar with https://github.com/obsproject/obs-studio/pull/6577 (cc @chippydip ) that was exactly what my plugin needed at the time, for potentially doing multiple outputs with separate encoder instances at the time.

Later on, the requirements changed and I'd prefer to encode an output with a custom resolution. I'm also using a scene with a custom resolution set as view source.

The multiple video mixes PR already did most of the work for making this happen, but this PR is just a small addition to ensure `obs_video_info` is a part of `obs_core_video_mix` context and reduce the direct use of `obs->video.main_mix`.

### How Has This Been Tested?

- Tested on a Mac and PC (x264/nvenc variants, with and without scaling)
- Tested with one or multiple streaming outputs running in parallel, potentially with multiple different resolutions in use

### Types of changes

### Checklist:

### DIFF
--- a/libobs/obs-internal.h
+++ b/libobs/obs-internal.h
@@ -284,6 +284,7 @@ struct obs_core_video_mix {
 	volatile bool gpu_encode_stop;
 
 	video_t *video;
+	struct obs_video_info ovi;
 
 	bool gpu_conversion;
 	const char *conversion_techs[NUM_CHANNELS];
@@ -292,7 +293,6 @@ struct obs_core_video_mix {
 	float conversion_height_i;
 
 	float color_matrix[16];
-	enum obs_scale_type scale_type;
 };
 
 extern struct obs_core_video_mix *
@@ -324,9 +324,6 @@ struct obs_core_video {
 	uint32_t lagged_frames;
 	bool thread_initialized;
 
-	uint32_t base_width;
-	uint32_t base_height;
-
 	gs_texture_t *transparent_texture;
 
 	gs_effect_t *deinterlace_discard_effect;
@@ -338,7 +335,6 @@ struct obs_core_video {
 	gs_effect_t *deinterlace_yadif_effect;
 	gs_effect_t *deinterlace_yadif_2x_effect;
 
-	struct obs_video_info ovi;
 	float sdr_white_level;
 	float hdr_nominal_peak_level;
 
@@ -497,6 +493,8 @@ extern gs_effect_t *obs_load_effect(gs_effect_t **effect, const char *file);
 extern bool audio_callback(void *param, uint64_t start_ts_in,
 			   uint64_t end_ts_in, uint64_t *out_ts,
 			   uint32_t mixers, struct audio_output_data *mixes);
+
+extern struct obs_core_video_mix *get_mix_for_video(video_t *video);
 
 extern void
 start_raw_video(video_t *video, const struct video_scale_info *conversion,
--- a/libobs/obs-encoder.c
+++ b/libobs/obs-encoder.c
@@ -187,7 +187,9 @@ static inline bool has_scaling(const struct obs_encoder *encoder)
 
 static inline bool gpu_encode_available(const struct obs_encoder *encoder)
 {
-	struct obs_core_video_mix *video = obs->video.main_mix;
+	struct obs_core_video_mix *video = get_mix_for_video(encoder->media);
+	if (!video)
+		return false;
 	return (encoder->info.caps & OBS_ENCODER_CAP_PASS_TEXTURE) != 0 &&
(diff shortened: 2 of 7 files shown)


tt2468:
Please split the signals commit into a separate PR, since they cover different topics.

libobs/obs-internal.h
```diff
@@ -284,15 +284,14 @@ struct obs_core_video_mix {
 	volatile bool gpu_encode_stop;
 
 	video_t *video;
+	struct obs_video_info ovi;
 
-	bool gpu_conversion;
```
tt2468, line 288: Does this have to be relocated? Moving this to ovi means writing to ovi, which does technically change behavior for `obs_get_video_info()`, returning the current gpu conversion state instead of previously the requested gpu conversion state.

tt2468:
What does behavior look like if I have one scene active in two views of different resolutions/aspect ratios? Does the scene get rendered as the canvas resolution for each view independently?

libobs/obs-internal.h
aet: It doesn't have to be relocated, but thought it would be potentially nice to have if someone needs to query that state being turned off. I'll revert this to have less potential trouble

tt2468: Yeah, it's best to try to preserve behavior as much as possible with a change like this, since it complicates bug testing. If there's a need to get the gpu conversion state, then a new libobs function may be a better option.

aet:
When scenes are nested into another scene for rendering, they're using a custom resolution with scene->custom_size set. By default every scene is still using the main mix resolution, to play it safe. On a custom view with larger or smaller canvas area compared to main mix, the behavior should be the same as in the preview view after you've changed the base canvas size. :) You might be over or underfilling the entire area.

tt2468:
So if scenes still use the main mix canvas resolution, what exactly is the recommended usage of an auxiliary mix with a different resolution?

aet:
In my use case, I'm creating a private scene with a custom resolution that follows the aux view resolution. Then creating a view similar to OBS Multiview through adding nested scenes for rendering into this private scene. Though typically I'm creating a scene with double wide resolution to what the user settings are, to show a custom scene and OBS program scene side-by-side.

tt2468:
Please add a commit which documents the additional functionality.

Looking this over, the FPS info in *ovi for additional views is ignored?

jp9000:
I love this PR. This is very well done.

We can add documentation a bit later. Please make sure to do so, but I'm going to merge this for now so we can get some master branch testing and experimentation.

Well done!

RytoEX:
To add to the above, we've tentatively merged this to make it available to a wider testing pool. We'll be looking at this during the upcoming beta period in case it requires any follow-ups.

As mentioned, please be sure to add the requested documentation in a separate PR.

aet:
@jp9000 @tt2468 Thanks for merging and to get some testing mileage!

I was trying to look for some starting pointers where to add obs_view related documentation additions before making the PR, but seems it's one of the areas that's missing entirely, maybe being so low-level, so I followed the inline comments to the header. Any suggestions how to help? Can always expand the oneliner comment from the header as bare minimum to explain potential uses and things not expected to work (fps override below).

The FPS info from ovi should be passed to encoder specific instances, but I only tested configuration variations to base resolution/output resolution from main mix ovi. Abstracting `video_frame_interval_ns` and friends seems to get messy fast and there's more public APIs involved without a context, but maybe in the meantime the additional views could override critical parts of the user provided ovi by using the values from main mix? Looking at the struct, maybe fps_num/fps_den would be enough.

I usually code on a Mac but started to do more testing on a PC too once my plugin (unreleased, in development) started to prefer actual modifications to OBS core. I just realized that current OBS 28.x releases don't handle multiple encoders and separate views properly when using GPU encoder path like nvenc, everything is just getting the main mix view.

Looks like fixing gpu_encode_available/start_gpu_encode/stop_gpu_encode to do `get_mix_for_video` lookup should fix the issue, changes part of this PR.

I've only been getting familiar with OBS code a few months now so not all familiar with release cycles and plans, but hopefully there's some releases left for 28.x where my changes could make it. :)

tt2468:
It sounds like we may have an issue then if the fps details are being passed to encoders, as there is one graphics thread which drives all frame output, and it will run at the rate of the main view. If encoders are thinking they are receiving 30fps, but the graphics thread is outputting at 60fps, that could cause problems.

Perhaps we need to explicitly override any additional views with the fps_num and fps_den value of the main mix?

Issue #7784, opened 2022-11-15
AV1: SVT and AOM encoders does not respect CQ/CQP value


flaeri:
### Operating System Info

Windows 11

### OBS Studio Version

28.1.2

### OBS Studio Log URL

https://obsproject.com/logs/X5mQQyP4Z0bSdBmC

### Expected Behavior

When using the CQ/CQP rate control I expect the value to influence the actual quant and size of the files produced.

### Current Behavior

The CQ/CQP value for both encoders are ignored, and it seems some default is used instead.

### Steps to Reproduce

1. Pick SVT/AOM av1 encoder
2. Change the rate control to CQP
3. Set value to 50, and record. Note the size/quant
4. Set the value to 1, and record. Note the size/quant

### Anything else we should know?

A user in the discord mentioned this issue, and I could easily replicate.

libAOM reports in the log: `warning: Neither bitrate nor constrained quality specified, using default CRF of 32`

libSVT does not report anything in normal nor verbose log.

Example videos, neatly labeled:
[Videos.zip](https://github.com/obsproject/obs-studio/files/10016423/Videos.zip)

aom-av1, cqp 1:
![libaom-cqp1](https://user-images.githubusercontent.com/50419942/202032713-13c2b338-c871-404a-b535-ee7bcf80330d.png)

aom-av1, cqp 50:
![libaom-cqp50](https://user-images.githubusercontent.com/50419942/202032760-c7e09b4c-dd28-440f-adf5-b19a1e7f2473.png)

Using a bitstream analyzer, I compared the quantizer of cqp 1 vs 50 in libaom:
```
aom cqp 1, 
first visible frame quant: 87
second and third frame: 176

aom cqp 50, 
first visible frame quant: 87
second and third frame: 176
```

flaeri:
LibAOM:

If I replace https://github.com/obsproject/obs-studio/blob/301730f40c6c250eafbcedcf37dfc89125a0c795/plugins/obs-ffmpeg/obs-ffmpeg-av1.c#L102 with `av_opt_set_int(enc->ffve.context->priv_data, "crf", cqp, 0);` it seems to work as expected, and does not throw errors, nor crash (at least with very basic testing).

This should result in AOM_Q rate control (we dont want CQ (constrained)), as bitrate is `0`

This is probably not the correct way to do this, but I am unfortunately not capable enough PR a proper fix.
[...]

PR #7799, opened 2022-11-18
obs-ffmpeg: Fix SVT-AV1 rate control mode selection

RytoEX:
### Description

In updated versions of FFmpeg (5.1.x) and SVT-AV1 (1.2.0/1.3.0), the rate contol mode selection mechanism has changed, causing CBR to throw an "Invalid Parameter" error and the encoder will fail to start. This appears to be caused by us specifying rc_max_rate, which will also fail on FFmpeg CLI. Adjust how we manage rate control methods, appropriately version-gated, to work with both our current dependency versions and the versions that we'd like to use.

Alternatively, if it's more appealing as a simpler temporary fix, we could just ifdef specifying `rc_max_rate`, and review all of these changes as a separate effort later.

### Motivation and Context

Want to update to newer FFmpeg and SVT-AV1. Want to work seamlessly with current deps and newer deps.

### How Has This Been Tested?

Tested builds locally on Windows 11 against FFmpeg 5.1.2 + SVT-AVT 1.3.0. Recording in SVT-AV1 works with all rate control modes, though there may be other existing issues that are unresolved (e.g., how constant CBR really is).

### Types of changes

### Checklist:

### DIFF
--- a/plugins/obs-ffmpeg/obs-ffmpeg-av1.c
+++ b/plugins/obs-ffmpeg/obs-ffmpeg-av1.c
@@ -66,6 +66,7 @@ static bool av1_update(struct av1_encoder *enc, obs_data_t *settings)
 	int cqp = (int)obs_data_get_int(settings, "cqp");
 	int keyint_sec = (int)obs_data_get_int(settings, "keyint_sec");
 	int preset = (int)obs_data_get_int(settings, "preset");
+	AVDictionary *svtav1_opts = NULL;
 
 	video_t *video = obs_encoder_video(enc->ffve.encoder);
 	const struct video_output_info *voi = video_output_get_info(video);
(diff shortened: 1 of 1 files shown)


flaeri:
Did you by check happen to check if the CQP value gets respected with this change and the newer deps?

If yes, then half of https://github.com/obsproject/obs-studio/issues/7784 can be considered solved by this PR.

Can confirm this PR runs with the old deps, and produces files av1 files.

RytoEX:
> Did you by check happen to check if the CQP value gets respected with this change and the newer deps?
> 
> If yes, then half of #7784 can be considered solved by this PR.
> 
> Can confirm this PR runs with the old deps, and produces files av1 files.

No, I did not. This was written before that Issue was raised. If you'd like to investigate that question, go for it. I had planned to look into it at some point in the future if no one else had.

PR #7804, opened 2022-11-18
UI: Remove Qt taskbar overlay

gxalpha:
### Description

Removes the Qt implementation of the taskbar overlay on Windows.
This was removed with Qt 6, which is why we now have a native integration for that. There's no reason for this native integration to not be used with Qt 5 (except that we're not using Qt 5 anymore anyways).

### Motivation and Context

+0 -35

### How Has This Been Tested?

Millions of users used this "new" codepath with OBS 28 which I consider to be enough of a test.
That being said, this PR itself hasn't been tested, but I consider it to be very low risk (and even in the tiny chance that it does somehow break, it would only break Qt 5 which we're not using anymore on Windows anyways).

### Types of changes

### Checklist:

### DIFF
--- a/UI/platform-windows.cpp
+++ b/UI/platform-windows.cpp
@@ -37,11 +37,6 @@
 #include <util/windows/HRError.hpp>
 #include <util/windows/ComPtr.hpp>
 
-#if QT_VERSION < QT_VERSION_CHECK(6, 0, 0)
-#include <QWinTaskbarButton>
-#include <QMainWindow>
-#endif
-
 using namespace std;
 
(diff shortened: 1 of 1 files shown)


RytoEX:
As far as I understand, the native implementation should work whether or not a user happens to be compiling with Qt5 or on an older, unsupported version of Windows, so removing this shouldn't affect bisecting or troubleshooting in any meaningful way (I hope).

WizardCM:
This might also allow us to remove QtWinExtras in CMake.

PR #7805, opened 2022-11-18
UI: Add Apple Hardware Encoder to AutoConfig

gxalpha:
### Description

Adds the M1/Pro/Max/Ultra/M2 Hardware encoder to AutoConfig on macOS 13 or newer.

### Motivation and Context

We added this to simple mode with OBS 28, but not to AutoConfig.
This meant that when running AutoConfig, people would still only get recommended x264 even if the hardware encoder was available.

### How Has This Been Tested?

Ran AutoConfig, both for recording and streaming.
Optimizing for streaming, with "Prefer hardware encoder" checked the encoder got selected and applied correctly.
Optimizing for recording, the hardware encoder got selected and applied correctly.
When the encoder was set to unavailable, x264 got set.

### Types of changes

### Checklist:

### DIFF
--- a/UI/window-basic-auto-config-test.cpp
+++ b/UI/window-basic-auto-config-test.cpp
@@ -894,6 +894,8 @@ void AutoConfigTestPage::TestStreamEncoderThread()
 			wiz->streamingEncoder = AutoConfig::Encoder::NVENC;
 		else if (wiz->qsvAvailable)
 			wiz->streamingEncoder = AutoConfig::Encoder::QSV;
+		else if (wiz->appleAvailable)
+			wiz->streamingEncoder = AutoConfig::Encoder::Apple;
 		else
 			wiz->streamingEncoder = AutoConfig::Encoder::AMD;
 	} else {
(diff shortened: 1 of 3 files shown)


UI/window-basic-auto-config.cpp
```diff
@@ -1000,6 +1000,17 @@ void AutoConfig::TestHardwareEncoding()
 			hardwareEncodingAvailable = qsvAvailable = true;
 		else if (strcmp(id, "h264_texture_amf") == 0)
 			hardwareEncodingAvailable = vceAvailable = true;
+#ifdef __APPLE__
+		else if (strcmp(id,
+				"com.apple.videotoolbox.videoencoder.ave.avc")
```
derrod, line 1005: I think you're missing a `== 0` here

gxalpha: You're definitely right. I'm honestly a bit baffled why it still worked.

gxalpha: fixed

derrod (approved):

UI/window-basic-auto-config-test.cpp
```diff
@@ -894,6 +894,8 @@ void AutoConfigTestPage::TestStreamEncoderThread()
 			wiz->streamingEncoder = AutoConfig::Encoder::NVENC;
 		else if (wiz->qsvAvailable)
 			wiz->streamingEncoder = AutoConfig::Encoder::QSV;
+		else if (wiz->appleAvailable)
+			wiz->streamingEncoder = AutoConfig::Encoder::Apple;
```
RytoEX, line 898: Is this a different order from the enum because the diff is smaller?

gxalpha: yes

RytoEX: Works for me.

PR #7808, opened 2022-11-19
win-dshow: Recognise higher FPS values from devices

WizardCM:
### Description

This updates the whitelist of allowed frame rates to provide additional, modern frame rates supported by capture cards.

Device Support: 

- UCEC GAM LIVE ULTRA (HZ20GAU01): 100, 120
- ezcap GameLink RAW (ezcap321): 120
- DigitNow U601: 120
- ASUS TUF CU4K30: 120, 144
- AVerMedia Live Gamer Ultra (GC553): 100, 120
- Magewell Pro Capture Quad HDMI: 100, 120, 144
- Elgato 4K60 Pro Mk2: 100, 120, 144, 240
- ACASIS AC-VS2583: 100, 120, 144, 240
- Magewell USB Capture HDMI 4K Plus: 100, 120, 144, 240
- Magewell Pro Capture Dual HDMI 4K Plus LT: 100, 120, 144, 240, 360

List provided by @EposVox.

More frame rates could be added in future, but I believe this is a good starting point.

### Motivation and Context

The list of supported frame rates hasn't been updated since win-dshow's inception in 2014. Higher FPS capture cards are becoming more common.

Technically users could use these before by setting OBS to one of these FPS values and then using the "Highest FPS" option, but that was a terrible user experience.

### How Has This Been Tested?

It has not. Should be tested this weekend.

### Types of changes

### Checklist:


derrod:
Thismlist should probably include 199.88 as that is what broadcast standards and thus modern consoles use.

WizardCM:
> This list should probably include 199.88 as that is what broadcast standards and thus modern consoles use.

I agree, but wasn't exactly sure if we wanted to include both 199.88 and 239.76. I unfortunately have no hardware to test with and therefore couldn't verify on my own. If I do include 199, how should I label it?

derrod:
Afaik there's no broadcast standard for 240 Hz, so just 119.88. Not sure how to label it, maybe stick to the existing pattern wnd name it "119.88 NTSC" (while not directly related to NTSC, this shit is still a result of it)?

WizardCM:
Added 119.88 NTSC at request. Ready for review.

derrod (approved):
LGTM, don't really see any notable omissions. Although I'm sure *someone* will complain about 165 or 175 Hz not being listed.

PR #7828, opened 2022-11-22
cmake: Improve Windows SDK detection handling

computerquip:
We rely on CMAKE_VS_WINDOWS_TARGET_PLATFORM_VERSION which isn't defined when using the Ninja generator.

Changes in a list:
* Some of the messages used the terminology "Windows API" instead of "Windows SDK". I've changed those to "Windows SDK".
* There is now a Windows 11 SDK so specifying Windows 10 specifically in the error message I think is a touch misleading. Simply referring to it as "Windows SDK" is more generic but still concise in what we're referring to. I've removed the "10" as a result.
* If CMAKE_VS_WINDOWS_TARGET_PLATFORM_VERSION wasn't defined, it would give an odd message saying Windows SDK wasn't found even if it's available and being used in the toolchain. Now it uses a fallback that uses an environment variable that's required to be present in this scenario. If that doesn't work, it assumes the SDK isn't present.
* If no SDK is found at all, it now errors out after logging a message as such.
* Since Windows SDK version fits within CMake's idea of a version, I've changed the comparison logic to use CMake's built-in version comparison.

There's another problem here with the FATAL_ERROR message when the found Windows SDK version isn't high enough. It only displays the first line because obs_status appears to only allow two arguments but three are provided. No fix for that is provided here.

The environment variable I use here is WindowsSDKLibVersion which the only documentation for is in winsdk.bat in the VS developer prompt tools. I learned about this trick from the apitrace folks.

### Motivation and Context
Allows building with Ninja while using cl.exe (MSVC) while improving the resulting errors in a problematic environment.

### How Has This Been Tested?
Built with both Ninja and Visual Studio 2022 generators using cl version 19.34.31933

### Types of changes

### Checklist:

### DIFF
--- a/cmake/Modules/CompilerConfig.cmake
+++ b/cmake/Modules/CompilerConfig.cmake
@@ -21,33 +21,26 @@ if(OS_WINDOWS AND MSVC)
     set(THREADS_HAVE_PTHREAD_ARG OFF)
   endif()
 
-  # Check for Win SDK version 10.0.20348 or above
-  obs_status(
-    STATUS "Windows API version is ${CMAKE_VS_WINDOWS_TARGET_PLATFORM_VERSION}")
-  string(REPLACE "." ";" WINAPI_VER
-                 "${CMAKE_VS_WINDOWS_TARGET_PLATFORM_VERSION}")
-
-  list(GET WINAPI_VER 0 WINAPI_VER_MAJOR)
-  list(GET WINAPI_VER 1 WINAPI_VER_MINOR)
-  list(GET WINAPI_VER 2 WINAPI_VER_BUILD)
-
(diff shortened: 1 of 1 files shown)


PatTheMav:
While the code itself seems correct, I don’t know if we will merge this without comprehensive support for building the Windows version with Ninja.

Right now, the official stance is that Ninja is not supported on Windows and I would not want to merge bits and pieces of changes to allow it to work (except if this change is literally the only thing required to make it work).

I’d rather see a ‘Enable Windows builds with Ninja’ PR that adds comprehensive support and some commitment to maintain it.

computerquip:
Push this off to add more Ninja-specific support later on.

Issue #7827, opened 2022-11-22
AMF: HEVC CQP results in very poor quality


flaeri:
### Operating System Info

Windows 11

### OBS Studio Version

Git

### OBS Studio Log URL

https://gist.githubusercontent.com/flaeri/2c6d9db648fdb5dd318008ea015e6da8/raw/16a8348781e96adf53ff6b80d3225305d4d165da/gistfile1.txt

### Expected Behavior

When recording via AMF HEVC, using the CQP rate control at a low value (QP 12) I expect to get nice and clean looking footage.

### Current Behavior

Visual quality looks very poor due to very visible macrosblocks and other strange encoding artefacts.

### Steps to Reproduce

1. Select the AMD HW H.265 (HEVC) encoder
2. Select the CQP rate control
3. Pick CQ Level: 12
4. Record and inspect footage
...

### Anything else we should know?

The issue is more visible at lower resolutions like 720p, but exists on higher resolutions as well. Happens on latest master (as of 2022-11-22), as well as current release (28.1.2), as well as in AMD's own TranscodeHW (if enabling VBAQ).

I believe this is down to VBAQ being enabled by default for all rate controls. This seems to be fine for h264, but it is not for HEVC. See below:

Stock:
https://user-images.githubusercontent.com/50419942/203286588-9d04f19a-8b69-46b1-9600-9eba8a2e54b1.mp4
VBAQ off:
https://user-images.githubusercontent.com/50419942/203286598-3a7b0b40-2db5-40b7-b714-7424b66282cd.mp4

These files were recorded with identical settings, except for changing VBAQ using `HevcEnableVBAQ=false`.

I plan to PR a fix for this by checking for CQP rate control, and disabling VBAQ for that RC.

PR #7829, opened 2022-11-22
obs-ffmpeg: Disable VBAQ for HEVC CQP rate control

flaeri:
### Description
Check rate control, if cqp (0), dont use vbaq.

### Motivation and Context
Fixes https://github.com/obsproject/obs-studio/issues/7827

Want CQP to work properly for HEVC AMF.

### How Has This Been Tested?
On win 11 22H2. Checked all rate controls using breakpoints, making sure only CQP disables VBAQ.

Recorded videos and made sure output was ok (fixed).

### Types of changes

### Checklist:

### DIFF
--- a/plugins/obs-ffmpeg/texture-amf.cpp
+++ b/plugins/obs-ffmpeg/texture-amf.cpp
@@ -1580,7 +1580,8 @@ static bool amf_hevc_init(void *data, obs_data_t *settings)
 	int rc = get_hevc_rate_control(rc_str);
 
 	set_hevc_property(enc, RATE_CONTROL_METHOD, rc);
-	set_hevc_property(enc, ENABLE_VBAQ, true);
+	if (rc != AMF_VIDEO_ENCODER_HEVC_RATE_CONTROL_METHOD_CONSTANT_QP)
+		set_hevc_property(enc, ENABLE_VBAQ, true);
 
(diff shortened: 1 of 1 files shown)


plugins/obs-ffmpeg/texture-amf.cpp
```diff
@@ -1580,7 +1580,9 @@ static bool amf_hevc_init(void *data, obs_data_t *settings)
 	int rc = get_hevc_rate_control(rc_str);
 
 	set_hevc_property(enc, RATE_CONTROL_METHOD, rc);
-	set_hevc_property(enc, ENABLE_VBAQ, true);
+	if (rc != 0) {
```
WizardCM, line 1583: Should this not check for `AMF_VIDEO_ENCODER_HEVC_RATE_CONTROL_METHOD_CONSTANT_QP` directly? (Damn that's a long key)

derrod: `AMF_VIDEO_ENCODER_HEVC_RATE_CONTROL_METHOD_CONSTANT_QP` will just be replaced with the literal value `0`by the compiler, there is no performance difference. But it's more "correct" and should the enum ever change (god forbid) it will continue to work.

flaeri: Alright, understood :)

Will try to figure out how to use that approach.

PR #7830, opened 2022-11-22
CI: Create Sparkle appcast and deltas on tag

derrod:
### Description

Adds CI job to download previous builds, generate deltas, and create a new appcast when a notarised build is created.

Based on changes in #7723 (new URL, channels, combined arm/x86 feed).

### Motivation and Context

Stop Jim from editing XML manually. Get Mac users some deltas.

### How Has This Been Tested?

Has not. I do not have the necessary credentials and haven't done a mock setup yet.

### Types of changes

### Checklist:

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -444,6 +444,9 @@ jobs:
     env:
       HAVE_CODESIGN_IDENTITY: ${{ secrets.MACOS_SIGNING_IDENTITY != '' && secrets.MACOS_SIGNING_CERT != '' }}
       BUILD_FOR_DISTRIBUTION: 'ON'
+      HAVE_SPARKLE_KEY: ${{ secrets.SPARKLE_PRIVATE_KEY != '' }}
+    outputs:
+      run_sparkle: ${{ steps.sparkle_check.outputs.run_sparkle }}
     if: ${{ startsWith(github.ref, 'refs/tags/') && github.event_name != 'pull_request' }}
     strategy:
       matrix:
@@ -460,6 +463,12 @@ jobs:
         id: setup
         run: |
           echo "commitHash=$(git rev-parse --short=9 HEAD)" >> $GITHUB_OUTPUT
+
+      - name: 'Determine if Sparkle should run'
+        if: env.HAVE_CODESIGN_IDENTITY == 'true'
+        id: sparkle_check
+        run: |
+          echo 'run_sparkle=${{ env.HAVE_SPARKLE_KEY }}' >> $GITHUB_OUTPUT
 
       - name: 'Download artifact'
         if: env.HAVE_CODESIGN_IDENTITY == 'true'
@@ -492,3 +501,109 @@ jobs:
         with:
           name: 'obs-studio-macos-${{ matrix.arch }}-notarized'
           path: '${{ github.workspace }}/${{ env.FILE_NAME }}'
+
+  macos_sparkle:
+    name: '04 - macOS Sparkle Updates'
+    runs-on: [macos-12]
+    needs: [macos_release]
+    if: fromJSON(needs.macos_release.outputs.run_sparkle)
+    strategy:
+      matrix:
+        arch: ['x86_64', 'arm64']
+    env:
+      SPARKLE_VERSION: '2.3.2'
+      SPARKLE_HASH: '2b3fe6918ca20a83729aad34f8f693a678b714a17d33b5f13ca2d25edfa7eed3'
+    defaults:
+      run:
+        shell: bash
+    steps:
+      - name: 'Checkout'
+        uses: actions/checkout@v3
+        with:
+          path: 'repo'
+          ref: ${{ github.ref }}
+
+      - name: 'Download artifact'
+        uses: actions/download-artifact@v3
+        with:
+          name: 'obs-studio-macos-${{ matrix.arch }}-notarized'
+          path: 'artifacts'
+
+      - name: 'Install Python requirements'
+        run: pip3 install requests xmltodict
+
+      - name: 'Install Brew requirements'
+        run: brew install coreutils pandoc
+
+      - name: 'Setup Sparkle'
+        run: |
+          curl -L "https://github.com/sparkle-project/Sparkle/releases/download/${{ env.SPARKLE_VERSION }}/Sparkle-${{ env.SPARKLE_VERSION }}.tar.xz" -o Sparkle.tar.xz
+
+          if [[ '${{ env.SPARKLE_HASH }}' != "$(sha256sum Sparkle.tar.xz | cut -d " " -f 1)" ]]; then
+              echo "Sparkle download hash does not match!"
+              exit 1
+          fi
+
+          mkdir sparkle && cd sparkle
+          tar -xf ../Sparkle.tar.xz
+
+      - name: 'Setup folder structure'
+        run: |
+          mkdir builds
+          mkdir -p output/appcasts/stable
+          mkdir -p output/sparkle_deltas/${{ matrix.arch }}
+
+      - name: 'Determine branch and tag'
+        id: branch
+        run: |
+          pushd repo
+          
+          GIT_TAG="$(git describe --tags --abbrev=0)"
+          if [[ ${GIT_TAG} == *'beta'* || ${GIT_TAG} == *'rc'* ]]; then
+            echo "branch=beta" >> $GITHUB_OUTPUT
+            echo "deltas=0" >> $GITHUB_OUTPUT
+          else
+            echo "branch=stable" >> $GITHUB_OUTPUT
+            echo "deltas=1" >> $GITHUB_OUTPUT
+          fi
+          # Write tag description to file
+          git tag -l --format='%(contents)' ${GIT_TAG} >> ../notes.rst
+
+      - name: 'Download existing Appcast and builds'
+        run: python3 repo/CI/macos/appcast_download.py --branch "${{ steps.branch.outputs.branch }}" --max-old-versions ${{ steps.branch.outputs.deltas }}
+
+      - name: 'Prepare release notes'
+        run: |
+          # Insert underline at line 2 to turn first line into heading
+          sed -i '' '2i\'$'\n''###################################################' notes.rst
+          pandoc -f rst -t html notes.rst -o output/appcasts/notes_${{ steps.branch.outputs.branch }}.html
+
+      - name: 'Setup Sparkle key'
+        run: echo -n "${{ secrets.SPARKLE_PRIVATE_KEY }}" >> eddsa_private.key
+
+      - name: 'Generate Appcast'
+        run: |
+          mv artifacts/*.dmg builds/
+          ./sparkle/bin/generate_appcast \
+              --verbose \
+              --ed-key-file ./eddsa_private.key \
+              --download-url-prefix "https://cdn-fastly.obsproject.com/downloads/" \
+              --full-release-notes-url "https://obsproject.com/osx_update/notes_${{ steps.branch.outputs.branch }}.html" \
+              --maximum-versions 0 \
+              --maximum-deltas ${{ steps.branch.outputs.deltas }} \
+              --channel "${{ steps.branch.outputs.branch }}" builds/
+          # Move deltas, if any
+          if compgen -G "builds/*.delta" > /dev/null; then
+            mv builds/*.delta output/sparkle_deltas/${{ matrix.arch }}
+          fi
+          # Move appcasts
+          mv builds/*.xml output/appcasts/
+
+      - name: 'Create 1.x Appcast'
+        run: python3 repo/CI/macos/appcast_convert.py
+
+      - name: 'Upload Appcast and Deltas'
+        uses: actions/upload-artifact@v3
+        with:
+          name: 'macos-sparkle-updates'
+          path: '${{ github.workspace }}/output'
--- a/CI/macos/appcast_convert.py
+++ b/CI/macos/appcast_convert.py
@@ -1,0 +1,96 @@
+import os
+from copy import deepcopy
+
+import xmltodict
+
+
+DELTA_BASE_URL = "https://cdn-fastly.obsproject.com/downloads/sparkle_deltas"
+
+
+def convert_appcast(filename):
+    print("Converting", filename)
+    in_path = os.path.join("output/appcasts", filename)
+    out_path = os.path.join("output/appcasts/stable", filename.replace("_v2", ""))
+    with open(in_path, "rb") as f:
+        xml_data = f.read()
+    if not xml_data:
+        return
+
+    appcast = xmltodict.parse(xml_data, force_list=("item",))
+    out_appcast = deepcopy(appcast)
+
+    # Remove anything but stable channel items.
+    new_list = []
+    for _item in appcast["rss"]["channel"]["item"]:
+        item = deepcopy(_item)
+        branch = item.pop("sparkle:channel", "stable")
+        if branch != "stable":
+            continue
+        # Remove delta information (incompatible with Sparkle 1.x)
+        item.pop("sparkle:deltas", None)
+        new_list.append(item)
+
+    out_appcast["rss"]["channel"]["item"] = new_list
+
+    with open(out_path, "wb") as f:
+        xmltodict.unparse(out_appcast, output=f, pretty=True)
+
+    # Also create legacy appcast from x86 version.
+    if "x86" in filename:
+        out_path = os.path.join("output/appcasts/stable", "updates.xml")
+        with open(out_path, "wb") as f:
+            xmltodict.unparse(out_appcast, output=f, pretty=True)
+
+
+def adjust_appcast(filename):
+    print("Adjusting", filename)
+    file_path = os.path.join("output/appcasts", filename)
+    with open(file_path, "rb") as f:
+        xml_data = f.read()
+    if not xml_data:
+        return
+
+    arch = "arm64" if "arm64" in filename else "x86_64"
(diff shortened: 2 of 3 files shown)


derrod:
Updated to work with separate arm64/x86 appcasts, and also includes a script to generate backwards-compatible appcasts for the old feeds (only containing the latest "stable" item).

derrod:
I have trialed this locally now and made some adjustments/fixes bsaed on that.

The following output will now be produced by this script:
* `deltas/`
   + `x86_64`
      - `OBS [...] .delta` - x86 delta files
   + `arm64`
      - `OBS [...] .delta` - arm delta files
 * `appcasts/`
   + `updates_arm64_v2.xml` new Sparkle v2 appcast with channels
   + `updates_arm64.xml` legacy Sparkle v1 appcast
   + `updates_x86_64_v2.xml` new Sparkle v2 appcast with channels
   + `updates_x86_64.xml` legacy Sparkle v1 appcast

Together with the updates to #7723 this now assumes the following updated layout on the OBS server:

* `obsproject.com/osx_update/stable/updates_(x86_64|arm64).xml` - legacy v1 appcasts *(unchanged)*
* `obsproject.com/osx_update/updates_(x86_64|arm64)_v2.xml` - new v2 appcasts (with channels)
* `obsproject.com/osx_update/notes_<channel>.html` - Channel-specific changelogs

And for downloads:
* `cdn-fastly.obsproject.com/downloads/obs [...].dmg` - DMG downloads *(unchanged)*
* `cdn-fastly.obsproject.com/downloads/sparkle_deltas/(x86_64|arm64)/OBS [...].delta` - architecture separated delta files

derrod:
Updated with the following:
- Workflow is now skipped if Sparkle key not available (annoyingly, you cannot check for secrets in job-level `if`s...)
- Added fallback url to appcast downloader so it'll update the existing (old) appcasts until the new appcasts exist
- Do not generate deltas for beta/rc channel
- Fixed some typos

This could be merged before #7723 so we can start creating appcasts on CI going forward. The ones generated for Sparkle v2 can simply be discarded rather than uploaded to the server.

Test run on my repo (with modification to skip notarisation step): https://github.com/derrod/obs-studio/actions/runs/3557922082

derrod:
Took a while to figure things about, but it seems to work properly now: https://github.com/derrod/obs-studio/actions/runs/3563985210
(Best guess is that using the keychain meant that it would lock up when trying to access it)

A few things to note:
- Sparkle will not generate deltas between purely DSA and purely EdDSA builds
- Sparkle will not sign updates with EdDSA if they do not have a public key specified in their Info.plist

So for a smooth transition we should add the EdDSA key to OBS now (i.e. with 29) so that the v1 to v2 transition is a little smoother. In theory we should be able to use the appcasts created by this regardless since the developer identity matches between 28.1.2 and 29+ but to be safe at least that transitionary update should probably still be signed with DSA as well (manually).

derrod:
Reminder to self:
- Clean up scripts (some comments are out of date)
- Update to Sparkle 2.3.2 (`generate_appcast` was updated)
- Determine if we are okay with the Sparkle private key being on CI for the purposes of createing new appcasts

Might require some more feedback from @PatTheMav.

derrod:
This PR has now been updated again with the following changes:

- Updated Sparkle to 2.3.2
- Automatically create release notes HTML file by running the tag description through pandoc
- Updated python scripts

CI/macos/appcast_convert.py
```diff
@@ -0,0 +1,83 @@
+import os
+
+import xmltodict
```
RytoEX, line 3: Is there some Python standard/practice for spacing these apart?

RytoEX (changes requested):

CI/macos/appcast_convert.py
derrod: Yes, see PEP8 section for imports: https://peps.python.org/pep-0008/#imports

> Imports should be grouped in the following order:
> 1. Standard library imports.
> 2. Related third party imports.
> 3. Local application/library specific imports.

.github/workflows/main.yml
```diff
@@ -499,3 +499,127 @@ jobs:
...
+  macos_sparkle:
+    name: '04 - macOS Sparkle Updates'
+    runs-on: [macos-12]
+    needs: [macos_release]
+    if: ${{ startsWith(github.ref, 'refs/tags/') && github.event_name != 'pull_request' }}
+    strategy:
+      matrix:
+        arch: ['x86_64', 'arm64']
+    env:
+      HAVE_SPARKLE_KEY: ${{ secrets.SPARKLE_PRIVATE_KEY != '' }}
```
PatTheMav, line 512: I know it's not possible to check for a secret as a condition to run a job, does the same apply for a `needs` output? Because if that's possible we could check for the secret in one of the required steps and just skip the entire job.

.github/workflows/main.yml
```diff
@@ -499,3 +499,127 @@ jobs:
...
+        run: pip3 install requests xmltodict
+
+      - name: 'Install Brew requirements'
+        if: env.HAVE_SPARKLE_KEY == 'true'
+        run: brew install coreutils pandoc
+
+      - name: 'Setup Sparkle'
+        if: env.HAVE_SPARKLE_KEY == 'true'
+        run: |
+          wget "https://github.com/sparkle-project/Sparkle/releases/download/${{ env.SPARKLE_VERSION }}/Sparkle-${{ env.SPARKLE_VERSION }}.tar.xz" -O Sparkle.tar.xz
```
PatTheMav, line 544: Default to curl instead of wget. curl is guaranteed to exist on macOS and is our download CLI tool of choice.

.github/workflows/main.yml
```diff
@@ -499,3 +499,127 @@ jobs:
...
+        run: |
+          mkdir builds
+          mkdir -p output/appcasts
+          mkdir -p output/deltas/${{ matrix.arch }}
+
+      - name: 'Determine branch and tag'
+        if: env.HAVE_SPARKLE_KEY == 'true'
+        id: branch
+        run: |
+          cd repo/
```
PatTheMav, line 565: Make it a habit to use `pushd`/`popd` pairs to switch directory scopes instead of `cd` on CI (it won't make a difference here, but I'd like to remove any occurrence of `cd` calls moving forward).

.github/workflows/main.yml
```diff
@@ -499,3 +499,127 @@ jobs:
...
+          pandoc -f rst -t html notes.rst -o output/appcasts/notes_${{ steps.branch.outputs.branch }}.html
+
+      - name: 'Setup Sparkle key'
+        if: env.HAVE_SPARKLE_KEY == 'true'
+        run: echo -n "${{ secrets.SPARKLE_PRIVATE_KEY }}" >> eddsa_private.key
+
+      - name: 'Generate Appcast'
+        if: env.HAVE_SPARKLE_KEY == 'true'
+        run: |
+          mv artifacts/*.dmg builds/
```
PatTheMav, line 600: Ensure that you only handle a single `dmg` file - use a GLOBbing expression that yields only what you need (e.g. store all files sorted by date into an array, then use the first element of the resulting array).

CI/macos/appcast_convert.py
```diff
@@ -0,0 +1,83 @@
...
+
+
+DELTA_BASE_URL = 'https://cdn-fastly.obsproject.com/downloads/sparkle_deltas'
+
+
+def convert_appcast(filename):
```
PatTheMav, line 9: Run this file through `black` without any specific options.

CI/macos/appcast_convert.py
```diff
@@ -0,0 +1,83 @@
...
+    for item in appcast['rss']['channel']['item']:
+        branch = item.pop('sparkle:channel', 'stable')
+        if branch != 'stable':
+            continue
+        # remove delta information (incompatible with Sparkle 1.x)
+        item.pop('sparkle:deltas', None)
+        new_list.append(item)
+        break
+
+    appcast['rss']['channel']['item'] = new_list
```
PatTheMav, line 33: I'd prefer if this function were "pure" and not modify the appcast data structure at all. Operate on a copy, then pass the copy to `xmltodict` or use functional patterns to begin with.

CI/macos/appcast_convert.py
```diff
@@ -0,0 +1,83 @@
...
+    appcast['rss']['channel']['link'] = 'https://obsproject.com/'
+
+    new_list = []
+    for item in appcast['rss']['channel']['item']:
+        # Fix changelog URL
+        # Sparkle doesn't allow us to specify the URL for a specific update,
+        # so we set the full release notes link instead and then rewrite the
+        # appcast. Yay.
+        if release_notes_link := item.pop('sparkle:fullReleaseNotesLink', None):
+            item['sparkle:releaseNotesLink'] = release_notes_link
```
PatTheMav, line 59: Same as above - don't modify existing data structures, convert them into a copy.

CI/macos/appcast_download.py
```diff
@@ -0,0 +1,55 @@
...
+
+import requests
+import xmltodict
+
+
+def download_build(url):
+    print('Downloading', url)
+    filename = url.rpartition('/')[2]
+    r = requests.get(url)
+    if r.status_code == 200:
```
PatTheMav, line 11: Should handle any status except `200 OK`, otherwise the script silently continues, leading to undefined state.

CI/macos/appcast_download.py
```diff
@@ -0,0 +1,55 @@
...
+    if r.status_code == 404:
+        r = requests.get(fallback)
+        if r.status_code == 404:
+            return
+
+    filename = url.rpartition('/')[2]
+    with open(f'builds/{filename}', 'wb') as f:
+        f.write(r.content)
+
+    appcast = xmltodict.parse(r.content, force_list=('item',))
```
PatTheMav, line 27: Code continues unchecked even if target file wasn't written, resulting in undefined state.

.github/workflows/main.yml
derrod: In this case we want all downloaded DMG files to be in the builds directory as they are used for delta creation (currently this is limited to 1 previous build).

.github/workflows/main.yml
derrod: I think reading outputs *should* work in conditions based on https://docs.github.com/en/actions/learn-github-actions/contexts#context-availability

CI/macos/appcast_convert.py
derrod: Black actually does nothing to this file except changing the quote from `'` to `"` (although my default is to run it with that disabled).

CI/macos/appcast_download.py
derrod: I would expect an exception to be thrown if writing the file fails for whatever reason, which doesn't have a pretty error here, but would abort the script execution and set a non-zero exit code.

.github/workflows/main.yml
derrod: I now added this check to `macos_release` in a step that sets an output determining whether or not Sparkle should run. Tested this in my stripped-down test setup in my fork and validated that it works.

CI/macos/appcast_convert.py
derrod: Ran it through black with string normalisation enabled, as I said the only thing changed was the quotes.

RytoEX (changes requested):
Does this handle the legacy appcast URL (https://obsproject.com/osx_update/stable/updates.xml), or do we need to do some manual work for that for a transition period?

.github/workflows/main.yml
```diff
@@ -492,3 +501,113 @@ jobs:
         with:
           name: 'obs-studio-macos-${{ matrix.arch }}-notarized'
           path: '${{ github.workspace }}/${{ env.FILE_NAME }}'
+
+  macos_sparkle:
+    name: '04 - macOS Sparkle Updates'
+    runs-on: [macos-12]
+    needs: [macos_release]
+    if: needs.macos_release.outputs.run_sparkle == 'true'
```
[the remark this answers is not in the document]
derrod: That does not work since it's a string, not a boolean, meaning that `"false"` would also evaluate to a truthy value. I think in the future we want to change such evaluations to `toJSON( needs.macos_release.outputs.run_sparkle)` which will evaluate `"true"` to `true`.

CI/macos/appcast_convert.py
```diff
@@ -0,0 +1,87 @@
...
+        xml_data = f.read()
+    if not xml_data:
+        return
+
+    appcast = xmltodict.parse(xml_data, force_list=("item",))
+    out_appcast = deepcopy(appcast)
+
+    # Remove anything but the first stable channel item.
+    # Mostly due to some issues with older Sparkle versions
+    # we still have out there.
```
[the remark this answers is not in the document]
derrod: Yeah in that case keeping all other stable channel items would probably be fine, if a bit pointless.

derrod:
> Does this handle the legacy appcast URL (https://obsproject.com/osx_update/stable/updates.xml), or do we need to do some manual work for that for a transition period?

Yes. They are derived from the v2 appcast by ignoring everything except the first "stable" channel item (since Sparkle 1.x does not support channels) and removing some properties that Sparkle 1.x does not support, such as delta patches created for Sparkle 2.

.github/workflows/main.yml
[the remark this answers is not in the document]
PatTheMav: Yeah using the `toJSON` pattern seems to be the suggested way to handle this, but it's effectively the same.

We could argue "why not use it already if we _add_ code" and I wouldn't have a good answer to that. 😄

CI/macos/appcast_download.py
```diff
@@ -0,0 +1,62 @@
...
+            continue
+
+        if dl == max_old_vers:
+            break
+        download_build(item["enclosure"]["@url"])
+        dl += 1
+
+
+if __name__ == "__main__":
+    appcast_urls = {
```
PatTheMav, line 49: Wouldn't it be better to use `plutil` to read the _actual_ appcast URLs from the app bundles that are downloaded in a prior step?

CI/macos/appcast_download.py
```diff
@@ -0,0 +1,62 @@
...
+        "x86_64": (
+            "https://obsproject.com/osx_update/updates_x86_64_v2.xml",
+            "https://obsproject.com/osx_update/stable/updates_x86_64.xml",
+        ),
+        "arm64": (
+            "https://obsproject.com/osx_update/updates_arm64_v2.xml",
+            "https://obsproject.com/osx_update/stable/updates_arm64.xml",
+        ),
+    }
+    target_branch = os.getenv("BRANCH")
```
PatTheMav, line 59: As always, I _really_ don't like relying on environment variables - if these are parameters for the script they should be parsed and handled as such.

CI/macos/appcast_download.py
PatTheMav: We used to do it for notarisation:

https://github.com/obsproject/obs-studio/blob/6bee92d0545a73509c79af7ce5ea083e3d1e540f/CI/macos/03_package_obs.sh#L61

CI/macos/appcast_download.py
derrod: Yeah that could just be a simple bit of argparse, didn't really think much of it at the time since it's only ever meant to run on CI.

PatTheMav: It's not a blocker IMO. I just prefer any code (even little scripts) to be fully-fledged and production-quality.

.github/workflows/main.yml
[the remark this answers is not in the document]
derrod: I can certainly just use `fromJSON` instead when I update this PR tomorrow.

CI/macos/appcast_download.py
derrod: The new version of this script mounts the DMG, reads the plist, and then uses the feed URL fromt here.

For this to correctly include existing releases when it is created for the first time we need to get R1CH or Jim to copy the existing appcasts to the new locations before tagging the first release with this workflow.

derrod:
Updated with the following changes/amendments:

- The workflow now uses `fromJSON( ... )` for the condition
- Output folder structure now matches what is expected on the server
  + This now also includes the legacy `updates.xml` file
- `appcast_download.py`
    + Now uses command line arguments rather than environment varaibles
    + Now mounts the DMG and read the `SUFeedURL` from the plist file 
        * This removes the fallback, but that shouldn't be an issue as long as somebody copies the old appcasts before we tag 30.0

PatTheMav:
> Updated with the following changes/amendments:
> 
> * The workflow now uses `fromJSON( ... )` for the condition
> * Output folder structure now matches what is expected on the server
>   
>   * This now also includes the legacy `updates.xml` file
> * `appcast_download.py`
>   
>   * Now uses command line arguments rather than environment varaibles
>   * Now mounts the DMG and read the `SUFeedURL` from the plist file
>     
>     * This removes the fallback, but that shouldn't be an issue as long as somebody copies the old appcasts before we tag 30.0

Awesome.. I guess the final test will be a proper release, but I guess this is fine to merge now.

derrod:
> Awesome.. I guess the final test will be a proper release [...]

Yep, don't have codesigning credentials to test on my fork, so I just ran it with a stripped-down version of the `main` workflow that downloaded some prebuilt DMGs and generated the appcasts/deltas only. It seemed to work as expected: https://github.com/derrod/obs-studio/actions/runs/4292178311

Fingers crossed. Main reason I could see for a run failing is a 404 on the appcast URLs because we forget to copy the "input" ones over, but in that case we can just do that and rerun the failed job.

RytoEX (approved):
I think there's probably nothing to do but try it out. Will circle back around next week.

PR #7833, opened 2022-11-22
libobs,UI: Add bool list and radio button list to properties

gxalpha:
### Description

Adds `OBS_COMBO_FORMAT_BOOL` to combo formats, making it possible to have bool values in a properties list (dropdown).

Adds `OBS_COMBO_TYPE_RADIO`, which displays a properties list as radio buttons instead of as a combobox. For the API user, input and output behaves the same as `OBS_COMBO_TYPE_LIST`.
Hypothetical case of how this would look (this isn't the actual use case, which will follow in a second separated PR to make reviewing easier.):
<p float="left">
Bildschirm­foto 2022-11-22 um 17 52 45
Bildschirm­foto 2022-11-22 um 17 53 11
</p>

Technically, these changes are unrelated, but one modifies each other, making it simpler to have them combined (cause the radio type also needs bool support if that would be merged first, so I'd have to fix conflicts from my own PR).

### Motivation and Context

Having radio buttons in properties would be beneficial for some use cases, and a year ago at least was something on Warchamp's wishlist.
I then made this branch about half a year ago, then forgot it.

This is preparation for a second PR which I will open shortly.

### How Has This Been Tested?

Tested with the properties of the ft2-text source, as well as the browser source (see screenshot above). Also tested with a properties view in encoder settings.

### Types of changes

### Checklist:

### DIFF
--- a/UI/properties-view.cpp
+++ b/UI/properties-view.cpp
@@ -511,6 +511,9 @@ static void AddComboItem(QComboBox *combo, obs_property_t *prop,
 
 	} else if (format == OBS_COMBO_FORMAT_STRING) {
 		var = QByteArray(obs_property_list_item_string(prop, idx));
+	} else if (format == OBS_COMBO_FORMAT_BOOL) {
+		bool val = obs_property_list_item_bool(prop, idx);
+		var = QVariant::fromValue<bool>(val);
 	}
 
 	combo->addItem(QT_UTF8(name), var);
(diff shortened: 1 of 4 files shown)


RealIndrit:
The idea of radio button list is cool, but I don't really know if they are good for lists with more than 4 buttons, at some point the list of radio buttons is just too much and makes the window look bloated, I personally prefer the drop down on the permissions settings, but still do see the usage of the radio button list in other instances of use cases.

gxalpha:
> I don't really know if they are good for lists with more than 4-5 buttons, at some point the list of radio buttons is just too much and makes the window look bloated. Personally I prefer the look of the drop down list for the permission setting

As mentioned, this isn't the actual use case.

You can find a real use case of this in #7835

RealIndrit:
> > I don't really know if they are good for lists with more than 4-5 buttons, at some point the list of radio buttons is just too much and makes the window look bloated. Personally I prefer the look of the drop down list for the permission setting
> 
> 
> 
> As mentioned, this isn't the actual use case.
> 
> 
> 
> You can find a real use case of this in #7835

Just realized, my bad, completely missed that part 😂.

CodeYan01:
I think these should be added to reference-properties.rst

gxalpha:
Fixed.

PR #7838, opened 2022-11-23
cmake: Forward extra args in obs_status to message

computerquip:
### Description
Forwards extra arguments given to obs_status to message who will then concatenate them together.

### Motivation and Context
So far, I see this only making a difference in exactly one place: https://github.com/obsproject/obs-studio/blob/510f0116a62c12ea6c3812bb88a672ebbebef6a6/cmake/Modules/CompilerConfig.cmake#L48

The `"Please download the most recent Windows 10 SDK in order to compile."` line is missing from the output since it's a third argument.

Alternatively, I could just fix that one place. I have no preference.

### How Has This Been Tested?
I've run CMake in way that would trigger that message so it displays the correct text. It gives the output:
```
CMake Error at cmake/Modules/ObsHelpers.cmake:446 (message):
  OBS: OBS requires Windows 10 SDK version 10.0.20348.0 and above to compile.

  Please download the most recent Windows 10 SDK in order to compile.
```
which appears to be correct.

Other messages haven't changed at all.

### Types of changes

### Checklist:


PatTheMav:
The observation is correct, seems to me like a better fix would remove the `text` parameter and use `ARGN` entirely though.

RytoEX:
> The observation is correct, seems to me like a better fix would remove the `text` parameter and use `ARGN` entirely though.

Circling back around to this item. Let's do this, barring any major issues.

computerquip:
Sorry, this somehow kept slipping by me, I'll clean it up soon™.

PatTheMav:
Probably not required as that call was used in legacy CMake only and we’re gonna move Linux off that soon as well.

RytoEX:
@PatTheMav Do we still want this?

PatTheMav:
> @PatTheMav Do we still want this?

IMO the legacy build path is in "maintenance mode" now, so we accept bugfixes but no new features.

PR #7839, opened 2022-11-23
libobs: Override fps ovi for aux views

aet:
### Description
- libobs: Override fps ovi for aux views

### Motivation and Context

Previous iteration at https://github.com/obsproject/obs-studio/pull/7797 had discussion that since the one graphics thread drives all outputs, aux views should inherit the FPS settings from the main view

### How Has This Been Tested?
- Tested on a custom plugin that aux view do get settings from the main mix when available
- Tested what happens when video output settings are changed in the UI and obs_reset_video() is triggered

### Types of changes

### Checklist:

### DIFF
--- a/libobs/obs.c
+++ b/libobs/obs.c
@@ -589,6 +589,18 @@ static int obs_init_video_mix(struct obs_video_info *ovi,
 
 	make_video_info(&vi, ovi);
 	video->ovi = *ovi;
+
+	/* main view graphics thread drives all frame output,
+	 * so share FPS settings for aux views */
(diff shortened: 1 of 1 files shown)


tt2468:
Circling back to this. I'm working with the view code experimentally. Did you ever consider what might happen if the main view's FPS changes? It seems like this only synchronizes fps_num and fps_den initially, but changes to the main view's fractionals may not be propagating to all other views automatically.

aet:
@tt2468 if the only way to alter them is through obs_reset_video, it should be covered

PR #7841, opened 2022-11-23
CI: Add missing libappstream-glib in Flatpak workflow

tytan652:
### Description

Deps changes on the Fedora flatpak package breaking our workflow for release.

### Motivation and Context

Fix it

### How Has This Been Tested?

I will test it on my fork wit a fake release.

### Types of changes

### Checklist:

### DIFF
--- a/.github/workflows/flatpak.yml
+++ b/.github/workflows/flatpak.yml
@@ -66,6 +66,7 @@ jobs:
         run: |
           git config --global --add safe.directory $GITHUB_WORKSPACE
           echo "commitHash=$(git rev-parse --short=9 HEAD)" >> $GITHUB_OUTPUT
+          dnf install -y -q libappstream-glib
 
       - name: Build Flatpak Manifest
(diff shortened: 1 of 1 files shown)


tytan652:
My test CI passed: https://github.com/tytan652/obs-studio/actions/runs/3530133119/jobs/5921797439

RytoEX:
> > This seems like the wrong place to fix
> 
> If you mean the action images is where it should be fixed, note that the image no longer requires/provides it so this is theoretically not the wrong place.

If `appstream-util` previously ran from within the docker container image, then I'd like to just report this to the container image maintainer and see it missing now is intentional or accidental.

GeorgesStavracas:
It is the Flatpak action that must provide the necessary tooling for publishing Flatpaks. This was the case before, should continue to be the case now. https://github.com/flatpak/flatpak-github-actions/pull/110 landed, and the Docker image should have this tool again.

PR #7842, opened 2022-11-23
Frozen screen capture Mac OS Release/28.1 

amddesigns:
your f--in soft capturing nothing and gives no sign of it, only after I open the vid I see screenshot not a video. Don't need to pull anything, just wanna say f--k you, and this f--in scum bag soft.


gxalpha:
Yup, thanks for insulting the developers! Very kind of you.

PR #7859, opened 2022-11-25
UI: Disable replay buffer checkbox when using custom FFmpeg

### LINKED ISSUE #4948: Replay Buffer Missing in Advanced Mode
### Operating System Info

Windows 10

### Other OS

_No response_

### OBS Studio Version

27.0.1

_No response_

### OBS Studio Log URL

https://obsproject.com/logs/gTndw0cnxWUXq2FT

### OBS Studio Crash Log URL

_No response_

### Expected Behavior

Enable replay buffer

### Current Behavior

Replay Buffer is missing completely when in Advanced Mode
image

### Steps to Reproduce

1. Switch from simple to advanced mode
2. Enable Replay Buffer
3. Nothing happens
...

Replay buffer works in simple mode

gxalpha:
### Description

Disables the "Enable Replay Buffer" checkbox in Advanced output mode if the recording type is set to "Custom Output (FFmpeg)" and adds a label stating why it's disabled.

Bildschirm­foto 2022-11-25 um 10 42 02

### Motivation and Context

Closes #4948

### How Has This Been Tested?

Switched back and forth between normal and custom FFmpeg Output, checked that the checkbox is disabled and the text is visible, and that the checkbox is enabled and the text invisible otherwise.

### Types of changes

### Checklist:

### DIFF
--- a/UI/data/locale/en-US.ini
+++ b/UI/data/locale/en-US.ini
@@ -945,6 +945,7 @@ Basic.Settings.Output.ReplayBuffer.EstimateTooLarge="Warning: Estimated memory u
 Basic.Settings.Output.ReplayBuffer.EstimateUnknown="Cannot estimate memory usage. Please set maximum memory limit."
 Basic.Settings.Output.ReplayBuffer.Prefix="Replay Buffer Filename Prefix"
 Basic.Settings.Output.ReplayBuffer.Suffix="Suffix"
+Basic.Settings.Output.ReplayBuffer.UnavailableCustomFFmpeg="Replay Buffer cannot be used when recording type is set to Custom Output (FFmpeg)."
 Basic.Settings.Output.Simple.SavePath="Recording Path"
 Basic.Settings.Output.Simple.RecordingQuality="Recording Quality"
 Basic.Settings.Output.Simple.RecordingQuality.Stream="Same as stream"
--- a/UI/forms/OBSBasicSettings.ui
+++ b/UI/forms/OBSBasicSettings.ui
@@ -4594,6 +4594,16 @@
                         <number>9</number>
                        </property>
                        <item>
+                        <widget class="QLabel" name="advReplayBufCustomFFmpeg">
+                         <property name="text">
+                          <string>Basic.Settings.Output.ReplayBuffer.UnavailableCustomFFmpeg</string>
+                         </property>
+                         <property name="themeID" stdset="0">
+                          <string>warning</string>
+                         </property>
+                        </widget>
+                       </item>
+                       <item>
                         <widget class="QCheckBox" name="advReplayBuf">
                          <property name="text">
                           <string>Basic.Settings.Output.UseReplayBuffer</string>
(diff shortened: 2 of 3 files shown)


UI/data/locale/en-US.ini
```diff
@@ -946,6 +946,7 @@ Basic.Settings.Output.ReplayBuffer.EstimateTooLarge="Warning: Estimated memory u
 Basic.Settings.Output.ReplayBuffer.EstimateUnknown="Cannot estimate memory usage. Please set maximum memory limit."
 Basic.Settings.Output.ReplayBuffer.Prefix="Replay Buffer Filename Prefix"
 Basic.Settings.Output.ReplayBuffer.Suffix="Suffix"
+Basic.Settings.Output.ReplayBuffer.UnavailableCustomFFmpeg="Replay Buffer is unavailable when using Custom Output (FFmpeg)."
```
WizardCM, line 949: Potentially for clarity. Up to Warchamp in the end.
```suggestion
Basic.Settings.Output.ReplayBuffer.UnavailableCustomFFmpeg="Replay Buffer is unavailable when Recording is set to Custom Output (FFmpeg)."
```

gxalpha: I agree with this

Fenrirthviti:
Is it not possible to use the normal warning label we use elsewhere in the program, at the bottom of the window, in the warning label color?

gxalpha:
If this question is about the UX: Sure, I can make it yellow. As for it being on the bottom, this doesn't make sense to me? Putting it away from the disabled section to the bottom of the window visually disconnects it (created with some quick image manipulation):
203950039-8a724a59-c850-480c-bb21-3fb340bfda40
That being said, I don't have hard feelings about it, if you want me to put it on the bottom I can put it there. I just created it since that's how it felt right to me (and since the original guidance on this was "Tab content disabled with an inline warning").


If this question is about potential code duplication: No, such a "normal" warning label does not generally exists, every label of this type is custom. See #7709 for a potential fix, bearing in mind that that PR still has quite a few questions that need to be resolved.

Fenrirthviti:
Having a consistent UI is valuable. If warnings are always in the same place, as we have been trying to do in the settings panels, then users will know where to find them and what they mean more intuitively. I think consistency is a good tradeoff to what might "feel" better for this specific case.

However, as always, I defer to @Warchamp7 if there is an exception to be made here.

Warchamp7 (changes requested):
Position of this information is a bit nebulous currently, yeah. I think the fixed position tackled by #7709 should be used for things that are actual warnings the user should be made aware of. There needs to be a larger pass on classifying things into categories like `ERROR`, `WARNING` and `INFO`

Something like this message I would classify as `INFO`; informational details makes sense closer to the relevant context.

For now, I'm okay with this at the top as it was originally, but I'd like it styled like our warnings in the bolded yellow text.

gxalpha:
Updated the PR to now look like this (includes wording-adjustments as discussed off-thread):

image

Warchamp7 (approved):

PR #7860, opened 2022-11-25
UI: Guard GetMonitorName behind Qt < 6.4

gxalpha:
### Description

d27cda1 made `GetMonitorName` unused on newer Qt versions, so it only needs to be defined for older ones.

### Motivation and Context

Potential code cleanup in the (far) future.

### How Has This Been Tested?

Checked via `git grep` that the only other instance of `GetMonitorName` is in `win-capture`, where the function is defined again (since it can't access UI code anyways).

### Types of changes

### Checklist:

### DIFF
--- a/UI/platform-windows.cpp
+++ b/UI/platform-windows.cpp
@@ -356,6 +356,7 @@ static BOOL CALLBACK GetMonitorCallback(HMONITOR monitor, HDC, LPRECT,
 	return true;
 }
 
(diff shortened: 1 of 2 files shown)


WizardCM:
My original hope was that this capability could eventually become a utility function in libobs, for startup logging, Display Capture in win-capture, **and** projectors. However, it seems all 3 implementations are different enough that it may not be worth it.

Overall I agree with this change.

WizardCM (approved):

PR #7869, opened 2022-11-28
cmake/CI: Add Sparkle EdDSA key

derrod:
### Description

Backports some changes from #7723 to add EdDSA keys to future OBS versions sooner rather than later.

### Motivation and Context

For a cleaner transition to Sparkle 2 and purely EdDSA signed updates it would be helpful to start including our EdDSA key now.

Also makes #7830 usable to generate the appcast going forward without waiting on #7723 (even if we might want to cross-sign 29.0 itself with DSA still).

### How Has This Been Tested?

Builds fine.

### Types of changes

### Checklist:

### DIFF
--- a/cmake/Modules/ObsHelpers_macOS.cmake
+++ b/cmake/Modules/ObsHelpers_macOS.cmake
@@ -354,13 +354,21 @@ function(setup_obs_bundle target)
     set(_CODESIGN_ENTITLEMENTS \"${CMAKE_SOURCE_DIR}/cmake/bundle/macOS\")"
     COMPONENT obs_resources)
 
-  if(ENABLE_SPARKLE_UPDATER)
+  if(SPARKLE_APPCAST_URL AND SPARKLE_PUBLIC_KEY)
     add_custom_command(
       TARGET ${target}
       POST_BUILD
       COMMAND
         /bin/sh -c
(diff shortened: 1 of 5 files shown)


derrod:
Updated to include the now generated public key and also retargeted it to the release branch in case we can to do a 29.0.X with a EdDSA key. Otherwise this PR is superseeded by the now finalised #7723

derrod:
Kepping open in case we want to do a 29.0.3

cmake/Modules/ObsHelpers_macOS.cmake
```diff
@@ -354,13 +354,21 @@ function(setup_obs_bundle target)
     set(_CODESIGN_ENTITLEMENTS \"${CMAKE_SOURCE_DIR}/cmake/bundle/macOS\")"
     COMPONENT obs_resources)
 
-  if(ENABLE_SPARKLE_UPDATER)
+  if(DEFINED SPARKLE_APPCAST_URL AND DEFINED SPARKLE_PUBLIC_KEY)
```
PatTheMav, line 357: Remove the `DEFINED` part, because this will enable Sparkle even if the values are empty strings.

`0`, `OFF`, `FALSE`, ``, and `<NAME>-NOTFOUND` are all treated as `false` in CMake.

derrod: I guess we should fix that in master as well then 😅

PatTheMav (approved):

RytoEX:
Seems fine to me. I'm going to set this aside for a bit. I'd like a firm decision on whether or not we'll release another 29.0.x.

RytoEX:
Closing as we are unlikely to be releasing a 29.0.3. We can revisit this if we change our minds.

PR #7881, opened 2022-12-01
docs: Add versionadded for 29.0.0 functions

CodeYan01:
### Description

Checked https://github.com/obsproject/obs-studio/compare/28.1.2...29.0.0-beta1 to see which were added (although I didn't check each commit, just judged if they seem to be adding functions to the API, so I could have missed a few).
Also fixed this by changing the tabs to spaces
image

Sample of versionadded directive:
image

### Motivation and Context

This makes it easier for developers to know which version supports the functions added. Otherwise plugin/script developers would just get support questions and spend time debugging each version to find out when it was added. This then allows them to put a minimum OBS version. I myself needed the `source_update` event.

### How Has This Been Tested?

Built and viewed locally.

### Types of changes

### Checklist:

### DIFF
--- a/docs/sphinx/reference-frontend-api.rst
+++ b/docs/sphinx/reference-frontend-api.rst
@@ -79,7 +79,7 @@ Structures/Enumerations
    - **OBS_FRONTEND_EVENT_TRANSITION_DURATION_CHANGED**
 
      Triggered when the transition duration has been changed by the
-	 user.
+     user.
 
    - **OBS_FRONTEND_EVENT_TBAR_VALUE_CHANGED**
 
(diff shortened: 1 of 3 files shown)


gxalpha:
Personally I don’t think the patch version is needed since there shouldn’t be cases where new APIs are added in patches, just "29.0" should be enough

CodeYan01:
> Personally I don’t think the patch version is needed since there shouldn’t be cases where new APIs are added in patches, just "29.0" should be enough

I'm fine either way, but existing `deprecated` directives use "27.2.0"
image

So I should I change mine to "29.0"?

CodeYan01:
Added `os_get_sys_total_size`.
Still waiting on a confirmation on whether we should use the patch version in `deprecated` or `versionadded` directives.

PR #7887, opened 2022-12-02
docs: Fix reference count info of obs_frontend_get_streaming_service

gxalpha:
### Description

Changes docs to say that `obs_frontend_get_streaming_service` does not increase the reference.

### Motivation and Context

Current docs claim that `obs_frontend_get_streaming_service` increases the reference count. That's not actually the case.
https://github.com/obsproject/obs-studio/blob/ba27474027f6a116a83a190862224aad5fbaaf3e/UI/window-basic-main.cpp#L4284-L4292
### How Has This Been Tested?

CI build:

image

Also checked that other methods in that file do in fact increase the reference if they claim so.

### Types of changes

### Checklist:

### DIFF
--- a/docs/sphinx/reference-frontend-api.rst
+++ b/docs/sphinx/reference-frontend-api.rst
@@ -655,8 +655,8 @@ Functions
 
 .. function:: obs_service_t *obs_frontend_get_streaming_service(void)
 
-   :return: A new reference to the current streaming service object.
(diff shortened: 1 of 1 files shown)


gxalpha:
As a note, I agree that it probably *should* increase the reference. But that's a breaking change that could break any plugin/script using that function.

PR #7891, opened 2022-12-03
frontend-tools: Fix crash on non X11 windowing systems

WarmUpTill:
### Description
On non X11 windowing systems OBS will crash when the automatic scene switcher is enabled and the current foreground window is changed due to invalid Window parameters being passed to various X11 functions:

```
X Error of failed request:  BadWindow (invalid Window parameter)
  Major opcode of failed request:  20 (X_GetProperty)
  Resource id in failed request:  0x0
  Serial number of failed request:  29
  Current serial number in output stream:  29
```

### Motivation and Context

Add error handling to avoid passing invalid windows to various X11 functions and thus crashes of OBS.

### How Has This Been Tested?

OBS no longer crashes on Ubuntu 22 with the Wayland windowing system when the automatic scene switcher is enabled and the foreground window is changed.

The automatic scene switcher still works as expected on Ubuntu 22 with the X11 windowing system.

### Types of changes

### Checklist:

### DIFF
--- a/UI/frontend-plugins/frontend-tools/auto-scene-switcher-nix.cpp
+++ b/UI/frontend-plugins/frontend-tools/auto-scene-switcher-nix.cpp
@@ -46,8 +46,14 @@ static bool ewmhIsSupported()
 	unsigned long num = 0, bytes = 0;
 	unsigned char *data = NULL;
 	Window ewmh_window = 0;
+	Window root_window = 0;
 
-	int status = XGetWindowProperty(display, DefaultRootWindow(display),
+	root_window = DefaultRootWindow(display);
+	if (!root_window) {
+		return false;
+	}
+
+	int status = XGetWindowProperty(display, root_window,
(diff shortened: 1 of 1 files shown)


kkartaltepe:
It seems like we should disable this whenever the `obs_nix_get_platform` is wayland, is there a specific reason you are not doing that instead?

WarmUpTill:
> It seems like we should disable this whenever the `obs_nix_get_platform` is wayland, is there a specific reason you are not doing that instead?

Seems like _a few_ windows are still supported in the automatic scene switcher even with Wayland as the active windowing system on Ubuntu. (For example, VS code)

Just guessing, but I assume those windows are likely using XWayland instead of Wayland directly.

So the tool might still have some, although rather limited, use.

kkartaltepe:
Sounds fine to me, thanks.

kkartaltepe (approved):

GeorgesStavracas (approved):

PR #7892, opened 2022-12-03
Allow hiding the frame of projectors

kflynn:
### Description
This changes adds a setting to hide the frame of new projectors:
Screenshot 2022-12-03 at 13 38 45

Enabling frame-hiding uses Qt to hide the frame of the projector window (not that a frameless empty projector is a terribly exciting screenshot):
Screenshot 2022-12-03 at 13 39 07

One problem with disabling the frame is that (at least on a Mac) it's very difficult to move a
frameless window. So this change also allows moving a frameless projector by left-dragging
anywhere in the projector window, except that for MultiView projectors, if you click on a scene
selector, that won't start a drag.

### Motivation and Context
I tend to use OBS a lot for demos and workshops shared via Zoom or the like: by defining OBS
scenes that share specific bits of content, I can switch scenes programmatically or using hotkeys,
and  that is _much_ more graceful during a live presentation than messing about with the mouse
(to say nothing of the fact that it gives me much, much more control over what is visible and what
is not).

But the only way that works for my setup and for all conferencing systems is to share OBS'
windowed projector, and having the window frame on the projector always showing up in the
share has been annoying. Hence this change.

### How Has This Been Tested?
This change works fine - meaning I can enable and disable the frame on projectors, and drag
around frameless projector windows - on my M1 Pro MacBook Pro running MacOS 13.0.1 Ventura.
I don't have Windows or Linux lying around to test on, sadly.

### Types of changes

### Checklist:

### DIFF
--- a/plugins/rtmp-services/data/package.json
+++ b/plugins/rtmp-services/data/package.json
@@ -1,11 +1,11 @@
 {
     "$schema": "schema/package-schema.json",
     "url": "https://obsproject.com/obs2_update/rtmp-services/v5",
-    "version": 248,
+    "version": 249,
     "files": [
         {
             "name": "services.json",
-            "version": 248
+            "version": 249
         }
     ]
 }
--- a/plugins/rtmp-services/data/services.json
+++ b/plugins/rtmp-services/data/services.json
@@ -2706,8 +2706,14 @@
             ],
             "recommended": {
                 "keyint": 1,
-                "profile": "high",
-                "bframes": 0
+                "bframes": 0,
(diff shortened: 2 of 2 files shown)


RytoEX:
See also #5818.

kflynn:
🤦‍♂️  I should've searched open PRs for "projector" too, instead of just issues. 😂 

I'm going to refactor this on top of #5818 – it seems like a better base for this, and I like its ability to change the frame after creation. I do still want the drag-anywhere functionality, though.

PR #7897, opened 2022-12-05
UI: Fix Qt AutoUic warning

RytoEX:
### Description

Currently, Qt's User Interface Compiler (UIC) emits a warning about the tabstop assignment for horizontalLayout_sdrPaperWhite, which is a QLayout rather than a QWidget. Remove the tabstop assignment to fix the warning.

### Motivation and Context

Want less warnings while building. Get rid of this warning:
```
Automatic MOC and UIC for target obs
AutoUic: C:\dev\obsproject\obs-studio\UI\forms\OBSBasicSettings.ui: Warning: Tab-stop assignment: 'horizontalLayout_sdrPaperWhite' is not a valid widget.
```

cc @norihiro since this was added in #7486.

### How Has This Been Tested?

On Windows 11, I built OBS before and after this change. Verified that the warning is gone and that OBS builds and runs.

### Types of changes

### Checklist:

### DIFF
--- a/UI/forms/OBSBasicSettings.ui
+++ b/UI/forms/OBSBasicSettings.ui
@@ -7664,7 +7664,6 @@
   <tabstop>colorSpace</tabstop>
   <tabstop>colorRange</tabstop>
   <tabstop>sdrWhiteLevel</tabstop>
-  <tabstop>horizontalLayout_sdrPaperWhite</tabstop>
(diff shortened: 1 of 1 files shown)


norihiro:
I was modifying the tabstop with a text editor and checked the behavior at #7486. Just removing `horizontalLayout_sdrPaperWhite` should be fine as this PR.

Issue #7247, opened 2022-09-01
Cursor for Scene item resize handles don't respect rotation


Richard-Quinlivan:
### Operating System Info

Windows 11

### OBS Studio Version

27.2.4

### OBS Studio Log URL

https://obsproject.com/logs/DzZ4Yps15coeWonj

### Expected Behavior

The resize handles for the webcam source should face outwards from the corner

### Current Behavior

The resize handles for the webcam source are flipped from what they should be 

### Steps to Reproduce

1. add webcam source
2. resize webcam source
3.
...

WizardCM:
Could you post a screenshot example?

Richard-Quinlivan:
Sure, here you go

WizardCM:
Confirmed, resize handles don't respect rotation.

![image](https://user-images.githubusercontent.com/941350/189506014-ec5b3010-7a5c-471d-9bed-1de3e048dbd1.png)

PR #7910, opened 2022-12-09
UI: Fix the resize cursor with respecting the item transformation

BGM99:
### Description
When rotating or flipping a scene item the resize cursor is not updated correctly. With this fix the cursor gets set according to the item rotation, the horizontal scale and the vertical scale. It handles all 90° and 45° rotations. All rotations in between are rounded up or down.

Before:
doubledip

After:

after1
after2
after3
after4

### Motivation and Context

fixes [#7247](https://github.com/obsproject/obs-studio/issues/7247)

### How Has This Been Tested?

I've tested it with nearly every rotation and transformation that is possible on windows with dev version.

What else should I test?

### Types of changes

### Checklist:

### DIFF
--- a/UI/window-basic-preview.cpp
+++ b/UI/window-basic-preview.cpp
@@ -658,8 +658,47 @@ void OBSBasicPreview::UpdateCursor(uint32_t &flags)
 
 	if (!flags && (cursor().shape() != Qt::OpenHandCursor || !scrollMode))
 		unsetCursor();
-	if (cursor().shape() != Qt::ArrowCursor)
+	if ((cursor().shape() != Qt::ArrowCursor) || flags == 0)
 		return;
+
+	if (flags & ITEM_ROT) {
+		setCursor(Qt::OpenHandCursor);
+		return;
+	}
+
+	float rotation = obs_sceneitem_get_rot(stretchItem);
+	vec2 scale;
+	obs_sceneitem_get_scale(stretchItem, &scale);
+
+	if (rotation < 0.0f)
+		rotation = 360.0f + rotation;
+
+	int octant = int(std::round(rotation / 45.0f));
+	bool isCorner = (flags & (flags - 1)) != 0;
+
+	if ((scale.x < 0.0f) && isCorner)
+		flags ^= ITEM_LEFT | ITEM_RIGHT;
+	if ((scale.y < 0.0f) && isCorner)
+		flags ^= ITEM_TOP | ITEM_BOTTOM;
+
+	if (octant % 4 >= 2) {
+		if (isCorner) {
+			flags ^= ITEM_TOP | ITEM_BOTTOM;
+		} else {
+			flags = (flags >> 2) | (flags << 2);
+		}
+	}
+
+	if (octant % 2 == 1) {
+		if (isCorner) {
+			flags &= (flags % 3 == 0) ? ~ITEM_TOP & ~ITEM_BOTTOM
+						  : ~ITEM_LEFT & ~ITEM_RIGHT;
+		} else {
+			flags = (flags % 4 == 0)
+					? flags | flags >> ((flags / 2) - 1)
+					: flags | ((flags >> 2) | (flags << 2));
+		}
+	}
 
 	if ((flags & ITEM_LEFT && flags & ITEM_TOP) ||
 	    (flags & ITEM_RIGHT && flags & ITEM_BOTTOM))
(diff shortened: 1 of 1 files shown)


RytoEX:
If this fixes a specific GitHub Issue, please use the keywords "Fixes #\<Issue Number>". See:

Please provide both before and after screenshots of the behavior.

RytoEX:
How does this handle non-90° rotations? I suppose an equally valid question is how do we currently handle non-90° rotations?

BGM99:
> How does this handle non-90° rotations? I suppose an equally valid question is how do we currently handle non-90° rotations?

There is no need to change anything on the non-90° rotations. When a item is rotated by 180°, then Top Left will be at the former position of Bottom Right. If you look at [window-basic-preview.cpp L.659 and L.660](https://github.com/obsproject/obs-studio/blob/master/UI/window-basic-preview.cpp#L659), you can see that both handles use the same cursor. The same applies to horizontal and vertical cursors. But with a 90° clockwise rotation there is the problem that Top Left will be displayed as Bottom Left, what causes the bug.

RytoEX:
> > How does this handle non-90° rotations? I suppose an equally valid question is how do we currently handle non-90° rotations?
> 
> There is no need to change anything on the non-90° rotations. When a item is rotated by 180°, then Top Left will be at the former position of Bottom Right. If you look at [window-basic-preview.cpp L.659 and L.660](https://github.com/obsproject/obs-studio/blob/master/UI/window-basic-preview.cpp#L659), you can see that both handles use the same cursor. The same applies to horizontal and vertical cursors. But with a 90° clockwise rotation there is the problem that Top Left will be displayed as Bottom Left, what causes the bug.

To be clear, I'm not asking for changes. I'm just asking, "how does this handle non-90° rotations?" Put another way, could you please describe and/or provide screenshots of the before and after for that scenario? For example, 45° rotations.

BGM99:
As I already mentioned, just rotations that are in between 45° and 135° or 225° and 315° are inverted.

So 45° remains the same:
45grad

The next possible rotation is around 50°, from here my fix inverts the cursor:
50grad

70°, as we get closer to 90°, my fix makes more sense:
70grad

125° with another handle:
125grad


I hope it is now clearer how it currently works and what the scope of my change is.

UI/window-basic-preview.cpp
```diff
@@ -658,9 +658,48 @@ void OBSBasicPreview::UpdateCursor(uint32_t &flags)
...
-	if (cursor().shape() != Qt::ArrowCursor)
+	if ((cursor().shape() != Qt::ArrowCursor) || flags == 0)
 		return;
 
+	if (flags & ITEM_ROT) {
+		setCursor(Qt::OpenHandCursor);
+		return;
+	}
+
+	float rot = obs_sceneitem_get_rot(stretchItem);
```
PatTheMav, line 669: Rename `rot` to `rotation`, `q` to `quadrant`, and `corner` to `isCorner`. We have the storage, we can write out variable names. 😄

PatTheMav (commented):
Pretty nifty amount of bit-shifting but this all seems correct - Opposing diagonal pairs are treated as equivalent to simplify calculations and at some more extreme angles (or right at the edge of quadrants) the chosen handle might not be perfect, but it's still in the overall correct direction (compared to the current status quo).

BGM99:
@PatTheMav Thx for the review. Since I use a circular sector of 45°, the correct term would be "octant". By "q" I actually meant quotient, but octant is more precise. It is true that opposing diagonal pairs are treated as equal, but it shouldn't be a problem since it's irrelevant for the UI and the flag won't be used afterwards. But its important to mention that the flag is incorrect afterwards (when there is a rotation), regardless of this. If this is a problem somehow, then we should call UpdateCursor() with call by value.

PatTheMav:
"Octant" would be actually correct, yeah.

I agree that the interchangeable use of the handles is fine, insofar as it doesn't have a _visual_ impact (the function is entirely unaffected by this).

Fixing that would require a bit more complex bit arithmetic (e.g. just shifting the left/right part to the top/bottom part for non-perpendicular angles wouldn't suffice).

BGM99:
I have tested it visually for all 8 rotation cases, and the resize functionality still works fine. So is this approach mergeable?

PatTheMav (approved):

PatTheMav:
@BGM99 could you please change the commit title to something like `UI: Fix resize cursor to respect item transformation`.

RytoEX (approved):
Behavior with this PR is certainly better than it is now. Seems fine.

PR #7920, opened 2022-12-12
UI: Disable properties button in source toolbar

cg2121:
### Description
If a scene or group is selected, disable the properties button in the source toolbar. This also does a refactor of the function to enable/disable the toolbar buttons.

Screenshot from 2022-12-11 22-49-45

### Motivation and Context
Makes the source toolbar consistent with the source context bar.

### How Has This Been Tested?
Selected/unselected a nested scene in the sources list to make sure the toolbar updated correctly.

### Types of changes

### Checklist:

### DIFF
--- a/UI/window-basic-main.cpp
+++ b/UI/window-basic-main.cpp
@@ -3089,10 +3089,26 @@ void OBSBasic::UpdateContextBarDeferred(bool force)
 				  Qt::QueuedConnection, Q_ARG(bool, force));
 }
 
-void OBSBasic::SourceToolBarActionsSetEnabled(bool enable)
+void OBSBasic::SourceToolBarActionsSetEnabled()
 {
+	bool enable = false;
+	bool disableProps = false;
+
+	OBSSceneItem item = GetCurrentSceneItem();
+
+	if (item) {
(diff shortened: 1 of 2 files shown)


COOLIGUAY:
Should it be the same for groups?
Groups also have no properties.

cg2121:
Updated to also include groups.

UI/window-basic-main.cpp
```diff
@@ -3089,10 +3089,29 @@ void OBSBasic::UpdateContextBarDeferred(bool force)
...
+		OBSSource source = obs_sceneitem_get_source(item);
+		const char *id = obs_source_get_id(source);
+
+		if (strcmp(id, "scene") == 0 || strcmp(id, "group") == 0)
+			disableProps = true;
+
+		enable = true;
+	}
+
+	if (disableProps)
```
gxalpha, line 3109: You can use `obs_source_configurable` for this instead of checking the source id.

gxalpha (changes requested):

UI/window-basic-main.cpp
cg2121: Fixed

PR #7929, opened 2022-12-15
obs-scripting: Add config file functions

exeldro:
### Description
Add config file functions to scripting

### Motivation and Context
Allow to read and write config files from scripts

### How Has This Been Tested?
On windows 10 64bit with lua code: `obs.config_set_uint(obs.obs_frontend_get_profile_config(), "SimpleOutput", "VBitrate", 8000)`

### Types of changes

### Checklist:

### DIFF
--- a/deps/obs-scripting/obslua/obslua.i
+++ b/deps/obs-scripting/obslua/obslua.i
@@ -25,6 +25,7 @@
 #include "cstrcache.h"
 #include "obs-scripting-config.h"
 #include <util/platform.h>
+#include <util/config-file.h>
 
 #if defined(ENABLE_UI)
 #include "obs-frontend-api.h"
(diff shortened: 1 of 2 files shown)


WizardCM (approved):

PatTheMav:
The change itself is fine, but the more interesting question to me is if we really want to expose an entire header's worth of C API calls without checking every single one for compatibility, because inevitably _someone_ will implement a call in a way we didn't think about.

That said, I don't think we put in that effort for the rest of the scripting API and "exposed first, fixed later", so by that token we could merge it.

But moving forward I'd really like to only expose C API calls that we _know_ are safe to use with scripts.

Lain-B:
Yea, I agree. For the time being I think it's fine. I'll just merge it. There's just nothing you can really do with configs via script right now due to the lack of these functions.

PR #7933, opened 2022-12-16
UI: Use stream encoder when resetting encoders

tytan652:
### Description

Use stream encoder when resetting encoder rather than the recording one.

While resetting, OBS will try to restore the actual stream encoder stored in the `lastAdvEnc` variable but it contains the record one which will restored the wrong encoder.

### Motivation and Context

It actually use recording encoder which is wrong.

It could cause issue if the record and steam encoder differ and a reset happen.

"reset": Action that usually happen while the user change services and the actual encoder is no longer available because of compatibility (e.g. HEVC).

### How Has This Been Tested?

Everything work as expected.

### Types of changes

### Checklist:


RytoEX (approved):
Finally tested this. Looks good to me.

PR #7942, opened 2022-12-19
UI: Disable qt5ct when compiled with qt6

kkartaltepe:
### Description

Qt is not smart enough to disable qt5 platform plugins when users blindly ask for them, which causes users to be unable to start OBS. Instead lets kindly disable broken platforms on their behalf.

### Motivation and Context

Users prefer obs to not crash.

### How Has This Been Tested?

It doesnt do anything when I load obs normally. It doesnt fail to load when I set the platform theme.

### Types of changes

### Checklist:

### DIFF
--- a/UI/obs-app.cpp
+++ b/UI/obs-app.cpp
@@ -2302,6 +2302,15 @@ static int run_program(fstream &logFile, int argc, char *argv[])
 
 	setenv("QT_STYLE_OVERRIDE", "Fusion", false);
 
+#if OBS_QT_VERSION == 6
+	/* NOTE: Users blindly set this, but this theme is incompatble with Qt6 and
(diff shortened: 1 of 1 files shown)


tytan652:
My only issue here is that this PR is fixing a situation created by the user itself.

And it is even more complicated since (IIRC) qt6ct also use the name `qt5ct` along with `qt6ct.

kkartaltepe:
> My only issue here is that this PR is fixing a situation created by the user themself.
> 
> And it is even more complicated since (IIRC) qt6ct also use the name `qt5ct` along with `qt6ct`.

Fair, but its by far the biggest crash source since we switched to qt6 and it has not going away. So it seems better to just prevent users from crashing and provide them our default gui. They can still override this by specify the platform theme on the command line if they "know what they are doing".

PR #7950, opened 2022-12-21
Enable realtime mode for graphics thread on OSX

kc5nra:
# DRAFT, just for consideration as I witnessed this issue as I was adding more profiling code

### Description
Fix to inaccurate timer in the main graphics thread.  There are likely different solutions that could apply here like reducing wake up time (sleep_time / 2) so it becomes more "spinny" but often the sleep time is in the 40-50ms off.

Master: **15.0728% within ±2% (43.6836% lower, 41.2436% higher)**
```
info: obs_graphics_thread(16.6667 ms): min=0.205 ms, median=16.623 ms, max=63.61 ms, 15.0728% within ±2% of 16.667 ms (43.6836% lower, 41.2436% higher)
```

PR: **99.6557% within ±2% (0.176358% lower, 0.16796% higher)**
```
info: obs_graphics_thread(16.6667 ms): min=0.078 ms, median=16.667 ms, max=60.686 ms, 99.6557% within ±2% of 16.667 ms (0.176358% lower, 0.16796% higher)
```

## Master-branch timing:
Screen Shot 2022-12-20 at 7 40 52 PM

## Timeline of master branch (magnitude of each item is ms taken during each graphics loop iter)
Screen Shot 2022-12-20 at 7 42 03 PM

## PR-branch timing
Screen Shot 2022-12-20 at 7 38 34 PM

## Timeline of PR branch
Screen Shot 2022-12-20 at 7 43 15 PM

## Flamegraph of several iterations of PR branch
Screen Shot 2022-12-20 at 7 43 35 PM

### Motivation and Context
Stabilize thread timing to be more in line with Linux and Windows timings

### How Has This Been Tested?
This has been profiled from a graphics thread performance standpoint but little real-life testing of how a realtime graphics thread interacts with games running on the system.

### Types of changes

### Checklist:

### DIFF
--- a/libobs/obs-cocoa.m
+++ b/libobs/obs-cocoa.m
@@ -27,6 +27,11 @@
 #include <Carbon/Carbon.h>
 #include <IOKit/hid/IOHIDDevice.h>
 #include <IOKit/hid/IOHIDManager.h>
+#include <Foundation/NSProcessInfo.h>
+
+#include <mach/mach.h>
+#include <mach/mach_time.h>
+#include <pthread.h>
 
 #import <AppKit/AppKit.h>
 
@@ -889,8 +894,38 @@ bool obs_hotkeys_platform_is_pressed(obs_hotkeys_platform_t *plat,
 	return plat->is_key_down[key];
 }
 
+static void mach_set_thread_realtime()
+{
+	mach_timebase_info_data_t timebase_info;
+	mach_timebase_info(&timebase_info);
+
+	const uint64_t NANOS_PER_MSEC = 1000000ULL;
+	double clock2abs =
+		((double)timebase_info.denom / (double)timebase_info.numer) *
+		NANOS_PER_MSEC;
+
+	thread_time_constraint_policy_data_t policy;
+	policy.period = 0;
+	policy.computation = (uint32_t)(5 * clock2abs);
+	policy.constraint = (uint32_t)(10 * clock2abs);
+	policy.preemptible = FALSE;
+
+	int kr = thread_policy_set(pthread_mach_thread_np(pthread_self()),
+				   THREAD_TIME_CONSTRAINT_POLICY,
+				   (thread_policy_t)&policy,
+				   THREAD_TIME_CONSTRAINT_POLICY_COUNT);
+	if (kr != KERN_SUCCESS) {
+		blog(LOG_ERROR,
+		     "error setting rt thread, thread_policy_set: %s",
+		     mach_error_type(kr));
+		exit(1);
+	}
+}
+
 void *obs_graphics_thread_autorelease(void *param)
 {
+	// Set the priority of this thread for osx for better timer precision
+	mach_set_thread_realtime();
 	@autoreleasepool {
 		return obs_graphics_thread(param);
 	}


libobs/obs-cocoa.m
```diff
@@ -889,8 +894,38 @@ bool obs_hotkeys_platform_is_pressed(obs_hotkeys_platform_t *plat,
...
+	mach_timebase_info_data_t timebase_info;
+	mach_timebase_info(&timebase_info);
+
+	const uint64_t NANOS_PER_MSEC = 1000000ULL;
+	double clock2abs =
+		((double)timebase_info.denom / (double)timebase_info.numer) *
+		NANOS_PER_MSEC;
+
+	thread_time_constraint_policy_data_t policy;
+	policy.period = 0;
```
jpark37, line 908: I don't think 0 is a valid value for period. I'm looking at how Chromium is using this API for audio, and one code path shows:
period = 2.9 ms
computation = 75% * period
constraint = 85% * period
preemptible = NO

Maybe OBS could do similar for the graphics thread, but use the reciprocal of frame rate instead of 0 or 2.9 ms.

kc5nra: Hey @jpark37 , thanks for looking into that!
These were the values as described by apple documentation but looking at the chromium source I would tend to agree.

Here is a comparison of 0 period vs 16ms period:
Screen Shot 2022-12-21 at 8 03 08 AM

## 0ms period:
Screen Shot 2022-12-21 at 8 06 17 AM

## 16ms period
Screen Shot 2022-12-21 at 8 07 19 AM

I'll go ahead and move this to 16ms for now

kc5nra:
> 1. Use "macOS" when describing the current Apple operating system. It hasn't been called "OSX" in a while.
> 2. Use an appropriate commit prefix (in this case, "libobs: ").

Fixed

libobs/obs-cocoa.m
PatTheMav: Wouldn't that lock it into supporting 60fps as a maximum though?

kc5nra: It uses the obs interval now, during the draft I was hardcoding things

kc5nra: I wasn't able to actually witness any demonstrable difference between different periods (16ms period with 30fps stream or 33ms period with 60fps stream).

PatTheMav:
If we're working with thread policies already, would it be possible to add correct QoS values so the graphics thread gets the scheduling priority it needs, especially on Apple Silicon systems?

RytoEX:
> @kc5nra
[libobs: Enable rt for graphics thread on macOS](https://github.com/obsproject/obs-studio/pull/7950/commits/bb35a000ac9d4e510ec040c1b511f42143a0d985)

Please don't abbreviate realtime to rt. Commit messages should be easily understood. "libobs: Enable realtime for graphics thread on macOS" is short enough.

kc5nra:
@PatTheMav info: obs_graphics_thread(33.3333 ms): min=0.205 ms, median=33.334 ms, max=79.73 ms, 77.3852% within ±2% of 33.333 ms (10.689% lower, 11.9258% higher)
Adding QoS ```pthread_set_qos_class_self_np(QOS_CLASS_USER_INTERACTIVE, 0);``` by itself does reduce the jitter somewhat. I'm unclear what the interaction between QoS and thread_policy are

kc5nra:
Added QoS prio 47 for rendering (0 offset) and renamed the commit.

libobs/obs-cocoa.m
```diff
@@ -889,8 +894,48 @@ bool obs_hotkeys_platform_is_pressed(obs_hotkeys_platform_t *plat,
...
+	}
+
+	pthread_set_qos_class_self_np(QOS_CLASS_USER_INTERACTIVE, 0);
+}
+
 void *obs_graphics_thread_autorelease(void *param)
 {
+	// Set the priority of this thread for macOS for better timer precision
+	const double GRAPHICS_THREAD_BUSY = 0.5;
+	const double GRAPHICS_THREAD_BUSY_LIMIT = 1.0;
```
jpark37, line 931: I don't know if the OS is going to like period == constraint. That's like saying you may never give up the core. This isn't audio, but I wouldn't mind .75/.85 is here.

jpark37: And 15% might seem like a lot of slop, but OBS isn't a game where the render thread could need a lot of frame time. The two CPU-intensive tasks that OBS may do depending on configuration, FFmpeg swscale resizing and software encoding, happens on other threads. Assuming I'm not forgetting something.

PatTheMav: `swscale` should happen on the media-io or video thread iirc.

libobs/obs-cocoa.m
```diff
@@ -889,8 +894,48 @@ bool obs_hotkeys_platform_is_pressed(obs_hotkeys_platform_t *plat,
...
+				   THREAD_TIME_CONSTRAINT_POLICY,
+				   (thread_policy_t)&policy,
+				   THREAD_TIME_CONSTRAINT_POLICY_COUNT);
+	if (kr != KERN_SUCCESS) {
+		blog(LOG_ERROR,
+		     "error setting rt thread, thread_policy_set: %s",
+		     mach_error_type(kr));
+	}
+
+	pthread_set_qos_class_self_np(QOS_CLASS_USER_INTERACTIVE, 0);
```
jpark37, line 924: I don't know if we want pthread_set_qos_class_self_np. If it's not a huge win, I'd rather not cross the streams and make things more complicated.

kc5nra: It does not appear to alter the behavior of manually specifying constraints

libobs/obs-cocoa.m
kc5nra: @jpark37 From light testing, 0.75/0.85 appears to gain the same benefits under light load. Where the CPU is more contested the values will likely diverge but I agree that, in the short term, it's both safer from a regression standpoint and more polite the system as a whole to reduce the busy max guarantee

libobs/obs-cocoa.m
jpark37: My impression of the Apple video is that it's using pthread because it doesn't want games to use realtime priority. An argument could be made that the OBS graphics thread hitting its deadline is more critical than a game starting its render thread from a consistent point.

We should probably use one or the other, either the mach APIs for realtime priority, or pthread APIs for high priority, and I wouldn't mind staying with mach. At least, that seems cleaner to me from this documentation:
https://developer.apple.com/library/archive/documentation/Darwin/Conceptual/KernelProgramming/scheduler/scheduler.html

jpark37:
> These were the values as described by apple documentation but looking at the chromium source I would tend to agree.

Not that it matters much since you've already changed it, but I'm curious where you saw period=0 because the only Apple example I see has this:
```
ttcpolicy.period=period; // HZ/160
ttcpolicy.computation=computation; // HZ/3300;
ttcpolicy.constraint=constraint; // HZ/2200;
ttcpolicy.preemptible=1;
```

kc5nra:
> > These were the values as described by apple documentation but looking at the chromium source I would tend to agree.
> 
> Not that it matters much since you've already changed it, but I'm curious where you saw period=0 because the only Apple example I see has this:
> 
> ```
> ttcpolicy.period=period; // HZ/160
> ttcpolicy.computation=computation; // HZ/3300;
> ttcpolicy.constraint=constraint; // HZ/2200;
> ttcpolicy.preemptible=1;
> ```

https://developer.apple.com/library/archive/technotes/tn2169/_index.html

libobs/obs-cocoa.m
kc5nra: Sorry, by my statement I was in support of removing as it appears to add nothing. (Just for clarification)

jpark37: No worries. I hadn't even seen your reply as I was writing that wall of text. :P

libobs/obs-cocoa.m
jpark37: Sounds good.

libobs/obs-cocoa.m
PatTheMav: Yeah I think James is right that QoS has probably no bearing on realtime threads, though it sounds like we should not use them in the first place (if games are discouraged from it).

Note that all these "old" documents from the Developer archive do not take the changes/differences of the new Apple Silicon architecture into account, so things might actually not fully apply anymore.

jpark37:
Thanks, I think I get understand period=0. If the thread is woken to do stuff, it should only stay awake according to the other values before voluntarily giving itself up, and when it gets woken is not on an interval.

libobs/obs-cocoa.m
kc5nra: Though, from my light testing, utilizing only QoS doesn't quite get us into the accuracy of timers we'd prefer

PatTheMav: Did you try with `SCHED_RR` and QoS 45 as the video suggested? Otherwise I'm happy to defer to James' judgement.

jpark37: Then I say Mach it up!

kc5nra: Sched RR disables qos from my understanding, I'm afk but I'll double check

PatTheMav: Yeah it's then called "priority" it seems:

Bildschirm­foto 2022-12-21 um 18 41 30

kc5nra: info: obs_graphics_thread(16.6667 ms): min=0.22 ms, median=16.666 ms, max=73.189 ms, 63.9489% within ±2% of 16.667 ms (18.4185% lower, 17.6326% higher)

Sched RR 45

kc5nra: Reverted QoS addition and switched to .75/.85

PatTheMav:
@kc5nra I asked Apple about this issue and received some feedback:

* Add `pthread_attr_set_qos_class_np` with `QOS_CLASS_USER_INTERACTIVE` to the `phread` creation in https://github.com/obsproject/obs-studio/blob/167f53941643f570b5b01e13b34dfa8d7ab3a567/libobs/obs.c#L430
* This should make the timer code run better without any changes to `os_sleepto_ns` or other tweaks
* Overall we should not use real time for the graphics thread, as this might also preempt all other work done on the machine, including the game/software being captured (so OBS would severely impact other apps on the system, as those should usually not use real time threading this way either).

I will still need to clarify some things, as the graphics thread is "abused" for other scheduling tasks that are not related to GPU work.

kc5nra:
> @kc5nra I asked Apple about this issue and received some feedback:
> 
> * Add `pthread_attr_set_qos_class_np` with `QOS_CLASS_USER_INTERACTIVE` to the `phread` creation in https://github.com/obsproject/obs-studio/blob/167f53941643f570b5b01e13b34dfa8d7ab3a567/libobs/obs.c#L430
> * This should make the timer code run better without any changes to `os_sleepto_ns` or other tweaks
> * Overall we should not use real time for the graphics thread, as this might also preempt all other work done on the machine, including the game/software being captured (so OBS would severely impact other apps on the system, as those should usually not use real time threading this way either).
> 
> I will still need to clarify some things, as the graphics thread is "abused" for other scheduling tasks that are not related to GPU work.

I would say this is still not accurate enough. 

```info: obs_graphics_thread(16.6667 ms): min=0.207 ms, median=16.667 ms, max=196.966 ms, 67.9839% within ±2% of 16.667 ms (16.1638% lower, 15.8523% higher)```

Screen Shot 2023-01-12 at 8 04 30 AM

PatTheMav:
I received further feedback from Apple who tested this PR, their recommended changes, and OBS v29 with real-life usage scenarios and the data was inconclusive (in the provided metrics the QoS variant performed slightly better than the realtime variant). 

On the whole the feedback by Apple and its developers is "that's a bad idea" and "not the right way to use RT threads". If we need high-precision timers, we should use macOS' (or one of the specialised frameworks') timers (dispatched by the OS itself) instead of relying on a thread of ours - using our graphics thread for that is very "un-macOS-ish".

PatTheMav:
@kc5nra I discussed this PR internally and if you could update the PR to use the QoS attribute with `pthread_create` instead, we could merge this soon-ish. Also feel free to retain the current change in a different branch so we could possibly revisit that solution later.

PatTheMav (approved):

PR #7953, opened 2022-12-21
obs-ffmpeg: Add quality preset for VA-API

nowrep:
### Description
Added support for VA-API quality preset in Mesa >= 22.3.
https://gitlab.freedesktop.org/mesa/mesa/-/commit/a727ec83ba01776e9c6f61c912c4caac280e968e

### Motivation and Context
Better encoding quality on AMD cards on Linux.

### How Has This Been Tested?
On ArchLinux (mesa 22.3.1, ffmpeg 5.1.2).

### Types of changes

### Checklist:

### DIFF
--- a/plugins/obs-ffmpeg/vaapi-utils.c
+++ b/plugins/obs-ffmpeg/vaapi-utils.c
@@ -185,6 +185,25 @@ bool vaapi_device_rc_supported(VAProfile profile, VADisplay dpy, uint32_t rc,
 	return false;
 }
 
+bool vaapi_quality_preset_supported(VAProfile profile, VADisplay dpy)
+{
+	const char *vendor_string = vaQueryVendorString(dpy);
+	if (strncmp("Mesa Gallium", vendor_string, 12) != 0)
+		return false;
+
+	VAStatus va_status;
+	VAConfigAttrib attrib[1];
+	attrib->type = VAConfigAttribEncQualityRange;
+
+	va_status = vaGetConfigAttributes(dpy, profile, VAEntrypointEncSlice,
+					  attrib, 1);
+
+	if (va_status != VA_STATUS_SUCCESS)
+		return false;
+
+	return attrib->value != VA_ATTRIB_NOT_SUPPORTED;
+}
+
 #define CHECK_PROFILE(ret, profile, va_dpy, device_path)                      \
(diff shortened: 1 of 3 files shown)


tytan652:
Screenshot of the UI change are missing.

tytan652:
> On ArchLinux with mesa-git.
Please test on a real release not a git package.

nowrep:
Done.

plugins/obs-ffmpeg/vaapi-utils.c
```diff
@@ -185,6 +185,25 @@ bool vaapi_device_rc_supported(VAProfile profile, VADisplay dpy, uint32_t rc,
 	return false;
 }
 
+bool vaapi_quality_preset_supported(VAProfile profile, VADisplay dpy)
+{
+	const char *vendor_string = vaQueryVendorString(dpy);
+	if (strncmp("Mesa Gallium", vendor_string, 12) != 0)
+		return false;
```
tytan652, line 192: Sorry, but driver filters is a no-go.

nowrep: It has to be like this, since the quality value has different meaning on other va drivers (intel).

tytan652: If this is not supported any drivers, it is going to be an issue.

kkartaltepe: The problem here appears to be that AMD decided to just ignore the defined meaning of this value and come up with their own new and exciting meaning for the parameter. 

https://intel.github.io/libva/structVAEncMiscParameterBufferQualityLevel.html - A lower value means higher quality, and a value of 1 represents the highest quality. The quality level setting is used as a trade-off between quality and speed/power consumption, with higher quality corresponds to lower speed and higher power consumption.

https://gitlab.freedesktop.org/mesa/mesa/-/blob/a727ec83ba01776e9c6f61c912c4caac280e968e/src/gallium/frontends/va/config.c#L213 - AMD decided to use this to control a whole suite of encoder modes without regard to the API definition.

I'm not a fan of trying to divine the meaning of a parameter after AMD intentionally broke the api for their implementation, better to just ignore this if drivers want to use this a special out of spec signal since they may change it any time on us.

PR #7961, opened 2022-12-23
Add 8-band Equalizer

phkahler:
### Description
This PR adds an 8-channel equalizer to OBS studio

### Motivation and Context
I used OBS once and wanted an EQ badly. It turns out this is a common request for many years.

### How Has This Been Tested?
I tested this by playing an .mp3 in the default gnome music player, applying this filter to the desktop audio in OBS and monitoring in headphones. All sliders work.

### Types of changes
Strictly the addition of a new audio filter. Similar to the recently added 3-band.

### Checklist:


pkviet:
The property labels should be localized. This means you'll have to add entries to en-us.ini in the data subfolder. Check the other filters to have a model.

phkahler:
> The property labels should be localized. This means you'll have to add entries to en-us.ini in the data subfolder. Check the other filters to have a model.

I was wondering why my eq8. prefix kept appearing in the UI when that didn't happen in the 3-band one. I ended up removing it in the code. I'll put it back and do the same - removing during localization.

pkviet (commented):
Please change the commit title to follow our guidelines.

pkviet (changes requested):
also run clang-format

notr1ch:
The commit title needs a space after the module prefix.

phkahler:
Is the "appropriate documentation" checkbox for PRs for the code or user docs? I don't really see a place for either and the code seems much like the other filters.

Updated the commit message and localized the properties.

notr1ch:
Documentation is primarily for libobs changes, e.g. if you introduce a new API or change behavior of an existing API.  Not necessary for internal components like this.

phkahler:
@pkviet  Is there anything more you'd like from me on this?

phkahler:
Fixed the commit message.

phkahler:
@pkviet  I consider this done, unless someone finds an issue with it or has suggestions to make it better. I'm not sure how this project operates, but it might be good to merge after a major release so some set of people (developers) get to test it prior to the next one. I'll be around if they find anything even though this is kind of drive-by PR from me ;-)

Fenrirthviti:
After internal discussion and now that we have a basic 3 band EQ, we don't feel this is adding anything meaningful to do natively in OBS, and users who want that extra level of control will probably have a preferred VST that would work better anyway.

Thank you for the submission.

PR #7965, opened 2022-12-26
mac-capture: add audio-only toggle to macOS Screen Capture

jcm93:
### Description

Adds a toggle to macOS Screen Capture properties for audio-only stream output for the selected window/display/application, rather than both an audio and video stream output.

I have discussed this to a limited degree previously with gxalpha and Pat. This was one suggested possible route to achieve the desired functionality that I decided to go ahead and implement to get the ball rolling. If macOS 13 is available, the checkbox is presented to discard video. If ``discard_video`` is ``true``, only the audio stream output is activated rather than both audio and video streams. The capture struct's ``frame`` is also set to ``CGRectZero``. This is my first contribution so I hope I have not violated any patterns with this, but I tested it on my machine to make sure it behaves as expected.

I hope that even if this is not implemented it serves to start the discussion for how it can be properly implemented!

Screenshot 2022-12-25 at 11 54 50 PM

### Motivation and Context

The overall motivation for this PR is to provide a way for program and desktop audio to be captured by itself on macOS, without associated video. Currently this requires the use of third party programs like BlackHole, or alternately requires the user to add an SCK source and then resize it to zero or place it off the canvas.

There has been some discussion as to the best way to achieve this goal. This PR represents one way to achieve the functionality, with minimal changes to the existing SCK source to accommodate audio-only streams.

### How Has This Been Tested?

Tested locally on my machine with macOS 13.2, M1 MacBook Pro.

### Types of changes

### Checklist:


GeorgesStavracas:
There's been agreement while discussing this pull request that this is not how we want to proceed with this feature. The main problem here is that this breaks the paradigm of screen captures providing video. The more appropriate way forward is adding a new audio-only source type, which is what we present on Windows.

jcm93:
Understood. Would a separate audio source still attach to the same SCStream object as the video capture when possible, as SCK seems to "want"? When I was thinking about a dedicated audio source solution, it seemed to me like there would be some complexities in the implementation there, what with SCK seemingly designed around the audio and video streams being attached to the same SCStream object. Curious what the project's thoughts are there on how to handle those shared objects, or whether it's fine to simply create multiple separate SCStream objects, one for audio and one for video.

No worries if this isn't the venue for this discussion, I am just curious to know more.

gxalpha:
This is a question that probably only Apple can answer with certainty, but it seems fine to me to use two completely different instances.
Trying to share them between sources feels like too much work and added code complexity.

gxalpha:
Closing in favor of #8778. We'll get around to reviewing #8778 again hopefully soon, apologies for the delay on that.

jcm93:
No worries, let me know whenever if you want a rebase.

PR #7986, opened 2022-12-30
linux-pipewire: Check format availablity against DRM only for dmabufs

### LINKED ISSUE #7985: pipewire format negotiation can fail to fallback to shared memory
### Operating System Info

Fedora 37

### Other OS

_No response_

### OBS Studio Version

29.0.0-beta2

_No response_

### OBS Studio Log URL

https://obsproject.com/logs/GJUwf1iqZLQRDfBv

### OBS Studio Crash Log URL

_No response_

### Expected Behavior

Fallback to shared memory should always work, and the negotiation should eventually succeed (or fail)

### Current Behavior

We enter an infinite loop attempting to negotiate format. We also attempt to bmalloc 0 bytes which we shouldnt do.

### Steps to Reproduce

1. Run your compositor on a driver that supports modifiers
2. Run obs on a software renderer, or possibly older mesa versions with different modifier support than native.
3. Observe negotiation continues infinitely.
...

_No response_

columbarius:
### Description

Fixes pipewire-screencast when using a software renderer

### Motivation and Context

Whe using a software renderer there will be no formats available for using with dmabufs. We should only consider those formats wrt. to modifiers and as such move this check after adding general support for that format.

Fixes #7985 

### How Has This Been Tested?

Run obs with `LIBGL_ALWAYS_SOFTWARE=true`

### Types of changes

### Checklist:

### DIFF
--- a/plugins/linux-pipewire/pipewire.c
+++ b/plugins/linux-pipewire/pipewire.c
@@ -359,16 +359,14 @@ static void init_format_info(obs_pipewire_data *obs_pw)
 	for (size_t i = 0; i < N_SUPPORTED_FORMATS; i++) {
 		struct format_info *info;
 
-		if (!drm_format_available(supported_formats[i].drm_format,
-					  drm_formats, n_drm_formats))
-			continue;
-
 		info = da_push_back_new(obs_pw->format_info);
 		da_init(info->modifiers);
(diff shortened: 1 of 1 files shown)


kkartaltepe (approved):
Works well on my machine, thanks.

PR #7994, opened 2023-01-01
NVENC error handling improvements

notr1ch:
### Description
This copies the result of `nvEncGetLastErrorString` to the last encoder error instead of only writing it to the log file. This makes it user-visible, which should help users diagnose and solve the problem.

Also strips off leading colons from NVENC error strings as these can look odd given our format string. An example of such an error:

`::NV_ENC_TWO_PASS_FULL_RESOLUTION is only supported multipass encoding on this architecture.`

Additionally handles `NV_ENC_ERR_NO_ENCODE_DEVICE` with the unsupported device message and blacklists a GeForce MX350 variant.

### Motivation and Context
While the full error message in the log file is useful, it's not really visible to users. Showing it in the encoder failure message box should help users solve the problem.

### How Has This Been Tested?
Not tested very well as I couldn't easily trigger error conditions on my GPU, code seems OK though?

### Types of changes

### Checklist:

### DIFF
--- a/plugins/obs-ffmpeg/jim-nvenc-helpers.c
+++ b/plugins/obs-ffmpeg/jim-nvenc-helpers.c
@@ -37,11 +37,22 @@ bool nv_failed2(obs_encoder_t *encoder, void *session, NVENCSTATUS err,
 		const char *func, const char *call)
 {
 	struct dstr error_message = {0};
+	const char *nvenc_error = NULL;
 
-	switch (err) {
-	case NV_ENC_SUCCESS:
+	if (err == NV_ENC_SUCCESS)
 		return false;
 
+	if (session) {
+		nvenc_error = nv.nvEncGetLastErrorString(session);
+		if (nvenc_error) {
+			// Some NVENC errors begin with :: which looks
+			// odd to users. Strip it off.
+			while (*nvenc_error == ':')
+				nvenc_error++;
+		}
+	}
+
+	switch (err) {
 	case NV_ENC_ERR_OUT_OF_MEMORY:
 		obs_encoder_set_last_error(
 			encoder, obs_module_text("NVENC.TooManySessions"));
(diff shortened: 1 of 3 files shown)


flaeri:
Seems to be working fine for me. Some examples that I could think of:

1. Too many encodes running
image

2. Resolution too high
image

3. Incorrect GPU chosen
image

RytoEX:
Any other specific items you'd like to try to verify @notr1ch ?

notr1ch:
Seeing the error in the dialog like that makes me wonder if we should remove the function call information, users may begin to read the error text and immediately give up as soon as it starts looking too technical. Perhaps adding a couple of newlines between the parts could help too.

RytoEX:
> Seeing the error in the dialog like that makes me wonder if we should remove the function call information, users may begin to read the error text and immediately give up as soon as it starts looking too technical. Perhaps adding a couple of newlines between the parts could help too.

Move the technical parts to the bottom, if possible?

notr1ch:
I honestly don't think the technical info is useful at all in the message box, it still gets sent to the log file for developers to review. I'm worried that exposing too much technical information will still confuse users regardless of where we position it. How about something like this:

> NVENC Error: Frame dimension greater than the maximum supported value (NV_ENC_ERR_INVALID_PARAM)

On another note it also seems the GPU selection error has regressed at some point, that used to give explicit feedback telling the user to set their GPU index back to 0. I'll see if I can fix that in this PR too.

notr1ch:
Some more updates: GPU index error is now visible again, `NV_ENC_ERR_NO_ENCODE_DEVICE` is handled with a nicer error message and a GeForce MX350 variant without NVENC was added to the unsupported devices list.

RytoEX:
Do you need re-tests for this?

notr1ch:
I don't think so, as long as everyone is happy with the new error format `NVENC Error: text (error code)` I think this is good to go.

RytoEX (approved):
Perfect is the enemy of done. Looks good to me.

PR #8002, opened 2023-01-03
libobs: Fix bug with scene_audio_render() 

oceancx:
### Description

In obs-scene.c file, function scene_audio_render() has following code.
```
pos = (size_t)ns_to_audio_frames(sample_rate,
					source_ts - timestamp);
count = AUDIO_OUTPUT_FRAMES - pos;

if (!apply_buf && !item->visible &&
		  !transition_active(item->hide_transition)) {
	item = item->next;
	continue;
}

obs_source_get_audio_mix(source, &child_audio);

for (size_t mix = 0; mix < MAX_AUDIO_MIXES; mix++) {
	if ((mixers & (1 << mix)) == 0)
		continue;

	for (size_t ch = 0; ch < channels; ch++) {
		float *out = audio_output->output[mix].data[ch];
		float *in = child_audio.output[mix].data[ch];

		if (apply_buf)
			mix_audio_with_buf(out, in, buf, pos,
							 count);
		else
			mix_audio(out, in, pos, count);
	}
}
```

what this snippet code basically do is to align source's audio_output_buf to the mininum timestamp, so here `pos` and `count` was calculated. then, apply them to mix_audio() function. `audio_output->output[mix].data[ch]` and `child_audio.output[mix].data[ch]` both are one audio frame's channel data, which contains 1024 floats. the difference between them is `out` corresponding to timestamp, while `in`  corresponding to source_ts. we should use `pos` and `count` to align them in mix_audio. what currently do is following code, but I this is incorrect.

```
static void mix_audio_with_buf(float *p_out, float *p_in, float *buf_in,
			       size_t pos, size_t count)
{
	register float *out = p_out;
	register float *buf = buf_in + pos;
	register float *in = p_in + pos;
	register float *end = in + count;

	while (in < end)
		*(out++) += *(in++) * *(buf++);
}

static inline void mix_audio(float *p_out, float *p_in, size_t pos,
			     size_t count)
{
	register float *out = p_out;
	register float *in = p_in + pos;
	register float *end = in + count;

	while (in < end)
		*(out++) += *(in++);
}
```

what we should do is to use out+pos instead of in+pos to align.

```
static void mix_audio_with_buf(float *p_out, float *p_in, float *buf_in,
			       size_t pos, size_t count)
{
	register float *out = p_out + pos;
	register float *buf = buf_in;
	register float *in = p_in;
	register float *end = in + count;

	while (in < end)
		*(out++) += *(in++) * *(buf++);
}

static inline void mix_audio(float *p_out, float *p_in, size_t pos,
			     size_t count)
{
	register float *out = p_out + pos;
	register float *in = p_in;
	register float *end = in + count;

	while (in < end)
		*(out++) += *(in++);
}
```
when `pos` is 0, we got no problem, but when `pos` is not 0, there is a bug.

### Motivation and Context

this bug happens when there are two audio sources in obs, one's audio stamp is ahead, and play the other source whose audio stamp is far behind to trigger add_buffering logic, which in most cases will cause the played source `pos` not equal 0 and trigger this bug.

### How Has This Been Tested?

has tested and solve the audio blast issue we have been encountered.

### Types of changes

### Checklist:

- [ x ] The code has been tested.

### DIFF
--- a/libobs/obs-scene.c
+++ b/libobs/obs-scene.c
@@ -1363,9 +1363,9 @@ static void process_all_audio_actions(struct obs_scene_item *item,
 static void mix_audio_with_buf(float *p_out, float *p_in, float *buf_in,
 			       size_t pos, size_t count)
 {
-	register float *out = p_out;
-	register float *buf = buf_in + pos;
-	register float *in = p_in + pos;
+	register float *out = p_out + pos;
+	register float *buf = buf_in;
+	register float *in = p_in;
 	register float *end = in + count;
 
 	while (in < end)
@@ -1375,8 +1375,8 @@ static void mix_audio_with_buf(float *p_out, float *p_in, float *buf_in,
 static inline void mix_audio(float *p_out, float *p_in, size_t pos,
 			     size_t count)
 {
-	register float *out = p_out;
-	register float *in = p_in + pos;
+	register float *out = p_out + pos;
+	register float *in = p_in;
 	register float *end = in + count;
 
 	while (in < end)


RytoEX:
> has tested and solve the audio blast issue we have been encountered.

What "audio blast issue"? How does one reproduce the bug?

oceancx:
> > has tested and solve the audio blast issue we have been encountered.
> 
> What "audio blast issue"? How does one reproduce the bug?
image

you can test this bug by adding two sources in obs. one is "NDI source", the other is "Media source". 
image

"NDI source" sync mode choose to "Network", by doing this, we expect NDI's audio timestamp is earlier than audio sound, in this situation, when you replay audio sound, you will align these two sources audio timestamps in scene_audio_render() written in obs-scene.c file, and you will get audio sound's `pos` not equal 0 and you will get a wrong result in the first audio frame in obs record file which was demonstrated above screenshot.

tt2468:
Would you be able to post another screenshot comparing the before and after waveforms?

oceancx:
image

image

image

@tt2468

oceancx:
any follow up on this issue? @gxalpha @RytoEX @tt2468 @eric

RytoEX:
Frankly, I do not understand the issue. @tt2468, @norihiro, or @pkviet are better suited to understand this.

norihiro:
I can reproduce the case `pos` becomes non-zero by following steps on Linux, and I confirmed this PR fixes the bug.

1. Create an empty scene collection.
2. Add `Audio Input Capture (Pulse)` to the scene.
3. Add `Media Source` with an audio file. (I used a chirp starting from the first sample.)
4. Start recording
5. Replay the media source by clicking the replay button after it stops several times.

(Step 2 can be replaced by an [asynchronous audio source](https://github.com/norihiro/obs-asynchronous-audio-source/). I guess any audio source should be fine.)

The variable `pos` indicates `source_ts - timestamp` in samples.
When `pos > 0`, the source samples are lagged like this figure. It makes sense to add `pos` to `out`.
pr8002-pos

I saw other issues that should be fixed by other PRs.
- Sometimes `pos` becomes a huge number exceeding `AUDIO_OUTPUT_FRAMES` and underflow happens. Since we use the comparison `in < end`, maybe it won't make an actual issue.
- Last samples in the wave file are omitted with and without this PR. This is an unrelated issue.

Issue #3929, opened 2020-12-18
Virtual camera fails to start


topongo:
## Platform

Operating system and version: Arch Linux - Rolling
OBS Studio version: 26.1.0.r9

## Expected Behavior

Clicking on "Start Virtual Camera" button should start the virtual cam flawlessly

## Current Behavior

Pressing the button does not create any effect.

## Steps to Reproduce

1. Install obs-studio-git from AUR
2. Remove the old CatxFish/obs-v4l2sink plugin
3. Install v4l2loopback-dkms
4. Start OBS and click "Start Virtual Camera" button.

## Additional information

The log says "warning: Failed to start virtual camera" when clicking on the button.
I've installed droidcam alongside OBS, and it currently uses the /dev/video0 device. I thought maybe obs is trying to use that device and it conflicts with droidcam, but disabling (not removing) droidcam doesn't solve the issue.

pardomi:
I am experiencing the same issue.

**Platform**
Operating system and version: Kubuntu 20.10
OBS Studio version: 26.1

**Expected Behavior**
Clicking on "Start Virtual Camera" button should start the virtual cam flawlessly

**Current Behavior**
Pressing the button does not create any effect.  (button text remains "Start Virtual Camera"

**Steps to Reproduce**
    Install obs-studio 26.1 from ppa:obsproject/obs-studio
    Install v4l2loopback-dkms (previously installed)
    Start OBS and click "Start Virtual Camera" button.

**Additional information**

- The log says "warning: Failed to start virtual camera" when clicking on the button.
- I do not have droidcam or any other camera manipulation software installed (as far as I know)
- I had a previous version of OBS Studio installed (from Ubuntu 20.10 repository - uninstalled before installing 26.1 from PPA).

**Log File Contents**
03:19:54 PM.696: CPU Name: Intel(R) Core(TM) i5-7200U CPU @ 2.50GHz
03:19:54 PM.696: CPU Speed: 1830.528MHz
03:19:54 PM.696: Physical Cores: 2, Logical Cores: 4
03:19:54 PM.696: Physical Memory: 11850MB Total, 2914MB Free
03:19:54 PM.696: Kernel Version: Linux 5.8.0-33-generic
03:19:54 PM.696: Distribution: "Ubuntu" "20.10"
03:19:54 PM.696: Session Type: x11
03:19:54 PM.697: Window System: X11.0, Vendor: The X.Org Foundation, Version: 1.20.9
[...]

PR #8008, opened 2023-01-05
linux-v4l2: Create & use an OBS specific loopback card

aramg:
### Description
Create and use an OBS specific loopback card by taking advantage of `v4l2loopback-ctl add` command. Fallback to any available option if that fails.

### Motivation and Context
Improved user experience when multiple loopback cards are present.

### How Has This Been Tested?
1. Click [Start Virtual Cam], v4l2loopback is loaded with label "OBS Virtual Camera".
2. Stop and Re-Start virtual cam, "OBS Virtual Camera" is reused.

--

1. Unload v4l2loopback, and re-add manually ("Dummy" card is created).
2. Launch OBS and click [Start Virtual Cam], a new card is added with label "OBS Virtual Camera" and used.
3. Stop and Re-Start virtual cam, "OBS Virtual Camera" is reused.

--

1. Delete the "OBS Virtual Camera" card and move the `v4l2loopback-ctl` executable away.
2. Launch OBS and click [Start Virtual Cam], the existing Dummy card is selected.

--

### Types of changes

### Checklist:


aramg:
Potentially closes #3929

plugins/linux-v4l2/v4l2-output.c
```diff
@@ -220,18 +209,63 @@ static bool virtualcam_start(void *data)
...
 			     " This can be ignored since it is quite improbable.");
 
-		if (try_connect(vcam, device)) {
-			success = true;
-			break;
+		int fd = open(device, O_RDWR);
+		if (fd < 0)
+			continue;
+
+		if (ioctl(fd, VIDIOC_QUERYCAP, &v4l2cap) < 0)
```
kkartaltepe, line 216: leaving these checks that must jump inside try_configure to avoid these goto's seems a bit cleaner.

plugins/linux-v4l2/v4l2-output.c
```diff
@@ -111,51 +121,42 @@ static void *virtualcam_create(obs_data_t *settings, obs_output_t *output)
...
-		goto fail_close_device;
+		return false;
+
+	uint32_t width = obs_output_get_width(vcam->output);
+	uint32_t height = obs_output_get_height(vcam->output);
+	vcam->frame_size = width * height * 2;
 
 	format.fmt.pix.width = width;
 	format.fmt.pix.height = height;
+	format.fmt.pix.field = V4L2_FIELD_NONE;
```
kkartaltepe, line 154: what is the purpose of this change?

kkartaltepe:
> Potentially closes #3929

Yea feel free to add a fixes marker to the commit. I think scanning for our label is fine as a solution, though users who did not heed the warning to use exactly our module loading config will likely be upset.

plugins/linux-v4l2/v4l2-output.c
```diff
@@ -220,18 +209,63 @@ static bool virtualcam_start(void *data)
...
+	if (!loopback_module_loaded()) {
+		if (loopback_module_load() != 0) {
+			return false;
+		}
+	}
+
+	bool success = open_loopback_card(data, OBS_V4L2_CARD_LABEL);
+
+	if (!success) {
+		loopback_module_add_card();
```
kkartaltepe, line 255: Can we not get the device number back out of this operation?

plugins/linux-v4l2/v4l2-output.c
aramg: tbh I am not sure if this makes a difference in any specific scenarios, but the default is V4L2_FIELD_ANY (0) whereas V4L2_FIELD_NONE (1) is for progressive formats (not interlaced). 
Setting this value to hopefully prevent potential odd behavior.
  
https://linuxtv.org/downloads/legacy/video4linux/API/V4L2_API/spec/ch03s06.html

plugins/linux-v4l2/v4l2-output.c
aramg: Added a TODO: need to refactor `run_command` to return the cmd output, as one way of getting the new device id.

kkartaltepe:
I was able to test this out, I think the only problem is that there is no checking for v4l2loopback-ctl having the required `add` command or that v4l2loopback-ctl has failed. This means that it will continually prompt the user for permission when they start the webcam if v4l2loopback-ctl fails and they have an existing incorrect v4l2loopback device loaded.

1) unload v4l2loopback
1) load up a non-obs device e.g. `modprobe v4l2loopback video_nr=0 card_label=Video-Loopback exclusive_caps=1`
1) Start virtual cam in obs, get pkexec prompt
1) stop virtual cam
1) Start virtual cam, get another pkexec. (We should probably remember this failed and just give up on v4l2loopback-ctl)

kkartaltepe:
Also where does this work, as I was testing on archlinux and there is no add function. Is this still unreleased by upstream?

aramg:
I just did some checking with a few images, and it seems the `add` command is largely missing, even on Ubuntu which I swear is where I'd originally seen it :/

The current binary in my dev environment was built from source it seems.

Looking at [v4l2loopback/utils](https://github.com/umlaeute/v4l2loopback/tree/main/utils), the master branch has `add` support implemented from ~2020, but when switching to [release tags](https://github.com/umlaeute/v4l2loopback/tree/v0.12.7/utils) then the utils directory instead contains a very old prebuilt binary. Not sure what the deal here is, I will follow up with the maintainer. 

RE "We should probably remember this failed and just give up on v4l2loopback-ctl", ack, I'll adjust the logic to only try this once.

aramg:
Relevant issue: https://github.com/umlaeute/v4l2loopback/issues/514

aramg:
Added logic to check for v4l2loopback-ctl existence, and only try adding a device once. The current v4l2loopback-ctl exits with a '1' on `v4l2loopback-ctl -v`, so this change will be a noop until the new rewritten one from the main branch is released.

(also rebased master out of habit, sorry the incremental diff is messed)

plugins/linux-v4l2/v4l2-output.c
```diff
@@ -139,18 +160,21 @@ static bool try_connect(void *data, const char *device)
...
 	if (ioctl(vcam->device, VIDIOC_S_PARM, &parm) < 0)
 		goto fail_close_device;
 
+	uint32_t width = obs_output_get_width(vcam->output);
+	uint32_t height = obs_output_get_height(vcam->output);
+	vcam->frame_size = width * height * 2;
+
 	format.fmt.pix.width = width;
 	format.fmt.pix.height = height;
+	format.fmt.pix.field = V4L2_FIELD_NONE;
```
Lain-B, line 177: Is this line change related to what this PR is about? If it's not related it should probably be a separate PR. Or at least a separate commit.

Lain-B (changes requested):
@kkartaltepe will probably need to give another look at this again for actual functionality, distro compability, and testing, but in terms of actual code these are the things I spotted.

plugins/linux-v4l2/v4l2-output.c
```diff
@@ -230,8 +249,43 @@ static bool virtualcam_start(void *data)
...
+		}
+	}
+
+	bool success = loopback_card_open(data, OBS_V4L2_CARD_LABEL);
+
+	if (!success && v4l2loopback_ctl_avaialble) {
+		// Try to dynamically add a new device using the `v4l2loopback-ctl`
+		// utility, if available and if possible. Only try this once.
+
+		// TODO: Parse the output of add command and connect directly
```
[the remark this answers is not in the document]
aramg: Its a "nice to have", based on one of the previous reviews. 
From a technical point it would be nicer, but code wise its just extra LOC achieving the same functionality.

I will try to carve some time out to address the latest comments, soon.

aramg:
Compiling the new `v4l2loopback-ctl` utility is basic, just clone [v4l2loopback](https://github.com/umlaeute/v4l2loopback) and `make -C utils`. 
Again, unfortunate that for some reason that hasn't been released despite being present in the repo for a while now.

aramg:
There is also the option of pulling in the code v4l2loopback-ctl and replicating the 'add' command directly.  It's just an ioctl.
Not sure how we feel about that.

stephematician:
This might be worth a revisit soon, _Debian Bookworm_ has _v4l2loopback-utils_ at 0.12.6 which should have the `add` command. That means that _Ubuntu_ 22.04 might get it soon, too, although I note that it is still at 0.12.5 at the time of writing.

PR #8015, opened 2023-01-08
UI: Consistently reference the software H264 encoder properly

Conan-Kudo:
### Description

This change adjusts the encoder string to indicate that it's an H.264 encoder from x264.

### Motivation and Context

The code here assumes that the only software encoder is the x264-based H.264 encoder.
That may not always remain true.

### How Has This Been Tested?

Built and ran this on Fedora Linux 37 on x86_64.
Observed that strings loaded as they should for the x264 encoder from locale data.

### Types of changes

### Checklist:


gxalpha (changes requested):
Please only change the English translation file, the other files will be updated automatically.

Conan-Kudo:
@gxalpha Done.

notr1ch:
This feels somewhat unnecessary to me. It's very unlikely there will be a software H264 encoder competitive with x264 any time soon, all R&D has moved on to newer codecs.

Conan-Kudo:
Well, it's also a problem for introducing software VP9/AV1 codecs too.

jp9000:
Due to the way the locale system is designed, we may just have to live with the old ID. I realize it's a bit annoying, but it would just needlessly clear out all translations for that string and force translators to retranslate that string. Are you sure there's no other way to handle this situation that doesn't require changing the translation identifier?

tt2468:
> Due to the way the locale system is designed, we may just have to live with the old ID. I realize it's a bit annoying, but it would just needlessly clear out all translations for that string and force translators to retranslate that string. Are you sure there's no other way to handle this situation that doesn't require changing the translation identifier?

Aren't we able to rename translation string IDs in crowdin without wiping the values?

Conan-Kudo:
@jp9000 I intend to introduce changes to add more software encoders, particularly ones that I can ship in Fedora out of the box. Ideally, I'd like to eliminate the assumption the "one" software codec is x264 especially since it won't be in the near future. I'm considering implementing OpenH264 support since Fedora and openSUSE have arrangements with Cisco<sup>[1] [2]</sup> to support that. Additionally, I am looking to expose ffmpeg-aom as a software codec in the UI too.

In an ideal world, everything would be going through `obs-ffmpeg` and this would be easy. 😢 

[1]: https://fedoraproject.org/wiki/OpenH264
[2]: https://en.opensuse.org/OpenH264

PR #8017, opened 2023-01-08
libobs/util: Fix text-lookup not always case-insensitive

norihiro:
### Description

The text-lookup library is supposed to handle the key case-insensitive but the function `text_node_bychar` compared the key case-sensitive. This PR fixes the bug by converting the key to upper-case before creating the data structure.

When converting to upper-case, uses `toupper` so that the behavior is consistent with `astrcmpi_n`, which is utilized when looking-up the text.

Also removes the code below because both `val1` and `val2` have already been converted to upper-case.
https://github.com/obsproject/obs-studio/blob/783bd5219758f4eb031d9bfba19d92dd120ea7d8/libobs/util/text-lookup.c#L162-L165

The definition of `astrcmpi_n` is as below.
https://github.com/obsproject/obs-studio/blob/783bd5219758f4eb031d9bfba19d92dd120ea7d8/libobs/util/dstr.c#L122-L142

### Motivation and Context

I'm making another PR #8018 having translations below, which reveals this bug; some texts are not translated.
```
FilenameFormatting.TT.MM.month="Month as a decimal number (01-12)"
FilenameFormatting.TT.mm.minute="Minute (00-59)"
FilenameFormatting.TT.m.month="Month as a decimal number (01-12)"
FilenameFormatting.TT.M.minute="Minute (00-59)"
```

### How Has This Been Tested?

OS: Fedora 36.
Confirmed above translations are all translated.
Also briefly checked there are no wrong translations on the front UI and the settings dialog.

### Types of changes

### Checklist:

### DIFF
--- a/libobs/util/text-lookup.c
+++ b/libobs/util/text-lookup.c
@@ -14,6 +14,7 @@
  * OR IN CONNECTION WITH THE USE OR PERFORMANCE OF THIS SOFTWARE.
  */
 
+#include <ctype.h>
 #include "dstr.h"
 #include "text-lookup.h"
 #include "lexer.h"
@@ -158,11 +159,6 @@ static void lookup_addstring(const char *lookup_val, struct text_leaf *leaf,
 		for (len = 0; len < child->str.len; len++) {
 			char val1 = child->str.array[len],
 			     val2 = lookup_val[len];
-
-			if (val1 >= 'A' && val1 <= 'Z')
-				val1 += 0x20;
-			if (val2 >= 'A' && val2 <= 'Z')
-				val2 += 0x20;
 
 			if (val1 != val2)
 				break;
@@ -322,6 +318,9 @@ static void lookup_addfiledata(struct text_lookup *lookup,
 		leaf = bmalloc(sizeof(struct text_leaf));
 		leaf->lookup = bstrdup_n(name.array, name.len);
 		leaf->value = convert_string(value.array, value.len);
+
+		for (size_t i = 0; i < name.len; i++)
+			leaf->lookup[i] = toupper(leaf->lookup[i]);
 
 		lookup_addstring(leaf->lookup, leaf, lookup->top);
 


jp9000:
You know, I'm genuinely surprised I made the ID comparisons case insensitive. I feel like that was a silly decision on my part. Really no reason to have done that. Oh well, guess we're stuck with it that way, not a huge deal.

libobs/util/text-lookup.c
```diff
@@ -71,7 +72,7 @@ static struct text_node *text_node_bychar(struct text_node *node, char ch)
 
 	while (subnode) {
 		if (!dstr_is_empty(&subnode->str) &&
-		    subnode->str.array[0] == ch)
+		    toupper(subnode->str.array[0]) == toupper(ch))
```
jp9000, line 75: I'd recommend doing something like `ch = toupper(ch);` only once right after the `struct text_node *subnode = node->first_subnode;` line rather than needlessly calling it every time in the while loop. After that, `toupper(subnode->str.array[0]) == ch` should be fine here. Just a minor nitpick, but it's best to avoid needless extra calls to a function that just gets the same value over and over again for no reason when you could just do it once at the beginning of the function call.

Ideally it'd be nice if we stored all identifier strings in uppercase so we can minimize the toupper() calls overall, but at this point I'm just nitpicking at my own code.

norihiro: > I'd recommend doing something like ch = toupper(ch); only once right after the struct text_node *subnode = node->first_subnode; line rather than needlessly calling it every time in the while loop.
Ideally it'd be nice if we stored all identifier strings in uppercase so we can minimize the toupper() calls overall, but at this point I'm just nitpicking at my own code.

I didn't care about the time to call `toupper`.
As you suggested, I moved `toupper` to the beginning of creating the data structure to avoid calling `toupper` needlessly.

norihiro:
> You know, I'm genuinely surprised I made the ID comparisons case insensitive.

Maybe it could be able to switch to case-sensitive in the future major version. However, I'd like to fix it earlier.

PR #8018, opened 2023-01-08
UI: Separate translation of filename format tooltip

norihiro:
### Description

This PR depends other PRs so that marked as draft:
- https://github.com/obsproject/obs-studio/pull/8017
- https://github.com/obsproject/obs-studio/pull/8229

The tooltip for the filename format is written in HTML and each format has now each translation.

| Current tooltip | PR tooltip |
| -- | -- |
| format-tooltip-text | format-tooltip-html |

### Motivation and Context

This PR try to improve these two things.
- Makes the tooltip looks more beautiful. `%CCYY` is not aligned as above screenshot.
- it is a pain to translate such a long line. When adding a new format character, whole string has to be translated. Or, if a translator copied from the old string, it could miss the new format character without knowing.

### How Has This Been Tested?

OS: Fedora 36
Qt: 6.3.1-3.fc36.x86_64
DE: XFCE4
Also set an environment variable `QT_STYLE_OVERRIDE`=`Fusion`.

Tested with merging the PR #8229: https://github.com/norihiro/obs-studio/tree/ui-separate-format-tt-hashed-tables

### Types of changes

### Checklist:


UI/data/locale/en-US.ini
```diff
@@ -1057,7 +1057,34 @@ Screenshot.Source="Screenshot (Source)"
...
+FilenameFormatting.TT.CCYY="Year, four digits"
+FilenameFormatting.TT.YY.year="Year, last two digits (00-99)"
+FilenameFormatting.TT.MM.month="Month as a decimal number (01-12)"
+FilenameFormatting.TT.DD.day="Day of the month, zero-padded (01-31)"
+FilenameFormatting.TT.hh="Hour in 24h format (00-23)" # fail
+FilenameFormatting.TT.mm.minute="Minute (00-59)"
+FilenameFormatting.TT.ss="Second (00-61)"
+FilenameFormatting.TT.%="A % sign"
```
WizardCM, line 1067: For safety, I'd probably use the word "Percent" rather than "%" in the key.

WizardCM (changes requested):

UI/data/locale/en-US.ini
norihiro: I've revised it to `FilenameFormatting.TT.Percent`.
Also removed some comments ` # fail`, which was added while debugging the case-sensitivity bug and should not be left in the file.

WizardCM (approved):

yarons:
Hi, there's a bit of confusion among the translators, why is the maximum value for seconds is 61?

norihiro:
> why is the maximum value for seconds is 61?

Even considering the leap seconds, I think the maximum is 60.
`61` comes from the old combined text and I didn't touch it in this PR.

norihiro:
For translators, I should have noted that the translated documents of these articles would be helpful to translate the strings added by this PR.

- https://learn.microsoft.com/en-US/cpp/c-runtime-library/reference/strftime-wcsftime-strftime-l-wcsftime-l?view=msvc-170
- https://developer.apple.com/library/archive/documentation/System/Conceptual/ManPages_iPhoneOS/man3/strftime.3.html
- https://docs.python.org/3/library/datetime.html#strftime-strptime-behavior
- manpage of strftime

PR #8026, opened 2023-01-09
CI: Factorize Github labels checks

tytan652:
### Description

Factorize all 'Check for Github Labels' as one job so this step/job is done once.

It also fixes macOS and Ubuntu caches without dates when the CI is run on master.

NOTE: The use of `fromJSON()` is required to make GHA read the value as a boolean.

### Motivation and Context

I'm working on aarch64 Flatpak CI. And seeing the same step being repeated multiple times, I ask myself if it could be factorized.

> 1 is less than 4, so it's better

So my work on Flatpak aarch64 will rely on this CI change.

### How Has This Been Tested?

Tested on CI on my fork while I was working on Flatpak aarch64 with and without `Seeking Testers` label.

### Types of changes

- CI Bug fix (non-breaking change which fixes an issue)

- CI Tweak (non-breaking change to improve existing functionality)

### Checklist:


.github/workflows/main.yml
```diff
@@ -64,14 +64,35 @@ jobs:
...
+    runs-on: [ubuntu-22.04]
+    outputs:
+      create_artifacts: ${{ steps.config.outputs.create_artifacts }}
+      cache_date: ${{ steps.config.outputs.cache_date }}
+    steps:
+      - name: 'Configure Build Jobs'
+        id: config
+        run: |
+          if [[ "${{ github.event_name == 'pull_request' }}" == "true" ]]; then
+            if test -n "$(curl -H "Authorization: Bearer ${{ secrets.GITHUB_TOKEN }}" -s "${{ github.event.pull_request.url }}" | jq -e '.labels[] | select(.name == "Seeking Testers")')"; then
```
PatTheMav, line 78: Can be replaced by using `${{ contains(github.event.pull_request.labels.*.name, 'Seeking Testers') }}`, which should yield `true` if the PR has a label of that name.

PatTheMav (approved):
Had a similar idea and also switched to using `fromJSON` to parse "truthy" strings for conditions.

.github/workflows/main.yml
tytan652: Change done.

PR #8027, opened 2023-01-09
deps/media-playback: Enable CUDA HW decoder

rjgtav:
### Description

Enables FFmpeg's CUDA hardware decoder.

### Motivation and Context

To save CPU while playing media files on Linux environments with a NVIDIA GPU.

### How Has This Been Tested?

Tested on a n1-standard machine on Google Cloud Platform with a T4 NVIDIA GPU, on Debian 11, by looping a video of Big Buck Bunny while monitoring the GPU usage using nvidia-smi

### Types of changes

### Checklist:


tt2468:
CUDA hardware acceleration was known to cause crashes on Windows during initial testing a few years back. Please modify your commit to only enable CUDA for Linux until Windows has been adequately tested.

Fenrirthviti:
Well, if we need testing on Windows, I think it's fair to leave this enabled on Windows so we can actually test there. If we need to removed because we run in to similar issues, we can do so as part of the testing process.

RytoEX:
For reference to the exact commit where CUDA decoding was dropped:
https://github.com/obsproject/obs-studio/commit/86572935afda66442ea2a24c47d1dcca21185d97

tytan652:
CUDA should be after VDPAU and VAAPI.

On Linux VAAPI (any vendor) and VDPAU (Nvidia) shall be prioritized over CUDA.

tt2468:
> CUDA should be after VDPAU and VAAPI.
> 
> On Linux VAAPI (any vendor) and VDPAU (mainly Nvidia) shall be prioritized over CUDA.

I disagree. I have had significantly more issues with VAAPI and VDPAU on Linux than I have with CUDA. And by significantly more, I mean that I can crash the NVIDIA driver using VDPAU, and I've never experienced a single issue with CUDA on Ubuntu 20/22.04.

tytan652:
@tt2468 what do you think about VAAPI before CUDA and letting VDPAU after those two ?

PR #8028, opened 2023-01-09
obs-vst Added reporting of vendor name

crocodileGena:
### Description
Added support for plugins querying about Vendor Name to return "OBS Studio"

### Motivation and Context
It is a standard thing every VST2 host responds to.
Might help some plugins vendors.

### How Has This Been Tested?
It is a very safe change, only affects the return value when plugins ask for Vendor Name.

### Types of changes

### Checklist:

### DIFF
--- a/plugins/mac-virtualcam/src/dal-plugin/OBSDALMachClient.mm
+++ b/plugins/mac-virtualcam/src/dal-plugin/OBSDALMachClient.mm
@@ -93,6 +93,8 @@
 
 - (void)handlePortMessage:(NSPortMessage *)message
 {
+	__strong id<MachClientDelegate> strongDelegate = self.delegate;
+
 	VLogFunc(@"");
 	NSArray *components = message.components;
 	switch (message.msgid) {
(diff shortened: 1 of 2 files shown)


plugins/obs-vst/VSTPlugin.cpp
```diff
@@ -38,6 +38,10 @@ intptr_t VSTPlugin::hostCallback_static(AEffect *effect, int32_t opcode,
 	case audioMasterGetCurrentProcessLevel:
 		return 1;
 
+	case audioMasterGetVendorString:
+		strncpy((char*)ptr, "OBS Studio", 11);
```
WizardCM, line 42: Code format probably wants a space before the asterisk. Otherwise this looks good to me.

crocodileGena: thanks @WizardCM !
sorry about that, committed the fix and it looks ok now.

RytoEX (changes requested):
Please squash the commits and use the following commit message:
```
obs-vst: Add reporting of vendor name
```

jp9000:
How do we know that the allocation we're modifying has the required size that we're setting? I don't really like just modifying some random allocation we ourselves did not allocate, and that we don't know the size of. Is there a specification to this somewhere? I tried searching but didn't really come up with anything useful.

jp9000:
Huh, I think I found out. Seems the max string length should in theory be 64. Makes me a bit uncomfortable that the API was designed this way, but not much we can do about it I suppose. But if I'm reading things correctly this should be fine.

crocodileGena:
thanks @RytoEX . things got massed up locally for me, I will create new pull request with a single commit

Issue #8030, opened 2023-01-09
Expander produce "clicks" after attack event happens, after update from OBS v28 to v29


FelikZ:
### Operating System Info

macOS 13.1

### Other OS

M1 Pro

### OBS Studio Version

Other

### OBS Studio Version (Other)

29.0.0

### OBS Studio Log URL

https://obsproject.com/logs/Dei5QTHkC3co7Vhi

### Expected Behavior

Expander does not produce extra "clicks" in sound

### Current Behavior

Produces a "click" sound right after "Attack" happens

### Steps to Reproduce

1. Create similar expander config as attached, but configure threshold to "background noise level you have in the room"
2. Monitor your voice input (in sound settings you can choose device to monitor OBS output)
3. It should produce clicks

Sample config json and as screenshot:

```
            {
                "balance": 0.5,
                "deinterlace_field_order": 0,
                "deinterlace_mode": 0,
                "enabled": true,
                "flags": 0,
                "hotkeys": {},
                "id": "expander_filter",
                "mixers": 255,
                "monitoring_type": 0,
                "muted": false,
                "name": "Expander",
                "prev_ver": 469827585,
                "private_settings": {},
                "push-to-mute": false,
                "push-to-mute-delay": 0,
                "push-to-talk": false,
                "push-to-talk-delay": 0,
                "settings": {
                    "attack_time": 1,
                    "output_gain": 0.0,
                    "presets": "expander",
                    "ratio": 8.0,
                    "release_time": 400,
                    "threshold": -60.0
                },
                "sync": 0,
                "versioned_id": "expander_filter",
                "volume": 1.0
            },
```

<img width="854" alt="image" src="https://user-images.githubusercontent.com/1080269/211358960-dd299d21-8426-4370-b98d-d7f03c541ab4.png">

### Anything else we should know?

Used to work normally, until I have upgraded to v29 from v28.

If I disable expander, the issue disappears but I need expander for a reason :)

Also a sample audio file, where you can clearly hear "clicks".

v29
https://user-images.githubusercontent.com/1080269/211359822-479a6ce9-29a9-474a-a0e6-0098c8b5281f.mp4
v28.1.2
https://user-images.githubusercontent.com/1080269/211363567-eb5f860c-47af-4c7f-a730-49c416a713fa.mp4

pkviet:
Fixed here: https://github.com/obsproject/obs-studio/pull/8031

FelikZ:
[...]

PR #8031, opened 2023-01-09
obs-filters: Reapply audio processing above threshold for expander

pkviet:
### Description
Some code was applied to both expander and upward compressor. We limit it now to the compressor to keep the same behaviour as previous versions.
The said code was setting the gain to 0 dB above threshold.
Fixes issue #8030 .
### Motivation and Context
Try not to break existing behaviour.
Some users report clicks : https://github.com/obsproject/obs-studio/issues/8030

### How Has This Been Tested?
Works as v28.

### Types of changes

### Checklist:

### DIFF
--- a/plugins/obs-filters/expander-filter.c
+++ b/plugins/obs-filters/expander-filter.c
@@ -384,8 +384,8 @@ static inline void process_sample(size_t idx, float *samples, float *env_buf,
 			   : 0.0f;
 	gain = db_to_mul(fminf(min_val, gain_db[idx]));
 
-	// above threshold, don't process expander nor upward compressor
-	if (threshold - env_db <= 0)
+	// above threshold, don't process upward compressor
+	if (is_upwcomp && threshold - env_db <= 0)
 		gain = 1.0f;
 
 	samples[idx] *= gain * output_gain;


notr1ch:
Nit: The commit prefix should be "obs-filters"

RytoEX:
Was the original commit not cleanly revertable? The commit message initially struck me as "removing expander filter", so I'm trying to understand the scope of this change and the original commit to see if this commit message can be clarified.

pkviet:
Mmh the wording is probably unclear.
The previous commit applied a change to both expander and downward compressor since they share code.
The current commit makes the change only for the compressor.
I'll try to find some other wording.

FelikZ:
Yes, this seems to fix it.

v29-this-build
https://user-images.githubusercontent.com/1080269/212062471-4411eef6-1092-4026-8a27-e98e9dc924be.mp4

v28.1.2
https://user-images.githubusercontent.com/1080269/212062487-f5b7c449-7d55-454a-a711-86d646902c3b.mp4

Thanks!
Not sure if it breaks something though.

sandr1x:
I don't use expander, but a lot of guys on the forum are reporting the problem. Correction required.
https://obsproject.com/forum/threads/expander-causing-pops-and-clicks-after-upgrade-to-obs-29.163096/

pkviet:
Closed in favour of https://github.com/obsproject/obs-studio/pull/8121

PR #8041, opened 2023-01-11
UI: Move transition preview button to button box

exeldro:
### Description
Move transition preview button to button box
Before:
image
After:
image

### Motivation and Context
It is more clear that the button does not belong to the transition properties

### How Has This Been Tested?
On windows 64 bit by opening the properties of a stinger transition

### Types of changes

### Checklist:


WizardCM:
Overall I think this is a good idea, my only concern UX-wise is that it's right next to the Defaults button, which a user might accidentally click instead (though Cancel would undo it, so it's probably safe).

exeldro:
If people prefer this:
image
Only have to change the insert position from 1 to 2
The insert position should work as long as there is no help button added according to the order of buttons on:
https://doc.qt.io/qt-6/qdialogbuttonbox.html

jp9000:
Your proposed change sounds fairly reasonable exeldro. what do you think matt?

WizardCM:
I'd worry about the non-standard UX, it'd be a bit inconsistent with buttons on other dialogs. Overall though, I'm OK with either option.

GeorgesStavracas:
@Warchamp7 prefer the bottom-right approach

exeldro:
Changed to the bottom-right version:
image

Warchamp7 (approved):

Warchamp7:
I've rebased this PR

RytoEX (approved):
Seems fine.

frontend/dialogs/OBSBasicProperties.cpp
```diff
@@ -159,11 +159,12 @@ OBSBasicProperties::~OBSBasicProperties()
...
 
-void OBSBasicProperties::AddPreviewButton()
+void OBSBasicProperties::addTransitionPreviewButton()
 {
 	QPushButton *playButton = new QPushButton(QTStr("PreviewTransition"), this);
-	VScrollArea *area = view;
-	area->widget()->layout()->addWidget(playButton);
+
+	QBoxLayout *buttonBoxLayout = dynamic_cast<QBoxLayout *>(ui->buttonBox->layout());
+	buttonBoxLayout->insertWidget(2, playButton);
```
sebastian-s-beckmann:
```suggestion
	ui->buttonBox->addButton(playButton, QDialogButtonBox::ButtonRole::ActionRole);
```

Warchamp7: I would barely call this hacking the internals of `QDialogButtonBox`. It's a composite widget and has a layout. The cast is only needed to be able to insert at a specific index, which is specific to box layouts.

`addButton` doesn't work because roles are placed at predetermined positions on the right side.

image

Warchamp7: In any case, I added a new QFrame and placed the Defaults and Transition buttons in there instead.

PR #8048, opened 2023-01-12
updater: Deduplicate Update Downloads

derrod:
### Description

Only download files once.

### Motivation and Context

Skips unnecessary downloads and removes potential race condition when two threads attempt to write a download with the same hash at once.

### How Has This Been Tested?

Tested locally with update from 27.2.4 to 29.0.0 (i.e. without delta patches).

### Types of changes

### Checklist:

### DIFF
--- a/UI/win-update/updater/updater.cpp
+++ b/UI/win-update/updater/updater.cpp
@@ -24,6 +24,7 @@
 #include <vector>
 #include <string>
 #include <mutex>
+#include <unordered_set>
 
 using namespace std;
 using namespace json11;
@@ -246,6 +247,7 @@ enum state_t {
 	STATE_PENDING_DOWNLOAD,
 	STATE_DOWNLOADING,
 	STATE_DOWNLOADED,
+	STATE_ALREADY_DOWNLOADED,
 	STATE_INSTALL_FAILED,
 	STATE_INSTALLED,
 };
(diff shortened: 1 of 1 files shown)


notr1ch:
We may also need to subtract the size of the skipped files from `totalFileSize` to avoid a misleading progress indicator, though this is low risk since the duplicate files are usually tiny.

derrod:
I missed that, but easy enough to do.

UI/win-update/updater/updater.cpp
```diff
@@ -1564,6 +1565,32 @@ static bool Update(wchar_t *cmdLine)
 					   source.c_str(), size);
 	}
 
+	/* ------------------------------------- *
+	 * Deduplicate Downloads                 */
+
+	vector<BYTE *> hashes;
```
notr1ch, line 1571: Are we able to use something like `unordered_set` instead of `vector`? We don't care about the actual values, only existence, so it should provide constant-time performance instead of iterating over everything.

derrod: Updated it to simply use the `tempPath` with an unordered set since `std::hash` contains an implementation for `wstring` so no additional code or conversion necessary.

derrod:
I've now tested this locally, seems to work fine.

notr1ch (approved):

PR #8055, opened 2023-01-12
libobs/media-io: Check codec tag compatibility for out stream

### LINKED ISSUE #8029: Saved recording does not get remuxed successfully
### Operating System Info

Windows 11

### Other OS

_No response_

### OBS Studio Version

Other

29.0.0

### OBS Studio Log URL

https://obsproject.com/logs/iBjekFuN2emZQtXo

### OBS Studio Crash Log URL

_No response_

### Expected Behavior

I have enabled the setting `Advanced > Recording > Automatically remux to mp4`.
I do record in the `.ts` file format.

After saving a recording or the replay buffer, I expect OBS to create two files: A `Replay 2023-01-09 15-37-58.ts` and a short time after that a `Replay 2023-01-09 15-37-58.mp4`. 

### Current Behavior

The `.mp4` file is empty (`0 KB`). 
This bug has not been present in my previous version of OBS (latest release of `28.Y.Z` I believe), only since yesterday, since I upgraded to `29.0.0`.

### Steps to Reproduce

I am unsure since I do not know which settings need to be active (maybe it has to do something with the ts file format? Maybe with my encoder? I really don't know).

At least, this is what is needed:
1. Activate `Advanced > Recording > Automatically remux to mp4`
2. Save a recording or the replay buffer to disk
3. Empty `.mp4` file gets created

_No response_

tuduweb:
### Description

Check codec tag compatibility for out stream.

### Motivation and Context

Fixes #8029 

When the codec tag is not supported by out stream codec, the error happen.

introduced by https://github.com/obsproject/obs-studio/commit/bbb3dc00590a572cf49f97db3397221f0828f9ad

So add judgement to check compatibility.

https://github.com/FFmpeg/FFmpeg/blob/51efa68ec0b4f42b5b124b8987fb68f60a929c4f/libavformat/segment.c#L174-L180

### How Has This Been Tested?

Follow steps mentioned in issue to test, run well. (`ts, flv, m3u8 ...`)

~Not check hevc behavior.~

### Types of changes

### Checklist:

### DIFF
--- a/libobs/media-io/media-remux.c
+++ b/libobs/media-io/media-remux.c
@@ -149,14 +149,17 @@ static inline bool init_output(media_remux_job_t job, const char *out_filename)
 		av_dict_copy(&out_stream->metadata, in_stream->metadata, 0);
 
 #if LIBAVCODEC_VERSION_INT >= AV_VERSION_INT(57, 48, 101)
-		if (in_stream->codecpar->codec_tag != 0) {
-			out_stream->codecpar->codec_tag =
-				in_stream->codecpar->codec_tag;
-		} else if (in_stream->codecpar->codec_id == AV_CODEC_ID_HEVC) {
+		if (in_stream->codecpar->codec_id == AV_CODEC_ID_HEVC &&
+		    job->ofmt_ctx->oformat->codec_tag &&
+		    av_codec_get_id(job->ofmt_ctx->oformat->codec_tag,
+				    MKTAG('h', 'v', 'c', '1')) ==
+			    out_stream->codecpar->codec_id) {
 			// Tag HEVC files with industry standard HVC1 tag for wider device compatibility
+			// when HVC1 tag is supported by out stream codec
 			out_stream->codecpar->codec_tag =
 				MKTAG('h', 'v', 'c', '1');
 		} else {
+			// Otherwise tag 0 to let FFmpeg automatically select the appropriate tag
 			out_stream->codecpar->codec_tag = 0;
 		}
 #else


libobs/media-io/media-remux.c
```diff
@@ -149,13 +149,20 @@ static inline bool init_output(media_remux_job_t job, const char *out_filename)
...
 			// Tag HEVC files with industry standard HVC1 tag for wider device compatibility
 			out_stream->codecpar->codec_tag =
 				MKTAG('h', 'v', 'c', '1');
+		} else if (!job->ofmt_ctx->oformat->codec_tag ||
+			   av_codec_get_id(job->ofmt_ctx->oformat->codec_tag,
+					   in_stream->codecpar->codec_tag) ==
+				   out_stream->codecpar->codec_id ||
+			   av_codec_get_tag(job->ofmt_ctx->oformat->codec_tag,
+					    in_stream->codecpar->codec_id) <=
+				   0) {
```
jp9000, line 163: av_codec_get_tag returns an unsigned integer. 0 if no tag found, otherwise non-zero. Are you sure you meant to check for 0? or did you mean to check for non-zero?

tuduweb: I noticed this problem when I ported the code from `segment.c`, ffmpeg Repo.
I think it check for 0, not for non-zero.

should I repush this PR to modify this.

tuduweb:
Sorry @PatTheMav, I accidentally canceled the review request. :(

libobs/media-io/media-remux.c
PatTheMav: I'm currently trying to logically parse what's happening here, please correct me if I got this wrong - based on the change the following should happen:

* When the input stream has no codec tag but the codec id is HEVC, then set the codec_tag to `hvc1`.
* When the output format has no codec tag set, OR
  *  When the input stream has a codec tag, that codec tag has a supported codec id in the output format, and that codec id is equal to the codec id of the output stream, OR
  * When the output format has a supported codec tag given the codec id of the input stream, THEN
  * Set the codec tag of the output stream to the codec tag of the input stream
*  Otherwise set the codec tag to `0`

Shouldn't the compatibility check happen in every case then? Because right now we set the codec tag to `hvc1` regardless of whether the output format supports it (iirc `mkv` might not support it) and with your change only an input stream with an existing tag gets these checks.

PatTheMav: Also, maybe I'm missing something, but if an output format supports `hev1` but not `hvc1` (for whatever reason), wouldn't the third check yield `true`? You pass in `AV_CODEC_ID_HEVC` to `av_codec_get_tag` and might get `hev1`, wouldn't setting it to `hvc1` break the file then?

Also FFmpeg's code checks for "less then or equal to zero", which means it requires a mismatch ("no tags for codec id found in output format"), which confuses me a bit.

tuduweb: Thanks for your analysis, I found some flaws in this PR.

Back to the version before this PR:
Has Tag, directly equal assignment.
Not Has Tag, But `codec_id == HEVC`, assign `tag` to `hvc1`(for wider compatibility)
Else assign `tag` to `0`(ffmpeg will auto select in init_muxer)

Normal version(obs 28):
assign `codec` to `0` in any case.(ffmpeg will auto select)

https://github.com/FFmpeg/FFmpeg/blob/51efa68ec0b4f42b5b124b8987fb68f60a929c4f/libavformat/mux.c#L360-L381


So solution 1:
For wider compatibility:
1. Check OutputTags exist
2. CodecID equal HEVC
3. Check Codec support HVC1

then assign `out stream tag` to `HVC1`(for apple support? )

otherwise, set `0` (ffmpeg auto select)

Or solution 2:
Check and Select outstream tag as possible in OBS remux.

I think solution 1 is much better.
What do you think? Thx.
Look forward to your reply.

tuduweb: @PatTheMav 
Excuse me, is this code block for `respect input tags`?
https://github.com/obsproject/obs-studio/commit/bbb3dc00590a572cf49f97db3397221f0828f9ad#R152-R154
```c++
if (in_stream->codecpar->codec_tag != 0) {
			out_stream->codecpar->codec_tag =
				in_stream->codecpar->codec_tag;
		}
```
So need remain this code block and add compatibility check if I understand right?

PatTheMav: Yeah that one just naively retains the tag - it might have been there before my change even (haven't checked yet), but from the sound of it it could also break the output file if the selected output format does not support the tag.

Given that most people convert straight to `mp4` (and don't change the suffix in the remux window) it might not have occurred that often.

PatTheMav:
I take it you simplified the check to only apply `hvc1` if the codec is HEVC and the output format supports the tag?

libobs/media-io/media-remux.c
tuduweb: > Thanks for your analysis, I found some flaws in this PR.
> 
> Back to the version before this PR: Has Tag, directly equal assignment. Not Has Tag, But `codec_id == HEVC`, assign `tag` to `hvc1`(for wider compatibility) Else assign `tag` to `0`(ffmpeg will auto select in init_muxer)
> 
> Normal version(obs 28): assign `codec` to `0` in any case.(ffmpeg will auto select) _avformat_write_header -> init_output.. -> init_muxer_ https://github.com/FFmpeg/FFmpeg/blob/51efa68ec0b4f42b5b124b8987fb68f60a929c4f/libavformat/mux.c#L360-L381
> 
> So solution 1: For wider compatibility:
> 
> 1. Check OutputTags exist
> 2. CodecID equal HEVC
> 3. Check Codec support HVC1
> 
> then assign `out stream tag` to `HVC1`(for apple support? )
> 
> otherwise, set `0` (ffmpeg auto select)
> 
> Or solution 2: Check and Select outstream tag as possible in OBS remux.
> 
> I think solution 1 is much better. What do you think? Thx. Look forward to your reply.

@PatTheMav If you perfer to `solution 2 : Check and Select outstream tag as possible in OBS remux.`, I will update soon. Thanks for your guidance in advance.

PR #8066, opened 2023-01-15
Add support for multiple encoders per output

kc5nra:
### Description
This adds the internal plumbing to allow outputs multiple video encoders/tracks in addition to existing ability to add multiple audio tracks.

### Motivation and Context
This will enable future output plugins to support sending video in multiple resolutions, codecs, and configurations.  Many protocols benefit from having multiple encoded outputs:
* Variant programs in mpegts
* WebRTC simulcast
* MoQ/QUIC-based
* Emerging protocols/codecs that require communicating with a diverse set of connection profiles

This also aligns and simplifies a lot of the code that handled multiple audio tracks but hardcoded support for only one encoder in the output code.

### How Has This Been Tested?
Tested with example implementations found here ____

### Types of changes

### Checklist:


libobs/obs-output.c
```diff
@@ -798,17 +819,19 @@ void obs_output_remove_encoder(struct obs_output *output,
...
+		if (output->audio_encoders[i] == encoder)
+			output->audio_encoders[i] = NULL;
 	}
 }
 
-void obs_output_set_video_encoder(obs_output_t *output, obs_encoder_t *encoder)
+void obs_output_set_video_encoder2(obs_output_t *output, obs_encoder_t *encoder,
+				   size_t idx)
 {
 	if (!obs_output_valid(output, "obs_output_set_video_encoder"))
```
tt2468, line 836: This should be renamed to `obs_output_set_video_encoder2`. It's only for logging, but other functions have their "v2" functions with their own valid checks too.

libobs/obs-output.c
```diff
@@ -875,11 +916,28 @@ void obs_output_set_audio_encoder(obs_output_t *output, obs_encoder_t *encoder,
 	output->audio_encoders[idx] = encoder;
 }
 
+obs_encoder_t *obs_output_get_video_encoder2(const obs_output_t *output,
+					     size_t idx)
+{
+	if (!obs_output_valid(output, "obs_output_get_video_encoder"))
```
tt2468, line 922: requires rename too

libobs/obs-output.c
```diff
@@ -825,18 +848,36 @@ void obs_output_set_video_encoder(obs_output_t *output, obs_encoder_t *encoder)
...
+	if (idx == 0) {
+		/* set the preferred resolution on the encoder */
+		if (output->scaled_width && output->scaled_height)
+			obs_encoder_set_scaled_size(output->video_encoders[idx],
+						    output->scaled_width,
+						    output->scaled_height);
+	}
+}
+
+void obs_output_set_video_encoder(obs_output_t *output, obs_encoder_t *encoder)
```
tt2468, line 878: missing validity check

libobs/obs-output.c
```diff
@@ -875,11 +916,28 @@ void obs_output_set_audio_encoder(obs_output_t *output, obs_encoder_t *encoder,
...
+	} else {
+		if (idx > 0) {
+			return NULL;
+		}
+	}
+
+	return output->video_encoders[idx];
+}
+
 obs_encoder_t *obs_output_get_video_encoder(const obs_output_t *output)
```
tt2468, line 938: Missing `obs_output_valid()`

libobs/obs-output.c
```diff
@@ -875,11 +916,28 @@ void obs_output_set_audio_encoder(obs_output_t *output, obs_encoder_t *encoder,
 	output->audio_encoders[idx] = encoder;
 }
 
+obs_encoder_t *obs_output_get_video_encoder2(const obs_output_t *output,
+					     size_t idx)
+{
+	if (!obs_output_valid(output, "obs_output_get_video_encoder"))
+		return NULL;
+
+	if ((output->info.flags & OBS_OUTPUT_MULTI_TRACK_VIDEO) != 0) {
```
tt2468, line 925: Technically this logic can be simplified only to a single `if (idx >= MAX_VIDEO_ENCODERS)` because the array always exists, and the items in `idx > 0` are NULL anyway if the output is single track video, right?

libobs/media-io/video-io.h
```diff
@@ -24,6 +24,8 @@
 extern "C" {
 #endif
 
+#define MAX_VIDEO_ENCODERS 4
```
kc5nra, line 27: Just a reasonable sounding number

libobs/obs-output.c
```diff
@@ -978,46 +1038,69 @@ void obs_output_set_preferred_size(obs_output_t *output, uint32_t width,
 		return;
 	}
 
-	output->scaled_width = width;
-	output->scaled_height = height;
+	// Only really matters for non-encoded
```
tt2468, line 1041: I would move this to the `if` statement below, only changing it if the output is raw, assuming these parameters only affect raw outputs.

libobs/media-io/video-io.h
tt2468: I think MAX_VIDEO_STREAMS or MAX_VIDEO_TRACKS might be a better name. Number seems fine to me

libobs/obs-output.c
```diff
@@ -1151,12 +1278,17 @@ get_audio_conversion(struct obs_output *output)
 	return output->audio_conversion_set ? &output->audio_conversion : NULL;
 }
 
-static size_t get_track_index(const struct obs_output *output,
-			      struct encoder_packet *pkt)
+static size_t get_encoder_index(struct obs_output *output,
+				struct encoder_packet *pkt)
 {
+	// Mutiple audio tracks are more common
```
tt2468, line 1284: Is it possible to use `pkt->type` to not have to iterate multiple times depending on packet type?

libobs/obs-output.c
```diff
@@ -1592,7 +1721,7 @@ static bool initialize_interleaved_packets(struct obs_output *output)
 
 	/* ensure that there is audio past the first video packet */
```
tt2468, line 1722: This should probably detect the first video packet on *any* video encoder, to keep with the original intention of only sending audio after video has started.

Additionally, the current implementation assumes that there is a video encoder at idx 0, which it may be possible that there isn't one? This bug would also apply to the multiple audio track implementation if true.

libobs/obs-output.c
```diff
@@ -1625,7 +1755,7 @@ static bool initialize_interleaved_packets(struct obs_output *output)
 
 	/* subtract offsets from highest TS offset variables */
 	output->highest_audio_ts -= audio[0]->dts_usec;
```
tt2468, line 1757: Possibly unsafe assumption that there are encoders at idx 0

libobs/obs.h
```diff
@@ -2204,6 +2258,13 @@ EXPORT void
 obs_output_set_audio_conversion(obs_output_t *output,
 				const struct audio_convert_info *conversion);
 
+/** Returns the number of video encoders associated with this output */
+EXPORT size_t
+obs_output_get_video_encoder_count(const struct obs_output *output);
```
tt2468, line 2263: This provides a count, but does not provide any info on the exact indexes of how tracks are routed. For example, if you have a layout like `- + + -`, you have two audio encoders, but the output has no clue where they are. There is `obs_output_get_mixers()` which is a badly named function, but provides a bitmask specifying exactly what mixers are routed.

I need to understand this more personally, because I'm pretty sure that function is used for raw output only, but for some reason there are multiple encoded outputs that are using it. This may actually cause audio bugs if say you route track 0 to idx 1, and idx 1 to track 0.

libobs/obs-output.c
kc5nra: Ya, just missed that. Will fix

kc5nra:
Closing in favor of chained PRs

PR #8070, opened 2023-01-16
updater: Pass AppData path to elevated process

derrod:
### Description

Pass AppData location to elevated process via command line parameter.

(This is classed as a bug fix as the previous method doesn't work [anymore] and this replaces it.)

### Motivation and Context

When a separate administrator account is used for elevation the appdata path would be set to the wrong user's directory.

Previously this was attempted via an environment variable, but those are not retained for the elevated process.

### How Has This Been Tested?

Verified correct path is passed and read by the elevated process.

### Types of changes

### Checklist:

### DIFF
--- a/UI/win-update/updater/updater.cpp
+++ b/UI/win-update/updater/updater.cpp
@@ -1352,6 +1352,7 @@ static bool Update(wchar_t *cmdLine)
 
 	bool bIsPortable = false;
 	wstring branch = L"stable";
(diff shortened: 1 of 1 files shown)


RytoEX (changes requested):
Nits on a comment. Otherwise seems okay at a glance. Would like @notr1ch 's approval before merging.

notr1ch (approved):
C

PR #8071, opened 2023-01-16
libobs, UI: Collection of encoder/output related fixes

tt2468:
### Description
Various issues I noticed while going down a rabbit hole implementing some obs_view_t stuff. See commit descriptions for individual details.

### Motivation and Context
I dislike when things do not work correctly.

### How Has This Been Tested?
- Start streaming and recording in different combinations
- Modify code to run tests on added error handling (encoder funcs, remaining view logging)

### Types of changes

### Checklist:

### DIFF
--- a/UI/window-basic-main-outputs.cpp
+++ b/UI/window-basic-main-outputs.cpp
@@ -1007,7 +1007,7 @@ bool SimpleOutput::StartStreaming(obs_service_t *service)
 	else
 		lastError = string();
 
-	const char *type = obs_service_get_output_type(service);
+	const char *type = obs_output_get_id(streamOutput);
 	blog(LOG_WARNING, "Stream output type '%s' failed to start!%s%s", type,
 	     hasLastError ? "  Last Error: " : "", hasLastError ? error : "");
 	return false;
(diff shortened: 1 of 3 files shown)


UI/window-basic-main-outputs.cpp
```diff
@@ -1482,7 +1482,6 @@ inline void AdvancedOutput::SetupStreaming()
 
 	obs_output_set_audio_encoder(streamOutput, streamAudioEnc, 0);
 	obs_encoder_set_scaled_size(videoStreaming, cx, cy);
-	obs_encoder_set_video(videoStreaming, obs_get_video());
```
kc5nra, line 1485: Nit, but probably should be a separate commit

libobs/obs.c
```diff
@@ -774,13 +774,16 @@ void obs_free_video_mix(struct obs_core_video_mix *video)
...
-		obs_free_video_mix(obs->video.mixes.array[i]);
+	size_t num_views = 0;
+	for (size_t i = 0; i < obs->video.mixes.num; i++) {
+		struct obs_core_video_mix *video = obs->video.mixes.array[i];
+		if (video && video->view)
+			num_views++;
+		obs_free_video_mix(video);
 		obs->video.mixes.array[i] = NULL;
 	}
+	if (num_views)
```
kc5nra, line 785: Probably clearer to not cast to boolean and just do > 0. I think there is some UB or truth shenanigans than can occur since size_t is > int (which is what types are cast to for boolean coercion for very large numbers, so not really something we'd run into but.. ya)

PR #8074, opened 2023-01-16
updater: Use Zstandard for delta update and file compression

derrod:
### Description

Adds support for delta patches created with Zstandard rather than bsdiff.

ToDo:
- [x] Add support to [bouf](https://github.com/obsproject/bouf)
   + Supported added in [v0.4.0](https://github.com/obsproject/bouf/releases/tag/v0.4.0)
- [x] Add Zstandard to [obs-deps](https://github.com/obsproject/obs-deps)
   + PR opened: https://github.com/obsproject/obs-deps/pull/150

### Motivation and Context

- Faster patch generation (nearly 2x at level 22, 4x at 19 which offers comparable size to bsdiff)
- Smaller patches (~10 MiB less for 27.2.4 => 29.0)
- Lower memory usage on client (TBC)

### How Has This Been Tested?

In-dev bouf branch and local update server setup.

### Types of changes

### Checklist:


derrod:
With the last two comments I switched from using zlib to using zstandard for compressing the JSON body of the patch manifest request and removed zlib entirely.

The manual gzip decompression has been removed as WinHTTP supports doing this for you as of Windows 8.1 (via `WINHTTP_OPTION_DECOMPRESSION`).

derrod:
Note for CMake stuff: zstd outputs `zstdConfig.cmake` and `zstdTargets.cmake` which become part of obs-deps once that PR is merged. When I was building locally CMake was able to automatically pick those up and building with zstd simply required adding the `find_package(zstd)` and the static target to `CMakeLists.txt`

derrod:
Here's a branch that additionally removes LZMA as well (both from `deps/` and the updater): https://github.com/derrod/obs-studio/compare/updater-zstd..updater-remove-lzma

If we ever want to feel better about the number of lines of code in our repo that's an easy -27,000 :P

notr1ch:
What do you think about sending `Accept-Encoding: zstd` and decompression of `Content-Encoding: zstd` for non-patch files? Basically we'd be implementing our own decompression again like we previously had for gzip, but using zstd which should achieve speed and compression wins.

derrod:
I've thought about using compression for non-patch files, but I didn't consider of doing it as part of the download itself. That seems fine though.

I assume we'd want to do ahead-of-time compression there as well. I'd just need to know what that would mean for bouf as far as creating the directory structure goes. For example could I just output both `file.ext` and `file.ext.zst` into the directory and nginx could automatically rewrite the download to the zst file if the client advertises support?

derrod:
Now updated with support for compressed non-patch files. Support for building compatible update files and manifests was added in [bouf v0.4.1](https://github.com/obsproject/bouf/releases/tag/v0.4.1).

Using compressed update files brings the download size for OBS 29.0 down from ~352 MiB to ~114 MiB. Zstandard delta patches in my test case (27.2.4 => 29.0.0) would come in at around ~65 MiB - around 8 MiB less than the old bsdiff + lzma based deltas.

The implementation here assumes files are compressed if a `compressed_hash` field is present in the manifest. That hash is used to verify the downloaded file before decompression, which happens in the download worker thread.

derrod:
Comparison of speed of this PR with #8115 (#8106 merged in either case):
This is running with patches served from my NAS so there's some lower latency of course, but it did shave off another 3,5-ish seconds off, bringing us to below 5 seconds for 28.1.2 => 29.0.0!
Updater speed comparison_ ST, MT, MT+ZSTD

derrod:
Conflicted due to #8115 now being merged, will wait #8142/#8141 to be merged and obs-deps update with zstd to be released before rebasing and fixing it.

RytoEX (approved):
@notr1ch Tentatively approved off-thread. We'll have to test thoroughly during the next beta period.

PR #8076, opened 2023-01-17
libobs: Refactor obs-output in preparation of multiple video encoders

kc5nra:
### Description
Refactors the obs-output to better separate the usage of MAX_AUDIO_MIXES (raw audio) and the audio encoders.

### Motivation and Context
This is in preparation for adding multiple video encoders per output as well as generally clarifying usage and allowing potentially more encoders than audio mix tracks (6).

Also fixes some bugs around removing encoders from outputs while active

### How Has This Been Tested?
* Tested multi track consecutive and sparse tracks (ffmpeg muxer)
* Tested multi track consecutive and sparse tracks (ffmpeg adv recording)
### Types of changes

### Checklist:

### DIFF
--- a/libobs/obs-output.c
+++ b/libobs/obs-output.c
@@ -792,10 +792,10 @@ size_t obs_output_get_mixers(const obs_output_t *output)
 		       : 0;
 }
 
-void obs_output_remove_encoder(struct obs_output *output,
-			       struct obs_encoder *encoder)
+void obs_output_remove_encoder_internal(struct obs_output *output,
+					struct obs_encoder *encoder)
 {
-	if (!obs_output_valid(output, "obs_output_remove_encoder"))
+	if (!obs_output_valid(output, "obs_output_remove_encoder_internal"))
 		return;
 
 	if (output->video_encoder == encoder) {
@@ -806,6 +806,18 @@ void obs_output_remove_encoder(struct obs_output *output,
 				output->audio_encoders[i] = NULL;
 		}
 	}
+}
+
+void obs_output_remove_encoder(struct obs_output *output,
+			       struct obs_encoder *encoder)
+{
+	if (!obs_output_valid(output, "obs_output_remove_encoder"))
+		return;
+	if (active(output)) {
+		return;
+	}
+
+	obs_output_remove_encoder_internal(output, encoder);
 }
 
 void obs_output_set_video_encoder(obs_output_t *output, obs_encoder_t *encoder)
--- a/libobs/obs-output.c
+++ b/libobs/obs-output.c
@@ -180,7 +180,7 @@ static inline void free_packets(struct obs_output *output)
 	da_free(output->interleaved_packets);
 }
 
-static inline void clear_audio_buffers(obs_output_t *output)
+static inline void clear_raw_audio_buffers(obs_output_t *output)
 {
 	for (size_t i = 0; i < MAX_AUDIO_MIXES; i++) {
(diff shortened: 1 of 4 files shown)


libobs/obs-output.c
```diff
@@ -888,14 +901,8 @@ obs_encoder_t *obs_output_get_audio_encoder(const obs_output_t *output,
...
-		if (idx >= MAX_AUDIO_MIXES) {
-			return NULL;
-		}
-	} else {
-		if (idx > 0) {
-			return NULL;
-		}
+	if (idx >= MAX_OUTPUT_AUDIO_ENCODERS) {
+		return NULL;
 	}
```
tuduweb, line 906: maybe shouldn't use braces?

kc5nra: Fixed in 6333146988f2f6f2b5558bd2e129b691f9b5cb85

libobs/obs-output.c
```diff
@@ -792,22 +792,35 @@ size_t obs_output_get_mixers(const obs_output_t *output)
 		       : 0;
 }
 
-void obs_output_remove_encoder(struct obs_output *output,
-			       struct obs_encoder *encoder)
+void obs_output_remove_encoder_unchecked(struct obs_output *output,
```
tt2468, line 795: The going procedure for un-exported functions like this is `obs_output_remove_encoder_internal`

libobs/obs-output.c
```diff
@@ -1042,42 +1048,18 @@ void obs_output_set_audio_conversion(
...
-		for (size_t i = 0; i < mix_count; i++) {
-			if (!output->audio_encoders[i]) {
-				return false;
+		for (size_t i = 0; i < MAX_OUTPUT_AUDIO_ENCODERS; i++) {
+			if (output->audio_encoders[i]) {
+				return true;
 			}
 		}
+		return false;
 	} else {
```
tt2468, line 1060: This doesn't even need to be an else, it can just return since the `if` statement will always return.

libobs/obs-output.c
```diff
@@ -1907,8 +1922,9 @@ static void reset_packet_data(obs_output_t *output)
 	output->highest_video_ts = 0;
 	output->video_offset = 0;
 
-	for (size_t i = 0; i < MAX_AUDIO_MIXES; i++)
+	for (size_t i = 0; i < MAX_OUTPUT_AUDIO_ENCODERS; i++) {
```
tt2468, line 1925: This is adding unnecessary brackets, though I forget what the project consensus is on brackets and single statement for loops.

kc5nra: Yeah, I saw both... and it _seemed_ like they added braces around fors more often than not?

kc5nra: But happy to do whatever

libobs/obs-output.c
kc5nra: k, I'll fix that up

libobs/obs-output.c
kc5nra: Fixed in 69da75d32b061ff6a177f4177cbecbc06a024948

libobs/obs-output.c
kc5nra: Fixed in 537ed6df1a161ee9322a6106bcc83385fc20a5d5

libobs/obs-output.c
kc5nra: Fixed in 6f2f1567e1e68e1b054a6e5e649fa6a32e64dc7f

PR #8079, opened 2023-01-17
UI: Don't load global plugins in portable mode; Minor portable mode fixes

gxalpha:
### Description

Consists of three small commits that all are related / kinda depend on each other. I'm all for atomic commits but don't think separate PR's are needed here. Feel free to disagree with me and let me know.

1. Nearly everything that involves portable mode is ifdef'ed, just not actually setting the variable. This would mean that if the `--portable` argument was passed on unsupported systems, the log would show portable mode as enabled even though it's effectively no-op. This can lead to confusion and shouldn't be the case.
2.  Portable mode is meant to be separated from the rest of the system, and as such it doesn't make sense to load globally installed plugins there. On Windows, there currently are only two major plugins that I'm aware of (StreamDeck and StreamFX) that install themselves there. Plugins installing themselves there is good, but it currently also means that even portable instances load them which can make testing and debugging in a clean environment annoying.

### Motivation and Context

Read about the annoyance of global plugins loading in portable mode multiple times over the last few days.
Noticed the other things while working on that.

### How Has This Been Tested?

1 has been tested on my Mac (macOS 13.1).

2 was tested by Flaeri, who confirmed that the streamdeck plugin no longer loads.

### Types of changes

### Checklist:

### DIFF
--- a/UI/window-basic-main.cpp
+++ b/UI/window-basic-main.cpp
@@ -173,6 +173,9 @@ static void AddExtraModulePaths()
 				    data_path_with_module_suffix.c_str());
 	}
 
+	if (portable_mode)
+		return;
+
 	char base_module_dir[512];
 #if defined(_WIN32)
 	int ret = GetProgramDataPath(base_module_dir, sizeof(base_module_dir),
--- a/UI/obs-app.cpp
+++ b/UI/obs-app.cpp
@@ -3220,12 +3220,14 @@ int main(int argc, char *argv[])
 	obs_set_cmdline_args(argc, argv);
 
 	for (int i = 1; i < argc; i++) {
-		if (arg_is(argv[i], "--portable", "-p")) {
-			portable_mode = true;
-
-		} else if (arg_is(argv[i], "--multi", "-m")) {
+		if (arg_is(argv[i], "--multi", "-m")) {
 			multi = true;
 
+#if ALLOW_PORTABLE_MODE
+		} else if (arg_is(argv[i], "--portable", "-p")) {
+			portable_mode = true;
+
+#endif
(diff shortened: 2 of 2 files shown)


flaeri:
Tested on windows 10 22H2. I have the streamdeck plugin installed in `%programdata%` as is per default. It would load on 29 release, but does not load with the code in this pr. I have not tested extensively, but the portable part seems to work as intended, and with a cursory glance, nothing has exploded, yet.

[2023-01-17 18-22-11.txt](https://github.com/obsproject/obs-studio/files/10438259/2023-01-17.18-22-11.txt)

derrod:
I would quite like at least 79bd265223db10c550a9267994ed426d4dcfe1b6 and 841a1e28f6fb5131b36639d8952faa8807ff217b to be picked to the release branch when merged.

gxalpha:
Dropped the first commit as per off-thread discussion on Discord.

UI/window-basic-main.cpp
```diff
@@ -173,6 +173,9 @@ static void AddExtraModulePaths()
 				    data_path_with_module_suffix.c_str());
 	}
 
+	if (portable_mode)
+		return;
+
```
notr1ch, line 178: `portable_mode` has external visibility, so there's no need to pass it to this function, it can be tested directly.

gxalpha: you're right, fixed.

derrod (approved):
Seems fine to me. I think not loading global plugins in portable mode is reasonable.

Technically might break some setups of people using portable mode with Stream Deck, but I think it's worth having an actually self-contained portable mode and if those people complain too much we can always reconsider it.

gxalpha:
R1ch had no objections off-thread.

PR #8082, opened 2023-01-18
UI: Add space for right arrow in menu

exeldro:
### Description
Add space for right arrow in menu

### Motivation and Context
When making custom menu I noticed that the right arrow was on top of the text.
It looks to me that it is caused by setting the padding for QMenu::item to 6px in the theme.
Now I override only the right padding for QMenu::item to 12px
Before:
image
After:
image

The dark and system theme don't set the padding and don't have the issue.

### How Has This Been Tested?
On windows 64 by having only 1 menu item that has a right arrow.

### Types of changes

### Checklist:

### DIFF
--- a/UI/data/themes/Acri.qss
+++ b/UI/data/themes/Acri.qss
@@ -201,6 +201,10 @@ SceneTree::item {
     padding: 6px;
 }
 
+QMenu::item {
+    padding-right: 20px;
+}
+
 QListWidget::item,
 SourceTreeItem,
 QMenu::item,
--- a/UI/data/themes/Grey.qss
+++ b/UI/data/themes/Grey.qss
@@ -201,6 +201,10 @@ SceneTree::item {
     padding: 6px;
 }
 
+QMenu::item {
+    padding-right: 20px;
+}
+
 QListWidget::item,
 SourceTreeItem,
 QMenu::item,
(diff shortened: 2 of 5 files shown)


GeorgesStavracas:
Does this behave properly with right-to-left locales?

exeldro:
> Does this behave properly with right-to-left locales?
No idea, what would be a good way to test that?

GeorgesStavracas:
I don't really know, sorry. On Linux, usually launching the application with an RTL locale (e.g. `LC_ALL=he-IL.utf8 obs`) would be enough, but I'm not sure if Qt will respect that.

RytoEX:
> Does this behave properly with right-to-left locales?

I came here to ask the same question. Ideally, I'd like to know how this behaves in an RTL language/locale context.

exeldro:
Tested it by adding `program.setLayoutDirection(Qt::RightToLeft);` to `obs-app.cpp`
Al seems to work correct:
image

WizardCM:
It's worth noting this also occurs on the system tray context menu on Windows. Reportedly this is because none of the QMenu items have an icon/checkbox.

image

I've tested this PR on Windows and verified it improves the behaviour.

exeldro:
Made it 20px
image
image

GeorgesStavracas (approved):
LGTM from the screenshots

RytoEX (approved):
Seems better than the current situation.

Looks good to me.

PR #8083, opened 2023-01-18
obs-ffmpeg: Add option to disable settings logging

CodeYan01:
Allows plugins to disable the logging of settings every time the ffmpeg source is updated, by passing "log_changes" as true in the settings.

### Description

Adds "log_changes" to the ffmpeg source settings. If false, it will not log changes every time the settings of the media source is updated. This does not affect logging of errors and warnings. I did not add this to the properties window because I assume it's used in support channels to check the settings of media sources, so it would be bad if users can set this directly.

### Motivation and Context

I am writing a playlist plugin that uses media sources internally, just as the Image Slideshow source uses Image sources internally. However, Media source spams the log when the settings are updated, which is not ideal for this sort of plugin. Image Slideshow does not have this problem because Image sources do not log their settings.

### How Has This Been Tested?

I made my plugin pass "log_changes" as false to see that it disables logging when the settings of the internal media source are modified. I also verified that Media Sources still log as usual, as users can't set the property.

### Types of changes

### Checklist:

### DIFF
--- a/plugins/obs-ffmpeg/obs-ffmpeg-source.c
+++ b/plugins/obs-ffmpeg/obs-ffmpeg-source.c
@@ -53,6 +53,7 @@ struct ffmpeg_source {
 	bool close_when_inactive;
 	bool seekable;
 	bool is_stinger;
+	bool log_changes;
 
 	pthread_t reconnect_thread;
 	pthread_mutex_t reconnect_mutex;
@@ -127,6 +128,7 @@ static void ffmpeg_source_defaults(obs_data_t *settings)
 	obs_data_set_default_int(settings, "reconnect_delay_sec", 10);
 	obs_data_set_default_int(settings, "buffering_mb", 2);
 	obs_data_set_default_int(settings, "speed_percent", 100);
+	obs_data_set_default_bool(settings, "log_changes", true);
 }
 
 static const char *media_filter =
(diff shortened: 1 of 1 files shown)


plugins/obs-ffmpeg/obs-ffmpeg-source.c
```diff
@@ -233,6 +235,11 @@ static obs_properties_t *ffmpeg_source_getproperties(void *data)
 static void dump_source_info(struct ffmpeg_source *s, const char *input,
 			     const char *input_format)
 {
+	if (!s->log_changes) {
+		UNUSED_PARAMETER(input);
+		UNUSED_PARAMETER(input_format);
+		return;
+	}
```
norihiro, line 242:
```suggestion
	if (!s->log_changes)
		return;
```
`UNUSED_PARAMETER` is a macro to suppress a warning that the parameter is not used in the function. We don't need to have this in the `if` block.

CodeYan01: Resolved, thank you!

derrod:
There's a merge conflict, please resolve it. Generally this seems fine.

Lain-B:
Although this is a very niche feature, because it affects your plugin's users, and because it may be useful for us in the future, I'll merge this.

CodeYan01:
Thank you very much, and I apologize for not seeing the merge conflict earlier

PR #8084, opened 2023-01-18
vlc-video: Refactor playback behavior

CodeYan01:
### Description

Used the enum behavior directly, converting the list from using strings internally to just ints. Added check if behavior is greater than `BEHAVIOR_ALWAYS_PLAY` (last enum item) to set it back to `BEHAVIOR_ALWAYS_PLAY`, to keep previous behavior (even though this case is only achievable if user edits the json file directly, or a script/plugin sets behavior to an invalid value.

### Motivation and Context

Removes unnecessary code and string comparison, although performance difference should be negligible.

### How Has This Been Tested?

Verified that changing the behavior works.

### Types of changes

Probably a breaking change only for plugins that set the behavior. If this change is not worth it, it's fine.

### Checklist:

### DIFF
--- a/plugins/vlc-video/vlc-video-source.c
+++ b/plugins/vlc-video/vlc-video-source.c
@@ -16,9 +16,6 @@
 #define S_LOOP                         "loop"
 #define S_SHUFFLE                      "shuffle"
 #define S_BEHAVIOR                     "playback_behavior"
-#define S_BEHAVIOR_STOP_RESTART        "stop_restart"
-#define S_BEHAVIOR_PAUSE_UNPAUSE       "pause_unpause"
-#define S_BEHAVIOR_ALWAYS_PLAY         "always_play"
 #define S_NETWORK_CACHING              "network_caching"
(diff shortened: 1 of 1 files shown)


CodeYan01:
On second thought, this would also slightly break saved settings. Behavior of old sources would probably be changed to the first enum value (because get_int would probably return 0), so this might indeed not be a worthwhile change.

PatTheMav:
What you _could_ do is add some code that "upgrades" legacy settings to the new (simplified) settings - so when a user launches OBS after this merge, their old configuration is interpreted correctly, but the value actually stored in the settings object (and then written to the configuration file) is already the new value.

This has the benefit that code _handling_ that value can use the simplified handling and you only need to add some upgrade code path for the plugin creation.

See as an example:

https://github.com/obsproject/obs-studio/blob/8fc0489e4ad33e57ae04b8ea80fe18cdf22c178b/plugins/mac-capture/window-utils.m#L249-L280

Warchamp7:

PR #8089, opened 2023-01-19
RFC 45-4: Use protocol to select output type for streaming

tytan652:
### Description

**This PR is about the [RFC 45](https://github.com/obsproject/rfcs/pull/45).**

Depends on:
- https://github.com/obsproject/obs-studio/pull/5104
- https://github.com/obsproject/obs-studio/pull/8087
- https://github.com/obsproject/obs-studio/pull/8088

Make OBS Studio rely on the notion of protocol to select the output while starting streaming rather than only asking the services.

The output id provided by the service is considered as preferred.

### Motivation and Context

Part of the implementation of the RFC 45.

### How Has This Been Tested?

Only testing trying to stream SRT and RIST, need more testing

### Types of changes

- New feature (non-breaking change which adds functionality)

### Checklist:

- [x] My code has been run through [clang-format](https://github.com/obsproject/obs-studio/blob/master/.clang-format).
- [x] I have read the [**contributing** document](https://github.com/obsproject/obs-studio/blob/master/CONTRIBUTING.rst).
- [x] My code is not on the master branch.
- [x] The code has been tested.
- [x] All commit messages are properly formatted and commits squashed where appropriate.
- [x] I have included updates to all appropriate documentation.

PR #8088, opened 2023-01-19
RFC 45-3: Use protocol for video codec selections

tytan652:
### Description

**This PR is about the [RFC 45](https://github.com/obsproject/rfcs/pull/45).**

Depends on:
- https://github.com/obsproject/obs-studio/pull/5104
- https://github.com/obsproject/obs-studio/pull/8087

Make OBS Studio rely on the notion of protocol and service compatibility to enable some codec to be used with the stream service.

### Motivation and Context

Part of the implementation of the RFC 45.

Remove the hardcoded list and allow AV1 on SRT/RIST once MPEG-TS has it's standard ready.

### How Has This Been Tested?

Changing services and check if codecs shown match with compatible ones.

### Types of changes

- New feature (non-breaking change which adds functionality)

### Checklist:

- [x] My code has been run through [clang-format](https://github.com/obsproject/obs-studio/blob/master/.clang-format).
- [x] I have read the [**contributing** document](https://github.com/obsproject/obs-studio/blob/master/CONTRIBUTING.rst).
- [x] My code is not on the master branch.
- [x] The code has been tested.
- [x] All commit messages are properly formatted and commits squashed where appropriate.
- [x] I have included updates to all appropriate documentation.

PR #8087, opened 2023-01-19
RFC 45-2: Add protocol in Services API and enum functions in Outputs API

tytan652:
### Description

**This PR is about the [RFC 45](https://github.com/obsproject/rfcs/pull/45).**

Depends on:
- #5104

Additions (those will be used in other parts of RFC 45 implementation):
- Adds `get_protocol` to services
- Add three enumeration functions:
  - `obs_enum_output_protocols` to enumerate protocols
  - `obs_enum_output_types_with_protocol` to enumerate outputs with the given protocol.

### Motivation and Context

Part of the implementation of the RFC 45.

### How Has This Been Tested?

Testing by logging enumeration functions outputs and through other part of RFC 45 implementation.

### Types of changes

- New feature (non-breaking change which adds functionality)

### Checklist:

- [x] My code has been run through [clang-format](https://github.com/obsproject/obs-studio/blob/master/.clang-format).
- [x] I have read the [**contributing** document](https://github.com/obsproject/obs-studio/blob/master/CONTRIBUTING.rst).
- [x] My code is not on the master branch.
- [x] The code has been tested.
- [x] All commit messages are properly formatted and commits squashed where appropriate.
- [x] I have included updates to all appropriate documentation.

PR #5104, opened 2021-08-09
RFC 45-1: Add protocol in Outputs API and check service protocol in the "UI"

tytan652:
### Description

**This PR is about the [RFC 45](https://github.com/obsproject/rfcs/pull/45).**

- Add a "protocols" and "protocols_prefixes" fields in `obs_output_info` and add protocols to outputs meant for serv
- Make the UI skip services where the protocol is not registered and also skip RTMPS servers for service providing RTMP and RTMPS server when the later is not registered.

### Motivation and Context

Part of the implementation of the RFC 45.

### How Has This Been Tested?

**Artifacts from Github actions will not much help with testing this PR since it build with all protocol support.**
- Build OBS without FTL support and check if "Glimesh" service is not shown.
- Build OBS without RTMPS support and check if "YouTube - RTMPS" RTMPS servers, "Facebook Live" service and "STAGE TEN" are not shown, RTMPS servers from other services are also not shown.
- Build OBS with support for both and check those three services are shown.

### Types of changes

- Bug fix (non-breaking change which fixes an issue)
  - Showing services that are not supported is sort of a bug
- New feature (non-breaking change which adds functionality)
- Documentation (a change to documentation pages)

### Checklist:

- [x] My code has been run through [clang-format](https://github.com/obsproject/obs-studio/blob/master/.clang-format).
- [x] I have read the [**contributing** document](https://github.com/obsproject/obs-studio/blob/master/CONTRIBUTING.rst).
- [x] My code is not on the master branch.
- [x] The code has been tested.
- [x] All commit messages are properly formatted and commits squashed where appropriate.
- [x] I have included updates to all appropriate documentation.

PR #8090, opened 2023-01-19
RFC 45-5: Allow selection of the audio codec/encoder

tytan652:
### Description

**This PR is about the [RFC 45](https://github.com/obsproject/rfcs/pull/45).**

Depends on:
- https://github.com/obsproject/obs-studio/pull/5104
- https://github.com/obsproject/obs-studio/pull/8087
- https://github.com/obsproject/obs-studio/pull/8088
- https://github.com/obsproject/obs-studio/pull/8089

Allow users to choose AAC or Opus to stream and record.

Stream side, codec are shown depending on the protocol used and the service compatibility.

If using different encoders in advanced, only common bitrates will be shown, this is made to avoid to over-complexify the UI.

SimpleAudio
SimpleAudioOpen

AdvAudio
AdvAudioOpened

AdvAudioRec

### Motivation and Context

Part of the implementation of the RFC 45.

### How Has This Been Tested?

Only testing trying to stream SRT with opus and AAC, might need more testing.

### Types of changes

### Checklist:


UI/audio-encoders.cpp
```diff
@@ -162,52 +227,119 @@ static void PopulateBitrateMap()
...
+			   << entry.second << ')';
+
+		blog(LOG_DEBUG, "AAC simple encoder bitrate mapping:%s",
+		     ss.str().c_str());
+	});
+}
+
+static map<int, std::string> simpleOpusBitrateMap;
+
+static void PopulateSimpleOpusBitrateMap()
```
derrod, line 275: What is the reasoning behind introducing a bitrate map for Opus? For AAC it largely exists because some encoders are picky about support (mainly MediaFoundation), but for Opus we only have one, and that one does not seem to have any restrictions on bitrate that would be relevant here.

UI/window-basic-main.cpp
```diff
@@ -1570,6 +1576,15 @@ void OBSBasic::InitBasicConfigDefaults2()
 				  useNV ? SIMPLE_ENCODER_NVENC
 					: SIMPLE_ENCODER_X264);
 
+	const char *aac_default = "ffmpeg_aac";
+	if (EncoderAvailable("libfdk_aac"))
+		aac_default = "libfdk_aac";
+	else if (EncoderAvailable("CoreAudio_AAC"))
+		aac_default = "CoreAudio_AAC";
```
derrod, line 1583: Probably fine because libfdk is generally not available on platforms where we expect CoreAUdio, but this does change the priority from the current (CoreAudio > libfdk > MF > FFmpeg)

UI/window-basic-settings-stream.cpp
```diff
@@ -1272,6 +1356,22 @@ static QString get_adv_fallback(const QString &enc)
 	return "obs_x264";
 }
 
+static QString get_adv_audio_fallback(const QString &enc)
+{
+	const char *codec = obs_get_encoder_codec(QT_TO_UTF8(enc));
+
+	if (strcmp(codec, "aac") == 0)
+		return "ffmpeg_opus";
```
derrod, line 1364: I assume this is meant to fall back to opus if AAC is used and AAC if opus is used, wouldn't it make sense to try other AAC encoders first before switching to a diferent codec?

UI/window-basic-settings-stream.cpp
```diff
@@ -1272,6 +1356,22 @@ static QString get_adv_fallback(const QString &enc)
...
+
+	QString aac_default = "ffmpeg_aac";
+	if (EncoderAvailable("libfdk_aac"))
+		aac_default = "libfdk_aac";
+	else if (EncoderAvailable("CoreAudio_AAC"))
+		aac_default = "CoreAudio_AAC";
```
derrod, line 1370: Same as above, changes order from what it currently is.

UI/window-basic-settings-stream.cpp
tytan652: If the unsupported codec is AAC, why return an AAC encoder ?

We should fallback to the opus encoder.

UI/audio-encoders.cpp
tytan652: The settings as it is relies heavily on the bitrate map workflow.

Working around it is worse than applying the same flow as AAC to opus.

UI/window-basic-main.cpp
tytan652: I forgot to change that when I learn about it.

tytan652:
Remove prefix related changes as it was done in the RFC.

UI/window-basic-main-outputs.cpp
```diff
@@ -1290,17 +1363,16 @@ bool SimpleOutput::ReplayBufferActive() const
...
 	OBSEncoder streamArchiveEnc;
-	OBSEncoder aacTrack[MAX_AUDIO_MIXES];
+	OBSEncoder recordTrack[MAX_AUDIO_MIXES];
 	OBSEncoder videoStreaming;
 	OBSEncoder videoRecording;
 
 	bool ffmpegOutput;
 	bool ffmpegRecording;
 	bool useStreamEncoder;
+	bool useStreamAudioEncoder;
```
derrod, line 1373: Not sure how much sense it really makes to have this option. For video encoding it's to save resources, but in the case of audio you're still creating separate encoders for streaming and recording. And by making it the default you're likely breaking existing setups for streaming FTL + recording locally since it would now use Opus wouldn't it?

tytan652: > Not sure how much sense it really makes to have this option. For video encoding it's to save resources, but in the case of audio you're still creating separate encoders for streaming and recording.

The point of this is just to allow the user to indicate to OBS to use the same codec/encoder type as the stream output, if the user change the streaming one the recording change too. This is not a question of saving resources.

> And by making it the default you're likely breaking existing setups for streaming FTL + recording locally since it would now use Opus wouldn't it?

I should maybe add a piece of code to make those user use AAC on recording if the config string does not exist in the ini file.

derrod: I think the default should just remain AAC. There really is no reason for it to follow the stream output by default since that has the potential of breaking setups when switching to a service that uses Opus, which would be unexpected and not visible in the UI until starting the recording fails.

PR #8093, opened 2023-01-19
libobs: Add support to outputs for multi video encoders

kc5nra:
### Description
This adds the internal plumbing to allow outputs multiple video encoders/tracks in addition to existing ability to add multiple audio tracks.

### Motivation and Context
This will enable future output plugins to support sending video in multiple resolutions, codecs, and configurations.  Many protocols benefit from having multiple encoded outputs:
* Variant programs in mpegts
* WebRTC simulcast
* MoQ/QUIC-based
* Emerging protocols/codecs that require communicating with a diverse set of connection profiles

This also aligns and simplifies a lot of the code that handled multiple audio tracks but hardcoded support for only one encoder in the output code.

### How Has This Been Tested?
Tested with example implementations found here ____

### Types of changes

### Checklist:


jp9000:
Hmm, I skimmed a bit over this a bit and it seems good. It's just a good isolated and focused back-end change. If you want to have support for multiple video encodings in a single stream then this is the only real way you can do it. As long as the old APIs still function identically to how they used to, then I feel comfortable with it. The code is fairly simple and straightforward.

That's just my initial take on this though. As long as this is tested with our existing code and doesn't change current functionality I don't see any real problem with it.

libobs/obs.h
```diff
@@ -2105,6 +2105,17 @@ EXPORT size_t obs_output_get_mixers(const obs_output_t *output);
 EXPORT void obs_output_set_video_encoder(obs_output_t *output,
 					 obs_encoder_t *encoder);
 
+/**
+ * Sets the current video encoder associated with this output,
+ * required for encoded outputs
```
RytoEX, line 2110: Use ending punctuation if comment is more than a single sentence. See the comment for `obs_output_set_audio_encoder` below.

Fix throughout.

libobs/obs-output.c
```diff
@@ -888,11 +970,27 @@ void obs_output_set_audio_encoder(obs_output_t *output, obs_encoder_t *encoder,
...
+}
+
 obs_encoder_t *obs_output_get_video_encoder(const obs_output_t *output)
 {
-	return obs_output_valid(output, "obs_output_get_video_encoder")
-		       ? output->video_encoder
-		       : NULL;
+	if (!obs_output_valid(output, "obs_output_get_video_encoder"))
+		return NULL;
+	size_t first_venc_idx;
```
RytoEX, line 989: Prefer a blank line after `return;` when followed by statements.

RytoEX (changes requested):
Mostly nits or clarifications.

> How Has This Been Tested?
> Tested with example implementations found here ____

Was this intended to be filled in at some point?

kc5nra:
Rebased. I still need to rebase my other changes that use this PR to confirm that everything still works. I'll do this tomorrow and follow up here

Lain-B (approved):
Did a thorough review, it's deceptively straightforward. Does exactly what the title says and nothing more.

Lain-B (approved):
Good catch on the fixes. Especially `interleave_packets()`. I didn't even think of that myself. Will have to be very meticulous about testing just because of "gotcha" situations like that.

libobs/obs-output.h
```diff
@@ -28,8 +28,13 @@ extern "C" {
 #define OBS_OUTPUT_SERVICE (1 << 3)
 #define OBS_OUTPUT_MULTI_TRACK (1 << 4)
```
tt2468, line 29: Wonder if we can somehow apply a DEPRECATED flag to this to throw a warning for anything using the old name?

kc5nra: Yeah, I thought about this but couldn't come up with anything that fit that usage pattern in a macro

kc5nra:
> Mostly nits or clarifications.

> > How Has This Been Tested?
> > Tested with example implementations found here ____

> Was this intended to be filled in at some point?

@RytoEX I have a PR that enhances ffmpeg-mux to support multiple video tracks, but since it depends on this PR it was a bit hard to keep that updated with changes to this PR. ffmpeg-mux is also debatably useful (maybe hls output?). SRT might be a better example

PR #8095, opened 2023-01-19
RFC 45-6: Enable more connection info in Services API

tytan652:
### Description

**This PR is about the [RFC 45](https://github.com/obsproject/rfcs/pull/45).**

Depends on:
- #5104
- #8087
- #8088
- #8089
- #8090

Or if merge as is, contains:
- Closes #5104
- Closes #8087
- Closes #8088
- Closes #8089
- Closes #8090

Replace and deprecate url, key, username and password functions to replace with only one `obs_service_get_connect_info()`.

This allow to ask the service for various connection information like:
- Server URL (Commonly used in various protocol)
- Stream key (Usually used with RTMP/FTL/HLS)
- Username (Can be use with RMTP/~~FTL~~/RIST)
- Password (Can be use with RMTP/~~FTL~~/RIST)
- Stream ID (Can be used for SRT)
- Encryption passphrase (Can be used for SRT/RIST)
- It also allow third-party protocol to use custom connection infos when third-party outputs+services will be possible.

Those info type are not bound to protocol.

Adds a function that allow the service to indicate if it has all the connection info that it needs to start streaming: `obs_service_can_try_to_connect`.

### Motivation and Context

Part of the implementation of the RFC 45 recent changes.

### How Has This Been Tested?

Only tested trying to stream SRT, might need more testing.

### Types of changes

### Checklist:


tytan652:
Rebased on #8090 to be able to work on top of all RFC 45 PRs.

tytan652:
Remove prefix related changes as it was done in the RFC.

docs/sphinx/reference-outputs.rst
jp9000, line 1: Two issues:
1.) Protocols can be duplicated when enumerating.
2.) Protocol enumeration is a bit inconsistent depending on what the caller does.
3.) Callers have to deal with allocations in some cases.
4.) If the caller does not manage the allocation, the lifetime of a string returned by protocol enumeration is only guaranteed to last until the next enumeration, which can cause crashes.

Proposed solution:
Keep some sort of array of allocated strings in, for example, maybe `obs_core_data`. When an output is successfully registered, it adds its protocols to that list. An array of string pointers, each string being an individual allocation. It detects duplicates and does not create a new value in that array if it is a duplicate. On obs shutdown, the strings in the array are all freed, and then the array storing those string pointers are all freed.

That way, you can enumerate all protocols without duplicates, all protocol strings have a lifetime that lasts until the end of the program (preventing the possibility of crashes), and callers do not have to deal with allocations (they can just get `const char *` pointers with a guaranteed lifespan).

libobs/obs-output.c
```diff
@@ -290,7 +290,8 @@ bool obs_output_start(obs_output_t *output)
 		return false;
 
 	has_service = (output->info.flags & OBS_OUTPUT_SERVICE) != 0;
-	if (has_service && !obs_service_initialize(output->service, output))
+	if (has_service && !obs_service_can_try_to_connect(output->service) &&
```
tt2468, line 293: This seems like it might supposed to be operator||?

docs/sphinx/reference-outputs.rst
tytan652: Changes done in #5104 and #8087, I will rebase others soon.

libobs/obs-output.c
tytan652: Good catch, should be good now.

tytan652:
After off-thread discussion `OBS_SERVICE_CONNECT_INFO_STREAM_KEY` is now an alias of `OBS_SERVICE_CONNECT_INFO_STREAM_ID`.

UI/data/locale/en-US.ini
```diff
@@ -917,10 +917,8 @@ Basic.Settings.Stream.TTVAddon.BTTV="BetterTTV"
 Basic.Settings.Stream.TTVAddon.FFZ="FrankerFaceZ"
 Basic.Settings.Stream.TTVAddon.Both="BetterTTV and FrankerFaceZ"
 Basic.Settings.Stream.MissingSettingAlert="Missing Stream Setup"
+Basic.Settings.Stream.MissingSetting="The service is misconfigured. \n\nOpen settings to and check the 'Stream' tab."
 Basic.Settings.Stream.StreamSettingsWarning="Open Settings"
-Basic.Settings.Stream.MissingUrlAndApiKey="URL and Stream Key are missing.\n\nOpen settings to enter the URL and Stream Key in the 'stream' tab."
-Basic.Settings.Stream.MissingUrl="Stream URL is missing.\n\nOpen settings to enter the URL in the 'Stream' tab."
-Basic.Settings.Stream.MissingStreamKey="Stream key is missing.\n\nOpen settings to enter the stream key in the 'Stream' tab."
```
jp9000, line 922: > Open settings to and check
A bit of an English error here. I'm guessing this was originally copied/pasted.

But this error reporting could probably use some reworking. I'm not entirely sure what to suggest here though. This all-encompassing error message is actually *less* helpful than the old errors. At least with the old errors the users knew what to change or knew what to do. With this error message, the users have a more vague message that sounds like "your settings are bad, try again". What settings are bad? The users probably need to know that.

UI/window-basic-settings.cpp
```diff
@@ -448,6 +512,7 @@ OBSBasicSettings::OBSBasicSettings(QWidget *parent)
 	HookWidget(ui->advOutNoSpace,        CHECK_CHANGED,  OUTPUTS_CHANGED);
 	HookWidget(ui->advOutRecFormat,      COMBO_CHANGED,  OUTPUTS_CHANGED);
 	HookWidget(ui->advOutRecEncoder,     COMBO_CHANGED,  OUTPUTS_CHANGED);
+	HookWidget(ui->advOutRecAudioEncoder,     COMBO_CHANGED,  OUTPUTS_CHANGED);
```
jp9000, line 515: Again here:
`advOutRecAudioEncoder -> advOutRecAEncoder`
then make these parameters align with the others to cause us less physical pain.

jp9000 (changes requested):
Mostly just nitpicks, but there are a few things that need to be adjusted in order to finish this PR.

plugins/rtmp-services/rtmp-common.c
```diff
@@ -478,6 +482,32 @@ static void fill_stream_key_link(json_t *service, obs_data_t *settings)
...
+	server = json_array_get(servers, 0);
+	url = get_string_val(server, "url");
+
+	if (strncmp(url, RTMPS_PREFIX, strlen(RTMPS_PREFIX)) == 0) {
+		obs_data_set_string(settings, "protocol", "RTMPS");
+		return;
+	}
+
+	obs_data_set_string(settings, "protocol", "RTMP");
+}
```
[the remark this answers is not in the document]
tytan652: The use `obs_data_set_*` is necessary for now, since protocol selection is a thing only in RFC 39.
We need to manage the protocol in this service this way for now.

UI/window-basic-settings-stream.cpp
```diff
@@ -520,6 +531,43 @@ void OBSBasicSettings::ServiceChanged()
...
+			obs_get_service_properties("rtmp_common");
+		obs_property_t *services = obs_properties_get(props, "service");
+
+		OBSDataAutoRelease settings = obs_data_create();
+
+		obs_data_set_string(settings, "service",
+				    QT_TO_UTF8(ui->service->currentText()));
+		obs_property_modified(services, settings);
+
+		obs_properties_destroy(props);
```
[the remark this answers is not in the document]
tytan652: This is the only way for now, we already do that for most of the service settings, until RFC 39 replaces this kind of hack.

UI/data/locale/en-US.ini
```diff
@@ -992,6 +993,7 @@ Basic.Settings.Output.Warn.EnforceResolutionFPS.Resolution="Resolution: %1"
 Basic.Settings.Output.Warn.EnforceResolutionFPS.FPS="FPS: %1"
 Basic.Settings.Output.Warn.ServiceCodecCompatibility.Title="Incompatible Encoder"
 Basic.Settings.Output.Warn.ServiceCodecCompatibility.Msg="The streaming service \"%1\" does not support the encoder \"%2\". The encoder will be changed to \"%3\".\n\nDo you want to continue?"
+Basic.Settings.Output.Warn.ServiceCodecCompatibility.Msg2="The streaming service \"%1\" does not support encoders \"%2\" and \"%3\". Encoders will be changed to \"%4\" and \"%5\".\n\nDo you want to continue?"
```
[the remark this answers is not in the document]
tytan652: Unresolve it if there is still an issue.

UI/window-basic-settings-stream.cpp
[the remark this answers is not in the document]
tytan652: Unresolve it if it needs more discussion.

plugins/rtmp-services/rtmp-common.c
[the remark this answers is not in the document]
tytan652: Unresolve it if it needs more discussion.

libobs/obs-output.c
```diff
@@ -2699,3 +2699,34 @@ const char *obs_output_get_protocols(const obs_output_t *output)
...
+					*id = obs->output_types.array[i].id;
+					return true;
+				}
+				count++;
+			}
+			substr = next ? next + 1 : NULL;
+		}
+	}
+	return false;
+}
```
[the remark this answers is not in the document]
tytan652: This enum has future purpose/usage in RFC 39, I put it in RFC 45 to avoid creating something like `obs_get_first_output_type_with_protocol` just for a transition period.

I will check the `enum_cb` suggestion.

UI/data/locale/en-US.ini
tytan652: In RFC 39, we no longer to be able to guess what is wrong since services will no longer be only first party.
But I'll move this change to RFC 39 since we can add warnings and errors in the service property view.

jp9000 (approved):
LGTM™

notr1ch:
A crash when using custom output was reported on Discord (https://discord.com/channels/348973006581923840/374636084883095554/1087384798894501939)

```
	if (!codecs || IsCustomService()) {
		const char *output;
		char **output_codecs;

		obs_enum_output_types_with_protocol(QT_TO_UTF8(protocol),
						    &output, return_first_id);

		output_codecs = strlist_split(
			obs_get_output_supported_video_codecs(output), ';',
			false);

		ret = service_supports_codec((const char **)output_codecs,
					     codec);
```

`obs_enum_output_types_with_protocol` fails to find an output, so `output` remains uninitialized, causing a crash in `obs_get_output_supported_audio_codecs`. I'm not sure what the intention is here since the code clearly wants to go this path with the `IsCustomService` condition. Changing it to `if (!codecs && !IsCustomService())` might be more correct if we want to allow all codec compatibility for custom outputs.

The same issue exists in `ServiceAndACodecCompatible` too, perhaps these functions could be merged to avoid logic duplication.

tytan652:
> if we want to allow all codec compatibility for custom outputs.

We don't.

tytan652:
The main problem is why no output were found.

PR #8106, opened 2023-01-21
updater: Hash files with multiple threads

derrod:
### Description

Breaks out the hashing of existing files into a separate step that runs with multiple threads.

On my system this results in about 4x faster hashing when the files are not cached (7800 ms -> 1900 ms) and still about 2x faster hashing when files are in the OS cache (980 ms -> 480 ms).

### Motivation and Context

Faster hashing. Further speeding up the updating process.

### How Has This Been Tested?

Tested locally with changes applied on top of #8074.

### Types of changes

### Checklist:

### DIFF
--- a/UI/win-update/updater/updater.cpp
+++ b/UI/win-update/updater/updater.cpp
@@ -25,6 +25,7 @@
 #include <string>
 #include <mutex>
 #include <unordered_set>
+#include <queue>
 
 using namespace std;
 using namespace json11;
@@ -356,6 +357,7 @@ struct deletion_t {
 	}
 };
 
+static unordered_map<string, wstring> hashes;
 static vector<update_t> updates;
 static vector<deletion_t> deletions;
 static mutex updateMutex;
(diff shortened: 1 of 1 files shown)


UI/win-update/updater/updater.cpp
```diff
@@ -594,6 +596,78 @@ static inline bool WideToUTF8(char *utf8, int utf8Size, const wchar_t *wide)
...
+static void RunHasherWorkers(int num, const Json &packages)
+try {
+
+	for (const Json &package : packages.array_items()) {
+		for (const Json &file : package["files"].array_items()) {
+			if (!file["name"].is_string())
+				continue;
+			hashQueue.push(file["name"].string_value());
+		}
+	}
```
derrod, line 651: I haven't included this, but I found that by reversing the order of packages (thus adding `obs-browser` first) I can actually shave off another 100 ms or so since CEF will be processed first rather than at the end, meaning that we spend less time with only a single thread working on CEF while the others are already finished.

```c++
	/* Okay I know this looks mad. But there's actually a somewhat sane reason
	 * to do this:
	 *
	 * libcef.dll takes the longest of all files to hash, so by ensuring the processing
	 * of it gets started as soon as possible we don't spend as long waiting for the last
	 * thread to complete as if it was added in its normal position (at the end).
	 *
	 * (Technically this will add any file > 10 MiB to the start, which is only three )
	 * 
	 * This saves a whole 50 ms on my machine!
	 */
	Json::array pVec = packages.array_items();
	for_each(pVec.rbegin(), pVec.rend(), [](const Json &package) {
		for (const Json &file : package["files"].array_items()) {
			if (!file["name"].is_string())
				continue;
			hashQueue.push(file["name"].string_value());
		}
	});
```

notr1ch (approved):

PR #8109, opened 2023-01-21
libobs-opengl:  Do not close X11 platform display on error

pobrn:
### Description

The platform display retrieved in `OBSApp::OBSInit()` belongs to Qt, therefore libobs components should not destroy it. Because when it is destroyed, it may result in UAFs when Qt components try to use it. Not closing the platform display is the main change, but there are also two other small changes that close X displays that weren't previously closed. See the commit messages for more details.

### Motivation and Context

Fixes a use-after-free issue that manifests itself when `gl_x11_egl_platform_create()` fails.

### How Has This Been Tested?

Tested locally on top of the master branch. Linux 6.1.7, with X11, NVIDIA.

### Types of changes

### Checklist:

-- 

@GeorgesStavracas @kkartaltepe

### DIFF
--- a/libobs-opengl/gl-x11-egl.c
+++ b/libobs-opengl/gl-x11-egl.c
@@ -385,6 +385,8 @@ static void gl_x11_egl_platform_destroy(struct gl_platform *plat)
 
 	gl_context_destroy(plat);
 	eglTerminate(plat->edisplay);
+	if (plat->close_xdisplay)
+		XCloseDisplay(plat->xdisplay);
 	bfree(plat);
 }
 
--- a/libobs-opengl/gl-x11-egl.c
+++ b/libobs-opengl/gl-x11-egl.c
@@ -82,6 +82,7 @@ struct gl_platform {
 	EGLConfig config;
 	EGLContext context;
 	EGLSurface pbuffer;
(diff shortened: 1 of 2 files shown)


kkartaltepe:
this seems good to me. You can also merge the two libobs-opengl commits.

kkartaltepe (approved):

jpark37:
Can you put X11 in the commit title? I thought this was a cross-platform issue, but it's not.

pobrn:
> Can you put X11 in the commit title? I thought this was a cross-platform issue, but it's not.

Is it better now? Or did you have something else in mind?

jpark37:
That was it. Thanks!

GeorgesStavracas (approved):

RytoEX:
As a matter of personal preference, I would probably not have included the full stack trace in the commit message. I would probably have mentioned the AddressSanitizer heap-use-after-free error, and then documented the full stack trace on the PR. As an aside, you can wrap the commit message bodies at 72 characters rather than 60 characters. That said, I will not let any of those items block merging.

pobrn:
Sorry, I was meaning to change the commit messages but have not had the time yet.

PR #8110, opened 2023-01-21
UI: Don't open properties dialog if item is scene

cg2121:
### Description
This disables opening of the properties dialog if a scene is double clicked in the source tree.

### Motivation and Context
The scene properties dialog is empty, so it is unnecessary to open it.

### How Has This Been Tested?
Double clicked scene item that was scene.

### Types of changes

### Checklist:


notr1ch:
It might be better to use `obs_source_configurable` to match existing behavior to determine whether a source has properties.

cg2121:
Updated to use the suggestion from @notr1ch

PR #8115, opened 2023-01-22
updater: Multi-threaded delta patching

derrod:
### Description

Makes applying delta patches multi-threaded.

On its own this shaves off about 3 seconds when doing 28.1.2 => 29.0.0 on my system. Together with #8106 this speeds up the entire process quite a bit.

The speedup when patching multiple large files (e.g. CEF) as in my 27.2.4 => 29.0.0 test case I use for #8074 is a decent bit larger. 

### Motivation and Context

Gotta go fast(er).

### How Has This Been Tested?

28.1.2 -> 29.0.0 with official update server.

### Types of changes

### Checklist:

### DIFF
--- a/UI/win-update/updater/updater.cpp
+++ b/UI/win-update/updater/updater.cpp
@@ -141,8 +141,8 @@ try {
 	WinHandle hSrc;
 	WinHandle hDest;
 
-	hSrc = CreateFile(src, GENERIC_READ, 0, nullptr, OPEN_EXISTING,
-			  FILE_FLAG_SEQUENTIAL_SCAN, nullptr);
+	hSrc = CreateFile(src, GENERIC_READ, FILE_SHARE_READ, nullptr,
+			  OPEN_EXISTING, FILE_FLAG_SEQUENTIAL_SCAN, nullptr);
 	if (!hSrc.Valid())
 		throw LastError();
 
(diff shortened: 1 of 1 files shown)


UI/win-update/updater/updater.cpp
```diff
@@ -1085,6 +1086,72 @@ static bool UpdateFile(update_t &file)
...
+				completedUpdates - updateQueue.size();
+			int position = (int)(((float)updatesInstalled /
+					      (float)completedUpdates) *
+					     100.0f);
+			if (position > lastPosition) {
+				lastPosition = position;
+				SendDlgItemMessage(hwndMain, IDC_PROGRESS,
+						   PBM_SETPOS, position, 0);
+			}
+			ulock.unlock();
```
derrod, line 1124: The locks here are *probably* not necessary but I wanted to be on the safe side.

derrod: Removed the locks and avoided accessing the queue (even though `size()` is probably fine). `SendMessage` is supposed to be thread safe and duplicate progress messages probably aren't an issue worth worrying about.

derrod:
Little comparison with this PR and #8106 which is now merged (note the different axis scales):
**Before:**
before
**After:**
after

So all in all, cuts down the time to update from ~22 seconds to ~8 seconds :)

derrod:
This has been rebased and updated to support #8169

RytoEX (approved):
@notr1ch approved off-thread, so let's merge this.

PR #8126, opened 2023-01-23
UI: Add new scene at current row

cg2121:
### Description
Instead of having a new scene at the bottom of the list, have it added after the current item.

### Motivation and Context
https://ideas.obsproject.com/posts/2111/new-scene-here

### How Has This Been Tested?
Added new scene to see where it was added in the list

### Types of changes

### Checklist:

### DIFF
--- a/UI/window-basic-main.cpp
+++ b/UI/window-basic-main.cpp
@@ -2872,7 +2872,7 @@ void OBSBasic::AddScene(OBSSource source)
 
 	QListWidgetItem *item = new QListWidgetItem(QT_UTF8(name));
 	SetOBSRef(item, OBSScene(scene));
-	ui->scenes->addItem(item);
+	ui->scenes->insertItem(ui->scenes->currentRow() + 1, item);
 
 	obs_hotkey_register_source(
 		source, "OBSBasic.SelectScene",


WizardCM:
I consider this a breaking change as it's a pretty drastic difference in behaviour, though I agree it's a positive improvement.

I do worry that it further differentiates from the Sources list, which *always* places items at the top of the list..

CodeYan01:
I hope this change is also applied for duplicating scenes. Makes a lot more sense when duplicating scenes to have it be after the scene i just duplicated. Even if it requires a separate PR

Warchamp7:
> I consider this a breaking change as it's a pretty drastic difference in behaviour, though I agree it's a positive improvement.
> 
> I do worry that it further differentiates from the Sources list, which _always_ places items at the top of the list..

It is worth nothing though that the source tree differs in that the order of it's items matters.


While I don't dislike this PR conceptually, I'm not sure I like the behavior change

Warchamp7 (approved):
After consideration I'm approving this from a UX perspective.

Let's apply this behaviour to duplicating as well.

cg2121:
It also works for duplicating, as it uses the same function.

sebastian-s-beckmann (approved):
Works fine, change seems correct.

Warchamp7:
@tt2468 @WarmUpTill Do we foresee this being a breaking change for anything? 

I vaguely recall a Discord discussion about this with @tt2468 at least

WarmUpTill:
> @tt2468 @WarmUpTill Do we foresee this being a breaking change for anything?
> 
> I vaguely recall a Discord discussion about this with @tt2468 at least

I don't think this would break anything on advanced scene switcher side.
But thanks for the the heads up! 👍

tt2468:
This to me definitely classifies as a breaking change, as there are definitely applications out there which make assumptions about where scenes are added, but I don't believe that it would cause anything more than minor graphical bugs. For example, an application might receive a scene creation event and ignore other events, and put the scene at the top of its own list visually. This would be corrected during any resync of the scene order though.

So basically, this is fine to merge, but it's worth mentioning in the next obs-websocket release notes.

PR #8113, opened 2023-01-22
deps/media-playback: use of deprecated ffmpeg field

kkartaltepe:
### Description

`pkg_duration` was deprecated in favor of simply `duration`. 

### Motivation and Context

With warnings as errors this fails to build on systems with bleeding edge ffmpeg.

### How Has This Been Tested?

Compiles fine on my machine and we are already using `last_duration` so this seems fine. I'm hoping the non-deprecated field exists back on ubuntu.

### Types of changes

- Bug fix (non-breaking change which fixes an issue)

### Checklist:

- [x] My code has been run through [clang-format](https://github.com/obsproject/obs-studio/blob/master/.clang-format).
- [x] I have read the [**contributing** document](https://github.com/obsproject/obs-studio/blob/master/CONTRIBUTING.rst).
- [x] My code is not on the master branch.
- [x] The code has been tested.
- [x] All commit messages are properly formatted and commits squashed where appropriate.
- [x] I have included updates to all appropriate documentation.

PR #8127, opened 2023-01-23
Add CALM_DEPRECATION cmake option and fix some warnings

tytan652:
### Description

Contains a rework of #8113

Add an option `CALM_DEPRECATION` to enable `-Wno-error=deprecated-declarations` for Linux users that might use bleeding-edge dependencies compared to obs-deps.

Fix a few other warning/errors.

### Motivation and Context

Improve #6759 and fix warnings/errors.

### Types of changes

### Checklist:

### DIFF
--- a/cmake/Modules/CompilerConfig.cmake
+++ b/cmake/Modules/CompilerConfig.cmake
@@ -90,6 +90,7 @@ else()
     set(CMAKE_CUDA_COMPILER_LAUNCHER "${CCACHE_PROGRAM}") # CMake 3.9+
   endif()
 
+  option(CALM_DEPRECATION "Keep deprecated-declarations as warnings" OFF)
   add_compile_options(
     -Werror
     -Wextra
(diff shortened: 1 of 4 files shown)


PatTheMav:
Changes themselves look correct - can you educate me how a developer might end up in a situation necessitating this?

tytan652:
On Linux (non-Flatpak), we could end up with a new FFmpeg (Python …) version which add can add deprecations.

This option enable the use to allows those deprecations.

jp9000:
Confirmed working, thank you very much

jp9000:
@PatTheMav to be more specific, I asked tytan to add something like this because I'm building locally with a newer version of ffmpeg and am getting unexpected deprecation warnings. This just makes it more convenient to bypass the warning if by chance you just so happen to use a newer version of some library that just so happens to deprecate functions we're currently using.

PR #8130, opened 2023-01-23
UI: Add transition override to transitions dock

cg2121:
### Description
This adds the per-scene transitions to the transitions dock, so that they are not just in the context menu.

Screenshot from 2023-05-28 01-37-10

### Motivation and Context
Better UI

### How Has This Been Tested?
Set per-scene transitions on different scenes.

### Types of changes

### Checklist:


GeorgesStavracas:
Some comments from the UX discussion:

 - [ ] Rename the dock title to "Transitions"
 - [ ] Add a subtitle for the global transitions, named tentatively "Default transition"

@Warchamp7 requested being included here

WizardCM:
My biggest UX concern here is that the Transitions dock is very much not tied to a scene in any visual way, so just having two dropdowns with a transition name is super confusing. I'm not even sure if renaming the section to something like "Transition Override for '<scene>'" would be enough.

Additionally, I thought that there was a hope to move the Studio Mode controls (the center area) to the Transitions dock? Or is the plan to put it in a separate dock named Quick Transitions and move the gear that contains a bunch of Studio Mode settings into the Settings window?

Warchamp7:
> Additionally, I thought that there was a hope to move the Studio Mode controls (the center area) to the Transitions dock? Or is the plan to put it in a separate dock named Quick Transitions and move the gear that contains a bunch of Studio Mode settings into the Settings window?
>

I don't think I've ever taken any swings at the studio mode controls

> My biggest UX concern here is that the Transitions dock is very much not tied to a scene in any visual way, so just having two dropdowns with a transition name is super confusing. I'm not even sure if renaming the section to something like "Transition Override for ''" would be enough.
> 

That's the main purpose of adding an additional label for more context.

---

With the new extra controls added here, I'd like to avoid layout shifting by not having the Duration input field disappear for Cut/Stinger and instead be disabled with either no value listed, or 0ms / Automatic respectively

cg2121:
Updates:
- Set the title of the transitions dock to "Transitions"

- Added "Default" label to global transition
- Added "Current Scene" label to transition override
- These are different than suggested because I feel like they are clearer to users what they are

- Disabled the transition duration, instead of hiding it, when transition is a cut
- Disabled the transition override duration when transition is a cut or none

Screenshot from 2023-01-24 05-49-57

Warchamp7:
> * These are different than suggested because I feel like they are clearer to users what they are

Can we go with the originally proposed labels please

cg2121:
Updated it to use the suggested text. My only concern now is that it makes the dock too big.

Screenshot from 2023-05-28 01-37-10

cg2121:
Closing, as this makes the dock too big. Will reopen if a better UI for this can be made.

PR #8134, opened 2023-01-24
UI: Hide menu items if source is only audio

cg2121:
### Description
Hides the following items if source is only audio:
- Transform menu
- Resize output (size of source)
- Scale filter menu
- Blending mode menus
- Projector menus
- Screenshots
- Interact

### Motivation and Context
These are not needed if the source is only audio.

### How Has This Been Tested?
Right clicked audio source to make sure the items were hidden.

### Types of changes

### Checklist:


norihiro:
The interact menu is hidden by the `OBS_SOURCE_INTERACTION` flag. I don't think we need to hide it for audio-only sources.

WizardCM:
I could have sworn this was intentional, but overall I like this change due to how different audio and video sources are conceptually.

RytoEX:
> I could have sworn this was intentional, but overall I like this change due to how different audio and video sources are conceptually.

If I recall correctly, we have generally avoided changing the context menus to avoid disrupting muscle memory, particularly for users who navigate them via keyboard or with accessibility tools. As I do not personally navigate context menus with keyboard or accessibility tools, I cannot say whether or not it matters in practice.

PR #8141, opened 2023-01-25
updater: Remove 32-bit Support

derrod:
### Description

Removes remaining support for updating a 32 bit OBS build from the updater.

(Ideally this should be merged after

* #8070
* #8114 
* #8115 
 
so I only have to solve conflicts (if any) in this PR and 

* #8074

)

### Motivation and Context

Code cleanup before more updates.

OBS has not provided 32 bit releases since 28.0.

### How Has This Been Tested?

Updater still works.

### Types of changes

- Code cleanup (non-breaking change which makes code smaller or more readable)
- Breaking change (fix or feature that would cause existing functionality to change)

### Checklist:

- [x] My code has been run through [clang-format](https://github.com/obsproject/obs-studio/blob/master/.clang-format).
- [x] I have read the [**contributing** document](https://github.com/obsproject/obs-studio/blob/master/CONTRIBUTING.rst).
- [x] My code is not on the master branch.
- [x] The code has been tested.
- [x] All commit messages are properly formatted and commits squashed where appropriate.
- [x] I have included updates to all appropriate documentation.

PR #8142, opened 2023-01-25
updater: Check if awaited instance matches current install 

derrod:
### Description

When checking for running OBS instance ensure the executable of an OBS process is actually residing the the directory we want to update.

Based on an old WIP commit: https://github.com/notr1ch/obs-studio/commit/95909ee25c3bbab9f6adc2a5c3795b0c115dd185

Contains and closes #8141 if merged first.

### Motivation and Context

Allow updating OBS even if some other `obs64.exe` is running (e.g. updating portable install while non-portable version is running).

Also selfishly, this means I won't have to use ShareX when recording my screen while working on updater updates :P

### How Has This Been Tested?

Verified portable install can be updated while existing OBS install is running

### Types of changes

### Checklist:

### DIFF
--- a/UI/win-update/updater/updater.cpp
+++ b/UI/win-update/updater/updater.cpp
@@ -48,6 +48,7 @@ int completedFileSize = 0;
 static int completedUpdates = 0;
 
 static wchar_t tempPath[MAX_PATH];
+static wchar_t obs_base_directory[MAX_PATH];
 
 struct LastError {
 	DWORD code;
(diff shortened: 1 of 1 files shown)


UI/win-update/updater/updater.cpp
```diff
@@ -531,7 +527,12 @@ static inline DWORD WaitIfOBS(DWORD id, const wchar_t *expected)
 	if (!proc.Valid())
 		return WAITIFOBS_WRONG_PROCESS;
 
-	if (!GetProcessImageFileName(proc, path, _countof(path)))
+	if (!GetModuleFileNameExW(proc, nullptr, path, _countof(path)))
```
notr1ch, line 530: Perhaps `QueryFullProcessImageNameW` is better here?

derrod: Either seems to work. Given the note in the MS Docs about `GetModuleFileName` being primarily intended for use in debuggers I've updated it to use `QueryFullProcessImageName` instead.

derrod:
Rebased now that #8141 is merged. Should be good to merge™

RytoEX (approved):
@notr1ch tentatively approved off-thread. Will merge soon.

PR #8146, opened 2023-01-25
UI: Use native color dialog on macOS

### LINKED ISSUE #4081: Picking Screen Color outside of window
## Platform
The Pick Screen Color option in any color choosing source doesn't register anything outside of the select color window.

Operating system and version: Mac OS Big Sur 11.2 Beta 
OBS Studio version: 26.1.2

## Expected Behavior
 the pick screen color option should work everywhere on the display.

## Current Behavior
 the screen color picker only works in the pick screen color window, no where else on the display which makes it pretty useless. 
 https://obsproject.com/logs/ucaqNtU89dSVC0C1 

## Steps to Reproduce

1. create a new color source
2. press select color
3. press pick screen color
here's a video of it to make it more clear
https://user-images.githubusercontent.com/76796294/104387227-ab790f80-5504-11eb-8210-4c8d0ecc2d2a.mp4

gxalpha:
### Description

Currently, we use the Qt provided color dialog on all platforms except Windows.
This was because of the native dialog "closing other open QDialogs" and events not working properly.

The former is either not true anymore, or has never been true. What happens is that other open QDialogs get put behind the main window, making it look like they were closed. This can be worked around with a `raise()` call on the main window (like in #8116).
[QTBUG-34532](https://bugreports.qt.io/browse/QTBUG-34532) is about events not working when subclassed. We're not subclassing the QColorDialog so this isn't be relevant to what we're doing.

### Motivation and Context

The Qt color picker on macOS has the problem that it can't pick screen colors from outside the color picker window, making it useless for picking screen colors.
macOS' color picker does not have this problem, making it the better option.

The original commit, adf6beb, only used the non-native dialog for macOS. a9a19cf changes this to include Linux (and others) to fix a hang on Ubuntu. ~~Such a hang is very likely to be caused or catalyzed by Qt, it would be nice to be able to reference a Qt Bug here so that we can revisit it in the future (if no such report exists but the problem is still present, it would be useful to both us and Qt to create one). cc @WizardCM~~
Update: The hang doesn't *seem* to trigger anymore, but let's leave it as-is for now and only focus on macOS.

Fixes #4081

### How Has This Been Tested?

macOS 13.1

### Types of changes

### Checklist:

### DIFF
--- a/UI/properties-view.cpp
+++ b/UI/properties-view.cpp
@@ -1807,15 +1807,18 @@ bool WidgetInfo::ColorChangedInternal(const char *setting, bool supportAlpha)
 		options |= QColorDialog::ShowAlphaChannel;
 	}
 
-	/* The native dialog on OSX has all kinds of problems, like closing
-	 * other open QDialogs on exit, and
-	 * https://bugreports.qt-project.org/browse/QTBUG-34532
-	 */
-#ifndef _WIN32
+#ifdef __linux__
+	// TODO: Revisit hang on Ubuntu with native dialog
 	options |= QColorDialog::DontUseNativeDialog;
 #endif
 
 	color = QColorDialog::getColor(color, view, QT_UTF8(desc), options);
+
+#ifdef __APPLE__
+	// TODO: Revisit when QTBUG-42661 is fixed
+	widget->window()->raise();
+#endif
+
 	if (!color.isValid())
 		return false;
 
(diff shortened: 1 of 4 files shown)


UI/properties-view.cpp
```diff
@@ -1807,15 +1807,18 @@ bool WidgetInfo::ColorChangedInternal(const char *setting, bool supportAlpha)
...
-#ifndef _WIN32
+#ifdef __linux__
+	// TODO: Revisit hang on Ubuntu with native dialog
 	options |= QColorDialog::DontUseNativeDialog;
 #endif
 
 	color = QColorDialog::getColor(color, view, QT_UTF8(desc), options);
+
+#ifdef __APPLE__
+	// TODO: Revisit when QTBUG-42661 is fixed
```
PatTheMav, line 1818: Does this have negative side-effects once the bug is fixed?

gxalpha: It shouldn’t, no.

Issue #7552, opened 2022-10-07
VLC Media Source Lag


Chromagram:
### Operating System Info

macOS 12

### OBS Studio Version

28.0.3

### OBS Studio Log URL

https://obsproject.com/logs/Ol1XxkvDPm8Qo3eE

### Expected Behavior

When VLC plays through a set list of files, the on screen controls should respond instantly, like a normal file playback should.

### Current Behavior

Currently, pressing any of the buttons like skip, pause, play, stop, etc. completes the action, but not before a minimum of at least 3 seconds of delay, usually longer. This causes plugins like Tuna that display song metadata to change its displayed data instantly as expected, but the music doesn’t actually catch up until many seconds later.

### Steps to Reproduce

1. Add a new VLC media source.
2. Add either a group of songs or a directory into the playlist.
3. Start the playback, and turn on Audio Monitoring to hear the upcoming issue.
4. Maybe wait about 15 seconds, then start trying to control the playback in any way. It will delay.

Opt. If you can, try running this with the Tuna extension as well to see the visual results too. If not, the delay can still be heard without installing Tuna.

### Anything else we should know?

I have been trying to fix this for multiple years, updating OBS and VLC as I go, but the issue persists for some reason. It makes a stream playlist practically useless.

I was hoping with the new macOS Capture, I could just load up a separate player like VLC or IINA and let that play the audio, while I captured the metadata from the player’s window. However, with the restriction of macOS capture still being unable to see anything that isn’t directly under it in windowed mode, it’s also not really an option for me.

I’d love to know if a fix for the VLC source is considerable, or if not, what alternative I can use to have a simple running playlist of music for my stream with some actively running metadata to display. This is dead simple on Windows but an absolute nightmare on Mac.

PatTheMav:
Issue is confirmed, though it only seems to affect the audio monitoring - I did a recording of my VLC source and skipped around the file and every skip was correctly recorded without any delays.

The audio monitoring however buffered over 10 seconds of audio, leading to delayed reaction to skips, stopping playback, and other interactions.

I have no idea how to fix this right now, but it would probably require clearing the circle buffer when such actions are taken by users.

usuallyno:
same problem here with mac

PatTheMav:
Thank you for your comment, but at this point we are not in need of further confirmations but bug fixes. 😅

Martiszz:
Same problem here with the mac

PigDan20:
[...]

Issue #3410, opened 2020-09-04
VLC Video Source unexpected behaviors when having mixed audio and video files in the playlist.


CodeYan01:
## Platform

Operating system and version: Windows 10 Pro 64-bit (Build 19041)
OBS Studio version: 25.0.8 64-bit and 26.0.0-rc1 64-bit

## Expected Behavior
After playing a video file and the next is an audio file, the video should disappear. The audio should then play normally (monitored through internal speakers).

## Current Behavior
The video file keeps the frame where it stopped (either by reaching the end of the video, or using the "Next in Playlist" Button. The audio _seems_ to be captured as the audio meter shows audio levels, but nothing is heard from the monitoring device. Furthermore, if the next one after the audio is another video, a little bit (a little quicker than should be, and has cuts) of the audio file is heard while the video plays. The audio from the video is not heard during this time. Then, the audio file stops and the audio from the video starts (totally unsynced).

## Steps to Reproduce

1. Start OBS
2. Start with a new Scene
3. Add a VLC Video Source
4. Add files to it in this order: 1 video file, 1 audio file, 1 video file
5. In the Advanced Audio Properties, set its Audio Monitoring type to Monitor and Output
6. In Settings > Audio > Monitoring Device, select your internal speakers (or whatever you want to use).
7. Start it, and press the "Next in Playlist" button (or use the hotkey). Waiting for the video to end also does it.
8. The video frame will freeze, while the audio file supposedly works. Nothing is heard, however, even if you skim through it.
9. Press the "Next in Playlist" once more to go the second video file. The video will show properly, but the audio will come from the **audio file** (will be heard). 
10. Let it be like that for a few moments, and the audio from the audio file will stop while the video continues. The audio from the video will start (out of sync). 

## Additional information
I tried both skimming the video and audio, and letting it finish on its own, but the results are the same.

Seems to occur even if I use a .m3u8 file.

The audio issue seems to be affecting only the monitoring, as the audio from the recording seems fine. The video issue (keeping its frame even if its playing the audio file) however affects both.

Justafowl:
I made an account just to report the very issue above, when I found this older post describing it perfectly, yet left dormant.
I can confirm this issue _still persists_ in OBS Studio version 29.1.1 in macOS Ventura 13.3.1 (a) on an M1 Max.

[...]

PR #8147, opened 2023-01-25
Add pause callbacks for coreaudio

usakhelo:
### Description
When playing VLC source on MacOS audio starts lagging after pause, stop or timeline seeking actions.
This is caused by VLC delaying the call to `buffer_audio()` callback. After each pause or stop the delay keeps adding up resulting in audio lag.
This change is adding an additional callback for "media_pause" and "media_stop" events to clean the `monitor->new_data` buffer and prevent audio data accumulation caused by VLC callback delay.

### Motivation and Context
Fixes issue #7552

### How Has This Been Tested?
Tested on MacBook Air M1 with different files in multiple sources including VLC

### Types of changes

### Checklist:

### DIFF
--- a/libobs/audio-monitoring/osx/coreaudio-output.c
+++ b/libobs/audio-monitoring/osx/coreaudio-output.c
@@ -54,6 +54,15 @@ static inline bool fill_buffer(struct audio_monitor *monitor)
 	return true;
 }
 
+static void on_audio_pause(void *data, calldata_t *calldata)
+{
+	UNUSED_PARAMETER(calldata);
+	struct audio_monitor *monitor = data;
+	pthread_mutex_lock(&monitor->mutex);
+	circlebuf_free(&monitor->new_data);
+	pthread_mutex_unlock(&monitor->mutex);
+}
+
 static void on_audio_playback(void *param, obs_source_t *source,
 			      const struct audio_data *audio_data, bool muted)
 {
@@ -261,6 +270,8 @@ static void audio_monitor_free(struct audio_monitor *monitor)
 	if (monitor->source) {
 		obs_source_remove_audio_capture_callback(
 			monitor->source, on_audio_playback, monitor);
+		obs_source_remove_audio_pause_callback(monitor->source,
+						       on_audio_pause, monitor);
 	}
 	if (monitor->active) {
 		AudioQueueStop(monitor->queue, true);
@@ -288,6 +299,8 @@ static void audio_monitor_init_final(struct audio_monitor *monitor)
 
 	obs_source_add_audio_capture_callback(monitor->source,
 					      on_audio_playback, monitor);
+	obs_source_add_audio_pause_callback(monitor->source, on_audio_pause,
+					    monitor);
 }
 
 struct audio_monitor *audio_monitor_create(obs_source_t *source)
--- a/libobs/obs-source.c
+++ b/libobs/obs-source.c
@@ -5732,6 +5732,32 @@ void obs_source_get_audio_mix(const obs_source_t *source,
 	}
 }
 
+void obs_source_add_audio_pause_callback(obs_source_t *source,
+					 signal_callback_t callback,
+					 void *param)
+{
+	if (!obs_source_valid(source, "obs_source_add_audio_pause_callback"))
+		return;
+
+	signal_handler_t *handler = obs_source_get_signal_handler(source);
+
+	signal_handler_connect(handler, "media_pause", callback, param);
+	signal_handler_connect(handler, "media_stopped", callback, param);
+}
+
+void obs_source_remove_audio_pause_callback(obs_source_t *source,
+					    signal_callback_t callback,
+					    void *param)
+{
+	if (!obs_source_valid(source, "obs_source_remove_audio_pause_callback"))
+		return;
+
+	signal_handler_t *handler = obs_source_get_signal_handler(source);
+
+	signal_handler_disconnect(handler, "media_pause", callback, param);
+	signal_handler_disconnect(handler, "media_stopped", callback, param);
+}
+
 void obs_source_add_audio_capture_callback(obs_source_t *source,
 					   obs_source_audio_capture_t callback,
 					   void *param)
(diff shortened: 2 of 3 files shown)


usakhelo:
> Please use one of our template options under "Types of changes" so that we can more easily categorize this PR.

Done.

libobs/audio-monitoring/osx/coreaudio-output.c
```diff
@@ -54,6 +54,13 @@ static inline bool fill_buffer(struct audio_monitor *monitor)
 	return true;
 }
 
+static void on_pause_audio(void *data, calldata_t *calldata)
+{
+	UNUSED_PARAMETER(calldata);
```
PatTheMav, line 59: This should lock and release the monitor mutex as this is executed on the UI thread whereas the other calls happen either on the media IO thread or CoreAudio's own audio thread.

PatTheMav:
On the whole this seems to work as advertised - I guess the issue this fixes is the VLC source not being able to feed new data fast enough before the `buffer_audio` callback empties it again and all buffers are "unused" and it pauses the audio queue again.

libobs/audio-monitoring/osx/coreaudio-output.c
usakhelo: Added mutex lock/unlock

usakhelo:
If I remember correctly `buffer_audio` is not called for a few seconds when VLC source starts playback. When it's called for the first time circle buffer already contains few extra seconds of audio data. On pause or seek VLC delays calling `buffer_audio` for another few seconds. `buffer_audio` keeps calling `fill_buffer` until circle buffer is empty. But because it contains extra audio data pause or seek starts to lag. My fix just empties circle buffer on every pause or seek.

PatTheMav:
IIRC `buffer_audio` is never called by any source - it's a callback method registered with CoreAudio by the macOS monitoring implementation. It will be called by CoreAudio independently of anything else in OBS and will pause its queue as soon as there are 3 empty buffers in the monitoring data struct.

The buffers are filled via `on_audio_playback` (which is called via `vlcs_audio_play`) and what happens is that CoreAudio empties the 3 buffers quicker than `on_audio_playback` can fill them, resulting in a flip-flop between pausing and starting the monitoring queue until a race condition ensures that there are less than 3 empty buffers at any time.

usakhelo:
I can be wrong but I think VLC has some issues opening files on macOS M1. It shows errors in the log. That's why I thought it can be that VLC somehow affects CoreAudio and delays `buffer_audio` calls.
image

BTW I added the mutex lock/unlock. But I didn't rebase the branch. Please let me know if it's needed.

libobs/audio-monitoring/osx/coreaudio-output.c
```diff
@@ -288,6 +296,8 @@ static void audio_monitor_init_final(struct audio_monitor *monitor)
 
 	obs_source_add_audio_capture_callback(monitor->source,
 					      on_audio_playback, monitor);
+	obs_source_add_audio_pause_callback(monitor->source, on_pause_audio,
```
PatTheMav, line 299: The callback is added but never removed (even though the corresponding function is added) - is that on purpose?

usakhelo: That was an overlook. Added removing the callback in `static void audio_monitor_free()` function.

libobs/audio-monitoring/osx/coreaudio-output.c
```diff
@@ -122,7 +131,6 @@ static void on_audio_playback(void *param, obs_source_t *source,
 static void buffer_audio(void *data, AudioQueueRef aq, AudioQueueBufferRef buf)
 {
 	struct audio_monitor *monitor = data;
-
```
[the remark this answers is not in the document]
usakhelo: Reverted in the new commit.

RytoEX (changes requested):
A question about naming.

libobs/audio-monitoring/osx/coreaudio-output.c
```diff
@@ -54,6 +54,15 @@ static inline bool fill_buffer(struct audio_monitor *monitor)
 	return true;
 }
 
+static void on_pause_audio(void *data, calldata_t *calldata)
```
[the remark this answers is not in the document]
usakhelo: I agree, it makes sense. Added another commit.

PatTheMav (approved):
Looks good to me.

PatTheMav:
Final thing before we can merge this:

* Please combine all commits adding the callbacks for libobs into a single commit
* Leave a second commit that implements the functionality to the coreaudio monitoring

usakhelo:
Combined commits. Hope commit naming is correct.

RytoEX (changes requested):
Final nit: the first commit message subject is too long.

Please reword the commit message so that the commit message subject is no longer than 72 characters. Ideally, the commit message subject text after "libobs: " would be 50 characters or less. For additional detail, you can add a commit message body.

usakhelo:
Just a friendly reminder that this PR is ready for merge.

RytoEX:
> Just a friendly reminder that this PR is ready for merge.

We're aware of this PR, thank you. We'll merge it when we have time to circle back around to it and re-review it and find that it's good to merge.

PatTheMav:
Please rename the first commit - it should say something like "libobs: Add pause callback support for sources" because the current title doesn't seem to make sense given the changes.

The second commit should probably be called something like "libobs: Add use of pause callbacks to CoreAudio monitoring".

usakhelo:
Renamed commits, rebased the branch.

Justafowl:
It's been nearly a year since the last update on this potential fix, could this **please** be added to a release version soon? I'm getting very, very tired of having to turn audio monitoring off and on again on VLC Video Source playlists _every time_ a new file in that playlist loads, in order to reset the monitoring sync.

derrod:
This was merged and is part of release versions, what you are experiencing is probably unrelated.

Justafowl:
Then why were #3410 and #7552 closed and merged into this request? The exponential playlist latency in Mac audio monitoring is **still** present.

derrod:
#3410 isn't closed.

PR #8149, opened 2023-01-25
CI,plugins: Remove enc-amf

derrod:
### Description

The legacy AMD encoder has been replaced. We still kept it around for a bit while the new one was getting battle-tested. With one of the next major or minor releases we want to remove it.

### Motivation and Context

One. Less. Submodule.

### How Has This Been Tested?

CI will tell me it's okay I'm sure

### Types of changes

- Code cleanup (~~non~~-breaking change which makes code smaller or more readable)

### Checklist:


gxalpha:
Commit prefix should probably be CI,plugins instead of CI/plugins, the latter would indicate a subdirectory plugins in CI.

WizardCM:
I feel this needs some kind of migration path; I agree we can't just keep this submodule forever. Maybe a shim that migrates users automatically or with a UI flow or some kind to the new one?

derrod:
I've changed the commit prefix.

Migrations might be difficult, the options both encoders offer are vastly different. Perhaps we could at least display a warning *before* a users tries to start a recording/stream (which will fail) that the encoder no longer exists and needs to be updated.

We may also exclude it from removals in the updater/installer for now so that users that already have it will keep having access to it until we decide to fully remove it or it ultimately becomes incompatible.

PR #1254, opened 2018-04-07
UI: Convert properties to tabbed interface

cg2121:
TODO: 
- ~~Try to add hotkeys tab~~
- ~~Fix segfault when opening and closing properties multiple times~~
- ~~Add interface for scenes and transitions~~
As a byproduct, transitions now have filters.
screenshot from 2018-04-07 04-13-14

PR #8155, opened 2023-01-26
UI: Add tabbed properties

cg2121:
### Description
This adds tabbed properties for sources. The tabs are properties, filters and advanced (deinterlace settings).

Properties:
Screenshot from 2023-06-18 03-20-03

Filters:
Screenshot from 2023-06-18 03-24-24

Screenshot from 2023-06-18 03-25-18

Advanced:
Screenshot from 2023-06-18 03-26-08

### Motivation and Context
Redo of #1254

This one is simpler because it only is done for sources, not scene items. Scene items properties can be done in the future.

### How Has This Been Tested?
Opened properties of different sources.

### Types of changes

### Checklist:

### DIFF
--- a/UI/data/locale/en-US.ini
+++ b/UI/data/locale/en-US.ini
@@ -115,6 +115,9 @@ MixerToolbarMenu="Audio Mixer Menu"
 SceneFilters="Open Scene Filters"
 List="List"
 Grid="Grid"
+Mode="Mode"
+AudioVideo="Audio/Video"
+Effects="Effects"
 
 # warning for plugin load failures
 PluginsFailedToLoad.Title="Plugin Load Error"
@@ -556,6 +559,7 @@ Deinterlacing.Yadif="Yadif"
 Deinterlacing.Yadif2x="Yadif 2x"
 Deinterlacing.TopFieldFirst="Top Field First"
 Deinterlacing.BottomFieldFirst="Bottom Field First"
+Deinterlacing.FieldOrder="Field Order"
 
 # volume control accessibility text
 VolControl.SliderUnmuted="Volume slider for '%1':"
--- a/UI/forms/OBSBasicFilters.ui
+++ b/UI/forms/OBSBasicFilters.ui
@@ -1,7 +1,7 @@
 <?xml version="1.0" encoding="UTF-8"?>
 <ui version="4.0">
  <class>OBSBasicFilters</class>
- <widget class="QDialog" name="OBSBasicFilters">
+ <widget class="QWidget" name="OBSBasicFilters">
   <property name="geometry">
    <rect>
     <x>0</x>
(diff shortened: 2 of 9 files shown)


Warchamp7:
I'll try to find time to test but I'm super excited to move towards this change, especially when we can get scene item things added.

Looking at the diff, this removes the Filters button from the context bar. I'd like to hold off on removing that for at least one major release to give users time to learn Filters in part of the properties window. Can that button open to the Filters tab?

How does the defaults button behave in the bottom bar? Does it properly respect the context of what tab you're in and what filter might be selected?

Filters tab is a bit cramped now for source types that support both audio/video filters and effect filters. As a pure hypothetical question, how feasible is it to have each type of filters in their own tab?

cg2121:
> Looking at the diff, this removes the Filters button from the context bar. I'd like to hold off on removing that for at least one major release to give users time to learn Filters in part of the properties window. Can that button open to the Filters tab?

Yes, I will add back the filters button. It will also directly go to the filters tab.

> How does the defaults button behave in the bottom bar? Does it properly respect the context of what tab you're in and what filter might be selected?

The defaults button resets whatever the current tab is.

> Filters tab is a bit cramped now for source types that support both audio/video filters and effect filters. As a pure hypothetical question, how feasible is it to have each type of filters in their own tab?

It would probably be possible, but it would take a lot more code. I got some ideas that might work, with just re-arranging few a things in the filters.

derrod:
> Filters tab is a bit cramped now for source types that support both audio/video filters and effect filters. As a pure hypothetical question, how feasible is it to have each type of filters in their own tab?

Tabs within tabs, all the way down

cg2121:
I have updated the filters tab to have the async and effect lists to be tabs.

Screenshot from 2023-01-28 12-31-16

Warchamp7:
Tabs within an already tabbed navigation feels like one of those scenarios where you're made a mistake somewhere.

> It would probably be possible, but it would take a lot more code. I got some ideas that might work, with just re-arranging few a things in the filters.

If it's possible, I'd prefer to go down that path

Wiethoofd:
I would go even further with adding more tabs for individual settings on the sources, the show/hide transition overrides, scale filters and the 'Edit Transform' dialog values could get a tab.

This would potentially result in 2 places where you can find the properties for a source, having the tabs in a single properties dialog as well as the individual popup dialogs or context menu foldouts;, could do as suggested by making the current context menu options open the correct tab in the new properties dialog, which would get rid of duplicate info dialogs.

Warchamp7:
> I would go even further with adding more tabs for individual settings on the sources, the show/hide transition overrides, scale filters and the 'Edit Transform' dialog values could get a tab.
> 
> This would potentially result in 2 places where you can find the properties for a source, having the tabs in a single properties dialog as well as the individual popup dialogs or context menu foldouts;, could do as suggested by making the current context menu options open the correct tab in the new properties dialog, which would get rid of duplicate info dialogs.

This is part of the eventual goal

cg2121:
Updated to make async and effect filters separate tabs.

Screenshot from 2023-06-18 03-20-03

cg2121:
Closing, until we have a better idea of what this should look like.

PR #8161, opened 2023-01-27
UI: Use native combobox popup on macOS

gxalpha:
### Description

Sets the `SH_ComboBox_UseNativePopup` to true on macOS. By doing this, Qt's combobox popups get replaced by native macOS ones. Before/After:
<p>
image
image
</p>

This flag currently only has an effect on macOS, but it's still probably safer to ifdef-guard it.

As recommended by Qt on QTBUG-106162, this fixes QComboBox popups not working with VoiceOver.
Together with Qt 6.5 (which fixes comboboxes not opening in the first place, see [QTBUG-106162](https://bugreports.qt.io/browse/QTBUG-106162)) this should make non-editable comboboxes work flawlessly.

Renamed `OBSIgnoreWheelProxyStyle` to `OBSProxyStyle` since it's no longer exclusive to ignoring wheel scrolling.

### Motivation and Context

Want OBS to be more accessible. Making VoiceOver work properly is a long and difficult (maybe impossible) road, but every step counts.

### How Has This Been Tested?

macOS 13.1
Tested with Qt dev branch and this PR that comboboxes work well now with VoiceOver.

### Types of changes

### Checklist:

### DIFF
--- a/UI/obs-app.cpp
+++ b/UI/obs-app.cpp
@@ -1242,6 +1242,10 @@ bool OBSApp::SetTheme(std::string name, std::string path)
 		themeDarkMode = !(color.redF() < 0.5);
 	}
 
+#ifdef __APPLE__
+	SetMacOSDarkMode(themeDarkMode);
+#endif
+
 	emit StyleChanged();
 	return true;
 }
--- a/UI/obs-app.cpp
+++ b/UI/obs-app.cpp
@@ -1249,7 +1249,7 @@ bool OBSApp::SetTheme(std::string name, std::string path)
 bool OBSApp::InitTheme()
 {
 	defaultPalette = palette();
-	setStyle(new OBSIgnoreWheelProxyStyle());
+	setStyle(new OBSProxyStyle());
 
 	const char *themeName =
 		config_get_string(globalConfig, "General", "CurrentTheme3");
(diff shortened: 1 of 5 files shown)


UI/obs-app.cpp
```diff
@@ -1249,7 +1249,7 @@ bool OBSApp::SetTheme(std::string name, std::string path)
 bool OBSApp::InitTheme()
 {
 	defaultPalette = palette();
-	setStyle(new OBSIgnoreWheelProxyStyle());
+	setStyle(new OBSProxyStyle());
```
WizardCM, line 1252: This reverts an intentional design decision to stop users from accidentally scrolling combo boxes when trying to scroll the page they're in.

gxalpha: It doesn't revert it (I was the one who wrote the code to stop users from wheel scrolling so I made sure that that doesn't happen :P).
I just renamed the ProxyStyle class since it's no longer exclusive to ignoring wheel scroll (so the old name would have been misleading), probably should have mentioned that in the PR though.

PatTheMav:
Looks fine to me - supplied a patch off-thread to make the native appearance match the theme (dark vs light) which would otherwise look a bit jarring.

gxalpha:
Added the patch (modified to use the correct variable (`themeDarkMode` instead of `themeMeta->dark`, the existence of `themeMeta` isn't guaranteed with third-party overlays and would crash otherwise) and to just pass a bool).

PR #8171, opened 2023-01-29
Linux: virtual audio support via ALSA Loopback card

aramg:
NOTE: This PR includes the changes in #8008, which is expected to get merged first.

### Description
Adds virtual audio support via the ALSA loopback module, `snd_aloop`. As implemented, audio is optional and the virtualcam depends on v4l2loopback still. `snd_aloop` is not auto-loaded and an audio-only mode is not available.

### Motivation and Context
Allows sharing audio output with other programs as a virtual microphone.

### How Has This Been Tested?

1. Load the loopback modules: `modprobe v4l2loopback` `modprobe snd-aloop`
2. Start OBS and click [Start Virtual Cam]. Both audio and video are activated:
```
info: v4l2-output: Using device 'Dummy video device (0x0000)' at '/dev/video0'
info: ALSA output: hw:Loopback,1,0
info: Virtual camera starting: audio=1 video=1
```

3. Add a media source, and observe video output as before (in this case, via '/dev/video0' **).
4. Record audio: ` arecord -D hw:Loopback,1,0 --duration=30 -fS16_LE -c2 -r48000 /tmp/test.pcm`
5. Replay audio: ` aplay -fS16_LE -c2 -r48000 /tmp/test.pcm`
6. This varies by distro, but to load the device into PluseAudio:
`pacmd load-module module-alsa-source source_properties=device.description=OBS  device=hw:Loopback,1,0`

** ideally the audio and video device names will be added to the OBS status bar, a to-do.

### Types of changes

### Checklist:


david672orford:
I have been testing this PR. Here are some initial observations:

* This PR includes #8008 which is not really related. Worse, it changes behavior and is broken. When the virtual camera is started, it attempts to create a new V4L2 loopback interface called "OBS Virtual Camera". This produces nuisance pkexec authorization popup for a command which ultimately fails do to the use of a non-existent "add" subcommand.
* While it makes a kind of sense that since the video output goes to an in-kernel loopback devices, the audio should too, I am not sure this is the best decision. The only monitoring output for Linux I see in the code goes to Pulseaudio. Shouldn't this use Pulseaudio as well?
* When I play a video file I hear frequent clicking.

david672orford:
I am really interested in getting the main audio output working. You can get by with the monitor output, but it bypasses the built-in mixer in OBS and there are no fades for scene transitions.

Audio for the virtual camera was requested nearly three years ago:

https://ideas.obsproject.com/posts/1415/obs-virtual-camera-audio

I don't really understand the response from Lain about why a satisfactory implementation is not currently possible. When I asked for details of what needed to be done, I was told that "someone willing, and available, to spend time to do it". I found that completely uninformative.

Your patch is the closest thing we have to an implementation. I wish we could find a way to move it forward. I can program in C, but if nobody is willing to discuss whatever is blocking a properly-working implementation, I'm stuck. Maybe you have some insights.

david672orford:
Wanted to follow up on the clicking. After more tests I find that when I read audio directly from the Alsa loopback device there is no clicking. I get the clicking when I read the audio through PipeWire.

david672orford:
Just posted #10495 which implements audio output through PulseAudio. Look a Lua script in the PR decription which starts and stops a virtual audio cable when the virtual camera starts and stops.

Issue #2855, opened 2020-05-02
[BUG] Multi track audio recording generates empty output file at high FPS


notr1ch:
## Platform

Operating system and version: Windows 10 1909
OBS Studio version: 25.0.8

## Expected Behavior
Recordings at high FPS should function normally.

## Current Behavior
The OBS `audio_thread` is single threaded - it processes and encodes each track sequentially, encoding and sending packets to be interleaved. `audio_time` gets progressively more and more behind `cur_time`, the user has no feedback this is occurring and trying to stop the output gets stuck since `audio->stop_event` is not checked due to the backlog. Meanwhile, the video thread is still outputting packets to be interleaved as the outputs are only stopped in `ffmpeg_mux_data` once `packet->sys_dts_usec >= stream->stop_ts`.

High FPS seems needed to reproduce this, the high rate of packets being sent to interleave is likely causing bottlenecks in the audio thread, as each audio packet must iterate the array of interleaved packets once per track.

Despite the audio packets eventually catching up, the output file is completely trashed - in testing I only got a 1.4KB MKV. I've not yet looked into why this happens.

## Steps to Reproduce
1. Set OBS to 240 FPS, 6 audio tracks in advanced output.
2. Record for 30-60 seconds.
3. Stop recording.

## Additional information
Couple of potential fixes - making audio encoding multi-threaded seems like a good place to start, however the video thread continuing to send packets even after the output has been stopped should also be investigated.

Xaymar:
Is this perhaps the same underlying issue that causes these old Mantis issues where not even an output file is generated, or is this a different issue? Talking about any of these:

- https://obsproject.com/mantis/view.php?id=789
- https://obsproject.com/mantis/view.php?id=953 (ignore my comments from then, I was completely wrong)
- https://obsproject.com/mantis/view.php?id=1268
- https://obsproject.com/mantis/view.php?id=1348
- (I'm sure there's more of these, but the titles and description of the Mantis issues is a lacking a lot of info.)

lizardpeter:
Yes, this issue has been driving me CRAZY! Sometimes it works and sometimes it doesn't. I just can't have that kind of inconsistency and have to guess if it's actually recording or not. It sucks because OBS with a single track at a high frame rate works for me, but I have wanted to start using multiple audio tracks for the sake of keeping game and voice audio separate for use in editing.

lizardpeter:
[...]

PR #8175, opened 2023-01-30
libobs: Reduce synchronization limit for multiple audio tracks

kkartaltepe:
### Description

Since 65eb3c0815929866dde7d398bec7738d21e9b211 we tried to get as close a sync between audio and video tracks as we could before starting to send frames to the output.

But in be717dbb2c19aa0e91a164fa52e3d4467b64be9d when multi-track audio was considered for synchronization it continued to try and line ALL audio packets up with one video packet. While audio and video packets are similar size this will work out. But once video packets duration is smaller than 1/2 audio packet duration this may fail forever. In practice with AAC's 20-23ms frame duration we can often get tracks producing frames 5-10ms out of phase. For 60fps video this mostly fine as 16ms frame duration will cover the gap, but for framerates as low as 100 its possible to fail to synchronize. In practice this is ~50% for 6 audio tracks and 120fps video on my system, or 100% at 240fps.

Since this doesnt change the logic for single audio tracks I think it's relatively safe but you can never be too safe around audio sync issues.

### Motivation and Context

No one likes when OBS eats a recording. Largely mitigates #2855 but by just picking even the worst possible starting frame. Ideally we could reduce the phase difference in the audio encoders and continue using the old logic. This also doesnt solve the issue with us marking the recording as started despite it not actually outputting any frames while waiting for the sync point.

### How Has This Been Tested?

I started and stopped a lot of recordings with a fast hardware encoder and 6 audio tracks enabled. If the video encoder is too slow this may not occur as I couldnt get it to happen with software encoders.

### Types of changes

### Checklist:

### DIFF
--- a/libobs/obs-output.c
+++ b/libobs/obs-output.c
@@ -1388,14 +1388,21 @@ static size_t get_interleaved_start_idx(struct obs_output *output)
 	return video_idx < idx ? video_idx : idx;
 }
 
+static int64_t get_encoder_duration(struct obs_encoder *encoder)
+{
+	return (encoder->timebase_num * 1000000LL / encoder->timebase_den) *
+	       encoder->framesize;
+}
+
 static int prune_premature_packets(struct obs_output *output)
 {
 	struct encoder_packet *video;
 	int video_idx;
 	int max_idx;
-	int64_t duration_usec;
+	int64_t duration_usec, max_audio_duration_usec = 0;
 	int64_t max_diff = 0;
 	int64_t diff = 0;
+	int audio_encoders = 0;
 
 	video_idx = find_first_packet_type_idx(output, OBS_ENCODER_VIDEO, 0);
 	if (video_idx == -1) {
@@ -1410,9 +1417,11 @@ static int prune_premature_packets(struct obs_output *output)
 	for (size_t i = 0; i < MAX_OUTPUT_AUDIO_ENCODERS; i++) {
 		struct encoder_packet *audio;
 		int audio_idx;
+		int64_t audio_duration_usec = 0;
 
 		if (!output->audio_encoders[i])
 			continue;
+		audio_encoders++;
 
 		audio_idx = find_first_packet_type_idx(output,
 						       OBS_ENCODER_AUDIO, i);
@@ -1428,6 +1437,21 @@ static int prune_premature_packets(struct obs_output *output)
 		diff = audio->dts_usec - video->dts_usec;
 		if (diff > max_diff)
 			max_diff = diff;
+
+		audio_duration_usec =
+			get_encoder_duration(output->audio_encoders[i]);
+		if (audio_duration_usec > max_audio_duration_usec)
+			max_audio_duration_usec = audio_duration_usec;
+	}
+
+	/* Once multiple audio encoders are running they are almost always out
+	 * of phase by ~Xms. If users change their video to > 100fps then it
+	 * becomes probable that this phase difference will be larger than the
+	 * video duration preventing us from ever finding a synchronization
+	 * point due to their larger frame duration. Instead give up on a tight
+	 * video sync. */
+	if (audio_encoders > 1 && duration_usec < max_audio_duration_usec) {
+		duration_usec = max_audio_duration_usec;
 	}
 
 	return diff > duration_usec ? max_idx + 1 : 0;


kkartaltepe:
And an example bad audio encoder frame generation series is https://gist.github.com/kkartaltepe/d9bc973b03eda915829d4df10f289439

At the first attempted sync (`Pruning! 0`) we have

```
debug: packet: audio 1, ts: 149505163945, pruned = false, size = 3
debug: packet: audio 2, ts: 149505163945, pruned = false, size = 3
debug: packet: audio 3, ts: 149505163945, pruned = false, size = 3
debug: packet: audio 4, ts: 149505163945, pruned = false, size = 3
debug: packet: audio 5, ts: 149505163945, pruned = false, size = 3
debug: packet: audio 0, ts: 149505185226, pruned = false, size = 524
debug: packet: video 0, ts: 149505187726, pruned = false, size = 25586
```

Here it seems the packets are out of phase by 21.281ms which is basically the worst case but seems fairly common in my testing at least.  And we choose this audio/video packet as our possible sync point and begin pruning packets.

 later after some more audio packets we end up with another possible sync point at

```
debug: packet: audio 1, ts: 149505183945, pruned = false, size = 3
debug: packet: audio 2, ts: 149505183945, pruned = false, size = 3
debug: packet: audio 3, ts: 149505183945, pruned = false, size = 3
debug: packet: audio 4, ts: 149505183945, pruned = false, size = 3
debug: packet: audio 5, ts: 149505183945, pruned = false, size = 3
debug: packet: audio 0, ts: 149505185226, pruned = false, size = 524
debug: packet: video 0, ts: 149505187726, pruned = false, size = 25586
debug: packet: video 0, ts: 149505191892, pruned = false, size = 2164
```

And we see the phase issue, audio encoders 1-5 are all in sync but audio 0 is out of phase by 1.281ms and audio 0 and video 0 are out sync by 2.500ms. This one might work if we were smart but we clear the interleaving array up to the audio 0 sample, removing the earlier audio 1/2/3/5 that were without our hopeful 4ms window. So this sync point fails and we need to wait for the next samples, which we saw earlier was 21ms out of phase and thus we never find a valid <4ms (for 240fps) sync point. (see pruning occurring at `Pruning! 10` and `Pruning! 11`)

If we wanted to make the bigger change to try and catch this better sync point where the order of audio 0/1/2/3/4/5 might be different that would good too but likely a bit more complicated than I care to do within the audio sync system.

flaeri:
Tested on win10, 240fps, 6 audio tracks. Normally the recording will die very quickly with this setup.
I did 1 recording attempt with lots of render lag, and it did not die. I then did 4 more recordings, up to 5 min long, and none of them died.

This is looking very promising in so far as that problem goes (muxer/encoder hang, or whatever it is). On a cursory glance, audio sync is looking no worse for wear.

norihiro:
Frame duration comes from 1024 samples for each packet; 1024 / 48kHz = 21.3 ms or 1024 / 44.1kHz = 23.2 ms.
Did you intentionally put the magic number `1024 * 21`?

tytan652:
Nitpik on the commit name `Reduce` with a uppercase `R`.

kkartaltepe:
> Frame duration comes from 1024 samples for each frame; 1024 / 48kHz = 21.3 ms or 1024 / 44.1kHz = 23.2 ms. Did you intentionally put the magic number `1024 * 21`?

Yes the AAC_FRAME_DURATION_USEC is in microseconds, so the 21 comes from 1/48kHz=20.8333us. It possible this is too short for 44.1kHz Ill test and see maybe we need to be more conservative.

norihiro:
> It possible this is too short for 44.1kHz Ill test and see maybe we need to be more conservative.

Instead of having a hard coded constant, cann't we use `1000000 / output->audio->info->samples_per_sec`?

kkartaltepe:
> Instead of having a hard coded constant, cann't we use 1000000 / output->audio->info->samples_per_sec?

We could but this is a per encoder thing, hardcoding (hopefully) makes it clear its not really sustainable if we change the way we do multiple encoders. We can instead query every encoder and pick the largest period if we are ok with a larger change.

norihiro:
> We can instead query every encoder and pick the largest period if we are ok with a larger change.

I misunderstood the motivation of this PR. Sounds like current hard-code implementation is sufficient enough.

tytan652:
Is it really a good idea to hardcode with an AAC value when in the future (while ignoring FTL) using an Opus encoder will be possible for recording, SRT and RIST ?

derrod:
> Is it really a good idea to hardcode with an AAC value when in the future (while ignoring FTL) using an Opus encoder will be possible for recording, SRT and RIST ?

Opus uses a frame size of 20ms by default, which is probably *close enough* for now.

kkartaltepe:
> Is it really a good idea to hardcode with an AAC value when in the future (while ignoring FTL) using an Opus encoder will be possible for recording, SRT and RIST ?

None of these are supported, if/when we do support multiple audio codecs in the future when we do that would be a good time to revisit this code. But also AAC has the largest frame size of currently available codecs at 1024 samples. Opus is only 960 samples at it default period.

prgmitchell:
I tested at 240fps with 6 audio tracks and as mentioned a few messages above, this would normally be a guaranteed way to crash the encoder/muxer on my system. I was able to let the recording run for about an hour to test with no issues and no crashes.

kkartaltepe:
> > It possible this is too short for 44.1kHz Ill test and see maybe we need to be more conservative.
> 
> Instead of having a hard coded constant, cann't we use `1000000 / output->audio->info->samples_per_sec`?

I have updated it to grab the actual max duration from the audio encoders instead of hardcoding AAC's framesize and 48khz.

energizerfellow:
Would there be any issues with enabling 96 kHz at some future date with the new sync logic?

kkartaltepe:
> Would there be any issues with enabling 96 kHz at some future date with the new sync logic?

The current version is no longer hardcoded this should be fine for whatever (baring numerical/precision issues).

PR #8179, opened 2023-01-30
libobs: Fix obs_remove_data_path memory leak

orgcrab:
### Description

This change adjusts fix obs_remove_data_path. if developer want to add custom path for core_module_paths, would memory leak. the last one item of core_module_paths can not clear.

### Motivation and Context

When sorting out the directory structure, I found obs_add_data_path can add a path to search libobs data files.
and remove path would leak memory.

### How Has This Been Tested?

1. use obs_add_data_path to add one or more path
2. remove path by obs_remove_data_path. the last one item of core_module_paths can not clear, then memory leak.

### Types of changes

Fix bug (non-breaking change to improve existing functionality)

### Checklist:

### DIFF
--- a/libobs/obs.c
+++ b/libobs/obs.c
@@ -1190,7 +1190,10 @@ bool obs_remove_data_path(const char *path)
 
 		if (result == 0) {
 			dstr_free(&core_module_paths.array[i]);
-			da_erase(core_module_paths, i);
+			if (core_module_paths.num == 1)
+				da_free(core_module_paths);
+			else
+				da_erase(core_module_paths, i);
 			return true;
 		}
 	}


notr1ch:
This seems specific to one instance of a darray. A darray does not release the array even when elements are removed in order to avoid continuous allocation and freeing of memory since they are often used in patterns where frequent insertion and removal occurs. As such I don't think this change is appropriate - either all darrays should behave this way or `core_module_paths` should be cleaned up elsewhere.

orgcrab:
Thanks, i agree with you. This changed is not best way, just for the sake of fix core_module_paths (memory leak). 
My english is average, if i have any grammatical mistakes, please forgive me. After your prompt, i will think about the problem more comprehensively in the future.

Lain-B:
Were you actually using `obs_add_data_path()` and `obs_remove_data_path()`? Honestly I'm not sure why I added those functions considering we don't use them.

PR #8180, opened 2023-01-30
rtmp-services: Update Switchboard Live Servers & Name

one0ten:
### Description

The Switchboard Live Server URLs have been recently updated in the application, so the URLs need to be updated in OBS to match.

### Motivation and Context

There were recent changes to the ingestion servers  at Switchboard Live

### How Has This Been Tested?

This was tested using the latest build of OBS on Windows.
Testing consisted of going live with every server to ensure connectivity  was stable.

### Types of changes

### Checklist:

### DIFF
--- a/plugins/rtmp-services/data/services.json
+++ b/plugins/rtmp-services/data/services.json
@@ -1107,51 +1107,17 @@
             ],
             "servers": [
                 {
-                    "name": "Global Zone (geo based)",
+                    "name": "Global - Recommended",
+                    "url": "rtmps://live.sb.zone:443/live"
+                },
+                {
+                    "name": "Global - Legacy",
                     "url": "rtmp://ingest-global.switchboard.zone/live"
-                },
-                {
-                    "name": "US Zone (geo based)",
-                    "url": "rtmp://ingest-us.switchboard.zone/live"
-                },
-                {
-                    "name": "US West",
-                    "url": "rtmp://ingest-us-west.switchboard.zone/live"
-                },
-                {
-                    "name": "US East",
-                    "url": "rtmp://ingest-us-east.switchboard.zone/live"
-                },
-                {
-                    "name": "US Central",
-                    "url": "rtmp://ingest-us-central.switchboard.zone/live"
-                },
-                {
-                    "name": "South America",
-                    "url": "rtmp://ingest-sa.switchboard.zone/live"
-                },
-                {
-                    "name": "EU (auto)",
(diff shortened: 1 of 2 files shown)


plugins/rtmp-services/data/services.json
```diff
@@ -1103,55 +1103,21 @@
         {
             "name": "Switchboard Live",
             "alt_names": [
-                "Switchboard Live (Joicaster)"
+                "Switchboard Live"
```
RytoEX, line 1106: I'm pretty sure this will break user configurations if they still have the old Service dropdown entry selected. `alt_names` is used to leave the Service selected for the user, and fetching the `servers` list attached.

RytoEX (changes requested):

plugins/rtmp-services/data/services.json
one0ten: For old selected URLs, we'll create CNAMEs. I've also reverted the `alt_names` change

one0ten: We're also informing our OBS users about this update to make necessary changes

Fenrirthviti: This isn't about your service, but how it appears in OBS. You can change the name, you just need the old one to be listed in `alt_names`, which is invisible to the user and only used internally for anyone with an old config to know what service is selected.

one0ten: Understood. I reverted the changes to the `alt_names`

RytoEX (changes requested):
Please squash the commits into a single commit.

one0ten:
> Please squash the commits into a single commit.

done

plugins/rtmp-services/data/services.json
```diff
@@ -1107,51 +1107,17 @@
...
+                    "name": "Global - Legacy",
+                    "url": "rtmp://ingest-global.switchboard.zone/live"
+                }                            
             ],
             "recommended": {
                 "keyint": 2,
-                "profile": "high",
-                "max audio bitrate": 160,
-                "max video bitrate": 6000
+                "profile": "high"
```
gxalpha, line 1120: Just to confirm since you didn't mention them anywhere else, is the removal of these bitrate restrictions intentional?

one0ten: yes, removal is intentional

Fenrirthviti (approved):

PR #8190, opened 2023-02-01
CI/docs: Some fixes for CF Pages publishing

derrod:
### Description

- Disables beta/rc tags being published to live page
- Runs version update step for `workflow_dispatch` as well
- Sphinx output is now created without `.html` suffix in Links to avoid unnecessary redirects on CF's platform

Note: This would break browing the docs locally unless they're hosted on a webserver that is configured to support extension-less URLs.

### Motivation and Context

Fixing good.

### How Has This Been Tested?

Tested on my fork: https://obs-docs-test.pages.dev/

### Types of changes

### Checklist:

### DIFF
--- a/.github/workflows/docs.yml
+++ b/.github/workflows/docs.yml
@@ -34,7 +34,7 @@ jobs:
 
       - name: Bump Version Number
         shell: bash
-        if: github.event_name == 'push'
+        if: github.event_name != 'pull_request'
         run: |
           VERTEST="\#define\sLIBOBS_API_\w+_VER\s([0-9]{1,2})"
           VER=""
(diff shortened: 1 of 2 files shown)


RytoEX:
> Note: This would break browing the docs locally unless they're hosted on a webserver that is configured to support extension-less URLs.

This is unfortunate. If I'm checking docs builds, I load them locally in a browser without running a webserver. @WizardCM thoughts?

derrod:
I think there are two ways to keep offline docs working:

- Only use extension-less links on `workflow_disaptch` or `tags` so normal pushes create regular docs
- Generate second set of docs with different settings for publishing only

WizardCM:
@derrod I vote for the second option. Docs should always generate user-usable ZIPs, including alongside CF Pages runs.

derrod:
Updated based on @WizardCM's and @RytoEX's feedback so that the workflow generates a separate artifact for CF pages and the normal artifact remains locally browseable.

RytoEX (approved):
Seems fine to me.

derrod:
> Artifact has HTML files, though the version in the docs is seemingly outdated.

The step to update the version number doesn't run for PRs. Not sure why that was originally done.

RytoEX:
> > Artifact has HTML files, though the version in the docs is seemingly outdated.
> 
> The step to update the version number doesn't run for PRs. Not sure why that was originally done.

Then I suppose it's a non-issue. Will merge shortly.

PR #8208, opened 2023-02-04
obs-filters: Ensure gain is positive for upward compressor

pkviet:
### Description
The gain should be positive for an upward compressor. Initially, the gain would be zeroe'd below -60 dB to ensure noise is not amplified by the upward compressor. This created a discontinuity at -60 dB since just above -60 dB, the audio is boosted by +20 dB (at default settings). This was fixed in commit 50db097 which decreased smoothly the gain so that it's 0 dB at -60 dB. However that commit forgot to limit the gain decrease which was negative below -60 dB. This is fixed by the current commit.
Additionally initialization allowed -inf gain. We hard limit to positive gains only as they should be for an upward compressor. 

### Motivation and Context
Bug found in audio-support, confirmed by @RytoEX and @pkviet .
Fix found by @notr1ch and put in form by @pkviet 

### How Has This Been Tested?
Tested against a test tone.
Added a new upward compressor: checked the filter works.
Restarted obs; checked the filter keeps working.
Previously, the filter would work on initial addition but would fail upon restart of obs due to
a bad initialization propagating -inf gains which are forbidden for the upward compressor.

### Types of changes

### Checklist:

### DIFF
--- a/plugins/obs-filters/expander-filter.c
+++ b/plugins/obs-filters/expander-filter.c
@@ -360,12 +360,15 @@ static inline void process_sample(size_t idx, float *samples, float *env_buf,
 	float diff = threshold - env_db;
 
 	if (is_upwcomp && env_db <= (threshold - 60.0f) / 2)
-		diff = env_db + 60.0f;
+		diff = env_db + 60.0f > 0 ? env_db + 60.0f : 0.0f;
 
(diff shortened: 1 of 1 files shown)


pkviet:
The bug reporter on discord confirms the bug is fixed.

PR #8217, opened 2023-02-05
Add option to fully decode and preload stinger video files into RAM

jp9000:
### Description
Adds the ability to fully preload stinger videos to RAM before playback, greatly improving stinger playback performance.

### Motivation and Context
Stingers -- and especially track matte stingers -- are subject to real time decoding, thus if there are not enough computing power available, decoding can lag, causing stingers to play back slowly or stutter, which just kind of ruins a nice broadcast. It's even worse when track matte stingers are in use, because they require twice the decoding power of normal videos.

To solve this, add an option to fully decode and preload video files into RAM to completely eliminate decoding from the equation, and ensure smooth playback of the stinger's video and greatly improving stinger playback performance (at the cost of RAM usage, of course).

### How Has This Been Tested?
Tested with various stinger transition and track matte stinger video files.

### Types of changes

### Checklist:

### DIFF
--- a/deps/media-playback/CMakeLists.txt
+++ b/deps/media-playback/CMakeLists.txt
@@ -9,8 +9,14 @@ add_library(OBS::media-playback ALIAS media-playback)
 
 target_sources(
   media-playback
-  INTERFACE media-playback/media.c media-playback/media.h
-            media-playback/decode.c media-playback/decode.h
+  INTERFACE media-playback/media.c
+            media-playback/media.h
+            media-playback/cache.c
+            media-playback/cache.h
+            media-playback/decode.c
+            media-playback/decode.h
+            media-playback/media-playback.c
+            media-playback/media-playback.h
             media-playback/closest-format.h)
 
 target_link_libraries(media-playback INTERFACE FFmpeg::avcodec FFmpeg::avdevice
--- a/deps/media-playback/media-playback/cache.c
+++ b/deps/media-playback/media-playback/cache.c
@@ -1,0 +1,698 @@
+#include <media-io/audio-io.h>
+#include <util/platform.h>
+
+#include "media-playback.h"
+#include "cache.h"
+#include "media.h"
+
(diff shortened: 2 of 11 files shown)


flaeri:
Seems to be a potential regression of the previous frame flicker (f48795439504ac). Can usually get it in less than 5 attempts.

Could not reproduce the flicker without preload. Tried about 10-20 transitions.
Log: https://obsproject.com/logs/Gox6wFa703S0TyED
Video: 

https://user-images.githubusercontent.com/50419942/217084331-0dc00f43-481f-4f18-acc1-ec14ee0174b4.mp4

jp9000:
I looked at that commit, was surprised to see my name on it as the author. Cue the old Gandalf meme: "...I have no memory of this place"

I'll take a look at it though.

jp9000:
Okay, I think I've fixed the issue. Probably needs a lot more testing, both with preload on and off.

flaeri:
Testing the new build (from artifacts).

So far, without caching/preload enabled, I've seen this happen twice. The stinger will get "stuck". Its the same transition I used last time (and you have), you might recognize it.
https://user-images.githubusercontent.com/50419942/218340683-ffdfe725-52e5-4a91-b061-c786c302b6b4.mp4

The 2nd issue I've ran into, which happens both with and without preload is the following:

I'm on scene 1. Im going to scene 2.
I start the transition, I see scene one, then it it will flash the destination scene (2) for a single, return to source scene (1), and then start the transition, and successfully transition. 

See slowed down example:
 https://user-images.githubusercontent.com/50419942/218340910-365abf04-25bd-465c-9732-ad54872f4c28.mp4
I am just slowing going frame by frame forwards.

flaeri:
After the last update [988af7b](https://github.com/obsproject/obs-studio/pull/8217/commits/988af7b743848f464feaba37b0d6520155d64001) it now seems rock solid, both with and without preload.

PR #8220, opened 2023-02-06
A first foray into reproducible builds

derrod:
### Description

Makes Windows builds *mostly* reproducible using undocumented (but working) MSVC compiler/linker options and dropping the Qt Resource Compiler version to 1 (since 2+ includes last-modified timestamps).

PDBs won't be identical due to lambdas having random identifiers but they are compatible, meaning that it would be possible to also create debug symbols for any build after the fact.

ToDo:
- [x] Figure out how to make cmake behave when it comes to escaping the `/PDBALTPATH` linker option
  + CMake bug: https://gitlab.kitware.com/cmake/cmake/-/issues/15865 (7 years old D:)
  + Possible workaround: https://github.com/boostorg/stacktrace/issues/55#issuecomment-984782636
- [x] Figure out why `coreaudio-encoder` is not currently reproducible
  + Seems to be fixed by #8377 

### Motivation and Context

We eventually want to have fully reproducible builds for OBS and its dependencies. For the reason why you may want that, see https://reproducible-builds.org/docs/buy-in/

Aside from the security implications and verifiability, a side-benefit of reproducible builds would be that delta patches as we employ on Windows (and possibly macOS in the future) would be smaller, and determining which files haven't changed to exclude them from an update could be more easily automated. (Note: Ensuring that timestamps are always the same even for adjacent builds [ducible](https://github.com/jasonwhite/ducible) may still be a required, it can be added to CMake for example like so: https://github.com/derrod/obs-studio/commit/5f4dee7423f50ea47f073831fdd7165cd1d10e67)

Getting OBS itself into a state where building it with our precompiled Windows and macOS dependencies will create identical binaries is a first step.

### How Has This Been Tested?

Ran various tests on my fork, for the same commit, all binaries except `obs-websocket` and `coreaudio-encoder` generate identical files (sha512 checked).

A workflow run with only two x64 Windows jobs that have produced mostly-identical builds is here: https://github.com/derrod/obs-studio/actions/runs/4099136967

### Types of changes

### Checklist:

### DIFF
--- a/UI/CMakeLists.txt
+++ b/UI/CMakeLists.txt
@@ -82,6 +82,10 @@ set_target_properties(
              AUTORCC ON
              AUTOUIC_SEARCH_PATHS "forms;forms/source-toolbar")
 
+if(_QT_VERSION EQUAL 6 AND OS_WINDOWS)
+  set_target_properties(obs PROPERTIES AUTORCC_OPTIONS "--format-version;1")
+endif()
+
 target_include_directories(obs PRIVATE ${CMAKE_CURRENT_SOURCE_DIR}
                                        ${CMAKE_CURRENT_BINARY_DIR})
 
--- a/cmake/Modules/CompilerConfig.cmake
+++ b/cmake/Modules/CompilerConfig.cmake
@@ -52,6 +52,7 @@ if(OS_WINDOWS AND MSVC)
   endif()
 
   add_compile_options(
+    /Brepro
     /MP
     /W3
     /WX
@@ -75,6 +76,7 @@ if(OS_WINDOWS AND MSVC)
     /std:c17)
 
   add_link_options(
+    "LINKER:/Brepro"
     "LINKER:/OPT:REF"
     "LINKER:/WX"
     "$<$<NOT:$<EQUAL:${CMAKE_SIZEOF_VOID_P},8>>:LINKER\:/SAFESEH\:NO>"
--- a/cmake/Modules/ObsHelpers_Windows.cmake
+++ b/cmake/Modules/ObsHelpers_Windows.cmake
@@ -29,6 +29,9 @@ function(setup_binary_target target)
   endif()
 
   if(MSVC)
+    target_link_options(${target} PRIVATE
+                        /PDBALTPATH:$<TARGET_PDB_FILE_NAME:${target}>)
+
     install(
       FILES $<TARGET_PDB_FILE:${target}>
       CONFIGURATIONS "RelWithDebInfo" "Debug"
@@ -56,6 +59,9 @@ function(setup_plugin_target target)
   _setup_plugin_target(${ARGV})
 
   if(MSVC)
+    target_link_options(${target} PRIVATE
+                        /PDBALTPATH:$<TARGET_PDB_FILE_NAME:${target}>)
+
     install(
       FILES $<TARGET_PDB_FILE:${target}>
       CONFIGURATIONS "RelWithDebInfo" "Debug"
@@ -97,6 +103,9 @@ function(setup_script_plugin_target target)
   _setup_script_plugin_target(${ARGV})
 
   if(MSVC)
+    target_link_options(${target} PRIVATE
+                        /PDBALTPATH:$<TARGET_PDB_FILE_NAME:${target}>)
+
     install(
       FILES $<TARGET_PDB_FILE:${target}>
       CONFIGURATIONS "RelWithDebInfo" "Debug"
--- a/UI/frontend-plugins/aja-output-ui/CMakeLists.txt
+++ b/UI/frontend-plugins/aja-output-ui/CMakeLists.txt
@@ -18,6 +18,11 @@ set_target_properties(
              AUTORCC ON
              AUTOUIC_SEARCH_PATHS "forms")
 
+if(_QT_VERSION EQUAL 6 AND OS_WINDOWS)
+  set_target_properties(aja-output-ui PROPERTIES AUTORCC_OPTIONS
+                                                 "--format-version;1")
+endif()
+
 target_sources(aja-output-ui PRIVATE forms/output.ui)
 
 target_sources(
--- a/UI/frontend-plugins/decklink-captions/CMakeLists.txt
+++ b/UI/frontend-plugins/decklink-captions/CMakeLists.txt
@@ -18,6 +18,11 @@ set_target_properties(
              AUTORCC ON
              AUTOUIC_SEARCH_PATHS "forms")
 
+if(_QT_VERSION EQUAL 6 AND OS_WINDOWS)
+  set_target_properties(decklink-captions PROPERTIES AUTORCC_OPTIONS
+                                                     "--format-version;1")
+endif()
+
 target_compile_features(decklink-captions PRIVATE cxx_std_17)
 
 target_sources(decklink-captions PRIVATE forms/captions.ui)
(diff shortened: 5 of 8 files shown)


UI/CMakeLists.txt
```diff
@@ -80,6 +80,7 @@ set_target_properties(
   PROPERTIES AUTOMOC ON
              AUTOUIC ON
              AUTORCC ON
+             AUTORCC_OPTIONS "--format-version;1"
```
tytan652, line 83:
```suggestion
             AUTORCC_OPTIONS "--format-version;1;--compress-algo;zlib"
```

tytan652: Context: obs-deps build Qt only with zlib but Linux have all features and then default to zstd which does not work with version less than 3.

derrod: Yeah, that works. Apparently Qt also supports `SOURCE_DATE_EPOCH` nowadays to force a specific timestamp, but setting that from within cmake did not appear to work for Windows so I'm sticking with overriding the version for now.

derrod:
Went back to just setting the `AUTORCC` options for now, and made it limited to Windows. Reproducible builds aren't really a thing on macOS and the main motivation of this PR is to make Windows builds reproducible. I'd still like to incorporate https://github.com/jasonwhite/ducible to ensure that EXE/PDB metadata is always consistent (which also reduces delta update sizes!) but I would probably have to add it to obs-deps first.

derrod:
Indeed #8377 seems to have fixed `coreaudio-encoder.dll` not being reproducible. I will need to also submit the Qt RCC options to obs-websocket for our CI builds to be fully reproducible.

I will consider this PR ready for review now.

derrod:
The workflow run here can now be checked to see the results of this PR with two parallel builds: https://github.com/obsproject/obs-studio/actions/runs/4314253892

When checking the hashes, only the PDBs and `obs-websocket.dll` are mismatched, the former is not an issue, the latter will require a PR to the websockets repo at some point.

cmake/Modules/ObsHelpers_Windows.cmake
```diff
@@ -29,6 +29,9 @@ function(setup_binary_target target)
   endif()
 
   if(MSVC)
+    target_link_options(${target} PRIVATE
+                        /PDBALTPATH:$<TARGET_PDB_FILE_NAME:${target}>)
```
PatTheMav, line 33: I take it `PDBALTPATH` is used to avoid having that info written into the binary itself?

derrod: By default MSVC writes the absolute path of the PDB file into the binary. This overrides that to be just the name (i.e. relative) so that binaries build in different environments remain identical.

PatTheMav (approved):
Looks good to me - CMake changes use the fixes suggested to make builds reproducible and also use the correct target properties for it.

cmake/Modules/CompilerConfig.cmake
```diff
@@ -75,6 +76,10 @@ if(OS_WINDOWS AND MSVC)
     /std:c17)
 
   add_link_options(
+    # ToDo figure out how to properly escape this (currently it gets turned into
+    # %%%%PDB%%%%, thanks cmake!)
```
RytoEX: Remove this todo from the commit.

cmake/Modules/CompilerConfig.cmake
```diff
@@ -75,6 +76,10 @@ if(OS_WINDOWS AND MSVC)
...
+    "LINKER:/PDBALTPATH:%_PDB%"
```
RytoEX: As this is removed in a later commit, either squash as needed, or remove from here.

RytoEX (changes requested):
I think I commented off-thread about a concern with forcing Qt RCC to v1 and wanting to know if this was a known deficiency to Qt (i.e., in a QTBUG) with v2. Ideally, we'd be able to use v2 (or some later v3) without having to keep ourselves stuck on this older format.

Other than that and the inline review comments, looks good.

derrod:
> I think I commented off-thread about a concern with forcing Qt RCC to v1 and wanting to know if this was a known deficiency to Qt (i.e., in a QTBUG) with v2. Ideally, we'd be able to use v2 (or some later v3) without having to keep ourselves stuck on this older format.

v2 introduces last-modified timestamps, so this isn't a bug, but a known problem for reproducability. Qt 5.11 introduced an environment variable (`QT_RCC_SOURCE_DATE_OVERRIDE `) to set a fixed timestamp, but you cannot set it from the `rcc` command line and setting it from CMake has not worked for me.

There is no benefit to using v2 or v3 for us. v2 introduces timestamps, v3 zstd compression which we do not build. If we ever do want to use a newer version we'll have to either:
 * Somehow set `SOURCE_DATE_EPOCH`/`QT_RCC_SOURCE_DATE_OVERRIDE` before running `rcc`
 * Modify the last-modified timestamp of all resource files to a fixed value before running `rcc`
 * Patch our obs-deps `rcc` to have a fixed timestamp

RytoEX:
> > I think I commented off-thread about a concern with forcing Qt RCC to v1 and wanting to know if this was a known deficiency to Qt (i.e., in a QTBUG) with v2. Ideally, we'd be able to use v2 (or some later v3) without having to keep ourselves stuck on this older format.
> 
> v2 introduces last-modified timestamps, so this isn't a bug, but a known problem for reproducability. Qt 5.11 introduced an environment variable (`QT_RCC_SOURCE_DATE_OVERRIDE`) to set a fixed timestamp, but you cannot set it from the `rcc` command line parameters and setting it from CMake has not worked for me.
> 
> There is no benefit to using v2 or v3 for us. v2 introduces timestamps, v3 zstd compression which we do not build. If we ever do want to use a newer version we'll have to either:
> 
>     * Somehow set `SOURCE_DATE_EPOCH`/`QT_RCC_SOURCE_DATE_OVERRIDE` before running `rcc`
> 
>     * Modify the last-modified timestamp of all resource files to a fixed value before running `rcc`
> 
>     * Patch our obs-deps `rcc` to have a fixed timestamp

Could we not [set the environment variable early via CMake](https://cmake.org/cmake/help/v3.26/command/set.html#set-environment-variable)? I suppose it is not something that needs addressed now, I'm just wary of pinning on an old version causing problems for us down the road.

If each commit builds on its own, then no need to further squash.

derrod:
> Could we not [set the environment variable early via CMake](https://cmake.org/cmake/help/v3.26/command/set.html#set-environment-variable)? I suppose it is not something that needs addressed now, I'm just wary of pinning on an old version causing problems for us down the road.

I've tried that and it didn't work (at least on Windows). We can revisit if there are any changes on Qt's side that make using the old version untenable for us.

> If each commit builds on its own, then no need to further squash.

They all build on their own.

RytoEX:
> I've tried that and it didn't work (at least on Windows). We can revisit if there are any changes on Qt's side that make using the old version untenable for us. See their note (emphasis mine):
> 
> > This command affects **only the current CMake process**, not the process from which CMake was called, nor the system environment at large, nor the environment of subsequent build or test processes.

Ah. I see. That's curious, if not a bit unfortunate. We can, of course, set this in CI environments elsewhere. To set it in a non-CI environment, we could probably set it from a build script, or else we would have to set it manually (yuck).

As an aside, [QTBUG-105926](https://bugreports.qt.io/browse/QTBUG-105926) seems to be the Qt place to track work on reproducible builds.

RytoEX (approved):
Tentatively approved. @PatTheMav will revisit tomorrow to be sure we have no lingering questions or concerns.

PatTheMav:
Alas there is no support for _any_ environment variables in Qt's CMake modules, so there is no way to influence this behaviour with them (which makes sense given that Qt doesn't consider reproducible builds "supported" and they haven't considered support for it yet).

The only other way would be to set it explicitly in the environment before running `cmake`.

Issue #6081, opened 2022-03-04
Settings loading slowly on massive scene collections


flaeri:
### Operating System Info

Windows 11

### OBS Studio Version

27.2.3

### OBS Studio Log URL

https://obsproject.com/logs/6z4OmpDLvjpHUSYf

### Expected Behavior

Expected behavior would be for buttons/functions to load with somewhat similar performance to previous version.

### Current Behavior

Currently, 27.2.0-beta1 onwards loads about half as quickly as previous versions when clicking the "settings" button.

### Steps to Reproduce

1. Stock OBS (no plugins etc)
2. Create a very large scene collection (a lot of scenes and sources)
OR import the one provided: [Massive.zip](https://github.com/obsproject/obs-studio/files/8186981/Massive.zip)
3. Test performance on 27.2.0-beta1 and newer vs 27.1.3 and older

### Anything else we should know?

With the provided scene collection ([Massive.zip](https://github.com/obsproject/obs-studio/files/8186981/Massive.zip)) I've observed the following approximate values on a modern'ish Laptop.

5-6 sec opening settings on 27.1.3
10-14 sec opening settings on 27.2.0-beta1

On an even larger scene collection I got:
9-10 sec opening settings 27.1.3
25-30 sec opening settings 27-2-beta1

The performance numbers seem identical from 27-2-beta1 compared to latest release (27.2.3 at the time of writing).

The scene collection is just a bunch of simple Color Sources.

I believe the extra time loading is primarily spent on the changes in https://github.com/obsproject/obs-studio/pull/5297

By no means is this meant to be a usable scene scene collection (opening hotkeys or shutting down OBS takes a very very long time). 
I've observed a couple of people visiting the OBS support channels with similarly sized scene collections complaining of OBS hanging (long loading times).

jeansuizo:
I have the same issue. After updating from 27.2.2 to 27.2.3 it works well, except that going to the Settings leads to crash on my massive scene collection (58 scenes) 
This only occurs with massive collections. In version 27.2.2 it needed about 8 seconds with the same scenes.
Win 11, I7 9.thGen, Asus Zenbook ProDuo

flaeri:
@jeansuizo This might be a different issue. 27.2.2 and 27.2.3 did not have changes in this regard as far as I know.
You're welcome to join the discord, and hit us up in the support channels, and we could have a look.
https://obsproject.com/discord

WizardCM:
[...]

PR #8229, opened 2023-02-07
Hashtable Adventures

derrod:
### Description

This is a finished implementation of @jp9000's [hash-table](https://github.com/obsproject/obs-studio/tree/hash-table) branch.

**Overview:**
Switched various components from using linked lists or radix/trie trees to hashmaps (via `uthash`):
  + Config file utility (INI read/write)
  + Hotkeys and their name map
  + OBS data (`obs_data_item`)
  + Translation text lookup (`lookup_t`)
  + Public and Private Sources (public by name and UUID, private by UUID)
  + Properties

**Breaking changes**
- N/A

**_Potentially_ breaking changes**
- `obs_source_create()` now deduplicates the provided name for public sources
  + Also affects `obs_source_duplicate()` and `obs_source_set_name()`
  + This makes the functions behave as they are described in the documentation, however, some people may rely on the previous behaviour
- Private groups are no longer enumerated in `obs_enum_sources()`
  + OBS itself does not create groups as private sources
- `config_t` is now case-sensitive
  + Only example I found was `SysTrayEnabled` in `window-basic-main.cpp`, this has been fixed in this PR
- `text_lookup` is now case-sensitive
  + The only issue I could find in OBS itself was `obs-ffmpeg-source.c` using an incorrectly-capitalised translation key for the name (`FFMpegSource`), this has been fixed in this PR

### Motivation and Context

Faster and cleaner code. Especially for translation lookups (10x speedup in my tests).

Additionally replacing the darrays in hotkeys resulted in a massive speedup when unloading a scene collection which makes working with large ones much nicer.

### How Has This Been Tested?

- Launched OBS
- Made some changes to config/sources
- Checked there are no memory leaks

### Types of changes

### Checklist:


tytan652:
Then uthash header only deps should be added to obs-deps. Ubuntu has the `uthash-dev` package.

derrod:
> uthash (headers only library) deps should be added to obs-deps. Ubuntu has the `uthash-dev` package.

The version in ubuntu 20.04 is older and is missing some stuff. I think for something as fundamental and small as this vendoring it inside libobs itself like SIMDe might be the most sensible choice since we really only need the `uthash.h` file. Unless we want to use it in plugins of course.

tytan652:
Then we drop Ubuntu 20.04, if it becomes a blocker for improvements.

Adding deps inside obs-studio tree is bad and is a bad practice, most of the deps that are like this are not kept up-to-date.

Adding to obs-deps allows to centralize those deps in one place.

libobs/util/hash/README.libobs
```diff
@@ -0,0 +1,31 @@
+# `uthash`
+
+Vendored from https://github.com/troydhanson/uthash for use in OBS.
+
+Modifications inlucde:
+ - Removed unused files
+ - Formatted with OBS clang-format settings
```
tytan652, line 7: We should stop including deps in-tree and rely more on obs-deps, Flatpak modules and Ubuntu packages when possible.

Ubuntu 20.04 does not have a recent enough `uthash-dev` package. If it becomes a blocker for improvements I guess 20.04 support should be dropped.

tytan652 (changes requested):

kkartaltepe:
> Unless we want to use it in plugins of course.

I would like to use this in xcomposite to replace our poor implementation of a set.  But also agree with just keeping the header in-project.

derrod:
> I would like to use this in xcomposite to replace our poor implementation of a set. But also agree with just keeping the header in-project.

In that case I could move it back to `deps`. That would also make it easier to update in the future since you just need to replace the header files there with the latest release.

I also don't know if we care about Elbrus CPU support still, but that would require using the uthash version from git since their latest release doesn't have the fix for that yet (https://github.com/troydhanson/uthash/commit/4d01591ed9b67ff153410d0f0cfa47f51b735eb4).

derrod:
Just for fun, a little test to use xxHash with SSE2 vector instructions instead of uthash's built-in hash functions: https://github.com/derrod/obs-studio/commit/0fb32c56f06e21f31ccee69ac9eb7d1f2e2c2b29 (obs-deps commit to add it for Windows here: https://github.com/derrod/obs-deps/commit/befd6564ae659fca586927f9d0e43430131bb94b).

It's a decent bit faster (~25%) but probably overkill for our purposes. Just felt like leaving it here since it's an interesting experiment.

derrod:
Moved hotkeys to hashtables, this is one of the things where you can see an absolutely massive improvement when it comes to saving/closing a scene collection. In particular the example linked in #6081 (`Massive.json`) on my system now takes < 1 second to unload/save, from over 1 minute and 15 seconds!

norihiro:
Can we add `HASH_ITER` to `ForEachMacros` in `.clang-format`?
```patch
diff --git a/.clang-format b/.clang-format
index c9dfc48e9..89cbc19a6 100644
--- a/.clang-format
+++ b/.clang-format
@@ -55,10 +55,11 @@ DisableFormat: false
 FixNamespaceComments: false
 ForEachMacros: 
   - 'json_object_foreach'
   - 'json_object_foreach_safe'
   - 'json_array_foreach'
+  - 'HASH_ITER'
 IncludeBlocks: Preserve
 IndentCaseLabels: false
 IndentPPDirectives: None
 IndentWidth: 8
 IndentWrappedFunctionNames: false
```

derrod:
We can, yeah. Not sure I like the result more than the default but it's fine:
```diff
diff --git a/libobs/obs-hotkey.c b/libobs/obs-hotkey.c
index d8c694189..d2e7e0346 100644
--- a/libobs/obs-hotkey.c
+++ b/libobs/obs-hotkey.c
@@ -1009,8 +1009,7 @@ void obs_hotkeys_context_release(struct obs_context_data *context)
 void obs_hotkeys_free(void)
 {
        obs_hotkey_t *hotkey, *tmp;
-       HASH_ITER(hh, obs->hotkeys.hotkeys, hotkey, tmp)
-       {
+       HASH_ITER (hh, obs->hotkeys.hotkeys, hotkey, tmp) {
                HASH_DEL(obs->hotkeys.hotkeys, hotkey);
                bfree(hotkey->name);
                bfree(hotkey->description);
```

libobs/util/hash/README.libobs
derrod: Based on off-chain discussions we believe that right now vendoring is the simplest way forward, and means not having to worry about package availability in Linux distributions and not having to write finders etc. for CMake.

Once the CMake rework for Windows and Linux are finalised and merged and we drop support for 20.04 we can revisit that and move the headers over to obs-deps.

libobs/obs-source.c
```diff
@@ -339,6 +345,26 @@ static void obs_source_init_audio_hotkeys(struct obs_source *source)
...
+{
+	struct obs_context_data *tmp;
+	HASH_FIND_STR((struct obs_context_data *)obs->data.sources, *name, tmp);
+
+	if (tmp) {
+		struct dstr new_name = {0};
+		int suffix = 2;
+
+		while (tmp) {
+			dstr_free(&new_name);
```
notr1ch, line 358: I don't think it's necessary to call `dstr_free` here.

libobs/obs-data.c
```diff
@@ -27,11 +27,20 @@
...
 
 #include <jansson.h>
+#include <uthash/uthash.h>
+
+#ifdef UTHASH_USE_OBS_ALLOCATOR
+#undef uthash_malloc
+#undef uthash_free
+#define uthash_malloc(sz) bmalloc(sz)
+#define uthash_free(ptr, sz) bfree(ptr);
+#endif
```
notr1ch, line 37: This is duplicated across many source files, since we're vendoring uthash could we move it to `uthash.h`? Also doesn't this need to be set before the header is included?

derrod: It does not need to be set before the header is included, this works as expected. Since we want to eventually move away from vendored deps and want to avoid modifications I'd probably rather just remove the commit for this entirely, it was mostly just to sanity check what I'm doing anyway.

libobs/obs-source.c
derrod: Yeah it probably isn't since `dstr_printf` will just overwrite or reallocate the memory, I'll remove it.

derrod:
As a general note, some of these commits could be squashed, but I'd prefer #8345 to be merged beforehand so I only have to resolve conflicts once (as a bonus the diff will be slightly cleaner if this is applied on top of 8345 than the other way around).

derrod:
As mentioned in the previous commit I'd prefer if #8345 is merged first and to that effect I have now rebased this PR on that one. This makes a few things slightly easier/cleaner. There also were a number of minor changes and code cleanup compared to the original PR.

derrod:
With #8345 merged I updated and cleaned up this PR a bit. I'm now fairly happy with where it's at. Hopefully it can make it for 29.1!

tt2468 (approved):
Approving in regards to compatibility in obs-websocket. obs-websocket currently performs its own checks for `obs_source_create()` and `obs_source_duplicate()`, and will deny requests if a source already exists with the specified name. As such, no behavior will be changed.

norihiro:
I got a crash at exit saying `double free or corruption (out)`.

I tried to reproduce this without 3rd party plugin but I couldn't do that so far.

obs-studio commit: 67c93f43
OBS log: https://obsproject.com/logs/Mx4I5sOo1TwCfdb3
3rd party plugin: [obs-ptz](https://github.com/glikely/obs-ptz/) commit c904d0391

Steps:
0. Build and install obs-ptz plugin.
1. Start OBS
2. Add a dock `PTZ Controls`
3. Continue stop and start OBS until the crash happen. (roughly once for 10 times)

```
[Current thread is 1 (Thread 0x7fe732c9c480 (LWP 422197))]
(gdb) bt
#0  0x00007fe73df8e490 in unregister_hotkey (id=id@entry=24) at /home/kamae/repo/obs-studio/libobs/obs-hotkey.c:914
#1  0x00007fe73df9191d in obs_hotkey_unregister (id=24) at /home/kamae/repo/obs-studio/libobs/obs-hotkey.c:956
#2  0x00007fe6a927ba8b in PTZControls::~PTZControls() (this=0x2689c40, __in_chrg=<optimized out>)
    at /home/kamae/repo/obs-ptz/src/ptz-controls.cpp:296
#3  0x00007fe6a927bb69 in PTZControls::~PTZControls() (this=0x2689c40, __in_chrg=<optimized out>)
    at /home/kamae/repo/obs-ptz/src/ptz-controls.cpp:301
#4  0x00007fe73bdab75a in QObjectPrivate::deleteChildren() () at /lib64/libQt6Core.so.6
#5  0x00007fe73d20d4b8 in QWidget::~QWidget() () at /lib64/libQt6Widgets.so.6
#6  0x000000000053baa9 in OBSBasic::~OBSBasic() (this=0xf5f040, __in_chrg=<optimized out>)
    at /home/kamae/repo/obs-studio/UI/window-basic-main.cpp:2787
#7  0x00007fe73bda3693 in QObject::event(QEvent*) () at /lib64/libQt6Core.so.6
#8  0x00007fe73d1b9a35 in QApplicationPrivate::notify_helper(QObject*, QEvent*) () at /lib64/libQt6Widgets.so.6
#9  0x00007fe73bd50c48 in QCoreApplication::notifyInternal2(QObject*, QEvent*) () at /lib64/libQt6Core.so.6
#10 0x00007fe73bd53f94 in QCoreApplicationPrivate::sendPostedEvents(QObject*, int, QThreadData*) () at /lib64/libQt6Core.so.6
#11 0x00007fe73c00a147 in postEventSourceDispatch(_GSource*, int (*)(void*), void*) () at /lib64/libQt6Core.so.6
#12 0x00007fe73b318faf in g_main_context_dispatch () at /lib64/libglib-2.0.so.0
#13 0x00007fe73b36e2c8 in g_main_context_iterate.constprop () at /lib64/libglib-2.0.so.0
#14 0x00007fe73b316940 in g_main_context_iteration () at /lib64/libglib-2.0.so.0
#15 0x00007fe73c009a10 in QEventDispatcherGlib::processEvents(QFlags<QEventLoop::ProcessEventsFlag>) () at /lib64/libQt6Core.so.6
#16 0x00007fe73bd5da1b in QEventLoop::exec(QFlags<QEventLoop::ProcessEventsFlag>) () at /lib64/libQt6Core.so.6
#17 0x00007fe73bd59118 in QCoreApplication::exec() () at /lib64/libQt6Core.so.6
#18 0x000000000046996a in run_program (argv=0x7ffeee641088, argc=<optimized out>, logFile=...)
    at /home/kamae/repo/obs-studio/UI/obs-app.cpp:2484
#19 main(int, char**) (argc=<optimized out>, argv=0x7ffeee641088) at /home/kamae/repo/obs-studio/UI/obs-app.cpp:3376
```
When I have more plugins, I saw a crash at another location.
obs-studio commit: 67c93f43
OBS log: https://obsproject.com/logs/1Cwh5KCeF6vlQsEw

backtrace:
```
Program terminated with signal SIGABRT, Aborted.
#0  0x00007f060228ebec in __pthread_kill_implementation () from /lib64/libc.so.6
[Current thread is 1 (Thread 0x7f05dbfff640 (LWP 412422))]
(gdb) bt
#0  0x00007f060228ebec in __pthread_kill_implementation () at /lib64/libc.so.6
#1  0x00007f060223e956 in raise () at /lib64/libc.so.6
#2  0x00007f06022287f4 in abort () at /lib64/libc.so.6
#3  0x00007f0602282d3e in __libc_message () at /lib64/libc.so.6
#4  0x00007f060229893c in  () at /lib64/libc.so.6
#5  0x00007f060229aa60 in _int_free () at /lib64/libc.so.6
#6  0x00007f060229d133 in free () at /lib64/libc.so.6
#7  0x00007f0604d8eadb in a_free (ptr=<optimized out>) at /home/kamae/repo/obs-studio/libobs/util/bmem.c:84
#8  0x00007f0604d2d707 in unregister_hotkey (id=59) at /home/kamae/repo/obs-studio/libobs/obs-hotkey.c:930
#9  0x00007f0604d30c01 in context_release_hotkeys (context=0x382c920) at /home/kamae/repo/obs-studio/libobs/obs-hotkey.c:979
#10 obs_hotkeys_context_release (context=0x382c920) at /home/kamae/repo/obs-studio/libobs/obs-hotkey.c:1002
#11 0x00007f0604d1bc3e in obs_context_data_free (context=context@entry=0x382c920) at /home/kamae/repo/obs-studio/libobs/obs.c:2537
#12 0x00007f0604d553a8 in obs_source_destroy_defer (source=0x382c920) at /home/kamae/repo/obs-studio/libobs/obs-source.c:779
#13 0x00007f0604d9c838 in tiny_tubular_task_thread (param=0x235dfe0) at /home/kamae/repo/obs-studio/libobs/util/task.c:161
#14 0x00007f060228cdcd in start_thread () at /lib64/libc.so.6
#15 0x00007f0602312630 in clone3 () at /lib64/libc.so.6
```

At the trace `#12`, the ID of the source is `pulse_input_capture`.

libobs/obs-hotkey.c
```diff
@@ -1021,59 +905,55 @@ static void release_registerer(obs_hotkey_t *hotkey)
...
+	if (hotkey->registerer_type == OBS_HOTKEY_REGISTERER_SOURCE)
+		obs_weak_source_release(hotkey->registerer);
+
 	bfree(hotkey->name);
 	bfree(hotkey->description);
+	bfree(hotkey);
 
-	if (hotkey->registerer_type == OBS_HOTKEY_REGISTERER_SOURCE)
-		obs_weak_source_release(hotkey->registerer);
+	HASH_DEL(obs->hotkeys.hotkeys, hotkey);
```
norihiro, line 929: Should `HASH_DEL` be called before `bfree`?
```suggestion
	HASH_DEL(obs->hotkeys.hotkeys, hotkey);

	bfree(hotkey);
```

derrod: Oh yes, this is a use after free so they must be the other way around.

norihiro: With this fix, the failures I reported at https://github.com/obsproject/obs-studio/pull/8229#issuecomment-1473194490 never happen again :-)

derrod:
Rebased on master and fixed the use-after-free @norihiro pointed out. Also ran it with AddressSanitizer just to see if it would catch anything else, but it didn't (it just made running OBS nearly unbearable on my machine).

PR #8238, opened 2023-02-09
UI: Fix scene item edit drag & drop bug

cg2121:
### Description
If an user exits the source tree edit mode, without changing the name, the scene item would become undraggable until the scene is refreshed.

### Motivation and Context
Bug reported by @Warchamp7 

### How Has This Been Tested?
Entered the source tree edit mode and exited, without changing the name, and made sure the item was draggable.

### Types of changes

### Checklist:


cg2121:
I've figured out that if we just set the focus to the whole widget, not just the label, it also fixes this bug.

RytoEX:
Focus changes make me a little nervous, particularly on macOS or with screen readers. We'll just have to keep an eye out and hope for the best.

PR #8242, opened 2023-02-09
libobs: Allow gs_texrender_t shared texture creation

MrMahgu:
### Description
A new `gs_texrender_create2()` function adds the ability for an underlying texrender texture to be created with and additional `GS_SHARED_TEX` flag.

- Add `boolean` named `share` to `gs_texrender_t` struct to provide ability to set `GS_SHARED_TEX` flag on texture creation.
- Adds `gs_texrender_create2()` to build `gs_texrender_t` struct with new shared boolean set to true.
- Modifies `texrender_resetbuffer()` to check if texture is shared and adds `GS_SHARED_TEX` to texture creation flags.
- Adds helper function `gs_texrender_has_shared_texture()` which returns  a `bool` value set in original `gs_texrender _t` creation.

### Motivation and Context
This change allows plugins to create a texture which can be shared with external processes without creating and additional texture and without requiring and additional copy to that texture.

### How Has This Been Tested?
The changes have been tested only locally on my windows 11 machine. My current dev machine is an AMD 2700, Nvidia 1070, 32GB, Windows 11 Home.

I have tested the changes for many hours against various directx software processes.

### Types of changes

### Checklist:

### DIFF
--- a/libobs/graphics/graphics.h
+++ b/libobs/graphics/graphics.h
@@ -465,6 +465,8 @@ EXPORT void gs_effect_set_color(gs_eparam_t *param, uint32_t argb);
 
 EXPORT gs_texrender_t *gs_texrender_create(enum gs_color_format format,
 					   enum gs_zstencil_format zsformat);
+EXPORT gs_texrender_t *gs_texrender_create2(enum gs_color_format format,
+					    enum gs_zstencil_format zsformat);
 EXPORT void gs_texrender_destroy(gs_texrender_t *texrender);
 EXPORT bool gs_texrender_begin(gs_texrender_t *texrender, uint32_t cx,
 			       uint32_t cy);
@@ -476,6 +478,7 @@ EXPORT void gs_texrender_reset(gs_texrender_t *texrender);
 EXPORT gs_texture_t *gs_texrender_get_texture(const gs_texrender_t *texrender);
 EXPORT enum gs_color_format
 gs_texrender_get_format(const gs_texrender_t *texrender);
+EXPORT bool gs_texrender_has_shared_texture(const gs_texrender_t *texrender);
 
 /* ---------------------------------------------------
  * graphics subsystem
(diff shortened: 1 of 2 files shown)


kkartaltepe:
> without creating and additional texture and without requiring and additional copy to that texture

You have already created the extra texture via gs_texrender_create and are performing a copy via render. Why not just create a plain pair of shared textures and copy directly via gs_copy_texture. Either way it doesnt seem to me that v2 of this function just always creates a shared texture, such a change might instead be named `_shared` instead of v2.

MrMahgu:
> You have already created the extra texture via gs_texrender_create and are performing a copy via render. Why not just create a plain pair of shared textures and copy directly via gs_copy_texture. 

The end goal with this change is so that I can share the underlying texture used inside my callback from obs_add_main_render_callback.  Currently I have the following in my callback:

```
gs_texrender_reset(filter->texrender_current_ptr);
if (gs_texrender_begin(filter->texrender_current_ptr, width, height)) {
	// ... 
	gs_texrender_end(filter->texrender_current_ptr);
}
```

I would happily replace the texture inside my gs_texrender_t *elm if I could/knew how.  Which would provide me the same functionality.

> Either way it doesnt seem to me that v2 of this function should just always create a shared texture, such a new function might instead be named `_shared` instead of v2.

I choose to simply add 2 to the function name as I don't think it would be very useful to 99.99% of plugin writers. I am terrible with names and have no preference.

kkartaltepe:
> I would happily replace the texture inside my gs_texrender_t *elm if I could/knew how. Which would provide me the same functionality.

You can look at the implementation of `gs_texrender_begin`, basically just `gs_set_render_target_with_color_space` your shared texture directly.

MrMahgu:
> You can look at the implementation of `gs_texrender_begin`, basically just `gs_set_render_target_with_color_space` your shared texture directly.

Thank you very much.  That is exactly what I'm trying to do!

I will close this pull request.

PR #8243, opened 2023-02-09
mac-videotoolbox: Load encoders from system asynchronously

gxalpha:
### Description

Copying the encoder list takes a while which blocks the main thread. Doing the copying asynchronously removes instead of blocking the main thread improves startup performance by about 60ms (from ~60-70ms to less than 1ms including the post_load function).
Most importantly, it stops Xcode from throwing 18 runtime warnings at us.

### Motivation and Context

Xcode complained *loudly* about us running the `VTCopyVideoEncoderList` function on the main thread.

### How Has This Been Tested?

macOS 13.1
Checked that all encoders are still there.

Checked with profiler that loading times for the module that previously were about 60-70ms are down to less than 1ms for both module_load and module_post_load.

### Types of changes

### Checklist:

### DIFF
--- a/plugins/mac-videotoolbox/encoder.c
+++ b/plugins/mac-videotoolbox/encoder.c
@@ -1477,8 +1477,23 @@ compare_encoder_list(const void *left_val, const void *right_val, void *unused)
 
 OBS_DECLARE_MODULE()
 OBS_MODULE_USE_DEFAULT_LOCALE("mac-videotoolbox", "en-US")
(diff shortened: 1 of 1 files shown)


PatTheMav (approved):
The speed improvement is really noticeable. I wish we could use this pattern for more modules to make the heavy load operations run on other threads.

PatTheMav:
This has the potential to "deadlock" the application if the queued block never finishes, though that would have also affected the app when it could've happened on the main thread.

PR #8244, opened 2023-02-10
UI: Add check for null widgetForAction result

zavitax:
### Description

In some cases, `QToolbar::widgetForAction` may return `nullptr`. This causes a crash on affected system.

To mitigate this, we check for `widgetForAction` result and skip operating on a NULL widget.

### Motivation and Context

Not checking for the return value causes the program to crash in cases where `widgetForAction` returns nullptr

### How Has This Been Tested?

Crash reproduced locally, added fix, crash disappeared.

### Types of changes

### Checklist:

### DIFF
--- a/UI/qt-wrappers.cpp
+++ b/UI/qt-wrappers.cpp
@@ -412,6 +412,10 @@ void RefreshToolBarStyling(QToolBar *toolBar)
 {
 	for (QAction *action : toolBar->actions()) {
 		QWidget *widget = toolBar->widgetForAction(action);
+
+		if (!widget)
+			continue;
+
 		widget->style()->unpolish(widget);
 		widget->style()->polish(widget);
 	}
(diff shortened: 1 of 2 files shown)


gxalpha:
Please also add the `UI: ` prefix to the commit itself.

Seeing that this is probably meant to fix the crash for the SE.live plugin there might not be, but are you aware of a way to trigger the original issue with just OBS itself or a simple modification to it? Would be interesting to know what causes the widget to be null in the first place.

Other than that, this looks fairly straightforward.

zavitax:
> Please also add the `UI: ` prefix to the commit itself.
Added

> Seeing that this is probably meant to fix the crash for the SE.live plugin there might not be, but are you aware of a way to trigger the original issue with just OBS itself or a simple modification to it? Would be interesting to know what causes the widget to be null in the first place.
> 
> Other than that, this looks fairly straightforward.
What seems to trigger it is a sequence of QToolBar::addWidget() followed by QToolBar::layout()->removeWidget() possibly followed by another QToolBar::addWidget() call

Issue #7280, opened 2022-09-03
Scene transitions are grayed out when scene collection is changed


style1925:
### Operating System Info

Windows 10

### OBS Studio Version

28.0.0

### OBS Studio Version (Other)

28.0.1

### OBS Studio Log URL

https://obsproject.com/logs/jf8KR4FuHCoRqgss

### Expected Behavior

Scene transitions can be changed even after changing the scene collection.

### Current Behavior

Scene transitions are grayed out.

### Steps to Reproduce

1. change the scene collection
2. Scene transitions are grayed out

28.0-rc2

https://user-images.githubusercontent.com/1337354/188270494-c5e9a288-cf17-4477-9433-92f0c6cf3a54.mp4

28.0.1

https://user-images.githubusercontent.com/1337354/188270496-55ec9e91-32fa-4531-86ee-e62e8ae044e9.mp4

### Anything else we should know?

It does not occur until 28.0 rc2.
Changing the scene cures the gray out. (It will be grayed out only when changing scene collections.)

kkartaltepe:
Replicated on linux.

yoshinorisano:
I also reproduced the bug and investigated it using Visual Studio's debugger.
I found that the following pull request produces this bug.
https://github.com/obsproject/obs-studio/pull/7200/

PatTheMav:
The observation is correct - the change introduced by that commit forces a scene item to be selected even though we purposefully clear the selection before importing the new scene selection.

This leads to a forced scene transition twice, which starts a transition (disabling the transition dock), but as no actual transition takes place, the dock is not activated again until users interact with the scene setup.

PR #8246, opened 2023-02-10
UI: Remove workaround for current scene being deselectable on Qt 6.4.3+

gxalpha:
### Description

Qt 6.2 introduced an issue where SingleSelection item views would deselect the current item if the user clicked on empty area in the widget.
This was very confusing in the scene tree as it was now possible to unselect the current scene. A workaround for this in OBS was added in 08e4ee6 (#7200) and expanded on in dc30cf0 (#7499), but being quite hacky it never was the perfect solution.

I since dug into Qt and fixed the issue upstream at [Qt Gerrit 459511](https://codereview.qt-project.org/c/qt/qtbase/+/459511). ~~That hasn't actually landed in Qt (yet), but first review by the maintainer suggests it's in relatively decent shape, so I'm preparing this PR now.~~ That patch has now landed.
It's being cherry-picked to 6.4 and 6.5, so it will be in 6.4.3 and newer meaning we can remove the workaround when using those versions with the long term goal of removing the code altogether (In 6.5 it will only be in Beta 3 or newer (since 1 and 2 are already out), but I don't care a lot about people using Beta Qt with OBS).

~~TODO: Wait until the patch actually lands. Fix the commit description to reference the actual commit in qtbase and potentially adjust the version in code.~~ Done.

### Motivation and Context

We should be fixing issues upstream instead of working around them downstream when possible. This improves both our lives by reducing the need for workarounds and the lives of other people using the same dependencies.

### How Has This Been Tested?

macOS 13.1, Qt dev branch.
It's no longer possible to unselect the current scene by clicking on empty area below the scenes, as well as the current page in the settings window (which we didn't work around in the first place since it wasn't particularly important and not worth the extra complexity).
Verified #7280 didn't magically regress somehow by changing scene collections and changing transitions.

### Types of changes

### Checklist:

### DIFF
--- a/UI/item-widget-helpers.cpp
+++ b/UI/item-widget-helpers.cpp
@@ -37,8 +37,10 @@ void DeleteListItem(QListWidget *widget, QListWidgetItem *item)
 
 void ClearListItems(QListWidget *widget)
 {
+#if QT_VERSION < QT_VERSION_CHECK(6, 4, 3)
 	// Workaround for the SceneTree workaround for QTBUG-105870
 	widget->setProperty("clearing", true);
+#endif
 
 	widget->setCurrentItem(nullptr, QItemSelectionModel::Clear);
 
@@ -46,5 +48,8 @@ void ClearListItems(QListWidget *widget)
 		delete widget->itemWidget(widget->item(i));
 
 	widget->clear();
+#if QT_VERSION < QT_VERSION_CHECK(6, 4, 3)
+	// Workaround for the SceneTree workaround for QTBUG-105870
 	widget->setProperty("clearing", false);
+#endif
 }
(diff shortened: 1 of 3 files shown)


UI/item-widget-helpers.cpp
```diff
@@ -37,14 +37,18 @@ void DeleteListItem(QListWidget *widget, QListWidgetItem *item)
...
+#endif
 
 	widget->setCurrentItem(nullptr, QItemSelectionModel::Clear);
 
 	for (int i = 0; i < widget->count(); i++)
 		delete widget->itemWidget(widget->item(i));
 
 	widget->clear();
+#if QT_VERSION < QT_VERSION_CHECK(6, 4, 3)
 	widget->setProperty("clearing", false);
```
RytoEX, line 52: Maybe add a comment here that this is also related to QTBUG-105870 just to make it clear.

RytoEX (changes requested):

UI/item-widget-helpers.cpp
gxalpha: Added.

gxalpha:
The patch landed in Qt (qt/qtbase@f11e5435c7), I'm very happy. Updated the commit description to include the upstream commit.

PR #8251, opened 2023-02-11
libobs-opengl: Accelerate dmabuf import

kkartaltepe:
### Description

Previously we would actually initialize a texture memory that would then also have to be deleted when we bound the EGLImage to the texture during dmabuf import.

Instead simply do not create the dummy texture memory. One odd thing is that we must still query the texture to ensure its initialized or binding the EGLImage will not work. So we leave the TEXTURE_MAX_LEVEL check.

### Motivation and Context

This makes screencapture up to 100x faster on discrete intel cards and likely has some performance benefit for amd/integrated cards. Without this dmabufs are actually slower than shared memory for these intel cards.

### How Has This Been Tested?

Tested with screencapture and texture encoding (pending another PR) and both continue to work with this change. If someone on AMD wants to confirm a performance improvement that would be nice.

### Types of changes

### Checklist:

### DIFF
--- a/libobs-opengl/gl-egl-common.c
+++ b/libobs-opengl/gl-egl-common.c
@@ -186,7 +186,7 @@ struct gs_texture *gl_egl_create_texture_from_eglimage(
 
 	struct gs_texture *texture = NULL;
 	texture = gs_texture_create(width, height, color_format, 1, NULL,
-				    GS_DYNAMIC);
+				    GS_GL_DUMMYTEX);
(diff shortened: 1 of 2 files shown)


kkartaltepe:
Also since this flag is not used anywhere in OBS I assume no one actually depends on the behavior of initializing this texture. But its possible someone depended on this outside of our repo and would be broken by this change.

kkartaltepe:
I can also confirm a ~50% reduction in capture overhead for xcomposite capture in a simple case.

GeorgesStavracas (approved):

PR #8253, opened 2023-02-11
UI: Defer loading hotkeys until tab is switched

derrod:
### Description

Waits with loading hotkeys until the user clicks on the settings tab.

The user sees a message while the tab is still loading:
2023-02-12_01-37-46_fIunMq

### Motivation and Context

Enables the settings window to still open really quickly even with tons of hotkeys (like #6081).

### How Has This Been Tested?

Opened settings, it was quick!

### Types of changes

### Checklist:

### DIFF
--- a/UI/data/locale/en-US.ini
+++ b/UI/data/locale/en-US.ini
@@ -1209,6 +1209,7 @@ Basic.Settings.Hotkeys.Pair="Key combinations shared with '%1' act as toggles"
 Basic.Settings.Hotkeys.Filter="Filter"
 Basic.Settings.Hotkeys.FilterByHotkey="Filter by Hotkey"
 Basic.Settings.Hotkeys.DuplicateWarning="This hotkey is shared by one or more other actions, click to show conflicts"
+Basic.Settings.Hotkeys.PleaseWait="Loading hotkeys, please wait..."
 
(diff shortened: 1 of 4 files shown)


PatTheMav:
Would it be possible to a) increase the font size of the message and b) make it horizontally and vertically centred in the container?

Because right now it's stuck in a visually busy area (the top with two text input boxes) that it's guaranteed to be overlooked.

derrod:
> Would it be possible to a) increase the font size of the message and b) make it horizontally and vertically centred in the container?

a) I think the size is fixed by QSS so that would require creating a new attribute and then adding to our QSS
b) It's harder than you'd think! (That is to say, I tried, but it didn't work)

PatTheMav:
> > Would it be possible to a) increase the font size of the message and b) make it horizontally and vertically centred in the container?
> 
> a) I think the size is fixed by QSS so that would require creating a new attribute and then adding to our QSS b) It's harder than you'd think! (That is to say, I tried, but it didn't work)

Meh, but good-enough then.

tuduweb:
It seems that the new feature will frozen the properties settings dialog, so can't change the current tab via click left tab items.
And I just run this in debug mode in vs studio, not release.

PR #8254, opened 2023-02-12
UI: Massively improve hotkey search performance

derrod:
### Description

This PR significantly improves search performance in the hotkey settings tab.

The original author had the right idea with disabling updates so the list would not be continuously updated/redrawn while the search loop is updating visibilities, but they set it on the wrong element. Disabling the layout instead improves things significantly with large scene collections.

Also includes some smaller optimisations to avoid some operations (`toLower()`/getting combinations) unless necessary which speeds up the loop a tiny bit.

### Motivation and Context

Makes search usable with the example collection linked in #6081.

### How Has This Been Tested?

Searched some hotkeys, no longer was afraid of dying of old age while it updates

### Types of changes

### Checklist:

### DIFF
--- a/UI/window-basic-settings.cpp
+++ b/UI/window-basic-settings.cpp
@@ -4519,7 +4519,6 @@ void OBSBasicSettings::SearchHotkeys(const QString &text,
 	std::vector<obs_key_combination_t> combos;
 	bool showHotkey;
 	ui->hotkeyScrollArea->ensureVisible(0, 0);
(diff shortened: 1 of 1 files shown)


flaeri:
As it says on the tin, massive improvement over current stock. Seems to be alright, didnt crash or missbehave. Its difficult to put this into numbers, but it feels a lot better. 

On stock, in certain situations I am waiting what feels like several minutes. With this PR I hardly ever wait more than a few sec.

Testing in win10 with our favortie Massive.zip.

PR #8258, opened 2023-02-13
UI: Remove unused Qt crash reporter code

gxalpha:
### Description

Removes the Qt crash reporter. The crash reporter made with Qt was introduced in 824c7b0, but then was quickly replaced by a native OS implementation in d42a7ce. This code has been lying around ever since.

### Motivation and Context

Stumbled across this while working on another PR.
Want less dead code, since despite being unused it still always gets processed by the Qt Meta-Object Compiler (and possibly C++-compiled as well, unsure).

### How Has This Been Tested?

macOS 13.2
Still compiles.
CI also still compiles.

Checked `git grep` for other occurrences of `OBSCrashReport` and `crash-report`, and there are none.

### Checklist:


cg2121:
I think we should test if this is more stable now, with newer versions of Qt. This way it might be possible to use also on macOS and Linux.

RytoEX (approved):
Seems fine to me.

RytoEX:
Merging this per @notr1ch off-thread:
> definitely remove, it's unsafe to do anything like further UI operations after a crash

PR #8261, opened 2023-02-13
UI: Correct browse behavior in non-empty input edit line

tuduweb:
### Description

Correct browse behavior in non-empty input edit line,
to keep single-file and multi-file (selected) behavior consistent

### Motivation and Context

Found when debugged

before:

https://user-images.githubusercontent.com/30559085/218437695-31b663bb-35bf-421f-818e-06c7a8b86c4d.mp4

after:

https://user-images.githubusercontent.com/30559085/218437754-34fbde2a-7c1c-41b3-ba53-63542d376b42.mp4

### How Has This Been Tested?

Just run again
win10, latest branch

### Types of changes

### Checklist:

### DIFF
--- a/UI/window-remux.cpp
+++ b/UI/window-remux.cpp
@@ -349,7 +349,7 @@ bool RemuxQueueModel::setData(const QModelIndex &index, const QVariant &value,
 				endRemoveRows();
 			}
 		} else {
-			if (pathList.size() > 1 &&
+			if (pathList.size() >= 1 &&
 			    index.row() < queue.length()) {
 				queue[index.row()].sourcePath = pathList[0];
 				checkInputPath(index.row());


PatTheMav:
Could you please put in writing what the observed behaviour is vs. the expected behaviour? The videos left me confused about what the supposed issue is (I'd also like to not exclude maintainers with disabilities).

tuduweb:
I'm sorry for confusing you.
The observed behaviour:
When click the non-empty row *browse* button, the file(s) selected via dialog:
files > 1:
input file in current row will be **replaced** by new one, and other files will be appended

file == 1:
just be appended, but not replaced.

This behavior make me confused. I think this is a bug.

And after this PR, those two cases will be similar, **be replaced**.

@PatTheMav

PatTheMav:
Got it, so it fixes the browse-button of existing entries to replace the given entry instead of appending by default?

@RytoEX would you agree that this is the expected behaviour for that button?

tuduweb:
> Got it, so it fixes the browse-button of existing entries to replace the given entry instead of appending by default?
> 
> @RytoEX would you agree that this is the expected behaviour for that button?

Yes. and this PR fixes it.

tuduweb:
Just fix the browse-button of existing entries when selecting one file.

And when selecting more than one file, it already replaced the given entry.

RytoEX (approved):
Looks good to me.

PR #8266, opened 2023-02-13
UI: Rename P1-P7 presets

matoi974:
### Description
Changes NVENC's P1-P7 presets to remove mention of "best quality" or "worst quality".

### Motivation and Context
People tend to choose P7 because it says "best quality" and then often end up having issues. Convincing someone to switch from P7 then becomes needlessly difficult, and so is explaining why it doesn't make a major difference in the CQP rate control mode.

### How Has This Been Tested?
N/A

### Types of changes
* Origin translation change

### Checklist:
* [ ]  My code has been run through [clang-format](https://github.com/obsproject/obs-studio/blob/master/.clang-format).
* [x]  I have read the [**contributing** document](https://github.com/obsproject/obs-studio/blob/master/CONTRIBUTING.rst).
* [x]  My code is not on the master branch.
* [ ]  The code has been tested.
* [x]  All commit messages are properly formatted and commits squashed where appropriate.
* [ ]  I have included updates to all appropriate documentation.


RytoEX:
Not describing P1-P7 just means people will have to look them up, inevitably leading to advice that says P7 is best, or come and ask in Discord or the Forums which one gives best quality and why they shouldn't pick the one farthest down when suggested to pick something else. On its own, I don't see how this is significantly better.

matoi974:
The "fastest"/"slowest" parts are still there; in context of troubleshooting, it's much more useful to a person without background knowledge on this encoder, as GPU usage will be lower for presets below P3, and higher for presets above. As it stands currently, no one wants to switch to P1 because it says "worst quality", even though it has virtually no impact on quality in CQP. Besides, the answer to "which one yields best quality" will depend on what kind of preset a given PC can reliably manage in a given workload, and lower presets may yield "better" quality in the sense that they would have fewer performance issues and therefore less duplicated frames. I've seen people try to push 4K60 at p7, even though it makes no sense for them to do so, and failed to convince them that they don't stand to gain anything practical from such efforts.

Fenrirthviti:
I'm open to the idea of adjusting the wording to something else, but I think removing the "layman" description entirely is not going to improve anything. I also question that this is even an issue in the first place, and there is some level of bias that it is.

dodgepong:
I think the "quality" labels are a red herring. The problem is that "fast" vs. "slow" are bad descriptors of what the difference is. "Fast" means "best performance" and "slow" means "worst performance", but I don't think that's intuitive for users. If users know that "best performance" also corresponds with "worst quality", I suspect they would be less likely to make the mistake. Same for P7 with "worst performance, best quality". I think as long as users better understand the tradeoff, they will be better equipped to make the right decision. Removing information, as in this PR, does not equip them to make that decision.

flaeri:
Labeling the presets with "quality" makes sense for VBR and CBR, but not for CQP.

P1 is the highest quality for CQP, but the largest, and quality will ever so slightly degrade, while filesize also comes down as you increase (+) the preset, which is counter to what the user would expect. You could argue the quality difference is negligable, and I'd probably concede that point.

Most users I've spoken to about this was firmly under the impression that their recordings would suffer a sizable hit to quality if they were to pick the lower presets. Its often a difficult to convince them because at face value, OBS is telling you that you're choosing low/lowest quality.

I personally wouldnt mind removing the "quality" term from the presets, because I selfishly do not wish to spend time arguing whether a users videos will be low quality when selecting p1-p3, which is more or less required for recording 4k60.

lindenkron:
As per the discussion on Discord:
I don't think the laymans term for quality is the problem here. The issue is there's no layman indication of the tradeoff when selecting any of these settings. **Slower** and **Faster** does not indicate, to anyone not tech savvy, that it significantly increases the processing power of your hardware.

I'm not even entirely sure this should be available in simple. I can see an argument for switching to a faster preset on worse hardware, but in majority of cases unless you _know_ what you're doing, you should only be moving this to a slower preset on more advanced setups/equipment. Which would warrant a move to advanced.

Relabeling things without causing a whole write-up per item is also hard, eg:
P6: Slower (Better Quality, High processing power requirement)
P7: Slowest (Best Quality, Highest processing power requirement)

If this is available in **Simple** there should be a better/alternative description on the labels, not less.

Fenrirthviti:
After internal discussion, we do not feel this is the right way to handle this, and don't feel the problem of "users will slam all settings to max and be confused why it doesn't work" is something that is really solvable. As such, it should not drive design or UI/UX choices. We do, however, feel there are ways to improve on the current presentation, just not like this.

Thank you for the submission.

PR #8267, opened 2023-02-13
UI: Remove bitness strings

cg2121:
### Description
This removes the bitness strings from the title bar and about dialog.

### Motivation and Context
These are now pointless because we are now only 64 bit.

### How Has This Been Tested?
Checked UI to see if the strings were removed.

### Types of changes

### Checklist:

### DIFF
--- a/UI/obs-app.cpp
+++ b/UI/obs-app.cpp
@@ -1722,11 +1722,6 @@ string OBSApp::GetVersionString() const
 	ver << " (";
 
 #ifdef _WIN32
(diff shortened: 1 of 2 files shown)


WizardCM:
I'm in favour of removing unnecessary text from the titlebar, but it's worth noting that even now, it's possible for a user to be using, say, the x64 build on an M1 Mac, when they should be using the ARM build., I think we should keep these strings in the About dialog (and the Scripting dialog), and just trim down the unnecessary strings in the titlebar (imo bitness and platform).

PR #8271, opened 2023-02-14
libobs-opengl: Fix projector crash with external macOS displays

PatTheMav:
### Description
Fixes a crash occurring whenever a full screen projector was placed on an attached display, which is then disconnected while OBS runs.

### Motivation and Context
When an external display is disconnected with a fullscreen projector attached to it, Qt will trigger a resize of the window twice, which makes the renderer queue 2 resize event blocks.

At the time when those blocks are run, Qt's window destructor will have reset all pointers, and the block directly accesses pointers within structures identified by pointers, which are invalid by that point.

This commit makes a block return early if the associated window has been destroyed already and also explicitly checks for valid pointers after.

### How Has This Been Tested?
Tested with an external display connected via HDMI on macOS Ventura.

### Types of changes

### Checklist:

### DIFF
--- a/libobs-opengl/gl-cocoa.m
+++ b/libobs-opengl/gl-cocoa.m
@@ -228,6 +228,10 @@ void gl_update(gs_device_t *device)
 	NSOpenGLContext *parent = device->plat->context;
 	NSOpenGLContext *context = swap->wi->context;
 	dispatch_async(dispatch_get_main_queue(), ^() {
+		if (!swap || !swap->wi) {
+			return;
+		}
+
 		CGLContextObj parent_obj = [parent CGLContextObj];
 		CGLLockContext(parent_obj);
 


libobs-opengl/gl-cocoa.m
```diff
@@ -228,6 +228,10 @@ void gl_update(gs_device_t *device)
 	NSOpenGLContext *parent = device->plat->context;
 	NSOpenGLContext *context = swap->wi->context;
 	dispatch_async(dispatch_get_main_queue(), ^() {
+		if (!swap || !swap->wi) {
+			return;
+		}
```
gxalpha, line 233: How do we feel about curly braces around one liners again? I thought we wanted to avoid them but I'm not sure.

gxalpha (approved):
Womm.

libobs-opengl/gl-cocoa.m
PatTheMav: Discussed this off-thread, we're safely in "let's be explicit" territory now. 😄

PR #8272, opened 2023-02-14
linux-pipewire: Demote error to debug message

GeorgesStavracas:
### Description

Receiving buffers flagged as corrupted, or empty, is a casualty of how things are implemented in various compositors, and it's not a critical situation to be afraid of. We can expect a few buffers to be flagged as corrupted here and there, and that's fine.

Demote these warnings to debug messages, as they're still useful for debugging.

We were spamming the log with messages that are actually fine.

### Motivation and Context

Spamming log bad, clean log good

### How Has This Been Tested?

 - Run OBS Studio in Wayland
 - Notice that the stdout / stderr log is decent

### Types of changes

### Checklist:

### DIFF
--- a/plugins/linux-pipewire/pipewire.c
+++ b/plugins/linux-pipewire/pipewire.c
@@ -567,13 +567,13 @@ static void on_process_cb(void *user_data)
 
 		if ((buffer->datas[0].chunk->flags & SPA_CHUNK_FLAG_CORRUPTED) >
 		    0) {
-			blog(LOG_ERROR,
+			blog(LOG_DEBUG,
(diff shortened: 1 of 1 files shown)


RytoEX (approved):
Seems straightforward enough. Looks good to me.

GeorgesStavracas:
Thank you for the review

fzwoch:
What about this one:

https://github.com/obsproject/obs-studio/blob/19a0afe5268f55cf360b8c6ce324991ae92ea5cc/plugins/linux-pipewire/pipewire.c#L530-L534

I think this one is in the DMABUF case and I believe that one is still spamming my log.

PR #8276, opened 2023-02-16
UI: Force Wayland usage on Ubuntu GNOME

tytan652:
### Description

Apparently XDG_CURRENT_DESKTOP returns `ubuntu:GNOME` on Ubuntu.

### Motivation and Context

Ubuntu bypass our workaround which is needed for Ubuntu versions that we support that provides Qt version earlier than 6.3.

### How Has This Been Tested?

Needs to be tested.

### Types of changes

### Checklist:

### DIFF
--- a/UI/obs-app.cpp
+++ b/UI/obs-app.cpp
@@ -2322,7 +2322,7 @@ static int run_program(fstream &logFile, int argc, char *argv[])
 
 	const char *desktop = getenv("XDG_CURRENT_DESKTOP");
 	const char *session_type = getenv("XDG_SESSION_TYPE");
-	if (session_type && desktop && strcmp(desktop, "GNOME") == 0 &&
+	if (session_type && desktop && strstr(desktop, "GNOME") != nullptr &&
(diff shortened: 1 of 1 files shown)


kkartaltepe (approved):
Seems fine, i feel like we usually do the check against `nullptr` explicitly instead of `!!` but either way is fine.

UI/obs-app.cpp
```diff
@@ -2322,7 +2322,7 @@ static int run_program(fstream &logFile, int argc, char *argv[])
 
 	const char *desktop = getenv("XDG_CURRENT_DESKTOP");
 	const char *session_type = getenv("XDG_SESSION_TYPE");
-	if (session_type && desktop && strcmp(desktop, "GNOME") == 0 &&
+	if (session_type && desktop && !!strstr(desktop, "GNOME") &&
```
norihiro, line 2325: There are more usage of `strstr() != NULL`  than `!!strstr`.
```suggestion
	if (session_type && desktop && strstr(desktop, "GNOME") != NULL &&
```

PR #8277, opened 2023-02-16
UI: Fix transform dialog not being closable

cg2121:
### Description
If the user clicked away from the source, while the transform dialog is open, the whole dialog would be disabled. The user then couldn't click close or press escape to close the dialog. The only way for them to close it was to click on another source to re-enable it.

### Motivation and Context
Fixes bug I've found.

### How Has This Been Tested?
Click away from a source and made sure the dialog was closable.

### Types of changes

### Checklist:

### DIFF
--- a/UI/forms/OBSBasicTransform.ui
+++ b/UI/forms/OBSBasicTransform.ui
@@ -15,510 +15,628 @@
   </property>
   <layout class="QVBoxLayout" name="verticalLayout" stretch="0,5,0">
    <item>
-    <layout class="QFormLayout" name="formLayout">
-     <property name="fieldGrowthPolicy">
-      <enum>QFormLayout::AllNonFixedFieldsGrow</enum>
-     </property>
-     <property name="labelAlignment">
-      <set>Qt::AlignRight|Qt::AlignTrailing|Qt::AlignVCenter</set>
-     </property>
-     <item row="0" column="0">
-      <widget class="QLabel" name="label">
-       <property name="sizePolicy">
-        <sizepolicy hsizetype="Preferred" vsizetype="Minimum">
-         <horstretch>0</horstretch>
-         <verstretch>0</verstretch>
-        </sizepolicy>
-       </property>
-       <property name="minimumSize">
-        <size>
-         <width>170</width>
-         <height>0</height>
-        </size>
-       </property>
-       <property name="accessibleName">
-        <string>Basic.TransformWindow.Position</string>
-       </property>
-       <property name="text">
-        <string>Basic.TransformWindow.Position</string>
-       </property>
-       <property name="alignment">
-        <set>Qt::AlignRight|Qt::AlignTrailing|Qt::AlignVCenter</set>
-       </property>
-      </widget>
-     </item>
-     <item row="0" column="1">
-      <widget class="QFrame" name="widget">
-       <property name="sizePolicy">
-        <sizepolicy hsizetype="Fixed" vsizetype="Preferred">
-         <horstretch>0</horstretch>
-         <verstretch>0</verstretch>
-        </sizepolicy>
-       </property>
-       <layout class="QHBoxLayout" name="horizontalLayout">
-        <property name="leftMargin">
-         <number>0</number>
+    <widget class="QWidget" name="container" native="true">
+     <layout class="QFormLayout" name="formLayout">
+      <property name="fieldGrowthPolicy">
+       <enum>QFormLayout::AllNonFixedFieldsGrow</enum>
+      </property>
+      <property name="labelAlignment">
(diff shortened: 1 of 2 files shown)


sebastian-s-beckmann (commented):
I think the better fix here would be encapsulating `formLayout` in another `QWidget` and then just disabling that widget plus the "Reset" button instead. The resulting git diff would be a bit annoying since everything would be indented by one more space, but it'd be much less and much cleaner code than this current PR.

cg2121:
I've put the layout inside of a QFrame, so we can now just disable it instead.

jp9000:
Why QFrame rather than QWidget?

gxalpha:
To elaborate on Jim's comment, I don't see the value in using a QFrame here.

As stated by the [documentation](https://doc.qt.io/qt-6/qframe.html), "QFrame [...] is the base class of widgets that can have a frame". Using a QFrame with `NoFrame` is a bit pointless, it could just be a simple QWidget instead.
As far as I understand (and I'm by no means an expert), `NoFrame` is made for cases where you sometimes want a frame and sometimes not, and have to dynamically switch between them.

Of course, if there are specific QSS-related concerns that can only be solved with QFrames, it would be alright to use it; but if it's just because we use QFrame as a container everywhere that seems like a general problem across the program to me.

cg2121:
QWidgets can't be styled with QSS.

gxalpha:
> QWidget containers can't be styled with QSS.

I get that part.
What I'm asking is what the advantages of a QFrame are *in this specific scenario*, since - unless I'm mistaken - we're not looking to do any styling that would require a QFrame here. Is there something that requires the QFrame?

cg2121:
Updated to change the QFrame to a QWidget.

UI/forms/OBSBasicTransform.ui
```diff
@@ -6,213 +6,172 @@
    <rect>
     <x>0</x>
     <y>0</y>
-    <width>564</width>
-    <height>313</height>
+    <width>619</width>
+    <height>329</height>
```
gxalpha, line 10: Is this change intentional?

UI/forms/OBSBasicTransform.ui
```diff
@@ -226,592 +185,638 @@
...
-     <property name="orientation">
-      <enum>Qt::Vertical</enum>
-     </property>
-     <property name="sizeHint" stdset="0">
-      <size>
-       <width>0</width>
-       <height>10</height>
-      </size>
-     </property>
-    </spacer>
```
gxalpha, line 814: Is the removal of this intentional?

sebastian-s-beckmann (commented):
In addition to the two instances I marked below, there are a few buddy properties (cropLeft/cropBottom/cropTop/cropRight) getting removed. Are those intentional or just something Qt Creator decided to do on its own?
If they are intentional, could you quickly explain the motivation behind them? It doesn't *look* like anything's broken at first glance (We're not using ampersands for the label texts so the buddies currently [are likely a no-op anyways](https://doc.qt.io/qt-6/qlabel.html#setBuddy)) but a short explanation might be useful for future reference.

Other than that, this looks fine.

UI/forms/OBSBasicTransform.ui
cg2121: Fixed

UI/forms/OBSBasicTransform.ui
cg2121: Fixed

UI/forms/OBSBasicTransform.ui
```diff
@@ -15,204 +15,157 @@
...
-        <property name="rightMargin">
-         <number>0</number>
-        </property>
-        <property name="bottomMargin">
-         <number>0</number>
-        </property>
-        <item>
-         <widget class="QDoubleSpinBox" name="positionX">
+    <widget class="QWidget" name="container" native="true">
+     <layout class="QGridLayout" name="gridLayout_2">
```
gxalpha, line 19: Apologies for only seeing this now, but is this additional grid layout intentional?
I hope "Hide Whitespaces" isn't deceiving me here, but it looks like only (0, 0) of the layout gets filled with an item (the form layout), in which case I think the grid layout would be redundant? Hope I'm not missing something in the diff here.

cg2121: A widget needs to have a layout to have other widgets inside of it.

gxalpha: Sure, but - as far as I understand - here we have

`container` (widget)
\- `gridLayout_2` (layout)
\- \- [0, 0] `formLayout` (layout)

So gridLayout seems redundant, doesn't it?

cg2121: Updated to remove the extra layout.

sebastian-s-beckmann (approved):

PR #8278, opened 2023-02-16
UI: Connect Reset button in Transform menu with main window directly

gxalpha:
### Description

Removes the detour of going via an extra slot and connects the clicked signal of the reset button in the Edit Transform dialog to the main window directly.

### Motivation and Context

Qt threw `QMetaObject::connectSlotsByName: No matching signal for on_resetButton_clicked` at us because the slot wasn't connected automatically. This made me sad.
I was going to just rename it (which would be a correct fix, the `on_foo_bar` names should only be used if the signal is (primarily) connected automatically), however I noticed that the slot function just called a slot of the main window. We can get rid of this complication by connecting directly.

### How Has This Been Tested?

Tested that "Reset" button in the Edit Transform menu still works.

### Types of changes

### Checklist:

### DIFF
--- a/UI/window-basic-transform.cpp
+++ b/UI/window-basic-transform.cpp
@@ -65,7 +65,8 @@ OBSBasicTransform::OBSBasicTransform(OBSSceneItem item, OBSBasic *parent)
 	ui->buttonBox->button(QDialogButtonBox::Close)->setDefault(true);
 
 	connect(ui->buttonBox->button(QDialogButtonBox::Reset),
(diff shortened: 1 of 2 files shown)


RytoEX (changes requested):
Nit in commit message: menu -> dialog

While that ends up just barely under 72 characters, if that's too long for comfort, maybe split up the message into a more concise subject and a brief explainer message body.

Otherwise, looks good to me.

PR #8281, opened 2023-02-16
UI: Add settings reset button

cg2121:
### Description
This adds a reset button in the settings dialog to reset the current tab to its default settings.

Screenshot from 2023-02-16 15-13-55

### Motivation and Context
Currently there is no way to reset the settings to defaults.

### How Has This Been Tested?
Clicked the reset button to make sure the settings were set to the defaults.

### Types of changes

### Checklist:

### DIFF
--- a/UI/data/locale/en-US.ini
+++ b/UI/data/locale/en-US.ini
@@ -890,6 +890,8 @@ Basic.Settings.General.MultiviewLayout.4Scene="Scenes only (4 Scenes)"
 Basic.Settings.General.MultiviewLayout.9Scene="Scenes only (9 Scenes)"
 Basic.Settings.General.MultiviewLayout.16Scene="Scenes only (16 Scenes)"
 Basic.Settings.General.MultiviewLayout.25Scene="Scenes only (25 Scenes)"
+Basic.Settings.General.Reset.Title="Reset General Settings"
+Basic.Settings.General.Reset.Text="Are you sure you want to reset the general settings?"
 
 # default channel name translations
 Basic.Settings.General.ChannelName.stable="Stable"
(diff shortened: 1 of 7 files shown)


Fenrirthviti:
For clarification here, when you say "the settings" you are referring specifically to general settings, correct?

I'm not sure putting the reset button in the larger dialog is correct here, unless it's expected to work in every tab in settings. Many of the other tabs are already controlled by scene collections and profiles, so there might be a small conflict/overlap here in user expectations.

cg2121:
Works for whatever the current tab is.

prgmitchell:
Tested with a few different collections and all works as expected. People often ask how to reset their settings in the support channels and the only current solution is to create a new profile. This of course isn't ideal when someone only wants to reset one part of their settings. I can see what Fenrir is saying considering the "Apply" button in the same section of the UI would apply all settings, not just the settings on that tab. I personally don't see this as an issue for the user experience as the confirmation dialog makes it clear what is happening and what exactly is being reset.

Fenrirthviti:
Screenshots of said dialog in the PR description would be useful then, as I didn't even know there was a confirmation based on the information provided.

WizardCM:
I'm personally not sold on this feature. While the UI provides no way to reset settings, it can be done by deleting the `global.ini` or creating a new Profile (depending on the tab). Additionally, we don't often have users request the ability to reset to default.

WizardCM:
Thinking about this further, personally the only areas that would benefit strongly from a Reset button are the Advanced "Video" and "Recording" sections. Users experiment with colour settings and filename formatting, and have no clear way to revert to something functional. Every other setting has much clearer default values, either a "Default" item in the dropdown or a blank field as default.

GeorgesStavracas:
As per today's discussion, Settings needs a bigger reorganization and by the time it happens, this PR will probably not apply. Also, this direction for Defaults is not consensus. Let's discuss how we handle resetting defaults further in the context of updating the Settings dialog.

PR #8283, opened 2023-02-17
libobs: Render audio for all views

exeldro:
### Description
Render audio for all views

### Motivation and Context
At the moment you can add views to obs via `obs_view_add` or `obs_view_add2`, but the audio for those views was not rendered.
Not rendered audio causes transitions not to finish and scenes to forever wait on audio to deactivate sources.

### How Has This Been Tested?
With the vertical plugin on windows

### Types of changes

### Checklist:

### DIFF
--- a/libobs/obs-audio.c
+++ b/libobs/obs-audio.c
@@ -542,18 +542,34 @@ bool audio_callback(void *param, uint64_t start_ts_in, uint64_t end_ts_in,
 #endif
 
 	/* ------------------------------------------------ */
-	/* build audio render order
-	 * NOTE: these are source channels, not audio channels */
-	for (uint32_t i = 0; i < MAX_CHANNELS; i++) {
-		obs_source_t *source = obs_get_output_source(i);
-		if (source) {
+	/* build audio render order */
+
+	pthread_mutex_lock(&obs->video.mixes_mutex);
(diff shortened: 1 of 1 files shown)


tt2468:
Wouldn't this mean that audio for a source on one view and not on the main program mix would play anyway on the audio feed from that mix?

exeldro:
No, all audio stays on its own view.
All audio sources are always rendered, this only makes scenes and transitions that are on other views mix their audio so that view works correctly and allows to use the audio for that view (which is not required).

exeldro:
comment style fixed and empty newlines added

tt2468 (approved):

PatTheMav:
If I understand correctly, this change ensures that audio is rendered (but not output) for all sources even if they are not visible or part of an active output, because many sources rely on the audio render loop happening?

derrod:
This PR still skips inactive sources, but checks all views instead of only the main one. So if you happen to have another view, say with the [vertical output plugin](https://obsproject.com/forum/resources/aitum-vertical.1715/), it will also render the audio on that view.

PatTheMav:
Yep I saw the check for inactive sources, so I guess this should be good to merge?

Lain-B (approved):

PR #8293, opened 2023-02-19
pipewire: calc spa buffer size

columbarius:
### Description

Calculate dynamically the size for the PipeWire params.

### Motivation and Context

Modern GPUs might support a lot of modifiers. As such the hardcoded size for the ENUM_Formats PipeWire params are not enough and we should calculate the required size manually.

Added logs if a ENUM_Format can't be assembled, which happens when the buffer size was not large enough.

### How Has This Been Tested?

Start a screencast. If the buffer is to small for all formats, it will quit it.

### Types of changes

### Checklist:

### DIFF
--- a/plugins/linux-pipewire/pipewire.c
+++ b/plugins/linux-pipewire/pipewire.c
@@ -144,6 +144,41 @@ static void update_pw_versions(obs_pipewire *obs_pw, const char *version)
 		blog(LOG_WARNING, "[pipewire] failed to parse server version");
 }
 
+static uint32_t calc_spa_enumformat_buffer_size(obs_pipewire *obs_pw)
+{
+	// from spa/pod/pod.h
+	uint32_t spa_pod_object_size = sizeof(struct spa_pod_object);
+	uint32_t spa_pod_prop_size = sizeof(struct spa_pod_prop);
+	uint32_t spa_pod_value_size = 1 * sizeof(uint64_t);
+	uint32_t spa_pod_choice_size = 2 * sizeof(uint64_t);
+	// This is counted from the build_format function
+	uint32_t n_objects = 0;
+	uint32_t n_props = 0;
+	uint32_t n_values = 0;
+	uint32_t n_choices = 0;
+	uint32_t size;
+
+	for (size_t i = 0; i < obs_pw->format_info.num; i++) {
+		if (obs_pw->format_info.array[i].modifiers.num == 0)
+			continue;
+		n_objects += 1;
+		n_props += 5 + 1;
+		n_values += 9 + 2 + obs_pw->format_info.array[i].modifiers.num;
+		n_choices += 2 + 1;
+	}
+	for (size_t i = 0; i < obs_pw->format_info.num; i++) {
+		n_objects += 1;
+		n_props += 5;
+		n_values += 9;
+		n_choices += 2;
+	}
+	size = n_objects * spa_pod_object_size + n_props * spa_pod_prop_size +
+	       n_values * spa_pod_value_size + n_choices * spa_pod_choice_size;
+	blog(LOG_DEBUG, "[pipewire]: calculated spa enumFormat buffer size %u",
+	     size);
+	return size;
+}
+
 static void teardown_pipewire(obs_pipewire *obs_pw)
 {
 	if (obs_pw->thread_loop) {
@@ -408,6 +443,10 @@ static bool build_format_params(obs_pipewire *obs_pw,
 			obs_pw->format_info.array[i].spa_format,
 			obs_pw->format_info.array[i].modifiers.array,
 			obs_pw->format_info.array[i].modifiers.num);
+		if (!params[params_count - 1]) {
+			blog(LOG_ERROR, "[pipewire] Failed to format param");
+			return false;
+		}
 	}
 
 build_shm:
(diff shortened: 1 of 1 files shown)


columbarius:
It works, but my calculated buffer sizes are 1 `uint64_t` bigger than the one shown by `spa_pod_debug`. For shm buffer I had to add the difference manually, might have to take a look again if I missed an element.

plugins/linux-pipewire/pipewire.c
```diff
@@ -521,7 +564,7 @@ static void renegotiate_format(void *data, uint64_t expirations)
 
 	pw_thread_loop_lock(obs_pw->thread_loop);
 
-	uint8_t params_buffer[2048];
+	uint8_t params_buffer[calc_spa_enumformat_buffer_size(obs_pw)];
```
GeorgesStavracas, line 567: At this point, I'd prefer that we bzalloc / bfree instead...

plugins/linux-pipewire/pipewire.c
```diff
@@ -144,6 +144,41 @@ static void update_pw_versions(obs_pipewire *obs_pw, const char *version)
 		blog(LOG_WARNING, "[pipewire] failed to parse server version");
 }
 
+static uint32_t calc_spa_enumformat_buffer_size(obs_pipewire *obs_pw)
+{
+	// from spa/pod/pod.h
+	uint32_t spa_pod_object_size = sizeof(struct spa_pod_object);
+	uint32_t spa_pod_prop_size = sizeof(struct spa_pod_prop);
+	uint32_t spa_pod_value_size = 1 * sizeof(uint64_t);
+	uint32_t spa_pod_choice_size = 2 * sizeof(uint64_t);
```
GeorgesStavracas, line 153: I think I prefer if you didn't use variables for these

plugins/linux-pipewire/pipewire.c
```diff
@@ -144,6 +144,41 @@ static void update_pw_versions(obs_pipewire *obs_pw, const char *version)
...
+		n_props += 5 + 1;
+		n_values += 9 + 2 + obs_pw->format_info.array[i].modifiers.num;
+		n_choices += 2 + 1;
+	}
+	for (size_t i = 0; i < obs_pw->format_info.num; i++) {
+		n_objects += 1;
+		n_props += 5;
+		n_values += 9;
+		n_choices += 2;
+	}
```
GeorgesStavracas, line 174: Can you merge both loops, please?

plugins/linux-pipewire/pipewire.c
```diff
@@ -144,6 +144,41 @@ static void update_pw_versions(obs_pipewire *obs_pw, const char *version)
...
+		n_objects += 1;
+		n_props += 5 + 1;
+		n_values += 9 + 2 + obs_pw->format_info.array[i].modifiers.num;
+		n_choices += 2 + 1;
+	}
+	for (size_t i = 0; i < obs_pw->format_info.num; i++) {
+		n_objects += 1;
+		n_props += 5;
+		n_values += 9;
+		n_choices += 2;
```
GeorgesStavracas, line 173: I'm a bit concerned about these numbers. Where do they come from? Can they change beneath our feet?

GeorgesStavracas (changes requested):

plugins/linux-pipewire/pipewire.c
columbarius: These are counted from how we build our `EnumFormat` pods.

columbarius:
As an alternative PipeWire supports the `dynamic_builder`. It will reallocate memory when required, but this has the drawback that we can only build one thing per builder, since a rebase might create dangling pointers from previous created pods. We would have to create 2*n_formats `dynamic_builders`.

plugins/linux-pipewire/pipewire.c
columbarius: for value_pods we could use a specific one since they are all padded to uint64_t.
choice_pods I'd have to investigate.

columbarius:
The `dynamic_builder` uses the provided buffer until it's size is exceeded. If that happens, it will allocate a new, larger buffer and copy the first buffer into it. If the size of the second buffer is exceeded, realloc is called on it. At this time the third buffer (after realloc) might be at another location in memory. `spa_builder_*_pop` will return a pointer to the current memory location of a pod. Any pointer to the second buffer might be invalid after the reallocation to the third buffer. Thus we can't use one `builder` for more than one pod.

norihiro:
Would you revise the commit message like below?
The module name prefix should be `linux-pipewire`. The first commit word of the commit title should be capitalized. The article of ENUM_Format should be `an` instead of `a`.
```plain
linux-pipewire: Calculate SPA buffer size

Modern GPUs might support a lot of modifiers. As such the hardcoded size
for the ENUM_Formats PipeWire params are not enough and we should
calculate the required size manually.

Added logs if an ENUM_Format can't be assembled, which happens when the
buffer size was not large enough.
```

PatTheMav:
Closing this as open feedback comments have not been addressed in over half a year.

PR can be reopened if feedback is addressed.

columbarius:
This MR is stalled, waiting for an answer to https://github.com/obsproject/obs-studio/pull/8293#issuecomment-1470761876

The issue is still unresolved. Please reopen.

@norihiro Thanks, will update the message after an answer

PR #8296, opened 2023-02-19
UI: Don't double-delete children of deleted widgets

gxalpha:
### Description

These widgets are all children of updateSettingsGroupBox, meaning they get deleted when updateSettingsGroupBox gets deleted. This means that the first two calls are unnecessary and the third one segfaults as updateChannelLabel is gone already.
Just delete the parent widget instead.

### Motivation and Context

Noticed by @fzwoch in https://github.com/obsproject/obs-studio/commit/12a27d8b99473d00d8939a78601c3ebcd6b2a2a9#r101210111.

### How Has This Been Tested?

Hasn't.
But should work according to my knowledge of Qt and also it can't get any worse than segfaulting.

### Types of changes

### Checklist:


fzwoch:
Works for me. Thanks, I was about to look up widget ownership in Qt..

PR #8301, opened 2023-02-19
UI: Support platform-specific WhatsNew entries

derrod:
### Description

Adds support for targeting specific platforms with WhatsNew entries, e.g. in case an issue or feature only applies to one or multiple but not all.

It adds a simple `os` object that can be added to each entry. If `os` is unspecified, an entry is assumed to be OS-agnostic. If an OS is ommited from the list it is implicitly assumed to be `false`.

Example:
```diff
[
  {
    "version": "29.0.0",
    "url": "https://obsproject.com/startup/hypothetical-windows-issue",
    "increment": 2,
+   "os": {
+     "windows": true,
+     "macos": false
+   }
  }
]
```

### Motivation and Context

Without this we'd rely on users reading comprehension, which is a bad idea.

### How Has This Been Tested?

Tested whatsnew files with various combinations, verified the notification only shows when it should.

### Types of changes

### Checklist:

### DIFF
--- a/UI/window-basic-main.cpp
+++ b/UI/window-basic-main.cpp
@@ -2194,6 +2194,20 @@ void OBSBasic::ReceivedIntroJson(const QString &text)
 	/* check to see if there's an info page for this version */
 	const Json::array &items = json.array_items();
 	for (const Json &item : items) {
+		if (item["os"].is_object()) {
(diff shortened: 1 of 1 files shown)


RytoEX:
On the surface, this seems fine.

PR #8303, opened 2023-02-20
libobs: Fix pulseaudio monitoring, once and for all

### LINKED ISSUE #7574: Crash in circlebuf_peek_front during obs_source_output_audio
### Operating System Info

Other

### Other OS

Linux Mint 21 Cinnamon

### OBS Studio Version

28.0.3

_No response_

### OBS Studio Log URL

https://obsproject.com/logs/yMvMEwcRm3V2u_mF https://obsproject.com/logs/Lnk8HfTB4pMAP0W9 https://obsproject.com/logs/bIVWQ-oPx7I1B9pD

### OBS Studio Crash Log URL

_No response_

### Expected Behavior

To not close down and/or crash during a stream

### Current Behavior

OBS stalls out for about 10 seconds and then closes without any error message.

### Steps to Reproduce

1.Start streaming on twitch
2.stream for about an hour
3.OBS crashes
...

I use the current version of obs teleport and I compile and use the current version of obs multi rtmp from github.

kkartaltepe:
### Description

This makes multiple changes to resolve issues with monitoring. We have added a number of things over time to resolve various issues due to our misunderstanding of the pulse api and the explicitly wrong pulse api documentation.

First is removing the prior stream write tracking. Write tracking was incorrect as pulse sends you the total number of bytes it wants written at that time and not a incremental, but also this write tracking is unhelpful since we can just let begin_write signal when we should stop writing.

Second removing underflow tracking, this sort of worked to detect high latency sources like vlc which buffer ~1s of data before writing audio data. It worked because it would grow the buffer in PA until we could write enough data into it, but it didn't allow us to leverage that larger buffer because we never paused the PA stream.

To solve the buffering/high latency issue instead start the PA stream corked and only uncork it once we have enough data to reach the targeted latency. We also check how much data we recieve from the source and if it appears to be high latency providing us with much more data than our target latency we increase the latency to match the source. Fixing the VLC source issue and resulting in a smooth start to audio instead of underruns while we play and increase latency.

Finally add some handling for pulses incorrectly documented API. Notably that begin_write may request more or less data from you so handle both cases. And stop attempting to use attr.maxlength which is not updated after set_buffer_attr calls and was always -1, and tlength is just more appropriate.

old mantis resulting in some changes, still fixed. https://obsproject.com/mantis/view.php?id=1076
fixes #7574

### Motivation and Context

It finally seems we are using the api as intended and the PA no longer spams us requested more data for a setup with a browser+vlc+media source playing audio, we are correctly keeping ahead of the stream.

And more importantly we no longer crash when PA requests more data than we have in our buffer.

### How Has This Been Tested?

Run with a setup of vlc/media/browser sources all being monitored.

### Types of changes

### Checklist:

### DIFF
--- a/libobs/audio-monitoring/pulse/pulseaudio-output.c
+++ b/libobs/audio-monitoring/pulse/pulseaudio-output.c
@@ -20,7 +20,6 @@ struct audio_monitor {
 
 	struct circlebuf new_data;
 	audio_resampler_t *resampler;
-	size_t bytesRemaining;
 
 	bool ignore;
 	pthread_mutex_t playback_mutex;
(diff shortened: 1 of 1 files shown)


libobs/audio-monitoring/pulse/pulseaudio-output.c
```diff
@@ -190,26 +189,51 @@ static void do_stream_write(void *param)
...
+		data->attr.fragsize = (uint32_t)-1;
+		data->attr.maxlength = (uint32_t)-1;
+		data->attr.prebuf = (uint32_t)-1;
+		data->attr.minreq = (uint32_t)-1;
+		data->attr.tlength = data->new_data.size;
+		pa_stream_set_buffer_attr(data->stream, &data->attr, NULL,
+					  NULL);
+	}
+
+	// buffer up enough data before we start playing.
```
RytoEX, line 206: If writing in sentence style, capitalize the first word.

libobs/audio-monitoring/pulse/pulseaudio-output.c
```diff
@@ -190,26 +189,51 @@ static void do_stream_write(void *param)
...
+		size_t bytesToFill = data->new_data.size;
 		if (pa_stream_begin_write(data->stream, (void **)&buffer,
-					  &bytesToFill)) {
-			pulseaudio_unlock();
-			return;
+					  &bytesToFill))
+			goto finish;
+
+		// PA may request we submit more or less data than we have.
+		// wait for more data if we cannot perform a full write.
```
RytoEX, line 222: If writing in sentence style, capitalize the first word.

RytoEX (changes requested):
Do not use sentence final punctuation in the commit message subject.

tuduweb:
https://github.com/obsproject/obs-studio/blob/master/CONTRIBUTING.rst#commit-guidelines

It seems that the commit title shouldnt be followed by punctuation.

ogmkp:
In test here, looks promising.

kkartaltepe:
another user in discord tested this and had success with avoiding crashes. For latency it seems some sources can still produce faster than PA can consume (browser source in this case) and fall out of sync. This is equivalent to the old behavior but now the buffer is mostly in OBS instead of PA so we could take some kind of action. I don't have any intention to fix this though if someone knows how to tweak playback rate in pulse that would be a good follow up to manage latency better.

ogmkp:
> another user in discord tested this and had success with avoiding crashes. For latency it seems some sources can still produce faster than PA can consume (browser source in this case) and fall out of sync after 10+ hours. This is equivalent to the old behavior but now the buffer is mostly in OBS instead of PA so we could take some kind of action. I don't have any intention to fix this though if someone knows how to tweak playback rate in pulse that would be a good follow up to manage latency better.

Yes, we need a true pipewire monitoring output.
For this PR, no latency drift after 3,5 hours here on MIC monitoring and a very very little latency on media source plugin.
And unfortunatly VLC is too bad like said.

CodexHere:
Hey this originated from my repeated requests in Discord...

I've tested multiple times, and while it doesn't crash 🙏 there is still noticable monitor drift (~1s) after some unknown time.

What was interesting is it was only noticible on a single random `Jack Input Client` source (I have 4), while others were seemingly fine. Since it would crash after 3 hours, it's not exactly the easiest thing to monitor, but restarting OBS would immediately resolve the latency. 

Reviewing a recording and multiple Twitch VODs, however, indicates the latency *may not* propagated to final output (ie file nor VOD) for persisted audio tracks (which is just my mic, sans the other 3 sources). Personally, I can live with this over crashing, but it would be nice one day to see it resolved 🤞.

Thank you for your help @kkartaltepe!

VictorKruger:
I've tested this patch out for 3 hours today and didn't have a single crash.

PR #8312, opened 2023-02-20
Linux texture encoding

kkartaltepe:
### Description

This enables the texture encoding path for opengl, my target is linux but I think we probably can/should support these on  mac/opengl so I did not restrict these to just windows+linux.

It is largely a rebase of some of commits in  #4974 that were specific just to enabling the texture_encode path on opengl.

It also adds a profiler root for the gpu encoder thread which is helpful when bringing up a new encoder.

### Motivation and Context

I was looking into texture encoding on QSV and the existing work in #4974 worked well in my implementation so it seems worth pulling into its own PR.

Considering I am working with vaapi/mesa and  #4974 is nvenc, I think this should be a workable solution for all platforms. But it would be good to have more discussion on whether this is the best way to pass textures. 

At least one consideration was whether we should pass DMABUFs directly, but in both my implementation and NVENC we need to use the opengl context to blit data into the encoder surfaces so passing a DMABUF just results in extra work to re-import it into opengl.

### How Has This Been Tested?

Tested in the implementation of QSV texture encoding in https://github.com/kkartaltepe/obs-studio/tree/qsv-linux-pt3 which "Works on my machine" at least.

### Types of changes

### Checklist:

### DIFF
--- a/libobs/obs-encoder.h
+++ b/libobs/obs-encoder.h
@@ -29,6 +29,9 @@
 extern "C" {
 #endif
 
+struct obs_encoder;
+typedef struct obs_encoder obs_encoder_t;
+
 #define OBS_ENCODER_CAP_DEPRECATED (1 << 0)
 #define OBS_ENCODER_CAP_PASS_TEXTURE (1 << 1)
 #define OBS_ENCODER_CAP_DYN_BITRATE (1 << 2)
--- a/libobs/obs-encoder.h
+++ b/libobs/obs-encoder.h
@@ -120,6 +120,16 @@ struct obs_encoder_roi {
 	float priority;
 };
 
+struct gs_texture;
+
+/** Encoder input texture */
+struct encoder_texture {
+	/** Shared texture handle, only set on Windows */
+	uint32_t handle;
+	/** Textures, length determined by format */
+	struct gs_texture *tex[4];
+};
+
 /**
  * Encoder interface
  *
@@ -280,6 +290,12 @@ struct obs_encoder_info {
 			       uint64_t lock_key, uint64_t *next_key,
 			       struct encoder_packet *packet,
 			       bool *received_packet);
+
+	bool (*encode_texture2)(void *data, struct encoder_texture *texture,
+				int64_t pts, uint64_t lock_key,
+				uint64_t *next_key,
+				struct encoder_packet *packet,
+				bool *received_packet);
 };
 
 EXPORT void obs_register_encoder_s(const struct obs_encoder_info *info,
--- a/libobs/obs-video-gpu-encode.c
+++ b/libobs/obs-video-gpu-encode.c
@@ -17,8 +17,11 @@
 
 #include "obs-internal.h"
 
-static void *gpu_encode_thread(struct obs_core_video_mix *video)
+#define NBSP "\xC2\xA0"
+static const char *gpu_encode_frame_name = "gpu_encode_frame";
+static void *gpu_encode_thread(void *data)
 {
+	struct obs_core_video_mix *video = data;
 	uint64_t interval = video_output_get_frame_time(video->video);
 	DARRAY(obs_encoder_t *) encoders;
 	int wait_frames = NUM_ENCODE_TEXTURE_FRAMES_TO_WAIT;
@@ -26,6 +29,10 @@ static void *gpu_encode_thread(struct obs_core_video_mix *video)
 	da_init(encoders);
 
 	os_set_thread_name("obs gpu encode thread");
+	const char *gpu_encode_thread_name = profile_store_name(
+		obs_get_profiler_name_store(),
+		"obs_gpu_encode_thread(%g" NBSP "ms)", interval / 1000000.);
+	profile_register_root(gpu_encode_thread_name, interval);
 
 	while (os_sem_wait(video->gpu_encode_semaphore) == 0) {
 		struct obs_tex_frame tf;
@@ -41,6 +48,8 @@ static void *gpu_encode_thread(struct obs_core_video_mix *video)
 			wait_frames--;
 			continue;
 		}
+
+		profile_start(gpu_encode_thread_name);
 
 		os_event_reset(video->gpu_encode_inactive);
 
@@ -125,6 +134,7 @@ static void *gpu_encode_thread(struct obs_core_video_mix *video)
 			else
 				next_key++;
 
+			profile_start(gpu_encode_frame_name);
 			if (encoder->info.encode_texture2) {
 				struct encoder_texture tex = {0};
 
@@ -142,6 +152,8 @@ static void *gpu_encode_thread(struct obs_core_video_mix *video)
 					encoder->cur_pts, lock_key, &next_key,
 					&pkt, &received);
 			}
+			profile_end(gpu_encode_frame_name);
+
 			send_off_encoder_packet(encoder, success, received,
 						&pkt);
 
@@ -178,6 +190,9 @@ static void *gpu_encode_thread(struct obs_core_video_mix *video)
 			obs_encoder_release(encoders.array[i]);
 
 		da_resize(encoders, 0);
+
+		profile_end(gpu_encode_thread_name);
+		profile_reenable_thread();
 	}
 
 	da_free(encoders);
@@ -186,7 +201,6 @@ static void *gpu_encode_thread(struct obs_core_video_mix *video)
 
 bool init_gpu_encoding(struct obs_core_video_mix *video)
 {
-#ifdef _WIN32
 	const struct video_output_info *info =
 		video_output_get_info(video->video);
 
@@ -210,7 +224,11 @@ bool init_gpu_encoding(struct obs_core_video_mix *video)
 			return false;
 		}
 
+#ifdef _WIN32
 		uint32_t handle = gs_texture_get_shared_handle(tex);
+#else
+		uint32_t handle = (uint32_t)-1;
+#endif
 
 		struct obs_tex_frame frame = {.tex = tex,
 					      .tex_uv = tex_uv,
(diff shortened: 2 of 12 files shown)


kkartaltepe:
Note this is draft because I think there probably needs more discussion on if this is the best way to solve the texture encoding on opengl problem, but otherwise the code is done.

norihiro:
I'd like to suggest revising the commit message of 3ec5a3a82 like below.
```plain
libobs,libobs-opengl: Enable GPU encoding for OpenGL

Enable all of the previously Windows only paths for OpenGL backends that
support encode_texture2
```

kkartaltepe:
> I'd like to suggest revising the commit message of [3ec5a3a](https://github.com/obsproject/obs-studio/commit/3ec5a3a826c4a5548abf1a096d7a9bcf575b5f28) like below.
> 
> ```
> libobs,libobs-opengl: Enable GPU encoding for OpenGL
> 
> Enable all of the previously Windows only paths for OpenGL backends that
> support encode_texture2
> ```

Thanks, done and taking it out of WIP based on some discussion in discord today. As it seems most implementations seem fine with this structure and we can work out any more changes on the PR.

kkartaltepe:
Rebased over the frameskip PR, we just happened to touch the same block to add variables. So this is good to go again.

kkartaltepe:
Rebased after ROI conflicts and updated comments.

kkartaltepe:
The qsv impl using this is also rebased and tested now.

libobs/obs-video-gpu-encode.c
```diff
@@ -125,10 +134,36 @@ static void *gpu_encode_thread(struct obs_core_video_mix *video)
...
+			if (encoder->info.encode_texture2) {
+				union {
+					struct encoder_texture tex;
+					/* MSVC complains about
+					   offsetof(..., tex[3]) */
+					char dummy[offsetof(struct encoder_texture,
+							    tex) +
+						   sizeof(struct gs_texture *) *
+							   3];
+				} u = {0};
```
derrod, line 147: What's this about?

kkartaltepe: The most difficult code to read is always your past self... I believe msvc blew up when trying to zero initialize this struct, so this added a byte-array of sufficient size. Lets remove it and see the error in CI.

derrod: It seems like even if MSVC hates it that it would be saner to just `memset` the struct rather than... this.

kkartaltepe: Actually i think this was left over from when we used a VLA instead of a sized array for gs_texture pointers. At least its working with msvc so monstrosity removed. Thanks.

libobs/obs-encoder.h
```diff
@@ -280,6 +295,27 @@ struct obs_encoder_info {
...
+	 * If this function is not defined, it is assumed that only textures in
+	 * NV12 format are supported.
+	 *
+	 * @param          data  Data associated with this encoder context
+	 * @param[in]      info  Video format information
+	 * @return               Whether the encoder supports texture encoding
+	 *                       with this video format
+	 */
+	bool (*encode_texture_available)(void *data,
+					 const struct video_scale_info *info);
```
derrod, line 312: This is a leftover from the ARGB PR and can be removed for now as it's unused. We shall revisit ARGB texture support later but right now this is unused.

libobs/obs-encoder.h
```diff
@@ -120,6 +123,18 @@ struct obs_encoder_roi {
 	float priority;
 };
 
+struct gs_texture;
+
+/** Encoder input texture */
+struct encoder_texture {
+	/** Texture format and size */
+	struct video_scale_info info;
```
derrod, line 131: This appears to be unused in all existing implementations and can probably be removed. Even if we add ARGB support we probably don't need this as all textures passed on an encoder must have the same format which should be negotiated on encoder init.

libobs/obs-video-gpu-encode.c
```diff
@@ -125,10 +134,27 @@ static void *gpu_encode_thread(struct obs_core_video_mix *video)
...
 
-			success = encoder->info.encode_texture(
-				encoder->context.data, tf.handle,
-				encoder->cur_pts, lock_key, &next_key, &pkt,
-				&received);
+			profile_start(gpu_encode_frame_name);
+			if (encoder->info.encode_texture2) {
+				struct encoder_texture tex = {0};
+
+				obs_encoder_get_video_info(encoder, &tex.info);
```
derrod, line 141: As noted above we don't really need the video info in the texture metadata so this can be removed.

libobs/obs-encoder.h
```diff
@@ -120,6 +123,18 @@ struct obs_encoder_roi {
...
+	/** Shared texture handle, only set on Windows */
+	uint32_t handle;
+	/** Textures, NULL-terminated */
+	struct gs_texture *tex[5];
```
derrod, line 135: Also not really sure about this, Why does it need to be NULL terminates? The currently in use texture format should determine the number of textures in use. And at most this would then be 4 (Y + U + V + Alpha).

derrod (changes requested):
A couple leftovers from the old ARGB and texture encoding PRs that I think can be removed until we actually want to add support for more formats.

libobs/obs-encoder.h
kkartaltepe: The 5th element was an earlier discussion to just be safe and include a terminal, i dont feel strongly.

As this is OBS's api we can decide to always use multiple textures, one per plane. In theory (in reality on vulkan) some platforms only do planar as multiple views on a single alloc, but we can cross that bridge when we get to it.

kkartaltepe:
> A couple leftovers from the old ARGB and texture encoding PRs that I think can be removed until we actually want to add support for more formats.

Good catches, updated.

derrod (approved):
Should be fine as it is now, we have some time befire it'll make it into a stable release so revision should be no problem.

libobs/obs-encoder.c
```diff
@@ -694,6 +694,25 @@ void obs_encoder_shutdown(obs_encoder_t *encoder)
...
+	info->format = voi->format;
+	info->colorspace = voi->colorspace;
+	info->range = voi->range;
+	info->width = obs_encoder_get_width(encoder);
+	info->height = obs_encoder_get_height(encoder);
+
+	if (encoder->info.get_video_info)
+		encoder->info.get_video_info(encoder->context.data, info);
+
+	if (info->width != voi->width || info->height != voi->height)
```
tt2468, line 712: What's happening here? I don't think it's a good idea to be changing any data with a `get` method.

libobs/obs-encoder.c
```diff
@@ -694,6 +694,25 @@ void obs_encoder_shutdown(obs_encoder_t *encoder)
 	pthread_mutex_unlock(&encoder->init_mutex);
 }
 
+void obs_encoder_get_video_info(struct obs_encoder *encoder,
```
tt2468, line 697: Is this necessary, or just a helpful util?

libobs/obs-encoder.h
```diff
@@ -280,6 +293,12 @@ struct obs_encoder_info {
 			       uint64_t lock_key, uint64_t *next_key,
 			       struct encoder_packet *packet,
 			       bool *received_packet);
+
+	bool (*encode_texture2)(void *data, struct encoder_texture *texture,
+				int64_t pts, uint64_t lock_key,
```
tt2468, line 298: If we're introducing an encoder_texture struct, it seems like `pts`, `lock_key`, and `next_key` should be values of that struct.

tt2468 (changes requested):
I've tested this with Rodney's PR on Ubuntu 22.04 with an RTX A4000 GPU, running X11. It works well, and I tried a number of different codepaths like color space and other formats, and everything appears to work as desired. Great to see PCIe bus bandwidth staying in the 10's of MB/s while streaming...

libobs/obs-encoder.c
derrod: This is just a copy of the existing `get_video_info`, also a leftover from the original ARGB stuff and I think it could be removed as well.

https://github.com/obsproject/obs-studio/blob/fbf6c8d243714b5d5409826f69c633c627d3535c/libobs/obs-encoder.c#L168

kkartaltepe: It seemed useful and I assumed it was used in the ARGB stuff, but I also cannot find it used in the PR (and the old get_video_info is still leftover). Given its not actually used Ill remove it and revert to the internal.

That said do we still need that set on this path?

derrod: The video info feature is to allow the encoder to specify the input format it wants, in theory I suppose this could include a different size. I'd leave the existing internal function alone for now.

libobs/obs-encoder.h
derrod: pts sure, the next_key is technically an output variable so making that part of the struct feeels kind of ugly.

kkartaltepe: im not sure i agree with any of them, its just that we dont share/reuse textures that all of these happen to have the same lifetimes.

derrod: I see. What I have been wondering if perhaps we can move at least the pts to a struct similar to `encoder_frame`, the motivation here is that for the multi-encoder streaming Twitch is doing and #10096 we'll need a way to force IDR frames. 
For non-texture encoders we can simply extend the existing struct without breaking ABI, but for GPU encoders that isn't possible. So perhaps we can make this API more future-looking by adding a means to add more information later without requiring a new ABI?

I realise this is getting into bikeshedding territory, I won't block this PR on that, just wanted to hear your opinion.

tt2468 (approved):

RytoEX (approved):
Seems fine. Want to land this to unblock other ongoing work to give that work some lead time. Reviews here plus off-thread discussion indicate that this is fine to land.

PR #8321, opened 2023-02-22
UI: Fix disabled text color in dark theme

shiina424:
### Description

Fixed an issue where some text colors in settings were black at the start of output in dark themes.
If this is incorrect, we would appreciate it if you could correct it.

Steps to Reproduce
1. Start output
2. open settings
3. change theme to dark
4. View output or Advanced settings

| Before |  After |
| ------------- | ------------- |
| before | after |

Before (Output Mode Advanced)

https://user-images.githubusercontent.com/24505855/220629103-7a9d87be-b991-46d6-9385-52cd41d3025e.mp4

After (Output Mode Advanced)

https://user-images.githubusercontent.com/24505855/220629280-95909f98-05df-44c2-b061-745d77cb7a43.mp4

### Motivation and Context

I want displayed correctly.

### How Has This Been Tested?

I changed the qss file and tested it.

### Types of changes

### Checklist:

### DIFF
--- a/UI/data/themes/Dark.qss
+++ b/UI/data/themes/Dark.qss
@@ -89,7 +89,7 @@ QWidget {
 /* Misc */
 
 QWidget::disabled {
-    color: 2px solid palette(bright-text);
+    color: rgb(165,164,165); /* disabledText */
 }
 
 QAbstractItemView, QStackedWidget#stackedMixerArea QWidget {


Warchamp7 (approved):
That malformed line of qss has technically been there [since Dark theme was added](https://github.com/obsproject/obs-studio/blame/58caf406cd26613103ec0f4cfea52067a647d71e/UI/data/themes/Dark.qss#L83-L89)

https://github.com/obsproject/obs-studio/blob/a62f079541bfbdeaa2c3e233caad212a7bfcc50f/obs/data/themes/Dark.qss#L41-L46

Looks like something in v28 caused it to actually start causing issues. I suspect something to do with theme meta.

Initially I thought the fix here would be to remove this rule altogether, but since we explicitly set `color` for QWidget, this leads to disabled using the non-disabled text, rather than the the OBSTheme defined colors.

Removing the QWidget rule for `color` leads to the expected behaviour of the OBSTheme text color definitions being used everywhere, except for titles of all QGroupBoxes (Not just disabled).

I do not know why. I do not care why. QSS is an enigma.

GeorgesStavracas (approved):

PR #8323, opened 2023-02-22
UI: Set QT_NO_SUBTRACTOPAQUESIBLINGS env var 

derrod:
### Description

This feels like a dirty hack, and it probably is, but it doesn't seem to break anything (as far as I tested).

The function in Qt this effectively disables is annotated with `this is too expensive !!!` - they're right about that one!

### Motivation and Context

Make hotkeys faster for large scene collections.

### How Has This Been Tested?

Checked with profiler. This function used to be the single slowest thing and now it's gone.

For my favourite stress test (`Massive.json`) this speeds up loading the hotkeys list by eleminating most of the time the UI would spend unresponsive *after* it has finished creating all the widgets (creating the widgets itself is still slow tho).

### Types of changes

### Checklist:

### DIFF
--- a/UI/obs-app.cpp
+++ b/UI/obs-app.cpp
@@ -2329,6 +2329,13 @@ static int run_program(fstream &logFile, int argc, char *argv[])
 #endif
 #endif
 
+	/* NOTE: This disables an optimisation in Qt that attempts to determine if
+	 * any "siblings" intersect with a widget when determining the approximate
+	 * visible/unobscured area. However, by Qt's own admission this is slow
+	 * and in the case of OBS it significantly slows down lists with many
+	 * elements (e.g. Hotkeys) and it is actually faster to disable it. */
+	qputenv("QT_NO_SUBTRACTOPAQUESIBLINGS", "1");
+
 	OBSApp program(argc, argv, profilerNameStore.get());
 	try {
 		QAccessible::installFactory(accessibleFactory);


derrod:
The Qt function this disables is here: https://github.com/qt/qtbase/blob/0627ab17277ad5fa83fe6fcab5af5cd2748162a3/src/widgets/kernel/qwidget.cpp#L2024

As somebody pointed out to me, this environment variable is very much undocumented, and the only other case where I could find somebody using it is RenderDoc:
https://github.com/baldurk/renderdoc/blob/52e9404961277d1bb1ed03a376b722c2b91e3762/qrenderdoc/Code/qrenderdoc.cpp#L161-L164

The comment there claims that this function is an optimisation in Qt, and not strictly necessary. Looking at what the function does that appears to be correct; It attempts to determine if any child widgets of its parents intersect with the widget's region to avoid unnecessary draws. Sadly the commit history for pre-Qt5 stuff does not appear to be public and this function hasn't been touched in over a decade, so the reasoning is lost to time.

Given that widgets generally shouldn't overlap each other this does seem like something we probably don't have to worry about in OBS and can safely disable for the rather large time save in hotkeys.

Conditionally enabling it by setting the environment variable just before hotkeys are loaded does not appear to work unfortunately (at least not the ways I tried [via `qputenv`]).

PR #8332, opened 2023-02-22
mac-virtualcam: Fix incorrect PTS on Apple Silicon as well as PTS being rounded

mac-cain13:
👋 First time contributing to the OBS project, please let me know if anything is unclear or needs improvement. Happy to improve this PR where necessary to get this merged.

#### 1. Stable presentation time stamp on Apple Silicon
The `fillFrame` method of the mac-virtualcam plugin is creating samples directly using the value from `mach_absolute_time()` as `hostTime`. This assumes this value is in nanoseconds, while it is in fact in mach tick units. On Intel macs mach tick units will be exactly 1 nanosecond resulting in valid values, but on Apple Silicon macs this is no longer the case.

This results in sample buffers with the placeholder image that have much lower presentation time stamps than the samples containing content produced by OBS. In previews/live streams this shows due to the last content sample being shown frozen until the whole buffer is filled with samples containing the placeholder image. Applications recording the video stream are even more confused and crash or record videos with wildly incorrect lengths.

In this PR ~`mach_timebase_info`~ `clock_gettime_nsec_np` is used to convert from mach tick units to nanoseconds. This will make sure the `hostTime` value is correct on both Apple Silicon and Intel macs. Making sure we produce stable presentation time stamps from the virtual webcam plugin at all times.

#### 2. Prevent rounding of presentation time stamp to whole seconds
All presentation time stamps are rounded to whole seconds during the convertion from nanoseconds to seconds, because of the immediate cast to `int64_t`. This results in the same presentation time stamp being send to consumers for a whole second.

In previews/live streams this isn’t super visible as last frame is often assumed to be the newest and the stream is updated. It’s more problematic when recording since APIs like Apples AVFoundation don’t allow duplicate presentation time stamps or it can look like frames are produces in huge bursts once per second.

In this PR `CMTimeMakeWithSeconds` is used instead of `CMTimeMake` to make sure the conversion is done correctly and simplify the calculation we have to do a little.

### Motivation and Context
- Solves incorrect length in recordings when the recording is started with a stopped virtual camera
- Some recorders will also crash if they expect presentation time stamps to increase/be in reasonable range of each other
- Prevents seconds long delay before placeholder shows up when stopping the virtual webcam/closing OBS

Further reading:
- [`mach_absolute_time` documentation](https://developer.apple.com/documentation/driverkit/3438076-mach_absolute_time)
- [`mach_timebase_info` documentation](https://developer.apple.com/documentation/driverkit/3433733-mach_timebase_info)
- [Blogpost about the clock change between Intel & Apple Silicon macs](https://eclecticlight.co/2020/09/08/changing-the-clock-in-apple-silicon-macs/)

### How Has This Been Tested?
Tested on both an Apple Silicon Mac with macOS 13 and an Intel Mac with macOS 11.
Recorded and observed the samples outputted from the virtual camera to a 3rd party app in different scenarios;
- OBS app not started
- OBS started + virtual webcam live
- OBS started + virtual webcam stopped

Also transitioned between these states while observing the stream. All looks good with this change.

### Types of changes
- Bug fix

### Checklist:

### DIFF
--- a/plugins/mac-virtualcam/src/dal-plugin/CMSampleBufferUtils.mm
+++ b/plugins/mac-virtualcam/src/dal-plugin/CMSampleBufferUtils.mm
@@ -13,7 +13,7 @@ CMSampleTimingInfo CMSampleTimingInfoForTimestamp(uint64_t timestampNanos,
 {
 	// The timing here is quite important. For frames to be delivered correctly and successfully be recorded by apps
 	// like QuickTime Player, we need to be accurate in both our timestamps _and_ have a sensible scale. Using large
-	// timestamps and scales like mach_absolute_time() and NSEC_PER_SEC will work for display, but will error out
+	// timestamps and scales like clock_gettime_nsec_np() and NSEC_PER_SEC will work for display, but will error out
 	// when trying to record.
 	//
 	// 600 is a common default in Apple's docs https://developer.apple.com/documentation/avfoundation/avmutablemovie/1390622-timescale
--- a/plugins/mac-virtualcam/src/dal-plugin/CMSampleBufferUtils.mm
+++ b/plugins/mac-virtualcam/src/dal-plugin/CMSampleBufferUtils.mm
@@ -21,9 +21,8 @@ CMSampleTimingInfo CMSampleTimingInfoForTimestamp(uint64_t timestampNanos,
 	CMSampleTimingInfo timing;
 	timing.duration =
 		CMTimeMake(fpsDenominator * scale, fpsNumerator * scale);
-	timing.presentationTimeStamp = CMTimeMake(
-		((int64_t)(timestampNanos / (double)NSEC_PER_SEC) * scale),
-		scale);
+	timing.presentationTimeStamp = CMTimeMakeWithSeconds(
+		timestampNanos / (double)NSEC_PER_SEC, scale);
 	timing.decodeTimeStamp = kCMTimeInvalid;
 	return timing;
 }
--- a/plugins/mac-virtualcam/src/dal-plugin/OBSDALStream.mm
+++ b/plugins/mac-virtualcam/src/dal-plugin/OBSDALStream.mm
@@ -316,7 +316,7 @@
 	CVPixelBufferRef pixelBuffer =
 		[self createPixelBufferWithTestAnimation];
 
-	uint64_t hostTime = mach_absolute_time();
+	uint64_t hostTime = clock_gettime_nsec_np(CLOCK_UPTIME_RAW);
 	CMSampleTimingInfo timingInfo =
 		CMSampleTimingInfoForTimestamp(hostTime, (uint32_t)self.fps, 1);
 
@@ -365,9 +365,9 @@
 	CMSampleTimingInfo timingInfo = CMSampleTimingInfoForTimestamp(
 		timestamp, fpsNumerator, fpsDenominator);
 
-	err = CMIOStreamClockPostTimingEvent(timingInfo.presentationTimeStamp,
-					     mach_absolute_time(), true,
-					     self.clock);
+	err = CMIOStreamClockPostTimingEvent(
+		timingInfo.presentationTimeStamp,
+		clock_gettime_nsec_np(CLOCK_UPTIME_RAW), true, self.clock);
 	if (err != noErr) {
 		DLog(@"CMIOStreamClockPostTimingEvent err %d", err);
 	}


mac-cain13:
Seems I got caught off guard by de release version of OBS replacing my local version of the DAL plugin. So I half of my tests were against the my changes, the other half against the current release. 😱 

Now that I took a better look it seems there is an issue with the placeholder image being shown at all times and the presentation time increasing only once per second. Unsure if this is introduced by this PR or it's an issue on the main branch, I will test further apply fixes where necessary and report back. 🧑‍💻

RytoEX:
> * Some recorders will also crash if they expect presentation time stamps to increase/be in reasonable range of each other

Could you give specific examples of such recorders so that we could observe this behavior?

PatTheMav:
`mach_absolute_time` is deprecated - as the documentation states the correct replacement would be to simply use `clock_gettime_nsec_np(CLOCK_UPTIME_RAW)`.

mac-cain13:
> > * Some recorders will also crash if they expect presentation time stamps to increase/be in reasonable range of each other
> 
> Could you give specific examples of such recorders so that we could observe this behavior?

Don't have production apps that show this behaviour, easiest way to reproduce it is create a small Mac app using `AVAssetWriterInput.append()` to write the data you get from OBS to a MP4 file. That will crash as `AVAssetWriterInput` does expect the presentation time to always move forward (when there is no decoded time stamp). So duplicate/backward times will crash.

> `mach_absolute_time` is deprecated - as the documentation states the correct replacement would be to simply use `clock_gettime_nsec_np(CLOCK_UPTIME_RAW)`.

Indeed nicer to use `clock_gettime_nsec_np` instead of `mach_absolute_time` in combination with `mach_timebase_info`. Will adjust the PR!

mac-cain13:
### Updated the PR:
- Switched from using `mach_timebase_info` to using the simpler `clock_gettime_nsec_np` to get correct nsec on Apple Silicon.
- Restested correctly this time; discovered that presentation time stamps are rounded to whole seconds.
- Added fix by using `CMTimeMakeWithSeconds` instead of `CMTimeMake` to reduce the calculation we need to do and prevent errors.

Also updated the PR description above to reflect this.

### Logs that show the change in behaviour
<details>
  <summary>PTS log from current master:</summary>

```log
PTS: 6836.0
mac-virtualcam(DAL): Server is not available
PTS: 6836.0
PTS: 6836.0
[...removed some repetitive lines for readability...]
PTS: 6836.0
PTS: 6836.0
mac-virtualcam(DAL): Attempting connection
mac-virtualcam(DAL): -[OBSDALMachClient connectToServer]
mac-virtualcam(DAL): Initialized mach port 76551 for receiving
PTS: 6836.0 // <-- Last "placeholder" frame from fillFrame
mac-virtualcam(DAL): -[OBSDALStream stopServingDefaultFrames]
PTS: 284849.0 // <-- First "real" frame from OBS Studio
PTS: 284849.0
PTS: 284850.0 // <-- Only increases every second
PTS: 284850.0
[...removed some repetitive lines for readability...]
PTS: 284851.0 // <-- Only increases every second
PTS: 284851.0
```
</details>

<details>
  <summary>PTS log from this PR:</summary>

```log
PTS: 284428.8233333333
mac-virtualcam(DAL): Server is not available
PTS: 284428.85333333333
PTS: 284428.88666666666
[...removed some repetitive lines for readability...]
PTS: 284429.7866666667
PTS: 284429.82
mac-virtualcam(DAL): Attempting connection
mac-virtualcam(DAL): -[OBSDALMachClient connectToServer]
mac-virtualcam(DAL): Initialized mach port 76551 for receiving
PTS: 284429.8566666667 // <-- Last "placeholder" frame from fillFrame
mac-virtualcam(DAL): -[OBSDALStream stopServingDefaultFrames]
PTS: 284429.83166666667 // <-- First "real" frame from OBS Studio
PTS: 284429.865
PTS: 284429.8983333333
PTS: 284429.93166666664
```
</details>

### Known issues

The last "placeholder" frame can (and will very often) be time stampt with a later PTS than the first "real" frame from OBS. This is not optimal, however the new behaviour already is a major improvement and quite a small & simple change. Preventing that last frame from being emitted might complicate this PR too much to quickly/easily merge it. We can always add that improvement later on.

---

@PatTheMav hope we can merge this in soon! Let me know if you need anything from me to help out with the review.

RytoEX (changes requested):
Commit message nits:
* macs -> Macs
* virtual webcam -> virtual camera
* convertion -> conversion
* isn’t -> isn't
* It’s -> It's
* Apples -> Apple's
* produces -> produced

mac-cain13:
@RytoEX fixed the commit message nits.

RytoEX (approved):
Commit messages seem good enough. Will leave code review to @PatTheMav and/or @gxalpha . I know we've been bitten by unclear documentation for the CMTime functions before, and they're better suited to review this code than I am.

mac-cain13:
@PatTheMav no rush, but please let me know if I can do anything to make a review from you easier.
Would be nice if we can merge this in and make recording the OBS virtual camera less fragile.

PR #8337, opened 2023-02-23
UI: Make T-Bar unclickable

gxalpha:
### Description

Changes the T-Bar so that it can only be dragged and not clicked at all.
Follow-up to #7347.

### Motivation and Context

Requested by @Warchamp7 in off-thread discussions about #7347.
*Edit: Maybe. That was what I remembered at least. On the PR itself he wasn't completely sure. So we'll see, maybe this PR will get closed in the end.*

### How Has This Been Tested?

macOS 13.2
Tested by clicking on the T-Bar and seeing that nothing happens, and by dragging it and seeing that it still functions when dragging.

### Types of changes

### Checklist:

### DIFF
--- a/UI/slider-ignorewheel.cpp
+++ b/UI/slider-ignorewheel.cpp
@@ -94,3 +94,42 @@ QAccessible::Role VolumeAccessibleInterface::role() const
 {
 	return QAccessible::Role::Slider;
 }
+
+void SliderIgnoreClick::mousePressEvent(QMouseEvent *event)
+{
+	QStyleOptionSlider styleOption;
+	initStyleOption(&styleOption);
+	QRect handle = style()->subControlRect(QStyle::CC_Slider, &styleOption,
+					       QStyle::SC_SliderHandle, this);
+#if QT_VERSION >= QT_VERSION_CHECK(6, 0, 0)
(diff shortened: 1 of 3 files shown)


Warchamp7:
Curious and open to hearing people's thoughts on this.

On a hardware mixer it's impossible to instantaneously jump the bar to a specific partial position as dictated by the laws of physics. There may be control surfaces with buttons or macros for such behaviour but I am not aware of them.

With a digital interface, OBS is able to do the "impossible" in this kind of scenario.

Generally speaking, transitions are created with their actual progression from start to finish being in mind. The ability to skip to 25% or 50% into a transition could open up creative uses with things like Luma Wipes.

Prior to #7347, doing so was a cumbersome process. That fix opens the door to that unique usage to be viable.

Whether that's useful or desired is the underlying question here.

Fenrirthviti:
I think this is a far better end state for the UI, and my understanding is that the API will allow plugins/websockets/etc. to modify and set the % manually, which IMO covers the niche use case for when you'd actually want to do that. You'd know why, and you'd be able to find out how pretty easily.

Warchamp7 (approved):

RytoEX (approved):
Seems okay to me.

PR #8339, opened 2023-02-24
UI: Add native dark titlebar capability on Windows

Arastais:
### Description

This PR is a simple change that allows the titlebar/caption of OBS on Windows 10 and up to be natively dark. It was done using qt's built-in system and a `qt.conf` file that just specifies the appropriate settings/toggle. Doing it using qt means that the titlebar of obs is now based on the user's **system** color theme:

obs_dark

It's important to note that because this feature using qt's built-in functionality, **the titlebar color is independent of the theme set in OBS**, and only relies on the color set in the user's system settings (or the "default app mode" if a custom color theme is used). This means that, for example, if a user has "light" as their system theme in Window's system setting, but is using a light obs theme, it will be "mismatched" once again:

obs_light_mismatch_vice_versa

And vice versa:
obs_light_mismatch

This may seem like a bug, but I would actually argue that basing the titlebar color on the system setting (rather than the obs theme) should be intentional; This is how most Win32 applications do it, and it would make more sense IMO if obs stayed consistent with the system theme. Doing it based on obs' theme instead would be significantly more complicated, requiring manually changing it by leveraging the Desktop Window Manager API within the Windows SDK (a la [`DwmSetWindowAttribute`](https://learn.microsoft.com/en-us/windows/win32/api/dwmapi/nf-dwmapi-dwmsetwindowattribute) and [`DWMWINDOWATTRIBUTE::DWMWA_USE_IMMERSIVE_DARK_MODE`](https://learn.microsoft.com/en-us/windows/win32/api/dwmapi/ne-dwmapi-dwmwindowattribute) in `dwmapi.h`).

### Motivation and Context

Before this change, on Windows 10, the titlebar is always light regardless of the obs theme or system setting; obs "ignores" (since it has no way to tell the system color) both the theme and system setting and just always has a light titlebar:
 
image

### How Has This Been Tested?

All screenshots and test cases were done on Windows 10 21H2 (build 19044.2604) on the same build of OBS (except for the movtivation and context screenshot, which was done on release 29.0.1)

### Types of changes

### Checklist:

### DIFF
--- a/UI/forms/obs.qrc
+++ b/UI/forms/obs.qrc
@@ -80,4 +80,7 @@
     <file>fonts/OpenSans-Bold.ttf</file>
     <file>fonts/OpenSans-Italic.ttf</file>
   </qresource>
+  <qresource prefix="/qt/etc/">
+	<file>qt.conf</file>
+  </qresource>
 </RCC>
--- a/UI/forms/qt.conf
+++ b/UI/forms/qt.conf
@@ -1,0 +1,2 @@
+[Platforms]
+WindowsArguments = darkmode=1


WizardCM:
Does this follow the system toggle in realtime, or does it require an OBS restart?

Personally I agree it makes more sense to follow the system option vs going based on the OBS theme, though I expect there are users who would prefer the latter.

RytoEX:
I'm fairly certain that without this change we get this functionality with Qt 6.4+, which we will update to in the next OBS version anyway. You can see this behavior in the OBS Studio 29 beta releases where we briefly used Qt 6.4.1:
OS Dark + OBS Light:
image

OS Light + OBS Light:
image

No OBS restarts required.

What would this PR get us that Qt 6.4+ by itself does not?

Arastais:
> Does this follow the system toggle in realtime, or does it require an OBS restart?

Realtime; if you change the theme color in windows settings, any open obs windows will immediately update without needing a restart.

<br/>

> What would this PR get us that Qt 6.4+ by itself does not?

I did not test on the beta versions, only the latest `master`, so I would not have noticed that it is supported without the need of a toggle. Regardless, I'm surprised that it works within 6.4.1, because iirc it was 6.4.2 that allowed qt to switch the titlebar in realtime; 6.4 required a restart.

Anyway, I think at a bare minimum there is a good argument for this pr: It is small enough that it is worth adding so that this feature gets implemented before the upgrade to 6.4.x. That way this feature doesn't have to depend on using a specific version of qt.

derrod:
> Anyway, I think at a bare minimum there is a good argument for this pr: It is small enough that it is worth adding so that this feature gets implemented before the upgrade to 6.4.x. That way this feature doesn't have to depend on using a specific version of qt.

Well the next release is targeted to use 6.4.3, and even if there would be another patch release on 6.3.x we wouldn't merge this for that. We can leave this open should the 6.4 upgrade not work out for whatever reason, but it's unlikely to be merged at this point.

RytoEX:
> I did not test on the beta versions, only the latest `master`, so I would not have noticed that it is supported without the need of a toggle. Regardless, I'm surprised that it works within 6.4.1, because iirc it was 6.4.2 that allowed qt to switch the titlebar in realtime; 6.4 required a restart.

The OBS Studio 29 beta releases used Qt 6.4.1. This can be confirmed by checking the DLL versions. Windows Dark Mode support is an ongoing work in progress with Qt.

> Anyway, I think at a bare minimum there is a good argument for this pr: It is small enough that it is worth adding so that this feature gets implemented before the upgrade to 6.4.x. That way this feature doesn't have to depend on using a specific version of qt.

The next major/minor release of OBS will use Qt 6.4 on Windows, so it seems like a moot point to me, unless we can confirm that something is amiss in Qt 6.4+.

norihiro:
Would you revise the commit title as below? Line breaks are revised and one word `be` is removed.
```plain
UI: Add native dark titlebar capability on Windows

Make the titlebar of obs dark if the system's color setting is "dark".
Titlebar/caption was previously always light on Windows 10+ regardless
of system setting. The titlebar can now be natively dark if the user's
system color setting is set to "dark". It will still be light otherwise.
```

derrod:
We have now updated to Qt 6.4.3 so we don't really need this. We can always revisit in the future should we need a Qt configuration file for some reason.

Issue #7812, opened 2022-11-19
Video Capture window freezes constantly


ImaraRavenfire:
### Operating System Info

Windows 10

### OBS Studio Version

28.1.2

### OBS Studio Log URL

https://obsproject.com/logs/6Mv1l3Ax1gV-JTAu

### Expected Behavior

I expected to be able to add my camera as a source, but the second I did, OBS froze entirely and I had to task manager shut it down.

### Current Behavior

I attempted to add my camera as a source, only for it to decide it wasn't going to allow that, freezing the entire program to the point that it had to be manually shut down in the task manager

### Steps to Reproduce

1. Open OBS
2. Add Source > Video Capture Device
3. Alternatively, open the properties on your video capture device
...

### Anything else we should know?

It worked fine in the previous version of OBS and has only started happening since the latest update as well as the latest windows update

Fenrirthviti:
Driver Version: 31.0.15.2698

This is a known issue with certain nvidia drivers, including the version you are on. Either rolling back, or removing Unity Virtual Camera (the most common triggering cause), or other non-OBS Virtual Cameras might resolve this, however it's a driver bug and outside our control to fix on our side.

I will leave this open for tracking purposes, as we do have an open line of communication with NVIDIA regarding the issue.

AudricSan:
Going back to an older version of nvidia driver indeed solves the problem..

Put an update of Obs at this level would it be planned?Because it can be annoying for some games not to have the latest version of the driver, I find it hard to uninstall the driver each time I want to add a Video capture in Obs to reinstall it afterwards 😅

Fenrirthviti:
In the next release of OBS, we are hoping to have merged a new feature to block the offending third-party tool from loading in OBS https://github.com/obsproject/obs-studio/pull/8343. This should help reduce this. It is not a bug with NVIDIA, or with OBS, but rather a third-party virtual camera device (UnityCapture, which is commonly used by vtuber applications and has already been patched in newer versions of those).

guiglass:
> In the next release of OBS, we are hoping to have merged a new feature to block the offending third-party tool from loading in OBS #8343. This should help reduce this. It is not a bug with NVIDIA, or with OBS, but rather a third-party virtual camera device (UnityCapture, which is commonly used by vtuber applications and has already been patched in newer versions of those).

[...]

PR #8343, opened 2023-02-24
Add DLL blocking functionality for Windows

notr1ch:
### Description
This PR implements a way to block specific DLLs from loading into OBS by way of a `NtMapViewOfSection` hook. This is useful to block known problematic DLLs from injecting that end up causing OBS crashes or other undesirable behavior. DLLs are matched by path suffix, for example `\BadApplication\StupidHook.dll` to only block StupidHook from the BadApplication directory, or `\GenericHook.dll` to match all DLLs with that name.

Currently the list of DLLs is hard coded - this list has been built up over the years since OBS Classic, so while initially it looks quite large, the rate of change is expected to be low. As the blocking hook has to be installed as early as possible to be effective, using methods like loading JSON from an auto updated list is less feasible as the libobs API is not yet available, and the code executes for every `NtMapViewOfSection` call so it should be fast and simple.

There is also an option to match on the PE header timestamp - this allows selectively blocking earlier versions of a DLL that may have since been fixed and similar. An "inverse" mode is also possible where only DLLs matching an exact timestamp are allowed. The timestamp matching is currently used to match older VSeeFace DShow filter DLLs which have known deadlocks in their enumeration routine.

Detours is used for hooking for maximum safety.

### Motivation and Context
Many OBS crashes are caused by buggy or outdated 3rd party modules that have injected into OBS and hooked or otherwise caused unwanted behavior. This PR allows us to finally stop that from happening.

### How Has This Been Tested?
Tested on my own system (Windows 10 22H2), no obvious issues found. This needs more extensive testing on more systems though as a bug in this code will essentially prevent OBS from starting.

### Types of changes

### Checklist:

Fixes https://github.com/obsproject/obs-studio/issues/7812

### DIFF
--- a/UI/win-dll-blocklist.c
+++ b/UI/win-dll-blocklist.c
@@ -1,0 +1,334 @@
+/******************************************************************************
+    Copyright (C) 2023 by Richard Stanway
+
+    This program is free software: you can redistribute it and/or modify
+    it under the terms of the GNU General Public License as published by
+    the Free Software Foundation, either version 2 of the License, or
+    (at your option) any later version.
+
+    This program is distributed in the hope that it will be useful,
+    but WITHOUT ANY WARRANTY; without even the implied warranty of
+    MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+    GNU General Public License for more details.
+
+    You should have received a copy of the GNU General Public License
+    along with this program.  If not, see <http://www.gnu.org/licenses/>.
+******************************************************************************/
+
+#include <Windows.h>
+#include <psapi.h>
+#include <stdint.h>
+#include <stdbool.h>
+#include <inttypes.h>
+#include "detours.h"
+#include "obs.h"
+
+// Undocumented NT structs / function definitions !
+typedef enum _SECTION_INHERIT { ViewShare = 1, ViewUnmap = 2 } SECTION_INHERIT;
+
+typedef enum _SECTION_INFORMATION_CLASS {
+	SectionBasicInformation = 0,
+	SectionImageInformation
+} SECTION_INFORMATION_CLASS;
+
+typedef struct _SECTION_BASIC_INFORMATION {
+	PVOID BaseAddress;
+	ULONG Attributes;
+	LARGE_INTEGER Size;
+} SECTION_BASIC_INFORMATION;
+
+typedef NTSTATUS(STDMETHODCALLTYPE *fn_NtMapViewOfSection)(
+	HANDLE, HANDLE, PVOID, ULONG_PTR, SIZE_T, PLARGE_INTEGER, PSIZE_T,
+	SECTION_INHERIT, ULONG, ULONG);
+
+typedef NTSTATUS(STDMETHODCALLTYPE *fn_NtUnmapViewOfSection)(HANDLE, PVOID);
+
+typedef NTSTATUS(STDMETHODCALLTYPE *fn_NtQuerySection)(
+	HANDLE, SECTION_INFORMATION_CLASS, PVOID, SIZE_T, PSIZE_T);
+
+static fn_NtMapViewOfSection ntMap;
+static fn_NtUnmapViewOfSection ntUnmap;
+static fn_NtQuerySection ntQuery;
+
+#define STATUS_UNSUCCESSFUL ((NTSTATUS)0xC0000001L)
+
+// Method of matching timestamp of DLL in PE header
+typedef enum {
+	TS_IGNORE = 0,      // Ignore timestamp; block all DLLs with this name
+	TS_EQUAL,           // Block only DLL with this exact timestamp
+	TS_LESS_THAN,       // Block all DLLs with an earlier timestamp
+	TS_GREATER_THAN,    // Block all DLLs with a later timestamp
+	TS_ALLOW_ONLY_THIS, // Invert behavior: only allow this specific timestamp
+} ts_compare_t;
+
+typedef struct {
+	// DLL name, lower case
+	const wchar_t *name;
+
+	// Length of name, calculated at startup - leave as zero
+	size_t name_len;
+
+	// PE timestamp
+	const uint32_t timestamp;
+
+	// How to treat the timestamp field
+	const ts_compare_t method;
+
+	// Number of times we've blocked this DLL, for logging purposes
+	uint64_t blocked_count;
+} blocked_module_t;
+
+/*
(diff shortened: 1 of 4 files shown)


ma3uk:
It works perfectly, I don't know which library was blocked, but this PR reduced the QSV encoding time by almost 2 times from 2 ms to 1-1.1 ms, and it seems that random obs closures without notifications have stopped.

Fenrirthviti:
For support purposes, would it be possible to add logging of blocked DLLs?

UI/win-dll-blocklist.c
```diff
@@ -0,0 +1,300 @@
...
+* exists in a certain directory. A name should always include a path component
+* so that e.g. fraps.dll doesn't match notfraps.dll.
+*/
+
+blocked_module_t blocked_modules[] = {
+	// Dell / Alienware Backup & Recovery, crashes during "Browse" dialogs
+	{L"\\dbroverlayiconbackuped.dll", 0, 0, TS_IGNORE},
+
+	// Intel Graphics Performance Analyzer, reported crashes
+	{L"\\gpa.dll", 0, 0, TS_IGNORE},
```
RytoEX, line 88: Is this still occurring with recent versions of Intel GPA? I think some people do use (or have used) it to debug/profile OBS in the past.

UI/win-dll-blocklist.c
```diff
@@ -0,0 +1,300 @@
...
+} blocked_module_t;
+
+/*
+* Note: The name matches at the end of the string based on its length, this allows
+* for matching DLLs that may have generic names but a problematic version only
+* exists in a certain directory. A name should always include a path component
+* so that e.g. fraps.dll doesn't match notfraps.dll.
+*/
+
+blocked_module_t blocked_modules[] = {
```
norihiro, line 83: Can we add `static` qualifier?
```suggestion
static blocked_module_t blocked_modules[] = {
```

UI/win-dll-blocklist.c
```diff
@@ -0,0 +1,300 @@
...
+				if (timestamp == b->timestamp)
+					return false;
+			}
+		}
+	}
+
+	return should_block;
+}
+
+NTSTATUS NtMapViewOfSection_hook(HANDLE SectionHandle, HANDLE ProcessHandle,
```
norihiro, line 202: Can we add `static` qualifier?
```suggestion
static NTSTATUS NtMapViewOfSection_hook(HANDLE SectionHandle, HANDLE ProcessHandle,
```

RytoEX (commented):
Just a question about Intel GPA.

notr1ch:
> For support purposes, would it be possible to add logging of blocked DLLs?

Not really, libobs hasn't been initialized when this code runs, and writing to a log file during a DLL loader lock is not something I want to consider. Additionally, some applications may attempt to constantly re-hook if unsuccessful causing log spam.

UI/win-dll-blocklist.c
notr1ch: Unsure - I couldn't find any information on this, only that it was reported in the past. Perhaps we could block anything older than 2022 or something?

Fenrirthviti:
Understood, thanks for the explanation!

RytoEX:
> > For support purposes, would it be possible to add logging of blocked DLLs?
> 
> Not really, libobs hasn't been initialized when this code runs, and writing to a log file during a DLL loader lock is not something I want to consider. Additionally, some applications may attempt to constantly re-hook if unsuccessful causing log spam.

Instead of logging at the time it was blocked, can we store the names of blocked modules and retrieve the list later in the OBS startup process when logging is available? No need to log in realtime every time they are blocked, just "hey these were blocked on startup". However, if there is no real value in knowing what was blocked, then I'm not bothered by it.

notr1ch:
The injected DLLs can be blocked at any time, not just at startup. It may be misleading to log since if OBS starts before the 3rd party hook has tried to inject, nothing will be logged but it will still be blocked. I'm aware it would be nice to log this data, but I could not come up with a clean / safe implementation of doing so, given that this code exists independently of libobs.

RytoEX:
> The injected DLLs can be blocked at any time, not just at startup. It may be misleading to log since if OBS starts before the 3rd party hook has tried to inject, nothing will be logged but it will still be blocked. I'm aware it would be nice to log this data, but I could not come up with a clean / safe implementation of doing so, given that this code exists independently of libobs.

Hmm. The only other thing I can think of would be to try to log it on OBS shutdown then. That would at least give an overview of what was blocked throughout the entire session.

notr1ch:
I've added a log on shutdown that shows which DLLs have a non-zero blocked count. The OBS log file remains open after libobs has shutdown as the logfile technically belongs to the OBS UI and not libobs. The code also runs outside of the DLL loader lock, so this seems like the safest place to do it.

UI/win-dll-blocklist.c
notr1ch: I don't recall seeing any gpa.dll crashes in a long time so it might be best to remove this one.

RytoEX: If we see such crashes suddenly reappear, we can always block it at that time.

UI/win-dll-blocklist.c
```diff
@@ -0,0 +1,337 @@
...
+
+	// Number of times we've blocked this DLL, for logging purposes
+	uint64_t blocked_count;
+} blocked_module_t;
+
+/*
+* Note: The name matches at the end of the string based on its length, this allows
+* for matching DLLs that may have generic names but a problematic version only
+* exists in a certain directory. A name should always include a path component
+* so that e.g. fraps.dll doesn't match notfraps.dll.
```
RytoEX, line 85: Typically, "e.g." in a sentence when not inside a parenthetical would be used like so:
> so that, e.g., fraps.dll doesn't match notfraps.dll.

UI/win-dll-blocklist.c
```diff
@@ -0,0 +1,337 @@
...
+*/
```
RytoEX, line 86: Asterisks should align:
```
/*
 * Comments go here,
 * and here,
 * for more lines.
 */
```
or
```
/* Comments go here,
 * and here,
 * for more lines.
 */
```

RytoEX (changes requested):
Nits.

UI/win-dll-blocklist.c
notr1ch: Let my British English shine!

RytoEX (approved):
Looks fine to me.

PR #8344, opened 2023-02-25
docs: Add clarifications

CodeYan01:
### Description

Added more info on some functions

### Motivation and Context

It is not immediately clear how to get `obs_scene_t` objects from `obs_enum_scenes`, since it just `obs_source_t` just like `obs_enum_sources`. Developers may also find `obs_enum_scenes` to enumerate in an unexpected order, especially as plugins need to enumerate scenes to show them as options, in which case it has to be in order. Adding a link to the frontend function reduces a lot of development time. The same way, some developers would not expect `obs_scene_enum_items` to start at the bottom, I've seen a few plugins list scene items in a reversed order because this info is not in the docs. Also added the info that `obs_source_get_weak_source` and `obs_weak_source_get_source` returns an incremented reference, which was not clear, especially since we have functions like `obs_scene_get_source` and `obs_sceneitem_get_source` don't increment the reference. The clearer the docs, the less development time for us, especially considering that we'd have to search for code using the functions, along with getting unexpected behavior the first time.

### How Has This Been Tested?

Built and viewed locally, tested the enum functions myself, the unreliable order of `obs_enum_scenes` was confirmed by Rodney (my testing showed that it went either most recent to first created or the reverse).

### Types of changes

### Checklist:

### DIFF
--- a/docs/sphinx/reference-scenes.rst
+++ b/docs/sphinx/reference-scenes.rst
@@ -265,7 +265,8 @@ General Scene Functions
 
 .. function:: void obs_scene_enum_items(obs_scene_t *scene, bool (*callback)(obs_scene_t*, obs_sceneitem_t*, void*), void *param)
 
-   Enumerates scene items within a scene.
+   Enumerates scene items within a scene in order of the bottommost scene item
+   to the topmost scene item.
 
    Callback function returns true to continue enumeration, or false to end
    enumeration.
--- a/docs/sphinx/reference-core.rst
+++ b/docs/sphinx/reference-core.rst
@@ -284,7 +284,8 @@ Libobs Objects
 
 .. function:: void obs_enum_scenes(bool (*enum_proc)(void*, obs_source_t*), void *param)
 
-   Enumerates all scenes.
+   Enumerates all scenes. Use :c:func:`obs_scene_from_source()` if the scene is
+   needed as an :c:type:`obs_scene_t`.
   
    Callback function returns true to continue enumeration, or false to end
    enumeration.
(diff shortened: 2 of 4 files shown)


docs/sphinx/reference-scenes.rst
```diff
@@ -264,7 +264,9 @@ General Scene Functions
 
 .. function:: void obs_scene_enum_items(obs_scene_t *scene, bool (*callback)(obs_scene_t*, obs_sceneitem_t*, void*), void *param)
 
-   Enumerates scene items within a scene.
+   Enumerates scene items within a scene, starting with the bottommost scene
+   item, contrary to the Sources dock showing the topmost scene item as
+   the first, because rendering a scene starts with the bottommost scene item.
```
RytoEX, line 269:
```suggestion
   Enumerates scene items within a scene in order of the bottommost scene item
   to the topmost scene item.
```

While rewriting this, I began to think it was unnecessary to describe the Source dock behavior. As such, I've removed that portion. If you think it is actually needed, perhaps something like this, even though I'm not really happy with this either:
> This enumeration order is contrary to the Sources dock showing the topmost scene item as the first, because rendering a scene starts with the bottommost scene item.

RytoEX (changes requested):

docs/sphinx/reference-scenes.rst
CodeYan01: That is incredibly precise and indeed removes the need to describe the Source dock behavior! Thank you!

CodeYan01:
Applied suggested changes exactly as said, but squashed it onto the last commit

RytoEX (approved):
I wonder a little bit if it would be more clear to refer to the `scene_order` key/array in the scene collection data/JSON. However, this is at least clearer than it used to be, so this is fine to me for now.

PR #8345, opened 2023-02-25
Add unique identifier (UUID) to sources

derrod:
### Description

- Adds libobs platform util for creating UUIDs
- Adds `uuid` to `obs_context_data`
  + Only saves/restores UUIDs for sources
- Adds `obs_get_source_by_uuid()` to retrieve a source by it's UUID

If #8229 gets merged this PR will be updated to use a separate hash table for GUID based lookups.

### Motivation and Context

Developers of plugins and third-party integrations have asked for UUIDs for a long time so that, for example, name changes done via the UI do not result in a desynchronisation between their app's and OBS's state.

#### Why UUIDs/strings?

UUIDv4 is a standardised format that is supported and understood by various frameworks, databases, and standard libraries. They can be serialised/deserialised for storing without any extra work, and can be used in databases with better performance than a generic string type. Additionally, most software engineers should recognise one when they see it, due to their recognisable standard formatting.

While UUIDs in OBS could be stored in a binary representation, this would require us to provide various extra utility functions to convert from/to string representations, as well as potentially having to provide additional functions that do so for the caller (e.g. `obs_get_source_by_uuid_str()`).
Additionally, many API users such as `obs-websocket` would have to convert UUIDs form/to a string format anyway for their use, and even OBS itself would have to convert back and forth when storing/loading scene collections since we use JSON.

In general, I do believe that using a standardised format is beneficial and that storing them in strings is simpler while not being meaningfully slower or memory intensive.

### How Has This Been Tested?

Ran on macOS and Windows, seemed to work fine. UUIDs are saved and restored to scene collection.

Nothing on Linux yet, but any good distro should come with `libuuid`, right?

### Types of changes

### Checklist:

### DIFF
--- a/cmake/Modules/FindLibUUID.cmake
+++ b/cmake/Modules/FindLibUUID.cmake
@@ -1,0 +1,34 @@
+# Stripped down version of
+# https://gitlab.kitware.com/cmake/cmake/blob/e1409101c99f7a3487990e9927e8bd0e275f564f/Source/Modules/FindLibUUID.cmake
+#
+# Distributed under the OSI-approved BSD 3-Clause License.  See accompanying
+# file Copyright.txt or https://cmake.org/licensing for details.
+#
+# Once done these will be defined:
+#
+# LibUUID_FOUND LibUUID_INCLUDE_DIRS LibUUID_LIBRARIES
+
+find_library(LibUUID_LIBRARY NAMES uuid)
+mark_as_advanced(LibUUID_LIBRARY)
+
+find_path(LibUUID_INCLUDE_DIR NAMES uuid/uuid.h)
+mark_as_advanced(LibUUID_INCLUDE_DIR)
+
+include(FindPackageHandleStandardArgs)
+find_package_handle_standard_args(
+  LibUUID
+  FOUND_VAR LibUUID_FOUND
+  REQUIRED_VARS LibUUID_LIBRARY LibUUID_INCLUDE_DIR)
+set(LIBUUID_FOUND ${LibUUID_FOUND})
+
+if(LibUUID_FOUND)
+  set(LibUUID_INCLUDE_DIRS ${LibUUID_INCLUDE_DIR})
+  set(LibUUID_LIBRARIES ${LibUUID_LIBRARY})
+  if(NOT TARGET LibUUID::LibUUID)
+    add_library(LibUUID::LibUUID UNKNOWN IMPORTED)
+    set_target_properties(
+      LibUUID::LibUUID
+      PROPERTIES IMPORTED_LOCATION "${LibUUID_LIBRARY}"
+                 INTERFACE_INCLUDE_DIRECTORIES "${LibUUID_INCLUDE_DIRS}")
+  endif()
+endif()
--- a/libobs/obs-internal.h
+++ b/libobs/obs-internal.h
@@ -520,6 +520,7 @@ typedef void (*obs_destroy_cb)(void *obj);
 
 struct obs_context_data {
 	char *name;
+	const char *uuid;
 	void *data;
 	obs_data_t *settings;
 	signal_handler_t *signals;
@@ -545,8 +546,8 @@ struct obs_context_data {
 
 extern bool obs_context_data_init(struct obs_context_data *context,
 				  enum obs_obj_type type, obs_data_t *settings,
-				  const char *name, obs_data_t *hotkey_data,
-				  bool private);
+				  const char *name, const char *uuid,
+				  obs_data_t *hotkey_data, bool private);
 extern void obs_context_init_control(struct obs_context_data *context,
 				     void *object, obs_destroy_cb destroy);
 extern void obs_context_data_free(struct obs_context_data *context);
@@ -834,7 +835,8 @@ extern struct obs_source_info *get_source_info2(const char *unversioned_id,
 						uint32_t ver);
 extern bool obs_source_init_context(struct obs_source *source,
 				    obs_data_t *settings, const char *name,
-				    obs_data_t *hotkey_data, bool private);
+				    const char *uuid, obs_data_t *hotkey_data,
+				    bool private);
 
 extern bool obs_transition_init(obs_source_t *transition);
 extern void obs_transition_free(obs_source_t *transition);
@@ -851,8 +853,9 @@ extern void audio_monitor_destroy(struct audio_monitor *monitor);
 
 extern obs_source_t *
 obs_source_create_set_last_ver(const char *id, const char *name,
-			       obs_data_t *settings, obs_data_t *hotkey_data,
-			       uint32_t last_obs_ver, bool is_private);
+			       const char *uuid, obs_data_t *settings,
+			       obs_data_t *hotkey_data, uint32_t last_obs_ver,
+			       bool is_private);
 extern void obs_source_destroy(struct obs_source *source);
 
 enum view_type {
--- a/libobs/obs.c
+++ b/libobs/obs.c
@@ -2019,6 +2019,26 @@ obs_source_t *obs_get_source_by_name(const char *name)
 				   obs_source_addref_safe_);
 }
 
+obs_source_t *obs_get_source_by_uuid(const char *uuid)
+{
+	struct obs_source **first = &obs->data.first_source;
+	struct obs_source *source;
+
+	pthread_mutex_lock(&obs->data.sources_mutex);
+
+	source = *first;
+	while (source) {
+		if (strcmp(source->context.uuid, uuid) == 0) {
+			source = obs_source_get_ref(source);
+			break;
+		}
+		source = (struct obs_source *)source->context.next;
+	}
+
+	pthread_mutex_unlock(&obs->data.sources_mutex);
+	return source;
+}
+
 obs_source_t *obs_get_transition_by_name(const char *name)
 {
 	struct obs_source **first = &obs->data.first_source;
@@ -2203,6 +2223,7 @@ static obs_source_t *obs_load_source_type(obs_data_t *source_data,
 	obs_data_array_t *filters = obs_data_get_array(source_data, "filters");
 	obs_source_t *source;
 	const char *name = obs_data_get_string(source_data, "name");
+	const char *uuid = obs_data_get_string(source_data, "uuid");
 	const char *id = obs_data_get_string(source_data, "id");
 	const char *v_id = obs_data_get_string(source_data, "versioned_id");
 	obs_data_t *settings = obs_data_get_obj(source_data, "settings");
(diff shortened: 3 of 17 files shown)


notr1ch:
I'm not a big fan of treating UUIDs as strings, ideally the binary representation should be used everywhere but that's probably not feasible with JSON not having a proper UUID type. I wonder if a simple random string with more entropy would make more sense.

derrod:
> I'm not a big fan of treating UUIDs as strings, ideally the binary representation should be used everywhere but that's probably not feasible with JSON not having a proper UUID type.

I thought about that, but I ultimately figured that most users of the API would ultimately convert it back to a string representation for their own use anyway (e.g. websockets) which just ends up requiring a lot of conversions back and forth.

If we were to store it in binary form then we should also provide functions to convert from/to that form as well as utility wrappers that accept strings and do the conversion for the user (e.g. `obs_get_source_by_uuid_str()`).

> I wonder if a simple random string with more entropy would make more sense.

I suppose any sufficiently random string would do (e.g. created with the help of #5289), but UUIDs are pretty convenient as most platforms provide APIs to create them without us having to do all the heavy lifting when it comes to randomness and such. They are also a format that many databases have native support for so they can be used as primary keys without additional conversions.

cg2121:
I would suggest having UUIDs also for scene items. This way we can easily keep track of them, without also knowing what scene they are in.

cg2121:
Also just for completeness, it would probably be good to also have functions to get the encoder, service, etc. objects by the UUID.

derrod:
> I would suggest having UUIDs also for scene items. This way we can easily keep track of them, without also knowing what scene they are in.

Yes that's a fair point. Should hopefully not be too difficult.

> Also just for completeness, it would probably be good to also have functions to get the encoder, service, etc. objects by the UUID.

I was thinking to actually change it so that if `NULL` is explicitly passed to the context creation it will not create a UUID (the source load function will provide at least an empty string if no UUID is assigned yet).

libobs/obs-internal.h
```diff
@@ -520,6 +520,7 @@ typedef void (*obs_destroy_cb)(void *obj);
 
 struct obs_context_data {
 	char *name;
+	char *uuid;
```
cg2121, line 523: Should probably be a const char, so it can't be modified.

derrod: Since this is allocated at runtime, making it const would mean you could not free the pointer later on. The context data of any OBS object must also never be accessed directly outside the internals, and `obs_source_get_uuid()` should be used instead.

cg2121: See https://github.com/obsproject/obs-studio/blob/fea4a19c8b146d01a0357ba59a1e6003e7018264/libobs/obs-source.h#L223 and https://github.com/obsproject/obs-studio/blob/fea4a19c8b146d01a0357ba59a1e6003e7018264/libobs/obs-source.c#L743 as an example of how it would be done.

derrod: I'm surprised the compiler doesn't complain about the recasting of a const variable to a non-const one here.

derrod:
Figured out the Flatpak build issue by adding a finder for libuuid to our cmake modules. I think this is "done" as far as the actual implementation goes and ready for review.

cg2121:
My idea for the scene item UUID is to have each obs_sceneitem_t struct have its own UUID, separate from the source UUID, so you could find a scene item by obs_find_sceneitem_by_uuid, without knowing what scene it is in or the scene item name, like you have to with obs_scene_find_source(obs_scene_t *scene, const char *name).

derrod:
> My idea for the scene item UUID is to have each obs_sceneitem_t struct have its own UUID, separate from the source UUID, so you could find a scene item by obs_find_sceneitem_by_uuid, without knowing what scene it is in or the scene item name, like you have to with obs_scene_find_source(obs_scene_t *scene, const char *name).

I can certainly see some utility in that, but perhaps that would also be a question for @tt2468 and @Warchamp7 (as is this PR in general).

But for this PR I believe that is out of scope. If global scene item search by UUID is something developers want then I think we can introduce it after #8229 so we don't have to implement a linked list based solution just to replace it wtih hashtables afterwards.

tt2468:
I think that a change like that would only be worthwhile if we fixed up the utilities for moving scene items between scenes and/or duplicating them. There's not much use otherwise, since theoretically if you have the scene item id, you would also have the scene name anyway (in obs-websocket)

Perhaps we could use it to help stabilize the currently horrendous handling of scene item ids for things like groups.

RytoEX (changes requested):
One documentation nit.

Ideally, I'd like someone more familiar with the problem this is solving to review this, but I would like to get this and #8229 merged for the next release.

derrod:
> How is duplicating a scene collection handled?

Since duplicating works by just changing the target file name and saving again it would currently also duplicate the ids of the original. I suppose it is valid to expect them to remain unique even when duplicating an existing collection, so perhaps there could be a step there invalidating and regenerating them.

derrod:
Updated to regenerate UUIDs when a collection is duplicated @notr1ch

cmake/Modules/FindLibUUID.cmake
```diff
@@ -0,0 +1,34 @@
+# Stripped down verison of
```
RytoEX, line 1:
```suggestion
# Stripped down version of
```

libobs/obs.c
```diff
@@ -2552,6 +2590,12 @@ static inline bool obs_context_data_init_wrap(struct obs_context_data *context,
 	if (!context->procs)
 		return false;
 
+	if (uuid && strlen(uuid) == UUID_STR_LENGTH)
+		context->uuid = bstrdup(uuid);
+	/* Only automatically generate UUIDs for sources */
```
RytoEX, line 2595: I personally dislike placing comments between `if` and `else if` portions, but I understand if placing it at the top would be too disconnected.

RytoEX (changes requested):
Nits.

libobs/obs.c
derrod: Yeah there didn't really seem to be any "nice" place to put that comment that didn't look even weirder.

RytoEX (approved):
Nits are addressed from my perspective. Would still like @notr1ch or someone else to give an approval on the implementation.

libobs/obs.c
```diff
@@ -2019,6 +2019,26 @@ obs_source_t *obs_get_source_by_name(const char *name)
...
+
+	pthread_mutex_lock(&obs->data.sources_mutex);
+
+	source = *first;
+	while (source) {
+		if (strcmp(source->context.uuid, uuid) == 0) {
+			source = obs_source_addref_safe_(source);
+			break;
+		}
+		source = (void *)source->context.next;
```
notr1ch, line 2035: Shouldn't this be cast to `obs_source_t *` or `struct obs_source *`? It's inconsistent across libobs, but `obs_source_t *` seems to be more common.

libobs/obs.c
```diff
@@ -2511,6 +2535,21 @@ obs_data_array_t *obs_save_sources(void)
...
+	while (source) {
+		bfree((void *)source->context.uuid);
+		source->context.uuid = os_get_uuid();
+
+		source = (void *)source->context.next;
+	}
+
+	pthread_mutex_unlock(&obs->data.sources_mutex);
+}
+
```
notr1ch, line 2552: This also assigns uuids to scenes, filters, transitions and groups which seems unintended given the "Only automatically generate UUIDs for sources" comment. Perhaps using `obs_enum_sources` with a callback would be safer?

libobs/util/platform.h
```diff
@@ -201,6 +201,10 @@ EXPORT bool os_get_proc_memory_usage(os_proc_memory_usage_t *usage);
 EXPORT uint64_t os_get_proc_resident_size(void);
 EXPORT uint64_t os_get_proc_virtual_size(void);
 
+#define UUID_STR_LENGTH 36
+
+EXPORT char *os_get_uuid(void);
```
notr1ch, line 206: Perhaps `os_generate_uuid` would be a clearer name for this function, `get` could imply it returns something that already exists.

libobs/obs.c
derrod: Scenes, filters, etc. are also sources (`OBS_OBJ_TYPE_SOURCE`), and have UUIDs as a result, which is intentional.
Perhaps the commit title could be reworded to more clearly differentiate between what we expose to the user as "sources" and the internal `struct obs_source` objects which is what it actually refers to.

libobs/obs.c
derrod: Yeah it is inconsistent, this is essentially the same as `obs_get_transition_by_name` immediately below it, which uses `void *`, but I can change it to be clearer.

libobs/util/platform.h
derrod: Makes sense. I'll update the name.

libobs/obs.c
notr1ch: Ah I'd mixed up `OBS_OBJ_TYPE_SOURCE` and `OBS_SOURCE_TYPE_INPUT`. This is fine then, though also has the same `void *` pattern.

libobs/obs.c
derrod: Updated to use `struct obs_source *` now. Seems we generally don't use the typedef'd versions here, but as you said, it's not consistent.

libobs/obs.c
```diff
@@ -2019,6 +2019,26 @@ obs_source_t *obs_get_source_by_name(const char *name)
...
+{
+	struct obs_source **first = &obs->data.first_source;
+	struct obs_source *source;
+
+	pthread_mutex_lock(&obs->data.sources_mutex);
+
+	source = *first;
+	while (source) {
+		if (strcmp(source->context.uuid, uuid) == 0) {
+			source = obs_source_addref_safe_(source);
```
notr1ch, line 2032: I have the feeling that `obs_source_addref_safe_` was only intended to be used with `get_context_by_name` calls. I realise this was probably copied from transitions but `obs_source_get_ref` seems like it might be better here.

derrod: Yeah it probably just exists since the context functions need a callback that takes `void *`, I just updated it to use `obs_source_get_ref()` directly now.

notr1ch (approved):

gxalpha:
> (The function isn't documented since I'm not sure it really needs to be. It's basically just for the OBS Studio UI)

If this is only meant for internal use and not by anyone else, could it be marked as such more clearly, e.g. by a comment in the function header or even an `internal` in the function name?
I'm fairly sure I'll make it unused again with #8446 (since duplicating is implemented differently there to make it work with collections that are currently not active) and it would be nice if I could just remove the function again there without having to worry about API users (if there is a release in which this PR is in but the collection manager is not).

derrod:
> If this is only meant for internal use and not by anyone else, could it be marked as such more clearly, e.g. by a comment in the function header or even an `internal` in the function name?

Perhaps, I don't know if we have any kind of existing convention to deal with that. I'm fine at least adding a comment to the export saying it's only meant to be used by the UI and that it will be removed in a future update.

> I'm fairly sure I'll make it unused again with #8446 (since duplicating is implemented differently there to make it work with collections that are currently not active) and it would be nice if I could just remove the function again there without having to worry about API users (if there is a release in which this PR is in but the collection manager is not).

With your changes you'd still have to remove the UUIDs during duplication. Perhaps by iterating over all the sources as in f55b251fa8da4bc1304e48293b947f02998ad038. Though you'd also have to iterate over all scene items in scenes/groups and remove their `source_uuid` property.

derrod:
Gonna merge in a bit given the approvals. Note that the last force-push that github shows is a bit misleading, the only change since yesterday was the comment `obs.h`, the other change was something I did not mean to push to the PR.

PR #8347, opened 2023-02-25
Disable encoder scaling if scaled resolution matches output resolution

notr1ch:
### Description
Adds an additional check to `obs_encoder_set_scaled_size` that disables scaling if the requested scale resolution matches the output resolution.

DRAFT: Is this the correct place to do this? Should the logic also / instead be in the UI to not request the scaling in the first place?

### Motivation and Context
Users would often tick "Enable scaling" in Advanced Output mode and then leave the resolution the same as their main resolution. OBS would dutifully call `obs_encoder_set_scaled_size` with the same resolution as the output, causing an unnecessary pass through swscale and skipping hardware  encoders like jim-nvenc that don't handle rescaled output.

### How Has This Been Tested?
Ran OBS, tried to "rescale" 1080p to 1080p, scaling was disabled and jim-nvenc worked.

### Types of changes

### Checklist:

### DIFF
--- a/libobs/obs-encoder.c
+++ b/libobs/obs-encoder.c
@@ -694,6 +694,18 @@ void obs_encoder_set_scaled_size(obs_encoder_t *encoder, uint32_t width,
 		return;
 	}
 
+	const struct video_output_info *voi;
+	voi = video_output_get_info(encoder->media);
+	if (voi && voi->width == width && voi->height == height) {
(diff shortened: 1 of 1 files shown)


jp9000:
It's fine, I'd rather just have this code for the time being rather than try to mess around with the UI at this point trying to exclude some resolution.

PR #8351, opened 2023-02-26
UI: Add ability to re-color icons

cg2121:
### Description
This adds the ability to change icon colors by qss.

### Motivation and Context
@Warchamp7 really wants this feature because it prevents having to maintain several different sets of icons. It currently doesn't work with combobox or spinbox up/down icons.

### How Has This Been Tested?
Changed icon color in qss and made sure they were the correct color.

### Types of changes

### Checklist:

### DIFF
--- a/UI/obs-frontend-api/obs-frontend-api.h
+++ b/UI/obs-frontend-api/obs-frontend-api.h
@@ -242,6 +242,9 @@ EXPORT void obs_frontend_add_undo_redo_action(
 	const char *name, const undo_redo_cb undo, const undo_redo_cb redo,
 	const char *undo_data, const char *redo_data, bool repeatable);
 
+EXPORT void obs_frontend_set_icon(void *obj, const char *path,
+				  const char *theme_id);
+
 /* ------------------------------------------------------------------------- */
 
 #ifdef __cplusplus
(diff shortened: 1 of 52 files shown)


UI/obs-frontend-api/obs-frontend-api.h
```diff
@@ -242,6 +242,9 @@ EXPORT void obs_frontend_add_undo_redo_action(
 	const char *name, const undo_redo_cb undo, const undo_redo_cb redo,
 	const char *undo_data, const char *redo_data, bool repeatable);
 
+EXPORT void obs_frontend_set_icon(void *obj, const char *path,
+				  const char *theme_id);
+
```
tytan652, line 247: Needs documentation.

It also feels out of scope without context.

tytan652 (changes requested):

WizardCM:
The more I read through this, the less maintainable it looks. While the implementation seems alright enough, it's a lot of code for something that _should_ be relatively simple, and requires extra consideration whenever a new icon is added & styled. Additionally, if Qt ever does add proper SVG support, this would have to be cleaned up in reverse.

While I like the idea of this feature, I don't think it's a good idea for us to implement it on our side. Ideally, we push for Qt to implement proper `currentColor` support on their end.

cg2121:
After thinking about it, this shouldn't be something we should be doing, as Qt should be doing it. Just too much code. This is an old branch that I had laying around, as mostly experimental, so I don't really care much about it anyway.

PR #8364, opened 2023-02-27
Explicitly close RTMP socket on send error

notr1ch:
### Description
Explicitly closes the RTMP socket if `send()` returns an error.

### Motivation and Context
Some `send()` errors such as a timeout with no data sent are not fatal, but librtmp will still shut down the connection on an error. During shutdown, librtmp attempts to send an `FCunpublish` message which would usually not make it to the service since the socket has been torn down due to the error. But depending on the type of `send()` error, the `FCunpublish` message could actually make it through to the remote side and cause it to think a clean shutdown occurred, disabling features like "disconnect protection".

### How Has This Been Tested?
Not tested but Seems Good™

### Types of changes

### Checklist:

### DIFF
--- a/plugins/obs-outputs/librtmp/rtmp.c
+++ b/plugins/obs-outputs/librtmp/rtmp.c
@@ -1567,6 +1567,7 @@ static int
 WriteN(RTMP *r, const char *buffer, int n)
 {
     const char *ptr = buffer;
+    struct linger l;
 
     while (n > 0)
     {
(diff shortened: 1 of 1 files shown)


jdeng89:
Did some initial testing with the same process used to initially repro this bug and no longer seem to be able to repro. Will do a rebuild with both this and the extended logging to confirm later this week and will attach afterwards

jdeng89:
I re-tested this again with additional RTMP logging and could not re-produce the issue. Below is the OBS log.

https://obsproject.com/logs/fAeHv2aFZ6IUE041

RytoEX:
That sounds like as much confirmation as we're likely to get. I don't know how else we might test this.

tt2468:
The underlying bug is timing-based and also depends on the network conditions that caused the send error, so reproduction of the bug is not expected to be reliable.

notr1ch:
To clarify on the "not tested" part, I can't trigger the error condition but I believe this code to be the proper solution to the problem. This shouldn't be able to make the problem any worse than it already is - the socket is being closed regardless during shutdown after the error, this just makes it more explicit and earlier. I'm confident that merging this will be fine.

PR #8367, opened 2023-02-27
Add mutex for writing to the log file

notr1ch:
### Description
Adds a mutex that is locked when writing to the log file and removes inline on the function.

### Motivation and Context
fstream is not thread-safe and the act of writing the string and the newline are two separate operations which could execute concurrently in multiple threads, resulting in lines joining together followed by two newlines (example: https://obsproject.com/logs/H9kyOhPqQEOwEXeW)

### How Has This Been Tested?
Ran OBS, it worked.

### Types of changes

### Checklist:

### DIFF
--- a/UI/obs-app.cpp
+++ b/UI/obs-app.cpp
@@ -277,14 +277,17 @@ string CurrentDateTimeString()
 	return buf;
 }
 
-static inline void LogString(fstream &logFile, const char *timeString,
-			     char *str, int log_level)
+static void LogString(fstream &logFile, const char *timeString, char *str,
+		      int log_level)
 {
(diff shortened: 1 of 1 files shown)


prgmitchell:
Tested this, seems to work as advertised!
I opened OBS on my 29.0.2 install 5 times and each log had the StreamDeck plugin logging inserted into the loaded modules list like so: https://obsproject.com/logs/GafBxatjDtu1nGBn

I then opened OBS running the code from this PR and the StreamDeck logging was no longer in the middle of the loaded modules list as seen here: https://obsproject.com/logs/Wgg4kzYfZ3rYGnz8

PR #8372, opened 2023-02-28
UI: Use OBSBasic.ui generated by Qt6 Creator

cg2121:
### Description
Apparently, the file has been manually updated, and Qt Creator generates a different result, so it is annoying to edit commits. And some of the file was generated by Qt5 Creator, which produces a slightly different file than Qt6 Creator.

### Motivation and Context
Gets annoying to use git add -p when editing the OBSBasic.ui file in Qt Creator.

### How Has This Been Tested?
Compiled and ran OBS, and everything still works.

### Types of changes

### Checklist:

### DIFF
--- a/UI/forms/OBSBasic.ui
+++ b/UI/forms/OBSBasic.ui
@@ -487,7 +487,7 @@
      <x>0</x>
      <y>0</y>
      <width>1079</width>
-     <height>21</height>
+     <height>23</height>
     </rect>
    </property>
    <widget class="QMenu" name="menu_File">
@@ -666,6 +666,11 @@
     <property name="title">
      <string>Basic.MainMenu.View</string>
     </property>
+    <widget class="QMenu" name="multiviewProjectorMenu">
+     <property name="title">
+      <string>MultiviewProjector</string>
+     </property>
+    </widget>
     <action name="resetUI">
      <property name="text">
       <string>Basic.MainMenu.View.ResetUI</string>
@@ -693,11 +698,6 @@
     <addaction name="multiviewProjectorWindowed"/>
     <addaction name="separator"/>
     <addaction name="actionAlwaysOnTop"/>
-   </widget>
-   <widget class="QMenu" name="multiviewProjectorMenu">
-    <property name="title">
-     <string>MultiviewProjector</string>
-    </property>
    </widget>
    <widget class="QMenu" name="menuTools">
     <property name="title">
@@ -732,7 +732,7 @@
   <widget class="OBSBasicStatusBar" name="statusbar"/>
   <widget class="OBSDock" name="scenesDock">
    <property name="features">
-    <set>QDockWidget::DockWidgetClosable|QDockWidget::DockWidgetMovable|QDockWidget::DockWidgetFloatable</set>
+    <set>QDockWidget::DockWidgetClosable|QDockWidget::DockWidgetFloatable|QDockWidget::DockWidgetMovable</set>
    </property>
    <property name="windowTitle">
     <string>Basic.Main.Scenes</string>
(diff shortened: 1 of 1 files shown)


jp9000:
You probably want to use Qt Designer, not Qt Creator.

cg2121:
Qt Designer is a part of Qt Creator. I don't think they are separate programs as of Qt6.

derrod:
> You probably want to use Qt Designer, not Qt Creator. We don't use Qt Creator with OBS.

Qt Designer is part of Qt Creator, the standalone Qt Designer app no longer exists.

jp9000:
I'm almost totally positive Qt Designer does still exist. It's usually built with Qt. I don't think they distribute it with normal Qt distributions but I'm almost totally positive it still exists and that they do not force people to use Qt Creator. The reason why I asked about Qt Designer is because we had a case previously where Qt Creator was not producing exactly the same output as Qt Designer. I use Qt Designer personally. It's much cleaner and more simple to use.

derrod:
I think we could run this process on all `.ui` files since many of the format changes apply to most of them.

> I'm almost totally positive Qt Designer does still exist. It's usually built with Qt. I don't think they distribute it with normal Qt distributions but I'm almost totally positive it still exists and that they do not force people to use Qt Creator. The reason why I asked about Qt Designer is because we had a case previously where Qt Creator was not producing exactly the same output as Qt Designer. I use Qt Designer personally. It's much cleaner and more simple to use.

You're right, it's not installed a sa standalone application but it is included as `designer.exe` in the `bin/` directory.

Anyway, there are no differences between the output of Qt Designer and Qt Creator that I can find. The only odd thing is that for me both remove `multiviewProjectorMenu` entirely rather than just moving it, and I don't know why.

Chiitoo:
> It's usually built with Qt. I don't think they distribute it with normal Qt distributions

A bit more specifically, it's an optional component of Qt Tools, and it may or may not be included depending on who packaged/built Qt (would imagine most pre-built distributions would include it).

I've not seen any talk about it going away any time soon while following Qt development someawhats, and packaging it for a Linux distribution (can't promise I never miss a thing of course).

1. https://code.qt.io/cgit/qt/qttools.git/

Warchamp7:
> I'm almost totally positive Qt Designer does still exist. It's usually built with Qt. I don't think they distribute it with normal Qt distributions but I'm almost totally positive it still exists and that they do not force people to use Qt Creator. The reason why I asked about Qt Designer is because we had a case previously where Qt Creator was not producing exactly the same output as Qt Designer. I use Qt Designer personally. It's much cleaner and more simple to use.

Qt Designer is the Design tab of Qt Creator. It's the same functionality just rolled into a larger app. The mode is even [referred to as Qt Designer in the official docs(https://doc.qt.io/qtcreator/creator-using-qt-designer.html).

qtcreator-formedit

UI/forms/OBSBasic.ui
```diff
@@ -487,7 +487,7 @@
      <x>0</x>
      <y>0</y>
      <width>1079</width>
-     <height>21</height>
+     <height>23</height>
```
PatTheMav, line 490: This seems to be the only change that actually also changes a value? Is this intended?

PatTheMav:
I agree that using Qt Creator/Designer as the "ground truth" for UI files is a decent choice (not everybody might agree with its formatting, but at least it would be consistent) and as outlined in the OP would make diffs easier to handle.

If the value change is addressed I think this just needs another pass through Qt Creator to update the file and should be good to merge then.

Warchamp7:
> Anyway, there are no differences between the output of Qt Designer and Qt Creator that I can find. The only odd thing is that for me both remove `multiviewProjectorMenu` entirely rather than just moving it, and I don't know why.

I believe this is due to that entry being in the wrong place in the file.

cg2121:
#10419 is replacing this.

cg2121:
https://github.com/obsproject/obs-studio/pull/10419 is replacing this.

Issue #8199, opened 2023-02-03
Full/partial color range switch is broken when using custom output through FFmpeg


mia-is-baka:
### Operating System Info

Windows 11

### OBS Studio Version

Other

### OBS Studio Version (Other)

29.0.1

### OBS Studio Log URL

https://obsproject.com/logs/BgY3fRZrYgONbSLt

### Expected Behavior

Full/partial color range switch is working properly with custom FFmpeg output

### Current Behavior

If you select full color range in advanced settings and try to record video through custom FFmpeg output, OBS marks output file as full range in metadata (you can see it with MediaInfo, for example), but actually encodes it always as partial range, so colors get washed out

### Steps to Reproduce

1. Launch OBS
2. Setup custom FFmpeg output (for example, mkv container with plain H.264 through libx264 or h264_nvenc)
3. Set color range to full
4. Record short video
5. Look at washed out result

### Anything else we should know?

It worked properly until release 28.x.x

#### Broken full range example

<details>
    <summary>OBS settings</summary>
    <img src="https://user-images.githubusercontent.com/109112234/216574642-5c904d38-23a0-4449-9b64-d13d610277db.png" alt=""/>
    <img src="https://user-images.githubusercontent.com/109112234/216574661-ad90c1c9-efbc-4249-a9fc-70204405715a.png" alt=""/>
    <img src="https://user-images.githubusercontent.com/109112234/216574674-9c4a664e-029e-4436-aea2-abcc33a73a8e.png" alt=""/>
</details>

<details>
    <summary>MediaInfo</summary>
    <img src="https://user-images.githubusercontent.com/109112234/216576115-1ea868d7-b895-4cbd-be62-a232a2067359.png" alt=""/>
</details>

<details>
    <summary>Reference image</summary>
    <img src="https://user-images.githubusercontent.com/109112234/216579504-0a452fa9-d45d-4120-9e54-792e304196bf.jpg" alt=""/>
</details>

<details>
    <summary>Frame from video (screenshot of clsid2/mpc-hc  with madVR)</summary>
    <img src="https://user-images.githubusercontent.com/109112234/216579657-d6646497-6d85-436c-98b3-e93680da75ae.jpg" alt=""/>
</details>

mia-is-baka:
I did a little research on the code and found that in video-io.c:video_output_connect video_scale_info field of input is not initialized properly

```c
-- snip --

    if (video_get_input_idx(video, callback, param) == DARRAY_INVALID) {
[...]

Issue #7602, opened 2022-10-15
FFmpeg output + full range results in limited range video with full range metadata


tearfur:
### Operating System Info

Windows 11

### OBS Studio Version

28.0.3

### OBS Studio Log URL

https://obsproject.com/logs/rZsqb0V-izUSsph2

### Expected Behavior

A full range video is produced.

### Current Behavior

A limited range video marked as full range in metadata is produced.

### Steps to Reproduce

1. Create a new profile.
2. Open "Settings", go to the "Output" tab.
3. Choose "Advanced" output mode and "Custom Output (FFmpeg)" type.
4. Choose "matroska" as the container format.
5. Go to the "Advanced" tab and choose "Full" colour range with everything else as default.
6. Save the settings and record a test clip.

### Anything else we should know?

I did not test other formats, so it may not just be applicable to the matroska format.

Issue #5363, opened 2021-10-01
Setting full range video results in files that claim limited range video


mzso:
### Operating System Info

Windows 10

### OBS Studio Version

Other

### OBS Studio Version (Other)

26.1.1

### OBS Studio Log URL

https://obsproject.com/logs/9Ieoxas0WCh9m4lL

### Expected Behavior

Recorded video being full range, and also the playback software being signalled as such.

### Current Behavior

The video is in fact full range, but the decoder receives the information that it's limited range.
I have LAV decoder and madVR. The latter shows "limited range (says upstream)".
This in past experience meant that some data is set wrongfully which misleads the decoder.

I can verify that the video is actually proper full range, by setting madVR to override to PC range, in which case the video displays properly.

### Steps to Reproduce

1. Set lossless quality in output, simple mode.
2. Set I444, 709 color space and Full range in advanced > video.
3. record a large size video.
3. observe the the recorded video, being borked, they way I described.
...

### Anything else we should know?

Also weird, and potentially a bug, is that if I record a smaller video size (don't know exact dimensions) the recorded video will be RGB as indicated by madVR's OSD, not sure if it's right. In which case madVR also shows "full range (best guess)", so I guess range data is not set. But the video definitely appears normally.

I don't know if the problem is container related, because I see no option to change container in simple mode. And archaic .avi is used.
I made a very short sample: https://drive.google.com/file/d/1RjlVCOMPsIi6ycOSWLp-P8F0pLT5Bm1h/view?usp=sharing

SuslikV:
The forum's origin:
https://obsproject.com/forum/threads/i-want-to-record-actully-lossless-video-full-range-rgb-no-subsampling-how-do-i-do-it.148637/

lextra2:
NV12 with bt709 & Full range works just fine for Nvidia,
[as you can see](https://i.imgur.com/fB4Iqfh.jpeg)

Now the question is, how to tag the stream correctly to make this work for other encoders?

flaeri:
@lextra2 This issue is in relation to the AVI container and UTVideo. Unrelated to nvidia/amd, what have you.

lextra2:
So turns out, the actual issue is that twitch does not accept Full for AMD, while it works for Nvidia.

I never actually tested local recordings (until now) which does work just fine.

ValZapod:
>also shows "full range (best guess)", so I guess range data is not set

[...]

PR #8202, opened 2023-02-03
Fix full/limited range and RGB color space handling

mia-is-baka:
### Motivation and Context

It fixes #8199 

### How Has This Been Tested?

I recorded videos with multiple combinations of color range, color space and output methods and then verified that all ok with image and file metadata

### Types of changes

- Bug fix (non-breaking change which fixes an issue)

### Checklist:

- [x] My code has been run through [clang-format](https://github.com/obsproject/obs-studio/blob/master/.clang-format).
- [x] I have read the [**contributing** document](https://github.com/obsproject/obs-studio/blob/master/CONTRIBUTING.rst).
- [x] My code is not on the master branch.
- [x] The code has been tested.
- [x] All commit messages are properly formatted and commits squashed where appropriate.
- [ ] I have included updates to all appropriate documentation.

PR #8374, opened 2023-02-28
libobs/media-io: Add color range and colorspace data to conversion

flaeri:
### Description
By adding the range data we can get successful handling of full range color via custom ffmpeg output. Previously, it would always default to 0 which would would yield partial/limited output.

Adds colorspace info. Currently the the output file is correctly tagged, but not undergone the actual conversion.

Big thank you to @RytoEX who helped me work out some of this libobs stuff, and to @mia-is-baka for their code submission.

### Motivation and Context
Fixes https://github.com/obsproject/obs-studio/issues/7602
Fixes https://github.com/obsproject/obs-studio/issues/8199
Fixes https://github.com/obsproject/obs-studio/issues/5363

https://github.com/obsproject/obs-studio/pull/8202 was closed.

### How Has This Been Tested?
On Win10 22H2 I ran thru a battery of codecs and containers to test and observe said change, and compare with expected behavior from FFmpeg commandline.

I used ffprobe to check for tagging, as well as ffplay and mpv to confirm output was valid, by recording a black screen with this image 0-16-doubleBar

The background will be 0,0,0 and 0-16 and 235-255 is visible (the bars are not the same). Background is 0,0,0, not 16,16,16.

There are several codec and container variations that do *not* work with full range, and they do not work in FFmpeg CLI either. This is expected behavior, and users would be expected to be mindful of these combinations.

#### Crashes

In #8202 I noted several crashes. I've investigated both of these thoroughly, and it is two distinct crashes that turned out to be unrelated to the color range change (not sure about the other changes).

Crash in swscale happens when you go from yuv to RGB24, which happens by default with qtrle, both on release, master and with this change.

Crash in avcodec happens specifically with ffv1, on release, current master and with this change. Happens regardless of partial/full prior to this change, and does not seem to worsen in any way by this change.

### Types of changes

### Checklist:

### DIFF
--- a/libobs/media-io/video-io.c
+++ b/libobs/media-io/video-io.c
@@ -392,6 +392,8 @@ bool video_output_connect(
 			input.conversion.format = video->info.format;
 			input.conversion.width = video->info.width;
 			input.conversion.height = video->info.height;
+			input.conversion.range = video->info.range;
+			input.conversion.colorspace = video->info.colorspace;
 		}
 
 		if (input.conversion.width == 0)


flaeri:
Updated to include colorspace. Setting as draft until testing is complete.

flaeri:
Tested both nv12 and i444, in the most common codecs (h264 and hevc) and containers (mp4 and mkv).
SRGB, rec601 and rec709 now matches what we get in in "standard" mode.

Can't really correctly test HDR space, since I have no supported monitor.

JonasCz:
I can confirm that this fixes my problem: https://github.com/obsproject/obs-studio/issues/8199#issuecomment-1447974174

tearfur:
Thank you so much for this. I mainly use fffmpeg output to have lossless audio tracks, having full range again was a nice surprise.

PR #8377, opened 2023-02-28
Support MSVC standard conforming mode

derrod:
### Description

Enable compiling OBS with `/permissive- /Zc:__cplusplus /Zc:preprocessor /std:c17`.

This is the bare-minimum to get OBS (minus `enc-amf`) to compile and run with these options.

See: https://learn.microsoft.com/en-us/cpp/preprocessor/preprocessor-experimental-overview

Draft because I'm not 100% happy/sure about some of these. 

### Motivation and Context

Want to not use `/permissive-` anymore and use more consistent build options with *nix.

### How Has This Been Tested?

Build and ran OBS.

### Types of changes

### Checklist:

### DIFF
--- a/UI/frontend-plugins/frontend-tools/captions-handler.hpp
+++ b/UI/frontend-plugins/frontend-tools/captions-handler.hpp
@@ -25,8 +25,8 @@ public:
 
 typedef std::function<void(const std::string &)> captions_cb;
 
-#define captions_error(s) std::string(obs_module_text("Captions.Error."##s))
-#define CAPTIONS_ERROR_GENERIC_FAIL captions_error("GenericFail")
+#define CAPTIONS_ERROR_GENERIC_FAIL \
+	std::string(obs_module_text("Captions.Error.GenericFail"))
 
 /* ------------------------------------------------------------------------- */
 
--- a/UI/window-basic-settings.cpp
+++ b/UI/window-basic-settings.cpp
@@ -603,20 +603,21 @@ OBSBasicSettings::OBSBasicSettings(QWidget *parent)
 		ui->hideOBSFromCapture = nullptr;
 	}
 
-#define PROCESS_PRIORITY(val)                                                \
-	{                                                                    \
-		"Basic.Settings.Advanced.General.ProcessPriority."##val, val \
-	}
-
 	static struct ProcessPriority {
 		const char *name;
 		const char *val;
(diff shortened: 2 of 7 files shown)


RytoEX (changes requested):
The changes overall looked fine to me. We are still shipping minor fixes for enc-amf for now, particularly for build fixes (such as `/utf-8`), so feel free to PR those changes on that repo. Please move the `cmake: Enable stricter MSVC compiler options` commit to be last so that each commit can compile (assuming the changes work with both compiler configurations).

derrod:
In that case would you mind merging https://github.com/obsproject/obs-amd-encoder/pull/435 so that I can rebase make this PR update the submodule to keep a working build?

notr1ch (approved):
LGTM.

RytoEX (approved):

derrod:
Ignore that force push, the changes are exactly the same, I just found a typo in the first commit message that I *needed* to fix.

SetoKaiba:
https://youtrack.jetbrains.com/issue/CPP-32997/Reload-CMake-Project-problem-in-CLion
Hi, when I use CLion to open the project. It won't reload the project. 
I can reload the project with /std:c17 removed from CompilerConfig.cmake.

```
Command line error D8016 : '/std:c17' and '/std:c++17' command-line options are incompatible
```

SetoKaiba:
@derrod Any idea on this? Is it a bug of CLion? Or the param shouldn't be here?

derrod:
I have no idea. I haven't used CLion on Windows for OBS.

SetoKaiba:
In windows, /std:c17 and /std:c++17 is incompatible. Maybe the /std:c17 should be only applied to the project that need /std:c17. Previously, there's no /std:c17 in the CompilerConfig.cmake.

RytoEX:
> In windows, /std:c17 and /std:c++17 is incompatible. Maybe the /std:c17 should be only applied to the project that need /std:c17. Previously, there's no /std:c17 in the CompilerConfig.cmake. And it works for CLion.

It works fine for MSVC, which is the compiler that we support on Windows. That said, we have made an adjustment in #8694 to use CMake flags to set the C standard instead of using a compiler flag directly.

SetoKaiba:
I tried #8684. It fixes the issue. Thank you.

PR #8380, opened 2023-03-01
libobs/media-io: Use inputs_mutex during close

kkartaltepe:
### Description

This code assumed that no one would be racing after we stopped the input thread, however the obs output thread also races and can free the inputs during disconnect before we finish closing the video mix.

Instead keep the locks around after stop and use them during close to avoid races.

### Motivation and Context

Fixes crashes when stopping virtual webcam on linux.

### How Has This Been Tested?

Starting and stopping virtual webcam no longer crashes with double free.

### Types of changes

### Checklist:

### DIFF
--- a/libobs/media-io/video-io.c
+++ b/libobs/media-io/video-io.c
@@ -69,8 +69,6 @@ struct video_output {
 	volatile long skipped_frames;
 	volatile long total_frames;
 
-	bool initialized;
-
 	pthread_mutex_t input_mutex;
 	DARRAY(struct video_input) inputs;
 
@@ -236,7 +234,6 @@ int video_output_open(video_t **video, struct video_output_info *info)
 	memcpy(&out->info, info, sizeof(struct video_output_info));
 	out->frame_time =
 		util_mul_div64(1000000000ULL, info->fps_den, info->fps_num);
-	out->initialized = false;
 
 	if (pthread_mutex_init_recursive(&out->data_mutex) != 0)
 		goto fail0;
@@ -249,7 +246,6 @@ int video_output_open(video_t **video, struct video_output_info *info)
 
 	init_cache(out);
 
-	out->initialized = true;
 	*video = out;
 	return VIDEO_OUTPUT_SUCCESS;
 
@@ -260,7 +256,7 @@ fail2:
 fail1:
 	pthread_mutex_destroy(&out->data_mutex);
 fail0:
-	video_output_close(out);
+	bfree(out);
 	return VIDEO_OUTPUT_FAIL;
 }
 
@@ -271,12 +267,19 @@ void video_output_close(video_t *video)
 
 	video_output_stop(video);
 
+	pthread_mutex_lock(&video->input_mutex);
+
 	for (size_t i = 0; i < video->inputs.num; i++)
 		video_input_free(&video->inputs.array[i]);
 	da_free(video->inputs);
 
 	for (size_t i = 0; i < video->info.cache_size; i++)
 		video_frame_free((struct video_frame *)&video->cache[i]);
+
+	pthread_mutex_unlock(&video->input_mutex);
+	os_sem_destroy(video->update_semaphore);
+	pthread_mutex_destroy(&video->data_mutex);
+	pthread_mutex_destroy(&video->input_mutex);
 
 	bfree(video);
 }
@@ -535,14 +538,10 @@ void video_output_stop(video_t *video)
 	if (!video)
 		return;
 
-	if (video->initialized) {
-		video->initialized = false;
+	if (!video->stop) {
 		video->stop = true;
 		os_sem_post(video->update_semaphore);
 		pthread_join(video->thread, &thread_ret);
-		os_sem_destroy(video->update_semaphore);
-		pthread_mutex_destroy(&video->data_mutex);
-		pthread_mutex_destroy(&video->input_mutex);
 	}
 }
 


libobs/media-io/video-io.c
```diff
@@ -271,12 +271,23 @@ void video_output_close(video_t *video)
...
+		video->initialized = false;
+		pthread_mutex_lock(&video->input_mutex);
 
-	for (size_t i = 0; i < video->info.cache_size; i++)
-		video_frame_free((struct video_frame *)&video->cache[i]);
+		for (size_t i = 0; i < video->inputs.num; i++)
+			video_input_free(&video->inputs.array[i]);
+		da_free(video->inputs);
+
+		for (size_t i = 0; i < video->info.cache_size; i++)
```
tt2468, line 282: Is this a memory leak that was previously unhandled?

kkartaltepe: no the diff just looks weird you can see we were freeing video->inputs and video->cache before and are still freeing them after. But the diff interpreted the if statement as rewriting the freeing of video->inputs.

tt2468: Oh ok yeah I see that now, sorry.

notr1ch:
What's the intended difference between `video_output_close` and `video_output_stop`? Both seem to shut things down with no way to restart and it's unclear to me which one should be called when. Perhaps some of this could be consolidated?

kkartaltepe:
> What's the intended difference between `video_output_close` and `video_output_stop`? Both seem to shut things down with no way to restart and it's unclear to me which one should be called when. Perhaps some of this could be consolidated?

thats a good question, it seems the purpose is to stop to the input processing thread and can end up being called from a couple places (and is exported). So removing it didnt seem entirely reasonable or I would have just removed it as I agree its confusing. But since its exported  just moved the destruction of resources out.

_close is simply the pair to _open which is the allocation and initialization of the mixer.

libobs/media-io/video-io.c
```diff
@@ -533,14 +544,10 @@ void video_output_stop(video_t *video)
 	if (!video)
 		return;
 
-	if (video->initialized) {
-		video->initialized = false;
+	if (video->initialized && !video->stop) {
```
notr1ch, line 547: You mentioned there were race conditions - do we need to be concerned with thread safety here, for example two threads calling `video_output_stop` on the same video output?

kkartaltepe: There is no indication that we have that occuring, but we could add another mutex here to avoid that in the future.

kkartaltepe:
From discussion:

Failure in the video-io:

```
#0  bfree (ptr=0x7ffebb8f7020) at /home/kk/ceph-data/projects/mine/obs-studio-main/libobs/util/bmem.c:142
#1  0x00007ffff65af2d6 in video_frame_free (frame=0x5555573158e0) at /home/kk/ceph-data/projects/mine/obs-studio-main/libobs/media-io/video-frame.h:35
#2  video_frame_free (frame=0x5555573158e0) at /home/kk/ceph-data/projects/mine/obs-studio-main/libobs/media-io/video-frame.h:32
#3  video_input_free (input=0x5555573158c0) at /home/kk/ceph-data/projects/mine/obs-studio-main/libobs/media-io/video-io.c:56
#4  video_output_close (video=0x5555572c3020) at /home/kk/ceph-data/projects/mine/obs-studio-main/libobs/media-io/video-io.c:275
#5  video_output_close (video=0x5555572c3020) at /home/kk/ceph-data/projects/mine/obs-studio-main/libobs/media-io/video-io.c:267
#6  0x00007ffff653a4cc in obs_free_video_mix (video=0x55555732cb00) at /home/kk/ceph-data/projects/mine/obs-studio-main/libobs/obs.c:838
#7  0x00007ffff658695f in output_frames () at /home/kk/ceph-data/projects/mine/obs-studio-main/libobs/obs-video.c:935
#8  obs_graphics_thread_loop (context=context@entry=0x7fff7ceb44c0) at /home/kk/ceph-data/projects/mine/obs-studio-main/libobs/obs-video.c:1149
#9  0x00007ffff65887e8 in obs_graphics_thread (param=<optimized out>) at /home/kk/ceph-data/projects/mine/obs-studio-main/libobs/obs-video.c:1219
#10 0x00007ffff429ebb5 in  () at /usr/lib/libc.so.6
#11 0x00007ffff4320d90 in  () at /usr/lib/libc.so.6
```

This results from access the video capture thread that is also accessing frames.:

```
#0  0x00007ffff42a08ec in  () at /usr/lib/libc.so.6
#1  0x00007ffff4251ea8 in raise () at /usr/lib/libc.so.6
#2  0x00007ffff423b53d in abort () at /usr/lib/libc.so.6
#3  0x00007ffff423c29e in  () at /usr/lib/libc.so.6
#4  0x00007ffff42aa657 in  () at /usr/lib/libc.so.6
#5  0x00007ffff42ac7bc in  () at /usr/lib/libc.so.6
#6  0x00007ffff42aee63 in free () at /usr/lib/libc.so.6
#7  0x00007fffe7459d71 in sws_freeContext () at /usr/lib/libswscale.so.6
#8  0x00007ffff65afd51 in video_scaler_destroy (scaler=0x555556f90a20) at /home/kk/ceph-data/projects/mine/obs-studio-main/libobs/media-io/video-scaler-ffmpeg.c:234
#9  video_scaler_destroy (scaler=0x555556f90a20) at /home/kk/ceph-data/projects/mine/obs-studio-main/libobs/media-io/video-scaler-ffmpeg.c:231
#10 0x00007ffff65aef52 in video_input_free (input=0x55555728f4e0) at /home/kk/ceph-data/projects/mine/obs-studio-main/libobs/media-io/video-io.c:57
#11 video_output_disconnect (video=0x5555572b3700, callback=<optimized out>, param=<optimized out>)
    at /home/kk/ceph-data/projects/mine/obs-studio-main/libobs/media-io/video-io.c:448
#12 0x00007ffff653d7b2 in stop_raw_video (v=<optimized out>, callback=<optimized out>, param=<optimized out>) at /home/kk/ceph-data/projects/mine/obs-studio-main/libobs/obs.c:2854
#13 0x00007ffff655fee6 in end_data_capture_thread (data=0x555556ec5700) at /home/kk/ceph-data/projects/mine/obs-studio-main/libobs/obs-output.c:2284
#14 0x00007ffff429ebb5 in  () at /usr/lib/libc.so.6
#15 0x00007ffff4320d90 in  () at /usr/lib/libc.so.6
```

norihiro:
I feel removing `initialized` and directly calling `bfree` from `video_output_open` would be a simpler implementation.
Since `video_output_open` never returns with `initialized` = `false`, `initialized` is not necessary outside `video_output_open`, we can remove the variable `initialized`.
```patch
diff --git a/libobs/media-io/video-io.c b/libobs/media-io/video-io.c
index 99fb79cea..a33ce72f1 100644
--- a/libobs/media-io/video-io.c
+++ b/libobs/media-io/video-io.c
@@ -69,8 +69,6 @@ struct video_output {
 	volatile long skipped_frames;
 	volatile long total_frames;
 
-	bool initialized;
-
 	pthread_mutex_t input_mutex;
 	DARRAY(struct video_input) inputs;
 
@@ -236,7 +234,6 @@ int video_output_open(video_t **video, struct video_output_info *info)
 	memcpy(&out->info, info, sizeof(struct video_output_info));
 	out->frame_time =
 		util_mul_div64(1000000000ULL, info->fps_den, info->fps_num);
-	out->initialized = false;
 
 	if (pthread_mutex_init_recursive(&out->data_mutex) != 0)
 		goto fail0;
@@ -249,7 +246,6 @@ int video_output_open(video_t **video, struct video_output_info *info)
 
 	init_cache(out);
 
-	out->initialized = true;
 	*video = out;
 	return VIDEO_OUTPUT_SUCCESS;
 
@@ -260,7 +256,7 @@ fail2:
 fail1:
 	pthread_mutex_destroy(&out->data_mutex);
 fail0:
-	video_output_close(out);
+	bfree(video);
 	return VIDEO_OUTPUT_FAIL;
 }
 
@@ -271,23 +267,20 @@ void video_output_close(video_t *video)
 
 	video_output_stop(video);
 
-	if (video->initialized) {
-		video->initialized = false;
-		pthread_mutex_lock(&video->input_mutex);
+	pthread_mutex_lock(&video->input_mutex);
 
-		for (size_t i = 0; i < video->inputs.num; i++)
-			video_input_free(&video->inputs.array[i]);
-		da_free(video->inputs);
+	for (size_t i = 0; i < video->inputs.num; i++)
+		video_input_free(&video->inputs.array[i]);
+	da_free(video->inputs);
 
-		for (size_t i = 0; i < video->info.cache_size; i++)
-			video_frame_free(
-				(struct video_frame *)&video->cache[i]);
+	for (size_t i = 0; i < video->info.cache_size; i++)
+		video_frame_free(
+			(struct video_frame *)&video->cache[i]);
 
-		pthread_mutex_unlock(&video->input_mutex);
-		os_sem_destroy(video->update_semaphore);
-		pthread_mutex_destroy(&video->data_mutex);
-		pthread_mutex_destroy(&video->input_mutex);
-	}
+	pthread_mutex_unlock(&video->input_mutex);
+	os_sem_destroy(video->update_semaphore);
+	pthread_mutex_destroy(&video->data_mutex);
+	pthread_mutex_destroy(&video->input_mutex);
 
 	bfree(video);
 }
@@ -544,7 +537,7 @@ void video_output_stop(video_t *video)
 	if (!video)
 		return;
 
-	if (video->initialized && !video->stop) {
+	if (!video->stop) {
 		video->stop = true;
 		os_sem_post(video->update_semaphore);
 		pthread_join(video->thread, &thread_ret);
```

jp9000:
Okay, I understand the problem now, just a race condition on shutdown because an output stops in a separately spawned thread. This is a perfectly fine fix to solve it.

In addition, I also applied the code that @norihiro suggested.

PR #8383, opened 2023-03-01
libobs: Fix device functions not marked as EXPORT

PatTheMav:
### Description
Marks function symbols intended to be exported to _actually_ be exported even with visibility set to "hidden".

### Motivation and Context
As clang and gcc use "default" visibility (as in "not hidden") when no visibility is specified, all functions are implicitly exported.

When changing default visibility to "hidden", the corresponding symbols are not exposed anymore and linking fails. This PR updates all functions signatures that were added in recent commits but not set to be EXPORTed.

Also makes required OpenGL imports for texture sharing on macOS mandatory.

### How Has This Been Tested?
Tested on macOS and Linux with clang and gcc.

### Types of changes

### Checklist:

### DIFF
--- a/libobs/graphics/device-exports.h
+++ b/libobs/graphics/device-exports.h
@@ -177,6 +177,15 @@ EXPORT void device_debug_marker_begin(gs_device_t *device,
 				      const char *markername,
 				      const float color[4]);
 EXPORT void device_debug_marker_end(gs_device_t *device);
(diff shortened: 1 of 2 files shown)


tytan652:
Same as @RytoEX.

PatTheMav:
So no objection to merge as-is? The functions are used anyway, but until visibility is set to hidden the issue won't show itself on current master.

PR #8388, opened 2023-03-02
UI: Add apply button to properties dialog

cg2121:
### Description
This duplicates the source that is in the properties, so that updating it doesn't update in the preview. To update it in the preview, the user would click the Apply or Ok button.

### Motivation and Context
Idea from @Warchamp7 

### How Has This Been Tested?
Updated the source properties and made sure it didn't update until hit the Apply button.

### Types of changes

### Checklist:

### DIFF
--- a/UI/forms/OBSBasicProperties.ui
+++ b/UI/forms/OBSBasicProperties.ui
@@ -111,7 +111,7 @@
    <item alignment="Qt::AlignBottom">
     <widget class="QDialogButtonBox" name="buttonBox">
      <property name="standardButtons">
-      <set>QDialogButtonBox::Cancel|QDialogButtonBox::Ok|QDialogButtonBox::RestoreDefaults</set>
+      <set>QDialogButtonBox::Apply|QDialogButtonBox::Cancel|QDialogButtonBox::Ok|QDialogButtonBox::RestoreDefaults</set>
      </property>
     </widget>
    </item>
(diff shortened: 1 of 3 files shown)


tytan652:
Not all sources can be duplicated.
I'm knowledgeable about this, but in this PR, it doesn't look like that case with `OBS_SOURCE_DO_NOT_DUPLICATE` flagged sources are managed.

And so in this case the change is done to the original and logically bypass "Apply" button.

tytan652:
Not all sources can be duplicated.
I'm not knowledgeable about this, but in this PR, it doesn't look like that case with `OBS_SOURCE_DO_NOT_DUPLICATE` flagged sources are managed.

And so in this case the change is done to the original and logically bypass the "Apply" button.

cg2121:
Updated to hide the Apply button if the source can't be duplicated and its properties updates are not deferred, so it acts the same as before with these types of sources.

RytoEX:
> Updated to hide the Apply button if the source can't be duplicated and its properties updates are not deferred, so it acts the same as before with these types of sources.

This feels weird to me, and I expect at least some users will ask "where's the apply button" especially if told while receiving support to "change a setting and click apply".

Please provide screenshots of how this looks.

cg2121:
Unfortunately, @Warchamp7, I don't think this will work how we want it to, as many sources in OBS can't be duplicated. And it would be bad UX if it only worked for some sources and not others.

cg2121:
Unless you think it is acceptable if we only have the apply button for sources that have deferred updates.

PR #8392, opened 2023-03-02
obs-ffmpeg: Fix crash during ratecontrol check

kkartaltepe:
### Description

When iterating over an array of structs we incorrectly check for a null pointer, but the actual sentinal value is the empty struct.

### Motivation and Context

Crash fix when starting recording due to trying to copy the null string from the sentinel.

### How Has This Been Tested?

The crash when starting recording due to copying a null char * no longer occurs.

### Types of changes

### Checklist:


RytoEX (changes requested):
Commit message nit:
sentinal -> sentinel

PR #8393, opened 2023-03-03
obs-vst: Fix incorrect VST window size on HiDPI display

### LINKED ISSUE #8092: VST windows don't behave properly on initial open with non-100% DPI
### Operating System Info

Windows 10

### Other OS

_No response_

### OBS Studio Version

29.0.0

_No response_

### OBS Studio Log URL

https://obsproject.com/logs/NwSSvb3cBRuiiFfP

### OBS Studio Crash Log URL

_No response_

### Expected Behavior

This issue is very similar to https://github.com/obsproject/obs-vst/issues/107

I'm developing a DPI-aware plugin, and am hindered by an issue that seems to have the same root cause.

- When opening the plugin, getRect(ERect** rect) gets called.
- I've got some dimensions to resize my contents to (let's say: my plugin needs 2000x2000 pixels at 100% scaling).
- My main monitor is at 175%, so I know I actually have to resize my contents to 3500x3500.
- I fill the ERect structure with 3500x3500.

Expected behavior is that the host window gets resized to 3500x3500

### Current Behavior

The host window gets resized to 6125x6125 (175% of the size I returned in the ERect structure)

This is aberrant behavior: other VST hosts treat the ERect that I return as the true dimensions of my window.
Only OBS thinks it still needs to compensate for DPI.

### Steps to Reproduce

1. Add a VST audio plugin that only reports its own size and doesn't resize its contents in response to window resizes (e.g. FabFilter)
2. Open plugin interface

_No response_

xlinshan:
### Description
On Windows, the VST plugins' window sizes are rendered larger than the actual content on displays that have UI scale factor. The sizes are larger by the scale factor, for example, 100x100 content will have a 200x200 window on a 200% scaled screen, and 150x150 on a 150% scaled screen. This change adjust the window size to fit the content size. The screenshot is a loaded VST plugin with an extremely big window.
screenshot-1

### Motivation and Context
The problem described in the description above makes VST plugins difficult to use on HiDPI screen. This change makes the window as big as its content.
Issue #8092 talked about this problem. This change should fix it.

### How Has This Been Tested?
I tested the change with VST2 plugins from MeldaProduction's MFreeFXBundle, and Youlean Loudness Meter 2. The MeldaProduction plugins are HiDPI aware plugins that adjust UI size according to OS setting, and the Youlean one is a HiDPI unaware plugin that provides a custom UI scale option. I loaded those plugins to see if they are created correctly, and I resized them, moved them from one display to another, and also changed the OS's UI scale factor while they are opened.

On a single display environment, the plugins are created and resized correctly under different UI scale factor settings. On a dual display environment, the plugins are created incorrectly if the main window of obs is on a sub-display with an UI scale different from the main display, that the window could be bigger or smaller than the content depending on the UI scale difference, but resizing, if applicable, works correctly and fixes the incorrect initial creation, and for HiDPI aware VST plugins, moving the window between the two displays fixes it too and the window adjusts to the different UI scale correctly.

The type of VST plugins that adjust its content to the window size (which is mentioned in #8092) is not tested due to my lack of access.

The test is done on a 2560x1600 laptop connecting to a 3840x2160 display on Windows 11 22h2.

### Types of changes

### Checklist:

Maintainer edit: Fixes #8092

### DIFF
--- a/plugins/obs-vst/win/EditorWidget-win.cpp
+++ b/plugins/obs-vst/win/EditorWidget-win.cpp
@@ -49,10 +49,16 @@ void EditorWidget::buildEffectContainer(AEffect *effect)
 	VstRect *vstRect = nullptr;
 	effect->dispatcher(effect, effEditGetRect, 0, 0, &vstRect, 0);
 	if (vstRect) {
-		widget->resize(vstRect->right - vstRect->left,
-			       vstRect->bottom - vstRect->top);
-		resize(vstRect->right - vstRect->left,
-		       vstRect->bottom - vstRect->top);
+		// on Windows, the size reported by 'effect' is larger than
+		// its actuall size by a factor of the monitor's ui scale,
+		// so the window size should be divided by the factor
+		qreal scale_factor = devicePixelRatioF();
+		int width = vstRect->right - vstRect->left;
+		int height = vstRect->bottom - vstRect->top;
+		width = static_cast<int>(width / scale_factor);
+		height = static_cast<int>(height / scale_factor);
+		widget->resize(width, height);
+		resize(width, height);
 	} else {
 		widget->resize(300, 300);
 	}
@@ -72,6 +78,14 @@ void EditorWidget::handleResizeRequest(int, int)
 	effect->dispatcher(effect, effEditGetRect, 0, 0, &rec, 0);
 
 	if (rec) {
-		resize(rec->right - rec->left, rec->bottom - rec->top);
+		// on Windows, the size reported by 'effect' is larger than
+		// its actuall size by a factor of the monitor's ui scale,
+		// so the window size should be divided by the factor
+		qreal scale_factor = devicePixelRatioF();
+		int width = rec->right - rec->left;
+		int height = rec->bottom - rec->top;
+		width = static_cast<int>(width / scale_factor);
+		height = static_cast<int>(height / scale_factor);
+		resize(width, height);
 	}
 }


WizardCM:
To confirm, this change has only been tested on Windows?

xlinshan:
> To confirm, this change has only been tested on Windows?

Yes, the change has only been tested on Windows. I don't have a MacOS or Linux build environment set up, and I assume that the source file modified is platform specific, so I didn't have the thought. I will try to build it for MacOS and Linux (probably on Debian 11) to see if it has unintentional effects.

xlinshan:
I tried to build it on macOS Monterey on an Intel MacBook and Ubuntu 22.04 on a VMware virtual machine.

On macOS, the modified obs-vst plugin works correctly with MeldaProduction plugins and Bias FX under every screen scale options.

On Ubuntu, obs-vst crashes obs when trying to open the window for a selected VST plugin. The error message on terminal is:
```
info: User selected new VST plugin: '/home/lbx/.lxvst/lsp-plugins/art-delay-mono.so'
qt.qpa.wayland: Wayland does not support QWindow::requestActivate()
X Error of failed request:  BadWindow (invalid Window parameter)
  Major opcode of failed request:  3 (X_GetWindowAttributes)
  Resource id in failed request:  0x7
  Serial number of failed request:  1407
  Current serial number in output stream:  1408
Segmentation fault (core dumped)
```
The obs installed from the ppa source has its VST plugin window opened without problems. However, the version that I built from the master branch of obsproject/obs-studio.git crashed too, with same error message. So I don't know if the crash is because of the change from this commit or is because of my build configuration.

WizardCM:
On Windows, I can confirm current `master` does not handle scaling correctly, and this PR fixes the initial size on the primary display and subsequent resizes triggered by the VST. Tested with BlueCat, Voxengo SPAN, DSEQ3, and FabFilter.

I will do some further experimentation to try and improve behaviour on secondary screens (especially when dragging between displays) and I'll also test on other platforms.

It's worth noting that VST also has a `pluginRespondsToDPIChanges` flag that can be checked using `effCanDo` (which we don't currently use), which may be necessary for some VSTs (though I don't know what's special about it).

PatTheMav:
@WizardCM given that it worked well for you in tests, would you recommend merging this? The change only affects Windows (Linux and macOS use a different source file) anyway.

Tom-Evers:
There's something still not quite right with this. Please check my remark on 542cb876dce6077bb24fa5ee0439e7d21ac53f59

PR #8395, opened 2023-03-03
UI: Add volume controls to properties window

cg2121:
### Description
This makes it easier for users to edit volume controls of a source, without first closing the properties window.

With preview:
Screenshot from 2023-03-06 07-32-51

Without preview:
Screenshot from 2023-03-02 21-06-58

### Motivation and Context
Seems like a good idea. It's a small change, so if others don't like it, that's fine to me.

### How Has This Been Tested?
Open properties of sources with audio.

### Types of changes

### Checklist:

### DIFF
--- a/UI/window-basic-properties.cpp
+++ b/UI/window-basic-properties.cpp
@@ -21,6 +21,7 @@
 #include "qt-wrappers.hpp"
 #include "display-helpers.hpp"
 #include "properties-view.hpp"
+#include "volume-control.hpp"
 
 #include <QCloseEvent>
 #include <QScreen>
@@ -51,6 +52,12 @@ OBSBasicProperties::OBSBasicProperties(QWidget *parent, OBSSource source_)
 			OBSBasicProperties::SourceRemoved, this),
 	  renamedSignal(obs_source_get_signal_handler(source), "rename",
 			OBSBasicProperties::SourceRenamed, this),
+	  audioActivated(obs_source_get_signal_handler(source),
+			 "audio_activate", OBSBasicProperties::AudioActivated,
+			 this),
+	  audioDeactivated(obs_source_get_signal_handler(source),
+			   "audio_deactivate",
+			   OBSBasicProperties::AudioDeactivated, this),
 	  oldSettings(obs_data_create())
 {
 	int cx = (int)config_get_int(App()->GlobalConfig(), "PropertiesWindow",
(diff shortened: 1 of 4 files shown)


WizardCM:
Love this idea. Couple of unhandled edge cases:

If a source _can_ support audio, but doesn't currently, the volume meter is still shown. This includes:

1. Browser source with "Control audio via OBS" **off**
2. Video Capture Device source on Windows with "Audio Output Mode" set to either DirectSound or WaveOut

UI/window-basic-properties.cpp
```diff
@@ -166,6 +168,22 @@ OBSBasicProperties::OBSBasicProperties(QWidget *parent, OBSSource source_)
 
 	} else {
 		ui->preview->hide();
+		verticalMixer = false;
+	}
+
+	if ((caps & OBS_SOURCE_AUDIO) == OBS_SOURCE_AUDIO) {
```
WizardCM, line 174:
```suggestion
	if (obs_source_audio_active(source)) {
```

This'll check whether the source is actually set to play audio. Will require some experimentation to ensurte it works how we expect.

WizardCM (changes requested):

UI/window-basic-properties.cpp
cg2121: Updated to handle sources such as browser and Windows video capture.

WizardCM (approved):

Warchamp7 (changes requested):
Can we get a frame or widget containing the volume widget in the With Preview version, so we can have some spacing between the video preview and the volume bar. It's a bit cramped right now.

How does this behave with private sources like Transitions?

cg2121:
@Warchamp7 updated to have spacing between preview and volume controls. It doesn't work with transitions at this time, so it is hidden for them.

Screenshot from 2023-03-06 07-32-51

Wiethoofd:
Would also not mind the ability to change the Monitor/output status in the properties here, without having to go in to the Advanced Audio Properties dialog.

Personally I would like to see the db/% value input field instead of (or in addition to) the volume monitor, that way I can set it to positive db values (in case a gain filter only isn't enough to boost audio levels on quiet sources).

Would the orientation of the volume slider in this dialog match the setting in the Audio Mixer dock (horizontal/vertical)?

cg2121:
This will be part of a bigger properties redesign, so I will close until that is finished.

PR #8402, opened 2023-03-04
libobs: Simplify balance float comparison

cg2121:
### Description
This now uses the close_float function, instead of comparing floats manually.

### Motivation and Context
Better and cleaner code

### How Has This Been Tested?
Used balance on sources to make sure nothing broke

### Types of changes

### Checklist:


derrod:
Not sure if I'd call this simpler or cleaner.

The current code is two simple comparison, whereas the replacement is a function call to something in the graphics code, using some magic constant (`LARGE_EPSILON`), which in turn calls another function (`fabsf()`). This is adding like three levels of indirection to something that is otherwise a fairly trivial check.

PR #8407, opened 2023-03-04
UI: Sort and pretty-print exported collections

derrod:
### Description

When a scene collection is exported, sort sources by name and pretty-print the output.

This also adds new functions to `obs_data` to get/write JSON in a prettified format.

### Motivation and Context

As of 0dc1e01b5bed35f1514ad32ed4beb5155ca53eb4 the JSON has been saved in minified/"compact" form meaning that it's just one long line.

Opening those files in editors can perform extremely poorly, and tracking changes between them is additionally made difficult due to the order of sources being unreliable.

Some users use git to store and version their scene collection, or work on them collaboratively (e.g. GDQ), and have complained about the instability of the exported collections resulting in large diffs.

### How Has This Been Tested?

Exported some collections, verified they're now pretty-printed and sorted by name.

### Types of changes

### Checklist:

### DIFF
--- a/UI/window-basic-main-scene-collections.cpp
+++ b/UI/window-basic-main-scene-collections.cpp
@@ -397,10 +397,46 @@ void OBSBasic::on_actionExportSceneCollection_triggered()
 	string file = QT_TO_UTF8(exportFile);
 
 	if (!exportFile.isEmpty() && !exportFile.isNull()) {
-		if (QFile::exists(exportFile))
-			QFile::remove(exportFile);
+		QString inputFile = path + currentFile + ".json";
 
-		QFile::copy(path + currentFile + ".json", exportFile);
+		OBSDataAutoRelease collection =
+			obs_data_create_from_json_file(QT_TO_UTF8(inputFile));
+
+		OBSDataArrayAutoRelease sources =
+			obs_data_get_array(collection, "sources");
+		if (!sources) {
+			blog(LOG_WARNING,
+			     "No sources in exported scene collection");
+			return;
+		}
+		obs_data_erase(collection, "sources");
+
+		// We're just using std::sort on a vector to make life easier.
+		vector<OBSData> sourceItems;
+		obs_data_array_enum(
+			sources,
+			[](obs_data_t *data, void *pVec) -> void {
+				auto &sourceItems =
+					*static_cast<vector<OBSData> *>(pVec);
+				sourceItems.push_back(data);
+			},
+			&sourceItems);
+
+		std::sort(sourceItems.begin(), sourceItems.end(),
+			  [](const OBSData &a, const OBSData &b) {
+				  return astrcmpi(obs_data_get_string(a,
+								      "name"),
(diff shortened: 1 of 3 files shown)


gxalpha:
Do we really have to introduce Qt to the mix here?
Jansson has `JSON_SORT_KEYS` , which combined with `JSON_INDENT(4)` should yield a similar result, perhaps implemented in a separate `obs_data_get_json_pretty` function. That way we wouldn't need JSON library no 4, save on code, and give other places the option to save pretty jsons.

derrod:
> Do we really have to introduce Qt to the mix here? Jansson has `JSON_SORT_KEYS` , which combined with `JSON_INDENT(4)` should yield a similar result

It does not. That only sorts the keys, not the items within an array.

> perhaps implemented in a separate `obs_data_get_json_pretty` function. That way we wouldn't need JSON library no 4, save on code, and give other places the option to save pretty jsons.

I've done that as part of some experiments with hashtables (since those have a simple sort function): https://github.com/derrod/obs-studio/compare/f2fa29ab146815d93755de1688265be26396ee4e..c32ddc9c13c55203feca9002644ce8ccd4a163d5

But in this context it would probably be easier to just use `jansson` directly rather than also dealing with `obs_data`, it's just a even more cumbersome than using Qt's stuff.

gxalpha:
Ah, unfortunate.

derrod:
Updated it to use `obs_data` instead, wasn't actually too bad. Qt's JSON also has some weird formatting choices, so this is actually a nicer output.

derrod:
Rebased on master and tested with `Massive.json` which revealed a mistake in ref counting that resulted in a crash to desktop due to attempting to release and already free'd object.
This was fixed by replacing `vector<OBSDataAutoRelease>` with `vector<OBSData>`, since the former does not add a reference.

RytoEX:
Does this only take effect for scene collections that are explicitly exported?

RytoEX (approved):
Seems fine at a glance.

joelsgp:
The JSON_PRESERVE_ORDER flag can be removed, since this is now the default behaivour and the flag has been deprecated.

Anyway, I would appreciate if this was the default behaivour, not just for exports, since I manage my scenes in-place with a dotfiles git repo, and I don't see the benefit of ever compacting it. since on any system modern enough run OBS, a few kilobytes on disk will be completely negligible

joelsgp:
(source: https://jansson.readthedocs.io/en/latest/apiref.html?highlight=dumps#encoding)

data-enabler:
I also version-control my OBS configs, and agree that this isn't super useful if it only affects exports and not the actual config files. Having to import/export constantly seems like more of a hassle to me than just sorting the config file directly, [which is what I currently do](https://github.com/data-enabler/obs-version-control/blob/master/sort_sources.ps1).

Also, if y'all are really interested in having better VCS support I'm happy to open issues with other suggestions (relative file paths would be a big one).

PR #8413, opened 2023-03-06
UI: Fix restore srt stream on youtube HLS

dbbrito:
Description
Hello, our cameras work in SRT mode: srt://192.168.3.200:20000?mode=listener using NVIDIA NVENC H.264 encoder using "Youtube - RTMPs" service.
In this configuration above, after starting the transmission, we can stop and start the SRT stream again without any problems.
But when we use the "Youtube - HLS" service with the NVIDIA NVENC HEVC encoder and we start the transmission, when we stop the SRT stream on the camera we can only restore the stream again if we stop the transmission, after stopping the transmission we start the srt stream on the camera again, the image reappears in the obs and then we have to click on start transmission again, the fact is that this problem only occurs in the "Youtube HLS" service using NVIDIA NVENC HEVC encoder.
Thanks and regards.

Motivation and Context
The problem is that if one of the 5 cameras loses the srt network connection, we will only be able to reconnect it if the transmission stops, thus causing a huge problem.

How Has This Been Tested?
Windows 11
OBS 29.0.2
NVIDIA RTX 3060

Types of changes

Checklist:


pkviet:
There's no original code to fix the bug you describe. Seems like you wanted to create an issue rather ?

PR #8426, opened 2023-03-08
UI: Limit preview scrolling

cg2121:
### Description
This prevents the preview from being scrolled on forever.

### Motivation and Context
Makes preview zoom mode more tolerable.

### How Has This Been Tested?
Moved preview around

### Types of changes

### Checklist:

### DIFF
--- a/UI/window-basic-main.cpp
+++ b/UI/window-basic-main.cpp
@@ -4665,6 +4665,7 @@ void OBSBasic::ResizePreview(uint32_t cx, uint32_t cy)
 	obs_get_video_info(&ovi);
 
 	if (isFixedScaling) {
+		ui->preview->ClampScrollingOffsets();
 		previewScale = ui->preview->GetScalingAmount();
 		GetCenterPosFromFixedScale(
 			int(cx), int(cy),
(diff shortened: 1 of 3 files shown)


Warchamp7:
I'd like to see this factor in the size of the preview area if possible, rather than lock going past the center point like it does now

cg2121:
Updated to limit the offset to the edges of the preview instead of the center, per @Warchamp7 suggestion.

PatTheMav:
Is it intentional that the device pixel ratio isn't taken into account here (because it's unnecessary maybe)?

cg2121:
That is what the GetPixelSize function does.

PatTheMav:
> That is what the GetPixelSize function does.

Got it, might require additional check if it's thread safe (because Qt calls a UI function to get this on macOS and that's only allowed from the main thread). I will try and check this later tonight.

PatTheMav:
Doesn't seem to be called from the graphics thread so far, so should be good.

```
OBSBasic *main = OBSBasic::Get();

QSize targetSize = this->size() * main->GetDevicePixelRatio();
```

Would be the thread-safe variant, if we ever need it.

PR #8445, opened 2023-03-10
UI: Fix Dark and System themes select list height

TianQiBuTian:
### Description

Before:
before
After:
after
### Motivation and Context

Limit the influence range of padding in `QListView` to fix the select list height.
### How Has This Been Tested?

Open a select list.
### Types of changes

### Checklist:

### DIFF
--- a/UI/data/themes/Dark.qss
+++ b/UI/data/themes/Dark.qss
@@ -770,7 +770,7 @@ OBSBasicSettings {
     qproperty-advancedIcon: url(./Dark/settings/advanced.svg);
 }
 
-OBSBasicSettings QListView::item {
+OBSBasicSettings QListWidget::item {
     padding-top: 5px;
     padding-bottom: 5px;
 }
--- a/UI/data/themes/System.qss
+++ b/UI/data/themes/System.qss
@@ -200,7 +200,7 @@ OBSBasicSettings {
     qproperty-advancedIcon: url(:settings/images/settings/advanced.svg);
 }
 
-OBSBasicSettings QListView::item {
+OBSBasicSettings QListWidget::item {
     padding-top: 5px;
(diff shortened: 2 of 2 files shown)


Warchamp7:
The size of QComboBox items is currently smaller on all themes in master. This is likely due to some other change by us, or in Qt itself.

The proposed changes here don't actually have any effect on the size/padding of QComboBox dropdowns as suggested by the screenshots.

TianQiBuTian:
No, in other themes, the option bar has almost no padding.

Also, the select list on the non-setting page is the correct height.


For reference, this is 27.2.4.

UI/data/themes/Dark.qss
```diff
@@ -770,7 +770,7 @@ OBSBasicSettings {
     qproperty-advancedIcon: url(./Dark/settings/advanced.svg);
 }
 
-OBSBasicSettings QListView::item {
+OBSBasicSettings QListView#listWidget::item {
```
Warchamp7, line 773: This should be

```OBSBasicSettings QListWidget::item {}```

to match how the other themes handle this

File: UI/data/themes/System.qss [hunk 2]
```diff
@@ -200,7 +200,7 @@ OBSBasicSettings {
     qproperty-advancedIcon: url(:settings/images/settings/advanced.svg);
 }
 
-OBSBasicSettings QListView::item {
+OBSBasicSettings QListView#listWidget::item {
```
Warchamp7, line 203: This should be

```OBSBasicSettings QListWidget::item {}```

to match how the other themes handle this

Warchamp7 (changes requested):
Apologies, I see now what happened. There's a typo in Dark and System for handling this padding.

I've left notes with the correct fix

PR #8446, opened 2023-03-11
UI: Add Manage Scene Collections dialog

gxalpha:
### Description

Adds a Scene Collection Manager dialog and moves the scene collection menu to the scenes dock.

<p>
image
image
</p>
<p>
image
image
</p>

This solves three major UX problems with current scene collection management:
- The place where you'd manage scene collections was visually disconnected from the scene tree
- All actions were in a context menu, making them a bit cumbersome
- You could only act on the collection you're currently on. You couldn't rename, export, delete, or duplicate any other collection without switching to it first.

The new dialog solves these issues. Scene collection that aren't active can now be renamed, duplicated, exported, or deleted. However, the currently active scene collection can no longer be deleted as that would then require switching to *some* other collection, which can be a bit unpredictable to the user.

It also adds the ability to perform bulk actions on scene collections. This means that it's possible to duplicate, export, or delete multiple scene collections at once.
The collections shown in the menu is now capped at the 10 most recent ones to avoid unreasonably large context menus, other collections can still be switched to by double-clicking them in the dialog.

Driven by a design first approach, based on [this](https://raw.githubusercontent.com/GeorgesStavracas/mockups/main/mockups/scene-collections-dialog/scene-collections-dialog.png) mockup. Thanks to @GeorgesStavracas for creating the mockup and to @Warchamp7 for his assistance.

This PR utilizes new widgets by @derrod. The PR for those is still pending, this PR contains a squashed and slightly modified version of a WIP branch (`UI: Add custom OBS Widgets` and `[TMP] Add selected border style to OBSActionRow`). Until said pending PR has been created, reviewed, and merged, this one will stay draft. However, reviews are still very much desired.

### Motivation and Context

See the problems mentioned above.

### How Has This Been Tested?

macOS 13.2.1

Tested all new features. This includes:
- Opening the importer and importing
- Creating a new collection
- Renaming a collection (both current and not-current)
- Duplicating a collection (both current and not-current)
- Exporting a collection (both current and not-current) and making sure it's still sorted and pretty-printed (see #8407)
- Deleting a not-current collection
- Making sure the "Delete" field is greyed out and that the collection doesn't get deleted even if it isn't (by temporarily changing other code)
- Exporting multiple collections and making sure potential existing collections in the target folder aren't deleted
- Duplicating multiple collections
- Deleting multiple collections
- Switching scene collections via the menu in the scenes dock
- Switching scene collections by double-clicking a scene collection in the dialog
- Sorting works both by last used and alphabetically by name, and is remembered even after the dialog or OBS altogether is closed

### Types of changes

### Checklist:


gxalpha:
Added Seeking Testers and re-triggered CI to get downloadable artifacts for easier testing.

gxalpha:
Rebased on the updated #8447.

New screenshot:
image

With the updated base branch, this should now actually have downloadable build artifacts once CI finishes.

gxalpha:
Rebased on top of updated #8447 and integrated the changes from #9200.

Warchamp7:
Noting down feedback I left on Discord

image

Having the scene collection dropdown in the bottom toolbar was the proposed design but don't feel it works in practice. If it's possible to instead have this as the dock titlebar itself, that would be great.

----------

image

Since we allow the Scenes dock to be hidden, we should probably keep the `Scene Collection` menu in the window top bar for accessibility reasons. Same entries as the dropdown menu.

-----

https://cdn.discordapp.com/attachments/940868766345084928/1130713502407991406/obs64_5lTW7Kg94K.gif

Left border disappears when hovering (Suspected issue with the widgetry itself in #8447)

-----

image

We should add `Open` to the top of this list

-----

image

The padding to the left of the collection names is too large right now (Approximately 80px~), it should be reduced to match the padding on the right of the row (Approximately 28px~)

When the checkboxes are visible, they should have the 28px~ on the left and right

Examples:
image
image

gxalpha:
> Having the scene collection dropdown in the bottom toolbar was the proposed design but don't feel it works in practice. If it's possible to instead have this as the dock titlebar itself, that would be great.
> Since we allow the Scenes dock to be hidden, we should probably keep the Scene Collection menu in the window top bar for accessibility reasons. Same entries as the dropdown menu.

@Warchamp7 Added back the Scene Collection menu to the menu bar.
As noted off-thread, adding the dropdown to the title bar itself involves work much beyond the scope of this PR. For the concern of crowding the toolbar, I removed the name of the scene collection from it. Before completely removing the button and the associated code changes from this PR, I'd like to have an explicit "please remove the button" from you or @GeorgesStavracas if it's still too crowed (as adding it back in at a later stage would be quite a bit of work fighting either Qt or git).

gxalpha:
Adjusted the margins to match better.

<p>
Bildschirmfoto 2023-09-08 um 14 06 12
Bildschirmfoto 2023-09-08 um 14 06 18
</p>

As suspected this involves changes to the widgetry (and I believe is not fixable without them), so I'd definitely want @derrod's opinion on what I'm doing [here](https://github.com/obsproject/obs-studio/pull/8446/commits/890892a159c7dbe04d9fe4a20d997bda123d0f5e)

derrod:
That seems fine to me, I can incorporate those changes in the widgets PR itself.

UI/data/locale/en-US.ini
```diff
@@ -1531,3 +1523,44 @@ YouTube.Errors.rateLimitExceeded="You are sending messages too quickly."
...
+SceneCollections.New.Title="Create Scene Collection"
+SceneCollections.GenericNamePrompt.Text="Please enter the name of the scene collection"
+SceneCollections.Duplicate.Title="Duplicate Scene Collection"
+SceneCollections.Duplicate.Default="%1 (Duplicate)"
+SceneCollections.Rename.Title="Rename Scene Collection"
+SceneCollections.Delete.Title="Delete Scene Collection"
+SceneCollections.Delete.Text="Do you want to delete the scene collection \"%1\"?"
+SceneCollections.Export.Title="Export Scene Collection"
+SceneCollections.BulkDuplicate.Title="Duplicate Scene Collections"
+SceneCollections.BulkDuplicate.Text="Do you want to duplicate selected %1 scene collections?"
```
Warchamp7, line 1549: Are you sure you wish to duplicate %1 scene collections?

UI/data/locale/en-US.ini
```diff
@@ -1531,3 +1523,44 @@ YouTube.Errors.rateLimitExceeded="You are sending messages too quickly."
...
+SceneCollections.BulkDelete.Title="Delete Scene Collections"
+SceneCollections.BulkDelete.Text="Do you want to delete selected %1 scene collections?"
```
Warchamp7, line 1551: Are you sure you wish to delete %1 scene collections?

UI/data/locale/en-US.ini
```diff
@@ -1531,3 +1523,44 @@ YouTube.Errors.rateLimitExceeded="You are sending messages too quickly."
...
+SceneCollections.Sort.LastUsed="Last used"
+SceneCollections.Sort.Name="Name"
+
+SceneCollections.New.Title="Create Scene Collection"
+SceneCollections.GenericNamePrompt.Text="Please enter the name of the scene collection"
+SceneCollections.Duplicate.Title="Duplicate Scene Collection"
+SceneCollections.Duplicate.Default="%1 (Duplicate)"
+SceneCollections.Rename.Title="Rename Scene Collection"
+SceneCollections.Delete.Title="Delete Scene Collection"
+SceneCollections.Delete.Text="Do you want to delete the scene collection \"%1\"?"
```
Warchamp7, line 1546: Are you sure you wish to delete '%1'?

We use this phrasing and single quotes when deleting scenes, sources and filters.

Warchamp7 (changes requested):
- Final verdict on the toolbar button is to remove it
image


- Last Used time doesn't work yet, always says `Just now` for all collections
  - Let's also have the current collection list `Currently Open` for the last used time


- Left border indicator is missing from the currently active collection (Even on the Grey version of Yami now)
image


- When filtering the collection list, there are gaps in the list where the non-matches are
image


- Let's move the multi-select button to the left side of the search bar
image


- In multi-selection mode, there's no indicator of the currently active scene collection
image


- Right now if you select the current collection in multi-select mode, we disable the Delete button and it's not immediately clear as to why. Instead we should leave the Delete button enabled and display an error when they try to Delete with the current collection selected.
```You cannot delete a scene collection while it is open```

**Comment by gxalpha** [comment]
@Warchamp7 addressed part of the feedback. Will investigate the other things soon :tm:. Please see the last question.

> Final verdict on the toolbar button is to remove it

Gone

> Last Used time doesn't work yet, always says Just now for all collections

Will still have to debug this. Works fine on macOS, this is probably operating system related.

> Let's also have the current collection list Currently Open for the last used time

Done

> Left border indicator is missing from the currently active collection (Even on the Grey version of Yami now)

This seems to have gone lost in the Idian PR. Maybe I'll just add it back here.

> When filtering the collection list, there are gaps in the list where the non-matches are

Will investigate why this happens.

> Let's move the multi-select button to the left side of the search bar

Done

> In multi-selection mode, there's no indicator of the currently active scene collection

I think this is also the problem with the left border indicator having gone missing - or are you also looking for the checkmark-thing on the right?

> Right now if you select the current collection in multi-select mode, we disable the Delete button and it's not immediately clear as to why. Instead we should leave the Delete button enabled and display an error when they try to Delete with the current collection selected.

Would it maybe make sense to display such a message somewhere in the window and leave the button disabled? It feels a bit wrong to me to have an enabled button only for that button to be disabled which the user will only know when they click it.

File: UI/data/locale/en-US.ini (hunk 3 above)
Comment (gxalpha) [reply UI/data/locale/en-US.ini]: Done

File: UI/forms/OBSSceneCollections.ui [hunk 4]
```diff
@@ -0,0 +1,185 @@
...
+  </property>
+  <property name="windowTitle">
+   <string>SceneCollections.Title</string>
+  </property>
+  <property name="sizeGripEnabled">
+   <bool>true</bool>
+  </property>
+  <layout class="QVBoxLayout" name="verticalLayout" stretch="1,1,0,1">
+   <item>
+    <layout class="QHBoxLayout" name="horizontalLayout_2" stretch="1,1,1">
```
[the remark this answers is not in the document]
gxalpha: Done

UI/data/locale/en-US.ini
gxalpha: Done

UI/data/locale/en-US.ini
gxalpha: Done

Warchamp7:
> Would it maybe make sense to display such a message somewhere in the window and leave the button disabled? It feels a bit wrong to me to have an enabled button only for that button to be disabled which the user will only know when they click it.

Pulling in off-thread discussion on this below:

If the only action here was Delete, I would agree with disabling the button and displaying a notice. Since there's multiple possible actions in Export/Duplicate/Delete, an error notice makes less sense. If a user is only trying to export a few collections, it doesn't make sense to warn them they can't delete the selection.

Similarly if it was only possible to duplicate one collection at a time, disabling Duplicate upon making two or more selections would be fine, but because Deletion is selection dependent I don't think it's appropriate to take the same approach

If they've ONLY selected the current collection, that is fine to disable the Delete button

As soon as two or more are selected, it gets murky, so I'd rather enable the button and then inform the user their selection is invalid when trying to perform the Delete action

gxalpha:
Rebased on latest Idian branch. There are a few TODOs still left which I will fix later, plus a conflict to current master, but that would require me to rebase the Idian branch as well.

PR #8447, opened 2023-03-11
frontend: Add custom OBS widgets

derrod:
### Description

Adds the custom widgets #8446 is waiting for. These are *very* work in progress and subject to significant changes.

Currently only a subset of them is used in the aforementioned PR, and development will focus on finalising those widgets first. Accessibility will still need some work as does the styling, which is currently only implemented for Yami and Dark.

This PR adds an "idian playground" that can be built by setting `ENABLE_IDIAN_PLAYGROUND` to `ON` in cmake.

Note that the name is a working title (and a pun, "OBS Idian", get it?) and subject to change. I'm open to suggestions!

**ToDo:**
- [x] Accessibility pass, mainly for widgets used in #8446 
- [x] Stylesheet updates to accomodate at least action rows and properties lists in all themes

Demo:
Screenshot 2023-04-08 at 19 00 41

### Motivation and Context

Things could look a little nicer, couldn't they?

### How Has This Been Tested?

Built and fiddled with them.

### Types of changes

### Checklist:


UI/idian/obs-widgets.hpp
```diff
@@ -0,0 +1,19 @@
+#pragma once
+
+/*
+* Idian - A family of custom widgets for OBS implementing the "Yami" UI design.
+*
+* (OBS Idian, get it?)
+*/
+
```
tytan652, line 8:
```suggestion

#define IDIAN_INSIDE

```

UI/idian/obs-widgets.hpp
```diff
@@ -0,0 +1,19 @@
...
+#include "obs-actionrow.hpp"
+#include "obs-controls.hpp"
+#include "obs-groupbox.hpp"
+#include "obs-propertieslist.hpp"
+#include "obs-toggleswitch.hpp"
+
+/// Note: This file serves as an all-in-one include for custom OBS widgets.
+///       It is not intended to define any widgets by itself.
```
tytan652, line 16:
```suggestion
#undef IDIAN_INSIDE
```

UI/idian/obs-toggleswitch.hpp
```diff
@@ -0,0 +1,75 @@
+#pragma once
+
```
tytan652, line 2:
```suggestion
#if !defined(IDIAN_INSIDE)
# error "Only <obs-widgets.h> can be included directly."
#endif
```

UI/idian/obs-propertieslist.hpp
```diff
@@ -0,0 +1,35 @@
+#pragma once
+
```
tytan652, line 2:
```suggestion
#if !defined(IDIAN_INSIDE)
# error "Only <obs-widgets.h> can be included directly."
#endif
```

UI/idian/obs-groupbox.hpp
```diff
@@ -0,0 +1,42 @@
+#pragma once
+
```
tytan652, line 2:
```suggestion
#if !defined(IDIAN_INSIDE)
# error "Only <obs-widgets.h> can be included directly."
#endif
```

UI/idian/obs-controls.hpp
```diff
@@ -0,0 +1,39 @@
+#pragma once
+
```
tytan652, line 2:
```suggestion
#if !defined(IDIAN_INSIDE)
# error "Only <obs-widgets.h> can be included directly."
#endif
```

UI/idian/obs-actionrow.hpp
```diff
@@ -0,0 +1,99 @@
+#pragma once
+
```
tytan652, line 2:
```suggestion
#if !defined(IDIAN_INSIDE)
# error "Only <obs-widgets.h> can be included directly."
#endif
```

tytan652 (commented):
Here is a way to ensure that only obs-widgets.hpp can be included.

UI/idian/obs-propertieslist.hpp
```diff
@@ -0,0 +1,35 @@
...
+	void addRow(OBSActionBaseClass *ar);
+	void clear();
+
+	QList<OBSActionBaseClass *> rows() const { return rowsList; }
+
+private:
+	QVBoxLayout *layout;
+	QList<OBSActionBaseClass *> rowsList;
+};
```
tytan652, line 23: Example:
```suggestion
class OBSPropertiesList : public QFrame {
	Q_OBJECT

	QVBoxLayout *layout;
	QList<OBSActionBaseClass *> rowsList;
public:
	OBSPropertiesList(QWidget *parent = nullptr);

	void addRow(OBSActionBaseClass *ar);
	void clear();

	QList<OBSActionBaseClass *> rows() const { return rowsList; }
};
```

tytan652 (commented):
Why adding a `private:` at the end where it is already private by default at the start (before any other keyword).

UI/idian/obs-propertieslist.hpp
derrod: OBS is pretty inconsistent with it, but it's a stylistic choice that you'll also find in Qt to have an explicit `private:` section and order declarations in ascending order of privacy (i.e. public -> protected -> private).

derrod:
> Here is a way to ensure that only obs-widgets.hpp can be included.

The intention is that singular widgets can be included where other are not required, so I don't think that is necessary. The `obs-widgets.hpp` files exists as a shortcut for UI where we may want to use most if not all of the widgets, e.g. if we rework the properties or settings UI.

derrod:
I'm open to suggestions for the name, I just didn't like a generic `obs-widgets`, and I wasn't sure about just going with `yami` since that's the theme/visual identity, and while those widgets are a part of the development of that, they will also have to work with other styles (And I like to give my own projects little fun names rather than using something existing, see also: [bouf](https://github.com/obsproject/bouf)).

gxalpha:
This may have been noted before (but in that case I can’t find where), "zoo" feels like an unusual term for a sample / test / demonstration? I think "Playground" or "Sandbox" would be more common maybe?

derrod:
"playground" perhaps?

gxalpha:
> "playground" perhaps?

That would make sense to me.

derrod:
Updated to rename it to "Playground"

GeorgesStavracas:
To do:

 - [ ] Implement the updated arrow style in combo boxes in action rows

gxalpha:
Rebasing #8446 I noticed a couple of small things:
1) The styling for the border color of selected rows appears to have gotten lost (example for Grey):
```
OBSActionRow[rowSelected="true"] {
    border-color: rgb(40,76,184);
}

OBSActionRow[rowSelected="true"]:hover {
    border-color: rgb(54,92,192);
}
```
2) Hover colors on actionrows seem wrong in Grey and Acri. Grey has what appears to be the Yami shade, and Acri has a color that might be the one intended for Grey.
3) The `first` and `last` effect seems to have gone kaputt on hover. In the Manage Scene Collection dialog, but also for some items in the Playground (e.g. "Toggle Switch"), Rachni, Acri, and Yami show them as being the last (rounding off the bottom corners when hovered over); while Grey and Light never round off the corners on hover (even for the first and last one).

Also commit 3 (`UI: Make comboboxes in action rows flat` (which probably should be `UI/idian: (...)` :P)) seems a bit weird as it only contains Yami while the changes to other themes are already in a previous commit, but maybe that's just nitpicking at this point.

derrod:
1. Restored it for themes that had it, it wasn't used in this PR and I didn't realise it was used by urs
2. Yeah colours and styling should be considered not final, I assume Warchamp will want to go over everything before we ship it
3. Mh, gotta look into that, I'm also thinking about maybe doing the whole rounded corner thing differently using a custom paint event, but not sure if that'll work out yet.

Warchamp7:
I've rebased this PR with a few additional bits of cleanup. The commits have been squashed in preparation for merge but mainly to make rebasing easier.

The branch prior to rebasing is archived at
https://github.com/Warchamp7/obs-studio/tree/idian-backup

Warchamp7:
I've rebased this again. Here is also the current appearance of the widgets, as they changed quite a while ago from what's in the original description.

image

shared/qt/idian/widgets/Row.hpp
```diff
@@ -0,0 +1,211 @@
...
+#include <QScrollArea>
+#include <QMouseEvent>
+#include <QCheckBox>
+
+#include "../components/ToggleSwitch.hpp"
+#include "../components/ComboBox.hpp"
+#include "../components/SpinBox.hpp"
+#include "../components/DoubleSpinBox.hpp"
+#include "../Utils.hpp"
+#include "PropertiesList.hpp"
```
sebastian-s-beckmann: Includes still need to be changed to not be relative (in many files, not just this one). Marking this for myself.

sebastian-s-beckmann:
I've made some changes based on off-thread discussion on naming. Classes were changed from prefix naming (`OBS*`) to being in a namespace (`idian::`, not `obs::` to make it clearer that these do not depend on OBS). Some names have also been made more clear as to what they are (e.g. `ActionRow` -> `GenericRow`, `OBSActionRowWidget` -> `Row`, etc).
Reviews are appreciated.

I also rebased this on master to fix the conflicts that have occurred due to changes to the Yami theme files.

Although I don't believe it will be necessary, I've retained the previous version of this PR [here](https://github.com/gxalpha/obs-studio/tree/obs-idian-backup).

shared/qt/idian/widgets/Row.hpp
sebastian-s-beckmann: Changed this. (again, [previous version retained](https://github.com/gxalpha/obs-studio/tree/idian-rebase-backup-before-include-changes))

b1bradders:
Sorry for this useless comment; I couldn't help but notice the branch name! 😂 `obs-idian`!

b1bradders:
> I've rebased this again. Here is also the current appearance of the widgets, as they changed quite a while ago from what's in the original description.
> 
> image

Oof - I prefer the first screenshot's design imo. Here, it feels too cramped. Kind of looks like you sucked the GNOME-like look out of the original design, although maybe that was intentional? 🤔

frontend/dialogs/OBSIdianPlayground.cpp
```diff
@@ -0,0 +1,142 @@
...
+    along with this program.  If not, see <http://www.gnu.org/licenses/>.
+******************************************************************************/
+
+#include <QTimer>
+
+#include <Idian/Idian.hpp>
+#include "OBSIdianPlayground.hpp"
+
+// This file is a temporary playground for building new widgets,
+// it may be removed in the future.
```
PatTheMav: Remove comment - I think it's absolutely fine to ship the playground as a developer feature and maybe there's even value for Ui development past that.

frontend/forms/OBSBasic.ui
```diff
@@ -2165,9 +2166,9 @@
     <string>PasteDuplicate</string>
    </property>
   </action>
-  <action name="autoConfigure2">
```
PatTheMav, line 2168: Was the `autoConfigure2` action dormant?

shared/qt/idian/components/ComboBox.cpp
```diff
@@ -0,0 +1,64 @@
...
+void ComboBox::showPopup()
+{
+	if (allowOpeningPopup) {
+		QComboBox::showPopup();
+	}
+}
+
+void ComboBox::hidePopup()
+{
+	/* ToDo: Find a better way to do this
```
PatTheMav: `TODO` is usually automatically detected by IDEs and thus this line is potentially detected and highlighted in some way. In that case the actual to-do summary should be more expressive about what should be done.

Also is this a to-do to finish up Idian or will this stay as-is for the foreseeable future?

shared/qt/idian/include/Idian/Group.hpp
```diff
@@ -0,0 +1,75 @@
...
+	Q_OBJECT
+
+public:
+	Group(QWidget *parent = nullptr);
+
+	PropertiesList *properties() const { return propertyList; }
+
+	// ToDo add event for checkable group being enabled/disabled
+	// ToDo allow setting enabled state
+	// (Maybe) ToDo add option for hiding properties list when disabled
```
PatTheMav: See above regarding `TODO` blocks, also remove if this is not a necessary to-do, but actually just ideas for future features.

frontend/dialogs/OBSIdianPlayground.cpp
```diff
@@ -0,0 +1,142 @@
...
+    GNU General Public License for more details.
+
+    You should have received a copy of the GNU General Public License
+    along with this program.  If not, see <http://www.gnu.org/licenses/>.
+******************************************************************************/
+
+#include <QTimer>
+
+#include <Idian/Idian.hpp>
+#include "OBSIdianPlayground.hpp"
```
PatTheMav: Apply the include order introduced with the frontend framework:

```C++
#include "OBSIdianPlayground.hpp" // Counterpart
#include "SomeOtherFileInTheSameDirectory.hpp"
#include <Some/OtherFileInSameTargetButDifferentDirectory.hpp>

#include <Idian/Idian.hpp>
#include <Some/OtherFileFromAnotherTarget.hpp>

#include <3rdPartyLibraryHeader.hpp>

#include <StandardLibraryHeader.hpp>
```

This should be applied to all files (I've seen more instances of the order not being followed below).

shared/qt/idian/widgets/PropertiesList.cpp
```diff
@@ -0,0 +1,83 @@
...
+	setSizePolicy(QSizePolicy::MinimumExpanding, QSizePolicy::Minimum);
+
+	rowsList = QList<GenericRow *>();
+
+	setLayout(layout);
+}
+
+/* Note: This function takes ownership of the added widget
+ * and it may be deleted when the properties list is destroyed
+ * or the clear() method is called! */
```
PatTheMav: I'd prefer if C++ comment style is used here. Functionally this seems to be in line with Qt's ownership model (parents always own their children and thus parents take care of destroying them as well).

shared/qt/idian/widgets/Row.cpp
```diff
@@ -0,0 +1,371 @@
...
+	labelLayout->addWidget(descriptionLabel);
+
+	layout->addLayout(labelLayout, 0, 1, Qt::AlignLeft);
+}
+
+void Row::setPrefix(QWidget *w, bool auto_connect)
+{
+	setSuffixEnabled(false);
+
+	_prefix = w;
```
PatTheMav: Don't use underscore prefixes in C++, those might clobber reserved internal names used by the platform library.

shared/qt/idian/widgets/Row.cpp
```diff
@@ -0,0 +1,371 @@
...
+}
+
+void Row::connectBuddyWidget(QWidget *widget)
+{
+	setAccessibleName(nameLabel->text());
+	setFocusProxy(widget);
+	setBuddy(widget);
+
+	/* If element is a ToggleSwitch and checkable, forward
+	 * clicks to the widget */
```
PatTheMav: Use C++ comment style, apply throughout the PR.

PatTheMav (changes requested):

frontend/forms/OBSBasic.ui
sebastian-s-beckmann: Yes. However, this was either a faulty rebase, an automatic Qt Creator change on save, or a drive-by by someone (that I at least can't remember doing). I'll undo the change and PR this separately.

RytoEX: If you check the blame, `autoConfigure2` dates back to the introduction of the Auto-Configuration Wizard.

PatTheMav: Got it.

shared/qt/idian/components/ComboBox.cpp
sebastian-s-beckmann: I'll replace the `ToDo:` with something else. This is a Rodney comment but I didn't have a better solution either. As far as I'm concerned, this is a tad-bit hacky but works correctly and imo is fine to merge as-is.

Edit: Replaced the `ToDo`.

frontend/dialogs/OBSIdianPlayground.cpp
sebastian-s-beckmann: Removed.

sebastian-s-beckmann:
I think all the comments from @PatTheMav above should be addressed ([complete diff without rebase on master](https://github.com/obsproject/obs-studio/compare/b04da7cc9bbabaa4cb30fa195f6f79b65ba35419..0f862475e088a1a47ceca6291f091072a42bf3cd)).


Also for reference, here are some screenshots with the different density levels (from the new OBS 31.1 appearance options):


"Classic" (smallest / minimum padding):
image

"Compact" (small / between Classic and Normal):
image

"Normal" (default):
image

"Comfortable" (largest / most padding):
image

frontend/dialogs/OBSIdianPlayground.hpp
```diff
@@ -0,0 +1,34 @@
...
+    MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+    GNU General Public License for more details.
+
+    You should have received a copy of the GNU General Public License
+    along with this program.  If not, see <http://www.gnu.org/licenses/>.
+******************************************************************************/
+
+#pragma once
+
+#include <ui_OBSIdianPlayground.h>
```
PatTheMav: In my experience those UI headers play by their own rules - according to our rules they should be placed up here, but I've encountered quite a few situations where I _had_ to place it at the very bottom of includes.

YMMV, but just wanted to point this out that it wouldn't be "wrong" to put this at the bottom of the includes for that reason.

shared/qt/idian/components/CheckBox.cpp
```diff
@@ -0,0 +1,24 @@
...
+    along with this program.  If not, see <http://www.gnu.org/licenses/>.
+******************************************************************************/
+
+#include <Idian/CheckBox.hpp>
+
+using idian::CheckBox;
+
+CheckBox::CheckBox(QWidget *parent) : QCheckBox(parent), Utils(this) {}
+
+#include <Idian/moc_CheckBox.cpp>
```
PatTheMav: Why is this include at the bottom of the file? Does the namespace use require it?

frontend/dialogs/OBSIdianPlayground.hpp
sebastian-s-beckmann: I hadn't encountered any issues thus far, but moved it to the bottom just to be sure.

shared/qt/idian/components/CheckBox.cpp
sebastian-s-beckmann: Having the `moc_XYZ.cpp` includes at the very bottom follows what Qt does in their own code. See e.g. https://github.com/qt/qtbase/blob/ba29d7bf99ac917b52b5d947802025751b5c6e6a/src/widgets/widgets/qabstractspinbox.cpp#L2208
It's not required, it could be moved to the top.

PatTheMav: IIRC that's because they use an automatic script to add those includes after `QT_END_NAMESPACE`:

https://github.com/qt/qtbase/blob/dev/util/includemocs/includemocs6.sh

But for our own frontend code the generated MOC header is usually included as the very last thing in the include section. That has proven to be the most robust place during the frontend reorganisation.

sebastian-s-beckmann: I've moved them up.

RytoEX (approved):

PR #8458, opened 2023-03-12
flatpak: Enable obs-libfdk

tytan652:
### Description

Enable obs-libfdk since Freedesktop runtime provides it.

### Motivation and Context

Enable an available feature and do some fixup.

### How Has This Been Tested?

Builds fine and obs-libfdk is correctly loaded.

### Types of changes

### Checklist:

### DIFF
--- a/CI/flatpak/com.obsproject.Studio.json
+++ b/CI/flatpak/com.obsproject.Studio.json
@@ -505,7 +505,8 @@
         "-DENABLE_JACK=ON",
         "-DENABLE_RTMPS=ON",
         "-DENABLE_VLC=OFF",
-        "-DENABLE_AJA=ON"
+        "-DENABLE_AJA=ON",
+        "-DENABLE_LIBFDK=ON"
       ],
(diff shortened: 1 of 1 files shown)


derrod:
Do you know if the libfak version included in the runtime is the stripped-down Fedora one?

They removed some features that their lawyers believed still were subject to patents (though they have since expired iirc) and I'm not sure if that gutted version actually still is better than frmpeg aac.

tytan652:
Here is Freedesktop fork branch in use: https://cgit.freedesktop.org/~wtay/fdk-aac/log/?h=stripped4

> They removed some features that their lawyers believed still were subject to patents (though they have since expired iirc) and I'm not sure if that gutted version actually still is better than frmpeg aac.

I don't know either.

tytan652:
Moved CMake change to another PR (#8461).

derrod:
Since we now allow users to select which AAC encoder they want to use this is probably fine to merge to get some feedback during the beta.

PR #8463, opened 2023-03-13
Inotify image and text

based-a-tron:
### Description
Added a core structure for storing file watch info.
Files can be watched with inotify on linux systems or polled, pending support for kqueues and whatever windows uses.
Files removed from the filesystem (that cannot be immediately rewatched) fall back to polling, and restore themselves to the OS specific notification mechanism when they are restored.
File watches are identified by a type and a long watch id.
Added some additional utilities for file locking to prevent races, for image sources.
Allows users to manually fall back to polling by selecting "network mode" (for files that are over NFS or likewise and don't receive OS level notification events).

### Motivation and Context
File polling is unreliable for files that update more often than once a second.
This commit was cleaned up for merge, but I've been using the base system for around 2 years on my own machine.

### How Has This Been Tested?
Only the linux version with the following build options has been tested, and even then not thoroughly. I would appreciate it if someone else would try compiling this and testing it as well.
```
-DENABLE_JACK=ON -DENABLE_RTMPS=ON -DENABLE_FREETYPE=ON -DENABLE_BROWSER=OFF -DENABLE_PULSEAUDIO=OFF -DENABLE_WAYLAND=OFF -DENABLE_WEBSOCKET=OFF -DENABLE_VLC=OFF -DENABLE_AJA=OFF -DENABLE_DECKLINK=OFF -DUNIX_STRUCTURE=1 -DENABLE_V4L2=OFF -DENABLE_PIPEWIRE=OFF -DENABLE_NEW_MPEGTS_OUTPUT=OFF -DENABLE_LIBDFK=OFF -DENABLE_SPEEXDSP=OFF -DENABLE_VST=ON -DQT_VERSION=6
```
I know of one possible "bug" but I couldn't decide if I wanted to patch it or not.
When a source is added and a file is not present, it will fail to watch the file and return an invalid watch. Maybe it should fall back to polling instead? It wouldn't hurt anything, probably. Use your best judgement.
Unfortunately, it doesn't look like OBS offloads image loading to another thread while processing a frame. So for large, compressed files that update extremely frequently, the framerate slows to a crawl while they process. I actually don't use this for anything but files in shared memory. It's better than nothing.

### Types of changes
Changes the text-freetype source and the image-source source to facilitate generic file watching.
I have documented nothing. I have tested nothing. It all Works on My Machine (tm).

### Checklist:

- [ x ] I have read the [**contributing** document](https://github.com/obsproject/obs-studio/blob/master/CONTRIBUTING.rst).


PatTheMav:
This PR seems incomplete (there are no implementations for Windows, macOS, or FreeBSD), the Linux code is by its own admission "not thoroughly tested" and the code itself seems like a work in progress and fails formatting checks.

As there has been no progress on this PR since March, this PR is closed.

If support for other platforms has been implemented and the code made final, it can be reopened on request.

Issue #4895, opened 2021-06-13
Cursor doesn't properly disappear with Pipewire capture


rmnvgr:
### Operating System Info

Other

### Other OS

Fedora 34

### OBS Studio Version

27.0.1

### OBS Studio Log URL

https://obsproject.com/logs/AUrBGETVVGn7lrbQ

### Expected Behavior

When the cursor disappears in the captured window/screen, it also disappears in the output.

### Current Behavior

The cursor is still visible in the output.

### Steps to Reproduce

1. Open an application where the cursor disappears (for example OpenTTD or Minecraft)
2. Add a Pipewire capture (either window or screen)
3. See that cursor still appears in the output

### Anything else we should know?

In OpenTTD (cursor in the bottom right):
![openttd](https://user-images.githubusercontent.com/66537460/121820380-ba6e1480-cc92-11eb-8926-26310b6fa886.png)

In Minecraft (cursor at the top left of the pointer):
![minecraft](https://user-images.githubusercontent.com/66537460/121820389-c35ee600-cc92-11eb-9edc-4eea80656d70.png)

RytoEX:
Just to confirm, did you uncheck "Show Cursor" on the source properties in OBS?

![Screenshot from 2021-06-13 19-06-42](https://user-images.githubusercontent.com/624931/121824575-1cbb1b00-cc7b-11eb-8569-e2c47ebaf2da.png)

rmnvgr:
No, I did not uncheck "Show Cursor". That would do for OpenTTD, which provides its own cursor, but not for Minecraft, which has to display the cursor in menus/inventory.

GeorgesStavracas:
I've been trying to track this down for months now. It is almost certainly a Mutter issue. For some reason, it keeps sending cursor data even when the cursor should be invisible. It seems not even https://gitlab.gnome.org/GNOME/mutter/-/merge_requests/1417 helped this case...

RytoEX:
Is Minecraft perhaps switching from a custom cursor to the system cursor for things like the inventory screen?  If so, I'm not sure how that would be handled, at least on our end.

rmnvgr:
@GeorgesStavracas Is there a Mutter issue I can subscribe to already? Or do you want me to open one?

> Is Minecraft perhaps switching from a custom cursor to the system cursor for things like the menus and inventory screen? If so, I'm not sure how that would be handled, at least on our end.

[...]

PR #8467, opened 2023-03-14
linux-pipewire: Clear cursor texture on empty bitmap

rmader:
If we receive an empty cursor bitmap - one without valid size - we should hide the cursor. Do so by clearing the texture.

This fixes visible cursors when recording various games with Wayland compositors.

Closes https://github.com/obsproject/obs-studio/issues/4895

### How Has This Been Tested?
Locally tested, asking other people to test as well.

### Types of changes

### Checklist:

### DIFF
--- a/plugins/linux-pipewire/pipewire.c
+++ b/plugins/linux-pipewire/pipewire.c
@@ -719,6 +719,10 @@ read_metadata:
 			bitmap = SPA_MEMBER(cursor, cursor->bitmap_offset,
 					    struct spa_meta_bitmap);
 
+		if (bitmap)
+			g_clear_pointer(&obs_pw->cursor.texture,
+					gs_texture_destroy);
+
 		if (bitmap && bitmap->size.width > 0 &&
 		    bitmap->size.height > 0 &&
 		    lookup_format_info_from_spa_format(
@@ -732,8 +736,7 @@ read_metadata:
 			obs_pw->cursor.width = bitmap->size.width;
 			obs_pw->cursor.height = bitmap->size.height;
 
-			g_clear_pointer(&obs_pw->cursor.texture,
-					gs_texture_destroy);
+			assert(obs_pw->cursor.texture == NULL);
 			obs_pw->cursor.texture = gs_texture_create(
 				obs_pw->cursor.width, obs_pw->cursor.height,
 				gs_format, 1, &bitmap_data, GS_DYNAMIC);


plugins/enc-amf
tytan652, line 1: Please remove this unintentional changes.

tytan652 (changes requested):

plugins/enc-amf
rmader: Oops, done!

GeorgesStavracas (approved):
Would you mind adding the `Closes https://github.com/obsproject/obs-studio/issues/4895` to the commit message too?

The code LGTM. Thank you very much. This bug has been pestering OBS for 2 years.

rmader:
Should we mark this for backporting somehow? Or is that done automatically?

plugins/linux-pipewire/pipewire.c
```diff
@@ -719,27 +719,33 @@ static void on_process_cb(void *user_data)
...
-			bitmap_data =
-				SPA_MEMBER(bitmap, bitmap->offset, uint8_t);
-			obs_pw->cursor.hotspot_x = cursor->hotspot.x;
-			obs_pw->cursor.hotspot_y = cursor->hotspot.y;
-			obs_pw->cursor.width = bitmap->size.width;
-			obs_pw->cursor.height = bitmap->size.height;
-
+		if (bitmap) {
 			g_clear_pointer(&obs_pw->cursor.texture,
 					gs_texture_destroy);
```
GeorgesStavracas, line 724: Actually, please let me suggest a stylistic change. Looking at the file itself, instead of the diff, seems like there's too much nesting happening. So I think I'd prefer if this was written as:

```c
if (bitmap)
	g_clear_pointer(&obs_pw->cursor.texture, gs_texture_destroy);

if (bitmap && width > 0 && height > 0 && [...]) {
	[all of the current code]

	assert(obs_pw->cursor.texture == NULL);
	obs_pw->cursor.texture = ...;
}
```

What do you think?

GeorgesStavracas (changes requested):

GeorgesStavracas:
> Should we mark this for backporting somehow? Or is that done automatically?

We'll think that separatedly, don't worry

plugins/linux-pipewire/pipewire.c
rmader: Sure, I don't mind. Certainly makes the diff more readable, done.

plugins/linux-pipewire/pipewire.c
```diff
@@ -732,8 +736,6 @@ static void on_process_cb(void *user_data)
 			obs_pw->cursor.width = bitmap->size.width;
 			obs_pw->cursor.height = bitmap->size.height;
 
-			g_clear_pointer(&obs_pw->cursor.texture,
-					gs_texture_destroy);
```
GeorgesStavracas, line 736: Can you please add a `assert(obs_pw->cursor.texture == NULL)` here? :)

rmader: Done!

GeorgesStavracas (approved):

PR #8473, opened 2023-03-15
Add old Vtuber Maker versions to DLL blocklist

notr1ch:
### Description
Blocks old versions of the Vtuber Maker `lib64_new2.dll` which don't have the Unity Capture deadlock fix implemented.

### Motivation and Context
Freezing OBS is bad.

### How Has This Been Tested?
Simulated using the NVIDIA Broadcast filter as I don't wish to install Vtuber Maker. Seemed to work.

### Types of changes

### Checklist:


notr1ch:
Note that Vtuber Maker also ship filters `lib64.dll` and `lib64_new.dll` but these don't seem used in the current release. I'm awaiting feedback from the developers as to what these are for, it might be safe to also block these older ones as they sound like they belong to older versions of the software.

PR #8487, opened 2023-03-18
libobs/util: simplify implementation of os_get_path_extension

norihiro:
### Description

This PR includes these changes.
- libobs/util: Simplify implementation of `os_get_path_extension`
  In current implementation, it allocate memory, replace `\` with `/`, and calls two `strrchr`. This PR will just iterate two times without allocating the memory.
- docs: Clarify a dot is included in the extension
- test: Add a test for `os_get_path_extension`

Also considered other implementations below but the code in PR would be the best balance of readability and simplicity.

This code has just one interaction but I feel it is worse to read.
```c
const char *os_get_path_extension(const char *path)
{
	const char *ext = NULL;

	for (; *path; path++) {
		switch (*path) {
		case '.':
			ext = path;
			break;
		case '/':
		case '\\':
			ext = NULL;
		}
	}

	return ext;
}
```

This is similar to the original code in terms of using `strrchr`. This has 3 iteration and I feel the ease of reading is not so different from the PR.
```c
const char *os_get_path_extension(const char *path)
{
	if (!path[0])
		return NULL;

	char *slash = strrchr(path, '/');
	char *bslash = strrchr(path, '\\');
	char *period = strrchr(path, '.');

	if (!period || slash > period || bslash > period)
		return NULL;

	return period;
}
```

### Motivation and Context

I'm enabling GCC's `-Wuse-after-free` option to detect use-after-free of `bfree` and `obs_*_release` in [this branch](https://github.com/norihiro/obs-studio/commits/bfree-warning).

I want to avoid memory allocation in that function because it can be replaced by a simple implementation.

When rewriting the function, I noticed an ambiguity so that I also fixed documentation.

### How Has This Been Tested?

Some test cases are added and checked the test has passed for both old and new code.

### Types of changes

### Checklist:

### DIFF
--- a/libobs/util/platform.c
+++ b/libobs/util/platform.c
@@ -654,28 +654,17 @@ int os_mkdirs(const char *dir)
 
 const char *os_get_path_extension(const char *path)
 {
-	struct dstr temp;
-	size_t pos = 0;
-	char *period;
-	char *slash;
+	for (size_t pos = strlen(path); pos > 0; pos--) {
+		switch (path[pos - 1]) {
+		case '.':
+			return path + pos - 1;
+		case '/':
+		case '\\':
+			return NULL;
+		}
+	}
 
-	if (!path[0])
-		return NULL;
-
-	dstr_init_copy(&temp, path);
-	dstr_replace(&temp, "\\", "/");
-
-	slash = strrchr(temp.array, '/');
-	period = strrchr(temp.array, '.');
-	if (period)
-		pos = (size_t)(period - temp.array);
-
-	dstr_free(&temp);
-
-	if (!period || slash > period)
-		return NULL;
-
-	return path + pos;
+	return NULL;
 }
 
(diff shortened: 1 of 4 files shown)


libobs/util/platform.c
```diff
@@ -654,28 +654,17 @@ int os_mkdirs(const char *dir)
...
-	dstr_free(&temp);
-
-	if (!period || slash > period)
-		return NULL;
+	for (size_t pos = strlen(path); pos > 0; pos--) {
+		switch (path[pos - 1]) {
+		case '.':
+			return path + pos - 1;
+		case '/':
+		case '\\':
```
norihiro, line 662: For non-Windows system, I want to remove `\\`. However, removing `\\` is a breaking change and also require other codes to be changed as well. So, I don't change it for now.

derrod (approved):
Seems like a nice little optimisation that also makes it much easier to read.

PR #8495, opened 2023-03-20
libobs: Keep mixer data for placeholder sources

derrod:
### Description

Placeholder sources are created when a plugin providing a source type is unavailable. Since they're not marked as having audio any audio track settings for them get lost. With this change we retain mixer information if the source owns its id to avoid that.

(This commit was pulled from #8455)

### Motivation and Context

Lost most audio on a 4+ hour D&D session, probably because I ran OBS without the `win-capture-audio` plugin once which lost all of the track assignments.

### How Has This Been Tested?

Loaded up OBS without `win-wasapi`, sources still had assignment after relaunching with it.

### Types of changes

### Checklist:


RytoEX (approved):
Approved off-thread by @norihiro and @notr1ch .

PR #8497, opened 2023-03-20
UI: Add unassigned indicator and warning to mixer

derrod:
### Description

Adds a new icon to the volume meter indicating that a source is unmuted, but inaudible due to not being assigned to any tracks:
2023-03-24_15-00-06_Qd4YJB

When attempting to unmute a source in this state, the following message is shown:
2023-03-24_15-00-08_st9Zlg

### Motivation and Context

Had no audio in a 4+ hour D&D session recording because the track assignment got lost (possibly fixed by #8495), probably a good idea to warn users about such cases.

### How Has This Been Tested?

Unassigned a soure, tried to start output.

### Types of changes

### Checklist:

### DIFF
--- a/UI/data/locale/en-US.ini
+++ b/UI/data/locale/en-US.ini
@@ -559,6 +559,8 @@ VolControl.SliderUnmuted="Volume slider for '%1':"
 VolControl.SliderMuted="Volume slider for '%1': (currently muted)"
 VolControl.Mute="Mute '%1'"
 VolControl.Properties="Properties for '%1'"
+VolControl.UnassignedWarning.Title="Unassigned Audio Source"
+VolControl.UnassignedWarning.Text="\"%1\" is not assigned to any audio tracks and it will not be audible in streams or recordings.\n\nTo assign an audio source to a track, open the Advanced Audio Properties via the right-click menu or the cog button in the mixer dock toolbar."
 
 # add scene dialog
 Basic.Main.AddSceneDlg.Title="Add Scene"
--- a/UI/data/themes/Acri.qss
+++ b/UI/data/themes/Acri.qss
@@ -986,6 +986,10 @@ MuteCheckBox::indicator:checked {
     image: url(./Dark/mute.svg);
 }
 
+MuteCheckBox::indicator:indeterminate {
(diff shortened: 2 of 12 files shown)


Warchamp7:
Discussed on Discord. Proposal is to change the unmuted icon whenever a source is disabled on all audio tracks.

Non-final mockup idea
image

This indicates to the user the source is unmuted, but with a warning. A user will be inclined to click the icon if they're trying to confirm the state. Potentially with a tooltip.

Clicking the icon will then mute as normal, if the user wants that source heard, they will click it again. This will return it to the yellow warning icon state, and at that stage we can pop up this warning. We'll likely want a 'Don't show again' for that as well.

Warchamp7:
Looks good 👍

Warchamp7 (approved):

UI/data/locale/en-US.ini
```diff
@@ -559,6 +559,8 @@ VolControl.SliderUnmuted="Volume slider for '%1':"
 VolControl.SliderMuted="Volume slider for '%1': (currently muted)"
 VolControl.Mute="Mute '%1'"
 VolControl.Properties="Properties for '%1'"
+VolControl.UnassignedWarning.Title="Unassigned Audio Source"
+VolControl.UnassignedWarning.Text="\"%1\" is not assigned to any audio tracks and it will not be audible in streams or recordings.\n\nTo assign an audio source to a track, open the advanced audio properties via the right-click menu or the cog button in the mixer dock toolbar."
```
RytoEX, line 563:
```suggestion
VolControl.UnassignedWarning.Text="\"%1\" is not assigned to any audio tracks and it will not be audible in streams or recordings.\n\nTo assign an audio source to a track, open the Advanced Audio Properties via the right-click menu or the cog button in the mixer dock toolbar."
```

RytoEX (changes requested):
Looks fine at a glance except for one locale string nit.

UI/data/locale/en-US.ini
derrod: Updated.

RytoEX (approved):
Nit addressed. Looks good to me.

PR #8289, opened 2023-02-18
libobs,obs-ffmpeg: Add option for recording lossless audio

derrod:
### Description

Adds support for manually selecting the recording audio encoder in the UI.

Currently adds 16, 24, and 32 bit PCM as well as 24-bit ALAC and 16-bit FLAC (their respective default bit-depths in FFmpeg's order of formats).

Screenshot (note: out of date, minor changes with RFC 45, but overall still the same UX):
2023-02-18_06-19-28_xvjboa

#8498 adds the necessary UI to prevent users from selecting invalid container and audio codec combinations.

### Motivation and Context

Lossless audio recording is a highly requested feature (e.g. ideas page posts [271](https://ideas.obsproject.com/posts/271/add-flac-as-selectable-audio-codec-for-recordings), [1342](https://ideas.obsproject.com/posts/1342/record-lossless-audio), [1680](https://ideas.obsproject.com/posts/1680/allow-audio-codec-choice-on-advanced-type-standard-mode-recording-tab-only-lossless-audio), and probably many more) due to the additional flexibility in editing and avoidance of generational loss.

Additionally, PCM floating point audio allows for some additional flexibility when it comes to editing and fixing of audio in post, as clipping audio (i.e. >1.0f or <-1.0f) is retained and can be adjusted down to no longer be clipping. Aside from being able to make up for wrong recording levels in editing, this is also useful in situations where constructive interference resulted in the final output clipping while individually sources were fine (this currently happens invisibly due to the lack of a master mixer).

### How Has This Been Tested?

Recorded some videos.

### Types of changes

- New feature (non-breaking change which adds functionality)

### Checklist:

- [x] My code has been run through [clang-format](https://github.com/obsproject/obs-studio/blob/master/.clang-format).
- [x] I have read the [**contributing** document](https://github.com/obsproject/obs-studio/blob/master/CONTRIBUTING.rst).
- [x] My code is not on the master branch.
- [x] The code has been tested.
- [x] All commit messages are properly formatted and commits squashed where appropriate.
- [x] I have included updates to all appropriate documentation.

PR #8498, opened 2023-03-20
UI: Disable incompatible codecs in settings

derrod:
### Description

Disable codecs that are incompatible based on the container drop down state. If an incompatible codec is selected when changing container, the encoder selection is reset to an invalid state and a warning is shown.

**Screenshot:**
2023-03-24_11-30-51_cCsexx

### Motivation and Context

Wanted to avoid having to add more warnings for Opus and #8289

### How Has This Been Tested?

Selected different contaiers, ensured codecs got disabled as necessary. Also tested with #8289 and made sure that works correctly.

### Types of changes

### Checklist:

### DIFF
--- a/UI/window-basic-settings.cpp
+++ b/UI/window-basic-settings.cpp
@@ -4919,47 +4919,17 @@ void OBSBasicSettings::AdvOutRecCheckWarnings()
 			errorMsg = QTStr("OutputWarnings.NoTracksSelected");
 	}
 
-	QString recEncoder = ui->advOutRecEncoder->currentText();
+	if (recFormat == "mp4" || recFormat == "mov") {
+		if (!warningMsg.isEmpty())
+			warningMsg += "\n\n";
 
-	if (recEncoder.contains("ProRes")) {
-		if (recFormat == "mkv") {
-			ui->autoRemux->setText(
-				QTStr("Basic.Settings.Advanced.AutoRemux")
-					.arg("mov"));
-		} else if (recFormat == "mov") {
-			if (!warningMsg.isEmpty())
-				warningMsg += "\n\n";
-			warningMsg += QTStr("OutputWarnings.MP4Recording");
-
-			ui->autoRemux->setText(
-				QTStr("Basic.Settings.Advanced.AutoRemux")
-					.arg("mov") +
-				" " +
-				QTStr("Basic.Settings.Advanced.AutoRemux.MP4"));
-		} else {
-			if (!errorMsg.isEmpty()) {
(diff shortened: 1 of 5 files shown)


notr1ch:
I think it might be nicer if the container was always shown in uppercase in the incompatible warning.

derrod:
> I think it might be nicer if the container was always shown in uppercase in the incompatible warning.

Sure that can be done. See the updated screenshot.

derrod:
OK'd by @Warchamp7 off-thread, looking to merge during weekly PR review.

derrod:
Based on feedback from @Warchamp7 the following changes have been implemented:

- When an invalid codec is selected, the selection is now set to "nothing" (-1) and a placeholder text is shown
- A warning is visible as long as either encoder selection is invalid
- The same behaviour now applies to simple mode as long as recording encoders are used
- When using stream encoders in simple mode recording formats are disabled instead

See updated screenshots for the placeholder and warning.

UI/window-basic-settings.cpp
```diff
@@ -4890,6 +4911,116 @@ void OBSBasicSettings::AdvOutSplitFileChanged()
...
+	DisableIncompatibleCodecs(ui->advOutRecEncoder, format, streamEncoder);
+	DisableIncompatibleCodecs(ui->advOutRecAEncoder, format,
+				  streamAudioEncoder);
+	ui->advOutRecEncoder->blockSignals(false);
+	ui->advOutRecAEncoder->blockSignals(false);
+
+	AdvOutRecCheckWarnings();
+}
+
+#ifdef __APPLE__
```
gxalpha, line 5003: Please add something along the lines of
"TODO: Revisit when QTBUG-XXXXX is fixed"
so that we know when we can remove it

derrod: I did where the function is invoked rather than at the function declaration

gxalpha: Might be useful in both places, otherwise the function might get left, but probably not that important then.

Warchamp7 (approved):

notr1ch:
Closing the settings without picking a replacement encoder seems to stop recording from functioning at all (not even anything in the log file). After re-opening settings, there is no obvious error as the list index is reset to "Use stream encoder" (but internally not).

https://user-images.githubusercontent.com/876345/227736102-898b41f2-a1e9-4aea-a8b1-66a1d2dd5b0f.mp4

derrod:
Should be fixed now. Apparently the way the comboboxes are restored is slightly different so this only affected the audio one.

PR #8499, opened 2023-03-20
UI: Fix simple fallback audio codec name

tytan652:
### Description

Fix this while switching from an RTMP to an FTL service (and vice versa):
Bug

I shouldn't have relied on the combobox since it will not contain the other codec.

### Motivation and Context

Fix a bug.

### How Has This Been Tested?

Set an RTMP service and then switch to an FTL one, the dialog should no longer have an empty string for the second codec.

### Types of changes

### Checklist:


UI/window-basic-settings-stream.cpp
```diff
@@ -1429,8 +1429,7 @@ bool OBSBasicSettings::ServiceSupportsCodecCheck()
 
 		cur_audio_name = ui->simpleOutStrAEncoder->itemText(
 			ui->simpleOutStrAEncoder->findData(cur_enc));
-		fb_audio_name = ui->simpleOutStrAEncoder->itemText(
-			ui->simpleOutStrAEncoder->findData(fb_enc));
+		fb_audio_name = (cur_audio_name == "Opus") ? "AAC" : "Opus";
```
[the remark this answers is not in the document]
tytan652: It doesn't really matter but I still made the change.

UI/window-basic-settings-stream.cpp
```diff
@@ -1429,8 +1429,7 @@ bool OBSBasicSettings::ServiceSupportsCodecCheck()
 
 		cur_audio_name = ui->simpleOutStrAEncoder->itemText(
 			ui->simpleOutStrAEncoder->findData(cur_enc));
-		fb_audio_name = ui->simpleOutStrAEncoder->itemText(
-			ui->simpleOutStrAEncoder->findData(fb_enc));
+		fb_audio_name = (cur_enc == "opus") ? "AAC" : "Opus";
```
RytoEX, line 1432: Isn't fb_audio_name a displayed/translated string? Is there be a way to get the appropriate value without hard-coding like this? We do this in the `else` branch, though that also seems to rely on some hard-coding from `get_adv_audio_fallback()`:
```
fb_audio_name = obs_encoder_get_display_name(QT_TO_UTF8(fb_enc));
```

RytoEX (commented):
A question about the approach here.

tytan652:
I answer by e-mail so it might not be in the related conversation.

In Simple mode, only codec name are used (Opus, AAC) and are not sourced from translations. Those name are harcoded.
The user choose a codec not a specific implementation.

In Advanced mode, encoder implementation names are used since the user can choose the implementation in this mode.

UI/window-basic-settings-stream.cpp
tytan652: Old Message:
> I answer by e-mail so it might not be in the related conversation.
>
>In Simple mode, only codec name are used (Opus, AAC) and are not sourced from translations. Those name are harcoded.
>The user choose a codec not a specific implementation.
>
>In Advanced mode, encoder implementation names are used since the user can choose the implementation in this mode.

New message:
I made audio codec names shown in the UI translatable.

> though that also seems to rely on some hard-coding from `get_adv_audio_fallback()`

Like the video one.

tytan652: > Is there be a way to get the appropriate value without hard-coding like this?

In simple mode, no. If AAC is not available we need to switch to Opus and vice versa. There is `get_simple_fallback` for video which is also hardcoded.

In advanced mode,

derrod: Not sure if it's really necessary to make the codec name translateable since as proper nouns they shouldn't be translated.

tytan652:
Added `(Default)` to the AAC string.

PR #8500, opened 2023-03-20
UI: Fix inconsistencies in FindProtocol 

tytan652:
### Description

Fix inconsistencies that might lead to issue or even a crash.

Note: The else branch should not have been taken in custom server scenarios.

### Motivation and Context

Fix a crash reported on discord: https://discord.com/channels/348973006581923840/374636084883095554/1087384798894501939

### How Has This Been Tested?

Switch from a FTL service to custom (with empty server URL) and the encoder warning shown up. Without the fix it doesn't show.

### Types of changes

### Checklist:

### DIFF
--- a/UI/window-basic-settings-stream.cpp
+++ b/UI/window-basic-settings-stream.cpp
@@ -533,11 +533,14 @@ void OBSBasicSettings::ServiceChanged()
 
 QString OBSBasicSettings::FindProtocol()
 {
-	if (IsCustomService() && !ui->customServer->text().isEmpty()) {
+	if (IsCustomService()) {
+		if (ui->customServer->text().isEmpty())
+			return QString("RTMP");
 
 		QString server = ui->customServer->text();
 
-		if (server.startsWith("rtmps://"))
+		if (obs_is_output_protocol_registered("RTMPS") &&
+		    server.startsWith("rtmps://"))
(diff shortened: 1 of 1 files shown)


notr1ch:
I'm generally not a fan of branches that use the same variable in different conditions like this. Personally I think an early return in the first branch or letting it fall through would be better.

notr1ch (approved):

UI/window-basic-settings-stream.cpp
```diff
@@ -533,11 +533,15 @@ void OBSBasicSettings::ServiceChanged()
 
 QString OBSBasicSettings::FindProtocol()
 {
-	if (IsCustomService() && !ui->customServer->text().isEmpty()) {
+	if (IsCustomService()) {
+
+		if (ui->customServer->text().isEmpty())
```
RytoEX, line 538: Nit: remove the blank line here.

RytoEX (changes requested):

UI/window-basic-settings-stream.cpp
tytan652: Done

RytoEX (approved):
Typically, no need to end list items with periods, but not going to block on this.

Looks good to me.

PR #8508, opened 2023-03-22
Make sure that protocol is set in non-custom services

tytan652:
### Description

I'm an idiot that forgot some migration code in `rtmp_custom` and code to save the protocol in settings.

### Motivation and Context

Fix bugs that I introduced.

### How Has This Been Tested?

Without the fix service like Twitch don't return a protocol and so start streaming fail.

With it, streaming works and protocol is saved in the JSON in the settings.

### Types of changes

### Checklist:


RytoEX (approved):
Approved off-thread by @notr1ch .

PR #8522, opened 2023-03-25
Enable AV1, HEVC via RTMP to YouTube

nquah:
Implementation credited to yuriy-chumak

### Description

- Adds support for HEVC and AV1 codecs to RTMP, according to the [Enhanced RTMP spec](https://github.com/veovera/enhanced-rtmp)

- Adds flag-guarded support for HDR metadata frame to RTMP, according to the [Enhanced RTMP spec](https://github.com/veovera/enhanced-rtmp)

### Motivation and Context

Veovera Software Organization (aka VSO), a not-for-profit organization, was formed to help enhance software for the public good. VSO members currently include Adobe, Google and Veriskope. VSO published an [Enhanced RTMP spec](https://github.com/veovera/enhanced-rtmp), which adds long-awaited support for HEVC and AV1 to the FLV container, enabling their use in RTMP. 

Furthermore, with the advent of GPUs that support real-time AV1 encoding, we can now enable AV1 streaming. 

This change updates the RTMP implementation in OBS to support HEVC and AV1.  It additionally enables these codecs for RTMP ingestion to YouTube. This YouTube feature is currently in beta.

Note: YouTube transcodes live streams, and does not currently support AV1 for live delivery. Creators can still benefit from maintaining higher ingestion quality, with fewer compression artifacts, even at lower bitrates.

### How Has This Been Tested?

- Tested using FFmpeg as the RTMP server, and FFplay as the client.

- Tested against YouTube’s RTMP server, up to 4K 60fps. 

- Tested AV1 live encoding using NVENC and Intel encoders. 

### Types of changes

### Known Issues
- This feature is in beta at YouTube. Not all encoder settings may yield equal results.
- Playback of rich content at high resolution may struggle for reasons unrelated to encoding performance, e.g. network connectivity or device performance of the client playing the stream.
- HDR via RTMP is flag-guarded and not yet supported by YouTube.
- 10-bit AV1 streaming to YouTube may not work properly yet.

### Checklist:


RytoEX (changes requested):
Some quick feedback.

Commit messages must follow the 50/72 rule, outlined in our [Commit Guidelines](https://github.com/obsproject/obs-studio/blob/ddee7ff6feeea36ac449dc6077c6123f7889dc4e/CONTRIBUTING.rst#commit-guidelines).

Commit messages should have a prefix for the commit subject. In this case, as there are multiple components being modified, `obs-outputs` is probably most appropriate as the prefix. Then the commit message body should note that "this commit also modifies libobs, rtmp-services, and UI components". See other commits for examples.

Files must be formatted with clang-format 13.0.1. Currently, the file that is failing this check is `plugins/obs-outputs/rtmp-av1.h`.

I will leave code review to others.

plugins/rtmp-services/data/services.json
```diff
@@ -235,6 +235,11 @@
                 "YouTube - RTMP",
                 "YouTube - RTMPS (Beta)"
             ],
+            "supported video codecs": [
+                "h264",
+                "hevc",
+                "av1"
+            ],
```
tytan652, line 242: You need to do the opposite. You need to add the following to very RTMP/RTMPS service:
```
            "supported video codecs": [
                "h264"
            ],
```

To avoid adding HEVC and AV1 support yo every RTMP/RTMPS services.

tytan652 (changes requested):

plugins/obs-outputs/rtmp-stream.c
```diff
@@ -1234,6 +1424,24 @@ static void *connect_thread(void *data)
...
+			video_output_get_info(video);
+		enum video_format format = info->format;
+		enum video_colorspace colorspace = info->colorspace;
+
+		if (format >= VIDEO_FORMAT_I010) { // 10-bits video and more
+			obs_output_signal_stop(stream->output,
+					       OBS_OUTPUT_HDR_DISABLED);
+			return NULL;
+		}
+	}
```
tytan652, line 1443: I don't understand, it does not stop on HDR colorspace but on 10-bit format.

derrod: Yeah this would need clarification on whether or not only HDR colourspaces should be disallowed, or 10-bit video in its entirety (since 10-bit SDR is possible).

PatTheMav: Also don’t treat an enum like an integer - the order is not guaranteed and we could choose to change or reorder the entries - explicitly check for specific entries or use a `switch` statement.

libobs/obs-bits.h
```diff
@@ -0,0 +1,71 @@
...
+}
+static uint32_t ctz32(uint32_t x)
+{
+	return popcnt((x & -x) - 1);
+}
+
+#endif
+
+#ifndef MIN
+#define MIN(x, y) ((x) < (y) ? (x) : (y))
```
PatTheMav, line 51: Convert to a static inline function, please avoid macros and defines as much as possible, as there is no need for them in modern C/C++.

plugins/obs-outputs/flv-mux.c
```diff
@@ -32,6 +32,85 @@
...
 #define AUDIODATA_AAC 10.0
 
+#define VIDEO_FRAMETYPE_OFFSET 4
+enum video_frametype_t {
+	FT_KEY = 1 << VIDEO_FRAMETYPE_OFFSET,
+	FT_INTER = 2 << VIDEO_FRAMETYPE_OFFSET,
+};
+
+// Y2023 spec
+#define FRAME_HEADER_EX (8 << VIDEO_FRAMETYPE_OFFSET)
```
PatTheMav, line 42: Use `const` types as much as possible, avoid using defines (admittedly as C has no `constexpr` there’s
sometimes no other way).

plugins/rtmp-services/data/services.json
PatTheMav: It’s not the ‘opposite’ per-se, but I agree that we need to make this explicit (should’ve been in the first place, ‘assuming’ stuff from a lack of info is not a great design choice). As such we need to make that explicit now and have services explicitly state if they only support h264.

tytan652: @PatTheMav, ` "supported video codecs"` is meant to enable only a subset of what the protocol/output can handle (e.g. only H264 on HLS). Until now, RTMP was H264 only for streaming services.

So HEVC and AV1 are suddenly added to RTMP support and all RTMP services added until now only support H264 since it was the "standard".

PatTheMav: I understood that, but that means the configuration so far was very much tailored to a "there is only RTMP and thus there is only H264" mindset. Instead the mindset should be "services support specific codecs" and vanilla RTMP just happens to support H264 only, which should not be implicitly assumed by a config file or code parsing it.

The more pertinent question is whether we can add those "explicit" entries on master already, so we don't burden this PR with that cleanup work?

plugins/obs-outputs/rtmp-stream.c
nquah: Yes good call, we will change this to check on colorspace instead.  We have a bug with 10-bit AV1 streams but I will add that to Known Issues instead of guarding the behaviour here.

plugins/rtmp-services/data/services.json
nquah: Thanks for spotting, let us know, happy to do add them explicitly or wait for it on master.

derrod:
I pushed my changes now. These mainly split up the PR into different commits to make things a little easier to review. It also splits out the change blocking HDR into a separate commit which can more easily be reverted later if this limitation no longer exists.

Additionally there were a number of cleanup changes I made such as removing (now) unused code as well as removing the reliance on FFmpeg (which obs-outputs does not link to) for compilation.

derrod:
Whoops *somebody* forgot to run this through Linux CI so a breaking change was undetected because MSVC does things differently, yay. Hopefully should be fixed now.

plugins/obs-outputs/flv-mux.c
nquah: Done in the last commit from yuriy-chumak

libobs/obs-bits.h
nquah: Done in the last commit from yuriy-chumak

plugins/obs-outputs/rtmp-stream.c
nquah: Done

plugins/rtmp-services/data/services.json
nquah: Looks like this is done now, thanks.

UI/data/locale/en-US.ini
```diff
@@ -402,6 +402,7 @@ Output.ConnectFail.Title="Failed to connect"
 Output.ConnectFail.BadPath="Invalid Path or Connection URL. Please check your settings to confirm that they are valid."
 Output.ConnectFail.ConnectFailed="Failed to connect to server"
 Output.ConnectFail.InvalidStream="Could not access the specified channel or stream key, please double-check your stream key. If it is correct, there may be a problem connecting to the server."
+Output.ConnectFail.HdrDisabled="HDR output over RTMP is currently disabled for this codec"
```
tytan652, line 405: The message should not specify RTMP since any output (protocol) could use `OBS_OUTPUT_HDR_DISABLED`.

yuriy-chumak: Updated.

plugins/rtmp-services/data/services.json
```diff
@@ -366,15 +386,21 @@
...
+            "supported video codecs": [
+                "h264"
+            ]
         },
         {
             "name": "GoodGame.ru",
             "servers": [
                 {
-                    "name": "Моscow",
+                    "name": "\u041c\u043escow",
```
tytan652, line 398:
```suggestion
                    "name": "Моscow",
```

plugins/rtmp-services/data/services.json
```diff
@@ -746,12 +793,15 @@
...
+            },
+            "supported video codecs": [
+                "h264"
+            ]
         },
         {
             "name": "AfreecaTV",
             "alt_names": [
-                "아프리카TV",
+                "\uc544\ud504\ub9ac\uce74TV",
```
tytan652, line 804:
```suggestion
                "아프리카TV",
```

plugins/rtmp-services/data/services.json
```diff
@@ -1169,10 +1261,13 @@
...
                 "max fps": 30
-            }
+            },
+            "supported video codecs": [
+                "h264"
+            ]
         },
         {
-            "name": "Lahzenegar - StreamG | لحظه‌نگار - استریمجی",
+            "name": "Lahzenegar - StreamG | \u0644\u062d\u0638\u0647\u200c\u0646\u06af\u0627\u0631 - \u0627\u0633\u062a\u0631\u06cc\u0645\u062c\u06cc",
```
tytan652, line 1270:
```suggestion
            "name": "Lahzenegar - StreamG | لحظه‌نگار - استریمجی",
```

plugins/rtmp-services/data/services.json
```diff
@@ -1501,10 +1647,13 @@
...
                 "keyint": 2
-            }
+            },
+            "supported video codecs": [
+                "h264"
+            ]
         },
         {
-            "name": "niconico, premium member (ニコニコ生放送 プレミアム会員)",
+            "name": "niconico, premium member (\u30cb\u30b3\u30cb\u30b3\u751f\u653e\u9001 \u30d7\u30ec\u30df\u30a2\u30e0\u4f1a\u54e1)",
```
tytan652, line 1656:
```suggestion
            "name": "niconico, premium member (ニコニコ生放送 プレミアム会員)",
```

plugins/rtmp-services/data/services.json
```diff
@@ -1517,10 +1666,13 @@
...
                 "x264opts": "tune=zerolatency"
-            }
+            },
+            "supported video codecs": [
+                "h264"
+            ]
         },
         {
-            "name": "niconico, free member (ニコニコ生放送 一般会員)",
+            "name": "niconico, free member (\u30cb\u30b3\u30cb\u30b3\u751f\u653e\u9001 \u4e00\u822c\u4f1a\u54e1)",
```
tytan652, line 1675:
```suggestion
            "name": "niconico, free member (ニコニコ生放送 一般会員)",
```

plugins/rtmp-services/data/services.json
```diff
@@ -1932,10 +2117,13 @@
...
                 "x264opts": "scenecut=0"
-            }
+            },
+            "supported video codecs": [
+                "h264"
+            ]
         },
         {
-            "name": "OPENREC.tv - Premium member (プレミアム会員)",
+            "name": "OPENREC.tv - Premium member (\u30d7\u30ec\u30df\u30a2\u30e0\u4f1a\u54e1)",
```
tytan652, line 2126:
```suggestion
            "name": "OPENREC.tv - Premium member (プレミアム会員)",
```

plugins/rtmp-services/data/services.json
```diff
@@ -2034,20 +2228,26 @@
...
                 "max audio bitrate": 128
-            }
+            },
+            "supported video codecs": [
+                "h264"
+            ]
         },
         {
-            "name": "Bilibili Live - RTMP | 哔哩哔哩直播 - RTMP",
+            "name": "Bilibili Live - RTMP | \u54d4\u54e9\u54d4\u54e9\u76f4\u64ad - RTMP",
```
tytan652, line 2237:
```suggestion
            "name": "Bilibili Live - RTMP | 哔哩哔哩直播 - RTMP",
```

plugins/rtmp-services/data/services.json
```diff
@@ -2034,20 +2228,26 @@
...
             "more_info_link": "https://link.bilibili.com/p/help/index#/tools-tutorial?id=9",
             "stream_key_link": "https://link.bilibili.com/p/center/index#/my-room/start-live",
             "alt_names": [
                 "Bilibili Live"
             ],
             "servers": [
                 {
-                    "name": "Default | 默认",
+                    "name": "Default | \u9ed8\u8ba4",
```
tytan652, line 2245:
```suggestion
                    "name": "Default | 默认",
```

plugins/rtmp-services/data/services.json
```diff
@@ -2426,10 +2671,13 @@
...
                 "max audio bitrate": 192
-            }
+            },
+            "supported video codecs": [
+                "h264"
+            ]
         },
         {
-            "name": "Whowatch (ふわっち)",
+            "name": "Whowatch (\u3075\u308f\u3063\u3061)",
```
tytan652, line 2680:
```suggestion
            "name": "Whowatch (ふわっち)",

**Review by tytan652 — CHANGES_REQUESTED** [review]
Unicode character issues.

**Comment by flying1314** [comment]
could you provide the modified ffmpeg for us to test it

**Comment by Fenrirthviti** [comment]
> could you provide the modified ffmpeg for us to test it

We are not using a modified ffmpeg. The changes for our support are to librtmp.

**Comment by winlinvip** [comment]
SRS media server support this extended RTMP specification, so you can push HEVC via RTMP to SRS very easy:

```bash
git clone https://github.com/ossrs/srs.git
cd srs/trunk && ./configure --h265=on && make
./objs/srs -c conf/http.ts.live.conf
```

Then, you can use [OBS 29.1+](https://github.com/obsproject/obs-studio/releases) to push HEVC via RTMP.
Start OBS with the following settings in the `Settings > Stream` tab:

* Server: `rtmp://localhost/live`
* Stream Key: `livestream`
* Encoder: Please select the HEVC hardware encoder.

image

image

Finally, open the player http://localhost:8080/players/srs_player.html?stream=livestream.ts

Or use VLS or ffplay to play `http://localhost:8080/live/livestream.ts`

PR #8529, opened 2023-03-25
Add initial support for the OpenH264 H.264 software codec

Conan-Kudo:
### Description
This allows users to leverage the OpenH264 codec from Cisco to encode H.264 video content. It is significantly reduced in capability from alternatives, but it does the job.

This also provides a framework for adding support for other H.264 software codecs provided through FFmpeg.

### Motivation and Context
A number of distributions use [OpenH264 distributed by Cisco](http://www.openh264.org/) to provide a fully licensed H.264 implementation (in particular, both [Fedora Linux](https://fedoraproject.org/wiki/OpenH264) and [openSUSE Linux](https://en.opensuse.org/OpenH264) do). However, OBS Studio cannot currently use it.

This change allows OBS Studio to use OpenH264 through FFmpeg if it's available.

### How Has This Been Tested?

* Tested recording a screencast using OpenH264 + AAC with both Simple and Advanced settings modes
* Tested streaming a screencast using OpenH264 + AAC with both Simple and Advanced settings modes

This was tested on Fedora Linux 38 for x86_64 with ffmpeg 6.0.

### Types of changes

### Checklist:

### DIFF
--- a/frontend/data/locale/en-US.ini
+++ b/frontend/data/locale/en-US.ini
@@ -1053,6 +1053,7 @@ Basic.Settings.Output.Simple.Warn.Lossless="Warning: Lossless quality generates 
 Basic.Settings.Output.Simple.Warn.Lossless.Msg="Are you sure you want to use lossless quality?"
 Basic.Settings.Output.Simple.Warn.Lossless.Title="Lossless quality warning!"
 Basic.Settings.Output.Simple.Encoder.Software="Software (x264)"
+Basic.Settings.Output.Simple.Encoder.Software.OpenH264="Software (OpenH264)"
 Basic.Settings.Output.Simple.Encoder.Hardware.QSV.H264="Hardware (QSV, H.264)"
 Basic.Settings.Output.Simple.Encoder.Hardware.QSV.AV1="Hardware (QSV, AV1)"
 Basic.Settings.Output.Simple.Encoder.Hardware.AMD.H264="Hardware (AMD, H.264)"
--- a/frontend/settings/OBSBasicSettings.cpp
+++ b/frontend/settings/OBSBasicSettings.cpp
@@ -3343,6 +3343,11 @@ void OBSBasicSettings::SaveOutputSettings()
          do. This only exists to make sure that the x264 preset doesn't
          get overwritten with empty data. */
 		presetType = "ApplePreset";
+	else if (encoder == SIMPLE_ENCODER_OPENH264)
+		/* The OpenH264 encoder does not have presets like the other encoders
+         do. This only exists to make sure that the x264 preset doesn't
+         get overwritten with empty data. */
+		presetType = "OpenH264Preset";
 	else
 		presetType = "Preset";
 
@@ -4714,8 +4719,11 @@ void OBSBasicSettings::FillSimpleRecordingValues()
 	ADD_QUALITY("HQ");
 	ADD_QUALITY("Lossless");
 
-	ui->simpleOutRecEncoder->addItem(ENCODER_STR("Software"), QString(SIMPLE_ENCODER_X264));
-	ui->simpleOutRecEncoder->addItem(ENCODER_STR("SoftwareLowCPU"), QString(SIMPLE_ENCODER_X264_LOWCPU));
+	ui->simpleOutRecEncoder->addItem(ENCODER_STR("Software.OpenH264"), QString(SIMPLE_ENCODER_OPENH264));
+	if (EncoderAvailable("obs_x264")) {
+		ui->simpleOutRecEncoder->addItem(ENCODER_STR("Software"), QString(SIMPLE_ENCODER_X264));
+		ui->simpleOutRecEncoder->addItem(ENCODER_STR("SoftwareLowCPU"), QString(SIMPLE_ENCODER_X264_LOWCPU));
+	}
 	if (EncoderAvailable("obs_qsv11"))
 		ui->simpleOutRecEncoder->addItem(ENCODER_STR("Hardware.QSV.H264"), QString(SIMPLE_ENCODER_QSV));
 	if (EncoderAvailable("obs_qsv11_av1"))
@@ -4864,6 +4872,9 @@ void OBSBasicSettings::SimpleStreamingEncoderChanged()
 
 		defaultPreset = "balanced";
 		preset = curAMDAV1Preset;
+	} else if (encoder == SIMPLE_ENCODER_OPENH264) {
+		ui->simpleOutPreset->setVisible(false);
+		ui->simpleOutPresetLabel->setVisible(false);
 	} else {
 
 #define PRESET_STR(val) QString(Str("Basic.Settings.Output.EncoderPreset." val)).arg(val)
--- a/frontend/settings/OBSBasicSettings_Stream.cpp
+++ b/frontend/settings/OBSBasicSettings_Stream.cpp
@@ -1471,7 +1471,9 @@ static QString get_adv_fallback(const QString &enc)
 		return "com.apple.videotoolbox.videoencoder.ave.avc";
 	if (enc == "obs_qsv11_av1")
 		return "obs_qsv11";
-	return "obs_x264";
+	if (EncoderAvailable("obs_x264"))
+		return "obs_x264";
+	return "ffmpeg_openh264";
 }
 
 static QString get_adv_audio_fallback(const QString &enc)
@@ -1500,7 +1502,9 @@ static QString get_simple_fallback(const QString &enc)
 		return SIMPLE_ENCODER_APPLE_H264;
 	if (enc == SIMPLE_ENCODER_QSV_AV1)
 		return SIMPLE_ENCODER_QSV;
-	return SIMPLE_ENCODER_X264;
+	if (EncoderAvailable("obs_x264"))
+		return SIMPLE_ENCODER_X264;
+	return SIMPLE_ENCODER_OPENH264;
 }
 
 bool OBSBasicSettings::ServiceSupportsCodecCheck()
@@ -1675,7 +1679,9 @@ void OBSBasicSettings::ResetEncoders(bool streamOnly)
 
 #define ENCODER_STR(str) QTStr("Basic.Settings.Output.Simple.Encoder." str)
 
-	ui->simpleOutStrEncoder->addItem(ENCODER_STR("Software"), QString(SIMPLE_ENCODER_X264));
+	ui->simpleOutStrEncoder->addItem(ENCODER_STR("Software.OpenH264"), QString(SIMPLE_ENCODER_OPENH264));
+	if (service_supports_encoder(vcodecs, "obs_x264"))
+		ui->simpleOutStrEncoder->addItem(ENCODER_STR("Software"), QString(SIMPLE_ENCODER_X264));
 #ifdef _WIN32
 	if (service_supports_encoder(vcodecs, "obs_qsv11"))
 		ui->simpleOutStrEncoder->addItem(ENCODER_STR("Hardware.QSV.H264"), QString(SIMPLE_ENCODER_QSV));
--- a/frontend/utility/SimpleOutput.cpp
+++ b/frontend/utility/SimpleOutput.cpp
@@ -80,7 +80,9 @@ void SimpleOutput::LoadStreamingPreset_Lossy(const char *encoderId)
 /* mistakes have been made to lead us to this. */
 const char *get_simple_output_encoder(const char *encoder)
 {
-	if (strcmp(encoder, SIMPLE_ENCODER_X264) == 0) {
+	if (strcmp(encoder, SIMPLE_ENCODER_OPENH264) == 0) {
+		return "ffmpeg_openh264";
+	} else if (strcmp(encoder, SIMPLE_ENCODER_X264) == 0) {
 		return "obs_x264";
 	} else if (strcmp(encoder, SIMPLE_ENCODER_X264_LOWCPU) == 0) {
 		return "obs_x264";
@@ -112,7 +114,7 @@ const char *get_simple_output_encoder(const char *encoder)
 #endif
 	}
 
-	return "obs_x264";
+	return "ffmpeg_openh264";
 }
 
 void SimpleOutput::LoadRecordingPreset()
--- a/frontend/widgets/OBSBasic.cpp
+++ b/frontend/widgets/OBSBasic.cpp
@@ -519,6 +519,8 @@ static const double scaled_vals[] = {1.0, 1.25, (1.0 / 0.75), 1.5, (1.0 / 0.6), 
 #define DEFAULT_CONTAINER "hybrid_mp4"
 #endif
 
+extern bool EncoderAvailable(const char *encoder);
+
 bool OBSBasic::InitBasicConfigDefaults()
 {
 	QList<QScreen *> screens = QGuiApplication::screens();
@@ -680,7 +682,9 @@ bool OBSBasic::InitBasicConfigDefaults()
 	config_set_default_bool(activeConfiguration, "AdvOut", "UseRescale", false);
 	config_set_default_uint(activeConfiguration, "AdvOut", "TrackIndex", 1);
 	config_set_default_uint(activeConfiguration, "AdvOut", "VodTrackIndex", 2);
-	config_set_default_string(activeConfiguration, "AdvOut", "Encoder", "obs_x264");
+
+	bool useX264 = EncoderAvailable("obs_x264");
+	config_set_default_string(activeConfiguration, "AdvOut", "Encoder", (useX264 ? "obs_x264" : "ffmpeg_openh264"));
 
 	config_set_default_string(activeConfiguration, "AdvOut", "RecType", "Standard");
 
@@ -796,10 +800,13 @@ void OBSBasic::InitBasicConfigDefaults2()
 	bool oldEncDefaults = config_get_bool(App()->GetUserConfig(), "General", "Pre23Defaults");
 	bool useNV = EncoderAvailable("ffmpeg_nvenc") && !oldEncDefaults;
 
+	bool useX264 = EncoderAvailable("obs_x264");
+	const char *h264_fallback = (useX264 ? SIMPLE_ENCODER_X264 : SIMPLE_ENCODER_OPENH264);
+
 	config_set_default_string(activeConfiguration, "SimpleOutput", "StreamEncoder",
-				  useNV ? SIMPLE_ENCODER_NVENC : SIMPLE_ENCODER_X264);
+				  useNV ? SIMPLE_ENCODER_NVENC : h264_fallback);
 	config_set_default_string(activeConfiguration, "SimpleOutput", "RecEncoder",
-				  useNV ? SIMPLE_ENCODER_NVENC : SIMPLE_ENCODER_X264);
+				  useNV ? SIMPLE_ENCODER_NVENC : h264_fallback);
 
 	const char *aac_default = "ffmpeg_aac";
 	if (EncoderAvailable("CoreAudio_AAC"))
--- a/frontend/widgets/OBSBasic.hpp
+++ b/frontend/widgets/OBSBasic.hpp
@@ -75,6 +75,7 @@ struct Rect;
 
 #define SIMPLE_ENCODER_X264 "x264"
 #define SIMPLE_ENCODER_X264_LOWCPU "x264_lowcpu"
+#define SIMPLE_ENCODER_OPENH264 "ffmpeg_openh264"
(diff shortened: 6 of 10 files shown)


RytoEX:
Does this require #8015?

Conan-Kudo:
> Does this require #8015?

Not yet. This is only the wiring for the ffmpeg plugin, which I think means it only shows up in the "advanced view" right now? I literally just banged this out and am taking a bit of a break before continuing to iterate on it. I've got local builds going so I can test the functionality...

Conan-Kudo:
@RytoEX That said, having #8015 land first would make it tremendous easier for the UI wiring I need to do...

Conan-Kudo:
I've pulled in #8015 into this pull request to simplify things for me and to show the full context of this change.

Conan-Kudo:
I've now tested that this works, thanks to @Gawdl3y's help! 🎉

tytan652:
We recently merged a rework of our CMake build-system.

Now, `plugins/obs-ffmpeg/CMakeLists.txt` and `plugins/obs-ffmpeg/cmake/legacy.txt` should be updated.

tytan652 (changes requested):
This PR adds FFmpeg OpenH264 because Fedora can't "easily" provide x264.

- But it does not mean that all OBS (or FFmpeg) builds will have OpenH264 support (e.g., Windows, macOS, Flatpak, other distro). There is no reason to have enabled OpenH264 and x264 when the later is easily available.

- In simple mode, the user must not be able to choose between x264 and OpenH264. If x264 is available, OpenH264 should not be shown.
  - If a user loads a profile that was using (in simple mode) OpenH264 and x264 is available, it might be better to make the user switch to x264.

- When it comes to defaults, x264 (if available) should be prioritized over OpenH264.

- Some parts of this PR make the assumption that FFmpeg OpenH264 is always available.
The assumption should be: FFmpeg OpenH264 is always available **if** x264 is not.

UI/window-basic-main.cpp
```diff
@@ -1542,7 +1544,15 @@ bool OBSBasic::InitBasicConfigDefaults()
...
-	config_set_default_string(basicConfig, "AdvOut", "Encoder", "obs_x264");
+
+	bool useX264 = EncoderAvailable("obs_x264");
+
+	if (useX264)
+		config_set_default_string(basicConfig, "AdvOut", "Encoder",
+					  "obs_x264");
+	else
+		config_set_default_string(basicConfig, "AdvOut", "Encoder",
+					  "ffmpeg_openh264");
```
tytan652, line 1555:
```suggestion
	config_set_default_string(basicConfig, "AdvOut", "Encoder", useX264 ? "obs_x264" : "ffmpeg_openh264");
```

Note: Not formatted

UI/window-basic-main.cpp
```diff
@@ -1674,12 +1683,14 @@ void OBSBasic::InitBasicConfigDefaults2()
 					      "Pre23Defaults");
 	bool useNV = EncoderAvailable("ffmpeg_nvenc") && !oldEncDefaults;
 
+	bool useX264 = EncoderAvailable("obs_x264");
+	const char *h264_fallback =
+		(useX264 ? SIMPLE_ENCODER_X264 : SIMPLE_ENCODER_OPENH264);
```
tytan652, line 1688:
```suggestion
		useX264 ? SIMPLE_ENCODER_X264 : SIMPLE_ENCODER_OPENH264;
```

UI/window-basic-settings-stream.cpp
```diff
@@ -1360,7 +1360,9 @@ static QString get_adv_fallback(const QString &enc)
 		return "h264_texture_amf";
 	if (enc == "com.apple.videotoolbox.videoencoder.ave.hevc")
 		return "com.apple.videotoolbox.videoencoder.ave.avc";
-	return "obs_x264";
+	if (enc == "obs_x264")
+		return "obs_x264";
+	return "ffmpeg_openh264";
```
tytan652, line 1365: I think you misunderstood the function purpose, it fallbacks HEVC encoders to H264 (some improvement needs to be done like you can see in the comment earlier in the code).

Last resort fallback must be obs_x264 if available (and ffmpeg_openh264 if no x264).

UI/window-basic-settings-stream.cpp
```diff
@@ -1583,8 +1587,14 @@ void OBSBasicSettings::ResetEncoders(bool streamOnly)
 
 #define ENCODER_STR(str) QTStr("Basic.Settings.Output.Simple.Encoder." str)
 
-	ui->simpleOutStrEncoder->addItem(ENCODER_STR("Software"),
-					 QString(SIMPLE_ENCODER_X264));
+	if (!service_supports_encoder(vcodecs, "obs_x264"))
+		ui->simpleOutStrEncoder->addItem(
+			ENCODER_STR("Software.OpenH264.H264"),
+			QString(SIMPLE_ENCODER_OPENH264));
+	if (service_supports_encoder(vcodecs, "obs_x264"))
```
tytan652, line 1594:
```suggestion
	else
```

UI/window-basic-settings.cpp
```diff
@@ -5218,10 +5218,18 @@ void OBSBasicSettings::FillSimpleRecordingValues()
...
 
-	ui->simpleOutRecEncoder->addItem(ENCODER_STR("Software"),
-					 QString(SIMPLE_ENCODER_X264));
-	ui->simpleOutRecEncoder->addItem(ENCODER_STR("SoftwareLowCPU"),
-					 QString(SIMPLE_ENCODER_X264_LOWCPU));
+	if (!EncoderAvailable("obs_x264"))
+		ui->simpleOutRecEncoder->addItem(
+			ENCODER_STR("Software.OpenH264.H264"),
+			QString(SIMPLE_ENCODER_OPENH264));
+	if (EncoderAvailable("obs_x264")) {
```
tytan652, line 5225:
```suggestion
	else {
```

tytan652 (changes requested):

UI/window-basic-main.cpp
tytan652: It's nitpicking, let it unresolved until a another person give his opinion.

Conan-Kudo: Huh, TIL that GitHub won't run CI if there are unresolved conversations if "all conversations must be resolved before merging" is set.

tytan652: No, it prevents merging but CI runs are not unlimited, so old runs need to finish before running newer.

plugins/obs-ffmpeg/obs-ffmpeg-h264.c
```diff
@@ -0,0 +1,237 @@
...
+	case VIDEO_FORMAT_I420:
+	case VIDEO_FORMAT_NV12:
+		break;
+	default:; // Make the compiler do the right thing
+		const char *const text =
+			obs_module_text("H264.UnsupportedVideoFormat");
+		obs_encoder_set_last_error(encoder, text);
+		blog(LOG_ERROR, "[H.264 encoder] %s", text);
+		return NULL;
+	}
```
derrod, line 133: This should allow all 8-bit formats, OBS will handle the conversion to I420.

derrod (changes requested):
As far as I could tell this will also still display the x264 presets in simple mode, which it shouldn't.

Conan-Kudo:
> As far as I could tell this will also still display the x264 presets in simple mode, which it shouldn't.

@derrod I've disabled the preset field for OpenH264, thanks for catching that.

Lain-B:
Hey @Conan-Kudo, remind me to come back to this after we make our v30 release branch. Sorry about the delay

PatTheMav:
I agree with @tytan652 that the PRa has its selection logic backwards: It installs OpenH264 as a fallback for x264, but x264 is a core module and the former is a very distro-specific option.

As such the PR needs to add OpenH264 as an optional codec and check for its availability and still fall back to x264 as the default.

Conan-Kudo:
> As such the PR needs to add OpenH264 as an optional codec and check for its availability and still fall back to x264 as the default.

I added it this way because OBS crashes the other way because there will not be x264 in the first place in Fedora's build unless you install the plugin from RPM Fusion.

Conan-Kudo:
And ffmpeg is a "core module" too. Both are always present in upstream OBS builds, it depends on detection logic and whether ffmpeg is compiled with OpenH264 support.

PatTheMav:
x264 is the built-in fallback and the app is structured around that.

The better way to solve this would be for Fedora to patch `obs-x264` to use OpenH264 so that the program is no wiser.

(Also I’m unsure as to whether that should be implemented in `obs-x264` itself with a specific check for Fedora, because I’m not happy about distro-specific workarounds while we have to maintain support for 3 operating systems already.)

Afaik we officially support Ubuntu and Flatpak and any distro-specific changes need to be made by distributors while packaging the app.

Conan-Kudo:
`obs-x264` is libx264 specific, an equivalent for OpenH264 would be to write a whole new plugin for libopenh264. I used ffmpeg because it already has an interface for this, we prefer OpenH264 to be accessed through ffmpeg anyway, and it would relatively straightforward to wire in, based on how AV1 was done.

PatTheMav:
Again, that will not work because you are creating a fallback for a fallback, you are not even checking if ffmpeg supports the encoder at runtime before making that decision and with that you also make macOS and Windows fall back to OpenH264 if x264 is not found.

It creates additional code complexity and maintenance burden for an edge case on a single distribution and IMO that’s not a good deal.

PatTheMav:
We can add generic support for OpenH264 if it is available (as in "checked if available via FFmpeg" because we do not ship FFmpeg with OpenH264 support on Windows and macOS as we do not support that codec at all), and that availability will probably be limited to Linux only and depends on the distribution-specific variant of FFmpeg available. If that check is positive, it can be added to the list of available encoders. 

But distribution-specific fixes need to be implemented/patched by distributors. Adding a global (as in: affecting all platforms) fallback on OpenH264 if the _actual_ fallback `x264` (which is commonly shipped by OBS Studio itself or expected to be available on the platform) is not a solution IMO.

Upstream distributors can then handle the added complexity of adding an OpenH264 fallback _if_ it is available at all and abort the OBS Studio launch if it cannot be found.

Also AV1 is not a "baseline/fallback" encoder, so the comparison with its implementation is not applicable.

derrod:
It would probably be sensible to split the encoder addition and the UI changes into separate PRs so we can move things along. The former is probably fine (with some minor code review nits).

Conan-Kudo:
I've split out the first half of this pull request as #10135, per @derrod's request.

DemiMarie:
> But distribution-specific fixes need to be implemented/patched by distributors. Adding a global (as in: affecting all platforms) fallback on OpenH264 if the _actual_ fallback `x264` (which is commonly shipped by OBS Studio itself or expected to be available on the platform) is not a solution IMO.

On desktop Linux, OpenH264 _is_ the worst-case fallback.  Software patent restrictions mean that it is the only H.264 codec that all distros can redistribute.

UI/data/locale/en-US.ini
```diff
@@ -1034,7 +1034,8 @@ Basic.Settings.Output.Simple.Warn.Encoder="Warning: Recording with a software en
 Basic.Settings.Output.Simple.Warn.Lossless="Warning: Lossless quality generates tremendously large file sizes! Lossless quality can use upward of 7 gigabytes of disk space per minute at high resolutions and framerates. Lossless is not recommended for long recordings unless you have a very large amount of disk space available. Replay buffer is unavailable when using lossless quality."
 Basic.Settings.Output.Simple.Warn.Lossless.Msg="Are you sure you want to use lossless quality?"
 Basic.Settings.Output.Simple.Warn.Lossless.Title="Lossless quality warning!"
-Basic.Settings.Output.Simple.Encoder.Software="Software (x264)"
+Basic.Settings.Output.Simple.Encoder.Software.OpenH264.H264="Software (OpenH264)"
+Basic.Settings.Output.Simple.Encoder.Software.X264.H264="Software (x264)"
```
tytan652, line 1038: Avoid having to re-translate because the key is changed, plus OpenH264 will only provide H264.
```suggestion
Basic.Settings.Output.Simple.Encoder.Software.OpenH264="Software (OpenH264)"
Basic.Settings.Output.Simple.Encoder.Software.X264="Software (x264)"
```

UI/data/locale/en-US.ini
```diff
@@ -1045,7 +1046,7 @@ Basic.Settings.Output.Simple.Encoder.Hardware.NVENC.AV1="Hardware (NVENC, AV1)"
 Basic.Settings.Output.Simple.Encoder.Hardware.NVENC.HEVC="Hardware (NVENC, HEVC)"
 Basic.Settings.Output.Simple.Encoder.Hardware.Apple.H264="Hardware (Apple, H.264)"
 Basic.Settings.Output.Simple.Encoder.Hardware.Apple.HEVC="Hardware (Apple, HEVC)"
-Basic.Settings.Output.Simple.Encoder.SoftwareLowCPU="Software (x264 low CPU usage preset, increases file size)"
+Basic.Settings.Output.Simple.Encoder.SoftwareLowCPU.X264.H264="Software (x264 low CPU usage preset, increases file size)"
```
tytan652, line 1049: Avoid having to re-translate because the key is changed.
```suggestion
Basic.Settings.Output.Simple.Encoder.SoftwareLowCPU="Software (x264 low CPU usage preset, increases file size)"
```

UI/window-basic-auto-config-test.cpp
```diff
@@ -981,7 +994,8 @@ void AutoConfigTestPage::TestRecordingEncoderThread()
 }
 
 #define ENCODER_TEXT(x) "Basic.Settings.Output.Simple.Encoder." x
-#define ENCODER_SOFTWARE ENCODER_TEXT("Software")
+#define ENCODER_OPENH264 ENCODER_TEXT("Software.OpenH264.H264")
+#define ENCODER_X264 ENCODER_TEXT("Software.X264.H264")
```
tytan652, line 998:
```suggestion
#define ENCODER_OPENH264 ENCODER_TEXT("Software.OpenH264")
#define ENCODER_X264 ENCODER_TEXT("Software")
```

UI/window-basic-settings-stream.cpp
```diff
@@ -1663,7 +1667,11 @@ void OBSBasicSettings::ResetEncoders(bool streamOnly)
 
 #define ENCODER_STR(str) QTStr("Basic.Settings.Output.Simple.Encoder." str)
 
-	ui->simpleOutStrEncoder->addItem(ENCODER_STR("Software"), QString(SIMPLE_ENCODER_X264));
+	if (!service_supports_encoder(vcodecs, "obs_x264"))
+		ui->simpleOutStrEncoder->addItem(ENCODER_STR("Software.OpenH264.H264"),
+						 QString(SIMPLE_ENCODER_OPENH264));
+	else
+		ui->simpleOutStrEncoder->addItem(ENCODER_STR("Software.X264.H264"), QString(SIMPLE_ENCODER_X264));
```
tytan652, line 1674:
```suggestion
	if (!service_supports_encoder(vcodecs, "obs_x264"))
		ui->simpleOutStrEncoder->addItem(ENCODER_STR("Software.OpenH264"),
						 QString(SIMPLE_ENCODER_OPENH264));
	else
		ui->simpleOutStrEncoder->addItem(ENCODER_STR("Software"), QString(SIMPLE_ENCODER_X264));
```

UI/window-basic-settings.cpp
```diff
@@ -4735,8 +4740,14 @@ void OBSBasicSettings::FillSimpleRecordingValues()
...
 
-	ui->simpleOutRecEncoder->addItem(ENCODER_STR("Software"), QString(SIMPLE_ENCODER_X264));
-	ui->simpleOutRecEncoder->addItem(ENCODER_STR("SoftwareLowCPU"), QString(SIMPLE_ENCODER_X264_LOWCPU));
+	if (!EncoderAvailable("obs_x264"))
+		ui->simpleOutRecEncoder->addItem(ENCODER_STR("Software.OpenH264.H264"),
+						 QString(SIMPLE_ENCODER_OPENH264));
+	else {
+		ui->simpleOutRecEncoder->addItem(ENCODER_STR("Software.X264.H264"), QString(SIMPLE_ENCODER_X264));
+		ui->simpleOutRecEncoder->addItem(ENCODER_STR("SoftwareLowCPU.X264.H264"),
+						 QString(SIMPLE_ENCODER_X264_LOWCPU));
```
tytan652, line 4749:
```suggestion
	if (!EncoderAvailable("obs_x264"))
		ui->simpleOutRecEncoder->addItem(ENCODER_STR("Software.OpenH264"),
						 QString(SIMPLE_ENCODER_OPENH264));
	else {
		ui->simpleOutRecEncoder->addItem(ENCODER_STR("Software"), QString(SIMPLE_ENCODER_X264));
		ui->simpleOutRecEncoder->addItem(ENCODER_STR("SoftwareLowCPU"),
						 QString(SIMPLE_ENCODER_X264_LOWCPU));
```

tytan652 (changes requested):
Translation nitpik that covers both PR.

tytan652 (changes requested):

Conan-Kudo:
This pull request has been rebased on top of current master and has the remaining commits to merge (frontend commits).

Conan-Kudo:
While this doesn't have a direct dependency on @asahilina's #12326, it does benefit from it.

Lain-B:
Just briefly looking at it, it seems pretty trivial now that everything else has been merged. Anyone else want to give further feedback?

Fenrirthviti:
For the fallback part of this, I would prefer if we just removed the entire fallback system in general in favor of a better error reporting that the encoder is missing. I don't want to make our already tenuous and fragile fallback that... really doesn't make sense anymore, more complex.

Fenrirthviti:
Also, can you please explain why we're intrinsically linking OpenH264's existence to x264 in the first place?

I would prefer that if the encoder is available, it just shows up. If it's not, it doesn't. I'm not sure what value there is in NOT displaying OpenH264 if x264 is available instead of just showing them both?

Conan-Kudo:
I was asked last time around to hide it if x264 is available. If we don't need that anymore, I can drop the commit that does that.

Fenrirthviti:
Thanks, found the comment in one of the old reviews. That is not something we'd prefer, it can be removed, yes.

Conan-Kudo:
I dropped it and rebased onto current master.

Conan-Kudo:
> For the fallback part of this, I would prefer if we just removed the entire fallback system in general in favor of a better error reporting that the encoder is missing. I don't want to make our already tenuous and fragile fallback that... really doesn't make sense anymore, more complex.

I would prefer we do this separately. I agree with you that the fallback logic is crufty and fragile, but purging it is currently beyond my capabilities. That said, @asahilina has taken the first steps to allow eliminating it with #12326.

Fenrirthviti:
I agree that it's a separate change, yes, and outside scope of what this PR is attempting to accomplish. 

We're doing some internal checking of the relevant code paths to determine if this PR is a change we're willing to accept short term, and if it is not, have a reasonable alternative proposal.

PatTheMav:
Given that the change seems to be mostly held up by the topic of how to properly handle the fallback, I did a little dive into our UI and settings code to check how deeply integrated the fallback selection actually is, and the answer is: "Very".

Without going into too much detail here, there are many pieces in the settings code that depend on a valid encoder being set as the "current" encoder at all times, and there are further parts of OBS Studio that require those parts being available as well, etc.

---

So my current thinking (after some deliberation) is that I still don't like `openh264` becoming a de-facto global fallback if `x264` isn't found, because it would only ever really be a working(!) fallback on Fedora systems.

So I'd propose that instead of this (pseudocode):
```
if (some encoder) {
   use some encoder
} else if (some other encoder) {
   use some other encoder
} else if (x264) {
   use x264
} else {
   use openh264
}
```
we'd do this:
```
if (some encoder) {
   use some encoder
} else if (some other encoder) {
   use some other encoder
} else {
   // Time to find a fallback
   if (x264) {
      use x264
   } else if (openh264) {
      use openh264
   } else {
      gracefully abort OBS with an error message because the app is in an unworkable state
   }
}
```
For one it decouples x264 from being the "de-facto" fallback encoder, but instead demotes it to a "fallback option" together with open264. Either would be fine, one of them _has_ to be present. And the code just has to ensure that _some_ fallback encoder is provided.

This also means that the code for simple mode output cannot just add "x264" as the very first encoder, but has to check for availability just like it has to do for all other encoders.

As far as the function of the encoder drop down is concerned, I agree that it requires a fix to allow people to change a possibly broken state, but wouldn't hoist that responsibility onto this PR.

PR #8535, opened 2023-03-26
libobs: Fix VIDEO_FORMAT_V210 shader for GLSL

norihiro:
### Description

Fixed these errors reported by GLSL.

- Type cast `(uint)pos.x` is not accepted by GLSL. It should be `uint(pos.x)`.
- GLSL is strictly compare types so that an implicit arithmetic operation with `int` and `uint` are not allowed.
- It was passing `float4` type variable to `float2` argument.

### Motivation and Context

OBS says error below and failed to initialize video.
```plain
error: Error compiling shader:
0:49(16): error: syntax error, unexpected ')', expecting '('
```

### How Has This Been Tested?

OS: Fedora 36
Hardware: AMD Radeon RX 6600 + Intel i7-7700.

### Types of changes

### Checklist:

  - Only tested on Linux. It should be tested with HLSL on Windows.

### DIFF
--- a/libobs/data/format_conversion.effect
+++ b/libobs/data/format_conversion.effect
@@ -927,56 +927,56 @@ float4 PSP010_HLG_2020_709_Reverse(VertTexPos frag_in) : TARGET
 
 float3 compute_v210_reverse(float2 pos)
 {
-	uint x = (uint)pos.x;
-	uint packed_x = x % 6;
-	uint base_x = (x / 6) * 4;
+	uint x = uint(pos.x);
+	uint packed_x = x % 6u;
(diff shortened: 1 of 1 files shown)


norihiro:
If it's ok to change `uint` to `int`, this commit also works for GLSL 3ce6b4da4d00fb9922a851fdbb618a153ff18ed8.

jp9000:
the unsigned integer literals appear to be supported in HLSL, so it seems to work with both HLSL and GLSL just fine.

jpark37:
Late to the party, but looks good. Thanks!

PR #8539, opened 2023-03-26
UI: Redirect Qt log messages to OBS log

gxalpha:
### Description

Installs a message handler for Qt that redirects Qt log messages to the OBS log, so that warnings thrown by Qt will be visible there.

### Motivation and Context

Developers sometimes aren't aware of warnings they're causing, since from time to time new ones appear. This might make them (and fellow developers) more aware.
It also makes it much easier to debug them since you can just breakpoint the blog statement, instead of having to step through the code and to see where stuff is going wrong.

### How Has This Been Tested?

macOS 13.2.1, Qt dev branch.
Tested that the ever-present `QWidget::setTabOrder: 'first' and 'second' must be in the same window` that I haven't gotten around to fixing yet now appears as a warning in the OBS log.

I'm also thinking maybe the message should get prefixed, but I'm not sure which prefix would be fitting.

### Types of changes

### Checklist:

### DIFF
--- a/UI/obs-app.cpp
+++ b/UI/obs-app.cpp
@@ -2452,6 +2452,32 @@ static int run_program(fstream &logFile, int argc, char *argv[])
 		if (!created_log)
 			create_log_file(logFile);
 
+		qInstallMessageHandler([](QtMsgType type,
+					  const QMessageLogContext &,
+					  const QString &message) {
+			switch (type) {
+#ifdef _DEBUG
+			case QtDebugMsg:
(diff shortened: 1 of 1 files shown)


jp9000:
I'd like to exclude debug/info I think. I'd prefer to only log warnings/errors.

derrod:
What about adding a `#ifdef _DBEUG` around info/debug so they're still visible in debug builds?

gxalpha:
Added an `ifdef _DEBUG` check to debug and info messages as suggested above.

PR #8547, opened 2023-03-27
cmake: Ignore all Qt darwin permission plugins

gxalpha:
### Description

Adds all variations of QDarwinPermissionPlugins to the ignorelist in cmake.
This already existed for QDarwinBluetoothPermissionPlugin but it looks like I didn't actually confirm that it generated correctly.

### Motivation and Context

Want OBS to work with Qt 6.5

### How Has This Been Tested?

macOS 13.2.1
The project *generates* correctly now with both Qt 6.5.0-rc and dev branch.

There still appears to be an issue with the 3.0 cmake, because no Qt versions apart from obs-deps appear to work.
I installed 6.4.3 and 6.5.0-rc from the Qt online installer, as well as building the dev branch, and while generating the project worked with all of them, they also failed to run (can't find QtDBus).
*However* using obs-deps Qt 6.4.3 works (while 6.4.3 from the online installer doesn't), so presumably using 6.5.0 from obs-deps would also work (with this change applied).

### Types of changes

### Checklist:

### DIFF
--- a/cmake/common/helpers_common.cmake
+++ b/cmake/common/helpers_common.cmake
@@ -251,8 +251,7 @@ function(find_dependencies)
         else()
           continue()
         endif()
-      elseif(library MATCHES "\\$<.*Qt6::EntryPointPrivate>" OR library MATCHES
-                                                                "\\$<.*Qt6::QDarwinBluetoothPermissionPlugin>")
+      elseif(library MATCHES "\\$<.*Qt6::EntryPointPrivate>" OR library MATCHES "\\$<.*Qt6::QDarwin.+PermissionPlugin>")
         # Known Qt6-specific generator expression, ignored.
         continue()
       else()


cmake/common/helpers_common.cmake
```diff
@@ -251,8 +251,9 @@ function(find_dependencies)
         else()
           continue()
         endif()
-      elseif(library MATCHES "\\$<.*Qt6::EntryPointPrivate>" OR library MATCHES
-                                                                "\\$<.*Qt6::QDarwinBluetoothPermissionPlugin>")
+      elseif(library MATCHES "\\$<.*Qt6::EntryPointPrivate>"
+             OR library MATCHES
+                "\\$<.*Qt6::QDarwin(Bluetooth|Calendar|Camera|Contacts|Location|Microphone)PermissionPlugin>")
```
PatTheMav, line 256: It's probably safe to use `Qt6::Darwin.+PermissionPlugin` because that's specific enough.

gxalpha: Changed

PatTheMav (approved):

PR #8557, opened 2023-03-28
UI: Improve AV1 fallbacks while changing services

tytan652:
### Description

This makes AV1 encoders fallback to their H264 counterpart rather than just falling back to x264 while changing services.

AV1 streaming was still not a thing, now it is. It just a avoids falling back directly to x264 in some service change scenarios.

### Motivation and Context

Better fallback, until we improve settings by avoiding hardcode.

### How Has This Been Tested?

I don't have AV1 HW encoders, sorry.

### Types of changes

### Checklist:

### DIFF
--- a/UI/window-basic-settings-stream.cpp
+++ b/UI/window-basic-settings-stream.cpp
@@ -1354,12 +1354,14 @@ bool OBSBasicSettings::ServiceAndACodecCompatible()
 /* we really need a way to find fallbacks in a less hardcoded way. maybe. */
 static QString get_adv_fallback(const QString &enc)
 {
-	if (enc == "jim_hevc_nvenc")
+	if (enc == "jim_hevc_nvenc" || enc == "jim_av1_nvenc")
 		return "jim_nvenc";
-	if (enc == "h265_texture_amf")
+	if (enc == "h265_texture_amf" || enc == "av1_texture_amf")
 		return "h264_texture_amf";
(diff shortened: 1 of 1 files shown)


RytoEX:
What is the current (buggy) behavior? What is the expected behavior after this change? Trying to understand this so that I don't draw the wrong conclusions.

tytan652:
When changing protocol and the encoder happens to be incompatible the codec is changed for a H264 one.
But the actual functions don't manage falling back all HW AV1 to their H264 counterpart but x264.

Yeah, the PR description is not helpful at all sorry.

Conan-Kudo:
Shouldn't we fallback to a software AV1 encoder instead? It seems weird to switch from AV1 to H.264...

tytan652:
> Shouldn't we fallback to a software AV1 encoder instead?

This is a bad idea, while x264 can be done by most CPU, AOM and SVT AV1 are another story.

Like I said earlier, I did a pretty bad PR description, but the scope of the PR is only about function that are used if we switch from a service that supports AV1/HEVC to one that does not.

> It seems weird to switch from AV1 to H.264...

It is not weird if the protocol in use does not support AV1.

And…
> Better fallback, until we improve settings by avoiding hardcode.

tytan652:
@RytoEX, I changed PR/commit title. I hope it is now less confusing.

RytoEX (approved):
Looks fine at a glance.

PR #8563, opened 2023-03-28
decklink: Query for preroll frame count

jpark37:
### Description
Three is recommended though, so use at least that many.

Also add commit to stop swscale from kicking in because of color space mismatch add chewing up CPU without actually fixing the colors.

### Motivation and Context
Correctness.

### How Has This Been Tested?
1080p and 2160p tested.

### Types of changes

### Checklist:

### DIFF
--- a/plugins/decklink/decklink-device.hpp
+++ b/plugins/decklink/decklink-device.hpp
@@ -21,6 +21,7 @@ class DeckLinkDevice {
 	decklink_bool_t supportsInternalKeyer = false;
 	int64_t subDeviceIndex = 0;
 	int64_t numSubDevices = 0;
+	int64_t minimumPrerollFrames = 3;
 	int64_t supportedVideoInputConnections = -1;
 	int64_t supportedVideoOutputConnections = -1;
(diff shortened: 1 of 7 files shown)


plugins/decklink/decklink-device.hpp
```diff
@@ -21,6 +21,7 @@ class DeckLinkDevice {
 	decklink_bool_t supportsInternalKeyer = false;
 	int64_t subDeviceIndex = 0;
 	int64_t numSubDevices = 0;
+	int64_t minimumPrerollFrames = 3;
```
tt2468, line 24: This seems redundant given the FAILED() check above. Otherwise, looks good to me

jpark37: Yeah, I thought about it, and I figured it was better to have a reasonable value here "just in case."

tt2468 (approved):

jpark37:
Bolted on another commit to stop unnecessary swscale, which was causing the 4K slowdown I mentioned in the testing section (which has been removed).

PR #8582, opened 2023-03-29
obs-ffmpeg: Disable b-frames in FFmpeg NVENC (for now)

derrod:
### Description

Disables b-frames when FFmpeg NVENC is used (e.g. as fallback or on Linux).

### Motivation and Context

B-frames result in wrong timestamps with FFmpeg NVENC, until we know how to fix that (and if that's even on our end) force them to be disabled as to not produce files with decoding issues.

### How Has This Been Tested?

Recorded file in I444 with b-frames set to 2, recording had no b-frames and log noted they were disabled.

### Types of changes

### Checklist:

### DIFF
--- a/plugins/obs-ffmpeg/obs-ffmpeg-nvenc.c
+++ b/plugins/obs-ffmpeg/obs-ffmpeg-nvenc.c
@@ -185,6 +185,14 @@ static bool nvenc_update(struct nvenc_encoder *enc, obs_data_t *settings,
 
 	set_psycho_aq(enc, psycho_aq);
 
+	/* B-Frames in FFmpeg NVENC result in pts/dts being wrong,
+	 * disable them for now. */
+	if (bf != 0) {
+		warn("B-Frames enabled, disabling as they are currently broken"
+		     " in FFmpeg NVENC.");
+		bf = 0;
+	}
+
 	enc->ffve.context->max_b_frames = bf;
 
 	const char *ffmpeg_opts = obs_data_get_string(settings, "ffmpeg_opts");


tt2468:
While timestamps appear to be broken no matter what with ffmpeg nvenc, streaming protocols like RTMP and SRT don't seem to expose any playback issues for users. If we can perhaps enforce disabling b-frames via ui code for the record output, that seems like a better approach.

derrod:
It seems that this line in FFmpeg is the issue:
https://github.com/FFmpeg/FFmpeg/blob/c221036502751d7a6e1403a9868c60f1ff357963/libavcodec/nvenc.c#L2269
If we change it from
`pkt->dts = dts - FFMAX(ctx->encode_config.frameIntervalP - 1, 0) * FFMAX(avctx->ticks_per_frame, 1);`
to
`pkt->dts = dts - FFMAX(ctx->encode_config.frameIntervalP - 1, 0) * FFMAX(avctx->time_base.num * avctx->ticks_per_frame, 1);`
The calculation works correctly.

An alternative hack found by @RytoEX is to manually set the frame rate and `ticks_per_frame`, we could potentially ifdef that for older versions of FFmepg on platforms where we don't/can't patch FFmpeg.

derrod:
Closing in favor of #8622 as its a proper workaround for this issue.

obs-deps will soon have a patch to fix FFmpeg's DTS calculation: https://github.com/obsproject/obs-deps/pull/177
This patch will hopefully soon be upstreamed to FFmpeg itself as well.

PR #8587, opened 2023-03-30
linux-pipewire: Use premultiplied alpha

asahilina:
### Description
Pipewire desktop portal capture assumes premultiplied alpha, so make sure to use the right blend equation to make translucent pixels work properly.

This is still broken for emissive pixels (alpha < color) since OBS seems to unpremultiply at some point during blending, but it works properly for translucent pixels (emissive pixel support requires an end-to-end premultiplied pipeline).

There is actually partial support in PipeWire for specifying whether the video format is premultiplied or not, but it seems to be incomplete and nobody uses it. I opened an issue [here](https://gitlab.freedesktop.org/pipewire/pipewire/-/issues/3126). In the future if this is properly implemented as a property, OBS could support both formats.

### Motivation and Context

Makes transparent window capture with PipeWire/Wayland work properly. Without this change, translucent pixels are too dark.

### How Has This Been Tested?

Tested on an Apple M2 running Asahi Linux and KDE Plasma Wayland.

### Types of changes

### Checklist:

### DIFF
--- a/plugins/linux-pipewire/pipewire.c
+++ b/plugins/linux-pipewire/pipewire.c
@@ -1090,6 +1090,18 @@ void obs_pipewire_video_render(obs_pipewire *obs_pw, gs_effect_t *effect)
 	rotated = push_rotation(obs_pw);
 
 	flip = get_buffer_flip(obs_pw);
+
+	/* There is a SPA_VIDEO_FLAG_PREMULTIPLIED_ALPHA flag, but it does not
+	 * seem to be fully implemented nor ever set. Just assume premultiplied
+	 * always, which seems to be the default convention.
+	 *
+	 * See https://gitlab.freedesktop.org/pipewire/pipewire/-/issues/3126
+	 *
+	 * The cursor bitmap alpha mode is also not specified, and the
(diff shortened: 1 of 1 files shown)


jp9000:
Thanks for the PR!

Also for @GeorgesStavracas -- if this is ever changed in pipewire, this will probably need to be reverted, but I'm just going to merge for the time being because it's a super trivial fix.

asahilina:
Just a comment: I've noticed some buggy behavior where it looks like PipeWire sometimes does not negotiate a proper alpha channel (at least with KWin). That leads to sometimes getting the texture sampling as alpha=1.0 (which leads to the expected behavior with no alpha), but sometimes you get alpha=0.0 (which with this change ends up doing weird additive blending). Re-selecting the captured window a few times eventually makes it work properly.

I'm not sure exactly where the problem is yet (it could even be in our graphics drivers), but I wanted to leave a note here in case someone else runs into this and/or notices the relationship to this commit. I don't *think* the bug is here, but this change does affect the behavior in the alpha=0.0 case regardless.

PR #8597, opened 2023-03-30
cmake: Sort large target_sources lists alphabetically

gxalpha:
### Description

The `# cmake-format: sortable` lets cmake know that a list can be sorted. By default `autosort` only sorts `add_library` `add_executable` (per the [documentation](https://cmake-format.readthedocs.io/en/latest/format-features.html#sort-argument-lists)).

In the UI, adds the `sortable`  hint to large `target_sources` lists to have them be sorted alphabetically.

In libobs, just sort the lists. Unfortunately the hint doesn't differentiate between files and directories and mixes them, making it quite ugly. As such, I added the hint, undid some of the changes manually and removed it again. I don't think for libobs it's as bad not to have it as the lists aren't that volatile.

### Motivation and Context

Wanted to add a new file and wondered where to put it, noticing that some lists were only close to being sorted.

### How Has This Been Tested?

macOS 13.2.1.
Ran cmake and built successfully.

### Types of changes

### Checklist:

### DIFF
--- a/libobs/CMakeLists.txt
+++ b/libobs/CMakeLists.txt
@@ -19,12 +19,9 @@ add_subdirectory("${CMAKE_SOURCE_DIR}/deps/uthash" "${CMAKE_BINARY_DIR}/deps/uth
 
 target_sources(
   libobs
-  PRIVATE obs.c
-          obs.h
-          obs.hpp
-          obs-audio.c
-          obs-audio-controls.c
+  PRIVATE obs-audio-controls.c
           obs-audio-controls.h
+          obs-audio.c
           obs-avc.c
           obs-avc.h
           obs-data.c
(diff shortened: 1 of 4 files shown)


RytoEX (approved):
Seems okay at a glance. Would like @PatTheMav 's opinion.

libobs/CMakeLists.txt
```diff
@@ -19,12 +19,9 @@ add_subdirectory("${CMAKE_SOURCE_DIR}/deps/uthash" "${CMAKE_BINARY_DIR}/deps/uth
 
 target_sources(
   libobs
-  PRIVATE obs.c
-          obs.h
-          obs.hpp
-          obs-audio.c
-          obs-audio-controls.c
+  PRIVATE obs-audio-controls.c
```
PatTheMav, line 22: What's happening here? The sortable switch seems to have been applied here as well.

gxalpha: Yesn't. The switch has been applied, but it also messes up directories and conditionals by mixing them together with normal files, which doesn't make much sense.

For example:
```
a.c
z.c
b.c
myfolder/whatever.c
$<$<BOOL:${SOME_VALUE}>:conditional.c>
```
would become
```
$<$<BOOL:${SOME_VALUE}>:conditional.c>
a.c
b.c
myfolder/whatever.c
z.c
```
which isn't desirable.

So as mentioned in the description, I added the flag, then undid the directory changes manually (and also had to of course remove the flag again). I don't think for libobs it's as bad not to have it as the lists aren't that volatile as in the UI, where files get added rather frequently in comparison.

PatTheMav (approved):

PR #8371, opened 2023-02-28
UI: Remove platform string from title bar

cg2121:
### Description
This cleans up the title bar text.

### Motivation and Context
Doesn't seem necessary to have in the title bar.

### How Has This Been Tested?
Looked at title bar.

### Types of changes
- Code cleanup (non-breaking change which makes code smaller or more readable)

### Checklist:
- [x] My code has been run through [clang-format](https://github.com/obsproject/obs-studio/blob/master/.clang-format).
- [x] I have read the [**contributing** document](https://github.com/obsproject/obs-studio/blob/master/CONTRIBUTING.rst).
- [x] My code is not on the master branch.
- [x] The code has been tested.
- [x] All commit messages are properly formatted and commits squashed where appropriate.
- [x] I have included updates to all appropriate documentation.

PR #8600, opened 2023-03-31
Revert "UI: Remove bitness strings"

RytoEX:
### Description

This reverts commit 64d4ae0106ff9c61978bdb7b55dc35679edb5a13.

### Motivation and Context

@WizardCM 's comment on #8267 indicated that it would be preferable to remove the display of extra build information from the application window's title bar (platform and bitness) but that we should keep it in the log and the About dialog. Removing this string caused the log analyzer to throw false positives for the "32-bit OBS on 64-bit OS" warning (see https://github.com/obsproject/loganalyzer/issues/122 and https://github.com/obsproject/loganalyzer/pull/124). Per discussion with @Fenrirthviti , I've reverted the changes in #8267 accounting for the new `platform` bool in #8371. The string will still be absent from the application window's title bar. We can reconsider removing the strings from the log and About dialog for a later release and also ensure that any future changes are well-tested with the log analyzer.

### How Has This Been Tested?

Tested the build locally.

### Types of changes

### Checklist:

### DIFF
--- a/UI/obs-app.cpp
+++ b/UI/obs-app.cpp
@@ -1754,6 +1754,11 @@ string OBSApp::GetVersionString(bool platform) const
 	if (platform) {
 		ver << " (";
 #ifdef _WIN32
+		if (sizeof(void *) == 8)
+			ver << "64-bit, ";
+		else
+			ver << "32-bit, ";
+
 		ver << "windows)";
 #elif __APPLE__
 		ver << "mac)";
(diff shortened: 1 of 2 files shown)


WizardCM:
I personally don't think this is necessary. The log analyzer PR linked does a good enough job in solving the problem. Any build over 29 can safely be assumed to be 64bit, and if the log contains 32bit then it's an old enough build that is 1) 32bit and 2) no longer supported by us anyway.

Fenrirthviti:
Just to state for posterity, since I didn't get a chance before the other PR was merged, I'm in favor of leaving all the bitness strings in place. I don't feel removing them does anything but remove helpful, at-a-glance information.

PR #8602, opened 2023-03-31
Fix Codec Compatibilty Check

derrod:
### Description

Unfortunately FFmpeg's check is too unreliable, so just use a hardcoded list instead *sigh*.

### Motivation and Context

Don't want people to be able to select invalid combnations (e.g. AV1 in MOV).

### How Has This Been Tested?

Checked that correct codecs are now marked as invalid.

### Types of changes

### Checklist:

### DIFF
--- a/UI/window-basic-settings.cpp
+++ b/UI/window-basic-settings.cpp
@@ -23,6 +23,7 @@
 #include <initializer_list>
 #include <sstream>
 #include <unordered_map>
+#include <unordered_set>
 #include <QCompleter>
 #include <QGuiApplication>
 #include <QLineEdit>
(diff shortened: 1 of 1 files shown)


Conan-Kudo:
Well, at least it'll be easy to enable AV1 for MPEG-TS once that's in... 😅

derrod:
On second thought, this should fall back to the ffmpeg check for codecs that we don't ship to avoid breaking some third-party plugins that add additional encoders.

> Not necessarily asking for this to be solved in this PR, but would it make sense to think about consolidating this with the code in obs-ffmpeg that handles some of this?

Not sure what you're referring to.

Conan-Kudo:
> > Not necessarily asking for this to be solved in this PR, but would it make sense to think about consolidating this with the code in obs-ffmpeg that handles some of this?
> 
> Not sure what you're referring to.

There are codec filters in obs-ffmpeg-hls-mux.c and obs-ffmpeg-mpegts.c as part of obs-ffmpeg.

energizerfellow:
Neither [MPEG-TS](https://www.iso.org/standard/83239.html) nor [HLS](https://developer.apple.com/documentation/http_live_streaming/http_live_streaming_hls_authoring_specification_for_apple_devices) appear to support Opus in their spec sheets, so Opus should be probably be removed for those.

While TS doesn't support LPCM, [M2TS supports LPCM (and AAC)](https://exiftool.org/TagNames/M2TS.html) due to Blu-ray (more info on [HD](http://www.disc-group.com/wp-content/uploads/2011/05/Blu-Ray-4-LogicalAudioVisual-867KB.pdf) and [UHD](https://docplayer.net/48882577-White-paper-blu-ray-disc-read-only-format-ultra-hd-blu-ray-audio-visual-application-format-specifications-for-bd-rom-version-3.html)) and can be created by FFmpeg. Adding M2TS should probably be a separate PR, however.

derrod:
Seems like FFmpeg is perfectly happy to mux Opus into HLS but I suspect at least Apple devices won't support that, so I'll remove it.

norihiro:
Nitpick: The commit description has 73 characters in a line. Could you fix it?

Conan-Kudo:
Shouldn't `hevc` be filtered out if `ENABLE_HEVC` is disabled?

derrod:
> Shouldn't `hevc` be filtered out if `ENABLE_HEVC` is disabled?

Doesn't really matter here and would just make things uglier.

Issue #8603, opened 2023-03-31
OBS Crashes when saving a replay with Fragmented MP4 or Fragmented MOV


Breadsniffer:
### Operating System Info

Windows 10

### OBS Studio Version

29.1.0-beta1

### OBS Studio Log URL

https://obsproject.com/logs/ptfygNUUrb5vdQsN

### OBS Studio Crash Log URL

https://obsproject.com/logs/HujtCQVkB5XnLEUx

### Expected Behavior

OBS to not crash when saving a replay with Fragmented MP4 or Fragmented MOV selected

### Current Behavior

OBS Crashes when saving a replay with Fragmented MP4 or Fragmented MOV selected

### Steps to Reproduce

1. Select Fragmented MP4 or Fragmented MOV as recording format
2. Start replay buffer
3. Save a replay
4. It crashes
...

RytoEX:
Could you confirm what encoder settings you had selected? The normal OBS log is cut off before that information is logged.

Breadsniffer:
nvidia nvenc h264

Breadsniffer:
oh I didn't catch what you meant

Rate control = CBR
Bitrate = 35000 Kbps
Keyframe Interval = 0 s
Preset = P7: Slowest
Tuning = High Quality
Multipass mode = Two Passes Quarter Resolution
Profile = high
Look-ahead = off
Psycho Visual Tuning = on
GPU = 0
Max B-frames = 2

PR #8615, opened 2023-04-01
UI: Suppress multiple message boxes

norihiro:
### Description

While a message box is shown and the user has not closed it, another message box won't be shown unless it has a higher severity.

### Motivation and Context

When multiple errors happen and each error triggers a message box, it will create a huge number of popup dialogs.
On XFCE4 on Linux, this will results whole desktop not responding.

There is an issue #8603 reporting OBS crashes on Windows.

### How Has This Been Tested?

OS: Fedora 36

1. Set fmp4 format, enable replay buffer.
2. Start replay buffer
3. Wait for a while
4. Save the replay buffer
5. Try 3-4 several times until a dialog box appears.

### Types of changes

### Checklist:


gxalpha:
I can’t help but think that this is fighting symptoms, not the causes. If we’re showing multiple message boxes at once, something else has already gone wrong (and should, in my opinion, be fixed there instead).

notr1ch:
I don't like this idea, we should ideally never be showing multiple message boxes at once, but if something terrible does happen and two errors occur at once for example, we equally shouldn't ignore one of them.

norihiro:
I will reconsider the behavior.

Possible better implementation would be one or combination of  them
- Filter the same messages like log-filter does.
- Queue the messages and show the dialog once the previous one was closed.

Warchamp7:

PR #8617, opened 2023-04-01
win-capture: Add and tweak apps in compatibility.json

TianQiBuTian:
### Description

Add:
Counter-Strike 2, UWP applications, Gaming Services applications, Microsoft 365(Excel, PowerPoint, Word, Access, OneNote, Outlook, Publisher and 365 software), Adobe After Effects, Adobe Character Animator, Adobe Photoshop, Adobe Premiere Pro, Steam, Epic Games Launcher, Ubisoft Connect, Tencent GUI applications, WeChat and YY.

Tweak:
Change "Electron" name to "Chromium";
Change "%name% cannot be captured via Game Capture." to "%name% cannot be captured using Game Capture.".

Remove:
Chrome and Edge (reason is based on Chromium).

Also add win-capture json in `.editorconfig`
### Motivation and Context

Before tweak: (Edge is not using `Compatibility.WindowCapture.BitBlt`)
QQ截图20230402055155

### How Has This Been Tested?

Local build and use 29.1.0-beta1 game capture.
### Types of changes

### Checklist:


plugins/win-capture/window-capture.c
derrod, line 1: Please split this change out into another commit since it modifies the auomatic WGC list rather than compatibility data.

plugins/win-capture/data/locale/en-US.ini
```diff
@@ -48,6 +48,11 @@ Compatibility.WindowCapture.BitBlt="%name% may not be capturable using the selec
 
 # Specific compatibility messages
 Compatibility.Application.CSGO="CS:GO may require the <code>-allow_third_party_software</code> launch option to use Game Capture."
-Compatibility.Application.ElectronGameCapture="Games built on Electron cannot be captured using Game Capture. Use Window Capture or Display Capture instead."
-Compatibility.Application.ElectronBitBlt="Electron-based applications may not be capturable using the selected Capture Method (BitBlt)."
+Compatibility.Application.ChromiumGameCapture="Games built on Chromium cannot be captured using Game Capture. Use Window Capture or Display Capture instead."
+Compatibility.Application.ChromiumBitBlt="Chromium-based applications may not be capturable using the selected Capture Method (BitBlt)."
+Compatibility.Application.UWPBitBlt="UWP applications may not be capturable using the selected Capture Method (BitBlt)."
+Compatibility.Application.UWPGameCapture="UWP applications cannot be captured using Game Capture. Use Window Capture or Display Capture instead."
```
derrod, line 54: This may be misleading, UWP games can be game captured, but I'm not sure if they have a different window class.

plugins/win-capture/data/locale/en-US.ini
```diff
@@ -48,6 +48,11 @@ Compatibility.WindowCapture.BitBlt="%name% may not be capturable using the selec
...
+Compatibility.Application.GamingServicesBitBlt="Gaming Services applications may not be capturable using the selected Capture Method (BitBlt)."
+Compatibility.Application.GamingServicesGameCapture="Gaming Services applications cannot be captured using Game Capture. Use Window Capture or Display Capture instead."
 Compatibility.Application.Minecraft="If you're having issues capturing Minecraft: Java Edition, check our troubleshooting guide."
+Compatibility.Application.TXGuiBitBlt="Tencent GUI applications may not be capturable using the selected Capture Method (BitBlt)."
```
derrod, line 58: You could use the generic translation strings for those, they do not need special ones.

derrod (changes requested):
Mostly fine, though I do question the need for some of these.

Generally we'd want to keep this list to apps where users often run into issues. I'm not sure how often items like the Tencent apps come up, and I don't think many people want to 
specifically capture e.g. Steam or the Epic Launcher.

Others like PowerPoint and the generic UWP ones are certainly valid however.

Finally, you also need to increment the version in `package.json`.

TianQiBuTian:
OK, I tested UWP games before submitting pr and couldn't use game capture. After downloading the OBS again, you can capture it. My bad.

The Tencent GUI is used on [QQ](https://im.qq.com/) and other Tencent software. Confirm that most software cannot use BitBlt.

And confirm that I need to add a version number? I checked https://obsproject.com/obs2_update/win-capture/v1/package.json status is still not found. And this feature is not currently available in the release version.

derrod:
If you bump the version number now then we can push it independently of the next beta release and make sure before release that pulling the compatibility data updates works as expected.

plugins/win-capture/data/compatibility.json
```diff
@@ -17,8 +17,8 @@
             "url": "https://help.steampowered.com/faqs/view/09A0-4879-4353-EF95#whitelist"
         },
         {
-            "name": "Electron",
-            "translation_key": "Compatibility.Application.ElectronBitBlt",
+            "name": "Chromium",
+            "translation_key": "Compatibility.Application.ChromiumBitBlt",
```
derrod, line 21: Does this actually affect Chromium? Either way I'd prefer if you kept the existing translation key and just updated the text to say "Electron/Chromium" instead.

derrod (changes requested):
Overall looks fine now, just one more thing.

plugins/win-capture/data/compatibility.json
TianQiBuTian: `Electron` includes `Chromium`. `Edge` and `Chrome` are also recognized as `Electron`, but they are not built on `Electron`.
https://wikipedia.org/wiki/Electron_(software_framework)

That's why I changed `Electron` to `Chromium`.

Like Steam, it can actually capture frames, but cannot capture pages. Because their sub windows use `Chromium`.
QQ截图20230402213709

norihiro:
Please fold the commit descriptions at 72-column.

plugins/win-capture/data/compatibility.json
derrod: While Electron is using Chromium, the average user is more familiar with Electron than Chromium. Renaming it to "Electron/Chromium" and retaining the current translation key is probably the best way forward for clarity and translation workflow.

My question was more about whether Chromium/CEF based applications even use the `Chrome_WidgetWin_0` window class.

derrod (approved):
Looks fine to me now, I don't have a Windows machine handy right now so if somebody could confirm that Chrome/Electron still get correctly identified based on window class that would be great.

TianQiBuTian:
I feel a bit strange about displaying Electron in front of only based on Chromium software. Can I swap Electron and Chromium tomorrow?

plugins/win-capture/data/locale/en-US.ini
```diff
@@ -42,12 +42,14 @@ Mode="Mode"
...
 Compatibility.GameCapture.WrongGPU="If the preview is blank, make sure %name% is running on the same GPU as OBS."
 Compatibility.WindowCapture.BitBlt="%name% may not be capturable using the selected Capture Method (BitBlt)."
+Compatibility.WindowCapture.BitBlt.Applications="%name% applications may not be capturable using the selected Capture Method (BitBlt)."
 
 # Specific compatibility messages
 Compatibility.Application.CSGO="CS:GO may require the <code>-allow_third_party_software</code> launch option to use Game Capture."
-Compatibility.Application.ElectronGameCapture="Games built on Electron cannot be captured using Game Capture. Use Window Capture or Display Capture instead."
-Compatibility.Application.ElectronBitBlt="Electron-based applications may not be capturable using the selected Capture Method (BitBlt)."
+Compatibility.Application.ElectronGameCapture="Games built on Electron/Chromium cannot be captured using Game Capture. Use Window Capture or Display Capture instead."
+Compatibility.Application.ElectronBitBlt="Based on Electron/Chromium applications may not be capturable using the selected Capture Method (BitBlt)."
```
derrod, line 54: Actually the grammar here is wrong, it should either be "Applications based on Electron/Chromium" or "Electron/Chromium-based applications".

derrod:
> I feel a bit strange about displaying Electron in front of only based on Chromium software. Can I swap Electron and Chromium tomorrow?

These filters are more likely to be displayed for Electron applications because those are simply more common and we already have an automatic switch to WGC for chrome/edge anyway. But perhaps @Warchamp7 has an opinion here.

TianQiBuTian:
Idea: Can I remove specific message and change to `Games built on %name%`, `Applications based on %name%`?
Perhaps other engines will use this message in the future, so there is no need to add a special message. And there is no need to change the translation key when changing the name.

derrod (approved):
I think this is now mostly fine as-is, would love to get a second opinion from @Warchamp7 on the wording.

Warchamp7:
My only feedback here is to simplify the Chromium/Electron warning to just 'Chromium'

A user is more likely to understand what Chromium means in the context of 'browser-based', and I think anyone who knows what the word Electron means in this context, will understand Chromium as a broader catch all

TianQiBuTian:
Okay, this should be the last change (CS2 has been released).
It has been a long time since the first approval, so please review the code I changed again to see if there is any problem.

derrod (approved):

Warchamp7 (approved):

PR #8623, opened 2023-04-02
mac-capture: Add option to hide OBS windows in desktop capture

PatTheMav:
### Description
Adds the option to hide all windows by OBS.app in desktop capture to avoid a hall of mirrors effect and enable users to change their stream setup without having those steps appear on stream as well.

### Motivation and Context
Implement common use case for streamers that either use desktop capture as a choice (e.g. creative streams that show workflows, or improvisational streams that require streamers to react to their chat) or for technical reasons (e.g. having to capture multiple windows as the same time).

### How Has This Been Tested?
Tested on macOS 13.

### Types of changes

### Checklist:


plugins/mac-capture/mac-screen-capture.m
```diff
@@ -1009,6 +1050,14 @@ static bool content_settings_changed(void *data, obs_properties_t *props,
 		props, "show_hidden_windows",
 		obs_module_text("WindowUtils.ShowHidden"));
 
+	obs_properties_add_bool(props, "show_cursor",
+				obs_module_text("DisplayCapture.ShowCursor"));
+
+	obs_property_t *hide_obs = obs_properties_add_bool(
+		props, "hide_obs", "Hide OBS from capture");
```
gxalpha, line 1057: This will need to be localized

gxalpha (changes requested):

RytoEX:
Probably needs a discussion between @PatTheMav and @gxalpha for macOS platform UX knowledge and @GeorgesStavracas and @Warchamp7 for general UX.

GeorgesStavracas:
This behaviour seems fine, from an interaction perspective. :+1: if the code is fine.

plugins/mac-capture/mac-screen-capture.m
```diff
@@ -1009,6 +1050,14 @@ static bool content_settings_changed(void *data, obs_properties_t *props,
 		props, "show_hidden_windows",
 		obs_module_text("WindowUtils.ShowHidden"));
 
+	obs_properties_add_bool(props, "show_cursor",
+				obs_module_text("DisplayCapture.ShowCursor"));
+
+	obs_property_t *hide_obs = obs_properties_add_bool(
+		props, "hide_obs", obs_module_text("DisplayCapture.HideOBS"));
+	obs_property_set_modified_callback2(hide_obs, content_settings_changed,
+					    sc);
```
gxalpha, line 1059: I think this also needs to be in the `if (sc)` block below to avoid crashes when `sc` is null, e.g. by calling obs_get_source_properties on the source type.

gxalpha (changes requested):
Two comments, looks good otherwise.

As a note, when I tried this a while back with similar looking code I ran into issues with OBS not existing in the `sharable_content.applications` list when a source was created at launch. I wonder if there were changes in OBS that made it get detected earlier or in SCK (or my code being bad), so it might be worth testing this PR on macOS 13.0 to confirm everything works there as well.

plugins/mac-capture/mac-screen-capture.m
PatTheMav: Yeah the modified callbacks shouldn't even be added if there is no valid data object.

PatTheMav:
> Two comments, looks good otherwise.
> 
> As a note, when I tried this a while back with similar looking code I ran into issues with OBS not existing in the `sharable_content.applications` list when a source was created at launch. I wonder if there were changes in OBS that made it get detected earlier or in SCK (or my code being bad), so it might be worth testing this PR on macOS 13.0 to confirm everything works there as well.

I did test it on macOS 13, or do you mean 13.0 specifically? I don't think we should necessarily test for specific minor versions of an OS. If you're on macOS 13, you should be on the latest minor version and that's what we should target.

gxalpha:
I did think think 13.0 specifically, but you’re probably right.

plugins/mac-capture/mac-screen-capture.m
```diff
@@ -1009,7 +1049,19 @@ static bool content_settings_changed(void *data, obs_properties_t *props,
...
+
+	obs_property_t *hide_obs = obs_properties_add_bool(
+		props, "hide_obs", obs_module_text("DisplayCapture.HideOBS"));
+
 	if (sc) {
+		obs_property_set_modified_callback2(
+			capture_type, content_settings_changed, sc);
+
+		obs_property_set_modified_callback2(
+			hide_obs, content_settings_changed, sc);
```
gxalpha, line 1063: I want to apologize for not realizing earlier, but looking at this again I’m not sure it needs a modified_callback in the first place. As far as I can tell (and I haven’t tested this myself yet), the only things that need these callbacks are properties which have a direct effect on other properties, e.g. the capture type changing which other comboboxes are shown, or the "Show hidden windows" toggle changing the list of available windows.

PatTheMav: Given that the checkbox only affects display capture, the other dropdowns are unaffected by this change, correct.

PR #8626, opened 2023-04-02
frontend-plugins: Fix auto scene switcher not saving settings on close

PatTheMav:
### Description
Fixes settings in the auto scene switcher window not being saved when the window is closed.

### Motivation and Context
The `done` event does not emit a `closeEvent`, but saving is only implemented in that event handler. Changing this to `close` correctly triggers the default QWidget event and thus settings are saved.

### How Has This Been Tested?
Tested on macOS 13.

### Types of changes

### Checklist:

### DIFF
--- a/UI/frontend-plugins/frontend-tools/auto-scene-switcher.cpp
+++ b/UI/frontend-plugins/frontend-tools/auto-scene-switcher.cpp
@@ -128,9 +128,10 @@ SceneSwitcher::SceneSwitcher(QWidget *parent)
 		SetStopped();
 
 	loading = false;
+	connect(this, &QDialog::finished, this, &SceneSwitcher::finished);
 }
 
-void SceneSwitcher::closeEvent(QCloseEvent *)
+void SceneSwitcher::finished()
 {
 	obs_frontend_save();
 }
--- a/UI/frontend-plugins/frontend-tools/auto-scene-switcher.hpp
+++ b/UI/frontend-plugins/frontend-tools/auto-scene-switcher.hpp
@@ -21,8 +21,6 @@ public:
 
 	SceneSwitcher(QWidget *parent);
 
-	void closeEvent(QCloseEvent *event) override;
-
 	void SetStarted();
 	void SetStopped();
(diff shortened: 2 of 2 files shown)


RytoEX:
One thing that stands out from the docs on [QDialog::done()](https://doc.qt.io/qt-6/qdialog.html#done):
> If this dialog is shown with [exec](https://doc.qt.io/qt-6/qdialog.html#exec)(), done() also causes the local event loop to finish, and [exec](https://doc.qt.io/qt-6/qdialog.html#exec)() to return r.

Does `close()` also does this?

UI/frontend-plugins/frontend-tools/auto-scene-switcher.cpp
```diff
@@ -128,9 +128,12 @@ SceneSwitcher::SceneSwitcher(QWidget *parent)
 		SetStopped();
 
 	loading = false;
+
+	QObject::connect(this, SIGNAL(finished(int)), this,
+			 SLOT(finished(int)));
 }
 
-void SceneSwitcher::closeEvent(QCloseEvent *)
+void SceneSwitcher::finished(int)
```
gxalpha, line 136: Unused parameters can be omitted entirely in slots.

UI/frontend-plugins/frontend-tools/auto-scene-switcher.cpp
```diff
@@ -128,9 +128,12 @@ SceneSwitcher::SceneSwitcher(QWidget *parent)
 		SetStopped();
 
 	loading = false;
+
+	QObject::connect(this, SIGNAL(finished(int)), this,
+			 SLOT(finished(int)));
```
gxalpha, line 133: It would be nice to use new syntax (along the lines of `connect(this, &QDialog::finished, this, &SceneSwitcher::finished)`)

gxalpha (commented):
Seems fine.
The PR changed substantially since the above comment so I don't believe it's relevant (to this PR at least) anymore.

PR #8633, opened 2023-04-03
obs-ffmpeg: Use event to cancel media source reconnect sleep (first dropped frames fix)

tt2468:
### Description
Fixes an issue where update/destroy of a media source during reconnect would block the parent thread until the next reconnect attempt. This would result in significant quantities of dropped frames, delayed OBS shutdown, or frozen UI during this period.

Fixes #7521

### Motivation and Context
We should avoid blocking the graphics thread for meaningless tasks. Blocking the UI thread also is not very desirable.

### How Has This Been Tested?
- Create media source
- Specify invalid network-based playback URL (like `rtmp://stream.global.irl.run/ingest/invalidkey`)
- Click Ok
- Immediately open properties dialog of media source
- Click Ok
- UI will freeze until the reconnect delay is over, and the graphics thread will too

With this PR, it no longer happens.

With time measurement attached to `pthread_join()` call in `ffmpeg_source_update()` callback:
- Current master: `info: time taken to join: 5733ms`
- This PR: `info: time taken to join: 0ms`

### Types of changes
- Bug fix (non-breaking change which fixes an issue)
- Performance enhancement (non-breaking change which improves efficiency)

### Checklist:
- [x] My code has been run through [clang-format](https://github.com/obsproject/obs-studio/blob/master/.clang-format).
- [x] I have read the [**contributing** document](https://github.com/obsproject/obs-studio/blob/master/CONTRIBUTING.rst).
- [x] My code is not on the master branch.
- [x] The code has been tested.
- [x] All commit messages are properly formatted and commits squashed where appropriate.
- [x] I have included updates to all appropriate documentation.

PR #8634, opened 2023-04-03
obs-ffmpeg: Use destroy thread to destroy media player (second dropped frames fix)

tt2468:
### Description
**Depends #8633**

Destroying the media player can take anywhere from 30-80ms, regardless of underlying media type (tested rtmp, srt, mp4). Previously, the thread tasked to destroying the media player would be the graphics thread. This would lead to dropped frames from that thread in the following cases:
- Source update and playback is running or previously ended
- Playback finishes and "close file when inactive" was enabled

(No frames would be dropped on source destroy due to that callback being offloaded to the destroy thread in https://github.com/obsproject/obs-studio/commit/8b3416c1e70bd4dad6baf1d424357a6af540efe0)

### Motivation and Context
I don't like anything that has the possibility of causing the graphics thread to drop frames.

### How Has This Been Tested?
Reproduce the fixed behavior:
- Load OBS
- Open stats dialog
- Create a media source
- Select a local file, keep "Loop" disabled, enable "Close file when inactive"
- Click ok
- Wait for file to end (or skip to a few seconds before the file ends)
- Watch the stats dialog as dropped render frames increases by 1-4 frames when playback ends

### Types of changes

### Checklist:

### DIFF
--- a/plugins/obs-ffmpeg/obs-ffmpeg-source.c
+++ b/plugins/obs-ffmpeg/obs-ffmpeg-source.c
@@ -32,7 +32,6 @@
 
 struct ffmpeg_source {
 	media_playback_t *media;
-	bool destroy_media;
 
 	enum video_range_type range;
 	bool is_linear_alpha;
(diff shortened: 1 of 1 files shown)


tt2468:
Closing this as there are associated threading issues which require more extensive effort into refactoring the media-playback usage. Will tackle at a later date (likely between 29.1 and 30)

PR #8640, opened 2023-04-03
mac-syphon: Update syphon, use obs-deps framework, remove submodule

gxalpha:
### Description

Updates the syphon to use the Syphon.framework from obs-deps instead of the submodule, and removes the submodule.

#### Draft TODO-list:
- [x] Merge https://github.com/obsproject/obs-deps/pull/176
- [x] Update obs-deps in obs-studio

### Motivation and Context

The submodule was not updated in 9 years and additionally had custom patches. We want to get away from having submodule dependencies, and also from having custom patches for dependencies.

### How Has This Been Tested?

macOS 13.3
Tested with a custom obs-deps that includes `Syphon.framework` framework.
Ran their [sample server](https://github.com/Syphon/Simple) and had it correctly get picked up, announce creation and retirement; and received the correct frames.

### Types of changes

### Checklist:

### DIFF
--- a/cmake/macos/helpers.cmake
+++ b/cmake/macos/helpers.cmake
@@ -65,6 +65,13 @@ function(set_target_properties_obs target)
           PROPERTY XCODE_EMBED_FRAMEWORKS ${SPARKLE})
       endif()
 
+      if(TARGET mac-syphon)
+        set_property(
+          TARGET ${target}
+          APPEND
+          PROPERTY XCODE_EMBED_FRAMEWORKS ${SYPHON})
+      endif()
+
       get_property(obs_executables GLOBAL PROPERTY _OBS_EXECUTABLES)
       add_dependencies(${target} ${obs_executables})
       foreach(executable IN LISTS obs_executables)
--- a/plugins/mac-syphon/CMakeLists.txt
+++ b/plugins/mac-syphon/CMakeLists.txt
@@ -11,76 +11,19 @@ else()
   target_enable_feature(mac-syphon "Syphon sharing support")
 endif()
 
-find_library(COCOA Cocoa)
-find_library(IOSURF IOSurface)
-find_library(SCRIPTINGBRIDGE ScriptingBridge)
-mark_as_advanced(COCOA IOSURF SCRIPTINGBRIDGE)
-
-add_library(syphon-framework STATIC EXCLUDE_FROM_ALL )
-add_library(Syphon::Framework ALIAS syphon-framework)
-
-target_sources(
-  syphon-framework
-  PRIVATE syphon-framework/Syphon_Prefix.pch
-          syphon-framework/Syphon.h
-          syphon-framework/SyphonBuildMacros.h
-          syphon-framework/SyphonCFMessageReceiver.m
-          syphon-framework/SyphonCFMessageReceiver.h
-          syphon-framework/SyphonCFMessageSender.h
-          syphon-framework/SyphonCFMessageSender.m
-          syphon-framework/SyphonClient.m
-          syphon-framework/SyphonClient.h
-          syphon-framework/SyphonClientConnectionManager.m
-          syphon-framework/SyphonClientConnectionManager.h
-          syphon-framework/SyphonDispatch.c
-          syphon-framework/SyphonDispatch.h
-          syphon-framework/SyphonIOSurfaceImage.m
-          syphon-framework/SyphonIOSurfaceImage.h
-          syphon-framework/SyphonImage.m
-          syphon-framework/SyphonImage.h
-          syphon-framework/SyphonMachMessageReceiver.m
-          syphon-framework/SyphonMachMessageReceiver.h
-          syphon-framework/SyphonMachMessageSender.m
-          syphon-framework/SyphonMachMessageSender.h
-          syphon-framework/SyphonMessageQueue.m
-          syphon-framework/SyphonMessageQueue.h
-          syphon-framework/SyphonMessageReceiver.m
-          syphon-framework/SyphonMessageReceiver.h
-          syphon-framework/SyphonMessageSender.m
-          syphon-framework/SyphonMessageSender.h
-          syphon-framework/SyphonMessaging.m
-          syphon-framework/SyphonMessaging.h
-          syphon-framework/SyphonOpenGLFunctions.c
-          syphon-framework/SyphonOpenGLFunctions.h
-          syphon-framework/SyphonPrivate.m
-          syphon-framework/SyphonPrivate.h
-          syphon-framework/SyphonServer.m
-          syphon-framework/SyphonServer.h
-          syphon-framework/SyphonServerConnectionManager.m
-          syphon-framework/SyphonServerConnectionManager.h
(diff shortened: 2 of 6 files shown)


gxalpha:
Updated to use the new `2023-06-01` obs-deps release, which means it now works without any additional dependencies locally and also compiles on CI. For this to undraft, the new obs-deps release would need to be used on master (although that's probably not too far off).

gxalpha:
With the deps updated on master, this now no longer needs to be in draft.

cmake/macos/helpers.cmake
```diff
@@ -65,6 +65,13 @@ function(set_target_properties_obs target)
           PROPERTY XCODE_EMBED_FRAMEWORKS ${SPARKLE})
       endif()
 
+      if(ENABLE_SYPHON)
```
PatTheMav, line 68: Check for the presence of the `mac-syphon` target instead, as the inclusion depends on whether the plugin is configured and not whether some variable is set.

(The variable _might_ trigger the creation of the target, but that should be of no concern for the application target).

plugins/mac-syphon/CMakeLists.txt
```diff
@@ -11,76 +11,19 @@ else()
...
 add_library(OBS::syphon ALIAS mac-syphon)
 
 target_sources(mac-syphon PRIVATE syphon.m plugin-main.c)
 
-target_compile_options(mac-syphon PRIVATE -include "${CMAKE_CURRENT_SOURCE_DIR}/syphon-framework/Syphon_Prefix.pch"
-                                          -fobjc-arc)
-
-target_link_libraries(mac-syphon PRIVATE OBS::libobs Syphon::Framework ${SCRIPTINGBRIDGE})
+target_compile_options(mac-syphon PRIVATE -fobjc-arc)
+target_link_libraries(mac-syphon PRIVATE OBS::libobs ${APPKIT} ${IOSURFACE} ${SCRIPTINGBRIDGE} ${SYPHON})
```
PatTheMav, line 26: If we're changing these parts here, we might as well clean them up correctly:

Instead of `find_library` and a CMake variable, use `$<LINK_LIBRARY:Framework,AppKit.framework>` to link against system frameworks and with `${SYPHON}` to make it explicit.

plugins/mac-syphon/syphon.m
```diff
@@ -147,10 +160,12 @@ static inline void check_description(syphon_t s, NSDictionary *desc)
...
 
-static inline void handle_new_frame(syphon_t s,
-				    SYPHON_CLIENT_UNIQUE_CLASS_NAME *client)
+static inline void handle_new_frame(syphon_t s, SyphonOpenGLClient *client)
 {
-	IOSurfaceRef ref = [client IOSurface];
+	SyphonOpenGLImage *image = [client newFrameImage];
+	IOSurfaceRef ref = (IOSurfaceRef)CFBridgingRetain(
+		/* Rules are made to be broken... like buildings! Or people! */
+		[image performSelector:@selector(surface)]);
```
PatTheMav, line 168: It's either that..

Or you properly create a new category for the original class (say.. `SyphonOpenGLImage+Surface.m`) - there are canonical ways to do this:

https://developer.apple.com/library/archive/documentation/Cocoa/Conceptual/ProgrammingWithObjectiveC/CustomizingExistingClasses/CustomizingExistingClasses.html#//apple_ref/doc/uid/TP40011210-CH6-SW2

plugins/mac-syphon/CMakeLists.txt
gxalpha: Changed to use `LINK_LIBRARY:FRAMEWORK` for all frameworks.

cmake/macos/helpers.cmake
gxalpha: Changed to check for `TARGET mac-syphon`

plugins/mac-syphon/syphon.m
gxalpha: I will admit that I don't understand how that would avoid the `performSelector` call, as `surface` is still hidden in `SyphonOpenGLImage`.
I created a category for the class, but as far as I can tell I'd still have to call `surface` via the ObjC runtime. The code does indeed look nicer but I don't think it's what you intended.
I might very well be missing something though, please apologize my ignorance here. I'll push what I did as a temporary commit here so you can see what I mean.

Looking into this more I believe that the "proper" way to do this is to create our own subclasses of `SyphonServerBase` and `SyphonImageBase`. I can probably do that if we don't want to call private APIs, but would like to know that I didn't miss something major with what you mentioned above because it does look at least like a bit of work.

PatTheMav: By implementing a new category for `SyphonOpenGLImage` (and importing that header) you can add a public getter method that returns the property.

As Apple's docs specify:

> If you need to add a method to an existing class, perhaps to add functionality to make it easier to do something in your own application, the easiest way is to use a category.

So precisely what we need here. You effectively "overwrite" the existing class definition with your extended one. It's a common pattern in ObjC to add functionality to existing classes (e.g. adding encoders/decoders to the `NSString` class and similar use cases).

PatTheMav: The result would be that you would just do `[image getSurface]` and the category would just extend via:

```
@public
- (IOSurfaceRef)getSurface
{
    return _surface;
}
```

gxalpha: `_surface` would still be invisible though?

Either way it doesn't matter, I changed it to subclass `SyphonServerBase` which I believe is the real proper way anyways. This has the advantage of cutting out the OpenGL middleman, removing the need for an OpenGL context and allowing us to access the IOSurfaces in a direct way.

plugins/mac-syphon/syphon.m
```diff
@@ -1,13 +1,19 @@
-#import <Cocoa/Cocoa.h>
+#define GL_SILENCE_DEPRECATION
```
gxalpha, line 1: Note to self, as a conversation to avoid merging: This should be removed again

plugins/mac-syphon/SyphonOBSClient.m
```diff
@@ -0,0 +1,17 @@
+#import "SyphonOBSClient.h"
+
+@implementation SyphonOBSClient
+
+- (id)initWithServerDescription:(NSDictionary<NSString *, id> *)description
+                        options:(NSDictionary<NSString *, id> *)options
+                newFrameHandler:(void (^)(SyphonOBSClient *))handler
+{
+    self = [super initWithServerDescription:description options:options newFrameHandler:handler];
```
PatTheMav, line 9: It is canonical to always check if the assignment was actually successful:

```
    self = [super <SOME INITIALIZER>];
    if ( self ) {
        <SOME CODE>   
    }
    return self;
```

`self` can be returned safely because `nil` objects can still be used for message passing in ObjC.

plugins/mac-syphon/syphon.m
PatTheMav: Correct. :D

gxalpha: Fixed.

PatTheMav (approved):

PR #8647, opened 2023-04-04
CI: Fix build errors with Xcode 14.3 and platform SDK 13.3

PatTheMav:
### Description
Makes warning about unqualified `std::` calls a non-error and replaces `CGDisplayStream` header with correct lower macOS version boundaries.

### Motivation and Context
Xcode 14.3 and the macOS 13.3 platform SDK introduced a few breaking changes:

* Updated AppleClang emits warnings about unqualified std cast calls when using C++ - as `move` is too broad a word, developers are to use `std::move` to make this explicit. Alas this is exactly what `json11` uses and because that library is archived, there is no possibility of an upstream update.

* Apple guarded calls to old screen capture APIs as "available but deprecated", but seems to have chosen the wrong lower version boundary: The calls are flagged as being available for macOS 13 and macOS 14 only.

  To fix this, the existing macOS platform SDK header is replaced by a local copy that uses macOS 11 as the lower boundary (the oldest macOS version supported by obs-studio anyway)

Also requires https://github.com/obsproject/obs-browser/pull/402.
### How Has This Been Tested?
Tested with Xcode 14.3 and AppleClang 14.0.3.

### Types of changes

### Checklist:


RytoEX:
> * To fix this, the existing macOS platform SDK header is replaced by a local copy that uses macOS 11 as the lower boundary (the oldest macOS version supported by obs-studio anyway)

Is this something we can revert if fixed in the SDK? This feels gross.


> Also requires [obsproject/obs-browser#402](https://github.com/obsproject/obs-browser/pull/402).

Does that mean we need to update the submodule first? Or are you just saying we need both eventually in any order to be compatible with Xcode 14.3 and SDK 13.3?

UI/cmake/os-macos.cmake
```diff
@@ -15,3 +15,7 @@ target_link_libraries(obs-studio PRIVATE ${APPKIT} ${AVFOUNDATION} ${APPLICATION
 
 target_compile_options(obs-studio PRIVATE -Wno-error=float-conversion -Wno-error=implicit-int-conversion
                                           -Wno-error=shorten-64-to-32)
+
+if(CMAKE_C_COMPILER_VERSION VERSION_GREATER_EQUAL 14.0.3)
+  target_compile_options(obs-studio PRIVATE -Wno-error=unqualified-std-cast-call)
+endif()
```
gxalpha, line 21: Is this only relevant with AppleClang, or any new clang?

UI/platform-osx.mm
```diff
@@ -308,34 +308,15 @@ MacPermissionStatus CheckPermissionWithPrompt(MacPermissionType type,
 		break;
 	}
 	case kScreenCapture: {
-#if __MAC_OS_X_VERSION_MAX_ALLOWED >= 110000
-		if (@available(macOS 11.0, *)) {
-			permissionResponse = (CGPreflightScreenCaptureAccess()
+		permissionResponse = (CGPreflightScreenCaptureAccess()
+					      ? kPermissionAuthorized
+					      : kPermissionDenied);
+
```
gxalpha, line 314: Not directly completely related to this PR, but did we go through with raising the minimum runtime version to 11.0 for the 29.1 release? (I'm not against, just would like to know as it would also mean that we'll need to update support documents)

PatTheMav: The CMake 3.0 rework is based on that - the minimum system version is 11.0 already.

UI/cmake/os-macos.cmake
PatTheMav: The change was signed-off in clang in February last year, so it probably also affects LLVM-clang, but this file is for macOS only and you shouldn't use LLVM-clang on macOS.

UI/platform-osx.mm
gxalpha: Thanks for clarifying!

RytoEX:
> > * To fix this, the existing macOS platform SDK header is replaced by a local copy that uses macOS 11 as the lower boundary (the oldest macOS version supported by obs-studio anyway)
> 
> Is this something we can revert if fixed in the SDK? This feels gross.

I would still like clarification on this.

I'm otherwise okay with these changes for the most part.

PatTheMav:
> > > * To fix this, the existing macOS platform SDK header is replaced by a local copy that uses macOS 11 as the lower boundary (the oldest macOS version supported by obs-studio anyway)
> > 
> > 
> > Is this something we can revert if fixed in the SDK? This feels gross.
> 
> I would still like clarification on this.
> 
> I'm otherwise okay with these changes for the most part.

The `CGDisplayStream.h` header of the `CGDisplayStream` framework shipped with the macOS 13.3. platform SDK guards the functions we use for display capture:

`CG_EXTERN CGError CGDisplayStreamStart(CGDisplayStreamRef cg_nullable displayStream) CG_AVAILABLE_BUT_DEPRECATED(13.0, 14.0, "Please use ScreenCaptureKit API's startCaptureWithCompletionHandler: to start a stream instead");`

This marks the functions as being deprecated, but also only available in macOS 13 and 14 - the source files will not be compiled as long as the minimum deployment target is not macOS 13.

The choice was to either remove display capture entirely (per the deprecation notice using SCK instead) or use a local copy of the header that sets the lower limit to 11.0.

UI/cmake/os-macos.cmake
gxalpha: My question is basically "if someone were to try compiling with clang on another OS, would they also be running into this?"
(Not that I would want to block this PR over such a concern, for macOS it’s of course fine as-is.

PatTheMav: I haven't checked with llvm-clang, if the respective version contains the change, the answer would be "yes".

RytoEX (approved):
Okay with this as long as we remember to remove the local copy when possible.

PR #8648, opened 2023-04-04
UI: Use name instead of internal extension for incompatible codec check

gxalpha:
### Description

Uses the localized name instead of the internal extension to display that something is incompatible.

Before:
image

After:
image

I will say that I'm a tiny bit torn on the double brackets, but also didn't want to introduce another string for this.

(and yes, I realize my screenshots were from different comboboxes, but the result is the same for both)
 
### Motivation and Context

Current warning doesn't look great.

### How Has This Been Tested?

macOS 13.3
See screenshots above.

### Types of changes

### Checklist:

### DIFF
--- a/UI/window-basic-settings.cpp
+++ b/UI/window-basic-settings.cpp
@@ -5013,11 +5013,11 @@ static bool ContainerSupportsCodec(const string &container, const string &codec)
 }
 
 static void DisableIncompatibleCodecs(QComboBox *cbox, const QString &format,
+				      const QString &formatName,
 				      const QString &streamEncoder)
 {
 	QString strEncLabel =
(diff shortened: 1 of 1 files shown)


derrod:
This was an oversight when updating how formats are handled, this fix is probably fine, alternatively we could replace
```c++
QString formatUpper = format.toUpper();
```
with something like
```c++
string extension = GetFormatExt(QT_TO_UTF8(format));
QString formatUpper = QString::fromStdString(extension).toUpper();
```
Which I recognise is also ugly, but maybe a little more concise in the actual list than having the display name + double brackets.

derrod (approved):

PR #8650, opened 2023-04-04
obs-outputs: Fix HEVC/RTMP composition time

nquah:
Fixes bug in Enhanced RTMP implementation, where HEVC streams with B-frames enabled are not adhering to the spec.

Credit to derrod 

### Description

Populates non-zero composition times in the FLV extended video tag header for HEVC streams.

### Motivation and Context
See https://github.com/obsproject/obs-studio/pull/8522 for context.

The [Enhanced RTMP spec](https://github.com/veovera/enhanced-rtmp) provides an optimization where HEVC streams without B-frames do not need to send a composition time of 3 bytes (with value 0) for every frame.

This patch fixes a bug: Composition times are just not getting set at all, instead of only when the value of the composition time is 0.

### How Has This Been Tested?

We ran streams against YouTube, inspected the bitstream and confirmed that composition times are correctly set.

### Types of changes

### Checklist:

### DIFF
--- a/plugins/obs-outputs/flv-mux.c
+++ b/plugins/obs-outputs/flv-mux.c
@@ -326,14 +326,29 @@ void flv_packet_ex(struct encoder_packet *packet, enum video_id_t codec_id,
 	int32_t time_ms = get_ms_time(packet, packet->dts) - dts_offset;
 
 	// packet head
+	int header_metadata_size = 5;
+#ifdef ENABLE_HEVC
+	// 3 extra bytes for composition time offset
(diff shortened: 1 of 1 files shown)


jp9000:
Hi there! Just as a minor nitpick, please make sure to use present tense rather than past tense in PRs and commits.

derrod (approved):
Tested this locally on my MacBook and confirmed streams with b-frames now work correctly. Also made sure this still compiles/works without `ENABLE_HEVC`.

PR #8661, opened 2023-04-05
UI: Fix simple mode container check

derrod:
### Description

Fixes the compatibility check again, this time because of a silly mistake I should've caught because I made it before. `QT_TO_UTF8(x)` (which just expands to `x.toUtf8().constData()`) returns a pointer to memory that gets freed immediately at the end of the statement. This is valid in function calls but will cause issues in cases like this one.

Fun fact: This is a use-after-free and `rustc` would yell at you for this, but C++ compilers just go "nah seems goot mate".

### Motivation and Context

Fix bugs.

### How Has This Been Tested?

Opened settings a few times and validated things work now.

### Types of changes

### Checklist:

### DIFF
--- a/UI/window-basic-settings.cpp
+++ b/UI/window-basic-settings.cpp
@@ -5749,20 +5749,21 @@ static void DisableIncompatibleSimpleContainer(QComboBox *cbox,
 {
 	/* Similar to above, but works in reverse to disable incompatible formats
 	 * based on the encoder selection. */
-	const char *aCodec = QT_TO_UTF8(aEncoder);
-	const char *vCodec = obs_get_encoder_codec(
+	string vCodec = obs_get_encoder_codec(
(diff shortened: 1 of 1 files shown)


UI/window-basic-settings.cpp
```diff
@@ -5762,7 +5761,8 @@ static void DisableIncompatibleSimpleContainer(QComboBox *cbox,
 		QStandardItem *item = model->item(idx);
 
 		if (ContainerSupportsCodec(format.toStdString(), vCodec) &&
-		    ContainerSupportsCodec(format.toStdString(), aCodec)) {
+		    ContainerSupportsCodec(format.toStdString(),
+					   aEncoder.toStdString())) {
```
notr1ch, line 5765: Could this be pulled out of the loop so it's only creating one string?

PR #8667, opened 2023-04-05
libobs: Fix transition lookup by uuid

cg2121:
### Description
The obs_get_transition_by_uuid function would return a valid source even if it wasn't a transition.

### Motivation and Context
Fixes bug I've found

### How Has This Been Tested?
Got transitions by uuid

### Types of changes

### Checklist:

### DIFF
--- a/libobs/obs.c
+++ b/libobs/obs.c
@@ -2123,7 +2123,14 @@ obs_source_t *obs_get_transition_by_name(const char *name)
 
 obs_source_t *obs_get_transition_by_uuid(const char *uuid)
 {
-	return obs_get_source_by_uuid(uuid);
+	obs_source_t *source = obs_get_source_by_uuid(uuid);
+
+	if (source && source->info.type == OBS_SOURCE_TYPE_TRANSITION)
+		return source;
+	else if (source)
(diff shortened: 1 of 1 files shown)


libobs/obs.c
```diff
@@ -2123,7 +2123,14 @@ obs_source_t *obs_get_transition_by_name(const char *name)
 
 obs_source_t *obs_get_transition_by_uuid(const char *uuid)
 {
-	return obs_get_source_by_uuid(uuid);
+	obs_source_t *source = obs_get_source_by_uuid(uuid);
+
+	if (source && source->info.type == OBS_SOURCE_TYPE_TRANSITION)
+		return source;
+	else
+		obs_source_release(source);
```
notr1ch, line 2131: This could try to call `obs_source_release` with a NULL source. While this is protected in the API, I'd prefer we avoid passing parameters which we know will fail in the first place.

cg2121: Should by fixed now.

derrod (approved):
Seems fine. I mostly didn't bother adding a check here since once you have a UUID you probably should know what it's for.

Issue #8548, opened 2023-03-27
New AMF doesn't support reconfiguration properly


MattyFresh68:
### Operating System Info

Windows 11

### OBS Studio Version

29.0.2

### OBS Studio Log URL

https://obsproject.com/logs/O4PGzxUYDutvJB6C

### Expected Behavior

I expect to be able to adjust bitrate during an active stream output and the bitrate change to either apply, or at the very least, silently remain the same. 

### Current Behavior

With the change to the new version of amf, updating bitrate during an active output will hard lock at 0kbps. OBS will then also hang requiring killing it with task manager. 

### Steps to Reproduce

1. start stream at 3500kbps(I used twitch but other user in discord repro to YT)
2. adjust bitrate of stream output to 4500kbps
3.watch bitrate lock at 0kbps with green indicator
4.press stop stream
5.close obs(will then still be active in task manager until killed)

MattyFresh68:
As I can't personally test, but based on a report in discord, it seems this behavior also occurs using av1 on amd hardware.

RytoEX:
Can confirm on an RX 6600M.

Johl7:
Late to the party, tested here briefly with the mentioned OBS vers.. Couldnt repro on my Polaris gpu, using the latest drivers.

flaeri:
I'll add my confirmation as well. Pretty much instant encoder death on bitrate change reconfig.

https://obsproject.com/logs/Ik2S2z7XsnUiOXde

23.4.2 drivers, 29.0.2 OBS. Please always include a log if confirming/disproving :)

PR #8668, opened 2023-04-05
obs-ffmpeg: Disable AMF texture encoder runtime reconfiguration

derrod:
### Description

A number of people and their logs indicate that the AMD texture encoder will "crash" (drop to 0 kbps output) when reconfigured while running (e.g. #8548).

To prevent people's outputs from dropping disable this for now and rely on frame-dropping instead.

### Motivation and Context

Do not want outputs to die.

### How Has This Been Tested?

Has not.

### Types of changes

### Checklist:


RytoEX:
Should this have closed #8548 and

derrod:
No, the issue isn't identified/fixed, this just prevents the issue from happening.

RytoEX:
While the issue of "AMF doesn't support reconfiguration" is unresolved, but the encoder crash is solved by way of this PR. If we want to keep the Issue scoped on the former, then we can leave it open I suppose.

PR #8670, opened 2023-04-05
libobs: Use dlopen_preflight before loading dynamic libraries on macOS

PatTheMav:
### Description
Adds dynamic library checks with `dlopen_preflight` before actually loading a library.

### Motivation and Context
Checking for ABI compatibility and architecture is beneficial to avoid unnecessary calls to `dlopen` for libraries that will not be loadable and allows the program to gracefully skip those.

### How Has This Been Tested?
Tested on macOS 13.

### Types of changes

### Checklist:

### DIFF
--- a/libobs/util/platform-nix.c
+++ b/libobs/util/platform-nix.c
@@ -73,18 +73,25 @@ void *os_dlopen(const char *path)
 		return NULL;
 
 	dstr_init_copy(&dylib_name, path);
-#ifdef __APPLE__
-	if (!dstr_find(&dylib_name, ".framework") &&
-	    !dstr_find(&dylib_name, ".plugin") &&
-	    !dstr_find(&dylib_name, ".dylib") && !dstr_find(&dylib_name, ".so"))
-#else
-	if (!dstr_find(&dylib_name, ".so"))
-#endif
-		dstr_cat(&dylib_name, ".so");
 
 #ifdef __APPLE__
+#ifndef __aarch64__
+	if (dstr_find(&dylib_name, ".so")) {
+		bool check = dlopen_preflight(dylib_name.array);
+
+		if (!check) {
+			blog(LOG_ERROR, "Library failed preflight checks: %s",
+			     dylib_name.array);
+			dstr_free(&dylib_name);
+			return NULL;
+		}
+	}
+#endif
 	void *res = dlopen(dylib_name.array, RTLD_LAZY | RTLD_FIRST);
 #else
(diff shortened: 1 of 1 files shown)


notr1ch:
Doesn't `dlopen` perform the same checks? It seems redundant to call both functions.

> This is potentially an expensive call because it may internally do the same as dlopen/dlclose. Only use dlopen_preflight() if you need to show the user a list of potentially loadable plugins.

PatTheMav:
> Doesn't `dlopen` perform the same checks? It seems redundant to call both functions.
> 
> > This is potentially an expensive call because it may internally do the same as dlopen/dlclose. Only use dlopen_preflight() if you need to show the user a list of potentially loadable plugins.

That is correct, but we also got Qt5-based plugins still in the wild and alas also allow old-style plugins (the non-bundle variant with `.so` ending) to load on x86_64.

Apple Silicon is _mostly_ unaffected by that, but Intel-based Macs will run into those potentially "bad" plugins.

notr1ch:
I still don't really understand the purpose of this check. Does OBS crash if you `dlopen` an incompatible  .so on MacOS? What is this trying to accomplish / prevent?

PatTheMav:
The purpose is to have a bespoke error log entry when a user loads a heavily outdated plugin (e.g. linked against Qt5), as `dyld` error messages do not appear in our log otherwise (`dlopen` will just fail and skip the plugin).

This is simpler (and faster) compared to what we do on Windows already (https://github.com/obsproject/obs-studio/blob/e03a2751f47b160498a6a3941edfda6e3ce724c6/libobs/util/platform-windows.c#L142), but also adds transparency for support purposes.

The alternative would be to drop support for old Linux-style plugin libraries entirely (as they aren't supported by our Apple Silicon builds anyway).

notr1ch:
https://github.com/obsproject/obs-studio/blob/ae489cf39cae69a6ab96f058f86336cbf1516617/libobs/util/platform-nix.c#L97-L99

Doesn't this log the same kind of error (and in more detail with the `dlerror` call)?

PatTheMav:
> https://github.com/obsproject/obs-studio/blob/ae489cf39cae69a6ab96f058f86336cbf1516617/libobs/util/platform-nix.c#L97-L99
> 
> Doesn't this log the same kind of error (and in more detail with the `dlerror` call)?

It does, but it's also very verbose. It all comes down to whether we prefer EAFP or LBYL in this case, I'm fine with either.

PR #8671, opened 2023-04-05
obs-outputs: Fix keychain API deprecation warnings on macOS

PatTheMav:
### Description
Updates keychain API calls in `librtmp` to use non-deprecated calls (`SecKeychainOpen` is been deprecated since Mac OSX 10.10).

### Motivation and Context
Modernise code and avoid sudden breakage or unexpected results because of reliance on API functions explicitly marked as deprecated.

### How Has This Been Tested?
Tested on macOS 13 with an active RTMPS stream.

### Types of changes

### Checklist:

### DIFF
--- a/plugins/obs-outputs/librtmp/rtmp.c
+++ b/plugins/obs-outputs/librtmp/rtmp.c
@@ -314,42 +314,37 @@ RTMP_TLS_LoadCerts(RTMP *r) {
     CertFreeCertificateContext(pCertContext);
     CertCloseStore(hCertStore, 0);
 #elif defined(__APPLE__)
-    SecKeychainRef keychain_ref;
-    CFMutableDictionaryRef search_settings_ref;
-    CFArrayRef result_ref;
+    CFTypeRef keys[4] = {kSecClass, kSecMatchLimit, kSecReturnAttributes,
+                 kSecReturnData};
 
-    if (SecKeychainOpen("/System/Library/Keychains/SystemRootCertificates.keychain",
-                        &keychain_ref)
-        != errSecSuccess) {
-      goto error;
+    CFTypeRef values[4] = {kSecClassCertificate, kSecMatchLimitAll,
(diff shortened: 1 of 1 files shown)


plugins/obs-outputs/librtmp/rtmp.c
```diff
@@ -315,23 +315,15 @@ RTMP_TLS_LoadCerts(RTMP *r) {
     CertCloseStore(hCertStore, 0);
 #elif defined(__APPLE__)
     SecKeychainRef keychain_ref;
-    CFMutableDictionaryRef search_settings_ref;
     CFArrayRef result_ref;
 
-    if (SecKeychainOpen("/System/Library/Keychains/SystemRootCertificates.keychain",
+    if (SecKeychainCopyDomainDefault(kSecPreferencesDomainSystem,
```
derrod, line 320: `SecKeychainCopyDomainDefault ` is also deprecated https://developer.apple.com/documentation/security/1401993-seckeychaincopydomaindefault?language=objc

plugins/obs-outputs/librtmp/rtmp.c
```diff
@@ -314,42 +314,38 @@ RTMP_TLS_LoadCerts(RTMP *r) {
     CertFreeCertificateContext(pCertContext);
     CertCloseStore(hCertStore, 0);
 #elif defined(__APPLE__)
-    SecKeychainRef keychain_ref;
-    CFMutableDictionaryRef search_settings_ref;
-    CFArrayRef result_ref;
+#elif defined(__APPLE__)
```
RytoEX, line 317: Doesn't this make two `elif defined(__APPLE__)` in a row?

RytoEX (changes requested):

plugins/obs-outputs/librtmp/rtmp.c
PatTheMav: That is indeed an oversight.

PR #8672, opened 2023-04-05
obs-scripting: Replace macro adding frontend functions with plain C code

PatTheMav:
### Description
Replaces inline macro function to add frontend functions to the Lua script environment with a pure C-based implementation.

### Motivation and Context
Instead of using a non-debuggable macro function, keep function references and desired function names in a static array that can then be iterated over to add desired functions.

This allows for proper debugging as well as makes the combination of function and function name more explicit.

### How Has This Been Tested?
Used obs test-script used internally to check for Lua functionality (which explicitly checks for availability and return types of frontend API functions).

### Types of changes

### Checklist:

### DIFF
--- a/deps/obs-scripting/obs-scripting-lua-frontend.c
+++ b/deps/obs-scripting/obs-scripting-lua-frontend.c
@@ -297,37 +297,58 @@ static int add_save_callback(lua_State *script)
 
 /* ----------------------------------- */
 
+typedef struct lua_function_tuple {
+	const char *name;
+	lua_CFunction function;
+} obs_frontend_func;
+
+static const obs_frontend_func functions[] = {
+	{.name = "obs_frontend_get_scene_names", .function = get_scene_names},
+	{.name = "obs_frontend_get_scenes", .function = get_scenes},
+	{.name = "obs_frontend_get_current_scene",
+	 .function = get_current_scene},
+	{.name = "obs_frontend_set_current_scene",
+	 .function = set_current_scene},
+	{.name = "obs_frontend_get_transitions", .function = get_transitions},
+	{.name = "obs_frontend_get_current_transition",
+	 .function = get_current_transition},
+	{.name = "obs_frontend_set_current_transition",
+	 .function = set_current_transition},
+	{.name = "obs_frontend_get_transition_duration",
+	 .function = get_transition_duration},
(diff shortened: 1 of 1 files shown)


deps/obs-scripting/obs-scripting-lua-frontend.c
```diff
@@ -297,37 +297,45 @@ static int add_save_callback(lua_State *script)
 
 /* ----------------------------------- */
 
+typedef struct lua_function_tuple {
+	const char *name;
+	lua_CFunction function;
+} obs_frontend_func;
+
+static obs_frontend_func functions[19] = {
```
notr1ch, line 305:
```suggestion
static const obs_frontend_func functions[] = {
```

deps/obs-scripting/obs-scripting-lua-frontend.c
```diff
@@ -297,37 +297,45 @@ static int add_save_callback(lua_State *script)
...
-	add_func(set_current_scene_collection);
-	add_func(get_profiles);
-	add_func(get_current_profile);
-	add_func(set_current_profile);
-	add_func(remove_event_callback);
-	add_func(add_event_callback);
-	add_func(remove_save_callback);
-	add_func(add_save_callback);
-#undef add_func
+	size_t num_items = sizeof(functions) / sizeof(obs_frontend_func);
```
notr1ch, line 333: We have this macro as an alternative:
```suggestion
	size_t num_items = OBS_COUNTOF(functions);
```

deps/obs-scripting/obs-scripting-lua-frontend.c
PatTheMav: Ah, I remember again why I did it that way - Cl.exe doesn't seem to like that:

```
D:\a\obs-studio\obs-studio\obs-studio\deps\obs-scripting\obs-scripting-lua-frontend.c(305,46): error C2099: initializer is not a constant [D:\a\obs-studio\obs-studio\obs-studio\build64\deps\obs-scripting\obs-scripting.vcxproj]
```

PatTheMav: It's indeed the cast - `clang` and `gcc` don't mind, but `cl` doesn't like it. Easy enough to fix.

PR #8676, opened 2023-04-05
UI: Fix incorrect use of QT_TO_UTF8 in SpeakerLayoutChanged

notr1ch:
### Description
Fix an invalid pointer caused by holding on to freed memory.

### Motivation and Context
After #8661 I wanted to see if there were other places affected by this anti-pattern.

### How Has This Been Tested?
Compiled and ran OBS.

### Types of changes

### Checklist:

### DIFF
--- a/UI/window-basic-settings.cpp
+++ b/UI/window-basic-settings.cpp
@@ -4671,17 +4671,20 @@ void OBSBasicSettings::SpeakerLayoutChanged(int idx)
 		 */
 		PopulateSimpleBitrates(ui->simpleOutputABitrate, isOpus);
 
-		const char *encoder_id = QT_TO_UTF8(
-			ui->advOutAEncoder->currentData().toString());
-		QString rec_encoder_id =
-			ui->advOutRecAEncoder->currentData().toString();
+		string stream_encoder_id = ui->advOutAEncoder->currentData()
+						   .toString()
+						   .toStdString();
+		string record_encoder_id = ui->advOutRecAEncoder->currentData()
+						   .toString()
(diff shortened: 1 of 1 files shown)


UI/window-basic-settings.cpp
```diff
@@ -4671,16 +4671,16 @@ void OBSBasicSettings::SpeakerLayoutChanged(int idx)
 		 */
 		PopulateSimpleBitrates(ui->simpleOutputABitrate, isOpus);
 
-		const char *encoder_id = QT_TO_UTF8(
+		string encoder_id = QT_TO_UTF8(
 			ui->advOutAEncoder->currentData().toString());
```
derrod, line 4675: You could just replace the macro here with `.toStdString()` (internally it does pretty much the same), or keep it. a QString and wrap the parameters below themselves in `QT_TO_UTF8`. The latter would probably look nicer since `rec_encoder_id` already is, but it is technically probably less efficient.

gxalpha:
This could probably use QString's `toStdString()`, avoiding the `QT_TO_UFF8` macro entirely.

UI/window-basic-settings.cpp
notr1ch: It might be worth doing this across the entire codebase as part of another PR, we mostly use `QT_TO_UTF8` for `QString` -> `string` but there are places where we also use `toStdString`.

notr1ch: Changed to use `toStdString` for both to avoid use of `QT_TO_UTF8` entirely.

PR #8677, opened 2023-04-05
mac-videotoolbox: Enforce minimum keyframe interval in CRF mode

### LINKED ISSUE #8466: [macOS] Output with H265 hardware encoder with 'auto' keyframe interval lags QuickTime, FCP, etc.
### Operating System Info

macOS 13

### Other OS

_No response_

### OBS Studio Version

29.0.2

_No response_

### OBS Studio Log URL

https://obsproject.com/logs/tJC1sq4dO1Ooualk

### OBS Studio Crash Log URL

_No response_

### Expected Behavior

H265 output with the 0=auto keyframe interval should behave normally and have comparable performance with seeks, trims, etc. as prescribed keyframe intervals do (i.e. 2).

### Current Behavior

Output generated with 0=auto keyframes results in poor performance in Final Cut, QuickTime, VLC, etc. Recordings longer than approximately ~1 minute are unreasonably poorly performing when attempting seeks, trim operations, etc. Setting the keyframe interval to e.g. 2 resolves the issue.

### Steps to Reproduce

1. Set encoder to Apple VT HEVC Hardware Encoder.
2. Ensure keyframe interval is 0=auto. ~~Container and bitrate control method does not seem to matter~~ Seems to occur in ABR or CRF mode, with any container on my system. Occurs with remuxed mkvs, or when recording directly to mp4/mov.
3. Record for at least one minute.
4. Open the resulting file and try to manipulate it in editor of choice (FCP, QuickTime, and VLC tested).

Successfully trimming a small section at the end of these videos with QuickTime seems to result in an abnormally large file near in size to the original. This seems odd but I do not know enough to speculate about possible causes from that info.

Not sure if this is could be an encoder issue, or maybe some issue with ffmpeg or another part of OBS's H265 path.

Tested to occur with or without b-frames enabled in OBS, and with other reasonable keyframe intervals like 4 and 5 (which perform as expected)

Example video uploaded to Discord: https://cdn.discordapp.com/attachments/1016400504814837811/1084899034621280366/2023-03-13_12-57-04.mov

derrod:
### Description

Set a default keyframe interval when using Apple VT encoder in CRF mode. If not set, the encoder will not create *any* keyframes past the initial one, making the file nearly unusable in editors (including FCP) and video players as well as causing potential data loss issues with formats expecting to fragment on keyframes (in our case: fMOV, fMP4, HLS).

Fixes #8466

Sidenote: It seems that `kVTCompressionPropertyKey_MaxKeyFrameInterval` just straight up does nothing with the HEVC encoder on M1 even in other modes (haven't tested H.264). No matter what it's set no keyframes will be created.

### Motivation and Context

This is bad and Apple should feel bad.

### How Has This Been Tested?

Made a recording with and without it to confirm keyframes are now inserted in 10 second intervals.

### Types of changes

### Checklist:

### DIFF
--- a/plugins/mac-videotoolbox/encoder.c
+++ b/plugins/mac-videotoolbox/encoder.c
@@ -596,6 +596,17 @@ static bool create_encoder(struct vt_encoder *enc)
 
 	if (enc->codec_type == kCMVideoCodecType_H264 ||
 	    enc->codec_type == kCMVideoCodecType_HEVC) {
+		/* Apple's documentation states that a keyframe interval of 0 will result in
+		 * the encoder automatically picking times to insert them; However, Apple's
+		 * encoder, when in CRF mode, will never actually insert any keyframes past
+		 * the very first one, rendering the files near-unusable in editors or
+		 * video players. So to avoid that happening, enforce a reasonable default
+		 * of 10 seconds in CRF mode. */
+		if (enc->keyint == 0 && strcmp(enc->rate_control, "CRF") == 0) {
+			VT_BLOG(LOG_INFO,
+				"Enforcing non-zero keyframe interval in CRF mode");
+			enc->keyint = 10;
+		}
 
 		// This can fail when using GPU hardware encoding
 		code = session_set_prop_int(


andrewfitz:
Hi, I see your fix only applies to CRF. However, this issue also occurs for me in ABR.

PR #8678, opened 2023-04-05
cmake: Update project to enable compatibility with Xcode default build settings

PatTheMav:
### Description
Fixes code and silences warnings where appropriate to make obs-studio buildable with Xcode's suggested default settings.

### Motivation and Context
Xcode generates projects with a default set of compiler settings and warnings and warns about projects not using them. While these warnings tend to be strict, they also discover typical cases of code smell that are easy to remove or acknowledge.

This PR only silences the warnings for macOS builds or when using AppleClang, as some issues are not directly fixable (i.e. because they're inherent to the design of a framework or library).

The less warnings and code issues there are with Xcode, the less issues should occur upon regular Xcode/Clang updates (which happen more frequently on macOS than on Linux or Windows).

### How Has This Been Tested?
Compiled successfully in Debug and Release configurations, successfully archived+exported the package to check for valid code signatures.

### Types of changes

### Checklist:


PatTheMav:
Requires https://github.com/obsproject/obs-studio/pull/8678 to successfully compile.

UI/window-basic-main.cpp
```diff
@@ -8219,8 +8219,10 @@ void OBSBasic::GetConfigFPS(uint32_t &num, uint32_t &den) const
 		GetFPSInteger(num, den);
 	else if (type == 2) //"Fraction"
 		GetFPSFraction(num, den);
-	else if (false) //"Nanoseconds", currently not implemented
-		GetFPSNanoseconds(num, den);
+	/*
+     *	else if (false) //"Nanoseconds", currently not implemented
+	 *		GetFPSNanoseconds(num, den);
+	 */
```
RytoEX, line 8225: Realign whitespace.

cmake/macos/xcode.cmake
```diff
@@ -67,22 +67,29 @@ set(CMAKE_XCODE_ATTRIBUTE_DEBUG_INFORMATION_FORMAT[variant=Release] dwarf-with-d
 set(CMAKE_XCODE_ATTRIBUTE_DEBUG_INFORMATION_FORMAT[variant=MinSizeRel] dwarf-with-dsym)
 
 # Strip generated and installed products only in Release or MinSizeRel configuration
-set(CMAKE_XCODE_ATTRIBUTE_STRIP_INSTALLED_PRODUCT[variant=Release] YES)
-set(CMAKE_XCODE_ATTRIBUTE_STRIP_INSTALLED_PRODUCT[variant=MinSizeRel] YES)
+# set(CMAKE_XCODE_ATTRIBUTE_STRIP_INSTALLED_PRODUCT[variant=Release] YES)
+# set(CMAKE_XCODE_ATTRIBUTE_STRIP_INSTALLED_PRODUCT[variant=MinSizeRel] YES)
```
RytoEX, line 71: Is there any point to leaving these here commented rather than removing them?

plugins/mac-avcapture/av-capture.mm
```diff
@@ -2396,7 +2396,7 @@ bool obs_module_load(void)
 	av_capture_info.output_flags = OBS_SOURCE_ASYNC_VIDEO |
 				       OBS_SOURCE_AUDIO |
 				       OBS_SOURCE_DO_NOT_DUPLICATE;
-	av_capture_info.get_defaults = av_capture_defaults_v2,
+	av_capture_info.get_defaults = av_capture_defaults_v2;
```
RytoEX, line 2399: This seems like a typo that could be fixed on its own?

plugins/mac-capture/mac-display-capture.m
```diff
@@ -366,6 +366,7 @@ static void display_capture_video_tick(void *data, float seconds)
 		end.y -= dc->crop_rect.size.height;
 		end.x -= dc->crop_rect.size.width;
 	case CROP_TO_WINDOW:
+
```
RytoEX, line 369: Remove this whitespace to align with `case CROP_NONE:` above.

RytoEX (changes requested):
A few items. The CMake changes overall look okay. I'd appreciate someone else's review on some of the code changes.

UI/window-basic-main.cpp
PatTheMav: That’s how clang-format does it.

plugins/mac-avcapture/av-capture.mm
PatTheMav: Sure, can move it into a separate PR.

PatTheMav: Will automatically be removed once https://github.com/obsproject/obs-studio/pull/8682 is merged and the PR will be rebased.

RytoEX:
> Requires #8678 to successfully compile.

This is #8678.

PatTheMav:
> > Requires #8678 to successfully compile.
> 
> This is #8678.

Updated the link.

cmake/macos/compilerconfig.cmake
```diff
@@ -30,40 +30,98 @@ set(CMAKE_VISIBILITY_INLINES_HIDDEN TRUE)
...
+      -Wconversion
+      -Wdeprecated-implementations
+      -Wduplicate-method-match
+      -Wfloat-conversion
+      -Wfour-char-constants
+      -Wimplicit-retain-self
+      -Winvalid-offsetof
+      -Wmove
+      -Wno-arc-maybe-repeated-use-of-weak
+      -Wno-c++11-extensions
```
RytoEX, line 93: Out of curiosity, why do we disable this warning? Shouldn't setting `CMAKE_CXX_EXTENSIONS OFF` (like we currently do) prevent us from being able to disable warnings about using C++11 extensions? Also, don't we *want* to avoid using extensions, so shouldn't we want the warnings? I'd actually almost rather we use `-Wpedantic`, since my understanding was that we want to avoid using extensions at all.

RytoEX (changes requested):

cmake/macos/compilerconfig.cmake
PatTheMav: I just explicitly added the warnings that Xcode enables by default so that compilers complain about the same things.

RytoEX: This is disabling a warning though, unless I'm mistaken?

PatTheMav: That is correct - I just enabled the preferred project settings and copied the compiler flags. I wonder if this one has any effect given that we enforce C++17 anyway, so I wouldn't be too fussed about removing it.

UI/window-basic-main.cpp
gxalpha: Are you certain? In my experience it doesn’t do anything inside block comments, I’d expect that it would leave a manually inserted tab here intact.

PatTheMav: Looks like it, might've been something else interfering with that then (probably Xcode).

RytoEX (approved):
Looks good to me.

PR #8684, opened 2023-04-07
cmake: Set C standard to C17

derrod:
### Description

Set default compiler options C standard to C17. For MSVC this was previously manually set via the compiler flags but we can probably use this on all platforms (I can't quite remember why it was necessary for MSVC).

### Motivation and Context

Might fix issue reported here: https://github.com/obsproject/obs-studio/pull/8377#issuecomment-1499495578

### How Has This Been Tested?

Compiled on macOS with Xcode 14.3, this PR will be the test for Windows/Linux.

### Types of changes

### Checklist:

### DIFF
--- a/cmake/Modules/CompilerConfig.cmake
+++ b/cmake/Modules/CompilerConfig.cmake
@@ -5,6 +5,13 @@ set(CMAKE_CXX_STANDARD 17)
 set(CMAKE_CXX_STANDARD_REQUIRED ON)
 set(CMAKE_CXX_EXTENSIONS OFF)
 
+set(CMAKE_C_STANDARD 17)
+set(CMAKE_C_STANDARD_REQUIRED ON)
+# TODO/FIXME: Investigate disabling C extensions on Linux/POSIX
+if(OS_MACOS OR NOT OS_POSIX)
+  set(CMAKE_C_EXTENSIONS OFF)
+endif()
+
 # Set compile options for MSVC
 if(OS_WINDOWS AND MSVC)
   if(NOT EXISTS "${CMAKE_BINARY_DIR}/ALL_BUILD.vcxproj.user")
@@ -66,8 +73,7 @@ if(OS_WINDOWS AND MSVC)
     /utf-8
     /permissive-
     /Zc:__cplusplus
-    /Zc:preprocessor
-    /std:c17)
+    /Zc:preprocessor)
 
   add_link_options(
     "LINKER:/Brepro" "LINKER:/OPT:REF" "LINKER:/WX" "$<$<NOT:$<EQUAL:${CMAKE_SIZEOF_VOID_P},8>>:LINKER\:/SAFESEH\:NO>"
--- a/cmake/macos/xcode.cmake
+++ b/cmake/macos/xcode.cmake
@@ -85,10 +85,8 @@ set(CMAKE_XCODE_ATTRIBUTE_LINKER_DISPLAYS_MANGLED_NAMES[variant=Debug] YES)
 set(CMAKE_XCODE_ATTRIBUTE_CLANG_ENABLE_OBJC_ARC NO)
 # Disable strict aliasing
 set(CMAKE_XCODE_ATTRIBUTE_GCC_STRICT_ALIASING NO)
-# cmake-format: off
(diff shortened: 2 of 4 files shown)


RytoEX (approved):
Looks okay to me at a glance.

PatTheMav (approved):

PatTheMav:
Looks good to me as well.

cmake/macos/xcode.cmake
```diff
@@ -85,12 +85,10 @@ set(CMAKE_XCODE_ATTRIBUTE_LINKER_DISPLAYS_MANGLED_NAMES[variant=Debug] YES)
...
-# cmake-format: off
-# Re-enable once the bad goto statements are fixed
-# set(CMAKE_XCODE_ATTRIBUTE_GCC_C_LANGUAGE_STANDARD c99)
-# cmake-format: on
-# Set C++ language default to c++17
-set(CMAKE_XCODE_ATTRIBUTE_CLANG_CXX_LANGUAGE_STANDARD c++17)
+# Set C language default to GNU flavour C17
+set(CMAKE_XCODE_ATTRIBUTE_GCC_C_LANGUAGE_STANDARD gnu17)
+# Set C++ language default to GNU flavour C++17
+set(CMAKE_XCODE_ATTRIBUTE_CLANG_CXX_LANGUAGE_STANDARD gnu++17)
```
RytoEX, line 91: For the record, I would very much like if we stuck to ISO C17 and C++17. However, since apparently Xcode does not list C17 as an option, we're using GNU flavors here (which apparently we were defaulting to on macOS anyway). If that's the case, we should consider using `-pedantic` and `-Wpedantic` so that devs do not accidentally make use of a GNU extension in cross-platform code that will not work elsewhere. That can be done separately, so I won't block on this here.

RytoEX (approved):
Still looks okay, though I'd prefer if we figured out why we can't use C17 with Xcode and stick to ISO C/C++ and/or add pedantic warnings to avoid issues such as accidentally using GNU extensions.

cmake/macos/xcode.cmake
PatTheMav: Xcode does support c17 and c++17, however c17 is stricter than gnu17 and would require us to fix code in some places. Also CMake defaults to gnu17 when the language standard is set via `CMAKE_C_STANDARD`.

RytoEX: If this is something that must be fixed in parts, that's fine with me. I just think that ISO C/C++ should be the longterm goal here for compatibility across our supported compilers and avoiding accidental extension usage.

PatTheMav (approved):

cmake/Modules/CompilerConfig.cmake
```diff
@@ -5,6 +5,13 @@ set(CMAKE_CXX_STANDARD 17)
 set(CMAKE_CXX_STANDARD_REQUIRED ON)
 set(CMAKE_CXX_EXTENSIONS OFF)
 
+set(CMAKE_C_STANDARD 17)
+set(CMAKE_C_STANDARD_REQUIRED ON)
+# TODO/FIXME: Investigate disabling C extensions on Linux
+if(OS_WINDOWS AND MSVC)
+  set(CMAKE_C_EXTENSIONS OFF)
+endif()
```
RytoEX, line 13: To maintain similar functionality between the OBS_CMAKE 2 and 3 paths (until we retire OBS_CMAKE 2 as an option), shouldn't this be:
```
if((OS_WINDOWS AND MSVC) OR OS_MACOS)
  set(CMAKE_C_EXTENSIONS OFF)
endif()
```
or:
```
if(NOT OS_LINUX)
  set(CMAKE_C_EXTENSIONS OFF)
endif()
```

derrod: If we still care about CMake 2.0 working, then yeah, the latter would be fine.

Conan-Kudo:
This change broke compatibility with CMake 3.16~3.20 because recognition of C17 wasn't added until CMake 3.21: https://cmake.org/cmake/help/latest/prop_tgt/C_STANDARD.html#prop_tgt:C_STANDARD

PR #8689, opened 2023-04-07
UI: SIGNAL and SLOT macros begone!

gxalpha:
### Description

Replaces all but three instances of `SIGNAL` and `SLOT` macros in the OBS codebase with the modern connection style (see https://wiki.qt.io/New_Signal_Slot_Syntax and https://doc.qt.io/qt-6/signalsandslots-syntaxes.html).

Commits are pretty small to make for easier reviewing. Can either be squashed later or left atomic for easier bisecting (for when inevitably something did go wrong), I don't mind either way.

#### Draft todo-list:
- [x] Drop Ubuntu 20.04. The problem is that in Qt 5.12, there are overloaded functions for the `QComboBox::currentIndexChanged` and `QSpinBox::valueChanged` signals. This can be dealt with by using `qOverload<>` but looks quite ugly. Alternatively, we could exclude the changes to these connections. Both solutions however are quite unclean and I, at least personally, would rather do it properly. In Qt 5.15, the overloads still exist but are deprecated, so we could disable deprecated Qt features via cmake if we wanted to retain support for Qt 5.15 after dropping Ubuntu 20.04 for some reason.

### Motivation and Context

The `SIGNAL` and `SLOT` macros expand to strings that are then interpreted at runtime. This design choice means that any bugs only occur at runtime, and even then aren't immediately obvious.
There were several examples of this happening in obs-studio in the past.

The new connection style prevents this, as signals and slots are checked at compile time. This means that incompatibilities between signal and slot arguments, nonexistent signals or slots, etc., will now be caught by the compiler and never make their way into the program in the first place.

For reference, Qt is also moving towards this syntax in their code (https://github.com/qt/qtbase/commit/283cdb0e7017a846466620bb3f8c8370d75a72ea, [QTBUG-106893](https://bugreports.qt.io/browse/QTBUG-106893))

### How Has This Been Tested?

macOS 13.3 (and previous versions, I've been working off and on on this PR for a while)
Tested every. single. change. (Insofar as that I checked that the slots are not triggered without the call, and then triggered with it).
This does not include the commit `UI: Replace SIGNAL and SLOT macros in aja-/decklink-output-ui` as I don't have the necessary hardware.

### Types of changes

### Checklist:

### DIFF
--- a/UI/adv-audio-control.cpp
+++ b/UI/adv-audio-control.cpp
@@ -16,6 +16,8 @@
 
 #define MIN_DB -96.0
 #define MAX_DB 26.0
+static inline void setMixer(obs_source_t *source, const int mixerIdx,
+			    const bool checked);
 
 OBSAdvAudioCtrl::OBSAdvAudioCtrl(QGridLayout *, obs_source_t *source_)
 	: source(source_)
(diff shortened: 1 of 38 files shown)


gxalpha:
Double-checked a few cases where I replaced slots with lambdas, and I couldn't find instances where the connection outlives the `this` object used in the lambda, because generally either the sender is the `this` object (in which case the connection gets disconnected with the destruction of the object), or the sender is a child of the `this` object (in which case the child, and with it the connection, gets destroyed before the object itself does).

gxalpha:
Rebased on current master. 
Undrafting as Ubuntu 20.04 is gone now meaning this works on all supported platforms.

PR #8692, opened 2023-04-08
cmake: Fix build without Sparkle

gxalpha:
### Description

Sets `SPARKLE_UPDATE_INTERVAL` to 0 instead of not setting it if Sparkle is disabled.

### Motivation and Context

Plist files don't like empty integers (`<integer></integer>`), and an empty integer gets generated in Info.plist if `SPARKLE_UPDATE_INTERVAL` isn't set. This is a regression of 50d1b5ccb.
This means that OBS doesn't compile without Sparkle anymore which makes me sad because I quite like being able to compile OBS.

### How Has This Been Tested?

macOS 13.3
OBS now compiles on my Mac.

### Types of changes

### Checklist:

### DIFF
--- a/UI/cmake/feature-sparkle.cmake
+++ b/UI/cmake/feature-sparkle.cmake
@@ -14,5 +14,6 @@ if(SPARKLE_APPCAST_URL AND SPARKLE_PUBLIC_KEY)
 
   target_enable_feature(obs-studio "Sparkle updater" ENABLE_SPARKLE_UPDATER)
 else()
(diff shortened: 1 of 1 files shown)


derrod (approved):
Once again I've been silly, and not thinking of people who build OBS without Sparkle

PatTheMav:
Moving forward we'll probably tell Xcode to generate the plist file for us instead of generating it ourselves, which should hopefully reduce the issues around optional keys.

PR #8707, opened 2023-04-10
UI: Sort Add Source menu case insensitively

gxalpha:
### Description

Makes the sorting of the Add Source menu case insensitive.
Before / After:
<p>
image
image
</p>

### Motivation and Context

Was annoyed that the `macOS Screen Capture` source was at the bottom and not where it alphabetically should be.

### How Has This Been Tested?

macOS 13.3.1
See screenshots above.

### Types of changes

### Checklist:

### DIFF
--- a/UI/window-basic-main.cpp
+++ b/UI/window-basic-main.cpp
@@ -5980,7 +5980,8 @@ QMenu *OBSBasic::CreateAddSourcePopupMenu()
 		QList<QAction *> actions = menu->actions();
 
 		for (QAction *menuAction : actions) {
-			if (menuAction->text().compare(name) >= 0)
+			if (menuAction->text().compare(
+				    name, Qt::CaseInsensitive) >= 0)
 				return menuAction;
 		}
 


derrod:
Seems fine. Though we can probably also just remove the "macOS" part of the source name. It's not like you have other capture methods available on macOS and we don't call it "Windows Screen Capture" or "Linux Screen Capture" on other platforms either.

derrod (approved):

gxalpha:
We do call it `Screen Capture (XSHM)` (which is a type of Display Capture I think?) so there is precedent for inconsistent naming :p.
The `macOS` in the name was mostly to highlight that the source works differently from capture sources on other operating systems, as it's a unified capture source that includes multiple capture types (and audio). If we can agree on consistent naming schemes for capture sources across all platforms (answering the questions: What does "Display Capture", "Window Capture", and "Screen Capture" refer to? How do we name unified sources? How do we outline technical differences (XSHM vs Xcomposite vs Pipewire, CGDisplayStream vs ScreenCaptureKit)?) I'm totally in favour of that.
However I'd rather not just change one source name now, only to have to change it again (invalidating every guide we have and every YouTube video out there multiple times).

Issue #3385, opened 2020-09-01
obs_frontend_set_current_scene crashes if not used in Studio Mode


woganmay:
## Platform
Operating system and version: Windows 10 v1909 Build 18363.1016
OBS Studio version: 25.0.8 (64 bit)

## Expected Behavior
I'm using a Lua script to automatically change a scene after a certain period of time - see link here: https://gist.github.com/woganmay/78bad7577b52d0fbf09b2cefcd8a131f

I'm expecting that, after switching to Scene A, that OBS will automatically switch to Scene B after 5000 milliseconds. The callbacks and countdown timer work normally, and the scene does transition if I'm in Studio Mode.

## Current Behavior
If I trigger this script while not in Studio mode, OBS will crash when attempting to call obs_frontend_set_current_scene. If I'm in Studio Mode, the scene switches normally.

## Steps to Reproduce
1. Load the Lua plugin linked above
2. Create Scene A, Scene B and a third scene
3. Without Studio Mode enabled, switch to Scene A and wait 3 seconds
4. OBS becomes unresponsive and has to be force-quit
5. Re-launch OBS, enable Studio Mode, stage Scene A and Transition to it
6. Within 3 seconds, the current scene switches to Scene B

Video: https://www.youtube.com/watch?v=ljFxT26Pi9g

## Additional information
None

Fenrirthviti:
Can you also include the crashlog?

woganmay:
There aren't any crash logs. The last crash I have in my folder is from June:

![image](https://user-images.githubusercontent.com/225550/91908140-62d65d80-ecab-11ea-96fa-796b26139183.png)

Is there another place I can get that from?

Scrxtchy:
Doesn't crash for me, OBS stops responding

Scrxtchy:
It appears that in the OBS stops responding between within this block of code
https://github.com/obsproject/obs-studio/blob/master/UI/window-basic-main-transitions.cpp#L701-L704
In doing so, it's sending an event for `OBS_FRONTEND_EVENT_PREVIEW_SCENE_CHANGED`, where this should be sending a `OBS_FRONTEND_EVENT_SCENE_CHANGED`. It is likely that due to this, and the program being not being in PreviewProgram mode, something is bricking the UI

WizardCM:
@Scrxtchy Hmm, my wild guess is that highlighted block should have a `IsPreviewProgramMode()` check on it too (similar to further up in the function).

[...]

PR #8715, opened 2023-04-11
obs-frontend-api: Fix deadlock when setting current scene

cg2121:
### Description
OBS would freeze if obs_frontend_set_current_scene was called from a script. This fixes it by allowing Qt to select the best connection type in invokeMethod

### Motivation and Context
Fixes https://github.com/obsproject/obs-studio/issues/3385

### How Has This Been Tested?
Used script in above issue

### Types of changes

### Checklist:

### DIFF
--- a/UI/api-interface.cpp
+++ b/UI/api-interface.cpp
@@ -99,14 +99,14 @@ struct OBSStudioAPI : obs_frontend_callbacks {
 	void obs_frontend_set_current_scene(obs_source_t *scene) override
 	{
 		if (main->IsPreviewProgramMode()) {
-			QMetaObject::invokeMethod(
-				main, "TransitionToScene", WaitConnection(),
-				Q_ARG(OBSSource, OBSSource(scene)));
+			QMetaObject::invokeMethod(main, "TransitionToScene",
+						  Q_ARG(OBSSource,
+							OBSSource(scene)));
 		} else {
-			QMetaObject::invokeMethod(
-				main, "SetCurrentScene", WaitConnection(),
-				Q_ARG(OBSSource, OBSSource(scene)),
-				Q_ARG(bool, false));
+			QMetaObject::invokeMethod(main, "SetCurrentScene",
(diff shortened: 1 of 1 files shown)


notr1ch:
While this may fix this one instance, many other frontend and libobs APIs are also affected by similar behavior when invoked from a timer callback in the graphics thread. Scripts need a refactor in how they interact with OBS to properly fix the root cause of this issue.

cg2121:
Updated to remove the AutoConnection part, as invokeMethod uses it as default, anyway.

storm1er:
I'm currently using a custom build 29.1.3 with this fix (and nothing else). There's a weird issue: it does nothing if I'm recording of streaming, but works if I'm not.

notr1ch:
Looking at this again, this also constitutes a breaking API change: a plugin calling `obs_frontend_set_current_scene(x)` should expect to get `x` back from a subsequent call to `obs_frontend_get_current_scene`. With a queued connection this is no longer guaranteed.

storm1er:
@notr1ch I didn't encounter the case you're talking about but I may not understand fully what you mean by "_With a queued connection_" 

_Also, is there a way to retrieve the artifacts from this pull request (they expired) so [I don't have to share a build myself](https://gist.github.com/storm1er/09b8bee73cc6e31d1efd1f271e56d9bb?permalink_comment_id=4631620#gistcomment-4631620) <3_

cg2121:
As pointed out, this is not the correct fix. While it might seem like it fixes the problem, it has unintended consequences. If somebody were to get the current scene after setting the scene, it would return the wrong one as the QueuedConnection is async, meaning that calling the get_current_scene function is called before the set_current_scene one is finished. This is why the BlockingQueuedConnection was used, as this would wait while the set_current_scene function is running. The correct fix would to fix the race condition between the graphics thread and the scripting thread.

Somebody else would have to do that though, as I am not good when it comes to threading.

PR #8716, opened 2023-04-11
UI: Do not disable events when disabling codecs

derrod:
### Description

Previously Qt events were temporarily disabled while the codec disablement check ran to prevent `AdvOutRecCheckWarnings` being called several times in a row, however, this also resulted in `on_advOutRecEncoder_currentIndexChanged()` never being called (I didn't see this function previously as it does not use the `connect()` syntax but the `on_` syntax).

To avoid any more issues related to this simply accept that `AdvOutRecCheckWarnings()` may be called twice in quick succession, this should not have any negative effects aside from wasting CPU cycles. The call at the end of `AdvOutRecCheckCodecs()` will now only happen if neither encoder had been changed.

### Motivation and Context

Fix properties still showing after encoder has been invalidated.

### How Has This Been Tested?

Mucked about with settings for a bit.

### Types of changes

### Checklist:

### DIFF
--- a/UI/window-basic-settings.cpp
+++ b/UI/window-basic-settings.cpp
@@ -5088,17 +5088,18 @@ void OBSBasicSettings::AdvOutRecCheckCodecs()
 	QString streamAudioEncoder =
 		ui->advOutAEncoder->currentData().toString();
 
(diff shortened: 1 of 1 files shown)


RytoEX (approved):
This seems fine at a glance. Would appreciate a sanity check from someone else though.

Issue #7824, opened 2022-11-21
OBS Crashes when switching scenes by hotkey and an Image source file dialog box is open


coreyp1:
### Operating System Info

Windows 10

### Other OS

Duplicated on Windows 11

### OBS Studio Version

28.1.2

### OBS Studio Log URL

https://obsproject.com/logs/MAJHMyhbbbz2A56U

### OBS Studio Crash Log URL

It didn't create a crash log.

### Expected Behavior

1. Select Scene containing an image source.
2. In the "settings" area for the image source, select "Browse" (as in, browse for a new image).
    ![image](https://user-images.githubusercontent.com/168477/203176470-4c4cc58b-2f79-4dc4-8cc8-82c7cd998641.png)
3. Do not select a file or close the file dialog box.  Rather, use a hotkey to change to a scene that does not contain the image source.
4. OBS should change to that scene.  Wether or not the file browser window should remain open or functional is up to the OBS dev team

### Current Behavior

1. Select Scene containing an image source.
2. In the "settings" area for the image source, select "Browse" (as in, browse for a new image).
3. Do not select a file or close the file dialog box.  Rather, use a hotkey to change to a scene that does not contain the image source.
4. ***OBS crashes.***  Nothing appears in the logs.  No crash report is generated.

### Steps to Reproduce

1. Create 2 scenes.  In scene ***A***, add an ***Image*** source.  Do not add this source to scene ***B***.
2. Add a hotkey to change to scene ***B***.
3. Select scene ***A***.
4. Select the ***Image*** source in scene ***A***.
5. Click the "Browse" button to open the file dialog.
    ![image](https://user-images.githubusercontent.com/168477/203176470-4c4cc58b-2f79-4dc4-8cc8-82c7cd998641.png)
6. While the file dialog is open, press the hotkey to change to scene ***B***.
7. OBS will crash, but nothing will appear in the logs and no crash report will be generated.

### Anything else we should know?

I have replicated this on my desktop, too.  It is much more powerful than my laptop (the desktop has an actual GPU, much more RAM, Windows 11, doesn't have any OBS plugins installed, etc), and the results are the same.  It has completely different scenes, too, but the crash reliably happens from the same steps.

WizardCM:
This is caused by trying to clear the source toolbar.

https://github.com/obsproject/obs-studio/blob/03c837d448d55fbb242eaa87080a5dbe3090bd63/UI/window-basic-main.cpp#L3038

[...]

PR #8725, opened 2023-04-12
 UI: Delete selection dialogs when parent is deleted

cg2121:
### Description
In the source context bar, when an user opens a selection dialog
(font, color or file), and they switch scenes (causing the source
toolbar to be deleted), OBS would crash, as the dialog is no longer
valid. This is fixed by deleting the dialog when the toolbar is
deleted.

### Motivation and Context
Fixes https://github.com/obsproject/obs-studio/issues/7824

### How Has This Been Tested?
Made sure the selection dialogs were deleted when switching scenes.

### Types of changes

### Checklist:

### DIFF
--- a/UI/context-bar-controls.cpp
+++ b/UI/context-bar-controls.cpp
@@ -4,8 +4,6 @@
 #include "obs-app.hpp"
 
 #include <QStandardItemModel>
-#include <QColorDialog>
-#include <QFontDialog>
 
 #include "ui_browser-source-toolbar.h"
 #include "ui_device-select-toolbar.h"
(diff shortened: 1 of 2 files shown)


WizardCM:
Unfortunately this only fixes the crash with hotkeys. It'll still occur when switching scenes via WebSocket or third party plugins/scripts. The better way to fix this would be by properly using the relevant dialog constructors rather than the helper functions, and automatically closing the dialog during the destruction of the toolbar.

cg2121:
Updated it to delete the selection dialogs, when scenes are switched.

CodeYan01:
While this is about the toolbar, i've seen this happen with my media playlist plugin that creates an editable list property. i have code that will use `obs_source_update_properties` when the video has ended (as i want the property to show the file name of the current file in the editable list). if one were to add file/folder/url and have the dialog open while the video ends (thereby calling that code that calls `obs_source_update_properties`), obs crashes with a similar QT crash log. It also happens if the user is dragging an item in the list while the `obs_source_update_properties` is called. Of course, this is due to the properties being recreated, and what I did is likely unsupported, but I wanted to mention it anyway.

PatTheMav:
Could it be that this is a Windows-specific issue? Because in my tests the file browser, font picker, and colour picker are all modal dialogs, so I cannot even change the scene.

Or did we fix this in the meantime - @gxalpha does this ring a bell?

PatTheMav:
> It happens when you switch scenes with hotkeys.

Thanks, got it to crash. It also blocked keyboard input in all other applications, which is even worse..

I do wonder if the easier fix would be to ignore hotkeys affecting the main window when a modal window is open. Allowing hotkeys to affect the underlying (blocked) Window seems wrong and runs counter to the idea of a modal dialog.

cg2121:
Closing in favor of https://github.com/obsproject/obs-studio/pull/9981

PR #8735, opened 2023-04-13
Destroy frontend API on shutdown

### LINKED ISSUE #8370: Websocket crash when using Touch Portal
### Operating System Info

Windows 11

### Other OS

_No response_

### OBS Studio Version

Other

29.0.0

### OBS Studio Log URL

https://obsproject.com/logs/OaGo2P6PBwcKuYKL

### OBS Studio Crash Log URL

https://obsproject.com/logs/EfU7jNAvUBIjHvpV

### Expected Behavior

OBS should not crash.

### Current Behavior

OBS experiences a websocket related crash. I am posting this on behalf of a user in the community discord and although I don't have clear repro steps, I still think it would be best to track this.

### Steps to Reproduce

Sorry for the lack of repro steps, the only common thread amongst these crashes that I've seen have been that Touch Portal was involved in the setup. I don't use it personally and have not gotten a response from the user who originally posted about the crash when I inquired for more information. 

_No response_

notr1ch:
### Description
Destroys the frontend API after the `OBS_FRONTEND_EVENT_EXIT` event has finished and fixes some internal use of the frontend API that violates this.

### Motivation and Context
Many plugins (and even some parts of OBS itself) continue to call the frontend API after the `OBS_FRONTEND_EVENT_EXIT` event . This is very dangerous as OBS is in the process of shutting down so there are no guarantees that the functions they want to call will be valid or the data returned won't be in the middle of destruction, resulting in crashes.

Fixes #8370.

### How Has This Been Tested?
Ran OBS with default set of plugins, no errors logged or broken behavior noticed. Some 3rd party plugins may be relying on this previously dangerous behavior though so this needs testing well.

### Types of changes

### Checklist:

### DIFF
--- a/UI/obs-frontend-api/obs-frontend-api.cpp
+++ b/UI/obs-frontend-api/obs-frontend-api.cpp
@@ -13,7 +13,7 @@ void obs_frontend_set_callbacks_internal(obs_frontend_callbacks *callbacks)
 static inline bool callbacks_valid_(const char *func_name)
 {
 	if (!c) {
-		blog(LOG_WARNING, "Tried to call %s with no callbacks!",
+		blog(LOG_ERROR, "Tried to call %s with no callbacks!",
 		     func_name);
 		return false;
 	}
--- a/UI/source-tree.cpp
+++ b/UI/source-tree.cpp
@@ -641,6 +641,9 @@ void SourceTreeModel::OBSFrontendEvent(enum obs_frontend_event event, void *ptr)
 		stm->SceneChanged();
 		break;
 	case OBS_FRONTEND_EVENT_EXIT:
+		stm->Clear();
+		obs_frontend_remove_event_callback(OBSFrontendEvent, stm);
+		break;
 	case OBS_FRONTEND_EVENT_SCENE_COLLECTION_CLEANUP:
 		stm->Clear();
 		break;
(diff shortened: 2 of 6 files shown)


RytoEX:
Could we consider merging the documentation clarification/change immediately and then the behavior change for the release after 29.1?

tt2468:
I ended up implementing https://github.com/obsproject/obs-websocket/commit/a0ffe16e911a2cd4de22ed1bd44c80e8adc5fa49 which solves the crash issue on the obs-websocket side. This PR doesn't disable the frontend API during other times of undefined state (like scene collection change). I'm not aware of anything outside of obs-websocket where crashes were happening that this would fix, so in my opinion it's probably best to not change the frontend-api behavior.

notr1ch:
> it's probably best to not change the frontend-api behavior

Could you elaborate on your reasoning for this? That both OBS and obs-websocket were calling the API after shutdown highlights how easy it is to make this mistake, we should make this behavior a safe no-op instead of random crashes or invalid data.

tt2468:
I'm worried about:
- There are surely plugins which depend on the frontend API during shutdown to be able to release resources and other things
- This would not have actually fixed the crash happening in obs-websocket, as the crashes happening were due to `obs_frontend_get_current_scene()` returning NULL
- As mentioned previously, there's probably an equal chance of problems happening during scene collection switch, which this PR does not cover

notr1ch:
I don't think those are good reasons for not fixing this behavior though. Plugins depending on the API during shutdown are already in dangerous territory as the API functions are potentially pointing to destroyed garbage. If it's only for cleanup then it doesn't really matter as OBS is exiting anyway, we may get some additional memory leaks logged but at least we won't crash.

Scene collection switches won't have invalid API pointers like a shutdown will. While it's true that the data returned may be indeterminate, an API client can / should listen for the changing / changed events to know when it is safe to query the frontend values again. If it's deemed necessary we can also turn off the API during collection changes but I don't see that behavior as being as bad as the shutdown issues.

PR #8751, opened 2023-04-15
UI: Explicitly make Ok button in properties dialog default

### LINKED ISSUE #8749: Properties window begins with focus on the Defaults button rather than OK
### Operating System Info

Windows 10

### Other OS

_No response_

### OBS Studio Version

29.1.0-beta4

_No response_

### OBS Studio Log URL

https://obsproject.com/logs/FCMCeTQcghzaOLO2

### OBS Studio Crash Log URL

_No response_

### Expected Behavior

Just like previous behavior, focus should default to the OK button when a source is created.

### Current Behavior

image
When I create a new source, press ok, which opens the Properties window, Focus defaults to the Defaults buttons, so pressing Enter clicks it instead of OK. Also happens on simply opening the window. This breaks user workflows where they would add sources or edit properties and just press Enter to close the window.

### Steps to Reproduce

1. Create a source and press OK after the name prompt.
2. Press Enter
3. The defaults button warning dialog shows up, showing that the focus is on the Defaults button.
4. Close the window in any way.
5. Open it again and do step 2-3 again.
...

Has been there since beta1.

gxalpha:
### Description

Explicitly sets the Ok button in the source properties dialog to be the default button of the button box.

Interestingly, the [Qt docs](https://doc.qt.io/qt-6/qdialogbuttonbox.html#details) state that
> If you want a specific button to be default you need to call [QPushButton::setDefault](https://doc.qt.io/qt-6/qpushbutton.html#default-prop)() on it yourself.

but also that 
> if there is no default button set [...] the first push button with the accept role is made the default button when the QDialogButtonBox is shown

which would indicate that Ok should already be the default (since RestoreDefaults shouldn't have the accept role), so either we're running into a Qt bug or this fix isn't actually the fix. However I'm not running into the reported problem myself, so I need someone (ideally the reporter of the bug) to confirm this.

#### Draft TODO-list:
- [ ] Have someone actually test this change on an effected operating system.

### Motivation and Context

Hopefully fixes #8749

### How Has This Been Tested?

On my Mac: Tested with a small change to this PR that using `QDialogButtonBox::RestoreDefaults` would make the "Defaults" button the default when clicking enter. However currently the "OK" button already is the default on my machine so this PR itself doesn't make any difference.

Needs testing from effected people.

### Types of changes

### Checklist:

### DIFF
--- a/UI/window-basic-properties.cpp
+++ b/UI/window-basic-properties.cpp
@@ -63,6 +63,7 @@ OBSBasicProperties::OBSBasicProperties(QWidget *parent, OBSSource source_)
 	setWindowFlags(windowFlags() & ~Qt::WindowContextHelpButtonHint);
 
 	ui->setupUi(this);
+	ui->buttonBox->button(QDialogButtonBox::Ok)->setDefault(true);
 
 	if (cx > 400 && cy > 400)
 		resize(cx, cy);


CodeYan01:
https://user-images.githubusercontent.com/65320293/232244002-a4978945-7e05-4c80-8e74-1578413f146f.mp4

It does not fix it. Pressing enter still clicks the Defaults button.

CodeYan01:
However, now that I retested on 29.0.2, it does seem that the behavior I was expecting is not actually the case. Opening Properties window and pressing Enter does nothing, and I have to press tab and then Enter so it focuses the OK button. I have no idea how I got the idea that it contradicted my muscle memory.

gxalpha:
Nothing I can do then, as I’m not experiencing it myself, as stated.

PR #8752, opened 2023-04-15
libobs-d3d11: Better HAGS detection

derrod:
### Description

Queries and logs HAGS status directly rather than relying on registry entries.

Largely derived from capabilities check in NVIDIA's Streamline Framework (MIT licensed): https://github.com/NVIDIAGameWorks/Streamline/blob/f33459c2342498b6123d702f8a9dd0ce1733ea02/source/plugins/sl.common/commonInterface.cpp#L123

### Motivation and Context

Registry is unrealiable. HAGS causes issues.

### How Has This Been Tested?

Ran on my machine, status was logged correctly:
```
	  HAGS Supported: true
	  HAGS Enabled:   false
```

### Types of changes

### Checklist:

### DIFF
--- a/libobs-d3d11/d3d11-subsystem.cpp
+++ b/libobs-d3d11/d3d11-subsystem.cpp
@@ -26,6 +26,7 @@
 #include <d3d9.h>
 #include "d3d11-subsystem.hpp"
 #include <shellscalingapi.h>
+#include <d3dkmthk.h>
 
 struct UnsupportedHWError : HRError {
 	inline UnsupportedHWError(const char *str, HRESULT hr)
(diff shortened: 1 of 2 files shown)


derrod:
In case we don't care about logging supported/enabled status for every adapter this could be superseeded by #8754 with additional logging added.

prgmitchell:
Working fine on Windows 11! 

https://obsproject.com/logs/VIVM7Rv3wPLwXCdz

libobs-d3d11/d3d11-subsystem.cpp
```diff
@@ -1326,6 +1332,16 @@ static inline void LogD3DAdapters()
...
 
+	/* D3DKMT has its own handles that we need to enumerate */
+	d3dkmt_enum.NumAdapters = MAX_ADAPTERS;
+	d3dkmt_enum.pAdapters = d3dkmt_infos;
+	res = D3DKMTEnumAdapters2(&d3dkmt_enum);
+	if (FAILED(res)) {
+		blog(LOG_INFO,
+		     "D3DKMTEnumAdapters2 failed, HAGS information will be unavailable");
+		d3dkmt_enum.NumAdapters = 0;
+	}
```
RytoEX, line 1343: Do we know why it could fail? I suppose we might just have to see how often this fails in the wild, if ever.

RytoEX (commented):
Seems okay at a glance. Curious in what situations some parts of this could fail, if at all. Would like to see the logging for #8754 in comparison.

libobs-d3d11/d3d11-subsystem.cpp
derrod: In theory it would fail if the user has more than 16 GPUs. But like with most Win32 API usage, it's checked here "just in case".

RytoEX: Good to know.

derrod:
Closing this in favour of a reworked version that ended up being a bit larger than initially thought (but also more detailed!).

PR #8754, opened 2023-04-15
UI: Add HAGS warning on startup

derrod:
### Description

Adds the following warning on startup:
2023-05-07_03-30-24_gJcbin

Uses the same detection method as #8752, but only checks the device OBS loads up on.

### Motivation and Context

HAGS causes a number of issues with OBS, so we should tell users.

### How Has This Been Tested?

Only validated that off detection works correctly, not tested on a system with HAGS enabled.

### Types of changes

### Checklist:

### DIFF
--- a/UI/data/locale/en-US.ini
+++ b/UI/data/locale/en-US.ini
@@ -119,6 +119,10 @@ SceneFilters="Open Scene Filters"
 PluginsFailedToLoad.Title="Plugin Load Error"
 PluginsFailedToLoad.Text="The following OBS plugins failed to load:\n\n%1\nPlease update or remove these plugins."
 
+# warning if HAGS is enabled
+HAGSEnabled.Title="HAGS Enabled"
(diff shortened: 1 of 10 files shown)


prgmitchell:
Tested and seems to work fine. Shows the message when it is on, doesn't when it is off. My only annoyance with this would be it popping up every time you use OBS. I don't experience any issues with HAGS on my setup currently so maybe a "do not show this message again" box?

derrod:
@WizardCM suggested that rather than using the instructions from our bot factoids (which are outdated anway, thanks Microsoft!) we could have a deep link to the correct settings page (`ms-settings:display-advancedgraphics`).

derrod:
Wording updated and added "Do not show again" checkbox.

derrod:
Closing as #8923 was merged, leaving #8752 open for more verbose logging if we want it later on.

PR #8757, opened 2023-04-15
libobs: Improve util_mul_div64 outside MSVC

jpark37:
### Description
Use inline assembly for x86_64 speed/accuracy.

Sacrifice speed for accuracy on other processors.

Continue to use original implementation for ARM on Windows.

### Motivation and Context
Make Linux behave like Windows. Make Mac behave closer to Windows.

### How Has This Been Tested?
Checked Windows (x64), Mac (ARM64), and WSL2 Ubuntu (x86_64). Modified code temporarily to verify previous failure case works now.

### Types of changes

### Checklist:

### DIFF
--- a/libobs/util/util_uint64.h
+++ b/libobs/util/util_uint64.h
@@ -22,7 +22,17 @@
 
 static inline uint64_t util_mul_div64(uint64_t num, uint64_t mul, uint64_t div)
 {
-#if defined(_MSC_VER) && defined(_M_X64) && (_MSC_VER >= 1920)
+#if !defined(_MSC_VER)
+#if defined(__x86_64__)
+	uint64_t rax, rdx;
+	__asm__("mulq %2" : "=a"(rax), "=d"(rdx) : "r"(num), "a"(mul));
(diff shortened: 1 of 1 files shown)


derrod:
clang arm64 (`-O2`) actually compiled the new code to just two instructions:
```asm
        umulh   x8, x0, x8
        lsr     x0, x8, #3
```
vs the old variant:
```asm
        umulh   x8, x0, x8
        lsr     x8, x8, #11
        msub    w9, w8, w9, w0
        and     w9, w9, #0xffff
        mul     w9, w9, w10
        mov     w10, #1000
        lsr     w9, w9, #19
        madd    x0, x8, x10, x9
```
So chances are it might actually be faster on ARM as well? 😛

jpark37:
godbolt is useful for seeing what the non-inlined code generation looks like.
It's probably more than nanoseconds since `__udivti3` has to do a lot of juggling to emulate 128 bit division, and it's even more precise than DIVQ since it handles results larger than 64 bits.
I did look at the ARM assembly list, but it didn't seem to have any 128 -> 128 or 128 -> 64 bit divide instructions, despite having 64 -> 128 multiply as a combination of MUL and UMULH.

PR #8761, opened 2023-04-16
obs-outputs: Fix enhanced RTMP frame type not being set

derrod:
### Description

Per the spec, non-metadata packets still use the frame type. In OBS's use case this should either be 1 (key frame) or 2 (inter frame), but never 0 (reserved/undefined). The enum for inter frames existed but was not used here due to an oversight.

### Motivation and Context

Send spec-conforming stream.

### How Has This Been Tested?

Tested HEVC streaming to YouTube, still worked as expected.

### Types of changes

### Checklist:


TroyKomodo:
👍 thanks

RytoEX (approved):
Off-thread discussion indicates that this seems fine.

PR #8766, opened 2023-04-17
libobs: Add new signals and private support for output, encoder and service

aet:
### Description

- Add output_update/update signals when obs_output_update called
- Add obs_source_create_private equivalent for output, encoder and service and update UI to use it for test instances
- Add obs_source_dosignal equivalent for output, encoder and service
- Add source_create/source_destroy signal equivalent for output, encoder and service

### Motivation and Context

OBS is missing the functionality to listen UI handled service/output changes through signals. You could use a tick callback to continuously probe for seeing when the instance has been changed or created at all, but it's not really efficient when we could have global signals similar to what sources have.

Second, they're lazily created. When someone is starting streaming/recording the first time, that's when you have limited a time window to react. Without modifications, it's too late to validate or enhance output settings through a plugin, to change output to use a custom obs_view for example.

### How Has This Been Tested?

- Tested with a custom plugin on a Mac and a PC

### Types of changes

### Checklist:

### DIFF
--- a/UI/window-basic-auto-config-test.cpp
+++ b/UI/window-basic-auto-config-test.cpp
@@ -198,12 +198,12 @@ void AutoConfigTestPage::TestBandwidthThread()
 	const char *serverType = wiz->customServer ? "rtmp_custom"
 						   : "rtmp_common";
 
-	OBSEncoderAutoRelease vencoder = obs_video_encoder_create(
-		"obs_x264", "test_x264", nullptr, nullptr);
-	OBSEncoderAutoRelease aencoder = obs_audio_encoder_create(
-		"ffmpeg_aac", "test_aac", nullptr, 0, nullptr);
-	OBSServiceAutoRelease service = obs_service_create(
-		serverType, "test_service", nullptr, nullptr);
+	OBSEncoderAutoRelease vencoder = obs_video_encoder_create_private(
+		"obs_x264", "test_x264", nullptr);
+	OBSEncoderAutoRelease aencoder = obs_audio_encoder_create_private(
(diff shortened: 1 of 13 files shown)


PatTheMav:
As there has been a chain of PRs and issues this seems to be related to, I struggle to find an example of what these signals would be used for? Is it for arbitrary plugins to hook into events related to recordings/broadcasts?

aet:
@PatTheMav It's understandable that core plugins have a little more breathing room to build functionality, when there's not always a solid API or interfaces in place yet that 3rdparty plugins could also use.

I have a plugin that would like to achieve few things consistently (see side note about lazily created instances above):
- Validate encoder settings, the core encoder instances created by UI or potentially other plugins that provide custom output                                                                                                 
- Allow plugins to avoid dealing with creating custom UI for output/encoder settings, tracking support for other encoders than x264/nvenc. An earlier PR at https://github.com/obsproject/obs-studio/pull/7797 added extra flexibility for obs_views, but you can't easily change UI encoder/output view from main_mix into something custom.

For core encoder/output, you could use a callback tick to observe when things are created but it's not efficient and you're bound to miss your validation window the first time recording/streaming is triggered with lazily created instances. One could even add ugly workarounds to restart the recording/streaming on first instance, but it's definitely not ideal - extending signal behavior from sources to other types would help with this and it wouldn't change any existing behavior.  Felt it'd be nice for overall consistency between instance types (source, output, encoder, service) behavior

I have been given a workaround that you can build a separate UI, create custom encoder/outputs in the plugin but I'd like to have an option that works without relying on UI/ops-frontend-api

PR #8773, opened 2023-04-19
CI: Update obs-crowdin-sync to latest

Vainock:
### Description
Updates obs-crowdin-sync to the latest commit https://github.com/obsproject/obs-crowdin-sync/commit/30b5446e3b5eb19595aa68a81ddf896a857302cf which includes all changes from https://github.com/obsproject/obs-crowdin-sync/pull/13.
I'm also re-adding the `fetch-depth` option, as it was removed with the workflow overhaul.

### Motivation and Context
Fix the translations pull not working because Crowdin changed stuff in their API and some QOL changes described in the PR.

### How Has This Been Tested?
By testing all scenarios I could come up with in a forked org on GitHub and a private Crowdin project + unit tests.

### Types of changes

### Checklist:

### DIFF
--- a/.github/workflows/dispatch.yaml
+++ b/.github/workflows/dispatch.yaml
@@ -60,7 +60,8 @@ jobs:
         with:
           submodules: recursive
           token: ${{ secrets.CROWDIN_SYNC_GITHUB_PAT }}
-      - uses: obsproject/obs-crowdin-sync/download@0.2.1
+          fetch-depth: 0
+      - uses: obsproject/obs-crowdin-sync/download@30b5446e3b5eb19595aa68a81ddf896a857302cf
 
   steam-upload:
     name: Upload Steam Builds 🚂
--- a/.github/workflows/push.yaml
+++ b/.github/workflows/push.yaml
@@ -98,7 +98,7 @@ jobs:
 
       - name: Upload US English Language Files 🇺🇸
         if: fromJSON(steps.checks.outputs.hasChangedFiles)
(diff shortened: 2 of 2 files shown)


PatTheMav:
The main question is if that repository can be breached/exploited independently of the main repo. Because branch names are highly insecure for this as you can easily get compromised code to run on your own repo (and the primary reason why this is discouraged) and thus also exfiltrate all our repository secrets (and those are per-repo, not per org).

Vainock:
I understand that concern though aren't tags equally dangerous (which the obs-studio repo uses lots of)? In which way exploitable? IMO unrealistic.

PatTheMav:
> I understand that concern though aren't tags equally dangerous (which the obs-studio repo uses lots of)? In which way exploitable? IMO unrealistic.

Correct, which is why we use commit hashes for actions that use actual repository secrets (or at least we should and it should be corrected where we don't).

But as I said, I'm not sure how isolated a breach would be if it occurred just on the crowdin-action repo

gxalpha:
If we want to avoid versioning it might make sense to pin to specific commit instead.
I would agree that using the `main` branch isn't a great idea, not because of security concerns (as noted above this is an obsproject repo after all), but because using a tag or commit would allow quickly reverting an update if something explodes for whatever reason by simply reverting the updating commit in obs-studio.

Vainock:
Okay. I'm going to update this PR to use the latest commit introduced by https://github.com/obsproject/obs-crowdin-sync/pull/13 after it got merged.

PR #8777, opened 2023-04-19
Add license declaration files for bundled dependencies

Conan-Kudo:
### Description
This change adds license declaration files for bundled dependencies that included them only in their source files so that Linux packagers can capture them for compliance purposes.

### Motivation and Context
As part of getting OBS reviewed and packaged for Fedora, I had to audit the licensing for OBS and identify the licenses for applicable bundled dependencies.

### How Has This Been Tested?
N/A

### Types of changes

### Checklist:


plugins/obs-qsv11/obs-qsv11-LICENSE.txt
tytan652, line 1: libmfx is maybe an another license. But `obs-qsv11` can be only GPL 2 and can't be dual-licensed since it will go against libobs GPL 2.

Also if it's only for Linux, this License is not needed since if I followed correctly system headers are plan to be used for Linux QSV

deps/libff/LICENSE.libff
tytan652, line 1: libff got modified through PRs so it might no longer be ISC licensed and might requires PR/commit authors agreement.

Conan-Kudo: I just grabbed what the license headers say. I would expect that the "inbound=outbound" principle applies here, and code modified that has that license declaration still follows that license.

plugins/obs-qsv11/obs-qsv11-LICENSE.txt
Conan-Kudo: That's true, I need to fix what this license notice says.

Conan-Kudo: Actually, no, I can't change this, because this is the license header that the `obs-qsv11.c` file states.

Conan-Kudo: libmfx is MIT licensed, and I didn't create a file for that.

Conan-Kudo: It is not true that the plugin code can't be dual licensed BSD/GPL. It just means as part of the larger collection, the effective license is GPL. The plugin still could be extracted and put into a project where the effective license is BSD instead.

tytan652: > The plugin still could be extracted and put into a project where the effective license is BSD instead.

The plugin would still use GPL 2 code since it relies on libobs, which explain my confusion seeing this dual license.

deps/libff/LICENSE.libff
tytan652: Can be put as resolved.

Conan-Kudo:
Refreshed this to add `media-playback` to the content set.

plugins/obs-qsv11/obs-qsv11-LICENSE.txt
tytan652, line 1: After making a deeper dive in how licensed is obs-studio repo, and the obs-qsv11 folder contains files that are under `GPL-2.0-or-later OR BSD-3-Clause` and files that are under `GPL-2.0-or-later`.

Putting a license file like this one mismatch completely how licensed this folder is.

Conan-Kudo: Where did you see something different?

tytan652: Any files that don't have an header indicating the dual license are potentially `GPL-2.0-or-later` only.

PatTheMav:
Is the pattern used in this PR documented somewhere:

```
<Description of license used>

/*
 * <Actual license content>
 * /
```

It seems weird to me to use C-style block comments for text files as well as explicitly specifying the type of license (as that should usually be obvious from the license itself). We commonly simply copy the `LICENSE`, `COPYING`, and other files relating to licensing straight from the original source tree without any local modifications.

Conan-Kudo:
I copied the only license references verbatim. If there was something else more natural in-tree, I already used them.

PatTheMav:
> I copied the only license references verbatim. If there was something else more natural in-tree, I already used them.

That's not what I see - each file mentions the license type and keeps the original license as a block comment, but none of the original license files are added.

`libnsgif` comes with a file `COPYING`, which is not even correctly replicated in this PR: https://github.com/netsurf-plan9/libnsgif/blob/master/COPYING

The `simde` license also differs from the `COPYING` file available in the official repository, so where does the variant quoted in the file added in the PR come from?

Conan-Kudo:
From the sources in this codebase. I can't use external ones, I had to use the ones in the project repository.

PatTheMav:
I have to admit I'm confused by what's going on here - I don't understand why the license information contained in the source files needs to be copied into a separate file instead of using the _actual_ license files (because that's what they're for).

Conan-Kudo:
The actual license files do not exist in the vendored copies in this repository. As I don't know whether they apply or not, I am only using the ones that are actually included in the repository and pulling out copies into their own files.

PatTheMav:
> The actual license files do not exist in the vendored copies in this repository. As I don't know whether they apply or not, I am only using the ones that are actually included in the repository and pulling out copies into their own files.

Got it - would it help if we add the license files as they _should_ apply to the respective subdirectories?

Conan-Kudo:
Yes, if we had them, then some of what I'm doing here would not be needed at all.

deps/libff/LICENSE.libff
```diff
@@ -0,0 +1,18 @@
+libff is licensed under the ISC license. The license terms are fully
```
RytoEX, line 1: libff has been removed since this PR was opened. This addition is no longer necessary.

plugins/decklink/LICENSE.decklink-sdk
```diff
@@ -0,0 +1,30 @@
+decklink-sdk is licensed under to Boost Software License 1.0 (BSL-1.0).
```
RytoEX, line 1: "licensed under to" -> "licensed under the" ?

deps/media-playback/LICENSE.media-playback
```diff
@@ -0,0 +1,19 @@
+media-playback is licensed under the ISC license. The license terms are fully
+reproduced below:
+
+
```
RytoEX, line 4: Remove extra blank line.

deps/media-playback/LICENSE.media-playback
```diff
@@ -0,0 +1,19 @@
...
+/*
+ * Copyright (c) 2023 Hugh Bailey <obs.jim@gmail.com>
```
RytoEX, line 6: Please see updated Copyright in `deps/media-playback`.

plugins/obs-qsv11/obs-qsv11-LICENSE.txt
```diff
@@ -0,0 +1,60 @@
...
+A PARTICULAR PURPOSE ARE DISCLAIMED. IN NO EVENT SHALL THE COPYRIGHT
+OWNER OR CONTRIBUTORS BE LIABLE FOR ANY DIRECT, INDIRECT, INCIDENTAL,
+SPECIAL, EXEMPLARY, OR CONSEQUENTIAL DAMAGES (INCLUDING, BUT NOT
+LIMITED TO, PROCUREMENT OF SUBSTITUTE GOODS OR SERVICES; LOSS OF USE,
+DATA, OR PROFITS; OR BUSINESS INTERRUPTION) HOWEVER CAUSED AND ON ANY
+THEORY OF LIABILITY, WHETHER IN CONTRACT, STRICT LIABILITY, OR TORT
+(INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE
+OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+*/
+
```
RytoEX, line 60: Remove extra blank line.

RytoEX (changes requested):

Conan-Kudo:
@RytoEX I've addressed your feedback.

RytoEX (changes requested):
`deps` is not the appropriate commit message subject prefix for files in `libobs/` and `plugins/`. For changes to files in `libobs/`, use "libobs: ". For changes to files in `plugins/`, use the appropriate plugin directory name if isolated to their own commits, or use "plugins: " for changes to multiple plugins at once.

This could probably be three commits, one each for: deps, libobs, plugins.

Conan-Kudo:
@RytoEX Restructured the commits as requested.

plugins/obs-qsv11/obs-qsv11-LICENSE.txt
Conan-Kudo: I've adjusted the file to note that only portions of it are under this license.

RytoEX:
Will take a look at this today.

RytoEX (approved):
Seems to accurately reflect the licenses of these components.

Conan-Kudo:
Thanks! 🎉

PR #8778, opened 2023-04-19
mac-capture: Add macOS Audio Capture

jcm93:
### Description

This PR adds a new source type to macOS, macOS Audio Capture, an audio-only source for desktop-wide or application-specific audio.

In the first two commits, ``mac-screen-capture.m`` is refactored into ``mac-sck-video-capture.m``, ``mac-sck-common.h`` and ``mac-sck-common.m``, so the common parts of OBS's SCK implementation may be shared between the combined SCK source and the audio-only SCK source. Code is moved and functions renamed but no code should be functionally changed.

Next, ``mac-sck-audio-capture.m`` is added, which functions similarly to ``mac-sck-video-capture.m``, but drops all video frames. It has a similar structure to the video capture source and shares much of the same boilerplate, as well as leveraging many of the functions in ``mac-sck-common``.

Screenshot 2023-04-19 at 1 39 30 PM

Only two basic options are implemented for now, Desktop and Application audio capture. The refactor allows for more functionality to easily be extended in the future to further leverage SCK features, such as allow lists and exclude lists of programs for audio capture.

macOS Screen Capture functionality is not changed. Audio streams will still be added to new macOS Screen Capture sources.

### Motivation and Context

This addresses a longstanding need for audio-only sources on macOS without the added overhead of video streams. It is one of the most commonly requested features by Mac users, and addresses a real performance issue with combined video+audio streams when only audio is wanted, particularly for users on older hardware.

Further context: https://github.com/obsproject/obs-studio/pull/7965

### How Has This Been Tested?

Tested locally on an M1 MacBook Pro running Ventura 13.3. 

### Known Issues

~~Changing a property within the source properties does not update the source properties dialog. I implemented callbacks to update the properties window the same way that macOS Screen Capture did, but it seems that with audio sources, the source's ``.update`` function is called before the property changed callback, so when the callback is reached, it's not known that a value is changed. Was unsure of the proper resolution here, so flagging this as WIP and in draft mode until this issue is resolved.~~

resolved, see https://github.com/obsproject/obs-studio/pull/8778#issuecomment-1526155833

As another note, when refactoring ``mac-screen-capture.m``, I made several functions that were previously ``static`` and ``inline`` no longer. I am not sure if that was the best way to handle this and would defer to maintainers for whether my refactor was done correctly. There may also be further bloat in the header file, but I generally left things as they were to avoid any breaking changes.

### Types of changes

### Checklist:

Maintainer edit: Closes #7965


jcm93:
Fixed CMakeLists.txt formatting.

plugins/mac-capture/mac-sck-audio-capture.m
```diff
@@ -0,0 +1,386 @@
...
+
+	switch (sc->capture_type) {
+	case ScreenCaptureAudioDesktopStream: {
+		SCDisplay *target_display = get_target_display();
+
+		content_filter = [[SCContentFilter alloc]
+			 initWithDisplay:target_display
+			excludingWindows:[[NSArray alloc] init]];
+
+		set_display_mode(sc, target_display);
```
PatTheMav, line 111: Is this even necessary when we're just interested in audio samples?

plugins/mac-capture/mac-sck-audio-capture.m
```diff
@@ -0,0 +1,386 @@
...
+	} break;
+	}
+	os_sem_post(sc->shareable_content_available);
+	[sc->stream_properties setQueueDepth:8];
+
+	[sc->stream_properties setCapturesAudio:TRUE];
+	[sc->stream_properties setExcludesCurrentProcessAudio:TRUE];
+	[sc->stream_properties setChannelCount:2];
```
PatTheMav, line 151: If we're refactoring SCK around this, maybe investigate if this is bound to always be stereo, or if we could pass the channel count configured in OBS here.

plugins/mac-capture/mac-sck-audio-capture.m
```diff
@@ -0,0 +1,386 @@
...
+	}];
+	os_event_wait(sc->stream_start_completed);
+
+	MACCAP_ERR("init closing, returning %d\n", did_stream_start);
+	return did_stream_start;
+}
+
+static void sck_audio_capture_defaults(obs_data_t *settings)
+{
+	CGDirectDisplayID initial_display = 0;
```
PatTheMav, line 211: Please investigate if selecting a specific display makes a difference when it comes to audio output, i.e. if selecting a specific display, but another app outputs audio but is on another display, whether this influences the audio buffers we receive.

PatTheMav:
I haven't looked into the entire PR in detail yet, but I'll have to figure out if I'm happy with a semi-refactor that is somewhat hamstrung by being written mostly in C vs a more thorough refactor that moves most functionality into ObjC with all the benefits it brings (native API usage, easier String handling, ARC, atomic class properties, thread safety, ephemeral worker threads, easier handling of blocks, etc).

The approach is fine as-is, so it's probably a decent stepping stone to a more complete overhaul.

plugins/mac-capture/mac-sck-audio-capture.m
jcm93: I investigated this one during testing. Regardless of the display selected, SCK seems to pick up audio system-wide. I double checked with "Displays have separate Spaces" on and made sure that regardless of which display was selected, audio from other displays/Spaces would be picked up. I will double check and see if I missed any weird edge cases.

plugins/mac-capture/mac-sck-audio-capture.m
jcm93: Yeah ``set_display_mode`` definitely seems unnecessary, will remove.

plugins/mac-capture/mac-sck-audio-capture.m
jcm93: According to the [docs](https://developer.apple.com/documentation/screencapturekit/scstreamconfiguration/3931901-channelcount), acceptable values are just mono or stereo. Looks like I could at least handle the mono case as well.

jcm93: Force pushed so this became outdated, my git usage is a bit hamfisted, sorry if I did something wrong. I replaced the last line with the [following block](https://github.com/obsproject/obs-studio/blob/b4c920bda7174292fc1d0d7867e87c1c57fadf4e/plugins/mac-capture/mac-sck-audio-capture.m#L137-L150):

```objc
        [sc->stream_properties setCapturesAudio:TRUE];
	[sc->stream_properties setExcludesCurrentProcessAudio:TRUE];

	struct obs_audio_info audio_info;
	BOOL did_get_audio_info = obs_get_audio_info(&audio_info);
	if (!did_get_audio_info) {
		MACCAP_ERR(
			"init_audio_screen_stream: No audio configured, returning %d\n",
			did_get_audio_info);
		return did_get_audio_info;
	}
	int channel_count = get_audio_channels(audio_info.speakers);
	if (channel_count == 1) {
		[sc->stream_properties setChannelCount:1];
	} else {
		[sc->stream_properties setChannelCount:2];
	}
```

If this looks acceptable I'll let you resolve.

plugins/mac-capture/mac-sck-audio-capture.m
jcm93: Yeah, as far as I can tell, Display Capture will always pick up all system audio regardless of selected display. Even if the external monitor has an audio output and system audio is set to that external output, a Display Capture on the native display will still pick up that audio output, including for programs launched within and from that external Space (also vice versa). Tested with Music.app, Safari, and Firefox. I'll resolve this for now unless anyone can think of another edge case.

norihiro:
Please ensure the first word for each commit title is capitalized.

plugins/mac-capture/mac-sck-audio-capture.m
PatTheMav: Logically this has the potential to set the `channelCount` to `2`, even though the actual count is `0`, which would be logically wrong.

A simpler solution could be to use something along the lines of `[sc->stream_properties setChannelCount:((channel_count > 1) ? 2 : channel_count)];`.

plugins/mac-capture/mac-sck-audio-capture.m
```diff
@@ -0,0 +1,385 @@
...
+		[sc->stream_properties setChannelCount:1];
+	} else {
+		[sc->stream_properties setChannelCount:2];
+	}
+
+	sc->disp = [[SCStream alloc] initWithFilter:content_filter
+				      configuration:sc->stream_properties
+					   delegate:nil];
+
+	//add a dummy video stream output to silence errors from SCK. frames are dropped by the delegate
```
PatTheMav, line 156: What kind of errors do you receive? It seems wasteful to me to create a `SCStreamOutputTypeScreen` if we only need an `SCStreamOutputTypeAudio`.

If there is an error, we need to report it to Apple and get it fixed.

plugins/mac-capture/mac-sck-audio-capture.m
```diff
@@ -0,0 +1,385 @@
...
+					return *stop;
+				}];
+		}
+		NSArray *target_application_array = [[NSArray alloc]
+			initWithObjects:target_application, nil];
+
+		content_filter = [[SCContentFilter alloc]
+			      initWithDisplay:target_display
+			includingApplications:target_application_array
+			     exceptingWindows:[[NSArray alloc] init]];
```
PatTheMav, line 128: See above, the empty array leaks.

plugins/mac-capture/mac-sck-audio-capture.m
```diff
@@ -0,0 +1,385 @@
...
+						*stop = TRUE;
+					}
+					return *stop;
+				}];
+		}
+		NSArray *target_application_array = [[NSArray alloc]
+			initWithObjects:target_application, nil];
+
+		content_filter = [[SCContentFilter alloc]
+			      initWithDisplay:target_display
```
PatTheMav, line 126: `target_display` and `target_application_array` need to manually released, otherwise they leak memory.

plugins/mac-capture/mac-sck-audio-capture.m
```diff
@@ -0,0 +1,385 @@
...
+				      configuration:sc->stream_properties
+					   delegate:nil];
+
+	//add a dummy video stream output to silence errors from SCK. frames are dropped by the delegate
+	NSError *error = nil;
+	BOOL did_add_output = [sc->disp addStreamOutput:sc->capture_delegate
+						   type:SCStreamOutputTypeScreen
+				     sampleHandlerQueue:nil
+						  error:&error];
+	if (!did_add_output) {
```
PatTheMav, line 162: If possible release `content_filter` after it's been passed to `initWithFilter` as it leaks otherwise.

plugins/mac-capture/mac-sck-audio-capture.m
jcm93: Without adding a video stream output, ``[ERROR] _SCStream_RemoteVideoQueueOperationHandlerWithError:712 stream output NOT found. Dropping frame`` will be reported by SCK on every frame.

My understanding from conversations with Cassidy on Discord is that the performance penalty for starting a video stream is not material, since either way, the video frames are in system memory and are being exposed for SCK to use (i.e. no further copies are done just by starting a video stream). Anyone with greater knowledge should jump in here though.

I also looked at setting ``SCStreamConfiguration``'s ``minimumFrameInterval`` to something high to minimize whatever work is being done on the back end to expose the frames. However, on OBS's end, I could not tell any difference in profiling between starting a dummy video stream and only an audio stream, so I figured it wiser to leave it at the default to avoid any potential side effects.

PatTheMav: Fine by me then - I would probably use a dummy delegate that returns immediately when the protocol method is called, but that's a micro-optimisation then.

plugins/mac-capture/mac-sck-audio-capture.m
```diff
@@ -0,0 +1,385 @@
...
+						       "application")];
+	pthread_mutex_init(&sc->mutex, NULL);
+
+	if (!init_audio_screen_stream(sc))
+		goto fail;
+
+	return sc;
+
+fail:
+	obs_leave_graphics();
```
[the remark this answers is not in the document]
jcm93: Removed.

plugins/mac-capture/mac-sck-audio-capture.m
```diff
@@ -0,0 +1,385 @@
...
+	struct screen_capture *sc = data;
+
+	ScreenCaptureAudioStreamType capture_type =
+		(ScreenCaptureAudioStreamType)obs_data_get_int(settings,
+							       "type");
+	NSString *application_id = [[NSString alloc]
+		initWithUTF8String:obs_data_get_string(settings,
+						       "application")];
+
+	obs_enter_graphics();
```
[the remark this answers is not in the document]
jcm93: Seems no; removed.

plugins/mac-capture/mac-sck-video-capture.m
```diff
@@ -606,11 +319,11 @@ typedef void (^shareable_content_callback)(SCShareableContent *,
 
 fail:
 	obs_leave_graphics();
```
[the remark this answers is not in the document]
jcm93: Removed. Technically should be part of the first refactor commit since this is in video-capture but resolving for now. If I have to do commit surgery later, or if you want me to, I will add it to the refactor commit.

plugins/mac-capture/mac-sck-audio-capture.m
```diff
@@ -0,0 +1,385 @@
...
+		return target_display;
+	};
+
+	switch (sc->capture_type) {
+	case ScreenCaptureAudioDesktopStream: {
+		SCDisplay *target_display = get_target_display();
+
+		content_filter = [[SCContentFilter alloc]
+			 initWithDisplay:target_display
+			excludingWindows:[[NSArray alloc] init]];
```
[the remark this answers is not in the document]
jcm93: Fixed.

plugins/mac-capture/mac-sck-audio-capture.m
jcm93: Fixed.

plugins/mac-capture/mac-sck-audio-capture.m
```diff
@@ -0,0 +1,385 @@
...
+	[sc->stream_properties setCapturesAudio:TRUE];
+	[sc->stream_properties setExcludesCurrentProcessAudio:TRUE];
+
+	struct obs_audio_info audio_info;
+	BOOL did_get_audio_info = obs_get_audio_info(&audio_info);
+	if (!did_get_audio_info) {
+		MACCAP_ERR(
+			"init_audio_screen_stream: No audio configured, returning %d\n",
+			did_get_audio_info);
+		return did_get_audio_info;
```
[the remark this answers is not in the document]
jcm93: Fixed.

plugins/mac-capture/mac-sck-audio-capture.m
jcm93: Fixed to release ``target_application_array``. I don't see an alloc call for ``target_display`` though, and it seems to come from ``sc->shareable_content.displays[index]``, so I did not add a ``release`` for it. I will double check it later, currently having issues with the Xcode leaks instrument which I think will require a macOS or Xcode update.

plugins/mac-capture/mac-sck-audio-capture.m
jcm93: Fixed to release ``content_filter`` immediately after the ``initWithFilter`` call.

plugins/mac-capture/mac-sck-audio-capture.m
jcm93: Resolved with:

https://github.com/obsproject/obs-studio/blob/839a86c40e7c7085337ed0268a5a00f3d764a362/plugins/mac-capture/mac-sck-audio-capture.m#L148-L152

plugins/mac-capture/mac-sck-audio-capture.m
PatTheMav: You can just do an analytical build (cmd+shift+b) to have it check for these kind of leaks.

jcm93: Alright, analysis seems clear for leaks, got the leaks Instrument working as well and ran some tests and things seem clear there as well, no warnings about ``target_display``.

jcm93:
I just fixed the outstanding issue with the properties dialog not updating when the user changed the capture method.

I decided to take the approach of not caring about the old value and simply restructuring the callback to be a dedicated callback for that value, with the thought that future callbacks can simply use dedicated functions instead of a blanket ``content_settings_changed`` callback such as the one used by SCK video. The callback is now called ``audio_capture_method_changed``.

The issue, briefly repeated, is that the way SCK video works, the previous state of the struct was captured in the ``modified2`` callback. Its ``content_settings_changed`` function acts as a blanket callback for all changed properties. In SCK audio, ``.update`` is not deferred, so it is called before the callback, and the previous state of the struct is not captured when the callback is reached.

I don't know if there is a better way here, or if there is indeed a way to capture the old value with the callback for audio sources that I just missed. This works, however, and I don't see any issues with it. After I double check things, I will probably take this PR out of draft mode since it doesn't have any more outstanding functional issues (from my perspective).

jcm93:
Rebased on master and removed draft status.

norihiro:
The first letter in the commit title should be capitalized.
refactor macOS Screen Capture --> Refactor macOS Screen Capture

plugins/mac-capture/mac-sck-audio-capture.m
```diff
@@ -0,0 +1,372 @@
+#include "mac-sck-common.h"
+
+const char *sck_audio_capture_getname(void *unused __attribute__((unused)))
```
PatTheMav, line 3: You can use just `__unused` in ObjC files.

PatTheMav:
> I just fixed the outstanding issue with the properties dialog not updating when the user changed the capture method.
> 
> I decided to take the approach of not caring about the old value and simply restructuring the callback to be a dedicated callback for that value, with the thought that future callbacks can simply use dedicated functions instead of a blanket `content_settings_changed` callback such as the one used by SCK video. The callback is now called `audio_capture_method_changed`.
> 
> The issue, briefly repeated, is that the way SCK video works, on a property update, the previous state of the struct is captured in the `modified2` callback. Its `content_settings_changed` function acts as a blanket callback for all changed properties. In SCK audio, `.update` is not deferred, so it is called before the callback, and the previous state of the struct is not captured when the callback is reached.
> 
> I don't know if there is a better way here, or if there is indeed a way to capture the old value with the callback for audio sources that I just missed. This works, however, and I don't see any issues with it. After I double check things, I will probably take this PR out of draft mode since it doesn't have any more outstanding functional issues (from my perspective).

I'm in the process of rewriting another core macOS plugin in ObjC and I came to a similar conclusion:

* In general it seems easiest to just rebuild all properties from scratch if changes appear versus iterating over current and new values in nested loops to figure out any changes
* Property update callbacks tend to happen _before_ the internal state of the source changed

Also please rebase your PR as we recently changed the `clang-format` rules for ObjC files. As there are some incoming code updates to the capture plugin and necessary fixes for macOS 14 compatibility, there might be a few more things you will need to pick apart given that you broke apart the plugin to share code.

jcm93:
Yarp, looks like a decent number of conflicts but I will rebase at my earliest opportunity.

jcm93:
Alright, rebased successfully. I integrated all recent changes in ``mac-capture``, and any parts of functions that were mirrored between ``mac-sck-audio-capture.m`` and ``mac-sck-video-capture.m`` should also integrate new changes (cleaning up the block formatting, other code cleanups).

I integrated the display UUID changes for audio capture as well, though they should not be relevant since, as we've discussed, audio is captured systemwide regardless of the selected display, which still seems to be the case in Sonoma.

I think I fixed CMake formatting, but the runners are still doing their thing, if they complain I will push another fix.

plugins/mac-capture/mac-sck-common.m
```diff
@@ -0,0 +1,339 @@
...
+                screen_stream_audio_update(self.sc, sampleBuffer);
+            }
+        }
+#endif
+    }
+}
+
+- (void)stream:(SCStream *)stream didStopWithError:(NSError *)error
+{
+    NSString *errorMessage;
```
PatTheMav, line 34: `errorMessage` is created, but never used. Using `error.domain.UTF8String` doesn't seem to make much sense below (because it doesn't give you an actual error description, that'd be `error.localizedDescription`.

PatTheMav:
Tested this right now and looks quite good apart from the error message comment I made above.

plugins/mac-capture/mac-sck-common.m
jcm93: Hmm, looks like this came from https://github.com/obsproject/obs-studio/pull/9065/. I just pushed a change with ``error.domain.UTF8String`` changed to ``errorMessage.UTF8String``, since I assume that was Sebastian's intention. I'll ping @gxalpha just in case as well.

gxalpha: > I just pushed a change with error.domain.UTF8String changed to errorMessage.UTF8String, since I assume that was Sebastian's intention

Yes, that was an embarrassing mistake in the original which resulted from a test that then somehow made its way in the implementation. See https://github.com/obsproject/obs-studio/pull/9107 where I included a fix for that.

PatTheMav: Alright, will probably merge this later tonight then.

PR #8781, opened 2023-04-20
win-capture: Add frame timestamp and a new frame signal

lostmsu:
### Description

Two new features in win-capture:
- frame timestamps, that enable synchronizing with other events like user input or audio
- `CaptureHook_Frame` signal to let the capture receiver be notified that a new frame was captured

### Motivation and Context
Both features are necessary for precise video/audio synchronization.
With the frame signal it should be possible to lower latency by 1/2 of a frame on average by picking the new frame up ASAP.

### How Has This Been Tested?
Tested on Windows 11 22H2. OBS can still capture games using win-capture.
I wrote a small test program, that ensures the timestamps and signals are set correctly.

### Types of changes

@jp9000 tagging you as requested in graphics-hook-ver.h. The minor version bump is due to the new features. I am assuming adding a new field to the shared texture is not a breaking change as long as the receivers know to either read size or to limit the size to 1.8 structure's size.

### Checklist:

### DIFF
--- a/plugins/win-capture/graphics-hook-info.h
+++ b/plugins/win-capture/graphics-hook-info.h
@@ -11,6 +11,8 @@
 #define EVENT_CAPTURE_RESTART L"CaptureHook_Restart"
 #define EVENT_CAPTURE_STOP L"CaptureHook_Stop"
 
+#define EVENT_NEW_FRAME L"CaptureHook_Frame"
+
 #define EVENT_HOOK_READY L"CaptureHook_HookReady"
 #define EVENT_HOOK_EXIT L"CaptureHook_Exit"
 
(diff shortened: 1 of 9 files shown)


jp9000:
Hi there, could you specify more context for this PR please?

I'm not saying it's unlikely that we'll merge this but changing the graphics hook ABI is something we try to avoid as much as possible.

lostmsu:
@jp9000 I am trying to use the hook in a standalone tool (e.g. not OBS). The tool needs low latency, hence the change that introduces new frame signal (that one btw is completely backwards compatible - you can simply ignore the signal).

The bit with timestamps I am using for machine learning to correlate frames with user input.

I could probably just keep the change in my fork, but as a comaintainer of Python.NET I feel it is generally better experience for everyone when changes get upstreamed.

jp9000:
Alrighty, well I'd just recommend keeping this forked for now. You can always bring this up with us again later, and/or chat with us any time if you want to discuss further! But thank you in either case.

Issue #3011, opened 2020-05-31
Text blanks, fails to read from file


Meerkov:
## Platform
Windows 10 
OBS Studio version: 25.0.1 64 bit

## Expected Behavior
1) Create a Text Source with "Read File" enabled.
2) Switch to a scene where the Text Source is deleted
3) Update the text file manually in another program (or use a script in a separate process).
4) Switch back to the first scene
5) Repeat step 2-4 another 10x

Result: The text updates every time

## Current Behavior
Sometimes the TextSource will become blank even though the file is not empty.
The TextSource may sometimes recover if content is written to the file again, but sometimes will not.

## Steps to Reproduce
1) run this in OBS. Make sure to have a text source with "Read File" on one scene, but deleted on the other.
```lua
obs = obslua

function change_scene(sceneName)
	source = obs.obs_get_source_by_name(sceneName)
	if source ~= nil then
		obs.obs_frontend_set_current_scene(source)
		obs.obs_source_release(source)
	else
		print("Can't find scene: "..sceneName)
	end
end

function display_timer_tick()
	if(state == "DisplayText") then
                state == "HideTextDuringUpdate"
		change_scene("s1")
        else
                state == "DisplayText"
                change_scene("s2")
        end
end

function script_load(settings)
	print("Loaded Script.")
        obs.timer_add(display_timer_tick, 10000)
end
```
2) Update the textfile while the text is hidden.
3) Watch to see if the text appears as expected.
4) if it appears, edit the textfile and save again when it disappears. (I used a separate script to synchronize the file-write with OBS's scene switching, but you should be able to do this manually instead).
5) Repeat until it shows up blank.
6) Stop editing the file and wait to see it probably doesn't read the file until you edit it again.
If it doesn't work, try "refreshing" the lua script in OBS.

## Additional information

It might be something related to the source getting disabled during scene switching, which the documentation acknowledges "is confusing to users because it's not surfaced in the UI and sources will fail to render even though they appear to be marked visible". But I would have no idea what might cause that to happen here.

Meerkov:
It triggered for me again. This time with manually timed file-reads via a script.

```lua

function update_text(sourcename, value )
  print("Updating text for "..sourcename)
  text=obs.obs_get_source_by_name(sourcename)
  if(text ~= nil) then
	newRef=obs.obs_data_create_from_json('{"text":"'..value..'"}')
	ref=obs.obs_save_source(text)
	obs.obs_source_update(text,newRef)
[...]

PR #8783, opened 2023-04-21
image-source: Retry image load if not loaded

adampoit:
### Description

Allows an image source to attempt another load, at the regular polling interval, even if the timestamp is unchanged. This handles situations where a previous load attempt may have failed, such as the file being locked by another process.

### Motivation and Context

If an image source attempts to load an image and it fails, the image source will never reattempt to load the image until the file timestamp changes. 

I came across this issue when using an external program to update images for my stream. I noticed that occasionally an image update would coincide with the image not showing in OBS. I dug into the logs and noticed the following entries:

```
19:48:23.144: Failed to open file 'C:/Users/.../rightSongMetadata.png': Permission denied
19:48:23.145: WIC: Failed to create IWICBitmapDecoder from file: C:/Users/.../rightSongMetadata.png
19:48:23.145: gs_image_file_init_internal: Failed to load file 'C:/Users/.../rightSongMetadata.png'
19:48:23.145: [image_source: 'Right Song'] failed to load texture 'C:/Users/.../rightSongMetadata.png'
```

Upon further investigation, I saw this problem described in [this issue comment](https://github.com/obsproject/obs-studio/issues/3011#issuecomment-1437666466). While I could not recreate the text source problem described in the original issue, I was able to consistently reproduce the problem with an image source.

### How Has This Been Tested?

I tested on Windows 11 using the following PowerShell script:

```
param([bool] $lock)

$source = Join-Path $PSScriptRoot "example1.jpg"
$target = Join-Path $PSScriptRoot "example.jpg"

Remove-Item $target -Force
Start-Sleep -Milliseconds 1500
Copy-Item $source $target -Force
if ($lock)
{
	Write-Host "Locking $target"
	$fileStream = New-Object IO.FileStream($target, [System.IO.FileMode]::Open, [System.IO.FileAccess]::ReadWrite, [System.IO.FileShare]::None)
	Start-Sleep -Milliseconds 1500
	$fileStream.Close()
	Write-Host "Lock released"
}
```

If I ran the script with `-lock $true`, I was able to consistently reproduce the issue. The `lock` parameter was included in the script so I could easily reset the test.

### Types of changes

### Checklist:

### DIFF
--- a/plugins/image-source/image-source.c
+++ b/plugins/image-source/image-source.c
@@ -208,7 +208,8 @@ static void image_source_tick(void *data, float seconds)
 			time_t t = get_modified_timestamp(context->file);
 			context->update_time_elapsed = 0.0f;
 
-			if (context->file_timestamp != t) {
+			if (context->file_timestamp != t ||
+			    !context->if4.image3.image2.image.loaded) {
 				image_source_load(context);
 			}
 		}
--- a/plugins/image-source/image-source.c
+++ b/plugins/image-source/image-source.c
@@ -209,7 +209,8 @@ static void image_source_tick(void *data, float seconds)
 			context->update_time_elapsed = 0.0f;
 
 			if (context->file_timestamp != t ||
-			    !context->if4.image3.image2.image.loaded) {
+			    (!context->if4.image3.image2.image.loaded &&
+			     t != -1)) {
 				image_source_load(context);
 			}
 		}


plugins/image-source/image-source.c
```diff
@@ -208,7 +208,8 @@ static void image_source_tick(void *data, float seconds)
 			time_t t = get_modified_timestamp(context->file);
 			context->update_time_elapsed = 0.0f;
 
-			if (context->file_timestamp != t) {
+			if (context->file_timestamp != t ||
+			    !context->if4.image3.image2.image.loaded) {
```
adampoit, line 212: Another downside to this approach is that we will continually retry to load the image at the polling interval if the file is missing. This could be avoided if we confirm that `t != -1` before attempting the load.

jp9000: Hm, if it causes log spam or something then I'd say it should be handled.

adampoit: I've updated my PR to check for `t != -1` so that we can avoid attempting the image load in a loop. I'd still prefer the other approach I outlined above, but this is at least better now.

PatTheMav:
I guess the underlying root issue is that Windows has mandatory file locks (POSIX commonly doesn't use them) and if a file is locked one cannot even get the modification timestamp?

Because on POSIX systems the modification timestamp only changes if the actual file is renamed, which then also updates `context->file_timestamp` and once the file becomes available again, the actual timestamp will be different to `-1` and the file appears again.

PR #8788, opened 2023-04-21
Adjust AV1 tiles at 4K+

RytoEX:
### Description

Use 2x2 tiles in NVENC AV1 for 4K+. When resolution is 4K or higher, use 2x2 uniform tiles for NVENC AV1.

Use 2 tiles per frame in AMF AV1 for 4K+. When resolution is 4K or higher, use 2 tiles per frame in AMF AV1.

### Motivation and Context

Single-tile 4K videos are more difficult to decode. Using 1x2 or 2x2 tiles for 4K should make decoding easier.

### How Has This Been Tested?

Tested with NVENC AV1 at 4K and verified that the output was 2x2 (2 columns, 2 rows) tiles. I have not tested this with AMF AV1, but I expect this to produce 1x2 (1 column, 2 rows) tiles at 4K.

### Types of changes

### Checklist:

### DIFF
--- a/plugins/obs-ffmpeg/jim-nvenc.c
+++ b/plugins/obs-ffmpeg/jim-nvenc.c
@@ -922,6 +922,17 @@ static bool init_encoder_av1(struct nvenc_data *enc, obs_data_t *settings,
 	if (config->rcParams.rateControlMode == NV_ENC_PARAMS_RC_CBR)
 		av1_config->enableBitstreamPadding = 1;
 
+#define FOUR_K_CX 3840
+#define FOUR_K_CY 2160
+#define PIXELCOUNT_4K (FOUR_K_CX * FOUR_K_CY)
+
+	/* If 4K+, set tiles to 2x2 uniform tiles. */
+	if ((voi->width * voi->height) >= PIXELCOUNT_4K) {
+		av1_config->enableCustomTileConfig = 0;
+		av1_config->numTileColumns = 2;
+		av1_config->numTileRows = 2;
+	}
+
 	switch (voi->colorspace) {
 	case VIDEO_CS_601:
 		av1_config->colorPrimaries = 6;
--- a/plugins/obs-ffmpeg/texture-amf.cpp
+++ b/plugins/obs-ffmpeg/texture-amf.cpp
@@ -2018,6 +2018,15 @@ static void amf_av1_create_internal(amf_base *enc, obs_data_t *settings)
 	const bool is10bit = enc->amf_format == AMF_SURFACE_P010;
 	const char *preset = obs_data_get_string(settings, "preset");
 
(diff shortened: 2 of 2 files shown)


notr1ch:
Should we perhaps use pixel count instead of the h/v resolution? 2160x3840 should probably also be tiled for example.

RytoEX:
> Should we perhaps use pixel count instead of the h/v resolution? 2160x3840 should probably also be tiled for example.

We could do pixel count or check 3840x2160 and 2160x3840.

plugins/obs-ffmpeg/jim-nvenc.c
```diff
@@ -922,6 +922,17 @@ static bool init_encoder_av1(struct nvenc_data *enc, obs_data_t *settings,
 	if (config->rcParams.rateControlMode == NV_ENC_PARAMS_RC_CBR)
 		av1_config->enableBitstreamPadding = 1;
 
+#define FOUR_K_CX 3840
+#define FOUR_K_CY 2160
+#define PIXELCOUNT_4K (FOUR_K_CX * FOUR_K_CY)
+
+	/* If 4K+, set tiles to 2x2 uniform tiles. */
+	if ((voi->width * voi->height) >= PIXELCOUNT_4K) {
+		av1_config->enableCustomTileConfig = 0;
```
notr1ch, line 931: Any reason the NVENC code uses macros and the AMD code uses `constexpr`? I'd prefer they are consistent over the PR.

jp9000: C doesn't have constexpr, otherwise I'd use it. They're functionally identical otherwise.

jp9000: To answer your question, it's cleaner, and constexpr doesn't leave scope. Acts like a normal constant but is instead a compile-time constant

notr1ch: Ah, I missed that. I suppose we could use `const int` instead and let compiler optimizations take care of it for a similar result.

RytoEX: As a note for the future, C23 introduces constexpr, and constexpr ints should work once we support C23. Until then, as mentioned above, C does not have constexpr.

derrod:
Why even make it a macro/contexpre rather than just putting `(3840 * 2160)` into the if clause? It's not like it'll ever change.

jp9000:
I donno, habit I guess? It's not like it matters one way or another

RytoEX:
We decided not to merge this implementation for now. If we find there's still a need for this feature, we may revisit this in the future, and we may take a different approach. I'll keep the branch available for a while yet.

PR #8806, opened 2023-04-25
CI: Log actions runner public IP and RTMP server IP (on fail)

tt2468:
### Description
Whether or not an issue is happening can have a lot to do with the source and destination IPs of a socket. Many services use load balancing on their ingest hostnames which may make network-related issues difficult to troubleshoot. Logging both source and destination IP addresses allows parties to troubleshoot from their end.

### Motivation and Context
Intended to help troubleshoot #8549

### How Has This Been Tested?
Ran `check-services.py` - it printed my public IP address and the IP address of a domain which was intentionally offline.

### Types of changes

### Checklist:

### DIFF
--- a/CI/check-services.py
+++ b/CI/check-services.py
@@ -19,6 +19,7 @@ SERVICES_FILE = 'plugins/rtmp-services/data/services.json'
 PACKAGE_FILE = 'plugins/rtmp-services/data/package.json'
 CACHE_FILE = 'other/timestamps.json'
 GITHUB_OUTPUT_FILE = os.environ.get('GITHUB_OUTPUT', None)
+PUBLIC_IP_ENDPOINT = 'https://ifconfig.me'
 
 DO_NOT_PING = {'jp9000'}
 PR_MESSAGE = '''This is an automatically created pull request to remove unresponsive servers and services.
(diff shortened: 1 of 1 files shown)


notr1ch:
I think `socket.create_connection` should also use the resolved IP, in the case of multiple IP addresses it is not guaranteed that the one that was just resolved will be the same one returned in the next call.

CI/check-services.py
```diff
@@ -58,6 +59,12 @@
 
 context = ssl.create_default_context()
 
+def print_public_ip():
+    try:
+        resp = requests.get(PUBLIC_IP_ENDPOINT)
```
derrod, line 64: Maybe also add the timeout here like with the HLS check, just to avoid it getting stuck for a long time.

CI/check-services.py
```diff
@@ -95,7 +102,9 @@ def check_rtmp_server(uri) -> bool:
     else:
         port = 1935
 
+    ip_address = '[none]'
     try:
+        ip_address = socket.gethostbyname(hostname)
```
derrod, line 107: Technically this IP could end up being a different one than what `create_connection()` resolves to, to ensure they're the same replace `hostname` with `ip_address` when the connection is created below.

derrod (changes requested):

PR #8807, opened 2023-04-25
CI: Switch service checker to macOS

derrod:
### Description

Switches the service checker to macOS, which runs on different networks and seems to have fewer connection issues that aren't actually the service's fault.

### Motivation and Context

Finally get #8549 over with by removing the false-positive.

### How Has This Been Tested?

Ran on my fork, confirmed the wpstream server can now be connected to normally.

### Types of changes

### Checklist:

### DIFF
--- a/.github/workflows/services-json.yml
+++ b/.github/workflows/services-json.yml
@@ -45,7 +45,7 @@ jobs:
 
   service_check:
     name: Service Check
-    runs-on: ubuntu-20.04
+    runs-on: macos-latest
     needs: schema
     if: ${{ github.repository_owner == 'obsproject' && github.event_name == 'schedule' || github.event_name == 'workflow_dispatch' }}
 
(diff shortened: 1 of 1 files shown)


RytoEX:
Just a note, the macOS runners have a concurrency limit of 5 active jobs. If other macOS jobs have filled that limit at the time this tries to run, it will be stuck waiting. Have you checked if the ubuntu-22.04 image suffers from these same issues? What about the Windows image (which, unlike macOS, has no specific concurrency limit)? Can we report these issues on the GHA runner images repo, or do we still not know what the issue actually is?

derrod:
> Just a note, the macOS runners have a concurrency limit of 5 active jobs. If other macOS jobs have filled that limit at the time this tries to run, it will be stuck waiting. Have you checked if the ubuntu-22.04 image suffers from these same issues? What about the Windows image (which, unlike macOS, has no specific concurrency limit)? Can we report these issues on the GHA runner images repo, or do we still not know what the issue actually is?

This runs once a day and takes about 2 1/2 minutes, so I'm not too worried about concurrency limits. Windows/Ubuntu runners still have concurrency limits, they're just higher (20 instead of 5).

This seems to be more of an issue with Azure's network rather than the runner image and affects both Windows and Linux runner images, so I don't think it's worth reporting. The reason Mac runners work is that they're not actually hosted on Azure.

RytoEX:
I suppose my main annoyance here is that it will hide whatever issue was actually occurring with the Linux runners, and it really only seemed to affect *one* service.

derrod:
As mentioned it seems to affect Windows as well, so probably some network issue with GHA/Azure that we can do very little about. I agree though that I'd like a better solution. Perhaps we can stop relying on GHA for those checks and use some external service that's more globally distributed as well.

Issue #8665, opened 2023-04-05
P416 color format is not compatible with HDR


arm64-v9a:
### Operating System Info

Windows 11

### OBS Studio Version

29.1.0-beta1

### OBS Studio Log URL

https://obsproject.com/logs/Fu8UvOyHJM0v9Gvf

### Expected Behavior

Normal output 4:4:4 16bit/10bit video

### Current Behavior

"OBS does not support 8bit output of Rec.2100"

### Steps to Reproduce

1. Settings > Adanced > Color Format > P416 (I would prefer to use P410, but OBS does not have this. As far as I know, NVENC will automatically convert P416 to P410)
2. Settings > Adanced > Color Space > Rec.2100
3. Add Display Capture Source
4. Start Recoding

gxalpha:
P416 is only compatible with macOS ProRes encoders at this time. NVENC is probably silently falling back to NV12, causing this message. I do see room for improvement, but the expectation here should be a better error message and not NVENC working.

derrod:
That's FFmpeg, not OBS. The encoder itself does not support P416 and the obs-ffmpeg encoder plugin does not use the conversion that the FFmpeg CLI would do, instead it just falls back to NV12.

gxalpha:
Maybe in the future. However, that is not relevant to this issue as-is and would be a feature request.

For this issue to be considered fixed, PRs #8811, #8812, and similar not-yet-existing PRs for NVENC, AMF, and other encoders need to be merged to provider better error message and to avoid opaque conversions.

PR #8812, opened 2023-04-27
obs-x264: Disallow 16-bit color formats

gxalpha:
### Description

Extends the 10-bit check in x264 to 16-bit formats, which aren't supported either.
The "high-precision" terminology is used elsewhere in OBS already as well (see https://github.com/obsproject/obs-studio/commit/463bf0dff59a452edf66bbdb3ab20995bcfd25b1#diff-cd50d66da1e02db4fa473b80d6109b5799b4e619d09ac4e4487e4492ded51fc8).

### Motivation and Context

Selecting 16-bit formats with the x264 encoder would result in opaque conversion to NV12 (and as such an 8-bit recording) which we don't want.

### How Has This Been Tested?

macOS 13.3.1
Tested that instead of recording NV12, OBS now errors out when trying to start a 16-bit x264 recording with the new message:

image

### Types of changes

### Checklist:


RytoEX:
It looks like this and #8811 partially address #8665, but that we'd still need to address it in the other encoders as well?

gxalpha:
Correct, I didn't attempt to do this with NVENC and AMF as I do not know how the compatibility matrices look there and also don't have a machine to test with.

PR #8814, opened 2023-04-28
UI: Only defer property updates for input and transition sources

### LINKED ISSUE #5139: OBS_PROPERTIES_DEFER_UPDATE does not properly work for scripts
### Operating System Info

Windows 10

### Other OS

_No response_

### OBS Studio Version

27.0.1

_No response_

### OBS Studio Log URL

https://obsproject.com/logs/FjsVbgdGv7X53Zwa

### OBS Studio Crash Log URL

_No response_

### Expected Behavior

If the `OBS_PROPERTIES_DEFER_UPDATE` flag was set on the properties created in the `script_properties` function, the `script_update` should be called once when the "Scripts" window is closed.

As per OBS API [documentation](https://obsproject.com/docs/reference-properties.html): "OBS_PROPERTIES_DEFER_UPDATE - A hint that tells the front-end to defers updating the settings until the user has finished editing all properties rather than immediately updating any settings."

### Current Behavior

With the `OBS_PROPERTIES_DEFER_UPDATE` flag set, the `script_update` function is only called when the script is loaded, afterwards it is **never** called even if the settings are changed when the "Scripts" window is closed.

Looking at the [source code](https://github.com/alibaba/tblive/blob/master/libobs/obs-properties.h#L43), it states: "Only update when the user presses OK or Apply". However, the "Scripts" window only has a "Close" button. I think this is the root cause.
To be noted that changes in the settings are still persisted as they would if the `OBS_PROPERTIES_DEFER_UPDATE` wasn't set.

### Steps to Reproduce

1. Add the script below to OBS.
2. Change the setting defined.
3. Notice how the `script_update` is not called.

Script to replicate the issue.
```
import obspython as obs

def script_description():
	return "Test script."

def script_update(settings):
	print("Called script_update")

def script_defaults(settings):
	obs.obs_data_set_default_string(settings, "message", "Hello!")

def script_properties():
	props = obs.obs_properties_create()
	obs.obs_properties_set_flags(props, obs.obs_properties_get_flags(props) | obs.OBS_PROPERTIES_DEFER_UPDATE)
	obs.obs_properties_add_text(props, "message", "Message", obs.OBS_TEXT_DEFAULT)
	return props
```

gxalpha:
### Description

Deferring properties was only ever made for input sources. As other property dialogs do not have an "Ok" button that would cause an update to happen, if the deferred flag was set the callback would never be called.
Also clarifies the docs to reflect this.

### Motivation and Context

Fixes #5139

### How Has This Been Tested?

Tested that whereas before, setting the flag on scripts and filters would prevent the update signal from being called basically ever, it now is a no-op and the signal is called in all cases.
Tested that with normal sources, the flag still defers saving until pressing Ok/Apply.

### Types of changes

### Checklist:

### DIFF
--- a/UI/properties-view.cpp
+++ b/UI/properties-view.cpp
@@ -88,18 +88,32 @@ Q_DECLARE_METATYPE(media_frames_per_second);
 
 void OBSPropertiesView::ReloadProperties()
 {
+	deferUpdate = false;
 	if (weakObj || rawObj) {
 		OBSObject strongObj = GetObject();
 		void *obj = strongObj ? strongObj.Get() : rawObj;
-		if (obj)
+		if (obj) {
 			properties.reset(reloadCallback(obj));
+
+			if (obs_obj_get_type(obj) == OBS_OBJ_TYPE_SOURCE) {
+				enum obs_source_type type = obs_source_get_type(
+					(obs_source_t *)obj);
+				if (type == OBS_SOURCE_TYPE_INPUT ||
+				    type == OBS_SOURCE_TYPE_TRANSITION) {
+					uint32_t flags =
+						obs_properties_get_flags(
+							properties.get());
+					deferUpdate =
+						(flags &
+						 OBS_PROPERTIES_DEFER_UPDATE) !=
+						0;
+				}
+			}
+		}
 	} else {
 		properties.reset(reloadCallback((void *)type.c_str()));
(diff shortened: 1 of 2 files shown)


gxalpha:
Updated this to also allow the flag for transitions. I think there are some bugs/issues related to transitions and this flag (where the duplicated source would always update immediately (or rather, recreated entirely?)), but those already on master.

PR #8815, opened 2023-04-28
updater: Support delta patch bundles & Refactor

derrod:
### Description

Adds support for delta patch bundles to the OBS updater. These bundle all the delta patches required for patching between two versions into a single file, improving update speeds by reducing total I/O and downloading one larger file instead of potentially hundreds of small ones (very slow, especially HTTP/1.1).

Additionally, a second commit refactors a number of a things to be cleaner/easier to read, this includes:

- New `B2Hash` typedef for blake 2 hashes (`std::array<std::byte, 20>`)
  + Also added `std::hash` specialisation to allow using it in maps or sets
- Stop using temporary files for downloads (uses RAM instead)
  + VC redistributables installation still uses a temp file if necessary, so the folder in `%TEMP%` will still be created/removed
- Moved some defines to global `constexpr` variables
- Moved some hardcoded values to global variables
- Removed a lot of utf8<->wide conversions that were unnecessary now that temporary files aren't used anymore
- Changed some for loops to use iterators instead of indices

### Motivation and Context

- Significant speedup when downloading lots of small delta patches
- Cleanup and simplification of existing code

### How Has This Been Tested?

Tested with https://github.com/obsproject/bouf/pull/30 on my test deployment.

### Types of changes

### Checklist:

### DIFF
--- a/UI/win-update/updater/hash.cpp
+++ b/UI/win-update/updater/hash.cpp
@@ -21,33 +21,33 @@
 
 using namespace std;
 
-void HashToString(const uint8_t *in, wchar_t *out)
+void HashToString(const B2Hash &in, string &out)
 {
-	const wchar_t alphabet[] = L"0123456789abcdef";
+	const char alphabet[] = "0123456789abcdef";
+	out.resize(kBlake2StrLength);
 
-	for (int i = 0; i != BLAKE2_HASH_LENGTH; ++i) {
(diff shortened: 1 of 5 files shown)


derrod:
I'm probably gonna remove the bundle feature (for now) since it doesn't provide *that* much of a speedup with our CDN anyway, but I'll keep a lot of the cleanup work to make code more readable.

PR #8823, opened 2023-04-30
obs-ffmpeg: Add FFmpeg Options for VA-API

nowrep:
### Description

Add FFmpeg Options for VA-API.

### Motivation and Context

Add option to set custom FFmpeg options for VA-API encoder, which is already available for other FFmpeg encoders.

### How Has This Been Tested?

Checked the options were applied correctly. ArchLinux.

### Types of changes

### Checklist:

### DIFF
--- a/plugins/obs-ffmpeg/obs-ffmpeg-vaapi.c
+++ b/plugins/obs-ffmpeg/obs-ffmpeg-vaapi.c
@@ -29,6 +29,7 @@
 #ifdef ENABLE_HEVC
 #include <obs-hevc.h>
 #endif
+#include <opts-parser.h>
 
 #include <unistd.h>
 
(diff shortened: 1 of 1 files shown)


kkartaltepe:
Personally, I would rather see PRs working on resolving acceleration device selection issues in the custom ffmpeg encoder than attempting to turn our supported encoder into the custom ffmpeg encoder. Options like this are precisely why the custom ffmpeg is unsupported, so I'm not sure I would agree with adding this,

nowrep:
> Personally, I would rather see PRs working on resolving acceleration device selection issues in the custom ffmpeg encoder than attempting to turn our supported encoder into the custom ffmpeg encoder

Custom ffmpeg encoder isn't available as stream encoder, so that wouldn't help.

> Options like this are precisely why the custom ffmpeg is unsupported, so I'm not sure I would agree with adding this,

There are already encoders (x264, AMF, ...) that allow to set more options than only what is available in GUI, so I don't see how VAAPI encoder would be a different case here.

kkartaltepe:
> Custom ffmpeg encoder isn't available as stream encoder, so that wouldn't help.

This is precisely the reason "Output to URL" was added, so this doesnt seem like blocker to me.

nowrep:
> This is precisely the reason "Output to URL" was added, so this doesnt seem like blocker to me.

Then you can't stream and record locally at the same time, so it doesn't seem to be usable alternative.

PR #8824, opened 2023-04-30
Set error message for NVENC, AMF, and QSV with P216/P416

### LINKED ISSUE #8665: P416 color format is not compatible with HDR

RytoEX:
### Description

If P216 or P416 color formats are selected with NVENC, AMF, or QSV, these color formats were not explicitly handled, so the switch statements would end up in the default case. If the user had also selected a Rec. 2100 color space, this would result in the strange error message:
> OBS does not support 8-bit output of Rec. 2100.

This message is confusing and does not correctly reflect the chosen settings. Let's explicitly handle the P216/P416 cases and provide a more accurate error message.

----

An alternative wording/approach such as with #8811 and #8812 could be taken instead, if preferred.

### Motivation and Context

Fixes #8665

### How Has This Been Tested?

Tested locally on Windows 10 with NVENC.

### Types of changes

### Checklist:


RytoEX:
Added a commit to also handle these cases in AMF. If we'd prefer to split that into a separate PR or change the implementation, that can be done.

RytoEX:
Added a commit to also handle these cases for QSV.

gxalpha:
> An alternative wording/approach such as with https://github.com/obsproject/obs-studio/pull/8811 and https://github.com/obsproject/obs-studio/pull/8812 could be taken instead, if preferred.

I have no particular preference on the wording here vs in #8812 but I think they should at least match, so if this wording here is preferred I'd go ahead and adjust #8812 to match.

RytoEX:
> > An alternative wording/approach such as with #8811 and #8812 could be taken instead, if preferred.
> 
> I have no particular preference on the wording here vs in #8812 but I think they should at least match, so if this wording here is preferred I'd go ahead and adjust #8812 to match.

For my part, this was the original code draft I'd had, so I went with it because it was quickest for me to toss into a PR yesterday. I also couldn't determine if "high-precision color formats" was a universally understood term or if we'd made it up, so I kept my original wording pending further discussion.

notr1ch (approved):

RytoEX:
Off-thread discussion indicates that this is good to merge as-is. We can adjust or align error messages later.

PR #8838, opened 2023-05-03
UI: Relax mc_trans_video_imagescaler.dll DLL block

notr1ch:
### Description
This is actually a MainConcept redistributable and not related to Adobe. Unfortunately Elgato Game Capture HD software relies on this dependency when presenting the DirectShow device to OBS, so we unintentionally blocked it from loading.

Instead of outright blocking, we now block only versions older than the one shipped by Elgato, which has hopefully been patched to fix the random crashes.

### Motivation and Context
Blocking legit capture devices isn't great.

### How Has This Been Tested?
Hasn't as I don't have the necessary hardware.

### Types of changes

### Checklist:

### DIFF
--- a/UI/win-dll-blocklist.c
+++ b/UI/win-dll-blocklist.c
@@ -124,8 +124,9 @@ static blocked_module_t blocked_modules[] = {
 	// Wacom / Other tablet driver, locks up UI
 	{L"\\wintab32.dll", 0, 0, TS_IGNORE},
 
-	// Adobe Dynamic Link (Adobe CC), crashes in its own thread
-	{L"\\mc_trans_video_imagescaler.dll", 0, 0, TS_IGNORE},
+	// MainConcept Image Scaler, crashes in its own thread. Block versions
+	// older than the one Elgato uses (2016-02-15).
+	{L"\\mc_trans_video_imagescaler.dll", 0, 1455495131, TS_LESS_THAN},
 
 	// Weird Polish banking "security" software, breaks UI
 	{L"\\wslbscr64.dll", 0, 0, TS_IGNORE},


RytoEX (approved):
Looks good to me.

UI/win-dll-blocklist.c
```diff
@@ -124,8 +124,8 @@ static blocked_module_t blocked_modules[] = {
 	// Wacom / Other tablet driver, locks up UI
 	{L"\\wintab32.dll", 0, 0, TS_IGNORE},
 
-	// Adobe Dynamic Link (Adobe CC), crashes in its own thread
-	{L"\\mc_trans_video_imagescaler.dll", 0, 0, TS_IGNORE},
+	// MainConcept Image Scaler, crashes in its own thread
+	{L"\\mc_trans_video_imagescaler.dll", 0, 1455495131, TS_LESS_THAN},
```
tt2468, line 128: Seems like it might be appropriate to #define the version number or at least add a comment as to how the number is calculated

WizardCM: It's a unix timestamp, and the next perameter (`TS_LESS_THAN`) determines the matching rule.

In this case, it's set to `UTC Monday, February 15, 2016` (ie. DLL older than Feb 2016 gets blocked)

notr1ch: This is documented in the struct:

https://github.com/obsproject/obs-studio/blob/63e4cdafad3d44a8a605c055b68ae2665401720c/UI/win-dll-blocklist.c#L71-L75

tt2468: Ah I suppose that's fine then

notr1ch: I added a comment anyway just for consistency with all the others that have a timestamp.

RytoEX:
This can't be worse than the current situation. Merging.

Issue #8843, opened 2023-05-03
Switching collections causes sources to change order and change names. Very weird behaviour


nacalai:
### Operating System Info

Windows 10

### OBS Studio Version

29.1.0

### OBS Studio Log URL

https://obsproject.com/logs/LMo6rn3pE80jleEt

### Expected Behavior

Switching scene collections should have no effect on the order and names of sources in the collection.

### Current Behavior

I have a scene collection with a lot of "Kamera 1", "Kamera 2", and so on, as seen here:

![image](https://user-images.githubusercontent.com/4357694/235877902-e6f16ce6-3fa1-4180-b282-c0f914b6bc78.png)

If I switch to for example an empty scene collection, and then back again, suddenly things are not the way they were. "Kamera 1" is now at the  bottom of the list, and is named "Kamera 1 2", as if it's been duplicated, but in reality it is not. Some sources are moved around and have a "2" added to them, some are not.

![image (1)](https://user-images.githubusercontent.com/4357694/235877946-7545e40d-9bed-4f62-94c5-584099178fe6.png)

Incredibly weird behaviour.

In the logs I can see stuff like this

`
11:22:12.908: Attempted to insert context with duplicate name "NDI Kam 4"! Name has been changed to "NDI Kam 4 2"

11:22:12.908: Attempted to insert context with duplicate UUID "2213c270-102a-4f0e-a26f-4df8e660fe9d"!
11:22:12.908: Attempted to insert context with duplicate UUID "72a0de21-1e01-4309-94df-512a56a97a00"!
11:22:12.908: A/V thread for 'NDI Kam 4 2' started
11:22:12.909: Attempted to insert context with duplicate UUID "6c0c3437-be0f-4c6f-a489-8f7358d8b756"!
11:22:12.909: Attempted to insert context with duplicate UUID "5f06c5d2-8304-4fb1-9c45-4280dec7fc6f"!
11:22:12.909: started A/V threads for source 'CAM-2 (Channel 1)'
11:22:12.909: Attempted to insert context with duplicate name "NDI Kam 3"! Name has been changed to "NDI Kam 3 2"
11:22:12.909: Attempted to insert context with duplicate UUID "945391de-36c4-4365-bb7b-768cafcb0a98"!
11:22:12.909: Attempted to insert context with duplicate UUID "4643243d-61fc-40a8-ab3f-3974a346476b"!
11:22:12.910: A/V thread for 'NDI Kam 3 2' started
[...]

PR #8846, opened 2023-05-03
UI: Exit if clearing scene data fails

derrod:
### Description

Faulty plugins or other bugs resulting in references to source outliving the intended lifetime can result in data corruption or loss as in #8843. Instead of corrupting data just crash as something has gone horribly wrong.

### Motivation and Context

Crashing > Data corruption.

### How Has This Been Tested?

Manually incremented ref counter of some sources. Couldn't reproduce #8843 directly.

### Types of changes

### Checklist:

### DIFF
--- a/UI/data/locale/en-US.ini
+++ b/UI/data/locale/en-US.ini
@@ -502,6 +502,10 @@ MacPermissions.Continue="Continue"
 UpdateAvailable="New Update Available"
 UpdateAvailable.Text="Version %1.%2.%3 is now available. <a href='%4'>Click here to download</a>"
 
+# Source leak error message
+SourceLeak.Title="Source Cleanup Error"
+SourceLeak.Text="There was a problem while changing scene collections and some sources could not be unloaded. This issue is typically caused by plugins that are not releasing resources properly. Please ensure that any plugins you are using are up to date.\n\nOBS Studio will now exit to prevent any potential data corruption."
+
 # audio device names
 Basic.DesktopDevice1="Desktop Audio"
 Basic.DesktopDevice2="Desktop Audio 2"
--- a/UI/window-basic-main.cpp
+++ b/UI/window-basic-main.cpp
@@ -1065,8 +1065,17 @@ static inline void AddMissingFiles(void *data, obs_source_t *source)
 void OBSBasic::LoadData(obs_data_t *data, const char *file)
 {
 	ClearSceneData();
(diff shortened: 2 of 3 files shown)


notr1ch:
Note that this will also trigger during shutdown for any leaking plugins. Given how many 3rd party plugins seem to cause crashes on shutdown, we may want to change the messaging to make references to plugins causing the issue.

derrod:
> Note that this will also trigger during shutdown for any leaking plugins. Given how many 3rd party plugins seem to cause crashes on shutdown, we may want to change the messaging to make references to plugins causing the issue.

Originally I was thinking about showing a Qt Error dialog that says something like:
`OBS detected orphaned sources when trying to clear scene data, this is likely caused by a third-party plugin not releasing sources correctly. Once clicking "OK" OBS will exit to avoid corrupting your scene collection.`

Maybe this exit here can be more graceful than relying on `bcrash()`, I'm not sure how it behaves on Linux/macOS.

notr1ch:
As long as all user data has been saved (and we've given plugins a chance to clean up) perhaps we could do a clean `exit()` or similar to avoid triggering an actual crash. There might be some log files that don't get closed cleanly but having OBS crash is probably a worse user experience.

derrod:
Updated to show this warning instead and then gracefully (more or less) exit:
2023-05-05_09-52-44_YGXTjN

Saving is kept disabled to avoid overwriting the scene collection with garbage data, but we allow for a clean-ish shutdown, though it does result in OBS reporting several hundred memory leaks in my log. Not sure what exactly those are but it probably doesn't matter once we're in a bad state anyway.

UI/data/locale/en-US.ini
```diff
@@ -501,6 +501,10 @@ MacPermissions.Continue="Continue"
 UpdateAvailable="New Update Available"
 UpdateAvailable.Text="Version %1.%2.%3 is now available. <a href='%4'>Click here to download</a>"
 
+# Source leak error message
+SourceLeak.Title="Memory Leak Detected"
+SourceLeak.Text="OBS has detected a memory leak. This is likely due to faulty a third-party plugin not working correctly. To avoid data corruption, OBS will now exit."
```
Vainock, line 506: I'm not native but I think you switched the "faulty" and "a".

derrod:
Updated wording based on off-thread discussions. Also updated comments to reflect what the current version of this PR actually does.

PR #8867, opened 2023-05-08
libobs: Add optional audio/video support to outputs via new flags

tt2468:
### Description
Adds the following new output flags:
- `OBS_OUTPUT_OPTIONAL_VIDEO`
- `OBS_OUTPUT_OPTIONAL_AUDIO`
- `OBS_OUTPUT_OPTIONAL_AV`

### Motivation and Context
Some outputs (like SRT/MPEG-TS and WHIP) can theoretically support audio-only or video-only operation if made to do so. Allowing this to happen requires libobs to be aware of when an output that supports video but has no video encoder is okay.

So basically, audio-only or video-only streams are sometimes nice to have for various reasons, so it's nice to have that support natively without having to do workarounds like defining multiple output infos.

### How Has This Been Tested?
- Tested normal streaming, recording - No behavioral change
- Tested video-only and audio-only streaming using a fork of the WHIP output PR - Worked great with Cloudflare Stream

### Types of changes

### Checklist:


libobs/obs-output.c
```diff
@@ -2045,8 +2048,24 @@ static inline void convert_flags(const struct obs_output *output,
 	else
 		flags &= output->info.flags;
 
-	*has_video = (flags & OBS_OUTPUT_VIDEO) != 0;
-	*has_audio = (flags & OBS_OUTPUT_AUDIO) != 0;
+	if ((flags & OBS_OUTPUT_VIDEO) != 0) {
+		if ((flags & OBS_OUTPUT_OPTIONAL_VIDEO) != 0)
+			*has_video = output->video_encoder != NULL;
```
tytan652, line 2053: This make the assumption that `OBS_OUTPUT_ENCODED` is set which is not a good thing.

tytan652 (changes requested):

notr1ch:
Why is the OBS_OUTPUT_OPTIONAL_AV flag standalone? Wouldn't OBS_OUTPUT_OPTIONAL_AUDIO | OBS_OUTPUT_OPTIONAL_VIDEO mean the same thing?

tytan652:
> Why is the OBS_OUTPUT_OPTIONAL_AV flag standalone? Wouldn't OBS_OUTPUT_OPTIONAL_AUDIO | OBS_OUTPUT_OPTIONAL_VIDEO mean the same thing?

Maybe because `OBS_OUTPUT_AV` exists.

tt2468:
Welp, went to try to fix the issues with raw outputs and realized that there's a lot more flaws to the implementation than originally intended. Found myself too deep into what would be more of a refactor of obs-output.c than I'd like to with this PR. For now, I'm going to just do appropriate workarounds (registering audio-only and video-only output types) on my end and I'll maybe tackle this again if I get around to doing a dedicated refactor.

PR #8873, opened 2023-05-09
libobs: Add sanity checks to some obs_output_t functions

tt2468:
### Description
Adds sanity checks for cases where we shouldn't let certain set functions be performed, when the output doesn't support the associated features. This also does a mild amount of code cleanup, adding inline functions for checking the flags mentioned below.

### Motivation and Context
Since OBS outputs can have any combination of flags for encoded, video, audio, and service, there are a number of cases where it would be a good idea to validate that we're not allowing output changes that risk undefined behavior given the supported flags.

### How Has This Been Tested?
Ubuntu 20.04
- Streaming
- Recording
- Virtual cam

### Types of changes

### Checklist:

### DIFF
--- a/libobs/obs-output-delay.c
+++ b/libobs/obs-output-delay.c
@@ -28,6 +28,22 @@ static inline bool delay_capturing(const struct obs_output *output)
 	return os_atomic_load_bool(&output->delay_capturing);
 }
 
+static inline bool flag_encoded(const struct obs_output *output)
+{
+	return (output->info.flags & OBS_OUTPUT_ENCODED) != 0;
+}
+
+static inline bool log_flag_encoded(const struct obs_output *output,
+				    const char *func_name, bool inverse_log)
+{
+	const char *prefix = inverse_log ? "n encoded" : " raw";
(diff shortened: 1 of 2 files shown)


notr1ch:
I'm not a big fan of APIs failing silently as it can hide further bugs in the calling application. I think we should either log or assert for these cases where it's clearly undefined behavior.

tt2468:
The reason I did not add logging for these checks is because of simple code complexity reasons- the documentation is pretty good at stating which functions are meant for raw outputs and which are meant for encoded. A number of the guarded functions largely do nothing if the output capabilities are not there (deeper obs_output_t logic seems to do its own checks in most cases already). I'm just trying to move these checks further "up the chain" so that they can be removed from other less-obvious places later on.

notr1ch:
I unfortunately didn't get to see the new version before the merge, but I think the `log_` prefixes are quite confusing as the function name suggests it's just a logging function and it's kind of unclear how it should be used. There's also some code duplication with the logging functions across files. Something to keep in mind if we ever want to refactor this.

PR #8874, opened 2023-05-09
libobs, plugins: Deprecate obs_output_t functions with flag parameters

tt2468:
### Description
Depends #8873

This deprecates the following functions, replacing them with new versions:
- `obs_output_can_begin_data_capture()` - now `*capture2()`
- `obs_output_initialize_encoders()` - now `*encoders2()`
- `obs_output_begin_data_capture()` - now `*capture2()`

An implementation to allow audio/video to be optional is best done using the flag technique described below, with audio/video enablement specified by whether media (raw, `video_t/audio_t`) or encoder (`obs_encoder_t`) objects are specified.

Since every implementation I could find always specifies `flags` as 0, I was able to safely conclude that immediately removing the parameters' functionality is safe to do.

### Motivation and Context
The flags parameter was reportedly initially designed to support audio-only or video-only operation of an output which had the `OBS_OUTPUT_AV` flag, however, full support for that was never implemented, and there are likely fundamental issues with finishing the implementation, mainly that most outputs are programmed assuming that there will always be at least one audio and one video track. This requires new flags specifying support for optional audio/video, among other things.

To protect libobs from weird, unpredictable/undefined behavior, deprecating these parameters seems appropriate.

### How Has This Been Tested?
Ubuntu 20.04
- Streaming
- Recording
- Virtualcam

### Types of changes

(breaking change is subjective here. ABI has not directly changed but it signifies a time when ABI will eventually be broken by its removal)

### Checklist:

### DIFF
--- a/docs/sphinx/reference-outputs.rst
+++ b/docs/sphinx/reference-outputs.rst
@@ -882,44 +882,35 @@ Functions used by outputs
 
 ---------------------
 
-.. function:: bool obs_output_can_begin_data_capture(const obs_output_t *output, uint32_t flags)
+.. function:: bool obs_output_can_begin_data_capture2(const obs_output_t *output)
 
    Determines whether video/audio capture (encoded or raw) is able to
-   start.  Call this before initializing any output data to ensure that
+   start.  Call this before initializing any output state to ensure that
    the output can start.
 
-   :param flags: Set to 0 to initialize both audio/video, otherwise a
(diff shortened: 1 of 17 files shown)


notr1ch:
I'm not really sure this is a code cleanup when we have to keep the previous functions to avoid an API break. Documenting that the `flags` parameter is reserved and must be zero would have been the better option IMO.

PR #8878, opened 2023-05-09
cmake: Update macOS build system to fix issues introduced with last upgrade

PatTheMav:
### Description
Updates the new CMake 3.0.0 build system for macOS to simplify code and also fix newly introduced issues:

* Fix missing OBS icon for generated `dmg` distribution disk image
* Simplify dependency handling by using universal builds of `obs-deps` to build the binary
* Simplify CMake presets for macOS
* Remove Ninja-based builds and use Xcode as only build generator supported for fully distributable builds of `obs-studio`
* Add include guards for the CMake bootstrap to avoid files being included multiple times
* Fix policies not being applied to the global scope when set in included file
* Clean up compiler flag definition by introducing a cross-platform set of `clang` compiler flags which are used on Windows, macOS, and Linux alike
* Properly document some Xcode-specific CMake settings that are either being ignored or lead to undefined results at the moment
* Add support for `CMAKE_COMPILE_WARNING_AS_ERROR` variable (introduced in CMake 3.24) to enable the desired functionality (enabled by default)
* Remove `Info.plist` template files and use Xcode's built-in file generation instead
* Remove legacy macOS CMake support files
* Remove legacy macOS CMake script files
* Update Framework link definition to use `$<LINK_LIBRARY:Framework,SomeFramework.framework>` generator expression
* Add helper function for Xcode-specific properties to make associated code easier to maintain

### Motivation and Context
Ensures good practices in our CMake scripts and also reduces complexity by having Xcode take care of most macOS-specific build steps.

### How Has This Been Tested?
Tested on macOS, Windows, and Linux by running builds via the build scripts.

### Types of changes

### Checklist:


cmake/common/buildspec_common.cmake
```diff
@@ -0,0 +1,172 @@
+# Common build dependencies module
```
RytoEX, line 1: Missing "OBS" in the comment here?

plugins/mac-virtualcam/src/camera-extension/CMakeLists.txt
```diff
@@ -15,36 +15,42 @@ endforeach()
...
+  PROPERTIES SWIFT_VERSION 5.0
+             MACOSX_DEPLOYMENT_TARGET 13.0
+             CODE_SIGN_ENTITLEMENTS "${CMAKE_CURRENT_SOURCE_DIR}/cmake/macos/entitlements.plist"
+             PRODUCT_NAME com.obsproject.obs-studio.mac-camera-extension
+             PRODUCT_BUNDLE_IDENTIFIER com.obsproject.obs-studio.mac-camera-extension
+             CURRENT_PROJECT_VERSION ${OBS_BUILD_NUMBER}
+             MARKETING_VERSION ${OBS_VERSION_CANONICAL}
+             COPY_PHASE_STRIP NO
+             GENERATE_INFOPLIST_FILE YES
+             INFOPLIST_KEY_NSHumanReadableCopyright "(c) 2012-${CURRENT_YEAR} Sebastian Beckmann, Patrick Heyer"
```
RytoEX, line 54: Should say 2022?

plugins/mac-virtualcam/src/dal-plugin/CMakeLists.txt
```diff
@@ -32,38 +23,45 @@ target_sources(
           Defines.h
           Logging.h)
 
-if(NOT XCODE)
-  target_compile_options(obs-dal-plugin PRIVATE -fobjc-arc -fobjc-weak)
-endif()
+target_compile_options(obs-dal-plugin PRIVATE -fobjc-arc -fobjc-weak)
```
RytoEX, line 26: Does Xcode now require this as well? This looks backwards to the other changes to remove Ninja CMake paths.

RytoEX (changes requested):
Did a visual pass of the changes. I also compiled this locally and it compiled and ran fine on my M2 system. I didn't do any in-depth runtime testing beyond that.

cmake/common/buildspec_common.cmake
```diff
@@ -0,0 +1,172 @@
...
+
+# _check_deps_version: Checks for obs-deps VERSION file in prefix paths
+function(_check_deps_version version)
+  # cmake-format: off
+  set(found FALSE PARENT_SCOPE)
+  # cmake-format: on
+
+  foreach(path IN LISTS CMAKE_PREFIX_PATH)
+    if(EXISTS "${path}/share/obs-deps/VERSION")
+      if(dependency STREQUAL qt6 AND NOT EXISTS "${path}/lib/cmake/Qt6/Qt6Config.cmake")
```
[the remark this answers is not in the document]
PatTheMav: Quoting is necessary when compositing strings with user-provided input and the possibility of space-separated tokens contained in the string.

Paths are usually considered "user-provided" as users can have paths with spaces on their local machines.

CMakePresets.json
```diff
@@ -8,93 +8,37 @@
...
-        "ENABLE_SPARKLE": true,
-        "OBS_SPARKLE_KEY": {"type": "STRING", "value": ""},
-        "OBS_CODESIGN_IDENTITY": "$penv{CODESIGN_IDENT}",
-        "OBS_CODESIGN_TEAM": "$penv{CODESIGN_TEAM}",
-        "SPARKLE_APPCAST_URL": {"type": "STRING", "value": "https://obsproject.com/osx_update/updates_arm64_v2.xml"},
-        "SPARKLE_PUBLIC_KEY": {"type": "STRING", "value": "HQ5/Ba9VHOuEWaM0jtVjZzgHKFJX9YTl+HNVpgNF0iM="},
+        "CMAKE_OSX_DEPLOYMENT_TARGET": {"type": "STRING", "value": "11.0"},
+        "OBS_CMAKE_VERSION": {"type": "STRING", "value": "3.0.0"},
+        "OBS_CODESIGN_TEAM": {"type": "STRING", "value": "$penv{CODESIGN_TEAM}"},
+        "OBS_CODESIGN_IDENTITY": {"type": "STRING", "value": "$penv{CODESIGN_IDENT}"},
```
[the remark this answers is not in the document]
PatTheMav: The order is on purpose - `TEAM` should always be specified and takes precedence over the `IDENTITY` - the latter is just a fallback if no team is provided (or for special environments like CI where only specific identities are available in the keychain).

cmake/common/ccache.cmake
```diff
@@ -1,4 +1,6 @@
-# OBS CMake ccache module
+# CMake ccache module
```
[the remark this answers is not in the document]
PatTheMav: Copy and Paste error as the plugintemplate files usually form the basis for this update.

plugins/mac-virtualcam/src/dal-plugin/CMakeLists.txt
PatTheMav: Nope, that looks more like an oversight.

CMakePresets.json
```diff
@@ -8,93 +8,37 @@
...
-        "ENABLE_SPARKLE": true,
-        "OBS_SPARKLE_KEY": {"type": "STRING", "value": ""},
-        "OBS_CODESIGN_IDENTITY": "$penv{CODESIGN_IDENT}",
-        "OBS_CODESIGN_TEAM": "$penv{CODESIGN_TEAM}",
-        "SPARKLE_APPCAST_URL": {"type": "STRING", "value": "https://obsproject.com/osx_update/updates_x86_64_v2.xml"},
-        "SPARKLE_PUBLIC_KEY": {"type": "STRING", "value": "HQ5/Ba9VHOuEWaM0jtVjZzgHKFJX9YTl+HNVpgNF0iM="},
-        "VIRTUALCAM_DEVICE_UUID": {"type": "STRING", "value": "7626645E-4425-469E-9D8B-97E0FA59AC75"},
-        "VIRTUALCAM_SOURCE_UUID": {"type": "STRING", "value": "A8D7B8AA-65AD-4D21-9C42-66480DBFA8E1"},
-        "VIRTUALCAM_SINK_UUID": {"type": "STRING", "value": "A3F16177-7044-4DD8-B900-72E2419F7A9A"}
+        "CMAKE_COMPILE_WARNING_AS_ERROR": true
```
gxalpha, line 41: Since this is enabled by default this seems redundant. (Or is the idea to only have it enabled on CI? In that case it'd probably need to be set to off either in the `macos` preset or by default)

PatTheMav: It is only enabled by default if the cache value is not set at all (because the CMake code does not set `-Werror` anymore and relies entirely on the cache variable).

Thus it is more correct to explicitly enable this in the preset (as the default value used by CMake is `FALSE`). I've actually came around to the opinion that enforcing this actually violates good practises, hence why enforce it with a bit of a hack.

CMakePresets.json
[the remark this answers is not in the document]
RytoEX: Got it.

cmake/common/buildspec_common.cmake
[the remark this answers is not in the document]
RytoEX: Understood. Thanks.

RytoEX (changes requested):
Nit.

RytoEX (approved):
Looks good to me. Will let CI finish up.

PR #8881, opened 2023-05-09
CI: Update GitHub Actions workflows with repository actions and updated build scripts

PatTheMav:
### Description
Updates all GitHub Actions workflows to use repository actions and also utilise reusable workflows with updated build scripts:

* Updates macOS and Linux build scripts, based on ZSH, a safer, faster and more advanced shell than Bash
* Updates code-formatting script, based on ZSH and added support for `swift-format`
* Updates `gitignore` file to be opt-in (as in, everything is ignored apart from specific directory patterns)
* Removes legacy macOS and Linux build scripts
* Adds repository actions for
   * Running `clang-format`
   * Running `cmake-format`
   * Running `swift-format`
   * Building `obs-studio`
   * Packaging `obs-studio`
   * Setting up codesigning credentials
   * Validating Qt XML files
   * Steam Upload
   * Generate Sparkle Appcast
   * Compatibility file validation
   * Services file validation
* Adds reusable workflows (workflows that are triggered by other workflows) for
   * Building the project
   * Checking the formatting
   * Generating documentation
   * Services validation
* Add separate workflows for specific repository events:
   * Pull Requests, runs
       * Format check
       * Build project
       * Compatibility file validation
       * Services file validation
       * Update documentation
   * Push, runs
       * Format check
       * Build project
       * Compatibility file validation
       * Services file validation
       * Upload language files
       * Update documentation
   * Push with tag, runs
       * Format check
       * Build project
       * Compatibility file validation
       * Services file validation (schema only)
       * Upload language files
       * Update documentation (CloudFlare variant)
       * Deploy documentation to CloudFlare
       * Create Sparkle appcast
       * Create Release
   * Scheduled, runs
       * Services validation (contents)
       * Build project
       * Nightly steam upload
   * Dispatch, can run
       * Services file validation (schema and contents)
       * Download language files
       * Steam upload
       * Update documentation
       * Update documentation (CloudFlare variant)
       * Deploy documentation to CloudFlare

### Motivation and Context
The basic idea is to make workflows actually be composed of distinct actions or sub-workflows that should run based on specific events that happen in the repository:

* Checking the code formatting and building the project are common steps taken by most repository actions, but themselves are made up of _parallel_ steps - thus they are implemented as reusable workflows
* Building/packaging the project, handling documentation, uploading localisation files, and more, are distinct use cases whose functionality should be contained in an action and are implemented as such

This makes code (mostly) self-contained to the area of its own concern with limited "leakage" to consumers (mostly influenced by action inputs and outputs) and easier to maintain (albeit at the cost of having to make the code act more isolated).

### How Has This Been Tested?
Needs to be tested on CI with available secrets and maybe needs a temporary commit that enables actions taking place only when targeting the master branch and/or a release tag to happen on a separate test branch.

### Types of changes

### Checklist:


derrod:
Purely subjective, but curious what others think; Personally I prefer having empty lines between workflow steps and sections for readability. Otherwise it just get too dense of a wall of text.

PatTheMav:
> Purely subjective, but curious what others think; Personally I prefer having empty lines between workflow steps and sections for readability. Otherwise it just get too dense of a wall of text.

I learned to prefer it this way, also as most actions/workflows fit on a single page now, but I can see the benefit of spacing out the workflow jobs at the very least - will push an update for that tomorrow-ish.

.github/actions/build-obs/action.yaml
```diff
@@ -0,0 +1,109 @@
...
+        .github/scripts/build-macos ${build_args}
+
+    - name: Install Dependencies 🛍️
+      if: runner.os == 'Linux'
+      shell: bash
+      run: |
+        : Install Dependencies 🛍️
+        echo ::group::Install Dependencies
+        eval "$(/home/linuxbrew/.linuxbrew/bin/brew shellenv)"
+        brew install --quiet zsh
```
pheiduck, line 58: Maybe its better to install `zsh` via `apt`
```
apt-get install -y zsh > /dev/null
```

PatTheMav: This is on purpose - we manage pinned `clang-format` versions via Brew already and `apt` packages are usually severely out of date (especially on Ubuntu 22.04, which is still the most recent Linux runner version available on GitHub Actions).

On local machines, you can use any package manager of your choice to install some dependencies, but on CI it's easier for us to use Linuxbrew (which is also distribution-independent).

pheiduck: Sounds reasonable, but the Linux build CI currently has an error:
`Error: zsh: command not found`
I think it is because it was not linked to the right place.

PatTheMav: Yeah the branch is in a non-working state - I’m currently testing on my own fork because of all the different event types and moving parts. There are far more issues in its current state many of which are fixed already.

pheiduck:
```
sudo update-alternatives --install /bin/zsh zsh /home/linuxbrew/.linuxbrew/bin/zsh 100
sudo update-alternatives --set zsh /home/linuxbrew/.linuxbrew/bin/zsh
```

PatTheMav: That's not necessary - you just have to add the paths to the `GITHUB_PATH` so that it's persisted for the current runner (because every step in a workflow has its own state and `export`s have no effect).

.github/workflows/publish.yaml
```diff
@@ -0,0 +1,152 @@
...
+        run: |
+          : Set Up Environment 🔧
+          if [[ "${RUNNER_DEBUG}" ]]; then set -x; fi
+
+          git config --global --add safe.directory "${GITHUB_WORKSPACE}"
+
+          dnf install -y -q gh
+          gh extension install actions/gh-actions-cache
+
+          cache_key='flatpak-builder-${{ hashFiles('./build-aux/com.obsproject.Studio.json') }}-x86_64'
```
tytan652, line 79: The hash need to cover all manifest JSON files, feel free to use another argument for `hashFiles` it just needs to cover all the files.

```suggestion
          cache_key='flatpak-builder-${{ hashFiles('./build-aux/**/*.json') }}-x86_64'
```

.github/workflows/build-project.yaml
```diff
@@ -0,0 +1,287 @@
...
+          if [[ "${RUNNER_DEBUG}" ]]; then set -x; fi
+
+          git config --global --add safe.directory "${GITHUB_WORKSPACE}"
+
+          echo '::group::Install GitHub CLI tool'
+          dnf install -y -q gh
+          gh extension install actions/gh-actions-cache
+          echo '::endgroup::'
+
+          cache_key='flatpak-builder-${{ hashFiles('./build-aux/com.obsproject.Studio.json') }}-x86_64'
```
tytan652, line 235: The hash need to cover all manifest JSON files, feel free to use another argument for `hashFiles` it just needs to cover all the files.

```suggestion
          cache_key='flatpak-builder-${{ hashFiles('./build-aux/**/*.json') }}-x86_64'
```

tytan652 (changes requested):
Just quick review about #8765 changes.

I don't know if I just just missed it but I didn't find a workflow that calls `format-manifest.py`

TianQiBuTian:
QQ截图20230518180047

.github/scripts/.build.zsh
```diff
@@ -0,0 +1,347 @@
...
+  local -r _usage="
+Usage: %B${functrace[1]%:*}%b <option> [<options>]
+
+%BOptions%b:
+
+%F{yellow} Build configuration options%f
+ -----------------------------------------------------------------------------
+  %B-t | --target%b                     Specify target - default: %B%F{green}${host_os}-${CPUTYPE}%f%b
+  %B-c | --config%b                     Build configuration
+  %B--print-config%b                    Print composed cmake configuration parameters
```
RytoEX, line 100:
```suggestion
  %B--print-config%b                    Print composed CMake configuration parameters
```

.github/scripts/.build.zsh
```diff
@@ -0,0 +1,347 @@
...
+    }
+
+    if (( ${+YOUTUBE_CLIENTID} + ${+YOUTUBE_CLIENTID_HASH} + ${+YOUTUBE_SECRET} + ${+YOUTUBE_SECRET_HASH} > 3 )) {
+      cmake_args+=(
+        -DYOUTUBE_CLIENTID:STRING="${YOUTUBE_CLIENTID}"
+        -DYOUTUBE_CLIENTID_HASH:STRING="${YOUTUBE_CLIENTID_HASH}"
+        -DYOUTUBE_SECRET:STRING="${YOUTUBE_SECRET}"
+        -DYOUTUBE_SECRET_HASH:STRING="${YOUTUBE_SECRET_HASH}"
+      )
+    }
```
RytoEX, line 233: Does this need to be handled here, or can it be handled in a CMake preset?

.github/scripts/.build.zsh
```diff
@@ -0,0 +1,347 @@
...
+
+    case ${target} {
+      macos-*)
+        cmake_args+=(
+          --preset "macos${CI:+-ci}"
+          -DCMAKE_OSX_ARCHITECTURES=${target##*-}
```
RytoEX, line 239: `:TYPE=`?

.github/scripts/.build.zsh
```diff
@@ -0,0 +1,347 @@
...
+          autoload -Uz read_codesign_team && read_codesign_team
+
+          if [[ -z ${CODESIGN_TEAM} ]] {
+            autoload -Uz read_codesign && read_codesign
+          }
+        }
+
+        cmake_args+=(
+          -DOBS_CODESIGN_TEAM=${CODESIGN_TEAM:-}
+          -DOBS_CODESIGN_IDENTITY=${CODESIGN_IDENT:--}
```
RytoEX, line 254: `:TYPE=`?

.github/scripts/.build.zsh
```diff
@@ -0,0 +1,347 @@
...
+
+        if [[ ${generator} == 'Unix Makefiles' ]] {
+          cmake_build_args+=(--parallel $(( $(nproc) + 1 )))
+        } else {
+          cmake_build_args+=(--parallel)
+        }
+
+        cmake_args+=(
+          -DENABLE_AJA:BOOL=OFF
+          -DENABLE_WEBRTC=OFF
```
RytoEX, line 278: `:TYPE=`?

.github/scripts/utils.zsh/read_codesign_pass
```diff
@@ -0,0 +1,24 @@
+autoload -Uz read_codesign read_codesign_user log_info
```
RytoEX, line 1:
```suggestion
autoload -Uz read_codesign read_codesign_user log_info log_warning
```

RytoEX (changes requested):
Submitting a partial review. I'll finish up the rest next week.

I'm not sure if `name` fields or the run script labels/headers in the GitHub Actions are meant to be Title Case or not. If they are, Title Case all of them. If they aren't, then I suppose use Sentence case.

.gitignore
```diff
@@ -1,108 +1,49 @@
...
+!AUTHORS
+!buildspec.json
+!CMakeLists.txt
+!CMakePresets.json
+!COC.rst
+!COMMITMENT
+!CONTRIBUTING.rst
+!COPYING
+!INSTALL
+!README.rst
```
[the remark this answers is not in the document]
PatTheMav: Correct - the file is now opt-in instead of opt-out, which has the benefit of ignoring all kinds of user-generated directories.

.github/workflows/scheduled.yaml
```diff
@@ -0,0 +1,121 @@
+name: Scheduled
+run-name: Scheduled Repository Actions ⏰
+on:
+  schedule:
+    - cron: 0 0 * * *
```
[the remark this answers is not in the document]
PatTheMav: Yep, good idea.

.github/actions/build-obs/action.yaml
```diff
@@ -0,0 +1,118 @@
...
+    required: false
+    default: RelWithDebInfo
+  codesign:
+    description: Enable codesigning (macOS only)
+    required: false
+    default: 'false'
+  codesignIdent:
+    description: Developer ID for application codesigning (macOS only)
+    required: false
+    default: '-'
```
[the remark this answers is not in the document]
PatTheMav: Yes, a single dash means "ad-hoc" codesigning, so that's the correct fallback value.

.github/actions/check-changes/action.yaml
```diff
@@ -0,0 +1,57 @@
...
+    description: Glob expression to limit check to specific files
+    required: false
+  useFallback:
+    description: Use fallback compare against prior commit
+    required: false
+    default: 'true'
+outputs:
+  passed:
+    value: ${{ steps.checks.outputs.passed }}
+    description: True if specified files where changed in comparison to specified git reference
```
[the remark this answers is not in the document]
PatTheMav: The job checks if changes to the specified files exist. As such the check "passes" if changes are found. The explanation is handled by the `description`.

.github/actions/generate-docs/action.yaml
```diff
@@ -0,0 +1,62 @@
...
+        sed -i -E \
+          -e "s/version = '([0-9]+\.[0-9]+\.[0-9]+)'/version = '${major}.${minor}.${patch}'/g" \
+          -e "s/release = '([0-9]+\.[0-9]+\.[0-9]+)'/release = '${major}.${minor}.${patch}'/g" \
+          -e "s/copyright = '(2017-[0-9]+, Lain Bailey)'/copyright = '2017-$(date +"%Y"), Lain Bailey'/g" \
+          ${{ inputs.sourceDirectory }}/docs/sphinx/conf.py
+
+        if [[ '${{ inputs.disableLinkExtensions }}' == 'true' ]]; then
+          sed -i -e "s/html_link_suffix = None/html_link_suffix = ''/g" \
+           ${{ inputs.sourceDirectory }}/docs/sphinx/conf.py
+          echo "nameExtension=(No Extensions) " >> $GITHUB_OUTPUT
```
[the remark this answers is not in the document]
PatTheMav: That trailing space is by design, because it is composed in a compound statement in the upload artefact step:

`${{ steps.setup.outputs.nameExtension }}${{ steps.setup.outputs.commitHash }}`

This ensures that there is no double space if the output value is empty.

.github/actions/setup-macos-codesigning/action.yaml
```diff
@@ -0,0 +1,146 @@
+name: Set up macOS codesigning
+description: Sets up code signing certificates, provisioning profiles, and notarization information
```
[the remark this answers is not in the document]
PatTheMav: I just checked and "Code Signing" is the official term used by Apple, "codesign" the name of the program. So I guess I should ensure it's the former at least in normal sentences.

.github/scripts/.build.zsh
PatTheMav: We'd need to generate a preset on the fly by the workflow or action (either using the workflow secrets as input, or an entire preset stored as a base64-encoded string as a secret), but then also adjust the scripts to check for such a preset and use it.

The way the CMake script is set up, the variables should either be set or unset, but using a preset would always set them (albeit with empty values if the secrets are not configured), which would then lead to configuration errors.

.github/actions/check-changes/action.yaml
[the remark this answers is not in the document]
RytoEX: Use of the word "passed" Is strange because `passed=true` if files have changed. But if files have changed, we actually consider this a failed state, no?

.github/actions/generate-docs/action.yaml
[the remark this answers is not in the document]
RytoEX: Could you link me to the exact place(s) this comes up? This seems like an odd way to handle it.

.github/scripts/.build.zsh
RytoEX: I don't see how empty strings would break anything. Our CMake explicitly checks for a truthy value (empty strings are FALSE) _and_ that the _HASH variables are set to `(0|[a-fA-F0-9]+)`.

PatTheMav: IMO it's not worth the complexity of creating a user preset on the fly or editing a user preset template file - setting specific cache variables is a valid approach even when using presets.

.github/actions/check-changes/action.yaml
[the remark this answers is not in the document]
PatTheMav: The check is self-contained, so it makes no assumption about what file changes might mean.

That is a concern of the workflow that _calls_ the action, so from the action that checks for changed files, detected changes are a _success_ (because otherwise the check would need to be called "check-clean" or something).

.github/actions/generate-docs/action.yaml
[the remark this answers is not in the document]
PatTheMav: Line 59.

.github/workflows/push.yaml
```diff
@@ -0,0 +1,313 @@
...
+        with:
+          sparklePrivateKey: ${{ secrets.SPARKLE_PRIVATE_KEY }}
+          baseImage: ${{ github.workspace }}/obs-studio-*-macos-${{ steps.setup.outputs.cpuName }}.dmg
+          channel: ${{ steps.setup.outputs.channel }}
+          count: 1
+          urlPrefix: 'https://cdn-fastly.obsproject.com/downloads'
+          customTitle: 'OBS Studio'
+          customLink: 'https://obsproject.com/'
+
+      - name: Upload artifacts 📡
```
RytoEX, line 216:
```suggestion
      - name: Upload Artifacts 📡
```

.github/workflows/push.yaml
```diff
@@ -0,0 +1,313 @@
...
+              ;;
+            +([0-9]).+([0-9]).+([0-9])-@(beta|rc)*([0-9]) )
+              echo 'validTag=true' >> $GITHUB_OUTPUT
+              echo 'prerelease=true' >> $GITHUB_OUTPUT
+              echo "version=${GITHUB_REF_NAME}" >> $GITHUB_OUTPUT
+              ;;
+            *) echo 'validTag=false' >> $GITHUB_OUTPUT ;;
+          esac
+
+      - name: Download build artifacts 📥
```
RytoEX, line 252:
```suggestion
      - name: Download Build Artifacts 📥
```

.github/workflows/scheduled.yaml
```diff
@@ -0,0 +1,121 @@
...
+    name: Check Service Availability 🛜
+    if: github.repository_owner == 'obsproject'
+    runs-on: ubuntu-22.04
+    permissions:
+      checks: write
+    steps:
+      - uses: actions/checkout@v3
+        with:
+          fetch-depth: 0
+      - name: Set up Homebrew 🍺
```
RytoEX, line 22:
```suggestion
      - name: Set Up Homebrew 🍺
```

.github/workflows/scheduled.yaml
```diff
@@ -0,0 +1,121 @@
...
+        uses: Homebrew/actions/setup-homebrew@master
+      - name: Check for defunct services 📉
```
RytoEX, line 24:
```suggestion
      - name: Check for Defunct Services 📉
```

build-aux/format-manifest.py
```diff
@@ -1,40 +1,90 @@
-import json
+import argparse
+import glob
+import logging
 import os
+import json
```
RytoEX, line 5: Place `json` before `logging`.

.github/workflows/dispatch.yaml
```diff
@@ -0,0 +1,146 @@
...
+
+      - name: Set Up Redirects 🔄
+        run: |
+          : Set Up Redirects 🔄
+          if [[ "${RUNNER_DEBUG}" ]]; then set -x; fi
+
+          echo "/previous/27.2 https://obsproject.com/docs/27.2 302" >> docs/_redirects
+          echo "/previous/:major.:minor https://:major-:minor.${{ vars.CF_PAGES_PROJECT }}.pages.dev 302" >> docs/_redirects
+
+      - name: Publish To Live Page
```
RytoEX, line 140:
```suggestion
      - name: Publish to Live Page
```

.github/workflows/dispatch.yaml
```diff
@@ -0,0 +1,146 @@
...
+        run: |
+          : Get Commit Hash 🆔
+          echo "commitHash=${GITHUB_SHA:0:9}" >> $GITHUB_OUTPUT
+      - uses: ./.github/actions/generate-docs
+        with:
+          commitHash: ${{ steps.checks.setup.commitHash }}
+          disableLinkExtensions: true
+
+  deploy-documentation:
+    name: Deploy Documentation To CloudFlare ☁️
```
RytoEX, line 113:
```suggestion
    name: Deploy Documentation to CloudFlare ☁️
```

.github/workflows/dispatch.yaml
```diff
@@ -0,0 +1,146 @@
...
+        id: setup
+        run: |
+          : Get Commit Hash 🆔
+          echo "commitHash=${GITHUB_SHA:0:9}" >> $GITHUB_OUTPUT
+      - uses: ./.github/actions/generate-docs
+        with:
+          commitHash: ${{ steps.checks.setup.commitHash }}
+
+  update-documentation-cloudflare:
+    name: Update Documentation For CloudFlare ☁️
```
RytoEX, line 97:
```suggestion
    name: Update Documentation for CloudFlare ☁️
```

.github/workflows/dispatch.yaml
```diff
@@ -0,0 +1,146 @@
...
+jobs:
+  services-validation:
+    name: Check Services Configuration Files 🕵️
+    if: github.repository_owner == 'obsproject' && inputs.job == 'services'
+    runs-on: ubuntu-22.04
+    permissions:
+      checks: write
+    steps:
+      - uses: actions/checkout@v3
+      - name: Check for defunct services 📉
```
RytoEX, line 42:
```suggestion
      - name: Check for Defunct Services 📉
```

.github/workflows/build-project.yaml
```diff
@@ -0,0 +1,346 @@
...
+        with:
+          codesignIdentity: ${{ secrets.MACOS_SIGNING_IDENTITY }}
+          codesignCertificate: ${{ secrets.MACOS_SIGNING_CERT }}
+          certificatePassword: ${{ secrets.MACOS_SIGNING_CERT_PASSWORD }}
+          keychainPassword: ${{ secrets.MACOS_KEYCHAIN_PASSWORD }}
+          provisioningProfile: ${{ secrets.MACOS_SIGNING_PROVISIONING_PROFILE }}
+          notarizationUser: ${{ secrets.MACOS_NOTARIZATION_USERNAME }}
+          notarizationPassword: ${{ secrets.MACOS_NOTARIZATION_PASSWORD }}
+
+      - name: Build OBS-Studio 🧱
```
RytoEX, line 125: "OBS Studio" or "obs-studio"

.github/workflows/build-project.yaml
```diff
@@ -0,0 +1,346 @@
...
+          YOUTUBE_SECRET: ${{ secrets.YOUTUBE_SECRET }}
+          YOUTUBE_SECRET_HASH: ${{ secrets.YOUTUBE_SECRET_HASH }}
+        with:
+          target: ${{ matrix.target }}
+          config: ${{ needs.check-event.outputs.config }}
+          codesign: ${{ fromJSON(needs.check-event.outputs.codesign) }}
+          codesignIdent: ${{ steps.codesign.outputs.codesignIdent }}
+          codesignTeam: ${{ steps.codesign.outputs.codesignTeam }}
+
+      - name: Package OBS-Studio 📀
```
RytoEX, line 143: "OBS Studio" or "obs-studio"

.github/workflows/build-project.yaml
```diff
@@ -0,0 +1,346 @@
...
+
+      - uses: actions/cache/restore@v3
+        id: ccache-cache
+        with:
+          path: ${{ github.workspace }}/.ccache
+          key: ${{ runner.os }}-ccache-x86_64-${{ needs.check-event.outputs.config }}
+          restore-keys: |
+            ${{ runner.os }}-ccache-x86_64-
+
+      - name: Build OBS-Studio 🧱
```
RytoEX, line 195: "OBS Studio" or "obs-studio"

.github/workflows/build-project.yaml
```diff
@@ -0,0 +1,346 @@
...
+          RESTREAM_HASH: ${{ secrets.RESTREAM_HASH }}
+          YOUTUBE_CLIENTID: ${{ secrets.YOUTUBE_CLIENTID }}
+          YOUTUBE_CLIENTID_HASH: ${{ secrets.YOUTUBE_CLIENTID_HASH }}
+          YOUTUBE_SECRET: ${{ secrets.YOUTUBE_SECRET }}
+          YOUTUBE_SECRET_HASH: ${{ secrets.YOUTUBE_SECRET_HASH }}
+        with:
+          target: x86_64
+          config: ${{ needs.check-event.outputs.config }}
+
+      - name: Package OBS-Studio 📀
```
RytoEX, line 210: "OBS Studio" or "obs-studio"

.github/workflows/build-project.yaml
```diff
@@ -0,0 +1,346 @@
...
+          submodules: recursive
+          fetch-depth: 0
+          set-safe-directory: ${{ env.GITHUB_WORKSPACE }}
+
+      - name: Setup Environment 🔧
+        id: setup
+        env:
+          GH_TOKEN: ${{ github.token }}
+        run: |
+          : Setup Environment 🔧
```
RytoEX, line 265: s/Setup/Set Up/

.github/workflows/build-project.yaml
```diff
@@ -0,0 +1,346 @@
...
+      - uses: actions/cache@v3
+        id: ccache-cache
+        if: github.event_name == 'pull_request'
+        with:
+          path: ${{ github.workspace }}/.ccache
+          key: ${{ runner.os }}-ccache-x86_64-${{ needs.check-event.outputs.config }}
+          restore-keys: |
+            ${{ runner.os }}-ccache-x86_64-
+
+      - name: Build OBS-Studio 🧱
```
RytoEX, line 319: "OBS Studio" or "obs-studio"

.github/workflows/build-project.yaml
```diff
@@ -0,0 +1,346 @@
...
+          YOUTUBE_CLIENTID: ${{ secrets.YOUTUBE_CLIENTID }}
+          YOUTUBE_CLIENTID_HASH: ${{ secrets.YOUTUBE_CLIENTID_HASH }}
+          YOUTUBE_SECRET: ${{ secrets.YOUTUBE_SECRET }}
+          YOUTUBE_SECRET_HASH: ${{ secrets.YOUTUBE_SECRET_HASH }}
+          GPU_PRIORITY_VAL: ${{ secrets.GPU_PRIORITY_VAL }}
+        with:
+          target: x64
+          config: ${{ needs.check-event.outputs.config }}
+
+      - name: Package OBS-Studio 📀
```
RytoEX, line 335: "OBS Studio" or "obs-studio"

.github/workflows/check-format.yaml
```diff
@@ -0,0 +1,63 @@
...
+        with:
+          failCondition: error
+
+  qt-xml-validator:
+    runs-on: ubuntu-22.04
+    steps:
+      - uses: actions/checkout@v3
+        with:
+          fetch-depth: 0
+      - name: Qt Xml Check 🖼️
```
RytoEX, line 59:
```suggestion
      - name: Qt XML Check 🖼️
```

RytoEX (changes requested):
Nits:
"Setup" is a noun. "Set up" is a phrasal verb. In Title Case, "Set Up" is the appropriate capitalization.

In Title Case, prepositions four or fewer letters do not get capitalized (e.g., for, with, to).

.github/actions/check-changes/action.yaml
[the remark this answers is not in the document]
RytoEX: To me, the name `passed` is a cognitive disconnect. `filesChanged=true` would be fine because it's obviously communicating the intent of the boolean.
This logic, in general, is fine:
```
if (checkChanges)
// something is wrong and warrants action
else()
// everything is okay and there is nothing to do
```

This specific naming, however, warrants a double-take from me:
```
if (steps.checks.outputs.passed)
// something is wrong and warrants action
else()
// everything is okay and there is nothing to do
```

It's not the boolean true implying failure that bothers me the most here, it's the boolean name itself.

Hope that clarifies what I was trying to say. I'll probably not block on this, even if I do find it confusing.

.github/actions/generate-docs/action.yaml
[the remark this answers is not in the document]
RytoEX: Well, that's a bit unfortunate. I would almost suggest that nameExtension and commitHash are processed in a step into a title string or title string suffix, because I think having a trailing space here could cause confusion in the future (and someone will suggest removing it in the future not realizing that it's necessary). In my opinion, whitespace handling is the responsibility of a string builder, not of the individual components of a string. That said, it's not a critical issue, and I won't block on this here.

.github/actions/check-changes/action.yaml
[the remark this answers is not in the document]
PatTheMav: I'm not opposed to changing the term, but given that it checks for changed files, and it's "successful" when changes are detected, what would be the appropriate term for a positive check?

.github/scripts/.build.zsh
PatTheMav: > If we set cacheVariable entries that use $env or $penv in the CI presets that already exist

We don't right now, because these specific variables are not exposed in the preset (mainly because they are meant for "internal" use of official release builds).

So even if they are set in the environment, the current CMake scripts would ignore them entirely.

I'm not opposed to adding these as default cache variables to the default presets, but that'd be a different PR.

.github/scripts/utils.py/check-jsonschema.py
```diff
@@ -0,0 +1,131 @@
+import json
+import logging
+import os
+import sys
+import argparse
+from typing import Any
+
+from json_source_map import calculate
+from json_source_map.errors import InvalidInputError
+from jsonschema import Draft7Validator
```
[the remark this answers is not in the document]
PatTheMav: They were automatically added by the Python language server, I usually don't manage imports manually because the LSP takes care of all that.

In this case the LSP only imported the functions or objects as required by the code.

.github/workflows/pr-pull.yaml
```diff
@@ -0,0 +1,94 @@
...
+          fetch-depth: 0
+
+      - name: Check for Changed Files ✅
+        uses: ./.github/actions/check-changes
+        id: checks
+        with:
+          baseRef: origin/${{ github.base_ref }}
+          checkGlob: plugins/rtmp-services/data/*.json
+
+      - name: Check for defunct services 📉
```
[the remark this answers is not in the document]
PatTheMav: Correct, defunct services are only checked in scheduled workflow runs or manual dispatches. That's the current setup as well.

.github/workflows/publish.yaml
```diff
@@ -0,0 +1,150 @@
...
+    runs-on: macos-13
+    steps:
+      - uses: actions/checkout@v3
+      - uses: ./.github/actions/steam-upload
+        with:
+          steamSecret: ${{ secrets.STEAM_SHARED_SECRET }}
+          steamUser: ${{ secrets.STEAM_USER }}
+          steamPassword: ${{ secrets.STEAM_PASSWORD }}
+          workflowSecret: ${{ github.token }}
+          preview: false
```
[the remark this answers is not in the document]
PatTheMav: The publish workflow runs for the `release` event. The Steam upload action uses the GitHub CLI to download the assets attached to a release then:

```
gh release download \
              --pattern '*macOS*.dmg' \
              --pattern '*Windows*' \
              --pattern '*.zip' \
              --clobber
```

.github/workflows/push.yaml
```diff
@@ -0,0 +1,313 @@
...
+          fetch-depth: 0
+
+      - name: Check for Changed Files ✅
+        uses: ./.github/actions/check-changes
+        id: checks
+        with:
+          baseRef: ${{ github.event.before }}
+          checkGlob: plugins/rtmp-services/data/*.json
+
+      - name: Check for defunct services 📉
```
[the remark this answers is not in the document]
PatTheMav: See above, services are only checked in scheduled or dispatch runs.

.github/actions/generate-docs/action.yaml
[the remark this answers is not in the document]
PatTheMav: I switched the code to generate the entire string placed before the commit hash in the script block, which should remedy this issue.

.github/actions/flatpak-manifest-validator/action.yaml
```diff
@@ -0,0 +1,38 @@
...
+        : Validate Flatpak Manifest 🕵️
+
+        echo ::group::Run Validation
+        if [[ '${{ inputs.failCondition }}' == 'never' ]]; then set +e; fi
+        python3 -u \
+          build-aux/format-manifest.py \
+          build-aux/com.obsproject.Studio.json \
+          --check \
+          --loglevel INFO
+        echo ::endgroup::
```
tytan652, line 38: I may be mistaken, but I don't see a check if the JSON got it's format fixed (e.g. wrong tab/space).

tytan652:
First commit and flatpak validator commit are missing `build-aux` in their prefix.

.github/workflows/pr-pull.yaml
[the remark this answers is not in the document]
RytoEX: I more meant that the name "Check for defunct services" is misleading in this invocation, because the invoked call does not, in fact, check for defunct services. Am I mistaken? Do the current invocations also have this issue with a name/label?

.github/workflows/push.yaml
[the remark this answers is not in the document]
RytoEX: As replied in the other thread, it's the mismatch between the name/label and the action that I'm highlighting, not whether or not we're checking for defunct services by itself.

.github/actions/check-changes/action.yaml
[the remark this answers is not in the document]
RytoEX: I'd have gone with `filesChanged` instead of `passed`. If you want something more easily distinguished from `changedFiles`, then perhaps `areFilesChanged` or `hasChangedFiles`. Feel free to seek another opinion in case this is just something that bothers only me.

.github/actions/generate-docs/action.yaml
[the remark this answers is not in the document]
RytoEX: Looks fine now. Thanks.

.github/workflows/dispatch.yaml
```diff
@@ -0,0 +1,146 @@
...
+        id: setup
+        run: |
+          : Get Commit Hash 🆔
+          echo "commitHash=${GITHUB_SHA:0:9}" >> $GITHUB_OUTPUT
+      - uses: ./.github/actions/generate-docs
+        with:
+          commitHash: ${{ steps.checks.setup.commitHash }}
+
+  update-documentation-cloudflare:
+    name: Update Documentation for CloudFlare ☁️
```
RytoEX, line 97:
```suggestion
    name: Update Documentation for Cloudflare ☁️
```

Cloudflare does not have a capital "F".

.github/workflows/dispatch.yaml
```diff
@@ -0,0 +1,146 @@
...
+        run: |
+          : Get Commit Hash 🆔
+          echo "commitHash=${GITHUB_SHA:0:9}" >> $GITHUB_OUTPUT
+      - uses: ./.github/actions/generate-docs
+        with:
+          commitHash: ${{ steps.checks.setup.commitHash }}
+          disableLinkExtensions: true
+
+  deploy-documentation:
+    name: Deploy Documentation to CloudFlare ☁️
```
RytoEX, line 113:
```suggestion
    name: Deploy Documentation to Cloudflare ☁️
```

RytoEX (changes requested):
Nits.

.github/scripts/.build.zsh
RytoEX: I don't understand, but it's not a big deal. It works either way, I'm just confused why this approach is preferred over handling it in presets. Maybe I'll understand it some other day. Either way, it's not worth blocking over.

.github/actions/check-changes/action.yaml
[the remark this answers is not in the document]
PatTheMav: That would make the two outputs "filesChanged" and "changedFiles" - "hasChangedFiles" would be canonical with the output in the services validator.

.github/workflows/pr-pull.yaml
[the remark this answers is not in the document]
PatTheMav: Right, yeah I just copied the block around but there's no need for the name to be identical.

.github/actions/flatpak-manifest-validator/action.yaml
PatTheMav: That's what passing the `--check` switch is for - without it, the Python script would update the manifest.

But with it, the Python script will return non-zero if any changes would be necessary.

PatTheMav: It mimics the other formatting checker, which actually _updates_ the formatting of the files, _except_ if `--check` is passed - the parameter servers specifically the purpose used on CI.

PatTheMav:
> First commit and flatpak validator commit are missing `build-aux` in their prefix.
> 
> Edit: `build-aux` is not CI-only

The commit primarily targets CI by adding an action and updates the Python script to work with it. The only option would be to split up the changes into an earlier one updating the script and then the second one updating CI. I leave that to @RytoEX to decide if we want this granularity given that the commits will need to be squashed before merging anyway.

.github/scripts/.build.zsh
PatTheMav: As I mentioned above, the current presets don't contain these variables, so they will also not be adapted from the environment (and this PR was specifically designed not to make any CMake-level changes). And neither does the CMake work with them all being empty:

One of the recent bug fixes enforces the hash being either `0` or a hexadecimal string, but if the environment variable is not set (but used in the CMake preset), it would end up failing that check, so we'd need to add extra code to handle empty hash values (even though they _should_ fail) because they might be passed via the environment, and at this point I'd vote "no" based on complexity.

So my choice was to say "these are user-provided secrets, they don't need to be set to build OBS, and user-provided values arrive either via cache definitions on the command line or via local user presets".

RytoEX: As far as I'm concerned, setting these variables to empty strings is equivalent to not setting them at all. That is how the old CMake worked, and that is also how the new CMake works.

Legacy:
https://github.com/obsproject/obs-studio/blob/8d33da1fab53c362d82b9b2fcdcdb02ba4a3a006/UI/cmake/legacy.cmake#L27-L37

New:
https://github.com/obsproject/obs-studio/blob/8d33da1fab53c362d82b9b2fcdcdb02ba4a3a006/UI/cmake/feature-twitch.cmake#L1-L10

I still don't understand how setting them to empty strings would supposedly fail. It works fine in local testing. If empty strings are specified, the feature is (correctly, in my opinion) disabled. Empty strings do not cause a configure or build or runtime failure. No changes in CMakeLists.txt or *.cmake files are required for this. The only required to handle this in a preset is to set the cache variables using env or penv calls, which moves the complexity away from shell scripts.

As mentioned earlier, I won't block on this. I just didn't understand why we were bothering to control it in a script when it can be controlled just fine in a preset, and I thought we were deferring as much config complexity as possible to scripts. Doubly so here, where as far as I know, this script is _meant_ to be run only on CI, and we can create a CI preset that is conditionally only to be enabled on CI if we were so inclined. Or not.

.github/actions/flatpak-manifest-validator/action.yaml
tytan652: > _except_ if `--check` is passed

So there is no check if spacing (4 space per indentation) is correct, which is bad.

PatTheMav: Ah thanks for pointing that out - updated the script accordingly. It will also throw an error if the formatting changes by having it compare json string generated with a 4-space indent.

build-aux/.run-format.zsh
```diff
@@ -0,0 +1,190 @@
...
+      --)
+        shift
+        args+=($@)
+        break
+        ;;
+      -c|--check) check_only=1; shift ;;
+      -v|--verbose) (( verbosity += 1 )); shift ;;
+      -h|--help) log_output ${_usage}; exit 0 ;;
+      -V|--version) print -Pr "${_version}"; exit 0 ;;
+      --debug) _verbosity=3; shift ;;
```
tytan652, line 167: I may misunderstand the intention. But it should be:
```suggestion
      --debug) verbosity=3; shift ;;
```
No ? Otherwise it's completely unused.

PatTheMav: Correct, good catch.

RytoEX (changes requested):
Mostly nits.

@derrod are the steam-upload action and the dispatch/publish workflows fine?

.github/scripts/Build-Windows.ps1
```diff
@@ -46,57 +47,40 @@ function Build {
...
+    $CmakeInstallArgs = @()
+
+    if ( $DebugPreference -eq 'Continue' ) {
+        $CmakeArgs += ('--debug-output')
+        $CmakeBuildArgs += ('--verbose')
+        $CmakeInstallArgs += ('--verbose')
     }
 
+    $CmakeBuildArgs += @(
+        '--preset', "windows-${Target}"
```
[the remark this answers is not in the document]
PatTheMav: One is the configuration preset, the other the build preset and there are no CI-specific build presets.

build-aux/.run-format.zsh
```diff
@@ -0,0 +1,190 @@
...
+# setopt WARN_CREATE_GLOBAL
+# setopt WARN_NESTED_VAR
+# setopt XTRACE
+
+autoload -Uz is-at-least && if ! is-at-least 5.2; then
+  print -u2 -PR "%F{1}${funcstack[1]##*/}:%f Running on Zsh version %B${ZSH_VERSION}%b, but Zsh %B5.2%b is the minimum supported version. Upgrade zsh to fix this issue."
+  exit 1
+fi
+
+invoke_formatter() {
```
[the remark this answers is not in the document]
PatTheMav: That was the idea. In theory one could add the autoload steps anew (it will not load anything if the function was already loaded), but `invoke_formatter` is not supposed to work standalone anyway.

.github/scripts/Build-Windows.ps1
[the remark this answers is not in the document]
RytoEX: Ah, right. I misunderstood the section. This change appears to have been removed from this PR anyway.

.github/workflows/push.yaml
```diff
@@ -0,0 +1,313 @@
+name: Push to master
```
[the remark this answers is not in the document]
RytoEX: Recent push fixed this.

RytoEX (approved):
I have no further feedback at this time. Seems okay to me. We probably won't be able to really test the workflows until we try to do a tag/release.

derrod:
> @derrod are the steam-upload action and the dispatch/publish workflows fine?

As far as I can tell they should be fine. Pat has tested them in the dry-run configuration so unless something goes wrong with the secrets it should be good.

RytoEX:
This has some minor merge conflicts in the Linux CI scripts.

RytoEX (approved):

PR #8889, opened 2023-05-11
UI: Add obs_frontend_close_main_window()

BenJuan26:
### Description

Adds the ability to close the main window from the frontend API. The behaviour is the same as clicking the Exit button in OBS Studio.

### Motivation and Context

This is useful for scripting, where OBS is mainly run from the system tray and controlled through the script, i.e. through obs-websocket. Currently, the only way to close OBS from a script is to have the OS terminate the process. This is obviously not ideal: the program doesn't get the chance to properly free up resources, and in the case of Windows, the icon hangs around the system tray until it is moused-over.

Link to feature suggestion: https://ideas.obsproject.com/posts/2225/allow-clean-shutdown-via-api

### How Has This Been Tested?

OS: Windows
CPU: Intel i7 12700K
GPU: Gigabyte Nvidia GeForce RTX 3070 Ti 8 GB
RAM: 32GB

For the sake of simplicity, I hijacked the `StartStream` handler of obs-websocket to instead call `obs_frontend_close_main_window()`, then used OBS-web to trigger the handler. The OBS Studio window closed, and the log indicated that the shutdown process completed nominally, just as it would have if the Exit button was clicked.

### Types of changes

### Checklist:

### DIFF
--- a/docs/sphinx/reference-frontend-api.rst
+++ b/docs/sphinx/reference-frontend-api.rst
@@ -251,6 +251,13 @@ Functions
 
 ---------------------------------------
 
+.. function:: void obs_frontend_close_main_window(void)
+
+   Closes the main window. If there are active outputs, the user will be
+   prompted to confirm before closing.
+
+---------------------------------------
+
 .. function:: void *obs_frontend_get_main_window(void)
 
    :return: The QMainWindow pointer to the OBS Studio window
--- a/frontend/OBSStudioAPI.cpp
+++ b/frontend/OBSStudioAPI.cpp
@@ -24,6 +24,12 @@ inline size_t GetCallbackIdx(vector<OBSStudioCallback<T>> &callbacks, T callback
 	return (size_t)-1;
 }
 
+void OBSStudioAPI::obs_frontend_close_main_window()
+{
+	blog(LOG_INFO, "Closing main window via the API...");
+	QMetaObject::invokeMethod(main, "close");
+}
+
 void *OBSStudioAPI::obs_frontend_get_main_window()
 {
 	return (void *)main;
--- a/frontend/OBSStudioAPI.hpp
+++ b/frontend/OBSStudioAPI.hpp
@@ -38,6 +38,8 @@ struct OBSStudioAPI : obs_frontend_callbacks {
 
 	inline OBSStudioAPI(OBSBasic *main_) : main(main_) {}
 
+	void obs_frontend_close_main_window(void) override;
+
 	void *obs_frontend_get_main_window(void) override;
 
 	void *obs_frontend_get_main_window_handle(void) override;
--- a/frontend/api/obs-frontend-api.cpp
+++ b/frontend/api/obs-frontend-api.cpp
@@ -56,6 +56,12 @@ static char **convert_string_list(vector<string> &strings)
 
 /* ------------------------------------------------------------------------- */
 
+void obs_frontend_close_main_window(void)
+{
+	if (callbacks_valid())
+		c->obs_frontend_close_main_window();
+}
+
 void *obs_frontend_get_main_window(void)
 {
 	return !!callbacks_valid() ? c->obs_frontend_get_main_window() : nullptr;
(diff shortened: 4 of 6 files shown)


PatTheMav:
Is this supposed to end the application? Because on macOS an application is not automatically terminated just because it has no active windows - that’s an opt-in behaviour. It could be that Qt behaves this way in our case, but that needs to be verified.

BenJuan26:
For the use case that I'm targeting, yes, the intention is to end the application.

I just tested the release version of OBS on macOS, and indeed, clicking the Exit button and closing the window both end the application.

derrod:
> Is this supposed to end the application? Because on macOS an application is not automatically terminated just because it has no active windows - that’s an opt-in behaviour. It could be that Qt behaves this way in our case, but that needs to be verified.

Our main window's close event handler does some cleanup and ultimately calls `qApplication::quit()`

https://github.com/obsproject/obs-studio/blob/c289efd84a42887feba89c3579d39a9c264b749e/UI/window-basic-main.cpp#L5042

docs/sphinx/reference-frontend-api.rst
```diff
@@ -242,6 +242,12 @@ Functions
 
 ---------------------------------------
 
+.. function:: void obs_frontend_close_main_window(void)
+
+   Closes the main window.
```
CodeYan01, line 247: Perhaps some clarification here is in order. Like, would it prompt the user if there are active outputs, or ignore outputs.

BenJuan26: I added a line about active outputs. I didn't want to make things too wordy. I could have stated that behaviour is the same as the Exit button, but I think it's clearer to just state the behaviour.

CodeYan01: Currently looks good to me. However, I would prefer:
> Closes the main window**, as if the user clicked the Exit button**. If there are active outputs, the user will be prompted to confirm before closing.

Which may or may not be necessary, and I'll leave that opinion to you and the others.

BenJuan26: I'll leave it unresolved for now, so others can weigh in.

CodeYan01 (commented):
LGTM

Warchamp7:
I'd like us to include a log line indicating when something has closed OBS this way.

asoulliereHT:
> I'd like us to include a log line indicating when something has closed OBS this way.
> 
> How does this interact with things that prevent closing the window such as running outputs?

"Closes the main window. If there are active outputs, the user will be prompted to confirm before closing."

BenJuan26:
@Warchamp7 Log line added.

> How does this interact with things that prevent closing the window such as running outputs?

This was answered in the next comment. Let me know if you have any more questions/issues.

Warchamp7 (approved):
Fine from my end with the log line added

GeneralSpace:
@BenJuan26 @Warchamp7 this would actually really help with a project I'm working on. Any chance it could be considered for inclusion in 31.1 or another minor version soon?

BenJuan26:
@Warchamp7 @CodeYan01 This very simple change is nearly two years old and many parties have expressed interest in using it. Any idea whether there are plans to merge it anytime soon? Thanks.

RayneYoruka:
It seems I'm landing here too looking for a way to gracefully shutdown obs. I don't think taskill is the best way to do it. 

The log ends up being incomplete as it counts as a forced exit.

asoulliereHT:
I'm commenting here to bump.  I'm seeing wide support for this feature and no concerns on code.  🤷

RayneYoruka:
I'm returning back with a hackery solution yet this seems to be the best! 
NirCMD can gracefully exit obs without any issues, just make a bat file and automate in your favourite software and have it have "nircmd.exe closeprocess obs64.exe"

It exists as if you used the X button to close!

BenJuan26:
Closing as unnecessary. OBS now handles termination signals correctly.

PR #8893, opened 2023-05-11
CI: Download Qt5 conditionally on Linux

norihiro:
### Description

Qt6 libraries are installed if the package is available on the system.
Uses the inverted condition to install Qt5 libraries to avoid unnecessarily install both libraries.

### Motivation and Context

Installing Qt5 library is unnecessary but it took ~17 seconds to download and install the packages.

This will also reduce ~100 MB usage of storage.

### How Has This Been Tested?

Checked CI ran without errors.

Compared the log file of with and without this commit.
The difference is only that `WrapVulkanHeaders` is not found while loading Qt6-Widgets, which should not affect the results.

### Types of changes

### Checklist:

### DIFF
--- a/CI/linux/01_install_dependencies.sh
+++ b/CI/linux/01_install_dependencies.sh
@@ -37,7 +37,10 @@ install_qt5-deps() {
     status "Install Qt5 dependencies"
     trap "caught_error 'install_qt5-deps'" ERR
 
-    sudo apt-get install --no-install-recommends -y $@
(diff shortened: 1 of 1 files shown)


RytoEX:
> `libvulkan-dev` is aded to `obs-deps` because libvulkan-dev was installed by as a dependency of qt5-deps.

Is libvulkan-dev actually needed by anything if Qt5 is not used?

norihiro:
> Is libvulkan-dev actually needed by anything

The log from cmake differed when I didn't install libvulkan-dev. I will check its really required.

PR #8905, opened 2023-05-14
mac-videotoolbox: Fix incorrect keyframe interval calculation

gxalpha:
### Description

Fixes an accidental float to integer conversion which makes the values set for `kVTCompressionPropertyKey_MaxKeyFrameInterval` hilariously wrong. We're creating a number of type `kCFNumberSInt32Type` so we should give that an int instead of a float.

### Motivation and Context

It turns out that a keyframe every 1123024896 frames isn't super useful.
Hopefully fixes reports of encoders on Intel Macs not working properly and streaming services reporting incorrect keyframe intervals.

### How Has This Been Tested?

macOS 13.3.1
Tested that when setting a 2 second keyframe interval and then retrieving the value via `VTSessionCopyProperty`, it's now `120` instead of `1123024896`.
Don't really have an Intel Mac anymore so can't test if it actually fixes the issue reported, but will ask people reporting the problem to test CI builds of this PR.

Edit: CI builds tested by someone on Discord having the problem, they say it does indeed fix it.

### Types of changes

### Checklist:

### DIFF
--- a/plugins/mac-videotoolbox/encoder.c
+++ b/plugins/mac-videotoolbox/encoder.c
@@ -626,8 +626,9 @@ static bool create_encoder(struct vt_encoder *enc)
 			kVTCompressionPropertyKey_AllowFrameReordering,
 			kVTCompressionPropertyKey_ProfileLevel};
 
-		float key_frame_interval =
-			enc->keyint * ((float)enc->fps_num / enc->fps_den);
+		SInt32 key_frame_interval =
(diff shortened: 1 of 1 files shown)


PatTheMav (approved):

gxalpha:
Changed to use `SInt32` (Declaration `typedef signed int SInt32;` in `MacTypes.h´) explicitly.

RytoEX (approved):

PR #8907, opened 2023-05-14
UI: Use frontend api for multiview

cg2121:
### Description
Makes it possible for the multiview to work with plugins, such as Decklink, NDI, etc.

### Motivation and Context
Preparation for adding the multiview to plugins in the future

### How Has This Been Tested?
Ran multiview

### Types of changes

### Checklist:

### DIFF
--- a/UI/display-helpers.hpp
+++ b/UI/display-helpers.hpp
@@ -20,6 +20,8 @@
 #include <graphics/vec4.h>
 #include <graphics/matrix4.h>
 
+#include <QWidget>
+
 static inline void GetScaleAndCenterPos(int baseCX, int baseCY, int windowCX,
 					int windowCY, int &x, int &y,
 					float &scale)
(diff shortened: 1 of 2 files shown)


Lain-B:
I'm not sure how I feel about this if the context is to use these files directly from frontend plugins. The fault is mine for setting this precedent and originally using the properties view from within the frontend plugins, but generally we shouldn't really be duplicating files like this. So, hm. I'm not sure how to proceed. It'd probably be good to ask @DDRBoxman to see how he feels about this.

norihiro:
Thinking the situation other plugins uses the multiview, I feel the best implementation would be a obs-source type instead of implementing in UI.

cg2121:
Closing, as frontend plugins shouldn't include UI files directly. Ideally we would create a plugin agnostic UI.

PR #8913, opened 2023-05-15
.gitignore: Add CMakeUserPresets.json

gxalpha:
### Description

CMake lets you define a file called `CMakeUserPresets.json` which it treats like the `CMakePresets.json` file.
It's meant for local configuration that unlike common presets should not be added to the main repo, but only apply to one system. As presets defined by the user can also inherit common presets, this can be very useful.

[Documentation](https://cmake.org/cmake/help/latest/manual/cmake-presets.7.html#introduction):
> [...] CMakePresets.json is meant to specify project-wide build details, while CMakeUserPresets.json is meant for developers to specify their own local build details.
>
>CMakePresets.json may be checked into a version control system, and CMakeUserPresets.json should NOT be checked in. For example, if a project is using Git, CMakePresets.json may be tracked, and CMakeUserPresets.json should be added to the .gitignore.

### Motivation and Context

Using user presets can save you from long commands with a lot of cmake variables (virtualcam IDs, service integration secrets, (now additional) deps prefix path, etc).

### How Has This Been Tested?

macOS 13.3.1
Created the CMakeUserPresets.json file and used a preset I made. File is not in git.

### Types of changes

### Checklist:

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -19,6 +19,7 @@
 .dirstamp
 /cmake/.CMakeBuildNumber
 .deps
+CMakeUserPresets.json
 
 #xcode
 *.xcodeproj/


RytoEX (approved):
Looks good to me.

PR #8918, opened 2023-05-17
UI: Add output wrapper class

cg2121:
### Description
Makes a class for simplifying outputs, so there is not so much duplicate code.

### Motivation and Context
Less code

- Don't have an AJA card, so refactoring of that code will be done in the future
- Also planning in a separate PR to refactor the virtual camera output with this code

### How Has This Been Tested?
Tested with Decklink output

### Types of changes

### Checklist:


UI/output-obj.cpp
```diff
@@ -0,0 +1,149 @@
...
+	case OBS_FRONTEND_EVENT_SCENE_CHANGED:
+		if (!obs_frontend_preview_program_mode_active())
+			source = obs_frontend_get_current_scene();
+
+		break;
+	default:
+		break;
+	}
+
+	if (source)
```
tt2468, line 27: I think it would be best to remove this check altogether. If a scene is changed to nullptr, it's a good time to release the current source anyway.

cg2121: Fixed

UI/output-obj.hpp
```diff
@@ -0,0 +1,42 @@
+#pragma once
+
+#include "obs.hpp"
+
+enum class OutputType {
```
tt2468, line 5: It may be smart to namespace this to prevent enum values leaking into the main namespace of any files that include the header.

UI/output-obj.hpp
```diff
@@ -0,0 +1,42 @@
...
+	enum OutputType GetType();
+	void SetType(enum OutputType type);
+
+	bool Active();
+
+	bool Start();
+	void Stop();
+
+	void SetSource(OBSSource source);
+	OBSSource GetSource();
```
tt2468, line 36: Returning an RAII is technically not necessary, as it'd theoretically increment the reference then immediately decrement it again. `obs_source_t *` should work fine

UI/output-obj.hpp
```diff
@@ -0,0 +1,42 @@
...
+
+	void UpdateSettings(OBSData settings);
```
tt2468, line 38: Same here and below, it's safe to assume that the callers of these functions will own valid references already.

UI/output-obj.hpp
```diff
@@ -0,0 +1,42 @@
...
+
+	OutputObj(const char *id, OBSData settings, enum OutputType type);
```
tt2468, line 40: AFAIK most c++ headers in the codebase put the constructor and destructor declarations at the top of the list of functions in the class declaration.

UI/output-obj.cpp
```diff
@@ -0,0 +1,148 @@
...
+{
+	return output.Get();
+}
+
+enum OutputType OutputObj::GetType()
+{
+	return type;
+}
+
+void OutputObj::SetType(enum OutputType type_)
```
tt2468, line 51: This code does not remove the frontend event callback if the type is switched from `Preview`. I recommend using this code:
```c++
{
	if (type == type_)
		return;

  bool active = Active();
  if (active)
    Stop();

	type = type_;

	if (active)
		Start();
}
```

This stops the output before changing type, causing the event callback to be removed as necessary.

UI/output-obj.cpp
```diff
@@ -0,0 +1,148 @@
...
+{
+	return active;
+}
+
+bool OutputObj::Start()
+{
+	if (Active())
+		return false;
+
+	SetType(type);
```
tt2468, line 74: This appears to do nothing.

UI/output-obj.hpp
```diff
@@ -0,0 +1,42 @@
...
+	Multiview, // TODO
+};
+
+class OutputObj {
+	bool active = false;
+
+	enum OutputType type;
+
+	OBSOutputAutoRelease output;
+	OBSWeakSourceAutoRelease weakSource;
```
tt2468, line 19: AFAIK, while views hold strong references to sources when they are set, they do not handle the `remove` signal. I think you need to handle that signal yourself to prevent removed sources from being held indefinitely.

tt2468 (changes requested):

tt2468:
For the implementation allowing you to select a specific source, it may make more sense to create a private scene and add the source as a scene item with inner scaling applied.

It may be worthwhile to consider hooking frontend events on construct and removing on destruct, then adding a switch or if statement to only run `SetSource()` when the current type is `Preview`. It would also be possible to allow changing the type between `Preview` and `Source` without restarting the output, if you feel like that's worthwhile implementing.

jpark37:
The texrender stuff allows conversion from any pixel format type, e.g. NV12, not just BGRA, because it's converting from the main texture, and not the video output texture. It's also a lot faster to convert on the GPU than use swscale. From skimming over the DeckLink changes, it looks we're losing both in this PR.

UI/output-obj.hpp
cg2121: @Lain-B has stated before that RAII should always be used with cpp.

UI/output-obj.hpp
cg2121: Same explanation as above.

cg2121:
@jpark37 would it be possible for the Decklink gpu stuff you wrote to be implemented in libobs, so any output could use it?

jpark37:
Anything's possible, but I didn't think it would be easy, which is why I made as many changes in DeckLink as I did. I saw Palana do some things around GPU rescaling, but I didn't look at his changes, so maybe some progress has already been made.

cg2121:
Converted to draft, as we need to make this an output UI, so AJA, Decklink, etc. frontend plugins can be removed in favor of an agnostic output UI.

tt2468:
This PR really should get UI code added to it. It should be a class that a UI dialog class/whatever implements as a member variable.

paulh-aja:
> @jpark37 would it be possible for the Decklink gpu stuff you wrote to be implemented in libobs, so any output could use it?

We'd definitely appreciate being able to perform GPU format conversions in the AJA Output plugin. Right now, the swscale path is too slow for UHD/4Kp60 so I've disabled those formats in the AJA output plugin.

cg2121:
I've updated and refactored the code a bit.

cg2121:
Updated to latest frontend.

cg2121:
Updated to clean up code and add ability to set video/audio encoders, and ability to set service to the wrapper.

Issue #8290, opened 2023-02-18
timer_remove does not stop timer


honzaskypala:
### Operating System Info

macOS 13

### OBS Studio Version

29.0.1

### OBS Studio Log URL

https://obsproject.com/logs/BU2JOyggR8MBJ1rE

### Expected Behavior

With OBS scripting, `timer_remove` function should stop the timer, according to the documentation

### Current Behavior

Calling timer_remove does not stop the timer, the callback still continues running.

### Steps to Reproduce

Here is a simple Python script. If you load it to OBS and press the Start button, it will start calling the callback and it will print ticks into the log. Anyway, when you press Stop button, nothing changes, the ticks still continue to be written to the log.

```
import obspython as S

INTERVAL = 1

class RepeatAction:
    def __init__(self):
        self.timer_on = False
    
    def start_pressed(self, props, prop):
        if not self.timer_on:
            self.timer_on = True
            S.timer_add(self.ticker, INTERVAL * 1000)
    
    def stop_pressed(self, props, prop):
        if self.timer_on:
            self.timer_on = False
            S.timer_remove(self.ticker)

    def ticker(self):
        print("tick")

ra = RepeatAction()

def script_properties():  # ui
    props = S.obs_properties_create()
    S.obs_properties_add_button(props, "startButton", "Start", lambda x,y: RepeatAction.start_pressed(ra, x, y))
    S.obs_properties_add_button(props, "stopButton", "Stop", lambda x,y: RepeatAction.stop_pressed(ra, x, y))
    return props
```

CodeYan01:
Please check your log, your script has an error.

honzaskypala:
This error was from previous load of the script. You can notice in the log the script was loaded later second time, not producing the error anymore.

`10:58:58.175: [obs-scripting]: Loaded python script: timer-remove-bug.py`

CodeYan01:
So, I tried adding print statements like this:
```py
def stop_pressed(self, props, prop):
        print('stop')
        if self.timer_on:
            print('stopped')
            self.timer_on = False
            S.timer_remove(self.ticker)```
and sure enough, both print statements worked, proving that it's not the if condition (which ofc looks fine), but the timer is indeed still running.

also changed the statements at the bottom to:
```py
    S.obs_properties_add_button(props, "startButton", "Start", lambda x,y: ra.start_pressed(x, y))
[...]

PR #8920, opened 2023-05-17
 obs-scripting: Allow passing python methods as callbacks

Penwy:
### Description

When adding an object as callback, obs-scripting-python requires the
object to be of function type, which excludes bound method objects, 
although they are perfectly valid as callback objects.

This changes the type check in functions that add script callbacks so 
that bound methods pass it.

### Motivation and Context

There's no technical impossibility for bound methods to be used as 
callbacks, not allowing them to be used as such only removes potential 
functionality from python scripts.

### How Has This Been Tested?

This has been tested, on Ubuntu 22.04, by using the script linked below, and 
verifying  that the bound methods it requests to set as callbacks are correctly 
added, triggered and removed.
[method_callback.zip](https://github.com/obsproject/obs-studio/files/11495058/method_callback.zip)

### Types of changes

### Checklist:

### DIFF
--- a/deps/obs-scripting/obs-scripting-python-frontend.c
+++ b/deps/obs-scripting/obs-scripting-python-frontend.c
@@ -319,7 +319,7 @@ static PyObject *remove_save_callback(PyObject *self, PyObject *args)
 
 	if (!parse_args(args, "O", &py_cb))
 		return python_none();
-	if (!py_cb || !PyFunction_Check(py_cb))
+	if (!py_cb || (!PyFunction_Check(py_cb) && !PyMethod_Check(py_cb)))
 		return python_none();
 
 	struct python_obs_callback *cb =
@@ -343,7 +343,7 @@ static PyObject *add_save_callback(PyObject *self, PyObject *args)
 
 	if (!parse_args(args, "O", &py_cb))
 		return python_none();
-	if (!py_cb || !PyFunction_Check(py_cb))
+	if (!py_cb || (!PyFunction_Check(py_cb) && !PyMethod_Check(py_cb)))
 		return python_none();
 
 	struct python_obs_callback *cb = add_python_obs_callback(script, py_cb);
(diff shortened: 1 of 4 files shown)


norihiro:
Please remove a space at the beginning of the commit message.

Penwy:
I'm sorry, which commit message?
I don't see where exactly there is an extraneous space.

norihiro:
> I'm sorry, which commit message?

It's b711f1c.
The title line has a white space at the beginning.
```patch
- obs-scripting: Allow passing python methods as callbacks
+obs-scripting: Allow passing python methods as callbacks
```

Penwy:
Sorry for that, I had missed it. Thanks.

PatTheMav:
While this might seem to work on the face of it, it has one major issue:

Bound methods are not "stable" memory-wise and OBS internally uses pointers to the actual Python function objects to maintain its list of callbacks. This works well for functions defined in the module: Their PyObject is created once when the script is parsed and are "stable" for the entire lifetime of the script object.

Bound methods are created on demand, which has unfortunate consequences, as exhibited by your script:

The memory address of `obj1.frontend_callback` in the `script_load` block might be `0x11f97cc80`. The memory address of `obj1.frontend_callback` in the `script_unload` block is `0x11f97ce80`.

So from the point of view of `libobs`, these are two entirely different functions:

* The function _added_ as a callback is not removed
* The function _removed_ as a callback was never added

The only reason this works is that we mark all callbacks as removed when a script is unloaded. So the next time a signal is emitted, `libobs` will try to call the _added_ callbacks, but remove them because they were marked as "removed".

When it comes to just calling bound functions, this difference in memory addresses (Python will generate new bound method objects as needed) is not an issue. Alas `libobs` uses these addresses to identify functions and keep track of them and bound methods generate too much "garbage" data.

So from my POV we can't merge this because it already does lead to unexpected consequences with the provided test case and might lead to even more issues and quirks script developers might not expect (see also https://github.com/obsproject/obs-studio/issues/8290 which has the same core issue).

Scripting in OBS was designed with short and simple scripts and module-level functions, OOP and bound methods are out of scope as such (at least as far as using them for interactions with `libobs` directly).

PR #8930, opened 2023-05-19
libobs: Fix ccache misses due to unstable obsconfig header

PatTheMav:
### Description
Moves version and commit hash information out of `obsconfig.h` and into a slim static library to improve ccache hit rate for local compilations and CI runs, as well as reduce amount of necessary recompilations on new commits or reconfigured CMake runs.

### Motivation and Context
Currently CMake generates the `obsconfig.h` header file each time the project is configured (either manually or automatically because a file in the dependency graph has changed).

In its current state this leads to several architectural issues:

* `obsconfig.h` is included by `obs.h` (which is in turn included by `obs-module.h`), so almost every module or piece of OBS is affected by a change of this file and leads to recompilations even though the underlying files haven't changed
* `obsconfig.h` includes version strings but also git commit hashes, which can change quite regularly - this also forces a recompilation of a majority of code in the codebase as almost everything includes this file (see above)
* `ccache` can be configured to ignore modification and creation timestamps of included files and only go by content
   * In this case a CMake reconfiguration will not lead to cache misses, but a new git hash will

This is such a common issue that Craig Scott suggested a more elegant alternative in his book "Professional CMake":

Instead of providing compiler definitions or a header file generated by CMake, a slim static library is used, which fixes several issues at once:

* Proper `const char` arrays are used in place of preprocessor strings (you see what the compiler sees)
* Only the static library is affected by a reconfiguration or a different git commit hash
* Consumers only need to link the static library to get the updated information - compilation units stay unaffected

As the link step need to happen anyway, it cuts down the amount of recompilations down to a single file.

#### How this affects Ccache

This change positively impacts our Ccache performance in multiple ways:

* By default Ccache uses two modes: "direct" and "preprocessor"
    * At first it will attempt to find a cache hit using the "direct" mode, which uses a combined hash of source code, compiler type, working directory, as well as hashes of all included files to generate a "manifest" for any given source code file
    * If this manifest lookup fails, Ccache will attempt to look up a preprocessor result and use that one as a fallback
    * If the preprocessor lookup fails as well, Ccache invokes the compiler and stores the output
* Every time the project is configured by CMake (or compiled based on a new git commit hash), the direct mode will fail, as the hash of `obsconfig.h` will have changed
* The preprocessor fallback will yield _some_ cache hits because not every file including `obsconfig.h` seems to contain its contents in the preprocessed output.

As part of our modernisation of the macOS codebase we have enabled the use of ObjC modules, which is supported by recent versions of Ccache. Module support requires enabling "direct" and "depend" mode, which affects the above in a meaningful way:

* Enabling "depend" mode disables "preprocessor" mode, and as such even this fallback cannot be used anymore, leading to a much higher cache miss rate.

This is the core reason for the 70%+ cache miss rate of our CI builds and all of which is fixed by this PR.

*Note:* Even though Ccache supports MSVC since version 4.6, this is limited to builds with embedded debug information (`/Z7`). We need product databases (`PDB` files) to be generated however, so we still have to compile Windows unaccelerated. Windows is also most severely affected by configuration/header changes, which are mitigated to some degree by adding the changing compiler definitions to affected source code files only.

### How Has This Been Tested?
* Successful compilation and runs on macOS 13, Windows 11, Ubuntu 22.04, and FreeBSD 13
* Tested Ccache hit rates with and without this fix, using new empty commits to force content changes of `obsconfig.h`, as well as simple CMake reconfigurations to force inode cache misses.

### Types of changes

### Checklist:

### DIFF
--- a/UI/cmake/legacy.cmake
+++ b/UI/cmake/legacy.cmake
@@ -370,6 +370,7 @@ if(OS_WINDOWS)
   target_compile_definitions(obs PRIVATE UNICODE _UNICODE _CRT_SECURE_NO_WARNINGS _CRT_NONSTDC_NO_WARNINGS
                                          PSAPI_VERSION=2)
 
+  set_source_files_properties(update/win-update.cpp PROPERTIES COMPILE_DEFINITIONS OBS_COMMIT="${OBS_COMMIT}")
   if(MSVC)
     target_link_options(obs PRIVATE "LINKER:/IGNORE:4098" "LINKER:/IGNORE:4099")
     target_link_libraries(obs PRIVATE OBS::w32-pthreads)
--- a/UI/cmake/os-windows.cmake
+++ b/UI/cmake/os-windows.cmake
@@ -37,6 +37,7 @@ add_library(OBS::update-helpers ALIAS obs-update-helpers)
 
 target_sources(obs-update-helpers INTERFACE win-update/win-update-helpers.cpp win-update/win-update-helpers.hpp)
 target_include_directories(obs-update-helpers INTERFACE "${CMAKE_CURRENT_SOURCE_DIR}/win-update")
+set_source_files_properties(update/win-update.cpp PROPERTIES COMPILE_DEFINITIONS OBS_COMMIT="${OBS_COMMIT}")
 
 add_subdirectory(win-update/updater)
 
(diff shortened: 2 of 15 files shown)


notr1ch:
Makes sense to me as long as it doesn't break anything.

PR #8938, opened 2023-05-20
Add carla-patchbay as OBS plugin

falkTX:
This is an alternative to #8919
Please check that PR first for the initial discussion around this.

### Description
This adds a new input filter and input source OBS plugin (combined in a single module) called "carla-patchbay".
It uses publicly-exposed Carla APIs to run the entire thing inside OBS.

It is based on the same code foundation as done for #8919, but instead of having plugin hosting directly integrated in OBS we let Carla take care of all of that.
Code-wise it is basically a bridge between an API that Carla exposes, and OBS filter/source audio APIs.

Behind the scenes a custom Carla engine runs the audio, talking with the PyQt frontend in a separate process using pipes.
Most of the code for this engine is at https://github.com/falkTX/Carla/blob/main/source/backend/engine/CarlaEngineNative.cpp
That engine runs in the same process as OBS, but everything else (plugin hosting and GUI) can be placed in separate processes as to not crash OBS if plugins do something wrong (always-bridge option to be introduced for next major release)

Contrary to #8919 this PR requires to have Carla as a dependency, either on system-level or as part of obs-deps.
For now I made that using pkg-config, which is already in place for Linux systems. macOS and Windows needs manual integration work, builds for them are expected to fail.

### Motivation and Context
This is an alternative to the PR #8919 that uses Carla as dependency instead of tight integration.

### How Has This Been Tested?
I have tested running this Carla-as-dep approach in both Linux and macOS, with Windows not directly tested together with OBS but confirmed to work in a similar fashion inside the [Cardinal](https://github.com/DISTRHO/Cardinal) project that uses the same Carla APIs.
For Cardinal on Windows, a custom Carla build is done directly in CI containing Python, Qt5 and PyQt5.

### Types of changes
Pretty much the same as #8919 we have a new audio filter plus audio source for audio-generation plugins
The same kind of parameter support is possible, but with this approach we show the parameters of all plugins (up to 100 total params)

image

### Checklist:

### Final notes
While this approach is much cleaner in terms of code and maintenance, it is not as clean for the user in the end because there is always 1 window layer in between. But that layer (Carla main window) also provides a way to have a patchbay and plugin favorites, among other things, so perhaps not a bad compromise?
It does feel a bit alien and out of place though, from users POV it is not OBS loading plugins but OBS loading Carla which then loads plugins. Up to the OBS team to tell how comfortable they are with this idea.

For macOS and Windows where Carla is likely to not be installed/available, OBS would need to include it which brings yet another Qt version (Carla UI is not ready for Qt6 yet), plus PyQt5 too.
While the UI is self-contained (it is a separate process after all, no external deps after everything is packaged correctly), it is still a considerable amount of disk space used for it. From the latest Carla 2.5.4 win64 release binaries the patchbay-plugin UI and all needed DLLs/deps take 115Mb.

Also if it wasn't obvious, this PR is also not meant to be merged as-is, but already provides working code if one uses Linux with Carla installed. Plus I cleaned up the whole cmake stuff with the recommendations made on the previous PR, no more legacy cmake stuff in place.

If this approach and its compromises are good, I can forego the previous attempt and focus on this one.
Let me know what you all think.

### DIFF
--- a/plugins/CMakeLists.txt
+++ b/plugins/CMakeLists.txt
@@ -35,6 +35,7 @@ if(OBS_CMAKE_VERSION VERSION_GREATER_EQUAL 3.0.0)
 
   # Add plugins in alphabetical order to retain order in IDE projects
   add_subdirectory(aja)
+  add_subdirectory(carla)
   if(OS_WINDOWS OR OS_MACOS)
     add_subdirectory(coreaudio-encoder)
   endif()
@@ -191,3 +192,4 @@ add_subdirectory(obs-transitions)
 add_subdirectory(rtmp-services)
 add_subdirectory(text-freetype2)
 add_subdirectory(aja)
+add_subdirectory(carla)
--- a/plugins/carla/CMakeLists.txt
+++ b/plugins/carla/CMakeLists.txt
@@ -1,0 +1,39 @@
+cmake_minimum_required(VERSION 3.16...3.25)
+
+option(ENABLE_CARLA "Enable building OBS with carla plugin host" ON)
+
+if(NOT ENABLE_CARLA)
+  if(OBS_CMAKE_VERSION VERSION_GREATER_EQUAL 3.0.0)
+    target_disable(carla)
+  else()
+    message(STATUS "OBS:  DISABLED   carla")
+  endif()
+  return()
+endif()
+
+# Use pkg-config to find carla libraries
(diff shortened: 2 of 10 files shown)


falkTX:
The changes for this PR have been incorporated in #8919, closing

PR #8950, opened 2023-05-22
UI: Redesign status bar

cg2121:
### Description
Makes status bar look better

Screenshot from 2023-05-27 03-18-09

### Motivation and Context
Based on mockup by @GeorgesStavracas. The reason for this change is to better group similar items together in the status bar, and make it look better. For example, before, the network square was all the way to the right, next to the recording timer, when it should of been next to the streaming timer.

This is not 100% complete, a few minor things still need to be done
- [x] The streaming inactive/status inactive icons need a version for light themes
- [x] So far, only themed with Yami
- [x] Needs new recording icons

### How Has This Been Tested?
Ran stream/recorded to make sure everything works

### Types of changes

### Checklist:


GeorgesStavracas:
I've pushed various changes. There are now light and dark variants of various inactive icons, but there is no mechanism to switch between them depending on the theme. This should be implemented before landing this PR.

cg2121:
The icons are now updated based on theme. I also went through and optimized all of the svg files.

GeorgesStavracas (approved):
I cannot comment too much on the Qt code itself, so I'll assume it should be mostly fine. I've tested this and it looks great here! Thank you very much

UI/window-basic-status-bar.hpp
```diff
@@ -3,30 +3,40 @@
...
+	Q_OBJECT
+
+	friend class OBSBasicStatusBar;
+
+private:
+	std::unique_ptr<Ui_StatusBarWidget> ui;
+
+public:
+	StatusBarWidget(QWidget *parent = nullptr);
+	~StatusBarWidget();
```
tytan652, line 22:
```suggestion
	inline ~StatusBarWidget(){};
```

UI/window-basic-status-bar.cpp
```diff
@@ -6,92 +6,61 @@
...
+#define MEDIOCRE_THRESHOLD 0.6667f
+#define BAD_THRESHOLD 1.0f
+
+StatusBarWidget::StatusBarWidget(QWidget *parent)
+	: QWidget(parent), ui(new Ui::StatusBarWidget)
+{
+	ui->setupUi(this);
+}
+
+StatusBarWidget::~StatusBarWidget() {}
```
tytan652, line 26:
```suggestion
```

tytan652 (changes requested):
If the destructor is empty, just inline it in the header to avoid writing an empty function.

UI/window-basic-status-bar.hpp
cg2121: The destructor needs to be in the cpp file for the std::unique_ptr to work.

Warchamp7 (approved):

UI/window-basic-status-bar.cpp
```diff
@@ -233,11 +225,30 @@ void OBSBasicStatusBar::UpdateCPUUsage()
...
+	struct obs_video_info ovi;
+	obs_get_video_info(&ovi);
+	float targetFPS = (float)ovi.fps_num / (float)ovi.fps_den;
+
+	QString text;
+	text += QString::number(obs_get_active_fps(), 'f', 2);
+	text += QString(" / ");
+	text += QString::number(targetFPS, 'f', 2);
+	text += QString(" FPS");
+
```
notr1ch, line 250: What is the point of showing the selected output FPS in the status bar? This is a static value once set, so doesn't really belong in a status bar IMO. The current FPS will also match the output FPS in the majority of cases so this looks confusing to the user ("why do I have two FPS numbers?").

Also this seems to be missing on the main screenshot of the PR - UI changes should always have accurate screenshots.

notr1ch:
The alignment of the icons looks slightly off in Dark (didn't test other themes).

align

UI/window-basic-status-bar.cpp
```diff
@@ -6,92 +6,61 @@
...
+#include "qt-wrappers.hpp"
+#include "platform.hpp"
+
+#include "ui_StatusBarWidget.h"
+
+#define CONGESTION_UPDATE_SECONDS 10
+#define EXCELLENT_THRESHOLD 0.0f
+#define GOOD_THRESHOLD 0.3333f
+#define MEDIOCRE_THRESHOLD 0.6667f
+#define BAD_THRESHOLD 1.0f
```
notr1ch, line 18: We should prefer `constexpr` over macros for new code.

UI/window-basic-status-bar.cpp
```diff
@@ -331,28 +349,37 @@ void OBSBasicStatusBar::UpdateDroppedFrames()
...
-		statusSquare->setPixmap(pixmap);
+	if (congestionArray.size() >= CONGESTION_UPDATE_SECONDS) {
+		congestionOverTime = accumulate(congestionArray.begin(),
+						congestionArray.end(), 0.0f) /
+				     (float)congestionArray.size();
+		congestionArray.clear();
+		update = true;
+	} else {
+		congestionArray.emplace_back(avgCongestion);
 	}
```
notr1ch, line 368: I understand the desire to avoid rapid state changes, but taking up to 10 seconds for the UI to notify about connection issues feels way too long. Congestion typically comes in bursts, so a user will see dropped frames increasing but the congestion icon will still show an excellent connection until the next update which doesn't seem right.

UI/window-basic-status-bar.cpp
cg2121: Updated to use constexpr

UI/window-basic-status-bar.cpp
cg2121: Updated it to 4 seconds, which seems more reasonable.

cg2121:
The alignment issues have now been fixed with the latest update of the code.

UI/window-basic-status-bar.cpp
cg2121: The screenshot has now been updated.

WizardCM:
I finally had a chance to experiment with this a bit, and have a couple of minor pieces of feedback.

The first is that I worry about elements jumping around too much when numbers change. Such as
* When active FPS changes
* When CPU usage changes

I assume this is why the kbps counter is to the left of the list of elements. What happens when bitrate drops from 1000 to 911? Will the "Dropped Frames" counter shift too?

You can see this behaviour here with CPU usage:


https://github.com/obsproject/obs-studio/assets/941350/22411367-a284-4e39-a824-0185b3c681e4

Imagine CPU usage repeatedly jumping between 9% and 10% for example. In all cases, setting a slightly wider minimum width would do the trick. Though this will depend on fonts - I know macOS has a habit of rendering fonts slightly thicker/bolder.

Secondly, I have minor concerns about re-arranging items that we've had for over 10 years without clear reasoning for the user. We've often been hurt by PR "Motivation & Context" being very barebones and unclear. This is one such example. I personally know a lot of effort goes into Georges' mockups, but we need to verbalise the reasoning in the PR and in the release notes.

UI/window-basic-status-bar.cpp
GeorgesStavracas: We went with `current / max FPS` after Fenrir suggested switching to `render time / max render time ms` upon clicking. I think it's good to maintain the direct link between both. This behaviour is not implemented in this PR, but eventually it will.

Warchamp7:
The intent here is grouping related information together. Let's get that added to Motivation as Matt mentioned for when patch notes time comes along.

As a proposal for the size changes, I'd suggest checking the size of the widget after each update. If it is larger, set that as the minimum size of the widget. Those minimum sizes should then get reset whenever that output is stopped.

This should mean sizes only "jump" up a very limited number of times and stay at that size to potentially accommodate that jump again later.

cg2121:
Updated to prevent labels from jumping around

UI/data/themes/Acri.qss
```diff
@@ -337,8 +337,20 @@ QScrollArea {
     border-radius: 4px;
 }
 
+/* Qt enforces a padding inside it's status bar, so we
```
RytoEX, line 340:
```suggestion
/* Qt enforces a padding inside its status bar, so we
```

UI/data/themes/Dark.qss
```diff
@@ -644,6 +644,22 @@ QStatusBar::item {
     border: none;
 }
 
+/* Qt enforces a padding inside it's status bar, so we
```
RytoEX, line 647:
```suggestion
/* Qt enforces a padding inside its status bar, so we
```

UI/data/themes/Grey.qss
```diff
@@ -337,8 +337,20 @@ QScrollArea {
     border-radius: 4px;
 }
 
+/* Qt enforces a padding inside it's status bar, so we
```
RytoEX, line 340:
```suggestion
/* Qt enforces a padding inside its status bar, so we
```

UI/data/themes/Light.qss
```diff
@@ -337,8 +337,20 @@ QScrollArea {
     border-radius: 4px;
 }
 
+/* Qt enforces a padding inside it's status bar, so we
```
RytoEX, line 340:
```suggestion
/* Qt enforces a padding inside its status bar, so we
```

UI/data/themes/Rachni.qss
```diff
@@ -339,8 +339,20 @@ QScrollArea {
     border-radius: 4px;
 }
 
+/* Qt enforces a padding inside it's status bar, so we
```
RytoEX, line 342:
```suggestion
/* Qt enforces a padding inside its status bar, so we
```

UI/data/themes/System.qss
```diff
@@ -402,3 +402,8 @@ QCalendarWidget #qt_calendar_nextmonth {
     qproperty-icon: url(./Dark/right.svg);
     icon-size: 16px, 16px;
 }
+
+/* Status bar */
```
RytoEX, line 406:
```suggestion
/* Status Bar */
```

UI/data/themes/Yami.qss
```diff
@@ -341,8 +340,20 @@ QScrollArea {
     border-radius: 4px;
 }
 
+/* Qt enforces a padding inside it's status bar, so we
```
RytoEX, line 343:
```suggestion
/* Qt enforces a padding inside its status bar, so we
```

UI/window-basic-status-bar.hpp
```diff
@@ -3,30 +3,40 @@
 #include <QStatusBar>
 #include <QPointer>
 #include <QTimer>
-#include <util/platform.h>
 #include <obs.h>
+#include <memory>
 
+class Ui_StatusBarWidget;
 class QLabel;
```
RytoEX, line 10: Is QLabel still used here? I don't see any class members of this type.

RytoEX (changes requested):
Mostly nits. One or two actual code questions. May take another deeper look early next week.

UI/window-basic-status-bar.hpp
cg2121: Fixed

Warchamp7 (approved):

UI/data/themes/Yami.qss
```diff
@@ -135,7 +135,6 @@ SourceTree QWidget {
     border: 1px solid palette(dark);
 }
 
-
```
RytoEX, line 138: This extra line appears in every theme file that used the same base, and it is not the only double blank line before a comment header/label. Might as well leave it be.

RytoEX (changes requested):
One nit.

PatTheMav:
> The intent here is grouping related information together. Let's get that added to Motivation as Matt mentioned for when patch notes time comes along.
> 
> As a proposal for the size changes, I'd suggest checking the size of the widget after each update. If it is larger, set that as the minimum size of the widget. Those minimum sizes should then get reset whenever that output is stopped.
> 
> This should mean sizes only "jump" up a very limited number of times and stay at that size to potentially accommodate that jump again later.

Doesn't need to be addressed in this PR necessarily, but given that `CPU` is not a localised label, the maximum width of the widget is naturally limited:

`CPU: 000.0%`

As such a fixed width that should work? (The absolute "correct" way would be to have a fixed width widget, which contains the left-aligned label, and the right-aligned value enclosed).

mufunyo:
> `CPU` is not a localised label

Shouldn't it be? I can imagine a few languages like Quebecois having a different abbreviation for it.

PatTheMav:
> > `CPU` is not a localised label
> 
> Shouldn't it be? I can imagine a few languages like Quebecois having a different abbreviation for it.

Maybe? But it isn't right now and the PR doesn't change it:

https://github.com/obsproject/obs-studio/blob/fcf0d377195a39ef4062b88506bb90e6ebe88de9/UI/window-basic-status-bar.cpp#L225

UI/window-basic-status-bar.cpp
```diff
@@ -233,11 +223,34 @@ void OBSBasicStatusBar::UpdateCPUUsage()
...
-	cpuUsage->setText(text);
-	cpuUsage->setMinimumWidth(cpuUsage->width());
+	UpdateCurrentFPS();
+}
+
+void OBSBasicStatusBar::UpdateCurrentFPS()
+{
+	OBSBasic *main = qobject_cast<OBSBasic *>(parent());
+	if (!main)
+		return;
```
gxalpha, line 239: This can't ever be `NULL` I don't think.
Either way, the variable is unused here and can get removed.

UI/window-basic-status-bar.cpp
```diff
@@ -233,11 +223,34 @@ void OBSBasicStatusBar::UpdateCPUUsage()
...
+
+	struct obs_video_info ovi;
+	obs_get_video_info(&ovi);
+	float targetFPS = (float)ovi.fps_num / (float)ovi.fps_den;
+
+	QString text;
+	text += QString::number(obs_get_active_fps(), 'f', 2);
+	text += QString(" / ");
+	text += QString::number(targetFPS, 'f', 2);
+	text += QString(" FPS");
```
gxalpha, line 249: It could be worth doing `QString::asprintf` instead to avoid all the appending.

sebastian-s-beckmann (commented):
Wouldn't block this PR on it but the notification text looks misaligned here, should probably be centered:
image
Other minor notes below.

UI/data/themes/Yami.qss
RytoEX: I've force-pushed the change for this nit.

RytoEX:
I've force pushed the change to resolve my nit and also rebased the PR.

RytoEX:
I've force-pushed a change to resolve the clang-format failure and rebase the PR again. CI should hopefully be green now.

RytoEX (approved):
Approving to clear my previous "changes requested" status. My nits have been addressed.

PR #8955, opened 2023-05-22
frontend: Add frontend event and output signal for replay buffer saving

Penwy:
### Description

This adds a frontend event and a signal on the replay buffer's signal 
handler that trigger when the replay buffer is saving (i.e. when a save 
has been requested, but before it saves).
The frontend event triggers the moment the frontend requests a save 
(either through hotkey or the button), whereas the signal triggers when 
the replay buffer actually starts saving (separated by the encoding delay).

N.B.: Allowing the save hotkey to trigger a frontend event required 
moving its registration from obs-ffmpeg (in the definition of the replay 
buffer output) to the UI (with all the frontend hotkeys for outputs).
This actually constitutes the biggest change in this PR.

### Motivation and Context

Motivation is mainly to bring replay buffer save to feature parity with 
other output events, which almost all have both "before" and "after" events.
A specific usecase envisionned was to allow the dynamic changing of 
replay buffer saving path/filename rather than having one set-in-stone
save formatting for the whole session, but many other usecases could be 
possible.

Moving over the hotkey registration also has the added benefit of moving 
its location next to the other replay buffer hotkeys in the settings UI, 
which creates a more consistent UX.

### How Has This Been Tested?

This has been tested on Ubuntu 22.04.
Changes to hotkey registration seem to not have affected its behaviour.
Event and signal trigger correctly from hotkey and button.

### Types of changes

### Checklist:

### DIFF
--- a/frontend/widgets/OBSBasic.hpp
+++ b/frontend/widgets/OBSBasic.hpp
@@ -492,7 +492,7 @@ private:
 
 	obs_hotkey_pair_id streamingHotkeys, recordingHotkeys, pauseHotkeys, replayBufHotkeys, vcamHotkeys,
 		togglePreviewHotkeys, contextBarHotkeys;
-	obs_hotkey_id forceStreamingStopHotkey, splitFileHotkey, addChapterHotkey;
+	obs_hotkey_id forceStreamingStopHotkey, splitFileHotkey, addChapterHotkey, replayBufSaveHotkey;
 
 	void InitHotkeys();
 	void CreateHotkeys();
--- a/frontend/widgets/OBSBasic.hpp
+++ b/frontend/widgets/OBSBasic.hpp
@@ -492,7 +492,7 @@ private:
 
 	obs_hotkey_pair_id streamingHotkeys, recordingHotkeys, pauseHotkeys, replayBufHotkeys, vcamHotkeys,
 		togglePreviewHotkeys, contextBarHotkeys;
-	obs_hotkey_id forceStreamingStopHotkey, splitFileHotkey, addChapterHotkey;
+	obs_hotkey_id forceStreamingStopHotkey, splitFileHotkey, addChapterHotkey, saveReplayBufferHotkey;
 
 	void InitHotkeys();
 	void CreateHotkeys();
--- a/frontend/widgets/OBSBasic_Hotkeys.cpp
+++ b/frontend/widgets/OBSBasic_Hotkeys.cpp
@@ -206,6 +206,30 @@ void OBSBasic::CreateHotkeys()
 						  this, this);
 	LoadHotkeyPair(replayBufHotkeys, "OBSBasic.StartReplayBuffer", "OBSBasic.StopReplayBuffer");
 
+	auto replayBufferCallback = [](void *data, obs_hotkey_id, obs_hotkey_t *, bool pressed) {
+		OBSBasic *basic = static_cast<OBSBasic *>(data);
+		if (basic->outputHandler->ReplayBufferActive() && pressed) {
+			blog(LOG_INFO, "Saving replay buffer due to hotkey");
+			basic->ReplayBufferSave();
+		}
+	};
+
+	replayBufSaveHotkey = obs_hotkey_register_frontend("OBSBasic.SaveReplayBuffer", Str("Basic.Main.SaveReplay"),
+							   replayBufferCallback, this);
+
+	const char *exists = config_get_string(activeConfiguration, "Hotkeys", "OBSBasic.SaveReplayBuffer");
+	if (exists) {
+		LoadHotkey(replayBufSaveHotkey, "OBSBasic.SaveReplayBuffer");
+	} else {
+		OBSDataArrayAutoRelease array = obs_data_get_array(LoadHotkeyData("ReplayBuffer"), "ReplayBuffer.Save");
+		obs_hotkey_load(replayBufSaveHotkey, array);
+
+		OBSDataAutoRelease newData = obs_data_create();
+		obs_data_set_array(newData, "bindings", array);
+		config_set_string(activeConfiguration, "Hotkeys", "OBSBasic.SaveReplayBuffer",
+				  obs_data_get_json(newData));
+	}
+
 	if (vcamEnabled) {
 		vcamHotkeys = obs_hotkey_pair_register_frontend(
 			"OBSBasic.StartVirtualCam", Str("Basic.Main.StartVirtualCam"), "OBSBasic.StopVirtualCam",
--- a/frontend/widgets/OBSBasic_Hotkeys.cpp
+++ b/frontend/widgets/OBSBasic_Hotkeys.cpp
@@ -206,6 +206,30 @@ void OBSBasic::CreateHotkeys()
 						  this, this);
 	LoadHotkeyPair(replayBufHotkeys, "OBSBasic.StartReplayBuffer", "OBSBasic.StopReplayBuffer");
 
+	auto replayBufferCallback = [](void *data, obs_hotkey_id, obs_hotkey_t *, bool pressed) {
+		OBSBasic *basic = static_cast<OBSBasic *>(data);
+		if (basic->outputHandler->ReplayBufferActive() && pressed) {
+			blog(LOG_INFO, "Saving replay buffer due to hotkey");
+			basic->ReplayBufferSave();
+		}
+	};
+
+	saveReplayBufferHotkey = obs_hotkey_register_frontend("OBSBasic.SaveReplayBuffer", Str("Basic.Main.SaveReplay"),
+							      replayBufferCallback, this);
+
+	bool hasSaveReplayHotkey = config_has_user_value(activeConfiguration, "Hotkeys", "OBSBasic.SaveReplayBuffer");
+	if (hasSaveReplayHotkey) {
+		LoadHotkey(saveReplayBufferHotkey, "OBSBasic.SaveReplayBuffer");
+	} else {
+		OBSDataArrayAutoRelease array = obs_data_get_array(LoadHotkeyData("ReplayBuffer"), "ReplayBuffer.Save");
+		obs_hotkey_load(saveReplayBufferHotkey, array);
+
+		OBSDataAutoRelease newData = obs_data_create();
+		obs_data_set_array(newData, "bindings", array);
+		config_set_string(activeConfiguration, "Hotkeys", "OBSBasic.SaveReplayBuffer",
+				  obs_data_get_json(newData));
+	}
+
 	if (vcamEnabled) {
 		vcamHotkeys = obs_hotkey_pair_register_frontend(
 			"OBSBasic.StartVirtualCam", Str("Basic.Main.StartVirtualCam"), "OBSBasic.StopVirtualCam",
--- a/frontend/widgets/OBSBasic_ReplayBuffer.cpp
+++ b/frontend/widgets/OBSBasic_ReplayBuffer.cpp
@@ -147,6 +147,8 @@ void OBSBasic::ReplayBufferSave()
 	if (!outputHandler->ReplayBufferActive())
 		return;
 
+	OnEvent(OBS_FRONTEND_EVENT_REPLAY_BUFFER_SAVING);
+
 	calldata_t cd = {0};
 	proc_handler_t *ph = obs_output_get_proc_handler(outputHandler->replayBuffer);
 	proc_handler_call(ph, "save", &cd);
--- a/plugins/obs-ffmpeg/data/locale/en-US.ini
+++ b/plugins/obs-ffmpeg/data/locale/en-US.ini
@@ -100,7 +100,6 @@ MediaFileFilter.AudioFiles="Audio Files"
 MediaFileFilter.AllFiles="All Files"
 
 ReplayBuffer="Replay Buffer"
-ReplayBuffer.Save="Save Replay"
 
 HelperProcessFailed="Unable to start the recording helper process. Check that OBS files have not been blocked or removed by any 3rd party antivirus / security software."
 UnableToWritePath="Unable to write to %1. Make sure you're using a recording path which your user account is allowed to write to and that there is sufficient disk space."
--- a/plugins/obs-ffmpeg/obs-ffmpeg-mux.c
+++ b/plugins/obs-ffmpeg/obs-ffmpeg-mux.c
@@ -914,30 +914,20 @@ static const char *replay_buffer_getname(void *type)
 	return obs_module_text("ReplayBuffer");
 }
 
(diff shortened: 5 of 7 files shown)


UI/obs-frontend-api/obs-frontend-api.h
```diff
@@ -50,6 +50,7 @@ enum obs_frontend_event {
 
 	OBS_FRONTEND_EVENT_TRANSITION_DURATION_CHANGED,
 	OBS_FRONTEND_EVENT_REPLAY_BUFFER_SAVED,
+	OBS_FRONTEND_EVENT_REPLAY_BUFFER_SAVING,
```
RytoEX, line 53: You must add new events to the end of the enum.

UI/window-basic-main.cpp
```diff
@@ -2670,6 +2670,20 @@ void OBSBasic::CreateHotkeys()
 	LoadHotkeyPair(replayBufHotkeys, "OBSBasic.StartReplayBuffer",
 		       "OBSBasic.StopReplayBuffer");
 
+	auto rb_sv_cb = [](void *data, obs_hotkey_id, obs_hotkey_t *,
```
RytoEX, line 2673: Use a meaningful variable name. Also we prefer camelCase in C++ and snake_case in C.

UI/window-basic-main.cpp
```diff
@@ -2670,6 +2670,20 @@ void OBSBasic::CreateHotkeys()
...
+			   bool pressed) {
+		OBSBasic &basic = *static_cast<OBSBasic *>(data);
+		if (basic.outputHandler->ReplayBufferActive() && pressed) {
+			blog(LOG_INFO, "Saving replay buffer due to hotkey");
+			basic.ReplayBufferSave();
+		}
+	};
+
+	replayBufSavHotkey = obs_hotkey_register_frontend(
```
RytoEX, line 2682: Just use `replayBufSaveHotkey`. Don't see the point in omitting `e`.

RytoEX (changes requested):
Is the `hotkey` member of `struct ffmpeg_muxer` used anywhere at all now? Should it be removed?

Penwy:
the `hotkey` member isn't used anywhere in obs code as far as I can see, but the commit that put it there, 87a64e6, states it does so to "Allows the ability to use it for more output types". So I preferred to not remove it in case it was used by some third party's output.
If you think that's mistaken, I'll gladly remove it.

UI/window-basic-main.cpp
```diff
@@ -2670,6 +2670,20 @@ void OBSBasic::CreateHotkeys()
 	LoadHotkeyPair(replayBufHotkeys, "OBSBasic.StartReplayBuffer",
 		       "OBSBasic.StopReplayBuffer");
 
+	auto replayBufferCallback = [](void *data, obs_hotkey_id,
+				       obs_hotkey_t *, bool pressed) {
+		OBSBasic &basic = *static_cast<OBSBasic *>(data);
```
gxalpha, line 2675: Personally not a fan of this syntax. Not sure what others say, I prefer just using pointers directly (which is how it’s done in most other places).

UI/window-basic-main.cpp
```diff
@@ -2670,6 +2670,20 @@ void OBSBasic::CreateHotkeys()
...
+		if (basic.outputHandler->ReplayBufferActive() && pressed) {
+			blog(LOG_INFO, "Saving replay buffer due to hotkey");
+			basic.ReplayBufferSave();
+		}
+	};
+
+	replayBufSaveHotkey = obs_hotkey_register_frontend(
+		"OBSBasic.SaveReplayBuffer", Str("Basic.Main.SaveReplay"),
+		replayBufferCallback, this);
```
gxalpha, line 2684: Are you *sure* that existing hotkeys are preserved here?
Not only is the registering function now _frontend instead of _output, but the id is also changed from ReplayBuffer.Save to OBSBasic.SaveReplayBuffer.

sebastian-s-beckmann (commented):
Are the changes here all dependent on each other? If not, it might make sense to split them into multiple PRs.
It might also make sense to combine "UI: Register the replay buffer save hotkey in the frontend" and "obs-ffmpeg: Clean up now-redundant hotkey registering code" into one commit, given that otherwise you’d have two replay buffer hotkeys at the same time if you only have one of the commits.

Would also like a @Warchamp7 review on the fact that the save replay buffer hotkey is now always available, even if replay buffer is disabled.

Penwy:
The two parts ("add signal and event" and "move hotkey registration") are not technically dependent on each other.

However, if the hotkey isn't moved, saving by pressing the hotkey will not trigger the firing of the frontend event (unless you can access the frontend api from within ffmpeg-mux). Event will only be sent when pressing the UI button, which does not feel to me like wishable behaviour.

Warchamp7:
> Would also like a @Warchamp7 review on the fact that the save replay buffer hotkey is now always available, even if replay buffer is disabled.

I am okay with this change

UI/window-basic-main.cpp
Penwy: resolved by checking on hotkey load whether "new" key data exists, and if not, attempting to load "old" key data.

Penwy:
(from what I can see the failed checks are just a result of temporary changes to the CI when this was last pushed, not of the changes introduced by this PR)

NullFragment:
I just wanted to selfishly poke over here and see what's up. It'd be nice to have this functionality for using plugins with the replay buffer in the main branch instead of having to build with these changes pulled or use custom build artifacts.

@gxalpha is the only concern with this change right now (other than merge conflicts) that the two commits for moving the hotkey should be squashed into one?

Midtan:
trying to bump this up again for equally selfish reasons. could we get this looked at again and ideally merged? as it stands right now, for many use cases we cannot use the official OBS release as a replacement for shadowplay, and using @Penwy's build works great, but has the problem that some anti cheat software blocks OBS from recording specific games (like Helldivers 2 for example) since the fingerprints don't match (at least I assume that's the reason, official builds work fine, but any fork seems to be blocked).
Me and a few others I know of would be very grateful to see this merged.

barrelltitor:
@gxalpha  @RytoEX  @Warchamp7 what is currently blocking this? I'd be happy to help make any changes required to get this merged. 

It seems like a very simple and small change, so I don't really understand why it's waiting for 2 years at this point

Fenrirthviti:
It's pending further review because it's a very low priority, and a niche use case. We don't know if any changes are necessary, because we haven't had time to look.

There's over 300 other PRs open at the time of writing, many of which are "simple changes" that have been waiting far longer than this PR to come up in review priority.

If this is critical to your workflow, we recommend pulling the change in to a fork and building yourself.

Thanks for your understanding.

Warchamp7 (approved):

frontend/widgets/OBSBasic_ReplayBuffer.cpp
```diff
@@ -147,6 +147,9 @@ void OBSBasic::ReplayBufferSave()
 	if (!outputHandler->ReplayBufferActive())
 		return;
 
+	if (api)
+		api->on_event(OBS_FRONTEND_EVENT_REPLAY_BUFFER_SAVING);
```
Warchamp7: Since this is part of OBSBasic it can call `OnEvent()` like all our other emitters.

Penwy: My apologies, had missed that being added since the initial PR.
Also took the opportunity to update commit and PR prefixes to the new ones.

plugins/obs-ffmpeg/data/locale/en-US.ini
```diff
@@ -100,7 +100,6 @@ MediaFileFilter.AudioFiles="Audio Files"
 MediaFileFilter.AllFiles="All Files"
 
 ReplayBuffer="Replay Buffer"
-ReplayBuffer.Save="Save Replay"
```
Warchamp7, line 103: This seems like an error

Warchamp7 (commented):
Looks fine at a glance. Tested Replay Buffer and it all still worked fine. Code needs a closer look from someone still (Maybe me later) and ideally a test of the new signal.

plugins/obs-ffmpeg/data/locale/en-US.ini
Penwy: It isn't, this locale string is only used for the hotkey registration [in the obs-ffmpeg plugin](https://github.com/obsproject/obs-studio/blob/84b8d16134753c9d5e7e12231b11503eb8f5fb43/plugins/obs-ffmpeg/obs-ffmpeg-mux.c#L957).

Since this PR moves the hotkey registration to the frontend instead of the obs-ffmpeg plugin, we can then use the locale string [from the frontend](https://github.com/obsproject/obs-studio/blob/84b8d16134753c9d5e7e12231b11503eb8f5fb43/frontend/data/locale/en-US.ini#L770), and this one becomes unused, hence the removal.

Penwy:
For the matter of testing the new signal, good to note is that I am distributing [an obspython script](https://github.com/Penwy/adv-ff) which makes use of the signal implemented by this PR, and since october 2024, shipping alongside it custom builds of OBS that includes this PR.

According to github release stats, those builds have been downloaded more than 800 times since then, and I have on multiple occasions interacted with users that were using it without issue.

plugins/obs-ffmpeg/data/locale/en-US.ini
Warchamp7: Aha, gotcha.

Warchamp7 (approved):

plugins/obs-ffmpeg/obs-ffmpeg-mux.c
```diff
@@ -914,30 +914,20 @@ static const char *replay_buffer_getname(void *type)
...
 		}
 
 		stream->save_ts = os_gettime_ns() / 1000LL;
 	}
-}
-
-static void save_replay_proc(void *data, calldata_t *cd)
-{
-	replay_buffer_hotkey(data, 0, NULL, true);
 	UNUSED_PARAMETER(cd);
```
RytoEX: If unused and there are early returns, this should probably be at the top.

plugins/obs-ffmpeg/obs-ffmpeg-mux.c
```diff
@@ -1154,6 +1139,12 @@ static void replay_buffer_save(struct ffmpeg_muxer *stream)
 	int64_t audio_offsets[MAX_AUDIO_MIXES] = {0};
 	int64_t audio_dts_offsets[MAX_AUDIO_MIXES] = {0};
 
+	calldata_t cd = {0};
+	calldata_set_ptr(&cd, "output", stream->output);
+	signal_handler_t *sh = obs_output_get_signal_handler(stream->output);
+	signal_handler_signal(sh, "saving", &cd);
+	calldata_free(&cd);
+
```
RytoEX: I think this could be set after the for loop, since muxing doesn't start until later, but in practice it's probably fine to send this signal a little early.

RytoEX (changes requested):

frontend/widgets/OBSBasic_Hotkeys.cpp
```diff
@@ -210,6 +206,30 @@ void OBSBasic::CreateHotkeys()
...
+		if (basic->outputHandler->ReplayBufferActive() && pressed) {
+			blog(LOG_INFO, "Saving replay buffer due to hotkey");
+			basic->ReplayBufferSave();
+		}
+	};
+
+	replayBufSaveHotkey = obs_hotkey_register_frontend("OBSBasic.SaveReplayBuffer", Str("Basic.Main.SaveReplay"),
+							   replayBufferCallback, this);
+
+	const char *exists = config_get_string(activeConfiguration, "Hotkeys", "OBSBasic.SaveReplayBuffer");
```
PatTheMav:
```suggestion
	bool hasSaveReplayHotkey = config_has_user_value(activeConfiguration, "Hotkeys", "OBSBasic.SaveReplayBuffer");
```
As far as I can tell the actual string value is not used by the code, so a boolean value should suffice.

frontend/widgets/OBSBasic.hpp
```diff
@@ -495,7 +492,7 @@ private slots:
 
 	obs_hotkey_pair_id streamingHotkeys, recordingHotkeys, pauseHotkeys, replayBufHotkeys, vcamHotkeys,
 		togglePreviewHotkeys, contextBarHotkeys;
-	obs_hotkey_id forceStreamingStopHotkey, splitFileHotkey, addChapterHotkey;
+	obs_hotkey_id forceStreamingStopHotkey, splitFileHotkey, addChapterHotkey, replayBufSaveHotkey;
```
PatTheMav:
```suggestion
	obs_hotkey_id forceStreamingStopHotkey, splitFileHotkey, addChapterHotkey, saveReplayBufferHotkey;
```
> Just use natural language and not nonHumanUseLanguage.

EDIT: This was unnecessarily snarky and I was rightfully called out for that off-stream.

I would prefer this variant because it fits with the scheme used for the existing variables.

PatTheMav (changes requested):
What's the order of function calls when saving a replay buffer via hotkey with these changes across the frontend and `obs-ffmpeg`?

Also ensure that all commits are atomic and represent complete units of change (so no commit leaves the project in an unusable or broken state).

Penwy:
Hotkey pressed > `replayBufferCallback` > `ReplayBufferSave` > `save_replay_proc`
`save_replay_proc` sets the save timestamp of the muxer to current time + 1 ms
When the muxer recieves a packet it runs `replay_buffer_data`, which checks the save timestamp, and runs `replay_buffer_save` if it has been passed
`replay_buffer_save` starts a thread that muxes and writes the file.
From `ReplayBufferSave` it becomes the exact same route as pressing the save button, or using the frontend API call.

Is there a specific commit that you consider not to be atomic, or was that general advice?

PatTheMav:
> Hotkey pressed > `replayBufferCallback` > `ReplayBufferSave` > `save_replay_proc` `save_replay_proc` sets the save timestamp of the muxer to current time + 1 ms When the muxer recieves a packet it runs `replay_buffer_data`, which checks the save timestamp, and runs `replay_buffer_save` if it has been passed `replay_buffer_save` starts a thread that muxes and writes the file. From `ReplayBufferSave` it becomes the exact same route as pressing the save button, or using the frontend API call.

I was mostly looking for the missing link between `ReplayBufferSave` and how we end up in `replay_buffer_save`, which is done via the `save` proc_handler.

> Is there a specific commit that you consider not to be atomic, or was that general advice?

General advice because it's part of our contribution guidelines and I've started making it part of my reviews.

PatTheMav (approved):

RytoEX (approved):

PR #8963, opened 2023-05-24
UI: Don't use ellipses in permissions window title

gxalpha:
### Description

beab803599e5b9e90e306b2da1f0484064ad9748 changed the string to add ellipses as recommend for menu items that require further action, however the same string was used as the window title (in fact, the key is `MacPermissions.Title`) so the ellipses would also appear in the window title:
image
That is not desired.

### Motivation and Context

Don't want ellipses in window title.

### How Has This Been Tested?

macOS 13.4
Ellipses is still in menu item but no longer in dialog title.
image

### Types of changes

### Checklist:


RytoEX (approved):
Seems fine at a glance.

PR #8979, opened 2023-05-27
text-freetype2: Fix crash after reaching buffer size

norihiro:
### Description

Test a pointer `src_glyph` is not NULL before the contents.

### Motivation and Context

A buffer `texbuf` has a fixed size that won't expand. When a lot of multi-byte characters has arrived, it overflow and `src_glyph` becomes NULL.

### How Has This Been Tested?

OS: Linux

- Configure as below.
  - Use a Japanese font (`IPAexMincho`, size 72).
  - Read from file (`ja.txt`)
  - Set `Word Wrap` on
- Prepare some Japanese text files (`ja-1.txt`, `ja-2.txt`, ...)
- On a shell, update `ja.txt` like below.
  ```
  for i in ja-?.txt; do sleep 2; cat $i > ja.txt; done
  ```
- Check log shows a message below. (Without this PR, obs crashed right after this message.)
  ```
  Out of space trying to render glyphs
  ```

### Types of changes

### Checklist:

### DIFF
--- a/plugins/text-freetype2/text-functionality.c
+++ b/plugins/text-freetype2/text-functionality.c
@@ -140,7 +140,8 @@ void set_up_vertex_buffer(struct ft2_source *srcdata)
 	next_char:;
 		glyph_index =
 			FT_Get_Char_Index(srcdata->font_face, srcdata->text[i]);
-		word_width += src_glyph->xadv;
+		if (src_glyph)
+			word_width += src_glyph->xadv;
 	eos_skip:;
 	}
 


kkartaltepe:
On overflow isnt `src_glyph` macro it may not be zeroed memory, is it possible to wrap up a length check into the `src_glyph` macro (maybe not since its used for assignment too?). 

https://github.com/obsproject/obs-studio/blob/1bda7614855224b556b5cbf028142759248155d3/plugins/text-freetype2/text-freetype2.h#L24

kkartaltepe (approved):

RytoEX:
> On overflow isnt `src_glyph` macro it may not be zeroed memory, is it possible to wrap up a length check into the `src_glyph` macro (maybe not since its used for assignment too?).
> 
> https://github.com/obsproject/obs-studio/blob/1bda7614855224b556b5cbf028142759248155d3/plugins/text-freetype2/text-freetype2.h#L24

I would rather find a way to replace the macro with anything other than a macro, but that's a separate problem.

norihiro:
There are a lot of macros and also a lot of goto statements that make it difficult to read the flow. The code should be cleaned-up but it's a different topic.

PR #8984, opened 2023-05-28
obs-scripting: Fix python save callback Py_BuildValue

Penwy:
### Description

Any frontend_save_callback added through a python script, when triggered,
doesn't recieve any calldata, and the triggering often causes errors in the log.
This is caused byPy_BuildValue being directly passed a C bool, which the function
doesn't accept, causing the building of the arguments passed to the callback to fail.
This fixes the callback by passing instead a Py_Bool object built from the C bool.

To reproduce the issue :
- add the script linked below to obs
- trigger a frontend save event (typically, changing settings and applying them)
- check the log for the arguments the callback recieved and errors

[Log showcasing the issue.](https://obsproject.com/logs/ESIvJQ8N2n4KR5e9)
tester script : [save_cb_tester.zip](https://github.com/obsproject/obs-studio/files/11585412/save_cb_tester.zip)

### Motivation and Context

Those function are included in the scripting API, they should be working as intended.

### How Has This Been Tested?

Tested on ubuntu 22.04, obs 29.1.1, python 3.10.
Tested with the script linked above, checking that it returns the correct calldata
and no error

### Types of changes

### Checklist:

### DIFF
--- a/deps/obs-scripting/obs-scripting-python-frontend.c
+++ b/deps/obs-scripting/obs-scripting-python-frontend.c
@@ -290,7 +290,8 @@ static void frontend_save_callback(obs_data_t *save_data, bool saving,
 	PyObject *py_save_data;
 
 	if (libobs_to_py(obs_data_t, save_data, false, &py_save_data)) {
-		PyObject *args = Py_BuildValue("(Op)", py_save_data, saving);
+		PyObject *args = Py_BuildValue("(ON)", py_save_data,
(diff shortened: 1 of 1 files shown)


norihiro:
Would you fix the title as below.
```patch
-obs-scripting: fix python save callback Py_BuildValue
+obs-scripting: Fix python save callback Py_BuildValue
```

PatTheMav (approved):

PR #8991, opened 2023-05-30
obs-ffmpeg: Fix AMF encoder reconfiguration

rhutsAMD:
### Description

For AMF, before a component can be reinitialized with the `ReInit` method, either `Flush` or `Drain` needs to be called on the same object to remove any frames which are still in the component for processing.

### Motivation and Context

Calls to either `Flush` or `Drain` must be done before any calls to `ReInit` for AMF. This was missing in the current code and would cause encoding to break if the bitrate was changed for example while recording or streaming. This resolves issue #8548.

### How Has This Been Tested?

Tested on Windows 10 with an AMD Radeon RX 7900 XTX.

### Types of changes

Bug fix for issue #8548 

### Checklist:


Lain-B:
LGTM :slightly_smiling_face:

PR #9001, opened 2023-06-01
Allow encoders to output at partial frame rates compared to main compositing frame rate

palana:
### Description
This allows setting `skip_frames` for encoders, which results in encoded frame rate being some fraction of the composition frame rate; e.g. using a base/composition frame rate of 60 FPS and setting `skip_frames = 1` on an encoder results in 30 FPS encoded stream being generated, `skip_frames = 2` would result in 20 FPS, `skip_frames = 3` gives 15 FPS etc.

### Motivation and Context
Outputting at different frame rates is potentially beneficial to various use cases, such as e.g. high FPS (60 or even 120 FPS) recording while only streaming at lower FPS to fit within bandwidth constraints (while at the same time reducing encoding load from two separate encoders). Another use case discussed on discord was allowing a lower FPS display for remote control.

Compared to enabling arbitrary frame rates, this change is somewhat smaller in scope, while still allowing experimentation with different frame rates, without having to recalculate timing for the composition thread "on the fly" and similar.

### How Has This Been Tested?
This was tested by augmenting output handlers with `obs_encoder_set_skip_frames` calls, using x264 and NVENC. 

### Types of changes

### Checklist:

### DIFF
--- a/libobs/media-io/video-io.c
+++ b/libobs/media-io/video-io.c
@@ -46,6 +46,16 @@ struct video_input {
 	struct video_frame frame[MAX_CONVERT_BUFFERS];
 	int cur_frame;
 
+	// allow outputting at fractions of main composition FPS,
+	// e.g. 60 FPS with frame_rate_divisor = 1 turns into 30 FPS
+	//
+	// a separate counter is used in favor of using remainder calculations
+	// to allow "inputs" started at the same time to start on the same frame
+	// whereas with remainder calculation the frame alignment would depend on
+	// the total frame count at the time the encoder was started
+	uint32_t frame_rate_divisor;
+	uint32_t frame_rate_divisor_counter;
+
 	void (*callback)(void *param, struct video_data *frame);
 	void *param;
 };
@@ -76,6 +86,8 @@ struct video_output {
 	size_t first_added;
 	size_t last_added;
 	struct cached_frame_info cache[MAX_CACHE_SIZE];
+
+	struct video_output *parent;
 
 	volatile bool raw_active;
 	volatile long gpu_refs;
@@ -135,6 +147,17 @@ static inline bool video_output_cur_frame(struct video_output *video)
 	for (size_t i = 0; i < video->inputs.num; i++) {
 		struct video_input *input = video->inputs.array + i;
 		struct video_data frame = frame_info->frame;
+
+		// an explicit counter is used instead of remainder calculation
+		// to allow multiple encoders started at the same time to start on
+		// the same frame
+		uint32_t skip = input->frame_rate_divisor_counter++;
+		if (input->frame_rate_divisor_counter ==
+		    input->frame_rate_divisor)
+			input->frame_rate_divisor_counter = 0;
+
+		if (skip)
+			continue;
 
 		if (scale_video_output(input, &frame))
 			input->callback(input->param, &frame);
@@ -371,13 +394,37 @@ static inline void reset_frames(video_t *video)
 	os_atomic_set_long(&video->total_frames, 0);
 }
 
+static const video_t *get_const_root(const video_t *video)
+{
+	while (video->parent)
+		video = video->parent;
+	return video;
+}
+
+static video_t *get_root(video_t *video)
+{
+	while (video->parent)
+		video = video->parent;
+	return video;
+}
+
 bool video_output_connect(
 	video_t *video, const struct video_scale_info *conversion,
 	void (*callback)(void *param, struct video_data *frame), void *param)
 {
+	return video_output_connect2(video, conversion, 1, callback, param);
+}
+
+bool video_output_connect2(
+	video_t *video, const struct video_scale_info *conversion,
+	uint32_t frame_rate_divisor,
+	void (*callback)(void *param, struct video_data *frame), void *param)
+{
 	bool success = false;
 
-	if (!video || !callback)
+	video = get_root(video);
+
+	if (!video || !callback || frame_rate_divisor == 0)
 		return false;
 
 	pthread_mutex_lock(&video->input_mutex);
@@ -388,6 +435,8 @@ bool video_output_connect(
 
 		input.callback = callback;
 		input.param = param;
+
+		input.frame_rate_divisor = frame_rate_divisor;
 
 		if (conversion) {
 			input.conversion = *conversion;
@@ -446,6 +495,8 @@ void video_output_disconnect(video_t *video,
 	if (!video || !callback)
 		return;
 
+	video = get_root(video);
+
 	pthread_mutex_lock(&video->input_mutex);
 
 	size_t idx = video_get_input_idx(video, callback, param);
@@ -468,7 +519,7 @@ bool video_output_active(const video_t *video)
 {
 	if (!video)
 		return false;
-	return os_atomic_load_bool(&video->raw_active);
+	return os_atomic_load_bool(&get_const_root(video)->raw_active);
 }
 
 const struct video_output_info *video_output_get_info(const video_t *video)
@@ -484,6 +535,8 @@ bool video_output_lock_frame(video_t *video, struct video_frame *frame,
 
 	if (!video)
 		return false;
+
+	video = get_root(video);
 
 	pthread_mutex_lock(&video->data_mutex);
 
@@ -518,6 +571,8 @@ void video_output_unlock_frame(video_t *video)
 	if (!video)
 		return;
 
+	video = get_root(video);
+
 	pthread_mutex_lock(&video->data_mutex);
 
 	video->available_frames--;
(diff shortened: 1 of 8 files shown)


libobs/media-io/video-io.c
```diff
@@ -136,6 +143,13 @@ static inline bool video_output_cur_frame(struct video_output *video)
 		struct video_input *input = video->inputs.array + i;
 		struct video_data frame = frame_info->frame;
 
+		uint32_t skip = input->fps_skipped_frames++;
+		if (input->fps_skipped_frames > input->fps_skip_frames)
+			input->fps_skipped_frames = 0;
```
palana, line 148: @tt2468: as far as `divisor` vs `skip_frames`, I currently find this terminology somewhat more straightforward, but if you have suggestions for better names that fit the divisor theme I'm not married to `skip_frames`

WizardCM: I do worry the term "skipped frames" is one we already use both internally and in the Stats dialog, and would agree to prefer something else.

norihiro: Maybe 'interleave' or 'decimate' would be the alternative word to describe the feature.

palana: `interleave` also feels used up in the codebase, as far as `decimate` vs `divisor` I think I'd prefer `divisor` instead? `decimate` feels somewhat 'irregular'/'random' to me if that makes sense

palana: @WizardCM, @norihiro, @tt2468: I renamed things to the `divisor` naming scheme in [fixup! libobs: Allow configuring skip frames for encoders](https://github.com/obsproject/obs-studio/pull/9001/commits/5d5aaf9343f89e05b778b6d03901a9c79a51149b), thoughts?

ErichDonGubler: Rando hot take: maybe `frame_interval` an interesting synonym?

tt2468: Thinking about this again, I'm pretty confident that `frame_divisor` or `fps_divisor` is the way to go here. It's very commonly used for cases like this, and IMO it's just the most accurate term for the operation we're doing here.

libobs/media-io/video-io.c
```diff
@@ -136,6 +143,14 @@ static inline bool video_output_cur_frame(struct video_output *video)
 		struct video_input *input = video->inputs.array + i;
 		struct video_data frame = frame_info->frame;
 
+		uint32_t skip = input->frame_rate_divisor_counter++;
```
tt2468, line 146: Seems like a better implementation would be to skip `frame_rate_divisor_counter` altogether and just do

```c
if (input->frame_rate_divisor > 1 && (video->total_frames % input->frame_rate_divisor) > 0)
  continue;
```

libobs/media-io/video-io.c
```diff
@@ -610,10 +667,30 @@ void video_output_dec_texture_encoders(video_t *video)
...
 void video_output_inc_texture_skipped_frames(video_t *video)
 {
-	os_atomic_inc_long(&video->skipped_frames);
+	os_atomic_inc_long(&get_root(video)->skipped_frames);
+}
+
+video_t *video_output_create_with_frame_rate_divisor(video_t *video,
+						     uint32_t divisor)
+{
+	if (!video || divisor == 0)
```
tt2468, line 681: I'm not sure it's a good idea to allow creating a dedicated `video_t` instance if the divisor is `1`. May encourage bad behavior if it's plugin-facing.

Other question is do we have to worry about deduplication of video outputs here, or might it be deduplicated elsewhere?

libobs/obs-video-gpu-encode.c
```diff
@@ -94,6 +96,13 @@ static void *gpu_encode_thread(struct obs_core_video_mix *video)
 						     encoder->context.settings);
 			}
 
+			skip = encoder->frame_rate_divisor_counter++;
```
tt2468, line 99: Same remainder calculation trick could be applied here

tt2468 (commented):
Just an idea, feel free to let me know if it already crossed your mind. The current method seems to reach quite deep into the media-io system in order to intercept a deep frame callback and stop it before it's performed.

I'm thinking it may be simpler while reaching the same goals by moving the frame skipping code out of media-io and into the individual implementations:
- obs-encoder.c - Standard encoders
- obs-video-gpu-encode.c - Shared texture encoders
- obs-output.c - Raw outputs (raw outputs could use this too! see decklink output for example)

I'm not seeing any places down the line in the callback chain line from media-io which do any significant computation or copying, so I don't think we'd be sacrificing any notable efficiency.

libobs/media-io/video-io.c
tt2468: Closing since I just saw `frame_rate_divisor` and that seems adequate.

tt2468:
I think I see a conundrum with the idea mentioned above, in that encoders and outputs are currently implemented in a way where they fetch video info from arbitrary, non-standardized places in order to know what they need in order to operate. With the current way OBS works, the above idea would arguably not be feasible without changing encoder code (like is done in the GPU scaling PR).

libobs/media-io/video-io.c
palana: Potentially, though an advantage of the current implementation should be that multiple encoders started at the same time start on the same frame, as opposed to starting frames being offset by ±a few frames

libobs/media-io/video-io.c
palana: Creating instances with `divisor` set to `1` isn't super useful, agreed; I can add this to the check

As far as deduplication, making parent contexts keep track of derives contexts should be straightforward enough; though it'd probably also require some form of refcounting or similar for those contexts, as opposed to the clear ownership semantics that are being used right now

Are you worried about lots of video outputs being created with the same parent and divisor using up additional memory (and creating additional unused(?) threads), or do you have some other use case in mind for deduplicating?

palana:
> I think I see a conundrum with the idea mentioned above, in that encoders and outputs are currently implemented in a way where they fetch video info from arbitrary, non-standardized places in order to know what they need in order to operate. With the current way OBS works, the above idea would arguably not be feasible without changing encoder code (like is done in the GPU scaling PR).

Indeed, that was the reason why I decided to put things where they are right now; i.e. you kind of need to have modified `video_t`s to communicate that information; changing how frame rate and other stuff is communicated is likely more api-breaking than this for now?

libobs/media-io/video-io.c
palana: Added to https://github.com/obsproject/obs-studio/pull/9001/commits/f2d41e4627a416e55a5cba3cd3b0e214a2897f80 to address `divisor == 1` case

libobs/media-io/video-io.c
palana: Added a comment in https://github.com/obsproject/obs-studio/pull/9001/commits/ac815904c8b387dfa0f083c5800716ca5de31fd9

libobs/media-io/video-io.c
```diff
@@ -371,13 +394,37 @@ static inline void reset_frames(video_t *video)
...
+	while (video->parent)
+		video = video->parent;
+	return video;
+}
+
 bool video_output_connect(
 	video_t *video, const struct video_scale_info *conversion,
 	void (*callback)(void *param, struct video_data *frame), void *param)
+{
+	return video_output_connect2(video, conversion, 0, callback, param);
```
Lain-B, line 415: Did you mean to use a divisor of 1? video_output_connect2() seems programmed to fail when you pass 0 to it

libobs/obs-output.c
```diff
@@ -2064,7 +2064,7 @@ static void hook_data_capture(struct obs_output *output)
 		if (has_video)
 			start_raw_video(output->video,
 					obs_output_get_video_conversion(output),
-					default_raw_video_callback, output);
+					0, default_raw_video_callback, output);
```
Lain-B, line 2067: Noting that it uses 0 here as well (also feel free to click resolve button on these if you'd like)

libobs/media-io/video-io.c
palana: You're right, missed updating this when changing from `skip_frames` to `frame_rate_divisor`

palana: fixed in https://github.com/obsproject/obs-studio/pull/9001/commits/0c1c17056952549c68fcbe9726584fe636bf1e06

libobs/obs-output.c
palana: also fixed in https://github.com/obsproject/obs-studio/pull/9001/commits/0c1c17056952549c68fcbe9726584fe636bf1e06

Lain-B (approved):

Lain-B:
Feel free to squash and rebase the changes whenever you're ready, seems like there hasn't been any further feedback. Briefly did some testing with normal settings and everything seemed fine.

palana:
@Lain-B: squashed and rebased; should be ready to merge

PR #9006, opened 2023-06-02
Fix non-common services not being selectable

derrod:
### Description

#8951 was intended to avoid a crash by resetting the service combobox to default if a service is not on the list. What we missed is that if a service is not shown by default, i.e. not has the `common` flag set, it would not be found and the selection reset.

This fixes the actual reason for the crash by adding a NULL check to `find_output`, which results in the behvaiour being as before where a stub service entry is created in the UI until the user selects a different one.

Note that the second commit (c5714e495d8c1c3a215c55b4b880e1ad393b1733) should probably be picked to master as well.

### Motivation and Context

Users want to be able to stream to services that are not listed by default.

### How Has This Been Tested?

Tested with outdated service file originally used for testing #8951 and selected uncommon service to verify it works as intended.

### Types of changes

### Checklist:

### DIFF
--- a/UI/window-basic-settings-stream.cpp
+++ b/UI/window-basic-settings-stream.cpp
@@ -140,10 +140,11 @@ void OBSBasicSettings::LoadStream1Settings()
 			QTStr("Basic.Settings.Stream.Custom.Password.ToolTip"));
 	} else {
 		int idx = ui->service->findText(service);
-		/* 29.1 crash workaround: Fall back to "Custom" if service not found. */
-		if (idx == -1)
-			idx = 0;
-
+		if (idx == -1) {
+			if (service && *service)
+				ui->service->insertItem(1, service);
+			idx = 1;
+		}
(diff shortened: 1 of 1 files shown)


notr1ch:
Are we sure this is the correct place for the fix? This affects how a lot of public APIs handle NULL as `obs_get_output_flags`, `obs_get_output_properties`, etc. all call `find_output`. Passing NULL into these API functions makes no sense to begin with, so IMO the caller of the function has a bug rather than the function itself.

derrod:
I could move the null check to the caller here, or possibly the UI. It seems that we also do not do any checks in similar functions (`get_source_info()`) so nit doing that here would be consistent at least.

Maybe the UI part can be changed to fall back to assuming rtmp as a default.

derrod:
Updated with a different approach that ensures that `OBSBasicSettings::protocol` is set to "RTMP" if the service does not have a protocol set (i.e. returns an empty string).

PR #9019, opened 2023-06-04
libobs: Fix crash when plugins delete properties in callback

ujifgc:
If a plugin removes a property when processing 'modified' property callback, OBS WILL CRASH.

### Description

This commit moves callback processing to after iterating over properties, thus removing the chance to crash OBS.

### Motivation and Context

Example of offending plugin: obs-source-record

Offending code: https://github.com/exeldro/obs-source-record/blob/76da992729ef6d6ed2f691235ca0e086127f2716/source-record.c#L1035-L1055

### How Has This Been Tested?

After applying the change OBS no longer crashes when loading the plugin settings.

### Types of changes

I have changed `obs_properties_apply_settings_internal` to put all properties with callbacks in a temporary array.

After iterating through properties `obs_properties_apply_settings` calls every callback from the temporary array.

The official API of `obs_properties_apply_settings` remains unchanged.

Reference to the change that might have introduced the crash: https://github.com/obsproject/obs-studio/commit/78aa4e03dca62c3e1439cae2eade1c6d247c8a78

### Checklist:

### DIFF
--- a/libobs/obs-properties.c
+++ b/libobs/obs-properties.c
@@ -360,22 +360,21 @@ void obs_properties_remove_by_name(obs_properties_t *props, const char *name)
 	}
 }
 
-void obs_properties_apply_settings_internal(obs_properties_t *props,
-					    obs_data_t *settings,
-					    obs_properties_t *realprops)
+typedef DARRAY(struct obs_property *) obs_property_da_t;
+
+void obs_properties_apply_settings_internal(
+	obs_properties_t *props, obs_property_da_t *properties_with_callback)
 {
 	struct obs_property *p = props->properties;
 
 	while (p) {
 		if (p->type == OBS_PROPERTY_GROUP) {
 			obs_properties_apply_settings_internal(
-				obs_property_group_content(p), settings,
-				realprops);
+				obs_property_group_content(p),
+				properties_with_callback);
 		}
-		if (p->modified)
-			p->modified(realprops, p, settings);
-		else if (p->modified2)
-			p->modified2(p->priv, realprops, p, settings);
+		if (p->modified || p->modified2)
+			da_push_back((*properties_with_callback), &p);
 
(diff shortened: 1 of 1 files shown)


derrod:
This is an interesting one, deleting items from a list while it's being iterated over is generally not a great idea, even before 78aa4e03dca62c3e1439cae2eade1c6d247c8a78 this could result in crashes if the current item were deleted so that `p = p->next` would have read from free'd memory.

One way to restore the previous behaviour would be to manually fix up the `tmp` pointer after the callbacks:
```c
if (tmp != p->hh.next)
    tmp = p->hh.next;
```
But perhaps deferring the callback like this is a better option.

ujifgc:
@derrod I agree that changing the list while iterating through it is a bad thing. But right now the API documentation here https://docs.obsproject.com/reference-properties#c.obs_property_set_modified_callback2 states that modifying the list in the callback is expected. I have not found in the current coddebase where the return value of the callback is checked. I assume the check is just not implemented right now.

PatTheMav:
Given that the properties can be changed while iterating over them, using a "stable" copy for iteration is the right approach. I did wonder if there are unforeseen consequences when a properties' callback is called _after_ a prior callback removed the property entirely, but that's where the crash would happen, so I don't think anybody might have relied on that behaviour.

It _does_ mean that such callbacks will still be executed though, which might be considered a side-effect.

Lain-B (approved):
I'm going to tentatively approve these changes, but one thing I will tell you is that it's probably better to make a property invisible than to delete it. That's what I typically did on all the core obs plugins (v4l2 being the exception, but I didn't write that one). It's not great, and that's my fault due to the way I programmed the properties system, but it's probably the safer than actually deleting a property.

I shouldn't have allowed property removal. But oh well, cat's out of the bag for now. I'll probably just deprecate the removal function or something at some point.

PR #9021, opened 2023-06-04
libobs,plugins: Remove new obs_output_*2 functions

gxalpha:
### Description

Effectively reverting parts of d314d47, this commit removes the new functions that got added to remove the flags parameter. Instead, it just marks the parameter as unused and documents this. Having what is effectively an API break just to remove a parameter is a bit overkill. The other parts of d314d47 which cleaned up the usage of the flags parameter are untouched here.

### Motivation and Context

People brought up that this effective API break is a bit extreme, and I agree.
If we want to merge this, we'll have to do so before the first beta of the next release, so opening this now to get the discussion going.

### How Has This Been Tested?

macOS 13.4
Tested that it still compiles, as well as making sure that recording and streaming still works.

### Types of changes

### Checklist:

### DIFF
--- a/docs/sphinx/reference-outputs.rst
+++ b/docs/sphinx/reference-outputs.rst
@@ -882,32 +882,38 @@ Functions used by outputs
 
 ---------------------
 
-.. function:: bool obs_output_can_begin_data_capture2(const obs_output_t *output)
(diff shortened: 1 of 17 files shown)


notr1ch (approved):
Changes look good, I'm in favor of reverting this API change too,

PR #9025, opened 2023-06-04
libobs: Fix crash when properties are deleted in callback

derrod:
### Description

Fixes the same issue as #9019 but in a minimal way that restores pre-29.1 behaviour.

As mentioned in https://github.com/obsproject/obs-studio/pull/9019#issuecomment-1575590865 this may still be problematic if the property whose callback is being executed is deleted during the callback. For that reason this is only targeting the `release/29.1` branch to leave discussion for an alternative approach open while mitigating possible crashes in 29.1.x.

### Motivation and Context

Fix crash.

### How Has This Been Tested?

Hasn't been.

### Types of changes

### Checklist:

### DIFF
--- a/libobs/obs-properties.c
+++ b/libobs/obs-properties.c
@@ -363,9 +363,9 @@ void obs_properties_apply_settings_internal(obs_properties_t *props,
 					    obs_data_t *settings,
 					    obs_properties_t *realprops)
 {
-	struct obs_property *p, *tmp;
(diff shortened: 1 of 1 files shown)


ujifgc:
Okay, I didn't test exactly this build, but when cherry-picked to master, this commit 5b418429bdda8f6b35d92b782012dc829e6b5a2d fixes the crash with obs-source-record

PR #9028, opened 2023-06-05
obs-ffmpeg: Enable multi-track audio support for mpegts

pkviet:
### Description
This adds support for multi-track audio for mpegts outputs (srt or rist).
Also modifies UI.
For mpegts compatible protocols, this changes radio buttons into checkboxes for the audio track selection in the Streaming output (only in Advanced output).

obs64_2023-06-05_21-27-40

### Motivation and Context
YouTube has added support for SRT ingest and allowed broadcast engineers to test it during the 2023 SRT Plugfest.
I was told during the zoom introduction that multi-track support is being worked on by YouTube (public discussion,
there was no NDA).
I had this code around for a while, so it seems it's the right time to add it to obs.
Maybe multi-track will be supported also by RTMP (who knows) and the  code would use the UI parts of this PR.

### How Has This Been Tested?
Tested against several services during 2023 SRT Plugfest.
Works as expected. One can play any track.

### Types of changes

### Checklist:

### DIFF
--- a/UI/window-basic-main-outputs.cpp
+++ b/UI/window-basic-main-outputs.cpp
@@ -19,6 +19,8 @@ volatile bool virtualcam_active = false;
 
 #define FTL_PROTOCOL "ftl"
 #define RTMP_PROTOCOL "rtmp"
+#define SRT_PROTOCOL "srt"
+#define RIST_PROTOCOL "rist"
 
 static void OBSStreamStarting(void *data, calldata_t *params)
 {
@@ -1473,6 +1475,7 @@ bool SimpleOutput::ReplayBufferActive() const
 struct AdvancedOutput : BasicOutputHandler {
 	OBSEncoder streamAudioEnc;
 	OBSEncoder streamArchiveEnc;
+	OBSEncoder streamTrack[MAX_AUDIO_MIXES];
 	OBSEncoder recordTrack[MAX_AUDIO_MIXES];
 	OBSEncoder videoStreaming;
 	OBSEncoder videoRecording;
@@ -1508,6 +1511,7 @@ struct AdvancedOutput : BasicOutputHandler {
 	virtual bool StreamingActive() const override;
 	virtual bool RecordingActive() const override;
 	virtual bool ReplayBufferActive() const override;
+	bool allowsMultiTrack();
 };
 
 static OBSData GetDataFromJsonFile(const char *jsonFile)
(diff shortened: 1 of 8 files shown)


GeorgesStavracas:
We've reviewed this and the question that was raised is if this change prevents unselecting all audio tracks. Recently OBS started enforcing that at least one audio track is selected for recordings, and this behaviour should be reflected here too.

pkviet:
> We've reviewed this and the question that was raised is if this change prevents unselecting all audio tracks. Recently OBS started enforcing that at least one audio track is selected for recordings, and this behaviour should be reflected here too.

@GeorgesStavracas i checked current behaviour with recording and there's a warning indeed when no track is selected. I see now what you meant. I've added that too in case some clueless user uncheck all audio tracks.

UI/window-basic-main-outputs.cpp
```diff
@@ -2151,7 +2212,12 @@ bool AdvancedOutput::StartStreaming(obs_service_t *service)
 #endif
 	bool enableDynBitrate =
 		config_get_bool(main->Config(), "Output", "DynamicBitrate");
-
+	bool is_mpegts = false;
+	obs_service_t *service_obj = main->GetService();
+	const char *url = obs_service_get_connect_info(
+		service_obj, OBS_SERVICE_CONNECT_INFO_SERVER_URL);
+	if (url != NULL)
+		is_mpegts = isMpegts(url);
```
tytan652, line 2220: You could just check the protocol that the service return and not the server URL.

tytan652 (commented):
I think it should not be coded as a MPEG TS only feature when, it comes to naming since in the future another container/protocol might do the same thing.

norihiro:
The commit 73dd9f82b has a too long line, 80 characters, where it's limit is 72.
> This enables audio multi-track support in UI for mpegts streams (srt, rist ...).

pkviet:
> I think it should not be coded as a MPEG TS only feature when it comes to naming, since in the future another container/protocol might do the same thing.

yeah, I had already thought about it. This is code I wrote a long while ago. But given there's a new rtmp spec which is still being worked on, I did plan on updating the code to be protocol neutral. I wanted the PR out sooner than later due to SRT PlugFest.

UI/window-basic-main-outputs.cpp
pkviet: ah yeah good point. I haven't adjusted yet to the new api

pkviet:
i'm gonna also use dedicated stream tracks instead of the recording tracks since some people might wan tto change encoders for streaming and recording.

UI/window-basic-main-outputs.cpp
pkviet: done

pkviet:
**Update**
- switched to neutral UI changes to accommodate other protocols with multi-track capability (ex: whip).
- removed the recourse to set_mixers function reserved to raw audio.
- I'll later add support for language tags

UI/window-basic-main-outputs.cpp
```diff
@@ -2060,7 +2114,6 @@ inline void AdvancedOutput::SetupVodTrack(obs_service_t *service)
 		if (!ServiceSupportsVodTrack(service))
 			vodTrackEnabled = false;
 	}
-
 	if (vodTrackEnabled && streamTrack != vodTrackIndex)
```
krakow10, line 2117: `int streamTrack` in this function is overshadowing `OBSEncoder streamTrack` used in the PR

pkviet: Thanks for catching this. `int streamTrack` is named int streamTrackIndex in other places; so I've renamed it accordingly in the two places where it's used.

krakow10:
I've been using this patch in one form or another since July 2022.  I can confirm that it continues to work on OBS v30.

AgainPsychoX:
When can we expect merge? Many streamers want to use this, but they are not technical enough to trust anything other than official build :C

pkviet:
> I've been using this patch in one form or another since July 2022.  I can confirm that it continues to work on OBS v30.

Well, I first have to address the comments, which I didn't because ... life.
I hope it can be merged in next point release. We'll see

t3pmd:
Would love to see this in the next version please. would help alot

pkviet:
Was so long ago, I had forgotten that I had actually addressed the comments. Will request that it moves forward.

pkviet:
Rebased to current master.

tt2468 (approved):

Warchamp7 (approved):

Lain-B (approved):

PR #9031, opened 2023-06-06
UI: Add YouTube Control Panel dock (based on CEF and studio.youtube.com)

yuriy-chumak:
### Description
Integration of YouTube Control Panel in the OBS (like YT chat).

image

### Motivation and Context
This new YT dock panel makes integration with YT  more responsive and gives the creator more control over the broadcast session. It also allows the YT side to faster deliver own changes to creators (including new features and possible bugfixes).

### Build instructions
Needs obs-browser plugin [PR410](https://github.com/obsproject/obs-browser/pull/410) code changes.

### How Has This Been Tested?
Tested manually.

* Windows 10 x64, Linux Ubuntu 22.04, macOS M1 Ventura, macOS x86_64 Catalina.

### Types of changes

### Checklist:

### DIFF
--- a/UI/cmake/feature-youtube.cmake
+++ b/UI/cmake/feature-youtube.cmake
@@ -2,8 +2,16 @@ if(YOUTUBE_CLIENTID
    AND YOUTUBE_SECRET
    AND YOUTUBE_CLIENTID_HASH MATCHES "(0|[a-fA-F0-9]+)"
    AND YOUTUBE_SECRET_HASH MATCHES "(0|[a-fA-F0-9]+)")
-  target_sources(obs-studio PRIVATE auth-youtube.cpp auth-youtube.hpp window-youtube-actions.cpp
-                                    window-youtube-actions.hpp youtube-api-wrappers.cpp youtube-api-wrappers.hpp)
+  target_sources(
+    obs-studio
+    PRIVATE auth-youtube.cpp
+            auth-youtube.hpp
+            window-dock-youtube-app.cpp
+            window-dock-youtube-app.hpp
+            window-youtube-actions.cpp
+            window-youtube-actions.hpp
+            youtube-api-wrappers.cpp
+            youtube-api-wrappers.hpp)
 
   target_enable_feature(obs-studio "YouTube API connection" YOUTUBE_ENABLED)
 else()
--- a/UI/cmake/legacy.cmake
+++ b/UI/cmake/legacy.cmake
@@ -330,8 +330,16 @@ endif()
 
 if(YOUTUBE_ENABLED)
   target_compile_definitions(obs PRIVATE YOUTUBE_ENABLED)
-  target_sources(obs PRIVATE auth-youtube.cpp auth-youtube.hpp youtube-api-wrappers.cpp youtube-api-wrappers.hpp
-                             window-youtube-actions.cpp window-youtube-actions.hpp)
+  target_sources(
+    obs
+    PRIVATE auth-youtube.cpp
+            auth-youtube.hpp
+            window-dock-youtube-app.cpp
+            window-dock-youtube-app.hpp
+            window-youtube-actions.cpp
+            window-youtube-actions.hpp
+            youtube-api-wrappers.cpp
+            youtube-api-wrappers.hpp)
 endif()
 
 if(OS_WINDOWS)
--- a/UI/data/locale/en-US.ini
+++ b/UI/data/locale/en-US.ini
@@ -1526,3 +1526,6 @@ YouTube.Errors.liveChatDisabled="Live chat is disabled on this stream."
 YouTube.Errors.liveChatEnded="Live stream has ended."
 YouTube.Errors.messageTextInvalid="The message text is not valid."
 YouTube.Errors.rateLimitExceeded="You are sending messages too quickly."
+# Browser Dock
+YouTube.DocksRemoval.Title="Clear Legacy YouTube Browser Docks"
+YouTube.DocksRemoval.Text="These browser docks will be removed as deprecated:\n\n%1\nUse \"Docks/YouTube Live Control Room\" instead."
--- a/UI/window-basic-auto-config.cpp
+++ b/UI/window-basic-auto-config.cpp
@@ -1050,6 +1050,13 @@ void AutoConfig::done(int result)
 		if (type == Type::Streaming)
 			SaveStreamSettings();
 		SaveSettings();
+
+#ifdef YOUTUBE_ENABLED
+		if (YouTubeAppDock::IsYTServiceSelected()) {
+			OBSBasic *main = OBSBasic::Get();
+			main->NewYouTubeAppDock();
+		}
+#endif
 	}
 }
 
--- a/UI/window-basic-main.cpp
+++ b/UI/window-basic-main.cpp
@@ -2170,6 +2170,9 @@ void OBSBasic::OBSInit()
 
 	/* ----------------------------- */
 	/* add custom browser docks      */
+#if defined(BROWSER_AVAILABLE) && defined(YOUTUBE_ENABLED)
+	YouTubeAppDock::CleanupYouTubeUrls();
+#endif
 
 #ifdef BROWSER_AVAILABLE
 	if (cef) {
@@ -2330,6 +2333,12 @@ void OBSBasic::OBSInit()
 
 	UpdatePreviewProgramIndicators();
 	OnFirstLoad();
+
+#ifdef YOUTUBE_ENABLED
+	/* setup YouTube app dock */
+	if (YouTubeAppDock::IsYTServiceSelected())
+		youtubeAppDock = new YouTubeAppDock();
+#endif
 
 	if (!hideWindowOnStart)
 		activateWindow();
@@ -6823,9 +6832,10 @@ void OBSBasic::DisplayStreamStartError()
 }
 
 #ifdef YOUTUBE_ENABLED
-void OBSBasic::YouTubeActionDialogOk(const QString &id, const QString &key,
-				     bool autostart, bool autostop,
-				     bool start_now)
+void OBSBasic::YouTubeActionDialogOk(const QString &broadcast_id,
+				     const QString &stream_id,
+				     const QString &key, bool autostart,
+				     bool autostop, bool start_now)
 {
 	//blog(LOG_DEBUG, "Stream key: %s", QT_TO_UTF8(key));
 	obs_service_t *service_obj = GetService();
@@ -6834,8 +6844,11 @@ void OBSBasic::YouTubeActionDialogOk(const QString &id, const QString &key,
 	const std::string a_key = QT_TO_UTF8(key);
 	obs_data_set_string(settings, "key", a_key.c_str());
 
-	const std::string an_id = QT_TO_UTF8(id);
-	obs_data_set_string(settings, "stream_id", an_id.c_str());
+	const std::string b_id = QT_TO_UTF8(broadcast_id);
+	obs_data_set_string(settings, "broadcast_id", b_id.c_str());
+
+	const std::string s_id = QT_TO_UTF8(stream_id);
+	obs_data_set_string(settings, "stream_id", s_id.c_str());
 
 	obs_service_update(service_obj, settings);
 	autoStartBroadcast = autostart;
@@ -7419,6 +7432,11 @@ void OBSBasic::StreamingStart()
 
 	OnActivate();
 
+#ifdef YOUTUBE_ENABLED
+	if (YouTubeAppDock::IsYTServiceSelected())
+		youtubeAppDock->IngestionStarted();
+#endif
+
 	blog(LOG_INFO, STREAMING_START);
 }
 
@@ -7498,6 +7516,11 @@ void OBSBasic::StreamingStop(int code, QString last_error)
 		api->on_event(OBS_FRONTEND_EVENT_STREAMING_STOPPED);
 
 	OnDeactivate();
+
+#ifdef YOUTUBE_ENABLED
+	if (YouTubeAppDock::IsYTServiceSelected())
+		youtubeAppDock->IngestionStopped();
+#endif
 
 	blog(LOG_INFO, STREAMING_STOP);
 
@@ -8461,6 +8484,27 @@ config_t *OBSBasic::Config() const
 {
 	return basicConfig;
 }
+
+#ifdef YOUTUBE_ENABLED
+YouTubeAppDock *OBSBasic::GetYouTubeAppDock()
+{
+	return youtubeAppDock;
+}
+
+void OBSBasic::NewYouTubeAppDock()
+{
+	if (youtubeAppDock)
+		delete youtubeAppDock;
+	youtubeAppDock = new YouTubeAppDock();
+}
+
+void OBSBasic::DeleteYouTubeAppDock()
+{
+	if (youtubeAppDock)
+		delete youtubeAppDock;
+	youtubeAppDock = nullptr;
+}
+#endif
 
 void OBSBasic::UpdateEditMenu()
 {
--- a/UI/window-basic-settings.cpp
+++ b/UI/window-basic-settings.cpp
@@ -48,6 +48,10 @@
 #include "window-basic-main-outputs.hpp"
 #include "window-projector.hpp"
 
+#ifdef YOUTUBE_ENABLED
+#include "youtube-api-wrappers.hpp"
+#endif
+
 #include <util/platform.h>
 #include <util/dstr.hpp>
 #include "ui-config.h"
@@ -4251,6 +4255,22 @@ void OBSBasicSettings::on_buttonBox_clicked(QAbstractButton *button)
 			return;
 
 		SaveSettings();
+
+#ifdef YOUTUBE_ENABLED
+		std::string service = ui->service->currentText().toStdString();
+		if (IsYouTubeService(service)) {
+			if (!main->GetYouTubeAppDock()) {
+				main->NewYouTubeAppDock();
+			}
+			main->GetYouTubeAppDock()->SettingsUpdated(
+				!IsYouTubeService(service) || stream1Changed);
+		} else {
+			if (main->GetYouTubeAppDock()) {
+				main->GetYouTubeAppDock()->AccountDisconnected();
+			}
+			main->DeleteYouTubeAppDock();
+		}
+#endif
 		ClearChanged();
 	}
 
--- a/UI/window-dock-youtube-app.cpp
+++ b/UI/window-dock-youtube-app.cpp
@@ -1,0 +1,482 @@
+#include <QUuid>
+
+#include "window-basic-main.hpp"
+#include "youtube-api-wrappers.hpp"
+#include "window-dock-youtube-app.hpp"
+
+#include "ui-config.h"
+#include "qt-wrappers.hpp"
+
+#include <nlohmann/json.hpp>
+using json = nlohmann::json;
+
+#ifdef YOUTUBE_WEBAPP_PLACEHOLDER
+static constexpr const char *YOUTUBE_WEBAPP_PLACEHOLDER_URL =
+	YOUTUBE_WEBAPP_PLACEHOLDER;
+#else
+static constexpr const char *YOUTUBE_WEBAPP_PLACEHOLDER_URL =
+	"https://studio.youtube.com/live/channel/UC/console?kc=OBS";
+#endif
+
+#ifdef YOUTUBE_WEBAPP_ADDRESS
+static constexpr const char *YOUTUBE_WEBAPP_ADDRESS_URL =
+	YOUTUBE_WEBAPP_ADDRESS;
+#else
+static constexpr const char *YOUTUBE_WEBAPP_ADDRESS_URL =
+	"https://studio.youtube.com/live/channel/%1/console?kc=OBS";
+#endif
+
+static constexpr const char *BROADCAST_CREATED = "BROADCAST_CREATED";
+static constexpr const char *BROADCAST_SELECTED = "BROADCAST_SELECTED";
+static constexpr const char *INGESTION_STARTED = "INGESTION_STARTED";
+static constexpr const char *INGESTION_STOPPED = "INGESTION_STOPPED";
+
+YouTubeAppDock::YouTubeAppDock()
+	: BrowserDock(),
+	  dockBrowser(nullptr),
+	  cookieManager(nullptr)
+{
+	OBSBasic::InitBrowserPanelSafeBlock();
+	AddYouTubeAppDock("YouTube Live Control Panel");
+}
+
+YouTubeAppDock::~YouTubeAppDock()
+{
+	if (cookieManager) {
+		cookieManager->FlushStore();
+		delete cookieManager;
+	}
+}
+
+bool YouTubeAppDock::IsYTServiceSelected()
+{
+	obs_service_t *service_obj = OBSBasic::Get()->GetService();
+	OBSDataAutoRelease settings = obs_service_get_settings(service_obj);
+	const char *service = obs_data_get_string(settings, "service");
+	return IsYouTubeService(service);
+}
+
+void YouTubeAppDock::AccountConnected()
+{
+	channelId.clear(); // renew channel id
+	UpdateChannelId();
+}
+
+void YouTubeAppDock::AccountDisconnected()
+{
+	SettingsUpdated(true);
+}
+
+void YouTubeAppDock::SettingsUpdated(bool cleanup)
+{
+	bool ytservice = IsYTServiceSelected();
+	SetVisibleYTAppDockInMenu(ytservice);
+
+	// definitely cleanup if YT switched off
+	if (!ytservice || cleanup) {
+		if (cookieManager)
+			cookieManager->DeleteCookies("", "");
+	}
+	if (ytservice)
+		Update();
+}
+
+std::string YouTubeAppDock::InitYTUserUrl()
+{
+	std::string user_url(YOUTUBE_WEBAPP_PLACEHOLDER_URL);
+
+	if (IsUserSignedIntoYT()) {
+		YoutubeApiWrappers *apiYouTube = GetYTApi();
+		if (apiYouTube) {
+			ChannelDescription channel_description;
+			if (apiYouTube->GetChannelDescription(
+				    channel_description)) {
+				QString url =
+					QString(YOUTUBE_WEBAPP_ADDRESS_URL)
+						.arg(channel_description.id);
+				user_url = url.toStdString();
+			} else {
+				blog(LOG_ERROR,
+				     "YT: InitYTUserUrl() Failed to get channel id");
+			}
+		}
+	} else {
+		blog(LOG_ERROR, "YT: InitYTUserUrl() User is not signed");
+	}
+
+	blog(LOG_DEBUG, "YT: InitYTUserUrl() User url: %s", user_url.c_str());
+	return user_url;
+}
+
+void YouTubeAppDock::AddYouTubeAppDock(const QString &title)
+{
+	QString bId(QUuid::createUuid().toString());
+	bId.replace(QRegularExpression("[{}-]"), "");
+	this->setProperty("uuid", bId);
+	this->setObjectName(title + "Object");
+	this->resize(580, 500);
+	this->setMinimumSize(400, 300);
+	this->setWindowTitle(title);
+	this->setAllowedAreas(Qt::AllDockWidgetAreas);
+
+	OBSBasic::Get()->addDockWidget(Qt::RightDockWidgetArea, this);
+	actionYTAppDock = OBSBasic::Get()->AddDockWidget(this);
+
+	if (IsYTServiceSelected()) {
+		const std::string url = InitYTUserUrl();
+		CreateBrowserWidget(url);
+
+		// reload panel layout
+		const char *dockStateStr = config_get_string(
+			App()->GlobalConfig(), "BasicWindow", "DockState");
+		if (dockStateStr) {
+			QByteArray dockState = QByteArray::fromBase64(
+				QByteArray(dockStateStr));
+			OBSBasic::Get()->restoreState(dockState);
+		}
+	} else {
+		this->setVisible(false);
+		actionYTAppDock->setVisible(false);
+	}
+}
+
+void YouTubeAppDock::CreateBrowserWidget(const std::string &url)
+{
+	std::string dir_name = std::string("obs_profile_cookies_youtube/") +
+			       config_get_string(OBSBasic::Get()->Config(),
+						 "Panels", "CookieId");
+	if (cookieManager)
+		delete cookieManager;
+	cookieManager = cef->create_cookie_manager(dir_name, true);
+
+	if (dockBrowser)
+		delete dockBrowser;
+	dockBrowser = cef->create_widget(this, url, cookieManager);
+	if (!dockBrowser)
+		return;
+
+	if (obs_browser_qcef_version() >= 1)
+		dockBrowser->allowAllPopups(true);
+
+	this->SetWidget(dockBrowser);
+	Update();
+}
+
+void YouTubeAppDock::SetVisibleYTAppDockInMenu(bool visible)
+{
+	if (!actionYTAppDock)
+		return;
+
+	actionYTAppDock->setVisible(visible);
+	this->setVisible(visible);
+}
+
+// only 'ACCOUNT' mode supported
+void YouTubeAppDock::BroadcastCreated(const char *stream_id)
+{
+	DispatchYTEvent(BROADCAST_CREATED, stream_id, YTSM_ACCOUNT);
+}
+
+// only 'ACCOUNT' mode supported
+void YouTubeAppDock::BroadcastSelected(const char *stream_id)
+{
+	DispatchYTEvent(BROADCAST_SELECTED, stream_id, YTSM_ACCOUNT);
+}
+
+// both 'ACCOUNT' and 'STREAM_KEY' modes supported
+void YouTubeAppDock::IngestionStarted()
+{
+	obs_service_t *service_obj = OBSBasic::Get()->GetService();
+	OBSDataAutoRelease settings = obs_service_get_settings(service_obj);
+	const char *service = obs_data_get_string(settings, "service");
+	if (IsYouTubeService(service)) {
+		if (IsUserSignedIntoYT()) {
+			const char *broadcast_id =
+				obs_data_get_string(settings, "broadcast_id");
+			this->IngestionStarted(broadcast_id,
+					       YouTubeAppDock::YTSM_ACCOUNT);
+		} else {
+			const char *stream_key =
+				obs_data_get_string(settings, "key");
+			this->IngestionStarted(stream_key,
+					       YouTubeAppDock::YTSM_STREAM_KEY);
+		}
+	}
+}
+
+void YouTubeAppDock::IngestionStarted(const char *stream_id,
+				      streaming_mode_t mode)
+{
+	DispatchYTEvent(INGESTION_STARTED, stream_id, mode);
+}
+
+// both 'ACCOUNT' and 'STREAM_KEY' modes supported
+void YouTubeAppDock::IngestionStopped()
+{
+	obs_service_t *service_obj = OBSBasic::Get()->GetService();
+	OBSDataAutoRelease settings = obs_service_get_settings(service_obj);
+	const char *service = obs_data_get_string(settings, "service");
+
+	if (IsYouTubeService(service)) {
+		if (IsUserSignedIntoYT()) {
+			const char *broadcast_id =
+				obs_data_get_string(settings, "broadcast_id");
+			this->IngestionStopped(broadcast_id,
+					       YouTubeAppDock::YTSM_ACCOUNT);
+		} else {
+			const char *stream_key =
+				obs_data_get_string(settings, "key");
+			this->IngestionStopped(stream_key,
+					       YouTubeAppDock::YTSM_STREAM_KEY);
+		}
+	}
+}
+
+void YouTubeAppDock::IngestionStopped(const char *stream_id,
+				      streaming_mode_t mode)
+{
+	DispatchYTEvent(INGESTION_STOPPED, stream_id, mode);
+}
+
+void YouTubeAppDock::showEvent(QShowEvent *)
+{
+	if (!dockBrowser)
+		Update();
+}
+
+void YouTubeAppDock::closeEvent(QCloseEvent *event)
+{
+	BrowserDock::closeEvent(event);
+	this->SetWidget(nullptr);
+}
+
+void YouTubeAppDock::DispatchYTEvent(const char *event, const char *video_id,
+				     streaming_mode_t mode)
+{
+	if (!dockBrowser)
+		return;
+
+	// update channelId if empty:
+	UpdateChannelId();
+
+	// notify YouTube Live Streaming API:
+	std::string script;
+	if (mode == YTSM_ACCOUNT) {
+		script = QString(R"""(
+			if (window.location.hostname == 'studio.youtube.com') {
+				let event = {
+					type: '%1',
+					channelId: '%2',
+					videoId: '%3',
+				};
+				console.log(event);
+				if (window.ytlsapi && window.ytlsapi.dispatchEvent)
+					window.ytlsapi.dispatchEvent(event);
+			}
+		)""")
+				 .arg(event)
+				 .arg(channelId)
+				 .arg(video_id)
+				 .toStdString();
+	} else {
+		const char *stream_key = video_id;
+		script = QString(R"""(
+			if (window.location.hostname == 'studio.youtube.com') {
+				let event = {
+					type: '%1',
+					streamKey: '%2',
+				};
+				console.log(event);
+				if (window.ytlsapi && window.ytlsapi.dispatchEvent)
+					window.ytlsapi.dispatchEvent(event);
+			}
+		)""")
+				 .arg(event)
+				 .arg(stream_key)
+				 .toStdString();
+	}
+	dockBrowser->executeJavaScript(script);
+
+	// in case of user still not logged in in dock panel, remember last event
+	SetInitEvent(mode, event, video_id, channelId.toStdString().c_str());
+}
+
+void YouTubeAppDock::Update()
+{
+	std::string url = InitYTUserUrl();
+
+	if (!dockBrowser) {
+		CreateBrowserWidget(url);
+	} else {
+		dockBrowser->setURL(url);
+	}
+
+	// if streaming already run, let's notify YT about past event
+	if (OBSBasic::Get()->StreamingActive()) {
+		obs_service_t *service_obj = OBSBasic::Get()->GetService();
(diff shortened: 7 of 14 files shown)


Juddex:
@yuriy-chumak does the live control panel dock also have the red "end stream" button that u have to manually click to end the stream on YouTube server side ?

yuriy-chumak:
> does the live control panel dock also have the red "end stream" button

I haven't seen such button. But I know that the content of dock panel (which is completely YT-side) may change at any time - and that's the one of the reason to make these changes.

UI/window-dock-youtube-app.cpp
```diff
@@ -0,0 +1,386 @@
...
+}
+
+void YouTubeAppDock::LoggedOut()
+{
+	// LOGGED_OUT;
+}
+
+void YouTubeAppDock::showEvent(QShowEvent *event)
+{
+	(void)event;
```
RytoEX, line 202: I think you can drop `event` from the function definition here and it would do the same thing.

UI/window-dock-youtube-app.cpp
```diff
@@ -0,0 +1,386 @@
+#include <sstream>
```
RytoEX, line 1: Is this include actually used?

RytoEX (changes requested):
Please address the following:
1. use a more descriptive name than "kiosk" panel in the commit message
2. do not end a commit message subject (the first line of a commit message) with punctuation
3. use the commit message prefix "UI: "
4. either split the CI changes to a second commit, or add a note at the end of the current commit's commit message that "this commit also modifies CI files"
5. use a descriptive PR title that is like a commit message subject

yuriy-chumak:
>     1. use a more descriptive name than "kiosk" panel in the commit message

Done.

>     2. do not end a commit message subject (the first line of a commit message) with punctuation

Done.

>     3. use the commit message prefix "UI: "

Done.

>     4. either split the CI changes to a second commit, or add a note at the end of the current commit's commit message that "this commit also modifies CI files"

Extended commit message with "this commit also modifies CI files".

>     5. use a descriptive PR title that is like a commit message subject

Done.

UI/data/locale/en-US.ini
```diff
@@ -1510,3 +1510,6 @@ YouTube.Errors.liveChatDisabled="Live chat is disabled on this stream."
 YouTube.Errors.liveChatEnded="Live stream has ended."
 YouTube.Errors.messageTextInvalid="The message text is not valid."
 YouTube.Errors.rateLimitExceeded="You are sending messages too quickly."
+# Browser Dock
+YouTube.DocksRemoval.Caption="Clear legacy YouTube browser docks"
```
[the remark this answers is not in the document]
yuriy-chumak: Done.

UI/window-basic-auto-config.cpp
```diff
@@ -504,6 +504,14 @@ void AutoConfigStreamPage::on_connectAccount_clicked()
 	auth = OAuthStreamKey::Login(this, service);
 	if (!!auth) {
 		OnAuthConnected();
+#if YOUTUBE_ENABLED && YT_WEBAPP_ENABLED
+		if (IsYouTubeService(service)) {
+			OBSBasic *main = OBSBasic::Get();
+			if (!main->GetYouTubeAppDock())
+				main->NewYouTubeAppDock();
+			main->GetYouTubeAppDock()->AccountConnected();
```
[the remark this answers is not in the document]
yuriy-chumak: These code block moved. Done for this code and for similar lines.

UI/window-dock-youtube-app.cpp
```diff
@@ -0,0 +1,386 @@
...
+}
+
+void YouTubeAppDock::DispatchYTEvent(const char *event, const char *video_id,
+				     streaming_mode_t mode)
+{
+	if (!dockBrowser)
+		return;
+	// update channelId if empty:
+	UpdateChannelId();
+	// notify Kiosk API:
```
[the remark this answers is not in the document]
yuriy-chumak: Thank you. Changed.

UI/window-dock-youtube-app.cpp
```diff
@@ -0,0 +1,386 @@
...
+	if (auth) {
+		YoutubeApiWrappers *apiYouTube(
+			dynamic_cast<YoutubeApiWrappers *>(auth));
+		if (apiYouTube) {
+			return apiYouTube;
+		} else
+			blog(LOG_ERROR,
+			     "YT: GetYTApi() Failed to get YoutubeApiWrappers");
+	} else
+		blog(LOG_ERROR, "YT: GetYTApi() Failed to get Auth");
```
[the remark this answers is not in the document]
yuriy-chumak: Done for this code block and similar.

UI/window-dock-youtube-app.cpp
```diff
@@ -0,0 +1,386 @@
...
+			}
+		}
+	}
+}
+
+void YouTubeAppDock::SetInitEvent(streaming_mode_t mode, const char *event,
+				  const char *video_id, const char *channelId)
+{
+	const std::string version = App()->GetVersionString();
+	blog(LOG_DEBUG, "YT: obs version: %s", version.c_str());
```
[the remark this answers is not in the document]
yuriy-chumak: We use this version inside cef session and this is the way we debug the code. This log inside "LOG_DEBUG" and should not affect release code.

UI/window-dock-youtube-app.cpp
```diff
@@ -0,0 +1,386 @@
...
+						type: '%1',
+						channelId: '%2',
+						videoId: '%3',
+					}
+			)""")
+					    .arg(event)
+					    .arg(channelId)
+					    .arg(video_id);
+		} else {
+			api_event = QString(R"""(,
```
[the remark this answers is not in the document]
yuriy-chumak: `api_event` is a part of javascript object inside `script` variable, and is a last element of this object. So, comma is a part of js object structure. We need it.

UI/window-dock-youtube-app.cpp
yuriy-chumak: Done.

UI/window-dock-youtube-app.cpp
yuriy-chumak: Removed, thank you.

UI/window-dock-youtube-app.cpp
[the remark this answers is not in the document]
RytoEX: Why is it needed inside the CEF session? Why is it needed for debugging? The Version of OBS is already emitted elsewhere. What is the benefit of adding it here?

UI/cmake/feature-youtube.cmake
```diff
@@ -11,3 +11,10 @@ else()
   set(YOUTUBE_SECRET_HASH 0)
   set(YOUTUBE_CLIENTID_HASH 0)
 endif()
+
+if(YT_WEBAPP_ADDRESS)
+  target_sources(obs-studio PRIVATE window-dock-youtube-app.cpp window-dock-youtube-app.hpp)
+  target_enable_feature(obs-studio "YouTube WebApp Dock" YT_WEBAPP_ENABLED)
+else()
+  target_disable_feature(obs-studio "YouTube WebApp Dock")
+endif()
```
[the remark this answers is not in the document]
yuriy-chumak: Thank you for this comment.
I rechecked YOUTUBE_ENABLED and YT_WEBAPP_ENABLED (renamed to the YOUTUBE_WEBAPP_ENABLED), added dependency in the cmake files (new one and legacy).
Changed double "#if YOUTUBE_ENABLED && YOUTUBE_WEBAPP_ENABLED" to the single "#if YOUTUBE_WEBAPP_ENABLED".

UI/window-basic-main.cpp
```diff
@@ -8380,6 +8483,25 @@ config_t *OBSBasic::Config() const
...
+	if (youtubeAppDock)
+		delete youtubeAppDock;
+	youtubeAppDock = new YouTubeAppDock();
+}
+void OBSBasic::DeleteYouTubeAppDock()
+{
+	if (youtubeAppDock)
+		delete youtubeAppDock;
+	youtubeAppDock = nullptr;
+}
```
RytoEX, line 8523: Leave a blank line between function definitions.

UI/window-dock-youtube-app.cpp
```diff
@@ -0,0 +1,386 @@
...
+#ifndef YT_WEBAPP_ADDRESS
+#define YT_WEBAPP_ADDRESS \
+	"https://studio.youtube.com/live/channel/%1/console?kc=OBS"
+#endif
+
+#define BROADCAST_CREATED "BROADCAST_CREATED"
+#define BROADCAST_SELECTED "BROADCAST_SELECTED"
+#define INGESTION_STARTED "INGESTION_STARTED"
+#define INGESTION_STOPPED "INGESTION_STOPPED"
+#define LOGGED_OUT "LOGGED_OUT"
```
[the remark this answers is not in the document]
yuriy-chumak: These constants made as "#define"s to be consisted with other youtube-related code. Like `youtube-api-wrappers.cpp`, for example. I think we can keep this as defines?

UI/window-dock-youtube-app.cpp
```diff
@@ -0,0 +1,383 @@
...
+					QString(YOUTUBE_WEBAPP_ADDRESS)
+						.arg(channel_description.id);
+				user_url = url.toStdString();
+			} else {
+				blog(LOG_ERROR,
+				     "YT: InitYTUserUrl() Failed to get channel id");
+			}
+		}
+	} else
+		blog(LOG_ERROR, "YT: InitYTUserUrl() User is not signed");
```
RytoEX, line 92: If you enclose one branch of an `if` statement, then all branches must be enclosed.

UI/window-dock-youtube-app.cpp
```diff
@@ -0,0 +1,383 @@
...
+	BrowserDock::closeEvent(event);
+	this->SetWidget(nullptr);
+}
+
+void YouTubeAppDock::DispatchYTEvent(const char *event, const char *video_id,
+				     streaming_mode_t mode)
+{
+	if (!dockBrowser)
+		return;
+	// update channelId if empty:
```
RytoEX, line 214: Prefer a blank line after an unenclosed `if` statement body.

UI/cmake/legacy.cmake
```diff
@@ -63,6 +63,12 @@ else()
   set(YOUTUBE_ENABLED ON)
 endif()
 
+if(YT_WEBAPP_ADDRESS)
+  set(YT_WEBAPP_ENABLED ON)
+else()
+  set(YT_WEBAPP_ENABLED OFF)
+endif()
```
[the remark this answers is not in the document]
yuriy-chumak: Done.

CI/windows/02_build_obs.ps1
```diff
@@ -110,6 +111,7 @@ function Configure-OBS {
             "-DYOUTUBE_CLIENTID_HASH=`"${Env:YOUTUBE_CLIENTID_HASH}`"",
             "-DYOUTUBE_SECRET=`"${Env:YOUTUBE_SECRET}`"",
             "-DYOUTUBE_SECRET_HASH=`"${Env:YOUTUBE_SECRET_HASH}`"",
+            "-DYT_WEBAPP_ADDRESS=${Env:YT_WEBAPP_ADDRESS}",
```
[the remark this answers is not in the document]
yuriy-chumak: Done. Thank you.

UI/window-dock-youtube-app.cpp
```diff
@@ -0,0 +1,386 @@
...
+				     channelId.toStdString().c_str());
+		} else {
+			const char *stream_key =
+				obs_data_get_string(settings, "key");
+			SetInitEvent(YTSM_STREAM_KEY, INGESTION_STARTED,
+				     stream_key);
+		}
+	} else
+		SetInitEvent(IsUserSignedIntoYT() ? YTSM_ACCOUNT
+						  : YTSM_STREAM_KEY);
```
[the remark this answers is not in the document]
yuriy-chumak: Done.

UI/window-dock-youtube-app.cpp
```diff
@@ -0,0 +1,386 @@
...
+}
+
+// both 'ACCOUNT' and "STREAM_KEY" modes supported
+void YouTubeAppDock::IngestionStarted(const char *stream_id,
+				      streaming_mode_t mode)
+{
+	DispatchYTEvent(INGESTION_STARTED, stream_id, mode);
+}
+
+// both 'ACCOUNT' and "STREAM_KEY" modes supported
```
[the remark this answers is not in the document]
yuriy-chumak: Done.

UI/window-dock-youtube-app.cpp
```diff
@@ -0,0 +1,386 @@
...
+	"https://studio.youtube.com/live/channel/%1/console?kc=OBS"
+#endif
+
+#define BROADCAST_CREATED "BROADCAST_CREATED"
+#define BROADCAST_SELECTED "BROADCAST_SELECTED"
+#define INGESTION_STARTED "INGESTION_STARTED"
+#define INGESTION_STOPPED "INGESTION_STOPPED"
+#define LOGGED_OUT "LOGGED_OUT"
+
+#define DEFAULT_RC_PORT "9999"
```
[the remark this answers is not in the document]
yuriy-chumak: Removed. Thank you.

UI/window-dock-youtube-app.cpp
```diff
@@ -0,0 +1,383 @@
...
+		YoutubeApiWrappers *apiYouTube(
+			dynamic_cast<YoutubeApiWrappers *>(auth));
+		if (apiYouTube) {
+			return apiYouTube;
+		} else {
+			blog(LOG_ERROR,
+			     "YT: GetYTApi() Failed to get YoutubeApiWrappers");
+		}
+	} else
+		blog(LOG_ERROR, "YT: GetYTApi() Failed to get Auth");
```
RytoEX, line 381: If you enclose one branch of an `if` statement, then all branches must be enclosed.

UI/window-dock-youtube-app.hpp
```diff
@@ -0,0 +1,51 @@
+#pragma once
+
+#include "window-dock-browser.hpp"
+#include "youtube-api-wrappers.hpp"
+class QCefWidget;
+class OBSBasic;
```
RytoEX, line 6: Is OBSBasic actually required here?

RytoEX (changes requested):

yuriy-chumak:
@RytoEX ,
Thank you very much for your time to review these changes.
Sorry for a bit late - it took some time to recheck changes under linux/windows/macOS.

UI/window-dock-youtube-app.cpp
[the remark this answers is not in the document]
yuriy-chumak: YT api inside CEF session requires obs version as parameter.

[added]
Ah, I see. The first lines of log contain version. Ok, this code removed.

UI/window-basic-main.cpp
yuriy-chumak: Done.

UI/window-dock-youtube-app.cpp
yuriy-chumak: Done.

UI/window-dock-youtube-app.cpp
yuriy-chumak: Done.

UI/cmake/feature-youtube.cmake
```diff
@@ -11,3 +11,10 @@ else()
   set(YOUTUBE_SECRET_HASH 0)
   set(YOUTUBE_CLIENTID_HASH 0)
 endif()
+
+if(YOUTUBE_WEBAPP_ADDRESS && YOUTUBE_ENABLED)
```
RytoEX, line 15: `YOUTUBE_ENABLED` is not a global bool. It is a private target compile definition for obs-studio. It cannot be used this way.

RytoEX (changes requested):

UI/window-dock-youtube-app.cpp
yuriy-chumak: Done.

UI/cmake/feature-youtube.cmake
yuriy-chumak: Hmmmmmmm.
I moved webapp-related block inside youtube-related. Is it ok?

Please wait a minute, I need to check building with and without YOUTUBE-related staff...

UI/window-dock-youtube-app.cpp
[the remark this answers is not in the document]
RytoEX: Whether they are consistent with other existing code is not relevant to the question at hand. We do not assume existing patterns to be "good" patterns.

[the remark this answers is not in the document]
PatTheMav: Correct - if there is a language construct for the purpose, it is to be used and always preferred over macro definitions. We are in the process of cleaning this up in our codebase step by step and expect new code to avoid the same old patterns.

Static strings should always be set up as character array constants and should not take the value of preprocessor definitions (which in this case would make the checks for `YT_WEBAPP_ADDRESS` superfluous as it should be changeable in the first place).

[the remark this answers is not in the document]
yuriy-chumak: Done.

UI/cmake/feature-youtube.cmake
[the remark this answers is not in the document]
yuriy-chumak: 2) Yes. This code uses YT api (via youtube-api-wrappers.cpp) and definitely requires YOUTUBE_ENABLED.

1) I think yes.
  * We need a flag to enable or disable this yt dock functionality. So we need some cmake variable.
  * I'm not sure that this URL remain same in the feature (it changed in the past, it may change in the feature), like YOUTUBE keys.

So, I think, it's good to combine both variables (url and flag) into one, url. And use this url as a flag.

norihiro:
As the contributing guideline describes, the commit title should have a verb in the present tense. Maybe `UI: Add YouTube Control Panel dock` would be fine. I don't think the double quotes (`"`) are necessary. And, I think the duplicated word `dock` can be omitted.
> Make sure commit titles are always in present tense, and are not followed by punctuation.

UI/cmake/feature-youtube.cmake
yuriy-chumak: Retested, done.

yuriy-chumak:
> As the contributing guideline describes, the commit title should have a verb in the present tense. Maybe `UI: Add YouTube Control Panel dock` would be fine. I don't think the double quotes (`"`) are necessary. And, I think the duplicated word `dock` can be omitted.
> 
> > Make sure commit titles are always in present tense, and are not followed by punctuation.

Updated.

UI/cmake/feature-youtube.cmake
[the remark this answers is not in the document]
yuriy-chumak: Done.

UI/window-basic-main.cpp
```diff
@@ -2138,6 +2138,82 @@ void OBSBasic::OBSInit()
...
+			if (!removedYTUrl.isEmpty()) {
+				Json::array save_array;
+
+				const QString msg_title =
+					QTStr("YouTube.DocksRemoval.Title");
+				const QString msg_text =
+					QTStr("YouTube.DocksRemoval.Text")
+						.arg(removedYTUrl);
+				OBSMessageBox::warning(this, msg_title,
+						       msg_text);
```
gxalpha, line 2184: I personally don't like the fact that we're forcing the removal here. Is there a reason for that? Wouldn't it be better to ask the user whether they want to remove them, with an explanation that there are built-in ones?

UI/window-basic-settings.cpp
```diff
@@ -4311,6 +4315,22 @@ void OBSBasicSettings::on_buttonBox_clicked(QAbstractButton *button)
...
+			if (!main->GetYouTubeAppDock()) {
+				main->NewYouTubeAppDock();
+			}
+			main->GetYouTubeAppDock()->SettingsUpdated(
+				!IsYouTubeService(service) || stream1Changed);
+		} else {
+			if (main->GetYouTubeAppDock()) {
+				main->GetYouTubeAppDock()->AccountDisconnected();
+			}
+			main->DeleteYouTubeAppDock();
```
gxalpha, line 4331: Should this really be here?
Would it not be better to have this as part of `YouTubeAuth`, similar to the existing chat dock and the integration docks for all other integrations (`TwitchAuth`, `RestreamAuth`)? I think such a solution would generally make this much more clean and mean less changes in places unrelated to the YouTube workflow authentication overall.

UI/window-basic-main.cpp
```diff
@@ -2138,6 +2138,82 @@ void OBSBasic::OBSInit()
 
 	/* ----------------------------- */
 	/* add custom browser docks      */
+#if BROWSER_AVAILABLE && YOUTUBE_ENABLED
+#define YOUTUBE_VIDEO_URL "://studio.youtube.com/video/"
```
PatTheMav, line 2142: Use `static constexpr` for string constants only used in the current compilation unit.

UI/window-dock-youtube-app.cpp
```diff
@@ -0,0 +1,392 @@
...
+#endif
+
+#ifdef YOUTUBE_WEBAPP_ADDRESS
+constexpr const char *YOUTUBE_WEBAPP_ADDRESS_URL = YOUTUBE_WEBAPP_ADDRESS;
+#else
+constexpr const char *YOUTUBE_WEBAPP_ADDRESS_URL =
+	"https://studio.youtube.com/live/channel/%1/console?kc=OBS";
+#endif
+
+constexpr const char *BROADCAST_CREATED = "BROADCAST_CREATED";
```
PatTheMav, line 24: Use `static constexpr` to ensure the variable is only initialised once and also only available in this compilation unit (it will probably moved into the `TEXT` area of the binary).

Indeed there are only a few situations where a `constexpr` should not be `static`.

UI/window-basic-settings.cpp
```diff
@@ -4311,6 +4315,22 @@ void OBSBasicSettings::on_buttonBox_clicked(QAbstractButton *button)
 			return;
 
 		SaveSettings();
+
+#if YOUTUBE_ENABLED
+		std::string service = QT_TO_UTF8(ui->service->currentText());
```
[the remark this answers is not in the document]
yuriy-chumak: Ok, done.

UI/window-basic-main.cpp
yuriy-chumak: The reason is that it was possible to get this functionality before using a direct URL (https://studio.youtube. ...), but now this is not a valid way. Moreover, such links will produce internal errors, and will increase questions "why it's not working" and, I think, more error situations on the YT backend.

So, the proper solution is to remove such links, with new dock panel introduction.

UI/window-basic-settings.cpp
yuriy-chumak: Hmmmmm. I'll check.

UI/window-basic-main.cpp
yuriy-chumak: Done.

UI/window-dock-youtube-app.cpp
yuriy-chumak: Done.

UI/window-basic-settings.cpp
yuriy-chumak: Unfotunately, yes. Here.
The reason - YouTubeAuth (and YoutubeApiWrappers) used only when user account connected to the YT. But provided code works with stream id without account connectiong. So, we need to include such calls into OBSBasic.

But I moved code above (and similar) into YouTubeAppDock function, and reduced YT related OBSBasic code to two lines per call.

yuriy-chumak:
Comments answered. Code rebased to the latest master.

UI/window-basic-main.cpp
```diff
@@ -2139,6 +2139,9 @@ void OBSBasic::OBSInit()
 
 	/* ----------------------------- */
 	/* add custom browser docks      */
+#if BROWSER_AVAILABLE && YOUTUBE_ENABLED
```
PatTheMav, line 2142: I _think_ this is not correct - at least `YOUTUBE_ENABLED` is only _defined_, but doesn't have an explicit value. So it is more correct to use `#if defined()`, because they don't carry any value (see line 2146 for the correct check).

Applies to the other occurrences of similar checks that just use `#if` for preprocessor definitions that don't carry a value.

UI/window-basic-settings.cpp
PatTheMav: I agree with @gxalpha that this is problematic, is it allows the settings window to directly interfere with the state of the main UI and thus requires the settings window to "know" that such things as YouTube AppDocks exist (it's not a concern of the settings window, so it should not have any concept of this).

IMO this should be handled in the main window when the settings window is closed.

UI/window-dock-youtube-app.cpp
```diff
@@ -0,0 +1,386 @@
...
+void YouTubeAppDock::SetInitEvent(streaming_mode_t mode, const char *event,
+				  const char *video_id, const char *channelId)
+{
+	const std::string version = App()->GetVersionString();
+	blog(LOG_DEBUG, "YT: obs version: %s", version.c_str());
+
+	QString api_event;
+	if (event) {
+		if (mode == YTSM_ACCOUNT) {
+			api_event = QString(R"""(,
```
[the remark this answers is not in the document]
PatTheMav: From what I could tell there will never be a situation in which `api_event` will be empty (it's either handling `YTSM_ACCOUNT` or anything else.

As such the string composition for `script` can assume that the `%4` placeholder will always be replaced by a string, so you can add the comma directly there:

```
console.log("initEvent:", {
 					initClientMode: client_mode,
 					%4 });
```

Adding the comma here is only necessary if `api_event` can be an empty string.

UI/window-basic-main.cpp
yuriy-chumak: YOUTUBE_ENABLED may be defined as "0" (cmake `set(YOUTUBE_ENABLED OFF)`, for example). That's why all "YOUTUBE_ENABLED" under "#if" and not under "defined()". Undefined constants are replaced with "0" after all. So it's ok, for me.

image

RytoEX: `set(YOUTUBE_ENABLED OFF)` does not propagate into C++ #if/#ifdef checks by itself. If `YOUTUBE_ENABLED` has been set to `OFF` in CMake, it is simply never propagated into C++ with `target_compile_definitions`.

norihiro: I suspect automoc generates `#define YOUTUBE_ENABLED 1` so that `#if YOUTUBE_ENABLED` works when the feature is enabled.
```text
$ grep -r YOUTUBE_ENABLED  build/
build/UI/obs_autogen/moc_predefs.h:#define YOUTUBE_ENABLED 1
```
When the feature is not enabled, GCC translate the undefined identifier to number `0`, according to [this document](https://gcc.gnu.org/onlinedocs/cpp/If.html). However, I cannot find how MSVC behaves when encountering the undefined identifier.
So, `#ifdef YOUTUBE_ENABLED` or `#if defined(YOUTUBE_ENABLED)` would be better.

yuriy-chumak: Guys, all similar OBS code uses "#if", not "#ifdef":
RESTREAM_ENABLED - "#if#"s, TWITCH_ENABLED - "#if", YOUTUBE_ENABLED - "#if"s (a lot of).

Do we really need to change only one of the 30 similar lines? The question is not about functionality, just a how code looks.

RytoEX: > Guys, all similar OBS code uses "#if", not "#ifdef": RESTREAM_ENABLED - "#if#"s, TWITCH_ENABLED - "#if", YOUTUBE_ENABLED - "#if"s (a lot of).
> 
> Do we really need to change only one of the 30 similar lines? The question is not about functionality, just a how code looks.

@PatTheMav said in the first post:
> Applies to the other occurrences of similar checks that just use #if for preprocessor definitions that don't carry a value.

As far as I can tell, that means "this feedback applies to all other similar instances in the PR".

`#if` and `#ifdef` are not cosmetically different, they are functionally different. `#if defined` and `#ifdef` are functionally identical.

norihiro: `#if TWITCH_ENABLED` was implemented by a88b4402903. At that time, the macro was defined to `0` or `1` as below.
https://github.com/obsproject/obs-studio/blob/a88b440290372cc6699734c1df89027333436158/UI/ui-config.h.in#L19
Later, a commit 349372b3b39e2346952732ab14122d23c7ec7a57 removes the macro `TWITCH_ENABLED`, `RESTREAM_ENABLED`, and `YOUTUBE_ENABLED`. So, it's incorrect to use `#if TWITCH_ENABLED` now. I think we need to make a separate PR to fix the preprocess.

UI/window-dock-youtube-app.cpp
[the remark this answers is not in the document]
RytoEX: This code is still here. It has not been removed.

> YT api inside CEF session requires obs version as parameter.

Why? For what purpose?

UI/window-basic-main.cpp
RytoEX: @norihiro If there are existing instances outside of the changes in this PR that require fixups, then yes. @PatTheMav is pointing out that the places in this PR that add `#if YOUTUBE_ENABLED` are not correct with the way our current CMake passes the definition to code.

yuriy-chumak: So, should I change `#if BROWSER_AVAILABLE && YOUTUBE_ENABLED` to
1. `#if BROWSER_AVAILABLE && defined(YOUTUBE_ENABLED)`
2. `#if defined(BROWSER_AVAILABLE) && defined(YOUTUBE_ENABLED)`
4. nothing
5. nothing, but it will be new PR with changed all "#if YOUTUBE_ENABLED" to all "#ifdef YOUTUBE_ENABLED"
?

yuriy-chumak:
> (I don't think the parentheses part `(based on CEF and studio.youtube.com)` is necessary for the commit title.)

PR title was not updated after change of commit message. Removed.

UI/window-basic-main.cpp
norihiro: > 2. `#if defined(BROWSER_AVAILABLE) && defined(YOUTUBE_ENABLED)`

This should be the appropriate one.

The merged PR #9242 has already fixed the existing preprocessor directives. You don't need to add a new wrong code.

norihiro:
I'm afraid we have a misunderstanding. I was talking about the commit message of 707ba8f1176e51b4055a626d6461052f66d06126, not the PR title.
> UI: "YouTube Control Panel" dock panel
>
> New dock panel with integrated youtube studio live control room.
> This commit also modifies CI files.

UI/window-basic-main.cpp
RytoEX: Option 2 is the only correct one.

yuriy-chumak: > Option 2 is the only correct one.

Changed.

yuriy-chumak:
> I'm afraid we have a misunderstanding. I was talking about the commit message of [707ba8f](https://github.com/obsproject/obs-studio/commit/707ba8f1176e51b4055a626d6461052f66d06126), not the PR title.

reverted ))

commit message had no

yuriy-chumak:
Rebased.

UI/window-dock-youtube-app.cpp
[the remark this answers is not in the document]
yuriy-chumak: Sorry. Somehow returned back. Removed.
I'm rechecking all comments...

UI/window-dock-youtube-app.cpp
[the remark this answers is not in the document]
yuriy-chumak: > there will never be a situation in which `api_event` will be empty

No, it will be.
We have two use cases:
1. dock panel opened while streaming is active. in this case api_event is set to latest event (INGESTION_STARTED, for example) and YT backend notified about situation (with current broadcast id or stream key). api_event is not empty.  
   in fact we remembering last sent event and sending it with "init" event when dock panel reopened (reloaded).
2. dock panel opened initially, without any other user actions before. in this case no api_event is required and created and no additional information sent to YT backend. api_event is empty.

Lines [322](https://github.com/yuriy-chumak/obs-studio/blob/818c0aedc82dfe57413889d863a6784f88364e96/UI/window-dock-youtube-app.cpp#L322)-344 of window-dock-youtube-app.cpp

UI/window-dock-youtube-app.cpp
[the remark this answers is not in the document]
yuriy-chumak: This is now resolved.

UI/window-basic-main.cpp
```diff
@@ -2138,6 +2138,61 @@ void OBSBasic::OBSInit()
...
+					};
+					save_array.push_back(obj);
+				}
+			}
+			std::string output = Json(save_array).dump();
+			config_set_string(App()->GlobalConfig(), "BasicWindow",
+					  "ExtraBrowserDocks", output.c_str());
+		}
+	}
+#endif
```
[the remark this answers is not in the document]
yuriy-chumak: This is now resolved.

UI/window-dock-youtube-app.cpp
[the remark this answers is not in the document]
yuriy-chumak: This is now resolved.

UI/window-dock-youtube-app.cpp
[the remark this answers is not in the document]
yuriy-chumak: This is now resolved.

UI/cmake/feature-youtube.cmake
yuriy-chumak: This is now resolved.

UI/window-basic-main.cpp
yuriy-chumak: This is now resolved. Isn't it?

UI/window-basic-main.cpp
yuriy-chumak: This is now resolved.

UI/window-basic-main.cpp
yuriy-chumak: This is now resolved.

UI/window-basic-settings.cpp
yuriy-chumak: For this particular case, it looks better to leave the related code in the settings rather than try to move it to the obsmain. The detailed explanation can be found below:

* dock panel should react to "disconnect account" immediately (this is instant action and can't be undone even if settings are not saved at all) because oauth keys are getting invalidated during disconnect. Similar functionality is already present in [on_disconnectAccount_clicked()](https://github.com/obsproject/obs-studio/blob/master/UI/window-basic-settings-stream.cpp#L784) (805 line, `main->SetBroadcastFlowEnabled(false);` call).
  No separate processing for such a case is present in obsmain, so we made it in a similar way (but without making a function because it is called only once).
* we keep reacting to "connect account" inside settings to not split on/off pair ("connect account"/"disconnect account") of functionality into completely different places. additionally, use case "connect"+"instant disconnect"+"connect"+"instant disconnect"+... is possible and we must process it immediately as described above.
* settings already have youtube-related code and talk with obsmain directly, so we do not introduce anything new in these two blocks(for "connect account" and "disconnect account") of code,
* all discussible functionality (6 and 7 lines of the code in two places) can be enclosed into two new obsmain functions and called as a functions not as a code blocks, but it doesn’t look to be a good idea, because these functions will be called only once and only from the settings.
* if we will try to create a standalone handler for "disconnect account" functionality inside obsmain (like save settings, or like create youtube broadcast), it will increase codebase complexity with only one related code consumer.

UI/window-basic-main.cpp
RytoEX: There are still instances in this PR where you are adding `#if YOUTUBE_ENABLED` to C/C++ files. Please use `#if defined(YOUTUBE_ENABLED)` as previously instructed.

UI/window-basic-main.cpp
RytoEX: > I personally don't like the fact that we're forcing the removal here. Is there a reason for that? Wouldn't it be better to ask the user whether they want to remove them, with an explanation that there are built-in ones?

As mentioned above, the old URLs may fail to work moving forward. In order to ensure the the dock functionality works as expected, they are being replaced as a one-time upgrade operation. We provided the acceptable options of "remove the old docks once ever or not at all" (options approved by @Warchamp7 ) in [an earlier review comment](https://github.com/obsproject/obs-studio/pull/9031#discussion_r1222208327), and the option for "remove old docks once ever" was chosen.

As such, I'll resolve this thread.

RytoEX (changes requested):
Please also rebase on the latest obs-studio master branch and fix any merge conflict resolutions.

UI/window-basic-main.cpp
yuriy-chumak: Rechecked all "YOUTUBE_ENABLED" in source changes. Made `#ifdef`s.

UI/cmake/feature-youtube.cmake
```diff
@@ -2,8 +2,16 @@ if(YOUTUBE_CLIENTID
...
+  target_sources(
+    obs-studio
+    PRIVATE auth-youtube.cpp
+            auth-youtube.hpp
+            youtube-api-wrappers.cpp
+            youtube-api-wrappers.hpp
+            window-youtube-actions.cpp
+            window-youtube-actions.hpp
+            window-dock-youtube-app.cpp
+            window-dock-youtube-app.hpp)
```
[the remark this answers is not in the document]
yuriy-chumak: Done.

UI/cmake/legacy.cmake
```diff
@@ -329,8 +329,16 @@ endif()
...
+  target_sources(
+    obs
+    PRIVATE auth-youtube.cpp
+            auth-youtube.hpp
+            youtube-api-wrappers.cpp
+            youtube-api-wrappers.hpp
+            window-youtube-actions.cpp
+            window-youtube-actions.hpp
+            window-dock-youtube-app.cpp
+            window-dock-youtube-app.hpp)
```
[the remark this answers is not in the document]
yuriy-chumak: Done.

UI/window-basic-main.cpp
```diff
@@ -2275,6 +2278,12 @@ void OBSBasic::OBSInit()
 	UpdatePreviewProgramIndicators();
 	OnFirstLoad();
 
+#if YOUTUBE_ENABLED
+	/* setup youtube app dock */
```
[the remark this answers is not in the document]
yuriy-chumak: Changed this and similar comments.

UI/window-basic-auto-config.cpp
```diff
@@ -1047,6 +1047,13 @@ void AutoConfig::done(int result)
 		if (type == Type::Streaming)
 			SaveStreamSettings();
 		SaveSettings();
+
+#if YOUTUBE_ENABLED
```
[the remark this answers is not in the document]
yuriy-chumak: Changed to the `#ifdef`, like all others.
image

yuriy-chumak:
> Please also rebase on the latest obs-studio master branch and fix any merge conflict resolutions.

Code rebased to the latest master. Conflicts solved.
clang-format and cmake-format checked with older scripts (no more such scripts in the "CI" folder?).

PatTheMav:
> > Please also rebase on the latest obs-studio master branch and fix any merge conflict resolutions.
> 
> Code rebased to the latest master. Conflicts solved. clang-format and cmake-format checked with older scripts (no more such scripts in the "CI" folder?).

We updated the scripts to Zsh and moved them into `build-aux`.

Bildschirmfoto 2023-07-21 um 16 48 12

UI/window-dock-youtube-app.cpp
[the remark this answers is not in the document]
RytoEX: I can still see this line/block present in the PR. At the very least, we do not need another line logging the OBS version when the OBS version is already written to the log file at OBS startup.

UI/window-basic-main.cpp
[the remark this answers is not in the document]
RytoEX: I'll resolve this thread since it was implemented as a one-time removal.

UI/cmake/feature-youtube.cmake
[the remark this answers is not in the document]
RytoEX: Alphabetically sorted, these would be:
```
    PRIVATE auth-youtube.cpp
            auth-youtube.hpp
            window-dock-youtube-app.cpp
            window-dock-youtube-app.hpp
            window-youtube-actions.cpp
            window-youtube-actions.hpp
            youtube-api-wrappers.cpp
            youtube-api-wrappers.hpp)
```

UI/cmake/legacy.cmake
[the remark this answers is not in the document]
RytoEX: Alphabetically sorted, these would be:
```
    PRIVATE auth-youtube.cpp
            auth-youtube.hpp
            window-dock-youtube-app.cpp
            window-dock-youtube-app.hpp
            window-youtube-actions.cpp
            window-youtube-actions.hpp
            youtube-api-wrappers.cpp
            youtube-api-wrappers.hpp)
```

UI/window-dock-youtube-app.cpp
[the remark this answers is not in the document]
yuriy-chumak: Removed again.
Sorry, looks like pull and merge issue.

UI/cmake/feature-youtube.cmake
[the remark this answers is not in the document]
yuriy-chumak: Done.

UI/cmake/legacy.cmake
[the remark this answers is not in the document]
yuriy-chumak: Done.

UI/window-dock-youtube-app.cpp
```diff
@@ -0,0 +1,482 @@
...
+		const std::string url = InitYTUserUrl();
+		CreateBrowserWidget(url);
+
+		// reload panel layout
+		const char *dockStateStr = config_get_string(
+			App()->GlobalConfig(), "BasicWindow", "DockState");
+		if (dockStateStr) {
+			QByteArray dockState = QByteArray::fromBase64(
+				QByteArray(dockStateStr));
+			OBSBasic::Get()->restoreState(dockState);
```
tytan652, line 135: This is bad, a dock type should not touch a dockState that is not the one used by the integration.

Sebasee009:
I used the YouTube Control Panel dock and could log in with Google Account but YouTube Chat unfortunately not, must be changed to finally be able to use functions and information is displayed in the chat.

Sebasee009:
I have already connected to YouTube and logged in with the dock, but chat is not logged in with.
image

RytoEX:
That looks normal to me. Our YouTube chat dock is a native implementation. It does not work or look the same as YouTube's web-based chat. You can confirm that you are connected to chat by simply trying to send a message to the chat.

Sebasee009:
So, the chat I can send messages without problems.

RytoEX:
> So, the chat I can send messages without problems.

If your YouTube account is connected, yes, you should be able to send messages without problems.

PR #9034, opened 2023-06-06
obs-ffmpeg: Consider requested throughput in AMF preset fallback

rhutsAMD:
### Description

Currently OBS only queries `...CAP_MAX_THROUGHPUT` property which returns the max throughput VCN can support regardless of any other encoding/decoding sessions running in parallel. 
OBS needs to query another property named `...CAP_REQUESTED_THROUGHPUT` to get the total requested throughput and subtracts it from the max throughput to obtain the available throughput. This PR implements this additional consideration of the already requested throughput in the quality preset fallback logic.

### Motivation and Context

This improves the quality preset level fallback logic to not only consider the maximum possible throughput of the VCN but to also consider the already requested throughput of the VCN.

### How Has This Been Tested?

Tested on Windows 10 with an AMD Radeon RX 7900 XTX.

### Types of changes

### Checklist:

### DIFF
--- a/plugins/obs-ffmpeg/texture-amf.cpp
+++ b/plugins/obs-ffmpeg/texture-amf.cpp
@@ -104,6 +104,7 @@ struct amf_base {
 	AMF_SURFACE_FORMAT amf_format;
 
 	amf_int64 max_throughput = 0;
+	amf_int64 requested_throughput = 0;
 	amf_int64 throughput = 0;
 	int64_t dts_offset = 0;
 	uint32_t cx;
@@ -440,6 +441,8 @@ static inline void refresh_throughput_caps(amf_base *enc, const char *&preset)
 	if (res == AMF_OK) {
 		caps->GetProperty(get_opt_name(CAP_MAX_THROUGHPUT),
 				  &enc->max_throughput);
+		caps->GetProperty(get_opt_name(CAP_REQUESTED_THROUGHPUT),
+				  &enc->requested_throughput);
 	}
 }
 
@@ -451,7 +454,8 @@ static inline void check_preset_compatibility(amf_base *enc,
 	 * throughput, switch to a lower preset. */
 
 	if (astrcmpi(preset, "highQuality") == 0) {
-		if (!enc->max_throughput) {
+		if (enc->max_throughput - enc->requested_throughput <
+		    enc->throughput) {
 			preset = "quality";
 			set_opt(QUALITY_PRESET, get_preset(enc, preset));
 		} else {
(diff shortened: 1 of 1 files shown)


tytan652:
The commit name can't contain two prefix.
Maybe: `obs-ffmpeg: Consider requested throughput in AMF preset fallback `

RytoEX:
Trying to understand how these values relate to each other to make sure I understand these checks. On my system with an RX 6600M, when recording with AVC, `requested_throughput` sometimes seems to be `108000` or `216000` for 720p30 up through 1440p60. `calc_throughput` sets `throughput` to the following values:
| Resolution | FPS | enc->throughput |
| ---------- | --- | --------------- |
| 1280x720   | 30  | 108000          |
| 1280x720   | 60  | 216000          |
| 1920x1080  | 30  | 244800          |
| 1920x1080  | 60  | 489600          |
| 3840x2160  | 30  | 972000          |
| 3840x2160  | 60  | 1944000         |

At 4K30+, `requested_throughput` seems to always be 0.

Should `requested_throughput` and `throughput` match, or are these expected to be different values?

I noticed that this changed the first branch of the check on `highQuality` but the second branch of the check on `quality`, so `quality` still has a check for `if (!enc->max_throughput) {`. Was that intentional? Is that covering a quirk of different presets or hardware?

rhutsAMD:
`enc->throughput` is a calculated amount of throughput that is required for the current encoder configuration including resolution and framerate. When changing the resolution and framerate, a change in the calculated throughput is expected.
 
`enc->requested_throughput` is a queried value that specifies the throughput that was already requested on the VCN. For example, if you have another application using the AMD HW encoder or decoder in the background, then you will see a non-zero value returned. If you had no other HW encoding or decoding sessions while running the 4k30 case, then you may have seen a requested throughput of zero for this reason.
 
Thanks for catching the wrong condition for one of the quality presets. Tagging the PR with the fix: #9087

RytoEX:
> `enc->throughput` is a calculated amount of throughput that is required for the current encoder configuration including resolution and framerate. When changing the resolution and framerate, a change in the calculated throughput is expected.
> 
> `enc->requested_throughput` is a queried value that specifies the throughput that was already requested on the VCN. For example, if you have another application using the AMD HW encoder or decoder in the background, then you will see a non-zero value returned. If you had no other HW encoding or decoding sessions while running the 4k30 case, then you may have seen a requested throughput of zero for this reason.

Thanks for the clarification. I had no other encode sessions running in any case, as far as I know. I will have to double check if I had any Media Sources running with hardware decoding.

PR #9043, opened 2023-06-08
Add GPU based output scaling

palana:
### Description
Allow GPU based scaling of output (textures) using on-demand created `obs_core_video_mix`es

### Motivation and Context
Currently if multiple outputs require different output resolution there's only one resolution that gets converted (and downloaded) on the GPU, which caused people to stream at 720p and upscale 720p to 1080p for recording in the past (not sure if that still happens).

Additionally, this allows encoders that take texture inputs to also be rescaled.

### How Has This Been Tested?
Modified `window-basic-outputs.cpp` to enable GPU scaling for the streaming encoder, set streaming encoder to rescale to 720p with a base output resolution of 1080p and observed `obs-nvenc` (as opposed to `FFmpeg NVENC`) log messages; also checked the resulting output file for proper resolution

### Types of changes

### Checklist:

### DIFF
--- a/libobs/obs-encoder.c
+++ b/libobs/obs-encoder.c
@@ -184,6 +184,116 @@ static inline bool gpu_encode_available(const struct obs_encoder *encoder)
 	       (video->using_p010_tex || video->using_nv12_tex);
 }
 
+/**
+ * GPU based rescaling is currently implemented via core video mixes,
+ * i.e. a core mix with matching width/height/format/colorspace/range
+ * will be created if it doesn't exist already to generate encoder
+ * input
+ */
+static void maybe_set_up_gpu_rescale(struct obs_encoder *encoder)
+{
+	struct obs_core_video_mix *mix = NULL;
+	bool create_mix = true;
+	struct obs_video_info ovi;
+	const struct video_output_info *info;
+
+	if (!encoder->media)
+		return;
+
+	info = video_output_get_info(encoder->media);
+
+	if (encoder->gpu_scale_type == OBS_SCALE_DISABLE)
+		return;
+
+	if (!encoder->scaled_height && !encoder->scaled_width)
+		return;
+
+	pthread_mutex_lock(&obs->video.mixes_mutex);
+	for (size_t i = 0; i < obs->video.mixes.num; i++) {
+		struct obs_core_video_mix *current = obs->video.mixes.array[i];
+		const struct video_output_info *voi =
+			video_output_get_info(current->video);
+		if (current->view != &obs->data.main_view)
+			continue;
+
+		if (voi->width != encoder->scaled_width ||
+		    voi->height != encoder->scaled_height)
+			continue;
+
+		if (voi->format != info->format ||
+		    voi->colorspace != info->colorspace ||
+		    voi->range != info->range)
+			continue;
+
+		current->encoder_refs += 1;
+		obs_encoder_set_video(encoder, current->video);
+		create_mix = false;
+		break;
+	}
+
+	if (!obs->video.main_mix) {
+		create_mix = false;
+	} else {
+		ovi = obs->video.main_mix->ovi;
+	}
+
+	pthread_mutex_unlock(&obs->video.mixes_mutex);
+
+	if (!create_mix)
+		return;
+
+	ovi.output_format = info->format;
+	ovi.colorspace = info->colorspace;
+	ovi.range = info->range;
+
+	ovi.output_height = encoder->scaled_height;
+	ovi.output_width = encoder->scaled_width;
+	ovi.scale_type = encoder->gpu_scale_type;
+
+	ovi.gpu_conversion = true;
+
+	mix = obs_create_video_mix(&ovi);
+	if (!mix)
+		return;
+
+	mix->encoder_only_mix = true;
+	mix->encoder_refs = 1;
+	mix->view = &obs->data.main_view;
+
+	pthread_mutex_lock(&obs->video.mixes_mutex);
+
+	// double check that nobody else added a matching mix while we've created our mix
+	for (size_t i = 0; i < obs->video.mixes.num; i++) {
+		struct obs_core_video_mix *current = obs->video.mixes.array[i];
+		const struct video_output_info *voi =
+			video_output_get_info(current->video);
+		if (current->view != &obs->data.main_view)
+			continue;
+
+		if (voi->width != encoder->scaled_width ||
+		    voi->height != encoder->scaled_height)
+			continue;
+
+		if (voi->format != info->format ||
+		    voi->colorspace != info->colorspace ||
+		    voi->range != info->range)
+			continue;
+
+		obs_encoder_set_video(encoder, current->video);
+		create_mix = false;
+		break;
+	}
+
+	if (!create_mix) {
+		obs_free_video_mix(mix);
+	} else {
+		da_push_back(obs->video.mixes, &mix);
+		obs_encoder_set_video(encoder, mix->video);
+	}
+
+	pthread_mutex_unlock(&obs->video.mixes_mutex);
+}
+
 static void add_connection(struct obs_encoder *encoder)
 {
 	if (encoder->info.type == OBS_ENCODER_AUDIO) {
@@ -459,6 +569,8 @@ static inline bool obs_encoder_initialize_internal(obs_encoder_t *encoder)
 
 	obs_encoder_shutdown(encoder);
 
+	maybe_set_up_gpu_rescale(encoder);
+
 	if (encoder->orig_info.create) {
 		can_reroute = true;
 		encoder->info = encoder->orig_info;
@@ -513,6 +625,31 @@ bool obs_encoder_initialize(obs_encoder_t *encoder)
 	return success;
 }
 
+/**
+ * free video mix if it's an encoder only video mix
+ * see `maybe_set_up_gpu_rescale`
+ */
+static void maybe_clear_encoder_core_video_mix(obs_encoder_t *encoder)
+{
+	pthread_mutex_lock(&obs->video.mixes_mutex);
+	for (size_t i = 0; i < obs->video.mixes.num; i++) {
+		struct obs_core_video_mix *mix = obs->video.mixes.array[i];
+		if (mix->video != encoder->media)
+			continue;
+
+		if (!mix->encoder_only_mix)
+			break;
+
+		obs_encoder_set_video(encoder, obs_get_video());
+		mix->encoder_refs -= 1;
+		if (mix->encoder_refs == 0) {
+			da_erase(obs->video.mixes, i);
+			obs_free_video_mix(mix);
+		}
+	}
+	pthread_mutex_unlock(&obs->video.mixes_mutex);
+}
+
 void obs_encoder_shutdown(obs_encoder_t *encoder)
 {
 	pthread_mutex_lock(&encoder->init_mutex);
@@ -523,6 +660,7 @@ void obs_encoder_shutdown(obs_encoder_t *encoder)
 		encoder->first_received = false;
 		encoder->offset_usec = 0;
 		encoder->start_ts = 0;
+		maybe_clear_encoder_core_video_mix(encoder);
 	}
 	obs_encoder_set_last_error(encoder, NULL);
 	pthread_mutex_unlock(&encoder->init_mutex);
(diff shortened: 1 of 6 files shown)


libobs/obs-internal.h
```diff
@@ -1205,6 +1208,9 @@ struct obs_encoder {
 
 	size_t mixer_idx;
 
+	bool gpu_scaling_enabled;
+	enum obs_scale_type scale_type;
+
```
notr1ch, line 1213: Do we need both an `enabled` flag and a `scale_type`? Wouldn't `OBS_SCALE_DISABLE` imply scaling is disabled?

libobs/obs-encoder.c
```diff
@@ -710,6 +837,30 @@ void obs_encoder_set_scaled_size(obs_encoder_t *encoder, uint32_t width,
...
+	if (encoder->info.type != OBS_ENCODER_VIDEO) {
+		blog(LOG_WARNING,
+		     "obs_encoder_enable_gpu_scaling: "
+		     "encoder '%s' is not a video encoder",
+		     obs_encoder_get_name(encoder));
+		return;
+	}
+	if (encoder_active(encoder)) {
+		blog(LOG_WARNING,
+		     "encoder '%s': Cannot enable GPU sclaing "
```
notr1ch, line 854:
```suggestion
		     "encoder '%s': Cannot enable GPU scaling "
```

libobs/obs-encoder.c
```diff
@@ -184,6 +184,110 @@ static inline bool gpu_encode_available(const struct obs_encoder *encoder)
...
+	if (!mix)
+		return;
+
+	mix->encoder_only_mix = true;
+	mix->encoder_refs = 1;
+	mix->view = &obs->data.main_view;
+
+	pthread_mutex_lock(&obs->video.mixes_mutex);
+
+	// double check that nobody else added a matching mix while we've created our mix
```
notr1ch, line 259: Why is a second check necessary here? Couldn't we hold the `mixes_mutex` while we create and add the mix?

notr1ch (commented):
Minor nit on the commit titles, the first word after the module prefix should start with a capital letter.

libobs/obs-internal.h
palana: I can't think of a good reason to keep both if we want `OBS_SCALE_DISABLE` to imply scaling is disabled; not sure when I'll get around to update this, probably on Monday

libobs/obs-encoder.c
palana: Not sure how expensive the individual parts inside `obs_create_video_mix` are, but it creates a few GPU resources and holds `mixes_mutex` for a while which would likely result in a deadlock if `mixes_mutex` is not a recursive mutex

I considered just creating the mix right away and only then checking whether it was necessary and adding it to the mixes list if needed, but traversing existing mixes seemed less expensive than unconditionally creating a new mix

I can double check whether holding the mutex while creating the new mix work though next week

norihiro:
The first word for each commit title should be capitalized.

libobs/obs-encoder.c
palana: @notr1ch: I've added a few profiler sections in <https://github.com/kc5nra/obs-studio/compare/ruwen/core-mix-gpu-scaling-standalone...kc5nra:obs-studio:ruwen/core-mix-gpu-scaling-standalone-profile> which gave the following numbers:

- In a debug build:
```
14:20:58.758: == Profiler Results =============================
14:20:58.758: run_program_init: 4536.06 ms
14:20:58.759:  ┣OBSApp::AppInit: 59.214 ms
14:20:58.759:  ┃ ┗OBSApp::InitLocale: 3.394 ms
14:20:58.759:  ┗OBSApp::OBSInit: 4417.01 ms
14:20:58.759:    ┣obs_startup: 17.556 ms
14:20:58.759:    ┗OBSBasic::OBSInit: 3569.51 ms
14:20:58.759:      ┣OBSBasic::InitBasicConfig: 0.986 ms
14:20:58.759:      ┣OBSBasic::ResetAudio: 0.882 ms
14:20:58.759:      ┣OBSBasic::ResetVideo: 1916.84 ms
14:20:58.759:      ┃ ┗obs_create_video_mix: 7.367 ms
14:20:58.759:      ┃   ┣copy_fps: 0 ms
14:20:58.760:      ┃   ┣open_output: 4.024 ms
14:20:58.760:      ┃   ┣lock_gpu: 0.002 ms
14:20:58.760:      ┃   ┗create_gpu_resources: 3.273 ms
[...snip...]
14:20:58.766: maybe_set_gpu_video_media: min=0 ms, median=0 ms, max=4.959 ms, 99th percentile=4.959 ms
14:20:58.766:  ┣find_mix_pre_create: 0.001 ms, 0.5 calls per parent call
14:20:58.766:  ┣create_mix: 4.952 ms, 0.5 calls per parent call
14:20:58.766:  ┃ ┗obs_create_video_mix: 4.951 ms
14:20:58.766:  ┃   ┣copy_fps: 0 ms
14:20:58.766:  ┃   ┣open_output: 1.848 ms
14:20:58.766:  ┃   ┣lock_gpu: 0.001 ms
14:20:58.767:  ┃   ┗create_gpu_resources: 3.088 ms
14:20:58.767:  ┗find_mix_post_create: 0.001 ms, 0.5 calls per parent call
```

- in a release (`RelWithDebInfo`) build:
```
14:28:55.499: == Profiler Results =============================
14:28:55.499: run_program_init: 4084.23 ms
14:28:55.499:  ┣OBSApp::AppInit: 51.2 ms
14:28:55.500:  ┃ ┗OBSApp::InitLocale: 1.741 ms
14:28:55.500:  ┗OBSApp::OBSInit: 3980.21 ms
14:28:55.500:    ┣obs_startup: 13.025 ms
14:28:55.500:    ┗OBSBasic::OBSInit: 3543.19 ms
14:28:55.500:      ┣OBSBasic::InitBasicConfig: 1.036 ms
14:28:55.500:      ┣OBSBasic::ResetAudio: 0.608 ms
14:28:55.500:      ┣OBSBasic::ResetVideo: 2099.66 ms
14:28:55.500:      ┃ ┗obs_create_video_mix: 3.422 ms
14:28:55.500:      ┃   ┣copy_fps: 0 ms
14:28:55.500:      ┃   ┣open_output: 0.68 ms
14:28:55.500:      ┃   ┣lock_gpu: 0 ms
14:28:55.500:      ┃   ┗create_gpu_resources: 2.721 ms
[...snip...]
14:28:55.505: maybe_set_gpu_video_media: min=0 ms, median=0 ms, max=2.887 ms, 99th percentile=2.887 ms
14:28:55.505:  ┣find_mix_pre_create: 0 ms, 0.5 calls per parent call
14:28:55.505:  ┣create_mix: 2.885 ms, 0.5 calls per parent call
14:28:55.505:  ┃ ┗obs_create_video_mix: 2.885 ms
14:28:55.505:  ┃   ┣copy_fps: 0 ms
14:28:55.505:  ┃   ┣open_output: 0.666 ms
14:28:55.505:  ┃   ┣lock_gpu: 0 ms
14:28:55.505:  ┃   ┗create_gpu_resources: 2.209 ms
14:28:55.505:  ┗find_mix_post_create: 0 ms, 0.5 calls per parent call
```

I'm leaning towards leaving this as they are right now based on those numbers, but I'm open to changing it if there's consensus on whether holding a lock while creating the mix (which would require factoring out the relevant lock calls inside `obs_create_video_mix`) or just creating a new mix unconditionally is the approach to take

libobs/obs-encoder.c
palana: Fixed in https://github.com/obsproject/obs-studio/pull/9043/commits/373a24f4b8bc29d6519e74a38041d6e6b67cac70

libobs/obs-internal.h
palana: @notr1ch: I changed this in <https://github.com/obsproject/obs-studio/pull/9043/commits/cbd606da0354af62b98cfaee7fb4836ed11e8c44>; at that point the name `obs_encoder_enable_gpu_scaling` didn't seem to fit super well anymore, let me know what you think

notr1ch: Looks good to me, I had the same thought but wanted to keep it to one discussion.

libobs/obs-encoder.c
notr1ch: Thanks for benchmarking, those numbers might indeed pose a risk to a 60 FPS render target. Though in practice, when would a video mix be created during an active stream / recording session? Wouldn't they be created ahead of time or on startup where this isn't a concern?

palana: They're created whenever new encoders are created (with output requirements that diverge from whatever is already active), so e.g. changing recording settings while the stream is active or changing stream settings while recording is active?
To be fair, the `create_gpu_resources` section might still pose a risk to rendering or similar since the graphics lock is being held for that duration

libobs/obs-encoder.c
```diff
@@ -710,6 +837,29 @@ void obs_encoder_set_scaled_size(obs_encoder_t *encoder, uint32_t width,
 	encoder->scaled_height = height;
 }
 
+void obs_encoder_set_gpu_scale_type(obs_encoder_t *encoder,
```
tt2468, line 840: Would I be correct in assuming that this GPU scaling feature is for shared texture encoders only? If so, we should impose that restriction here accordingly.

palana: Nope, it's for all encoders, though I haven't done any benchmarking to see how big the difference is when using e.g. x264

palana:
> I wonder if it might be nice to have a comment to explain this so the reader of the code understands what its doing in the future, but other than that I think this is probably the most simple way to add GPU scaling.

@Lain-B: do you have a suggestion on where to add the comment? above `maybe_set_gpu_video_media`?

Lain-B:
I just didn't realize what I was looking at in `maybe_set_gpu_video_media()` at first originally, probably also happened because I didn't write the video mix stuff so I was thrown off a bit at first. I think that function would be the most appropriate place if you did want to have a comment, but I think even just having a bit more elaboration of what's happening in the commit message would be good enough too depending on how you feel about it. But it's not critical to it being merged either way, mostly just something I noted because my slow brain didn't realize what was going on at first until I looked at the whole function. And to be honest, I'm not even exactly sure what comment I'd write (we all know how great I am at writing comments). I just didn't initially realize that it uses OBS' original output scaling through video mixes until I actually went over that whole function, and the commit message didn't really explain that, that's all.

PatTheMav:
Did I overlook something or does this PR implement the feature for the encoders used on Windows only?

If there is no technical reason for it not to work on macOS or Linux, support for them should be added as well.

palana:
> I just didn't realize what I was looking at in `maybe_set_gpu_video_media()` at first originally, probably also happened because I didn't write the video mix stuff so I was thrown off a bit at first. I think that function would be the most appropriate place if you did want to have a comment, but I think even just having a bit more elaboration of what's happening in the commit message would be good enough too depending on how you feel about it. But it's not preventing it being merged either way, mostly just something I noted because my slow brain didn't realize what was going on at first until I looked at the whole function. And to be honest, I'm not even exactly sure what comment I'd write (we all know how great I am at writing comments). I just didn't initially realize that it uses OBS' original output scaling through video mixes until I actually went over that whole function, and the commit message didn't really explain that, that's all.

I'll add a comment on Monday; there's potentially also a better name for that function

> Did I overlook something or does this PR implement the feature for the encoders used on Windows only?
> 
> If there is no technical reason for it not to work on macOS or Linux, support for them should be added as well.

@PatTheMav: I only "fixed" scaling restrictions for GPU encoders taking texture inputs on windows; are there macOS/linux encoders that currently take texture inputs? Encoders taking CPU inputs should work as is with these changes

PatTheMav:
> @PatTheMav: I only "fixed" scaling restrictions for GPU encoders taking texture inputs on windows; are there macOS/linux encoders that currently take texture inputs? Encoders taking CPU inputs should work as is with these changes

Got it - wasn't entirely sure there. Thanks for clearing that up - at least for VT we do a detour over the CPU iirc.

palana:
> I just didn't realize what I was looking at in `maybe_set_gpu_video_media()` at first originally, probably also happened because I didn't write the video mix stuff so I was thrown off a bit at first. I think that function would be the most appropriate place if you did want to have a comment, but I think even just having a bit more elaboration of what's happening in the commit message would be good enough too depending on how you feel about it. But it's not preventing it being merged either way, mostly just something I noted because my slow brain didn't realize what was going on at first until I looked at the whole function. And to be honest, I'm not even exactly sure what comment I'd write (we all know how great I am at writing comments). I just didn't initially realize that it uses OBS' original output scaling through video mixes until I actually went over that whole function, and the commit message didn't really explain that, that's all.

I've added comments in https://github.com/obsproject/obs-studio/pull/9043/commits/5ed4cde9eba97fd7e248ee8feb191308279f697a and queued up a commit title change in https://github.com/obsproject/obs-studio/pull/9043/commits/544a4bd9f294f199021a68b2b876c87f6778c03f, thoughts?

Lain-B (approved):

libobs/obs-encoder.c
Lain-B: Might be worth doing some more testing on this later, because it does does pose a valid question about whether we should consider this for all encoders in the future.

I don't feel this should block the PR for the time being though because it might require a lot more testing and possibly more code. But definitely something to consider.

PR #9045, opened 2023-06-09
libobs: Don't keep the sources mutex in tick_sources

exeldro:
### Description
Don't keep the sources mutex in tick_sources

### Motivation and Context
Allow other threads to use sources while the graphics thread does video_tick, update, activate and deactivate on sources
For example in a gdi+ text source the update and video_tick (when reading file from slow disk or network) can take long.
All this time the sources_mutex was locked, blocking other threads like obs websocket calls.

### How Has This Been Tested?
On windows 64 bit by trying to update a gdi+ text source via obs websocket multiple times a second.

### Types of changes

### Checklist:

### DIFF
--- a/libobs/obs-video.c
+++ b/libobs/obs-video.c
@@ -45,34 +45,41 @@ static uint64_t tick_sources(uint64_t cur_time, uint64_t last_time)
 	/* ------------------------------------- */
 	/* call tick callbacks                   */
 
-	pthread_mutex_lock(&obs->data.draw_callbacks_mutex);
+	pthread_mutex_lock(&data->draw_callbacks_mutex);
 
-	for (size_t i = obs->data.tick_callbacks.num; i > 0; i--) {
+	for (size_t i = data->tick_callbacks.num; i > 0; i--) {
 		struct tick_callback *callback;
-		callback = obs->data.tick_callbacks.array + (i - 1);
+		callback = data->tick_callbacks.array + (i - 1);
 		callback->tick(callback->param, seconds);
 	}
 
-	pthread_mutex_unlock(&obs->data.draw_callbacks_mutex);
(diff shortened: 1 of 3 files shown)


libobs/obs-video.c
```diff
@@ -56,24 +58,31 @@ static uint64_t tick_sources(uint64_t cur_time, uint64_t last_time)
 	pthread_mutex_unlock(&obs->data.draw_callbacks_mutex);
 
 	/* ------------------------------------- */
-	/* call the tick function of each source */
+	/* get a array of all sources to tick    */
```
norihiro, line 61:
```suggestion
	/* get an array of all sources to tick   */
```

Lain-B:
I know there isn't much difference but I'd feel a little more comfortable if this were a variable in the obs_core_data struct

exeldro:
moved `sources_to_tick` to `struct obs_core_data`

libobs/obs-video.c
```diff
@@ -56,24 +56,31 @@ static uint64_t tick_sources(uint64_t cur_time, uint64_t last_time)
 	pthread_mutex_unlock(&obs->data.draw_callbacks_mutex);
 
 	/* ------------------------------------- */
-	/* call the tick function of each source */
+	/* get an array of all sources to tick   */
+
+	obs->data.sources_to_tick.num = 0;
```
Lain-B, line 61: sorry for the additional nitpick, but you can just do `data->` in all these cases because there's already a pointer to the same at the top of the function. :sweat_smile:

Lain-B: Oh wait it seems that someone (myself?) already did the same myself right above. Well that's embarrassing

exeldro: fixed it for all cases in `tick_sources` now

PR #9048, opened 2023-06-09
docs: Add example for DARRAY as function parameter

CodeYan01:
### Description

Add example for DARRAY as function parameter.
image
image

While I'm at it, I also fixed the DARRAY macro definition in the docs (which previously did not show up because of a wrong directive).

### Motivation and Context

When I first started using DARRAY, it was very confusing, and there isn't any real documentation on how to use them other than code in the obs-studio repo.

### How Has This Been Tested?

Built and viewed locally. As for the code example, I tried that myself and it worked correctly, and the integers in the array were incremented.

### Types of changes

### Checklist:

### DIFF
--- a/docs/sphinx/reference-libobs-util-darray.rst
+++ b/docs/sphinx/reference-libobs-util-darray.rst
@@ -11,7 +11,7 @@ Dynamically resizing arrays (a C equivalent to std::vector).
 
    The base dynamic array structure.
 
-.. :c:macro:: DARRAY(type)
+.. macro:: DARRAY(type)
 
    Macro for a dynamic array based upon an actual type.  Use this with
    da_* macros.
@@ -49,6 +49,42 @@ dynamic array value with a reference (&) operator.  For example:
 
    /* free when complete */
    da_free(array_of_integers);
+
+To pass dynamic arrays to functions as parameters, create a typedef with the
+:c:macro:`DARRAY` macro and type so you can pass the dynamic array with a
+reference (&) operator. An alternative is to define a struct that will contain
+the dynamic array, and that struct will be passed to the functions, instead of
+the dynamic array directly. An example with the typedef method:
+
+.. code:: cpp
+
+   typedef DARRAY(int) int_array_t;
+
+   void generate_integers(int_array_t *integers, int start, int end)
+   {
+           for (int i = start; i < end; i++)
+                   da_push_back(*integers, &i);
+   }
+
+   [...]
+
+   int_array_t array_of_integers;
+   da_init(array_of_integers);
+
+   generate_integers(&array_of_integers, 0, 10);
+
+   /* free when complete */
+   da_free(array_of_integers);
+
+**IMPORTANT NOTE:** While it is also possible to accept the internal
+:c:struct:`darray` struct as a function parameter (via the ``da`` member
+variable of dynamic arrays) and redefine a variable with :c:macro:`DARRAY`
+inside the function, doing so is not safe and not recommended. One potential
+issue with it is having a type declaration in the function that is different
+than the type of the actual dynamic array that will be passed to the function,
+which will cause memory access issues that will not be caught by the compiler.
+As mentioned above, the recommended way is to create a typedef or a container
+struct, which will be safer in usage.
 
 .. function:: void da_init(da)
 


CodeYan01:
Just gonna retest. I'm not really sure why functions have to set the function parameter back to the .da of the dynamic array in the function at the end of the function, and removing it seems to cause stuff to not work, but in my example, i'm pretty sure the integers were edited.

norihiro:
> I'm not really sure why functions have to set the function parameter back to the .da of the dynamic array in the function at the end of the function

If you are talking a line like below, this is because the function `insert_packet` modified the dynamic array and the pointer in the structure could be reallocated. If the reallocation happens, the contents of the structure has to be copied back to the original structure.
https://github.com/obsproject/obs-studio/blob/a7dffb98a5dec3721b39612cad776dd13582426b/plugins/obs-ffmpeg/obs-ffmpeg-mux.c#L1154

CodeYan01:
@norihiro Thank you! I finally understand why as you've explained it clearly. I didn't know the array had to be realloc'd (but it makes sense, as otherwise we can't access it with array indexing). Will adjust the example to show this.

CodeYan01:
I made no mention of reallocation, because i think, even if the array isn't realloc'd (because sometimes its capacity is bigger than num, so adding items wouldn't resize it), darray size changes still need to be synced back to the original one, since the num in the original wouldn't be updated otherwise

Lain-B (commented):
I actually disagree with suggesting to programmers the `da` trick. The reason why is because it's very unsafe. I shouldn't even have done it in my prior code.

Instead, I would recommend a note like this:
```
**IMPORTANT NOTE:** If you need to pass dynamic arrays as a parameter to
a function, it's generally recommended to keep them as a member variable
of a structure and then pass a pointer to that structure, even if it's
the only variable within the structure. While you can access the
internal **darray** structure via the **da** member variable, using it
is much more unsafe, so if you must pass a singular dynamic array to a
function, it's recommended to just wrap it in a structure and pass that
structure instead.
```

What do you all think?

CodeYan01:
I agree that the da trick is unsafe, as I've seen first-hand multiple times, but I think I like @norihiro's approach better https://github.com/obsproject/obs-studio/pull/9072 where they simply added a define for the darray at the top of the file, which is used for every function argument. that way, you don't need the da trick, and you pass the darray directly as a parameter. I think that putting the darray in a struct instead is another level of indirection, which means longer code. What do you think about @norihiro's approach?

Lain-B:
I think @norihiro's method is indeed safer, I'm more comfortable with that. You could probably change my commentary to something like "if you need to pass it as a parameter, use it in a structure or create a typedef"

CodeYan01:
Updated. Tried to integrate the note suggested by @Lain-B, though i had to rephrase it in consideration of the example. I picked the typedef to write an example for, as that would make for a shorter example.

docs/sphinx/reference-libobs-util-darray.rst
```diff
@@ -50,6 +50,42 @@ dynamic array value with a reference (&) operator.  For example:
...
+the dynamic array directly. An example with the typedef method:
+
+.. code:: cpp
+
+   typedef DARRAY(int) int_array_t;
+
+   void generate_integers(int_array_t *integers, int start, int end)
+   {
+           for (int i = start; i < end; i++)
+                   da_push_back(*integers, &i);
```
CodeYan01, line 66: For this, I removed the parentheses that is necessary atm, like `da_push_back((*integers), &i);`, due to conversation at https://github.com/obsproject/obs-studio/pull/9072#discussion_r1236385202

CodeYan01:
I added back the parentheses around the darray in `da_push_back((*integers), &i);` despite the convo at https://github.com/obsproject/obs-studio/pull/9072#discussion_r1236385202, as I realized that most plugins still build against older versions.

CodeYan01:
If norihiro's PR adding parentheses in the macro gets merged, I could then update the docs and add `versionchanged::`

Lain-B:
Merge @norihiro's changes. Feel free to update.

Lain-B:
I'm just going to add the modification and merge it manually. You'll see "closed in [commit]"

CodeYan01:
I apologize, I didn't have time this week to look at it. I'll just write a draft PR for the versionchanged directive

Issue #8928, opened 2023-05-19
macOS Screen Capture stops working after waking from sleep


JacksonChen666:
### Operating System Info

macOS 13

### OBS Studio Version

29.1.1

### OBS Studio Log URL

https://obsproject.com/logs/6rGLVnQlA1EhExs7

### Expected Behavior

macOS Screen Capture continues to work (capture) after waking from sleep.

### Current Behavior

macOS Screen Capture freezes after waking from sleep.

### Steps to Reproduce

1. Open OBS
2. Have a macOS Screen Capture source (Add if necessary)
3. (Optional) Start recording
4. Close lid (or go to sleep)
5. Open lid (wake from sleep)
6. Experience the issue

### Anything else we should know?

- 100% reproduction chance
- *Only* macOS Screen Capture seems to stop working (freeze).
- This issue can be fixed by the end-user by restarting OBS Studio, or by switching to another scene collection and back (likely restarting the macOS Screen Capture source).

gxalpha:
We added a "Restart Capture" button to the properties of the source in OBS 30.0 that should be enabled when certain stoppages of capture (computer goes to sleep, user stops capture via menu bar in macOS 14) are detected. Could you try if that button is enabled when the computer wakes up again, and if pressing it fixes the problem?

mattymatty:
same issues happens to me

RytoEX:
The Apple sample app does the same thing as OBS here, so this seems like intended behavior for ScreenCaptureKit. Closing as such.

baardbi:
This has happened to me several times now. If I do some recording of webcam and macOS screen capture for a little while, it works fine. Then I stop the recording and go to lunch (leaving the OBS application open). Then I come back and start recording again, and when I'm done I see that (in the recorded video) only the webcam recording is shown in front of my desktop wallpaper. None of the applications (or mouse pointer, app windows etc.) I use during the recording is shown in the finished video. Just me in front of the desktop wallpaper. If I close the OBS recording before going to lunch and open it when I come back and start recording, it work just fine. The problem is that I sometimes forget to do that, and it's super frustrating when I have recorded an hour of a tutorial video for something.

thefrostypixel:
I don't understand why this has been closed. It's really annoying to have to manually go to OBS, right click the screen recording, select properties and press Restart Capture every time my mac wakes back up, and I don't see any reason for it being that way. Just because an Apple app does it this way that doesn't mean it's convenient or logical in any way.

[...]

PR #9065, opened 2023-06-10
mac-capture: Add SCStreamDelegate to stream to log capture errors

gxalpha:
### Description

Adds a SCStreamDelegate logging when SCK sources stop with an error.

### Motivation and Context

Having a delegate is needed for a future PR where we have to introduce some sort of UI feature that re-enables a stopped stream, as macOS 14 adds a UI for the user to stop SCK streams outside of OBS, or for when the computer goes to sleep (see https://github.com/obsproject/obs-studio/issues/8928).
I'm also low-key hoping (at least a tiny bit) that it could help us diagnose remaining freezing issues that still pop up very sporadically (assuming that they exit with error codes), and as such wouldn't mind this change making its way into a hotfix release (assuming other people think that's acceptable).

### How Has This Been Tested?

macOS 14
Put computer to sleep and (when waking it up again) saw the "Stream stopped with no capture source" message in the log.
Stopped the stream via the menu bar and saw the "User stopped stream" message.

### Types of changes

### Checklist:

### DIFF
--- a/plugins/mac-capture/mac-screen-capture.m
+++ b/plugins/mac-capture/mac-screen-capture.m
@@ -33,7 +33,7 @@ typedef enum {
     ScreenCaptureApplicationStream = 2,
 } ScreenCaptureStreamType;
 
-@interface ScreenCaptureDelegate : NSObject <SCStreamOutput>
+@interface ScreenCaptureDelegate : NSObject <SCStreamOutput, SCStreamDelegate>
 
 @property struct screen_capture *sc;
 
@@ -420,7 +420,8 @@ static bool init_screen_stream(struct screen_capture *sc)
         }
     }
 
-    sc->disp = [[SCStream alloc] initWithFilter:content_filter configuration:sc->stream_properties delegate:nil];
+    sc->disp = [[SCStream alloc] initWithFilter:content_filter configuration:sc->stream_properties
+                                       delegate:sc->capture_delegate];
 
     [content_filter release];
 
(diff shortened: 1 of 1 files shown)


plugins/mac-capture/mac-screen-capture.m
```diff
@@ -1334,6 +1334,30 @@ - (void)stream:(SCStream *)stream
...
+	NSString *errorMessage;
+	switch (error.code) {
+#if __MAC_OS_X_VERSION_MAX_ALLOWED >= 130000
+	case SCStreamErrorUserStopped:
+		errorMessage = @"User stopped stream.";
+		break;
+#endif
+	case SCStreamErrorNoCaptureSource:
+		errorMessage =
+			@"Stream stopped with no capture source. Did the computer go to sleep?";
```
PatTheMav, line 1348: Error messages are not the place to voice hunches or assumption, they should be matter-of-fact only.

If context is required, additional context should be logged instead (i.e. the `NSError` object's error description itself or the stack).

gxalpha: I'll just remove it then, there is no more context to add from the NSError side as far as I can tell. The localized description just says "No displays or windows found to record" (original message on my machine: "Keine Displays oder Fenster für die Aufnahme gefunden"), and the stack is not very helpful helpful:
image

PatTheMav: Bummer, but yeah that's kinda what the error code already spells out. Maybe rephrase to "Stream stopped as capture source was not found". "Capture source was lost" would also be possible, but that would make an assumption again.

gxalpha: Changed to "Stream stopped as no capture source was not found".

PatTheMav (approved):

PR #9072, opened 2023-06-11
Define types of DARRAY macro to carry element types

norihiro:
### Description

Defines type of DARRAY macros and changes function parameter types using `struct darray *` to the newly defined types.

Also add parentheses in DARRAY macros so that any expression can be passed to the macro as if it is a usual function.

### Motivation and Context

Inspired by #9048.

Currently, a pointer of `struct darray *` is passed to some functions because the unnamed union types cannot be identified as the same type even though the union definition is actually same.
However, using the type `struct darray *` looses the types of the element so that it requires type casts.

To avoid the type-casts, I'd like to suggest defining the DARRAY macro types at one place using `typedef` for each usage.

### How Has This Been Tested?

OS: Linux
Compiled with GCC and Clang. CI will also test the code can be built.

Image slideshow and VLC source are tested.

These files also have usages of `.da` but I didn't change because the amount is not so large or I don't have environment to test carefully.
- deps/file-updater/file-updater/file-updater.c
- libobs-d3d11/d3d11-shaderprocessor.cpp
- plugins/linux-capture/xcomposite-input.c
- plugins/mac-videotoolbox/encoder.c
- plugins/obs-ffmpeg/obs-ffmpeg-video-encoders.c

### Types of changes

### Checklist:

### DIFF
--- a/plugins/obs-ffmpeg/obs-ffmpeg-mux.c
+++ b/plugins/obs-ffmpeg/obs-ffmpeg-mux.c
@@ -813,12 +813,12 @@ static inline bool has_audio(struct ffmpeg_muxer *stream)
 	return !!obs_output_get_audio_encoder(stream->output, 0);
 }
 
-static void push_back_packet(struct darray *packets,
+static void push_back_packet(mux_packets_t *packets,
 			     struct encoder_packet *packet)
 {
 	struct encoder_packet pkt;
 	obs_encoder_packet_ref(&pkt, packet);
-	darray_push_back(sizeof(pkt), packets, &pkt);
+	da_push_back(*packets, &pkt);
 }
 
 static void ffmpeg_mux_data(void *data, struct encoder_packet *packet)
@@ -841,8 +841,7 @@ static void ffmpeg_mux_data(void *data, struct encoder_packet *packet)
 
 		if (pts_usec >= first_pts_usec) {
 			if (packet->type != OBS_ENCODER_AUDIO) {
-				push_back_packet(&stream->mux_packets.da,
-						 packet);
+				push_back_packet(&stream->mux_packets, packet);
 				return;
 			}
 
@@ -852,7 +851,7 @@ static void ffmpeg_mux_data(void *data, struct encoder_packet *packet)
 		}
 	} else if (stream->split_file && should_split(stream, packet)) {
 		if (has_audio(stream)) {
-			push_back_packet(&stream->mux_packets.da, packet);
+			push_back_packet(&stream->mux_packets, packet);
 			return;
 		} else {
 			if (!prepare_split_file(stream, packet))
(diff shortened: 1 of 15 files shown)


plugins/obs-ffmpeg/obs-ffmpeg-mux.c
```diff
@@ -1144,14 +1141,13 @@ static void insert_packet(struct darray *array, struct encoder_packet *packet,
...
-		struct encoder_packet *p = packets.array + (idx - 1);
+	for (idx = packets->num; idx > 0; idx--) {
+		struct encoder_packet *p = packets->array + (idx - 1);
 		if (p->dts_usec < pkt.dts_usec)
 			break;
 	}
 
-	da_insert(packets, idx, &pkt);
-	*array = packets.da;
+	da_insert((*packets), idx, &pkt);
```
CodeYan01, line 1150: I don't think the parentheses are needed in lines like this. It's just passed as is, so there's no added clarity, and instead I think it looks more complex. Especially since the third parameter used a `&` and also didn't use parentheses. There's only one similar circumstance in the repo https://github.com/obsproject/obs-studio/blob/a35e5aae4116f6570a72a3e30f57142e59566063/libobs/util/profiler.c#L468 (and two other calls in that same file) (you could search repo with `(\w+\(|,\s*)\(\*\w+\)[,)]`), but hundreds of occurrences without the parentheses in a function parameter (search with `(\w+\(|,\s*)\*\w+[,)]`). Such parentheses only makes sense imo for stuff like `(*p)->type` and `strtol((*argv)+3, ...)`

norihiro: > I don't think the parentheses are needed in lines like this.

You are right. Instead of having parentheses here, I think we need to have the parentheses inside the macro.

CodeYan01: oh, is that because the macro puts it in the sizeof call and uses its fields, so the parentheses are necessary? but yeah if it is okay to simply edit the macro to have the parentheses, i think that would be better for less confusion and for consistency with normal functions

Lain-B: Feel free to add parentheses to the macros if that would make code cleaner

CodeYan01: @norihiro I would appreciate if you could write a separate PR to add the parentheses to the macro and have this PR depend on that (especially since this still a draft anyway), so https://github.com/obsproject/obs-studio/pull/9048 will not have the parentheses anymore

CodeYan01: nevermind, realized i should still add the parentheses in the docs, as plugins would build against older versions

CodeYan01: It's fine to have that change in this same PR.

Lain-B (approved):

norihiro:
I'm sorry that I have some more places I want to replace the use of `.da`. Hopefully, I will push more changes later in this week.

plugins/obs-ffmpeg/obs-ffmpeg-mux.c
CodeYan01: I don't see an approve changes on me, nor "Resolve conversation", so it seems that you are the one who needs to resolve this conversation, @norihiro, just in case you forgot about this

PR #9088, opened 2023-06-17
UI: Release frontend-api callback object

norihiro:
### Description

Sets `nullptr` to `main` in the front-end API callback instance at the destructor of `OBSBasic`.
Also adds a warning message if an API is called to access the main window after the main window has been deleted.

The implementation of this PR is not thread-safe but this should be ok because the right plugins should not access `main` pointer after the main window has been deleted.

Also releases the callback instance at the end of `main` function.

### Motivation and Context

The instance of `OBSStudioAPI` holds a pointer to `OBSBasic`, which will be invalidated after `OBSBasic::~OBSBasic()`. The invalid pointer should not be held by anyone.

This change will detect potential errors in a 3rd party plugin, which tries to access main window after it has been deleted. For example, calling `obs_frontend_get_profile_config` to save something in `obs_module_unload`.

### How Has This Been Tested?

OS: Linux

Build with both GCC and Clang. Ran obs, enable obs-websocket, and exited obs for each build, and checked no errors.

### Types of changes

### Checklist:

### DIFF
--- a/UI/api-interface.cpp
+++ b/UI/api-interface.cpp
@@ -58,24 +58,40 @@ struct OBSStudioAPI : obs_frontend_callbacks {
 
 	inline OBSStudioAPI(OBSBasic *main_) : main(main_) {}
 
+	inline bool main_valid_(const char *func_name)
+	{
+		if (!main) {
+			blog(LOG_WARNING,
+			     "[%s] Tried to access main window after delete.",
+			     func_name);
+			return false;
+		}
+		return true;
+	}
+
+#define main_valid() main_valid_(__FUNCTION__)
+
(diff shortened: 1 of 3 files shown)


norihiro:
The commit 72f3a14 was not ok because obs-websocket (and maybe some other 3rd party plugins) calls the frontend API.
Instead of deleting the callback instance, just set null-pointer to `main` and output a warning if accessing `main`.

notr1ch:
Was there something wrong with #8735?

norihiro:
> Was there something wrong with https://github.com/obsproject/obs-studio/pull/8735?

That PR looks much better than this. I will close this PR.

PR #9090, opened 2023-06-17
mac-capture: Fix wrong CFString to NSNumber cast

gxalpha:
### Description

Fixes a new warning where the CFString `kCGWindowOwnerPID` is casted to an `NSNumber`.
This is wrong. Things most likely currently works because I think when using it it's implicitly casted to `NSString` but really it should never be casted to a number in the first place.

### Motivation and Context

Warning popped up, don't like warnings.

### How Has This Been Tested?

macOS 14
Checked that in the dictionaries where `OWNER_PID` is used that the key is indeed a string.
Tested in a debugger that the comparisons still work the same as before.

### Types of changes

### Checklist:


PatTheMav (approved):

PatTheMav:
This and the header fix in `mac-display-capture.m` were the only fixups required for macOS 14 _so far_.

PR #9091, opened 2023-06-17
CI: Add further packages for build-requirements on Linux

norihiro:
### Description

Adds two packages `lsb-release` and `dh-cmake` as the build requirements.

### Motivation and Context

When running `CI/03_package_obs.sh`, these packages are required but not available when building on docker container.

### How Has This Been Tested?

OS: Ubuntu 22.04.1 LTS

1. Remove these packages.
2. Run `CI/linux/01_install_dependencies.sh` and checked these packages are installed.

### Types of changes

### Checklist:


PatTheMav (approved):
I find it somewhat curious that CPack's `deb` generator doesn't generate entirely Debian-compliant packages despite it being the file suffix' namesake.

norihiro:
Due to the CI's failure on Ubuntu 20.04, I set this as draft.

derrod:
Is this superseeded with the CI changes now?

PatTheMav:
> Is this superseeded with the CI changes now?

I think this would need to be added to the new scripts then. I usually don't test with Docker but `lxc` containers.

PatTheMav:
Even though the build scripts should only be used on GitHub actions, we indeed have code to ensure that build dependencies are installed and available (just in case GitHub might decide to remove them from runners).

To that end I guess the change should be fine.

PR #9102, opened 2023-06-18
obs-scripting: Search native libraries in Lua script's location

### LINKED ISSUE #9100: Lua: OBS doesn't search for native libraries in script's location
### Operating System Info

Windows 10

### Other OS

_No response_

### OBS Studio Version

29.1.2

_No response_

### OBS Studio Log URL

https://obsproject.com/logs/V3rwlTs6CLrO4p0c

### OBS Studio Crash Log URL

_No response_

### Expected Behavior

Load the native binary just like the pure Lua files

### Current Behavior

When attempting to load native libraries from the script's location, OBS only attempts to search for .lua files and not native libraries. So if `local lib = require("mylib")` is in script's location and called `mylib.dll` OBS will only try to find `mylib.lua`

### Steps to Reproduce

1. Add any native library with Lua bindings into the script's location
2. Attempt to include that library in Lua's script using `require`

Currently there is a workaround to put the native library into any of the folders OBS does search, which are OBS's executable location and obs-scripting folder, but it's inconvenient to ask the script's user to go ahead and drag around library files.

Smartkin:
### Description
Adds to the search path of native libraries for OBS to look for.

### Motivation and Context
Currently when using Lua scripts and you include a module that is Lua code. OBS will properly resolve it by searching the script's location, however if the module is a native library OBS will only search its executable location as well as obs-scripting folder. This fixes the issue and allows OBS to resolve native libraries lying in Lua script's location. Fixes #9100

### How Has This Been Tested?
Tested on Windows 10, OBS 29.1.0, Lua 5.1
The script provided uses a native library for Windows 64-bit
[sample-script.zip](https://github.com/obsproject/obs-studio/files/11781606/sample-script.zip)

### Types of changes

### Checklist:

### DIFF
--- a/deps/obs-scripting/obs-scripting-lua.c
+++ b/deps/obs-scripting/obs-scripting-lua.c
@@ -52,6 +52,7 @@ static const char *get_script_path_func = "\
 function script_path()\n\
 	 return \"%s\"\n\
 end\n\
+package.cpath = package.cpath .. \";\" .. script_path() .. \"/?." SO_EXT "\"\n\
 package.path = package.path .. \";\" .. script_path() .. \"/?.lua\"\n";
 
 static char *startup_script = NULL;


WizardCM:
I generally think this is a good idea, but don't know any of the potential pitfalls. How common is it for lua scripts to come with pre-build DLLs that they depend on?

Smartkin:
@WizardCM I'd say it's pretty common for scripts that wanna utilize native libraries for certain tasks. Like doing http requests can be done using luasockets or curl which are written in C with Lua bindings and I haven't found any other solution that'd do it in pure Lua.

PatTheMav:
The major issue with that idea is that it enables developers to create scripts that go far beyond the scope of the scripts that were envisaged when scripting support was implemented.

The main goal for scripting is to automate internal OBS functionality, do little things with local files or at most do little things with data fetched from an URL.

We already get lots of issues and feature requests regarding scripts that bind resources for investigation and in many cases the end result is that people try to do things that the scripting language itself allows, but our implementation of the scripting system cannot support.

And for that reason I'm wary to actively enable _more_ functionality that will just lead to _more_ incompatibilities and issues that we have to research and handle (because once we allow loading of external libraries for extended functionality, there is the fair assumption that we _support_ such extended functionality).

The "official" way to do such things is to write a plugin, not use the scripting system to create a "pseudo"-plugin.

Smartkin:
@PatTheMav I understand the concerns regarding that.

From my point of view though it's an expected behaviour for scripting to support importing native modules from their local environment.

In general though I don't know why would someone write and use a native library for Lua that'd interact with OBS's API. At that point it definitely would just be a much better route to make a plugin.

PatTheMav:
> @PatTheMav I understand the concerns regarding that.
> 
> From my point of view though it's an expected behaviour for scripting to support importing native modules from their local environment.
> 
> In general though I don't know why would someone write and use a native library for Lua that'd interact with OBS's API. At that point it definitely would just be a much better route to make a plugin.

Ah, I might not have been entirely clear on this. My point was that scripts that need to load native Lua modules probably go beyond the scope of what we intend scripts to do, except if you happen to know of native Lua modules one needs to load to achieve such "simple" functionality (I don't know off the top of my head if there is some functionality that isn't available in the normal Lua runtime).

Smartkin:
@PatTheMav normal Lua runtime can't make http requests and requires additional libraries and dependencies which are native. At least from what I could find.

PatTheMav:
> @PatTheMav normal Lua runtime can't make http requests and requires additional libraries and dependencies which are native. At least from what I could find.

Ah well, that is indeed something one can do with Python, so it's fair to expect being able to do the same with Lua.

Are native Lua modules `.so` or `.dylib` on macOS by default?

Smartkin:
@PatTheMav yes, they are

PatTheMav:
> @PatTheMav yes, they are

Which of the two options? 😄

Smartkin:
@PatTheMav oh lol my bad, it's `.so`

PatTheMav:
> @PatTheMav oh lol my bad, it's `.so`

Got it - so for the lack of a better alternative and the need to load native modules for simple functionality, I guess it's fine to allow this and check on a case-by-case basis if limitations that might arise from this are intended or not.

PatTheMav (approved):

stephematician:
Can this be merged?

Lain-B (approved):

PR #9107, opened 2023-06-19
mac-capture: Add button to reactivate stopped capture

gxalpha:
### Description

Adds a "Restart Capture" button to the SCK source that reenables the capture if it has previously failed, e.g. because of the computer going to sleep, or more importantly with macOS 14 if the user stops the stream from outside of OBS.

https://github.com/obsproject/obs-studio/assets/59806498/c8ca21cd-4208-4712-a084-62434c29760a

This button is disabled if there has been no recorded failure (or if the button has been pressed and the capture restarted).

This UX was a suggestion from @Warchamp7.

### Motivation and Context

macOS 14 introduces a button to stop a capture, and we have no real way of reenabling it. Current fix would be updating something, e.g. toggling "Show Cursor", but this is suboptimal.
In the future, this button would also be added to the context bar once that exists for SCK.

### How Has This Been Tested?

macOS 14
Shut down the SCK source in multiple configurations via the menu bar and pressed the button to reenable.
Made sure the button enables even if the properties window is already open.
Made sure there are no memory leaks.

### Types of changes

### Checklist:


derrod:
Is there anything that prevents us from just restarting the capture without the user having to manually do it?

PatTheMav (approved):
Personally I'd rather hide the button if a restart is not necessary instead of disabling that, but I leave that to warchamp to decide.

gxalpha:
Rebased to fix conflicts with and include the recently added macOS Audio Capture source.

PR #9122, opened 2023-06-22
libobs: Add OS keychain API

derrod:
### Description

Implements OS keychain APIs per https://github.com/obsproject/rfcs/pull/54 for Windows, macOS, and Linux (via libsecret).

### Motivation and Context

Secure storage for sensitive data.

### How Has This Been Tested?

Tested on Windows, macOS, and Ubuntu 22.04 with # applied on top.

### Types of changes

### Checklist:

### DIFF
--- a/cmake/Modules/FindLibsecret.cmake
+++ b/cmake/Modules/FindLibsecret.cmake
@@ -1,0 +1,111 @@
+#[=======================================================================[.rst
+FindLibsecret
+-------------
+
+FindModule for libsecret and the associated library
+
+Imported Targets
+^^^^^^^^^^^^^^^^
+
+.. versionadded:: 2.0
+
+This module defines the :prop_tgt:`IMPORTED` target ``Libsecret::Libsecret``.
+
+Result Variables
+^^^^^^^^^^^^^^^^
+
+This module sets the following variables:
+
+``Libsecret_FOUND``
+  True, if the library was found.
+``Libsecret_VERSION``
+  Detected version of found Libsecret library.
+
+Cache variables
+^^^^^^^^^^^^^^^
+
+The following cache variables may also be set:
+
+``Libsecret_INCLUDE_DIR``
+  Directory containing ``secret.h``.
+
+#]=======================================================================]
+
+include(FindPackageHandleStandardArgs)
+
+find_package(PkgConfig QUIET)
+if(PKG_CONFIG_FOUND)
+  pkg_search_module(PC_Libsecret QUIET libsecret-1)
+endif()
+
+# Libsecret_set_soname: Set SONAME on imported library target
+macro(Libsecret_set_soname)
+  if(CMAKE_HOST_SYSTEM_NAME MATCHES "Linux")
+    execute_process(
+      COMMAND sh -c "${CMAKE_OBJDUMP} -p '${Libsecret_LIBRARY}' | grep SONAME"
+      OUTPUT_VARIABLE _output
+      RESULT_VARIABLE _result)
+
+    if(_result EQUAL 0)
+      string(REGEX REPLACE "[ \t]+SONAME[ \t]+([^ \t]+)" "\\1" _soname "${_output}")
+      set_property(TARGET Libsecret::Libsecret PROPERTY IMPORTED_SONAME "${_soname}")
+      unset(_soname)
+    endif()
+  endif()
+  unset(_output)
+  unset(_result)
+endmacro()
+
+find_path(
+  Libsecret_INCLUDE_DIR
+  NAMES libsecret/secret.h
+  HINTS ${PC_Libsecret_INCLUDE_DIRS}
+  PATHS /usr/include /usr/local/include
+  PATH_SUFFIXES libsecret-1
+  DOC "Libsecret include directory")
+
+find_library(
+  Libsecret_LIBRARY
+  NAMES secret secret-1 libsecret libsecret-1
+  HINTS ${PC_Libsecret_LIBRARY_DIRS}
+  PATHS /usr/lib /usr/local/lib
+  DOC "Libsecret location")
+
+if(PC_Libsecret_VERSION VERSION_GREATER 0)
+  set(Libsecret_VERSION ${PC_Libsecret_VERSION})
+elseif(EXISTS "${Libsecret_INCLUDE_DIR}/secret-version.h")
+  file(STRINGS "${Libsecret_INCLUDE_DIR}/secret-version.h" _version_string)
+  string(REGEX REPLACE "SECRET_[A-Z]+_VERSION \\(([0-9]+)\\)" "\\1\\.\\2\\.\\3" Libsecret_VERSION "${_version_string}")
+else()
+  if(NOT Libsecret_FIND_QUIETLY)
+    message(AUTHOR_WARNING "Failed to find Libsecret version.")
+  endif()
+  set(Libsecret_VERSION 0.0.0)
+endif()
+
+find_package_handle_standard_args(
+  Libsecret
+  REQUIRED_VARS Libsecret_INCLUDE_DIR Libsecret_LIBRARY
+  VERSION_VAR Libsecret_VERSION REASON_FAILURE_MESSAGE "Ensure libsecret-1 is installed on the system.")
+mark_as_advanced(Libsecret_INCLUDE_DIR Libsecret_LIBRARY)
+
+if(Libsecret_FOUND)
+  if(NOT TARGET Libsecret::Libsecret)
+    if(IS_ABSOLUTE "${Libsecret_LIBRARY}")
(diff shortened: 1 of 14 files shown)


derrod:
I know that there has been work ongoing for https://github.com/obsproject/rfcs/pull/39 to overhaul how services work, but since it's not submitted or reviewable yet I have added some things I deemed necessary for this to work properly. Namely, a means of removing authentication information from the config/keychain, which currently is not done. In addition to that I also added token invalidation, which isn't strictly necessary, but was easy enough to add once I added the `Delete()` method.

derrod:
Added `libsecret` to the flatpak by copying the [flathub/shared-modules](https://github.com/flathub/shared-modules/blob/master/libsecret/libsecret.json) module file, also tested and verified it works locally.

Since it is compiled with gcrypt it'll use the application-specific key provided by the secrets portal to encrypt and store data in a local file rather than storing it in the OS keyring. This makes them non-user-managable, but does not require access to the Secret Service D-Bus.

cmake/Modules/FindLibsecret.cmake
```diff
@@ -0,0 +1,111 @@
+#[=======================================================================[.rst
+FindLibsecret
+--------
```
PatTheMav, line 3: Align ruler with length of title.

cmake/Modules/FindLibsecret.cmake
```diff
@@ -0,0 +1,111 @@
...
+find_package(PkgConfig QUIET)
+if(PKG_CONFIG_FOUND)
+  pkg_search_module(PC_Libsecret QUIET libsecret-1)
+endif()
+
+# Libsecret_set_soname: Set SONAME on imported library target
+macro(Libsecret_set_soname)
+  if(CMAKE_HOST_SYSTEM_NAME MATCHES "Linux")
+    execute_process(
+      COMMAND sh -c "objdump -p '${Libsecret_LIBRARY}' | grep SONAME"
```
PatTheMav, line 45: We haven't updated that in the other finders yet (but we should): Use `${CMAKE_OBJDUMP}` here for cases where a toolchain file is provided, which might set an architecture-specific `objdump` binary.

cmake/finders/FindLibsecret.cmake
```diff
@@ -0,0 +1,111 @@
+#[=======================================================================[.rst
+FindLibsecret
+--------
```
PatTheMav, line 3: Align ruler with title length.

cmake/finders/FindLibsecret.cmake
```diff
@@ -0,0 +1,111 @@
...
+find_package(PkgConfig QUIET)
+if(PKG_CONFIG_FOUND)
+  pkg_search_module(PC_Libsecret QUIET libsecret-1)
+endif()
+
+# Libsecret_set_soname: Set SONAME on imported library target
+macro(Libsecret_set_soname)
+  if(CMAKE_HOST_SYSTEM_NAME MATCHES "Linux")
+    execute_process(
+      COMMAND sh -c "objdump -p '${Libsecret_LIBRARY}' | grep SONAME"
```
PatTheMav, line 45: See above regarding CMake variable for `objdump` binary.

libobs/cmake/os-macos.cmake
```diff
@@ -6,7 +6,8 @@ target_link_libraries(
           "$<LINK_LIBRARY:FRAMEWORK,AudioUnit.framework>"
           "$<LINK_LIBRARY:FRAMEWORK,AppKit.framework>"
           "$<LINK_LIBRARY:FRAMEWORK,IOKit.framework>"
-          "$<LINK_LIBRARY:FRAMEWORK,Carbon.framework>")
+          "$<LINK_LIBRARY:FRAMEWORK,Carbon.framework>"
```
PatTheMav, line 9: This is fine, but what required you to use Carbon (macOS keeps some functionality around in that framework, even though it "shouldn't" be used anymore).

derrod: Carbon was already there, this just shows a a new line because I moved the parentheses down because I added `Security` at the end (for keychain stuff).

libobs/util/platform-cocoa.m
```diff
@@ -505,3 +506,94 @@ bool cfstr_copy_dstr(CFStringRef cfstring, CFStringEncoding cfstring_encoding, s
...
+
+bool os_keychain_save(const char *key, const char *data)
+{
+    if (!key || !data)
+        return false;
+
+    NSString *nsKey = [[NSString alloc] initWithBytesNoCopy:(void *) key length:strlen(key)
+                                                   encoding:NSUTF8StringEncoding
+                                               freeWhenDone:NO];
+    NSData *nsData = [NSData dataWithBytesNoCopy:(void *) data length:strlen(data) freeWhenDone:NO];
```
PatTheMav, line 518: As you ensure that `key` is not an invalid pointer, you should be able to just box the char pointer to an `NSString` via `NSString *nsKey = @(key);`.

Alternatively you could just use CoreFoundation types throughout:

```
CFStringRef account = CFStringCreateWithCString(kCFAllocatorDefault, key, kCFStringEncodingUTF8);
CFStringRef service = CFStringCreateWithCString(kCFAllocatorDefault, "OBS Studio", kCFStringEncodingUTF8);

CFDataRef = CFDataCreate(kCFAllocatorDefault, data, strlen(data));

CFTypeRef keys[4] = {kSecAttrAccessible, kSecClass, kSecAttrService,
                 kSecAttrAccount, kSecValueData};

CFTypeRef values[4] = {kSecAttrAccessibleWhenUnlocked, kSecClassGenericPassword,
                   service, account, data};
```

It's functionally probably not much different, but personally I prefer to use CoreFoundation types when interacting in the CoreFoundation world.

You might need to `CFRelease` some of them after you're done of course.

libobs/util/platform-nix-libsecret.c
```diff
@@ -0,0 +1,95 @@
...
+ */
+
+#include <gio/gio.h>
+#include <libsecret/secret.h>
+
+#include "platform.h"
+#include "bmem.h"
+
+static const SecretSchema obs_schema = {
+	"com.obsproject.Studio",
```
tytan652, line 24: Unfortunately a reference to `com.obsproject.Studio` has nothing to do inside libobs, it can only be inside OBS Studio which is the UI.

libobs/util/platform-cocoa.m
derrod: I assume I could use `CFStringCreateWithCStringNoCopy`/`CFDataCreateWithBytesNoCopy` to avoid copying (since that not necessary here) in the same way as its done with `NSString`/`NSData`.
I'd still prefer to avoid using a CFDictionary directly since the readability suffers a lot.

PatTheMav: Fair enough - we use `CFDictionary` in `obs-outputs` already to access to root certificates in the keychain, so it would be "familiar". And the readability improves the longer you work with CoreFoundation. 😉

libobs/util/platform-nix-libsecret.c
derrod: We refer to libobs as the "OBS Studio Library" so I think it's fair to use it here as well. Because that's what it is.
https://github.com/obsproject/obs-studio/blob/master/libobs/pkgconfig/libobs.pc.in#L7

tytan652: `com.obsproject.Studio` represent an application not a library.

derrod: The name here is indeed incorrect, as it is supposed to identify a unique schema. I'll change it to something like `com.obsproject.Studio.Secret` I don't see any point in not using that here. The distinction between libobs and OBS Studio is a technicality that just doesn't matter since libobs should never be packaged or shipped as a separate library (I know some do, but that's irrelevant to us).

derrod: Schema name aside, currently entries are prefixed or identified as being from "OBS Studio" as well, if we wanted to avoid that, the API user would need to be able to specify the name so the user-visible entries in the keychain make sense.

For that purpose the API could be amended with another parameter, e.g. `os_keychain_save(const char* namespace, const char *key, const char *data)` so that the UI could then specify `os_keychain_save("OBS Studio", ...)`.

derrod:
Updated with changes made to obsproject/rfcs#54 to add a "label" to the parameters. This requires the API user to specify a user-facing group name for an entry which is then visible in keychain mangers, for example:
keychain access screenshot

Fenrirthviti:
Since this has an open and pending RFC, we are opting to close this until the RFC has been finalized.

The branch and discussion on implementation is welcome to take place on the RFC, and this may be reopened when the RFC is finalized.

PR #9126, opened 2023-06-22
obs-webrtc: Add user-agent, timestamp, and ssrc generation

mondain:
### Description

The following changes were made to provide compatibility and increase usability behind NAT / Firewall. Specific modifications and additions are as follows:

- Creates a User-Agent string for application identification
- Implements a maximum fragment size for video packetization
- Uses a randomized u32 generation routine for SSRC and RTP timestamps to fit RFC-3550
- Adds additional life-cycle methods for clearer flow; Init, Setup...

### Motivation and Context

My motivation is to remove hard-coded SSRC values and to set starting RTP timestamps to match RFC-3550.

### How Has This Been Tested?

All testing during development was performed on Windows 11 x64 against a modified private pre-release version of Red5 Pro Server. As of generation of this PR, I cannot build on this branch in current windows environment and I'm not sure why; its an error in the PowerShell extension within VS Code that does not appear in my separate fork.

### Types of changes

- New feature
- Tweak
- Code cleanup

### Checklist:

### DIFF
--- a/plugins/obs-webrtc/whip-output.cpp
+++ b/plugins/obs-webrtc/whip-output.cpp
@@ -1,16 +1,24 @@
 #include "whip-output.h"
+#include "whip-utils.h"
+
+/*
+ * Sets the maximum size for a video fragment. Effective range is
+ * 576-1470, with a lower value equating to more packets created,
+ * but also better network compatability.
+ */
+static uint16_t MAX_VIDEO_FRAGMENT_SIZE = 1200;
 
 const int signaling_media_id_length = 16;
 const char signaling_media_id_valid_char[] = "0123456789"
 					     "ABCDEFGHIJKLMNOPQRSTUVWXYZ"
 					     "abcdefghijklmnopqrstuvwxyz";
 
-const uint32_t audio_ssrc = 5002;
+const std::string user_agent = generate_user_agent();
+
 const char *audio_mid = "0";
 const uint32_t audio_clockrate = 48000;
 const uint8_t audio_payload_type = 111;
 
-const uint32_t video_ssrc = 5000;
 const char *video_mid = "1";
 const uint32_t video_clockrate = 90000;
 const uint8_t video_payload_type = 96;
@@ -23,6 +31,7 @@ WHIPOutput::WHIPOutput(obs_data_t *, obs_output_t *output)
 	  running(false),
 	  start_stop_mutex(),
 	  start_stop_thread(),
+	  base_ssrc(generate_random_u32()),
 	  peer_connection(-1),
 	  audio_track(-1),
 	  video_track(-1),
@@ -92,23 +101,28 @@ void WHIPOutput::ConfigureAudioTrack(std::string media_stream_id,
 {
 	auto media_stream_track_id = std::string(media_stream_id + "-audio");
 
+	uint32_t ssrc = base_ssrc;
+
 	rtcTrackInit track_init = {
 		RTC_DIRECTION_SENDONLY,
 		RTC_CODEC_OPUS,
 		audio_payload_type,
-		audio_ssrc,
+		ssrc,
 		audio_mid,
 		cname.c_str(),
 		media_stream_id.c_str(),
 		media_stream_track_id.c_str(),
 	};
 
-	rtcPacketizationHandlerInit packetizer_init = {audio_ssrc,
+	// Generate a random starting timestamp for the audio track
+	uint32_t rtp_audio_timestamp = generate_random_u32();
+
+	rtcPacketizationHandlerInit packetizer_init = {ssrc,
 						       cname.c_str(),
 						       audio_payload_type,
 						       audio_clockrate,
 						       0,
-						       0,
+						       rtp_audio_timestamp,
 						       RTC_NAL_SEPARATOR_LENGTH,
 						       0};
 
@@ -123,26 +137,32 @@ void WHIPOutput::ConfigureVideoTrack(std::string media_stream_id,
 {
 	auto media_stream_track_id = std::string(media_stream_id + "-video");
 
+	// More predictable SSRC values between audio and video
+	uint32_t ssrc = base_ssrc + 1;
+
 	rtcTrackInit track_init = {
 		RTC_DIRECTION_SENDONLY,
 		RTC_CODEC_H264,
 		video_payload_type,
-		video_ssrc,
+		ssrc,
 		video_mid,
 		cname.c_str(),
 		media_stream_id.c_str(),
 		media_stream_track_id.c_str(),
 	};
 
+	// Generate a random starting timestamp for the video track
+	uint32_t rtp_video_timestamp = generate_random_u32();
+
 	rtcPacketizationHandlerInit packetizer_init = {
-		video_ssrc,
+		ssrc,
 		cname.c_str(),
 		video_payload_type,
 		video_clockrate,
 		0,
-		0,
+		rtp_video_timestamp,
 		RTC_NAL_SEPARATOR_START_SEQUENCE,
-		0};
+		MAX_VIDEO_FRAGMENT_SIZE};
 
 	video_track = rtcAddTrackEx(peer_connection, &track_init);
 	rtcSetH264PacketizationHandler(video_track, &packetizer_init);
@@ -150,8 +170,19 @@ void WHIPOutput::ConfigureVideoTrack(std::string media_stream_id,
 	rtcChainRtcpNackResponder(video_track, 1000);
 }
 
-bool WHIPOutput::Setup()
+/**
+ * @brief Initialize encoders and store connect info provided by the service.
+ *
+ * @return bool
+ */
+bool WHIPOutput::Init()
 {
+	if (!obs_output_can_begin_data_capture(output, 0))
+		return false;
+
+	if (!obs_output_initialize_encoders(output, 0))
+		return false;
+
 	obs_service_t *service = obs_output_get_service(output);
 	if (!service) {
 		obs_output_signal_stop(output, OBS_OUTPUT_ERROR);
@@ -164,9 +195,20 @@ bool WHIPOutput::Setup()
 		obs_output_signal_stop(output, OBS_OUTPUT_BAD_PATH);
 		return false;
 	}
+
 	bearer_token = obs_service_get_connect_info(
 		service, OBS_SERVICE_CONNECT_INFO_BEARER_TOKEN);
 
+	return true;
+}
+
+/**
+ * @brief Set up the PeerConnection and media tracks.
+ *
+ * @return bool
+ */
+bool WHIPOutput::Setup()
+{
 	rtcConfiguration config;
 	memset(&config, 0, sizeof(config));
 
@@ -249,13 +291,22 @@ bool WHIPOutput::Connect()
 
 	std::string read_buffer;
 	std::string location_header;
+
 	char offer_sdp[4096] = {0};
 	rtcGetLocalDescription(peer_connection, offer_sdp, sizeof(offer_sdp));
+
+#ifdef DEBUG_SDP
+	do_log(LOG_DEBUG, "Offer SDP:\n%s", offer_sdp);
+#endif
+
+	// Add user-agent to our requests
+	headers = curl_slist_append(headers, user_agent.c_str());
 
 	CURL *c = curl_easy_init();
 	curl_easy_setopt(c, CURLOPT_WRITEFUNCTION, curl_writefunction);
 	curl_easy_setopt(c, CURLOPT_WRITEDATA, (void *)&read_buffer);
-	curl_easy_setopt(c, CURLOPT_HEADERFUNCTION, curl_headerfunction);
+	curl_easy_setopt(c, CURLOPT_HEADERFUNCTION,
+			 curl_header_location_function);
 	curl_easy_setopt(c, CURLOPT_HEADERDATA, (void *)&location_header);
 	curl_easy_setopt(c, CURLOPT_HTTPHEADER, headers);
 	curl_easy_setopt(c, CURLOPT_URL, endpoint_url.c_str());
@@ -318,6 +369,10 @@ bool WHIPOutput::Connect()
 		curl_url_cleanup(h);
 	}
 
+#ifdef DEBUG_SDP
+	do_log(LOG_DEBUG, "Answer SDP:\n%s", read_buffer.c_str());
+#endif
+
 	rtcSetRemoteDescription(peer_connection, read_buffer.c_str(), "answer");
 	cleanup();
 	return true;
@@ -325,6 +380,9 @@ bool WHIPOutput::Connect()
 
 void WHIPOutput::StartThread()
 {
+	if (!Init())
+		return;
+
 	if (!Setup())
 		return;
 
@@ -355,6 +413,9 @@ void WHIPOutput::SendDelete()
 		headers =
 			curl_slist_append(headers, bearer_token_header.c_str());
 	}
+
+	// Add user-agent to our requests
+	headers = curl_slist_append(headers, user_agent.c_str());
 
 	CURL *c = curl_easy_init();
 	curl_easy_setopt(c, CURLOPT_HTTPHEADER, headers);
@@ -403,12 +464,14 @@ void WHIPOutput::StopThread(bool signal)
 
 	SendDelete();
 
-	// "signal" exists because we have to preserve the "running" state
-	// across reconnect attempts. If we don't emit a signal if
-	// something calls obs_output_stop() and it's reconnecting, you'll
-	// desync the UI, as the output will be "stopped" and not
-	// "reconnecting", but the "stop" signal will have never been
-	// emitted.
+	/*
+	 * "signal" exists because we have to preserve the "running" state
+	 * across reconnect attempts. If we don't emit a signal if
+	 * something calls obs_output_stop() and it's reconnecting, you'll
+	 * desync the UI, as the output will be "stopped" and not
+	 * "reconnecting", but the "stop" signal will have never been
+	 * emitted.
+	 */
 	if (running && signal) {
 		obs_output_signal_stop(output, OBS_OUTPUT_SUCCESS);
 		running = false;
@@ -426,15 +489,15 @@ void WHIPOutput::Send(void *data, uintptr_t size, uint64_t duration, int track)
 	if (!running)
 		return;
 
-	// sample time is in us, we need to convert it to seconds
+	// Sample time is in microseconds, we need to convert it to seconds
 	auto elapsed_seconds = double(duration) / (1000.0 * 1000.0);
 
-	// get elapsed time in clock rate
+	// Get elapsed time in clock rate
 	uint32_t elapsed_timestamp = 0;
 	rtcTransformSecondsToTimestamp(track, elapsed_seconds,
 				       &elapsed_timestamp);
 
-	// set new timestamp
+	// Set new timestamp
 	uint32_t current_timestamp = 0;
 	rtcGetCurrentTrackTimestamp(track, &current_timestamp);
 	rtcSetTrackRtpTimestamp(track, current_timestamp + elapsed_timestamp);
--- a/plugins/obs-webrtc/whip-output.h
+++ b/plugins/obs-webrtc/whip-output.h
@@ -1,4 +1,5 @@
 #pragma once
+
 #include <obs-module.h>
 #include <util/curl/curl-helper.h>
 #include <util/platform.h>
@@ -11,13 +12,6 @@
 #include <thread>
 
 #include <rtc/rtc.h>
-
-#define do_log(level, format, ...)                              \
-	blog(level, "[obs-webrtc] [whip_output: '%s'] " format, \
-	     obs_output_get_name(output), ##__VA_ARGS__)
-#define do_log_s(level, format, ...)                            \
-	blog(level, "[obs-webrtc] [whip_output: '%s'] " format, \
-	     obs_output_get_name(whipOutput->output), ##__VA_ARGS__)
 
 class WHIPOutput {
 public:
@@ -37,10 +31,10 @@ private:
 				 std::string cname);
 	void ConfigureVideoTrack(std::string media_stream_id,
 				 std::string cname);
+	bool Init();
 	bool Setup();
 	bool Connect();
 	void StartThread();
-
 	void SendDelete();
 	void StopThread(bool signal);
 
@@ -57,6 +51,7 @@ private:
 	std::mutex start_stop_mutex;
 	std::thread start_stop_thread;
 
+	uint32_t base_ssrc;
 	int peer_connection;
 	int audio_track;
 	int video_track;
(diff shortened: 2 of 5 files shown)


gxalpha (changes requested):
Please squash the clang-format commit with your code changes and fix the commit titles in accordance with the [commit guidelines](https://github.com/obsproject/obs-studio/blob/master/CONTRIBUTING.rst#commit-guidelines).

mondain:
@gxalpha ok will do, whats the process for that, recreation of the PR itself?

.gitignore
```diff
@@ -38,6 +38,7 @@ ipch/
 GeneratedFiles/
 .moc/
 /UI/obs.rc
+*.code-workspace
```
tytan652, line 41: I may misunderstand something, but workspace are normally outside of the local repo because they agglomerate various folder.

I don't see the point of adding it to the `.gitignore`.

gxalpha:
No, keep the PR open and force-push to the branch you created it from. That will update it.

.gitignore
mondain: Developer preference catch-all; I keep workspaces self-contained, but I'm a new submitter here, I'll remove it from the ignore.

pkviet:
Hello, thanks again for this PR. 👍
There are currently 6 commits, the 1st of which is unnecessary, though i guess it's handy to keep it locally for your setup.
So you could run:
`git rebase -i HEAD~6`
This will open your editor and you'll get a screen like this:
notepad++_2023-06-23_08-19-12
Change to the following:
notepad++_2023-06-23_08-13-42

This will drop the 1st commit, squash all the other commits and retain the title of the 4th commit (notice the f -C).
Save your editor and exit. The rebase will complete.
Then you would run:
`git commit --amend`

and add the description you've written on github, so I assume that'd be:
`The following changes were made to provide compatibility and increase 
usability behind NAT / Firewall. Specific modifications and additions 
are as follows:

Creates a User-Agent string for application identification
Implements a maximum fragment size for video packetization
Uses a randomized u32 generation routine for SSRC and RTP timestamps 
to fit RFC-3550
Adds OPTIONS request support for collecting configuration data;
specifically for this use-case is the Link header containing a 
TURN/STUN server (only a single ICE server is used in this initial 
implementation)
Adds additional life-cycle methods for clearer flow; Init, Options, 
Setup...`
As you can see I've formatted the description to 72 chars max per line.
(That's the guideline only for commits, not on the PR github page itself.)

Also for future reference, the commit prefixes in the titles are per project or module (ex: libobs, UI, obs-webrtc, obs-ffmpeg, etc ...). The 50 chars limitation is for the title, excluding the prefix.
In general, one would split the PR in several commits if they deal with different modules (say, one commit in libobs, another in obs-webrtc etc...) or if there are different features for which one wants to facilitate later debugging (ex: obs-webrtc: Feature 1; obs-webrtc: Feature 2).

Lastly, you force push to your fork which will auto update the PR on github.

pkviet:
@Sean-Der @tt2468 could you review this PR ? thanks.

plugins/obs-webrtc/whip-output.cpp
```diff
@@ -323,21 +402,87 @@ bool WHIPOutput::Connect()
...
+	if (res == CURLE_OK) {
+		long response_code;
+		curl_easy_getinfo(c, CURLINFO_RESPONSE_CODE, &response_code);
+		// expected response per draft on OPTIONS is 200
+		if (response_code == 200) {
+			if (!link_header.empty()) {
+				do_log(LOG_INFO,
+				       "WHIP server provided a TURN server via the Link header");
+				struct curl_header *link_header_data;
+				CURLHcode h = curl_easy_header(
```
pkviet, line 449: there's an error here thrown in macOS clang because h is unused (warnings are turned into errors in obs).
So either just use the function without defining h if you're sure it can't fail, or if it's not a CURLHE_OK , do a cleanup.

mondain: That's in the next push.

mondain:
@pkviet I'm very clearly messing up this chain trying to meet these expectations. I don't know how to fix it.

pkviet:
> @pkviet I'm very clearly messing up this chain trying to meet these expectations. I don't know how to fix it.

No worry, Sean tt2468 and i will take a look.

mondain:
@pkviet okay got the short and long desc in; I'll see if I can squash and roll up now.

pkviet:
Mmh ci is failing for mac. But the runner has a recent enough version so it's odd.
@RytoEX tells me that the runner has libcurl  8.1.2.
7.84 is min required.

mondain:
@pkviet my M1 is failing / complaining about libcurl as well

pkviet:
> @pkviet my M1 is failing / complaining about libcurl as well

I don't have a mac but @PatTheMav checked with 
https://github.com/curl/curl/blob/master/docs/examples/headerapi.c
And it was compiled fine on his mac.

Not sure what's going on there.

pkviet:
can you check on your mac the curl version called in code with curl_version() ? seems curl_easy_header was added in april 22 with 7.83 (see https://curl.se/changes.html ); the CURLHcode errors date back from 8.0 march 23 (but you've removed them)

plugins/obs-webrtc/whip-output.cpp
```diff
@@ -134,24 +140,70 @@ void WHIPOutput::ConfigureVideoTrack(std::string media_stream_id,
...
+ * 
+ */
+bool WHIPOutput::Init()
 {
+	// construct operating system info string
+	std::stringstream os_info;
+#ifdef _WIN32
+	os_info << "Windows";
+#elif _WIN64
+	os_info << "Windows x86_64";
```
norihiro, line 173: Is there a case that `_WIN64` is defined but `_WIN32` is not defined? Unless such conditions met, `Windows x86_64` won't appear.

plugins/obs-webrtc/whip-output.cpp
```diff
@@ -134,24 +140,70 @@ void WHIPOutput::ConfigureVideoTrack(std::string media_stream_id,
...
 
-bool WHIPOutput::Setup()
+/**
+ * @brief Init before OPTIONS and Setup due to the need of the endpoint url for SendOptions.
+ * 
+ */
+bool WHIPOutput::Init()
 {
+	// construct operating system info string
+	std::stringstream os_info;
```
norihiro, line 169: Do we need to use `std::stringstream` instead of just `const char *` or `std::string`?

plugins/obs-webrtc/whip-output.cpp
```diff
@@ -134,24 +140,70 @@ void WHIPOutput::ConfigureVideoTrack(std::string media_stream_id,
...
+	// construct information for our user-agent
+	uint32_t obs_ver = obs_get_version();
+	// build the user-agent string
+	std::stringstream ua;
+	// user agent header prefix
+	ua << "User-Agent: Mozilla/5.0 ";
+	// obs version info
+	ua << "(OBS-Studio/" << std::to_string(obs_ver >> 24) << "."
+	   << std::to_string((obs_ver >> 16) & 0xFF) << "."
+	   << std::to_string(obs_ver & 0xFFFF) << "; ";
```
norihiro, line 194: Maybe `obs_get_version_string()` is ok.
```suggestion
	ua << "(OBS-Studio/" << obs_get_version_string() << "; ";
```

plugins/obs-webrtc/whip-output.cpp
```diff
@@ -134,24 +140,70 @@ void WHIPOutput::ConfigureVideoTrack(std::string media_stream_id,
...
+bool WHIPOutput::Init()
 {
+	// construct operating system info string
+	std::stringstream os_info;
+#ifdef _WIN32
+	os_info << "Windows";
+#elif _WIN64
+	os_info << "Windows x86_64";
+#elif __APPLE__
+	os_info << "MacOs";
```
norihiro, line 175: Did you intentionally use a lower case `s`?

plugins/obs-webrtc/whip-output.h
```diff
@@ -19,6 +21,11 @@
 	blog(level, "[obs-webrtc] [whip_output: '%s'] " format, \
 	     obs_output_get_name(whipOutput->output), ##__VA_ARGS__)
 
+// maximum size for a video fragment, keep it under a std MTU of 1500
+// effective range is 576-1470, with the lower the value equating to
+// more packets created
+const uint16_t max_fragment_size(1180);
```
norihiro, line 27: I think `constexpr` would be better.
I'm afraid `the` might be redundant in `with the lower the value`  but I'd like to leave it to be commented by another one.

plugins/obs-webrtc/whip-output.cpp
```diff
@@ -5,16 +5,19 @@ const char signaling_media_id_valid_char[] = "0123456789"
...
 const uint8_t audio_payload_type = 111;
 
-const uint32_t video_ssrc = 5000;
+const uint32_t video_ssrc = generate_random_u32();
 const char *video_mid = "1";
 const uint32_t video_clockrate = 90000;
 const uint8_t video_payload_type = 96;
 
+// user-agent to use for requests constructed at Init()
+std::string user_agent;
```
norihiro, line 19: Do we need to have this variable in the global scope? It looks this variable is updated from `Init`, which is called from `StartThread`. Is this thread safe?

plugins/obs-webrtc/whip-output.cpp
```diff
@@ -5,16 +5,19 @@ const char signaling_media_id_valid_char[] = "0123456789"
...
 const uint32_t audio_clockrate = 48000;
 const uint8_t audio_payload_type = 111;
 
-const uint32_t video_ssrc = 5000;
+const uint32_t video_ssrc = generate_random_u32();
 const char *video_mid = "1";
 const uint32_t video_clockrate = 90000;
 const uint8_t video_payload_type = 96;
 
+// user-agent to use for requests constructed at Init()
```
norihiro, line 18: This comment looks obvious. Is this necessary?

plugins/obs-webrtc/whip-output.cpp
PatTheMav: Correct would be either "macOS" or "Mac OS X", though the latter only identifies versions up to and including 10.15.

plugins/obs-webrtc/whip-output.cpp
```diff
@@ -164,12 +216,35 @@ bool WHIPOutput::Setup()
...
 		service, OBS_SERVICE_CONNECT_INFO_BEARER_TOKEN);
 
+	return true;
+}
+
+/**
+ * @brief Set up the PeerConnection and media tracks.
+ * 
+ * @return true 
+ * @return false
```
norihiro, line 230: Please remove trailing spaces in the lines.

PatTheMav:
macOS curl is 7.85 on CI runners (Xcode 14.2) and 7.87 on local machines (Xcode 14.3), so the version provided by the OS is recent enough to include the symbol.

Compiling the header API example file on CI and locally works without issue, so I have currently no clue what's the issue there.

PatTheMav:
Found the issue - the cURL header declaring the new header API did not contain `extern "C"` for C++ compatibility until version 7.86. GitHub's macOS runners are stuck with Xcode 14.2 by default, which ships with the macOS 13.1 SDK.

The macOS 13.3 SDK contains the more recent version of cURL which fixes the issue for C++ projects, but the macOS 12 runners _do not_ provide Xcode 14.3 (only the macOS 13 runners do, but those do not support running AppleScripts yet, which we require to package the application).

So short of manually installing Xcode 14.3 on the runners, this will not succeed on macOS until GitHub either provides Xcode 14.3 on macOS 12 runners _or_ updates macOS 13 runners to also run AppleScript.

plugins/obs-webrtc/whip-output.h
tt2468: I believe `with a lower value` would be a more correct way of phrasing it.

PatTheMav:
I've authored https://github.com/obsproject/obs-studio/pull/9143 which should fix the build issues once merged.

plugins/obs-webrtc/whip-output.cpp
mondain: @norihiro is this something to watch out for, I assumed clang-format handled this scenario?

plugins/obs-webrtc/whip-output.h
mondain: Darn it, failing at my native language :)

plugins/obs-webrtc/whip-output.cpp
mondain: I'd say no; sometimes I forget to remove "notes to self"

plugins/obs-webrtc/whip-output.cpp
pkviet: i get this for safari UA :
`Mozilla/5.0 (Macintosh; Intel Mac OS X 13_4_1) AppleWebKit/605.1.15 (KHTML, like Gecko) Version/16.5 Safari/605.1.15`
(the version you quote is from April 2014 😉 )

plugins/obs-webrtc/whip-output.cpp
pkviet: we're dropping support for windows x86 so any reference to arch can be dropped actually

mondain: @pkviet excellent; I'll remove it before my next push for clearing up these latest review items.

plugins/obs-webrtc/whip-output.cpp
norihiro: If I understand correctly, `StartThread` calls `Init` at first so that the variable `user_agent` is updated every time the WHIP output started.

plugins/obs-webrtc/whip-output.cpp
norihiro: I cannot find such a usage that the operator `<<` is used only once for a `std::stringstream` variable.

mondain: @norihiro right, but let's not forget your mod which I pulled in :)  I'll get it changed

plugins/obs-webrtc/whip-output.cpp
mondain: @norihiro true, what do you suggest as far as making it constant, or am I misunderstanding?

norihiro: If I were you, I would put `user_agent` as a member variable of the class `WHIPOutput` or construct the user agent everytime the user agent is required because it is required only 3 times for each pair of start-stop.
Thinking 3rd party plugins like obs-multi-rtmp and the possibility of more features in obs, in my opinion, the output should be designed to be thread-safe.

murillo128:
Note that using OPTIONS for turn server config is not the preferred method for WHIP:

> However, this method is not NOT RECOMMENDED and if supported by the underlying WHIP Client's webrtc implementation, the WHIP Client SHOULD wait for the information to be returned by the WHIP Endpoint on the response of the HTTP POST request instead.

It would be great if we could add support for the `setConfiguration` method in the libdatachannel peerconneciton object instead to add the turn server info received on the POST response.

murillo128:
The OPTIONS requests only provides the stun/turn server info, not the ICE candidates. You should not wait for the STUN/TURN server gathering/allocation before sending the POST request to the WHIP endpoint.

So you will have to send the PATCH request anyway regardless if you have retrieved the stun/turn server info via POST or OPTIONS.

mondain:
@murillo128 Yes, well aware of what comes in the OPTIONS. What happens in libdatachannel (not affiliated with OBS studio) with the iceServers information and process is in the library, not OBS. No PATCH is currently required when OPTIONS is used and its TURN server information is passed to iceServers in the context for libdatachannel. This process is very similar if not identical to how browsers work when provided with iceServers.

murillo128:
The ice gathering is an async operation, which according to the spec is inited when the remote description is set (chrome deviates from the spec and starts gathering when the PC is created).

Anyway, if you receive the STUN/TURN server info, and then create the peerconnection and generate the sdp offer, it won't have any stun/turn ICE candidates available on it. 

You will have to wait for libdatachannel to fire the `pc.onLocalCandidate()` callback and send it back on a PATCH request:
https://github.com/paullouisageneau/libdatachannel#signal-a-peerconnection

mondain:
@murillo128 I'll have to verify the libdatachannel process flow for that particular item, but for now its outside the scope of this PR; a PATCH process could most certainly increase the capability and value for the obs-webrtc plugin to be sure. As an aside, all browsers our team tested with for WHIP all play well with OPTIONS and provide srflx candidates in their offer SDP. There may be some magic that our front-end team did, but I'd imagine it was primarily a function of the browser itself (Not just Chrome).

mondain:
@murillo128 yeah, don't get me wrong, I prefer to hard-code nothing :)

murillo128:
I have started to implement the required changes to support STUN/TURN servers on the POST response here:

https://github.com/paullouisageneau/libdatachannel/pull/920

plugins/obs-webrtc/whip-output.cpp
```diff
@@ -5,12 +5,12 @@ const char signaling_media_id_valid_char[] = "0123456789"
...
 					     "abcdefghijklmnopqrstuvwxyz";
 
-const uint32_t audio_ssrc = 5002;
+const uint32_t audio_ssrc = generate_random_u32();
 const char *audio_mid = "0";
 const uint32_t audio_clockrate = 48000;
 const uint8_t audio_payload_type = 111;
 
-const uint32_t video_ssrc = 5000;
+const uint32_t video_ssrc = generate_random_u32();
```
murillo128, line 13: we need to make sure that `video_ssrc` is not the same as `audio_ssrc`

mondain: Its not

tt2468 (approved):

mondain:
@tt2468 if it gets us finally merged for such a minor PR yields a more capable WebRTC feature, I'm fine with "predictable" SSRC's; although without looking it up in the RFC's and locating citations, I'm fairly certain this is not any expectation of SSRC's. No browser that I'm aware of produces predictable SSRC's.

RytoEX (changes requested):
The commits need squashed before merging, and the final commit(s) will need to be appropriately prefixed with "obs-webrtc: ".

plugins/obs-webrtc/whip-output.cpp
murillo128: there is no warranty that the values returned by the function in two consecutive invocations can't be equal. The probability is low (very low), but nevertheless possible.

mondain:
I'm not aware of a requirement to have atomic decay level randomness on a
single instance on a single computer doing media streams with 1 audio and 1
video; there's good enough and there's nitpicking for no good purpose.

plugins/obs-webrtc/whip-output.cpp
```diff
@@ -92,23 +93,28 @@ void WHIPOutput::ConfigureAudioTrack(std::string media_stream_id,
 {
 	auto media_stream_track_id = std::string(media_stream_id + "-audio");
 
+	uint32_t ssrc = base_ssrc;
```
DDRBoxman, line 96: Is this variable necessary? I guess it helps with readability a bit when you look at the companion video function.

plugins/obs-webrtc/whip-output.h
```diff
@@ -109,3 +120,49 @@ static size_t curl_headerfunction(char *data, size_t size, size_t nmemb,
 
 	return real_size;
 }
+
```
DDRBoxman, line 123: Can we pull the below functions out into a util file?

DDRBoxman (approved):
Looks pretty solid, left some nitpicks. 👍

plugins/obs-webrtc/whip-output.cpp
norihiro: If we have multiple output from obs-webrtc, eg. stop and start again or start multiple outputs by a future change, the same SSRCs will be used among multiple outputs. Is that ok?
If not, `video_ssrc` and `audio_ssrc` should be generated for each instance and stored in the class `WHIPOutput`.

murillo128:
> I'm not aware of a requirement to have atomic decay level randomness on a
single instance on a single computer doing media streams with 1 audio and 1
video; there's good enough and there's nitpicking for no good purpose.

murillo128:
>I'm not aware of a requirement to have atomic decay level randomness on a
single instance on a single computer doing media streams with 1 audio and 1
video; there's good enough and there's nitpicking for no good purpose.

I assume this is about the ssrc geenreation. There is a hard requirement that the audio and video ssrc must be different, and generating two random values is not going to ensure that. 

I think I saw a proposal of just using `audio = random()` and `video = audio+1` that will trivially solve that. We would also want to avoid having an `ssrc = 0` as I have seen some lazy implementations failing in that case.

plugins/obs-webrtc/whip-output.cpp
```diff
@@ -23,6 +23,7 @@ WHIPOutput::WHIPOutput(obs_data_t *, obs_output_t *output)
 	  running(false),
 	  start_stop_mutex(),
 	  start_stop_thread(),
+	  base_ssrc(generate_random_u32()),
```
PatTheMav, line 26: As I'm not familiar with WebRTC, can you explain to me how this change is related to user agents and `OPTIONS` requests?

While the PR has been dutifully squashed into a single commit, its message will not suggest to me that `ssrc` generation was also changed as part of it. I'm not opposed to merging it as part of this PR, but I'd rather see those changes moved into its own commit (so _if_ issues arise with it, we don't have to manually pull the changes apart to revert it).

PatTheMav:
> > I'm not aware of a requirement to have atomic decay level randomness on a
> > single instance on a single computer doing media streams with 1 audio and 1
> > video; there's good enough and there's nitpicking for no good purpose.
> 
> I assume this is about the ssrc geenreation. There is a hard requirement that the audio and video ssrc must be different, and generating two random values is not going to ensure that.
> 
> I think I saw a proposal of just using `audio = random()` and `video = audio+1` that will trivially solve that. We would also want to avoid having an `ssrc = 0` as I have seen some lazy implementations failing in that case.

As implemented, at both video or audio `ssrc` can be zero (either because the randomizer yields `0` as the video `ssrc` or we overflow `UINT32_MAX` to `0` by adding `1`).

Using `std::uniform_int_distribution<uint32_t> dist(1, (UINT32_MAX-1));` should fix, as it should neither yield `0` itself, nor a value that would overflow to `0` (iirc `dist` uses inclusive boundaries).

DDRBoxman (changes requested):
Can you fix up the title and description on this PR?

Looks like it's just user agent, and timestamp / ssrc generation now.

DDRBoxman (approved):

DDRBoxman (changes requested):
See comment above: https://github.com/obsproject/obs-studio/pull/9126#issuecomment-1614543613

plugins/obs-webrtc/whip-output.h
```diff
@@ -19,6 +22,13 @@
 	blog(level, "[obs-webrtc] [whip_output: '%s'] " format, \
 	     obs_output_get_name(whipOutput->output), ##__VA_ARGS__)
 
+/*
+ * Sets the maximum size for a video fragment. Effective range is
+ * 576-1470, with a lower value equating to more packets created,
+ * but also better network compatability.
+ */
+#define MAX_VIDEO_FRAGMENT_SIZE 1200
```
PatTheMav, line 30: I'd prefer if this were a `static int` (there is no need to use a preprocessor define for this), but won't hold up merging if this isn't changed.

plugins/obs-webrtc/whip-output.cpp
tt2468: It's for code readability mainly, yes. I didn't want there to be confusion about the method being used for SSRC creation. As you can see in the video track function, it's the same style.

DDRBoxman (approved):

RytoEX (approved):
Merging based on approvals from @DDRBoxman and @tt2468 .

PR #6163, opened 2022-03-17
obs-libfdk: Set bitstream to ADTS for mpegts output

pkviet:
### Description
Also modifies rtmp-services.
This sets the aac bitstream to ADTS for the mpegts output.
This fixes a bug with mpegts ouput where the avformat muxer issues
an error with fdk-aac encoder.

### Motivation and Context
Fixes a bug.

### How Has This Been Tested?
Streams work with mpegts output.
Both srt & rist are fine.
Tested on win 10 to nimble server.

### Types of changes
- Bug fix (non-breaking change which fixes an issue)

### Checklist:
- [x] My code has been run through [clang-format](https://github.com/obsproject/obs-studio/blob/master/.clang-format).
- [x] I have read the [**contributing** document](https://github.com/obsproject/obs-studio/blob/master/CONTRIBUTING.rst).
- [x] My code is not on the master branch.
- [x] The code has been tested.
- [x] All commit messages are properly formatted and commits squashed where appropriate.
- [x] I have included updates to all appropriate documentation.

PR #9129, opened 2023-06-22
rtmp-services: Enforce encoder settings according to protocol

pkviet:
### Description
For mpegts output, used in rtmp-custom service, encoder settings
require:
- ADTS with fdk-aac.    

For all non-rtmp protocols, one also requires:
- video encoder header repetition;
This future proofs the code against protocol addition.

Additionally, I've added a check against NULL audio settings, which was
in my fork but that I forgot to add in the PR [1] to my utter shame.
🙈

[1] https://github.com/obsproject/obs-studio/pull/6163

### Motivation and Context
Force audio encoder settings only for SRT and RIST.
Force header repetition for all non-RTMP protocols.

### How Has This Been Tested?
Checked that an srt stream with nvenc h264 / fdk-aac does not errors out meaning
the encoders settings were enforced.

### Types of changes

### Checklist:

### DIFF
--- a/plugins/rtmp-services/rtmp-custom.c
+++ b/plugins/rtmp-services/rtmp-custom.c
@@ -134,18 +134,21 @@ static const char *rtmp_custom_get_protocol(void *data)
 	return "RTMP";
 }
 
-#define RTMP_PROTOCOL "rtmp"
-
 static void rtmp_custom_apply_settings(void *data, obs_data_t *video_settings,
 				       obs_data_t *audio_settings)
 {
 	struct rtmp_custom *service = data;
-	if (service->server != NULL && video_settings != NULL &&
-	    strncmp(service->server, RTMP_PROTOCOL, strlen(RTMP_PROTOCOL)) !=
-		    0) {
+	const char *protocol = rtmp_custom_get_protocol(service);
+	bool has_mpegts = false;
+	bool is_rtmp = false;
+	if (strcmp(protocol, "SRT") == 0 || strcmp(protocol, "RIST") == 0)
+		has_mpegts = true;
+	if (strcmp(protocol, "RTMP") == 0 || strcmp(protocol, "RTMPS") == 0)
+		is_rtmp = true;
+	if (!is_rtmp && video_settings != NULL)
 		obs_data_set_bool(video_settings, "repeat_headers", true);
+	if (has_mpegts && audio_settings != NULL)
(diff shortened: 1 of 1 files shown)


notr1ch:
What about any other protocols that require repeated headers like HLS? I would also expect WHIP to require headers to be repeated.

pkviet:
> What about any other protocols that require repeated headers like HLS? I would also expect WHIP to require headers to be repeated.

f... You're right

tytan652:
> I would also expect WHIP to require headers to be repeated.

`rtmp_custom` does not support WHIP.

pkviet:
> What about any other protocols that require repeated headers like HLS? I would also expect WHIP to require headers to be repeated.

wait, HLS is not available (for now) with rtmp-custom nor is whip

tytan652:
And HLS neither.

notr1ch:
Huh, I thought we allowed HLS in the custom server box. It's only supported to services.json services?

In that case, what is this change for? Doesn't the current behavior match what we want (repeat headers for all non-RTMP protocols)?

pkviet:
> Huh, I thought we allowed HLS in the custom server box. It's only supported to services.json services?
> 
> In that case, what is this change for? Doesn't the current behavior match what we want (repeat headers for all non-RTMP protocols)?

yeah, agreed.
Seems that all protocols require header repetition, except RTMP/RTMPS.
BUT adts is required only with mpegts for fdk-aac.
Also, I had forgotten to check against NULL audio settings.
Additionnally IIRC @tytan652  planned to have protocol selection with rtmp_custom, so HLS or WHIP will be supported by the service in the future ?
I've pushed new code taking all that into account.

tytan652:
> Additionnally IIRC @tytan652 planned to have protocol selection with rtmp_custom, so HLS or WHIP will be supported by the service in the future ?

I plan deprecate to `rtmp_custom` and replace it with a new service that support all of them and one service per protocol for internal use (e.g obs-websocket) so hidden.

pkviet:
> > Additionnally IIRC @tytan652 planned to have protocol selection with rtmp_custom, so HLS or WHIP will be supported by the service in the future ?
> 
> I plan deprecate to `rtmp_custom` and replace it with a new service that support all of them and one service per protocol for internal use (e.g obs-websocket) so hidden in UI.

Gotcha. Got mixed up. Anyway the current code of this PR will then be able to be directly transposed to the future multi proto service.

PR #9139, opened 2023-06-23
cmake: Set bundle name to "OBS Studio"

gxalpha:
### Description

Sets the bundle name to "OBS Studio" instead of "OBS".
Also adds `CFBundleName` everywhere that cmake generates `CFShortBundleName`s. This isn't technically needed for this PR but seems like it should be set as well.

### Motivation and Context

Recent changes to cmake appear to have set it to "OBS" again. We changed it to "OBS Studio" starting in OBS 28.

### How Has This Been Tested?

Tested that the menu bar app menu says "OBS Studio" again.

### Types of changes

### Checklist:

### DIFF
--- a/cmake/Modules/ObsDefaults_macOS.cmake
+++ b/cmake/Modules/ObsDefaults_macOS.cmake
@@ -116,7 +116,7 @@ macro(setup_obs_project)
   string(TIMESTAMP CURRENT_YEAR "%Y")
 
   # Set paths for distribution bundling
-  set(OBS_BUNDLE_NAME "OBS")
+  set(OBS_BUNDLE_NAME "OBS Studio")
   set(OBS_EXECUTABLE_DESTINATION "${CMAKE_INSTALL_BINDIR}")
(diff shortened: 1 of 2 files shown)


PatTheMav:
This does not work - the bundle name is automatically set by Xcode based on the "Product Name" - there is no Xcode setting for BundleName (so setting it this way has no effect).

So for this to work, the product name needs to be changed to "OBS Studio", and all other parts of the build system need to be adapted because the generated bundle will be called "OBS Studio.app".

This behaviour is correct, because "OBS.app"'s bundle name is "OBS", not "OBS Studio".

gxalpha:
I'll admit that I am very confused right now, because when I initially tested and submitted this, it did somehow work. But it doesn't now. I'm not sure how that happened, maybe I didn't clean-rebuild the cmake directory after some testing changes. I'm *very* annoyed about myself right now, because this absolutely shouldn't happen.
Sorry about the PR then, this does not make any sense to stay open.

PatTheMav:
Well the issue persists that the menu bar only shows "OBS" which we did fix before, so we need to fix/revisit this.

PR #9140, opened 2023-06-23
mac-capture: Fix  show hidden windows option which doesn't work immediately.

niw:
### Description
This patch fixes "Show fullscreen and hidden windows/applications", "Show windows with empty names" options works immediately on properties pane as UI behaviors expected.

### Motivation and Context

macOS Screen Capture properties pane has various check box but these are not working as expected until it's opend again. 

### How Has This Been Tested?

Add macOS Screen Capture to the Sources, grant Screen Capture permissions (and restart the application if necessary,) then open property pane.
Toggle these check mars and observe list of windows or applications are immediately updated.

### Types of changes

`screen_capture_build_content_list` or similar calls uses updated `sc->show_hidden_windows` or similar flags, but these are not set before calling these functions. Changing order of calls to solve the problem.

### Checklist:

### DIFF
--- a/plugins/mac-capture/mac-screen-capture.m
+++ b/plugins/mac-capture/mac-screen-capture.m
@@ -904,15 +904,14 @@ static bool content_settings_changed(void *data, obs_properties_t *props, obs_pr
         }
     }
 
+    sc->show_empty_names = obs_data_get_bool(settings, "show_empty_names");
+    sc->show_hidden_windows = obs_data_get_bool(settings, "show_hidden_windows");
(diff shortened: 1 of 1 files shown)


norihiro:
The commit title should start from `mac-capture: ` instead of `FIX: `.

niw:
> The commit title should start from `mac-capture: ` instead of `FIX: `.

Ah, forgot to update commit message not only PR title. Updated.

PatTheMav (approved):
Looks good to me - tested locally and does as advertised.

PR #9141, opened 2023-06-23
mac-capture: Improve various Objective-C code

niw:
### Description

Improve complicated Objective-C implementations in mac-screen-capture implementation.

### Motivation and Context

mac-screen-capture implementation contains some unnecessary complicated Objective-C
implementation, which can be simpler.
Also has some randomness for the value for each boolean like type.

### How Has This Been Tested?

Add macOS Screen Capture to the Sources, grant Screen Capture permissions (and restart the application if necessary,) then open property pane. Change each list and option to make sure it works as expected.

### Types of changes

Includes following Objective-C implementations improvements and cleanups.
- Use `for` instead of overuse of block based iteration code.
- Use `YES` and `NO` for `BOOL` type instead of wrong `TRUE`, `FALSE`, `true`, `false`.
- Use `length` to check empty string checks.

### Checklist:

### DIFF
--- a/plugins/mac-capture/mac-screen-capture.m
+++ b/plugins/mac-capture/mac-screen-capture.m
@@ -290,18 +290,14 @@ static bool init_screen_stream(struct screen_capture *sc)
     sc->stream_properties = [[SCStreamConfiguration alloc] init];
     os_sem_wait(sc->shareable_content_available);
 
-    SCDisplay * (^get_target_display)() = ^SCDisplay *()
+    SCDisplay * (^get_target_display)(void) = ^SCDisplay *
     {
-        __block SCDisplay *target_display = nil;
-        [sc->shareable_content.displays
-            indexOfObjectPassingTest:^BOOL(SCDisplay *_Nonnull display, NSUInteger idx, BOOL *_Nonnull stop) {
-                if (display.displayID == sc->display) {
-                    target_display = sc->shareable_content.displays[idx];
-                    *stop = TRUE;
-                }
-                return *stop;
-            }];
-        return target_display;
+        for (SCDisplay *display in sc->shareable_content.displays) {
+            if (display.displayID == sc->display) {
+                return display;
+            }
+        }
+        return nil;
     };
 
     void (^set_display_mode)(struct screen_capture *, SCDisplay *) =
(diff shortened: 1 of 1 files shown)


norihiro:
The commit title should start from `mac-capture: `.

plugins/mac-capture/mac-screen-capture.m
```diff
@@ -369,20 +358,13 @@ static bool init_screen_stream(struct screen_capture *sc)
...
-                [sc->shareable_content.applications
-                    indexOfObjectPassingTest:^BOOL(SCRunningApplication *_Nonnull application, NSUInteger idx,
-                                                   BOOL *_Nonnull stop) {
-                        if ([application.bundleIdentifier isEqualToString:sc->application_id]) {
-                            target_application = sc->shareable_content.applications[idx];
-                            *stop = TRUE;
-                        }
-                        return *stop;
-                    }];
+            NSMutableArray *target_application_array = [[NSMutableArray alloc] init];
```
PatTheMav, line 361: Given that the `target_application_array` will always contain a single element, I find it wasteful to use a mutable type for this.

Instead I'd rather declare an `NSArray` and then initialise it with the matching application object once it's found.

plugins/mac-capture/mac-screen-capture.m
```diff
@@ -317,19 +313,14 @@ static bool init_screen_stream(struct screen_capture *sc)
...
-                        if ([app.bundleIdentifier isEqualToString:[[NSBundle mainBundle] bundleIdentifier]]) {
-                            obsApp = app;
-                            *stop = TRUE;
-                        }
-                        return *stop;
-                    }];
-
-                NSArray *exclusions = [[NSArray alloc] initWithObjects:obsApp, nil];
-
+                NSMutableArray *exclusions = [[NSMutableArray alloc] init];
```
PatTheMav, line 316: For now, exclusions at most will contain a single element (namely OBS Studio itself), so using a mutable array seems wasteful to me.

I'd rather declare an `NSArray` and init it with the single element inside the loop.

PatTheMav (commented):
Apart from my other comments this code looks correct at first glance, haven't compiled it locally yet.

plugins/mac-capture/mac-screen-capture.m
niw: This is in fact intentional change. Logically, there is the case that we could have a chance that we don't have application that has bundle id that same as the main bundle.
In this case, in the original code we may have `nil` in `NSArray` which crashes the app. (`NSArray` can't hold `nil` in its item.) If we want to safely use `NSArray` and if it's not found use empty array, we can write such code, but it would be more complicated.
The cost of `NSMutableArray` is not really high and ignoreable in this use case I think.
How do you think?

niw: An, never mind actaully, `initWithObjects:nil, nil` may not crash the app because it's `nil` ended, uh, but still tricky code for sure.

PatTheMav: This code only affects OBS. OBS identifies itself through its main bundle, and hence as long as it's running its bundle identifier in SCKs application list will be identical to its main bundle ID.

I can't think of a generation where that bundle ID is _not_ the main bundle ID.

niw: Yeah, I agree with that point. I said logically... practically, it must be always found. Anways, reverthing to use `NSArray` instead now. As I found set `nil` is safe for `initWithObjects:`, I don't have strong opinion on this change anways. Thank you for the comments.

plugins/mac-capture/mac-screen-capture.m
niw: Updated!

plugins/mac-capture/mac-screen-capture.m
niw: Updated.

PatTheMav: Right, logically it could happen and the NSArray reference would be `nil` because it's never assigned an allocated NSArray object.

I guess I'm fine with keeping the mutable type, as it also makes the code probably easier to read.

niw: Let's keep `NSArray` approach here for now, I agree that it's not necessary to use `NSMutableArray` for sure.

PatTheMav (approved):
Looks good to me, didn't notice any impacts to functionality in my tests.

PR #9143, opened 2023-06-24
CI: Update main workflow to enable building with macOS 13 runners

PatTheMav:
### Description
Splits up macOS builds on CI between a macOS 13-based runner for building the project and a macOS 12-based runner for generating disk images for distribution.

### Motivation and Context
Since early 2023, Xcode 14.3 and later requires macOS 13 to run. This requires switching to macOS 13-based runners for support of macOS 14 and beyond.

Unfortunately macOS 13-based runners are still considered "beta" by GitHub themselves (though many issues are fixed purely by switching to a more recent macOS version). For obs-studio specifically, AppleScript support is required to create the disk images used for release, but this support is not available in those runners.

The common workaround to this issue is to generate application bundles on macOS 13 runners, upload them as artefacts, then download and package them on a macOS 12 runner.

### How Has This Been Tested?
Needs to be tested via CI runs as it mainly concerns CI.

### Types of changes

### Checklist:

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -88,7 +88,7 @@ jobs:
 
   macos_build:
     name: '02 - macOS'
-    runs-on: [macos-12]
+    runs-on: [macos-13]
     strategy:
       matrix:
         arch: ['x86_64', 'arm64']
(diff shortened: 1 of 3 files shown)


PatTheMav:
Updated the PR to use the workaround suggested in https://github.com/actions/runner-images/issues/7531#issuecomment-1609188199.

RytoEX (approved):
Seems fine for now. Once macos-13 runners work properly with Applescript without workarounds, we can remove the workaround portion of this.

PR #9148, opened 2023-06-25
libobs: Change return type for getting filter index

cg2121:
### Description
This changes the return type for getting the filter index from a size_t to an int. This makes it easier for developers to use, as an invalid index just returns a -1.

### Motivation and Context
Wanted to change this before an actual release, so we don't break api.

### How Has This Been Tested?
Tested by printing out the index in a script.

### Types of changes

### Checklist:


cg2121:
For the filters index PR, I chose int because DARRAY_INVALID isn't available in scripting. DARRAY_INVALID is (size_t)-1, which evaluates to max size_t, since size_t can't be less than 0. So -1 makes since here as indexes are always 0 and above.

PR #9157, opened 2023-06-26
UI: Fix crash radio list without items

niw:
### Description

Using `OBS_COMBO_TYPE_RADIO` but not adding any items app will crash
due to out-of-bound access.

### Motivation and Context

In case the conditional list UI component setup for example, and there are no items
assigned to the list with `OBS_COMBO_TYPE_RADIO`, app will crash due to
out-of-bounds access without checking nubmer of items.

This patch cheks number of items before accessing first item of items array to prevent
out-of-bounds access.

### How Has This Been Tested?

Create list property with `OBS_COMBO_TYPE_RADIO`, for example, changing existing
`OBS_COMBO_TYPE_LIST` to `OBS_COMBO_TYPE_RADIO`
and see if application works in case application doesn't have any items in the list.

### Types of changes

### Checklist:

### DIFF
--- a/UI/properties-view.cpp
+++ b/UI/properties-view.cpp
@@ -631,12 +631,15 @@ QWidget *OBSPropertiesView::AddList(obs_property_t *prop, bool &warning)
 		for (size_t idx = 0; idx < count; idx++)
 			AddRadioItem(buttonGroup, subLayout, prop, value, idx);
 
-		buttonGroup->setExclusive(true);
-		WidgetInfo *info =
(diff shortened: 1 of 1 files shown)


gxalpha (approved):
Seems correct.
I do wonder why the `WidgetInfo` constructor isn't just passed the `buttonGroup` directly, I'd imagine that is because `QButtonGroup` isn't a `QWidget` and changing the type would cause too much trouble.

PR #9173, opened 2023-06-29
obs-ffmpeg: Increase buffer size for VAAPI CBR

nowrep:
Default ffmpeg buffer size (equal to bitrate) is too low and results in significantly lower video quality on AMD cards.

### Description

Increase HRD buffer size to bitrate * 2 to fix video quality on AMD cards.

### Motivation and Context

gstreamer VAAPI plugin produces much better video quality than the built-in OBS VAAPI (ffmepg) encoder and the reason for that is the buffer size which defaults to bitrate * 2 in gstreamer.

### How Has This Been Tested?

Tested on ArchLinux with mesa 23.1.3 and 7900XTX.

### Types of changes

### Checklist:

### DIFF
--- a/plugins/obs-ffmpeg/obs-ffmpeg-vaapi.c
+++ b/plugins/obs-ffmpeg/obs-ffmpeg-vaapi.c
@@ -300,6 +300,8 @@ static bool vaapi_update(void *data, obs_data_t *settings, bool hevc)
 	enc->context->level = level;
 	enc->context->bit_rate = bitrate * 1000;
 	enc->context->rc_max_rate = maxrate * 1000;
+	if (cbr)
+		enc->context->rc_buffer_size = bitrate * 2000;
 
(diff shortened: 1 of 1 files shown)


flaeri:
Not sure this is a good idea. Vbv/hrd buffer size = bitrate on all other encoders by default. Using 2x would allow for twice the variance. While it may improve visual quality it strays further from what could be considered "CBR". In my personal opinion, cbr should stay cbr, and you should use vbr if you don't care about strict bitrate compliance.

nowrep:
It does increase peak bitrate a bit, that's true. On my test clip at 8000 kbps it increases the peak from 13100 to 16400.

Another way to fix the quality is to set initial buffer fullness to 100% (which is also what AMF encoder does). It does fix most of the quality issues and doesn't increase peak bitrate. I'll open a new PR and close this one.

PR #9177, opened 2023-06-29
cmake: Add flags for reproducible builds on Windows

PatTheMav:
### Description
Compiler and linker flags for reproducible builds on Windows were omitted from the recent CMake rework in error. This commit adds those flags back.

### Motivation and Context
Restores desired ability to create reproducible builds on Windows.

### How Has This Been Tested?
Will be tested by CI run on Windows.

### Types of changes

### Checklist:

### DIFF
--- a/cmake/windows/compilerconfig.cmake
+++ b/cmake/windows/compilerconfig.cmake
@@ -33,8 +33,15 @@ if(CMAKE_VS_WINDOWS_TARGET_PLATFORM_VERSION VERSION_LESS 10.0.20348)
                       "Please download and install the most recent Windows SDK.")
 endif()
 
+set(_obs_msvc_c_options /Brepro /MP /permissive- /Zc:__cplusplus /Zc:preprocessor)
+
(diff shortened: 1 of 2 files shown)


derrod (changes requested):
Reproducible build options work as expected and the files are identical between runs (build -> clean -> build), but the `PDBALTPATH` option does not seem to work, it still gets set to the absolute path.

Not sure to what extent we care about clang, but it supports `/Brepro` as well: https://clang.llvm.org/docs/UsersManual.html#id12

PatTheMav:
> Reproducible build options work as expected and the files are identical between runs (build -> clean -> build), but the `PDBALTPATH` option does not seem to work, it still gets set to the absolute path.
> 
> Not sure to what extent we care about clang, but it supports `/Brepro` as well: https://clang.llvm.org/docs/UsersManual.html#id12

Yeah my current experience is that as long as we use `clang-cl`, the same flags work.

derrod (approved):

RytoEX (approved):
Looks good to me.

PR #9180, opened 2023-06-30
rtmp-services: Add "wavezoo" RTMPS service

mariocoellocastro:
Wavezoo.com streaming server settings for DJ's

### Description
Adding 4 new types of server settings for with 10 servers each type

### Motivation and Context
Our DJ's need a simple way for configuring OBS properly

### How Has This Been Tested?
This was tested with the last release of OBS in a Linux environment (Ubuntu)

### Types of changes

New feature (non-breaking change which adds functionality): New Wavezoo specific stream settings and servers

### Checklist:


derrod (changes requested):
Please read the [Service Submission Guidelines](https://github.com/obsproject/obs-studio/wiki/Service-Submission-Guidelines).

You're also adding the service multiple times which is not necessary. On top of that, you have marked the service as common and not provided any valid ingest servers and the website listed does not appear to be available.

mariocoellocastro:
You're right. I missed the guidelines.

I'll do the proper changes and try again when is ready.

Thanks and sorry for the inconvenience.

Mario

Fenrirthviti:
You can update your branch and force push the changes, which is preferred rather than opening a new PR.

plugins/rtmp-services/data/services.json
```diff
@@ -2846,6 +2846,262 @@
...
+            "supported video codecs": [
+                "h264"
+            ],
+            "supported audio codecs": [
+                "aac"
+            ],
+            "servers": [
+                {
+                    "name": "Server #01 (RTMP)",
+                    "url": "rtmp://live-stream-server-01:1935/LiveApp"
```
norihiro, line 2863: Is this hostname (`live-stream-server-01`) correct? You might need to add your domain name, which is automatically added when you attempt to resolve the hostname.

mariocoellocastro:
No, I'm really ashamed for that. I did nothing well. The next commit will
be fine.

plugins/rtmp-services/data/services.json
```diff
@@ -2846,6 +2846,139 @@
...
+                }
+            ],
+            "protocol": "RTMPS",
+            "supported video codecs": [
+                "h264",
+                "hevc"
+            ],
+            "supported audio codecs": [
+                "aac",
+                "opus"
```
derrod, line 2943: Unless your servers support [enhanced RTMP](https://github.com/veovera/enhanced-rtmp) do not specify "hevc" and "opus" here (or omit those properties entirely).

plugins/rtmp-services/data/services.json
```diff
@@ -2846,6 +2846,139 @@
...
+                        "res": "852x480",
+                        "fps": 30,
+                        "max bitrate": 1500
+                    }
+                ],
+                "max fps": 60,
+                "max video bitrate": 14000,
+                "max audio bitrate": 192,
+                "bframes": 2,
+                "x264opts": "tune=zerolatency scenecut=0"
```
derrod, line 2980: `tune=zerolatency` negatively impacts the quality more than it saves latency, especially if you're allowing b-frames you're better off not having it.

plugins/rtmp-services/data/services.json
```diff
@@ -2846,6 +2846,139 @@
...
+        },
+        {
+            "name": "Wavezoo",
+            "common": false,
+            "more_info_link": "https://blog.wavezoo.com/obs-quality-guide/",
+            "stream_key_link": "https://oozevaw.com/user/bookings?filter=SETTLED_AND_ON_AIR",
+            "servers": [
+                {
+                    "name": "Live Stream Server #01",
+                    "url": "rtmps://live-stream-server-01.wavezoo.com:8443/LiveApp"
```
derrod, line 2858: None of the RTMP URLs resolve.

derrod (changes requested):

derrod:
The RTMP URLs you specified still don't resolve to any IPs.

mariocoellocastro:
Sure. It’s a pool of server names, the server is created on demand for an
specific event. Right now, there’s no server instance behind the URLs.

norihiro:
If the server is not running, it's eligible to be removed like the PR 9387.

mariocoellocastro:
Hi Nirihiro,

   As I explained before, we run the servers on demand, but the URLs are
fixed. If you want to try wavezoo.com I can give you a demo. When an event
(a party in our case) is about to start, we assign a server of the pool to
the party, and the server (a big and expensive one) starts. Then, the
streamer (DJ) connects to the assigned server using the provided stream ID
in order to authenticate the user.

As you can see, we cannot have on line 20 big and expensive servers 24/7.

I appreciate your understanding.

Kind regards,

    Mario

Fenrirthviti:
After some discussion, I am making the call to decline this PR for now.

Wavezoo operates on a non-standard way and, frankly, is a bit of an anti-pattern to how every other streaming service operates with servers that are not expected to be available unless explicitly requested.

It conflicts with the user expectations, and would require custom handling to avoid our service validation checks from failing (as these servers would be offline more often than not). No other services in OBS really operate this way, so this is very much a non-standard way of operating and we'd be carving out a lot of exceptions I would rather not set a precedent for.

If there is an update towards a more standard service setup, we will happily re-review at that time. Thank you for your understanding.

PR #9182, opened 2023-06-30
cmake: Fix configuration errors with Windows legacy path

PatTheMav:
### Description
Adds necessary code to allow legacy code paths to configure without changes to existing configurations.

### Motivation and Context
By default Windows builds use the legacy CMake code paths, which ideally work as-is with existing build directories. Switching to the new code paths should be opt-in (via preset selection), so at least for the time being existing configurations should continue to work after the code update.

### How Has This Been Tested?
Tested on Windows 11 with existing build configuration from _before_ the CMake 3.0 merge and reconfigured in place after updating to current master commit.

### Types of changes

### Checklist:


RytoEX (approved):
Looks okay at a glance. Tested locally and it works as described.

PR #9186, opened 2023-06-30
mac-capture: Mark CoreAudio output capture source as deprecated

gxalpha:
### Description

Marks the CoreAudio output capture source as deprecated when on macOS 13 or newer, hiding it away in the "Deprecated" folder of the Add Source menu along with the old screen capture sources.

image

### Motivation and Context

With macOS Audio Capture merged, users now have a clear and obvious way to capture desktop audio, which does not require any virtual audio cables shenanigans or alternatively hiding a screen capture source.
To avoid confusion, I think it makes sense to hide the old capture source away in the deprecated category and only show the new capture method when supported (macOS 13+). *Technically* the new source doesn't cover 100% of the use cases of the old method (you can't split a programs audio stream into different sources, which in theory could be achieved with the old method and supported applications) but I think it still makes to hide the old one.

### How Has This Been Tested?

macOS 14. The old source is now in the deprecated category.

### Types of changes

### Checklist:

### DIFF
--- a/plugins/mac-capture/plugin-main.c
+++ b/plugins/mac-capture/plugin-main.c
@@ -16,8 +16,6 @@ extern bool is_screen_capture_available() WEAK_IMPORT_ATTRIBUTE;
 
 bool obs_module_load(void)
 {
-	obs_register_source(&coreaudio_input_capture_info);
-	obs_register_source(&coreaudio_output_capture_info);
 #if __MAC_OS_X_VERSION_MAX_ALLOWED >= 120300 // __MAC_12_3
 	if (is_screen_capture_available()) {
 		extern struct obs_source_info sck_video_capture_info;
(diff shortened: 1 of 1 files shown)


PatTheMav:
I do wonder, what kind of devices were expected to show up for this source? Because even with external audio devices (with multiple outputs) connected, the device list is empty for me.

gxalpha:
BlackHole, VB-Cable, etc. There is an allowlist here:
https://github.com/obsproject/obs-studio/blob/aedc3aa64b4b4020b3dbfb850d8fb4574fa92a6a/plugins/mac-capture/audio-device-enum.c#L8-L20
So really just workarounds for the fact you couldn't record audio before SCK.

PatTheMav (approved):

PR #9188, opened 2023-07-01
cmake: Fix copy to rundir in Release and MinSizeRel configurations

PatTheMav:
### Description
Generated post build action batch scripts retained an empty token to the copy command with the used generator expression (as PDBs are only generated for Debug and RelWithDebInfo configurations).

Moving their copy step into its own command (and using the `true` CMake command, which is a no-op) ensures that builds in Release and MinSizeRel correctly finish this step.

### Motivation and Context
Fix builds in Release and MinSizeRel configurations.

### How Has This Been Tested?
Tested on Windows 11 with all 4 configurations.

### Types of changes

### Checklist:


flaeri:
Can confirm, this solves the issue on my machine.

PR #9192, opened 2023-07-02
UI: Enable control of transitions audio properties

Penwy:
### Description

This aims to implement a standardised way to control the audio properties (mono, volume, balance, offset, monitoring and mixer channels) of transition sources (or rather, of their child sources).

Audio properties are saved as data settings of the transition source (transitions, being composite sources cannot hold audio properties of their own), and it's left to the transition's implementation to apply those to the relevant child sources, either on `source_update`, or on a custom procedure, `transition_audio_update` (provided as a means to update the child sources audio properties without doing a full update).

From a UI/UX standpoint, this adds an entry in the transition properties menu (provided that the transition has the `OBS_TRANSITION_HAS_AUDIO` flag):
Workspace 1_008
This entry then opens a dialog similar to the advanced audio properties, tied to the transition's audio properties
Selection_1244

### Motivation and Context

Control of transition volume can be useful as a lot of transition are provided as premade packs and might not fit the user's audio setup.
Control of transition mixer tracks is undeniably useful for editing or VoD audio isolation purposes.
Control of transition monitoring was pre-existing for stingers, but this provides a standardised way to enable it for future or third-party transitions.
The rest is probably less useful, but felt a decent addition to have full advanced audio control.

This was based on and expanded from the suggestions :
https://ideas.obsproject.com/posts/495/stinger-transition-audio-controls
https://ideas.obsproject.com/posts/1435/stinger-transition-audio-controller

### How Has This Been Tested?

Tested on Ubuntu 22.04.
Tested by using a stinger video with audio and checking all audio setting applied correctly and persisted on save/load.
Linked below is a good media file for testing purposes if need be.
[stinger_countdown.webm](https://github.com/obsproject/obs-studio/assets/37129613/7303fede-810b-48cf-9bb2-897a48bad236)

### Types of changes

### Checklist:


LukiEnLive:
we need this ! 
Because sometimes we cannot directly modify the source file of the transition because some formats are not editable with software 😢

Warchamp7:
I think this is a feature we should add but I want to handle it in a different manner UX wise.

PR #9200, opened 2023-07-04
UI: Soft-delete scene collections

derrod:
### Description

Instead of actually removing the file, simply rename it to the .bak version which OBS ignores in most cases.

That allows users to recover accidentally deleted collections more easily, while only taking up a few kilobytes of disk space at most.

### Motivation and Context

Save users from themselves, at least a little. Suggested by @notr1ch on Discord.

### How Has This Been Tested?

Deleted collection, file was just renamed, and trivial to recover (just remove the `.bak`).

### Types of changes

### Checklist:

### DIFF
--- a/UI/window-basic-main-scene-collections.cpp
+++ b/UI/window-basic-main-scene-collections.cpp
@@ -343,11 +343,8 @@ void OBSBasic::on_actionRemoveSceneCollection_triggered()
 		api->on_event(OBS_FRONTEND_EVENT_SCENE_COLLECTION_CHANGING);
 
 	oldFile.insert(0, path);
-	oldFile += ".json";
-
-	os_unlink(oldFile.c_str());
-	oldFile += ".bak";
-	os_unlink(oldFile.c_str());
+	/* os_rename() overwrites if necessary, only the .bak file will remain. */
+	os_rename((oldFile + ".json").c_str(), (oldFile + ".json.bak").c_str());
 
 	Load(newPath.c_str());
 	RefreshSceneCollections();


WizardCM:
I do think this is a good idea overall. But I will mention that depending on how certain plugins (both first and third party) store data, a scene collection can be larger than a few KB. The most obvious example is the VST 2.0 plugin.

gxalpha:
What about moving it to the system trash? To me that seems like the most logical place where a user would check. (I think Qt should have a function for that already)

derrod:
> What about moving it to the system trash? To me that seems like the most logical place where a user would check. (I think Qt should have a function for that already)

It does, yeah: https://doc.qt.io/qt-6/qfile.html#moveToTrash-1

However, restoring from trash wouldn't be handled and it wouldn't show up again until scene collections are refreshed, so it wouldn't really work as users might expect.

SuslikV:
The option is not doing what it says. "Hide from the list" or "Remove from the list" is better description what it really do with this PR.

External backups is what users should do. It is universal.
Nobody knows where to look for the `.bak` files and what they means.

notr1ch:
"Remove from the list" is too long, so perhaps shorten it to "Remove". Or we can use "Delete from the list" and shorten it to "Delete". Users don't really need to care about the implementation details.

SuslikV:
> ...Users don't really need to care about the implementation details.

It is not about users. My comment was rather for developers (new developers, that will join you sooner or later). When you very first time looking at the option - you expecting exactly what it says.

Many users don't understands "magic" of OBS. It is hard to explain to the end user that Scene Collection is a file stored in the user's PC. Users who seen Notepad in Windows can't get why final product of OBS - recording file (or streaming) can't be saved through the File menu. Why to open media file you need to go not to the File menu but to click some tool buttons (at least drag-n-drop works, and this is good). For user - Scene Collection is some "magical" menu, you **don't** need to: create new project, to load saved project, the program **don't** ask if you want to save changes when you modified your current project and quit OBS. There are a lot of "magic" in OBS for end users. Users will not care if option will be named just "Remove" (as it was). In eyes of end user the name associated with Scene Collection simply disappears "from somewhere of OBS internals" and that's all. Of course, some advanced users would like to know what is behind the curtains. And they will learn it no matter what name will be given to the option. Old (advanced) users of OBS will expect file delete but likely do not notice the changes.

Lain-B:
I'll merge this for now, we can adjust it more later if needed

PR #9206, opened 2023-07-06
cmake: Detect environment variables for service configuration in CMake presets

PatTheMav:
### Description
Allows OAuth configuration and other environment-based variables be detected and used in CMake presets.

### Motivation and Context
This makes the use of repository secrets set as environment variables in CI runs easier to work with, as they are directly adapted by the CMake preset and used in the build configuration.

### How Has This Been Tested?
Tested on separate fork to check for proper masking of repository secrets in GitHub Actions output.

### Types of changes

### Checklist:


derrod:
I assume that if you have manually set these in a user preset or via command line parameters it'll overwrite whatever is set in this base preset?

PatTheMav:
> I assume that if you have manually set these in a user preset or via command line parameters it'll overwrite whatever is set in this base preset?

At least in my tests, that's what it does. Cache variables defined on the command line overrule any preset, followed by user presets and then default presets.

RytoEX (changes requested):
Nit: oAUTH -> OAuth

PatTheMav:
> Nit: oAUTH -> OAuth

Oh that one slipped through.

RytoEX (approved):
Seems fine.

PR #9211, opened 2023-07-06
libobs: Fix flip calculations in bounding box

Penwy:
### Description

This fixes flipping behaviour when positional alignement inside the bounding box is other than centered, so that the item gets flipped in-place within the bounding box, instead of being thrown out of the bounding box.

### Motivation and Context

Currently, when a sceneitem set in a bounding box with an alignment other than "center" gets flipped, it is thrown out of the bounding box, which does not feel like the expected behaviour of a bounding box.

### How Has This Been Tested?

Ubuntu 22.04
Tested various combinations of positional alignement and bounding box types, checked that all correctly flipped the item in-place on x and y.

### Types of changes

### Checklist:

### DIFF
--- a/libobs/obs-scene.c
+++ b/libobs/obs-scene.c
@@ -401,13 +401,19 @@ static void calculate_bounds_data(struct obs_scene_item *item,
 
 	width = (float)(*cx) * scale->x;
 	height = (float)(*cy) * scale->y;
-	width_diff = item->bounds.x - width;
-	height_diff = item->bounds.y - height;
+
+	/* Disregards flip when calculating size diff */
+	width_diff = item->bounds.x - fabsf(width);
+	height_diff = item->bounds.y - fabsf(height);
 	*cx = (uint32_t)item->bounds.x;
 	*cy = (uint32_t)item->bounds.y;
 
 	add_alignment(origin, item->bounds_align, (int)-width_diff,
 		      (int)-height_diff);
+
+	/* Makes the item stay in-place in the box if flipped */
+	origin->x += (width < 0.0f) ? width : 0.0f;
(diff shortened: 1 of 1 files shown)


norihiro:
The description has a 75-character line, where the limit is 72 characters.

libobs/obs-scene.c
```diff
@@ -401,13 +401,17 @@ static void calculate_bounds_data(struct obs_scene_item *item,
 
 	width = (float)(*cx) * scale->x;
 	height = (float)(*cy) * scale->y;
-	width_diff = item->bounds.x - width;
-	height_diff = item->bounds.y - height;
+	// Disregards flip when calculating size diff
```
cg2121, line 404: In c files it is suggested to /* Comment */ styled comments, instead of //

libobs/obs-scene.c
```diff
@@ -401,13 +401,17 @@ static void calculate_bounds_data(struct obs_scene_item *item,
...
+	width_diff = item->bounds.x - fabsf(width);
+	height_diff = item->bounds.y - fabsf(height);
 	*cx = (uint32_t)item->bounds.x;
 	*cy = (uint32_t)item->bounds.y;
 
 	add_alignment(origin, item->bounds_align, (int)-width_diff,
 		      (int)-height_diff);
+	// Makes the item stay in-place in the box if flipped
```
cg2121, line 412: Same here

libobs/obs-scene.c
RytoEX: That is only required for multiline comments. For single line comments, there is no set rule for `//` vs. `/* */`.

PR #9215, opened 2023-07-08
obs-qsv11: Add VP9 changes

shinji3:
### Description

- Add Intel VP9 hardware encoder
### Motivation and Context

- We should have a way to use Intel VP9 hardware encoder in streaming and recording.
### How Has This Been Tested?

- Intel Alder Lake CPU
### Types of changes

### Checklist:

### DIFF
--- a/plugins/obs-qsv11/QSV_Encoder.h
+++ b/plugins/obs-qsv11/QSV_Encoder.h
@@ -75,6 +75,10 @@ static const struct qsv_rate_control_info qsv_ratecontrols[] = {
 	{"AVBR", false},  {"ICQ", true},  {"LA_ICQ", true}, {"LA_CBR", true},
 	{"LA_VBR", true}, {0, false}};
 
+static const struct qsv_rate_control_info qsv_vp9_ratecontrols[] = {
+	{"CBR", false},   {"VBR", false},   {"CQP", false},   {"ICQ", true},
+	{"LA_ICQ", true}, {"LA_CBR", true}, {"LA_VBR", true}, {0, false}};
+
 static const struct qsv_rate_control_info qsv_av1_ratecontrols[] =
 	{{"CBR", false}, {"VBR", false}, {"CQP", false}, {0, false}};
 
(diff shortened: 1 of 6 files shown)


pkviet:
Is there a use case where av1 would not be more appropriate?

shinji3:
Enhanced RTMP supports VP9.
AV1 cannot be used without Intel Arc GPU.

pkviet:
> Enhanced RTMP supports VP9.
> AV1 cannot be used without Intel Arc GPU.

Ah fair enough i suppose

derrod:
> Enhanced RTMP supports VP9. AV1 cannot be used without Intel Arc GPU.

OBS does not support VP9 for enhanced RTMP or recording (not fully at least). And neither does YouTube or any of the other implementations I've seen yet.

shinji3:
The latest version of ffmpeg supports a feature called Enhanced RTMP.
I tried streaming to YouTube Live using this feature and found that HEVC and AV1 worked well,but VP9 did not.
It seems that YouTube Live does not support VP9.

shinji3:
It seems that you can debug Enhanced RTMP with the latest ffmpeg (B frame cannot be used)

`ffmpeg -listen 1 -i rtmp://127.0.0.1/live/livestream -c copy output.mkv`

settings

knock2one:
Sorry for necroposting but - QuickSync VP9 would be killer for A-Series cards! According to [this graph](https://www.intel.com/content/www/us/en/support/articles/000098345/graphics.html), it has support for 8-bit 4:4:4 (for example). That would probably be better than HEVC at the same job.

derrod:
> Sorry for necroposting but - QuickSync VP9 would be killer for A-Series cards! According to [this graph](https://www.intel.com/content/www/us/en/support/articles/000098345/graphics.html), it has support for 8-bit 4:4:4 (for example). That would probably be better than HEVC at the same job.

The VP9 encoder is quite bad, there's no reason to use it (or for us to support it). There's a PR for 4:4:4 support for HEVC already, which is significantly more useful: #13152

Fenrirthviti:
This PR is quite old, and I agree that there are better options available in OBS now, and this is not likely to be merged or worth spending much more efforts on.

PR #9218, opened 2023-07-08
obs-ffmpeg: Fix media source always STATE_ENDED

CodeYan01:
### Description

Previously would always end up with OBS_MEDIA_STATE_ENDED whether the
video reaches the end or the user stops the source (state is set to
STOPPED, but becomes ENDED right after). This fixes it so that if the
user stops the source, it will remain stopped.

### Motivation and Context

Found this bug after https://github.com/obsproject/obs-studio/pull/9137 got merged. Even Advanced Scene Switcher users are affected by this weirdness. My interpretation of the two states is that ENDED is when the video naturally ends, and STOPPED when the user stops the video manually.

### How Has This Been Tested?

I put a breakpoint in update function to watch the state.
- If video ended, state is `OBS_MEDIA_STATE_ENDED`
- If video stopped, state is `OBS_MEDIA_STATE_STOPPED`

This also made the referenced PR do what was intended; to only restart the media if the state is `OBS_MEDIA_STATE_ENDED`

### Types of changes

### Checklist:

### DIFF
--- a/plugins/obs-ffmpeg/obs-ffmpeg-source.c
+++ b/plugins/obs-ffmpeg/obs-ffmpeg-source.c
@@ -316,8 +316,10 @@ static void media_stopped(void *opaque)
 	if ((s->close_when_inactive || !s->is_local_file) && s->media)
 		s->destroy_media = true;
 
-	set_media_state(s, OBS_MEDIA_STATE_ENDED);
-	obs_source_media_ended(s->source);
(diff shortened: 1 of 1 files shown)


CodeYan01:
Also included `obs_source_media_ended(s->source);` in the if condition, as I noticed in my plugin that it triggered the ended signal

PR #5873, opened 2022-01-27
UI: Bunch of RAII stuff

cg2121:
### Description
- Add auto release for refs that don't have them
- Remove disconnect signals for OBSSignal
- Add RAII wrappers for volume meters, faders and properties
- Use BPtr for chars that don't have it

### Motivation and Context
WIth RAII, resources are automatically managed.

### How Has This Been Tested?
Went through the entire program with my setup to see if there were any bugs or memory leaks.
More testing on other setups would be appreciated.

### Types of changes
- Code cleanup (non-breaking change which makes code smaller or more readable)

### Checklist:
- [x] My code has been run through [clang-format](https://github.com/obsproject/obs-studio/blob/master/.clang-format).
- [x] I have read the [**contributing** document](https://github.com/obsproject/obs-studio/blob/master/CONTRIBUTING.rst).
- [x] My code is not on the master branch.
- [x] The code has been tested.
- [x] All commit messages are properly formatted and commits squashed where appropriate.
- [x] I have included updates to all appropriate documentation.

PR #9227, opened 2023-07-09
UI: Remove signal disconnect calls

cg2121:
### Description
The signals are wrapped in OBSSignal, so they are already automatically disconnected.

### Motivation and Context
Split from #5873

### How Has This Been Tested?
Ran program to make sure nothing broke

### Types of changes

### Checklist:


PatTheMav:
In general this looks fine, the only scenarios that I'm not fully sure about is when a manual disconnection might have protected against emitted signals before the destructor runs (or before a `Connect()` call implicitly disconnects it).

cg2121:
As @tt2468 mentioned off thread, I also am not quite sure if these are needed or not.

cg2121:
This PR is no longer needed as I converted these signals to vectors anyway.

PR #9235, opened 2023-07-11
Clang format update

derrod:
### Description

Updates the clang format file and affected file by uncommenting options that were previously commented out due to not being supported in the clang-format version used when the file was originally written. As we are now using clang-format 13 we can enable them.

Each option is its own commit to make it easier to review and decide which ones we might want.

Some options did not result in any changes at all since they were the default anyway, but perhaps we want to keep those for forwards-compatibility should clang defaults change.

### Motivation and Context

Cleanup.

### How Has This Been Tested?

Ran the formatting locally using `CI/check-format.sh`.

### Types of changes

### Checklist:

### DIFF
--- a/.clang-format
+++ b/.clang-format
@@ -15,7 +15,7 @@ AllowShortBlocksOnASingleLine: false
 AllowShortCaseLabelsOnASingleLine: false
 AllowShortFunctionsOnASingleLine: Inline
 AllowShortIfStatementsOnASingleLine: false
-#AllowShortLambdasOnASingleLine: Inline  # requires clang-format 9
+AllowShortLambdasOnASingleLine: Inline
 AllowShortLoopsOnASingleLine: false
 AlwaysBreakAfterDefinitionReturnType: None
 AlwaysBreakAfterReturnType: None
--- a/.clang-format
+++ b/.clang-format
@@ -8,7 +8,7 @@ AlignConsecutiveDeclarations: false
 AlignEscapedNewlines: Left
 AlignOperands: true
 AlignTrailingComments: true
-#AllowAllArgumentsOnNextLine: false  # requires clang-format 9
+AllowAllArgumentsOnNextLine: false
 #AllowAllConstructorInitializersOnNextLine: false  # requires clang-format 9
 AllowAllParametersOfDeclarationOnNextLine: false
 AllowShortBlocksOnASingleLine: false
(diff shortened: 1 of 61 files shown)


RytoEX (approved):
I don't have any particular objections to any of these. Don't know how others feel.

PatTheMav:
Are there options shared with the ObjC rules? The only reason I originally was so exhaustive was because those were based on clang-format 16, then adapted to 13. If the other rules are based on 13 as a baseline, there might be a set of shared rules now.

derrod:
There seems to be some duplication, like `SpaceAfterLogicalNot`, with how this file works, would be enough to specify them at the top and then remove them from the ObjC section?

derrod:
Added a commit to remove some of the now duplicated options. Technically part of `BraceWrapping` is also duplicated but I assume that is required for it to work correctly since it overrides the option of the language-agnostic block at the top.

derrod:
> I don't think the C++ block needs any of the ObjC rules when we have a separate ObjC block to begin with.

I updated it to move the ObjC rules that used to be in the main section to the ObjC section. I haven't checked through all the options so I assume that's all the ones relevant to ObjC?

derrod:
Moving `ObjCBlockIndentWidth: 8` to the ObjC block reformats a number of files, is that a setting we want or a legacy thing to omit?

gxalpha:
If opinions are being asked for here: It definitely does look more ugly, and we don't use 8 wide indents elsewhere

PatTheMav:
> Moving `ObjCBlockIndentWidth: 8` to the ObjC block reformats a number of files, is that a setting we want or a legacy thing to omit?

The ObjC rules need to stay as-is (otherwise we get back the issue of Xcode reformatting everything while coding). But the ObjC rules from the C++ block can be removed entirely and the ones that currently are in the ObjC-specific block need to be kept.

derrod:
I've removed that commit and squashed all the changes that only cleanup the config file and do not affect the formatting of code files into one.

PatTheMav:
> I kept the `ObjC*` options in the C/C++ block since removing them changes blocks such as
> 
> https://github.com/obsproject/obs-studio/blob/53cfe5cb7f2f9b6bb851265966ecfc70cd1c7875/plugins/mac-videotoolbox/encoder.c#L1601-L1603

That means the file should probably be renamed to `encoder.m` because I don't think block syntax is proper C code, that's ObjC. If the ObjC rules are not applied, then it's because the file is not correctly recognised as ObjC code..

PatTheMav:
Got it - PR looks fine as-is though.

PR #9238, opened 2023-07-12
libobs: Export more libobs/util headers

gxalpha:
### Description

Adds additional headers from `libobs/util` that plugins might want to use to the `public_headers` list.

These are the headers in `libobs/util` that with this PR are *not* exported:
```cmake
    util/array-serializer.h
    util/bitstream.h
    util/cf-lexer.h
    util/cf-parser.h
    util/crc32.h
    util/curl/curl-helper.h
    util/file-serializer.h
    util/lexer.h
    util/pipe.h
    util/serializer.h
    util/task.h
    util/utf8.h
    util/uthash.h
    util/util_uint128.h
```
If someone thinks some of them might be useful as well, I could add them.

### Motivation and Context

Fixes obsproject/obs-plugintemplate#79
Fixes obsproject/obs-plugintemplate#80
(This relies on the next OBS release of course)

### How Has This Been Tested?

Hasn't. Should be simple enough in theory.

### Types of changes

### Checklist:

### DIFF
--- a/libobs/CMakeLists.txt
+++ b/libobs/CMakeLists.txt
@@ -295,10 +295,17 @@ set(public_headers
     util/base.h
     util/bmem.h
     util/c99defs.h
+    util/circlebuf.h
+    util/config-file.h
     util/darray.h
+    util/dstr.h
+    util/platform.h
     util/profiler.h
+    util/profiler.hpp
     util/sse-intrin.h
     util/text-lookup.h
+    util/threading.h
+    util/util.hpp
     util/util_uint64.h)
 
 if(ENABLE_HEVC)


gxalpha:
I wouldn't call it kitchen sink, the libobs/util directory mostly implements stuff that is needed in all kinds of scenarios (hence the "util" name), having plugins reimplement things like threading or dynamic strings if they already exist seems like busywork.

gxalpha:
> I'd prefer if I could choose to _either_ include a "give me everything" header (which in turn should not only be made of `#include` directives itself) or be able to pick what I need (e.g. `#include <libobs-threading.h>`) with the benefit of `libobs` itself cherrypicking its own dependencies.

Yeah that’s fair

PR #9239, opened 2023-07-12
mac-capture: Remove internal display settings from SCK Audio Capture

gxalpha:
### Description

Removes the internal display settings from SCK Audio Capture.

### Motivation and Context

When capturing desktop audio via ScreenCaptureKit, this always includes the entire audio, independent of which display an application might be running on. As such we don't currently expose a setting to change the display, and don't need any code handling a display setting (as we can always simply use the primary display because it doesn't matter).

### How Has This Been Tested?

macOS 14 23A5286g
Checked that SCK Audio Capture still works.

### Types of changes

### Checklist:

### DIFF
--- a/plugins/mac-capture/mac-sck-audio-capture.m
+++ b/plugins/mac-capture/mac-sck-audio-capture.m
@@ -175,23 +175,6 @@ static bool init_audio_screen_stream(struct screen_capture *sc)
 
 static void sck_audio_capture_defaults(obs_data_t *settings)
 {
-    CGDirectDisplayID initial_display = 0;
-    {
-        NSScreen *mainScreen = [NSScreen mainScreen];
-        if (mainScreen) {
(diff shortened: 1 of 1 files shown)


PatTheMav (commented):
Did you test this with external displays and also headless modes (e.g. a Mac mini that has no display attached but is used via screen sharing)?

gxalpha:
I did test this with external displays. I did not test this headless (I don't have the required hardware), although I'm not aware that we're supporting headless mode.

PatTheMav:
> I did test this with external displays. I did not test this headless (I don't have the required hardware), although I'm not aware that we're supporting headless mode.

It's not true "headless" I guess, but I usually use a device without an active screen as the compositing machine and only use screen sharing or web sockets to set it up (there is an HDMI connection but it's not used).

PR #9246, opened 2023-07-13
linux-v4l2: Add all native V4L device selection methods

DarkDefender:
This is a continuation of: https://github.com/obsproject/obs-studio/pull/6493

@kkartaltepe @ggodlewski Sorry for the long wait, here is my clean up an minor rework of the old pull request.
I hope that the UI is now satisfactory. At least the artists I've talked to at my office seem fine with it.

### Description
Previously you could only select the devices via their "dev" file name in /dev (/sys/class/video4linux/ on linux). However these device file names are not consistent and depends on when the video source was detected by the kernel.

This means that users would have a really hard time to map a specific video source to a specific device as the file name the device would have is kinda random.

Thankfully V4L has native support for mapping devices via their hardware ID name or by which bus they are connected to.

By exposing this to the end user they will have a very robust and flexible way of mapping video sources and hot plugging them.

The old selection method is still available if anyone wants to still use it.

Screenshots of the new interface UI:

Devices listed by ID. The ID is their device name reported by the hardware. These should be unique but I have experienced some webcams where this wasn't the case. However in this scenario the "by bus path" option works nicely.

Devices listed by their bus path. This maps the device to their physical connection like the USB port they are connected to.
This will allow the user to easily replace a faulty webcam with a new one by plugging it into the same usb port.
(Or it can also be used if the devices do not have unique IDs)

"Other" is the old method that just looks for `video` devices and doesn't do any ID or path mapping:

### How Has This Been Tested?
I've tested this on my Gentoo Linux computer with 5 different webcams plugged into it.
I've also had some other people in my office provide some feedback on UI and naming.
As these changes are specific to this plugin, I don't think other areas will be affected by this change.

### Types of changes
- Bug fix (video sources not staying consistent between reboots or hotplugs)
- New feature

### Checklist:

I'm not sure about the last bullet point. Where is the V4L plugin documentation? I tried to search for it but couldn't find it.
It would be great if I could do a more through explanation of what the device selection types actually do to a end user.

### DIFF
--- a/plugins/linux-v4l2/data/locale/en-US.ini
+++ b/plugins/linux-v4l2/data/locale/en-US.ini
@@ -1,4 +1,8 @@
 V4L2Input="Video Capture Device (V4L2)"
+V4L.SelectionType="Device Detection"
+V4L.ById="By Device ID"
+V4L.ByPath="By Bus Path"
+V4L.Other="Other"
 Device="Device"
 Input="Input"
 VideoFormat="Video Format"
--- a/plugins/linux-v4l2/v4l2-helpers.c
+++ b/plugins/linux-v4l2/v4l2-helpers.c
@@ -318,3 +318,15 @@ int_fast32_t v4l2_set_dv_timing(int_fast32_t dev, int *timing)
 
 	return 0;
 }
+
+char *brealpath(const char *link_path)
+{
+	char *tmp_real_path = realpath(link_path, NULL);
+	if (NULL != tmp_real_path) {
+		char *real_path = bstrdup(tmp_real_path);
+		free(tmp_real_path);
+		return real_path;
+	} else {
+		return bstrdup(link_path);
+	}
+}
--- a/plugins/linux-v4l2/v4l2-helpers.h
+++ b/plugins/linux-v4l2/v4l2-helpers.h
@@ -319,6 +319,15 @@ int_fast32_t v4l2_enum_dv_timing(int_fast32_t dev, struct v4l2_dv_timings *dvt,
  */
 int_fast32_t v4l2_set_dv_timing(int_fast32_t dev, int *timing);
 
+/**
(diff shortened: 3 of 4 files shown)


GeorgesStavracas:
Thank you very much for the contribution and for the amount of effort put in this PR. I'm afraid I come with bad news though :( 

After deliberating about this feature, the interaction flow is not what we want to go as the intended user experience. What we want going forward is basically how the macOS integration in OBS Studio handles cameras. Each device gets a unique identifier - even devices of the same model - and that's what's used to address the device. That means that whatever device the user selects in OBS Studio is picked, regardless of where it is connected.

Picking the device of a specific bus might incur losing data, or other sorts of incompatibilities, as if the user swaps the device and the new device doesn't support the same properties of the old one.

As it is right now, I don't think the behaviour above is what our v4l2 code does. I think it should be what it does. Contributions that implement this behaviour would be welcomed.

Hope the user experience considerations that led to rejecting this feature were cleared up. Looking forward more PRs.

DarkDefender:
@GeorgesStavracas I'm a bit confused.

This PR does you said you want. IE the default option is that each camera gets it's own unique ID if possible and it doesn't matter which bus port they are plugged into. OBS will pick up the camera correctly with by its ID and you can hot plug it into which ever USB port you wish.
However there are some cases where assigning unique IDs are not possible (usually because the manufacture used the same serial number or ID tag for all devices of the same type).
In that case the user can choose to fall back to the bus address to be able to differentiate the devices if needed.

So what is implement is exactly what you want in addition some extra modes the user can manually choose to fallback to in case that is not good enough for them.

Which is why I am a bit confused because this PR implements features "Select by ID", "Select by Bus", and "Use old method that is currently in OBS which is neither".
However you are now closing this saying: We want feature "Select by ID", feel free to implement it if you want.

PR #9248, opened 2023-07-13
Fix defer encoder updates and NVENC VBR maxBitRate

flaeri:
### Description
MaxBitrate gets correctly set during encoder init. https://github.com/obsproject/obs-studio/blob/578dc46a797db8860e90789741bae40441d18575/plugins/obs-ffmpeg/obs-nvenc.c#L672

Gets promptly overwritten by nvenc_update, which set max = (avg)bitrate.

### Motivation and Context
I wish for the max bitrate value in the UI to be respect. There were at least 2 complaints.

### How Has This Been Tested?
Built on win10 22h2, it compiles. Using breakpoints in the debugger, I confirmed the contents of `maxBitRate` when running VBR and CBR. I checked the recordings pre and post this change, and confirmed the encoder now allows the bitrate to spike higher (I used 2500kbps bitrate, 10 000kbps max).

### Types of changes

### Checklist:

### DIFF
--- a/libobs/obs-encoder.c
+++ b/libobs/obs-encoder.c
@@ -505,12 +505,24 @@ void obs_encoder_update(obs_encoder_t *encoder, obs_data_t *settings)
 
 	obs_data_apply(encoder->context.settings, settings);
 
-	// Note, we don't actually apply the changes to the encoder here
-	// as it may be active in another thread. Setting this to true
-	// makes the changes apply at the next possible moment in the
-	// encoder / GPU encoder thread.
-	if (encoder->info.update)
+	// Encoder isn't initialized yet, only apply changes to settings
+	if (!encoder->context.data)
+		return;
+
+	// Encoder doesn't support updates
+	if (!encoder->info.update)
+		return;
+
+	// If the encoder is active we defer the update as it may not be
+	// reentrant. Setting reconfigure_requested to true makes the changes
+	// apply at the next possible moment in the encoder / GPU encoder
+	// thread.
+	if (encoder_active(encoder)) {
 		encoder->reconfigure_requested = true;
+	} else {
+		encoder->info.update(encoder->context.data,
+				     encoder->context.settings);
+	}
 }
 
 bool obs_encoder_get_extra_data(const obs_encoder_t *encoder,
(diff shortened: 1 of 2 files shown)


notr1ch:
I believe `nvenc_update` should only be called during a dynamic bitrate adjustment which doesn't apply for recording. It doesn't make sense to call it during initialization, perhaps that's something to fix too.

flaeri:
Well, I unfortunately cannot see where it gets called (not familiar enough with the system). I just observe that it does get called when you hit "start recording".

If I change the rate control (which is possible during recording) CBR->VBR, or VBR->CBR, it gets triggerd and gets inside the if can_change_bitrate.

If I alter the bitrate or the max bitrate, it fires as well, and gets set (at least as far as the breakpoint is concerned), and returns TRUE.

So, it is currently possible, and seems to be working, whether intentional or not. I don't feel comfortable vouching/PRing the removal of it, as I think its actually quite nice, and I would not be surprised if there are some people who make use of it.

notr1ch:
This actually seems like it might be my fault :). See if this helps: https://github.com/notr1ch/obs-studio/commit/c9ca1cdbab71fe76eaa0e3ee71f017591cc7143f

flaeri:
Neat :) Yes, that does indeed prevent the nvenc_update from trigger on record start. Changing bitrate still works as expected.

I will happily admit that I don't understand the full scale of impact to making changes like this in relation to context.data and info.update etc, so I don't have a good grasp on what needs testing. I can of course run thru all of the encoders I have, and check, but thats about it.

Do you want to PR your patch? Or do you want me to just force push with these changes? 

I don't think I am capable of writing a good commit message when I don't fully understand the scope of what I am commiting, hence my apprehension. I'm more than happy to do it, if thats what you'd like, and you give me a quick synopsis I can use.

I would also assume this should be two seperate commits, since it adresses two different things in two different areas, so yeah, any input/guidance would be helpful for progress.

SuslikV:
It seems that the one of the mentioned complaints was here: https://obsproject.com/forum/threads/limit-max-bitrate-in-cqp-mode.167660/post-616400

PR #9249, opened 2023-07-14
libobs-d3d11: Implement shader cache

notr1ch:
### Description
Implements a D3D shader cache. The shader script itself is hashed (using fnv1a for now as a quick hash) and if a file exists with a matching hash the compiled program is read from that instead of calling `D3DCompile` which is slow.

### Motivation and Context
A lot of time on OBS startup on Windows is spent compiling the same static shaders.

### How Has This Been Tested?
Ran OBS a bit and it worked. More testing across different drivers / vendors would be nice. We had a similar implementation of a shader cache in OBS Classic that seemed to always break things, but I have no idea why. Microsoft recommend precompiling and shipping compiled shaders as D3D compiles to [DXIL](https://github.com/microsoft/DirectXShaderCompiler/blob/main/docs/DXIL.rst) which is then further consumed by the GPU driver and compiled to native GPU code. So the D3D compiled shaders _should_ be entirely deterministic.

### Types of changes

### Checklist:

### DIFF
--- a/libobs-d3d11/d3d11-shader.cpp
+++ b/libobs-d3d11/d3d11-shader.cpp
@@ -21,6 +21,11 @@
 #include <graphics/vec3.h>
 #include <graphics/matrix3.h>
 #include <graphics/matrix4.h>
+#include <util/platform.h>
+#include <util/util.hpp>
+
+#include <filesystem>
+#include <fstream>
 
 void gs_vertex_shader::GetBuffersExpected(
 	const vector<D3D11_INPUT_ELEMENT_DESC> &inputs)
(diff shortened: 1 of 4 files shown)


libobs-d3d11/d3d11-shader.cpp
```diff
@@ -200,26 +201,71 @@ void gs_shader::BuildConstantBuffer()
...
-		if (errorsBlob != NULL && errorsBlob->GetBufferSize())
-			throw ShaderError(errorsBlob, hr);
-		else
-			throw HRError("Failed to compile shader", hr);
+	uint64_t hash = fnv1a_hash(shaderString);
+	snprintf(hashstr, sizeof(hashstr), "%02llx", hash);
+
+	dstr_init_move_array(&cache_path, os_get_program_data_path_ptr(
+						  "obs-studio/shader-cache"));
+	os_mkdirs(cache_path.array);
```
notr1ch, line 234: This should ideally be done somewhere else, but where?

derrod: Wheh running in portable mode, would this not still use appdata for the cache?

derrod: I suppose this could be done in `MakeUserDirs()` in `obs-app.cpp`, that should run before libobs-d3d11 is loaded. If this gets merged we should probably also add the clearing/removal of that folder to the NSIS/Steam install script like we do with `%PROGRAMDATA%\obs-studio-hook`.

libobs-d3d11/d3d11-shader.cpp
```diff
@@ -200,26 +201,71 @@ void gs_shader::BuildConstantBuffer()
...
+	} else {
+		hr = device->d3dCompile(shaderString, strlen(shaderString),
+					file, NULL, NULL, "main", target,
+					D3D10_SHADER_OPTIMIZATION_LEVEL3, 0,
+					shader, errorsBlob.Assign());
+		if (FAILED(hr)) {
+			if (errorsBlob != NULL && errorsBlob->GetBufferSize())
+				throw ShaderError(errorsBlob, hr);
+			else
+				throw HRError("Failed to compile shader", hr);
```
derrod, line 256: I think `cache_path` is not freed if this throws. I tried refactoring this with C++17 RAII stuff for funsies (https://github.com/derrod/obs-studio/commit/c94ca4d7bab6e53f89839753c311e90f0fac49a7), maybe some of that's a better option than using `dstr` here.

derrod:
If this gets merged we should probably also add the clearing/removal of the cache folder to the NSIS/Steam install script like we do with `%PROGRAMDATA%\obs-studio-hook`.

libobs-d3d11/d3d11-shader.cpp
notr1ch: I had considered that but thought it was bad design since there's no guarantee that libobs-d3d11 is the graphics backend that ends up being used. I suppose for now as it being the only supported one it's fine.

derrod: That is true, if we get back into the whole libobs vs UI thing then relying on the UI creating the folder may be considered bad practice as well, though in praxis it doesn't matter for us. Maybe there's a better place withing libobs-d3d11 itself during startup.

derrod: What about doing it in the `device_create` function like the logging of adapaters? https://github.com/derrod/obs-studio/commit/114bd0aa573f830a2c2644d5768497c3ab733a45

PR #9256, opened 2023-07-14
libobs: Enable fast clear on Windows with driver check

RytoEX:
### Description

On multi-adapter systems where at least one GPU is an NVIDIA GPU, using fast clear can cause the Properties window's preview pane to flicker rapidly if the NVIDIA driver is older than 527.37. This series of changes enables us to run a one-time check for if the graphics adapter can perform a fast clear, and if so, then always use that. The check is performed once and the result is stored for future use. If the adapter is determined to not be able to safely fast clear, continue to use the clear workaround.

### Motivation and Context

Want fast clear when possible. Don't want to try fast clear when it's not safe.

Follow-up to:
* #7601
* #7214
* #7177
* #7175

### How Has This Been Tested?

I tested OBS Studio 28.0.0-rc1 (a version where fast clear was universally enabled) on my Intel+NVIDIA laptop (Windows 10) with NVIDIA driver versions 522-528 to determine exactly which driver version fixes the issue. That testing determined that NVIDIA driver version 527.37 is the first driver version that is fixed (526.98 still flickers). I tested both driver versions while debugging in Visual Studio to confirm that `use_clear_workaround` remains `true` on driver version 526.98 and that it becomes `false` on driver version 527.37+.

### Types of changes

### Checklist:

### DIFF
--- a/libobs-d3d11/d3d11-subsystem.cpp
+++ b/libobs-d3d11/d3d11-subsystem.cpp
@@ -3593,3 +3593,8 @@ uint32_t gs_get_adapter_count(void)
 
 	return count;
 }
+
+extern "C" EXPORT bool device_can_adapter_fast_clear(gs_device_t *device)
+{
+	return device->fastClearSupported;
+}
--- a/libobs-d3d11/d3d11-subsystem.cpp
+++ b/libobs-d3d11/d3d11-subsystem.cpp
@@ -641,11 +641,28 @@ static bool CheckFormat(ID3D11Device *device, DXGI_FORMAT format)
 	       ((support & required) == required);
 }
 
+static bool FastClearSupported(UINT vendorId, uint64_t version)
+{
+	/* Always true for non-NVIDIA GPUs */
+	if (vendorId != 0x10de)
+		return true;
+
+	const uint16_t aa = (version >> 48) & 0xffff;
+	const uint16_t bb = (version >> 32) & 0xffff;
+	const uint16_t ccccc = (version >> 16) & 0xffff;
+	const uint16_t ddddd = version & 0xffff;
+
+	/* Check for NVIDIA driver version >= 31.0.15.2737 */
+	return aa >= 31 && bb >= 0 && ccccc >= 15 && ddddd >= 2737;
+}
+
 void gs_device::InitDevice(uint32_t adapterIdx)
 {
 	wstring adapterName;
 	DXGI_ADAPTER_DESC desc;
 	D3D_FEATURE_LEVEL levelUsed = D3D_FEATURE_LEVEL_10_0;
+	LARGE_INTEGER umd;
+	uint64_t driverVersion = 0;
 	HRESULT hr = 0;
 
 	adpIdx = adapterIdx;
@@ -662,6 +679,10 @@ void gs_device::InitDevice(uint32_t adapterIdx)
 	os_wcs_to_utf8_ptr(adapterName.c_str(), 0, &adapterNameUTF8);
 	blog(LOG_INFO, "Loading up D3D11 on adapter %s (%" PRIu32 ")",
 	     adapterNameUTF8.Get(), adapterIdx);
+
+	hr = adapter->CheckInterfaceSupport(__uuidof(IDXGIDevice), &umd);
+	if (SUCCEEDED(hr))
+		driverVersion = umd.QuadPart;
 
 	hr = D3D11CreateDevice(adapter, D3D_DRIVER_TYPE_UNKNOWN, NULL,
 			       createFlags, featureLevels,
@@ -725,6 +746,8 @@ void gs_device::InitDevice(uint32_t adapterIdx)
 	nv12Supported = CheckFormat(device, DXGI_FORMAT_NV12) &&
 			!HasBadNV12Output();
 	p010Supported = nv12Supported && CheckFormat(device, DXGI_FORMAT_P010);
+
+	fastClearSupported = FastClearSupported(desc.VendorId, driverVersion);
 }
 
 static inline void ConvertStencilSide(D3D11_DEPTH_STENCILOP_DESC &desc,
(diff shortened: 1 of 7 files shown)


libobs-d3d11/d3d11-subsystem.hpp
```diff
@@ -1002,8 +1002,15 @@ struct gs_device {
 	ComPtr<ID3D11Device> device;
 	ComPtr<ID3D11DeviceContext> context;
 	uint32_t adpIdx = 0;
+	string Name;
+	size_t DedicatedVideoMemory;
+	size_t SharedSystemMemory;
```
derrod, line 1007: These don't appear to be used for anything?

libobs-d3d11/d3d11-subsystem.cpp
```diff
@@ -454,6 +454,32 @@ try {
...
+
+bool gs_device::CheckFastClear()
+{
+	if (VendorId == 0x10de && IsNvidiaDriver527OrNewer(DriverVersion)) {
+		return true;
+	} else if (VendorId == 0x10de) {
+		return false;
+	} else {
+		return true;
+	}
```
derrod, line 480: Could be shortened to `return VendorId != 0x10de || IsNvidiaDriver527OrNewer(DriverVersion)`

libobs-d3d11/d3d11-subsystem.cpp
```diff
@@ -660,6 +687,19 @@ void gs_device::InitDevice(uint32_t adapterIdx)
 	blog(LOG_INFO, "Loading up D3D11 on adapter %s (%" PRIu32 ")",
 	     adapterNameUTF8.Get(), adapterIdx);
 
+	Name = adapterNameUTF8.Get();
+	DedicatedVideoMemory = desc.DedicatedVideoMemory;
+	SharedSystemMemory = desc.SharedSystemMemory;
+	VendorId = desc.VendorId;
+	DeviceId = desc.DeviceId;
```
derrod, line 694: Everything except `VendorId` appears unused. These + `DriverVersion` and `CheckFastClear()` probably don't need to be part of the strcut and could all just be variables/static functions since it's only ever evaluated when initialising.

libobs-d3d11/d3d11-subsystem.hpp
RytoEX: I think the were a holdover from previous experimentation. I'm indifferent if we keep or remove them.

libobs-d3d11/d3d11-subsystem.cpp
RytoEX: As mentioned in another comment, I think they were holdovers from previous experimentation (the branch itself is from about 4 months ago).

As for why they are part of the struct, the values should never change, so I felt they logically should be associated with the device on init. My recollection was that we've had other cases where we wished we could gate something on a driver version, so I wanted to enable that to be easily reusable for other cases.

That said, I'm open to alternatives here.

libobs-d3d11/d3d11-subsystem.cpp
RytoEX: That seems fine. Can change later.

libobs-d3d11/d3d11-subsystem.cpp
derrod: Until those are actually needed, they should probably just be left out to simplify the diff and save a couple bytes of memory.

RytoEX: Can try to simplify the diff some time later.

derrod: Here's what I'd cut it down to: https://github.com/derrod/obs-studio/commit/1cb357e1f64265ca38a694be2dd90de31e9dc060

jpark37:
I'm too lazy to think. I trust you guys.

libobs-d3d11/d3d11-subsystem.cpp
RytoEX: > Here's what I'd cut it down to: [derrod@1cb357e](https://github.com/derrod/obs-studio/commit/1cb357e1f64265ca38a694be2dd90de31e9dc060)

Used this implementation and squashed commits.

libobs-d3d11/d3d11-subsystem.cpp
RytoEX: Used your implementation suggested in the other review thread.

libobs-d3d11/d3d11-subsystem.hpp
RytoEX: Removed.

derrod (approved):
If somebody can test that this functionally works it should be good to go.

RytoEX:
> If somebody can test that this functionally works it should be good to go.

I double checked on my affected machine (laptop with Intel HD Graphics 630 + NVIDIA GeForce GTX 1060). With a simple test of adding a Media Source with no file selected, here is the behavior of the Properties preview with different driver and OBS versions.
| NVIDIA Driver Version | OBS Build  | Result                        |
| --------------------- | ---------- | ----------------------------- |
| 526.98                | 28.0.0-rc1 | :x: flicker                   |
| 526.98                | this patch | :white_check_mark: no flicker |
| 527.37                | 28.0.0-rc1 | :white_check_mark: no flicker |
| 527.37                | this patch | :white_check_mark: no flicker |

Using breakpoint inspection in Visual Studio, here are the values of `device->fastClearSupported` and `display->use_clear_workaround` with this patch on each driver version:
| NVIDIA Driver Version | `fastClearSupported` | `use_clear_workaround` |
| --------------------- | -------------------- | ---------------------- |
| 526.98                | `FALSE`              | `TRUE`                 |
| 527.37                | `TRUE`               | `FALSE`                |

I think you can also test this on desktop GPUs in any multi-GPU setup where the display that OBS is on is connected to a GPU that OBS itself is not running on, but the details are hazy.

Happy to do more tests since we know my machine is affected.

jpark37:
If you take renderdoc captures, you can verify it calls ClearRenderTargetView when the workaround is off and Draw when it's on.

RytoEX:
> If you take renderdoc captures, you can verify it calls ClearRenderTargetView when the workaround is off and Draw when it's on.

I suppose I'd have to learn how to use RenderDoc to do that.

Issue #7867, opened 2022-11-27
Intel QSV QuickSync H.264 CBR ignores target bitrate


JarodGu:
### Operating System Info

Windows 10

### OBS Studio Version

28.1.2

### OBS Studio Log URL

https://obsproject.com/logs/sLr1ezpFbDf0kOEa

### Expected Behavior

QSV encoder should respect the target bitrate when set below a threshold determined by the output resolution and framerate

### Current Behavior

OBS outputs video at around `4300Kbps` while streaming `1920x1080@120` after I lowered the CBR bitrate to `1800Kbps`.

QSV appears to enforce a minimum bitrate depending on the output resolution and framerate. Higher bitrate values take effect as expected. I was able to replicate the issue on CBR, VBR, and ICQ rate controls.

OBS Studio Log is from a local recording demonstrating the issue. It was also observed in a 3 hour livestream to Twitch. 
- Sample clip: https://clips.twitch.tv/RacyFurtiveGarageBrainSlug--hPP9JfqTGcZdP36
- Playback Info: 
![image](https://user-images.githubusercontent.com/43764205/204162937-34006ca4-7a7b-4599-ba96-aae60b9614c8.png)
- RTMP Header received by live-sea.twitch.tv
`map[2.1:false 3.1:false 4.0:false 4.1:false 5.1:false 7.1:false audiochannels:2 audiocodecid:10 audiodatarate:320 audiosamplerate:44100 audiosamplesize:16 duration:0 encoder:obs-output module (libobs version 28.1.0) fileSize:0 framerate:120 height:1080 stereo:true videocodecid:7 videodatarate:1700 width:1920]`
- Twitch Inspector Log: [inspector.pdf](https://github.com/obsproject/obs-studio/files/10099333/inspector.pdf)

### Steps to Reproduce

1. In the Advanced output menu, set QuickSync H.264 to CBR with an arbitrarily low bitrate like 100Kbps. If using ICQ, set ICQ quality to 51 (highest value, worse quality)
- Other settings used are target usage: slower, profile: high, keyframe interval: 4, latency: normal, bframes: 0, and no subjective video enhancements
![encodersettings](https://user-images.githubusercontent.com/43764205/204163063-f7581739-669f-48b7-8f37-0c4dde81c96d.png)
2. In the Video menu, set the output resolution to a desired value. I tested with 1920x1080@120fps, but each resolution and framerate has its own minimum bitrate
3. Start streaming or recording

### Anything else we should know?

[...]

PR #7704, opened 2022-11-01
obs-qsv11: Add AV1 changes

thyintel:
### Description

- Add Intel AV1 hardware encoder

### Motivation and Context

- We should have a way to use Intel AV1 hardware encoder in streaming and recording.

### How Has This Been Tested?

- Integrated graphics (to ensure the original code path works)
- Intel CPU and Intel Arc GPU
- Non-Intel and Intel Arc GPU

### Types of changes

- New feature (non-breaking change which adds functionality)

### Checklist:

- [x] My code has been run through [clang-format](https://github.com/obsproject/obs-studio/blob/master/.clang-format).
- [x] I have read the [**contributing** document](https://github.com/obsproject/obs-studio/blob/master/CONTRIBUTING.rst).
- [x] My code is not on the master branch.
- [x] The code has been tested.
- [x] All commit messages are properly formatted and commits squashed where appropriate.
- [x] I have included updates to all appropriate documentation.

PR #9258, opened 2023-07-15
obs-qsv11: Tune and clean up features

thyintel:
### Description

1. Improved AV1 quality by enabling bidirectional predictions for AV1 on newer Intel platforms and enabling content light meta data over the AV1 stream
2. Refactored look-ahead by enabling low power look-ahead for newer Intel platforms, changed the look-ahead depth for latencies, and disable VME look-ahead for older Intel platforms and HEVC
3. Refactored BRC by disabling AVBR, QVBR, and VCM
4. Enhanced video quality by disabling MBBRC and using AdaptiveCQM for CBR/VBR
5. Tuned bitstream by changing GopPicSize and removing GopOptFlag
6. Bug Fixes for InitBufferSize and BufferSize
7. Cleaned up redundant target usage settings

### Motivation and Context

- Improves recording and streaming for HEVC and AV1 codecs on Intel platforms 
- Bug fixes for rate control and look-ahead support on Intel platforms

### How Has This Been Tested?

Tested QSV encoder for AVC, HEVC, and AV1 recording for the following platforms:
- Kaby Lake (7th generation core)
- Coffee Lake (9th generation core)
- Ice Lake (10th generation core)
- Tiger Lake (11th generation core)
- Alder Lake (12th generation core)
- Raptor Lake (13th generation core)
- Meteor Lake (14th generation core)
- Raptor Lake + Intel Arc Graphics

Tested Gameplay Capture (CSGO, Fallout 4, and GTA 5) for AVC, HEVC, and AV1 recording for the following platforms:
- Raptor Lake (13th generation core)
- Meteor Lake (14th generation core)
- Raptor Lake + Intel Arc Graphics

### Types of changes

- Tweak (non-breaking change to improve existing functionality) -->
- Performance enhancement (non-breaking change which improves efficiency) -->
- Code cleanup (non-breaking change which makes code smaller or more readable) -->

### Checklist:


plugins/obs-qsv11/obs-qsv11.c
```diff
@@ -520,26 +551,20 @@ static void update_params(struct obs_qsv *obsqsv, obs_data_t *settings)
 
 	int width = (int)obs_encoder_get_width(obsqsv->encoder);
 	int height = (int)obs_encoder_get_height(obsqsv->encoder);
-	if (astrcmpi(target_usage, "quality") == 0)
+	if (astrcmpi(target_usage, "TU1: Slowest (Best Quality)") == 0)
```
kkartaltepe, line 557: This is not safe for users with config using the old names.

kkartaltepe: Woops i see the upgrade code above

kkartaltepe:
previously there were some version checks against the mfx version restricting some code branches. If someone of these depend on bug fixes/api features in vpl can we retain checks against the ubuntu 22 version (2022.1.0), which is likely behind the new VPL version imported.

RytoEX:
> previously there were some version checks against the mfx version restricting some code branches. If some of these depend on bug fixes/api features in vpl can we retain checks against the ubuntu 22 version (2022.1.0), which is likely behind the new VPL version imported for windows.

Echoing this statement above. Ubuntu 22.04 has VPL 2.6 (2022.1.0), which we cannot control. Since #8327 enables this plugin on Ubuntu 22.04+, we'll need to account for the differences between VPL versions on Ubuntu and Windows if needed.

norihiro:
Please check the commit guideline.
https://github.com/obsproject/obs-studio/blob/master/CONTRIBUTING.rst#commit-guidelines
- Make sure commit titles are always in present tense.
- Other than the first word should not be capitalized.

plugins/obs-qsv11/obs-qsv11.c
```diff
@@ -301,36 +319,57 @@ static bool update_enhancements(obs_data_t *settings)
 	}
 
 	if (update) {
-		bool enabled = (mbbrc && cqm);
+		bool enabled = (cqm);
```
RytoEX, line 322: As far as I understand, this is meant to keep Subjective Video Enhancements disabled if a user just upgraded from a version of OBS that had MBBRC and CQM settings set and they had previously disabled either MBBRC or CQM, while erasing both old settings. This can probably be left as-is.

RytoEX (changes requested):

RytoEX:
> Please check the commit guideline. https://github.com/obsproject/obs-studio/blob/master/CONTRIBUTING.rst#commit-guidelines
> * Make sure commit titles are always in present tense.
> * Other than the first word should not be capitalized.

I've made these changes and rebased the PR.

It does seem like it requires the new VPL headers though to fix the compile errors.
* mfxExtCodingOption3.AdaptiveCQM requires VPL 2.2 (v2021.2.2)
* MFX_PLATFORM_DG2 requires VPL requires VPL 2.6 (v2022.1.0), though it is deprecated in VPL 2.9

As VPL 2.6 is available on Ubuntu 22.04, this should be fine. @kkartaltepe do you have any specific concerns for version guards for anything else?

kkartaltepe:
> As VPL 2.6 is available on Ubuntu 22.04, this should be fine. @kkartaltepe do you have any specific concerns for version guards for anything else?

Nah if you have looked at it and nothing important was guarded that sounds good to me.

plugins/obs-qsv11/obs-qsv11.c
RytoEX: I've made this change and force pushed.

maginhart:
image
I've tried OBS 29.1.0-518-g0e75e7722 build and QSV fails to record with CBR/VBR (H264/AV1/HEVC) and for some reason CBR/VBR shows ICQ Quality selectable value field.
ICQ/CQP works 
https://obsproject.com/logs/hvQ5VqWpOB6L3ODq

RytoEX:
> I've tried OBS 29.1.0-518-g0e75e7722 build and QSV fails to record on Intel Arc A380 with CBR/VBR (H264/AV1/HEVC) and for some reason CBR/VBR shows ICQ Quality selectable value field.

I can reproduce this on Arc A770M and Iris Xe. Will debug.

RytoEX:
> I've tried OBS 29.1.0-518-g0e75e7722 build and QSV fails to record on Intel Arc A380 with CBR/VBR (H264/AV1/HEVC) and for some reason CBR/VBR shows ICQ Quality selectable value field.

This appears to be failing because `QSV_Encoder_Internal::InitParams` returns `MFX_ERR_UNDEFINED_BEHAVIOR` (which is returned from `m_pmfxENC->Query(&m_mfxEncParams, &validParams)`. Before https://github.com/obsproject/obs-studio/commit/319eac4bf51b6d2a1a447351d66ab5ee56974398 (#7704), we ignored this error code and simply changed the LowPower and LookAheadDepth settings in an attempt to set valid parameters. However, that commit/PR added an explicit status check for InitParams:
https://github.com/obsproject/obs-studio/commit/319eac4bf51b6d2a1a447351d66ab5ee56974398#diff-75e9eb48c657fc386ab4b2ccd37672a16b74691a0923799778b4673021ec752bR170-R171

RytoEX:
> I've tried OBS 29.1.0-518-g0e75e7722 build and QSV fails to record on Intel Arc A380 with CBR/VBR (H264/AV1/HEVC) and for some reason CBR/VBR shows ICQ Quality selectable value field.

@maginhart I believe I've fixed the specific issue with QSV failing to record in the most recent force-push. Please try again.

RytoEX:
> I've tried OBS 29.1.0-518-g0e75e7722 build and QSV fails to record on Intel Arc A380 with CBR/VBR (H264/AV1/HEVC) and for some reason CBR/VBR shows ICQ Quality selectable value field.

@maginhart I've also just fixed the issue with CBR/VBR showing the "ICQ Quality" property.

kkartaltepe:
Discussed some issues with usage of deprecated fields, but otherwise this seems good to go on my system (a750, libvpl 2023.3)

kkartaltepe (approved):

maginhart:
> > I've tried OBS 29.1.0-518-g0e75e7722 build and QSV fails to record on Intel Arc A380 with CBR/VBR (H264/AV1/HEVC) and for some reason CBR/VBR shows ICQ Quality selectable value field.
> 
> @maginhart I've also just fixed the issue with CBR/VBR showing the "ICQ Quality" property.

Yeah everything seems to be working now.

maginhart:
The only thing I noticed is that if I set key frame interval to 2 seconds and I switch capture source on and off which creates a bitrate spike in a visually demanding game. The bitrate spike stays far longer than the keyframe interval. I've recorded a clip with stats windows visible.

https://drive.google.com/file/d/1W-BMeTL24i0N3yYt2FPzcVvzFYvs0WJv/view?usp=sharing

https://obsproject.com/logs/U0hlxRsseWaOqJOU

maginhart:
Maybe max gop size is set higher? Also after the spikes hanging there for awhile you can see that the whole picture starts to pulsate.

RytoEX:
> The only thing I noticed is that if I set key frame interval to 2 seconds and I switch capture source on and off which creates a bitrate spike in a visually demanding game. The bitrate spike stays far longer than the keyframe interval. I've recorded a clip with stats windows visible.

> Maybe max gop size is set higher? Also after the spikes hanging there for awhile you can see that the whole picture starts to pulsate.

Setting keyframe interval to 2 seconds should result in a GopPicSize of `2 * fpsNum / fpsDen` both before and after this PR.

RytoEX:
I've force-pushed what is hopefully the last set of changes to appease MSVC and clang-format.

> The only thing I noticed is that if I set key frame interval to 2 seconds and I switch capture source on and off which creates a bitrate spike in a visually demanding game. The bitrate spike stays far longer than the keyframe interval. I've recorded a clip with stats windows visible.
> 
> https://drive.google.com/file/d/1W-BMeTL24i0N3yYt2FPzcVvzFYvs0WJv/view?usp=sharing
> 
> https://obsproject.com/logs/U0hlxRsseWaOqJOU

Does this also happen without this PR? The only thing that stands out to me is the removal of MFX_GOP_STRICT.

maginhart:
> I've force-pushed what is hopefully the last set of changes to appease MSVC and clang-format.
> 
> > The only thing I noticed is that if I set key frame interval to 2 seconds and I switch capture source on and off which creates a bitrate spike in a visually demanding game. The bitrate spike stays far longer than the keyframe interval. I've recorded a clip with stats windows visible.
> > https://drive.google.com/file/d/1W-BMeTL24i0N3yYt2FPzcVvzFYvs0WJv/view?usp=sharing
> > https://obsproject.com/logs/U0hlxRsseWaOqJOU
> 
> Does this also happen without this PR? The only thing that stands out to me is the removal of MFX_GOP_STRICT.

No it only happens with this PR

RytoEX:
> > I've force-pushed what is hopefully the last set of changes to appease MSVC and clang-format.
> > > The only thing I noticed is that if I set key frame interval to 2 seconds and I switch capture source on and off which creates a bitrate spike in a visually demanding game. The bitrate spike stays far longer than the keyframe interval. I've recorded a clip with stats windows visible.
> > > https://drive.google.com/file/d/1W-BMeTL24i0N3yYt2FPzcVvzFYvs0WJv/view?usp=sharing
> > > https://obsproject.com/logs/U0hlxRsseWaOqJOU
> > 
> > 
> > Does this also happen without this PR? The only thing that stands out to me is the removal of MFX_GOP_STRICT.
> 
> No it only happens with this PR.

Could you screenshot the QSV encoder settings? We unfortunately do not log all of them.

maginhart:
Settings with this PR?

RytoEX:
> Settings with this PR?

Both with this PR and without this PR would be ideal.

maginhart:
**With this PR**
image

**Without this PR**
image

RytoEX:
> **With this PR** image
> 
> **Without this PR** image

Could you try disabling Subjective Video Enhancements with this PR to get a closer match to the settings without this PR?

maginhart:
> > **With this PR** image
> > **Without this PR** image
> 
> Could you try disabling Subjective Video Enhancements with this PR to get a closer match to the settings without this PR?

Sure I've edited the comment.

ma3uk:
> > Я попробовал сборку OBS 29.1.0-518-g0e75e7722, и QSV не может записать на Intel Arc A380 с CBR / VBR (H264 / AV1 / HEVC), и по какой-то причине CBR / VBR показывает поле выбора качества ICQ.
> 
> Я считаю, что исправил конкретную проблему, из-за которой QSV не мог записывать в последнем форс-толчке. Пожалуйста, попробуйте еще раз.

By doing this, you simply removed the connection of the CO3 buffer to Extparams, which will cause ScenarioInfo not to be applied to the encoding. Therefore, co3.ScenarioInfo can also be deleted.

The encoding error with ScenarioInfo GAME_STREAMING is a floating problem that appears periodically and disappears in the same way, perhaps it is related to drivers. 

 The problem is that ScenarioInfo GAME_STREAMING is a trigger for Lookahead to work on ARC GPU and higher. There will be no coding error when installing lookaheaddepth, with ScenarioInfo turned off, but it is also not guaranteed that Lookahead will work.

maginhart:
Okay and by the way lookahead depth doesn't show up in log file with this PR i guess it doesn't work.

maginhart:
With PR
https://obsproject.com/logs/e2tuHnnwLcW7wHgi
https://drive.google.com/file/d/1XAoBOf7krcBVnVovxh1KiOLuPwydDlmF/view?usp=sharing


Without PR
https://obsproject.com/logs/E0j0brdkyELiapj5
https://drive.google.com/file/d/1WEM9i-cAelhDY2ja_QDXeT3WHi-g8m_l/view?usp=sharing
Lookahead is logged without this PR

RytoEX:
> With PR https://obsproject.com/logs/e2tuHnnwLcW7wHgi https://drive.google.com/file/d/1XAoBOf7krcBVnVovxh1KiOLuPwydDlmF/view?usp=sharing
> 
> Without PR https://obsproject.com/logs/E0j0brdkyELiapj5 https://drive.google.com/file/d/1WEM9i-cAelhDY2ja_QDXeT3WHi-g8m_l/view?usp=sharing

Thanks for these.

> Lookahead is logged without this PR

Ah, I see the removal now. Perhaps we can restore that if it makes sense.

ma3uk:
> > С пиаром https://obsproject.com/logs/e2tuHnnwLcW7wHgi https://drive.google.com/file/d/1XAoBOf7krcBVnVovxh1KiOLuPwydDlmF/view?usp=sharing
> > Без пиара https://obsproject.com/logs/E0j0brdkyELiapj5 https://drive.google.com/file/d/1WEM9i-cAelhDY2ja_QDXeT3WHi-g8m_l/view?usp=sharing
> 
> Спасибо за это.
> 
> > Lookahead регистрируется без этого PR
> 
> , теперь я вижу удаление. Возможно, мы сможем восстановить это, если это будет иметь смысл.

The operation of lookahead can be tracked by the number of dedicated Surfaces, 1 Surface should be allocated for each frame of Lookahead.
You may encounter a problem that ScenarioInfo = MFX_SCENARIO_GAME_STREAMING will give an Undefined behavior or Null PTR error, and Lookahead will not work without it. The solution may be the parameter CO2.ExtBRC = MFX_CODINGOPTION_ON, it will give a 100% workable Lookahead and a bonus will improve the quality of the output video due to the internal EncTools library, but ExtBRC with the AVC codec is not stable in OBS for some reason - after some time of recording or broadcasting, it begins to produce duplicates of the I frame, which are detected in OBS itself as missed frames, I have been trying to solve this problem for quite some time, but to no avail, perhaps you know what the reason may be.

RytoEX:
> With PR https://obsproject.com/logs/e2tuHnnwLcW7wHgi https://drive.google.com/file/d/1XAoBOf7krcBVnVovxh1KiOLuPwydDlmF/view?usp=sharing
> 
> Without PR https://obsproject.com/logs/E0j0brdkyELiapj5 https://drive.google.com/file/d/1WEM9i-cAelhDY2ja_QDXeT3WHi-g8m_l/view?usp=sharing Lookahead is logged without this PR

It appears that both video files (with and without the PR) are missing CBR filler data, so that hasn't changed with this PR. Both video files also exhibit a bitrate spike under the same circumstances. That being the case, I'm unlikely to block on this, but I have raised the issue to Intel.

RytoEX:
> > Lookahead is logged without this PR
> 
> Ah, I see the removal now. Perhaps we can restore that if it makes sense.

I have restored "Lookahead Depth" logging.

maginhart:
Yes bitrate spikes happen and are completely normal what is not normal is how with this PR the bitrate spike holds far longer than the 2 seconds keyframe interval. Which mean that on twitch the stream source quality will be removed and stream video will be unavailable. Already tested that.

RytoEX:
> Yes bitrate spikes happen and are completely normal what is not normal is how with this PR the bitrate spike holds far longer than the 2 seconds keyframe interval. Which means that on twitch the stream source quality will be removed and stream video will be unavailable. Already tested that.

Does it still happen if you are streaming at 6000 Kbps instead of 8100 Kbps? 8100 Kbps is already above Twitch's maximum bitrate recommendation/allowance. Any bitrate variance at your target bitrate is bound to exacerbate any already existing issues.

For what it's worth, I just spent the past hour testing this with OBS Studio 29.1.3 with your settings, and I managed to lose Source quality with QSV H.264 with OBS Studio 29.1.3, so I do not think this PR is to blame. QSV CBR has high variance before and after this PR. Additionally, Source quality eventually returned after the bitrate stabilized enough, both with and without this PR.

Twitch Inspector says:
> The average bitrate of the stream is too high. This can cause buffering for low bandwidth viewers. Source could also be removed if bitrate is too high.

With a high target bitrate and a high bitrate variance, you'll likely hit this penalty. Our recommendation would be to keep the target bitrate within Twitch's guidelines. Anything higher than that is unsupported.

The bitrate variance in CBR is still a concern, but I do not believe it should block this PR.

maginhart:
You can get a high spike with 6000 bitrate. I've streamed with x264/ nvenc and never had problem with streaming at 8000 bitrate of course it may also depend on the selected twitch ingest server. I've streamed with old QSV version with obs 29.1.3 and never had a problem with source video disappearing or key frame inconsistency.

maginhart:
I think you shouldn't block this PR but it would be interesting to hear from intel on this matter.

RytoEX:
> I've streamed with x264/ nvenc and never had problem with streaming at 8000 bitrate of course it may also depend on the selected twitch ingest server. I've streamed with old QSV version with obs 29.1.3 and never had a problem with source video disappearing or key frame inconsistency.

x264 and NVENC likely have less variance in their CBR output.

RytoEX (approved):
Seems fine in testing. Pre-existing issues with high bitrate variance for CBR have been escalated to Intel.

maginhart:
High spikes are similar I was getting 10-12k spikes but twitch doesn't care about that if you are under 8500 bitrate average and you have strict keyframe interval (2 seconds or less)

maginhart:
> > Yes bitrate spikes happen and are completely normal what is not normal is how with this PR the bitrate spike holds far longer than the 2 seconds keyframe interval. Which means that on twitch the stream source quality will be removed and stream video will be unavailable. Already tested that.
> 
> Does it still happen if you are streaming at 6000 Kbps instead of 8100 Kbps? 8100 Kbps is already above Twitch's maximum bitrate recommendation/allowance. Any bitrate variance at your target bitrate is bound to exacerbate any already existing issues.
> 
> For what it's worth, I just spent the past hour testing this with OBS Studio 29.1.3 with your settings, and I managed to lose Source quality with QSV H.264 with OBS Studio 29.1.3, so I do not think this PR is to blame. QSV CBR has high variance before and after this PR. Additionally, Source quality eventually returned after the bitrate stabilized enough, both with and without this PR.
> 
> Twitch Inspector says:
> 
> > The average bitrate of the stream is too high. This can cause buffering for low bandwidth viewers. Source could also be removed if bitrate is too high.
> 
> With a high target bitrate and a high bitrate variance, you'll likely hit this penalty. Our recommendation would be to keep the target bitrate within Twitch's guidelines. Anything higher than that is unsupported.
> 
> The bitrate variance in CBR is still a concern, but I do not believe it should block this PR.

I've tried with 6000 Kbps and bitrate also spikes up pretty high 10500 kbps and stays in range of 8700-8900 keyframe interval seems to be ignored.
https://obsproject.com/logs/369WSicWSR9RYpBk
https://drive.google.com/file/d/1oIXoaCqk7xxJIIf2YJ07H1GkJDsJaG-M/view?usp=sharing

Settings used
image

RytoEX:
@maginhart See, for example, #7867 for prior posts about issues with QSV CBR variance and target bitrate adherence/compliance.

In any case, in my test streams, I did not notice that the CBR variance was notably worse after this PR. In fact, with this PR, the average bitrate for the session was closer to the target bitrate. I did notice one higher peak than without this PR, but one observance is hardly sufficient testing to draw a solid conclusion. As mentioned, I've escalated the issue of CBR variance in general to Intel.

plugins/obs-qsv11/QSV_Encoder_Internal.cpp
```diff
@@ -257,110 +256,114 @@ mfxStatus QSV_Encoder_Internal::InitParams(qsv_param_t *pParams,
...
+	memset(&m_co3, 0, sizeof(mfxExtCodingOption3));
+	m_co3.Header.BufferId = MFX_EXTBUFF_CODING_OPTION3;
+	m_co3.Header.BufferSz = sizeof(m_co3);
+
+	if (pParams->nbFrames > 1)
+		m_co2.BRefType = MFX_B_REF_PYRAMID;
+
+	PRAGMA_WARN_PUSH
+	PRAGMA_WARN_DEPRECATION
+#pragma warning(disable : 4996)
```
tytan652, line 329:
```suggestion
```

Define the pragma for MSVC in `libobs/util/c99defs.h`

```c
#define PRAGMA_WARN_DEPRECATION __pragma(warning(disable:4996))
```

plugins/obs-qsv11/QSV_Encoder_Internal.cpp
```diff
@@ -257,110 +256,114 @@ mfxStatus QSV_Encoder_Internal::InitParams(qsv_param_t *pParams,
...
+	MFXVideoCORE_QueryPlatform(m_session, &platform);
 #if defined(_WIN32)
-	// TODO: Why isn't LowPower coding supported on VAAPI backend.
-	enum qsv_cpu_platform qsv_platform = qsv_get_cpu_platform();
-	if ((m_isDGPU || qsv_platform >= QSV_CPU_PLATFORM_ICL ||
-	     qsv_platform == QSV_CPU_PLATFORM_UNKNOWN) &&
-	    (pParams->nbFrames == 0) &&
-	    (m_ver.Major == 1 && m_ver.Minor >= 31)) {
+#pragma warning(push)
+#pragma warning(disable : 4996)
```
tytan652, line 265: Use `PRAGMA_` macros defines `libobs/util/c99defs.h` even if it's Windows-only.

plugins/obs-qsv11/QSV_Encoder_Internal.cpp
```diff
@@ -257,110 +256,114 @@ mfxStatus QSV_Encoder_Internal::InitParams(qsv_param_t *pParams,
...
+		if (platform.CodeName >= MFX_PLATFORM_DG2)
+			m_mfxEncParams.mfx.LowPower = MFX_CODINGOPTION_ON;
+	} else if (codec == QSV_CODEC_AV1) {
 		m_mfxEncParams.mfx.LowPower = MFX_CODINGOPTION_ON;
-		if (pParams->nRateControl == MFX_RATECONTROL_LA_ICQ ||
-		    pParams->nRateControl == MFX_RATECONTROL_LA_HRD ||
-		    pParams->nRateControl == MFX_RATECONTROL_LA)
-			pParams->nRateControl = MFX_RATECONTROL_VBR;
 	}
+#pragma warning(pop)
```
tytan652, line 272: Use `PRAGMA_` macros defines `libobs/util/c99defs.h` even if it's Windows-only.

tytan652 (changes requested):

plugins/obs-qsv11/QSV_Encoder_Internal.cpp
RytoEX: Fixed this in #9392.

plugins/obs-qsv11/QSV_Encoder_Internal.cpp
RytoEX: Fixed.

plugins/obs-qsv11/QSV_Encoder_Internal.cpp
RytoEX: Fixed.

RytoEX (approved):

RytoEX:
@maginhart I want to let you know that I will be chasing down the CBR issues, but I'll pursue that as a separate bug fix effort.

Issue #9262, opened 2023-07-16
Build failure with qt5


jmarca:
### Operating System Info

Other

### Other OS

Slackware Current

### OBS Studio Version

Git

### OBS Studio Log URL

NA

### Expected Behavior

I expected the program to compile cleanly when I chose -DQT_VERSION=5 in the cmake options, as per the build instructions.

### Current Behavior

Build fails:

```
AILED: UI/frontend-plugins/decklink-output-ui/CMakeFiles/decklink-output-ui.dir/__/__/slider-ignorewheel.cpp.o
[...]

PR #9263, opened 2023-07-16
cmake: Remove Qt 5 support

gxalpha:
### Description

Removes cmake code to support Qt version selection, and Qt 5 in general.
Removes libobs ifdefs for the Qt 5 plugin check, as well as a remaining UI ifdef that used `OBS_QT_VERSION` instead of `QT_VERSION_CHECK`.

### Motivation and Context

Our code requires Qt 6 now, so having cmake for finding Qt 5 does not make a lot of sense, and can be confusing for people who only have Qt 5 and then run into issues when compiling time vs when cmake is generating.
Fixes https://github.com/obsproject/obs-studio/issues/9262.

### How Has This Been Tested?

macOS 14, generated and compiled with Qt 6.
Had CI build successfully on all supported platforms (see [here](https://github.com/gxalpha/obs-studio/actions/runs/5568706718)). Re-added Ubuntu 20.04 and saw it fail to find Qt 6 (see [here](https://github.com/gxalpha/obs-studio/actions/runs/5569042122/jobs/10172205492)).

### Types of changes

### Checklist:

### DIFF
--- a/CMakePresets.json
+++ b/CMakePresets.json
@@ -71,7 +71,6 @@
       "cacheVariables": {
         "ENABLE_WAYLAND": true,
         "ENABLE_VLC": true,
-        "QT_VERSION": "6",
         "CMAKE_BUILD_TYPE": "Debug"
       }
     },
@@ -108,7 +107,6 @@
       "cacheVariables": {
         "ENABLE_WAYLAND": true,
         "ENABLE_VLC": true,
-        "QT_VERSION": {"type": "STRING", "value": "6"},
         "CMAKE_BUILD_TYPE": {"type": "STRING", "value": "Debug"}
       }
     },
--- a/UI/cmake/legacy.cmake
+++ b/UI/cmake/legacy.cmake
@@ -82,7 +82,7 @@ set_target_properties(
              AUTORCC ON
              AUTOUIC_SEARCH_PATHS "forms;forms/source-toolbar")
 
-if(_QT_VERSION EQUAL 6 AND OS_WINDOWS)
+if(OS_WINDOWS)
   set_target_properties(obs PROPERTIES AUTORCC_OPTIONS "--format-version;1")
 endif()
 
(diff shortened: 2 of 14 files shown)


RytoEX (commented):
Looks okay at a glance. One question.

PatTheMav:
If there is no need to differentiate between different Qt versions anymore, `find_qt` should be removed entirely and a native finder be used instead. Any Qt7 release should be far out enough and _if_ it comes around a new solution needs to be found anyway.

This would require updating `obs-browser`, and `obs-websocket` first, but they can independently use `find_package` already without `find_qt` removed, so the removal can happen after that.

gxalpha:
@PatTheMav seeing the required submodule changes, would you consider that to be in scope of this PR, or is that meant as a future cleanup?

PatTheMav:
> @PatTheMav seeing the required submodule changes, would you consider that to be in scope of this PR, or is that meant as a future cleanup?

That depends on whether we agree to remove the `find_qt` function entirely instead of removing just the part of its original purpose.

gxalpha:
> That depends on whether we agree to remove the find_qt function entirely instead of removing just the part of its original purpose.

That's probably your call. That said I think we should make sure that this PR gets merged before beta, so that packagers don't run into what the linked issue encountered (cmake finding Qt 5 and being happy, and then the program not compiling).

PatTheMav:
> > That depends on whether we agree to remove the find_qt function entirely instead of removing just the part of its original purpose.
> 
> That's probably your call. That said I think we should make sure that this PR gets merged before beta, so that packagers don't run into what the linked issue encountered (cmake finding Qt 5 and being happy, and then the program not compiling).

Right, in that case it should probably be fine as-is. The other stuff can be cleaned up separately then.

RytoEX (approved):
I'd like to get this fixed quickly. Further cleanups can follow-up, hopefully soon after.

RytoEX:
As far as I'm aware, @PatTheMav has approved this off-thread. Merging.

jmarca:
Thanks for this fix. This is exactly the kind of thing that would have helped me—full stop during cmake

PR #9271, opened 2023-07-19
CI: Fix Qt XML validator attempting to validate non-XML files

PatTheMav:
### Description
Explicitly check only for changes `.ui` files in forms subdirectory and pass only changed files with that suffix to validator.

### Motivation and Context
Fix validator being triggered by any changes to files in the `forms` subdirectory, which will inevitably lead to such files not passing validation.

### How Has This Been Tested?
Needs to be tested on CI.

### Types of changes

### Checklist:

### DIFF
--- a/.github/actions/qt-xml-validator/action.yaml
+++ b/.github/actions/qt-xml-validator/action.yaml
@@ -44,13 +44,14 @@ runs:
         : Validate XML 💯
         if [[ "${RUNNER_DEBUG}" ]]; then set -x; fi
         shopt -s extglob
+        shopt -s globstar
 
-        changes=($(git diff --name-only HEAD~1 HEAD -- UI/forms))
+        changes=($(git diff --name-only HEAD~1 HEAD -- UI/forms/**/*.ui))
         case "${GITHUB_EVENT_NAME}" in
-          pull_request) changes=($(git diff --name-only origin/"${GITHUB_BASE_REF}" HEAD -- UI/forms)) ;;
+          pull_request) changes=($(git diff --name-only origin/"${GITHUB_BASE_REF}" HEAD -- UI/forms/**/*.ui)) ;;
           push)
             if [[ "${GITHUB_EVENT_FORCED}" == false ]]; then
-              changes=($(git diff --name-only ${GITHUB_REF_BEFORE} HEAD -- UI/forms))
+              changes=($(git diff --name-only ${GITHUB_REF_BEFORE} HEAD -- UI/forms/**/*.ui))
             fi
             ;;
           *) ;;


tytan652:
You should also check if the file was not removed otherwise it check a non-existent file.

PatTheMav:
> You should also check if the file was removed otherwise it checks a non-existent file.

Deletions aren't listed by `git diff --name-only` in my checks:

```
      --name-only
           Show only names of changed files
```

A deletion is not a "change" in that interpretation.

RytoEX:
> You should also check if the file was removed otherwise it checks a non-existent file.

The old workflow didn't make this distinction and I don't recall it having a problem with checking removed files. The files are gone, so they shouldn't be able to be globbed.
https://github.com/obsproject/obs-studio/blob/c33fa8bfa2776712e1a56270fdc97a25a4ddeded/.github/workflows/qt-xml.yml

My priority is getting CI working for PRs that add files into `UI/forms`.

@PatTheMav This seems to require both `UI/forms/*.ui` and `UI/forms/**/*.ui`. I checked on a local branch that added a new file in `UI/forms` itself, and this syntax did not pick it up.
https://github.com/obsproject/obs-studio/blob/c33fa8bfa2776712e1a56270fdc97a25a4ddeded/.github/workflows/qt-xml.yml#L28-L30

PatTheMav:
> > You should also check if the file was removed otherwise it checks a non-existent file.
> 
> The old workflow didn't make this distinction and I don't recall it having a problem with checking removed files. The files are gone, so they shouldn't be able to be globbed. https://github.com/obsproject/obs-studio/blob/c33fa8bfa2776712e1a56270fdc97a25a4ddeded/.github/workflows/qt-xml.yml
> 
> My priority is getting CI working for PRs that add files into `UI/forms`.
> 
> @PatTheMav This seems to require both `UI/forms/*.ui` and `UI/forms/**/*.ui`. I checked on a local branch that added a new file in `UI/forms` itself, and this syntax did not pick it up.
> 
> https://github.com/obsproject/obs-studio/blob/c33fa8bfa2776712e1a56270fdc97a25a4ddeded/.github/workflows/qt-xml.yml#L28-L30

Works fine here:

```
git diff --name-only master HEAD -- UI/forms/**/*.ui 

UI/forms/AutoConfigFinishPage.ui
```

It _does_ seem to require `globstar` in Bash though (Zsh is more capable), which I added just now.

PatTheMav:
TEST commit broke a UI file on purpose, was correctly picked up.

tytan652:
> > You should also check if the file was removed otherwise it checks a non-existent file.
> 
> The old workflow didn't make this distinction and I don't recall it having a problem with checking removed files. The files are gone, so they shouldn't be able to be globbed. https://github.com/obsproject/obs-studio/blob/c33fa8bfa2776712e1a56270fdc97a25a4ddeded/.github/workflows/qt-xml.yml

This CI run proves that something is wrong a file is removed.
https://github.com/obsproject/obs-studio/actions/runs/5591042547/jobs/10221574362

The file no longer exist in the `UI/forms` folder

> My priority is getting CI working for PRs that add files into `UI/forms`.

Okay.

RytoEX:
> TEST commit broke a UI file on purpose, was correctly picked up.

Yes, apparently on bash `globstar` is required for this behavior, as you've noted. I personally tend to prefer more explicit invocations, but this should be fine.

PatTheMav:
> > A deletion is not a "change" in that interpretation.
> 
> Sorry, it was a rename git-wise (moved to another directory).

A-ha - yeah that might throw a wrench in the proceedings. It's good-enough for now, but I wonder if there is an elegant way to get that from git without going back to a full `diff` output and `grep`ping over it.

PR #9278, opened 2023-07-20
CI: Switch service availability checks to macOS runners

PatTheMav:
### Description
Switches service availability checks to macOS runners on GitHub Actions. Affects only scheduled and dispatched workflow runs, as other invocations only check schema validity.

### Motivation and Context
Linux runners exhibit connectivity issues with some services that are not present with macOS runners. Switching to macOS thus reduces chances for false positives.

### How Has This Been Tested?
Requires testing on CI.

### Types of changes

### Checklist:

### DIFF
--- a/.github/workflows/dispatch.yaml
+++ b/.github/workflows/dispatch.yaml
@@ -34,7 +34,7 @@ jobs:
   services-validation:
     name: Check Services Configuration Files 🕵️
     if: github.repository_owner == 'obsproject' && inputs.job == 'services'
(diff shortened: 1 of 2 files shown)


RytoEX:
For context, see #8807 (wpstream has an ingest that fails to respond to our service availability check).

This looks correct. Any run that is doing an actual service availability check will need to be on macOS unless/until we figure out how to fix the issue with the service check script and wpstream.

derrod (approved):

PR #9289, opened 2023-07-21
CI: Shorten Pull Request workflow names

gxalpha:
### Description

Shortens the names of the PR workflow and build/formatting jobs

### Motivation and Context

Long prefixes often make the important parts like OS get cut off, especially on mobile devices.

### How Has This Been Tested?

Works fine on this PR

### Types of changes

### Checklist:

### DIFF
--- a/.github/workflows/pr-pull.yaml
+++ b/.github/workflows/pr-pull.yaml
@@ -1,4 +1,4 @@
-name: Pull Request
+name: Pull
 run-name: ${{ github.event.pull_request.title }} pull request run 🚀
 on:
   workflow_dispatch:
@@ -14,20 +14,20 @@ concurrency:
   cancel-in-progress: true
 jobs:
   check-format:
-    name: Check Formatting 🔍
+    name: Format 🔍
     uses: ./.github/workflows/check-format.yaml
     permissions:
       contents: read
 
   build-project:
-    name: Build Project 🧱
+    name: Build 🧱
     uses: ./.github/workflows/build-project.yaml
     secrets: inherit
     permissions:
       contents: read
 
   compatibility-validation:
-    name: Validate Compatibility Data 🕵️
+    name: Validate Compatibility 🕵️
     if: github.base_ref == 'master'
     runs-on: ubuntu-22.04
     permissions:
@@ -49,7 +49,7 @@ jobs:
         uses: ./.github/actions/compatibility-validator
 
   services-validation:
-    name: Validate Services Data 🕵️
+    name: Validate Services 🕵️
     if: github.base_ref == 'master'
     runs-on: ubuntu-22.04
     permissions:
--- a/.github/workflows/build-project.yaml
+++ b/.github/workflows/build-project.yaml
@@ -3,7 +3,7 @@ on:
   workflow_call:
 jobs:
   check-event:
-    name: Check GitHub Event Data 📡
+    name: Event Data 🔎
     runs-on: ubuntu-22.04
     defaults:
       run:
@@ -58,7 +58,7 @@ jobs:
           echo "commitHash=${GITHUB_SHA:0:9}" >> $GITHUB_OUTPUT
 
   macos-build:
-    name: Build for macOS 🍏
+    name: macOS 🍏
     runs-on: macos-13
     needs: check-event
     strategy:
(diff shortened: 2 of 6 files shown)


gxalpha:
Hmm, not entirely sold on "Format Check" yet. Open to suggestions.

.github/workflows/pr-pull.yaml
```diff
@@ -14,13 +14,13 @@ concurrency:
   cancel-in-progress: true
 jobs:
   check-format:
-    name: Check Formatting 🔍
+    name: Format Check 🔍
```
RytoEX, line 17:
```suggestion
    name: Check Format 🔍
```
Verb + Noun

RytoEX (commented):
My two cents here. The other changes I don't hold a strong opinion on.

.github/workflows/pr-pull.yaml
```diff
@@ -14,13 +14,13 @@ concurrency:
...
   check-format:
-    name: Check Formatting 🔍
+    name: Check Format 🔍
     uses: ./.github/workflows/check-format.yaml
     permissions:
       contents: read
 
   build-project:
-    name: Build Project 🧱
+    name: BUILD 🧱
```
PatTheMav, line 23: I don't think this change is meaningful, as the first token is shortened to a single letter by GitHub anyway.

PatTheMav:
Going by this, the changes in this PR don't really move the needle much, instead other changes would be more beneficial:

Bildschirmfoto 2023-07-22 um 16 03 32

* Shorten the Flatpak label and label for the other build jobs to just the platform, so "Flatpak 📦", "macOS 🍏" and so forth.
* "Validate Services Data" could be shortened to "Validate Services", I don't know how best to do the same for compatibility data
* "Check GitHub Event Data" could be shortened to "Event Data 🔎"
* "Check Format" could be shortened to "Format 🔎"

Otherwise this is just a quirk of using workflow calls that make GitHub create this long composed label for each job and we shouldn't bend over backwards too much over it.

gxalpha:
> Going by this, the changes in this PR don't really move the needle much

The PR is for this view, which is greatly improved by this PR:
<p>image
image
</p>

The changes you suggested would further improve it though, I'll implement those later.

.github/workflows/pr-pull.yaml
gxalpha: See https://github.com/obsproject/obs-studio/pull/9289#issuecomment-1646612797

PatTheMav: I specifically did not use abbreviations and capital case names in the rework, so I'd prefer "Build" there (removing "Project" isn't detrimental IMO).

I do prefer keeping "Pull Request" though and while I don't wish to diminish the view on mobile phones, the overall GitHub UI is not really usable in portrait mode (and the site isn't really optimised for that anyway).

gxalpha:
Applied suggestions.

PatTheMav:
Could you also shorten

> name: Push to master and release branches

to just "Push" - that one mostly appears in the "Actions" tab when browsing for active workflows and should be distinct enough from the others we have in there.

.github/workflows/pr-pull.yaml
```diff
@@ -1,4 +1,4 @@
-name: Pull Request
+name: PR
```
PatTheMav, line 1:
```suggestion
name: Pull
```

I've looked at the other files and also logically went over this, and the correct label is "Pull", because the workflow actually "pulls" the pull-_request_ and compiles it. "PR" would be semantically wrong, and would also go against the naming scheme of the other workflows.

.github/workflows/pr-pull.yaml
RytoEX: I concur with @PatTheMav . Use "Build".

RytoEX: Looks like this has already been changed. Resolving.

gxalpha:
Updated, apologies for the delay.

.github/workflows/pr-pull.yaml
gxalpha: Applied change.

.github/workflows/pr-pull.yaml
gxalpha: Applied change.

PatTheMav:
Looks fine on mobile in portrait mode to me.

.github/workflows/pr-pull.yaml
PatTheMav, line 1: Could you apply these changes also to other workflows that might use these or similar steps to ensure the renaming is consistent across push/dispatch/scheduled workflows as well?

gxalpha: Applied to push, dispatch, scheduled, and publish.

PatTheMav (approved):

PatTheMav:
Final nit: I'd prefer if the commit title would look something like

> CI: Shorten job and step labels in workflow files

PR #9294, opened 2023-07-22
aja,aja-output-ui: Enforce -Wdangling-reference to never turn into an error

tytan652:
### Description

With how GCC 13.1 progress, `-Wdangling-reference` is victim of false positive, avoid letting it turn into an error on AJA plugins since those are the only one having this warning.

`" "` and `""` are temporaries considered bound to the returned reference but this is not the case.
```
[build] obs-studio/plugins/aja/aja-routing.cpp: In static member function ‘static bool aja::Routing::ParseRouteString(const std::string&, NTV2XptConnections&)’:
[build] /home/tytan652/Programming/OBS/obs-studio/plugins/aja/aja-routing.cpp:28:28: error: possibly dangling reference to a temporary [-Werror=dangling-reference]
[build]    28 |         const std::string &route_strip = aja::replace(route_lower, " ", "");
[build]       |                            ^~~~~~~~~~~
[build] obs-studio/plugins/aja/aja-routing.cpp:28:54: note: the temporary was destroyed at the end of the full expression ‘aja::replace(route_lower, std::__cxx11::basic_string<char>(((const char*)" "), std::allocator<char>()), std::__cxx11::basic_string<char>(((const char*)""), std::allocator<char>()))’
[build]    28 |         const std::string &route_strip = aja::replace(route_lower, " ", "");
[build]       |                                          ~~~~~~~~~~~~^~~~~~~~~~~~~~~~~~~~~~
```

### Motivation and Context

Reduction of false positive is not planned until GCC 14.

### How Has This Been Tested?

Build with ArchLinux GCC 13 package and no errors.

### Types of changes

- GCC false positive bypass

### Checklist:

### DIFF
--- a/UI/frontend-plugins/aja-output-ui/CMakeLists.txt
+++ b/UI/frontend-plugins/aja-output-ui/CMakeLists.txt
@@ -81,7 +81,18 @@ elseif(OS_LINUX OR OS_FREEBSD)
   find_package(X11 REQUIRED)
   target_link_libraries(aja-output-ui PRIVATE X11::X11 Qt::GuiPrivate)
 
-  target_compile_options(aja-output-ui PRIVATE -Wno-error=deprecated-declarations)
+  #[[
+    Note about -Wdangling-reference on GCC, this warning seems to be subject of false positives. This warning is set to not turn into an error.
+
+    - https://gcc.gnu.org/bugzilla/show_bug.cgi?id=107532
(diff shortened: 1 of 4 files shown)


norihiro:
The function `aja::replace` also modifies the 1st argument so that it's nonsense for the function to return something.
I'd like to suggest ignoring the return value of `aja:replace` instead of adding the option.

PR #9304, opened 2023-07-23
obs-qsv11: Downgrade minimum required CMake version

norihiro:
### Description

Set minimum required cmake version to 3.22 instead of 3.24.

### Motivation and Context

Ubuntu 22.04 provides cmake-3.22 so that the build is failing when compiling on Ubuntu 22.04 without extra package.

The ubuntu-22.04 runner on GitHub Actions has a newer cmake as described on the document.
https://github.com/actions/runner-images/blob/main/images/linux/Ubuntu2204-Readme.md#tools
So, CI is succeeding.

### How Has This Been Tested?

OS: Ubuntu 22.04 on docker
Installed necessary packages by `apt`.
Checked the build script `.github/scripts/build-linux` succeeded.

### Types of changes

- Bug fix (non-breaking change which fixes an issue)

### Checklist:

- [x] My code has been run through [clang-format](https://github.com/obsproject/obs-studio/blob/master/.clang-format).
- [x] I have read the [**contributing** document](https://github.com/obsproject/obs-studio/blob/master/CONTRIBUTING.rst).
- [x] My code is not on the master branch.
- [x] The code has been tested.
- [x] All commit messages are properly formatted and commits squashed where appropriate.
- [x] I have included updates to all appropriate documentation.

PR #9305, opened 2023-07-23
CI: Enforce the use of Ubuntu's CMake

tytan652:
### Description

Requires:
 - #9304

Use Ubuntu CMake package.

If it worked CI should fail while configuring obs-qsv until #9304 is merged.

### Motivation and Context

Github enforce an updated version of CMake, while we need to rely on what the distribution actually provides.

### How Has This Been Tested?

If it worked CI should fail while configuring obs-qsv until #9304 is merged.

The right CMake is indeed use since configure has failed on obs-qsv.

### Types of changes

### Checklist:

### DIFF
--- a/.github/scripts/.build.zsh
+++ b/.github/scripts/.build.zsh
@@ -185,6 +185,7 @@ build() {
       popd
       ;;
     linux-*)
+      local cmake_bin='/usr/bin/cmake'
       cmake_args+=(
         -S ${PWD} -B build_${target##*-}
         -G Ninja
@@ -199,15 +200,15 @@ build() {
       cmake_install_args+=(build_${target##*-} --prefix ${project_root}/build_${target##*-}/install/${config})
 
       log_group "Configuring ${product_name}..."
-      cmake -S ${project_root} ${cmake_args}
+      ${cmake_bin} -S ${project_root} ${cmake_args}
 
       log_group "Building ${product_name}..."
       if (( debug )) cmake_build_args+=(--verbose)
-      cmake ${cmake_build_args}
+      ${cmake_bin} ${cmake_build_args}
 
       log_group "Installing ${product_name}..."
       if (( debug )) cmake_install_args+=(--verbose)
-      cmake ${cmake_install_args}
+      ${cmake_bin} ${cmake_install_args}
       ;;
   }
   popd
--- a/.editorconfig
+++ b/.editorconfig
@@ -45,3 +45,11 @@ indent_size = 4
 [*.py]
 indent_style = space
 indent_size = 4
+
+[*.yaml]
+indent_style = space
+indent_size = 2
+
+[{*.zsh,.*.zsh,build-aux/.functions/*,.github/scripts/utils.zsh/*}]
+indent_style = space
+indent_size = 2
--- a/.github/scripts/.Aptfile
+++ b/.github/scripts/.Aptfile
@@ -1,5 +1,5 @@
 package 'ccache'
-package 'cmake'
+package 'cmake', bin: '/usr/bin/cmake'
 package 'curl'
 package 'git'
 package 'jq'
--- a/.github/scripts/.package.zsh
+++ b/.github/scripts/.package.zsh
@@ -209,13 +209,14 @@ package() {
     log_group
 
   } elif [[ ${host_os} == linux ]] {
+    local cmake_bin='/usr/bin/cmake'
     local -a cmake_args=()
     if (( debug )) cmake_args+=(--verbose)
 
     if (( package )) {
       log_group "Packaging obs-studio..."
       pushd ${project_root}
-      cmake --build build_${target##*-} --config ${config} -t package ${cmake_args}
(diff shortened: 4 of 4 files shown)


RytoEX:
I'm not sure how I feel about this. It would probably make some failure conditions easier to spot, but it makes me a little uneasy.

PatTheMav:
IMO this is not the correct approach, because CMake itself has very robust backwards support and specifically allows more recent versions to behave like older versions to ensure compatibility with older scripts.

If I understood it right the root cause of the issue was that QSV was made available to Linux, but its `CMakeLists.txt` was not updated to include older CMake versions available on Ubuntu natively.

In general it is advised to always use the most recent CMake version and not linger for too long on older versions, which is why GitHub preinstalls CMake not from distribution repositories but KitWare's own.

If we need/want "pure" Ubuntu runners, we need to move the Ubuntu builds to docker containers.

tytan652:
Sorry, answering through email.

> In general it is advised to always use the most recent CMake version and not linger for too long on older versions, which is why GitHub preinstalls CMake not from distribution repositories but KitWare's own.

Unfortunately, it can't work like that.
The distribution we support is Ubuntu. So we must deal with distribution packages to avoid missing issues (e.g
,  obs-qsv).

Installing ZSH or clang-format through brew is fine, but when it comes to CMake (which is a build deps) or OBS deps the limit is what the distribution provides.

RytoEX:
> The distribution we support is Ubuntu. So we must deal with distribution packages to avoid missing issues (e.g , obs-qsv).

IIRC, it's technically possible on the PPA builder to install newer CMake via Kitware's own apt repo. We just don't do it at present.

tytan652:
> it's technically possible on the PPA builder to install newer CMake via Kitware's own apt repo. We just don't do it at present

If it is, this apt repo should become a hard requirement and written all over the build wiki that we "require" a relatively recent CMake. So our limit became the CMake version provided by the Flatpak runtime/sdk.

---

> which is why GitHub preinstalls CMake not from distribution repositories but KitWare's own

(Little off-topic) It's not from KitWare's repo it's a custom install with a bash script with `/usr/local` as prefix.

PatTheMav:
> > it's technically possible on the PPA builder to install newer CMake via Kitware's own apt repo. We just don't do it at present
> 
> If it is, this apt repo should become a hard requirement and written all over the build wiki that we "require" a relatively recent CMake. So our limit became the CMake version provided by the Flatpak runtime/sdk.
> 
> Edit: Note, that distro packagers of fixed releases (e.g., Debian) might come complain of such requirements.
> 
> > which is why GitHub preinstalls CMake not from distribution repositories but KitWare's own
> 
> (Little off-topic) It's not from KitWare's repo it's a custom install with a bash script with `/usr/local` as prefix.

Shouldn't it be easier to install CMake via `apt` and use `update-alternatives` if necessary (usually `/usr/bin` should take precedence in `PATH` anyway)?

tytan652:
> and use `update-alternatives` if necessary

I don't know if it can cover custom installs and non-versioned binaries.

> (usually `/usr/bin` should take precedence in `PATH` anyway)?

It mind strange to you, but not `/usr/bin` is usually put last (`PATH=/my/new/bin:$PATH`).
We already install the Ubuntu package, but the local install takes over.

tytan652:
It's true that Kitware drop their PPA for a real apt repo.

PatTheMav:
Wouldn't it be easier to adjust the build script to explicitly use the `cmake` binary installed from Ubuntu's official packages (i.e. `/usr/bin/cmake`)?

We used to use the absolute paths on macOS to ensure that variants shipped by macOS (BSD variants) were used instead of GNU variants possibly installed via Homebrew.

tytan652:
Some script are common between macOS and Linux, so I don't know if it's possible.

PatTheMav:
> Some script are common between macOS and Linux, so I don't know if it's possible without adding a ton of switches.

I'd add a local variable `cmake_bin` and define it in the OS-specific branches and then use `${cmake_bin}` to call it.

tytan652:
Replace removal by an enforcement of the use of `/usr/bin/cmake`.

- For testing, `/usr/local/bin/cmake` was temporarily replaced by a script with `false` shell command
- For testing, 9ca55eea43be9e458bc65092d4e8d53f82518a7a temporarily

.github/scripts/.build.zsh
```diff
@@ -71,6 +71,7 @@ build() {
   local -i codesign=0
 
   if [[ ${host_os} == linux ]] {
+    local cmake_bin='/usr/bin/cmake'
```
PatTheMav, line 74: `local cmake_bin` needs to be defined outside the conditional (because it will also be used outside of them) and then assigned inside of them.

tytan652: Is it good now ?

PatTheMav: Yeah that should be fine.

tytan652:
Rebased and redone since #9405 completely broke the PR

PatTheMav (approved):

PR #9306, opened 2023-07-24
obs-webrtc: Fix WHIP request not following redirects

aggresss:
### Description

Fix issues that WHIP request did not follow 307 redirect.

Reference to [WHIP Section 4.3](https://datatracker.ietf.org/doc/html/draft-ietf-wish-whip-08#section-4.3)

> WHIP endpoints and Media Servers might not be colocated on the same server, so it is possible to load balance incoming requests to different Media Servers. WHIP clients SHALL support HTTP redirection via the "307 Temporary Redirect" response as described in [[RFC9110](https://www.rfc-editor.org/rfc/rfc9110)] section 6.4.7. The WHIP resource URL MUST be a final one, and redirections are not required to be supported for the PATCH and DELETE requests sent to it.

And libcurl documents [https://curl.se/libcurl/c/CURLOPT_FOLLOWLOCATION.html](https://curl.se/libcurl/c/CURLOPT_FOLLOWLOCATION.html) . CURLOPT_FOLLOWLOCATION should be set 1L to allow curl_easy_perform follow the redirection response.

### How Has This Been Tested?

Tested on separate branch as far as possible.

### Types of changes

Add a curl option to curl_sasy_perform.

### Checklist:

### DIFF
--- a/plugins/obs-webrtc/whip-output.cpp
+++ b/plugins/obs-webrtc/whip-output.cpp
@@ -290,7 +290,7 @@ bool WHIPOutput::Connect()
 	}
 
 	std::string read_buffer;
-	std::string location_header;
+	std::vector<std::string> location_headers;
 
 	char offer_sdp[4096] = {0};
 	rtcGetLocalDescription(peer_connection, offer_sdp, sizeof(offer_sdp));
@@ -307,12 +307,14 @@ bool WHIPOutput::Connect()
 	curl_easy_setopt(c, CURLOPT_WRITEDATA, (void *)&read_buffer);
 	curl_easy_setopt(c, CURLOPT_HEADERFUNCTION,
 			 curl_header_location_function);
-	curl_easy_setopt(c, CURLOPT_HEADERDATA, (void *)&location_header);
+	curl_easy_setopt(c, CURLOPT_HEADERDATA, (void *)&location_headers);
 	curl_easy_setopt(c, CURLOPT_HTTPHEADER, headers);
 	curl_easy_setopt(c, CURLOPT_URL, endpoint_url.c_str());
 	curl_easy_setopt(c, CURLOPT_POST, 1L);
 	curl_easy_setopt(c, CURLOPT_COPYPOSTFIELDS, offer_sdp);
 	curl_easy_setopt(c, CURLOPT_TIMEOUT, 8L);
+	curl_easy_setopt(c, CURLOPT_FOLLOWLOCATION, 1L);
+	curl_easy_setopt(c, CURLOPT_UNRESTRICTED_AUTH, 1L);
 
 	auto cleanup = [&]() {
 		curl_easy_cleanup(c);
@@ -347,13 +349,16 @@ bool WHIPOutput::Connect()
 		return false;
 	}
 
-	if (location_header.empty()) {
+	long redirect_count = 0;
+	curl_easy_getinfo(c, CURLINFO_REDIRECT_COUNT, &redirect_count);
+
+	if (location_headers.size() < static_cast<size_t>(redirect_count) + 1) {
 		do_log(LOG_WARNING,
 		       "WHIP server did not provide a resource URL via the Location header");
 	} else {
 		CURLU *h = curl_url();
-		curl_url_set(h, CURLUPART_URL, endpoint_url.c_str(), 0);
-		curl_url_set(h, CURLUPART_URL, location_header.c_str(), 0);
+		curl_url_set(h, CURLUPART_URL, location_headers.back().c_str(),
+			     0);
 		char *url = nullptr;
 		CURLUcode rc = curl_url_get(h, CURLUPART_URL, &url,
 					    CURLU_NO_DEFAULT_PORT);
--- a/plugins/obs-webrtc/whip-utils.h
+++ b/plugins/obs-webrtc/whip-utils.h
@@ -45,7 +45,7 @@ static size_t curl_writefunction(char *data, size_t size, size_t nmemb,
 static size_t curl_header_location_function(char *data, size_t size,
 					    size_t nmemb, void *priv_data)
 {
-	auto header_buffer = static_cast<std::string *>(priv_data);
+	auto header_buffer = static_cast<std::vector<std::string> *>(priv_data);
 
 	size_t real_size = size * nmemb;
 
@@ -54,8 +54,11 @@ static size_t curl_header_location_function(char *data, size_t size,
 
 	if (!astrcmpi_n(data, "location: ", LOCATION_HEADER_LENGTH)) {
 		char *val = data + LOCATION_HEADER_LENGTH;
-		header_buffer->append(val, real_size - LOCATION_HEADER_LENGTH);
-		*header_buffer = trim_string(*header_buffer);
+		auto header_temp =
+			std::string(val, real_size - LOCATION_HEADER_LENGTH);
+
+		header_temp = trim_string(header_temp);
+		header_buffer->push_back(header_temp);
 	}
 
 	return real_size;


Sean-Der:
Hi @aggresss thanks for contributing, this is a super important fix! Can you also set https://curl.se/libcurl/c/CURLOPT_UNRESTRICTED_AUTH.html

Otherwise authentication is dropped on redirect

aggresss:
> Hi @aggresss thanks for contributing, this is a super important fix! Can you also set https://curl.se/libcurl/c/CURLOPT_UNRESTRICTED_AUTH.html
> 
> Otherwise authentication is dropped on redirect

Updated.

notr1ch:
Does this mean all the `location_header` code is redundant? If so it should be removed.

Sean-Der:
@notr1ch Unrelated!

Location header is used to modify a session. Things like querying active viewers.

This PR just enables following redirects when creating

notr1ch:
It seems this PR will conflict with that then, per the curl docs:
> It's important to note that the callback will be invoked for **the headers of all responses received after initiating a request and not just the final response**. This includes all responses which occur during authentication negotiation. If you need to operate on only the headers from the final response, you will need to collect headers in the callback yourself and use HTTP status lines, for example, to delimit response boundaries.

https://curl.se/libcurl/c/CURLOPT_HEADERFUNCTION.html

This sounds like that the final location header will thus be made up of all the intermediary location lines too as we use `append` instead of overwriting the string.

notr1ch:
It does, however the documentation seems to suggest that the header callback will still be called for each redirection handled by curl. I haven't tested this though to confirm.

Sean-Der:
@aggresss Are you able to confirm/test that the `Location` header is properly captured across redirects?

If you need any help happy to write code/test. I think this fix is vital for v30, want to help however I can.

thanks!

aggresss:
@Sean-Der @tt2468 I Updated the code, and make a mini workbench for it [https://github.com/aggresss/playground-cmake/tree/main/workbench/libcurl_hello](https://github.com/aggresss/playground-cmake/tree/main/workbench/libcurl_hello)

On [WHIP Section 4](https://datatracker.ietf.org/doc/html/draft-ietf-wish-whip-09#section-4) , The 201(Created) response header field `Location` can been invoked directly.

aggresss:
@Sean-Der @tt2468 I updated the commit, and made a mini workbench for it [https://github.com/aggresss/playground-cmake/tree/main/workbench/obs_whip_demo](https://github.com/aggresss/playground-cmake/tree/main/workbench/obs_whip_demo)

- On [WHIP Section 4](https://datatracker.ietf.org/doc/html/draft-ietf-wish-whip-09#section-4) , The 201(Created) response header field `Location` can been invoked as a URL directly.
- [CURLOPT_HEADERFUNCTION](https://curl.se/libcurl/c/CURLOPT_HEADERFUNCTION.html) `The header callback will be called once for each header and only complete header lines are passed on to the callback. `

aggresss:
@Sean-Der @tt2468 I updated the commit, and made a mini workbench for it [https://github.com/aggresss/playground-cmake/tree/main/workbench/obs_whip_demo](https://github.com/aggresss/playground-cmake/tree/main/workbench/obs_whip_demo)

- On [WHIP Section 4](https://datatracker.ietf.org/doc/html/draft-ietf-wish-whip-09#section-4) , The 201(Created) response header field `Location` can been invoked as a URL directly.
- [CURLOPT_HEADERFUNCTION](https://curl.se/libcurl/c/CURLOPT_HEADERFUNCTION.html) `The header callback will be called once for each header and only complete header lines are passed on to the callback. `

notr1ch:
It seems like this will now use the last location header received, instead of the (possible) location header from the final URL. We're only interested in the location header on the final URL I believe.

DDRBoxman:
@notr1ch this logic seems to make sense to me. Curl follows the 307 requests and we store the location header each time, the final request that curl makes will response with a 201 status and will give us a location header in the callback that we append to the location_headers list. This URL then gets stored in the resource_url for later use.

DDRBoxman:
Also can we clean up the commit / title a little?
Something like the below might be nicer:
obs-webrtc: WHIP creation requests follow HTTP redirects

RytoEX:
> Also can we clean up the commit / title a little?
> Something like the below might be nicer:
> obs-webrtc: WHIP creation requests follow HTTP redirects

The first word after the prefix should be a verb in imperative form, per the commit guidelines.

aggresss:
> It seems like this will now use the last location header received, instead of the (possible) location header from the final URL. We're only interested in the location header on the final URL I believe.

location_headers.size() should equal redirect_count + 1

notr1ch:
> @notr1ch this logic seems to make sense to me. Curl follows the 307 requests and we store the location header each time, the final request that curl makes will response with a 201 status and will give us a location header in the callback that we append to the location_headers list. This URL then gets stored in the resource_url for later use.

The issue is if the WHIP endpoint does not return a Location header then we use the one from the redirect which is just wrong. The spec seems to suggest that a Location header is mandatory, so it should be a fatal error if there isn't one on the final request.

tt2468:
I believe we currently throw info or warning as more than half of WHIP servers I've tested do not return resource URLs.

aggresss:
> @aggresss Are you able to confirm/test that the `Location` header is properly captured across redirects?
> 
> If you need any help happy to write code/test. I think this fix is vital for v30, want to help however I can.
> 
> thanks!

@Sean-Der Have been tested.

derrod (changes requested):
Remove the merge commit from your PR, do a rebase instead.

aggresss:
> Remove the merge commit from your PR, do a rebase instead.

@derrod
Done

plugins/obs-webrtc/whip-output.cpp
```diff
@@ -347,13 +349,16 @@ bool WHIPOutput::Connect()
 		return false;
 	}
 
-	if (location_header.empty()) {
+	long redirect_count = 0;
+	curl_easy_getinfo(c, CURLINFO_REDIRECT_COUNT, &redirect_count);
+
+	if (location_headers.size() < static_cast<size_t>(redirect_count) + 1) {
 		do_log(LOG_WARNING,
 		       "WHIP server did not provide a resource URL via the Location header");
```
DDRBoxman, line 357: So I still have one question, when we make requests to a whip endpoint and they don't send us a location header does that mean that we actually cannot send delete / update requests? Or are the providers incorrectly implementing their endpoints so that the endpoint also receives those requests?

Any thoughts on if we could fail these and show an alert to put pressure on these services that have an incorrect implementation?

aggresss: Reference to [WHIP Section 4](https://datatracker.ietf.org/doc/html/draft-ietf-wish-whip-09#name-protocol-operation)

> To explicitly terminate a session, the WHIP client MUST perform an HTTP DELETE request to the resource URL returned in the Location header field of the initial HTTP POST. Upon receiving the HTTP DELETE request, the WHIP resource will be removed and the resources freed on the Media Server, terminating the ICE and DTLS sessions.

The purpose of `Location` in response header is to explicitly release resources. and if not, implicitly release resources.

DDRBoxman:
obs-webrtc: Fix WHIP request not follow redirect -> obs-webrtc: Fix WHIP request not following redirects

notr1ch:
> I believe we currently throw info or warning as more than half of WHIP servers I've tested do not return resource URLs.

IMO we should upgrade this to an error. https://datatracker.ietf.org/doc/html/draft-ietf-wish-whip-09#section-4-2 uses the slightly ambiguous "will" instead of "SHOULD" or "MUST" but it seems clear from all examples on the page that the Location header is a requirement. Without it, it's impossible to satisfy the "explicitly terminate a session" part of the protocol which has a "MUST". Now is the time to be strict about the protocol so we don't get all kinds of compatibility hacks that last forever.

aggresss:
> obs-webrtc: Fix WHIP request not following redirects

Done

RytoEX:
Closing as @Sean-Der stated that #9384 is meant to supersede this. I don't want a superseded PR to remain open as it will cause confusion as to which should be reviewed/merged.

PR #1465, opened 2018-09-02
Add compiler checks to avoid VLA

admshao:
There isn't any on Windows (maybe MSVC just doesn't allow/support it) but on Linux there were only 1 case of VLA and we should avoid it if possible.

Links:
lkml [here](https://lore.kernel.org/lkml/CA+55aFzCG-zNmZwX4A2FQpadafLfEzK6CC=qPXydAacU1RqZWA@mail.gmail.com/T/#u)
linux git [changes](https://git.kernel.org/pub/scm/linux/kernel/git/torvalds/linux.git/commit/?id=02361bc77888)
linux securiy summit [pages 6, 7 and 8](https://outflux.net/slides/2018/lss/danger.pdf)

PR #9339, opened 2023-07-30
Contain simde diagnostic pragmas and replace variable-length arrays

### LINKED ISSUE #9160: FTBFS: variable length arrays are treated as errors
### Operating System Info

Other

### Other OS

Debian

### OBS Studio Version

29.1.3

_No response_

### OBS Studio Log URL

n.n.

### OBS Studio Crash Log URL

_No response_

### Expected Behavior

i expect to be able to compile obs-studio with all it's bundled plugins

### Current Behavior

build fails with:

```
cd /build/obs-studio-2j3wxb/obs-studio-29.1.3+dfsg/obj-x86_64-linux-gnu/plugins/linux-pipewire && /usr/bin/ccache /usr/bin/cc -DENABLE_HEVC -DHAVE_OBSCONFIG_H -Dlinux_pipewire_EXPORTS -I/build/obs-studio-2j3wxb/obs-studio-29.1.3+dfsg/libobs -I/build/obs-studio-2j3wxb/obs-studio-29.1.3+dfsg/obj-x86_64-linux-gnu/config -I/build/obs-studio-2j3wxb/obs-studio-29.1.3+dfsg/deps/glad/include -isystem /usr/include/pipewire-0.3 -isystem /usr/include/spa-0.2 -isystem /usr/include/glib-2.0 -isystem /usr/lib/x86_64-linux-gnu/glib-2.0/include -isystem /usr/include/libmount -isystem /usr/include/blkid -isystem /usr/include/gio-unix-2.0 -g -O2 -ffile-prefix-map=/build/obs-studio-2j3wxb/obs-studio-29.1.3+dfsg=. -fstack-protector-strong -Wformat -Werror=format-security -DSIMDE_ENABLE_OPENMP -fopenmp-simd -O3 -Wdate-time -D_FORTIFY_SOURCE=2 -DFFMPEG_MUX_FIXED=\"/usr/lib/x86_64-linux-gnu/obs-plugins/obs-ffmpeg/obs-ffmpeg-mux\" -std=gnu17 -fPIC -Werror -Wextra -Wvla -Wswitch -Wno-error=switch -Wformat -Wformat-security -Wunused-parameter -Wno-unused-function -Wno-missing-field-initializers -fno-strict-aliasing -Werror-implicit-function-declaration -Wno-missing-braces -Wno-error=maybe-uninitialized -mmmx -msse -msse2 -I/usr/include/pipewire-0.3 -I/usr/include/spa-0.2 -D_REENTRANT -I/usr/include/glib-2.0 -I/usr/lib/x86_64-linux-gnu/glib-2.0/include -I/usr/include/libmount -I/usr/include/blkid -I/usr/include/gio-unix-2.0 -pthread -MD -MT plugins/linux-pipewire/CMakeFiles/linux-pipewire.dir/pipewire.c.o -MF CMakeFiles/linux-pipewire.dir/pipewire.c.o.d -o CMakeFiles/linux-pipewire.dir/pipewire.c.o -c /build/obs-studio-2j3wxb/obs-studio-29.1.3+dfsg/plugins/linux-pipewire/pipewire.c
/build/obs-studio-2j3wxb/obs-studio-29.1.3+dfsg/plugins/linux-pipewire/pipewire.c: In function ‘on_process_cb’:
/build/obs-studio-2j3wxb/obs-studio-29.1.3+dfsg/plugins/linux-pipewire/pipewire.c:596:17: error: ISO C90 forbids variable length array ‘offsets’ [-Werror=vla]
  596 |                 uint32_t offsets[planes];
      |                 ^~~~~~~~
/build/obs-studio-2j3wxb/obs-studio-29.1.3+dfsg/plugins/linux-pipewire/pipewire.c:597:17: error: ISO C90 forbids variable length array ‘strides’ [-Werror=vla]
  597 |                 uint32_t strides[planes];
      |                 ^~~~~~~~
/build/obs-studio-2j3wxb/obs-studio-29.1.3+dfsg/plugins/linux-pipewire/pipewire.c:598:17: error: ISO C90 forbids variable length array ‘modifiers’ [-Werror=vla]
  598 |                 uint64_t modifiers[planes];
      |                 ^~~~~~~~
/build/obs-studio-2j3wxb/obs-studio-29.1.3+dfsg/plugins/linux-pipewire/pipewire.c:599:17: error: ISO C90 forbids variable length array ‘fds’ [-Werror=vla]
  599 |                 int fds[planes];
      |                 ^~~
cc1: all warnings being treated as errors
```

The error is obviously triggered because
- the code uses variable length arrays https://github.com/obsproject/obs-studio/blob/19d7a96de38e40d89cc953db84452b53ec63487d/plugins/linux-pipewire/pipewire.c#L614-L617
- the code is compiled with `-Werror -Wvla`
presumable because of https://github.com/obsproject/obs-studio/blob/19d7a96de38e40d89cc953db84452b53ec63487d/cmake/Modules/CompilerConfig.cmake#L110-L113

cmake is invoked with:
```sh
cd obj-x86_64-linux-gnu && cmake -DCMAKE_INSTALL_PREFIX=/usr -DCMAKE_BUILD_TYPE=None -DCMAKE_INSTALL_SYSCONFDIR=/etc -DCMAKE_INSTALL_LOCALSTATEDIR=/var -DCMAKE_EXPORT_NO_PACKAGE_REGISTRY=ON -DCMAKE_FIND_USE_PACKAGE_REGISTRY=OFF -DCMAKE_FIND_PACKAGE_NO_PACKAGE_REGISTRY=ON -DFETCHCONTENT_FULLY_DISCONNECTED=ON -DCMAKE_INSTALL_RUNSTATEDIR=/run -DCMAKE_SKIP_INSTALL_ALL_DEPENDENCY=ON "-GUnix Makefiles" -DCMAKE_VERBOSE_MAKEFILE=ON -DCMAKE_INSTALL_LIBDIR=lib/x86_64-linux-gnu -DOBS_MULTIARCH_SUFFIX=/x86_64-linux-gnu -DUNIX_STRUCTURE=TRUE -DCMAKE_SKIP_RPATH=TRUE -DDISABLE_UPDATE_MODULE=TRUE -DBUILD_VST=OFF -DENABLE_AJA=OFF -DENABLE_SNDIO=OFF -DENABLE_JACK=TRUE -DOBS_VERSION_OVERRIDE=29.1.3.1-1 -DAPPDATA_RELEASE_DATE=2023-06-27 ..
```

### Steps to Reproduce

https://salsa.debian.org/multimedia-team/obs-studio/-/jobs/4368844

given that these compiler warnings and errors are mostly for development of obs-studio, should/could they be disabled for release-builds?

tytan652:
### Description

Fixes #9160

- Replace linux-pipewire variable-length arrays with dynamic arrays, those were not generating errors because of our in-tree simde that makes some warning ignored.
- Contain simde diagnostic pragmas

### Motivation and Context

I was trying to use my system simde rather than the in-tree and discover how to reproduce #9160.

So I tried to fix the vla issue and the pragma contamination.

### How Has This Been Tested?

Pipewire captures seems to work fine.

### Types of changes

### Checklist:

- [x] The code has been tested on Linux.
- [ ] The code has been tested on macOS.

### DIFF
--- a/libobs/obs-cocoa.m
+++ b/libobs/obs-cocoa.m
@@ -462,7 +462,7 @@ static bool code_to_str(int code, struct dstr *str)
 void obs_key_to_str(obs_key_t key, struct dstr *str)
 {
     const UniCharCount max_length = 16;
-    UniChar buffer[max_length];
+    UniChar buffer[16];
 
     if (localized_key_to_str(key, str))
         return;
--- a/plugins/mac-virtualcam/src/obs-plugin/plugin-main.mm
+++ b/plugins/mac-virtualcam/src/obs-plugin/plugin-main.mm
@@ -369,27 +369,25 @@ static bool virtualcam_output_start(void *data)
                                            .mScope = kCMIOObjectPropertyScopeGlobal,
                                            .mElement = kCMIOObjectPropertyElementMain};
         CMIOObjectGetPropertyDataSize(kCMIOObjectSystemObject, &address, 0, NULL, &size);
-        size_t num_devices = size / sizeof(CMIOObjectID);
-        CMIOObjectID cmio_devices[num_devices];
-        CMIOObjectGetPropertyData(kCMIOObjectSystemObject, &address, 0, NULL, size, &used, &cmio_devices);
+        NSMutableData *cmioDevices = [NSMutableData dataWithLength:size];
+        void *device_data = [cmioDevices mutableBytes];
+        CMIOObjectGetPropertyData(kCMIOObjectSystemObject, &address, 0, NULL, size, &used, device_data);
 
         vcam->deviceID = 0;
-
         NSString *OBSVirtualCamUUID = [[NSBundle bundleWithIdentifier:@"com.obsproject.mac-virtualcam"]
             objectForInfoDictionaryKey:@"OBSCameraDeviceUUID"];
-        for (size_t i = 0; i < num_devices; i++) {
-            CMIOObjectID cmio_device = cmio_devices[i];
+        for (size_t i = 0; i < (used * sizeof(CMIOObjectID)); i++) {
+            CMIOObjectID cmioDevice;
+            [cmioDevices getBytes:&cmioDevice range:NSMakeRange(i * sizeof(CMIOObjectID), sizeof(CMIOObjectID))];
+
             address.mSelector = kCMIODevicePropertyDeviceUID;
-
             UInt32 device_name_size;
-            CMIOObjectGetPropertyDataSize(cmio_device, &address, 0, NULL, &device_name_size);
-
+            CMIOObjectGetPropertyDataSize(cmioDevice, &address, 0, NULL, &device_name_size);
             CFStringRef uid;
-            CMIOObjectGetPropertyData(cmio_device, &address, 0, NULL, device_name_size, &used, &uid);
-
+            CMIOObjectGetPropertyData(cmioDevice, &address, 0, NULL, device_name_size, &used, &uid);
             const char *uid_string = CFStringGetCStringPtr(uid, kCFStringEncodingUTF8);
             if (uid_string && strcmp(uid_string, OBSVirtualCamUUID.UTF8String) == 0) {
-                vcam->deviceID = cmio_device;
+                vcam->deviceID = cmioDevice;
                 CFRelease(uid);
                 break;
             } else {
@@ -404,11 +402,15 @@ static bool virtualcam_output_start(void *data)
 
         address.mSelector = kCMIODevicePropertyStreams;
         CMIOObjectGetPropertyDataSize(vcam->deviceID, &address, 0, NULL, &size);
-        CMIOStreamID stream_ids[(size / sizeof(CMIOStreamID))];
+        NSMutableData *streamIds = [NSMutableData dataWithLength:size];
+        void *stream_data = [streamIds mutableBytes];
+        CMIOObjectGetPropertyData(vcam->deviceID, &address, 0, NULL, size, &used, stream_data);
 
-        CMIOObjectGetPropertyData(vcam->deviceID, &address, 0, NULL, size, &used, &stream_ids);
-
-        vcam->streamID = stream_ids[1];
+        if (used < (2 * sizeof(CMIOStreamID))) {
+            obs_output_set_last_error(vcam->output, obs_module_text("Error.SystemExtension.CameraNotStarted"));
+            return false;
+        }
+        [streamIds getBytes:&vcam->streamID range:NSMakeRange(sizeof(CMIOStreamID), sizeof(CMIOStreamID))];
 
         CMIOStreamCopyBufferQueue(
             vcam->streamID, [](CMIOStreamID, void *, void *) {
--- a/libobs/util/sse-intrin.h
+++ b/libobs/util/sse-intrin.h
@@ -17,6 +17,8 @@
 
 #pragma once
 
+#include "c99defs.h"
+
 #if (defined(_MSC_VER) || defined(__MINGW32__)) && \
 	((defined(_M_X64) && !defined(_M_ARM64EC)) || defined(_M_IX86))
 #include <emmintrin.h>
@@ -25,5 +27,7 @@
 #define WIN32_LEAN_AND_MEAN
 #endif
 #define SIMDE_ENABLE_NATIVE_ALIASES
+PRAGMA_WARN_PUSH
 #include "simde/x86/sse2.h"
+PRAGMA_WARN_POP
 #endif
--- a/plugins/linux-pipewire/pipewire.c
+++ b/plugins/linux-pipewire/pipewire.c
@@ -811,12 +811,21 @@ static void process_video_sync(obs_pipewire_stream *obs_pw_stream)
 
 	if (buffer->datas[0].type == SPA_DATA_DmaBuf) {
 		uint32_t planes = buffer->n_datas;
-		uint32_t offsets[planes];
-		uint32_t strides[planes];
-		uint64_t modifiers[planes];
-		int fds[planes];
+		DARRAY(uint32_t) offsets;
+		DARRAY(uint32_t) strides;
+		DARRAY(uint64_t) modifiers;
+		DARRAY(int) fds;
 		bool use_modifiers;
 		bool corrupt = false;
+
+		da_init(offsets);
+		da_reserve(offsets, planes);
+		da_init(strides);
+		da_reserve(strides, planes);
+		da_init(modifiers);
+		da_reserve(modifiers, planes);
+		da_init(fds);
+		da_reserve(fds, planes);
 
 #ifdef DEBUG_PIPEWIRE
 		blog(LOG_DEBUG,
@@ -838,11 +847,13 @@ static void process_video_sync(obs_pipewire_stream *obs_pw_stream)
 		}
 
 		for (uint32_t plane = 0; plane < planes; plane++) {
-			fds[plane] = buffer->datas[plane].fd;
-			offsets[plane] = buffer->datas[plane].chunk->offset;
-			strides[plane] = buffer->datas[plane].chunk->stride;
-			modifiers[plane] =
-				obs_pw_stream->format.info.raw.modifier;
+			da_push_back(fds, &buffer->datas[plane].fd);
+			da_push_back(offsets,
+				     &buffer->datas[plane].chunk->offset);
+			da_push_back(strides,
+				     &buffer->datas[plane].chunk->stride);
+			da_push_back(modifiers,
+				     &obs_pw_stream->format.info.raw.modifier);
 			corrupt |= (buffer->datas[plane].chunk->flags &
 				    SPA_CHUNK_FLAG_CORRUPTED) > 0;
 		}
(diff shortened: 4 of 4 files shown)


tytan652:
@PatTheMav, @gxalpha if you have a better fix rather than calming the virtualcam warning, please submit it.

PatTheMav:
There are no vla-related warnings in ObjC/ObjC++ code iirc, so I don't think those need to be fixed.

tytan652:
> I _do_ wonder why we warn about variable-length arrays though and why we consider that an error. If it's supposed to be a warning (because of the care it requires to be used correctly), then it shouldn't be an error, because you're not supposed to fix it, but acknowledge it.

`-Wvla` is added explicitly in CMake, simde just put some mess by making them ignored through pragmas. And `-Werror` intervene so they should be fixed at some point.

PatTheMav:
> > I _do_ wonder why we warn about variable-length arrays though and why we consider that an error. If it's supposed to be a warning (because of the care it requires to be used correctly), then it shouldn't be an error, because you're not supposed to fix it, but acknowledge it.
> 
> `-Wvla` is added explicitly in CMake, simde just put some mess by making them ignored through pragmas. And `-Werror` intervene so they should be fixed at some point.
> 
> Edit: I ping you and Sebastian because of the warning that I met in ObjC/ObjC++.

I get that, my question is why we make it a warning in the first place? I never spent much time pondering that and just carried it forward, but the question is whether we actually need the warning.

tytan652:
Related to #1465

PatTheMav:
> > vla must be avoided at all cost
> 
> Seems like Apple disagrees with that take, since I’m fairly sure what we’re doing is exactly what they expect you to do. I’d agree with Pat here, that warning shouldn’t be added then, at least on macOS (or to macOS-only modules).

I'm not so sure about that, an alternative would be to use `NSMutableData`:
```
NSMutableData *cmioDevices = [NSMutableData dataWithLength:size];
CMIOObjectGetPropertyData(kCMIOObjectSystemObject, &address, 0, NULL, size, &used, &cmioDevices);
        
for (size_t i = 0; i < num_devices; i++) {
    CMIOObjectID cmioDevice;
            
    [cmioDevices getBytes:&cmioDevice range:NSMakeRange(i * sizeof(CMIOObjectID), sizeof(CMIOObjectID))];
            
    [...]
}
```
(Or you go more low-level and `malloc`/`free` manually, though `NSMutableData` has the benefit of being ARC managed).

It is correct that Xcode's proposed compiler settings do not warn about VLAs though.

libobs/obs-cocoa.m
```diff
@@ -461,8 +461,11 @@ static bool code_to_str(int code, struct dstr *str)
 
 void obs_key_to_str(obs_key_t key, struct dstr *str)
 {
+#pragma clang diagnostic push
+#pragma clang diagnostic ignored "-Wvla"
     const UniCharCount max_length = 16;
```
PatTheMav, line 466: I wonder if in this case it wouldn't be simpler to just do `UniChar buffer[16]` given that the value is very much static.

plugins/mac-virtualcam/src/obs-plugin/plugin-main.mm
```diff
@@ -370,7 +370,10 @@ static bool virtualcam_output_start(void *data)
                                            .mElement = kCMIOObjectPropertyElementMain};
         CMIOObjectGetPropertyDataSize(kCMIOObjectSystemObject, &address, 0, NULL, &size);
         size_t num_devices = size / sizeof(CMIOObjectID);
+#pragma clang diagnostic push
```
PatTheMav, line 373: Use `NSMutableData` approach instead.

plugins/mac-virtualcam/src/obs-plugin/plugin-main.mm
```diff
@@ -404,7 +407,10 @@ static bool virtualcam_output_start(void *data)
 
         address.mSelector = kCMIODevicePropertyStreams;
         CMIOObjectGetPropertyDataSize(vcam->deviceID, &address, 0, NULL, &size);
+#pragma clang diagnostic push
```
PatTheMav, line 410: Use `NSMutableData` approach instead.

libobs/obs-cocoa.m
PatTheMav: This is not resolved - it replaced a variable-length array with a magic number and preprocessor statements, all of which are superfluous. The number doesn't change, so it can and should be represented in code.

tytan652: No, it doesn't become a VLA since it's a macro and the magic number was already there before the macro statement.

plugins/mac-virtualcam/src/obs-plugin/plugin-main.mm
```diff
@@ -370,15 +370,17 @@ static bool virtualcam_output_start(void *data)
                                            .mElement = kCMIOObjectPropertyElementMain};
         CMIOObjectGetPropertyDataSize(kCMIOObjectSystemObject, &address, 0, NULL, &size);
         size_t num_devices = size / sizeof(CMIOObjectID);
-        CMIOObjectID cmio_devices[num_devices];
-        CMIOObjectGetPropertyData(kCMIOObjectSystemObject, &address, 0, NULL, size, &used, &cmio_devices);
+        NSMutableData *cmioDevices = [NSMutableData dataWithLength:size];
```
PatTheMav, line 373: This code doesn't work - the example I posted was more or less pseudocode, this needs to be implemented properly and tested as well (`CMIOObjectGetPropertyData` expects a void pointer, not an `NSMutableData` pointer).

tytan652: I don't a have a Apple machine so I can't test this code.
I reverted to pragmas, feel free to do a PR which use NSMutableData.

gxalpha: If desired (I think it would be a better fix than throwing `pragma clang diagnostic push` all over the code), I can look into this.

tytan652: It surely does but I don't have what is needed to correctly test that, so feel free.

plugins/mac-virtualcam/src/obs-plugin/plugin-main.mm
PatTheMav, line 373:
```suggestion
        NSMutableData *cmioDevices = [NSMutableData dataWithLength:size];
        void *device_data = [cmioDevices mutableBytes];
        CMIOObjectGetPropertyData(kCMIOObjectSystemObject, &address, 0, NULL, size, &used, device_data);
        
        vcam->deviceID = 0;
        NSString *OBSVirtualCamUUID = [[NSBundle bundleWithIdentifier:@"com.obsproject.mac-virtualcam"] objectForInfoDictionaryKey:@"OBSCameraDeviceUUID"];

        for (size_t i = 0; i < (used * sizeof(CMIOObjectID)); i++) {
            CMIOObjectID cmioDevice;

            [cmioDevices getBytes:&cmioDevice range:NSMakeRange(i * sizeof(CMIOObjectID), sizeof(CMIOObjectID))];
            
            address.mSelector = kCMIODevicePropertyDeviceUID;

            UInt32 device_name_size;
            CMIOObjectGetPropertyDataSize(cmioDevice, &address, 0, NULL, &device_name_size);

            CFStringRef uid;
            CMIOObjectGetPropertyData(cmioDevice, &address, 0, NULL, device_name_size, &used, &uid);

            const char *uid_string = CFStringGetCStringPtr(uid, kCFStringEncodingUTF8);
            if (uid_string && strcmp(uid_string, OBSVirtualCamUUID.UTF8String) == 0) {
                vcam->deviceID = cmioDevice;
                CFRelease(uid);
                break;
            } else {
                CFRelease(uid);
            }
        }
```

plugins/mac-virtualcam/src/obs-plugin/plugin-main.mm
```diff
@@ -370,7 +370,10 @@ static bool virtualcam_output_start(void *data)
                                            .mElement = kCMIOObjectPropertyElementMain};
         CMIOObjectGetPropertyDataSize(kCMIOObjectSystemObject, &address, 0, NULL, &size);
         size_t num_devices = size / sizeof(CMIOObjectID);
+#pragma clang diagnostic push
+#pragma clang diagnostic ignored "-Wvla"
```
PatTheMav, line 374:
```suggestion
        NSMutableData *streamIds = [NSMutableData dataWithLength:size];
        void *stream_data = [streamIds mutableBytes];
        CMIOObjectGetPropertyData(vcam->deviceID, &address, 0, NULL, size, &used, stream_data);
        
        if (used < (2 * sizeof(CMIOStreamID))) {
            obs_output_set_last_error(vcam->output, obs_module_text("Error.SystemExtension.CameraNotStarted"));
            return false;
        }

        [streamIds getBytes:&vcam->streamID range:NSMakeRange(sizeof(CMIOStreamID), sizeof(CMIOStreamID))];
```

plugins/mac-virtualcam/src/obs-plugin/plugin-main.mm
tytan652: Patch applied

plugins/mac-virtualcam/src/obs-plugin/plugin-main.mm
tytan652: Patch applied

PatTheMav:
macOS changes look good to me, if @gxalpha can't find fault with them they might be good to go.

tytan652:
Added `da_reserve`s in PipeWire code.

PR #9344, opened 2023-07-31
mac-avcapture: Update plugin to ObjC and modern Cocoa APIs

PatTheMav:
### Description
This PR effectively reimplements the video capture device support in OBS Studio using ObjC classes and language features and deprecates the old variant to ensure functionality of existing scene setups.

> [!IMPORTANT] 
> This PR has been updated since its original publishing. The changes are detailed in the following section.

#### Updates to the PR

* New "Capture Card Device" source type was added.
    * This source type uses a "fast path" to render video output directly at the internal OBS Studio frame rate and does not use asynchronous video frames like the normal source. This improves frame pacing especially when using capture cards to work with fast-paced video game output.
* All source code has been amended with [DocC](https://developer.apple.com/documentation/docc?changes=la___4_6___8_5_4__3__5_1) style doc-headers, which gives extensive automatic code help in Xcode and would also allow Xcode to automatically generate documentation (and add it to the documentation viewer on a local system)
* Strict aliasing has been enabled for the plugin, adding additional checks for safe pointer aliasing.
* The property refresh handling was simplified into a single callback functions which checks and refreshes all available properties at once, reducing the overall amount of callback calls required when a property refresh is triggered

### Motivation and Context
Application development on macOS has become Swift-oriented, which means that many current and future Frameworks (and associated documentation) make the assumption of Swift being used by default.

To allow OBS Studio to move with the times, existing macOS-specific code needs to be migrated to more Apple-specific coding practices and languages, which can serve as a stepping stone toward full Swift implementations.

Specifically this requires moving C/C++-oriented code to ObjC (and also migrate ObjC++ code to ObjC, as Cocoa/CoreFoundation solutions are usually more elegant and easier to maintain than C++ code).

This also allowed separating plugin code functionality into specific blocks (plugin main entry point, plugin functionality, property API, native ObjC type classes), which makes the code more readable and avoids having a single big source code file.

Native ObjC types and helper functions also allow pure ObjC code to interact more elegantly with native C types provided by `libobs`, which will probably be necessary to correctly bridge to Swift at a later point as well.

> [!NOTE] 
> This PR also makes use of native async queues which allow otherwise slow or blocking operations (creating and starting a capture session in this case) be handled by Cocoa's event loop and  its ephemeral worker threads without the need for `libobs` bespoke threads (function calls exclusively happening off the main thread are not queued async because OBS Studio code expects code in those threads to be sync mostly).

### How Has This Been Tested?
Tested on macOS 13 with built-in Studio Display Camera and Logitech C922 external camera:

* Tested use of presets and distinct device configurations
* Tested attachment and detachment of external devices
* Tested scene setup restoration between OBS launches
* Tested preset changes _after_ device configurations

#### Notable Quirk

> [!NOTE] 
> There is a quirk present in the old video capture source plugin already and has not been fixed in the rewrite (I attempted to do so but this created an entire new slew of issues):

When a device is initialised with a _preset_, the sample buffer dimensions provided by `AVFoundation` is defined by that preset. This will be independent from the actual device resolution:

1. Select "1280x720" preset
2. Observe sample buffer resolution of 1280x720 pixels
3. Change internal _device resolution_ to 160x120 pixels
4. Observe sample buffer resolution is kept at 1280x720 pixels, but image is pillarboxed and upscaled from the 160x120 device image
5. Select "1920x1280" preset
6. Observe sample buffer resolution is changed to 1920x1080 pixels, but image is pillarboxed and upscaled still

This does noticeably not happen if preset usage is disabled _before_ selecting a device, as the capture session will be set up with a device format directly. Interestingly the sample buffer provided by `AVFoundation` will then keep track with the actual device resolution, but once presets are used, the buffer size is determined by that entirely.

#### No Automatic Selection of Properties

> [!IMPORTANT] 
> The automatic property selection of the old plugin has not been reimplemented, requiring users to explicitly select the colour format, frame rate, and other properties when not using presets. Presets are the canonical and best way to use video capture input.

#### Quirks with iOS Devices

Using iOS devices as camera sources via USB still works (including muxed audio), but retains some notable quirks:

* The device will not be automatically rediscovered/enabled between OBS restarts. 
* Only presets can be used, the device will not expose any specific resolutions or framerates
* Switching between landscape and portrait mode is supported

When used just for camera functionality, "Continuity Camera" should be preferred and capture via USB used for capture of screen contents.

### Types of changes

### Checklist:


jcm93:
I've put this through what paces I can with my few capture devices. Overall it seems to be functioning well, offering a noticeable improvement in latency in both the capture card device source as well as the new video capture device source! I hope this sees more testing from a wider base of Mac users, especially those with capture cards using other color formats, as my ability to test there was quite limited, but I did not see any regressions among my devices.

My primary criticism of the new sources is the parameter selection in the UI in non-preset mode. I have not analyzed the code to see why this hasn't been reimplemented, but it is somewhat tedious when switching devices to manually select all parameters, particularly when there is indeed only one valid parameter to choose from. Additionally, the FPS selector will always switch to rational input mode when switching devices, which seems perhaps unnecessary. In real-world use, this source is generally "set and forget", so this isn't a huge issue, just slightly tedious.

I'm also wondering if the new "Capture Card Device" is sufficiently different from the Video Capture Device source that it necessitates a second source type, instead of perhaps living as a toggle underneath Video Capture Device. I haven't analyzed the code here to see why this choice was made, but I do see this potentially introducing confusion from users as to what source type is appropriate for their uses.

PatTheMav:
> My primary criticism of the new sources is the parameter selection in the UI in non-preset mode. I have not analyzed the code to see why this hasn't been reimplemented, but it is somewhat tedious when switching devices to manually select all parameters, particularly when there is indeed only one valid parameter to choose from. 

That is by design - one of the main issues the rewrite attempts to fix is the dreaded "auto selection" feature that is to be phased out from the code base (because it is hard to maintain and never worked well to begin with). On top of that the values that were automatically selected by the prior implementation actually lead to colour inaccuracies because of forced multiple conversions (CMIO converting the native camera format to what the source thinks is "good" and then another conversion into what OBS uses).

In that regard not using a preset is considered an advanced feature and as such users should specify explicitly how they want the device to be configured.

_Another_ aspect is that any configuration made here affects all other applications also using the device, so OBS should not automatically change anything by itself.

> Additionally, the FPS selector will always switch to rational input mode when switching devices, which seems perhaps unnecessary. In real-world use, this source is generally "set and forget", so this isn't a huge issue, just slightly tedious.

This seems to be an unfortunate deficiency of OBS' property system, it automatically switches to this variant and I haven't found a way to prevent this yet.

> I'm also wondering if the new "Capture Card Device" is sufficiently different from the Video Capture Device source that it necessitates a second source type, instead of perhaps living as a toggle underneath Video Capture Device. I haven't analyzed the code here to see why this choice was made, but I do see this potentially introducing confusion from users as to what source type is appropriate for their uses.

The capture card source' lower latency and better frame pacing requires a UVC device to also have even frame pacing itself, otherwise it could potentially lead to worse visual quality on a stream. It's for that reason that the functionality is separated between two source types to make it clear that the new variant is better suited for capture cards than webcams.

PatTheMav:
I checked the OBS property code again and this seems to be by design: When the property window is refreshed and the frame rate selected for the prior device cannot be found in the list of available frame rates for the new device, the property automatically switches to the "rational" variant.

Only if the old frame rate is also available for the new device will the view stay in simple mode. I'd argue that this is a deficiency in the existing code (why should the property switch to rational frame rate selection just because the current one isn't supported?) but it's not for any plugin to change but needs to be changed in OBS' frontend code.

MrVelidi:
Hello @PatTheMav. I have tested the build artifact for apple silicon present on the PR for hardware combination of Macbook Pro with M3 Max with Elgato HD60X capture card. 

I have initially been facing the stutters with the latest stable and tried to get in touch with the community on discord and @yam has pointed out that i can test this build as he suspects that this would fix the issues. Regarding stutter with current stable while recording, i observed one frame drop immediately followed by 2 identical frames which the eye perceives as stutter, and also the output lagged randomly during game play.

Regarding the testing details:

Hardware:
Laptop: Macbook Pro with M3 Max
Capture Card: Elgato HD60X

Capture Details:
Source: Playstation 5
Resolution: 1440p 60fps
Encoders Used: H264 and H265
Bitrate Used: 60000 kbps
Duration of Capture: 15 mins minimum each.

Result: The result across both the captures are similar. The stutter is not observed what so ever. At least i could not make out any glitch like scenarios while recording. However on both the tests, i could observe that the lag is creeping in randomly and when it does, it continues for a 1-2 mins and automatically adjusts itself and goes back to clean output. During this 15 mins recording i could see the lag like 2-3 times but at random locations in the video. Yesterday i did a first test of 35 mins using H265 with same bitrate mentioned above, i could see the lag appearing at 4-5 different locations in the video but its short lived under 2 mins before it goes back to normal. 

All in all, this build is by far the most impressive one i have seen so far from OBS on Macbook Pro despite the occasional lag.

@yam form discord has encouraged me to add the results here so that it would help you guys. Cheers !

gxalpha:
(just fyi, you're looking to ping @jcm93 instead of `github.com/yam`)

PatTheMav:
@Warchamp7 Off-thread we discussed whether to move the frame rate selection in the property because in some situations it will only be populated with valid values if other properties have been changed to use valid values (e.g. colour format and input format).

However I also found many instances where the opposite is true (e.g. the selected frame rate  is correct, but the colour format doesn't match and requires adjustment), so changing the order of properties will always be detrimental to _some_ use-cases and just using the frame rate selection to the bottom is not beneficial in all cases.

WDYT?

plugins/mac-avcapture/OBSAVCapturePresetInfo.h
```diff
@@ -0,0 +1,24 @@
+//
+//  OBSAVCapturePresetInfo.h
+//  mac-avcapture
+//
+//  Created by Patrick Heyer on 07.03.23.
```
RytoEX, line 5: Could we make the date less ambiguous by using an ISO 8601 date (YYYY-MM-DD), here and elsewhere?

RytoEX (changes requested):
A comment nit.

plugins/mac-avcapture/OBSAVCapturePresetInfo.h
PatTheMav: Sure, it was autogenerated by Xcode so I didn't pay much attention.

RytoEX (approved):
Discussed with @PatTheMav , @gxalpha , and @jcm93 off-thread. What this mostly needs now is wider testing, and the best way to do that presently is to merge it and let users play with it. Since this is macOS-only, and it doesn't remove the existing source, the impact of merging should be minimal.

PR #9349, opened 2023-08-02
UI: Don't call obs_source_get_output_flags on a NULL source

gxalpha:
### Description

https://github.com/obsproject/obs-studio/commit/c33fa8b introduced a call to `obs_source_get_output_flags` which would always be called when the selection changes, even if things get unselected. This would result in `get_output_flags` being called on a `NULL` source which libobs yells about. This PR also inverts the `audioOnly` bool to make it a bit more obvious what is actually being checked (which is whether the source has video).
This PR leaves an existing bug w/r/t multiple scene items which I'm planning on addressing as well, but it may involve a minor refactor and I don't want this PR to get too large.

### Motivation and Context

Got annoyed at debug warnings each time I unselect the current scene item.

### How Has This Been Tested?

macOS 14
Tested that everything (including the bugs) works the same as before, with both video sources as well as audio only sources.

### Types of changes

### Checklist:

### DIFF
--- a/UI/window-basic-main.cpp
+++ b/UI/window-basic-main.cpp
@@ -3391,16 +3391,20 @@ void OBSBasic::SourceToolBarActionsSetEnabled()
 
 void OBSBasic::UpdateTransformShortcuts()
 {
+	bool hasVideo = false;
+
 	OBSSource source = obs_sceneitem_get_source(GetCurrentSceneItem());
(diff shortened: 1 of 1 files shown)


norihiro:
I feel `hasVideo` would be a better variable name than `hasTransform` for a future new feature to implement new menu-items only available for video source.

gxalpha:
You're right, `hasVideo` is probably even better. Changed.

PR #9350, opened 2023-08-02
Move non-deps and shared UI code in shared folder

tytan652:
### Description

- Move non-deps module from the `deps` folder to the `shared` folder
- Move UI code that was used in various frontend-plugins as module in the `shared` folder

~~Find Qt with versionless removed to avoid this issue on Ubuntu CI:~~ (fixed)
<!-- ```
  CMake Error at /usr/lib/x86_64-linux-gnu/cmake/Qt6Core/Qt6CoreVersionlessTargets.cmake:37 (message):
    Some (but not all) targets in this export set were already defined.
  
    Targets Defined: Qt::Core
  
    Targets not yet defined: Qt::CorePrivate
  
  Call Stack (most recent call first):
    /usr/lib/x86_64-linux-gnu/cmake/Qt6Core/Qt6CoreConfig.cmake:63 (include)
    /usr/lib/x86_64-linux-gnu/cmake/Qt6/Qt6Config.cmake:219 (find_package)
    cmake/Modules/ObsHelpers.cmake:333 (find_package)
    shared/properties-view/CMakeLists.txt:3 (find_qt)
``` -->

### Motivation and Context
<!--- Why is this change required? What problem does it solve? -->
<!--- If it fixes an open GitHub Issue, or implements feature request -->
<!--- from the Ideas page, please link to the issue here. -->

Based on ideas from this https://github.com/obsproject/obs-studio/pull/9220#issuecomment-1640500483

> I would also suggest not adding new things to the "deps" directory - its purpose was for vendors external dependencies, but this PR extends its use for "shared modules". I'd much rather have us use a new directory called "shared", "common", or some other that communicates this purpose.

> The solution to this would be to move UI elements shared by modules/plugins to be moved into the same "shared" directory and be made a standalone library

> That would also fix this outstanding issue with the frontend plugins which "magically" pull in sources from the UI source tree that they shouldn't actually know about

### How Has This Been Tested?
<!--- Please describe in detail how you tested your changes. -->
<!--- Include details of your testing environment (hardware, OS version, etc.),-->
<!--- and the tests you ran, including how it may affect other areas of code. -->

- Builds
- PropertiesView are translated correctly.

### Types of changes
<!--- What types of changes does your PR introduce? Uncomment all that apply -->
<!--- - Bug fix (non-breaking change which fixes an issue) -->
<!--- - New feature (non-breaking change which adds functionality) -->
<!--- - Performance enhancement (non-breaking change which improves efficiency) -->
<!--- - Breaking change (fix or feature that would cause existing functionality to change) -->
<!--- - Documentation (a change to documentation pages) -->

### Checklist:
<!--- Go over all the following points, and put an `x` in all the boxes that apply. -->
<!--- If you're unsure about any of these, don't hesitate to ask. We're here to help! -->

### CODE REVIEW

**Comment by tytan652** [comment]
I will push commit by commit to check if each commit build on Windows and macOS.

File: .gitignore [hunk 1]
```diff
@@ -9,6 +9,7 @@
 !/docs
 !/libobs*
 !/plugins
+!/shared
```
PatTheMav, line 12: @RytoEX The entire PR kinda hinges on us agreeing on a directory name for all the modules/headers/elements shared between libobs/plugins/UI, etc.

I'm indifferent to using "shared" but this should be agreed on quick.

UI/qt-display.cpp
```diff
@@ -62,6 +70,38 @@ static inline QColor rgba_to_color(uint32_t rgba)
 			       (rgba >> 16) & 0xFF, (rgba >> 24) & 0xFF);
 }
 
+static bool QTToGSWindow(QWindow *window, gs_window &gswindow)
```
PatTheMav, line 73: This looks like new code?

cmake/Modules/ObsHelpers.cmake
```diff
@@ -312,12 +312,6 @@ endfunction()
 macro(find_qt)
   set(multiValueArgs COMPONENTS COMPONENTS_WIN COMPONENTS_MAC COMPONENTS_LINUX)
   cmake_parse_arguments(FIND_QT "" "${oneValueArgs}" "${multiValueArgs}" ${ARGN})
-  set(QT_NO_CREATE_VERSIONLESS_TARGETS ON)
```
PatTheMav, line 315: Why is this removed?

deps/CMakeLists.txt
```diff
@@ -1,14 +1,8 @@
-if(OS_WINDOWS)
-  add_subdirectory(ipc-util)
+if(OS_WINDOWS AND NOT MINGW)
```
PatTheMav, line 1: We do not support compiling obs-studio via MINGW, so any breakage incurred here is desired.

shared/obs-scripting/cmake/legacy.cmake
```diff
@@ -176,3 +176,6 @@ target_compile_definitions(obs-scripting PRIVATE SCRIPT_DIR="${OBS_SCRIPT_PLUGIN
                                                  $<$<BOOL:${ENABLE_UI}>:ENABLE_UI>)
 
 setup_binary_target(obs-scripting)
+
+# Dirty workaround: CMake 2.0 seems to fail without this file
+file(TOUCH "${CMAKE_BINARY_DIR}/shared/cmake_install.cmake")
```
PatTheMav, line 181: That's probably because without its own `CMakeLists.txt` the `shared` directory has no scope within the CMake project and the file is not generated. Should be fine as a hack for the time until we remove the legacy path entirely.

cmake/Modules/ObsHelpers.cmake
tytan652: Read the PR description which explain why.

UI/qt-display.cpp
tytan652: Moved see the commit for more context about this "new code", QTToGSWindow is only used once by the UI there is no point keeping it in the wrapper.

shared/properties-view/properties-view.cpp
```diff
@@ -25,21 +25,30 @@
...
+#include <plain-text-edit.hpp>
+#include <slider-ignorewheel.hpp>
 #include <cstdlib>
 #include <initializer_list>
 #include <obs-data.h>
 #include <obs.h>
 #include <qtimer.h>
 #include <string>
 
+#ifdef INSIDE_UI
```
PatTheMav, line 42: Wouldn't it make more sense to include the frontend-API by default and use `obs_frontend_is_theme_dark` instead of making the distinction?

shared/properties-view/properties-view.hpp
```diff
@@ -143,6 +144,19 @@ class OBSPropertiesView : public VScrollArea {
...
+
+	static QWidget *
+	CreateRationalFPS(OBSFrameRatePropertyWidget *fpsProps, bool &selected,
+			  const media_frames_per_second *current_fps);
+	static OBSFrameRatePropertyWidget *
+	CreateFrameRateWidget(obs_property_t *prop, bool &warning,
+			      const char *option,
+			      media_frames_per_second *current_fps,
+			      frame_rate_ranges_t &fps_ranges);
+
```
PatTheMav, line 159: Why is this addition needed?

shared/qt/plain-text-edit/CMakeLists.txt
```diff
@@ -0,0 +1,11 @@
+cmake_minimum_required(VERSION 3.16...3.25)
+
+find_qt(COMPONENTS Core Widgets)
```
PatTheMav, line 3: `find_qt` is effectively deprecated, do not use it for new code.

deps/CMakeLists.txt
tytan652: Looks like it's a remnant of the past for `w32-pthread` through my rebase.
Because while I rebased `NOT MINGW` was there to skip `w32-pthread`.

cmake/Modules/ObsHelpers.cmake
PatTheMav: The underlying issue seems to be that the properties view seems to run `find_qt` twice in the same scope, which was never supported (we never had that issue so far), so the question is why Qt is looked for twice (otherwise it wouldn't run into the collision). That should be investigated/fixed first.

shared/properties-view/properties-view.cpp
tytan652: In my opinion, UI (`obs-studio`) must not depend on obs-frontend-api since it's implementation/interface is provided by the former.

shared/properties-view/properties-view.cpp
```diff
@@ -106,7 +115,7 @@ void OBSPropertiesView::ReloadProperties()
 	RefreshProperties();
 }
 
-#define NO_PROPERTIES_STRING QTStr("Basic.PropertiesWindow.NoProperties")
+#define NO_PROPERTIES_STRING tr("Basic.PropertiesWindow.NoProperties")
```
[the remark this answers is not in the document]
tytan652: `Q_OBJECT` macro in classes definition.

shared/properties-view/properties-view.hpp
tytan652: Mainly to access the `tr` function.

deps/CMakeLists.txt
tytan652: Should be good now.

.gitignore
RytoEX: Acknowledging. Let's discuss off-thread.

cmake/Modules/ObsHelpers.cmake
PatTheMav: The root cause for the issue is that if you add the subdirectory _after_ the `find_qt` call, then the already found Qt is inherited by the subdirectory (and thus you run into the collision).

If you add the subdirectory _before_ `find_qt`, then the targets stay in the subdirectory context and are not populated up. This would apply for every instance where the shared library is added after `find_qt`.

This obviously only applies to legacy CMake files and websockets/browser which have not been upgraded yet.

tytan652: It works, thank you pushed the changes.

shared/properties-view/properties-view.cpp
PatTheMav: I don't think that using an API internally is an issue - it would be a "bad" API if that weren't viable and this is precisely the use case for such an API that abstracts the implementation detail of the state variable.

IMO it's desirable to use the API instead of accessing global objects directly (that the implementation should not even know about in the first place).

shared/properties-view/properties-view.cpp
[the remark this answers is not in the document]
PatTheMav: Ah, so it's a Qt built-in?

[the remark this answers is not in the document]
tytan652: Yes `tr()` is the common static class function to have the localized version of a string in Qt, and the behavior of the function is already overridden inside the UI to use our translation function.

shared/properties-view/properties-view.cpp
tytan652: Since I want this to get merged, I'll go with removing `INSIDE_UI` and use only the API.

shared/properties-view/properties-view.hpp
tytan652: Replace this by using `QObject::tr` in static functions.

shared/properties-view/CMakeLists.txt
```diff
@@ -0,0 +1,51 @@
+cmake_minimum_required(VERSION 3.16...3.25)
```
PatTheMav, line 1: Our lower boundary is 3.22 iirc?

UI/CMakeLists.txt
```diff
@@ -20,6 +20,10 @@ if(NOT TARGET OBS::json11)
   add_subdirectory("${CMAKE_SOURCE_DIR}/deps/json11" "${CMAKE_BINARY_DIR}/deps/json11")
 endif()
 
+if(NOT TARGET OBS::qt-wrappers)
```
PatTheMav, line 23: This block (as well as adding the library target as a dependency) should be moved into `ui-qt.cmake` for encapsulation.

shared/properties-view/CMakeLists.txt
```diff
@@ -0,0 +1,51 @@
...
+  add_subdirectory("${CMAKE_SOURCE_DIR}/shared/qt/vertical-scroll-area"
+                   "${CMAKE_BINARY_DIR}/shared/qt/vertical-scroll-area")
+endif()
+
+if(NOT TARGET OBS::qt-slider-ignorewheel)
+  add_subdirectory("${CMAKE_SOURCE_DIR}/shared/qt/slider-ignorewheel"
+                   "${CMAKE_BINARY_DIR}/shared/qt/slider-ignorewheel")
+endif()
+
+add_library(properties-view INTERFACE)
```
PatTheMav, line 23: It might be more beneficial to make this (and the other qt-based) shared modules `OBJECT` libraries so they only have to be compiled once. `INTERFACE` type libraries just introduce the source and header files to the target, extending/duplicating compilation effort.

shared/properties-view/CMakeLists.txt
tytan652: I forgot to keep that up to date.

UI/CMakeLists.txt
tytan652: Done

shared/properties-view/CMakeLists.txt
PatTheMav: Just attempted this and it's not possible without a rewrite of the affected classes - they are all too intertwined to be compiled as independent units.

tytan652: Unfortunately, Qt needs to use at least MOC on those files (creating another object file) and this cause trouble in CMake (mising symbol from MOC objects) so I will not dig deeper in the scope of this PR.

RytoEX:
I've looked over this, and the one thing that gives me pause is the QTStr -> tr change in `shared/properties-view/properties-view.cpp`. This feels a little out-of-scope for this PR, and feels like it would make more sense as a follow-up PR.

The [`QObject::tr()`](https://doc.qt.io/qt-6/qobject.html#tr) docs say:
> Returns a translated version of sourceText, optionally based on a disambiguation string and value of n for strings containing plurals; otherwise returns [QString::fromUtf8](https://doc.qt.io/qt-6/qstring.html#fromUtf8)(sourceText) if no appropriate translated string is available.

So it seems like this function call would necessary introduce some additional overhead (which may be negligible) as it tries and fails to find a translated string in Qt's built-in locale/translation system.

tytan652:
> This feels a little out-of-scope for this PR, and feels like it would make more sense as a follow-up PR.

It is not out scope since the change is required to move out properties view code from the UI folder since it's used by various elements. `QTStr` is exclusive to OBS Studio.

As it was a said (IIRC) in a previous review, we already override Qt default translation system to use ours (`OBSTranslator`) so a `tr()` call with lead to text lookup in our system.

tt2468:
I think `common` would be a better name for these specific cases. Otherwise, I like this

tytan652:
> I think `common` would be a better name for these specific cases.

There is this thread https://github.com/obsproject/obs-studio/pull/9350#discussion_r1453982017 in the PR, about the name. It was resolved, so I guess it was indeed discussed.

If the name should be (re-)discussed, I would prefer it being done off-thread with a clear consensus on one name.

PatTheMav:
> > I think `common` would be a better name for these specific cases.
> 
> There is this thread [#9350 (comment)](https://github.com/obsproject/obs-studio/pull/9350#discussion_r1453982017) in the PR, about the name. It was resolved, so I guess it was indeed discussed.
> 
> If the name should be (re-)discussed, I would prefer it being done off-thread with a clear consensus on one name.

The result of the discussion boiled down to "we're not entirely happy with 'shared' but also haven't come up with a compelling alternative". Also given that it should be trivial to change this later, we didn't necessarily wanted to block the PR over it.

RytoEX (approved):
LGTM, and @PatTheMav has said off-thread that this is fine as-is. Neither of us have come up with a preferred alternative to "shared", so "shared" it will be. Rebase this and I'll use a merge commit.

PR #9352, opened 2023-08-02
obs-ffmpeg: New default settings for AMD encoders

rhutsAMD:
### Description

Introducing new default AMD encoder settings for improved perceptual quality for AVC/HEVC/AV1. These new default settings have been tuned to target stable recording on RX 5700 XT and up on the recent driver.

|                      |                     AVC                      |                     AVC                      |                     HEVC                     |                     HEVC                     |                     AV1                      |                     AV1                      |
| -------------------- | :------------------------------------------: | :------------------------------------------: | :------------------------------------------: | :------------------------------------------: | :------------------------------------------: | :------------------------------------------: |
|                      |               Up to 1080p60fps               |                 > 1080p60fps                 |               Up to 1080p60fps               |                 > 1080p60fps                 |               Up to 1080p60fps               |                 > 1080p60fps                 |
| RATE_CONTROL_METHOD  |                     CBR                      |                     CBR                      |                     CBR                      |                     CBR                      |                     CBR                      |                     CBR                      |
| PEAK_BITRATE         |           Same as   TARGET_BITRATE           |            Same as TARGET_BITRATE            |            Same as TARGET_BITRATE            |            Same as TARGET_BITRATE            |            Same as TARGET_BITRATE            |            Same as TARGET_BITRATE            |
| VBV_BUFFER_SIZE      |           Same as   TARGET_BITRATE           |            Same as TARGET_BITRATE            |            Same as TARGET_BITRATE            |            Same as TARGET_BITRATE            |            Same as TARGET_BITRATE            |            Same as TARGET_BITRATE            |
| FILLER_DATA_ENABLE   |                     TRUE                     |                     TRUE                     |                     TRUE                     |                     TRUE                     |                     TRUE                     |                     TRUE                     |
| ENFORCE_HRD          |                     TRUE                     |                     TRUE                     |                     TRUE                     |                     TRUE                     |                     TRUE                     |                     TRUE                     |
| MAX_B_FRAMES         |                      2                       |                      0                       |                      -                       |                      -                       |                      2                       |                      0                       |
| B_PIC_PATTERN        |                      2                       |                      0                       |                      -                       |                      -                       |                      2                       |                      0                       |
| SCREEN_CONTENT_TOOLS |                      -                       |                      -                       |                      -                       |                      -                       |                     TRUE                     |                     TRUE                     |
| PALETTE_MODE         |                      -                       |                      -                       |                      -                       |                      -                       |                     TRUE                     |                     TRUE                     |
| PRESET               |                   Quality                    |                   Quality                    |                   Quality                    |                   Quality                    |                 high quality                 |                   Balanced                   |
| PROFILE              |                     high                     |                     high                     |                 main/main10                  |                 main/main10                  |                     main                     |                     main                     |
| ENABLE_VBAQ/AQ_MODE  | TRUE for RCMethod != (CQP \| HQVBR \| HQCBR) | TRUE for RCMethod != (CQP \| HQVBR \| HQCBR) | TRUE for RCMethod != (CQP \| HQVBR \| HQCBR) | TRUE for RCMethod != (CQP \| HQVBR \| HQCBR) | TRUE for RCMethod != (CQP \| HQVBR \| HQCBR) | TRUE for RCMethod != (CQP \| HQVBR \| HQCBR) |
| ADAPTIVE_MINIGOP     |     TRUE for B-frames > 0 and PA enabled     |     TRUE for B-frames > 0 and PA enabled     |                   Not set                    |                   Not set                    |     TRUE for B-frames > 0 and PA enabled     |     TRUE for B-frames > 0 and PA enabled     |

### Motivation and Context

This PR provides new recommended defaults for the AMD encoders that have been optimized for perceptual quality. The settings are automatically applied depending on the resolution and framerate. The default settings can be overridden as expected and only provide an improved base foundation for encoder settings targeted at improving perceptual quality.

### How Has This Been Tested?

Tested on RX 5000, 6000, and 7000 series cards using public driver version 24.2.1.

### Types of changes

### Checklist:

### DIFF
--- a/plugins/obs-ffmpeg/texture-amf.cpp
+++ b/plugins/obs-ffmpeg/texture-amf.cpp
@@ -1160,14 +1160,36 @@ static void check_texture_encode_capability(obs_encoder_t *encoder, amf_codec_ty
 
 #include "texture-amf-opts.hpp"
 
-static void amf_defaults(obs_data_t *settings)
+static void amf_avc_defaults(obs_data_t *settings)
 {
+	// These are initial recommended settings that may be lowered later
+	// once we know more info such as the resolution and frame rate.
+	obs_data_set_default_string(settings, "rate_control", "CBR");
 	obs_data_set_default_int(settings, "bitrate", 2500);
 	obs_data_set_default_int(settings, "cqp", 20);
-	obs_data_set_default_string(settings, "rate_control", "CBR");
 	obs_data_set_default_string(settings, "preset", "quality");
 	obs_data_set_default_string(settings, "profile", "high");
-	obs_data_set_default_int(settings, "bf", 3);
+	obs_data_set_default_int(settings, "bf", 2);
+}
+
+static void amf_hevc_defaults(obs_data_t *settings)
+{
+	// These are initial recommended settings that may be lowered later
+	// once we know more info such as the resolution and frame rate.
+	obs_data_set_default_string(settings, "rate_control", "VBR");
+	obs_data_set_default_int(settings, "bitrate", 2500);
+	obs_data_set_default_int(settings, "cqp", 20);
+	obs_data_set_default_string(settings, "preset", "quality");
+}
+
+static void amf_av1_defaults(obs_data_t *settings)
+{
+	// These are initial recommended settings that may be lowered later
+	// once we know more info such as the resolution and frame rate.
+	obs_data_set_default_int(settings, "bitrate", 2500);
+	obs_data_set_default_int(settings, "cqp", 20);
+	obs_data_set_default_string(settings, "rate_control", "VBR");
+	obs_data_set_default_string(settings, "preset", "highQuality");
 }
 
 static bool rate_control_modified(obs_properties_t *ppts, obs_property_t *p, obs_data_t *settings)
@@ -1326,6 +1348,15 @@ static void amf_avc_update_data(amf_base *enc, int rc, int64_t bitrate, int64_t 
 
 		if (rc == AMF_VIDEO_ENCODER_RATE_CONTROL_METHOD_CBR) {
 			set_avc_property(enc, FILLER_DATA_ENABLE, true);
+		} else if (rc == AMF_VIDEO_ENCODER_RATE_CONTROL_METHOD_PEAK_CONSTRAINED_VBR ||
+			   rc == AMF_VIDEO_ENCODER_RATE_CONTROL_METHOD_HIGH_QUALITY_VBR) {
+			set_avc_property(enc, PEAK_BITRATE, bitrate * 1.5);
+			set_avc_property(enc, VBV_BUFFER_SIZE, bitrate * 1.5);
+		} else if (rc ==
+			   AMF_VIDEO_ENCODER_RATE_CONTROL_METHOD_LATENCY_CONSTRAINED_VBR) {
+			int64_t framerate = enc->fps_num / enc->fps_den;
+			set_avc_property(enc, VBV_BUFFER_SIZE,
+					 (bitrate / framerate) * 1.1);
 		}
 	} else {
 		set_avc_property(enc, QP_I, qp);
@@ -1366,6 +1397,23 @@ try {
 	amf_base *enc = (amf_base *)data;
 	error("%s: %s: %ls", __FUNCTION__, err.str, amf_trace->GetResultText(err.res));
 	return false;
+}
+
+static inline void adjust_recommended_avc_defaults(amf_base *enc,
+						   obs_data_t *settings)
+{
+	int64_t framerate = enc->fps_num / enc->fps_den;
+	if ((enc->cx * enc->cy > 1920 * 1088) || (framerate > 60)) {
+
+		// Recommended base defaults
+		obs_data_set_default_int(settings, "bitrate", 2500);
+		obs_data_set_default_int(settings, "cqp", 20);
+		obs_data_set_default_string(settings, "rate_control", "CBR");
+		obs_data_set_default_string(settings, "preset", "quality");
+		obs_data_set_default_string(settings, "profile", "high");
+		obs_data_set_default_int(settings, "bf", 0);
+		info("Original base default settings were used according to resolution and framerate.");
+	}
 }
 
 static void amf_set_codec_level(amf_base *enc)
@@ -1454,6 +1502,10 @@ static bool amf_avc_init(void *data, obs_data_t *settings)
 {
 	amf_base *enc = (amf_base *)data;
 
+	// Initial high perceptual quality settings are set in the UI.
+	// Adjust during init based on resolution and framerate.
+	adjust_recommended_avc_defaults(enc, settings);
+
 	int64_t bitrate = obs_data_get_int(settings, "bitrate");
 	int64_t qp = obs_data_get_int(settings, "cqp");
 	const char *preset = obs_data_get_string(settings, "preset");
@@ -1483,7 +1535,9 @@ static bool amf_avc_init(void *data, obs_data_t *settings)
 	int rc = get_avc_rate_control(rc_str);
 
 	set_avc_property(enc, RATE_CONTROL_METHOD, rc);
-	if (rc != AMF_VIDEO_ENCODER_RATE_CONTROL_METHOD_CONSTANT_QP)
+	if (rc != AMF_VIDEO_ENCODER_RATE_CONTROL_METHOD_CONSTANT_QP &&
+	    rc != AMF_VIDEO_ENCODER_RATE_CONTROL_METHOD_HIGH_QUALITY_VBR &&
+	    rc != AMF_VIDEO_ENCODER_RATE_CONTROL_METHOD_HIGH_QUALITY_CBR)
 		set_avc_property(enc, ENABLE_VBAQ, true);
 
 	amf_avc_update_data(enc, rc, bitrate * 1000, qp);
@@ -1681,7 +1735,7 @@ static void register_avc()
 	amf_encoder_info.destroy = amf_destroy;
 	amf_encoder_info.update = amf_avc_update;
 	amf_encoder_info.encode_texture = amf_encode_tex;
-	amf_encoder_info.get_defaults = amf_defaults;
+	amf_encoder_info.get_defaults = amf_avc_defaults;
 	amf_encoder_info.get_properties = amf_avc_properties;
 	amf_encoder_info.get_extra_data = amf_extra_data;
 	amf_encoder_info.caps = OBS_ENCODER_CAP_PASS_TEXTURE | OBS_ENCODER_CAP_DYN_BITRATE | OBS_ENCODER_CAP_ROI;
@@ -1748,6 +1802,15 @@ static void amf_hevc_update_data(amf_base *enc, int rc, int64_t bitrate, int64_t
 
 		if (rc == AMF_VIDEO_ENCODER_HEVC_RATE_CONTROL_METHOD_CBR) {
 			set_hevc_property(enc, FILLER_DATA_ENABLE, true);
+		} else if (rc == AMF_VIDEO_ENCODER_HEVC_RATE_CONTROL_METHOD_PEAK_CONSTRAINED_VBR ||
+			   rc == AMF_VIDEO_ENCODER_HEVC_RATE_CONTROL_METHOD_HIGH_QUALITY_VBR) {
+			set_hevc_property(enc, PEAK_BITRATE, bitrate * 1.5);
+			set_hevc_property(enc, VBV_BUFFER_SIZE, bitrate * 1.5);
+		} else if (rc ==
+			   AMF_VIDEO_ENCODER_HEVC_RATE_CONTROL_METHOD_LATENCY_CONSTRAINED_VBR) {
+			int64_t framerate = enc->fps_num / enc->fps_den;
+			set_hevc_property(enc, VBV_BUFFER_SIZE,
+					  (bitrate / framerate) * 1.1);
 		}
 	} else {
 		set_hevc_property(enc, QP_I, qp);
@@ -1789,9 +1852,28 @@ try {
 	return false;
 }
 
+static inline void adjust_recommended_hevc_defaults(amf_base *enc,
+						    obs_data_t *settings)
+{
+	const bool is10bit = enc->amf_format == AMF_SURFACE_P010;
+	const int64_t framerate = enc->fps_num / enc->fps_den;
+	if ((enc->cx * enc->cy > 1920 * 1088) || is10bit || (framerate > 60)) {
+
+		// Recommended base defaults
+		obs_data_set_default_int(settings, "bitrate", 2500);
+		obs_data_set_default_int(settings, "cqp", 20);
+		obs_data_set_default_string(settings, "preset", "quality");
+		info("Original base default settings were used according to resolution and framerate.");
+	}
+}
+
 static bool amf_hevc_init(void *data, obs_data_t *settings)
 {
 	amf_base *enc = (amf_base *)data;
+
+	// Initial high perceptual quality settings are set in the UI.
+	// Adjust during init based on resolution and framerate.
+	adjust_recommended_hevc_defaults(enc, settings);
 
 	int64_t bitrate = obs_data_get_int(settings, "bitrate");
 	int64_t qp = obs_data_get_int(settings, "cqp");
@@ -1801,7 +1883,9 @@ static bool amf_hevc_init(void *data, obs_data_t *settings)
 	int rc = get_hevc_rate_control(rc_str);
 
 	set_hevc_property(enc, RATE_CONTROL_METHOD, rc);
-	if (rc != AMF_VIDEO_ENCODER_HEVC_RATE_CONTROL_METHOD_CONSTANT_QP)
+	if (rc != AMF_VIDEO_ENCODER_HEVC_RATE_CONTROL_METHOD_CONSTANT_QP &&
+	    rc != AMF_VIDEO_ENCODER_HEVC_RATE_CONTROL_METHOD_HIGH_QUALITY_VBR &&
+	    rc != AMF_VIDEO_ENCODER_HEVC_RATE_CONTROL_METHOD_HIGH_QUALITY_CBR)
 		set_hevc_property(enc, ENABLE_VBAQ, true);
 
 	amf_hevc_update_data(enc, rc, bitrate * 1000, qp);
@@ -1903,6 +1987,7 @@ static void amf_hevc_create_internal(amf_base *enc, obs_data_t *settings)
 	const bool hlg = is_hlg(enc);
 	const bool is_hdr = pq || hlg;
 	const char *preset = obs_data_get_string(settings, "preset");
+	obs_data_set_string(settings, "profile", is10bit ? "main10" : "main");
 
 	set_hevc_property(enc, FRAMESIZE, AMFConstructSize(enc->cx, enc->cy));
 	set_hevc_property(enc, USAGE, AMF_VIDEO_ENCODER_USAGE_TRANSCODING);
@@ -2013,6 +2098,17 @@ try {
 	return nullptr;
 }
 
+/* These are initial recommended settings that may be lowered later
+ * once we know more info such as the resolution and frame rate.
+ */
+static void amf_hevc_defaults(obs_data_t *settings)
+{
+	obs_data_set_default_string(settings, "rate_control", "CBR");
+	obs_data_set_default_int(settings, "bitrate", 2500);
+	obs_data_set_default_int(settings, "cqp", 20);
+	obs_data_set_default_string(settings, "preset", "quality");
+}
+
 static void register_hevc()
 {
 	struct obs_encoder_info amf_encoder_info = {};
@@ -2024,7 +2120,7 @@ static void register_hevc()
 	amf_encoder_info.destroy = amf_destroy;
 	amf_encoder_info.update = amf_hevc_update;
 	amf_encoder_info.encode_texture = amf_encode_tex;
-	amf_encoder_info.get_defaults = amf_defaults;
+	amf_encoder_info.get_defaults = amf_hevc_defaults;
 	amf_encoder_info.get_properties = amf_hevc_properties;
 	amf_encoder_info.get_extra_data = amf_extra_data;
 	amf_encoder_info.caps = OBS_ENCODER_CAP_PASS_TEXTURE | OBS_ENCODER_CAP_DYN_BITRATE | OBS_ENCODER_CAP_ROI;
@@ -2108,6 +2204,12 @@ static void amf_av1_update_data(amf_base *enc, int rc, int64_t bitrate, int64_t 
 		} else if (rc == AMF_VIDEO_ENCODER_RATE_CONTROL_METHOD_PEAK_CONSTRAINED_VBR ||
 			   rc == AMF_VIDEO_ENCODER_AV1_RATE_CONTROL_METHOD_HIGH_QUALITY_VBR) {
 			set_av1_property(enc, PEAK_BITRATE, bitrate * 1.5);
+			set_av1_property(enc, VBV_BUFFER_SIZE, bitrate * 1.5);
+		} else if (rc ==
+			   AMF_VIDEO_ENCODER_AV1_RATE_CONTROL_METHOD_LATENCY_CONSTRAINED_VBR) {
+			int64_t framerate = enc->fps_num / enc->fps_den;
+			set_av1_property(enc, VBV_BUFFER_SIZE,
+					 (bitrate / framerate) * 1.1);
 		}
 	} else {
 		int64_t qp = cq_value * 4;
@@ -2150,9 +2252,30 @@ try {
 	return false;
 }
 
+static inline void adjust_recommended_av1_defaults(amf_base *enc,
+						   obs_data_t *settings)
+{
+	const bool is10bit = enc->amf_format == AMF_SURFACE_P010;
+	const int64_t framerate = enc->fps_num / enc->fps_den;
+	if ((enc->cx * enc->cy > 1920 * 1088) || is10bit || (framerate > 60)) {
+
+		// Recommended base defaults
+		obs_data_set_default_int(settings, "bitrate", 2500);
+		obs_data_set_default_int(settings, "cqp", 20);
+		obs_data_set_default_string(settings, "preset", "balanced");
+		obs_data_set_default_string(settings, "profile", "main");
+		obs_data_set_default_int(settings, "bf", 0);
+		info("Original base default settings were used according to resolution and framerate.");
+	}
+}
+
 static bool amf_av1_init(void *data, obs_data_t *settings)
 {
 	amf_base *enc = (amf_base *)data;
+
+	// Initial high perceptual quality settings are set in the UI.
+	// Adjust during init based on resolution and framerate.
+	adjust_recommended_av1_defaults(enc, settings);
 
 	int64_t bitrate = obs_data_get_int(settings, "bitrate");
 	int64_t qp = obs_data_get_int(settings, "cqp");
@@ -2171,6 +2294,11 @@ static bool amf_av1_init(void *data, obs_data_t *settings)
 
 	int rc = get_av1_rate_control(rc_str);
 	set_av1_property(enc, RATE_CONTROL_METHOD, rc);
+	if (rc != AMF_VIDEO_ENCODER_AV1_RATE_CONTROL_METHOD_CONSTANT_QP &&
+	    rc != AMF_VIDEO_ENCODER_AV1_RATE_CONTROL_METHOD_HIGH_QUALITY_VBR &&
+	    rc != AMF_VIDEO_ENCODER_AV1_RATE_CONTROL_METHOD_HIGH_QUALITY_CBR) {
+		set_av1_property(enc, AQ_MODE, 1);
+	}
 
 	amf_av1_update_data(enc, rc, bitrate * 1000, qp);
 
@@ -2209,18 +2337,23 @@ static bool amf_av1_init(void *data, obs_data_t *settings)
 	}
 
 	info("settings:\n"
-	     "\trate_control: %s\n"
-	     "\tbitrate:      %d\n"
-	     "\tcqp:          %d\n"
-	     "\tkeyint:       %d\n"
-	     "\tpreset:       %s\n"
-	     "\tprofile:      %s\n"
-	     "\tlevel:        %s\n"
-	     "\tb-frames:     %d\n"
-	     "\twidth:        %d\n"
-	     "\theight:       %d\n"
-	     "\tparams:       %s",
-	     rc_str, bitrate, qp, gop_size, preset, profile, level_str, bf, enc->cx, enc->cy, ffmpeg_opts);
+	     "\trate_control:         %s\n"
+	     "\tbitrate:              %d\n"
+	     "\tcqp:                  %d\n"
+	     "\tkeyint:               %d\n"
+	     "\tpreset:               %s\n"
+	     "\tprofile:              %s\n"
+	     "\tlevel:                %s\n"
+	     "\tb-frames:             %d\n"
+	     "\twidth:                %d\n"
+	     "\theight:               %d\n"
+	     "\tscreen content tools: %s\n"
+	     "\tpalette mode:         %s\n"
+	     "\tpre-analysis:         %s\n"
+	     "\tparams:               %s",
+	     rc_str, bitrate, qp, gop_size, preset, profile, level_str, bf, enc->cx, enc->cy,
+	     screen_content_tools_enabled ? "true" : "false", palette_mode_enabled ? "true" : "false",
+	     pa_enabled ? "true" : "false", ffmpeg_opts);
 
 	return true;
 }
@@ -2244,6 +2377,7 @@ static void amf_av1_create_internal(amf_base *enc, obs_data_t *settings)
 
 	const bool is10bit = enc->amf_format == AMF_SURFACE_P010;
 	const char *preset = obs_data_get_string(settings, "preset");
+	obs_data_set_string(settings, "profile", "main");
 
 	set_av1_property(enc, FRAMESIZE, AMFConstructSize(enc->cx, enc->cy));
 	set_av1_property(enc, USAGE, AMF_VIDEO_ENCODER_USAGE_TRANSCODING);
@@ -2256,7 +2390,8 @@ static void amf_av1_create_internal(amf_base *enc, obs_data_t *settings)
 	set_av1_property(enc, OUTPUT_COLOR_PROFILE, enc->amf_color_profile);
 	set_av1_property(enc, OUTPUT_TRANSFER_CHARACTERISTIC, enc->amf_characteristic);
 	set_av1_property(enc, OUTPUT_COLOR_PRIMARIES, enc->amf_primaries);
-	set_av1_property(enc, FRAMERATE, enc->amf_frame_rate);
+	set_av1_property(enc, SCREEN_CONTENT_TOOLS, true);
+	set_av1_property(enc, PALETTE_MODE, true);
 
 	amf_av1_init(enc, settings);
 
@@ -2346,13 +2481,15 @@ try {
 	return nullptr;
 }
 
+/* These are initial recommended settings that may be lowered later
+ * once we know more info such as the resolution and frame rate.
+ */
 static void amf_av1_defaults(obs_data_t *settings)
 {
 	obs_data_set_default_int(settings, "bitrate", 2500);
 	obs_data_set_default_int(settings, "cqp", 20);
 	obs_data_set_default_string(settings, "rate_control", "CBR");
-	obs_data_set_default_string(settings, "preset", "quality");
-	obs_data_set_default_string(settings, "profile", "high");
+	obs_data_set_default_string(settings, "preset", "highQuality");
 	obs_data_set_default_int(settings, "bf", 2);
 }
 
--- a/plugins/obs-ffmpeg/texture-amf.cpp
+++ b/plugins/obs-ffmpeg/texture-amf.cpp
@@ -543,6 +543,38 @@ static inline int64_t convert_to_obs_ts(amf_base *enc, int64_t ts)
 	return ts * (int64_t)enc->fps_den / amf_timebase;
 }
 
+static inline const char *convert_amf_memory_type_to_string(AMF_MEMORY_TYPE type)
+{
+	switch (type) {
+	case AMF_MEMORY_UNKNOWN:
+		return "Unknown";
+	case AMF_MEMORY_HOST:
+		return "Host";
+	case AMF_MEMORY_DX9:
+		return "DX9";
+	case AMF_MEMORY_DX11:
+		return "DX11";
+	case AMF_MEMORY_OPENCL:
+		return "OpenCL";
+	case AMF_MEMORY_OPENGL:
+		return "OpenGL";
+	case AMF_MEMORY_XV:
+		return "XV";
+	case AMF_MEMORY_GRALLOC:
+		return "Gralloc";
+	case AMF_MEMORY_COMPUTE_FOR_DX9:
+		return "Compute For DX9";
+	case AMF_MEMORY_COMPUTE_FOR_DX11:
+		return "Compute For DX11";
+	case AMF_MEMORY_VULKAN:
+		return "Vulkan";
+	case AMF_MEMORY_DX12:
+		return "DX12";
+	default:
+		return "Invalid";
+	}
+}
+
 static void convert_to_encoder_packet(amf_base *enc, AMFDataPtr &data, encoder_packet *packet)
 {
 	if (!data)
@@ -1160,36 +1192,16 @@ static void check_texture_encode_capability(obs_encoder_t *encoder, amf_codec_ty
 
 #include "texture-amf-opts.hpp"
 
+/* These are initial recommended settings that may be lowered later once we know more info such as the resolution and
+ * frame rate. */
 static void amf_avc_defaults(obs_data_t *settings)
 {
-	// These are initial recommended settings that may be lowered later
-	// once we know more info such as the resolution and frame rate.
 	obs_data_set_default_string(settings, "rate_control", "CBR");
 	obs_data_set_default_int(settings, "bitrate", 2500);
 	obs_data_set_default_int(settings, "cqp", 20);
 	obs_data_set_default_string(settings, "preset", "quality");
 	obs_data_set_default_string(settings, "profile", "high");
 	obs_data_set_default_int(settings, "bf", 2);
-}
-
-static void amf_hevc_defaults(obs_data_t *settings)
-{
-	// These are initial recommended settings that may be lowered later
-	// once we know more info such as the resolution and frame rate.
-	obs_data_set_default_string(settings, "rate_control", "VBR");
-	obs_data_set_default_int(settings, "bitrate", 2500);
-	obs_data_set_default_int(settings, "cqp", 20);
-	obs_data_set_default_string(settings, "preset", "quality");
-}
-
-static void amf_av1_defaults(obs_data_t *settings)
-{
-	// These are initial recommended settings that may be lowered later
-	// once we know more info such as the resolution and frame rate.
-	obs_data_set_default_int(settings, "bitrate", 2500);
-	obs_data_set_default_int(settings, "cqp", 20);
-	obs_data_set_default_string(settings, "rate_control", "VBR");
-	obs_data_set_default_string(settings, "preset", "highQuality");
 }
 
 static bool rate_control_modified(obs_properties_t *ppts, obs_property_t *p, obs_data_t *settings)
@@ -1255,6 +1267,9 @@ static obs_properties_t *amf_properties_internal(amf_codec_type codec)
 			add_profile("baseline");
 #undef add_profile
 	}
+
+	p = obs_properties_add_bool(props, "pre_analysis", obs_module_text("AMF.PreAnalysis"));
+	obs_property_set_long_description(p, obs_module_text("AMF.PreAnalysis.ToolTip"));
 
 	if (amf_codec_type::AVC == codec || amf_codec_type::AV1 == codec) {
 		obs_properties_add_int(props, "bf", obs_module_text("BFrames"), 0, 5, 1);
@@ -1352,11 +1367,9 @@ static void amf_avc_update_data(amf_base *enc, int rc, int64_t bitrate, int64_t 
 			   rc == AMF_VIDEO_ENCODER_RATE_CONTROL_METHOD_HIGH_QUALITY_VBR) {
 			set_avc_property(enc, PEAK_BITRATE, bitrate * 1.5);
 			set_avc_property(enc, VBV_BUFFER_SIZE, bitrate * 1.5);
-		} else if (rc ==
-			   AMF_VIDEO_ENCODER_RATE_CONTROL_METHOD_LATENCY_CONSTRAINED_VBR) {
+		} else if (rc == AMF_VIDEO_ENCODER_RATE_CONTROL_METHOD_LATENCY_CONSTRAINED_VBR) {
 			int64_t framerate = enc->fps_num / enc->fps_den;
-			set_avc_property(enc, VBV_BUFFER_SIZE,
-					 (bitrate / framerate) * 1.1);
+			set_avc_property(enc, VBV_BUFFER_SIZE, (bitrate / framerate) * 1.1);
 		}
 	} else {
 		set_avc_property(enc, QP_I, qp);
@@ -1399,12 +1412,10 @@ try {
 	return false;
 }
 
-static inline void adjust_recommended_avc_defaults(amf_base *enc,
-						   obs_data_t *settings)
+static inline void adjust_recommended_avc_defaults(amf_base *enc, obs_data_t *settings)
 {
 	int64_t framerate = enc->fps_num / enc->fps_den;
 	if ((enc->cx * enc->cy > 1920 * 1088) || (framerate > 60)) {
-
 		// Recommended base defaults
 		obs_data_set_default_int(settings, "bitrate", 2500);
 		obs_data_set_default_int(settings, "cqp", 20);
@@ -1512,6 +1523,11 @@ static bool amf_avc_init(void *data, obs_data_t *settings)
 	const char *profile = obs_data_get_string(settings, "profile");
 	const char *rc_str = obs_data_get_string(settings, "rate_control");
 	int64_t bf = obs_data_get_int(settings, "bf");
+	bool pa_enabled = obs_data_get_bool(settings, "pre_analysis");
+
+	if (pa_enabled) {
+		set_avc_property(enc, PRE_ANALYSIS_ENABLE, pa_enabled);
+	}
 
 	if (enc->bframes_supported) {
 		set_avc_property(enc, MAX_CONSECUTIVE_BPICTURES, bf);
@@ -1521,7 +1537,7 @@ static bool amf_avc_init(void *data, obs_data_t *settings)
 		 * as those with high motion. This only takes effect if
 		 * Pre-Analysis is enabled.
 		 */
-		if (bf > 0) {
+		if (bf > 0 && pa_enabled == true) {
 			set_avc_property(enc, ADAPTIVE_MINIGOP, true);
 		}
 
@@ -1537,8 +1553,9 @@ static bool amf_avc_init(void *data, obs_data_t *settings)
 	set_avc_property(enc, RATE_CONTROL_METHOD, rc);
 	if (rc != AMF_VIDEO_ENCODER_RATE_CONTROL_METHOD_CONSTANT_QP &&
 	    rc != AMF_VIDEO_ENCODER_RATE_CONTROL_METHOD_HIGH_QUALITY_VBR &&
-	    rc != AMF_VIDEO_ENCODER_RATE_CONTROL_METHOD_HIGH_QUALITY_CBR)
+	    rc != AMF_VIDEO_ENCODER_RATE_CONTROL_METHOD_HIGH_QUALITY_CBR) {
 		set_avc_property(enc, ENABLE_VBAQ, true);
+	}
 
 	amf_avc_update_data(enc, rc, bitrate * 1000, qp);
 
@@ -1595,8 +1612,10 @@ static bool amf_avc_init(void *data, obs_data_t *settings)
 	     "\tb-frames:     %d\n"
 	     "\twidth:        %d\n"
 	     "\theight:       %d\n"
+	     "\tpre-analysis: %s\n"
 	     "\tparams:       %s",
-	     rc_str, bitrate, qp, gop_size, preset, profile, level_str, bf, enc->cx, enc->cy, ffmpeg_opts);
+	     rc_str, bitrate, qp, gop_size, preset, profile, level_str, bf, enc->cx, enc->cy,
+	     pa_enabled ? "true" : "false", ffmpeg_opts);
 
 	return true;
 }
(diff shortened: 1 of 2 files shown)


flaeri:
Testing this PR, and unfortunately I ran into a crash that seems to occur in the AMD drivers.
Crash:
https://obsproject.com/logs/A1hmycbXo3FBhlJJ

Log:
https://obsproject.com/logs/JZHZe6rt8vkBVQyV

Occured when stopping the last recording (w hevc).

I've tried reproducing it, but it seems to be somewhat intermittent, so I have not managed to find specific reproduction steps so far. Will update if I find a reliable repro.

flaeri:
I am a little bit worried about these defaults. This is primarily because it enables Pre-Analysis (PA).
PA eats a decent chunk of traditional GPU compute, and this can in turn lead to encoder lag if there is contention for GPU resources.

In my opinion, people with older/weaker hardware, like rx4/500 series, or APUs, this is likely to cause some problems, which are not currently easily remedied.
As it stands, the only way to disable PA is to type params into the text box, and I worry about the amount of people who would manage to actually disable PA on their own without the assistance from OBS support or someone knowledgeable, or consulting a wiki/guide.

I would personally prefer one or a combination of the following:
* Keep PA off by default, change the default rate control for streaming to HQCBR
  * Rate controls requiring PA would automatically enable it (HQCBR, QVBR etc), which it currently does today
* Settings checkbox to enable/disable PA
* Keep PA off unless RC = CBR (other RCs requiring it functions as is, and enable PA (HQCBR, QVBR etc)

This would allow rate controls where PA is less impactful (VBR, CQP etc) to be easily (and/or automatically) be used without the additional GPU load.

I understand the reasoning and desire to have AMF look better by default for streaming, and at the very least be easier/simpler to achieve, and agree with that sentiment.

Just my two cents. Thank you for the PR, it seems to work well for me outside of that one crash, which seems to have been a one off (cannot reproduce it).

nowrep:
`FILLER_DATA_ENABLE=true` has no effect with `ENFORCE_HRD=false` (which is the default) - no filler data will be inserted.

Also `ENFORCE_HRD=false` will allow the bitrate to peak very high even with CBR, which doesn't seem to be a good thing for streaming.

rhutsAMD:
> I am a little bit worried about these defaults. This is primarily because it enables Pre-Analysis (PA). PA eats a decent chunk of traditional GPU compute, and this can in turn lead to encoder lag if there is contention for GPU resources.
> 
> In my opinion, people with older/weaker hardware, like rx4/500 series, or APUs, this is likely to cause some problems, which are not currently easily remedied. As it stands, the only way to disable PA is to type params into the text box, and I worry about the amount of people who would manage to actually disable PA on their own without the assistance from OBS support or someone knowledgeable, or consulting a wiki/guide.
> 
> I would personally prefer one or a combination of the following:
> 
> * Keep PA off by default, change the default rate control for streaming to HQCBR
>   
>   * Rate controls requiring PA would automatically enable it (HQCBR, QVBR etc), which it currently does today
> * Settings checkbox to enable/disable PA
> * Keep PA off unless RC = CBR (other RCs requiring it functions as is, and enable PA (HQCBR, QVBR etc)
> 
> This would allow rate controls where PA is less impactful (VBR, CQP etc) to be easily (and/or automatically) be used without the additional GPU load.
> 
> I understand the reasoning and desire to have AMF look better by default for streaming, and at the very least be easier/simpler to achieve, and agree with that sentiment.
> 
> Just my two cents. Thank you for the PR, it seems to work well for me outside of that one crash, which seems to have been a one off (cannot reproduce it).

Thank you for providing feedback on the PR and testing it.
 
Older/weaker hardware such as RX 4/500 series do not support PA and will not be allowed to enable it. To clarify, PA is only supported on Radeon RX 5000 Series or newer GPUs as well as Ryzen 2000 U/H series or newer APUs. [GPUOpen-LibrariesAndSDKs/AMF/blob/master/amf/doc/AMF_Video_PreAnalysis_API.md#13-supported-hardware](https://github.com/GPUOpen-LibrariesAndSDKs/AMF/blob/master/amf/doc/AMF_Video_PreAnalysis_API.md#13-supported-hardware)
 
The recommended quality settings are to improve the quality at 1080p60 or lower resolutions/throughput.
 
PA is kept off/not set for higher than 1080p60. The same idea applies for the preset where balanced mode is applied for all codecs for higher than 1080p60.
 
Since you mention having the RC method set to HQCBR for streaming, is there a way to identify the recording or streaming use case at the encoder plugin level? My understanding is that the AMF encoder plugin in OBS just receives textures and encodes them like the other encoders in OBS. It implements the encoder interface (`obs_encoder_info`) and is not aware of what OBS does with the textures after it is done encoding. Is there some way to tell if the encoded textures will be used for recording/streaming? If I remember correctly, OBS is currently enforcing CBR only for streaming [rtmp-common.c#L665](https://github.com/obsproject/obs-studio/blob/master/plugins/rtmp-services/rtmp-common.c#L665)
 
We are also interested in having a PA enable/disable checkbox but from past discussions it seems that UI/interface changes need to be done in a vendor agnostic way by OBS developers.

flaeri:
PA support: I'm particularly worried about the APUs. They don't have a lot of GPU to play with, and I believe basically any game would lead to GPU contention, and ultimately encoder lag.

HQCBR: I'm not certain, but I think you are correct. What I was envisioning was the defaults in the UI being different in streaming vs recording, but now that my memory has been jogged, that is not something that can be done in a simple manner.

PA bool/checkbox: Which leaves us back at the issue of the UI/UX, which is tricky. I know they don't want more checkboxes, but I am failing to see any other way to neatly handle it.

---
If it is just kept as is (PA enabled by default, no easy way to disable), then I truly believe that every APU user that tries streaming/recording gameplay will have a very poor experience, and be unlikely to figure it out on their own. There are a lot of APU users, and having the defaults be unusable(for gameplay), where the only solution is very arcane just really vexes me, personally. I think it would lead to substantially more support request as well.

At this point, I am just retreading the same ground, so I don't really have anything more to add. Thank you for your time and effort :)

flaeri:
[log.txt](https://github.com/obsproject/obs-studio/files/12311608/log.txt)

I figured it would be pertinent to add some data on my claims instead of people just having to take my word for it. Here is an example of the APUs I'm talking about.

Its a simple older game (assault android cactus), running at 1366x768, targeting 60fps. All else equal between the two OBS sessions. Running stock, it hovers around 50fps, and is pretty pleasant to play. With the PR its hovering 40fps, but the lows are very noticable, quite jarring/stuttery (big fluctuations), a poor experience.

Regarding OBS, as you can see in the log, current master/stock managed to stream/record without any render or encoder lag. With the PR it is unfortunately unusable `skipped frames due to encoding lag: 3895/4201 (92.7%)`

----
Side thought regarding PA handling. If bool/checkbox is out of the question, then perhaps we could at least somehow exclude APUs somehow? `MAX_THROUGHPUT ` is not supported on this chip, and returns 0. Maybe that would be a simple way to exclude APUs (assuming similar chips behave the same). As far as I know there is currently no way for AMF to identify low power/APU hardware.

hxzael:
I don't think having `AMF_PA_TAQ_MODE` set to `2` by default is a good idea considering how using `AMF_PA_LOOKAHEAD_BUFFER_DEPTH` with it will overload the encoder in OBS at resolutions > 720p. The only way to use it without overloading is to set the preset to `Balanced` or `Speed` but even then there will be constant frame skips present. Not to mention that TAQ Mode 2 will constantly crash OBS when you stop Recording/Streaming regardless if it overloads or not.

In ffmpeg, if `AMF_PA_TAQ_MODE` set to 2 is used in combination with `AMF_PA_LOOKAHEAD_BUFFER_DEPTH` at 1080p60, you'll frequently find the encoding speeds going below 1x, which would explain the overloads in OBS. Here too, it's necessary to dial down the preset back to `Balanced` or `Speed` to achieve real-time encoding speeds again.

OBS Log file with the suggested defaults above:
https://obsproject.com/logs/8kbzLRAOC2KdIOHQ

rhutsAMD:
> I don't think having `AMF_PA_TAQ_MODE` set to `2` by default is a good idea considering how using `AMF_PA_LOOKAHEAD_BUFFER_DEPTH` with it will overload the encoder in OBS at resolutions > 864p. The only way to use it without overloading is to set the preset to `Balanced` or `Speed` but even then there will be constant frame skips present. Not to mention that TAQ Mode 2 will constantly crash OBS when you stop Recording/Streaming regardless if it overloads or not.
> 
> In ffmpeg, if `AMF_PA_TAQ_MODE` set to 2 is used in combination with `AMF_PA_LOOKAHEAD_BUFFER_DEPTH` at 1080p60, you'll frequently find the encoding speed going below 1x, which would explain the overloads in OBS. Here too, it's necessary to dial down the preset back to `Balanced` or `Speed` to achieve real-time encoding speeds again.
> 
> OBS Log file with the suggested defaults above: https://obsproject.com/logs/ID4dOFS2O1OrJxSL

Thanks for helping to test these changes. Please attach the full log since the log you have shared cuts off at the name of the recoding file without any information on the number of dropped frames or any encoder overloading happening. There are also many errors unrelated to AMF encoding in your log from other plugin(s) or something else.

hxzael:
> I don't think having `AMF_PA_TAQ_MODE` set to `2` by default is a good idea considering how using `AMF_PA_LOOKAHEAD_BUFFER_DEPTH` with it will overload the encoder in OBS at resolutions > 864p. The only way to use it without overloading is to set the preset to `Balanced` or `Speed` but even then there will be constant frame skips present. Not to mention that TAQ Mode 2 will constantly crash OBS when you stop Recording/Streaming regardless if it overloads or not.
> 
> In ffmpeg, if `AMF_PA_TAQ_MODE` set to 2 is used in combination with `AMF_PA_LOOKAHEAD_BUFFER_DEPTH` at 1080p60, you'll frequently find the encoding speed going below 1x, which would explain the overloads in OBS. Here too, it's necessary to dial down the preset back to `Balanced` or `Speed` to achieve real-time encoding speeds again.
> 
> OBS Log file with the suggested defaults above: https://obsproject.com/logs/ID4dOFS2O1OrJxSL

I'm aware the log file I provided has a lot of clutter from other existing plugins, here is another log file from a clean Scene Collection where I reproduced the results I explained earlier.

https://obsproject.com/logs/O3mWvWmZsYf96CEL

hxzael:
Thank you for letting me know, I have replied to my earlier comment with a
cleaner log showing the number of dropped frames and the encoder overload,
as well as a crash log.

lextra2:
> HQCBR and HQVBR are almost unusable, even on the 7900XTX. For 1080p60 recording, they are all overloaded.

That is because rhuts forgot to tell you to use "Preset: Balanced" & "PASceneChangeDetectionEnable=false"
Otherwise the encoder overloads.

Here is a quick HEVC "quality" comparison used with the following settings:

Bitrate: 8000 Kbps
Keyframe Interval: 2s

AMF Options:
`HevcEnableVBAQ=true HevcDeBlockingFilter=false HevcFillerDataEnable=false HevcPeakBitrate=12000000 HevcVBVBufferSize=12000000`
AMD_HEVC_HQVBR

AMF Options:
`HevcEnableVBAQ=true HevcDeBlockingFilter=false HevcFillerDataEnable=false HevcPeakBitrate=12000000 HevcVBVBufferSize=12000000 HevcEnablePreAnalysis=true PASceneChangeDetectionEnable=false PALookAheadBufferDepth=20 PATemporalAQMode=2`
AMD_HEVC_HQVBR_PA

And here is x264 _-preset medium -profile high -threads 24_
_x264_medium

I'm seriously wondering who is the target audience for such a mediocre hardware encoder?
And its not like you can use it to encode 1440p or 4k. The encoder is already at its limit with 1080p, yet it can't even beat x264 (which has about 25% CPU usage on my Ryzen 3900x, which is no problem for game performance)

With AMF, Static elements look attrocious and god forbid you have to use a rate control preset that isn't "HQ"...

lextra2:
@rhutsAMD 
My suggestions; Set scenecut of PreAnalysis to false, e.g.
`PASceneChangeDetectionEnable=false`

Reason: x264 & [NVENC](https://github.com/obsproject/obs-studio/commit/8fb32c37ee170036a362a4772414e8060401a6fe) both have their scenecut for cbr/vbr disabled because a fixed keyint is preferred for streaming and having it disabled gives the encoder more headroom.

Also, use `Preset: Balanced` for all "HQ" presets.
Reason: The image quality difference of Balanced & Quality is minimal, but Quality is really pushing the encoder and so to maximize compatibility, Balanced would be the better option.

Johl7:
Yep, scenecut(scenehange) for real time livestreaming only hurt quality, also agree on bal preset, for higher end models qual is prob a no big deal but on my card for example(6650 xt), if i use PA stuff and bframes >2 qual preset can become quite a burden for the encoder, depending on the other PA params.

lextra2:
Hello,
any updates? Whats the hold up?
As previously discussed, using `AMF_VIDEO_ENCODER_QUALITY_PRESET_BALANCED` & `AMF_PA_SCENE_CHANGE_DETECTION_ENABLE=false` should make the new settings work for most people.

`QVBR`, `HQVBR `& `HQCBR `are already using PreAnalysis anyways.

Fenrirthviti:
Please refrain from comments asking why something hasn't been merged yet.

We know this is here, we'll get to it as soon as we have time, in appropriate priority.

Johl7:
To add some things ive noticed regarding some of the settings in this post. For the PA engine i would suggest either to use Vulkan or DX12, from some testings ive done, OpenCL doesnt seem to work with PA propper. DX12 and Vulkan had better performance compared to the current default(DX11), at least in the few testings ive done.

lextra2:
>PAEngineType=11(DX12) and PAEngineType=10(Vulkan).

Where are you getting those values from?
[The docs](https://github.com/GPUOpen-LibrariesAndSDKs/AMF/blob/master/amf/public/include/components/PreAnalysis.h#L95) don't say anything about DX12 support

[Nevermind found it](https://github.com/GPUOpen-LibrariesAndSDKs/AMF/issues/439#issuecomment-1917886326)
@rhutsAMD could you please update the docs of PreAnalysis.h#L95 with the relevant enumerations? Its very useful information which otherwise gets lost.

rhutsAMD:
> > PAEngineType=11(DX12) and PAEngineType=10(Vulkan).
> 
> Where are you getting those values from? [The docs](https://github.com/GPUOpen-LibrariesAndSDKs/AMF/blob/master/amf/public/include/components/PreAnalysis.h#L95) don't say anything about DX12 support
> 
> [Nevermind found it](https://github.com/GPUOpen-LibrariesAndSDKs/AMF/issues/439#issuecomment-1917886326) @rhutsAMD could you please update the docs of PreAnalysis.h#L95 with the relevant enumerations? Its very useful information which otherwise gets lost.

Thank you for noticing this. The header and the PreAnalysis API doc on GPUOpen have now been updated to reflect the additional PAEngineType options.

Regarding the new recommended defaults for AMF in OBS, there is still some internal testing going on and we expect to have an updated set of recommended defaults. I will update this PR with the new set once those are ready.

Johl7:
Regarding the engines, i was mistaken... seems like DX12 might also not be working properly, since quality atm is not as good as the default in OBS(DX11) and Vulkan.

rhutsAMD:
The PR has been updated and the settings have been retested on the recent public driver version 24.2.1.

hxzael:
I have a question regarding the recommended lookahead buffer-depth setting of 20 in these recommendations. 

The Preanalysis API (https://github.com/GPUOpen-LibrariesAndSDKs/AMF/blob/master/amf/doc/AMF_Video_PreAnalysis_API.pdf) specifies that AMF_PA_TAQ_MODE only works when AMF_PA_LOOKAHEAD_BUFFER_DEPTH is set to 11, 21 or 41. Do values other than these 3 work considering you're selecting 20 instead here? Would love some clarification on this, thank you.

rhutsAMD:
> I have a question regarding the recommended lookahead buffer-depth setting of 20 in these recommendations.
> 
> The Preanalysis API (https://github.com/GPUOpen-LibrariesAndSDKs/AMF/blob/master/amf/doc/AMF_Video_PreAnalysis_API.pdf) specifies that AMF_PA_TAQ_MODE only works when AMF_PA_LOOKAHEAD_BUFFER_DEPTH is set to 11, 21 or 41. Do values other than these 3 work considering you're selecting 20 instead here? Would love some clarification on this, thank you.

The three LAB depths ( short lookahead(11), medium lookahead(21) and long lookahead(41) )  are tested recommendations, rather than requirements as the original wording suggests.

I have updated the doc to clarify the wording of the example values as being suggestions.

RytoEX (changes requested):
The fourth commit ("@rhutsAMD
[CI: Pull in latest submodules from master branch](https://github.com/obsproject/obs-studio/pull/9352/commits/37722d9f31fdf2691c4c9267959fb82858d9e31b)) is superfluous. Just rebasing the branch should be sufficient.

plugins/obs-ffmpeg/texture-amf.cpp
```diff
@@ -2141,18 +2370,20 @@ static bool amf_av1_init(void *data, obs_data_t *settings)
...
-	     "\trate_control: %s\n"
-	     "\tbitrate:      %d\n"
-	     "\tcqp:          %d\n"
-	     "\tkeyint:       %d\n"
-	     "\tpreset:       %s\n"
-	     "\tprofile:      %s\n"
-	     "\twidth:        %d\n"
-	     "\theight:       %d\n"
-	     "\tparams:       %s",
+	info("Settings:\n"
```
RytoEX, line 2373: Revert case change.

plugins/obs-ffmpeg/texture-amf.cpp
```diff
@@ -2141,18 +2370,20 @@ static bool amf_av1_init(void *data, obs_data_t *settings)
...
+	     "\tpreset:                %s\n"
+	     "\tprofile:               %s\n"
+	     "\twidth:                 %d\n"
+	     "\theight:                %d\n"
+	     "\tscreen content tools:  %s\n"
+	     "\tpalette mode:          %s\n"
+	     "\toverriding params:  %s",
 	     rc_str, bitrate, qp, gop_size, preset, profile, enc->cx, enc->cy,
-	     ffmpeg_opts);
+	     "true", "true", ffmpeg_opts);
```
RytoEX, line 2386: Don't hard-code logged settings values.

plugins/obs-ffmpeg/texture-amf.cpp
```diff
@@ -1385,10 +1424,69 @@ try {
...
+			set_amf_property(enc, AMF_PA_ENGINE_TYPE,
+					 AMF_MEMORY_OPENCL);
+
+			info("High perceptual quality defaults:\n"
+			     "\tAdaptiveMiniGOP:        %s\n"
+			     "\tEnablePreAnalysis:      %s\n"
+			     "\tPALookAheadBufferDepth: %d\n"
+			     "\tPATemporalAQMode:       %d\n"
+			     "\tPAEngineType:           %s\n",
+			     "true", "true", 20, AMF_PA_TAQ_MODE_2, "OpenCL");
```
RytoEX, line 1469: I'm not sure if I'd prefer a short string for `PATemporalAQMode`, such as "MODE_NONE", "MODE_1", or "MODE_2", instead of the raw enum value. I'll have to think about it, but I wanted to comment so I didn't forget.

RytoEX (changes requested):

plugins/obs-ffmpeg/texture-amf.cpp
```diff
@@ -2281,15 +2514,6 @@ try {
...
 
-static void amf_av1_defaults(obs_data_t *settings)
-{
-	obs_data_set_default_int(settings, "bitrate", 2500);
-	obs_data_set_default_int(settings, "cqp", 20);
-	obs_data_set_default_string(settings, "rate_control", "CBR");
-	obs_data_set_default_string(settings, "preset", "quality");
-	obs_data_set_default_string(settings, "profile", "high");
-}
-
```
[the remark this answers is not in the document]
rhutsAMD: updated

plugins/obs-ffmpeg/texture-amf.cpp
```diff
@@ -1161,16 +1161,38 @@ static void check_texture_encode_capability(obs_encoder_t *encoder,
...
-	obs_data_set_default_string(settings, "rate_control", "CBR");
 	obs_data_set_default_string(settings, "preset", "quality");
 	obs_data_set_default_string(settings, "profile", "high");
 	obs_data_set_default_int(settings, "bf", 3);
 }
 
+static void amf_hevc_defaults(obs_data_t *settings)
+{
+	// These are initial recommended settings that may be lowered later
+	// once we know more info such as the resolution and frame rate.
```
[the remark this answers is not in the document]
rhutsAMD: updated

plugins/obs-ffmpeg/texture-amf.cpp
```diff
@@ -1161,16 +1161,38 @@ static void check_texture_encode_capability(obs_encoder_t *encoder,
...
+static void amf_av1_defaults(obs_data_t *settings)
+{
+	// These are initial recommended settings that may be lowered later
+	// once we know more info such as the resolution and frame rate.
+	obs_data_set_default_int(settings, "bitrate", 2500);
+	obs_data_set_default_int(settings, "cqp", 20);
+	obs_data_set_default_string(settings, "rate_control", "VBR");
+	obs_data_set_default_string(settings, "preset", "highQuality");
+}
+
```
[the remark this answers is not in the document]
rhutsAMD: updated

plugins/obs-ffmpeg/texture-amf.cpp
```diff
@@ -1161,16 +1161,38 @@ static void check_texture_encode_capability(obs_encoder_t *encoder,
...
 	obs_data_set_default_string(settings, "preset", "quality");
 	obs_data_set_default_string(settings, "profile", "high");
 	obs_data_set_default_int(settings, "bf", 3);
 }
 
+static void amf_hevc_defaults(obs_data_t *settings)
+{
+	// These are initial recommended settings that may be lowered later
+	// once we know more info such as the resolution and frame rate.
+	obs_data_set_default_string(settings, "rate_control", "VBR");
```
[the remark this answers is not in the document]
rhutsAMD: VBR offers better quality than CBR for video-on-demand content such as when recording where bits can be allocated according to the need/complexity of each scene. With the current code, VBR would apply for OBS recordings.

CBR is more suited for real time live applications such as streaming where OBS
would override the rate control method to a hard coded CBR when streaming anyways:

https://github.com/obsproject/obs-studio/blob/master/plugins/rtmp-services/rtmp-common.c#L665
```obs_data_set_string(settings, "rate_control", "CBR");```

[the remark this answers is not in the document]
Comment (RytoEX) [reply plugins/obs-ffmpeg/texture-amf.cpp]: While OBS currently overrides the rate control method to CBR for streaming, it may not always in the future. I prefer that all rate control methods are the same default and that users learn to pick the rate control that is most appropriate for their use case.

File: plugins/obs-ffmpeg/texture-amf.cpp [hunk 8]
```diff
@@ -2105,10 +2299,68 @@ try {
...
+			     "\tPAEngineType:           %s\n",
+			     pa_enabled ? "true" : "false", pa_lab_depth,
+			     pa_taq_mode,
+			     pa_engine_type == AMF_MEMORY_HOST     ? "Host"
+			     : pa_engine_type == AMF_MEMORY_DX11   ? "DX11"
+			     : pa_engine_type == AMF_MEMORY_OPENCL ? "OpenCL"
+			     : pa_engine_type == AMF_MEMORY_VULKAN ? "Vulkan"
+			     : pa_engine_type == AMF_MEMORY_UNKNOWN
+				     ? "Unknown (Auto)"
+				     : "Invalid");
```
RytoEX, line 2348: Consider a static inline function with a switch-case statement instead of chained ternaries.

plugins/obs-ffmpeg/texture-amf.cpp
```diff
@@ -2105,10 +2299,68 @@ try {
...
+
 static bool amf_av1_init(void *data, obs_data_t *settings)
 {
 	amf_base *enc = (amf_base *)data;
 
+	// We originally set the recommended defaults for high perceptual quality in
+	// the UI settings. Once the resolution and framerate are available here
+	// during init, we check if the defaults need to be lowered to the base
+	// defaults or the remaining recommended high perceptual quality settings can
+	// be set.
```
RytoEX, line 2361: Use the multi-line comment format.

plugins/obs-ffmpeg/texture-amf.cpp
```diff
@@ -2142,17 +2392,19 @@ static bool amf_av1_init(void *data, obs_data_t *settings)
...
+	     "\tbitrate:               %d\n"
+	     "\tcqp:                   %d\n"
+	     "\tkeyint:                %d\n"
+	     "\tpreset:                %s\n"
+	     "\tprofile:               %s\n"
+	     "\twidth:                 %d\n"
+	     "\theight:                %d\n"
+	     "\tscreen content tools:  %s\n"
+	     "\tpalette mode:          %s\n"
+	     "\toverriding params:  %s",
```
RytoEX, line 2405: Parameters should be aligned.

RytoEX (changes requested):

plugins/obs-ffmpeg/texture-amf.cpp
RytoEX: This is probably fine as-is for now.

plugins/obs-ffmpeg/texture-amf.cpp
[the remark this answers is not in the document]
rhutsAMD: Updated RATE_CONTROL_METHOD default to CBR

plugins/obs-ffmpeg/texture-amf.cpp
```diff
@@ -1384,10 +1437,77 @@ try {
...
 static bool amf_avc_init(void *data, obs_data_t *settings)
 {
 	amf_base *enc = (amf_base *)data;
 
+	/* We originally set the recommended defaults for high perceptual quality in
+	 * the UI settings. Once the resolution and framerate are available here
+	 * during init, we check if the defaults need to be lowered to the base
+	 * defaults or the remaining recommended high perceptual quality settings can
+	 * be set.
+	 */
```
RytoEX, line 1508: I dislike that we repeat this large block comment three times, but I'm not sure how to address that other than making it less verbose.

plugins/obs-ffmpeg/texture-amf.cpp
```diff
@@ -1384,10 +1437,77 @@ try {
 	return false;
 }
 
+static inline void check_recommended_avc_defaults(amf_base *enc,
```
RytoEX, line 1440: These functions should probably be called `set_` or `adjust_` rather than `check_`. The prefix `check_` implies that it's only doing settings value validation and allowing or stopping the init. What these functions actually do is change the settings values if "needed", even if it is only changing defaults, which should only apply if the user has not explicitly changed values.

RytoEX (changes requested):
Other than two nits, do not use merge commits to fast-forward history. Use `git rebase`.

plugins/obs-ffmpeg/texture-amf.cpp
rhutsAMD: Fixed in [409550a](https://github.com/obsproject/obs-studio/pull/9352/commits/409550a6bc3a9b1250ca0a538b90e23cd6757f20)

plugins/obs-ffmpeg/texture-amf.cpp
rhutsAMD: Fixed in [409550a](https://github.com/obsproject/obs-studio/pull/9352/commits/409550a6bc3a9b1250ca0a538b90e23cd6757f20)

rhutsAMD:
> Other than two nits, do not use merge commits to fast-forward history. Use `git rebase`.

Rebased to bring in latest changes.

rhutsAMD:
hi @RytoEX , I addressed your requested changes. Could you please revisit this PR?

Ryu0833:
hello my first time here just i am curious the "AMF_VIDEO_ENCODER_RATE_CONTROL_PREANALYSIS_ENABLE" in the site obs wiki:https://obsproject.com/wiki/AMF-Options , it by default it 'RateControlPreanalysisEnable=1' when the new default settings for amd encoders well be still enable by default ? and when it enbale it get alot stutter in obs and games thx and sorry for my bad english.

lextra2:
@Ryu0833 

PreAnalysis is enabled for
```
_RATE_CONTROL_METHOD_QUALITY_VBR
_RATE_CONTROL_METHOD_HIGH_QUALITY_VBR
_RATE_CONTROL_METHOD_HIGH_QUALITY_CBR
```
It has been this way forever.
If your encoder overloads with any of these presets, switch to `CBR`, `VBR` or `CQP`.
You may also try running OBS as Admin, or the following command AMF/FFMPEG Options
`PASceneChangeDetectionEnable=false`

Ryu0833:
thx for response @lextra2

 i am not talking about Preanalysis "AMF_VIDEO_ENCODER_PRE_ANALYSIS_ENABLE" ,ik it deafault disabled but i am talking about the "AMF_VIDEO_ENCODER_RATE_CONTROL_PREANALYSIS_ENABLE" the 'RateControlPreanalysisEnable' it enable without the Preanalysis and it hurts the performance. in the obs wiki it on by default just asking should leave it on by default or change it to off in the new default amd encoder

rhutsAMD:
I have pushed some changes to the new default settings PR which help improve performance. These changes include removing the new cases where PA would be enabled by default, as well as removing the addition of enabling the PA-dependent features by default. Additionally, the B-picture pattern for AVC was updated from 3 to 2.

Ryu0833:
Hello @rhutsAMD i did test the build and i have some feedback about it :

1) "ENABLE_VBAQ" it always enable even when use rc "CBR" and it hurt the image quality.in the line 1498 it well always set to true.

2)"PreAnalysis" why using "AMF_PA_TAQ_MODE" the need "AMF_PA_LOOKAHEAD_BUFFER_DEPTH" that use more compute power  why not using the "AMF_PA_PAQ_MODE" with "AMF_PA_CAQ_STRENGTH_LOW" or "AMF_PA_CAQ_STRENGTH_MEDIUM".

3) "PREENCODE_ENABLE" always true and it hurt performance with no change in quality maybe make true if the bitrate is high or using VBR.

4)"HIGH_MOTION_QUALITY_BOOST_ENABLE" it good without Preanaalysis and good combo if B-frame enable.

5)"B-frames" it only use in rx 6000 and 7000 why dont the rx 5000 at the minmum have only "1b-frame" it help maintain the video quality.

6)What are the changes made in driver version 24.x.x compared to 23.12.1 are not documented in the AMF repository. The 23.12.1 driver provided good quality, smooth video with no lag or rendering issues in OBS .it can any chane that roll back the encoder only not the driver 23.12.1.

and  my AMF option that i use to streaming twitch:
"RateControlPreanalysisEnable=0 EnableVBAQ=false" 
if using with the PreAnalysis:"RateControlPreanalysisEnable=0 EnablePreAnalysis=true PASceneChangeDetectionEnable=false PAPerceptualAQMode=1 PACAQStrength=0"

i hope my feedback it well help improve the amd encoder quality and performance and sorry for my bad english.have good day.

RytoEX (changes requested):
I've made another pass and caught some things that I'd previously missed.

plugins/obs-ffmpeg/texture-amf.cpp
```diff
@@ -1417,12 +1492,13 @@ static bool amf_avc_init(void *data, obs_data_t *settings)
 	int rc = get_avc_rate_control(rc_str);
 
 	set_avc_property(enc, RATE_CONTROL_METHOD, rc);
-	if (rc != AMF_VIDEO_ENCODER_RATE_CONTROL_METHOD_CONSTANT_QP)
+	if (rc != AMF_VIDEO_ENCODER_RATE_CONTROL_METHOD_CONSTANT_QP &&
+	    rc != AMF_VIDEO_ENCODER_RATE_CONTROL_METHOD_HIGH_QUALITY_VBR &&
+	    rc != AMF_VIDEO_ENCODER_RATE_CONTROL_METHOD_HIGH_QUALITY_CBR)
```
Ryu0833, line 1497: for working corectly
```suggestion
	    rc != AMF_VIDEO_ENCODER_RATE_CONTROL_METHOD_HIGH_QUALITY_CBR) {
```

plugins/obs-ffmpeg/texture-amf.cpp
```diff
@@ -1417,12 +1492,13 @@ static bool amf_avc_init(void *data, obs_data_t *settings)
...
 		set_avc_property(enc, ENABLE_VBAQ, true);
```
Ryu0833, line 1499: for working correctly
```suggestion
	if (rc != AMF_VIDEO_ENCODER_RATE_CONTROL_METHOD_CONSTANT_QP &&
	    rc != AMF_VIDEO_ENCODER_RATE_CONTROL_METHOD_HIGH_QUALITY_VBR &&
	    rc != AMF_VIDEO_ENCODER_RATE_CONTROL_METHOD_HIGH_QUALITY_CBR) {
		set_avc_property(enc, ENABLE_VBAQ, true);
	}
```

plugins/obs-ffmpeg/texture-amf.cpp
```diff
@@ -1740,12 +1859,13 @@ static bool amf_hevc_init(void *data, obs_data_t *settings)
 	int rc = get_hevc_rate_control(rc_str);
 
 	set_hevc_property(enc, RATE_CONTROL_METHOD, rc);
-	if (rc != AMF_VIDEO_ENCODER_HEVC_RATE_CONTROL_METHOD_CONSTANT_QP)
+	if (rc != AMF_VIDEO_ENCODER_HEVC_RATE_CONTROL_METHOD_CONSTANT_QP &&
+	    rc != AMF_VIDEO_ENCODER_HEVC_RATE_CONTROL_METHOD_HIGH_QUALITY_VBR &&
+	    rc != AMF_VIDEO_ENCODER_HEVC_RATE_CONTROL_METHOD_HIGH_QUALITY_CBR)
 		set_hevc_property(enc, ENABLE_VBAQ, true);
```
Ryu0833, line 1866: same for HEVC
```suggestion
	if (rc != AMF_VIDEO_ENCODER_HEVC_RATE_CONTROL_METHOD_CONSTANT_QP &&
	    rc != AMF_VIDEO_ENCODER_HEVC_RATE_CONTROL_METHOD_HIGH_QUALITY_VBR &&
	    rc != AMF_VIDEO_ENCODER_HEVC_RATE_CONTROL_METHOD_HIGH_QUALITY_CBR) {
		set_hevc_property(enc, ENABLE_VBAQ, true);
	}
```

Ryu0833 (commented):
Fix property "ENABLE_VBAQ" always true even when using "rc=CBR" for AVC and HEVC

plugins/obs-ffmpeg/texture-amf.cpp
rhutsAMD: fixed

Ryu0833:
hello @rhutsAMD i did test the new settings the h264 work normal but the h265 i get crachs whe ni start recording
[Crash 2024-11-07 21-43-08.txt](https://github.com/user-attachments/files/17669264/Crash.2024-11-07.21-43-08.txt)

rhutsAMD:
> hello @rhutsAMD i did test the new settings the h264 work normal but the h265 i get crachs when i start recording [Crash 2024-11-07 21-43-08.txt](https://github.com/user-attachments/files/17669264/Crash.2024-11-07.21-43-08.txt)

Hi @Ryu0833 , which GPU and AMD driver version are you using? I saw you mentioned RX 6000, 7000 and 5000 earlier. Have you changed any of the new default settings or are using all defaults? Please share a copy of the OBS log with AMF encoder settings.

Ryu0833:
@rhutsAMD i have rx5500 xt 4gb driver 24.10.1 cpu: ryzen 5 3500x  i didnt chane setting 
https://obsproject.com/logs/QBtlSp5QwXeHw2Gp==> using only the default setting h265 with no preanalysis
image
https://obsproject.com/logs/zFmjdTNDNEIwZnVA==> using only the default setting h264 with preanalysis
image
this is the cases(h265 and h264 with preanalysis)  i get crach 
image

rhutsAMD:
@Ryu0833 are you only seeing the crash when streaming and not recording? Your settings screenshots are from the streaming tab.

Also please be aware that B-frames are currently supported for the AMD encoders in H264, so the keyframe interval of 2s for HEVC is an invalid setting.

Ryu0833:
@rhutsAMD the recording streaming using the streaming settings 
"Also please be aware that B-frames are currently supported for the AMD encoders in H264, so the keyframe interval of 2s for HEVC is an invalid setting." i dindt understand the message like how b fram in h265 ? and did rx 5000 can use bfram now ?

lextra2:
> Also please be aware that B-frames are currently supported for the AMD encoders in H264, so the keyframe interval of 2s for HEVC is an invalid setting.

I'm not sure what the keyframe interval has to do with HEVC not supporting bframes?

rhutsAMD:
Thanks for catching that, sorry for the confusion, to clarify:

The log mentions B-frames are set to 2 but not supported on this device. B-frame support for H264 starts with VCN 3.0. @Ryu0833 's RX 5500 XT (VCN 2.0) does not support this feature.

[GPUOpen-LibrariesAndSDKs/AMF/wiki/GPU and APU HW Features and Support](https://github.com/GPUOpen-LibrariesAndSDKs/AMF/wiki/GPU%20and%20APU%20HW%20Features%20and%20Support#amd-gpu-multimedia-hardware-features)

```
22:06:04.499: [texture-amf-h264: 'advanced_video_stream'] B-Frames set to 2 but b-frames are not supported by this device
```

Ryu0833:
@rhutsAMD the crash is in the new setting 
obs 31.0.0 beta 3 no crash

rhutsAMD:
> @rhutsAMD the crash is in the new setting obs 31.0.0 beta 3 no crash

Crash fixed

plugins/obs-ffmpeg/texture-amf.cpp
```diff
@@ -1452,16 +1528,16 @@ static bool amf_avc_init(void *data, obs_data_t *settings)
...
+	     "\trate_control:           %s\n"
+	     "\tbitrate:                %d\n"
+	     "\tcqp:                    %d\n"
+	     "\tkeyint:                 %d\n"
+	     "\tpreset:                 %s\n"
+	     "\tprofile:                %s\n"
+	     "\tb-frames:               %d\n"
+	     "\twidth:                  %d\n"
+	     "\theight:                 %d\n"
+	     "\toverriding params:      %s",
```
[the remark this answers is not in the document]
rhutsAMD: reverted to `params` instead of `overriding params`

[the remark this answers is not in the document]
rhutsAMD: logged and normalized

plugins/obs-ffmpeg/texture-amf.cpp
```diff
@@ -1769,15 +1889,15 @@ static bool amf_hevc_init(void *data, obs_data_t *settings)
...
-	     "\tparams:       %s",
+	     "\trate_control:       %s\n"
+	     "\tbitrate:            %d\n"
+	     "\tcqp:                %d\n"
+	     "\tkeyint:             %d\n"
+	     "\tpreset:             %s\n"
+	     "\tprofile:            %s\n"
+	     "\twidth:              %d\n"
+	     "\theight:             %d\n"
+	     "\toverriding params:  %s",
```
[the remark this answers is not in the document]
rhutsAMD: done

plugins/obs-ffmpeg/texture-amf.cpp
```diff
@@ -2141,17 +2297,19 @@ static bool amf_av1_init(void *data, obs_data_t *settings)
...
+	     "\tbitrate:               %d\n"
+	     "\tcqp:                   %d\n"
+	     "\tkeyint:                %d\n"
+	     "\tpreset:                %s\n"
+	     "\tprofile:               %s\n"
+	     "\twidth:                 %d\n"
+	     "\theight:                %d\n"
+	     "\tscreen content tools:  %s\n"
+	     "\tpalette mode:          %s\n"
+	     "\toverriding params:     %s",
```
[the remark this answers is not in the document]
rhutsAMD: done

plugins/obs-ffmpeg/texture-amf.cpp
```diff
@@ -2141,17 +2297,19 @@ static bool amf_av1_init(void *data, obs_data_t *settings)
...
 	     rc_str, bitrate, qp, gop_size, preset, profile, enc->cx, enc->cy,
-	     ffmpeg_opts);
+	     "true", "true", ffmpeg_opts);
```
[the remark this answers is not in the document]
rhutsAMD: fixed

gregorytrentmartinjr:
@rhutsAMD Still crashes with AV1. Tried to record with default settings. I have an RX 7900XT driver 24.10.1 cpu: i7-13700K. https://obsproject.com/logs/IXBKuPhFdhfZMo3U 
[Crash.txt](https://github.com/user-attachments/files/17713152/Crash.txt)

rhutsAMD:
> @rhutsAMD Still crashes with AV1. Tried to record with default settings. I have an RX 7900XT driver 24.10.1 cpu: i7-13700K. https://obsproject.com/logs/IXBKuPhFdhfZMo3U [Crash.txt](https://github.com/user-attachments/files/17713152/Crash.txt)

AV1 crash fixed

Ryu0833:
@rhutsAMD the last changes working great no craches and bydefault gives the best quality even when not using "PreAnalysis".
i notice some thing in the code there are a lot of missing braces like in line 1484 -1500. just this example if the braces are missing what are the effect well get in obs or the encoder ? and if are breces what can changes ? and sorry for my bad english

rhutsAMD:
> @rhutsAMD the last changes working great no craches and bydefault gives the best quality even when not using "PreAnalysis". i notice some thing in the code there are a lot of missing braces like in line 1484 -1500. just this example if the braces are missing what are the effect will get in obs or the encoder ? and if are breces what can changes ? and sorry for my bad english

Thanks for helping with testing the changes @Ryu0833 😃 

Lines 1484 to 1500 are not modified in this PR.

Regardless, a single statement body without curly brackets is valid C++. It is up to codestyle preference whether to use or not use curly brackets in such case - it functions the same in both ways.

Ryu0833:
@rhutsAMD thank you for your maintenance and for answering my questions. I will report any changes or bugs.

lextra2:
@Ryu0833 
See here for which presets have it enabled by default
https://github.com/GPUOpen-LibrariesAndSDKs/AMF/blob/master/amf/doc/AMF_Video_Encode_HEVC_API.md#table-a-1-encoder-parameters
Screenshot_2024-11-19-15-48-23-826_com kiwibrowser browser-edit

Johl7:
Upon testing this PRs build, the PA checkbox doesnt seem to be working properly, enabling it didnt really seem to be turning on TAQ, semmed more like VBAQ.

lextra2:
@rhutsAMD 
Could you add a box to the UI for all modes which use peak bitrate?
And if you do, you should probably also change vbv buffer size to peak bitrate, instead of bitrate (which is used right now)

https://github.com/obsproject/obs-studio/blob/c8e215be3da8d0a5cfd83bb35db4be045a7669b5/plugins/obs-ffmpeg/texture-amf.cpp#L1247

https://github.com/obsproject/obs-studio/blob/c8e215be3da8d0a5cfd83bb35db4be045a7669b5/plugins/obs-ffmpeg/texture-amf.cpp#L1572

https://github.com/obsproject/obs-studio/blob/c8e215be3da8d0a5cfd83bb35db4be045a7669b5/plugins/obs-ffmpeg/texture-amf.cpp#L1914

Johl7:
> @rhutsAMD Could you add a peak bitrate field to the UI for all vbr modes? And if you do, you should probably also change vbv buffer size to peak bitrate, instead of bitrate (which is used right now)
> 
> https://github.com/obsproject/obs-studio/blob/c8e215be3da8d0a5cfd83bb35db4be045a7669b5/plugins/obs-ffmpeg/texture-amf.cpp#L1247
> 
> https://github.com/obsproject/obs-studio/blob/c8e215be3da8d0a5cfd83bb35db4be045a7669b5/plugins/obs-ffmpeg/texture-amf.cpp#L1572
> 
> https://github.com/obsproject/obs-studio/blob/c8e215be3da8d0a5cfd83bb35db4be045a7669b5/plugins/obs-ffmpeg/texture-amf.cpp#L1914
> 
> That way you still get the same behaviour for cbr but the correct behaviour for vbr based modes.

Just to add somethin ive noticed regarding TargetBitrate and PeakBitrate(on RDNA2, at least). When i set Target, it behaves like Peak, and Peak, like Target.

plugins/obs-ffmpeg/texture-amf.cpp
```diff
@@ -1323,12 +1391,14 @@ static bool amf_avc_init(void *data, obs_data_t *settings)
 	int rc = get_avc_rate_control(rc_str);
 
 	set_avc_property(enc, RATE_CONTROL_METHOD, rc);
-	if (rc != AMF_VIDEO_ENCODER_RATE_CONTROL_METHOD_CONSTANT_QP)
+	if (rc != AMF_VIDEO_ENCODER_RATE_CONTROL_METHOD_CONSTANT_QP &&
+	    rc != AMF_VIDEO_ENCODER_RATE_CONTROL_METHOD_HIGH_QUALITY_VBR &&
+	    rc != AMF_VIDEO_ENCODER_RATE_CONTROL_METHOD_HIGH_QUALITY_CBR) {
 		set_avc_property(enc, ENABLE_VBAQ, true);
+	}
```
Ryu0833, line 1398: change the  "&&" to "||" to the VBAQ work only in rc"CQP or HQCBR or CBR  #"

set_avc_property(enc, RATE_CONTROL_METHOD, rc);
	if (rc == AMF_VIDEO_ENCODER_RATE_CONTROL_METHOD_CONSTANT_QP ||
	    rc == AMF_VIDEO_ENCODER_RATE_CONTROL_METHOD_HIGH_QUALITY_VBR ||
	    rc == AMF_VIDEO_ENCODER_RATE_CONTROL_METHOD_HIGH_QUALITY_CBR) {
		set_avc_property(enc, ENABLE_VBAQ, true);
	}

Ryu0833:
@Johl7 and @lextra2 for the all the vbr is bitrate * 1.5

} else if (rc == AMF_VIDEO_ENCODER_RATE_CONTROL_METHOD_PEAK_CONSTRAINED_VBR ||
			   rc == AMF_VIDEO_ENCODER_RATE_CONTROL_METHOD_HIGH_QUALITY_VBR) {
			set_avc_property(enc, PEAK_BITRATE, bitrate * 1.5);
			set_avc_property(enc, VBV_BUFFER_SIZE, bitrate * 1.5);
		} else if (rc == AMF_VIDEO_ENCODER_RATE_CONTROL_METHOD_LATENCY_CONSTRAINED_VBR) {
			int64_t framerate = enc->fps_num / enc->fps_den;
			set_avc_property(enc, VBV_BUFFER_SIZE, (bitrate / framerate) * 1.1);
		}
so if you set bitrate t 1000kbps the peak bitrate=vbv buffer= 1500kbps

rhutsAMD:
> Upon testing this PRs build, the PA checkbox doesnt seem to be working properly, enabling it didnt really seem to be turning on TAQ, semmed more like VBAQ.

hi @Johl7 the PA checkbox is now fixed. Too many changes were accidentally removed in a previous revision when removing the enablement of PA and PA-dependent features by default.

Regarding TAQ, please note that the default value is generally `AMF_PA_TAQ_MODE_NONE` unless the high quality usage preset is applied.

https://github.com/GPUOpen-LibrariesAndSDKs/AMF/blob/master/amf/doc/AMF_Video_PreAnalysis_API.md#22-properties
**Default Value associated with usages:**
   - Transcoding: `AMF_PA_TAQ_MODE_NONE`
   - Ultra low latency: `AMF_PA_TAQ_MODE_NONE`
   - Low latency: `AMF_PA_TAQ_MODE_NONE`
   - Webcam: `AMF_PA_TAQ_MODE_NONE`
   - HQ: `AMF_PA_TAQ_MODE_2`
   - HQLL: `AMF_PA_TAQ_MODE_NONE`

rhutsAMD:
Rebased to pull in latest changes from base branch

rhutsAMD:
Removed setting ENFORCE_HRD to TRUE in order to maximize video quality. Also removed setting FILLER_DATA_ENABLE to TRUE to avoid wasting bits. The two settings are not performance sensitive features. Both properties can be manually enabled inside the “AMF/FFmpeg Options” field, for example for H264:

`FillerDataEnable=true EnforceHRD=true`

flaeri:
> Removed setting ENFORCE_HRD to TRUE in order to maximize video quality. Also removed setting FILLER_DATA_ENABLE to TRUE to avoid wasting bits. The two settings are not performance sensitive features. Both properties can be manually enabled inside the “AMF/FFmpeg Options” field, for example for H264:
> 
> `FillerDataEnable=true EnforceHRD=true`

Yes, that would improve video quality, but it would also mean that the bitrate would be able to spike a lot higher than when using strict CBR (with filler data). VBV/HRD would be near empty if you for instance pause/freeze on a still frame, and you would be able to spend double the normal bitrate for a short while.

I personally believe that this is unfortunate. OBS currently uses fillers with other encoders, like x264 and nvenc. In my opinion, this is to provide stability for platforms and networks that serve the stream as is (like Twitch). Having to deliver twice the amount of intended bits for a while can cause viewers to buffer, either because the network couldn't handle that, or because the viewers connection couldn't. 

Stream stability (not buffering) for a percentage of viewers, is in my personal opinion more important than slightly increasing video accuracy (quality).

When the initial implementation was made, I had the same complaint (although it had more downstream problems, which are probably solved): https://github.com/obsproject/obs-studio/pull/6508#issuecomment-1148919711

Just my two cents. I appreciate your continued work, and keeping up the spirits of getting new defaults merged. Happy holidays :)

Johl7:
MinQP=18 MaxQP=51 HighMotionQualityBoostEnable=true, and VBVBufferSize to 2x de bitrate, would be the changes i would suggest, as they wouldnt cause any issues on older hardware, or cause some buffering due to not bein a CBR without FillerData.

lextra2:
> HighMotionQualityBoostEnable=true

Its not for realtime encodes. How many times do you have to be told?

Johl7:
> > HighMotionQualityBoostEnable=true
> 
> Its not for realtime encodes. How many times do you have to be told?

Wrong, ur mistaking it for PAHighMotionQualityBoostMode

RytoEX:
> Removed setting ENFORCE_HRD to TRUE in order to maximize video quality. Also removed setting FILLER_DATA_ENABLE to TRUE to avoid wasting bits.



> Yes, that would improve video quality, but it would also mean that the bitrate would be able to spike a lot higher than when using strict CBR (with filler data). VBV/HRD would be near empty if you for instance pause/freeze on a still frame, and you would be able to spend double the normal bitrate for a short while.
> 
> I personally believe that this is unfortunate. OBS currently uses fillers with other encoders, like x264 and nvenc. In my opinion, this is to provide stability for platforms and networks that serve the stream as is (like Twitch).

I agree here. The point of CBR is to have constant bit rate. The defaults for CBR should aim to provide the most consistent bit rate possible and not require a user to provide custom configuration options to do so. If a user doesn't want constant bit rate, they can select another rate control method that does not aim for constant bit rate, such as VBR, instead.

rhutsAMD:
Rebased:
- Resolved merged conflicts from recent changes including throughput check fix which was merged today.
- Adjusted some of the logging to the new 120 char column limit.
- Brought in latest changes from target branch.

rhutsAMD:
Force pushed to resolve clang-format check

Ryu0833:
hello 
@rhutsAMD I tested it, and it works with no delay or issues, plus there’s an improvement in quality—especially when setting QualityPreset=highquality in both AVC and HEVC.

However, in HEVC, the PreAnalysis option is disabled by default, which is fine. But when using QualityPreset=highquality (QualityPreset=15 for those who want to try it), it creates a delay in-game.

rhutsAMD:
> hello @rhutsAMD I tested it, and it works with no delay or issues, plus there’s an improvement in quality—especially when setting QualityPreset=highquality in both AVC and HEVC.
> 
> However, in HEVC, the Preencode option is disabled by default, which is fine. But when using QualityPreset=highquality (QualityPreset=15 for those who want to try it), it creates a delay in-game.
> 
> and guys go check: [https://github.com/GPUOpen-LibrariesAndSDKs/AMF/wiki/Recommended%20FFmpeg%20Encoder%20Settings#recommended-ffmpeg-encoder-settings](url)
> 
> thank you for the good documentation and explication.
> 
> have good day

Thanks @ryu0833 for trying out the changes. The encoder automatically turns on PreAnalysis when the quality preset is high quality even if it is disabled by default, therefore a delay is expected in gaming. The high quality preset was recently added in the driver last month.

plugins/obs-ffmpeg/texture-amf.cpp
```diff
@@ -1089,14 +1121,17 @@ static void check_texture_encode_capability(obs_encoder_t *encoder, amf_codec_ty
 
 #include "texture-amf-opts.hpp"
 
-static void amf_defaults(obs_data_t *settings)
+/* These are initial recommended settings that may be lowered later
+ * once we know more info such as the resolution and frame rate.
+ */
```
RytoEX, line 1126: Rewrap lines at 120 characters/columns.

plugins/obs-ffmpeg/texture-amf.cpp
```diff
@@ -2058,7 +2232,8 @@ static void amf_av1_create_internal(amf_base *enc, obs_data_t *settings)
 	set_av1_property(enc, OUTPUT_COLOR_PROFILE, enc->amf_color_profile);
 	set_av1_property(enc, OUTPUT_TRANSFER_CHARACTERISTIC, enc->amf_characteristic);
 	set_av1_property(enc, OUTPUT_COLOR_PRIMARIES, enc->amf_primaries);
-	set_av1_property(enc, FRAMERATE, enc->amf_frame_rate);
```
RytoEX, line 2061: This removal is only made in the `amf_av1_create_internal` and none of the other codec variants. Was this intentional?

RytoEX (changes requested):
Nit: wrap lines at 120 columns/characters.

There is one change that seems unintentional. The rest looks fine.

plugins/obs-ffmpeg/texture-amf.cpp
rhutsAMD: thanks, updated

plugins/obs-ffmpeg/texture-amf.cpp
rhutsAMD: thanks for catching this, it must have been accidentally removed while rebasing/squashing

I've readded it back where it was

rhutsAMD:
Thanks Ryan for reviewing.

I have updated the comments to rewrap them to the 120 characters/columns limit and readded the line with setting of the AV1 `FRAMERATE` property.

@RytoEX Could you please take another look?

RytoEX:
This seems to have a merge conflict, which is probably my fault.

Johl7:
> > hello @rhutsAMD I tested it, and it works with no delay or issues, plus there’s an improvement in quality—especially when setting QualityPreset=highquality in both AVC and HEVC.
> > However, in HEVC, the Preencode option is disabled by default, which is fine. But when using QualityPreset=highquality (QualityPreset=15 for those who want to try it), it creates a delay in-game.
> > and guys go check: [https://github.com/GPUOpen-LibrariesAndSDKs/AMF/wiki/Recommended%20FFmpeg%20Encoder%20Settings#recommended-ffmpeg-encoder-settings](url)
> > thank you for the good documentation and explication.
> > have good day
> 
> Thanks @Ryu0833 for trying out the changes. The encoder automatically turns on PreAnalysis when the quality preset is high quality even if it is disabled by default, therefore a delay is expected in gaming. The high quality preset was recently added in the driver last month.

Hello, the high quality preset differs from the quality one just on that(PA)? or there are also some other changes?

Thanks in advance.

rhutsAMD:
> * VBAQ to TRUE for AV1 for appropriate rate control methods

No worries, I rebased the PR and fixed merge the conflict.

RytoEX (approved):
Seems okay.

RytoEX:
After another look, it looks like these commits were not cleanly structured. The second commit appears to be doing a lot of cleanup work, such as re-adding FRAMERATE and various other fixups, instead of simply adding the pre-analysis checkbox. I might revert these so they can be resubmitted with the changes in appropriate commits.

rhutsAMD:
> After another look, it looks like these commits were not cleanly structured. The second commit appears to be doing a lot of cleanup work, such as re-adding FRAMERATE and various other fixups, instead of simply adding the pre-analysis checkbox. I might revert these so they can be resubmitted with the changes in appropriate commits.

Restructured commits in new PR: https://github.com/obsproject/obs-studio/pull/12161

PR #9356, opened 2023-08-03
cmake: Mark source lists in feature files sortable

gxalpha:
### Description

Adds the `#cmake-format: sortable` flag to some longer lists ("The list is formatted to be separated into multiple lines" is my threshold for "long" here).

### Motivation and Context

Saw a conversation on a recent PR where it was suggested to sort the list in `feature-youtube.cmake`, adding this attribute can eliminate manual sorting in the future.
Went through other `feature-` files and looked for lists that are long enough to be worth sorting.

### How Has This Been Tested?

macOS 14, ran cmake-format. Lists are now sorted.

### Types of changes

### Checklist:

### DIFF
--- a/UI/cmake/feature-macos-update.cmake
+++ b/UI/cmake/feature-macos-update.cmake
@@ -8,14 +8,15 @@ endif()
 
 target_sources(
   obs-studio
-  PRIVATE update/crypto-helpers.hpp
+  PRIVATE # cmake-format: sortable
           update/crypto-helpers-mac.mm
+          update/crypto-helpers.hpp
+          update/models/branches.hpp
+          update/models/whatsnew.hpp
           update/shared-update.cpp
           update/shared-update.hpp
           update/update-helpers.cpp
-          update/update-helpers.hpp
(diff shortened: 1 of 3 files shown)


PatTheMav:
This is fine in theory, though I dunno if @RytoEX is happy about how it has no concept of subdirectories when sorting those entries (the sorting is entirely alphabetical, treating filepaths with subdirectories as a normal string).

IIRC I did add it manually to have them sorted once, then fixed them up to have entries from subdirectories at the bottom.

RytoEX:
My personal preference is that sub-directories are grouped at the top when they exist. However, I also recognize that cmake-format does not handle this, and that such grouping would require some manual effort. I'm not sure if my particular preference creating more work is agreeable to others. Happy to discuss.

gxalpha:
Hmm, a possible solution could be to have a separate `target_sources` call for each subdirectory? Since the files are atomic now that shouldn't hurt readability too much.

PatTheMav (approved):

Issue #9023, opened 2023-06-04
[macOS] Push-to-talk hotkey for audio sources does not seem to function correctly 


jcm93:
### Operating System Info

macOS 13

### OBS Studio Version

29.1.2

### OBS Studio Log URL

https://obsproject.com/logs/fUBR7DZXd4NgTuI8

### Expected Behavior

Push to talk should operate correctly, i.e. muting the audio source unless the selected button is currently pressed down.

### Current Behavior

I've seen different behavior. Currently, on my system, the push-to-talk hotkey seems to have no effect whatsoever. It neither mutes nor unmutes the audio source.

In the past, I've seen it momentarily engage the audio source for a fraction of a second and then immediately mute it even if the button is still pressed, but that behavior doesn't seem to be present now.

### Steps to Reproduce

1. Add a source with audio (for my testing, I used an Audio Input Source associated with my built-in microphone).
2. In Settings->Hotkeys, set a push-to-talk key for the audio source.
3. With or without OBS having focus, press the hotkey; observe no response in the audio mixer or in the output.

### Anything else we should know?

I've made sure OBS has Accessibility permissions as well as Input Monitoring permissions. I've tried removing those permissions and re-granting them, but it does not seem to have any effect.

The mute and unmute hotkeys seem to work correctly. Push-to-mute, however, also seems similarly broken.

Also: I believe that push-to-talk has not functioned correctly for some time; I *believe* at least as far back as 27.2.4 (though the issue presented slightly differently).

PatTheMav:
Preliminary investigation shows that the functionality works as intended _as long as the OBS window is not focused_. In that situation the key events are handled primarily by macOS itself, but when the window is in focus we receive the key events via Qt's internal key event system.

And in the latter case it seems like when pressing (and holding) the key combination we get a single event for the pressed keys, but after that the event is consumed and our hotkey system assumes that the keys are not pressed anymore, hence why "push to talk" only works for quick moment.

jcm93:
I am still having issues here, seeing behavior that is different from what Pat described. Even with Pat's PR, "push-to" hotkeys don't seem to do anything on my system, regardless of whether OBS is focused or not. Have not had time to investigate in detail yet.

PatTheMav:
> I am still having issues here, seeing behavior that is different from what Pat described. Even with Pat's PR, "push-to" hotkeys don't seem to do anything on my system, regardless of whether OBS is focused or not. Have not had time to investigate in detail yet.

[...]

PR #9364, opened 2023-08-04
libobs: Add local event monitor for hotkey thread

PatTheMav:
### Description
Fixes held-down hotkeys (such as "push to talk") on macOS.

Fixes https://github.com/obsproject/obs-studio/issues/9023.

### Motivation and Context
To avoid a mismatch between the state of pressed keys as held by the hotkey thread and keys pressed while the application is in focus, local key events need to be monitored as well (even though they are already handled by Cocoa's main event loop).

### How Has This Been Tested?
Tested on macOS 13.4 using "push to talk" functionality on a media source.

### Types of changes

### Checklist:

### DIFF
--- a/libobs/obs-cocoa.m
+++ b/libobs/obs-cocoa.m
@@ -160,6 +160,7 @@ static bool dstr_from_cfstring(struct dstr *str, CFStringRef ref)
 struct obs_hotkeys_platform {
     volatile long refs;
     CFTypeRef monitor;
+    CFTypeRef local_monitor;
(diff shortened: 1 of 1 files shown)


gxalpha:
Would/Could this remove the need for Qt hotkeys on macOS altogether on macOS?

PatTheMav:
> Would/Could this remove the need for Qt hotkeys on macOS altogether on macOS?

In theory, yes. But the application-specific code currently assumes that Qt's event handling also triggers hotkeys (hence why the `ProcessHotkeys` meta function is registered).

PR #9367, opened 2023-08-05
CI: Add write permissions to contents scope to service validator

PatTheMav:
### Description
Adds the "write" permission to the contents scope for the service validator job in scheduled and dispatched workflows.

The action's documentation unfortunately does not spell out which API permissions are required for the token and neither GitHub's documentation spells this out directly, so this fix is based on an educated guess.

### Motivation and Context
Current implementation seems to lack required permissions to push a new branch to the repository, which leads to the job failing when actual unrecoverable changes to services are detected.

### How Has This Been Tested?
Needs to be tested after merging unfortunately as it needs to be merged on the master branch and then run as part of a scheduled or dispatched workflow run.

### Types of changes

### Checklist:

### DIFF
--- a/.github/workflows/dispatch.yaml
+++ b/.github/workflows/dispatch.yaml
@@ -37,6 +37,7 @@ jobs:
     runs-on: macos-13
     permissions:
       checks: write
+      contents: write
     steps:
       - uses: actions/checkout@v3
       - name: Check for Defunct Services 📉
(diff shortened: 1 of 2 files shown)


RytoEX (approved):
Kind of a swing in the dark. If it doesn't fix things, we can just revert it.

PR #9374, opened 2023-08-06
UI: Use os_generate_uuid for new uuid

computerquip:
### Description
This uses os_generate_uuid for UUID generation in the shared-update code. 

### Motivation and Context
Currently, the UUID is a random array of bytes. The 6 bits meant for version are missing meaning it's currently just a 128-bit random set of bytes. It doesn't follow the UUIDv4 format. In addition, the logic to generate the UUID is duplicated and different from what os_generate_uuid does already for us. This change makes it more consistent.

### How Has This Been Tested?
* Started the application with an empty global ini.
* Started the application with a populated INI (with the UUID already present).

### Types of changes
- Tweak
- Code cleanup

### Checklist:
       (Hopefully I did this right...)

### DIFF
--- a/UI/update/shared-update.cpp
+++ b/UI/update/shared-update.cpp
@@ -127,18 +127,6 @@ try {
 
 /* ------------------------------------------------------------------------ */
 
-void GenerateGUID(std::string &guid)
-{
-	const char alphabet[] = "0123456789abcdef";
-	QRandomGenerator *rng = QRandomGenerator::system();
-
-	guid.resize(40);
-
-	for (size_t i = 0; i < 40; i++) {
-		guid[i] = alphabet[rng->bounded(0, 16)];
-	}
-}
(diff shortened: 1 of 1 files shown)


notr1ch:
This is not intended to be a UUID, if anything the function should be renamed rather than changing the format of the random identifier.

computerquip:
Changing the format wasn't the intention, I somehow missed that the UUID from the original didn't have hyphens and that the one that os_generate_uuid output did.

I can either:
 * Make a more flexible UUID API, as the current one can't output the needed format.
 * Use Qt's QUuid which already supports this use case via `QUuid::Id128`.
 * Drop it entirely, it was meant to be a minor improvement.

computerquip:
Gonna close this for now, works as it is, no point in changing it without a better reason.

PR #9391, opened 2023-08-09
obs-ffmpeg: Remove media source mpegts format override for SRT and RIST

### LINKED ISSUE #9388: SRT can only handle mpegts
### Operating System Info

Windows 11

### Other OS

_No response_

### OBS Studio Version

29.1.3

_No response_

### OBS Studio Log URL

https://obsproject.com/logs/mWRRIynk3pm88ZHN

### OBS Studio Crash Log URL

_No response_

### Expected Behavior

when adding an SRT:// stream via media source, you are forced to use mpegts for the input format

### Current Behavior

Allow selecting other formats. 

### Steps to Reproduce

1. add media source
2. add srt uri `srt://127.0.0.1:9990?mode=listener&transtype=live&latency=0&recv_buffer_size=0`
3. set input format to `matroska`
4. re-open media source settings and see its forced mpegts
...

SRT (and rist) are a container agnostic protocols, this means ogg/ogv and matroska should work with it. 

this can easily be verified by opening mpv using `mpv.com srt://127.0.0.1:9990?mode=listener&transtype=live&latency=0&recv_buffer_size=0 --no-cache --untimed --no-demuxer-thread --video-sync=audio --vd-lavc-threads=1` or using `ffplay srt://127.0.0.1:9990?mode=listener&transtype=live&latency=0&recv_buffer_size=0 ` and sending a compatible stream to it such as 

`ffmpeg -re -stream_loop -1 -i .\input.video -c copy -f matroska "srt://127.0.0.1:9990?mode=caller&transtype=live&latency=0&recv_buffer_size=0"`

tt2468:
### Description
Removes the hardcoded format override for the media source

Fixes #9388

### Motivation and Context
FFmpeg is already very effective at detecting the correct input format for socket-style protocols (SRT, RIST, TCP, UDP etc). By overriding the format of SRT and RIST manually to mpegts, the user is being prevented from using other container formats via these protocols.

In the rare case that libavformat is unable to detect the correct container format for an SRT or RIST stream, the user may manually specify that format in the existing format field.

With the FFmpeg options field which was recently added, probe options may be specified by the user to further tune format detection.

### How Has This Been Tested?
Wasn't able to test due to not updating my Ubuntu 20.04 machine to 22 yet, but the change is dead simple.

### Types of changes

### Checklist:

### DIFF
--- a/plugins/obs-ffmpeg/obs-ffmpeg-source.c
+++ b/plugins/obs-ffmpeg/obs-ffmpeg-source.c
@@ -425,14 +425,7 @@ static void ffmpeg_source_tick(void *data, float seconds)
 	}
 }
 
-#define SRT_PROTO "srt"
 #define RIST_PROTO "rist"
-
-static bool requires_mpegts(const char *path)
-{
-	return !astrcmpi_n(path, SRT_PROTO, sizeof(SRT_PROTO) - 1) ||
-	       !astrcmpi_n(path, RIST_PROTO, sizeof(RIST_PROTO) - 1);
-}
 
 static void ffmpeg_source_update(void *data, obs_data_t *settings)
 {
@@ -468,10 +461,6 @@ static void ffmpeg_source_update(void *data, obs_data_t *settings)
 		should_restart_media = true;
 		input = obs_data_get_string(settings, "input");
 		input_format = obs_data_get_string(settings, "input_format");
-		if (requires_mpegts(input)) {
-			input_format = "mpegts";
-			obs_data_set_string(settings, "input_format", "mpegts");
-		}
 		s->reconnect_delay_sec =
 			(int)obs_data_get_int(settings, "reconnect_delay_sec");
 		s->reconnect_delay_sec = s->reconnect_delay_sec == 0


pkviet (changes requested):
This will unfortunately cause crashes.
See the original PR https://github.com/obsproject/obs-studio/pull/5962

I'd thus suggest to default to mpegts in case the input format is empty because (many ?) users won't know which format to use.

So rename the requires_mpegts function into a default_to_mpegts function and change the data set function to its default variant.

I can't test since I don't have currently access to my dev machine. It'll have to wait for next week.

moo-the-cow:
seems like a very attractive fix/feature is there anything missing to get merged? is there any specific build in case it needs to be tested still?

pkviet:
I unfortunately didn't create an issue for the 2 bugs referenced in https://github.com/obsproject/obs-studio/pull/5962 .
From a quick test, Media Source still works for SRT both on listener and caller modes.
So I'm going to approve this PR.

pkviet (approved):

RytoEX:
> I unfortunately didn't create an issue for the 2 bugs referenced in #5962 . From a quick test, Media Source still works for SRT both on listener and caller modes. So I'm going to approve this PR.

I'll try to give it a quick test later.

RytoEX:
@pkviet I'd guess this is why we can't reproduce the crash with RIST and "restart on activate":
https://github.com/obsproject/obs-studio/commit/5b0c09a9f467fa6bfd8baa994d3ef9e7d0047080

pkviet:
> @pkviet I'd guess this is why we can't reproduce the crash with RIST and "restart on activate":
> https://github.com/obsproject/obs-studio/commit/5b0c09a9f467fa6bfd8baa994d3ef9e7d0047080

Mmh i didn't double check rist.
I should also check the reactivate.

RytoEX:
> > @pkviet I'd guess this is why we can't reproduce the crash with RIST and "restart on activate":
> > [5b0c09a](https://github.com/obsproject/obs-studio/commit/5b0c09a9f467fa6bfd8baa994d3ef9e7d0047080)
> 
> Mmh i didn't double check rist. I should also check the reactivate.

Since we're forcing it off in that commit, if you wanted to test the crash condition, you'd have to pull this PR with that commit reverted.

pkviet:
> > I unfortunately didn't create an issue for the 2 bugs referenced in #5962 . From a quick test, Media Source still works for SRT both on listener and caller modes. So I'm going to approve this PR.
> 
> I'll try to give it a quick test later. IIRC, the problem was specifically RIST, not SRT.

The "restart when activate" bug was an issue for RIST only indeed. This setting is still disabled for RIST. 
I double checked that:
- RIST in Media Source works even when the input format is not set (avformat therefore makes a good job at detecting the mpegts format);
- that activating, deactivating and then reactivating a RIST Media source still worked.
I also checked that removing the previous fix (force "restart when activate" bool to false for RIST) did trigger the bug that https://github.com/obsproject/obs-studio/commit/5b0c09a9f467fa6bfd8baa994d3ef9e7d0047080 fixed.
So in summary, I think it's safe to merge the current PR as I haven't been able to crash Media Source neither for SRT nor RIST even when leaving blank the 'input format' field.

PR #9392, opened 2023-08-09
libobs/util: Define PRAGMA_WARN_DEPRECATION for MSVC

RytoEX:
### Description

Define PRAGMA_WARN_DEPRECATION for MSVC to enable using that define for cross-platform code, which allows us to write slightly less code when we need to suppress this warning.

### Motivation and Context

For cross-platform code, instead of something like this:
```
#if defined(_WIN32)
#pragma warning(push)
#pragma warning(disable : 4996)
#elif defined(__clang__) || defined(__GNUC__)
#pragma GCC diagnostic push
#pragma GCC diagnostic ignored "-Wdeprecated-declarations"
#endif
...
#if defined(_WIN32)
#pragma warning(pop)
#elif defined(__clang__) || defined(__GNUC__)
#pragma GCC diagnostic pop
#endif
```
or this:
```
	PRAGMA_WARN_PUSH
	PRAGMA_WARN_DEPRECATION
#pragma warning(disable : 4996)
	...
	PRAGMA_WARN_POP
```
I just want to write this:
```
	PRAGMA_WARN_PUSH
	PRAGMA_WARN_DEPRECATION
	...
	PRAGMA_WARN_POP
```

References:
* https://learn.microsoft.com/en-us/cpp/error-messages/compiler-warnings/compiler-warning-level-3-c4996?view=msvc-170

### How Has This Been Tested?

Built locally on Windows 11.

### Types of changes

### Checklist:

### DIFF
--- a/libobs/util/c99defs.h
+++ b/libobs/util/c99defs.h
@@ -54,7 +54,7 @@
 #ifdef _MSC_VER
 #define PRAGMA_WARN_PUSH __pragma(warning(push))
 #define PRAGMA_WARN_POP __pragma(warning(pop))
-#define PRAGMA_WARN_DEPRECATION
+#define PRAGMA_WARN_DEPRECATION __pragma(warning(disable : 4996))
 #elif defined(__clang__)
 #define PRAGMA_WARN_PUSH _Pragma("clang diagnostic push")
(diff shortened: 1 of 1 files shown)


RytoEX:
Pretty sure this is fine as-is. Merging.

PR #9397, opened 2023-08-09
obs-qsv11: Fix adapter enumeration in test app

RytoEX:
### Description

MFXEnumImplementations expects the adapter index to be within a range of valid Intel adapters produced by MFXLoad and the config filter. If a system has one non-Intel high performance GPU and one Intel iGPU, OBS and the test app will see the Intel iGPU as index 1 due to high performance hints, but MFXEnumImplementations will expect only one valid index, 0. In this scenario, passing a value of 1 to MFXEnumImplementations will cause it to abort and return MFX_ERR_NOT_FOUND (Provided index is out of possible range). This causes subsequent capabilities testing to fail.

To avoid this, let's track how many non-Intel adapters we see and subtract that number from adapterIdx to only pass valid index to MFXEnumImplementations.

### Motivation and Context

Want QSV to be correctly detected on multi-GPU mixed vendor systems.

### How Has This Been Tested?

Tested on my Windows 10 machine (i7-7700HQ + NVIDIA GTX 1060 + Intel HD Graphics 630). Made sure that after this change `obs-qsv-test.exe` produced the correct output and that QSV options were correctly available in the OBS UI and that I could use them.

### Types of changes

### Checklist:

### DIFF
--- a/plugins/obs-qsv11/obs-qsv-test/obs-qsv-test.cpp
+++ b/plugins/obs-qsv11/obs-qsv-test/obs-qsv-test.cpp
@@ -60,6 +60,7 @@ static bool get_adapter_caps(IDXGIFactory *factory, mfxLoader loader,
 			     mfxSession m_session, uint32_t adapter_idx)
 {
 	HRESULT hr;
+	static uint32_t idx_adjustment = 0;
 
 	ComPtr<IDXGIAdapter> adapter;
 	hr = factory->EnumAdapters(adapter_idx, &adapter);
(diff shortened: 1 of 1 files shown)


kkartaltepe:
Ack, this is part of the platform abstraction as `void check_adapters(struct adapter_info *adapters, size_t *adapter_count)`

The test app is only called in windows, we directly call vaapi in the linux impl. So there shouldnt be any impact on linux. We discussed using the vaapi drm selector for linux so that will also be consistent across our enumeration and mfx (unlike this case in windows).

kkartaltepe (approved):

PR #9400, opened 2023-08-10
CONTRIBUTING.rst: Revise capitalization guideline text of commit titles

norihiro:
### Description

This PR adds a guideline that the 2nd and later words of the commit titles should not be capitalized.

### Motivation and Context

In some PRs such as #9258, each word of the commit titles is capitalized.

Current guideline does not clearly mention how the 2nd and later words should be capitalized or not but just saying to see examples.

### How Has This Been Tested?

Spells and grammers are checked on Gmail's spell-checker.
However, it would be better to be reviewed by a native English speaker.

### Types of changes

### Checklist:

### DIFF
--- a/CONTRIBUTING.rst
+++ b/CONTRIBUTING.rst
@@ -71,7 +71,8 @@ Commit Guidelines
 
 - Prefix each commit's titles with the module name, followed by a colon
   and a space (unless modifying a file in the base directory).  After
-  that, the first word should be capitalized.
+  that, the first word should be capitalized and the rest of the words
+  should not be capitalized.
 
   So for example, if you are modifying the obs-ffmpeg plugin::
 


CONTRIBUTING.rst
```diff
@@ -71,7 +71,8 @@ Commit Guidelines
 
 - Prefix each commit's titles with the module name, followed by a colon
   and a space (unless modifying a file in the base directory).  After
-  that, the first word should be capitalized.
+  that, the first word should be capitalized and the rest words should
```
CodeYan01, line 74: that, the first word should be capitalized and the rest of the words should
```suggestion
  that, the first word should be capitalized and the rest of the words should
```

norihiro: Fixed and wrapped at 72-column in the file.

TianQiBuTian:
It may be necessary to add "except for proper nouns and codes that require capitalization".

notr1ch:
This feels kind of unnecessary with the examples already there.

gxalpha (commented):
I'm not even sure the change is correct as-is, there are several examples of words (usually names) needing to be capitalized.
> Improve WHIP compliance
Follow HTTP redirects
Improve AV1 quality parameters

If you look at the [post](https://cbea.ms/git-commit) linked further up in the file regarding commit messages, it just says to use imperative mood, which is good probably good enough of an explanation (*usually* words should be lower case, but there may be instances where upper case is appropriate).
In my opinion just adding "and the rest of the words should not be capitalized" would lead to more confusion than it solves.

norihiro:
I will consider rephrasing or just closing this.

norihiro:
I checked some documents on IEEE and ACM but I couldn't find something helpful to describe in simple without confusion. I will close this.

PR #9401, opened 2023-08-10
CI: Update to clang-format 16

RytoEX:
### Description

Update to clang-format 16. I used clang-format 16.0.5 for the formatting because that's what VS 2022 17.7 ships.

This does not cover submodules (obs-browser, libdshowcapture, obs-websocket). Those can be done in tandem with this or separately.

### Motivation and Context

Want to stop pinning to clang-format 13.

### How Has This Been Tested?

Ran the formatter. I didn't do build tests, but CI will.

### Types of changes

### Checklist:

### DIFF
--- a/.clang-format
+++ b/.clang-format
@@ -1,4 +1,4 @@
-# please use clang-format version 13 or later
+# please use clang-format version 16 or later
 
 Standard: c++17
 AccessModifierOffset: -8
@@ -52,8 +52,8 @@ ContinuationIndentWidth: 8
 Cpp11BracedListStyle: true
 DerivePointerAlignment: false
 DisableFormat: false
-FixNamespaceComments: false
-ForEachMacros: 
+FixNamespaceComments: true
+ForEachMacros:
   - 'json_object_foreach'
   - 'json_object_foreach_safe'
   - 'json_array_foreach'
--- a/.github/actions/run-clang-format/action.yaml
+++ b/.github/actions/run-clang-format/action.yaml
@@ -28,7 +28,7 @@ runs:
         echo ::group::Install Dependencies
         eval "$(/home/linuxbrew/.linuxbrew/bin/brew shellenv)"
         echo "/home/linuxbrew/.linuxbrew/bin:/home/linuxbrew/.linuxbrew/sbin" >> $GITHUB_PATH
-        echo "/home/linuxbrew/.linuxbrew/opt/clang-format@13/bin" >> $GITHUB_PATH
+        echo "/home/linuxbrew/.linuxbrew/opt/clang-format@16/bin" >> $GITHUB_PATH
         brew install --quiet zsh
         echo ::endgroup::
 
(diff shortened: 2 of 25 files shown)


derrod:
There are an number of currently commented out options like https://github.com/obsproject/obs-studio/blob/master/.clang-format#L147-L149 that we could probably enable as part of an update.

The namespace comments could be turned off via `FixNamespaceComments` (added in clang-format 5!) but I'd almost prefer explicitly turning that on to see if it will then also add those in C++ code (right now it's just ObjC++).

RytoEX:
> There are an number of currently commented out options like https://github.com/obsproject/obs-studio/blob/master/.clang-format#L147-L149 that we could probably enable as part of an update.
> 
> The namespace comments could be turned off via `FixNamespaceComments` (added in clang-format 5!) but I'd almost prefer explicitly turning that on to see if it will then also add those in C++ code (right now it's just ObjC++).

That last highlighted line has no version marker. Should we uncomment it or ignore it?

derrod:
Seems like the line immediately afterwards sets it to a different value that's not valid based on the documentation (https://clang.llvm.org/docs/ClangFormatStyleOptions.html#breakbeforeconceptdeclarations), maybe the values changed with newer versions? Would be a question for @PatTheMav

RytoEX:
I've pushed the enablement of `FixNamespaceComments` for C++ as a fixup commit for now (to be squashed, if acceptable). I think I'll defer any ObjC adjustments to a separate PR to keep this scope down and also because clang-format on my Windows machine does not run with the ObjC options enabled. Specifically, I get this error:
```
error: unknown key 'QualifierOrder'
```

RytoEX:
> I've pushed the enablement of `FixNamespaceComments` for C++ as a fixup commit for now (to be squashed, if acceptable). I think I'll defer any ObjC adjustments to a separate PR to keep this scope down and also because clang-format on my Windows machine does not run with the ObjC options enabled. Specifically, I get this error:
> 
> ```
> error: unknown key 'QualifierOrder'
> ```

Figured this out with off-thread help from @tytan652 . We would need to change `QualifierAlignment` to `Custom`. Will still defer the ObjC changes to later, but this PR at least unblocks those changes.

PatTheMav:
> > I've pushed the enablement of `FixNamespaceComments` for C++ as a fixup commit for now (to be squashed, if acceptable). I think I'll defer any ObjC adjustments to a separate PR to keep this scope down and also because clang-format on my Windows machine does not run with the ObjC options enabled. Specifically, I get this error:
> > ```
> > error: unknown key 'QualifierOrder'
> > ```
> 
> Figured this out with off-thread help from @tytan652 . We would need to change `QualifierAlignment` to `Custom`. Will still defer the ObjC changes to later, but this PR at least unblocks those changes.

I suggest removing `QualifierOrder` entirely as their docs themselves say that `clang-format` does not have the necessary semantics to make this a non-breaking formatting change.

PatTheMav:
> Seems like the line immediately afterwards sets it to a different value that's not valid based on the documentation (https://clang.llvm.org/docs/ClangFormatStyleOptions.html#breakbeforeconceptdeclarations), maybe the values changed with newer versions? Would be a question for @PatTheMav
> 
> Edit: Indeed seems that it was changed from a bool to having multiple options somewhere between 14 and 16: https://releases.llvm.org/14.0.0/tools/clang/docs/ClangFormatStyleOptions.html

Yeah I think I commented out the variant that was _compatible_ with 16 only because I used 16.0.6 as the baseline and then commented out stuff not supported by 13 or used the old value type supported by that specific version.

RytoEX:
Approved off-thread. We'll handle submodules separately.

PR #9402, opened 2023-08-10
UI: Use QGroupBox::toggled signal for group changes

gxalpha:
### Description

The QGroupBox::clicked is specifically for mouse clicks. This means that it's not emitted if the "checked" property is modified through other means than the mouse, for example programatically or through accessibility software like VoiceOver.
However, we do want to catch such events, so the QGroupBox::toggle signal (which as per the Qt documentation is the notified signal for the "checked" property) is the appropriate one to use.

As for the first commit, loading needs to be set to true to avoid the signal from activating the button while the settings are loading in. This should have always been done but just wasn't detected as the correct signal wasn't used so it wasn't triggered until now (also because there are no other settings in Accessibility that would have revealed this bug).

### Motivation and Context

Noticed when rebasing https://github.com/obsproject/obs-studio/pull/6700 (yet to be force-pushed).

### How Has This Been Tested?

macOS 14
Tested that without the fix, using VoiceOver to toggle the "Use Different Colors" GroupBox in the Accessibility settings (a bit ironic I must say :p), the Apply button didn't activate.
Tested that with the fix, it does activate when using VoiceOver to toggle it.
Tested that both with and without the fix, using the mouse to toggle the checkbox changes activates the Apply button.

### Types of changes

### Checklist:

### DIFF
--- a/UI/window-basic-settings-a11y.cpp
+++ b/UI/window-basic-settings-a11y.cpp
@@ -50,6 +50,7 @@ void OBSBasicSettings::LoadA11ySettings(bool presetChange)
 {
 	config_t *config = GetGlobalConfig();
 
+	loading = true;
 	if (!presetChange) {
 		preset = config_get_int(config, "Accessibility", "ColorPreset");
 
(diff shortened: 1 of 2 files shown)


RytoEX (approved):
Looks good to me.

PR #9403, opened 2023-08-10
mac-virtualcam: Fix crash on macOS when starting virtualcam output

PatTheMav:
### Description
Fixes a crash on start of virtualcam output on macOS introduced by a recent change to use `NSData` instead of variable-length arrays.

### Motivation and Context
Array iterations need to be determined by _dividing_ the byte size of the returned data by the size of a single object (not multiplied).

The actual value of `used` is undocumented, so using the more obvious `size` parameter makes the code also safer to use.

### How Has This Been Tested?
Tested on macOS 13.4, actively running virtual camera output with the new camera extension.

### Types of changes

### Checklist:


RytoEX (approved):
Looks okay at a glance.

PR #9408, opened 2023-08-11
Docs: Add source-specific signals and procedures

Penwy:
### Description

This adds documentation for custom signals and procedures defined by specific source types.
Two additional minor changes have been bundled with it :
- Fixing the declaration strings of the procedures defined by the slideshow source in order to conform to preexisting conventions.
- Adding the calldata information for the media signals in the doc.

### Motivation and Context

Those signals and procedures were not documented or hinted at anywhere.

### How Has This Been Tested?

Ubuntu 22.04
Built the doc
Looked at it

### Types of changes

### Checklist:


CodeYan01:
i suggest having a link to the definition of the source-specific signals and procedures, so readers also learn where to find them, especially since docs doesn't reflect the actual code 100% of the time. Though it will be a problem hyperlinking to master, given that line numbers can change, procedure definitions can change, etc. On the other hand permalinks would also work, but would we want to take on the task of maintaining the permalinks when there are changes? 

For this next one, I don't have any suggestion, but I'd like to point this out, just in case anyone else has a better idea to present these two signals with the same name, but defined by different sources.
image

it makes sense that the entry would be repeated due to most of it being different, but i can see it look a bit confusing at first glance.

RytoEX (approved):
Seems fine. At first, the code change in `plugins/image-source/obs-slideshow.c` jumped out as possibly needing a separate commit or PR. However, on second glance, it is already a separate commit, and the changed string appears mostly to be for internally documenting the callback function rather than actually being a functional change.

RytoEX:
> i suggest having a link to the definition of the source-specific signals and procedures, so readers also learn where to find them, especially since docs doesn't reflect the actual code 100% of the time. Though it will be a problem hyperlinking to master, given that line numbers can change, procedure definitions can change, etc. On the other hand permalinks would also work, but would we want to take on the task of maintaining the permalinks when there are changes?
> 
> For this next one, I don't have any suggestion, but I'd like to point this out, just in case anyone else has a better idea to present these two signals with the same name, but defined by different sources.

This all seems like it could be done in a follow-up PR rather than being a required change that blocks this PR.

Penwy:
mb, had forgotten a proc in the dshow source

RytoEX (approved):

PR #9410, opened 2023-08-11
UI: Add high frame rates (120, 144, 240) to general video properties

FBirth:
### Description
I added higher frame rates 120, 144 and 240 fps as defaults in the device properties for macOS. The higher framerates are only shown if the capture device supports them.

Device with max. 120 fps
grafik

Device with max. 240 fps
grafik

### Motivation and Context
Newer capture devices from Elgato and other support capturing at these high framerates.
Currently only standard frame rates up to 60 fps are shown plus the highest possible framerate reported by the devices.
If a device supports 240 fps 144 and 120 fps were not shown in the list.

### How Has This Been Tested?
I connected a UVC capture card capable of 240 fps. It showed all the standard frame rates plus 240 fps.
After my change also 120 and 144 fps were shown. I tested preview and capturing from the device at trhe high frame rates and found no issues.
macOS

### Types of changes

### Checklist:


gxalpha:
We generally consider anything above 120 to be unsupported (on-your-own territory). I doubt this will get merged as-is.

derrod:
Specifically for consoles you might want to add 119.88 (i.e. 59.94 * 2) as that's what they will be using rather than actual 120.

FBirth:
> Specifically for consoles you might want to add 119.88 (i.e. 59.94 * 2) as that's what they will be using rather than actual 120.

@derrod, should I replace 120 with 119.88 or rather keep both?

derrod (approved):
Seems fine to me.

gxalpha:
The two commits are doing the same, simply adding different values, correct? If so, could you squash them together?

PR #9411, opened 2023-08-11
UI: Use inline function instead of macro to set color

gxalpha:
### Description

Changes the `SetStyle` macro in a11y settings to a static inline function.
Macros can be hard to read and are usually not very friendly to use in a debugger. Using a static inline function instead [gives the same result](https://gcc.gnu.org/onlinedocs/gcc-3.1.1/gcc/Inline.html) when compiling with optimizations,  with the advantage of better syntax highlighting in IDEs and better debugger support.

### Motivation and Context

Noticed this when working on https://github.com/obsproject/obs-studio/pull/9402. Like to fix (change? Not sure "fix" is the right word since it's not really a bug) small things like these when I see them.

### How Has This Been Tested?

macOS 14
Tested that colors in Accessibility settings are still correctly set, as before.

### Types of changes

### Checklist:

### DIFF
--- a/UI/window-basic-settings-a11y.cpp
+++ b/UI/window-basic-settings-a11y.cpp
@@ -133,26 +133,25 @@ void OBSBasicSettings::SaveA11ySettings()
 	main->RefreshVolumeColors();
 }
 
-#define SetStyle(label, colorVal)                                             \
-	color = color_from_int(colorVal);                                     \
-	color.setAlpha(255);                                                  \
-	palette = QPalette(color);                                            \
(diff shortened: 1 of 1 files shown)


notr1ch:
I'm not sure such a big function should be inline, we might get smaller code with it being a regular function.

gxalpha:
You're probably right, changed to just be `static` instead of `static inline`. And if some compiler still wants to inline it I think nothing's stopping it from doing so.

RytoEX (approved):

PR #9416, opened 2023-08-11
UI: Fix transform shortcuts with multiple items selected

gxalpha:
### Description

Firstly, removes the `UpdateTransformShortcuts()` method introduced in c33fa8b which was trying to re-implement the behavior of `UpdateEditMenu()`.
Secondly, updates `UpdateEditMenu()` to account for sources without video. Sources without video shouldn't be able to have their transform edited, copied, pasted, or changed in any other way (because they don't have one).

The Transform menu items are now enabled/disabled as per the following ("AV" in this context means both sources with audio *and* video as well as sources with only video, they behave the same. "Paste Transform if applicable" means "Paste Transform is enabled if a transform got copied before".):
- No item selected: 
  - No Transform items enabled.
- Single item selected:
  - Audio-only: No Transform items enabled
  - AV, locked: Edit and Copy Transform enabled. Paste Transform if applicable.
  - AV, unlocked: All Transform items enabled, except Paste Transform as applicable.
- Multiple items selected: Copy and Edit Transform always disabled
  - Only audio-only: No Transform items enabled.
  - Only AV, locked: No Transform items enabled, except Paste if Applicable.
  - Only AV, unlocked: All items enabled except Copy and Edit, and Paste if applicable.
  - Audio-only + AV, locked: No Transform items enabled.
  - Audio-only + AV, unlocked: All items enabled except Copy and Edit, and Paste if applicable.
  - AV, locked + AV, unlocked: All items enabled except Copy and Edit, and Paste if applicable.
  - Audio-only + AV, locked + AV, unlocked: All items enabled except Copy and Edit, and Paste if applicable.

(The combinations are the power set of the set of the relevant source/sceneitem types {"Audio-only", "AV, locked", "AV, unlocked"} , plus one instance of multiple items of one type for each type, making 2^3+3=11 different situations)
~~Also something something unit tests :p.~~

I noticed that the right-click menu something doesn't even include Transform when it should (it should always except when No Transform items are enabled), but that is a separate issue.

### Motivation and Context

As noted in https://github.com/obsproject/obs-studio/pull/9349, `UpdateTransformShortcuts()` was still buggy, it would enable and disable items when it shouldn't, particularly when multiple sceneitems are selected.
I now noticed that it isn't needed at all.

### How Has This Been Tested?

macOS 14
Checked all situations described above.

### Types of changes

### Checklist:

### DIFF
--- a/UI/window-basic-main.cpp
+++ b/UI/window-basic-main.cpp
@@ -3391,28 +3391,9 @@ void OBSBasic::SourceToolBarActionsSetEnabled()
 	RefreshToolBarStyling(ui->sourcesToolbar);
 }
 
-void OBSBasic::UpdateTransformShortcuts()
-{
-	bool hasVideo = false;
-
-	OBSSource source = obs_sceneitem_get_source(GetCurrentSceneItem());
-
-	if (source) {
-		uint32_t flags = obs_source_get_output_flags(source);
-		hasVideo = (flags & OBS_SOURCE_VIDEO) != 0;
-	}
-
-	ui->actionEditTransform->setEnabled(hasVideo);
-	ui->actionCopyTransform->setEnabled(hasVideo);
-	ui->actionPasteTransform->setEnabled(hasVideo ? hasCopiedTransform
-						      : false);
-	ui->actionResetTransform->setEnabled(hasVideo);
-}
-
 void OBSBasic::UpdateContextBar(bool force)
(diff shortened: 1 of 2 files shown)


RytoEX:
> removes the `UpdateTransformShortcuts()` method

It looks like this PR adds such a method? I don't see any removals.

Issue #9216, opened 2023-07-08
Auto-configuration wizard shows your current frame rate in scientific notation if FPS is >= 100


matoi974:
### Operating System Info

Windows 10

### OBS Studio Version

29.1.3

### OBS Studio Log URL

n/a

### Expected Behavior

Fractional frame rates should be displayed as-is (i.e. 120/1, 60000/1001).

### Current Behavior

Fractional frame rates are displayed in scientific notation (1.2e+02).
![image](https://github.com/obsproject/obs-studio/assets/63289008/05d9b9f9-28b8-4cb5-b74b-21a875914c77)

### Steps to Reproduce

1. In Settings>Video, set FPS type to fractional, numerator to 120, denominator to 1.
2. Run the auto-configuration wizard, choose "Optimize for recording".
3. Open the frame rate selection drop-down.

### Anything else we should know?

Reproduced on MacOS 13 and Windows 10, so presumably affects all platforms.

Xorshasia:
This happens anytime the framerate selected in Settings > Video has more than two digits in the integral portion. The actual framerate can be either using integer or fractional.

PR #9421, opened 2023-08-11
UI: Don't cap auto-config wizard FPS whole number precision at 2

gxalpha:
### Description

Capping the precision for whole numbers results in scientific notation being used for any number that's larger than what can be displayed with the given precision. In case of 2, that would be anything larger than 99, resulting in 120 for example to be shown as 1.2e+02. Let's not cap the precision for whole numbers ourselves and just use the default value (which at the time of writing would be 6).

### Motivation and Context

Fixes https://github.com/obsproject/obs-studio/issues/9216

### How Has This Been Tested?

macOS 14
Auto-config wizard now shows "120" instead of "1.2e+02" which is much more helpful. Also tested with 119.88, 60, 30, and 29.97 FPS.

### Types of changes

### Checklist:


RytoEX (approved):
Looks good to me.

PR #9422, opened 2023-08-12
obs-frontend-api: Silence errors when removing callbacks

norihiro:
### Description

This PR will remove the error message when the frontend API `c` is not available in `obs_frontend_remove_*_callback` functions.

### Motivation and Context

Less number of errors is better.

Some plugins, such as obs-websocket, try to remove their callbacks when unloading the module though the frontend API is no longer availbale.
When the frontend API is destroyed, the lists of the registered callbacks are also destroyed so that the callbacks are no longer called.

### How Has This Been Tested?

Run obs and exit it.
Checked there are no log lines like below.
```
Tried to call obs_frontend_remove_save_callback with no callbacks!
Tried to call obs_frontend_remove_event_callback with no callbacks!
Tried to call obs_frontend_remove_event_callback with no callbacks!
```

### Types of changes

### Checklist:

### DIFF
--- a/UI/obs-frontend-api/obs-frontend-api.cpp
+++ b/UI/obs-frontend-api/obs-frontend-api.cpp
@@ -360,7 +360,7 @@ void obs_frontend_add_event_callback(obs_frontend_event_cb callback,
 void obs_frontend_remove_event_callback(obs_frontend_event_cb callback,
 					void *private_data)
 {
-	if (callbacks_valid())
+	if (c)
 		c->obs_frontend_remove_event_callback(callback, private_data);
 }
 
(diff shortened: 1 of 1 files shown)


tytan652:
> Less number of errors is better.

No, if obs-websocket is doing something wrong we must not hide it under the carpet.
It's on the plugin side that it must be fixed.

norihiro:
I don't think the plugin does something wrong. The plugin registers some callbacks in `obs_module_load()` and remove them in `obs_module_unload()`.

After the frontend API is destroyed, just ignoring `obs_frontend_remove_*_callback` is not harmful at all. If it's not ok to ignore it, I would like to reorder unload plugins earlier than destroying the frontend API.

tytan652:
> and remove them in `obs_module_unload()`.

This is where it's wrong those should be removed on the `OBS_FRONTEND_EVENT_EXIT` the last moment where the API "exist".
The unload happens latter as expected.

gxalpha:
See also https://github.com/obsproject/obs-studio/pull/8735, which introduced the proper frontend api shutdown. I’d agree that this should be fixed on the plugin side instead.

norihiro:
I'm afraid it's not a good design that `obs_module_unload()` cannot cleanup something allocated at `obs_module_load()`. I will consider another solution.

PR #9426, opened 2023-08-12
obs-qsv11: Fix VPL initialization on intel-mediaSDK

kkartaltepe:
### Description

VPL does not have a consistent initialization process for the various
backends it dispatches to. So this resolves the new VPL dispatcher
failing to initialize the video session when it dispatches to
intel-mediaSDK backend. On the oneVPL-intel-gpu backend VA-API sessions
are correctly initialized so this wasnt noticed when tested on newer
hardware.

I also tried to sneak in a small fixup for the low power fallback inside `InitParams`.
we call Query internally and possibly fixup the low power state and expect the
next status check to be done after the 2nd Query call.

### Motivation and Context

QSV Should work on 7th gen like it did before the shift to the vpl dispatcher.

### How Has This Been Tested?

Arc card and 7th gen chip both work again.

### Types of changes

### Checklist:

### DIFF
--- a/plugins/obs-qsv11/CMakeLists.txt
+++ b/plugins/obs-qsv11/CMakeLists.txt
@@ -58,5 +58,5 @@ elseif(OS_LINUX)
 
   target_sources(obs-qsv11 PRIVATE common_utils_linux.cpp)
 
-  target_link_libraries(obs-qsv11 PRIVATE Libva::va Libva::drm)
+  target_link_libraries(obs-qsv11 PRIVATE Libva::va Libva::drm Libva::wayland Libva::x11)
 endif()
--- a/plugins/obs-qsv11/common_utils_linux.cpp
+++ b/plugins/obs-qsv11/common_utils_linux.cpp
@@ -5,6 +5,8 @@
 #include <util/c99defs.h>
 #include <util/dstr.h>
 #include <va/va_drm.h>
+#include <va/va_x11.h>
+#include <va/va_wayland.h>
 #include <va/va_str.h>
 #include <stdio.h>
 #include <stdlib.h>
@@ -12,6 +14,9 @@
 #include <fcntl.h>
 #include <string.h>
 #include <dirent.h>
+
+#include <obs.h>
+#include <obs-nix-platform.h>
 
 mfxStatus simple_alloc(mfxHDL pthis, mfxFrameAllocRequest *request,
 		       mfxFrameAllocResponse *response)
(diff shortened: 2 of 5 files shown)


plugins/obs-qsv11/CMakeLists.txt
```diff
@@ -58,5 +58,5 @@ elseif(OS_LINUX)
 
   target_sources(obs-qsv11 PRIVATE common_utils_linux.cpp)
 
-  target_link_libraries(obs-qsv11 PRIVATE Libva::va Libva::drm)
+  target_link_libraries(obs-qsv11 PRIVATE Libva::va Libva::drm Libva::wayland)
```
tytan652, line 61: Libva has it's own X11 library like DRM and Wayland.
To avoid linkage issue it should be added too.

kkartaltepe: It doesnt provide any actual linkages, vaGetDisplay is X11 by definition. I can remove the X11 header if we want to be clear about it.

kkartaltepe: Sorry poked the libraries a bit and yea we should link it, ill add it in a bit.

plugins/obs-qsv11/common_utils_linux.cpp
```diff
@@ -102,9 +107,29 @@ mfxStatus Initialize(mfxVersion ver, mfxSession *pSession,
...
+			vaGetDisplay((Display *)obs_get_nix_platform_display());
+	} else if (obs_get_nix_platform() == OBS_NIX_PLATFORM_WAYLAND) {
+		vaDisplay = vaGetDisplayWl(
+			(wl_display *)obs_get_nix_platform_display());
+	}
+
 	sts = MFXCreateSession(loader, 0, pSession);
 	MSDK_CHECK_RESULT(sts, MFX_ERR_NONE, sts);
 
+	// VPL Expects the VADisplay to be initialized.
```
RytoEX, line 122:
```suggestion
	// VPL expects the VADisplay to be initialized.
```

RytoEX (changes requested):
One comment nit.

Awaiting other changes.

plugins/obs-qsv11/common_utils_linux.cpp
kkartaltepe: Done.

RytoEX (approved):
Looks okay at a glance.

PR #9427, opened 2023-08-13
libobs: Add Source Profiling

derrod:
### Description

This PR adds libobs plumbing for profiling the performance of individual sources.

- [x] OpenGL (works on NVIDIA and AMD on Windows, does not work on macOS (Apple doesn't implement the GL command), untested on Linux)
- [x] Figure out what to do about views/mixes
  + Just collect all of them as samples for the current frame, it's a bit confusing, but there's no good way around it.

A GUI is being worked on separately, a build with the previous WIP GUI can be downloaded here: https://github.com/derrod/obs-studio/actions/runs/7045389326

### Motivation and Context

Diagnosing performance issues with OBS can be difficult. In many cases users get told to duplicate their current scene collection and delete stuff until things improve. That's obviously not ideal.

This pull request introduces an API for determining the performance characteristics of individual sources to find outliers and problematic ones, e.g. a complex shader filter that takes too long to render.

Additionally it can help spot bottlenecks with asynchronous sources, e.g. if a capture card or video does not reach its full frame rate due to resource limitations.

A UI will follow later once the libobs interface has been reviewed and finalised.

### How Has This Been Tested?

Rendered some sources, observed the numbers changing (Windows 11, AMD RDNA2, DX11 and OpenGL).

### Types of changes

### Checklist:

### DIFF
--- a/docs/sphinx/reference-libobs-util-source-profiler.rst
+++ b/docs/sphinx/reference-libobs-util-source-profiler.rst
@@ -1,0 +1,91 @@
+Source Profiler
+===============
+
+The source profiler is used to get information about individual source's performance.
+
+.. struct:: profiler_result
+
+.. member:: uint64_t profiler_result.tick_avg
+            uint64_t profiler_result.render_avg
+   
+   Average execution time of this source's tick and render functions within the sampled timeframe (5 seconds).
+   
+   Note that a source will be ticked only once per frame, but may be rendered multiple times.
+
+.. member:: uint64_t profiler_result.tick_max
+            uint64_t profiler_result.render_max
+
+   Maximum execution time of this source's tick and render functions within the sampled timeframe (5 seconds).
+
+.. member:: uint64_t profiler_result.render_gpu_avg
+            uint64_t profiler_result.render_gpu_max
+   
+   Average and maximum execution time for GPU rendering to execute within the sampled timeframe (5 seconds).
+   
+   Note that GPU timing is not supported on macOS and is of limited accuracy due to variations in GPU load/clock speed.
+
+.. member:: uint64_t profiler_result.render_sum
+            uint64_t profiler_result.render_gpu_sum
+
+   Sum of all CPU/GPU render time in a frame, averaged over the sampled timeframe (5 seconds).
+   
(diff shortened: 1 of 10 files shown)


cg2121:
Would it be possible to have percentage columns for both cpu and gpu? I believe this would be what most users would understand.

derrod:
> Would it be possible to have percentage columns for both cpu and gpu? I believe this would be what most users would understand.

Not really because they wouldn't add up to 100, for example a scene's render time also contains all the items contained within. And with GPU stuff it's not directly cumulative as far as I understand. And since this meant for power users it's probably fine for it to be more technical/accurate.

derrod:
We're back after an unscheduled interruption due to me messing up a rebase somehow.

alexandre-leites:
I just want to add that this is very needed. As a heavy user of OBS scenes, plugins, transitions and filters, I would love to know which one I can remove to have a reasonable resource saving and rethink my configuration. Sometimes removing one plugin is way more worthy than removing other 5s or so. Or reconfiguring some scenes.

jatedesigns:
> I just want to add that this is very needed. As a heavy user of OBS scenes, plugins, transitions and filters, I would love to know which one I can remove to have a reasonable resource saving and rethink my configuration. Sometimes removing one plugin is way more worthy than removing other 5s or so. Or reconfiguring some scenes.
> 
> Even if we get this only on logs (text or something) for the first release, it is worth merging.

Totally agree, and yes please! I've had rogue shaders and filters that were eating 80% of my GPU, but I thought everything I was adding after that was causing the issues (because it didn't lag until those were added). This would be extremely useful!

alinsavix:
Question: Would it be possible to expose these metrics right now, via something like obs-websocket, or obslua, or at least via "dump what we've gathered at shutdown time"? That would make this change both useful and more testable, without giving it a dependency on figuring out the proper UI for it (since UI changes take much longer and in this case are probably harder to do than the actual profiling itself).

This seems like a _very_ much needed feature, and it doesn't seem like there should be any hard requirement on having a UI for it finished (or even designed), for it to be merged and be extremely useful. Plus, if we can merge this and make the data actually accessible, it's virtually guaranteed that someone will come along and write some tools for making use of that data, even before there's an official UI.

Maybe that's a middle ground between "don't have it at all" and "waiting on a full-blown UI"?

Lain-B (approved):
Although I feel like I'm unable to digest everything going on here, I feel like I can give this my approval.

PR #9432, opened 2023-08-14
UI: Fix possible race condition in DrawSpacingHelpers

PatTheMav:
### Description
Replace direct lookup of currently selected scene item on the Qt selection model with data available through prior call to `obs_scene_enum_items`.

### Motivation and Context
The use of GetCurrentSceneItem can lead to a race condition between the graphics thread (where this function will be run) and the main UI thread, as both access the scene list model and iterate through its members which can change during iteration (as there is no lock around its access).

### How Has This Been Tested?
Tested on macOS 13, confirmed that spacing helpers are only drawn on a single selected item.

### Types of changes

### Checklist:

### DIFF
--- a/UI/window-basic-preview.cpp
+++ b/UI/window-basic-preview.cpp
@@ -2461,10 +2461,10 @@ void OBSBasicPreview::DrawSpacingHelpers()
 
 	obs_scene_enum_items(main->GetCurrentScene(), FindSelected, &data);
 
-	if (data.sceneItems.size() > 1)
+	if (data.sceneItems.size() != 1)
 		return;
 
-	OBSSceneItem item = main->GetCurrentSceneItem();
+	OBSSceneItem item = data.sceneItems[0];
 	if (!item)
 		return;
 


notr1ch:
I think we'll also need a local variable to hold the result of `main->GetCurrentScene()` which is an `OBSScene`, currently it gets immediately referenced and released so the next call in the function might also point to a different scene. And if the scene itself is destroyed from lack of references then the sceneitems will likely become invalid too.

PatTheMav:
> I think we'll also need a local variable to hold the result of `main->GetCurrentScene()` which is an `OBSScene`, currently it gets immediately referenced and released so the next call in the function might also point to a different scene. And if the scene itself is destroyed from lack of references then the sceneitems will likely become invalid too.

So we should hold a strong reference for as long as we interact with the underlying data and release once we done?

Lain-B:
I'm going to merge this, and then I'll add an extra commit for `GetCurrentScene()` after.

Issue #9118, opened 2023-06-21
No audio encoder fallback


prgmitchell:
### Operating System Info

Windows 11

### OBS Studio Version

29.1.2

### OBS Studio Log URL

https://obsproject.com/logs/vNVGSo3AmA663Shx

### Expected Behavior

Audio encoder should fallback to the best available audio encoder.

### Current Behavior

Out errors with a generic message and the user is unable to change their audio encoder. Sorry for the non-english screenshot but it was taken from a user in the support discord, along with their log.

![image](https://github.com/obsproject/obs-studio/assets/86465454/a4e97b29-7e31-48fc-b6af-80b22c29c9d0)
![image](https://github.com/obsproject/obs-studio/assets/86465454/06ff04f9-1465-4752-886d-a5406eeb5b4b)

### Steps to Reproduce

1. Install iTunes
2. Select CoreAudio AAC as the audio encoder and close OBS
3. Uninstall iTunes
4. Open OBS and try to start output
...

PR #9441, opened 2023-08-15
UI: Set advanced audio encoder to invalid if missing

derrod:
### Description

Somewhat addresses #9118 by making it at least possible to change the encoder if only one is available by setting the combobox to an invalid state. Users will be prevented from saving settings until they select a valid item.

### Motivation and Context

Want to make this issue recoverable without config editing.

### How Has This Been Tested?

Validated that when CoreAudio is missing the combobox will now show the placeholder instead, then once selecting "FFmpeg AAC" and saving the config will be in a working state again.

### Types of changes

### Checklist:

### DIFF
--- a/UI/window-basic-settings.cpp
+++ b/UI/window-basic-settings.cpp
@@ -920,6 +920,8 @@ OBSBasicSettings::OBSBasicSettings(QWidget *parent)
 		&OBSBasicSettings::AdvOutRecCheckCodecs);
 
 	// Set placeholder used when selection was reset due to incompatibilities
+	ui->advOutAEncoder->setPlaceholderText(
+		QTStr("CodecCompat.CodecPlaceholder"));
 	ui->advOutRecEncoder->setPlaceholderText(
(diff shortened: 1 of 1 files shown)


UI/window-basic-settings.cpp
```diff
@@ -4970,6 +4973,11 @@ static void DisableIncompatibleCodecs(QComboBox *cbox, const QString &format,
 						 ? strEncLabel
 						 : obs_encoder_get_display_name(
 							   encoderId.c_str());
+
+		/* Something has gone horribly wrong and there's no encoder */
+		if (encoderId.empty())
```
gxalpha, line 4978: Is the first part of this comment accurate?
If we’re explicitly dealing with the situation because people regular-ish-ly run into it with reliable steps to reproduce, saying "something has gone horribly wrong" sounds confusing to me. (All this is assuming that this code is directly effected by problem)

Lain-B: I'm just going to resolve this for now. We can sort out the details later if need be.

PR #9442, opened 2023-08-15
frontend, plugins: Set default bitrate to 6000 kbps

notr1ch:
### Description
Increases the default bitrate to 10000kbps (10mbps). For encoders that have a different `bitrate` / `max_bitrate`, use 8000 / 10000.

### Motivation and Context
The default of 2500 kbps was chosen 10 years ago and times have changed. Logs and forums posts show that a lot of users who use OBS for recording don't touch the default settings and end up with a 2.5 mbps recording which looks terrible.

With service bitrate enforcement, this will be automatically capped to the maximum bitrate for streaming services, so the only time this should result in a problem is if the user's upload speed is the limiting factor, hopefully rarer these days with increasing internet speeds.

### How Has This Been Tested?
Ran OBS.

### Types of changes

### Checklist:

### DIFF
--- a/frontend/widgets/OBSBasic.cpp
+++ b/frontend/widgets/OBSBasic.cpp
@@ -719,7 +719,7 @@ bool OBSBasic::InitBasicConfigDefaults()
 
 	config_set_default_string(activeConfiguration, "SimpleOutput", "FilePath", GetDefaultVideoSavePath().c_str());
 	config_set_default_string(activeConfiguration, "SimpleOutput", "RecFormat2", DEFAULT_CONTAINER);
-	config_set_default_uint(activeConfiguration, "SimpleOutput", "VBitrate", 2500);
+	config_set_default_uint(activeConfiguration, "SimpleOutput", "VBitrate", 6000);
 	config_set_default_uint(activeConfiguration, "SimpleOutput", "ABitrate", 160);
 	config_set_default_bool(activeConfiguration, "SimpleOutput", "UseAdvanced", false);
 	config_set_default_string(activeConfiguration, "SimpleOutput", "Preset", "veryfast");
@@ -751,7 +751,7 @@ bool OBSBasic::InitBasicConfigDefaults()
 	config_set_default_bool(activeConfiguration, "AdvOut", "FFOutputToFile", true);
 	config_set_default_string(activeConfiguration, "AdvOut", "FFFilePath", GetDefaultVideoSavePath().c_str());
 	config_set_default_string(activeConfiguration, "AdvOut", "FFExtension", "mp4");
-	config_set_default_uint(activeConfiguration, "AdvOut", "FFVBitrate", 2500);
+	config_set_default_uint(activeConfiguration, "AdvOut", "FFVBitrate", 6000);
 	config_set_default_uint(activeConfiguration, "AdvOut", "FFVGOPSize", 250);
 	config_set_default_bool(activeConfiguration, "AdvOut", "FFUseRescale", false);
 	config_set_default_bool(activeConfiguration, "AdvOut", "FFIgnoreCompat", false);
--- a/plugins/obs-ffmpeg/obs-ffmpeg-av1.c
+++ b/plugins/obs-ffmpeg/obs-ffmpeg-av1.c
@@ -229,7 +229,7 @@ static bool av1_encode(void *data, struct encoder_frame *frame, struct encoder_p
 
 void av1_defaults(obs_data_t *settings)
 {
-	obs_data_set_default_int(settings, "bitrate", 2500);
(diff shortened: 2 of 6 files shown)


derrod:
Good change.

Should we log if the "Ignore streaming service setting recommendation" box is checked? I forsee people coming into Discord going "why is my Twitch stream not working OBS sucks" because they checked that and the default of 10 Mbps means "Source" quality will be disabled (i.e. nothing at all for streams that don't get transcodes).

RytoEX:
> Should we log if the "Ignore streaming service setting recommendation" box is checked? I forsee people coming into Discord going "why is my Twitch stream not working OBS sucks" because they checked that and the default of 10 Mbps means "Source" quality will be disabled (i.e. nothing at all for streams that don't get transcodes).

Yes, we should be logging that.

I'd personally prefer a more conservative increase, such as to 5000 or 6000 Kbps, but I do agree that this default should be increased.

brunolopesdsilv:
I fully support the idea of increasing the bit rate 👍. I really know friends who made a bad stream without setting up OBS.

I only have doubts about **the number**. Until the 11th generation of Intel processors, such a high rate could easily break the coding of a low price Core i5, that I would venture to say that it has a considerable market share.

"High" rates tend to weigh a lot, however, this comment of mine is of little importance, as it depends on the sources inserted in the OBS.

Fenrirthviti:
> I fully support the idea of increasing the bit rate 👍. I really know friends who made a bad stream without setting up OBS.
> 
> I only have doubts about **the number**. Until the 11th generation of Intel processors, such a "high" rate could easily break the coding of a low price Core i5, that I would venture to say that it has a considerable market share.
> 
> "High" rates tend to weigh a lot, however, this comment of mine is of little importance, as it depends on the sources inserted in the OBS.

Bitrate has little to no performance impact, especially for hardware encoders, but that is a discussion to be had on our Discord or Forums.

brunolopesdsilv:
> but that is a discussion to be had on our Discord or Forums.

With all due respect, I don't think this is an off topic discussion. We are talking about relevant things so that the MR isn't reverted in the future.

WizardCM:
As some users may be on internet plans that have a 10mbps upload speed, this would saturate their connection if they choose to stream to YouTube or similar. I'd vote that we lower it to 9000 or 8500 to avoid the potential higher support requests. Outside of that, I think this is a good change.

basisbit:
> don't touch the default settings and end up with a 2.5 mbps recording which looks terrible
2.5 Mbit/s looks quite good if you use an encoder which has plenty of CPU/GPU time available (better compression) or when there is not much movement. If you think that 2.5 Mbit/s looks terrible for 1080p or 720p, then your encoder settings are just bad. Don't use verypoor / ultrafast and then expect good quality at low bitrate. 

I strongly suggest not setting the default so high. A higher bitrate increases the strain on the hardware used for encoding. With how often people use laptops with very low TDP, 10Mb/s will cause the system to overheat after some time. Besides that, especially for people who use wireless connections or DOCSIS based connection, an upload speed of more than a couple of Megabit/s is not reliably achieved. This additionally will increase the workload for those who provide support to users of OBS Studio.

What about using a default bitrate depending on video resolution and framerate, instead of a "one-size-fits-all" approach?
Maybe something roughly like this for 30fps:
- 150000 pixels (360p) or above -> 1Mb/s
- 900000 pixels (720p) or above -> 2.5 Mb/s
- 2000000 pixels (1080p) or above -> 5Mb/s
- and so on

If 60fps is used instead, increase those numbers by 50%.

This would be only a few lines of code, but give much more "quality-consistent" video.

derrod:
> I strongly suggest not setting the default so high. A higher bitrate increases the strain on the hardware used for encoding.

Bitrate has pretty much no impact on encoding workload.

Warchamp7:
> Good change.
> 
> Should we log if the "Ignore streaming service setting recommendation" box is checked? I forsee people coming into Discord going "why is my Twitch stream not working OBS sucks" because they checked that and the default of 10 Mbps means "Source" quality will be disabled (i.e. nothing at all for streams that don't get transcodes).

Do we default this checkbox off? This is the same concern that I have.

If this isn't going to impact people with fresh installs then it has my approval.

derrod:
> Do we default this checkbox off? This is the same concern that I have.
> 
> If this isn't going to negatively impact people with fresh installs then it has my approval.

The checkbox is on by default, and there's even a warning message when you try to uncheck it.

Warchamp7 (approved):

Warchamp7:
I would be fine with 6000. It'd still be a significant improvement over 2500.

RytoEX:
Rebased to get that done. May change defaults to 6000.

PR #9446, opened 2023-08-16
UI: Fix compiling without nlohmann JSON

Chiitoo:
### Description

This moves the inclusion of

`UI/update/models/whatsnew.hpp`

within

`#if !defined(_WIN32) && defined(WHATSNEW_ENABLED)`

to fix compiling without nlohmann JSON and `ENABLE_WHATSNEW=OFF`.

### Motivation and Context

This enables compiling without nlohmann JSON, which is used only for browser, websocket, and the whatsnew features (as far as I am aware).

### How Has This Been Tested?

Texted by compiling without nlohmann JSON installed, and the browser, websocket, and whatsnew features disabled on Gentoo Linux.

### Types of changes

### Checklist:

### DIFF
--- a/UI/window-basic-main.cpp
+++ b/UI/window-basic-main.cpp
@@ -84,6 +84,10 @@
 #include "windows.h"
 #endif
 
+#ifdef WHATSNEW_ENABLED
+#include "update/models/whatsnew.hpp"
+#endif
+
 #if !defined(_WIN32) && defined(WHATSNEW_ENABLED)
 #include "update/shared-update.hpp"
 #endif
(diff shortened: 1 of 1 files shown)


Chiitoo:
That is good to know.

Thank you!

Chiitoo:
Certainly wasn't thinking on most cylinders when staring at that 'if' line there...

As originally intended, I've added a conditional of its own for this include so that Windows gets to have it too.

UI/window-basic-main.cpp
```diff
@@ -84,6 +84,10 @@
 #include "windows.h"
 #endif
 
+#if defined(WHATSNEW_ENABLED)
```
derrod, line 87: Generally we use `#ifdef` instead of `#if defined(...)` (unless it's unavoidable).
```suggestion
#ifdef WHATSNEW_ENABLED
```

Chiitoo: D'oh, yeah, also kind of obvious from the other lines... sleepy eyes.  : ]

Thanks!

PR #9447, opened 2023-08-16
CI: Fix open issues and macOS disk image creation

PatTheMav:
### Description
Fixes open issues with CI workflows:

* Concurrency rules are enforced even for non-tagged pushes to master branch
* Checks against invalid git references fail in specific scenarios leading to false positives

Also downgrades the compression used for generated macOS disk images.

### Motivation and Context
The disk images created by the packaging script are based on APFS (which is the default as of 2023) and `lzma` compression. The latter is only supported on Mac OS X 10.15+, which leads to a cryptic error message about a corrupted disk image for users on those older systems.

Using `lzfse` is supported on Mac OS X 10.11+ (with APFS being supported on Mac OS X 10.13+) so those older versions will be able to open the disk image.

> [!NOTE] 
> Even though users can open the disk image and copy the application bundle to their Applications directory, the app will not run because it requires macOS 11+, but macOS is more explicit about that issue when launching an application vs opening the disk image.
>
> One might argue that failing earlier is better than allowing people to install the Application and then failing, but the cryptic nature of the error message for the incompatible disk image seems to create a higher support burden.

### How Has This Been Tested?
Needs testing on CI.

### Types of changes

### Checklist:

### DIFF
--- a/.github/actions/check-changes/action.yaml
+++ b/.github/actions/check-changes/action.yaml
@@ -38,7 +38,7 @@ runs:
         shopt -s extglob
         shopt -s dotglob
 
-        if ! git cat-file -e ${GIT_BASE_REF}; then
+        if ! git rev-parse --verify ${GIT_BASE_REF} &> /dev/null; then
           echo "::warning::Provided base reference ${GIT_BASE_REF} is invalid"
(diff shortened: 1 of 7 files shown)


RytoEX (approved):
Looks okay at a glance.

PR #9449, opened 2023-08-16
CI: Update checks for changed files to detect empty GHA references

PatTheMav:
### Description
Updates steps checking for changes to use `git cat-file` as the "empty" git reference passed by GitHub Actions will be considered a valid "commit-ish" hash by `git rev-parse --verify` (but not by `cat-file`).

Also updates the dispatch workflow to not create redirects to older (now non-existing) documentation pages and removes superfluous provision of the commit hash and fixes the condition for a documentation upload to always(!) happen for a pushed tag.

### Motivation and Context
Unbreak tagged pushes and checks in general.

### How Has This Been Tested?
Changes to `git cat-file` were tested locally, changes to dispatch workflow need to be tested on CI.

### Types of changes

### Checklist:

### DIFF
--- a/.github/actions/check-changes/action.yaml
+++ b/.github/actions/check-changes/action.yaml
@@ -38,7 +38,7 @@ runs:
         shopt -s extglob
         shopt -s dotglob
 
-        if ! git rev-parse --verify ${GIT_BASE_REF} &> /dev/null; then
+        if ! git cat-file -e "${GIT_BASE_REF}" &> /dev/null; then
           echo "::warning::Provided base reference ${GIT_BASE_REF} is invalid"
           if [[ "${USE_FALLBACK}" == 'true' ]]; then
             GIT_BASE_REF='HEAD~1'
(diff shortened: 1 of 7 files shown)


RytoEX:
Will merge this to fix CI.

```
When the internal git reference is nonexistent, GitHub Actions seems
to convert that value into a hash made up of only zeroes. As that
hash will be commit-ish it will pass verification.
```

To clarify a bit about `git rev-parse --verify` and this commit message, `--verify` by default only checks that the passed value can be turned into a raw 20-byte SHA-1 that _can_ be used to access the object database. By default, it does not verify that said hash exists in the object database. If you want to verify that the hash exists in the object database, you must use the `^{type}` peeling operator, such as adding `^{commit}` or `^{object}` to the end of the parameter.

Commit-ish refers to an object that is a commit or commit-like (a commit or a tag that points to a commit). When `git rev-list --verify "0000000000000000000000000000000000000000"` returns `0000000000000000000000000000000000000000` without error, it does not indicate that that hash "names an existing object that is a commit-ish", because we did not ask it to do so.
<https://git-scm.com/docs/git-rev-parse/2.41.0#Documentation/git-rev-parse.txt---verify>

RytoEX (approved):

PR #9453, opened 2023-08-17
CI: Fix Sparkle appcast release note generation

RytoEX:
### Description

The current steps attempted to remove SSH signatures from signed git commits/tags by making the release notes string into a single line, running sed, then restoring the newlines. However, this resulted in all n characters being missing from the resulting output. We can use a sed address range instead, which should handle this multiline removal for us without other actions.

### Motivation and Context

Want macOS release notes to be generated automatically without issue.

### How Has This Been Tested?

Tested the commands locally against some sample text. I have not tested this on CI.

### Types of changes

### Checklist:


.github/actions/sparkle-appcast/action.yaml
```diff
@@ -136,9 +136,7 @@ runs:
         if (( ${+RUNNER_DEBUG} )) setopt XTRACE
 
         git tag -l --format='%(contents)' ${GITHUB_REF_NAME} \
-          | tr '\n' '\\n' \
-          | sed 's/-----BEGIN SSH SIGNATURE-----.*-----END SSH SIGNATURE-----//g' \
-          | tr '\\n' '\n' > notes.rst
+          | sed '/-----BEGIN SSH SIGNATURE-----/,/-----END SSH SIGNATURE-----/d' > notes.rst
 
         sed -i '' '2i\'$'\n''###################################################' notes.rst
```
norihiro, line 141: Instead of modifying the output of `git tag`, I'd like to suggest using the format options `--format='%(contents:subject)` and `--format='%(contents:body)'` to retrieve the subject and body separately.
```suggest
git tag -l --format='%(contents:subject)' ${GITHUB_REF_NAME} > notes.rst
echo '###################################################' >> notes.rst
echo '' >> notes.rst
git tag -l --format='%(contents:body)' ${GITHUB_REF_NAME} >> notes.rst
```

PatTheMav: I agree, otherwise PGP signed tags will have that signature come up in the release notes as well. Better to explicitly extract only the fields we need/want then extract everything that can have unexpected fields or data.

RytoEX: I've used the multiline form suggested above, as I think it's more readable/decipherable.

PatTheMav (approved):

Issue #9342, opened 2023-07-31
Image Slide Show 'Randomise Playback' option loops the slideshow


biosmatrix123:
### Operating System Info

Windows 11

### OBS Studio Version

29.1.3

### OBS Studio Log URL

https://obsproject.com/logs/96avj0qWMfySH82T

### Expected Behavior

I expected the slide show to show the images in the playlist in a random order, then stop.

It seems the randomise option literally chooses an item from the playlist at random and continues indefinitely.

### Current Behavior

Using the 'randomise playback' option has the same effect as the 'Loop' option (though images from the playlist are shown randomly).

### Steps to Reproduce

1. Add the image slide show source and configure it to look at a folder of images
2. Enable the 'Randomise Playback' option and ensure 'Loop' is not enabled
3. After the slide show starts, it will never stop / finish

### Anything else we should know?

Could this be improved so that a playlist is built in a random order and 'Randomise Playback' option becomes 'Play the images in a random order'. There is no repeat protection currently, so random really is random and could in theory end up showing the same images repeatedly.

derrod:
Works as intended. Frature requests go on https://ideas.obsproject.com/

biosmatrix123:
This is working as intended? What does the loop function do if 'Randomise Playback' is in use? Seems a strange way of labeling things to me. I won't submit a feature request as I doubt it will get any traction but thanks anyway.

PR #9471, opened 2023-08-18
image-source: Move checkboxes to playback mode setting

derrod:
### Description

Changes image slideshow properties to use a combobox (displayed as radio buttons) for the playback mode to make it more obvious that "randomize" and "loop" are mutually exclusive.

screenshit

### Motivation and Context

Users get confused, e.g. #9342

### How Has This Been Tested?

Checked that settings get set and migrated correctly.

### Types of changes

### Checklist:


GeorgesStavracas (commented):
UI / UX review:

 * Rename "Random" to "Shuffle"
 * Use a combo box

derrod:
Also needs to be updated now that #10088

derrod:
Updated for Mk.2

cg2121:
I still have a couple of problems with it.
- Shuffle and random are two completely different things
- Shuffling is reordering the list of images in a random order, and they can only be shown once when going through the list of images
- Random is picking a image randomly out of the list, and in theory the same image can be twice in a row
- It should be like the VLC source where it uses a shuffle and has the ability to loop independent of that
- The slideshow source currently uses random selection and not shuffle

derrod:
Renamed back to "Random" because that's what it is.

Jacien:
Out of curiosity, the playback mode "Random" no longer loops a slideshow, it just plays it once and ends.  Curious if this is related to it.

Fenrirthviti:
Unrelated, that was a known issue that has a pending fix already: https://github.com/obsproject/obs-studio/pull/11629

PR #9474, opened 2023-08-19
obs-qsv11: Fix crash in QSV test process

notr1ch:
### Description
The code added in #9142 assumed `MFXCreateSession` would always succeed, but it fails on systems with no QSV implementations, causing a crash when we call `MFXClose` later. Additional success checks for the other API calls were also added.

### Motivation and Context
Crashes are bad. Even if users don't see them happening, this can trigger WER to collect dump files so can impact system performance on every OBS launch.

### How Has This Been Tested?
Only tested on my system without any QSV capable devices, but should be fine.

### Types of changes

### Checklist:


RytoEX (approved):
Looks fine at a glance.

PR #9482, opened 2023-08-20
docs: Add version added to filter index functions

cg2121:
### Description
These functions were added with OBS 30.0.0

### Motivation and Context
So developers know what version these functions were added

### How Has This Been Tested?
No testing really needed here

### Types of changes

### Checklist:

### DIFF
--- a/docs/sphinx/reference-sources.rst
+++ b/docs/sphinx/reference-sources.rst
@@ -1732,6 +1732,8 @@ Filters
 
    :param index: | The index to move the filter to.
 
+   .. versionadded:: 30.0
+
 ---------------------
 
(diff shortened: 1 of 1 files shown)


gxalpha:
Hmm, in existing documentation some version numbers only reference major.minor (29.1), while some have major.minor.patch (29.0.0).
Personally I’d go for the former as in theory the patch version shouldn’t matter, APIs usually shouldn’t be introduced in patches. I feel we should at least decide on one style for new additions (and then cleanups can happen later).

cg2121:
Updated it to 30.0 instead of 30.0.0

PR #9491, opened 2023-08-21
CI: Move language file upload to nightly runs and remove master push concurrency limits

PatTheMav:
### Description
Moves the upload of language files to the scheduled nightly workflow run to upload changes in bulk and removes concurrency limits on the push workflow.

### Motivation and Context
Current concurrency limits on the push workflow seem to prohibit GitHub Actions from _starting_ new workflow runs if there is an active one (even when canceling current runs is disabled) which means that desired actions are not run.

Language file upload is moved to nightly runs nevertheless because it might make more sense for translators to get changes in bulk as they will probably not look into untranslated strings after every push anyway.

### How Has This Been Tested?
Needs to be tested on CI with a nightly run and consecutive pushes to the master branch.

### Types of changes

### Checklist:

### DIFF
--- a/.github/workflows/push.yaml
+++ b/.github/workflows/push.yaml
@@ -11,9 +11,6 @@ on:
       - '*'
 permissions:
   contents: write
-concurrency:
-  group: '${{ github.workflow }} @ ${{ github.ref }}'
-  cancel-in-progress: false
 jobs:
   check-format:
     name: Format 🔍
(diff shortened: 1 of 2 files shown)


RytoEX (approved):
Seems okay at a glance. Would welcome additional review.

RytoEX:
@derrod approved off-thread. Merging.

PR #9496, opened 2023-08-21
rtmp-services: Use RTMPS for Twitch if available

derrod:
### Description

Switch Twitch to use RTMPS if protocol registered in OBS.

### Motivation and Context

Twitch recommends RTMPS now but apparently that got lost in translation somewhere.

### How Has This Been Tested?

Hasn't.

### Types of changes

### Checklist:


plugins/rtmp-services/service-specific/twitch.c
```diff
@@ -54,10 +55,13 @@ static bool load_ingests(const char *json, bool write_file)
...
 
+	rtmps_available = obs_is_output_protocol_registered("RTMPS");
 	for (size_t i = 0; i < count; i++) {
 		json_t *item = json_array_get(ingests, i);
 		json_t *item_name = json_object_get(item, "name");
-		json_t *item_url = json_object_get(item, "url_template");
+		json_t *item_url = json_object_get(
+			item, rtmps_available ? "url_template_secure"
+					      : "url_template");
 		struct ingest ingest = {0};
```
notr1ch, line 65: I think we should also check `url_template_secure` exists in the JSON in case an ingest has TLS support removed for whatever reason.

notr1ch:
We may want to expose this as an option in case the user is behind a restrictive HTTPS-intercepting network that chokes when it sees RTMP over port 443. RTMPS also disables Network Optimizations / TCP Pacing that some users might rely on.

plugins/rtmp-services/service-specific/twitch.c
derrod: Sure, easy enough.

derrod:
> We may want to expose this as an option in case the user is behind a restrictive HTTPS-intercepting network that chokes when it sees RTMP over port 443. RTMPS also disables Network Optimizations / TCP Pacing that some users might rely on.

Mh, I don't want to give @Warchamp7 another aneurysm with yet another checkbox, but the only other way I can really think of for doing this is to do what we did with YouTube and have separate "RTMP" and "RTMPS" service (or now server) entries, but that would of course result in a lot of duplication either way.

PR #9501, opened 2023-08-22
docs: Add precision on editable string lists

Penwy:
### Description

This adds a precision on the way editable string properties lists work.

### Motivation and Context

The doc states that obs_data_get_string() returns the val parameter of a string list item, which is not true of editable lists, causing potential confusion.

### How Has This Been Tested?

Built the doc.
Looked at the doc.

### Types of changes

### Checklist:

### DIFF
--- a/docs/sphinx/reference-properties.rst
+++ b/docs/sphinx/reference-properties.rst
@@ -664,6 +664,8 @@ Property Modification Functions
    :param    val:  The actual string value stored and will be returned by :c:func:`obs_data_get_string`
    :returns: The index of the list item.
 
+   `Note` : If ``p`` is of type OBS_COMBO_TYPE_EDITABLE, :c:func:`obs_data_get_string` will return ``name`` instead of ``val``.
+
 ---------------------
 
 .. function:: size_t obs_property_list_add_int(obs_property_t *p, const char *name, long long val)
@@ -693,6 +695,8 @@ Property Modification Functions
    :param    idx:  The index of the list item
    :param    name: Localized name shown to user
    :param    val:  The actual string value stored and will be returned by :c:func:`obs_data_get_string`
+
+   `Note` : If ``p`` is of type OBS_COMBO_TYPE_EDITABLE, :c:func:`obs_data_get_string` will return ``name`` instead of ``val``.
 
 ---------------------
 
--- a/docs/sphinx/reference-properties.rst
+++ b/docs/sphinx/reference-properties.rst
@@ -741,6 +741,9 @@ Property Modification Functions
    :param    val:  The actual string value stored and will be returned by :c:func:`obs_data_get_string`
    :returns: The index of the list item.
 
+   Note: If ``p`` is of type OBS_COMBO_TYPE_EDITABLE, :c:func:`obs_data_get_string` will return ``name`` instead of
+   ``val``.
+
 ---------------------
 
 .. function:: size_t obs_property_list_add_int(obs_property_t *p, const char *name, long long val)
@@ -770,6 +773,9 @@ Property Modification Functions
    :param    idx:  The index of the list item
    :param    name: Localized name shown to user
    :param    val:  The actual string value stored and will be returned by :c:func:`obs_data_get_string`
+
+   Note: If ``p`` is of type OBS_COMBO_TYPE_EDITABLE, :c:func:`obs_data_get_string` will return ``name`` instead of
+   ``val``.
 
 ---------------------
 
--- a/docs/sphinx/reference-settings.rst
+++ b/docs/sphinx/reference-settings.rst
@@ -164,6 +164,8 @@ Get Functions
 
 .. function:: const char *obs_data_get_string(obs_data_t *data, const char *name)
 
+   `Note` : If the data object was generated from a OBS_COMBO_TYPE_EDITABLE property, the property's ``name`` will be returned instead of its ``val``.
+
 ---------------------
 
 .. function:: long long obs_data_get_int(obs_data_t *data, const char *name)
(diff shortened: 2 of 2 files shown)


gxalpha:
Could you give an example of what exactly happens here? As in, which calls return what with every other type of list and what happens with editable lists?
I saw this being discussed on Discord and from what I understand on first glance, this seems more like a bug to me.

Penwy:
So when adding an item to a string list with `obs_property_list_add_string`, you give it two arguments, `name` which is what will get displayed when the user views the list, and `val` which is the "actual" value of the item. 
When an item is selected through the UI the corresponding item in the data settings object associated to it then gets set to `val`, and calling `obs_data_get_string` on it to get the item selected by the user returns `val`. That's also the behaviour of float and int lists.

But in the case of editable lists, the value set by the UI in the data settings object, and as such, what `obs_data_get_string` returns is `name`, not `val`.

I do perfectly agree that it can be argued to be a bug, but "fixing" it will presumably break a lot, it has been like that, as far as I can see for as long as editable lists have existed.
If you want I can create an issue, to have a tracked discussion on whether or not it should be fixed.

gxalpha:
Oh I understand what's going on, I'd call it expected, so this PR is probably correct. 
Personally I'd probably put the clarification in the `:param val:` line though, maybe as a note that the parameter is essentially unused?

Penwy:
I considered that, but found it a bit too bulky on trying it. Can change back to that if you so wish.

RytoEX:
Do we use `N.B.` anywhere else in the docs? Does Sphinx recognize this and use a special highlight/rendering? How does this note look when rendered into HTML docs?

Does `obs_property_list_insert_string` also need a note? Does `obs_data_get_string`?

The language is "good enough" as-is, but I personally prefer something like this:
> If `p` is of type OBS_COMBO_TYPE_EDITABLE, :c:func:`obs_data_get_string` will return `name` instead of `val`.

Highlight/Format as needed to make it look good in the rendered docs. Word wrap text as needed.

Penwy:
This is hat the current version looks like when rendered.
Selection_1425

N. B. is not used anywhere else and sphinx does not recognise it as something special. I couldn't find any similar equivalent of a "footnote" either in the existing docs or in sphinx itself, hence why I used that.

`obs_property_list_insert_string` does need a note, you're quite right, imo `obs_data_get_string` does not.

PatTheMav:
Shouldn't this notice be placed with `obs_data_get_string`? Because it is an implementation detail of that function that behaves differently given the property type.

This detail is entirely irrelevant to either `obs_property_list_add_string` or `obs_property_list_insert_string` and this quirk also does not impact the user of either function at all.

Given that `obs_data_get_string` is referenced, an interested user would click on that to look up information about that function and it is there that this notice should appear IMO.

Penwy:
My reasoning with this is that: 
- This information is relevant when setting up the property, since that's when those values are defined. A priori, the point of such a property is to be used, so, knowing what actually gets returned when it's used is very relevant to setting it up.
- `obs_data_get_string` is indeed referenced, but in a sentence that states that it will return `val`, so there would be no reason for any user to assume the behaviour is different and click on the reference to check.

If you still think that this information is better in `obs_data_get_string`, I will change the PR, but I do disagree here.

PatTheMav:
> My reasoning with this is that:
> 
> * This information is relevant when setting up the property, since that's when those values are defined. A priori, the point of such a property is to be used, so, knowing what actually gets returned when it's used is very relevant to setting it up.
> * `obs_data_get_string` is indeed referenced, but in a sentence that states that it will return `val`, so there would be no reason for any user to assume the behaviour is different and click on the reference to check.
> 
> If you still think that this information is better in `obs_data_get_string`, I will change the PR, but I do disagree here.

If I implement `obs_data_get_string` and look up its documentation, I expect it to tell me what it returns. It does not make sense to hide that information somewhere else.

If I implement `obs_property_list_add_string`, I don't necessarily also implement `obs_data_get_string`, so putting the information there just adds non-relevant information that has no value to me.

The fact that `obs_data_get_string` behaves differently is an implementation detail of _that_ function, not of the functions adding or inserting strings to a list.

PatTheMav:
@WizardCM Looks good to me, wdyt?

WizardCM (approved):

docs/sphinx/reference-properties.rst
```diff
@@ -664,6 +664,8 @@ Property Modification Functions
    :param    val:  The actual string value stored and will be returned by :c:func:`obs_data_get_string`
    :returns: The index of the list item.
 
+   `N.B.` : If ``p`` is of type OBS_COMBO_TYPE_EDITABLE, :c:func:`obs_data_get_string` will return ``name`` instead of ``val``.
```
RytoEX, line 667: Please use "Note: " instead of "N.B.: ". If "Note: " doesn't fit well, then just omit it entirely.

See here and throughout.

RytoEX (changes requested):

docs/sphinx/reference-properties.rst
```diff
@@ -664,6 +664,8 @@ Property Modification Functions
    :param    val:  The actual string value stored and will be returned by :c:func:`obs_data_get_string`
    :returns: The index of the list item.
 
+   `Note` : If ``p`` is of type OBS_COMBO_TYPE_EDITABLE, :c:func:`obs_data_get_string` will return ``name`` instead of ``val``.
```
RytoEX, line 667: Does "Note" need backticks? Is it just to provide highlight?

Penwy: It is just for highlight. Here's comparative screenshots. Do tell if you'd rather I remove the backticks.
Selection_1853

Selection_1854

RytoEX: I changed these to just "Note:" which aligns with other instances of such text in the docs.

RytoEX (approved):

RytoEX:
Fixed some text wrapping. LGTM

PR #9503, opened 2023-08-23
obs-qsv11: Fix unusual CBR bitrate issues

### LINKED ISSUE #9460: QSV CBR AVC Bug unusable for streaming
### Operating System Info

Windows 11

### Other OS

_No response_

### OBS Studio Version

30.0.0-beta1

29.1.3

### OBS Studio Log URL

https://obsproject.com/logs/0gd5UhqSh9CyR6J3

### OBS Studio Crash Log URL

_No response_

### Expected Behavior

CBR Spikes should still respect the set keyframe interval and should fall back to CBR target bitrate (in this example 6000+audio bitrate) over the course of 2 seconds.

### Current Behavior

With CBR set to 6000 Kbps bitrate spikes up pretty high on complex scenes 10500 kbps and **stays in the range of 8700-8900 kbps far longer than the set keyframe interval**. Also during those moments you can clearly see that the whole picture starts pulsating and creates additional picture blockiness.
If a person will try to stream to twitch the result of this kind of behaviour will lead to disabled video player on twitch if no transcodes were available. If transcodes were available people will only see those.

### Steps to Reproduce

1. Start recording
2. Switch to a scene with visually complex gameplay
3.
...

Video example of the issue with stats window
https://drive.google.com/file/d/19OqxZ5X6V3JruBDPVdJEUC_6uBybr7Go/view?usp=sharing

Settings used 
image

This issue was already reported in #9258 and RytoEX sent the reported issue to Intel.

RytoEX:
### Description

For some reason, using MFX_SCENARIO_GAME_STREAMING causes the keyframe quants to be higher than other frames, which is not desirable. In turn, this causes bitrate to be higher than the target bitrate for a sustained period of time after each keyframe.

Not setting the scenario removes this behavior and returns CBR to somewhat reasonable levels of consistency.

<details>
<summary>Initial approach no longer used</summary>

obs-qsv11: Adjust CBR parameters to reduce bitrate variance

For CBR, set MaxKbps to be the same as TargetKbps, and reduce the multiplier for BufferSizeInKB from 2 to 1. This should reduce the bitrate variance while using CBR.

obs-qsv11: Adjust CBR parameters to reduce initial bitrate spike

For CBR, reduce the InitialDelayInKB multiplier from 1 to 0.5. This should reduce the initial bitrate spike at the beginning of a CBR encoding session.
</details>

### Motivation and Context

Fixes #9460.

Want to address target bitrate issues with QSV CBR as much as we are able to do so from within OBS.

### How Has This Been Tested?

Tested locally on Windows 11 with an Intel Arc A770M. @flaeri checked the output of files before and after these changes and confirmed a noted improvement.

### Types of changes

### Checklist:

### DIFF
--- a/plugins/obs-qsv11/QSV_Encoder_Internal.cpp
+++ b/plugins/obs-qsv11/QSV_Encoder_Internal.cpp
@@ -345,15 +345,10 @@ mfxStatus QSV_Encoder_Internal::InitParams(qsv_param_t *pParams,
 		if (pParams->nLADEPTH &&
 		    m_mfxEncParams.mfx.LowPower == MFX_CODINGOPTION_ON) {
 			m_co2.LookAheadDepth = pParams->nLADEPTH;
-			m_co3.ScenarioInfo = MFX_SCENARIO_GAME_STREAMING;
 		}
 		// CQM to follow UI setting
 		if (pParams->bCQM && !pParams->bRepeatHeaders) {
 			m_co3.AdaptiveCQM = MFX_CODINGOPTION_ON;
-			if (m_co3.ScenarioInfo != MFX_SCENARIO_GAME_STREAMING) {
-				m_co3.ScenarioInfo =
-					MFX_SCENARIO_GAME_STREAMING;
-			}
 		} else {
 			m_co3.AdaptiveCQM = MFX_CODINGOPTION_OFF;
 		}


maginhart:
This PR still doesn't fix the bitrate spike getting stuck over target bitrate longer than keyframe interval. Reduction of the multiplier for BufferSizeInKB is detrimental to quality.

I believe related issue was fixed in recent PR in oneVPL github.
https://github.com/oneapi-src/oneVPL-intel-gpu/commit/249c854056273b88dd0417f97fff12a548c59c63

My test results video+log
https://drive.google.com/file/d/1LD75E_QT1usLU5Vqm2ieGhplKgvBOvtx/view?usp=sharing
https://obsproject.com/logs/5VDxM9QpfTwUvV5L

ma3uk:
@RytoEX You need to add initialization of the mfxCodingOption buffer and add the following parameters for CBR: 
m_co.VuiVclHrdParameters = MFX_CODINGOPTION_ON;
m_co.VuiNalHrdParameters = MFX_CODINGOPTION_ON;
m_co.NalHrdConformance = MFX_CODINGOPTION_ON;
This should solve the problem described by maginhart

flaeri:
> @RytoEX You need to add initialization of the mfxCodingOption buffer and add the following parameters for CBR: m_co.VuiVclHrdParameters = MFX_CODINGOPTION_ON; m_co.VuiNalHrdParameters = MFX_CODINGOPTION_ON; m_co.NalHrdConformance = MFX_CODINGOPTION_ON; This should solve the problem described by maginhart

Already tried it. It does not solve their issue. target 6000 kbps still hovers 7000 kbps after 2 or more static scenes.

maginhart:
https://github.com/obsproject/obs-studio/pull/9258/commits/4f338442233524498549bc13597bbf83ab8db02b

image
I wonder what was the reason for MFX_GOP_STRICT removal?

Because after addition of that PR I started seeing that problem with bitrate hovering.

RytoEX:
> @RytoEX [4f33844](https://github.com/obsproject/obs-studio/commit/4f338442233524498549bc13597bbf83ab8db02b)
> 
> image I wonder what was the reason for MFX_GOP_STRICT removal?
> 
> Because after addition of that PR I started seeing that problem with bitrate hovering.

If I recall correctly, removing it allows adaptive-B/I frames. I tested adding this back in very early in working on this PR, and I believe it didn't make a difference to the problem you were describing, but I can go back and investigate again.

maginhart:
Maybe as a test you should try adding both MFX_GOP_STRICT and NalHrdConformance & VuiNalHrdParameters

flaeri:
thanks to ma3uk and maginhart I got a plugin that did not have the problem, and by checking the differences between it and the current OBS code, and experimenting a little bit, I thought I found the core problem.

All that seemed to be required is disabling `mbbrc`. However, when I went back to the OBS implementation, I could no longer reproduce the original issue. It seems to now work perfectly fine even with mbbrc enabled. So I have to wonder if there may have been some sort of fix in the drivers at some point.

I tried the proper beta 2 build, both via steam and the zip from the OBS github, neither of which had the problem of the bitrate not returning to target. So while not really relating to the code in this PR, I think it might be fine to just omit/close it, and keep the code as is.

The current state has a bit of a bitrate spike, of course, but so do most hardware encoders. It returns to target pretty quickly and stays decently close while there is motion. I think its fine as is.

maginhart:
I still have hovering problem on obs 30.0 beta 2 with latest Intel Arc Driver

maginhart:
And this PR https://github.com/obsproject/obs-studio/pull/9258 removed MBBRC from OBS implementation and left only AdaptiveCQM

maginhart:
https://drive.google.com/file/d/1zXXMCtd_BVFUkJLgPdtUs4apEUoOjLMU/view?usp=sharing
https://obsproject.com/logs/j0uhGjH0rP9An45G
Video and log

RytoEX:
I have force pushed a completely different approach to this PR. This change is specifically only to address the keyframe pumping and the sustained high bitrate. It does not prevent very short temporary bitrate spikes that you might observe at the start of an encode session or at drastic scene changes.

maginhart:
This PR change fixed the keyframe problem i can't reproduce it anymore but removing game_streaming scenario seems to disable lookahead because number of surfaces now is 12 before the change it was 72 with normal latency which sets lookahead at 60.
image

RytoEX:
> This PR change fixed the keyframe problem i can't reproduce it anymore but removing game_streaming scenario seems to disable lookahead because number of surfaces now is 12 before the change it was 72 with normal latency which sets lookahead at 60.

Yes, I believe that is the expected outcome.

PR #9505, opened 2023-08-23
UI: Increase minimum output resolution to 32x32

### LINKED ISSUE #8934: Recording fails/hangs with older intel QSV H.264 using CQP or AVBR at <= 17x17 output resolution
### Operating System Info

Windows 10

### Other OS

_No response_

### OBS Studio Version

29.1.1

_No response_

### OBS Studio Log URL

https://obsproject.com/logs/5IBTa02tvaLSCkJG

### OBS Studio Crash Log URL

_No response_

### Expected Behavior

Either the recording works fully, or else OBS shows sort of warning or error message, and/or doesn't partly hang

### Current Behavior

I get `[qsv encoder: 'simple_video_recording'] Failed to reconfigure` in the logs right when I press "Start Recording".

Also: The recording never saves (no file is written out to disk for the recording session), and if you try to stop the recording, the button just says "Stopping Recording..." forever, the Recording ("REC") icon stays red and the recording length keeps incrementing in the bottom bar, and if you try to "Exit", regardless of trying to stop the recording or not, the visible window does close, but the OBS process keeps running in the background instead of fully exiting.

### Steps to Reproduce

1. Under "Video", set the output resolution to 17x17 or smaller (regardless of whether scaled or matching canvas)
2. Under "Output", set the recording encoder to QuickSync H.264
3. Set the "Quality" or "Rate Control" to something that uses CQP or AVBR
4. Press the "Start Recording" button

_This seems like a separate issue,_ but whenever I try to output to exactly 18x18 or 19x19 with QSV, regardless of CBR or CQP whatever (unlike the main issue I am reporting), I get a popup:

Window title: "Recording error" Message: "An encoder error occurred while recording." Button: "[OK]" (popup dialog)

RytoEX:
### Description

Increase the minimum required output resolution from 8x8 to 32x32.

Some hardware encoders fail in indecipherable ways if you try to at certain low resolutions such as 17x17 or 19x19, but works at 20x20. However, there is no good way to determine what the minimum working resolution is for every possible encoder without running extensive capabilities tests locally and storing the results. There isn't that much difference between 8x8 and 32x32, except that the latter will reduce or eliminate such odd failure cases, so let's increase the minimum required output resolution by just a little bit.

### Motivation and Context

Fixes #8934.

@Warchamp7 agreed with this in concept off-thread.

### How Has This Been Tested?

Built and tested locally on Windows 11. This uses a pre-existing mechanism for the resolution limit. The mechanism itself does not prevent a user from entering an output resolution below 32x32, but it will warn them and silently reject the change if Apply/OK are clicked anyway.

### Types of changes

### Checklist:

### DIFF
--- a/UI/window-basic-main.cpp
+++ b/UI/window-basic-main.cpp
@@ -4742,14 +4742,14 @@ int OBSBasic::ResetVideo()
 	ovi.gpu_conversion = true;
 	ovi.scale_type = GetScaleType(basicConfig);
 
-	if (ovi.base_width < 8 || ovi.base_height < 8) {
+	if (ovi.base_width < 32 || ovi.base_height < 32) {
 		ovi.base_width = 1920;
 		ovi.base_height = 1080;
 		config_set_uint(basicConfig, "Video", "BaseCX", 1920);
 		config_set_uint(basicConfig, "Video", "BaseCY", 1080);
 	}
 
-	if (ovi.output_width < 8 || ovi.output_height < 8) {
+	if (ovi.output_width < 32 || ovi.output_height < 32) {
 		ovi.output_width = ovi.base_width;
 		ovi.output_height = ovi.base_height;
 		config_set_uint(basicConfig, "Video", "OutputCX",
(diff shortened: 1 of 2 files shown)


DeeDeeG:
As the person who posted the initial issue, thanks for this fix! (Or at least as a solid workaround! Prevents the affected machines from my issue from getting stuck with a crashed or hung encoder/recording, that's what counts!) 👍

PR #9507, opened 2023-08-24
cmake: Restrict FindCEF path to omit default paths on Windows

RytoEX:
### Description

On Windows, if CEF libraries were present in default paths, they would be picked up in this find_program call, and it could later cause a build failure in the CMake custom copy commands on install if the components we are trying to copy do not exist. I recently ran into this after installing Intel PresentMon, which included CEF components, and was picked up by this find_program call. Let's apply NO_DEFAULT_PATH here the same as the find_library calls in this finder.

### Motivation and Context

Want OBS to build even when the dev has other software installed.

<https://cmake.org/cmake/help/v3.25/command/find_program.html>

### How Has This Been Tested?

Built OBS locally on Windows 11 with Intel PresentMon installed. Verified that the build failed before this PR, but succeeded after this PR. Manually inspected the paths picked up by CMake in both.

### Types of changes

### Checklist:

### DIFF
--- a/cmake/finders/FindCEF.cmake
+++ b/cmake/finders/FindCEF.cmake
@@ -98,6 +98,7 @@ if(CMAKE_HOST_SYSTEM_NAME STREQUAL Windows)
   find_program(
     CEF_LIBRARY_RELEASE
     NAMES cef.dll libcef.dll
+    NO_DEFAULT_PATH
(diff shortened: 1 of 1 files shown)


PatTheMav (approved):
Should be good because we always require our customised version of CEF and if there ever would be a system-wide variant installed, it's also not one we would want to use.

PR #9508, opened 2023-08-24
UI: Hide YouTube dock when switching profiles

joelgerard:
### Description
Follow the same behaviour as the YouTube chat window for hiding the YouTube Live Control Panel, e.g. when switching to a profile that doesn't use YT.  See https://pasteboard.co/gmNkucqueZQQ.png for which docks I mean.

### Motivation and Context
Feedback from 30 beta. Users were annoyed at having to manually close the YT dock when switching to a non-YT profile, e.g. Twitch.

### How Has This Been Tested?
Tested manually on  windows.

### Types of changes

### Checklist:

### DIFF
--- a/UI/window-basic-main-profiles.cpp
+++ b/UI/window-basic-main-profiles.cpp
@@ -784,13 +784,8 @@ void OBSBasic::ChangeProfile()
 	DestroyPanelCookieManager();
 
 #ifdef YOUTUBE_ENABLED
-	OBSBasic *main = OBSBasic::Get();
-	// This dock works with both stream key and auth
-	// so hide it here if required.
-	YouTubeAppDock *youtubeAppDock = main->GetYouTubeAppDock();
-	if (youtubeAppDock != NULL) {
+	if (youtubeAppDock)
 		youtubeAppDock->hide();
-	}
 #endif
 
 	config.Swap(basicConfig);
--- a/UI/window-basic-main-profiles.cpp
+++ b/UI/window-basic-main-profiles.cpp
@@ -783,6 +783,16 @@ void OBSBasic::ChangeProfile()
 	auth.reset();
 	DestroyPanelCookieManager();
 
+#ifdef YOUTUBE_ENABLED
+	OBSBasic *main = OBSBasic::Get();
+	// This dock works with both stream key and auth
+	// so hide it here if required.
+	YouTubeAppDock *youtubeAppDock = main->GetYouTubeAppDock();
+	if (youtubeAppDock != NULL) {
+		youtubeAppDock->hide();
+	}
+#endif
+
 	config.Swap(basicConfig);
 	InitBasicConfigDefaults();
 	InitBasicConfigDefaults2();
@@ -813,6 +823,16 @@ void OBSBasic::ChangeProfile()
 			close();
 		}
 	}
+
+#ifdef YOUTUBE_ENABLED
+	if (YouTubeAppDock::IsYTServiceSelected()) {
+		// This is re-initialized in some cases.
+		youtubeAppDock = main->GetYouTubeAppDock();
+		if (youtubeAppDock != NULL) {
+			youtubeAppDock->show();
+		}
+	}
+#endif
 }
 
 void OBSBasic::CheckForSimpleModeX264Fallback()
--- a/UI/window-basic-main-profiles.cpp
+++ b/UI/window-basic-main-profiles.cpp
@@ -820,13 +820,8 @@ void OBSBasic::ChangeProfile()
 	}
 
 #ifdef YOUTUBE_ENABLED
-	if (YouTubeAppDock::IsYTServiceSelected()) {
-		// This is re-initialized in some cases.
-		youtubeAppDock = main->GetYouTubeAppDock();
-		if (youtubeAppDock != NULL) {
-			youtubeAppDock->show();
-		}
-	}
+	if (YouTubeAppDock::IsYTServiceSelected() && youtubeAppDock)
+		youtubeAppDock->show();
 #endif
 }
 


joelgerard:
Been a while since I was in git. This work?

RytoEX:
> Been a while since I was in git. This work?

Use rebase and force-push instead of merge commits.

* YT -> YouTube (this is more searchable)
* behaviour -> behavior
* e.g. -> e.g., (US English uses a comma here, though I usually avoid "e.g." and "i.e." outside of parentheticals)

You can probably drop the word "appropriately" from the commit message subject if it is too long, or use a shorter equivalent like "as needed" to get as close to the 50 character subject length guideline as possible.

joelgerard:
Thanks. Updated. 



> > Been a while since I was in git. This work?
> 
> Use rebase and force-push instead of merge commits.
> 
> * YT -> YouTube (this is more searchable)
> * behaviour -> behavior
> * e.g. -> e.g., (US English uses a comma here, though I usually avoid "e.g." and "i.e." outside of parentheticals)
> 
> You can probably drop the word "appropriately" from the commit message subject if it is too long, or use a shorter equivalent like "as needed" to get as close to the 50 character subject length guideline as possible.

WizardCM:
Unfortunately the new YouTube Live Control Panel dock is not created exclusively by service integration, so if you populate a stream key manually then this won't clean up the dock. More likely the fix will have to go into the profile switching code itself.

joelgerard:
New to this code base. Any chance you can you point me to the right file and I will update.

RytoEX:
> New to this code base. Any chance you can you point me to the right file and I will update.

Probably somewhere in here or in a function called during profile change.
https://github.com/obsproject/obs-studio/blob/b2549e4036c8b9d1db0ea6e5cd22e47b5bc12294/UI/window-basic-main-profiles.cpp#L738

joelgerard:
> Probably somewhere in here or in a function called during profile change.
> 
> https://github.com/obsproject/obs-studio/blob/b2549e4036c8b9d1db0ea6e5cd22e47b5bc12294/UI/window-basic-main-profiles.cpp#L738


OK. Tested both "connected" and "streamkey". PTAL. Might need feedback/help on what you want me to do about the git history. I assume you squash when you approve the PR?

UI/auth-youtube.cpp
```diff
@@ -83,6 +83,7 @@ YoutubeAuth::~YoutubeAuth()
 	OBSBasic *main = OBSBasic::Get();
 
 	main->RemoveDockWidget(YOUTUBE_CHAT_DOCK_NAME);
+
```
RytoEX, line 86: Revert this change.

UI/window-basic-main-profiles.cpp
```diff
@@ -783,6 +783,16 @@ void OBSBasic::ChangeProfile()
 	auth.reset();
 	DestroyPanelCookieManager();
 
+#ifdef YOUTUBE_ENABLED
+	OBSBasic* main = OBSBasic::Get();
```
RytoEX, line 787: Place asterisk next to the variable name, not the type.

RytoEX (changes requested):

UI/auth-youtube.cpp
joelgerard: Done.

UI/window-basic-main-profiles.cpp
joelgerard: Done.

UI/window-basic-main-profiles.cpp
```diff
@@ -783,6 +783,16 @@ void OBSBasic::ChangeProfile()
...
+	// This dock works with both stream key and auth
+	// so hide it here if required.
+	YouTubeAppDock* youtubeAppDock = main->GetYouTubeAppDock();
```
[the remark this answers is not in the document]
joelgerard: Done.

UI/window-basic-main-profiles.cpp
```diff
@@ -783,6 +783,16 @@ void OBSBasic::ChangeProfile()
...
 	DestroyPanelCookieManager();
 
+#ifdef YOUTUBE_ENABLED
+	OBSBasic *main = OBSBasic::Get();
+	// This dock works with both stream key and auth
+	// so hide it here if required.
+	YouTubeAppDock *youtubeAppDock = main->GetYouTubeAppDock();
+	if (youtubeAppDock != NULL) {
+		youtubeAppDock->hide();
+	}
```
derrod, line 793: Since this method is in `OBSBasic` you can just access the app dock variable directly.
```suggestion
	if (youtubeAppDock)
		youtubeAppDock->hide();
```

UI/window-basic-main-profiles.cpp
```diff
@@ -813,6 +823,16 @@ void OBSBasic::ChangeProfile()
...
 	}
+
+#ifdef YOUTUBE_ENABLED
+	if (YouTubeAppDock::IsYTServiceSelected()) {
+		// This is re-initialized in some cases.
+		youtubeAppDock = main->GetYouTubeAppDock();
+		if (youtubeAppDock != NULL) {
+			youtubeAppDock->show();
+		}
+	}
```
derrod, line 834: Same as above.
```suggestion
	if (YouTubeAppDock::IsYTServiceSelected() && youtubeAppDock)
		youtubeAppDock->show();
```

derrod (changes requested):

UI/window-basic-main-profiles.cpp
joelgerard: I can make all these changes except for line 830. In the case where the user is logged in via a YouTube account (vs using streamkey), the reference to the dock changes and you need to recall main->GetYouTubeAppDock();

WDYT?

derrod: You are in `OBSBasic` here so `main->GetYouTubeAppDock()` and `youtubeAppDock` will point to the same object.

joelgerard: Shared repro steps here: https://youtu.be/Z23FcPHwUGA

Only crashes on YouTube connected profiles.

derrod: You need to apply both my suggested changes, otherwise you're still shadowing the member variable with the one declared in the first part of the diff, which may no longer point to a valid object.

PR #9509, opened 2023-08-25
UI: Improve UX of accessibility color settings

wooster0:
### Description

This gets rid of the "Use Different Colors" checkbox which was essentially there to allow the user to save their custom colors while using the default preset. This has the problem that it defaults to the default preset when the checkbox is not ticked which means you can't have custom colors if you enable the color blind color preset. This can be a problem for color blind people.

image

This, plus the first two commits of this PR which are unrelated to the main part of this PR described above and below and could be merged even if the main part is rejected.

### Motivation and Context

Currently, color blind people that use the Color Blind Alternative color preset are not able to retain their custom color preset because turning off "Use Different Colors" defaults to the Default preset instead of Color Blind Alternative,
so when they apply Color Blind Alternative they lose their custom colors.

This PR tries to solve this problem while keeping it simple by:

1. Getting rid of the "Use Different Colors" button.
2. Making it so that custom colors are saved as separate configuration values long-term along with the actually active colors.
   This means the stored active colors might be the same as the stored custom colors.

Now the behavior is:

When you have the custom preset selected and change a color, you
have to hit Apply for it to save, or else if you switch before that it
will reset to what the custom colors were before.

If you start on Default or Color Blind Alternative, then select a color,
click OK, it switches to the custom preset and replaces all previous custom colors with
the colors of the previously selected non-custom preset + the one selected color included.
*However*, this is not saved if you don't apply. When not applying/saving and switching back to Default or Color
Blind Alternative and then back again to the custom preset, the custom preset is as before.

It was also suggested adding a reset button. As for this, the problem is that it would only be able to reset to either Default or Color Blind Alternative. It would probably reset to Default.
What if I want to create custom colors starting off the Color Blind Alternative colors?
Should we let the user select which non-custom preset to reset to? At this point I feel like it's just getting too complicated for no good reason and if we can keep the current simplicity we should try to do so.

As for multiple custom colors, can't we just make the custom colors be different for each OBS profile?

### How Has This Been Tested?

I built this code many times and ran and tested it on my Linux machine. Any code this could affect or the changes of this PR should be pretty much platform-independent so this should work just as fine on any other operating system OBS supports.

### Types of changes

### Checklist:


RytoEX:
The appropriate commit prefix is "UI: ".

UI/window-basic-settings-a11y.cpp
```diff
@@ -209,191 +209,36 @@ void OBSBasicSettings::on_colorPreset_currentIndexChanged(int idx)
...
-
-	mixerRedActive = color_to_int(color);
-
-	preset = COLOR_PRESET_CUSTOM;
-	bool block = ui->colorPreset->blockSignals(true);
-	ui->colorPreset->setCurrentIndex(ui->colorPreset->count() - 1);
-	ui->colorPreset->blockSignals(block);
-
-	A11yChanged();
+#define GENERATE_ON_CHOOSE_CLICKED(color_override_var, color_override_str,                         \
```
gxalpha, line 227: Let's not do this. Macros are generally hard to read, debug and maintain.
If anything, create a function and connect the buttons to it manually, maybe in the style of https://github.com/obsproject/obs-studio/blob/075f3c40207fd29fa80ffd9eb8fd1396866374d0/UI/window-basic-main.cpp#L448-L467

gxalpha (changes requested):

gxalpha:
Besides making the current selection all black in the settings if "Custom" is already selected prior to this PR (which needs to be fixed), personally I think this PR makes the workflow even more confusing than it currently is in OBS.
With this PR, "Custom" colors get saved until the point where you change to a preset and then start adjusting colors from there, at which point those adjustments will override the custom colors and put you into "Custom" mode.

That said, before any changes are made here I'd want @Warchamp7's opinion on it.

UI/window-basic-settings-a11y.cpp
```diff
@@ -230,170 +236,110 @@ void OBSBasicSettings::on_choose1_clicked()
...
+
 void OBSBasicSettings::on_choose2_clicked()
 {
 	QColor color = GetColor(
 		selectGreen,
 		QTStr("Basic.Settings.Accessibility.ColorOverrides.SelectGreen"));
-
 	if (!color.isValid())
 		return;
-
```
RytoEX, line 241: Please add these lines back. We typically avoid placing new statements directly under an un-enclosed `if` statement.

UI/window-basic-settings-a11y.cpp
```diff
@@ -230,170 +236,110 @@ void OBSBasicSettings::on_choose1_clicked()
...
 }
 
+void OBSBasicSettings::on_choose1_clicked()
+{
+	QColor color = GetColor(
+		selectRed,
+		QTStr("Basic.Settings.Accessibility.ColorOverrides.SelectRed"));
+	if (!color.isValid())
+		return;
+	selectRed = color_to_int(color);
```
RytoEX, line 246: Please unsquash this. Separate the `if` statement by a blank line on each side.

UI/window-basic-settings-a11y.cpp
```diff
@@ -230,170 +236,110 @@ void OBSBasicSettings::on_choose1_clicked()
...
 
 void OBSBasicSettings::on_choose3_clicked()
 {
 	QColor color = GetColor(
 		selectBlue,
 		QTStr("Basic.Settings.Accessibility.ColorOverrides.SelectBlue"));
-
 	if (!color.isValid())
 		return;
-
```
RytoEX, line 262: Add these lines back.

UI/window-basic-settings-a11y.cpp
```diff
@@ -230,170 +236,110 @@ void OBSBasicSettings::on_choose1_clicked()
...
 
 void OBSBasicSettings::on_choose4_clicked()
 {
 	QColor color = GetColor(
 		mixerGreen,
 		QTStr("Basic.Settings.Accessibility.ColorOverrides.MixerGreen"));
-
 	if (!color.isValid())
 		return;
-
```
RytoEX, line 283: Add these lines back.

UI/window-basic-settings-a11y.cpp
```diff
@@ -230,170 +236,110 @@ void OBSBasicSettings::on_choose1_clicked()
...
 
 void OBSBasicSettings::on_choose5_clicked()
 {
 	QColor color = GetColor(
 		mixerYellow,
 		QTStr("Basic.Settings.Accessibility.ColorOverrides.MixerYellow"));
-
 	if (!color.isValid())
 		return;
-
```
RytoEX, line 304: Add these lines back.

UI/window-basic-settings-a11y.cpp
```diff
@@ -230,170 +236,110 @@ void OBSBasicSettings::on_choose1_clicked()
...
 
 void OBSBasicSettings::on_choose6_clicked()
 {
 	QColor color = GetColor(
 		mixerRed,
 		QTStr("Basic.Settings.Accessibility.ColorOverrides.MixerRed"));
-
 	if (!color.isValid())
 		return;
-
```
RytoEX, line 325: Add these lines back.

UI/window-basic-settings-a11y.cpp
```diff
@@ -230,170 +236,110 @@ void OBSBasicSettings::on_choose1_clicked()
...
 
 void OBSBasicSettings::on_choose7_clicked()
 {
 	QColor color = GetColor(
 		mixerGreenActive,
 		QTStr("Basic.Settings.Accessibility.ColorOverrides.MixerGreenActive"));
-
 	if (!color.isValid())
 		return;
-
```
RytoEX, line 346: Add these lines back.

UI/window-basic-settings-a11y.cpp
```diff
@@ -230,170 +236,110 @@ void OBSBasicSettings::on_choose1_clicked()
...
 
 void OBSBasicSettings::on_choose8_clicked()
 {
 	QColor color = GetColor(
 		mixerYellowActive,
 		QTStr("Basic.Settings.Accessibility.ColorOverrides.MixerYellowActive"));
-
 	if (!color.isValid())
 		return;
-
```
RytoEX, line 367: Add these lines back.

UI/window-basic-settings-a11y.cpp
```diff
@@ -230,170 +236,110 @@ void OBSBasicSettings::on_choose1_clicked()
...
 
 void OBSBasicSettings::on_choose9_clicked()
 {
 	QColor color = GetColor(
 		mixerRedActive,
 		QTStr("Basic.Settings.Accessibility.ColorOverrides.MixerRedActive"));
-
 	if (!color.isValid())
 		return;
-
```
RytoEX, line 388: Add these lines back.

RytoEX (changes requested):

wooster0:
Okay so for now I've reverted the main part of this PR because the other improvements unrelated to behavior I've made are already worth having a PR for on their own. I agree that the original main part of this PR might need some more thought.

For now I've left the PR description and title as it is in case @Warchamp7 thinks it's any good. Here also the original main part of this PR for reference: https://github.com/wooster0/obs-studio/commit/9990ecf3af873b0134f8b7fd57bf881ba1fd2077

With these commits now there should be no changes to the UI.

WizardCM:
For now please update the PR description to reflect the current changes in the PR, as edit history does remain so Warchamp (when he has time) can look there manually.

UI/window-basic-settings.hpp
```diff
@@ -408,6 +407,7 @@ private slots:
 
 	void on_disableOSXVSync_clicked();
 
+	void on_choose_clicked_update();
```
gxalpha, line 410: Avoid naming functions in Qt [auto-connection](https://doc.qt.io/qt-6/qobject.html#auto-connection) style (`on_sender_signal`) if you're not using Qt auto-connection, this will make Qt complain at runtime.

wooster0: Corrected. I don't think the new name I came up with is amazing but maybe it's good enough for now.

PR #9498, opened 2023-08-22
Cleanup obs-x264 plugin

quink-black:
### Description
1. Remove unused getparam() function
2. Refactor log callback
   1. Remove limitation of message length.
   2. Improve performance by remove the local copy.
   3. Map x264 log level to obs.

### Motivation and Context
1. Remove dead code
2. Improve performance

### How Has This Been Tested?

OS: Windows 10
Test steps:
1. Enable software encoder x264
2. Set x264 options to "log=3" to see some debug message of x264
3. Run recording.

Here is part of log file:
```
11:51:20.743: [x264 encoder: 'advanced_video_recording'] preset: fast
11:51:20.743: [x264 encoder: 'advanced_video_recording'] profile: main
11:51:20.743: [x264 encoder: 'advanced_video_recording'] settings:
...

11:51:20.743: 
11:51:20.744: using SAR=1/1
11:51:20.744: 
11:51:20.744: using cpu capabilities: MMX2 SSE2Fast SSSE3 SSE4.2 AVX FMA3 BMI2 AVX2
11:51:20.744: 
11:51:20.747: profile Main, level 3.1, 4:2:0, 8-bit
11:51:20.747: 
11:51:52.232: frame I:4     Avg QP:10.98  size: 48578
11:51:52.232: 
11:51:52.232: frame P:246   Avg QP:23.91  size:  3544
11:51:52.232: 
11:51:52.232: frame B:638   Avg QP:29.95  size:   514
11:51:52.232: 
11:51:52.232: consecutive B-frames:  1.9%  4.9%  4.7% 88.4%
```

### Types of changes
* New feature (non-breaking change which adds functionality: map x264 log level to obs log level
* Performance enhancement (non-breaking change which improves efficiency
* Code cleanup

### Checklist:
- [x] My code has been run through [clang-format](https://github.com/obsproject/obs-studio/blob/master/.clang-format).
- [x] I have read the [**contributing** document](https://github.com/obsproject/obs-studio/blob/master/CONTRIBUTING.rst).
- [x] My code is not on the master branch.
- [x] The code has been tested.
- [x] All commit messages are properly formatted and commits squashed where appropriate.
- [x] I have included updates to all appropriate documentation.

PR #9516, opened 2023-08-26
plugins, UI: Remove unused static functions

norihiro:
### Description

This PR removes unused static functions under UI and plugins.

### Motivation and Context

There are some unused static functions such as `getparam()` found by #9498.

### How Has This Been Tested?

Used `lcov` on macOS and Linux (Fedora 37 and Ubuntsu 22.04) and searched white lines (not counted lines).

When removing, I also checked the function is never found by `git grep`.

Built on Fedora 37 and run it.

CI will also test the code can be built...

Note:
`session_set_prop_str` is also unused but I left the function because this is just one type of `session_set_prop_*` functions.
https://github.com/obsproject/obs-studio/blob/08f5a7bd4d62c3aae459b59a02eb927001d4fdbb/plugins/mac-videotoolbox/encoder.c#L271-L279

### Types of changes

### Checklist:

### DIFF
--- a/UI/obs-app.cpp
+++ b/UI/obs-app.cpp
@@ -2965,51 +2965,6 @@ static bool update_reconnect(ConfigFile &config)
 	return false;
 }
 
-static void convert_x264_settings(obs_data_t *data)
-{
-	bool use_bufsize = obs_data_get_bool(data, "use_bufsize");
-
-	if (use_bufsize) {
-		int buffer_size = (int)obs_data_get_int(data, "buffer_size");
-		if (buffer_size == 0)
-			obs_data_set_string(data, "rate_control", "CRF");
-	}
-}
-
-static void convert_14_2_encoder_setting(const char *encoder, const char *file)
-{
-	OBSDataAutoRelease data =
-		obs_data_create_from_json_file_safe(file, "bak");
-	obs_data_item_t *cbr_item = obs_data_item_byname(data, "cbr");
-	obs_data_item_t *rc_item = obs_data_item_byname(data, "rate_control");
-	bool modified = false;
-	bool cbr = true;
-
-	if (cbr_item) {
-		cbr = obs_data_item_get_bool(cbr_item);
-		obs_data_item_unset_user_value(cbr_item);
-
-		obs_data_set_string(data, "rate_control", cbr ? "CBR" : "VBR");
-
-		modified = true;
-	}
-
-	if (!rc_item && astrcmpi(encoder, "obs_x264") == 0) {
-		if (!cbr_item)
-			obs_data_set_string(data, "rate_control", "CBR");
-		else if (!cbr)
-			convert_x264_settings(data);
-
-		modified = true;
-	}
-
-	if (modified)
-		obs_data_save_json_safe(data, file, "tmp", "bak");
-
-	obs_data_item_release(&rc_item);
-	obs_data_item_release(&cbr_item);
-}
-
 static void convert_nvenc_h264_presets(obs_data_t *data)
 {
 	const char *preset = obs_data_get_string(data, "preset");
--- a/plugins/linux-capture/xcomposite-input.c
+++ b/plugins/linux-capture/xcomposite-input.c
@@ -360,27 +360,6 @@ void xcomp_cleanup_pixmap(Display *disp, struct xcompcap *s)
 	}
 }
 
-static enum gs_color_format gs_format_from_tex()
-{
-	GLint iformat = 0;
-	// consider GL_ARB_internalformat_query
-	glGetTexLevelParameteriv(GL_TEXTURE_2D, 0, GL_TEXTURE_INTERNAL_FORMAT,
-				 &iformat);
-
-	// These formats are known to be wrong on Intel platforms. We intentionally
-	// use swapped internal formats here to preserve historic behavior which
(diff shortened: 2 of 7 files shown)


Lain-B:
Something tells me this is going to end up being a much larger PR than it currently is

norihiro:
> Something tells me this is going to end up being a much larger PR than it currently is

I will stop this for only UI and plugins so that the amount of change does not become huge.

UI/obs-app.cpp
```diff
@@ -2965,51 +2965,6 @@ static bool update_reconnect(ConfigFile &config)
...
-	bool use_bufsize = obs_data_get_bool(data, "use_bufsize");
-
-	if (use_bufsize) {
-		int buffer_size = (int)obs_data_get_int(data, "buffer_size");
-		if (buffer_size == 0)
-			obs_data_set_string(data, "rate_control", "CRF");
-	}
-}
-
-static void convert_14_2_encoder_setting(const char *encoder, const char *file)
```
norihiro, line 2979: This function was replaced with `convert_28_1_encoder_setting` by d7c7180a8c so that this became unused.

plugins/linux-capture/xcomposite-input.c
```diff
@@ -360,27 +360,6 @@ void xcomp_cleanup_pixmap(Display *disp, struct xcompcap *s)
 	}
 }
 
-static enum gs_color_format gs_format_from_tex()
```
norihiro, line 363: This function was called by `xcomp_create_pixmap` at 1604400e48. Later, the call was removed by ddb692ab2.

plugins/mac-avcapture/av-capture.mm
```diff
@@ -903,16 +903,6 @@ static bool init_preset(av_capture *capture, AVCaptureDevice *dev, obs_data_t *s
 static bool operator==(const CMVideoDimensions &a, const CMVideoDimensions &b);
 static CMVideoDimensions get_dimensions(AVCaptureDeviceFormat *format);
 
-static AVCaptureDeviceFormat *find_format(AVCaptureDevice *dev, CMVideoDimensions dims)
```
norihiro, line 906: This function was added by 3d558d65ba but was never used at that time.

plugins/obs-ffmpeg/obs-ffmpeg-mpegts.c
```diff
@@ -626,14 +626,6 @@ void ffmpeg_mpegts_data_free(struct ffmpeg_output *stream,
 	memset(data, 0, sizeof(struct ffmpeg_data));
 }
 
-static inline const char *safe_str(const char *s)
```
norihiro, line 629: This function was added by e7d097cab8 but was never used at that time.

plugins/obs-outputs/ftl-stream.c
```diff
@@ -109,8 +109,6 @@ struct ftl_stream {
 	frame_of_nalus_t coded_pic_buffer;
 };
 
-static void log_libftl_messages(ftl_log_severity_t log_level,
-				const char *message);
```
norihiro, line 113: This function was added by 6f8e38e0a08f2a7a1ad422b83516be106f771748 but was never used at that time.

plugins/obs-outputs/ftl-stream.c
```diff
@@ -122,12 +120,6 @@ static const char *ftl_stream_getname(void *unused)
 	return obs_module_text("FTLStream");
 }
 
-static void log_ftl(int level, const char *format, va_list args)
```
norihiro, line 125: This function was adde by 6f8e38e0a0 but was never used at that time.

plugins/obs-outputs/rtmp-stream.c
```diff
@@ -227,13 +227,6 @@ static void rtmp_stream_stop(void *data, uint64_t ts)
 	}
 }
 
-static inline void set_rtmp_str(AVal *val, const char *str)
```
norihiro, line 230: This function was added by 0cf9e0cfdd and was used in `try_connect`. Later, a commit 1ca065b0f replaced `set_rtmp_str` with `set_rtmp_dstr`.

plugins/rtmp-services/rtmp-common.c
```diff
@@ -837,12 +837,6 @@ static const char *rtmp_common_key(void *data)
 	return service->key;
 }
 
-static bool supports_multitrack(void *data)
```
norihiro, line 840: This function was added by 189fc7ab6a. But later commit af090573952eb0b removed the callback `supports_multitrack`.

norihiro (commented):
Added comments to refer to which commits left the unused static functions.

UI/obs-app.cpp
RytoEX: I wonder why we did that.

norihiro: The function `convert_14_2_encoder_setting` was added by commit 1f8fc8e8ef0.
Do you think we still need to call `convert_14_2_encoder_setting` so that a config file from 14.2 or older can still be correctly read?

derrod: I think at this point it's reasonable for us to remove ancient migrations such as this. tbh I don't like how migrations work at all and would prefer we rethink them at some point.

PR #9518, opened 2023-08-27
win-wasapi: Allow linking to game/window capture

Penwy:
### Description

This adds the ability to link an application audio output capture source to a given window/game capture source, so that it systematically captures the audio of whichever window is currently captured by the window/game capture.

### Motivation and Context

This feature (or equivalent ones) is frequently requested.
https://ideas.obsproject.com/posts/1923/application-specific-audio-capture-pick-game-capture-automatically-windows
https://ideas.obsproject.com/posts/1908/capture-application-audio-with-hotkey

### How Has This Been Tested?

Windows 11 22H2.
Linked sources together, tested behaviour on deleting/recreating both sources, changing window, unhooking, hooking and renaming on the linked source.
No observed crash, no memory leak.

### Types of changes

### Checklist:


Fenrirthviti:
I'm having a hard time understanding the intended workflow for this. Can you provide screenshots and a better explanation on how this is intended to be used?

Penwy:
The main usecase of this is with a game capture source in "capture any fullscreen" or hotkey mode.
When the window captured by such a source would change (by pressing the hotkey, for example), a wasapi source set to be linked to that game capture source will also change to capture the same window.
This removes the need to either go and manually change it in obs, or to have preemptively made wasapi sources for each and every potential window to be captured, greatly simplifying such workflows.

Admittedly, for window capture, equivalent behaviour to this could be had by ensuring window and priority are set to the same in both source, and this is naught more than just a convenience option to ensure both are always synced, but adding window capture to it was nothing more than adding a string so I figured it wouldn't harm.

Fenrirthviti:
You are explaining the what, which I understand, but not the how. What I don't understand is what "link" here means, and the actual steps to use this that a user would have to follow. Do they have to create a game capture and audio capture source and manually link them? Is this handled automatically?

It's unclear to me how this is *used* by an end user. If someone were to set this up from scratch, how would they do it? How would they manage it?

Penwy:
My apologies, I had misunderstood the question.
This is added as an alternative option for the standard  application audio capture source, so the workflow is to create such a source, check "Link to Game/Window capture source" in its properties, and select the pre-existing source to link to.
Screenshot 2023-08-27 185551

So yeah, from scratch it's just to create both sources then specify in the properties of the application audio capture that it has to follow the created game/window capture.

Fenrirthviti:
Thanks. I appreciate the effort here, but unfortunately this is not the approach we want to take. We've started poking at this internally to just add a simple checkbox in Game Capture to "Capture Audio" which has always been the preferred solution. Seems like it is mostly ready, look for a PR from derrod. If nothing else, this pushed us to implement it!

PR #9519, opened 2023-08-27
countdown: Add prefix + suffix script properties

psyb0t:
### Description
Added two new script properties, prefix_text and suffix_text, to the countdown script. These properties allow users to insert optional text before and after the countdown timer.
Screenshot_2023-08-27_03-38-50

### Motivation and Context
This change provides users with the flexibility to add contextual information alongside the countdown timer. It's a feature that enhances user customization without affecting existing functionality.

### How Has This Been Tested?
Manually tested on OS: Ubuntu 20.04 LTS, kernel: 5.8.0-53-generic, CPU arch: x86_64.
Added and removed prefix and suffix text, and verified that the timer displayed the text correctly in all scenarios.

### Types of changes

### Checklist:

### DIFF
--- a/UI/frontend-plugins/frontend-tools/data/scripts/countdown.lua
+++ b/UI/frontend-plugins/frontend-tools/data/scripts/countdown.lua
@@ -4,6 +4,8 @@ total_seconds = 0
 
 cur_seconds   = 0
 last_text     = ""
+prefix_text   = ""
+suffix_text   = ""
 stop_text     = ""
 activated     = false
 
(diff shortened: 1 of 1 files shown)


PatTheMav:
I might be mistaken, but I don't think the scripts that are bundled with the frontend-tools are meant to be used in production, they merely serve as examples for using the scripting API.

Adding a prefix, suffix, and final(?) text is a very modification very specifically suited to this stream setup and complicates the "generic" nature of the example code.

Given that the scripts can be freely copied and modified, I'd rather have the examples shipped with OBS be concise and implement a single simple use-case and leave it to users to adjust them to their needs.

Fenrirthviti:
I would agree that these scripts are not intended to really be used as-is, and are just examples.

This modification would be best submitted as a [resource on our forums](https://obsproject.com/forum/resources/) to share it with others.

Thanks for the submission either way!

PR #9520, opened 2023-08-27
CI: Fix ccache status output for workflow jobs

PatTheMav:
### Description
Removes custom Ccache configuration path for cache status output step in `build-obs` action.

### Motivation and Context
CCache configuration path was configurable until the CI script cleanup (as scripts were still intended to be used locally then), which is not needed anymore.

By specifying a custom path, the stats output of `ccache` will use the default cache directory (vs the custom one used on CI) and thus no cache hits will be reported.

### How Has This Been Tested?
Needs to be tested by CI run for pull request.

### Types of changes

### Checklist:


RytoEX (changes requested):
Commit message nits:
* Title Case: Ccache
* mid-sentence: ccache

Everything else looks fine.

PR #9524, opened 2023-08-27
win-capture: Add audio capture option to window/game capture

derrod:
### Description

Based on discussions and ideas from others I've come up with five differrent ways of doing this, this PR represents my favoured one but this is still up for discussion:

1. (first attempt) Making window/game capture a composite source ([branch](https://github.com/derrod/obs-studio/commits/game-cap-linked-audio-legacy))
2. Rerouting audio in libobs ([branch](https://github.com/derrod/obs-studio/commits/libobs-reroute-audio))
3. **(this PR)** Procedure call in win-wasapi to reroute its audio to another source ([branch](https://github.com/derrod/obs-studio/commits/wasapi-reroute-proc))
4. Using a submix in win-capture ([branch](https://github.com/derrod/obs-studio/commits/win-capture-audio-mix))
5. Using audio callbacks to re-emit the wasapi source's audio ([branch](https://github.com/derrod/obs-studio/commits/win-capture-audio-callback))

No. 5 is the only one that does not require modifications to win-wasapi, the UI, or libobs.

#### Demo
https://github.com/obsproject/obs-studio/assets/3123295/c18f67ea-f159-469e-8b4c-beed423dc3ea

### Motivation and Context

Inspired to revisit this by #9518, but with a UX more akin to what was originally floated in conversations on Discord when the "application audio capture" source was first introduced.

### How Has This Been Tested?

Tested with a few games, fullscreen, window, and hotkey mode.

### Types of changes

### Checklist:

### DIFF
--- a/plugins/win-capture/data/locale/en-US.ini
+++ b/plugins/win-capture/data/locale/en-US.ini
@@ -38,6 +38,9 @@ GameCapture.Rgb10a2Space="RGB10A2 Color Space"
 GameCapture.Rgb10a2Space.Srgb="sRGB"
 GameCapture.Rgb10a2Space.2100PQ="Rec. 2100 (PQ)"
 Mode="Mode"
+CaptureAudio="Capture Audio (BETA)"
+CaptureAudio.TT="When enabled, creates an \"Application Audio Capture\" source that automatically updates to the currently captured window/application. Note that if Desktop Audio is configured, this could result in doubled audio."
+AudioSuffix="Audio"
 
 # Generic compatibility messages
 Compatibility.GameCapture.Admin="%name% may require OBS to be run as admin to use Game Capture."
--- a/plugins/win-capture/CMakeLists.txt
+++ b/plugins/win-capture/CMakeLists.txt
@@ -44,6 +44,8 @@ target_sources(
   PRIVATE # cmake-format: sortable
           app-helpers.c
           app-helpers.h
+          audio-helpers.c
+          audio-helpers.h
           compat-format-ver.h
           compat-helpers.c
           compat-helpers.h
--- a/plugins/win-capture/audio-helpers.c
+++ b/plugins/win-capture/audio-helpers.c
@@ -1,0 +1,136 @@
+#include "audio-helpers.h"
+
+#include <util/dstr.h>
+
+static inline bool settings_changed(obs_data_t *old_settings,
+				    obs_data_t *new_settings)
+{
+	const char *old_window = obs_data_get_string(old_settings, "window");
+	const char *new_window = obs_data_get_string(new_settings, "window");
+
+	enum window_priority old_priority =
+		obs_data_get_int(old_settings, "priority");
+	enum window_priority new_priority =
+		obs_data_get_int(new_settings, "priority");
+
+	// Changes to priority only matter if a window is set
+	return (old_priority != new_priority && *new_window) ||
+	       strcmp(old_window, new_window) != 0;
+}
+
+static inline void reroute_wasapi_source(obs_source_t *wasapi,
+					 obs_source_t *target)
+{
+	proc_handler_t *ph = obs_source_get_proc_handler(wasapi);
+	calldata_t *cd = calldata_create();
+	calldata_set_ptr(cd, "target", target);
+	proc_handler_call(ph, "reroute_audio", cd);
+	calldata_free(cd);
+}
+
+void setup_audio_source(obs_source_t *parent, obs_source_t **child,
+			const char *window, bool enabled,
+			enum window_priority priority)
+{
+	if (enabled) {
+		obs_data_t *wasapi_settings = NULL;
+
+		if (window) {
+			wasapi_settings = obs_data_create();
+			obs_data_set_string(wasapi_settings, "window", window);
+			obs_data_set_int(wasapi_settings, "priority", priority);
+		}
+
+		if (!*child) {
+			struct dstr name = {0};
+			dstr_printf(&name, "%s (%s)",
+				    obs_source_get_name(parent),
+				    TEXT_CAPTURE_AUDIO_SUFFIX);
+
+			*child = obs_source_create_private(
+				AUDIO_SOURCE_TYPE, name.array, wasapi_settings);
+
+			// Ensure child gets activated/deactivated properly
+			obs_source_add_active_child(parent, *child);
+			// Reroute audio to come from window/game capture source
+			reroute_wasapi_source(*child, parent);
+			// Show source in mixer
+			obs_source_set_audio_active(parent, true);
+
+			dstr_free(&name);
+		} else if (wasapi_settings) {
+			obs_data_t *old_settings =
+				obs_source_get_settings(*child);
+			// Only bother updating if settings changed
+			if (settings_changed(old_settings, wasapi_settings))
+				obs_source_update(*child, wasapi_settings);
+
+			obs_data_release(old_settings);
+		}
+
+		obs_data_release(wasapi_settings);
+	} else {
+		obs_source_set_audio_active(parent, false);
+
+		if (*child) {
+			reroute_wasapi_source(*child, NULL);
+			obs_source_remove_active_child(parent, *child);
+			obs_source_release(*child);
+			*child = NULL;
+		}
+	}
+}
+
+static inline void encode_dstr(struct dstr *str)
+{
+	dstr_replace(str, "#", "#22");
+	dstr_replace(str, ":", "#3A");
+}
+
+void reconfigure_audio_source(obs_source_t *source, HWND window)
+{
+	struct dstr title = {0};
+	struct dstr class = {0};
+	struct dstr exe = {0};
+	struct dstr encoded = {0};
+
+	ms_get_window_title(&title, window);
+	ms_get_window_class(&class, window);
+	ms_get_window_exe(&exe, window);
+
+	encode_dstr(&title);
+	encode_dstr(&class);
+	encode_dstr(&exe);
+
+	dstr_cat_dstr(&encoded, &title);
+	dstr_cat(&encoded, ":");
+	dstr_cat_dstr(&encoded, &class);
+	dstr_cat(&encoded, ":");
+	dstr_cat_dstr(&encoded, &exe);
+
+	obs_data_t *audio_settings = obs_data_create();
+	obs_data_set_string(audio_settings, "window", encoded.array);
+	obs_data_set_int(audio_settings, "priority", WINDOW_PRIORITY_CLASS);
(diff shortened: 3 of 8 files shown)


derrod:
Updated to include window capture as well.

exeldro:
A small thing that might be strange for the users is if you rename the audio source or change properties from the Audio Mixer dock, that change will be gone as soon as the scene collection is loaded again.
Also the audio filters can not be done via the window/game capture filters, but have to be done on the audio source, which might not be clear to users.

derrod:
> A small thing that might be strange for the users is if you rename the audio source or change properties from the Audio Mixer dock, that change will be gone as soon as the scene collection is loaded again.

I removed the option to rename "hidden" sources, that seems like a sensible change anyway.

> Also the audio filters can not be done via the window/game capture filters, but have to be done on the audio source, which might not be clear to users.

Indeed, sadly I do not think there's a good way to solve that. This also affects other composite source such as groups and nested scenes, but those obviously communicate that more clearly by having children in the tree.

plugins/win-capture/data/locale/en-US.ini
```diff
@@ -38,6 +38,9 @@ GameCapture.Rgb10a2Space="RGB10A2 Color Space"
 GameCapture.Rgb10a2Space.Srgb="sRGB"
 GameCapture.Rgb10a2Space.2100PQ="Rec. 2100 (PQ)"
 Mode="Mode"
+CaptureAudio="Capture Audio"
+CaptureAudio.TT="When enabled an \"Application Audio Capture\" source will be created that automatically is updated to the currently captured window/application."
```
WizardCM, line 42: This wording tripped me up, took me 4 attempts to read it correctly & understand what it was trying to convey.
I'd also specify that the "Desktop Audio" source may conflict.

Something like:
```suggestion
CaptureAudio.TT="When enabled, creates an \"Application Audio Capture\" source that automatically updates to the currently captured window/application. Note that if Desktop Audio is configured, this could result in doubled audio."
```

libobs/obs-source.c
```diff
@@ -4192,6 +4203,33 @@ void obs_source_output_audio(obs_source_t *source,
 	pthread_mutex_unlock(&source->filter_mutex);
 }
 
+void obs_source_reroute_audio(obs_source_t *dst, obs_source_t *src)
```
gxalpha, line 4206: Just nitpicking, personally I'd make this return a bool indicating success or failure (and then return false on the following `return` calls below), just to be safe if it's needed in the future.

gxalpha (commented):
Seeing as the audio source is truly private now, would it make sense to remove the name handling (constructing the names, listening to the `rename` signal) and just giving it a `NULL` name? Seems like it would make the PR quite a bit smaller.

derrod:
> Seeing as the audio source is truly private now, would it make sense to remove the name handling (constructing the names, listening to the `rename` signal) and just giving it a `NULL` name? Seems like it would make the PR quite a bit smaller.

The name is still displayed in the log, if no name is provied one will be generated (which is something like `__unnamed001`), to make logs easier to understand I figured we should keep a source name that reflects what it is.

libobs/obs-source.c
derrod: Makes sense, it just wasn't useful here. If we can're reroute a source on creation something has gone horribly wrong.

cg2121:
If I remember correctly, you can set the output flags of the game/window capture sources to also use OBS_SOURCE_COMPOSITE, so you can use the .audio_render function to render the child audio source. This way you don't need all of the new audio reroute functions. See the slideshow source for an example of a composite source. At least I think this should work.

derrod:
> If I remember correctly, you can set the output flags of the game/window capture sources to also use OBS_SOURCE_COMPOSITE, so you can use the .audio_render function to render the child audio source. This way you don't need all of the new audio reroute functions. See the slideshow source for an example of a composite source. At least I think this should work.

That was the original version of this PR which is still on a branch in my fork[^1], the downside there is that the UX is slightly weird as the audio and video sources are separate, for example, you cannot directly add filters to the game capture source.

Based on feedback from @Warchamp7 and @Lain-B I added the rerouting feature to solve it more elegantly.

[^1] https://github.com/derrod/obs-studio/tree/game-cap-linked-audio-legacy

derrod:
Updated once again using an idea proposed by @norihiro to use a procedure call in the wasapi source instead for the rerouting, this skips having to modify libobs itself and is overall arguably cleaner.

Currently the three variations are available on different branches in my repo:
1. Composite source: https://github.com/derrod/obs-studio/commits/game-cap-linked-audio-legacy
2. libobs rerouting: https://github.com/derrod/obs-studio/commits/libobs-reroute-audio
3. proc call rerouting (this PR): https://github.com/derrod/obs-studio/commits/wasapi-reroute-proc

I guess 4. would be using `source_info.audio_mix` which has advantages of its own, mainly that neither libobs nor win-wasapi would need to be modified, but that's not something I'll deal with today.

cg2121:
> If I remember correctly, you can set the output flags of the game/window capture sources to also use OBS_SOURCE_COMPOSITE, so you can use the .audio_render function to render the child audio source. This way you don't need all of the new audio reroute functions. See the slideshow source for an example of a composite source. At least I think this should work.

This doesn't work like I thought it would, as a composite source can't also be an audio source, so the re-routing audio feature @derrod came up with is the preferred solution here.

cg2121:
Would the audio_line source work for this PR, as it has the OBS_SOURCE_SUBMIX flag? https://github.com/obsproject/obs-studio/blob/112adb0a73c968658336a23d8dac1d02c82f6892/libobs/obs.c#L1227

derrod:
You'd still need to get the audio from the wasapi source to that one though. Maybe another way could be getting the child's audio data via `obs_source_add_audio_capture_callback()` and then re-output it directly...

But I feel that all those approches ultimately are overcomplicating something.

derrod:
Well here's take #5 anyway, using a callback to re-output the audio from the window/game capture sources. This one is finally fully-self contained within `win-capture`: https://github.com/derrod/obs-studio/commits/win-capture-audio-callback

norihiro:
I feel negative about using `obs_source_add_audio_capture_callback` because the callback is called after the audio data has been placed into `audio_input_buf` for the private audio source, which might grow the audio buffering unnecessarily.

In the code below, `source_output_audio_push_back` and `source_output_audio_place` place the data into the buffer. `source_signal_audio_data` calls the callback.
https://github.com/obsproject/obs-studio/blob/5ad60809a850ff5829cbd23a6683f71e1d42c4fe/libobs/obs-source.c#L1561-L1570

derrod:
Mh not quite sure why it would increase audio buffering? It'll use a bit of extra RAM, yes, but I don't think that should impact audio buffering.

derrod:
I guess one downside of the current approach (audio callback) is that it'll get the resampled audio from OBS, i.e. if OBS is set to stereo but the audio capture source is non-stereo (e.g. 5.1) then the game capture audio will still show up as stereo. Dunno if that is really a concern though.

GeorgesStavracas (approved):
UI / UX approved (no code was reviewed though!)

norihiro:
> Mh not quite sure why it would increase audio buffering? It'll use a bit of extra RAM, yes, but I don't think that should impact audio buffering.

The audio buffering I'm referring is this one.
https://github.com/obsproject/obs-studio/blob/2308414bcc89975647c8db06609b3f03b43a422b/libobs/obs-audio.c#L384-L388


If audio from the private source is delayed, the above code will be triggered. Assuming the code is more likely called when the sync-offset is 0 than a positive sync-offset, the audio buffering will happen more than other approaches because the private audio source will always have 0 sync-offset.

derrod:
> If audio from the private source is delayed, the above code will be triggered. Assuming the code is more likely called when the sync-offset is 0 than a positive sync-offset, the audio buffering will happen more than other approaches because the private audio source will always have 0 sync-offset. The audio subsystem of libobs is complicated. I feel this would be another motivation to avoid putting audio into libobs and taking by the callback.

I don't think this would behave any differently from the other approaches since all of them retain the original timestamps. The sync offset is only used when the user changes it in the advanced audio properties and that generally shouldn't be required for WASAPI process capture.

Also not entirely sure what you mean by "avoid putting audio into libobs and taking by the callback".

derrod:
Rebased and updated to use the procedure call approach based on discussions with @norihiro, seems to be the safest as it avoids any libobs modifications and re-emitting the audio via callback may cause issues with audio buffering.

Lain-B (approved):

Lain-B:
Might need to be marked as a beta feature until process audio capture gets a MS-side fix.

DmitriyFrogo:
My OBS always crashing on startup with these code changes, even with deleted %appdata%\obs-studio\ folder

Crash Log: [Crash 2023-12-13 00-35-39.txt](https://github.com/obsproject/obs-studio/files/13653441/Crash.2023-12-13.00-35-39.txt)
Log: [2023-12-13 00-35-37.txt](https://github.com/obsproject/obs-studio/files/13653442/2023-12-13.00-35-37.txt)

DmitriyFrogo:
Now it's crashing on exit
Seems like, if you already have Window or Game Capture, before updating to this build, and then enable "Capture Audio (Beta)", it's messing with something and can't properly assign audio tracks
Even if you manually assign track - it crashes after exit
изображение

To fix this, I need to disable "Sound Capture (BETA)" first, then delete old capture source, and create new one
If I want to delete old capture source without disabling Sound Capture, it deletes from sources list, but continues to crash on exit, and I can't create new capture source with same name

[2023-12-13 19-51-43.txt](https://github.com/obsproject/obs-studio/files/13663462/2023-12-13.19-51-43.txt)
[Crash 2023-12-13 19-51-50.txt](https://github.com/obsproject/obs-studio/files/13663464/Crash.2023-12-13.19-51-50.txt)

DmitriyFrogo:
> To fix this, I need to disable "Sound Capture (BETA)" first, then delete old capture source, and create new one
If I want to delete old capture source without disabling Sound Capture, it deletes from sources list, but continues to crash on exit, and I can't create new capture source with same name, like it's still here

I sometimes forgot to trigger capture with active window, when I test this
Now it crashing on exit every time after I have captured any active window/game

DmitriyFrogo:
Tested for about 5 hours
No issues

DmitriyFrogo:
Still no issues after ~40 hours of tests

Nice feature that allows you to avoid mess in sources list
But one thing that can be improved
You can mute source when the game/window is visible
But you can't hide game/window so the sound keeps playing
So you still need to use separate Game/Window Capture + "App Sound Capture (Beta)" and make mess with sources list

PR #9528, opened 2023-08-28
CI: Suppress and fix majority of macOS-specific warnings

PatTheMav:
### Description
Either suppresses or fixes a majority of warnings currently emitted by macOS builds.

### Motivation and Context
After reviewing the warnings currently emitted by macOS builds, a majority of them falls into these categories:

* Integer truncation: `libobs` uses 64-bit integers internally which are (with a few exceptions) stored in 16-bit integers. Given that most values used by code should fit within that smaller type, the truncation is possible, but improbable and even if it occurs should not lead to runtime issues.
* Outdated or abandoned vendored dependencies: Maintainers are aware of the status of these dependencies, with plans to either remove or replace them in due time. The warnings have been acknowledged and can thus be silenced.

Given the widespread occurrence of the warnings, it was not feasible to silence them specifically on each occasion (while leaving the warning overall intact for new code), which is unfortunate but the most pragmatic choice right now.

Warnings for macOS-specific code have been fixed instead (e.g. by using explicit casts).

### How Has This Been Tested?
Code successfully compiled and ran on macOS 13.4.

### Types of changes

### Checklist:

### DIFF
--- a/deps/json11/CMakeLists.txt
+++ b/deps/json11/CMakeLists.txt
@@ -4,5 +4,6 @@ add_library(json11 INTERFACE)
 add_library(OBS::json11 ALIAS json11)
 
 target_include_directories(json11 INTERFACE "${CMAKE_CURRENT_SOURCE_DIR}")
+target_compile_options(json11 INTERFACE $<$<COMPILE_LANG_AND_ID:CXX,AppleClang,Clang>:-Wno-unqualified-std-cast-call>)
 
 target_sources(json11 INTERFACE json11.cpp json11.hpp)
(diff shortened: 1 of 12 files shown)


RytoEX:
Looks okay at a glance. Does any of this need mirrored in the old CMake paths to keep them relatively in sync?

PatTheMav:
macOS has no old CMake paths since 29.1 anymore and the CMake changes only affect Clang and AppleClang.

gxalpha (approved):
Looks reasonable.

PR #9529, opened 2023-08-28
CI: Add automatic code annotations for macOS arm64 builds

PatTheMav:
### Description
Adds automatic code annotations of warnings and errors for pull requests and scheduled nightly builds.

### Motivation and Context
Warnings and errors are usually buried within the build output and thus require more work to pick up for reviews.

By adding annotations the offending code parts can be reviewed directly in GitHub's CI, jumping directly to the code lines that generated the warning or error.

This is currently enabled for macOS arm64 builds only due to `xclogparser` only being available for Xcode build logs. But if other tools can generate similar JSON files (or files as similarly easy to convert using `jq`) the annotations can be added for other platforms as well.

### How Has This Been Tested?
Tested on local fork and successfully checked out generated warnings.

### Types of changes

### Checklist:


PatTheMav:
Annoyingly, pull requests with a 3rd party fork as the source are limited to `read` permissions, so PRs can never be annotated this way (PRs from the same fork do not have this limitation which is why I was able to test this on my own fork successfully).

As such this requires either a push, dispatch, or scheduled workflow run as only those are able to get `write` permissions for checks.

RytoEX (approved):
This is probably fine. I suppose we'll see how it goes later.

PR #9530, opened 2023-08-28
mac-videotoolbox: Set default keyframe interval to 2 seconds, Remove CRF 0 second keyint override

gxalpha:
### Description

0 seconds means "auto" according to the [documentation](https://developer.apple.com/documentation/videotoolbox/kvtcompressionpropertykey_maxkeyframeinterval), but this appears
to be broken in many configurations (more than just CRF mode on Apple
Silicon). With some encoders it means that the encoder sets a keyframes
every 31st frame, other times it just doesn't set any keyframes at all
after the first one, only rarely does the "auto" interval actually
appear to work.
Lets just set the default to 2 seconds and be happy. In theory this is
the maximum keyframe interval and encoders are allowed to set more if
they so wish, but they never appear to do so.

Also removes the 0 second keyint override in CRF mode.
The 0 second default interval no longer happens per default, if
someone explicitly sets the interval to 0 seconds ("auto"), we should
allow them to do that.

This change will effect existing configurations, but probably for the better.

### Motivation and Context

Videos without keyframes, from various encoders (Intel, Apple Silicon) and in various rate control configurations, still come up in support way too often.
This change should mostly eliminate them.

### How Has This Been Tested?

macOS 14
Tested that in Advanced the default is now 2 seconds; and in Simple the encoder always has 2 seconds.
Verified that recordings come out with a 2 second keyframe interval via ffprobe.

### Types of changes

### Checklist:

### DIFF
--- a/plugins/mac-videotoolbox/encoder.c
+++ b/plugins/mac-videotoolbox/encoder.c
@@ -596,18 +596,6 @@ static bool create_encoder(struct vt_encoder *enc)
 
 	if (enc->codec_type == kCMVideoCodecType_H264 ||
 	    enc->codec_type == kCMVideoCodecType_HEVC) {
-		/* Apple's documentation states that a keyframe interval of 0 will result in
-		 * the encoder automatically picking times to insert them; However, Apple's
-		 * encoder, when in CRF mode, will never actually insert any keyframes past
-		 * the very first one, rendering the files near-unusable in editors or
-		 * video players. So to avoid that happening, enforce a reasonable default
-		 * of 10 seconds in CRF mode. */
-		if (enc->keyint == 0 && strcmp(enc->rate_control, "CRF") == 0) {
-			VT_BLOG(LOG_INFO,
-				"Enforcing non-zero keyframe interval in CRF mode");
-			enc->keyint = 10;
-		}
-
 		// This can fail when using GPU hardware encoding
 		code = session_set_prop_int(
 			s,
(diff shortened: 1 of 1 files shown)


PatTheMav:
IIRC the setting actually sets keyframes according to actual scene changes (detected with some heuristics), so setting it to an automatic value will not do what people expect it to for the most part.

I think we reported this to Apple but I'm not aware if there will be any meaningful change to this behaviour on macOS 14.

derrod (commented):
On other encoders we use a nonzero keyframe default so that I'm not opposed to, but I'm wondering if this could break things for some users using CRF mode already and now would no longer have the safety blanket of a forced minimum keyframe interval.

gxalpha:
I think the impact would be pretty low? Seeing as we don't store defaults anyone who was set at the default 0/auto seconds is now at 2, it'd only be people who purposefully put it back to 0 probably?

basisbit:
+1 for changing the `default` keyframe setting value from 0 to 2. This change will improve simplicity of using OBS for many common use cases, thus also reducing support overhead, and it is one less setting which probably every video live streamer has to change every time they pick an encoder in the OBS video settings.

derrod:
In that case it's probably fine? Not a fan of even allowing the broken combination of a keyframe interval of 0 and CRF though.

gxalpha:
In that case I’d argue we should remove 0 completely. The point of this PR is that it’s *not* just 0 and CRF, but that the broken behavior as happened in various configurations with any rate control method.

Lain-B (approved):

PR #9535, opened 2023-08-29
mac-avcapture: Log if portrait or studio light effects are active

gxalpha:
### Description

These effects can remain enabled between devices (so enabling it on one device can mean it's later also enabled on another device), leading to cases of capture cards getting blurred.
Logging that the effects are enabled should make it easier to spot this in support.

### Motivation and Context

Blurred capture cards come up sometimes in support.
Medium term I'd also like to put a warning about it somewhere in the UI (maybe via the properties of the source, or the notifications system once it exists), but that brings UI/UX discussion with it. This PR just adds a log line on device selection (so on OBS startup or when a user switches the device).
Would like to get this into OBS 30.0 if possible (therefore the small scope).

### How Has This Been Tested?

macOS 14
Enabled the effects on my camera and saw it being logged upon device selection. Disabled the effects and saw that upon reselecting it would no longer be logged.

### Types of changes

### Checklist:

### DIFF
--- a/plugins/mac-avcapture/av-capture.mm
+++ b/plugins/mac-avcapture/av-capture.mm
@@ -1118,6 +1118,18 @@ static void capture_device(av_capture *capture, AVCaptureDevice *dev, obs_data_t
     obs_data_set_string(settings, "device_name", name);
     obs_data_set_string(settings, "device", dev.uniqueID.UTF8String);
     AVLOG(LOG_INFO, "Selected device '%s'", name);
+    if (@available(macOS 12.0, *)) {
+        if ([dev isPortraitEffectActive])
+            AVLOG(LOG_WARNING, "Portrait effect is active on selected device");
+    }
+    if (@available(macOS 12.3, *)) {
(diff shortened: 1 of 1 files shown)


PatTheMav:
Are those all known effects? I think there’s also the smart-follow mode, I’ll have to check if there’s a variable for it.

gxalpha:
>  I think there’s also the smart-follow mode, I’ll have to check if there’s a variable for it.

There is `isCenterStageActive`, yes. I didn't include it as I *believe* the camera must be explicitly able to use it, so at least for capture cards that wouldn't be an issue, but could include it as well.

PatTheMav:
> > I think there’s also the smart-follow mode, I’ll have to check if there’s a variable for it.
> 
> There is `isCenterStageActive`, yes. I didn't include it as I _believe_ the camera must be explicitly able to use it, so at least for capture cards that wouldn't be an issue, but could include it as well.

Right, I guess for capture cards checking for the two existing flags is sufficient.

PR #9536, opened 2023-08-30
libobs: Fix obs_output_video and obs_output_audio for encoded output

norihiro:
### Description

This PR partially reverts fb57eff21 and 645e31fa1 so that `obs_output_video` can return a valid pointer instead of NULL.

### Motivation and Context

Some plugins such as obs-websocket and obs-midi expect `obs_output_video` to return a valid pointer to calculate the duration of streaming and recording regardless of the output type is raw or encoded.

- https://github.com/obsproject/obs-websocket/blob/3cd81639457b789bda564cca58badf35f0285106/src/utils/Obs_NumberHelper.cpp#L31
- https://github.com/cpyarger/obs-midi/blob/6b9d12632370a994767284acf27482da73025b83/src/events.cpp#L340

The commit 645e31fa1 blocked `obs_output_set_media` to update `video` and `audio` pointers if it's an encoded type.  It resulted in the encoded output holding a released video pointer at the auto-config wizard.

Another commit fb57eff21 tried to fix the output context holding the released pointer. However, it broke the compatibility with the plugins.

### How Has This Been Tested?

OS: Fedora 37

1. Ran OBS and enabled obs-websocket.
2. Started streaming.
3. Called a request `GetStreamStatus` and checked it returns non-zero `outputDuration` field.

Also tested the auto-config wizard with the modification to accelerate the crash described in #9409 and checked it does not crash.

### Types of changes

### Checklist:

### DIFF
--- a/libobs/obs-output.c
+++ b/libobs/obs-output.c
@@ -201,10 +201,8 @@ obs_output_t *obs_output_create(const char *id, const char *name,
 	} else {
 		output->info = *info;
 	}
-	if (!flag_encoded(output)) {
-		output->video = obs_get_video();
-		output->audio = obs_get_audio();
-	}
+	output->video = obs_get_video();
+	output->audio = obs_get_audio();
 	if (output->info.get_defaults)
 		output->info.get_defaults(output->context.settings);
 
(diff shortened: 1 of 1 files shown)


tt2468:
Not sure this is the right approach. Will brainstorm today

norihiro:
Another solution could be this reverting PR + calling `obs_output_create` just before starting recording or streaming instead of calling `obs_output_create` at the beginning so that `video_t *`  won't be changed.

PR #9537, opened 2023-08-30
Revert "UI: Don't clear lastService immediately after setting it"

gxalpha:
### Description

This reverts commit 633b9e39f71a18fbf9c47817e608b8222c9c1aa7.
Reintroduces the bug solved by it, but that is a necessary evil.

### Motivation and Context

The commit broke stream keys for services with integration. Now, only YouTube seems to be broken, Twitch is fine.

### How Has This Been Tested?

macOS 14
When using Twitch, the "Use stream key" setting now works.
When using YouTube, it still doesn't.

### Types of changes

### Checklist:

### DIFF
--- a/UI/window-basic-settings-stream.cpp
+++ b/UI/window-basic-settings-stream.cpp
@@ -162,6 +162,7 @@ void OBSBasicSettings::LoadStream1Settings()
 	else
 		ui->key->setText(key);
 
+	lastService.clear();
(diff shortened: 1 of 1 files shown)


Lain-B:
No need to revert, your original fix was actually good, don't worry. I solved the problem in an alternate/better way just now in 112adb0a73c968.

PR #9539, opened 2023-08-30
UI: Increase YouTube API timeout 

joelgerard:
The OBS client side timeout for the YouTube API is too short. New changes proposed to the YouTube API will cause this to fail for creating a broadcast. Increase to a minimum of 60.

### Description
Increase the timeout for the API used to create a YouTube broadcast to a minimum of 60 seconds.

### Motivation and Context
The server side timeout is significantly higher. New changes to the create process are causing OBS
to prematurely timeout.

### How Has This Been Tested?
Tested on Windows.

### Types of changes

- Bug fix (non-breaking change which fixes an issue) -

### Checklist:

### DIFF
--- a/UI/youtube-api-wrappers.cpp
+++ b/UI/youtube-api-wrappers.cpp
@@ -90,7 +90,7 @@ bool YoutubeApiWrappers::TryInsertCommand(const char *url,
 	std::string output;
 	std::string error;
 	// Increase timeout by the time it takes to transfer `data_size` at 1 Mbps
-	int timeout = 5 + data_size / 125000;
+	int timeout = 60 + data_size / 125000;
 	bool success = GetRemoteFile(url, output, error, &httpStatusCode,
 				     content_type, request_type, data,
 				     {"Authorization: Bearer " + token},


derrod:
What changes to the create process are you referring to?

notr1ch:
Nit: Commit titles should not end with punctuation.

joelgerard:
> What changes to the create process are you referring to?

All YouTube server side, which is now using the channel level monetization defaults if no monetization settings are specified in the call. This is increasing the latency enough that it triggers the OBS client side timeout enough to cause issues.

derrod:
> All YouTube server side, which is now using the channel level monetization defaults if no monetization settings are specified in the call. This is increasing the latency enough that it triggers the OBS client side timeout enough to cause issues.

I see, technically this PR changes the timeout for *all* API calls, so the commit message should be updated to reflect that.

joelgerard:
> > All YouTube server side, which is now using the channel level monetization defaults if no monetization settings are specified in the call. This is increasing the latency enough that it triggers the OBS client side timeout enough to cause issues.
> 
> I see, technically this PR changes the timeout for _all_ API calls, so the commit message should be updated to reflect that.

Thanks. I misread that. Fixed.

norihiro:
I have another nitpick. The commit message should be wrapped at 72 columns.

RytoEX:
> I have another nitpick. The commit message should be wrapped at 72 columns.

I've fixed this and force-pushed.

PR #9546, opened 2023-09-01
UI: Fix settings Apply button always being enabled

derrod:
### Description

Fixes the settings window always having the "Apply" button enabled as though things have changed.

This was simply due to the state of the checkbox being changed before the "loading" flag was set.

### Motivation and Context

Fix UI.

### How Has This Been Tested?

Opened settings, apply button no longer enabled.

### Types of changes

### Checklist:


RytoEX (approved):
Makes sense. Looks like this was a regression in 30.0.0-beta1, though how/why is unclear since the order of the lines hasn't changed in years. Possibly either the Qt update, or changes we made elsewhere.

PR #9549, opened 2023-09-02
libobs: Remove old macOS guards for NSProcessInfo

PatTheMav:
### Description
Updates `NSProcessInfo` code to interact with the class directly instead of indirect ObjC message calls.

### Motivation and Context
With the lowest deployment target being macOS 11.0, there is no need to check for the availability of the selectors on the NSProcessInfo class anymore.

The API was added in Mac OSX 10.9, so we are safe:

```
- (id <NSObject>)beginActivityWithOptions:(NSActivityOptions)options reason:(NSString *)reason API_AVAILABLE(macos(10.9), ios(7.0), watchos(2.0), tvos(9.0));
```

### How Has This Been Tested?
Tested on macOS 13.5.1.

### Types of changes

### Checklist:

### DIFF
--- a/libobs/util/platform-cocoa.m
+++ b/libobs/util/platform-cocoa.m
@@ -223,13 +223,8 @@ void os_cpu_usage_info_destroy(os_cpu_usage_info_t *info)
 os_performance_token_t *os_request_high_performance(const char *reason)
 {
     @autoreleasepool {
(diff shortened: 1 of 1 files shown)


RytoEX (approved):
Discussed off-thread and it seems fine.

PR #9181, opened 2023-06-30
cmake: Add support for unobfuscated OAuth secrets

PatTheMav:
### Description
Unobfuscated secrets require the hash values to be set to 0, by default CMake will treat 0 as a falsy value. This commit adds support for _either_ 0 _or_ a valid hexadecimal hash.

### Motivation and Context
Allow maintainers and forks to provide their own unobfuscated OAuth credentials.

### How Has This Been Tested?
Tested locally with variants of empty hash values, using `0`, and using actual hash values.

### Types of changes
- Bug fix (non-breaking change which fixes an issue)

### Checklist:
- [x] My code has been run through [clang-format](https://github.com/obsproject/obs-studio/blob/master/.clang-format).
- [x] I have read the [**contributing** document](https://github.com/obsproject/obs-studio/blob/master/CONTRIBUTING.rst).
- [x] My code is not on the master branch.
- [x] The code has been tested.
- [x] All commit messages are properly formatted and commits squashed where appropriate.
- [x] I have included updates to all appropriate documentation.

PR #9568, opened 2023-09-06
UI: Restrict regex pattern for OAuth secrets

RytoEX:
### Description

The current regex pattern will successfully match if the cache variable contains the pattern anywhere inside the string. Let's restrict the regex pattern such that it requires the pattern to be precisely between the beginning and end of the string with no other characters in between those anchor points.

### Motivation and Context

Was testing locally and ran into this when I tried to build without obs-browser but the "YouTube API connection" feature was still enabled. I tried to disable that by deliberately prefixing the `YOUTUBE_CLIENTID_HASH` cache variable with what I thought were characters sure to avoid matching the pattern ("removemelater_"). However, the pattern still matched, and the feature was still enabled. I noticed that the pattern did not use beginning and end of string anchors, so I added them, and then the pattern match failed as expected.

The current regex pattern would also likely successfully match if `0` were present anywhere in the cache variables, instead of just matching a variable that was exactly `0`.

### How Has This Been Tested?

Tested locally on Windows 11. See Motivation and Context section.

### Types of changes

### Checklist:

### DIFF
--- a/UI/cmake/feature-restream.cmake
+++ b/UI/cmake/feature-restream.cmake
@@ -1,5 +1,5 @@
 if(RESTREAM_CLIENTID
-   AND RESTREAM_HASH MATCHES "(0|[a-fA-F0-9]+)"
+   AND RESTREAM_HASH MATCHES "^(0|[a-fA-F0-9]+)$"
    AND TARGET OBS::browser-panels)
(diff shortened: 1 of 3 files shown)


RytoEX:
Follow up to #9181 (https://github.com/obsproject/obs-studio/commit/c586c1d06f02912cb23871024bf576668c39f26a) where this pattern was added.

Issue #5282, opened 2021-09-14
Deadlock between threads using data->sources_mutex and current_lua_script->mutex


OldBaldGeek:
### Operating System Info

Windows 10

### OBS Studio Version

27.0.1

### OBS Studio Version (Other)

Also occurs with home-build using current source

### OBS Studio Log URL

OBS hangs; log file doesn't show this

### Expected Behavior

Non-looped Lua scripts shouldn't cause a hang or deadlock

### Current Behavior

Deadlock in at least the cases described below

### Steps to Reproduce

I am triggering a mutex deadlock, by running a diagnostic Lua script that hooks to frontend events and source activate/deactivate and show/hide signals. The script logs changes in scenes and sources, and is sometimes useful to help me debug other scripts. (Script attached)

[logorrhea.lua.txt](https://github.com/obsproject/obs-studio/files/7165294/logorrhea.lua.txt)

After a variable number of transitions (1 to over 50, and sometimes before the UI is even shown) the UI freezes. I still see "thread XXXX has exited" etc., so other threads are still running.

This happens both with current source and in version 27.0.1. Seems to occur more frequently in a debug build.

In a debug build of current (as of 14 Sept 2021) source, I break OBS after the hang giving these results:

Windows thread ID 3DD4 (Main Thread) is pthread_t* 0x000001d3db5b5f70
   Waiting for mutex 0x000001d3dd76a4d0 owned by 0x000001d3dfa25600 (This is data->sources_mutex)
   Stack has 
>w32-pthreads.dll!pthread_mutex_lock(pthread_mutex_t_ * * mutex) Line 127	C
 	obs.dll!obs_enum_sources(bool(*)(void *, obs_source *) enum_proc, void * param) Line 1451	C
 	obs-scripting.dll!enum_sources(lua_State * script) Line 589	C
 	[External Code]	
 	obs-scripting.dll!call_func_(lua_State * script, int reg_idx, int args, int rets, const char * func, const char * display_name) Line 234	C
 	obs-scripting.dll!frontend_event_callback(obs_frontend_event event, void * priv) Line 203	C
 	obs64.exe!OBSStudioAPI::on_event(obs_frontend_event event) Line 655	C++
 	obs64.exe!OBSBasic::TransitionStopped() Line 328	C++
 	obs64.exe!OBSBasic::qt_static_metacall(QObject * _o, QMetaObject::Call _c, int _id, void * * _a) Line 1197	C++
 	[External Code]	
[...]

PR #9570, opened 2023-09-07
obs-scripting: Add obs_enter_graphics_timed as a deadlock workaround

stephematician:
### Description

Adds a `obs_enter_graphics_timed` function to libobs that accepts a second argument for the time (in nanoseconds) to wait before abandoning the attempt to get the lock. The underlying timeout mechanism is managed by `pthread_mutex_timedlock`.

Also adds a 'realtime' clock accessor to libobs, called `os_realtime_ts`, that returns a `timespec` that can be passed on to functions like `pthread_mutex_timedlock`.

`pthread_mutex_timelock` is not implemented on MacOS, so instead a busy wait is used; this is a stop-gap measure for now (pending feedback on PR).

As an experiment, this PR also adds a `yield_script` and `yield_source_defn` function in `obslua` which provide a mechanism to release the lock on, respectively, the script, or the the script + source defn, for a given period of time (in ms). This allows the graphics thread to acquire locks on the script (or source) to perform an action that would have otherwise deadlocked; the script then attempts to re-acquire the locks.

Aside: I don't particularly like the 'yield', but it could be useful for people working on scripts that would otherwise experience deadlocks; it doesn't seem particularly advisable for a deployed script.

### Motivation and Context

The intention of these additions is to allow those writing Lua scripts (and eventually Python too) to handle failures to lock a mutex (specifically the graphics mutex) within a small amount of time when (re)loading the script, or adding/removing sources (see issue #6674). The experimental 'yields' allow a Lua script user/maintainer to respond to a failed lock attempt with a brief yield which would allow the graphics thread to perform some action, then (perhaps) attempt to acquire the lock again.

Note: there are still leftover deadlock issues in scripts that are _not_ addressed by this PR, in particular, those related to `sources_mutex` e.g. issue #5282. _Although I have yet to encounter one_. 

### How Has This Been Tested?

I have tested these changes with a custom `clock_source.lua` 
[clock-source-timed.tar.gz](https://github.com/obsproject/obs-studio/files/12550307/clock-source-timed.tar.gz)

The main change from the OG `clock_source.lua` are the parts where `obs_enter_graphics` is invoked, they now look like:

```lua
function ms_to_ns(ms) return ms * 1000000 end

source_def.destroy = function(data)
    while not (obs.obs_enter_graphics_timed(ms_to_ns(10)) == 0) do
        obs.script_log(obs.LOG_WARNING,
                       "destroy() yielding source definition")
        obs.yield_source_defn(10, source_def)
    end
-- clean up data ...
end
```

As per issue #6674 - repeatedly (re)loading the script shows that what would have been deadlocks (i.e. locks that took too long to acquire), however instead of locking, the script provides some warnings, yields the locks, and eventually re-acquires them. The graphics thread reports (sometimes) that it cannot render a source because it's definition lacks data - this is an expected side effect if the Lua (source) has to release the lock during construction. No other adverse behaviour is noted.

An alternative approach for the Lua (user) would be to have a flag for the state of the source data, e.g.:

```lua
source_def.create = function(source, settings)
    local data = {}
    data.is_valid_state = false
    if not (obs.obs_enter_graphics_timed(ms_to_ns(20)) == 0) then
        obs.script_log(obs.LOG_WARNING, "source_create failed to enter context")
        return data
    end
--  construct data ...
    data.is_valid_state = true
    return data
end

source_def.video_render = function(data)
    if not data.is_valid_state then return end
-- render...
end
```

### Types of changes

### Checklist:

I have not documented the new features as I'm awaiting further feedback on this PR.

### DIFF
--- a/deps/obs-scripting/obs-scripting-lua-source.c
+++ b/deps/obs-scripting/obs-scripting-lua-source.c
@@ -16,6 +16,7 @@
 ******************************************************************************/
 
 #include "obs-scripting-lua.h"
+#include <util/platform.h>
 #include "cstrcache.h"
 
 #include <obs-module.h>
@@ -694,15 +695,41 @@ fail:
 	return 0;
 }
 
+static int obs_lua_yield_source_defn(lua_State *script)
+{
+	struct obs_lua_script *data = current_lua_script;
+
+	int sleep_ms = (int)lua_tointeger(script, -2);
+	const char *id = get_table_string(script, -1, "id");
+
+	if (!id)
+		return 0;
+
+	struct obs_lua_source *existing = find_existing(id);
+
+	if (data)
+		pthread_mutex_unlock(&data->mutex);
+	if (existing)
+		pthread_mutex_unlock(&existing->definition_mutex);
+
+	os_sleep_ms(sleep_ms);
+
+	if (existing)
+		pthread_mutex_lock(&existing->definition_mutex);
+	if (data)
+		pthread_mutex_lock(&data->mutex);
+
+	return 0;
+}
+
 /* ========================================================================= */
 
 void add_lua_source_functions(lua_State *script)
 {
 	lua_getglobal(script, "obslua");
 
-	lua_pushstring(script, "obs_register_source");
-	lua_pushcfunction(script, obs_lua_register_source);
-	lua_rawset(script, -3);
+	add_func("obs_register_source", obs_lua_register_source);
+	add_func("yield_source_defn", obs_lua_yield_source_defn);
 
 	lua_pop(script, 1);
 }
--- a/deps/obs-scripting/obs-scripting-lua-source.c
+++ b/deps/obs-scripting/obs-scripting-lua-source.c
@@ -711,6 +711,9 @@ static inline void undef_source_type(struct obs_lua_script *data,
 				     struct obs_lua_source *ls)
 {
 	pthread_mutex_lock(&ls->definition_mutex);
+
+	struct obs_lua_script *__prev_script = current_lua_script;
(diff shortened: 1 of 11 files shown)


stephematician:
Oof - I missed something terribly obvious in `obs.c`. As I don't have a Windows machine I didn't have high hopes for successful compilation. I'll fix these problems soon (and dust off clang-format-16).

stephematician:
So it appears that MacOS doesn't implement `pthread_mutex_timedlock`; that means finding another function like it OR implementing our own; similar to https://github.com/elast0ny/shared_memory/commit/9c987b29827eb128589abdd59fe7a9faec83b806 ?

stephematician:
Tidied up the PR so it builds on the three test platforms.

I have _not_ tested the changes on MacOS or Windows as I do not have either system.

PatTheMav (commented):
I don't think we will merge this change - for one it only fixes a single symptom of a larger architectural issue (mainly that scripts are single-threaded and thus don't work well with OBS' multi-threaded infrastructure once they interact with other threaded elements of the program), only fixes it for Lua (and keeps Python unchecked) and by its own admission wasn't even tested on Windows (our major platform).

PatTheMav:
The entire scripting system was added as a nice little gift to our users but did not anticipate people writing functionality better relegated to actual plugins in Lua or Python scripts.

(The presence of circular locking also wasn't anticipated, which is how we arrived at the current status quo).

The entire scripting system needs a re-think and re-implementation, so we opt for keeping the current system as-is (for lack of a working alternative) but won't add any larger changes to it for the time being.

PR #9571, opened 2023-09-08
obs-ffmpeg: Allow the user to set metadata

Red-M:
Adds parsing for metadata to pass along to ffmpeg for stream outputs

### Description
This allows a user to provide options in the muxer settings under the recording tab using the ffmpeg output to set metadata on the output stream. Please note that I've only tested this with mpegts using both the mpegts and rtp_mpegts muxers with the format set to mpeg2video.

### Motivation and Context
Implements a feature that was sort of requested on the forums and that I needed.
This allows setting [the metadata options in mpegts](https://ffmpeg.org/ffmpeg-formats.html#Options-14) but also applies to other codecs that allow metadata. The use case I needed this for was easily compositing streams in OBS while feeding the output via the recording output with FFMpeg to a DVB converter to show video and audio on TVs hooked into a coax network.

One of the forum posts which was trying to use such a feature was [here](https://obsproject.com/forum/threads/problem-setting-metadata-in-mpeg-ts.100770/).

Another situation is allowing metadata to be added to files that are recorded to disk or sent across the network for other reasons as [the metadata standard is somewhat open to whatever the user wants to record](https://wiki.multimedia.cx/index.php/FFmpeg_Metadata).

### How Has This Been Tested?
I tested this using the above mentioned muxers and format with the receiver being a device to convert to DVB-T. I could confirm that both the service_name and service_provider were being set correctly and read correctly on the receiver's end.

### Types of changes

### Checklist:

I've tried to format with `clang-format-15` (also 14) but the versions I have access to in Debian produce errors with the project's provided config file. I've commented those options out and the CI seems to like the output.

Let me know of any documentation that needs to be provided and I can add that to the PR.

### DIFF
--- a/plugins/obs-ffmpeg/obs-ffmpeg-output.c
+++ b/plugins/obs-ffmpeg/obs-ffmpeg-output.c
@@ -421,6 +421,8 @@ static inline bool open_output_file(struct ffmpeg_data *data)
 {
 	const AVOutputFormat *format = data->output->oformat;
 	int ret;
+	const char *metadata_prefix = "metadata_";
+	const size_t len_metadata_prefix = strlen(metadata_prefix);
 
 	AVDictionary *dict = NULL;
 	if ((ret = av_dict_parse_string(&dict, data->config.muxer_settings, "=",
@@ -455,6 +457,27 @@ static inline bool open_output_file(struct ffmpeg_data *data)
 			av_dict_free(&dict);
 			return false;
 		}
+	}
+
+	// Handle metdata if requested by the user.
+	// We need to do this before the header write
+	if (av_dict_count(dict) > 0) {
+		struct dstr meta_str = {0};
+		AVDictionaryEntry *meta_entry = NULL;
+		while ((meta_entry = av_dict_get(dict, metadata_prefix,
+						 meta_entry,
+						 AV_DICT_IGNORE_SUFFIX))) {
+			if (strlen(meta_entry->key) > len_metadata_prefix) {
+				dstr_catf(&meta_str, "\n\t%s=%s",
+					  meta_entry->key, meta_entry->value);
+				av_dict_set(&data->output->metadata,
+					    (char *)(meta_entry->key +
+						     len_metadata_prefix),
+					    meta_entry->value, 0);
+			}
+		}
+		blog(LOG_INFO, "Using metadata settings: %s", meta_str.array);
+		dstr_free(&meta_str);
 	}
 
 	ret = avformat_write_header(data->output, &dict);


Lain-B:
The motivation and context section needs to be the full details behind your personal use case, and if you want, you can also include other context such as links to discussions and things like that, but we really need for full context.

Red-M:
I've updated the motivation and context section, happy to provide more information if required.

Red-M:
@Lain-B Can this please be reviewed again?

plugins/obs-ffmpeg/obs-ffmpeg-output.c
```diff
@@ -457,6 +459,27 @@ static inline bool open_output_file(struct ffmpeg_data *data)
...
+			if (strlen(meta_entry->key) > len_metadata_prefix) {
+				dstr_catf(&meta_str, "\n\t%s=%s",
+					  meta_entry->key, meta_entry->value);
+				av_dict_set(&data->output->metadata,
+					    (char *)(meta_entry->key +
+						     len_metadata_prefix),
+					    meta_entry->value, 0);
+			}
+		}
+		blog(LOG_INFO, "Using metadata settings: %s", meta_str.array);
```
notr1ch, line 479: This shouldn't log at all if no valid metadata was found. If muxer settings are present but none match `metadata`_, this leaves `meta_str.array` pointing to `NULL` which is technically UB when used in a format string.

plugins/obs-ffmpeg/obs-ffmpeg-output.c
```diff
@@ -457,6 +459,27 @@ static inline bool open_output_file(struct ffmpeg_data *data)
 		}
 	}
 
+	// Handle metdata if requested by the user.
```
notr1ch, line 462:
```suggestion
	// Handle metadata if requested by the user.
```

notr1ch (changes requested):
I'm not entirely convinced about using the muxer settings for metadata purposes, but it technically makes sense as the muxer is usually responsible for metadata in the output.

Is there a reason this has to go after the `avio_open2` call? I think it would make more sense logically to put it next to the muxer options in the same block.

Red-M:
> Is there a reason this has to go after the `avio_open2` call? I think it would make more sense logically to put it next to the muxer options in the same block.

From memory when I originally wrote this and please excuse me if I use the wrong terms but it has been a few years since I wrote this, its because the metadata for the output needs to be created and set before the header is made but after `&data->output->pb` is set, otherwise the audio and/or video streams can be `NULL` so the metadata stream won't reference them.

I can take another read of exactly why if you'd like.

plugins/obs-ffmpeg/obs-ffmpeg-output.c
Red-M: I can add a `NULL` check but I've also noticed that there is [a similar piece of code above this](https://github.com/obsproject/obs-studio/blob/master/plugins/obs-ffmpeg/obs-ffmpeg-output.c#L398).

Red-M: Upon further inspection, there is another `dstr` that isn't checked for being `NULL` before a formatted string print via `blog()`.

What would you like me to do?

plugins/obs-ffmpeg/obs-ffmpeg-output.c
Red-M: Will be fixed in my next push.

PR #9572, opened 2023-09-08
cmake: Make QR-Code-Generator finder debug aware.

admshao:
### Description

Similar to NTV2 lib files we ship 2 different files for the Release and Debug builds so this makes the finder aware of the changes and selects the correct file based on the build type.

v2: Updated to make the changes Windows only.

### Motivation and Context

I want to be able to build OBS Studio in Debug mode.

### How Has This Been Tested?

Windows:
After build OBS-Deps with the following changes: [obs-deps - PR#210](https://github.com/obsproject/obs-deps/pull/210)
I built OBS Studio with (Debug, Release, RelWithDebInfo, MinSize) and all launched OBS Successfully 

Macos, Linux:
Needs Tests.

### Types of changes

### Checklist:

### DIFF
--- a/cmake/finders/FindLibqrcodegencpp.cmake
+++ b/cmake/finders/FindLibqrcodegencpp.cmake
@@ -20,14 +20,18 @@ This module sets the following variables:
   True, if all required components and the core library were found.
 ``Libqrcodegencpp_VERSION``
   Detected version of found Libqrcodegencpp libraries.
+``Libqrcodegencpp_LIBRARIES``
+  Libraries needed to link to Libqrcodegencpp.
 
 Cache variables
 ^^^^^^^^^^^^^^^
 
 The following cache variables may also be set:
 
-``Libqrcodegencpp_LIBRARY``
-  Path to the library component of Libqrcodegencpp.
+``Libqrcodegencpp_LIBRARY_RELEASE``
+  Path to the library component of Libqrcodegencpp in non-debug configuration.
+``Libqrcodegencpp_LIBRARY_DEBUG``
+  Optional path to the library component of Libqrcodegencpp in debug configuration.
 ``Libqrcodegencpp_INCLUDE_DIR``
   Directory containing ``qrcodegen.hpp``.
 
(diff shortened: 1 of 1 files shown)


tytan652:
Rather than modifying the whole finder behavior just for Windows, try to replace `Libqrcodegencpp::Libqrcodegencpp` (our finder) by `qrcodegencpp::qrcodegencpp` (available CMake package) for Windows-only which should use the debug build automatically.

The finder is only meant to avoid having Ubuntu causing issue.

PatTheMav:
This PR is superfluous, as Windows builds should never use the finder and using a `d` suffix for debug variants is a Windows-specific thing.

RytoEX:
> This PR is superfluous, as Windows builds should never use the finder and using a `d` suffix for debug variants is a Windows-specific thing.

How is this superfluous? obs-studio cannot be built in Debug even with both qrcodegencpp.lib and qrcodegencppd.lib in the deps/lib folder. This is the same problem that we had with AJA and VPL, so why aren't changes necessary?

PatTheMav:
The library is called `qrcodegencpp` for Windows and macOS and is shipped as a CMake package with a `qrcodegencpp-targets-release.cmake` and `qrcodegencpp-targets-debug.cmake` file (with the associated `obs-deps` PR merged).

`libqrcodegencpp` and its finder can only be used on Linux.

RytoEX:
> The library is called `qrcodegencpp` for Windows and macOS and is shipped as a CMake package with a `qrcodegencpp-targets-release.cmake` and `qrcodegencpp-targets-debug.cmake` file (with the associated `obs-deps` PR merged).
> 
> `libqrcodegencpp` and its finder can only be used on Linux.

Then something isn't working correctly, because even with those targets files, I cannot build in Debug on Windows.

PatTheMav:
https://github.com/obsproject/obs-websocket/blob/ec2cdc84759d737e8612fdd0b3c3fcec6ac89761/CMakeLists.txt#L21

Needs to be changed into:

```
if(OS_MACOS OR OS_WINDOWS)
  find_package(qrcodegencpp REQUIRED)
else()
  find_package(Libqrcodegencpp REQUIRED)
endif()
```

Alternatively the finder needs to be updated to work as a drop-in replacement when the CMake package cannot be found, which means changing the name from "FindLibqrcodegencpp" to "FindQrcodegencpp" and ensure the target names match as well.

tytan652:
> Alternatively the finder needs to be updated to work as a drop-in replacement when the CMake package cannot be found,

The name was made different since there was zero guarantee that they work the same way (e.g. behavior if a version is put as find_package argument).

Maybe if `qrcodegencpp::qrcodegencpp` was not found (QUIET in find_package), fallback to `Libqrcodegencpp::Libqrcodegencpp`.

> which means changing the name from "FindLibqrcodegencpp" to "FindQrcodegencpp" and ensure the target names match as well.

"FindQrcodegencpp" does not work since it's awaiting for a `Qrcodegencpp` (uppercase Q) target.

RytoEX:
Going to go with #9601 instead. Apparently, the Finder should never have been invoked on Windows and was only being invoked due to an oversight.

PR #9575, opened 2023-09-09
UI: Disable edit transform if item is locked

cg2121:
### Description
The transform window was still accessible if the scene item was locked.

### Motivation and Context
You shouldn't be able to edit the transform of a locked item.

### How Has This Been Tested?
Locked/unlocked scene items to make sure the transform window was disabled.

### Types of changes

### Checklist:

### DIFF
--- a/UI/window-basic-main.cpp
+++ b/UI/window-basic-main.cpp
@@ -8566,8 +8566,11 @@ void OBSBasic::UpdateEditMenu()
 	}
 	const bool canTransformSingle = videoCount == 1 && totalCount == 1;
 
+	OBSSceneItem curItem = GetCurrentSceneItem();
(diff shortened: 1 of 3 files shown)


gxalpha:
I’d want @Warchamp7’s opinion on this, because as-is this is by design. IMO locking the preview should prevent accidental changes, but using the Edit Transform dialog isn’t accidental.

GeorgesStavracas (approved):
This seems fine

PatTheMav:
I also agree with this change, it mimics learned behaviour from other pro tools where locking a layer usually locks all input fields of associated property widgets and windows as well.

@cg2121 what happens if I lock an item with an associated transform window open?

cg2121:
If the window is opened, when the item is locked, the window is updated. This is what the lockSignal does in the code.

PR #9582, opened 2023-09-10
UI: Include sceneitem locked state in copy-paste

gxalpha:
### Description

When copy-pasting a source, so far it would always be unlocked. We already copy other attributes like whether the source is visible, so it makes sense to also include the locked state in there.

### Motivation and Context

Noticed that the status of whether a source is visible is copy-pasted, but not whether it's locked.

### How Has This Been Tested?

macOS 14
Copy-pasted with "Paste (Existing)" and "Paste (Duplicate)" and confirmed that a locked source would be locked when pasted.
Added new and existing sources via the dialog to confirm they were still always unlocked (since that code was modified).

### Types of changes

### Checklist:

### DIFF
--- a/UI/window-basic-main.cpp
+++ b/UI/window-basic-main.cpp
@@ -9976,6 +9976,7 @@ void OBSBasic::on_actionCopySource_triggered()
 		copyInfo.blend_method = obs_sceneitem_get_blending_method(item);
 		copyInfo.blend_mode = obs_sceneitem_get_blending_mode(item);
 		copyInfo.visible = obs_sceneitem_visible(item);
+		copyInfo.locked = obs_sceneitem_locked(item);
 
 		clipboard.push_back(copyInfo);
 	}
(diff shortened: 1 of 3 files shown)


GeorgesStavracas (approved):
Makes sense from a UI / UX perspective

PatTheMav:
While I'm not inherently opposed to this change, I'd be amiss not to notice that the current behaviour mimics that of Photoshop and other layer-based applications I'm aware of, where copying a locked layer creates a duplicate layer that is unlocked automatically (the purpose of the copy being able to make changes).

At least that's what I expect when copying a locked layer/element in an application mimicking that paradigm.

gxalpha:
bump on the above @Warchamp7 @GeorgesStavracas

Warchamp7:
> While I'm not inherently opposed to this change, I'd be amiss not to notice that the current behaviour mimics that of Photoshop and other layer-based applications I'm aware of, where copying a locked layer creates a duplicate layer that is unlocked automatically (the purpose of the copy being able to make changes).
> 
> At least that's what I expect when copying a locked layer/element in an application mimicking that paradigm.

I am generally in agreement with Pat here. I think in the majority of scenarios if not all, when a scene item is copied and pasted it's because someone wants to manipulate the copy. The current behaviour of not copying the lock status is therefore preferred.

PR #9584, opened 2023-09-11
UI: Replace SourceTreeSubItemCheckBox class with Qt property

gxalpha:
### Description

Replaces the empty `SourceTreeSubItemCheckBox` class which is only used for styling with a property.
IIRC @Warchamp7 wanted `themeID` to go the way of the dodo in favor of properties set to true, so I'm using this syntax. Needs confirmation though.

### Motivation and Context

Having classes just for QSS styling seems a bit overkill to me when it's possible to achieve the same result without them.

### How Has This Been Tested?

macOS 14
Tested in all themes that the expand checkbox still shows the icons, but when the group is expanded and when it's not.

### Types of changes

### Checklist:

### DIFF
--- a/UI/data/themes/Acri.qss
+++ b/UI/data/themes/Acri.qss
@@ -1291,24 +1291,24 @@ OBSHotkeyWidget QPushButton {
 
 /* Sources List Group Collapse Checkbox */
 
-SourceTreeSubItemCheckBox {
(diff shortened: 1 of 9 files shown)


GeorgesStavracas (approved):
Seems fine

Warchamp7 (approved):

PR #9594, opened 2023-09-14
libobs: Reuse existing memory when clearing async frame cache

derrod:
### Description

- If we hit the frame cache limit, reset the cache entries rather than freeing them and their memory
- Only run the `clean_cache` process once every second to give the cache enough time to fill up again after a reset

Sort of a newer take on 96c1a7652305784e4bf531933bc82535adcaf320 to avoid re-allocating a lot of memory if we have to invalidate the async frame cache, but keeps `clean_cache` to free unneeded memory allocations after a while.

### Motivation and Context

Want to avoid too many free/allocation loops if we have to churn through lots of frames at once (e.g. network issues, or after OBS was paused in a debugger)

### How Has This Been Tested?

Tested with additional logging in place to verify that memory is no longer allocated/free'd once we hit the cache size limit, and then is cleared once it actually isn't needed anymore.

Also tested an alternative of simply counting async frames since last clear and running clear once it hit `MAX_ASYNC_FRAMES` (https://github.com/derrod/obs-studio/commit/697a0a5bfd7fb248eeb2203da0627d5426050a4c), this also worked but using time seemed more like a more reliable option e.g. if the source is either a very low or very high frame rate.

### Types of changes

### Checklist:

### DIFF
--- a/libobs/obs-source.c
+++ b/libobs/obs-source.c
@@ -3554,11 +3554,16 @@ static inline void free_async_cache(struct obs_source *source)
 }
 
 #define MAX_UNUSED_FRAME_DURATION 5
+#define MINIMUM_CLEAN_INTERVAL 1000000000
 
 /* frees frame allocations if they haven't been used for a specific period
  * of time */
 static void clean_cache(obs_source_t *source)
 {
+	uint64_t time = obs_get_video_frame_time();
+	if (time - source->async_last_clean < MINIMUM_CLEAN_INTERVAL)
+		return;
+
 	for (size_t i = source->async_cache.num; i > 0; i--) {
 		struct async_frame *af = &source->async_cache.array[i - 1];
 		if (!af->used) {
@@ -3568,6 +3573,8 @@ static void clean_cache(obs_source_t *source)
 			}
 		}
 	}
+
+	source->async_last_clean = time;
 }
 
 #define MAX_ASYNC_FRAMES 30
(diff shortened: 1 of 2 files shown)


norihiro:
I have not tested this PR but I think `obs_get_video_frame_time()` would be sufficient and less overhead than `os_gettime_ns()`.

derrod:
@norihiro Yeah that sounds like a better option, I'll update it to use that instead.

@WizardCM I don't think this is a bug fix, the existing code works as intended, this is an optimisation to avoid doing unnecessary memory allocations.

libobs/obs-source.c
```diff
@@ -3579,11 +3586,17 @@ cache_video(struct obs_source *source, const struct obs_source_frame *frame)
...
 		source->last_frame_ts = 0;
-		pthread_mutex_unlock(&source->async_mutex);
-		return NULL;
+		source->async_last_clean = obs_get_video_frame_time();
+
+		da_clear(source->async_frames);
+		for (size_t i = 0; i < source->async_cache.num; i++) {
+			source->async_cache.array[i].used = false;
+			source->async_cache.array[i].unused_count = 0;
+		}
```
Lain-B, line 3599: The problem with doing this is that the point of the `used` member is that the frame is in use somewhere. For example, an async delay filter will use frames while they're in the cache.

How can one reproduce the original performance issue where it releases all the frames?

derrod: Pause OBS in a debugger for a bit while as async source is feeding it frames, then continue.

Lain-B: Well I meant like a real-world example of how this ends up happening. Because the max queued async frame count should be fairly big.

Also the async delay filter case is still a situation we probably have to think about. You could end up overwriting frames that are still technically referenced inside of an async delay filter. That's technically why the old frames are just released. The idea is "this situation *shouldn't* happen very often, but if it does, then just release the frames because they're probably in use somewhere." So that situation might need to be taken into account. I'm not ruling out this code as an optimization, I'm mostly just noting that's the situation that you may end up in if you force frames to an unused state.

derrod: As far as real-world examples go, network streams are on thing. e.g. if you run into buffering or stalls and then a ton of frames come in at once. I'll take a look at the async delay filter case, but I fear that taking that into account may complicate things to the point where it's not really worth it to work around that in this PR.

Lain-B: Yea, I know. I'm not sure what to do either.

derrod: This was mostly spawned out of a discussion with R1CH and his complaints about memory churn when we are in a catch-up scenario, so it's not super important. I'm sure we can improve upon the current situation, but I don't think it's worth keeping this PR around for now.

Issue #9574, opened 2023-09-09
Target Usage settings not retained when upgrading


hoga2443:
### Operating System Info

Windows 11

### OBS Studio Version

30.0.0-beta2

### OBS Studio Log URL

https://obsproject.com/logs/Pbtcp0Jx0Tt4EDyn

### Expected Behavior

When upgrading from OBS 29 to OBS 30, encoder settings, such as Target Usage, should be retained.

### Current Behavior

Target Usage was reset from the assigned setting "Fast" to "TU1: Slowest (Best Quality)"

### Steps to Reproduce

1.Set an Encoder Setting Target Usage to other than Slowest (Best Quality) in OBS 29.
2.Install OBS 30

### Anything else we should know?

Other encoder settings may have failed to migrate successfully as well.  I upgraded from 29 to 30, downgraded to 29, and then upgraded to 30 again.  I can't say for certain whether the setting was changed in one of the two upgrades or in the one downgrade.

gxalpha:
Confirmed. There are typos and missing `== 0`'s in the migration logic.

Lain-B:
Should now be fixed

Lain-B:
Ah okay, a misunderstanding on my part then.

RytoEX:
This *may* be somewhat intentional. While it is not reflected in the UI, the Target Usage is silently updated in the in-memory `obs_data_t *settings` data for the QSV session and a corresponding Target Usage value based the previously selected value will be used for the encoder session. I'm not sure if it was done this way on purpose to avoid a breaking change for users who downgrade OBS for some reason, or if it was simply an oversight.

As [I commented on PR 9656](https://github.com/obsproject/obs-studio/pull/9596#issuecomment-1720270859), this could be fixed by calling `update_targetusage` from `rate_control_modified`, and it will update/overwrite the existing user settings that way (though it is invoked twice, and I didn't finish figuring that out). If I recall correctly, because the settings are updated as soon as the Settings UI is loaded, this would make downgrading no longer retain that setting, but I think that's already somewhat the case anyway for users who have saved settings with the new target usage values.

hoga2443:
The issue appears unresolved in RC 1.  I just updated from Beta 3 to RC 1.  Target Usage was changed from TU6: Faster to TU1: Slowest (Best Quality).  If not resolved, I'm concerned a significant number of users will have recordings corrupted, as I did, if their systems cannot handle encoding at TU1.

gxalpha:
[...]

PR #9596, opened 2023-09-14
obs-qsv11: Fix target usage migration string comparisons

gxalpha:
### Description

Fixes problems with the astrcmpi logic, as well as a typo that said "vertfast" instead "veryfast".
Also makes the method a void because it always returns true and isn't checked anyways.

### Motivation and Context

Was found while investigating https://github.com/obsproject/obs-studio/issues/9574.

### How Has This Been Tested?

Hasn't.

### Types of changes

### Checklist:

### DIFF
--- a/plugins/obs-qsv11/obs-qsv11.c
+++ b/plugins/obs-qsv11/obs-qsv11.c
@@ -328,33 +328,31 @@ static bool update_enhancements(obs_data_t *settings)
 	return true;
 }
 
-static bool update_targetusage(obs_data_t *settings)
+static void update_targetusage(obs_data_t *settings)
 {
 	const char *target_usage =
 		obs_data_get_string(settings, "target_usage");
 
-	if ((astrcmpi(target_usage, "veryslow") ||
-	     astrcmpi(target_usage, "quality")) == 0)
+	if (astrcmpi(target_usage, "veryslow") == 0 ||
+	    astrcmpi(target_usage, "quality") == 0)
 		obs_data_set_string(settings, "target_usage",
 				    "TU1: Slowest (Best Quality)");
 	else if (astrcmpi(target_usage, "slower") == 0)
 		obs_data_set_string(settings, "target_usage", "TU2: Slower");
 	else if (astrcmpi(target_usage, "slow") == 0)
 		obs_data_set_string(settings, "target_usage", "TU3: Slow");
-	else if ((astrcmpi(target_usage, "medium") ||
-		  astrcmpi(target_usage, "balanced")) == 0)
+	else if (astrcmpi(target_usage, "medium") == 0 ||
+		 astrcmpi(target_usage, "balanced") == 0)
 		obs_data_set_string(settings, "target_usage",
 				    "TU4: Balanced (Medium Quality)");
 	else if (astrcmpi(target_usage, "fast") == 0)
 		obs_data_set_string(settings, "target_usage", "TU5: Fast");
 	else if (astrcmpi(target_usage, "faster") == 0)
 		obs_data_set_string(settings, "target_usage", "TU6: Faster");
-	else if ((astrcmpi(target_usage, "vertfast") ||
-		  astrcmpi(target_usage, "speed")) == 0)
+	else if (astrcmpi(target_usage, "veryfast") == 0 ||
+		 astrcmpi(target_usage, "speed") == 0)
 		obs_data_set_string(settings, "target_usage",
 				    "TU7: Fastest (Best Speed)");
-
-	return true;
 }
 
 static bool rate_control_modified(obs_properties_t *ppts, obs_property_t *p,


RytoEX:
Was looking into this myself. While these changes are fine, I think the core issue is that `update_targetusage` is only called via `update_settings`, which means that the target usage is not actually updated until that is invoked. If you start an encoder session, then `update_settings` is called, which calls `update_params`, which calls `update_targetusage` and internally sets the appropriate TU but doesn't change the UI. `update_latency`, `update_enhancements`, and `update_ratecontrol` are all called in `rate_control_modified` (those first two are also called in `update_params`). If we call `update_targetusage` from `rate_control_modified` like the others, it works fine in that the UI value is now updated.

plugins/obs-qsv11/obs-qsv11.c
```diff
@@ -328,33 +328,31 @@ static bool update_enhancements(obs_data_t *settings)
...
-static bool update_targetusage(obs_data_t *settings)
+static void update_targetusage(obs_data_t *settings)
 {
 	const char *target_usage =
 		obs_data_get_string(settings, "target_usage");
 
-	if ((astrcmpi(target_usage, "veryslow") ||
-	     astrcmpi(target_usage, "quality")) == 0)
+	if (astrcmpi(target_usage, "veryslow") == 0 ||
+	    astrcmpi(target_usage, "quality") == 0)
```
tytan652, line 337:
```suggestion
	if ((astrcmpi(target_usage, "veryslow") == 0) ||
	    (astrcmpi(target_usage, "quality") == 0))
```

plugins/obs-qsv11/obs-qsv11.c
```diff
@@ -328,33 +328,31 @@ static bool update_enhancements(obs_data_t *settings)
...
 		obs_data_set_string(settings, "target_usage",
 				    "TU1: Slowest (Best Quality)");
 	else if (astrcmpi(target_usage, "slower") == 0)
 		obs_data_set_string(settings, "target_usage", "TU2: Slower");
 	else if (astrcmpi(target_usage, "slow") == 0)
 		obs_data_set_string(settings, "target_usage", "TU3: Slow");
-	else if ((astrcmpi(target_usage, "medium") ||
-		  astrcmpi(target_usage, "balanced")) == 0)
+	else if (astrcmpi(target_usage, "medium") == 0 ||
+		 astrcmpi(target_usage, "balanced") == 0)
```
tytan652, line 345:
```suggestion
	else if ((astrcmpi(target_usage, "medium") == 0) ||
		 (astrcmpi(target_usage, "balanced") == 0))
```

plugins/obs-qsv11/obs-qsv11.c
```diff
@@ -328,33 +328,31 @@ static bool update_enhancements(obs_data_t *settings)
...
 		obs_data_set_string(settings, "target_usage",
 				    "TU4: Balanced (Medium Quality)");
 	else if (astrcmpi(target_usage, "fast") == 0)
 		obs_data_set_string(settings, "target_usage", "TU5: Fast");
 	else if (astrcmpi(target_usage, "faster") == 0)
 		obs_data_set_string(settings, "target_usage", "TU6: Faster");
-	else if ((astrcmpi(target_usage, "vertfast") ||
-		  astrcmpi(target_usage, "speed")) == 0)
+	else if (astrcmpi(target_usage, "veryfast") == 0 ||
+		 astrcmpi(target_usage, "speed") == 0)
```
tytan652, line 353:
```suggestion
	else if ((astrcmpi(target_usage, "veryfast") == 0) ||
		 (astrcmpi(target_usage, "speed") == 0))
```

tytan652 (changes requested):
`(astrcmpi || astrcmpi) == 0` is really bad because the returned 0 is casted as false but if one of the `astrcmpi` returns something different than 0 we get a cast to true and so even if one if them return 0/false the other returns true and so the `||` makes a true from both and is not equal to 0

derrod:
I'm not sure what you're saying, the code is doing `astrcmpi(...) == 0 || astrcmpi(...) == 0`, not `(astrcmpi || astrcmpi) == 0`.

tytan652:
Priorities in this situation are messy, feel free to resolve them.

gxalpha:
@tytan652 For reference, C++ operator precedence can be checked here: https://en.cppreference.com/w/cpp/language/operator_precedence
Equality is group 10, logical OR is group 14, so equality will be evaluated first.

@RytoEX do you want me to add another call to `update_targetusage` to this PR or do you consider that to be standalone? I could unlink the open issue in the latter case.

tytan652 (approved):
The function looks good to me

RytoEX:
> @tytan652 For reference, C++ operator precedence can be checked here: https://en.cppreference.com/w/cpp/language/operator_precedence Equality is group 10, logical OR is group 14, so equality will be evaluated first.
> 
> @RytoEX do you want me to add another call to `update_targetusage` to this PR or do you consider that to be standalone? I could unlink the open issue in the latter case.

Let's keep the scope small and just fix the obvious errors here. We can handle adding `update_targetusage` to `rate_control_modified` in a separate PR.

PR #9597, opened 2023-09-15
cmake: Allow CMake 3.20 and 3.21 to be used to build OBS

Conan-Kudo:
### Description
There does not appear to be anything preventing it from building on CMake 3.20, so let's permit it so that it can still be built for Fedora Extra Packages for Enterprise Linux 9.

### Motivation and Context
CentOS Stream 9 (and thus Red Hat Enterprise Linux 9) only have CMake 3.20.

### How Has This Been Tested?
Built OBS Studio 30~beta3 successfully in EPEL 9 (CMake 3.20) and Fedora 39+40 (CMake 3.27)

### Types of changes

### Checklist:

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -1,4 +1,4 @@
-cmake_minimum_required(VERSION 3.22...3.25)
+cmake_minimum_required(VERSION 3.20...3.25)
 
 if(CMAKE_HOST_SYSTEM_NAME MATCHES "(Darwin)" OR OBS_CMAKE_VERSION VERSION_GREATER_EQUAL 3.0.0)
   include("${CMAKE_CURRENT_SOURCE_DIR}/cmake/common/bootstrap.cmake" NO_POLICY_SCOPE)
--- a/UI/CMakeLists.txt
+++ b/UI/CMakeLists.txt
@@ -1,4 +1,4 @@
-cmake_minimum_required(VERSION 3.22...3.25)
+cmake_minimum_required(VERSION 3.20...3.25)
 
 legacy_check()
 
--- a/UI/frontend-plugins/aja-output-ui/CMakeLists.txt
+++ b/UI/frontend-plugins/aja-output-ui/CMakeLists.txt
@@ -1,4 +1,4 @@
-cmake_minimum_required(VERSION 3.22...3.25)
+cmake_minimum_required(VERSION 3.20...3.25)
 
 legacy_check()
 
--- a/UI/frontend-plugins/decklink-captions/CMakeLists.txt
+++ b/UI/frontend-plugins/decklink-captions/CMakeLists.txt
@@ -1,4 +1,4 @@
-cmake_minimum_required(VERSION 3.22...3.25)
+cmake_minimum_required(VERSION 3.20...3.25)
 
 legacy_check()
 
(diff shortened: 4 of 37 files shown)


RytoEX:
Is there a reason you can't patch this on the build system? The reason we chose 3.22 is that that is the version provided by Ubuntu 22.04, which is our current minimum supported OS where we do not control the version of CMake available on builders (such as with PPA). We don't like to encourage versions older than that, and we use our minimum as a baseline for which CMake features we can use.

Conan-Kudo:
I can, I just have a policy of trying to upstream everything I do unless it's absolutely not possible.

RytoEX:
> I can, I just have a policy of trying to upstream everything I do unless it's absolutely not possible.

It's _technically_ possible. It doesn't exactly hurt anything to make just this change, though it is one more thing to keep track of, and I'm wary of setting precedent.

We don't officially provide support outside of:
* Windows 10 [1]
* Windows 11 [1]
* macOS 11+ [1]
* Ubuntu 22.04+ [2]
* Flatpak

The oldest Ubuntu LTS that we support tends to have the oldest libraries/tools of those, so that's usually our baseline.

As those are our primary concerns, we tend to align our code, CMake included, with that list. For other platforms, such as FreeBSD, Debian, CentOS, or Fedora, we tend to prefer having packagers patch on their build systems for their own platform requirements unless it is prohibitively difficult or impossible to do so. Otherwise, we're at the mercy of everyone needing their own changes upstreamed because of their constraints.

While this is a seemingly innocuous change, I don't necessarily want to set precedent if someone with much harsher constraints asks for a similar change (such as only having CMake 3.16 or older). May have to discuss with @PatTheMav . As things stand, our CMake presets and all CMake code are currently written on the assumption that 3.22 is available.

1: Dev tools are somewhat determined by what the GitHub Actions runners provide, but we can otherwise have our own requirements/recommendations for these platforms.
2: Typically, whatever the PPA builder has available dictates what we do here.

PatTheMav:
I agree with @RytoEX - we specifically increased the lower CMake boundary to keep the (already complex) CMake code base in check and we intend to increase this lower boundary regularly (if it weren't for Ubuntu we'd be requiring at least CMake 3.24 and soon probably CMake 3.27). This also allows us to get rid of hacks and workarounds that were required in older CMake versions and have been replaced by proper first-party support in it, which reduces our maintenance burden.

This also serves as a "feature check" because Linux distributions that do not support at least CMake 3.22 tend to be quite outdated and the packages they provide (and we depend on) are as well and - again with our maintenance burden in mind - we choose not to support versions of our dependencies that are sometimes a full major version behind what we have available on Windows or macOS.

_If_ lowering the boundary does not seem to break anything, packagers are free to patch this on their end and hope for the best, but it is not something that we want to bear the support burden for.

As long as distributions treat GUI applications like system-level libraries or command-line utilities, this problem will persist (and tools like Flatpak exist to solve that specific issue).

Conan-Kudo:
This is no longer needed, as CMake was upgraded to 3.26.5 in CentOS Stream 9: https://kojihub.stream.centos.org/koji/buildinfo?buildID=40444

Since this means RHEL 9.4 will have CMake 3.26.5 or higher, the problem is short-term and I'll carry it downstream until I don't need it anymore.

PR #9601, opened 2023-09-16
cmake: Update Find module for qrcodegen and fix configuration fallbacks on Windows

PatTheMav:
### Description
Updates the Find module for qrcodegen to use the same overall name and target names as the CMake package generated by `obs-deps` for Windows and macOS.

Also adds explicit configuration mapping to fix potential build issues on Windows due to ABI breakage between Debug and non-Debug builds on Windows.

### Motivation and Context
qrcodegen is built on obs-deps for macOS and Windows, with the generated CMake package calling the dependency qrcodegen (not libqrcoden) and associated target qrcodegen::qrcodegen.

This change updates the finder (required for Linux) to create the same targets so consumers do not need to differentiate between different variations of the same dependency on Linux.

Windows is the only platform that has breaking ABI changes between Debug and non-Debug builds. If a package provides different libraries per configuration (but is non-exhaustive), CMake will fall back on the first found configuration.

Alphabetically this will make CMake always fall back to the "Debug" configuration, which will thus break on all other configurations on Windows.

With explicit maps, any other available non-Debug configuration will be used for Release, RelWithDebInfo, or MinSizeRel, to avoid this from happening.

Needs to be merged with a submodule update of `obs-websocket` after merging https://github.com/obsproject/obs-websocket/pull/1167.
Requires https://github.com/obsproject/obs-deps/pull/214 to be merged and released.

### How Has This Been Tested?
Tested on Windows 11, macOS 13.5.2, and Ubuntu 23.04.

### Types of changes

### Checklist:


RytoEX (approved):
Works fine on my Windows 11 machine when building in Debug.

PR #9609, opened 2023-09-19
obs-qsv11: Remove QSV D3D9 code

RytoEX:
### Description

obs-qsv11: Remove D3D9 allocator

On Windows 10 and up, D3D11 should never fail, so the D3D9 code should no longer be possible to hit and should no longer be needed.

Revert "obs-qsv11: Use d3d9 allocator on Win7"

This reverts commit https://github.com/obsproject/obs-studio/commit/b276b1633e7c37deec4e8a56dec65e4eb36bfc1e.

----

obs-qsv11: Remove D3D9 fallback

On Windows 10 and up, D3D11 should never fail, so the D3D9 code should no longer be possible to hit. As far as I can tell, this code was mostly for Windows 7/8/8.1 and it was part of the initial implementation. It should no longer be needed.

The first commit is just reverting the "Use d3d9 allocator" commit (b276b1633e7c37deec4e8a56dec65e4eb36bfc1e), though I had to do this manually because the code has diverged a bit since then.

The second commit is removing a D3D9 fallback that _should_ no longer be possible to hit. However, it was not part of the "Use d3d9 allocator" commit, which is why I am removing it separately here.

### Motivation and Context

Want cleaner code. Want to get rid of unused code.

### How Has This Been Tested?

Built and tested on Windows 11.

### Types of changes

### Checklist:


kkartaltepe (approved):
Consider removing the DX11 defines as well and there is a remaining DX9 define left.

RytoEX:
I force pushed a single change to add back a check for DX9_D3D because it wasn't actually part of the original D3D9 allocator commit, and I want to keep my manual revert commit scoped to just removing those changes. Will probably merge this as-is without additional cleanups. We can do more cleanup in a separate PR that hopefully isn't as entangled as this one is.

PR #9620, opened 2023-09-21
linux-v4l2: add support for linear alpha blending

SludgePhD:
### Description

This copies the linear alpha settings UI and translation strings from `image-source` and applies the setting to the V4L2 `obs_source_frame`.

### Motivation and Context

I am supplying non-premultiplied alpha to OBS, but OBS seems to be converting the frame to premulitplied alpha, and then render it as if it *wasn't* premulitplied, which gives a very wrong look. With this setting enabled, it looks correct.

I am not sure why this is called "linear" alpha in OBS - to my knowledge alpha is always *linear*.

### How Has This Been Tested?

By trying it with a v4l2loopback source, AR24 format, non-premultiplied alpha. I have also tested that changes to the checkbox are visible in the preview immediately.

### Types of changes

### Checklist:

  - (could not find any)

The changes to the V4L2 source UI look like this:

screenshot_2023-09-21_23:08:15

### DIFF
--- a/plugins/linux-v4l2/v4l2-input.c
+++ b/plugins/linux-v4l2/v4l2-input.c
@@ -125,6 +125,7 @@ static void v4l2_prep_obs_frame(struct v4l2_data *data,
 
 	const enum video_format format = v4l2_to_obs_video_format(data->pixfmt);
 
+	frame->flags = OBS_SOURCE_FRAME_LINEAR_ALPHA;
 	frame->width = data->width;
 	frame->height = data->height;
 	frame->format = format;


tytan652:
Remove all the translations except the en-US one, those are done through Crowdin.

tytan652 (changes requested):
Remove all the translations except the en-US one, those are done through Crowdin.

gxalpha (changes requested):
Please do not modify any ini files besides en-US. Any and all changes to them will get lost, translations are provided via our Crowdin.

derrod:
Also, don't modify libobs here.

tytan652:
Also you are modifying the the source property UI, please provide screenshots like it was asked in PR template.

GeorgesStavracas:
Which devices provide images with an alpha channel?

SludgePhD:
> Which devices provide images with an alpha channel?

I don't know of any, I use this with v4l2loopback to transfer image data from another application.

Warchamp7:
> > Which devices provide images with an alpha channel?
> 
> I don't know of any, I use this with v4l2loopback to transfer image data from another application.

This sounds like a particularly niche use case that is specific to your workflow. Is there any reason you don't simply make an alpha-supporting copy of the linux-v4l2 source for your own usage?

SludgePhD:
I am not sure why correct handling of the alpha channel is considered a "niche" use case – why does the V4L2 source support pixel formats with alpha channel at all, if the only way they can be displayed incorrectly handles the alpha channel?

derrod:
Instead of making this an option we should probably just always add the flag like we do in win-dshow[^1] because there probably is no situation where ingested alpha wouldn't require this. That'll turn it into a one line change.

[^1]: https://github.com/obsproject/obs-studio/blob/e85b257562f19e30fd89bcebf14dd37ccfdf69f3/plugins/win-dshow/win-dshow.cpp#L570

Lain-B:
I'd just say enable it all the time and remove the option. Would make the PR more simple and I don't see a case where we would want it otherwise to be perfectly honest.

Lain-B (changes requested):
Requesting to be changed to just remove the option and make this always on with frame->flags.

SludgePhD:
Okay, I've updated the PR to make that change

GeorgesStavracas (approved):

PR #9621, opened 2023-09-22
obs-qsv11: Redo session data releasing for Linux

### LINKED ISSUE #9611: (OBS 30.0.0-beta3) QSV crashes OBS
### Operating System Info

Windows 11

### Other OS

_No response_

### OBS Studio Version

Other

30.0.0-beta3

### OBS Studio Log URL

https://obsproject.com/logs/J7tEszLSxTkcgswc

### OBS Studio Crash Log URL

https://obsproject.com/logs/xjjyUQiHEcOLD5b0

### Expected Behavior

Encoder shouldn't crash obs.

### Current Behavior

OBS crashes when you stop recording while running stream.

### Steps to Reproduce

1.Select qsv as encoder both on streaming and recording tab
2. Start streaming
3. Start recording
4. Stop recording
5. Crash!
...

This issue started with OBS 30.0 beta-3
**OBS 30.0 beta-2** works without issues no crashes.

kkartaltepe:
### Description

This reverts the changes to Windows where Release() was called every time, since we need share a single DX context across multiple encoders.

Instead introduce a Release2() function and some platform specific session data that will be passed to it. We use this to track the VA-API display and fd to release them at the right time. Leaking displays will also lead to cache pollution in the intel-media-driver crashing users so we cannot do that.

### Motivation and Context

fixes #9611
Avoids crashes in drivers due to caching invalid buffers if we just leak these instead.

### How Has This Been Tested?

Starting a streaming and recording session works fine on linux again.

Starting and stopping different kinds of encoding sessions works (except h265 which is broken on discrete intel GPUs due to a driver bug).

### Types of changes

### Checklist:


kkartaltepe:
I guess the h265 issue is only with debug asserts on, with release drivers it will run happily.

Lordmau5:
Tested the build from the actions, can confirm this doesn't crash anymore on my machine 🎉

plugins/obs-qsv11/common_utils.h
```diff
@@ -162,10 +162,12 @@ int GetFreeTaskIndex(Task *pTaskPool, mfxU16 nPoolSize);
 mfxStatus Initialize(mfxVersion ver, mfxSession *pSession,
 		     mfxFrameAllocator *pmfxAllocator, mfxHDL *deviceHandle,
 		     bool bCreateSharedHandles, bool dx9hack,
-		     enum qsv_codec codec); //vpl change
+		     enum qsv_codec codec, void **data); //vpl change
 
-// Release resources (device/display)
+// Release global shared resources  (device/display)
```
RytoEX, line 167:
```suggestion
// Release global shared resources (device/display)
```

plugins/obs-qsv11/common_utils.h
```diff
@@ -162,10 +162,12 @@ int GetFreeTaskIndex(Task *pTaskPool, mfxU16 nPoolSize);
...
 void Release();
+// Release per session resources
+void Release2(void *data);
```
RytoEX, line 170: I kind of dislike the `Release2` naming here because it implies that it replaces `Release` (this is how we have historically named other replacement functions that deprecate an older version). I'd prefer something a name that better reflects its purpose. Maybe `ReleaseSessionData`?

plugins/obs-qsv11/common_utils_linux.cpp
```diff
@@ -145,24 +149,34 @@ mfxStatus Initialize(mfxVersion ver, mfxSession *pSession,
...
-// Release resources (device/display)
-void Release()
-{
-	if (default_fd > 0)
-		close(default_fd);
+// Release shared global resources.
+void Release() {}
 
-	default_fd = -1;
+void Release2(void *data)
```
RytoEX, line 172: Could use a comment similar to the above.

RytoEX (changes requested):

plugins/obs-qsv11/common_utils.h
kkartaltepe: It should replace Release, the current usage of globals is kind of insane.

RytoEX: Hmm. That seems like this is an intermediate fix and that the more complete fix would be to refactor the Release calls/definition.

kkartaltepe: Not fixing windows is indeed an intermediate state, but whether anyone will ever fix windows to avoid globals is a question.

kkartaltepe: Renamed to avoid confusion

plugins/obs-qsv11/common_utils.h
```diff
@@ -162,10 +162,12 @@ int GetFreeTaskIndex(Task *pTaskPool, mfxU16 nPoolSize);
 mfxStatus Initialize(mfxVersion ver, mfxSession *pSession,
 		     mfxFrameAllocator *pmfxAllocator, mfxHDL *deviceHandle,
 		     bool bCreateSharedHandles, bool dx9hack,
-		     enum qsv_codec codec); //vpl change
+		     enum qsv_codec codec, void **data); //vpl change
 
-// Release resources (device/display)
+// Release global shared resources (device/display)
```
RytoEX, line 167: Does this actually release device/display now? I think the comment is outdated.

kkartaltepe: This is still correct, it does the same thing it always did.

RytoEX (approved):
Did some quick testing and this seems fine. We can revisit refactoring the Release calls in the future.

PR #9622, opened 2023-09-22
UI: Add theme data search paths

derrod:
### Description

Adds search paths for Qt `theme:` prefix upon initialising style for the first time so theme data can be referenced with relative paths anywhere within Qt (e.g. stylesheets, QUrl).

### Motivation and Context

While exploring options for making the appearance of OBS more customisable without requiring users to write their own theme files I encountered the issue that all image files use relative paths.  
If we do not load QSS from a path but instead a runtime generated stylesheet this won't work, so instead use a search path with custom prefix that will work regardless of where the file is loaded from.

### How Has This Been Tested?

Loaded up themes, confirmed they still work.

### Types of changes

### Checklist:

### DIFF
--- a/UI/data/themes/Acri.qss
+++ b/UI/data/themes/Acri.qss
@@ -166,7 +166,7 @@ QMenu::item:disabled {
 }
 
 QMenu::right-arrow {
-    image: url(./Dark/expand.svg);
+    image: url(theme:Dark/expand.svg);
 }
 
 /* Top Menu Bar Items */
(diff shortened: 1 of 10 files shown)


derrod:
Updated to remove the moving of image files to an `images/` subfolder, that would break themes using relative paths that some users might have. At least until we rework how themes work I'd avoid breaking this.

GeorgesStavracas (approved):

PR #9624, opened 2023-09-23
mac-capture: Show physical instead of logical pixels in SCK properties

gxalpha:
### Description

Currently, we show the logical "looks like" size of displays in the SCK properties, which gets confusing for high-DPI displays that most people have. It's more useful to show the actual pixel dimensions as that's what people are used to.

Before: 
image
After:
image

Only downside is that the frame coordinates now don't match up. I think that tradeoff is worth it though.

### Motivation and Context

Confused users, also https://ideas.obsproject.com/posts/2379.

### How Has This Been Tested?

macOS 14
See screenshots above.
Tested that the source still works as before. We're only modifying the visible string.

### Types of changes

### Checklist:

### DIFF
--- a/plugins/mac-capture/mac-sck-common.m
+++ b/plugins/mac-capture/mac-sck-common.m
@@ -102,12 +102,16 @@ bool build_display_list(struct screen_capture *sc, obs_properties_t *props)
             continue;
         }
 
+        CGDisplayModeRef display_mode = CGDisplayCopyDisplayMode(display.displayID);
         char dimension_buffer[4][12] = {};
         char name_buffer[256] = {};
-        snprintf(dimension_buffer[0], sizeof(dimension_buffer[0]), "%u", (uint32_t) display_screen.frame.size.width);
-        snprintf(dimension_buffer[1], sizeof(dimension_buffer[0]), "%u", (uint32_t) display_screen.frame.size.height);
+        snprintf(dimension_buffer[0], sizeof(dimension_buffer[0]), "%u",
+                 (uint32_t) CGDisplayModeGetPixelWidth(display_mode));
+        snprintf(dimension_buffer[1], sizeof(dimension_buffer[0]), "%u",
+                 (uint32_t) CGDisplayModeGetPixelHeight(display_mode));
         snprintf(dimension_buffer[2], sizeof(dimension_buffer[0]), "%d", (int32_t) display_screen.frame.origin.x);
         snprintf(dimension_buffer[3], sizeof(dimension_buffer[0]), "%d", (int32_t) display_screen.frame.origin.y);
+        CGDisplayModeRelease(display_mode);
 
         snprintf(name_buffer, sizeof(name_buffer), "%.200s: %.12sx%.12s @ %.12s,%.12s",
                  display_screen.localizedName.UTF8String, dimension_buffer[0], dimension_buffer[1], dimension_buffer[2],


PatTheMav:
Will this correctly show the scaled resolution selected in the display settings? Because there is a reason why macOS doesn't use the physical pixel resolution anywhere in the system and applications are also not really supposed to dabble in that either.

gxalpha:
Oh neat yeah seems like I conveniently only chose the non-scaled resolutions only in testing. meh.
Fyi we use this in another place in SCK when actually setting the resolution, but that seems to not have hurt so far there.

jcm93:
I was just looking at this, couldn't we just use the `backingScaleFactor` from the `NSScreen` instance `display_screen`? I was just testing it and it seems to give correct results at scaled resolutions. Happy to PR it if so.

PatTheMav:
My point was that we should not use any concept of "physical resolution" at all - you set a desired point resolution in the system settings and that's what the OS runs with. Whether the underlying pixel density is 3x or 2x is irrelevant to the user and should not be exposed or supported.

Doing so would run counter to the whole reason _why_ macOS doesn't expose the actual pixel density in the first place. It doesn't make sense to show a value that has no representation in the system otherwise.

Case in point: The Studio Display display by default runs a 2560x1440 point resolution, but that is _not_ the same as 5120x2880, because you can actually set the display to just that resolution.

With this change, both settings would yield the same resolution description, which IMO just adds to the confusion.

jcm93:
Hmm. My personal sense is that the context of compositing and the OBS canvas, what's more important to convey is the actual pixel size of what the source is going to be, so the user has a full understanding of its size in relation to the canvas, etc. Which seemed like the original motivation for the PR; I also see people semi-frequently confused by macOS's scaling distinctions. Just my two cents though.

PatTheMav:
You could of course present both (e.g. "2560x1440 (Actual 5120x2880)") though you need to implement all the plumbing to dynamically react to macOS changing that (macOS own High Resolution guide mentions that the scale factor is not to be considered "fixed" coding-wise), especially when multiple displays with different scale factors are in play.

That would probably require subscribing to something like `NSWindowDidChangeBackingPropertiesNotification` though I don't know if that one also happens if the backing properties changed for a display the current window is not positioned on.

PatTheMav:
Another alternative would be to make the source itself aware of the scaling mode and automatically apply the factor to the capture size (i.e. if the scale factor is 1.7, a window capture texture would automatically be scaled down by 1.7 so it works on the same point size as the render canvas (which usually uses a 1.0 ratio).

Once again this would need to be reactive and change dynamically when users change their scaling or a window is a moved to a different screen with a different scale factor.

gxalpha:
> My point was that we should not use any concept of "physical resolution" at all - you set a desired point resolution in the system settings and that's what the OS runs with. Whether the underlying pixel density is 3x or 2x is irrelevant to the user and should not be exposed or supported.
>
>Doing so would run counter to the whole reason why macOS doesn't expose the actual pixel density in the first place. It doesn't make sense to show a value that has no representation in the system otherwise.
>
>Case in point: The Studio Display display by default runs a 2560x1440 point resolution, but that is not the same as 5120x2880, because you can actually set the display to just that resolution.
>
>With this change, both settings would yield the same resolution description, which IMO just adds to the confusion.


The problem that I see with that is that Apple (and every other manufacturer) market their displays with their physical resolutions. Imo the current "1920x1080" is harmful if anything, because the user goes "but that's not what the resolution of my display is!" (and as of currently, it's not even the resolution it's captured at, we capture at the display mode pixel resolutions, as alluded to in the PR description).

As for having to implement the plumbing to be reactive, that seems overkill to me. If the user is changing their scaling while OBS is running they should expect that things *could* break, and something that's only shown in the source properties that doesn't affect how the source operates isn't *that* important.

PR #9629, opened 2023-09-25
UI: Add setting to copy screenshots to clipboard

jpark37:
### Description
Add ability to save screenshots to the OS clipboard.

HDR images will be converted to SDR.

Skip Linux because the image is sometimes corrupt when tested on Ubuntu.

### Motivation and Context
Useful for working with programs that accept images from the clipboard.

### How Has This Been Tested?
- [x] Windows: Clipboard is not used when setting is disabled
- [x] Windows, HDR: Verified clipboard matches file and is correct
- [x] Windows, SDR: Verified clipboard is SDR tonemapped compared to JXR file and both are correct
- [x] Mac: Clipboard is not used when setting is disabled
- [x] Mac, SDR: Verified clipboard matches file and is correct
- [x] Mac, HDR: Verified clipboard is correct, and SDR file hasn't regressed
- [x] Linux: Setting is not visible
- [x] Linux: Screenshot to file still works

### Types of changes

### Checklist:

### DIFF
--- a/UI/data/locale/en-US.ini
+++ b/UI/data/locale/en-US.ini
@@ -658,6 +658,7 @@ Basic.StatusBar.DelayStartingStoppingIn="Delay (stopping in %1 sec, starting in 
 Basic.StatusBar.RecordingSavedTo="Recording saved to '%1'"
 Basic.StatusBar.ReplayBufferSavedTo="Replay buffer saved to '%1'"
 Basic.StatusBar.ScreenshotSavedTo="Screenshot saved to '%1'"
+Basic.StatusBar.ScreenshotSavedToClipboard="Screenshot saved to clipboard and '%1'"
 Basic.StatusBar.AutoRemuxedTo="Recording auto remuxed to '%1'"
 
 # filters window
@@ -1058,6 +1059,7 @@ Basic.Settings.Output.Adv.Recording.Type.FFmpegOutput="Custom Output (FFmpeg)"
 Basic.Settings.Output.Adv.Recording.UseStreamEncoder="(Use stream encoder)"
 Basic.Settings.Output.Adv.Recording.Filename="Filename Formatting"
 Basic.Settings.Output.Adv.Recording.OverwriteIfExists="Overwrite if file exists"
+Basic.Settings.Output.Adv.Recording.ClipboardScreenshots="Copy screenshots to clipboard"
 Basic.Settings.Output.Adv.FFmpeg.Type="FFmpeg Output Type"
 Basic.Settings.Output.Adv.FFmpeg.Type.URL="Output to URL"
 Basic.Settings.Output.Adv.FFmpeg.Type.RecordToFile="Output to File"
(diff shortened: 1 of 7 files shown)


Warchamp7:
Very much on board with screenshots going to the clipboard, but I'd like to request some UX adjustments

1. Do not add a new context menu entry
2. Add an option in Advanced -> Recording section for 'Copy screenshots to clipboard'
2b. Probably rename the Recording section of Advanced to 'Output', as the settings here apply to Replay Buffer and Screenshots as well these days. Might be better as it's own PR
2c. Default 'Copy screenshots to clipboard' to on/checked

By default screenshots should be saved to file, and placed on the clipboard when supported, with an option to disable copying to clipboard.

jpark37:
I'm fine with that except defaulting the setting to on. A user shouldn't expect to lose the existing contents of their clipboard just because they took a screenshot. If they opt in, then it's on them.

Personally, I don't want image files being written, but I'm happy to have this feature in any form.

Warchamp7:
> I'm fine with that except defaulting the setting to on. A user shouldn't expect to lose the existing contents of their clipboard just because they took a screenshot. If they opt in, then it's on them.
> 
> Personally, I don't want image files being written, but I'm happy to have this feature in any form.

On Windows at least, the built in Prnt Scr system hotkey ONLY goes to clipboard, and their Snipping Tool defaults to copying to clipboard on both Windows 10 and 11. I would never propose copying screenshots to the clipboard without an option to disable it, but I think there is reasonable precedent for it as the default (again, on Windows at least)

Speaking personally I was actually surprised (and frustrated 🥲) that OBS screenshots _didn't_ go to the clipboard when the feature was originally added, heh

jpark37:
I made the changes although I kept the default off.

Print Screen and Snipping Tool are copying to clipboard as the primary purpose of their actions, not as a side effect that the user doesn't expect. Messing with the clipboard of someone who is just trying to save a file without their consent is a bit aggressive.

jpark37:
On, snipping tool saves files to `Pictures\Screenshots`. Sorry, never used it before.

I'm still on the fence... for my use case I'd want it on all the time, but I don't want to take the heat for blowing up users' clipboards, especially since we're changing the behavior of an existing action.

Warchamp7:
> On, snipping tool saves files to `Pictures\Screenshots`. Sorry, never used it before.
> 
> I'm still on the fence... for my use case I'd want it on all the time, but I don't want to take the heat for blowing up users' clipboards, especially since we'd be changing the behavior of an existing action.

If anyone comes to you upset about it you are beyond welcome to direct them to me ;P

jpark37:
Ok, it's on by default.

PatTheMav:
> I'm fine with that except defaulting the setting to on. A user shouldn't expect to lose the existing contents of their clipboard just because they took a screenshot. If they opt in, then it's on them.
> 
> Personally, I don't want image files being written, but I'm happy to have this feature in any form.

FWIW I argued for the same, so thank you for pointing that out as well. 😄

Warchamp7 (approved):

ryantheleach:
Personally I wouldn't want old temporary debugging screenshots hogging up diskspace if I intend on using the copy to clipboard functionality. If copy to clipboard is an option, then saving to disk must also be IMO.

PR #9631, opened 2023-09-25
mac-capture: Don't crash when migrating unknown display IDs

### LINKED ISSUE #9606: 30.0.0 beta 3 SEGV on start with MacOS 12.6.6 & old profile from 29.x
### Operating System Info

macOS 12

### Other OS

_No response_

### OBS Studio Version

Other

3.0.0-beta3

### OBS Studio Log URL

https://obsproject.com/logs/AnbXRkDC55v0Ztib

### OBS Studio Crash Log URL

_No response_

### Expected Behavior

OBS should run - or at least not segfault.

### Current Behavior

Segfault on program start. 

### Steps to Reproduce

1. Run obs studio 30.0.0 beta 3 (with my profile from 29.1.2 still around)
2. Wait for it to crash.
3.
...

It's a beta, I get that.  Just thought y'all should know 30.0.0 beta 3 is SEGV on start for MacOS.  If I delete ${HOME}/Library/Application Support/obs-studio/, then it starts okay, but all my previously configured scenes are obviously gone.  Possibly OBS should nuke the old profile, or offer to do so?

Feel free to close this issue if you don't feel it's worth addressing.

gxalpha:
### Description

The current code assumes that a display UUID can be created with the stored ID, but that's not always the case, e.g. when the user doesn't have the display connected. As such, we need to null check this, and fall back to the invalid ID (0) when the ID cannot be migrated.

The current code also only migrates on source creation, which yields weird behaviour where if the user opens properties and then cancels it would still show the first display, but only for the session. This is why the code was factored out of the creation function and now is always used when an ID needs to be acquired from OBS Data settings, including when the source is updated.

The deprecated display capture still doesn't like when the display doesn't exist (it returns false/NULL on creation leading to the source basically being "dead" and the user not able to do anything with it but delete it), but this is the same on 29.1 and I'm not too worried about that source anyways.

### Motivation and Context

Got reports of crashes in mac-capture when the UUID gets created.
This can be reproduced easily by (first removing "display_uuid" if it exists and then) setting the "display" setting in the scene collection json to a number higher than the number of connected displays.

Fixes #9606

### How Has This Been Tested?

macOS 14
Tested that with a variety of different settings (existing "display", nonexistent "display", existing "display_uuid", nonexistent "display_uuid", and combinations of the keys) that previous behavior is restored, where when the display isn't found, nothing is shown but the properties let you select one of the available displays.

### Types of changes

### Checklist:

### DIFF
--- a/plugins/mac-capture/window-utils.h
+++ b/plugins/mac-capture/window-utils.h
@@ -31,3 +31,8 @@ void window_defaults(obs_data_t *settings);
 void add_window_properties(obs_properties_t *props);
 
 void show_window_properties(obs_properties_t *props, bool show);
+
+/** Get the display ID of a display and simultaneously migrate pre-30.0 display IDs to 30.0 UUIDs.
+ - Parameter settings: Pointer to `obs_data_t` object containing `display` int and/or `display_uuid` string
+ - Returns: `CGDirectDisplayID` of the display the user selected. May be 0 if the display cannot be found. */
+CGDirectDisplayID get_display_migrate_settings(obs_data_t *settings);
--- a/plugins/mac-capture/mac-display-capture.m
+++ b/plugins/mac-capture/mac-display-capture.m
@@ -286,27 +286,7 @@ static void *display_capture_create(obs_data_t *settings, obs_source_t *source)
     init_window(&dc->window, settings);
     load_crop(dc, settings);
 
-    bool legacy_display_id = obs_data_has_user_value(settings, "display");
-    if (legacy_display_id) {
-        CGDirectDisplayID display_id = (CGDirectDisplayID) obs_data_get_int(settings, "display");
-        CFUUIDRef display_uuid = CGDisplayCreateUUIDFromDisplayID(display_id);
-        CFStringRef uuid_string = CFUUIDCreateString(kCFAllocatorDefault, display_uuid);
-        obs_data_set_string(settings, "display_uuid", CFStringGetCStringPtr(uuid_string, kCFStringEncodingUTF8));
-        obs_data_erase(settings, "display");
-        CFRelease(uuid_string);
-        CFRelease(display_uuid);
-    }
-
-    const char *display_uuid = obs_data_get_string(settings, "display_uuid");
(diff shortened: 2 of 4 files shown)


PatTheMav:
So the crash is not related to an invalid UUID, but an invalid C-string being passed into the CoreFoundation method generating a `CFUUIDRef`?

gxalpha:
With the crash that I can reproduces (and where the crash stack pretty much matches the one in the report I've seen) `CGDisplayCreateUUIDFromDisplayID` returns `NULL` (because the `display_id` refers to a display that the system doesn't know), which is then passed into `CFUUIDCreateString` which crashes.

I'm not sure what you mean exactly with "invalid UUID" - are you referring to a string that's not formatted as a UUID or to a UUID that the system doesn't know? Although neither of these cases appear to crash on master on my system.

PatTheMav:
Ok, so the crash is limited to cases of migrating from the old displayID-based setting to the UUID-based setting but with an invalid displayID, got it.

PatTheMav (approved):

RytoEX:
Does this fix #9606?

gxalpha:
Yes.

plugins/mac-capture/window-utils.h
```diff
@@ -31,3 +31,8 @@ void window_defaults(obs_data_t *settings);
 void add_window_properties(obs_properties_t *props);
 
 void show_window_properties(obs_properties_t *props, bool show);
+
+/** Get the display ID of a display and simultaniously migrate pre-30.0 display IDs to 30.0 UUIDs.
```
RytoEX, line 35:
```suggestion
/** Get the display ID of a display and simultaneously migrate pre-30.0 display IDs to 30.0 UUIDs.
```

RytoEX (changes requested):

plugins/mac-capture/window-utils.h
gxalpha: Thanks.

RytoEX (approved):

PR #9643, opened 2023-09-28
libobs: Fix hotkeys with groups and duplicate sceneitems

### LINKED ISSUE #9639: Show/hide hotkeys for separate sceneitems of the same source on the same scene get merged after restart
### Operating System Info

Ubuntu 22.04

### Other OS

Also windows 10

### OBS Studio Version

30.0.0-beta2

Also confirmed on 29.1.3

### OBS Studio Log URL

https://obsproject.com/logs/uV7ol09LN88KVeEj

### OBS Studio Crash Log URL

_No response_

### Expected Behavior

Hotkeys for separate sceneitems should be saved and loaded separately.

### Current Behavior

Upon restart all sceneitems of the same source on the same scene get the same show/hide hotkeys regardless of what was assigned to them.

### Steps to Reproduce

1. Create a scene with two or more sceneitems of the same source.
2. In Settings > Hotkeys, assign different show/hide hotkeys to each.
3. Restart OBS.
4. Notice the hotkeys are now the same for each.
...

Log is from a 30.0 beta session because it was easier for me to provide, but behaviour is similar on 29.1.3.

derrod:
### Description

Use the locally unique ID of a scene item for its hotkey(s) rather than the name, which will conflict when multiple sceneitems for the same source exist.

Also addresses the following issues:
- IDs and counter were not restored before creating hotkeys, which needs to be the case for ID-based hotkeys to work
- Moving items into and out of groups did not move their hotkeys to the group/scene so they would get lost
- Adds migration for old name-based hotkeys to new ID-based ones

### Motivation and Context

Fixes #9639

### How Has This Been Tested?

Verified hotkeys are now saved and loaded correctly.

### Types of changes

### Checklist:

### DIFF
--- a/libobs/obs-scene.c
+++ b/libobs/obs-scene.c
@@ -960,7 +960,10 @@ static void set_visibility(struct obs_scene_item *item, bool vis)
 	pthread_mutex_unlock(&item->actions_mutex);
 }
 
-static void scene_load(void *data, obs_data_t *settings);
+static obs_sceneitem_t *obs_scene_add_internal(obs_scene_t *scene,
+					       obs_source_t *source,
+					       obs_sceneitem_t *insert_after,
+					       int64_t id);
 
 static void scene_load_item(struct obs_scene *scene, obs_data_t *item_data)
 {
@@ -971,6 +974,8 @@ static void scene_load_item(struct obs_scene *scene, obs_data_t *item_data)
 	const char *blend_method_str;
 	const char *blend_str;
 	struct obs_scene_item *item;
+	struct calldata params;
+	uint8_t stack[128];
 	bool visible;
 	bool lock;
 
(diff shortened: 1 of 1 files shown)


tt2468:
Just going to note that this is likely to break some obs-websocket scripts out there, but I quite honestly don't care since there's significantly better options for controlling hotkeys than triggering hotkeys via obs-websocket requests.

Penwy:
Tested on Ubuntu 22.04.
Correctly fixes the intended issue.
No additional issues noticed.

derrod:
> Edit : I was hasty and mistaken. Reordering sceneitems messes with hotkey attribution.
> 
> Steps to reproduce :
> 
> * have a scene with two items (don't need to be the same source)
> * attribute different show/hide hotkeys to them
> * switch their order in the list
> * close, reopen
> * the hotkeys have been swapped around

This should now be fixed, alongisde another issue relating to hotkeys when moving things in/out of groups.

RytoEX:
@Penwy does the recent force-push fix the issues you noted?

RytoEX (changes requested):
Commit message nit:
*  reodering -> reordering

Lain-B (approved):

PR #9647, opened 2023-09-29
obs-qsv11: Use translatable strings for target usage

RytoEX:
### Description

For Target Usage, instead of doing string comparisons against long descriptive strings, use translatable strings for descriptive UI text and simple designations for values used for comparisons and storage. This is similar to what we do for NVENC Presets.

### Motivation and Context

Having to do string comparisons against the full descriptive text all the time seems not ideal. We do something similar to this change for NVENC Presets, though we use a macro there.
https://github.com/obsproject/obs-studio/blob/5dda04ad5ee8733cf5b3073d8906f080647a9139/plugins/obs-ffmpeg/data/locale/en-US.ini#L44-L50
https://github.com/obsproject/obs-studio/blob/5dda04ad5ee8733cf5b3073d8906f080647a9139/plugins/obs-ffmpeg/obs-ffmpeg-nvenc.c#L547-L557

### How Has This Been Tested?

Tested locally and checked the stored values in my test profile's `recordEncoder.json`.
Before:
```
{"target_usage":"TU4: Balanced (Medium Quality)"}
```
After:
```
{"target_usage":"TU4"}
```
Verified that the updated string comparisons work as expected.

### Types of changes

### Checklist:


RytoEX:
Possible alternatives:
* Use an implementation like the `qsv_rate_control_info` struct
* Use a macro

I mostly just submitted this because it's what I wrote first.

kkartaltepe:
Seems ok to me, but now users of the beta might have broken presets since we dont upgrade those just the original preset names?

RytoEX:
> Seems ok to me, but now users of the beta might have broken presets since we dont upgrade those just the original preset names?

We generally consider settings in beta to be unstable and in-flux. The only upgrades we worry about are between stable releases.

RytoEX:
Received approvals off-thread. No major objections. I don't think we've historically handled settings migrations for anything except stable releases. Going to merge this as-is because I want to avoid the existing settings strings ending up in a stable release.

PR #9648, opened 2023-09-30
libobs,obs-filters: Fix NAN when tonemapping

jpark37:
### Description
Can happen when colors are wider than Rec. 2020.

### Motivation and Context
Don't want black pixels where colors should be.

### How Has This Been Tested?
Both shader paths looked buggy before and look fine after. Also verified Windows OpenGL runs.

### Types of changes

### Checklist:


PatTheMav:
Can confirm that this fixes the issue reported off-thread in our support channels - HDR test footage is correctly tonemapped in SDR as well as HDR with automatic tonemapping.

RytoEX:
Merging based on comment above.

PR #9654, opened 2023-10-02
rtmp-services: Update OnlyFans streaming service

artemm-dev:
### Description
Update OnlyFans streaming service urls 

### Motivation and Context
We have changed the our main url for streaming. Added it as a new entity. The old one has been renamed as an alternative.

### How Has This Been Tested?
Not tested. Since it was not possible to build the project on macOS 12.5 (a lot of build errors)

### Types of changes

### Checklist:

### DIFF
--- a/plugins/rtmp-services/data/package.json
+++ b/plugins/rtmp-services/data/package.json
@@ -1,11 +1,11 @@
 {
     "$schema": "schema/package-schema.json",
     "url": "https://obsproject.com/obs2_update/rtmp-services/v5",
-    "version": 234,
+    "version": 235,
     "files": [
         {
             "name": "services.json",
-            "version": 234
+            "version": 235
         }
     ]
 }
--- a/plugins/rtmp-services/data/services.json
+++ b/plugins/rtmp-services/data/services.json
@@ -1526,13 +1526,18 @@
         },
         {
             "name": "OnlyFans.com",
+            "stream_key_link": "https://onlyfans.com/my/settings/other",
             "servers": [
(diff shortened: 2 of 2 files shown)


gxalpha (commented):
Why does this create a new service entry? Are there reasons against just updating the existing one?

Also please fill out the PR template completely instead of removing parts of it.

Fenrirthviti:
I definitely don't like the idea of adding a new entry simple for changing endpoints. I personally don't see much difference in having to select a new destination server, as opposed to a new service entirely in the list.

Can you please explain why adding an entirely new entry is easier for your users?

artemm-dev:
> I definitely don't like the idea of adding a new entry simply for changing endpoints. I personally don't see much difference in having to select a new destination server, as opposed to a new service entirely in the list.
> 
> Can you please explain why adding an entirely new entry is easier for your users?

We use 2 different providers for rtmp streaming
Broadcast keys and how they are obtained are different for these providers
Therefore, we would like it to be separate entities
Also, we plan to add authorization through our service to the new entity, and leave the old one as an alternative and familiar option for users

notr1ch:
That seems like an implementation detail that shouldn't be visible to users. Have your different providers as different servers, there is no need for two separate services entries.

artemm-dev:
> That seems like an implementation detail that shouldn't be visible to users. Have your different providers as different servers, there is no need for two separate services entries.

ok, we have updated as you request

Lain-B (approved):

PR #9656, opened 2023-10-03
Remove semantic beta/RC versions

derrod:
### Description

Removes old `OBS_BETA_*` and `OBS_RELEASE_CANDIDATE_*` version macros and updates code that was previously using it.

Also fixes `OBS_COMMIT` not actually being set in the CMake rework and cleans up the version comparison in the windows updater a little bit.

> [!IMPORTANT]
> When merging before 30.0 final, remember to immediately increment `LIBOBS_API_VER` for the next release

### Motivation and Context

Historically we did not update `LIBOBS_API_VER` due to the Windows updater relying on it. Now that it has been reworked with #6907 this is no longer necessary.

Arguably we should bump the internal version as soon as we tag the first beta. This ensures that modules using the `LIBOBS_` version macros or `obs_get_version()` will receive the correct API version even in beta/rc releases.

### How Has This Been Tested?

Compiled and ran OBS, everything seems to work fine.

### Types of changes

### Checklist:


RytoEX (changes requested):
Does this call in `AutoUpdateThread::run()` in `UI/update/win-update.cpp` (need changed or is it inconsequential?
```
config_set_int(GetGlobalConfig(), "General", "SkipUpdateVersion", 0);
```

UI/window-basic-main.cpp
```diff
@@ -2509,7 +2498,6 @@ void OBSBasic::ReceivedIntroJson(const QString &text)
 #endif
 }
 
-#undef CUR_VER
 #undef LAST_INFO_VERSION_STRING
```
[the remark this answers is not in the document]
derrod: Yes

UI/window-basic-main.cpp
```diff
@@ -2448,19 +2447,9 @@ void OBSBasic::ReceivedIntroJson(const QString &text)
...
-		if (major == OBS_BETA_MAJOR && minor == OBS_BETA_MINOR &&
-		    item.Beta == OBS_BETA)
-#else
 		if (major == LIBOBS_API_MAJOR_VER &&
-		    minor == LIBOBS_API_MINOR_VER && item.RC == 0 &&
-		    item.Beta == 0)
-#endif
-		{
+		    minor == LIBOBS_API_MINOR_VER &&
+		    item.RC == OBS_RELEASE_CANDIDATE && item.Beta == OBS_BETA) {
```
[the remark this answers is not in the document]
derrod: `item.Beta`/`item.RC` are zero-initialised, the defines sre also set to 0 by default, so these always evaluate to true in case they're not set.

UI/update/win-update.cpp
```diff
@@ -283,8 +260,8 @@ try {
 	/* ----------------------------------- *
 	 * skip this version if set to skip    */
 
-	uint64_t skipUpdateVer = config_get_uint(GetGlobalConfig(), "General",
-						 "SkipUpdateVersion");
+	const char *skipUpdateVer = config_get_string(
```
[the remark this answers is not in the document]
derrod: Yes because config doesn't store types you can read anything saved to the config back as a string.

UI/window-basic-main.cpp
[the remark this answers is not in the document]
derrod: Effectively this will be the same as before for release builds (comparison with `0`).

UI/update/win-update.cpp
```diff
@@ -50,30 +50,13 @@ using namespace updater;
...
-#define PRE_RELEASE true
-#else
-#define CUR_VER ((uint64_t)LIBOBS_API_VER << 16ULL)
-#define PRE_RELEASE false
-#endif
-
-#ifndef CUR_COMMIT
-#define CUR_COMMIT "00000000"
-#endif
+constexpr uint64_t kCurVer = (uint64_t)LIBOBS_API_VER << 16ULL |
```
PatTheMav, line 53: It's not 1976 anymore, _if_ we refactor this, could we please use readable variable names (i.e. `kCurrentVersion`).

I'm not sure what `k` in Hungarian Notation means, but when it comes to the style it might be one of the very few things I agree with Linus Torvalds on. 😬

cmake/common/versionconfig.cmake
```diff
@@ -33,23 +33,22 @@ endif()
 
 # Set beta/rc versions if suffix included in version string
 if(_obs_version MATCHES "[0-9]+\\.[0-9]+\\.[0-9]+-rc[0-9]+")
-  string(REGEX REPLACE "([0-9]+)\\.([0-9]+)\\.([0-9]+)-rc([0-9]+).*$" "\\1;\\2;\\3;\\4" _obs_release_candidate
-                       ${_obs_version})
+  string(REGEX REPLACE "([0-9]+)\\.([0-9]+)\\.([0-9]+)-rc([0-9]+).*$" "\\4" _obs_release_candidate ${_obs_version})
```
PatTheMav, line 36: We match 4 tokens if you need only the last one - I'd just remove the parens around the first 3 tokens and only match the number after `rc` or `beta` respectively.

cmake/common/versionconfig.cmake
```diff
@@ -33,23 +33,22 @@ endif()
...
 elseif(_obs_version MATCHES "[0-9]+\\.[0-9]+\\.[0-9]+-beta[0-9]+")
-  string(REGEX REPLACE "([0-9]+)\\.([0-9]+)\\.([0-9]+)-beta([0-9]+).*$" "\\1;\\2;\\3;\\4" _obs_beta ${_obs_version})
+  string(REGEX REPLACE "([0-9]+)\\.([0-9]+)\\.([0-9]+)-beta([0-9]+).*$" "\\4" _obs_beta ${_obs_version})
+endif()
+
+# Set commit for untagged version comparisons in the Windows updater
+if(OS_WINDOWS)
```
PatTheMav, line 42: No OS-Specific code during bootstrapping (only exception is `osconfig` itself) - I don't think there's any harm in setting it by default.

Though if it is _only_ used in the Windows updater, that code should be moved inside that CMakeLists file instead.

PatTheMav:
Overall I think this is fine, but I can't shake the feeling that our handling of the program version is unnecessarily convoluted and instead of replacing it with something simpler we just removed some internal parts of it and kept the rest.

I'm also unhappy that this is all still straight-up handled in `window-basic-main.cpp`, which means that every time the version changes you invalidate the entire application binary and incur a costly recompile.

Handling that stuff in a static or object library and just referencing a version struct as an external symbol should have more structure (`obs_version_t` could hold all version relevant information) and would only require a recompile of that tiny object file and linking the rest of the binary to it.

cmake/common/versionconfig.cmake
derrod: Yeah I just forgot about that.

cmake/common/versionconfig.cmake
derrod: Alright, I'll just change it to work like all the other variables.

UI/update/win-update.cpp
derrod: `k` is a common prefix for global constants in various C++ projects, I think most just adapted that from the Google style guide, though it may predate that. Since we don't have a C++ style guide I just started using it in my C++ code to keep things clearer.

derrod: On second thought, these don't need to be globals though, I just replaced the macros very literally. I just moved this to the function scope.

derrod:
> Handling that stuff in a static or object library and just referencing a version struct as an external symbol should have more structure (`obs_version_t` could hold all version relevant information) and would only require a recompile of that tiny object file and linking the rest of the binary to it.

Yeah that may be something we should do in the future. This was mostly just a "get rid of those macros and old habits" cleanup that is small enough for 30.0 and by no means the end of it.

cmake/common/versionconfig.cmake
PatTheMav: The value seems only used in `UI/cmake/os-windows.cmake`, so the regex could be moved there. It shouldn't have any side-effects here, but it feels iffy adding this globally if it's really only used by a single module for a specific OS for a specific compilation unit.

PatTheMav:
> > Handling that stuff in a static or object library and just referencing a version struct as an external symbol should have more structure (`obs_version_t` could hold all version relevant information) and would only require a recompile of that tiny object file and linking the rest of the binary to it.
> 
> Yeah that may be something we should do in the future. This was mostly just a "get rid of those macros and old habits" cleanup that is small enough for 30.0 and by no means the end of it.

Got it, sounds good to me then.

PatTheMav (approved):

UI/update/win-update.cpp
RytoEX: I've seen a description that states:
> k - a constant or immutable variable

Globals are usually prefixed with `g` or `g_`.

RytoEX:
For what it's worth, there is libobs-version, though that may have a separate purpose (libobs versioning rather than application versioning).

UI/window-basic-main.cpp
```diff
@@ -2471,15 +2449,26 @@ void OBSBasic::ReceivedIntroJson(const QString &text)
...
 	}
 
+#if OBS_RELEASE_CANDIDATE > 0
+	const char *kLastInfoVersion = "InfoLastRCVersion";
+#elif OBS_BETA > 0
+	const char *kLastInfoVersion = "InfoLastBetaVersion";
+#else
+	const char *kLastInfoVersion = "InfoLastVersion";
+#endif
+	const uint64_t currentVersion = (uint64_t)LIBOBS_API_VER << 16ULL |
```
RytoEX, line 2459: This is constexpr in `ParseUpdateManifest` in `UI/update/win-update.cpp`. Should it also be constexpr here?

derrod: Sure, should work as well.

RytoEX: I mostly meant `currentVersion`, but if the others work too, by all means.

RytoEX (approved):
This seems fine. As noted, we'll need to bump `LIBOBS_API_VER` before the next release of any kind (beta/rc/stable) after merging this.

Lain-B:
> This seems fine. As noted, we'll need to bump `LIBOBS_API_VER` before the next release of any kind (beta/rc/stable) after merging this.

Seems like not a big deal. I'll make the commit for it after merging.

UI/window-basic-main.cpp
```diff
@@ -2471,15 +2449,26 @@ void OBSBasic::ReceivedIntroJson(const QString &text)
 		return;
 	}
 
+#if OBS_RELEASE_CANDIDATE > 0
+	constexpr const char *kLastInfoVersion = "InfoLastRCVersion";
+#elif OBS_BETA > 0
+	constexpr const char *kLastInfoVersion = "InfoLastBetaVersion";
+#else
+	constexpr const char *kLastInfoVersion = "InfoLastVersion";
+#endif
```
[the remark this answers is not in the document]
derrod: The variables were initially declared outside of the function, I forgot to remove it when moving them inside (this is fixed now).

It's a common prefix for constant values in C++ codebases (I assume because in hungarian notation `c` is already taken for `char`). Apple also uses it for instance (e.g. in CoreFoundation https://developer.apple.com/documentation/corefoundation/core_foundation_constants?language=objc), and so does Google in their style guide: https://google.github.io/styleguide/cppguide.html#Constant_Names

As OBS does not have a consistent C++ style, or style guide for that matter, separate authors use their own preferred syntax. Me and a few others often use this prefix for global constants, as it makes it easy to distinguish a global const from a member or local variable, as well as from macros which are all-uppercase.

PR #9657, opened 2023-10-03
linux-v4l2: Expand packed values to 4 bytes

kkartaltepe:
### Description

Expand packed values to 4 bytes, and wrap up static values with a macro.

### Motivation and Context

These days you might capture card a 175Hz monitor with a framerate defined as 10000000/57143 that doesnt fit in a short, as reported by a user. And reporting a negative FPS in that case isnt very helpful.

### How Has This Been Tested?

My logitech webcam still works.

### Types of changes

### Checklist:

### DIFF
--- a/plugins/linux-v4l2/v4l2-helpers.h
+++ b/plugins/linux-v4l2/v4l2-helpers.h
@@ -19,6 +19,7 @@ along with this program.  If not, see <http://www.gnu.org/licenses/>.
 
 #include <linux/videodev2.h>
 #include <libv4l2.h>
+#include <inttypes.h>
 
 #include <obs-module.h>
 #include <media-io/video-io.h>
@@ -26,6 +27,8 @@ along with this program.  If not, see <http://www.gnu.org/licenses/>.
 #ifdef __cplusplus
 extern "C" {
 #endif
+
+#define PACK64(a, b) (((uint64_t)a << 32) | ((uint64_t)b & 0xffffffff))
 
 /**
  * Data structure for mapped buffers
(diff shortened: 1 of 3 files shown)


kkartaltepe:
Actually this needs more work, we need to store resolution and framerate somewhere else and upgrade the old properties...

plugins/linux-v4l2/v4l2-helpers.h
```diff
@@ -91,24 +94,24 @@ static inline enum video_format v4l2_to_obs_video_format(uint_fast32_t format)
  * and the height in the low word.
  * The array is terminated with a zero.
  */
-static const int v4l2_framesizes[] = {
+static const uint64_t v4l2_framesizes[] = {
```
derrod, line 97: Why is this a `uint64_t` instead of `int64_t` as elsewhere in packed numbers?

derrod (commented):
> Actually this needs more work, we need to store resolution and framerate somewhere else and upgrade the old properties...

Sounds like you either migrate by changing the key to a new one and copy from the old one if it exists, or figure out a way to detect the old value's format.

plugins/linux-v4l2/v4l2-helpers.h
kkartaltepe: Fixed (in a second), it should be an int64_t.

kkartaltepe:
> > Actually this needs more work, we need to store resolution and framerate somewhere else and upgrade the old properties...
> 
> Sounds like you either migrate by changing the key to a new one and copy from the old one if it exists, or figure out a way to detect the old value's format.

I slept on it, and it should be safe to just treat any packed value with no high bits as the old 32bit form.

Lain-B (approved):

PR #9662, opened 2023-10-03
UI: Use C++11 ranged-for instead of Q_FOREACH

RytoEX:
### Description

Qt has been porting uses of Q_FOREACH/foreach to C++11's ranged-for. Let's do the same.

References:
 * https://bugreports.qt.io/browse/QTBUG-115796
 * https://doc.qt.io/qt-6/foreach-keyword.html
 * https://codereview.qt-project.org/q/Q_FOREACH
 * https://en.cppreference.com/w/cpp/language/range-for

### Motivation and Context

We target C++17. Q_FOREACH was previously suggested to be deprecated/replaced for Qt 6. While that did not happen, Qt's efforts to replace Q_FOREACH/foreach indicate that its days may be numbered.

### How Has This Been Tested?

Built and ran OBS on Windows 11. Invoked the Reset Stats hotkey a few times.

### Types of changes

### Checklist:

### DIFF
--- a/UI/window-basic-main.cpp
+++ b/UI/window-basic-main.cpp
@@ -10882,9 +10882,11 @@ void OBSBasic::CheckDiskSpaceRemaining()
 
 void OBSBasic::ResetStatsHotkey()
 {
-	QList<OBSBasicStats *> list = findChildren<OBSBasicStats *>();
+	const QList<OBSBasicStats *> list = findChildren<OBSBasicStats *>();
 
-	foreach(OBSBasicStats * s, list) s->Reset();
+	for (OBSBasicStats *s : list) {
+		s->Reset();
+	}
 }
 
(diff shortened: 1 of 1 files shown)


gxalpha (approved):
Good change.
Personally I’d at curly braces to the for-loop.

derrod:
Might as well skip the variable assignment and go straight for
```c++
	for (OBSBasicStats *s : findChildren<OBSBasicStats *>())
		s->Reset();
```

derrod (approved):

RytoEX:
I added the curly braces for the `for` loop (I usually prefer this anyway), but left in the variable assignment. While the assembly in Visual Studio seemed basically identical in Debug, removing the assignment produced about twice as much assembly in RelWithDebInfo. If there's an optimization to be made here by removing the assignment, maybe someone can revisit that in the future. For now, let's just let the compilers do their own thing

PR #1253, opened 2018-04-06
UI: Add headphone checkbox to mixer

cg2121:
This adds a headphone icon to the mixer area. It replaces the audio monitor selection in the advanced audio properties.

screenshot from 2018-04-06 14-34-48

PR #9666, opened 2023-10-04
libobs: Prevent audio doubling

cg2121:
### Description
If a source has output & monitor enabled and there is an audio output source active that has the same audio output device as the monitor, the audio would be doubled in the final output.

### Motivation and Context
Suggestion from @Warchamp7, as he says this should be implemented before #1253 is merged.

### How Has This Been Tested?
Created audio source and set the monitoring to output and monitor. Then added a audio output source the used the same device as the monitor. Then I recorded a video to make sure there was no audio doubling.

### Types of changes

### Checklist:

### DIFF
--- a/libobs/obs-source.c
+++ b/libobs/obs-source.c
@@ -172,6 +172,11 @@ static inline bool is_audio_source(const struct obs_source *source)
 	return source->info.output_flags & OBS_SOURCE_AUDIO;
 }
 
+static inline bool is_audio_output_source(const struct obs_source *source)
+{
+	return source->info.output_flags & OBS_SOURCE_DO_NOT_SELF_MONITOR;
+}
+
 static inline bool is_composite_source(const struct obs_source *source)
 {
 	return source->info.output_flags & OBS_SOURCE_COMPOSITE;
@@ -981,6 +986,49 @@ static void obs_source_deferred_update(obs_source_t *source)
 	}
 }
 
+void update_audio_output_dup_refs(obs_source_t *source)
+{
+	if (!is_audio_output_source(source))
+		return;
+
+	if (source->output_devices_match) {
+		os_atomic_dec_long(&obs->audio.duplicate_audio_device_refs);
+		source->output_devices_match = false;
+	}
+
+	if (obs_source_active(source) && !obs_source_muted(source)) {
+		obs_data_t *settings = obs_source_get_settings(source);
+		const char *device_id =
+			obs_data_get_string(settings, "device_id");
+		obs_data_release(settings);
+
+		bool match = devices_match(device_id,
+					   obs->audio.monitoring_device_id);
+
+		if (match) {
+			os_atomic_inc_long(
+				&obs->audio.duplicate_audio_device_refs);
+			source->output_devices_match = true;
+		}
+	}
+
+	bool ignore = os_atomic_load_long(
+			      &obs->audio.duplicate_audio_device_refs) > 0;
+
+	if (ignore == obs->audio.user_ignore_duplicate_audio)
+		return;
+
+	struct audio_action action = {.timestamp = os_gettime_ns(),
+				      .type = AUDIO_ACTION_IGNORE,
+				      .set = ignore};
+
+	obs->audio.user_ignore_duplicate_audio = ignore;
+
+	pthread_mutex_lock(&source->audio_actions_mutex);
+	da_push_back(source->audio_actions, &action);
+	pthread_mutex_unlock(&source->audio_actions_mutex);
+}
+
 void obs_source_update(obs_source_t *source, obs_data_t *settings)
 {
 	if (!obs_source_valid(source, "obs_source_update"))
@@ -995,6 +1043,9 @@ void obs_source_update(obs_source_t *source, obs_data_t *settings)
 	} else if (source->context.data && source->info.update) {
 		source->info.update(source->context.data,
 				    source->context.settings);
+
+		update_audio_output_dup_refs(source);
+
 		obs_source_dosignal(source, "source_update", "update");
 	}
 }
@@ -1100,6 +1151,8 @@ void obs_source_set_texcoords_centered(obs_source_t *source, bool centered)
 
 static void activate_source(obs_source_t *source)
 {
+	update_audio_output_dup_refs(source);
+
 	if (source->context.data && source->info.activate)
 		source->info.activate(source->context.data);
 	obs_source_dosignal(source, "source_activate", "activate");
@@ -1107,6 +1160,8 @@ static void activate_source(obs_source_t *source)
 
 static void deactivate_source(obs_source_t *source)
 {
+	update_audio_output_dup_refs(source);
+
 	if (source->context.data && source->info.deactivate)
 		source->info.deactivate(source->context.data);
 	obs_source_dosignal(source, "source_deactivate", "deactivate");
@@ -5263,6 +5318,8 @@ void obs_source_set_muted(obs_source_t *source, bool muted)
 
 	signal_handler_signal(source->context.signals, "mute", &data);
 
+	update_audio_output_dup_refs(source);
+
 	pthread_mutex_lock(&source->audio_actions_mutex);
 	da_push_back(source->audio_actions, &action);
 	pthread_mutex_unlock(&source->audio_actions_mutex);
@@ -5434,7 +5491,11 @@ static float get_source_volume(obs_source_t *source, uint64_t os_time)
 
 	bool muted = !source->enabled || source->muted ||
 		     (source->push_to_mute_enabled && push_to_mute_active) ||
-		     (source->push_to_talk_enabled && !push_to_talk_active);
+		     (source->push_to_talk_enabled && !push_to_talk_active) ||
+		     (obs->audio.ignore_duplicate_audio &&
+		      !is_audio_output_source(source) &&
+		      source->monitoring_type ==
+			      OBS_MONITORING_TYPE_MONITOR_AND_OUTPUT);
 
 	if (muted || close_float(source->volume, 0.0f, 0.0001f))
 		return 0.0f;
@@ -5482,6 +5543,9 @@ static inline void apply_audio_action(obs_source_t *source,
 		break;
 	case AUDIO_ACTION_PTM:
 		source->push_to_mute_pressed = action->set;
+		break;
+	case AUDIO_ACTION_IGNORE:
+		obs->audio.ignore_duplicate_audio = action->set;
 		break;
 	}
 }
--- a/libobs/audio-monitoring/osx/coreaudio-output.c
+++ b/libobs/audio-monitoring/osx/coreaudio-output.c
@@ -149,8 +149,6 @@ static void buffer_audio(void *data, AudioQueueRef aq, AudioQueueBufferRef buf)
 	UNUSED_PARAMETER(aq);
 }
 
-extern bool devices_match(const char *id1, const char *id2);
-
 static bool audio_monitor_init(struct audio_monitor *monitor,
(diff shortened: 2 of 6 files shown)


libobs/obs-source.c
```diff
@@ -1098,15 +1103,36 @@ void obs_source_set_texcoords_centered(obs_source_t *source, bool centered)
 	source->texcoords_centered = centered;
 }
 
+static void check_for_duplicate_audio(const obs_source_t *source, bool active)
```
PatTheMav, line 1106: I feel like this function should be renamed, a `check_something` function is usually expected to return a value (the result of the check) and be pure (aka it should not change state).

This function however checks _and_ changes state and does not return the result of the check.

So it should either return `match` with the caller then making the decision whether to change the ref count.

Also it seems like the ref count is increased when a source with a duplicate device is activated, but if the same source is deactivated, the ref count is not decreased (because `active` is `false`) - is that by design?

cg2121: Changed it to update_duplicate_audio_refs

> Also it seems like the ref count is increased when a source with a duplicate device is activated, but if the same source is deactivated, the ref count is not decreased (because active is false) - is that by design?

Since it compares (match && active) the refs will increment if both are true and will decrement if either is false. In testing everything is working as expected.

libobs/obs-source.c
```diff
@@ -5434,7 +5491,11 @@ static float get_source_volume(obs_source_t *source, uint64_t os_time)
 
 	bool muted = !source->enabled || source->muted ||
 		     (source->push_to_mute_enabled && push_to_mute_active) ||
-		     (source->push_to_talk_enabled && !push_to_talk_active);
+		     (source->push_to_talk_enabled && !push_to_talk_active) ||
+		     (obs->audio.ignore_duplicate_audio &&
```
PatTheMav, line 5495: I am confused by this check - `ignore_duplicate_audio` should never be set for a device that does not have `OBS_SOURCE_DO_NOT_SELF_MONITOR` set, as `update_audio_output_dup_refs` would return early without doing anything.

Also what is the difference between `user_ignore_duplicate_audio` and `ignore_duplicate_audio`? Is this essentially faking a user action for `apply_audio_action`?

PatTheMav:
As far as I understand this change, it will effectively ignore the entire source if it taps into the current monitoring output source?

I have some UX concerns around that: If I add an audio output capture source to my scene setup, I would expect it to be captured. OBS just internally deciding not to capture it (because I have the same device set as my monitoring device) is confusing because there is nothing in the program that would suggest to me that the source is ignored.

It should either _prevent_ me from adding the source in the first place or should be added in a disabled state ("Cannot add the capture source because the device is used for monitoring"). 

And the other way around selecting a monitoring device that is currently used as an audio capture device should be blocked or marked with a warning.

In the end "audio doubling" is the effect of explicit user choices and not something we should magically fixed. 

@Warchamp7 WDYT?

PS: This will also be to limited effect on macOS because we don't capture audio from audio devices anymore, but instead capture the entire system output or specific application's output (it's device-agnostic), so this check (and the same check in the CoreAudio-based monitoring implementation) will never match the monitored source.

Fenrirthviti:
It should only be ignoring a source if monitoring is enabled, and the source selected for monitoring output is also being captured by OBS. To an end user, there should not be a situation in which this makes a meaningful difference, or where you would actively want the audio to be doubled in this way. I am unable to review the code to see if that's exactly what it's doing, but that's what the end result should be.

PatTheMav:
> It should only be ignoring a source if monitoring is enabled, and the source selected for monitoring output is also being captured by OBS. To an end user, there should not be a situation in which this makes a meaningful difference, or where you would actively want the audio to be doubled in this way. I am unable to review the code to see if that's exactly what it's doing, but that's what the end result should be.

That's a choice we can make (monitoring trumps scene setup), but nothing in the app signals that. We should set a "muted" or "disabled" state on the source to communicate its state. Otherwise we have an internal state of source ("ignored") that is hidden from the user.

libobs/obs-source.c
cg2121: > Also what is the difference between user_ignore_duplicate_audio and ignore_duplicate_audio? Is this essentially faking a user action for apply_audio_action?

The ignore_duplicate_audio variable is used as async in the apply_audio_actions function. And user_ignore_duplicate_audio is used to check to if the audio is already ignored, so the update_audio_output_dup_refs function isn't ran if this is the case. This is also the case for muted and user_muted variables, for example.

cg2121: > I am confused by this check - ignore_duplicate_audio should never be set for a device that does not have OBS_SOURCE_DO_NOT_SELF_MONITOR set, as update_audio_output_dup_refs would return early without doing anything.

The ignore_duplicate_audio check is used for any audio source that is not OBS_SOURCE_DO_NOT_SELF_MONITOR. If duplicate audio is detected, for example in the media source,  the audio will not sent to the output and only be sent to the audio monitor.

ryantheleach:
The true UX problem, is OBS's users abuse Monitor functionality for something it was never intended for.

When monitoring in OBS, there are 3 main use-cases.

1. Permanently monitoring a source, because it's something OBS is making noise in that I can't hear, because it's running direct to output, and I need to hear it naturally.
2. Does X sound good? (true monitoring in broadcast terms) experienced broadcast users.
3. What does monitor mean?


Usecase 1:

People who are choosing to monitor to an output device that is being captured by OBS, are likely doing so in error, as they can't hear the OBS output, so don't realize it's double broadcasting.

If they have chosen monitor and output, then the likely intent is that they want to be able to hear the source, and output it, even though it would already be captured through the monitor device's source, so muting the capture output would seem appropriate.

Usecase 2: 

People who are choosing to monitor to an output device that is being captured by OBS, likely have the output device being captured as a separate track as a backup in case of audio problems, and I don't like the idea of 'magic' happening that could interfere with people's intended captures. Adding to the fact that Monitoring could sound different then the final output, depending on what stage of the audio mixing is being monitored (with hopeful PR's down the track to add mixing buses and master mixes) In this situation, they would likely want the capture to win, as they wouldn't want to compromise their recording for the sake of monitoring.

People intentionally monitoring something that is being captured already, are likely trying to either isolate the source to find issues or sound check, or hear things before later mixing is taking place. 

This user would be mad if their multi-track recordings suddenly didn't have sound on 1 track because they were monitoring something, so it MUST be the monitor that is rejected.

Usecase 3:

People unintentionally monitoring a source to the same output are probably trying to learn what the word 'monitor' means in OBS. It would make sense to educate the user in this case.

---

TLDR; there's good reasons to support either argument, so I propose that OBS simply doesn't fix this bug, but instead educates the user subtly if it's being done from the UI.

ryantheleach:
I'd propose that the UX that should happen instead, (and it's a big change) is that when adding a new audio source, when going to properties, that instead of going to advanced audio properties, where you are presented with a large intimidating mixing table, you can select multiple output devices for a given source.

These output devices would be a combination of virtual output devices (tracks) and physical output devices.

By default, 
1. microphones and recording devices would output to Track 1.
2. Line-in devices would default to Track 1 and the default Audio Output Device (if not captured).
3. Output devices would default to Track 1.
4. Browser sources, media sources, etc, would default to Track 1 and the default Audio Output Device (if not captured).

None of this would be considered monitoring, and if selecting a device that is being otherwise captured, a conflict warning would appear showing which devices are being double broadcast.

In broadcast terms, OBS currently conflates physical output devices as submixes.

Doing this would reveal the intent behind what the users wants when adding various audio sources.

---

Monitoring would then be finally free to be what it is supposed to be, a broadcast-level monitoring function that allows you to soundcheck and isolate audio sources. (not necessarily solo)

Warchamp7:
> > It should only be ignoring a source if monitoring is enabled, and the source selected for monitoring output is also being captured by OBS. To an end user, there should not be a situation in which this makes a meaningful difference, or where you would actively want the audio to be doubled in this way. I am unable to review the code to see if that's exactly what it's doing, but that's what the end result should be.
> 
> That's a choice we can make (monitoring trumps scene setup), but nothing in the app signals that. We should set a "muted" or "disabled" state on the source to communicate its state. Otherwise we have an internal state of source ("ignored") that is hidden from the user.

The intended goal is that we assume if a user adds a source that produces audio, they only want that audio in the output once.

Example: A user has their Desktop Audio device in the output and they add a media source. By default that media source only goes to the output. As a result, the user cannot hear it.

Now the user has their monitoring device set to their Desktop Audio device. If the user sets the media source to monitor and output, it will now be in the final output mix twice. Once from directly going to the output, and secondly from playing out of the monitoring device, which is then itself going to the output.

Most often the goal of a user in this scenario is to be able to hear a source themself and also output it to their stream or recording.

If they are trying to do true audio monitoring, it will be to another device that is most likely not included in the output mix.

There should never be a scenario where this causes a sources audio to actually disappear. It should only be preventing it being in the output twice.

The weird case we CAN end up with is when a user mutes a source that is also monitored to a device that is still outputting. In this case, the source would still be audible in the output. There was a PR somewhere that added an indicator icon for communicating this case.

Ultimately, the intent is for users to be able to say
- "I want this audio to go to stream" (Mute/Unmute toggle)
- "I want to be able to hear this audio myself" (Monitor toggle)
- without being able to footgun themselves into the audio being in the output twice.

PatTheMav:
In theory one should not be able to add a source for the device selected as monitoring nor should one be able to select a device that is being captured as a monitoring device.

If this is a bridge too far for some reason, then I agree that if the monitoring device is also added as a source to a scene, then all monitoring for all sources should be disabled to avoid the audio loop effect.

ryantheleach:
> > > It should only be ignoring a source if monitoring is enabled, and the source selected for monitoring output is also being captured by OBS. To an end user, there should not be a situation in which this makes a meaningful difference, or where you would actively want the audio to be doubled in this way. I am unable to review the code to see if that's exactly what it's doing, but that's what the end result should be.
> > 
> > 
> > That's a choice we can make (monitoring trumps scene setup), but nothing in the app signals that. We should set a "muted" or "disabled" state on the source to communicate its state. Otherwise we have an internal state of source ("ignored") that is hidden from the user.
> 
> The intended goal is that we assume if a user adds a source that produces audio, they only want that audio in the output once.
> 
> Example: A user has their Desktop Audio device in the output and they add a media source. By default that media source only goes to the output. As a result, the user cannot hear it.
> 
> Now the user has their monitoring device set to their Desktop Audio device. If the user sets the media source to monitor and output, it will now be in the final output mix twice. Once from directly going to the output, and secondly from playing out of the monitoring device, which is then itself going to the output.
> 
> Most often the goal of a user in this scenario is to be able to hear a source themself and also output it to their stream or recording.
> 
> If they are trying to do true audio monitoring, it will be to another device that is most likely not included in the output mix.
> 
> There should never be a scenario where this causes a sources audio to actually disappear. It should only be preventing it being in the output twice.
> 
> The weird case we CAN end up with is when a user mutes a source that is also monitored to a device that is still outputting. In this case, the source would still be audible in the output. There was a PR somewhere that added an indicator icon for communicating this case.
> 
> Ultimately, the intent is for users to be able to say
> 
> * "I want this audio to go to stream" (Mute/Unmute toggle)
> * "I want to be able to hear this audio myself" (Monitor toggle)
> * without being able to footgun themselves into the audio being in the output twice.

I can respect this, however it needs to work reliably in multi-track recordings. e.g. still output to track 1, even if track 2 was fixed because of doubling.

cg2121:
This is not the correct fix as audio monitoring is being output to the stream with this code, instead of the original source audio when doubling occurs. With audio monitoring the audio can be out of sync. I think a better solution would be to have a UI warning when doubling occurs, instead of automatically fixing the issue.

PR #9673, opened 2023-10-07
cmake: Depend entirely on pkg-config if found for VPL

### LINKED ISSUE #9573: Crash on QSV Recording start
### Operating System Info

Ubuntu 22.04

### Other OS

PopOS

### OBS Studio Version

30.0.0-beta2

_No response_

### OBS Studio Log URL

https://obsproject.com/logs/6NfDajK3QJACgmxo

### OBS Studio Crash Log URL

_No response_

### Expected Behavior

Selected QSV encode, expected recording to start using Arc GPU

### Current Behavior

When QSV settings are selected, AV1, 264 or 265, OBS closes without showing an error. In the terminal is the log linked before, and this last line not shown in the log file:
`obs: symbol lookup error: /usr//lib/x86_64-linux-gnu/obs-plugins/obs-qsv11.so: undefined symbol: MFXLoad`

### Steps to Reproduce

Testing was done on a fresh install of PopOS.
1. Install Intel GPU drivers from this link: https://dgpu-docs.intel.com/driver/installation.html#ubuntu-install-steps
```
sudo apt install bison flex intel-fw-gpu intel-gsc intel-level-zero-gpu intel-metrics-discovery intel-metrics-library level-zero libfl-dev libfl2 libigc1 libigdfcl1 libmetee libsigsegv2 libz3-4 m4 xpu-smi
sudo apt install intel-opencl-icd intel-level-zero-gpu level-zero intel-media-va-driver-non-free libmfx1 libmfxgen1 libvpl2 libegl-mesa0 libegl1-mesa libegl1-mesa-dev libgbm1 libgl1-mesa-dev libgl1-mesa-dri libglapi-mesa libgles2-mesa-dev libglx-mesa0 libigdgmm12 libxatracker2 mesa-va-drivers mesa-vdpau-drivers mesa-vulkan-drivers va-driver-all vainfo hwinfo clinfo
```

2. Install OBS beta using the PPA:
```
sudo add-apt-repository ppa:obsproject/obs-studio-unstable
sudo apt install obs-studio
```

3. Change recording settings to use QSV
4. Start recording, which causes OBS to close.

Not sure yet if this is just an issue with my setup, but QSV seems to work fine in ffmpeg (using jellyfin-ffmpeg6). Would be great to get this working with an Arc card in linux, thanks

kkartaltepe:
### Description

Released versions of VPL have a pkg-config that orders dependencies before the main VPL library. This causes our library searching with pkg-config hints to instead search for the first dependency of VPL.

Instead just import the pkg-config definition directly and use it. This is a bit more correct as all the dependencies will be correctly linked.

### Motivation and Context

fixes #9573.

### How Has This Been Tested?

builds and links vpl/dl/stdc++ instead of just vpl so it seems to work.

### Types of changes

### Checklist:


.github/scripts/utils.zsh/setup_linux
```diff
@@ -97,7 +97,7 @@ sudo apt-get install -y --no-install-recommends \
   libasound2-dev libfdk-aac-dev libfontconfig-dev libfreetype6-dev libjack-jackd2-dev \
   libpulse-dev libsndio-dev libspeexdsp-dev libudev-dev libv4l-dev libva-dev libvlc-dev \
   libpci-dev libdrm-dev \
-  nlohmann-json3-dev libwebsocketpp-dev libasio-dev libvpl-dev libqrcodegencpp-dev
+  nlohmann-json3-dev libwebsocketpp-dev libasio-dev libvpl-dev libvpl2 libqrcodegencpp-dev
```
PatTheMav, line 100: Why do we need to install `libvpl2` in addition to the dev package? Is that necessary for compilation?

cmake/Modules/FindVPL.cmake
```diff
@@ -44,7 +44,11 @@ The following cache variables may also be set:
 
 find_package(PkgConfig QUIET)
 if(PKG_CONFIG_FOUND)
-  pkg_search_module(_VPL QUIET vpl)
+  pkg_search_module(_VPL IMPORTED_TARGET QUIET vpl)
+  if(${_VPL_FOUND})
```
PatTheMav, line 48: You can use `if(_VPL_FOUND)` here.

PatTheMav: Also if I understood the issue right, this is just an issue in specific distributions and is fixed by later variants. As this circumvents the entire finder (and also the feature summary) I'd rather have this workaround limited to those "bad" distributions then.

kkartaltepe: Ace mentioned we just use one recipe for all of ubuntu, so its difficult to implement that there. In this case its just more correct then the rest of our finder so I think its fine to do this everywhere, even where pkg-configs are not formatted weirdly.

.github/scripts/utils.zsh/setup_linux
kkartaltepe: We need the library to link it. Since we only linked libdl previously and dont set errors on undefined functions it was fine to build with broken links.

For whatever reason ubuntu does not have a dependency between the -dev and actual library packages. So we need to manually specify it.

PatTheMav (approved):

PR #9638, opened 2023-09-28
mac-virtualcam: Always replace camera extension; Treat extension installation cancellation as error 

gxalpha:
### Description

Current code assumes that a newer version of the extension is always better. However that's not true - when the user installs an old version of OBS, we should also install the old version of the camera extension to have a version compatible with the installed OBS.
For normal users this method is only called when the versions in the Info.plist do not match. In system extensions developer mode it's called every time an installation is requested (meaning on every start), which is probably desired as well; and by never returning "ReplacementActionCancel" we can now always treat "ErrorRequestCanceled" as an error and not something that we requested to happen.

OSSystemExtensionErrorRequestCanceled is called when the activation delegate returns OSSystemExtensionReplacementActionCancel on request actionForReplacingExtension withExtension. As per above we no longer request cancellation, so we can be sure that receiving a "cancelled" error is an actual error and not something that is supposed to happen.

### Motivation and Context

Fixes #9636

### How Has This Been Tested?

macOS 14
Installed OBS 30.0 Beta 3 and allowed the virtualcam extension. Replaced OBS with my own build (which has a lower build number). Whereas before it would cancel the installation, it now replaces the extension.
Activated system extension developer mode and started OBS. Saw that the log reports that the installation request was cancelled, but the virtual camera still works.

Before, the first scenario would cancel the request.
The cancellation would result in an unintended error message on virtualcam start and prevent the virtualcam from starting.

### Types of changes

- Bug fix (non-breaking change which fixes an issue)

### Checklist:

- [x] My code has been run through [clang-format](https://github.com/obsproject/obs-studio/blob/master/.clang-format).
- [x] I have read the [**contributing** document](https://github.com/obsproject/obs-studio/blob/master/CONTRIBUTING.rst).
- [x] My code is not on the master branch.
- [x] The code has been tested.
- [x] All commit messages are properly formatted and commits squashed where appropriate.
- [x] I have included updates to all appropriate documentation.

PR #9675, opened 2023-10-07
mac-virtualcam: Check result of finished extension installation

gxalpha:
### Description

Adds a switch to distinguish between the possible results that could occur. While we expect to received RequestCompleted for our camera extension (RequestWillCompleteAfterReboot is usually more of a network extension thing), we can't be sure that this always is what we receive unless we check it.

I may have overdone it a bit by adding the locale string but if we're adding this handling we may as well also tell the user about it.

### Motivation and Context

@PatTheMav suggested this in #9638 

### How Has This Been Tested?

macOS 14.
Tested that the virtualcam still works as before because we always receive `OSSystemExtensionRequestCompleted`.
Tested the `OSSystemExtensionRequestWillCompleteAfterReboot` code path by flipping the switch cases, I'm not sure we can get actually get this case in reality (at least currently. who knows what happens in the future with these things).

### Types of changes

### Checklist:

### DIFF
--- a/plugins/mac-virtualcam/src/obs-plugin/plugin-main.mm
+++ b/plugins/mac-virtualcam/src/obs-plugin/plugin-main.mm
@@ -96,8 +96,17 @@ struct virtualcam_data {
 
 - (void)request:(nonnull OSSystemExtensionRequest *)request didFinishWithResult:(OSSystemExtensionRequestResult)result
 {
-    self.installed = YES;
-    blog(LOG_INFO, "macOS Camera Extension activated successfully.");
+    switch (result) {
+        case OSSystemExtensionRequestCompleted:
+            self.installed = YES;
+            blog(LOG_INFO, "macOS Camera Extension activated successfully.");
+            break;
+        case OSSystemExtensionRequestWillCompleteAfterReboot:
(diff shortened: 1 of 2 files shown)


plugins/mac-virtualcam/src/obs-plugin/plugin-main.mm
```diff
@@ -108,8 +108,20 @@ - (void)request:(nonnull OSSystemExtensionRequest *)request didFailWithError:(no
...
+            self.installed = YES;
+            blog(LOG_INFO, "macOS Camera Extension activated successfully.");
+            break;
+        case OSSystemExtensionRequestWillCompleteAfterReboot:
+            self.lastErrorMessage =
+                [NSString stringWithUTF8String:obs_module_text("Error.SystemExtension.CompleteAfterReboot")];
+            blog(LOG_INFO, "macOS Camera Extension will activate after reboot.");
+            break;
+        default:
+            blog(LOG_INFO, "macOS Camera Extension activation finished with unknown result %ld", (long) result);
```
PatTheMav, line 122: I don't think we need a `default` branch here because the switch statement should be exhaustive just with the two values checked above.

PR #9679, opened 2023-10-09
UI: Convert QTStr macro to inline function

gxalpha:
### Description

Converts the `QTStr` function-like macro to 

### Motivation and Context

Using inline functions improves IDE integration and avoids potential pitfalls of that using macros has.
Had this change lying around for a while but got reminded of it again yesterday when someone linked this article (which also explains a bit more of the rationale behind using inline functions instead of function-like macros): https://wiki.sei.cmu.edu/confluence/display/c/PRE00-C.+Prefer+inline+or+static+functions+to+function-like+macros

### How Has This Been Tested?

macOS 14, compiled the program.
Ran the program to see that strings still work. Was also able to breakpoint the function as it now no longer is a macro :)

### Types of changes

### Checklist:

### DIFF
--- a/UI/obs-app.hpp
+++ b/UI/obs-app.hpp
@@ -250,7 +250,10 @@ inline const char *Str(const char *lookup)
 {
 	return App()->GetString(lookup);
 }
-#define QTStr(lookupVal) QString::fromUtf8(Str(lookupVal))
+inline QString QTStr(const char *lookupVal)
+{
+	return QString::fromUtf8(Str(lookupVal));
+}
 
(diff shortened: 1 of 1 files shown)


RytoEX (approved):
Looks good to me.

PatTheMav (approved):

PatTheMav:
Should be noted that `inline` is a _request_ for the compiler to inline the function, but it might still elect not to (similar to how the compiler might actually choose to inline a function even though it wasn't explicitly requested).

But at least this way we can let the compiler optimise the code if necessary.

derrod (approved):

PR #9682, opened 2023-10-09
CI: Add support for dependency revisions to Linux CI builds

PatTheMav:
### Description
Adds support for revisions in buildspec file for Linux CI builds.

### Motivation and Context
Updating dependencies without increasing the version number requires a revision suffix to the file name (as hashes will change).

This support is already implemented in CMake for Windows and macOS but was missing support for Linux builds, which is added with this PR.

### How Has This Been Tested?
Tested on Ubuntu 23.04 with added revision to buildspec and without.

### Types of changes

### Checklist:

### DIFF
--- a/.github/scripts/utils.zsh/setup_linux
+++ b/.github/scripts/utils.zsh/setup_linux
@@ -28,9 +28,9 @@ local deps_baseurl
 local deps_label
 local deps_hash
 
-IFS=';' read -r deps_version deps_baseurl deps_label deps_hash <<< \
+IFS=';' read -r deps_version deps_baseurl deps_label deps_hash deps_revision <<< \
   "$(jq -r --arg target "${target}" \
-    '.dependencies["cef"] | {version, baseUrl, "label", "hash": .hashes[$target]} | join(";")' \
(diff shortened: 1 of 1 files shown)


.github/scripts/utils.zsh/setup_linux
```diff
@@ -40,7 +40,7 @@ if (( ! deps_version )) {
 log_group 'Setting up pre-built Chromium Embedded Framework...'
 
 pushd ${project_root}/.deps
-local _filename="cef_binary_${deps_version}_${target//-/_}.tar.xz"
+local _filename="cef_binary_${deps_version}_${target//-/_}${deps_revision:+"_${deps_revision}"}.tar.xz"
```
RytoEX, line 43: Doesn't this need a "v" prefix for revision?

RytoEX (changes requested):

.github/scripts/utils.zsh/setup_linux
PatTheMav: Correct, fixed.

RytoEX (approved):
Looks good to me.

PR #9691, opened 2023-10-11
UI: Use XDG_VIDEOS_DIR instead of HOME if set

derrod:
### Description

Updates `GetDefaultVideoSavePath()` to check `XDG_VIDEOS_DIR` on *nix instead of dumping things into`$HOME`

### Motivation and Context

Want to not clutter `$HOME` by default.

### How Has This Been Tested?

Has not been tested yet.

### Types of changes

### Checklist:

### DIFF
--- a/UI/platform-x11.cpp
+++ b/UI/platform-x11.cpp
@@ -203,7 +203,11 @@ bool GetDataFilePath(const char *data, string &output)
 
 string GetDefaultVideoSavePath()
 {
-	return string(getenv("HOME"));
+	const char *videoPath = getenv("XDG_VIDEOS_DIR");
+	if (!videoPath || videoPath[0] == '\0')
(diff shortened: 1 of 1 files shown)


kkartaltepe:
This was already proposed forever back #3746 fwiw, and i still feel the same... Code seems fine though.

derrod:
I missed that. 

I understand that `XDG_VIDEOS_DIR´ is not part of the XDG standard, but it seems the main desktop environments have adopted this practice, making it a de-facto standard on the Linux desktop. Perhaps there's a newer more standardised way (portals?) to get the default directories.

The suggested approach of using the `QStandardPaths` seems otherwise like perhaps a better choice. I'll close this for now and revisit as part of a larger change to how we handle default directories and output settings down the road.

PR #9693, opened 2023-10-11
mac-virtualcam: Log when extension replacement is requested

gxalpha:
### Description

Adds an info log when the system requests a replacement.

### Motivation and Context

We currently are seeing some issues of users not being able to use the macOS virtualcam at all.
This is known to happen after the extension is replaced. I don't *think* this is what happening to those users, but would like to know for sure.
Would be good to have this in RC2 so that effected users can post logs that would show whether this is happening.

### How Has This Been Tested?

macOS 14
Log line appears on upgrade.

### Types of changes

### Checklist:


plugins/mac-virtualcam/src/obs-plugin/plugin-main.mm
```diff
@@ -68,6 +68,7 @@ - (OSSystemExtensionReplacementAction)request:(nonnull OSSystemExtensionRequest
                   actionForReplacingExtension:(nonnull OSSystemExtensionProperties *)existing
                                 withExtension:(nonnull OSSystemExtensionProperties *)ext
 {
+    blog(LOG_INFO, "mac-camera-extension: Replacement requested, replacing existing extension.");
```
PatTheMav, line 71: I'd rather have the version information logged as well, that way we can tell whether the replacement is a downgrade vs an upgrade (as we allow both, we also need to check for both).

gxalpha: Ah yes, good idea

RytoEX: Looks like this was addressed?

PatTheMav: Yep.

PatTheMav (approved):

PR #9694, opened 2023-10-11
build-aux: Use fallback-x11 instead of x11

Etaash-mathamsetty:
### Description
Fixes a warning when compiling/building the obs flatpak
fallback-x11 only enables x11 if the application doesn't use wayland, and instead falls back to x11. If the app uses wayland then the x11 socket is never made available I am assuming.

### Motivation and Context
removes a warning

### How Has This Been Tested?
Testing is unnecessary because it just uses the same behavior but removes a build time warning

### Types of changes
Bug fix

### Checklist:

### DIFF
--- a/build-aux/com.obsproject.Studio.json
+++ b/build-aux/com.obsproject.Studio.json
@@ -6,7 +6,7 @@
     "command": "obs",
     "finish-args": [
         "--socket=wayland",
-        "--socket=x11",
+        "--socket=fallback-x11",
         "--socket=pulseaudio",
         "--device=all",
         "--share=network",


gxalpha:
Please adjust the commit message according to our [commit guidelines](https://github.com/obsproject/obs-studio/blob/master/CONTRIBUTING.rst#commit-guidelines).
Please also expand on the description/motivation here. How is `fallback-x11` different from `x11` (and in particular, why is it the correct fix)?, etc.

GeorgesStavracas:
The reason I specifically left `x11` and not `fallback-x11` is because some plugins may crash OBS if they can't connect to the X11 socket. One such example is the [Input Overlay plugin](https://github.com/flathub/com.obsproject.Studio.Plugin.InputOverlay).

Can you check if that's still the case?

Etaash-mathamsetty:
> The reason I specifically left `x11` and not `fallback-x11` is because some plugins may crash OBS if they can't connect to the X11 socket. One such example is the [Input Overlay plugin](https://github.com/flathub/com.obsproject.Studio.Plugin.InputOverlay).
> 
> Can you check if that's still the case?

ah I didn't know about this

GeorgesStavracas:
It would still be fantastic to replace `x11` with `fallback-x11` though, so it's worth checking and maybe even reporting the crash to the plugin authors. But I'm afraid of other plugins that could crash as well. Also, @gxalpha's comment still applies, a more comprehensive commit message would be great to have as well.

Etaash-mathamsetty:
> It would still be fantastic to replace `x11` with `fallback-x11` though, so it's worth checking and maybe even reporting the crash to the plugin authors. But I'm afraid of other plugins that could crash as well. Also, @gxalpha's comment still applies, a more comprehensive commit message would be great to have as well.

Input overlay still uses x11, I am still wondering why they didn't use evdev though

GeorgesStavracas:
I don't think it's possible to install polkit rules or elevate permissions from a Flatpak sandbox

tytan652:
> Input overlay still uses x11, I am still wondering why they didn't use evdev though. It does need permissions, but polkit can handle that right?

No, the author/maintainers of the plugin will have to find (or contribute to) a new way to do it without punching holes in the sandbox and relying on the x11 socket to work on both X11 and Wayland.

If a plugin requires the x11 socket, it must be fixed to not make OBS crash under Wayland with no access to the x11 socket.

build-aux/com.obsproject.Studio.json
```diff
@@ -6,7 +6,7 @@
     "command": "obs",
     "finish-args": [
         "--socket=wayland",
-        "--socket=x11",
```
tytan652, line 9: Keep unresolved until the commit tittle is changed.

tytan652 (changes requested):

tytan652:
Input Overlay still crash with this kind of change.

RytoEX:
cc @univrsal 
See above about an issue we're seeing when trying to clean up our CI warnings.
If I had to guess, the issue is with libuiohook, but I haven't actually checked.

univrsal:
I'm pretty sure it's libuiohook. It could also be SLD2 but I assume that library behaves correctly. I can add a check and not start libuiohook if obs is running under wayland. 
```C
getenv("WAYLAND_DISPLAY") || obs_get_nix_platform() == OBS_NIX_PLATFORM_WAYLAND
```
Would either of these checks be enough?

tytan652:
@univrsal

You should check if you are on X11/Xorg since it's the only supported "platform".
```
obs_get_nix_platform() != OBS_NIX_PLATFORM_X11_EGL
```
We dropped `_GLX` since a while so it can be ignored.

univrsal:
I've added the check. Seems to work from what I can tell but I can't test if it fixes the crash as I don't have the flatpak build.

RytoEX:
@GeorgesStavracas Is this something we want to try out and see what breakages occur in the wild? If so, when would you feel it's appropriate to do so?

GeorgesStavracas:
I think we can try merging this right after the current release

tytan652:
Blocked by #11019

build-aux/com.obsproject.Studio.json
RytoEX: I've fixed the commit message and PR title.

RytoEX (approved):
Approving based on prior conversation with @GeorgesStavracas .

PR #9698, opened 2023-10-12
cmake: Explicitly fail if macOS SDK is too old

gxalpha:
### Description

Adds a check to explicitly make sure that the macOS SDK being used is recent enough to build OBS. OBS code currently has ifdefs for older SDKs in some but not all places and relies on some frameworks existing that do not exist on older versions, this should clarify what we require.
This commit sets the minimum SDK to 13.0. This can be increased in the future if required.

I've put this code in `compilerconfig.cmake` because it feels to me like it makes the most sense, but am not really sure about that. Happy to change this if desired.

### Motivation and Context

The ifdef situation is a bit unclear. We have 12.3 ifdefs in some places, but then not others, same with 13.0.
The [wiki](https://github.com/obsproject/obs-studio/wiki/Build-Instructions-For-Mac#prerequisites) specifies Xcode 14.2 (which has the macOS 13.1 SDK) as the minimum, so let's use that explicitly.

This should also enable future cleanups of ifdefs that check for versions lower than our minimum, which still exist in some places.

### How Has This Been Tested?

macOS 14 with Xcode 15.0 and macOS 14.0 SDK
Generated the project to confirm it still works.
Manually increased the minimum to 14.1 and had it fail with:
> CMake Error at cmake/macos/compilerconfig.cmake:24 (message):
  Your macOS SDK version (14.0) is too low.  The macOS 14.1 SDK is required
  to build OBS.

### Types of changes

### Checklist:

### DIFF
--- a/cmake/macos/compilerconfig.cmake
+++ b/cmake/macos/compilerconfig.cmake
@@ -15,6 +15,20 @@ if(NOT CMAKE_OSX_ARCHITECTURES)
 endif()
 set_property(CACHE CMAKE_OSX_ARCHITECTURES PROPERTY STRINGS arm64 x86_64)
 
+# Make sure the macOS SDK is recent enough for OBS
+set(OBS_MACOS_MINIMUM_SDK "13.1") # Keep in sync with Xcode
+set(OBS_MACOS_MINIMUM_XCODE "14.2") # Keep in sync with SDK
+message(DEBUG "macOS SDK Path: ${CMAKE_OSX_SYSROOT}")
+string(REGEX MATCH ".+/MacOSX.platform/Developer/SDKs/MacOSX([0-9]+\.[0-9])+\.sdk$" _ ${CMAKE_OSX_SYSROOT})
+set(OBS_MACOS_CURRENT_SDK ${CMAKE_MATCH_1})
+message(DEBUG "macOS SDK version: ${OBS_MACOS_CURRENT_SDK}")
+if(OBS_MACOS_CURRENT_SDK VERSION_LESS OBS_MACOS_MINIMUM_SDK)
+  message(
+    FATAL_ERROR
+      "Your macOS SDK version (${OBS_MACOS_CURRENT_SDK}) is too low. The macOS ${OBS_MACOS_MINIMUM_SDK} SDK (Xcode ${OBS_MACOS_MINIMUM_XCODE}) is required to build OBS."
+  )
+endif()
+
 if(XCODE)
   # Enable dSYM generator for release builds
   string(APPEND CMAKE_C_FLAGS_RELEASE " -g")


cmake/macos/compilerconfig.cmake
```diff
@@ -15,6 +15,18 @@ if(NOT CMAKE_OSX_ARCHITECTURES)
...
 
+message(DEBUG "macOS SDK Path: ${CMAKE_OSX_SYSROOT}")
+string(REGEX MATCH ".+/MacOSX.platform/Developer/SDKs/MacOSX([0-9]+\.[0-9])+\.sdk$" _ ${CMAKE_OSX_SYSROOT})
+set(OBS_MACOS_SDK_CURRENT ${CMAKE_MATCH_1})
+message(DEBUG "macOS SDK version: ${OBS_MACOS_SDK_CURRENT}")
+set(OBS_MACOS_SDK_MINIMUM "13.1")
+if(OBS_MACOS_SDK_CURRENT VERSION_LESS OBS_MACOS_SDK_MINIMUM)
+  message(
+    FATAL_ERROR
+      "Your macOS SDK version (${OBS_MACOS_SDK_CURRENT}) is too low. The macOS ${OBS_MACOS_SDK_MINIMUM} SDK is required to build OBS."
```
PatTheMav, line 26: Maybe mention the minimum Xcode version that is required to be installed to fix this issue.

RytoEX: Would that require us to manually update this message each time or separately track the minimum Xcode version?

gxalpha: Both Xcode and macOS SDK Version would need to be updated here.
We could add this to buildspec or the default cmake preset but I’m not sure it makes sense in either? cc @PatTheMav

PatTheMav: Preset would be best suited, it's where we set the deployment target as well.

cmake/macos/compilerconfig.cmake
```diff
@@ -15,6 +15,21 @@ if(NOT CMAKE_OSX_ARCHITECTURES)
 endif()
 set_property(CACHE CMAKE_OSX_ARCHITECTURES PROPERTY STRINGS arm64 x86_64)
 
+if(NOT OBS_MACOS_MINIMUM_SDK OR NOT OBS_MACOS_MINIMUM_XCODE)
+  message(FATAL_ERROR "Required variable not found.")
+endif()
```
gxalpha, line 20: Wondering if this is fine or if we should explicify the preset situation somewhere else anyways.
Problem is that otherwise, an empty string is always a lower version than a string with a version, so we wouldn't catch the people who don't set presets.

cmake/macos/compilerconfig.cmake
gxalpha: Added as a preset variable.

cmake/macos/compilerconfig.cmake
PatTheMav: As discussed off-thread I'd actually lean towards hardcoding those values in CMake code and not use cache variables from the preset.

The expectation for cache variables is that they are set by the developer (the preset is just a convenience feature), so implementing this kinda expects the developer to always provide a minimum SDK and Xcode version.

But we don't want to make this a developer choice, it is a hard limit set by the project. As such it should be coded directly into the `compilerconfig` file.

gxalpha: This should be addressed.

cmake/macos/compilerconfig.cmake
gxalpha: Reverted this change.

PatTheMav (approved):

RytoEX (approved):
Looks good to me.

PR #864, opened 2017-03-19
UI: Add ability to copy scenes to other scene collections

RytoEX:
**Note:  This PR is not ready to be merged.**  I'm submitting it to get some eyes on it for review, testing, and cleanup.

This commit adds the ability for users to copy an entire scene and all of its child sources to another scene collection. This is for parity with OBS Classic.

I've been staring at this code and adjusting it on and off for a few months.  The functionality works, and I've compiled and tested it on Windows 10 64-bit.  At this point I've been looking at it so long that my eyes are just glossing over it now, so I'm in need of some fresh perspective.  I could use some feedback on everything here - project code style compliance, efficiency, lang strings, function/variable names, expected/actual functionality, user experience, etc.  I'm aware that there are still a lot of comments in the code.  Most of those I left in place intentionally for the purpose of this review.

To preemptively answer the questions, "Why so much new code? Why not use existing functions?"  Most of the new code in this PR was written because existing functions for copying scenes/sources seemed to rely on or affect the current UI state.  I needed to load and save data without those caveats, and I didn't want to alter any existing functions.

Any feedback is appreciated, though constructive feedback would _really_ be appreciated.  I will try to answer any questions as best as I can.

PR #9718, opened 2023-10-17
UI: Add ability to copy scenes to scene collections

cg2121:
### Description
This adds a menu in the scenes context menu to duplicate scenes to a different scene collection.

Menu:
Screenshot from 2023-10-16 19-23-37

Success Dialog:
Screenshot from 2023-10-16 19-27-50

### Motivation and Context
- https://ideas.obsproject.com/posts/20/add-the-ability-to-copy-and-paste-duplicate-sources-and-or-scenes-across-scene-collections

- This feature was also in OBS Classic

- I am not currently satisfied with the UI, as it is just a context menu item, which is not optimal. @GeorgesStavracas will have to create a mockup for a UI, so a scene can be duplicated to multiple scene collections at once.

### How Has This Been Tested?
Copied scene to scene collection. Switched to that scene collection and made sure the scene showed up properly.

### Types of changes

### Checklist:


derrod:
Given

> I am not currently satisfied with the UI, as it is just a context menu item, which is not optimal. @GeorgesStavracas will have to create a mockup for a UI, so a scene can be duplicated to multiple scene collections at once.

Should this be a draft?

Also there are a few other concerns I have:
- No checks for name collisions
- No checks for UUID collisions
- Right now it seems to use the current scene rather then the selected one?

cg2121:
Converted to draft

cg2121:
The updated code generates new uuids for copied sources and also generates new names, if names match between scene collections.

RytoEX:
For what it's worth, I had [previously tried to implement this](https://github.com/obsproject/obs-studio/pull/864) as dialog boxes, and I didn't like that UX either. One of the challenges this approach has would be with very large lists of scene collections causing a doubled-up/overflowed/multi-column menu. You can see this in the Scene Collections menu itself if you have a lot of Scene Collections. This could probably be solved by using a scrollable dropdown or a list box/view instead, though that raises the question of where it would go as they don't work well inside context menus. Additionally, it's my understanding, though perhaps I am mistaken, that we do not want to add more context-menu-only options, and that any new context menu entries should be accessible from another location. Hopefully a UI/UX discussion would be helpful here.

Warchamp7:
This will likely make more sense as part of a future revision to the Scene Collection Management rework #8446

Warchamp7:
Closing old PRs that require a rebase for the big frontend reorganization we did. I also don't believe this would be accepted as the context menu UX it currently has.

PR #9728, opened 2023-10-17
mac-capture: If window is not found, look for a matching one

jnkel:
### Description
If a macOS Screen Capture is started in windowed mode, and the target window does not exist, try to find one with a matching title and application bundle ID.

### Motivation and Context
I use a window capture for speedrun splits. Prior to this change, I had to reconfigure my window capture every time I started runs, due to OBS not automatically recognizing the window. This patch matches the behavior of the deprecated macOS Window Capture source (except it still does not watch for the window to be opened if it is closed when the capture starts).

### How Has This Been Tested?
I tested on my local machine, an M1 MBP running macOS 14.0. I verified it correctly finds a matching window, and falls back gracefully to the old behavior if no matching window is found. I also checked to make sure it does not break the behavior of the display capture or application capture.

### Types of changes

### Checklist:

### DIFF
--- a/plugins/mac-capture/mac-sck-common.h
+++ b/plugins/mac-capture/mac-sck-common.h
@@ -65,6 +65,7 @@ struct screen_capture {
     ScreenCaptureAudioStreamType audio_capture_type;
     CGDirectDisplayID display;
     CGWindowID window;
+    NSString *window_title;
     NSString *application_id;
 };
 
--- a/plugins/mac-capture/mac-sck-video-capture.m
+++ b/plugins/mac-capture/mac-sck-video-capture.m
@@ -74,7 +74,7 @@ static void sck_video_capture_destroy(void *data)
     bfree(sc);
 }
 
-static bool init_screen_stream(struct screen_capture *sc)
+static bool init_screen_stream(struct screen_capture *sc, obs_data_t *settings)
 {
     SCContentFilter *content_filter;
     if (sc->capture_failed) {
@@ -141,6 +141,16 @@ static bool init_screen_stream(struct screen_capture *sc)
                         break;
                     }
                 }
+                // If no window with a matching ID exists, look for one with a matching title and owning application
+                if (target_window == NULL) {
+                    for (SCWindow *window in sc->shareable_content.windows) {
+                        if ([window.title isEqualToString:sc->window_title] &&
+                            [window.owningApplication.bundleIdentifier isEqualToString:sc->application_id]) {
+                            target_window = window;
+                            break;
+                        }
+                    }
+                }
             } else {
                 target_window = [sc->shareable_content.windows objectAtIndex:0];
                 sc->window = target_window.windowID;
@@ -150,6 +160,11 @@ static bool init_screen_stream(struct screen_capture *sc)
             if (target_window) {
                 [sc->stream_properties setWidth:(size_t) target_window.frame.size.width];
                 [sc->stream_properties setHeight:(size_t) target_window.frame.size.height];
+                sc->window_title = [target_window.title retain];
+                sc->application_id = [target_window.owningApplication.bundleIdentifier retain];
+                obs_data_set_int(settings, "window", target_window.windowID);
+                obs_data_set_string(settings, "window_title", sc->window_title.UTF8String);
+                obs_data_set_string(settings, "application", sc->application_id.UTF8String);
             }
 
         } break;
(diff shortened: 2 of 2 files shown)


RytoEX:
Historically, we have not implemented such fallbacks because there is no way to get explicit consent for capturing a different window than the one originally selected.

cc @PatTheMav @gxalpha who are more familiar with macOS paradigms than I am.

NobodyNada:
Yeah, that makes a lot of sense -- it's a valid security concern. However, looking at the code the deprecated Mac capture plugin had this logic, and it seems the Windows capture plugin has similar logic too, unless I'm misreading it: https://github.com/obsproject/obs-studio/blob/master/libobs/util/windows/window-helpers.c#L505

There are a couple of mitigating factors: the code requires both window title and application bundle ID to match, so although we might capture a different *window* than the one originally selected, we will never capture a different application. Secondly, my code currently only checks for matching windows when the capture plugin is started, so it won't unexpectedly grab a window that you opened mid-stream (although the old mac capture plugin *did* continually check for matching windows). However, it would certainly still be possible for a window to get unexpectedly selected by a window capture in an inactive scene.

For my own personal use-case, if I select my scene for "capture card + speedrun splits", I know that I want OBS to capture whatever speedrun split window that I have open. I'm not going to have some other window full of sensitive information that happens to have the same application bundle ID and title as my splitter, so I would rather OBS guess and save me the trouble of manually reselecting my splits each time. And if I ever add additional window captures for things like input display or room timer, then I would rather not have to reselect several windows each time I start recording or streaming.

Perhaps a solution would be a checkbox in the window capture properties, allowing the user to explicitly give that consent for OBS to search for a window?

gxalpha:
> it's a valid security concern. However, looking at the code the deprecated Mac capture plugin had this logic

That doesn’t make it less of a security concern. Imo this was a mistake in the old capture plugin that doesn’t need repeating.

If this PR does what I think it does (keep in mind I have yet to test if that’s the case) I’d personally be against it.

PatTheMav:
As explained before, just switching the capture source based on a heuristic (i.e. "same window title") seems innocuous until it results in a bad leak on a stream.

Given the new share picker in Sonoma, it also seems that Apple intends users in general to _push_ content to a capture application versus a capture application pulling that content (the intent is more clear and users can revoke that permission in a central location in the menu bar).

FWIW I do agree that this makes speed running streams annoying. I have never used any speed running software, so I wonder why it needs to spawn a new window at every attempt? Could "Application Capture" with cropping solve that issue?

Debuggernaut:
Hello my good friends!  I hope this message finds you well.  I was wondering if there's some way to use OBS on MacOS without having to manually re-select every window capture source every single time I reboot OBS Studio.

It seems like we need something like this PR to get in for that to be possible

I was poking around at it in a local copy of the source, and was wondering if a solution along these lines would work for you guys:

image

If this looks like something you'd be interested in, I can do a machine translation for all the other locales and send a PR.

https://github.com/obsproject/obs-studio/commit/a489488a29a7e8875973fc0e2559cb18d26359d2

starryalley:
Thanks OP for this PR. I have been using this patch to build my own OBS for my own need: remember saved window captures across macOS system reboots. 

However, it isn't always working because saved windowID means nothing after system reboot. Only when the target window remains open will it be valid and meaningful, which we already know. 

During my extensive testing, when a saved windowID happens to be used again (most likely it will be another window) after system reboot, it will capture the wrong window. This is likely to happen more often when one tries to configure OBS mac-capture source right after the system starts, with limited number of windows open. In this case the windowID is likely to be within 100. (To add more data points, on a macbook with uptime of days with >20 apps running on and off, I usually see windowID in tens of thousands or more) When I configure my mac to auto-start my target apps, along with OBS, at macOS start, the collision of windowID is more likely to happen. 

The idea to fix this for the purpose of being able to capture the saved window across system reboot is to discard the saved windowID completely when creating the mac-capture source, and rely only on window title and owning application's ID matching. I've personally used this modifed patch for my own need.

I understand this "finding window" to capture idea violates the design principles of SCK or MacOS privacy rules, let alone being able to auto-capture across system reboots. So I just leave a comment here for anyone who happens to have the same purpose and encounters the issue. 

Thanks again for making this useful PR.

Debuggernaut:
Has anyone checked into the new picker Apple is trying to make us use?

It seems like it's not going to be able to gracefully handle recovering a complex multi-window configuration after a reboot either

I wonder how this works on Meld Studio or some of the other streaming apps.. it seems weird to me that this has been so broken for so long and there are only a few people complaining about it, I'm wondering if I'm just using OBS in a weird way

Warchamp7:

PR #9744, opened 2023-10-21
Add support for Lua plugin creators to create subcategories for Sources

karl-police:
### Description
Add support for Lua plugin creators to create subcategories for Sources. These categories always get added after "Group".

Plugins can now include this ``source_def.subcategory = "Test"`` a new property that can get passed to ``obs_register_source``.

This creates sub categories, and you can even put it inside Deprecated.

Here's an example .lua plugin to test this new feature: https://pastebin.com/gc2AeJfu

### Motivation and Context
This implements my own feature request after asking in the Discord, how I can put Plugin Sources into a Subcategory. Being told that I can't or being misunderstood.

This motivated me to add this feature.

I didn't like the idea that there's nothing that a Plugin Developer can do, **to not mix the original Sources with the custom-made sources**. Plugin Developers should have the freedom to decide how their Source should show up, which is also a good nice QoL for the user using the Plugin.

And other organization reasons.

So I wanted to add subcategories. Just don't ask whether one can put add subcategories in a subcategories.

### How Has This Been Tested?
Tested several cases like
``source_def.subcategory = nil``
``source_def.subcategory = "Test"``

and if it's not references at all
I had to check for ``\0`` and make it return ``NULL``

There's only one issue. The subcategories don't refresh if you refresh the plugin, everything else does refresh.
That's something with the Lua OBS C implementation script that I didn't figure out so far. Help is appreciated.

There's a test script above as well.

### Types of changes
* New Feature
* Small change in the Documentation

### Checklist:

### DIFF
--- a/UI/window-basic-main.cpp
+++ b/UI/window-basic-main.cpp
@@ -6130,6 +6130,8 @@ QMenu *OBSBasic::CreateAddSourcePopupMenu()
 	bool foundDeprecated = false;
 	size_t idx = 0;
 
+	std::map<const char *, Custom_Subcategory_info> customSubcategory_MAP;
+
 	QMenu *popup = new QMenu(QTStr("Add"), this);
 	QMenu *deprecated = new QMenu(QTStr("Deprecated"), popup);
 
@@ -6169,11 +6171,48 @@ QMenu *OBSBasic::CreateAddSourcePopupMenu()
 		const char *name = obs_source_get_display_name(type);
 		uint32_t caps = obs_get_source_output_flags(type);
 
+		const char *subcategoryName =
+			obs_source_get_subcategory_name(type);
+
+		if (subcategoryName != NULL) {
+
+			if (strcmp(subcategoryName, "Deprecated") == 0) {
+				/* This allows plugin creators to put
+				*  their Sources into Deprecated as well */
+				addSource(deprecated, unversioned_type, name);
+				foundDeprecated = true;
+			} else {
+				// If it doesn't already exist
+				if (customSubcategory_MAP.count(
+					    subcategoryName) == 0) {
+					Custom_Subcategory_info
+						custSubcategory_info;
+
+					QMenu *SubcategoryMenu = new QMenu(
+						QTStr(subcategoryName));
+
+					custSubcategory_info.m_QMenu_ref =
+						SubcategoryMenu;
+
+					customSubcategory_MAP[subcategoryName] =
+						custSubcategory_info;
+				}
+
+				QMenu *submenu =
+					customSubcategory_MAP[subcategoryName]
+						.m_QMenu_ref;
+
+				addSource(submenu, unversioned_type, name);
+			}
+		}
+
 		if ((caps & OBS_SOURCE_CAP_DISABLED) != 0)
 			continue;
 
 		if ((caps & OBS_SOURCE_DEPRECATED) == 0) {
-			addSource(popup, unversioned_type, name);
+			if (subcategoryName == NULL) {
+				addSource(popup, unversioned_type, name);
+			};
 		} else {
 			addSource(deprecated, unversioned_type, name);
 			foundDeprecated = true;
@@ -6189,6 +6228,16 @@ QMenu *OBSBasic::CreateAddSourcePopupMenu()
 	connect(addGroup, &QAction::triggered,
 		[this]() { AddSource("group"); });
 	popup->addAction(addGroup);
+
+	// Custom Subcategories
+	if (customSubcategory_MAP.size() >= 1) {
+		popup->addSeparator();
+
+		for (const auto &entry : customSubcategory_MAP) {
+			QMenu *submenu = entry.second.m_QMenu_ref;
+			popup->addMenu(submenu);
+		}
+	}
 
 	if (!foundDeprecated) {
 		delete deprecated;
--- a/UI/window-basic-main.hpp
+++ b/UI/window-basic-main.hpp
@@ -1281,3 +1281,10 @@ public:
 protected:
 	virtual bool eventFilter(QObject *editor, QEvent *event) override;
 };
+
+#ifndef CUSTOM_SUBCATEGORY_MAP
+#define CUSTOM_SUBCATEGORY_MAP
+struct Custom_Subcategory_info {
+	QMenu *m_QMenu_ref;
+};
+#endif
--- a/deps/obs-scripting/obs-scripting-lua-source.c
+++ b/deps/obs-scripting/obs-scripting-lua-source.c
@@ -599,6 +599,11 @@ static int obs_lua_register_source(lua_State *script)
 
 	info.id = v->id;
 	info.type = (enum obs_source_type)get_table_int(script, -1, "type");
+	info.subcategory = get_table_string(script, -1, "subcategory");
+
+	if (info.subcategory[0] == '\0') {
+		info.subcategory = NULL;
+	}
 
 	info.output_flags = get_table_int(script, -1, "output_flags");
 
--- a/docs/sphinx/reference-sources.rst
+++ b/docs/sphinx/reference-sources.rst
@@ -47,6 +47,10 @@ Source Definition Structure (obs_source_info)
    - **OBS_SOURCE_TYPE_FILTER**     - Filter
    - **OBS_SOURCE_TYPE_TRANSITION** - Transition
 
+.. member:: const char *obs_source_info.subcategory
+
+   Subcategory string for the Source to be put in at in the Popup UI.
+
 .. member:: uint32_t obs_source_info.output_flags
 
    Source output capability flags (required).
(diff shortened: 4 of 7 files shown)


karl-police:
my commit organization has been destroyed by all of this 😭

PatTheMav:
First of all: I applaud the effort - diving straight into C++ _and_ our UI code for this is no easy matter. 👏🏻

_However_.. 😅

I'm not sure whether I'm comfortable with allowing sources to set arbitrary categories for themselves - this has the unfortunate side-effect of sources filling up the source menu with all kinds of submenus, each _maybe_ containing a single entry.

There are a few more architectural things to consider:

* The "deprecated" flag is not a category, it should not be set by a plugin via its "category" string (it's a capability flag for a good reason)
* How would you handle a source that has a category _but is also deprecated_ (you'd need a "Deprecated" sub-menu inside the category sub-menu?
* `std::map` does not retain the order at which items have been added, iirc it uses an alphabetic comparator on map _keys_, so iterating over the map to add the sub menus has "undefined" behaviour in your code (as in "the UI does not define the order at which sub menus should be added")
* Submenus are always appended to the _bottom_ of the source menu, which is a valid choice but I'm not sure it's a good choice

I do agree that the source menu could benefit from sub-menus, but I would not allow sources to define arbitrary names for them. Instead we should probably either look at the capability flags and decide on a category based on that or add a category `ENUM` with specific categories that a source can choose from. 

And of course category support would then need to be added for _all_ sources then, not just script sources (for backwards compatibility you'd need to handle the case that no category was set by the plugin but then also suggest categories for our existing plugins to use).

karl-police:
> First of all: I applaud the effort - diving straight into C++ _and_ our UI code for this is no easy matter. 👏🏻
> 
> _However_.. 😅
> 
> I'm not sure whether I'm comfortable with allowing sources to set arbitrary categories for themselves - this has the unfortunate side-effect of sources filling up the source menu with all kinds of submenus, each _maybe_ containing a single entry.
> 
> There are a few more architectural things to consider:
> 
> * The "deprecated" flag is not a category, it should not be set by a plugin via its "category" string (it's a capability flag for a good reason)
> * How would you handle a source that has a category _but is also deprecated_ (you'd need a "Deprecated" sub-menu inside the category sub-menu?
> * `std::map` does not retain the order at which items have been added, iirc it uses an alphabetic comparator on map _keys_, so iterating over the map to add the sub menus has "undefined" behaviour in your code (as in "the UI does not define the order at which sub menus should be added")
> * Submenus are always appended to the _bottom_ of the source menu, which is a valid choice but I'm not sure it's a good choice
> 
> I do agree that the source menu could benefit from sub-menus, but I would not allow sources to define arbitrary names for them. Instead we should probably either look at the capability flags and decide on a category based on that or add a category `ENUM` with specific categories that a source can choose from.
> 
> And of course category support would then need to be added for _all_ sources then, not just script sources (for backwards compatibility you'd need to handle the case that no category was set by the plugin but then also suggest categories for our existing plugins to use).

I used ``std::map`` not because of a special reason. I literally haven't gone through the C++ book yet, so I am clueless about most of certain stuff. Not sure if Vectors would be a better thing.
``std::map`` sounds also like that it can be changed at any time. Point is to not break ``obs_source_info.subcategory``. ``std::map`` is just the backend.
First I had a _"functioning"_ version where I used ``std::map`` to cache those submenu pointers _for speed_, but scrapped it because the popup-menu gets created everytime anyways, that's it's a waste to even do that. If I parented QMenu's to something that gets _eliminated (popup closes, the var "popup" is gone, but not for saved QMenu's)_ I'd get an access violation crash. 

The "deprecated" thing, is nothing special. All I did in the code was, if ``subcategory`` is ``Deprecated`` then addSource inside deprecated.

If a Source is inside a custom category and they want to deprecate it, they can't create a subcategory inside of a subcategory, because I think this would require some more OOP stuff and system. They can just move it inside Deprecated.
Or make a new plugin and make everyone download the new one. At the end, the user installs the plugin.

But if someone ever wants to add subcategories within subcategories...
Honestly, at that point ``obs_source_info.subcategory`` can just support ``/`` and then you would do ``Submenu1/Submenu2`` and then it's just like Windows Explorer folders.
Anything that is _nested_ looks very complicated to add.



Regarding, whether it's good and nice that developers can create subcategories. You also asked if it doesn't sort of "clog" up too much.
Here's the thing. What's better, clogging subcategories with 500 sources. Or all the 500 sources mixed with all the other sources, and eventually other plugins.
Any sort of modification, like Plugin or Steam Workshop item in some game, eventually adds things and modifies stuff, so it's always expected that there's _stuff added_.


Point of the subcategories is to not mix the original stuff with the custom stuff, to have an organization. The Lua Sources **don't have to be in a subcategory**, it's optional.

You do not have to add category support for all sources. Any new source that gets added simply just needs to add ``subcategory`` in the script. If it doesn't have one, it will still show up in the popup menu, just not in a subcategory. Nothing changes.



The "ENUM" idea isn't bad. You could create presets. _I see Deprecated as such._
You could create a category called "Food & Drinks" and put it in there, yeah... but maybe not everyone is going to like the name of those pre-defined categories.

I'd be for a Plugins category. That would do it too. Aka. simply just copy paste Deprecated and then add, idk... ``subcategory = "Plugin"`` uhhhhh...
Which would then be optional again. If you prefer that. Then the _traffic_ for the Sources land always into _Plugin_ and no-one can create, _"Food & Drinks"_ and a bunch of other random-named categories.


Those submenus from my addition add below "Group" with a separator. Only if there's a source with a subcategory registered. The reason I decided that, is just due to OBS's current style. All the original things that someone needs are shown at first glance when opening the Popup menu.

Having _Item, Item, Item, Folder_ as a design, looks better than Item, _Folder, Item, Item, Folder_.

PatTheMav:
> I used `std::map` not because of a special reason. I literally haven't gone through the C++ book yet, so I am clueless about most of certain stuff. Not sure if Vectors would be a better thing. `std::map` sounds also like that it can be changed at any time. Point is to not break `obs_source_info.subcategory`. `std::map` is just the backend. First I had a _"functioning"_ version where I used `std::map` to cache those submenu pointers _for speed_, but scrapped it because the popup-menu gets created everytime anyways, that's it's a waste to even do that. If I parented QMenu's to something that gets _eliminated (popup closes, the var "popup" is gone, but not for saved QMenu's)_ I'd get an access violation crash.

The point is that you're currently not checking for _any_ order, but from a UX and UI point of view there might be requirements to that end (e.g. "Sub-menus need to follow alphabetical order") so your implementation has to fulfil that requirement. Whether you rely on the underlying storage mechanism to ensure that or your code ensures that while adding the sub-menus is up to the implementation, but it needs to be implemented _somewhere_ (and be explicit about it).


> The "deprecated" thing, is nothing special. All I did in the code was, if `subcategory` is `Deprecated` then addSource inside deprecated.
> 
> If a Source is inside a custom category and they want to deprecate it, they can't create a subcategory inside of a subcategory, because I think this would require some more OOP stuff and system. They can just move it inside Deprecated. Or make a new plugin and make everyone download the new one. At the end, the user installs the plugin.
> 
> But if someone ever wants to add subcategories within subcategories... Honestly, at that point `obs_source_info.subcategory` can just support `/` and then you would do `Submenu1/Submenu2` and then it's just like Windows Explorer folders. Anything that is _nested_ looks very complicated to add.

The "Deprecated" sub-menu is not a category, it is a capability flag. As such your code has to handle that implicitly and cannot rely on the category string. The scenario you need to cover for (in code) is a source that has a category set but _also_ the deprecation capability enabled.

> Regarding, whether it's good and nice that developers can create subcategories. You also asked if it doesn't sort of "clog" up too much. Here's the thing. What's better, clogging subcategories with 500 sources. Or all the 500 sources mixed with all the other sources, and eventually other plugins. Any sort of modification, like Plugin or Steam Workshop item in some game, eventually adds things and modifies stuff, so it's always expected that there's _stuff added_.

We can entertain that question when OBS has 500 distinct source types. And when it comes to Lua/Python sources, they are supposed to behave just like any other kind of source types. We make no distinction between sources created by scripts or by plugins/first party code and if you have more than - say - 5 distinct Lua sources you should look into writing a proper plugin and encapsulating functionality in that instead (i.e. like the mac-capture plugin encapsulates Desktop-, Window-, and Application-Capture.

Scripts are really just intended for simple automation but not as a drop-in replacement to proper plugins.

> The subcategories are simply named by string and will _merge_ into eachother, simply because it's just a string not an object.

And my point is that it should not be based on strings, but on pre-determined values instead - see also: ["Stringly-typed](https://wiki.c2.com/?StringlyTyped).

The bottom line is that if we land such a feature, it would need to work with _all_ source types and not just script sources so it needs to become a core functionality of the source menu implementation and also the core plugin API.

karl-police:
For the Deprecated thing, wouldn't it be required to add a sort of ``obs_source_info.deprecated`` boolean?


> And my point is that it should not be based on strings, but on pre-determined values instead - see also: ["Stringly-typed](https://wiki.c2.com/?StringlyTyped).

This wiki website looks interesting, it's the sort of information things that I've been looking for in general regarding coding. But I am not sure if I understood it right.

I had the idea to create subcategories the same way sources get created, like unique string IDs and etc. though I wasn't sure if it would fit in. However, I can see it allowing subcategories within subcategories, if implemented right.

Warchamp7:

PR #9745, opened 2023-10-21
linux-pipewire: Load glad symbols on start

### LINKED ISSUE #9733: OBS Studio Crashes on Arch Linux after selecting screen (pipewire)
### Operating System Info

Other

### Other OS

Arch Linux

### OBS Studio Version

Git

b29a44a67db76c6f665bcc0887730d689e6b908d

### OBS Studio Log URL

OBS won't let me upload an old-enough log file. Will attach instead.

### OBS Studio Crash Log URL

_No response_

### Expected Behavior

Full screen recording on selected monitor.

### Current Behavior

OBS crashes after the second time I select the screen (I'm not sure why, but OBS always asks me twice for which screen to use on start-up), with:
```
(lldb) bt
* thread #28, name = 'PipeWire thread', stop reason = signal SIGSEGV: invalid address (fault address: 0x30)
  * frame #0: 0x00007fffc7e3e8ec linux-pipewire.so`renegotiate_format(data=0x0000555557a102e0, expirations=<unavailable>) at pipewire.c:631:2
    frame #1: 0x00007fffbffe4cc9 libspa-support.so`source_event_func(source=0x00005555574dcc60) at loop.c:663:2
    frame #2: 0x00007fffbffe6666 libspa-support.so`loop_iterate(object=0x00005555574af308, timeout=<unavailable>) at loop.c:496:4
    frame #3: 0x00007fff88cda14d libpipewire-0.3.so.0`do_loop(user_data=0x0000555557315260) at thread-loop.c:295:14
    frame #4: 0x00007ffff3eaa9eb libc.so.6`start_thread(arg=<unavailable>) at pthread_create.c:444:8
    frame #5: 0x00007ffff3f2e7cc libc.so.6`__clone3 at clone3.S:78
```

I'm running:
- obs-studio @ b29a44a67db76c6f665bcc0887730d689e6b908d
- pipewire: 0.3.82
- hyprland: 0.30.0
- xdg-desktop-portal-hyprland: 1.2.2

on Arch Linux. The logs just prior to the crash are:

```
09:01:45.260: [pipewire] Asking for desktop
09:01:46.292: [pipewire] Asking for desktop
09:01:46.293: [pipewire] desktop selected, setting up screencast
09:01:46.294: [pipewire] desktop selected, setting up screencast
09:01:46.565: [pipewire] Server version: 0.3.82
09:01:46.565: [pipewire] Library version: 0.3.82
09:01:46.565: [pipewire] Header version: 0.3.82
09:01:46.565: [pipewire] Created stream 0x5555574dccd0
09:01:46.565: [pipewire] Stream 0x5555574dccd0 state: "connecting" (error: none)
09:01:46.566: [pipewire] Playing stream 0x5555574dccd0
09:01:46.602: [pipewire] Stream 0x5555574dccd0 state: "paused" (error: none)
09:01:46.603: [pipewire] Server version: 0.3.82
09:01:46.603: [pipewire] Library version: 0.3.82
09:01:46.603: [pipewire] Header version: 0.3.82
09:01:46.603: [pipewire] Negotiated format:
09:01:46.603: [pipewire]     Format: 8 (Spa:Enum:VideoFormat:BGRx)
09:01:46.603: [pipewire]     Modifier: 0x0
09:01:46.603: [pipewire]     Size: 3840x2160
09:01:46.603: [pipewire]     Framerate: 0/1
09:01:46.603: [pipewire] Created stream 0x555557511cf0
09:01:46.603: [pipewire] Stream 0x555557511cf0 state: "connecting" (error: none)
09:01:46.603: [pipewire] Playing stream 0x555557511cf0
09:01:46.603: [pipewire] Stream 0x555557511cf0 state: "paused" (error: none)
09:01:46.603: [pipewire] Negotiated format:
09:01:46.603: [pipewire]     Format: 8 (Spa:Enum:VideoFormat:BGRx)
09:01:46.603: [pipewire]     Modifier: 0x200000028a37f04
09:01:46.603: [pipewire]     Size: 3840x2160
09:01:46.603: [pipewire]     Framerate: 0/1
09:01:46.604: [pipewire] Stream 0x5555574dccd0 state: "streaming" (error: none)
09:01:46.605: [pipewire] Negotiated format:
09:01:46.605: [pipewire]     Format: 8 (Spa:Enum:VideoFormat:BGRx)
09:01:46.605: [pipewire]     Modifier: 0x0
09:01:46.605: [pipewire]     Size: 3840x2160
09:01:46.605: [pipewire]     Framerate: 0/1
09:01:46.606: [pipewire] Negotiated format:
09:01:46.606: [pipewire]     Format: 8 (Spa:Enum:VideoFormat:BGRx)
09:01:46.606: [pipewire]     Modifier: 0x200000028a37f04
09:01:46.606: [pipewire]     Size: 3840x2160
09:01:46.606: [pipewire]     Framerate: 0/1
09:01:46.623: Cannot create EGLImage: Arguments are inconsistent (for example, a valid context requires buffers not supplied by a valid surface).
09:01:46.623: [pipewire] Renegotiating stream
```

If I remove all the pipewire inputs from my scene JSON config, then OBS starts just fine (and doesn't prompt me to select a screen). If I then try to _add_ a window capture via Pipewire, it crashes again (after I select the screen).

Log file from crashing run: 
[2023-10-19 09-17-06.txt](https://github.com/obsproject/obs-studio/files/13045495/2023-10-19.09-17-06.txt)

### Steps to Reproduce

1. Open OBS Studio
1. Add pipewire as a source
1. Select a screen
1. Crash

Not sure if this is related to https://github.com/obsproject/obs-studio/issues/9617.

I'm using a build of Git HEAD.

Downgrading pipewire and xdg-desktop-portal-hyprland, restarting, and then starting OBS fixes the problem.

kkartaltepe:
### Description

After the cmake 3.0 rebuild glad was transitioned to a static library. This lead to the glad symbols being uninitialized and crashes when they are used in some cases.

### Motivation and Context

fixes shm capture for compositors that dont support dmabuf.

### How Has This Been Tested?

Loading with software gl no longer crashes when calling glBindTexture.

### Types of changes

### Checklist:

### DIFF
--- a/plugins/linux-pipewire/linux-pipewire.c
+++ b/plugins/linux-pipewire/linux-pipewire.c
@@ -21,6 +21,7 @@
 
 #include <obs-module.h>
 #include <obs-nix-platform.h>
+#include <glad/glad.h>
 
 #include <pipewire/pipewire.h>
 #include "screencast-portal.h"
@@ -34,6 +35,10 @@ MODULE_EXPORT const char *obs_module_description(void)
 
 bool obs_module_load(void)
 {
+	obs_enter_graphics();
+	gladLoadGL();
+	obs_leave_graphics();
+
 	pw_init(NULL, NULL);
 
 	screencast_portal_load();


tytan652:
> After the cmake 3.0 rebuild glad was transitioned to a shared library.

It's the opposite, Glad was turned into a static library so the symbols needs to be load per instance of the static library.

kkartaltepe:
> > After the cmake 3.0 rebuild glad was transitioned to a shared library.
> 
> It's the opposite, Glad was turned into a static library so the symbols needs to be load per instance of the static library.

Right, thanks for the catch fixed up the message.

plugins/linux-pipewire/linux-pipewire.c
```diff
@@ -34,6 +35,10 @@ MODULE_EXPORT const char *obs_module_description(void)
 
 bool obs_module_load(void)
 {
+	obs_enter_graphics();
+	gladLoadGL();
```
PatTheMav, line 39: I take that it was easier to just load all symbols vs loading specific ones?

PatTheMav (commented):
This _should_ be fine even though we only use `RTLD_LAZY` on Linux (and not `RTLD_LOCAL`) - `glad` is only provided once in the build tree, so all symbols should be identical.

But it would obviously be cleaner if we'd use `RTLD_LOCAL` on Linux as well, might also force some other architectural issues to appear but we can leave that for another PR.

One last note: _If_ we ever come around to implementing Vulkan, would this still be fine (as in: We use OpenGL here, because the compositor uses OpenGL)?

tytan652:
> One last note: _If_ we ever come around to implementing Vulkan, would this still be fine (as in: We use OpenGL here, because the compositor uses OpenGL)?

No, since it use OBS graphic context. `swap_texture_red_blue()` will need to be modified to rely only on libobs graphic API which I guess require to add new functions to it.

PatTheMav:
> > One last note: _If_ we ever come around to implementing Vulkan, would this still be fine (as in: We use OpenGL here, because the compositor uses OpenGL)?
> 
> No, since it uses OBS graphic context. `swap_texture_red_blue()` will need to be modified to rely only on libobs graphic API which I guess require to add new functions to it.

Got it - it's fine for now and will need to be revisited once we have alternative renderers on Linux.

PatTheMav (approved):

kkartaltepe:
> Got it - it's fine for now and will need to be revisited once we have alternative renderers on Linux.

Yea this is a hack, once we have vulkan we should have an RGBX/RGBA format, its only because opengl doesnt have this format in the first place that we need this hack.

plugins/linux-pipewire/linux-pipewire.c
kkartaltepe: Correct its easier and less error prone to keep glad actually initialized.

GeorgesStavracas (approved):
Seems fine, but I wished these symbols could be loaded automagically.

Anyway, @tytan652 feel free to pick this one to your cherry-pick branch

PR #9746, opened 2023-10-22
linux-pipewire: Add locks to unsafe pipewire functions

kkartaltepe:
### Description

These functions log that they should be used with the thread lock held, so lets hold the lock before calling them.

### Motivation and Context

Maybe it fixes the errors we see related to obs_pw being invalid.

### How Has This Been Tested?

adding a bunch of pipewire sources and removing a bunch and restarting obs and nothing seems impacted.

### Types of changes

 - Bug fix (non-breaking change which fixes an issue) (maybe)

### Checklist:


GeorgesStavracas (approved):
Looks correct to me

PR #9758, opened 2023-10-25
obs-webrtc: Allow non-CBR rate control with WHIP

### LINKED ISSUE #9753: WHIP usage forces CBR regardless of rate control settings or "Ignore recommendations"
### Operating System Info

Windows 10

### Other OS

_No response_

### OBS Studio Version

30.0.0-rc1

_No response_

### OBS Studio Log URL

https://obsproject.com/logs/tvOBVacXJqBDBtuz

### OBS Studio Crash Log URL

_No response_

### Expected Behavior

Configuring VBR or CQP along side WHIP should either actually utilize VBR or CQP, prevent these from being set to begin with, or throw a notification explaining that they cannot work.

### Current Behavior

Attempting to stream via WHIP using VBR or CQP will sliently revert to CBR at encoder start-up time. The only indication this has happened is in the logs.

### Steps to Reproduce

These are the steps I follow verbatim in the session reflected in the uploaded log

1. Create a new profile (mine is named TEST)
2. File->Settings->Stream, change Service to WHIP, no need to configure a server
3. Check "Ignore streaming service setting recommendations" just in case
4. Move to Output, change Output Mode to Advanced
5. Change Rate Control to VBR
6. Okay these changes
7. Record a blank scene for 30 seconds or so
8. Witness in the logs the encoder settings block reports `rate_control: CBR`
9. Inspect the recorded video and witness the constant 2500kbps bitrate; a blank scene should be <100kbps

Now, to pin down WHIP configuration is causing this...

10. Back to Stream, change Service to Custom
11. Begin a second recording
12. Witness in logs the encoder settings block reports `rate_control: VBR`
13. Inspect the recorded video and witness varying bitrate

As noted: this strikes if you set VBR or CQP. x264 or HVENC h264 are both susceptible.

I've replicated this on a friend's system as part of setting up and testing a private WebRTC backend. Same Windows version, same RC, same replication steps.

If the forced usage of CBR is not a bug, I'd propose a warning when starting stream or saving config that WHIP and VBR don't go together.

For my own use-cases though I'm hoping this is a config-handling bug :)

alinsavix:
### Description
Currently, the WHIP implementation in OBS forces the rate control mode to CBR. However, nothing about WHIP requires CBR (and many things using WHIP use VBR), and there's no specific upstream service to care about it (WHIP is a protocol, not a service, despite being exposed as a "service" in OBS), so there's no obvious reason to force this. After discussion in #9753 and in Discord, the consensus is that we should stop forcing the rate control method for WHIP and honor the user's requested rate control method.

### Motivation and Context
Allow WHIP functionality to work more like much of the world is using it by not arbitrarily limiting the user's choices when there doesn't seem to be a benefit to having that limitation.

Fixes #9753 

### How Has This Been Tested?
Using a test WHIP server, configured OBS for all four combinations of [CBR,VBR] and [x264, NVENC H.264] and verified that the output bitrate from OBS matched the requested rate control mode (steady at 5000kbit for CBR modes; varying from under 100kbit to 5000kbit  for VBR modes, depending on whether the test video stream had active movement or not).

Only tested on Windows, but since the only change is not forcing a rate control mode before initializing the encoder, should work fine on Mac & Linux as long as rate control modes worked on those platforms to start with.

### Types of changes

### Checklist:
- [x] I have included updates to all appropriate documentation. (no changes)

### DIFF
--- a/plugins/obs-webrtc/whip-service.cpp
+++ b/plugins/obs-webrtc/whip-service.cpp
@@ -33,7 +33,6 @@ void WHIPService::ApplyEncoderSettings(obs_data_t *video_settings, obs_data_t *)
 	// For now, ensure maximum compatibility with webrtc peers
 	if (video_settings) {
 		obs_data_set_int(video_settings, "bf", 0);
-		obs_data_set_string(video_settings, "rate_control", "CBR");
 		obs_data_set_bool(video_settings, "repeat_headers", true);
 	}
 }


alinsavix:
Probably worth noting that @Sean-Der, who has been the one working on the WHIP integration, agrees that (quote) "we should absolutely allow VBR" and this PR is submitted in line with that concurrence.

Sean-Der:
Thank you for submitting @alinsavix I am in support of this change!

alinsavix:
FWIW, I think it's probably worth it to merge this before v30 rc2 (and *definitely* before v30 final), since it would be a shame to ship this long-awaited feature in a basically crippled form, especially when there doesn't seam to be anyone that actually *knows* why it was limiting WHIP to CBR-only (just supposition), that it may well just have been a cut & paste error, and the person who has been shepherding WHIP through the system considers forcing CBR to be a mistake.

Fenrirthviti:
This likely won't make RC2 (we already pulled the cherry-pick PR in for that), but it should be a candidate as it's a pretty straightforward/simple fix for release.

Issue #7758, opened 2022-11-11
Dialog box text overwrites the Info badge


Xylopyrographer:
### Operating System Info

macOS 12

### OBS Studio Version

28.1.2

### Expected Behavior

Dialog box text should not overwrite the info badge

### Current Behavior

Dialog box text item  "Close file when inactive" overwrites the info badge
![SourcePropsGUI](https://user-images.githubusercontent.com/81269441/201293861-4cacc027-75bd-43da-8ab3-55c5d9ff918d.jpg)

### Steps to Reproduce

1. Create a scene
2. IN that scene, add a local media file as a source
3. Right click on that source
4. Select "Properties"
5. Observe the "Close file when inactive" item
...
You'll note the "circle with a question mark" badge is partially overwritten by the text.

nikola3211:
Hi, I think I can solve this. Should I make a PR?

Fenrirthviti:
We always welcome PRs to fix open issues.

PR #9759, opened 2023-10-25
UI:Properties view

nikola3211:
### Description

Added spacing to QHBoxLayout

### Motivation and Context

Some elements are overlapping on MAC due to 
no spacing in QHBoxLayout.

https://github.com/obsproject/obs-studio/issues/7758

### How Has This Been Tested?

Tested on MAC only.

### Types of changes

### Checklist:

- [ x] My code is not on the master branch.
- [ x] The code has been tested.
- [ x] All commit messages are properly formatted and commits squashed where appropriate.
- [ x] I have included updates to all appropriate documentation.

### DIFF
--- a/UI/properties-view.cpp
+++ b/UI/properties-view.cpp
@@ -1626,8 +1626,14 @@ void OBSPropertiesView::AddProperty(obs_property_t *property,
 			QHBoxLayout *boxLayout = new QHBoxLayout(newWidget);
 			boxLayout->setContentsMargins(0, 0, 0, 0);
 			boxLayout->setAlignment(Qt::AlignLeft);
+#ifdef __APPLE__
+			/* TODO: This fixes the issue of tooltip not aligning
+			 * correcty on macOS, the root cause needs further
+			 * investigation. */
+			boxLayout->setSpacing(10);
+#else
 			boxLayout->setSpacing(0);
-
+#endif
 			QCheckBox *check = qobject_cast<QCheckBox *>(widget);
 			check->setText(desc);
 			check->setToolTip(


Warchamp7:
Please post before and after screenshots if you're able

nikola3211:
Hi, 
Here are before and after screenshots:
BBD53B79-1B81-4571-8092-CC03CA4D82C7
BD8CC017-09C9-4704-9E6B-0BD3F21DD692

gxalpha (commented):
I arrived at a similar fix when looking into this a while back, the problem is that the problem appears to be exclusive to macOS. This leads me to believe that there's some kind of Qt issue which imo we should investigate first.
Also, changing this for every OS will probably add unnecessary spacing on other platforms - if it does (some screenshot of Windows or Linux would be nice), this should be ifdef-guarded to macOS only (but again, I'd be more happy about an investigation into what *actually* happens here on Mac).

nikola3211:
Example 2 on this link:
https://cpp.hotexamples.com/examples/-/QHBoxLayout/setContentsMargins/cpp-qhboxlayout-setcontentsmargins-method-examples.html
and quick google search showed me that many people have issues with QHBoxLayout  on Mac vs Win or Linux.
Seems to me that this is OS related. I can ifdef-guard it if you want?

WizardCM:
Yes, please guard this change to limit it to macOS. Otherwise, looks like a good fix.

RytoEX:
For reference, on Windows 11 right now, it looks like this:
image

With this PR, without guarding the change to affect macOS-only, it looks like this:
image

nikola3211:
On windows the box is also aligned properly. It seems that the box layout adds that tilt. When I tested this by adding help on another bool, it also shifted to the right a bit. I should look into that too.

nikola3211:
I've spent some time on this. Only solution I've found is to add box containers on all the checkboxes, so that they all shift a bit to the right, everything else I tried didn't work. If you want I can push those changes or we can leave it as it is.

gxalpha (commented):
I'm fine with this but would like a TODO comment that the root cause needs further investigation.

norihiro:
The first word after the prefix in the commit title should be a present-tense verb.
I'd like to suggest something like `UI: Fix ... on properties view`. Please replace `...` with something suitable word.

nikola3211:
@gxalpha  Shall we close this?

gxalpha:
Please see the comment above: https://github.com/obsproject/obs-studio/pull/9759#issuecomment-1806839734
Also please squash your commits together into one and force-push to this PR's branch.

nikola3211:
Done.

norihiro:
I'm sorry that we need a space between `:` and `Fix`.

nikola3211:
No problem, I didn't know.

nikola3211:
@gxalpha Is this ok to merge now?

gxalpha (approved):
seems fine to me

PR #9762, opened 2023-10-26
libobs: Export SIMDe headers unconditionally for plugin development

PatTheMav:
### Description
Changes `libobs` export to include SIMDe headers by default.

### Motivation and Context
Fixes https://github.com/obsproject/obs-plugintemplate/issues/95.

### How Has This Been Tested?
Checked import of `libobs` Framework with new settings and confirmed SIMDe headers were correctly copied over.

### Types of changes

### Checklist:

### DIFF
--- a/libobs/CMakeLists.txt
+++ b/libobs/CMakeLists.txt
@@ -303,6 +303,20 @@ set(public_headers
     util/platform.h
     util/profiler.h
     util/profiler.hpp
+    util/simde/check.h
+    util/simde/debug-trap.h
+    util/simde/hedley.h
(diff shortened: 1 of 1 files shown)


RytoEX (approved):
Seems fine at a glance if we want to always export these.

PatTheMav:
> Seems fine at a glance if we want to always export these.

Shouldn't hurt because headers are only actually pulled in by preprocessors if required, but we enable SIMD for `clang` and `gcc` by default because it will just fall back on native vector operations on x86_64 anyway (so there's no harm in enabling it by default and also using SIMDe by default).

PR #9765, opened 2023-10-26
CI: Add github-actions renderer option to xcbeautify

PatTheMav:
### Description
Adds the `github-actions` renderer options to `xcbeautify` which supports workflow commands for its output.

### Motivation and Context
Improve legibility and integration of CI builds with GitHub Actions.

### How Has This Been Tested?
Needs to be tested on CI.

### Types of changes

### Checklist:


PatTheMav:
As an unintended side-effect, this seems to achieve the kind of warning-annotation that we recently removed a manual implementation for.

PatTheMav:
@RytoEX looks like `xcbeautify` does this by default now, so there is no need to enable it explicitly.

We can merge this PR if we want to make it explicit in our codebase though.

RytoEX (approved):
Seems fine.

Issue #9760, opened 2023-10-25
Virtual Camera on OBS 30 is consuming more CPU than OBS itself


felippenardi:
### Operating System Info

macOS 14

### Other OS

Sonoma

### OBS Studio Version

30.0.0-rc1

### OBS Studio Log URL

couldn't get it

### Expected Behavior

The new camera process shouldn't consume more CPU than OBS

### Current Behavior

5 minutes after turning virtual camera on my Macbook M1 Max goes up to 99% cpu usage. In the task manager, the new camera process is taking more CPU than the OBS process

### Steps to Reproduce

1. Install OBS 30 on a Mac
2. Turn on virtual camera
3. Watch the CPU usage of the camera process

### Anything else we should know?

I had to format my mac in order to install OBS 29 again and be able to use the OBS 29 Virtual Camera. So I lost all logs and informations that could make this report more helpful.

Fenrirthviti:
A current log file that shows the setup and configuration is still useful, please provide one.

felippenardi:
As I've added to the notes above, I had to format my mac in order to install OBS 29 again and be able to use the OBS 29 Virtual Camera. So I lost all logs and informations that could make this report more helpful.

But this issue is replicable. I just can't afford to have to format my computer again just to test it.

Anyone with Mac OBS 30 can see this camera process taking more CPU than OBS, is very easily replicable.

gxalpha:
A log file from the current OBS would still be useful. Fyi, you can also use the new virtualcam from OBS 30 on macOS 13, so if that's what you went back to, please retry.

alinsavix:
I can reproduce this. On my laptop, OBS is using ~50% CPU, while the camera process is using ~175% CPU. I don't see the CPU stepping down below 2.9GHz, so this probably isn't a case of "percentage bigger because clock speed smaller". This is on an intel mac (OP was on a M2 mac), so not just ARM weirdness.

Interestingly, the camera process continues to use a lot of CPU (though slightly less, at 100%) even when OBS is closed; after closing the chrome tab that was reading the other end of the webcam (after OBS was already closed) the camera process did eventually exit, though took probably on the order of a minute to do so. 

OBS log and spindump attached, let me know if any other information or debugging and I'd be happy to oblige.

[2023-10-26 15-26-18.txt](https://github.com/obsproject/obs-studio/files/13183495/2023-10-26.15-26-18.txt)
[...]

PR #9768, opened 2023-10-27
mac-virtualcam: Correct device timer firing rate

jcm93:
### Description
Resolves the conversion to a Double in `OBSCameraDeviceSource`, such that the virtual camera device timer fires correctly every `1/60` seconds, instead of trying to fire every `0` seconds.

### Motivation and Context
Resolves https://github.com/obsproject/obs-studio/issues/9760. CPU use was very high with the new virtual camera extension at idle and while active because of the amount of polling the device timer was doing.

### How Has This Been Tested?

Tested locally on my machine. Reduced camera extension process CPU use by roughly 10x.

### Types of changes

### Checklist:

### DIFF
--- a/plugins/mac-virtualcam/src/camera-extension/OBSCameraDeviceSource.swift
+++ b/plugins/mac-virtualcam/src/camera-extension/OBSCameraDeviceSource.swift
@@ -141,7 +141,7 @@ class OBSCameraDeviceSource: NSObject, CMIOExtensionDeviceSource {
 
         _timer!.schedule(
             deadline: .now(),
-            repeating: Double(1 / OBSCameraFrameRate),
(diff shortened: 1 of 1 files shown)


PatTheMav:
That was indeed my mistake, the original example code correctly uses `1.0 / Double(kFrameRate)` as the loop interval.

PatTheMav (approved):

felippenardi:
@PatTheMav Do you happen to know when a new release candidate with this fix is going to be published? I'd like to try it

PatTheMav:
It's a critical fix, so we'll probably try to get it in as soon as possible.

RytoEX (approved):

PR #8832, opened 2023-05-02
obs-ffmpeg: Add AV1 support for VA-API

nowrep:
AV1 support for VA-API encoder.

### Description

Added FFmpeg VAAPI AV1 encoder.

### Motivation and Context

~~Draft for now as av1_vaapi FFmpeg encoder is not yet upstreamed.~~

### How Has This Been Tested?

Tested with Mesa driver.

### Types of changes

- New feature (non-breaking change which adds functionality)

### Checklist:

- [x] My code has been run through [clang-format](https://github.com/obsproject/obs-studio/blob/master/.clang-format).
- [x] I have read the [**contributing** document](https://github.com/obsproject/obs-studio/blob/master/CONTRIBUTING.rst).
- [x] My code is not on the master branch.
- [x] The code has been tested.
- [x] All commit messages are properly formatted and commits squashed where appropriate.
- [x] I have included updates to all appropriate documentation.

PR #9770, opened 2023-10-28
obs-ffmpeg: Add FFmpeg VAAPI AV1 encoder 

Jorropo:
### Description
This include refactors, update to current FFmpeg master (https://github.com/FFmpeg/FFmpeg/commit/e5f774268a06c637df5edb3dffa237caf6e678ea) and the new FFmpeg vaapi av1 decoder. (see commits and their descriptions)

### Motivation and Context
This allows to use the upcoming ffmpeg 6.1 `av1_vaapi` encoder (https://github.com/FFmpeg/FFmpeg/commit/3be81e3b449febfb04596f3f187aae27f18d02f7) for GPU accelerated av1 encoding using vaapi.

### How Has This Been Tested?
**This does not work** yet, when trying to be used `ffmpeg_mux_get_extra_data` fails crashing obs.
I think the issue is that I do not parse the headers from the av1 bitstream, I've tried copying what `obs-ffmpeg-av1.c` does and forward the `extradata` field or `AV_PKT_DATA_NEW_EXTRADATA` `sidedata` however theses are not present and by looking at ffmpeg's source I don't think they are touched at all by `av1_vaapi`, so I think parsing them from the bitstream is the only option, it is likely the keyframe flag also need to be fixed.
Help happily wanted, I have not touched av1 bitstream before.

### Types of changes

### Checklist:

### DIFF
--- a/deps/media-playback/media-playback/decode.c
+++ b/deps/media-playback/media-playback/decode.c
@@ -114,8 +114,11 @@ static uint16_t get_max_luminance(const AVStream *stream)
 {
 	uint32_t max_luminance = 0;
 
-	for (int i = 0; i < stream->nb_side_data; i++) {
-		const AVPacketSideData *const sd = &stream->side_data[i];
+
+	AVCodecParameters *cpar = stream->codecpar;
+
+	for (int i = 0; i < cpar->nb_coded_side_data; i++) {
+		const AVPacketSideData *const sd = &cpar->coded_side_data[i];
 		switch (sd->type) {
 		case AV_PKT_DATA_MASTERING_DISPLAY_METADATA: {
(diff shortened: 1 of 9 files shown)


Fenrirthviti:
Does this work differ from the work being done in https://github.com/obsproject/obs-studio/pull/8832 and https://github.com/obsproject/obs-studio/pull/9475?

Jorropo:
@Fenrirthviti good question, I had fun while doing it but I don't think this counts.
IMO there are small code organization issues with #8832 (`else if` chains should be `switch` so the compiler is able to warn if a case is not implemented) but it works, so ...

I'll comment or open a PR on #8832 later thx.

Issue #9754, opened 2023-10-24
Transformation of sources within a group results in unexpected movement (e.g. flying of the screen)


Zulleyy3:
### Operating System Info

Other

### Other OS

Arch Linux

### OBS Studio Version

Other

### OBS Studio Version (Other)

29.1.3-tytan652-6

### OBS Studio Log URL

https://obsproject.com/logs/2l2TAu4a1RkZcxzi

### Expected Behavior

Rotation of source (e.g. image) within group results simply results in some sort of rotation

### Current Behavior

The sources to be rotated move across or even out of the canvas while rotating.

### Steps to Reproduce

1. Add Group
2. Add Source (like an Image)
3. Move Source into Group
4. Rotate/Scale the Source

https://github.com/obsproject/obs-studio/assets/20947639/a1c0d24c-5407-488f-a1f3-cca221b39283

### Anything else we should know?

There is a chance that being close to the canvas (or group?) border matters for this. Here is a video of me experimenting with 2 similar scenes I have where I ran into the issue first.

https://github.com/obsproject/obs-studio/assets/20947639/ef678d31-aeeb-47c3-a1d8-1c968fec6059

(higher quality video on youtube: https://www.youtube.com/watch?v=MC-Z7Ac5hrg)

Issue #8768, opened 2023-04-18
UI: ALT cropping of the Grouped items moves sources to the wrong positions


SuslikV:
### Operating System Info

Other

### Other OS

Any

### OBS Studio Version

29.1.0-beta4

### OBS Studio Log URL

about:blank

### Expected Behavior

Mouse pointer for ALT-cropped item inside the group should follow the object when user interacts with the item and with normal speed/scale (1 px mouse move = 1px item position change).

### Current Behavior

When user changes ALT-crop of the item inside the group - the item may move(!) and at accelerated speed (1 px mouse move = ~50px item position change).

### Steps to Reproduce

1. Make two items (text/image sources)
2. Alt-crop Left and Top of the most left item(source) with the mouse
3. Group items from p1
4. Select whole group from p3 and move it up in the preview with the mouse
5. Select item(source) from p2 and move it up too with the mouse
6. Try to change Alt-crop of the item from p2 with the mouse
7. Notice that the item not only changes its own crop but also changes its own position while mouse pointer cannot follow these changes.

### Anything else we should know?

Bug was known far before the v28.

There is other bug in the video below - the "Transform" window is not always shows properties of the currently selected item, so the window was reopened by me intentionally (this UI bug also known before the v28 and not related to the original issue).

https://user-images.githubusercontent.com/19683044/232857052-df69aa02-1745-4e5f-b516-ffadc69e9371.mp4

(video was recorded in virtualbox environment, so I forgot about the log-file)

**Workaround:** Ungroup items before changing theirs cropping with the mouse

Forum's origin and possible use case:
https://obsproject.com/forum/threads/selection-of-sources-after-grouping-help-wanted.165785/

SuslikV:
Another steps to reproduce the issue are:
1) Add text source (GDI+) with "123" text.
2) Add text source (GDI+) with "456" text (now both texts are coexist at same position).
3) Over the "123" source hit Ctrl + E, and set Crop: Left: 100 px, Top: 100 px.
4) Group items "123" and "456".
5) Select both "123" and "456" sources and Group them.
6) Over the grouped source (from p4) hit Ctrl + E, and set Y position 100 px.
7) Select the source "123" in Sources list and move it **by mouse** at top left corner of the canvas in Preview.
8) Over the grouped source (from p4) hit Ctrl + E, and set Y position 100 px again.
[...]

PR #9775, opened 2023-10-29
UI: Fix transform of sources in groups

Zulleyy3:
### Description
Clear stretchGroup if the group is meant to be ignored.

### Motivation and Context
Group resize is not deferred until adjustments are completed by a mouseReleaseEvent in certain cases, resulting in unexpected movement of sources that are part of a group as they are resized or stretched beyond group bounds.

Related/Resolved Issue: https://github.com/obsproject/obs-studio/issues/9754

The bug can be described/reproduced as follow:

1. The user selects within a group in the sources dock
2. The user moves their mouse over the selected source such that the cursor would change.
   - This results in a cursor update and a call to `GetStrechHandle()`
   - as the source is part of a group, `stretchGroup` will be set to the group of the source

3. The user clicks on the canvas without touching a stretch/rotation handle
   - A `mouseReleaseEvent` is fired
   - `obs_sceneitem_defer_group_resize_end()` is called on the group as `stretchGroup` is still defined from earlier
   - the `defer_group_resize` member of the group will become negative
   - The deferal check in `resize_group()` (obs-scene.c) will always pass now
4. When scaling or rotating the source close to the group bounds, the group bounds will now dynamically update causing the source in to fly off the canvas.

### How Has This Been Tested?
I tested this by reproducing the steps outlined above and experiencing the expected behavior
This is unlikely to affect any other areas of the code.

OS: Arch Linux
Kernel: x86_64 Linux 6.5.9-arch2-1
CPU: AMD Ryzen 7 2700X Eight-Core @ 16x 3.7GHz
GPU: NVIDIA GeForce RTX 3070

### Types of changes

### Checklist:
- [] My code is not on the master branch.

### DIFF
--- a/UI/window-basic-preview.cpp
+++ b/UI/window-basic-preview.cpp
@@ -520,6 +520,8 @@ void OBSBasicPreview::GetStretchHandleData(const vec2 &pos, bool ignoreGroup)
 							 &invGroupTransform);
 			matrix4_inv(&invGroupTransform, &invGroupTransform);
 			obs_sceneitem_defer_group_resize_begin(stretchGroup);
+		} else {
+			stretchGroup = nullptr;
 		}
 	}
 }


SuslikV:
I will write soon if it related to: https://github.com/obsproject/obs-studio/issues/8768 too (or changes its behavior)

PatTheMav:
I might be doing something wrong because I cannot replicate the bug described here _or_ the bug described in https://github.com/obsproject/obs-studio/issues/8768 even following the reproduction steps.

Could this be actually related to the scaling factor? Because Retina displays are the default on macOS, so we always got a 2x scale factor on macOS (which is where the bug doesn't seem to occur).

Zulleyy3:
Preview-scale plays a role in the code that checks if a handle is present under the cursor and whether there is a need to update it's visuals. Could the difficulty in reproduction on MacOS be related to that?

I am consistently reproducing with the steps outlined above this with obs (29.1.3-tytan652-6), but i think i could have been a bit clearer on step 2. Those 3 points would change the cursor and I was able to make it work with all 3: 
image

PatTheMav:
I have been able to reproduce it _sometimes_ but not consistently, it's weird really.

But overall it makes sense that the code is problematic:

It changes state for a "mouse pressed" event (namely making `stretchGroup` a non-`NULL` reference) but there is no opposing mouse event (e.g. "mouse released" that would reset the state of the entire class.

Zulleyy3:
From my understanding (and tracing with GDB) it was set on mouse-move (call to `GetStretchHandleData()` is in line 1701, window-basic-preview.cpp). I don't have a Mac available to check, but maybe you can set a breakpoint there and see if the cursor change actually occurs when you expect it to?

PR #9785, opened 2023-11-01
libobs: Add more files to list of public headers

PatTheMav:
### Description
Adds more files to list of public header files to fix build errors for 3rd party plugins.

### Motivation and Context
`libobs` does not provide a bespoke set of public headers for consumers of the library - instead the internal structure of the library is exposed unchanged via its internal headers to consumers.

This requires all headers that export symbols to be included as well as all headers included by those, and also all headers providing functionality via static inline functions.

As these are distributed over almost all headers of `libobs` (there are no bespoke "export headers" that only contain symbols and inline functions needed for consumers), this unfortunately also requires providing a majority of the header files.

This change probably "overcorrects" the issue, but given that 3rd party plugins currently get full access to `libobs` internal API (there is no distinction between the required runtime modules and 3rd party plugins from an architectural point of view) we have no other choice for the time being.

### How Has This Been Tested?
Checked the includes of all header files and their recursive includes as well.

### Types of changes

### Checklist:

### DIFF
--- a/libobs/CMakeLists.txt
+++ b/libobs/CMakeLists.txt
@@ -261,21 +261,40 @@ target_compile_definitions(libobs PUBLIC HAVE_OBSCONFIG_H)
 set(public_headers
     # cmake-format: sortable
     callback/calldata.h
+    callback/decl.h
     callback/proc.h
     callback/signal.h
+    graphics/axisang.h
+    graphics/bounds.h
+    graphics/effect-parser.h
+    graphics/effect.h
     graphics/graphics.h
+    graphics/image-file.h
     graphics/input.h
+    graphics/libnsgif/libnsgif.h
     graphics/math-defs.h
+    graphics/math-extra.h
+    graphics/matrix3.h
+    graphics/matrix4.h
+    graphics/plane.h
+    graphics/quat.h
+    graphics/shader-parser.h
     graphics/srgb.h
     graphics/vec2.h
     graphics/vec3.h
     graphics/vec4.h
     media-io/audio-io.h
+    media-io/audio-math.h
+    media-io/audio-resampler.h
+    media-io/format-conversion.h
     media-io/frame-rate.h
     media-io/media-io-defs.h
+    media-io/media-remux.h
     media-io/video-frame.h
     media-io/video-io.h
+    media-io/video-scaler.h
     obs-audio-controls.h
+    obs-avc.h
     obs-config.h
     obs-data.h
     obs-defs.h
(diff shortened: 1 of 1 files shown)


paulpv:
https://github.com/occ-ai/obs-polyglot/ has an interesting technique.

They effectively patch libobs with the headers their plugin needs.
* https://github.com/occ-ai/obs-polyglot/blob/master/patch_libobs.diff
* https://github.com/occ-ai/obs-polyglot/blob/a86b5778cd200b6aeb4ef65a49e621905b18e823/cmake/common/buildspec_common.cmake#L77-L82

I made a variant of this at...
* https://github.com/obs-ndi/obs-ndi/blob/develop/libobs.patch
* https://github.com/obs-ndi/obs-ndi/blob/33020b2c3c6995c4a52f3cf61958af33b0107301/cmake/common/buildspec_common.cmake#L76-L92
I may still tweak this a bit, but it works for obs-ndi.

But, it gets me to thinking that it should be possible to automate this by defining/documenting/providing a mechanism for the libobs consumer's build to define in a config file all the headers that their code needs to use in libobs.
Their build could then read that list and patch the .deps/obs-studio-.../libobs/CMakeLists.txt public headers.
This way the public headers could possibly be shrunk back down to a minimal set and not have to be nearly exhaustive.

PatTheMav:
> https://github.com/occ-ai/obs-polyglot/ has an interesting technique.
> 
> They effectively patch libobs with the headers their plugin needs.
> 
> * https://github.com/occ-ai/obs-polyglot/blob/master/patch_libobs.diff
> * https://github.com/occ-ai/obs-polyglot/blob/a86b5778cd200b6aeb4ef65a49e621905b18e823/cmake/common/buildspec_common.cmake#L77-L82
> 
> I made a variant of this at...
> 
> * https://github.com/obs-ndi/obs-ndi/blob/develop/libobs.patch
> * https://github.com/obs-ndi/obs-ndi/blob/33020b2c3c6995c4a52f3cf61958af33b0107301/cmake/common/buildspec_common.cmake#L76-L92
>   I may still tweak this a bit, but it works for obs-ndi.
> 
> But, it gets me to thinking that it should be possible to automate this by defining/documenting/providing a mechanism for the libobs consumer's build to define in a config file all the headers that their code needs to use in libobs. Their build could then read that list and patch the .deps/obs-studio-.../libobs/CMakeLists.txt public headers. This way the public headers could possibly be shrunk back down to a minimal set and not have to be nearly exhaustive.

That's of course one way to solve this issue (especially immediately until this PR is merged), but overall it's easy enough to just ensure all headers that _could_ be required by a plugin (because plugins can theoretically use the entirety of the API exposed by `libobs`) are installed into the install prefix. They don't take up much space and having them available doesn't yield any negative effects.

So for now it's good enough to just copy all those files and look at a cleanup/creation of a distinct external API as a separate code cleanup.

RytoEX (approved):
Looks fine at a glance.

PR #9795, opened 2023-11-03
win-capture: Fix Vulkan race condition

sawillms:
This race condition is caused when one thread creates a swap chain, which calls OBS_CreateSwapchainKHR, at the same time another thread calls OBS_CreateImageView.

OBS_CreateSwapchainKHR allocates swap data, publishes this into the data->swaps linked list, then initializes it. Meanwhile, OBS_CreateImageView is iterating the swaps linked list, to see if the image matches any swap chain images. Due to the order in OBS_CreateSwapchainKHR, there's no guarantee this data is initialized so it often ends up running out of bounds on the swap_images array.

The fix is simply to defer the swap data publish to after init.

### Description

The issue is that we have one thread (our window thread) creating a swap chain, which calls OBS_CreateSwapchainKHR, at the same time another thread (the main thread) is calling OBS_CreateImageView.
OBS_CreateSwapchainKHR allocates swap data, publishes this into the data->swaps linked list, then initializes it. Meanwhile, OBS_CreateImageView is iterating the swaps linked list, to see if the image matches any swap chain images. Unfortunately, due to the order in OBS_CreateSwapchainKHR, there's no guarantee this data is initialized, so often ends up running out of bounds on the swap_images array.
The fix is simply to defer the swap data publish to after initialization

### Motivation and Context

We need to be able to consistently capture an in-development Vulkan project.

### How Has This Been Tested?

Tested against our Vulkan project.

### Types of changes

### Checklist:

### DIFF
--- a/plugins/win-capture/graphics-hook/vulkan-capture.c
+++ b/plugins/win-capture/graphics-hook/vulkan-capture.c
@@ -1796,7 +1796,6 @@ OBS_CreateSwapchainKHR(VkDevice device, const VkSwapchainCreateInfoKHR *cinfo,
 	if ((res == VK_SUCCESS) && (count > 0)) {
 		struct vk_swap_data *swap_data = alloc_swap_data(ac);
 		if (swap_data) {
-			init_swap_data(swap_data, data, sc);
 			swap_data->swap_images = vk_alloc(
 				ac, count * sizeof(VkImage), _Alignof(VkImage),
 				VK_SYSTEM_ALLOCATION_SCOPE_OBJECT);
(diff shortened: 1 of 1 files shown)


Lain-B:
Out of curiosity, why is this marked as draft?

sawillms:
I've worked through the checklist. I marked it as a draft initially because I had not done so and did not expect this to become public until I manually marked it ready.

RytoEX:
The fix looks good according to off-thread review.

Lain-B (approved):

PR #9798, opened 2023-11-04
UI: Log Virtual Camera output type

Penwy:
### Description

This logs the output type and output source of the virtual camera both when starting it and when changing it while it is running.

### Motivation and Context

Issues caused by an user mistakenly selecting the wrong output type/source are rather frequent, this would greatly streamline the troubleshooting process for those.

### How Has This Been Tested?

Ubuntu 22.04.
Started and changed the virtual camera output type, verified that the logging is correct.
No crash no memory leak.

### Types of changes

### Checklist:


Lain-B (approved):
Slightly refactored code.

Issue #9796, opened 2023-11-04
Screen sharing notification remains in macOS menu bar after quitting OBS 30.0.0 RC2


Iritscen:
### Operating System Info

macOS 14

### OBS Studio Version

30.0.0-rc2

### OBS Studio Log URL

https://obsproject.com/logs/iajBVDgolAIpEOiA

### Expected Behavior

I expected that the notification in the menu bar that shows the screen is being captured would go away when streaming/capturing stops and the app is quit.

### Current Behavior

After streaming or recording with OBS, the blue notification remains lodged in the menu bar until I log out of macOS.

### Steps to Reproduce

1. Simply open OBS and begin streaming anything from any video source – a window, etc.
2. Stop streaming.
3. Quit OBS.  The menu item will remain.

### Anything else we should know?

This also happened in the version of OBS which I was previously using, v29.1.3, so it's not a brand-new bug.  Visual of menu item attached below.

![Screenshot 2023-11-03 at 8 16 08 PM](https://github.com/obsproject/obs-studio/assets/1541542/0f0d609e-402f-43cd-a8f2-33d0497984a3)

gxalpha:
This is very likely a macOS bug. Please file a feedback to Apple.

Iritscen:
To be clear, other screen-capturing apps such as ScreenFlow do not leave the screen recording annunciator (if that's the correct term) visible in the menu bar after they're done capturing the screen.  I've also just noticed that simply opening OBS triggers the screen sharing annunciator.  I can quit OBS without ever having told the app to record or stream, and it remains in the menu bar.

soheilpro:
I also have the same problem on macOS 14.2 (23C64).

remyhunt:
Also having this issue on version 28.1.2 (64 bit), updated to 29.1.3 and the issue still persisted

M1 chip, Mac OS 14.1

<img width="349" alt="Screenshot 2024-01-11 at 4 40 21 PM" src="https://github.com/obsproject/obs-studio/assets/5217127/5d9b63de-935e-4ead-8a85-97e58598e8f9">

gxalpha:
*Copy-pasting my reply from https://github.com/obsproject/obs-studio/issues/9850#issuecomment-1891098202:*

[...]

PR #9803, opened 2023-11-04
mac-capture: Add handling for nil target window

jcm93:
### Description

Adds handling for when a macOS Screen Capture Window Capture source is trying to acquire a capture for a window where there is no `SCWindow` that corresponds to that source's window ID. This most commonly happens when trying to reacquire a capture for a window that has since been closed. The `nil` case is now handled by returning from `init_screen_stream` early, in similar fashion as we do when finding a pre-Ventura system.

Additionally, the existing fallback for window ID 0 has been modified to use this same early return. While, to my knowledge, this id 0 code isn't reachable under normal conditions, we now avoid any scenario where we could acquire a capture for a window that the user has not specifically picked.

### Motivation and Context
This resolves at least one prominent case where the macOS Sonoma capture indicator would linger onscreen after OBS has been closed (possibly https://github.com/obsproject/obs-studio/issues/9796, etc.). In my testing, *any* attempt to capture a nil window would place the system in a state where the capture indicator is stuck, even if the source was properly disposed of later. This was a fairly common code path and scenario, so I suspect that this case constitutes a lot of the "my capture indicator is stuck on" scenarios that have been seen in support channels.

According to the documentation, initializing an SCContentFilter in this way with a nil SCWindow shouldn't be allowed, at least in Swift. I would have expected this to have thrown an exception. Feedback has been filed with Apple, as FB13336034.

### How Has This Been Tested?
Tested locally on my machine running macOS Sonoma 14.1. In testing, the capture indicator no longer lingers after exiting OBS. The source that fails to initialize can still be modified after startup.

### Types of changes

### Checklist:

### DIFF
--- a/plugins/mac-capture/mac-sck-video-capture.m
+++ b/plugins/mac-capture/mac-sck-video-capture.m
@@ -134,24 +134,26 @@ static bool init_screen_stream(struct screen_capture *sc)
         } break;
         case ScreenCaptureWindowStream: {
             SCWindow *target_window = nil;
-            if (sc->window != 0) {
+            if (sc->window != kCGNullWindowID) {
                 for (SCWindow *window in sc->shareable_content.windows) {
                     if (window.windowID == sc->window) {
                         target_window = window;
                         break;
                     }
                 }
-            } else {
-                target_window = [sc->shareable_content.windows objectAtIndex:0];
-                sc->window = target_window.windowID;
             }
-            content_filter = [[SCContentFilter alloc] initWithDesktopIndependentWindow:target_window];
+            if (target_window == nil) {
+                os_sem_post(sc->shareable_content_available);
+                sc->disp = NULL;
+                os_event_init(&sc->disp_finished, OS_EVENT_TYPE_MANUAL);
+                os_event_init(&sc->stream_start_completed, OS_EVENT_TYPE_MANUAL);
+                return true;
+            } else {
+                content_filter = [[SCContentFilter alloc] initWithDesktopIndependentWindow:target_window];
 
-            if (target_window) {
                 [sc->stream_properties setWidth:(size_t) target_window.frame.size.width];
                 [sc->stream_properties setHeight:(size_t) target_window.frame.size.height];
             }
-
         } break;
         case ScreenCaptureApplicationStream: {
             SCDisplay *target_display = get_target_display();


plugins/mac-capture/mac-sck-video-capture.m
```diff
@@ -142,8 +142,14 @@ static bool init_screen_stream(struct screen_capture *sc)
                     }
                 }
             } else {
-                target_window = [sc->shareable_content.windows objectAtIndex:0];
-                sc->window = target_window.windowID;
+                target_window = nil;
```
PatTheMav, line 145: `target_window` should not be anything _but_ `nil` at this point (the condition for the only code changing the pointer value has not been met), so I don't think setting it to `nil` again is necessary.

PatTheMav:
> Additionally, the existing fallback for window ID 0 has been modified to use this same early return. While, to my knowledge, this id 0 code isn't reachable under normal conditions, we now avoid any scenario where we could acquire a capture for a window that the user has not specifically picked.

The fallback is not for Window id 0, but for the first element in the array of available windows. The only case where that should yield an invalid window id is if the array is empty.

plugins/mac-capture/mac-sck-video-capture.m
PatTheMav: The observed _root cause_ of the bug is correct, but this change also doesn't fix the issue:

When an invalid non-zero window ID is retained in `sc->window`, `target_window` would still be `nil`. The same would be true if `sc->window` contains an actual `0`.

So the easiest fix would be to change the `else` branch into an explicit check `target_window == nil` and then fallback to the first window ID found in `sc->shareable_content.windows` (functionality which this PR removes).

PatTheMav, line 145:
```suggestion
            SCWindow *target_window = nil;
            if (sc->window != 0) {
                for (SCWindow *window in sc->shareable_content.windows) {
                    if (window.windowID == sc->window) {
                        target_window = window;
                        break;
                     }
                 }
             }
             
             // sc->window was actually 0 or the window ID could not be found
             if (!target_window) {
                 // This assumes that there should be more than a single window
                 // in the array
                 target_window = [sc->shareable_content.windows objectAtIndex:0];
                 sc->window = target_window.windowID;
             }             
```

jcm93:
> > Additionally, the existing fallback for window ID 0 has been modified to use this same early return. While, to my knowledge, this id 0 code isn't reachable under normal conditions, we now avoid any scenario where we could acquire a capture for a window that the user has not specifically picked.
> 
> The fallback is not for Window id 0, but for the first element in the array of available windows. The only case where that should yield an invalid window id is if the array is empty.

> > Additionally, the existing fallback for window ID 0 has been modified to use this same early return. While, to my knowledge, this id 0 code isn't reachable under normal conditions, we now avoid any scenario where we could acquire a capture for a window that the user has not specifically picked.
> 
> The fallback is not for Window id 0, but for the first element in the array of available windows. The only case where that should yield an invalid window id is if the array is empty.

Sorry, what do you mean? `sc->window` is always set to the SCWindow's windowID, and that's what's compared against when initializing a window capture. Do you mean that id 0 is sometimes a legitimate id if the system has just initialized its internal array of windows like after a reboot?

plugins/mac-capture/mac-sck-video-capture.m
jcm93: Doesn't this introduce a lot more scenarios where we are capturing windows that the user didn't pick? Isn't it preferable to return early instead of capturing the first window?

I'm unclear if id=0 is even legitimate. I have not ever seen it in `SCShareableContent`'s window list, though I haven't done super thorough testing to see if it ever appears. I know that the system starts incrementing windowIDs on boot, but I'm not sure if it ever wraps around. There seem to be existing checks in this code ensuring that the window ID is nonzero as well?

This suggested code picks `[sc->shareable_content.windows objectAtIndex:0]` without checking if its windowID is actually 0; isn't that incorrect?

jcm93:
I'm also wondering if it would be good to store and check against something else when initializing a window capture, like maybe the owning application's bundle identifier. There's currently a fairly common scenario where a user picks windows soon after boot, then later reboots and starts OBS. Since SCWindow IDs increment beginning on boot, in this situation it's quite possible for the captures to pick up windows that were never chosen.

plugins/mac-capture/mac-sck-video-capture.m
PatTheMav: I think you're hitting an interesting point:

* Logically it should not be an issue to fall back on the very first item in the list, because you're configuring the source and you haven't committed the change yet - you're just setting it up. In that case there should not be any privacy concerns, because the list acts within the scope if "configuring the source"
* _However_ I don't know off the top of my head if any change in this list will actually immediately impact the current program output. Which would mean that "just browsing" through the list of available windows would be immediately visible in the stream. And _that_ would have a privacy impact, however that impact would include the pure act of configuring the source itself.

My expectation for the window ID would be that `0` is an invalid ID, so the first element in the array should never contain a window with that ID at the first index.

Looking at the documentation for `NSWindow windowNumber` it states:

> If the window doesn’t have a window device, the value of this property is equal to or less than 0

The docs for `windowNumbersWithOptions` however state that it would only return the window numbers for _visible_ windows, which implies that they all have a window device (because otherwise they wouldn't be visible).

Device in this case means the `NSDevice` a window is attached to.

PatTheMav:
> I'm also wondering if it would be good to store and check against something else when initializing a window capture, like maybe the owning application's bundle identifier. There's currently a fairly common scenario where a user picks windows soon after boot, then later reboots and starts OBS. Since SCWindow IDs increment beginning on boot, in this situation it's quite possible for the captures to pick up windows that were never chosen.

Logically any window capture-type source should be invalid after a boot, which would result in either empty captures or entirely different windows appearing in the preview. 

This is less of an issue if capturing windows by their window title, but that would require iterating over all window titles for every source, trying to find an exact match.

But IMO that would break macOS' consent-model, because having the same title cannot implicitly construe consent to capture that window.

plugins/mac-capture/mac-sck-video-capture.m
jcm93: > * Logically it should not be an issue to fall back on the very first item in the list, because you're configuring the source and you haven't committed the change yet - you're just setting it up. In that case there should not be any privacy concerns, because the list acts within the scope if "configuring the source"
> * _However_ I don't know off the top of my head if any change in this list will actually immediately impact the current program output. Which would mean that "just browsing" through the list of available windows would be immediately visible in the stream. And _that_ would have a privacy impact, however that impact would include the pure act of configuring the source itself.

I'm mainly thinking about the case where OBS has just launched and there are scenes containing window captures from a prior session. The current behavior is that these sources will hit this code on startup and initialize pointing to nil (this is what "usually" happens), or if there happens to be a window on the system with the same `SCWindow.windowID`, it captures that window. Both seem undesirable.

This change would cause all these sources in the latter case to capture whatever window happens to be the first in the list provided. To me this seems as undesirable as capturing a window that happens to have the same ID.

> Looking at the documentation for `NSWindow windowNumber` it states:

I'm not sure if this corresponds to SCWindow's windowID. I don't believe it does?

PatTheMav: That's a good catch, I didn't consider the use case of OBS launching and not being able to restore capture sessions for the old windows.

In that case the capture should indeed abort early like your PR suggests or - if we want to be fancy about it - show a placeholder.

jcm93:
> > I'm also wondering if it would be good to store and check against something else when initializing a window capture, like maybe the owning application's bundle identifier. There's currently a fairly common scenario where a user picks windows soon after boot, then later reboots and starts OBS. Since SCWindow IDs increment beginning on boot, in this situation it's quite possible for the captures to pick up windows that were never chosen.
> 
> Logically any window capture-type source should be invalid after a boot, which would result in either empty captures or entirely different windows appearing in the preview.
> 
> This is less of an issue if capturing windows by their window title, but that would require iterating over all window titles for every source, trying to find an exact match.
> 
> But IMO that would break macOS' consent-model, because having the same title cannot implicitly construe consent to capture that window.

I think that the window title is not good to use, because it can frequently change during use for some programs. But the window's `owningApplication`'s `bundleIdentifier` should never change during a session.

If it compares the ID of the window as well as the bundle identifier, you would need a large amount of coincidence for a window capture source to ever initialize pointing at a window the user had not picked.

To do that, we would also need to store and handle `sc->application_id` for window capture. Probably complex enough that it would be done in a separate PR, I think.

plugins/mac-capture/mac-sck-video-capture.m
jcm93: 👍 I will do an updated push in a bit.

PatTheMav:
> > > I'm also wondering if it would be good to store and check against something else when initializing a window capture, like maybe the owning application's bundle identifier. There's currently a fairly common scenario where a user picks windows soon after boot, then later reboots and starts OBS. Since SCWindow IDs increment beginning on boot, in this situation it's quite possible for the captures to pick up windows that were never chosen.
> > 
> > 
> > Logically any window capture-type source should be invalid after a boot, which would result in either empty captures or entirely different windows appearing in the preview.
> > This is less of an issue if capturing windows by their window title, but that would require iterating over all window titles for every source, trying to find an exact match.
> > But IMO that would break macOS' consent-model, because having the same title cannot implicitly construe consent to capture that window.
> 
> I think that the window title is not good to use, because it can frequently change during use for some programs. But the window's `owningApplication`'s `bundleIdentifier` should never change during a session.
> 
> If it compares the ID of the window as well as the bundle identifier, you would need a large amount of coincidence for a window capture source to ever initialize pointing at a window the user had not picked.
> 
> To do that, we would also need to store and handle `sc->application_id` for window capture. Probably complex enough that it would be done in a separate PR, I think.

Correct, though I think it's a good enough change. Might even make sense to always use a capture target struct that contains the application bundle id, window ID, and capture type. If the capture type is "Window Capture", we use the window ID to tell SCK exactly which Window we want and use the application bundle ID to avoid mismatches. For "Application Capture" we just use the bundle ID and use a `0` window ID that is otherwise invalid for "Window Capture".

jcm93:
Updated to resolve setting to nil erroneously, put the window initialization in a more explicit `else` block and still return early if target window is nil. Also substituted the constant name `kCGNullWindowID` in place of `0`, to make the meaning more clear, per the conversation above.

This change has the side effect of, when the user loads the source properties for an uninitialized window capture, it will show a blank preview by default until the user actually picks a window. This is slightly undesirable in my mind, but I didn't know offhand a good way for us to default to the first item in the list for only the source properties window and not in the case of source initialization in general.

PatTheMav (commented):
Looks good so far, will have to look into the semaphore-thing in detail to understand how it works again before final approval.

PatTheMav (approved):

Lain-B (approved):

PR #9809, opened 2023-11-05
build-aux: Remove StatusNotifierItem Flatpak permission

tytan652:
### Description

Since Qt 6.2, "org.kde.StatusNotifierItem-2-2" is no longer needed to make the tray feature work

`flatpak-builder-lint` output with exceptions on the manifest before:
```
{
    "errors": [
        "finish-args-broken-kde-tray-permission"
    ],
    "warnings": [
        "finish-args-contains-both-x11-and-wayland"
    ]
}
```
After:
```
{
    "warnings": [
        "finish-args-contains-both-x11-and-wayland"
    ]
}
```

### Motivation and Context

Make the Flatpak pass incoming Flathub validation. Keeping this permission will trigger a `"finish-args-broken-kde-tray-permission"` error while validating.

### How Has This Been Tested?

Ran flatpak-builder-lint (and on CI) locally against the manifest and ran the Flatpak without the permission and the tray icon still works.

### Types of changes

- Bug fix (non-breaking change which fixes an issue)

- Code cleanup (non-breaking change which makes code smaller or more readable)

### Checklist:

- [x] My code has been run through [clang-format](https://github.com/obsproject/obs-studio/blob/master/.clang-format).
- [x] I have read the [**contributing** document](https://github.com/obsproject/obs-studio/blob/master/CONTRIBUTING.rst).
- [x] My code is not on the master branch.
- [x] The code has been tested.
- [x] All commit messages are properly formatted and commits squashed where appropriate.
- [x] I have included updates to all appropriate documentation.

PR #9776, opened 2023-10-29
(Do not look at this)(Seriously) Flathub publishing updates

GeorgesStavracas:
### Description

We need to pass the build-log-url now, and validate our output with the flatpak-builder-linter tool. Let's do that.

### Motivation and Context

Flathub is changing, and so should we.

### How Has This Been Tested?

It hasn't.

### Types of changes

 - Tweak (non-breaking change to improve existing functionality)

### Checklist:

- [x] My code has been run through [clang-format](https://github.com/obsproject/obs-studio/blob/master/.clang-format).
- [x] I have read the [**contributing** document](https://github.com/obsproject/obs-studio/blob/master/CONTRIBUTING.rst).
- [x] My code is not on the master branch.
- [ ] The code has been tested.
- [ ] All commit messages are properly formatted and commits squashed where appropriate.
- [x] I have included updates to all appropriate documentation.

PR #9814, opened 2023-11-06
Flathub publishing updates

tytan652:
### Description

We need to pass the build-log-url now, and validate our output with the flatpak-builder-lint tool. Let's do that.

Depends on:
- #9809 to pass validation

Supersedes #9776 and the differences are:
- Avoid the use of the org.flatpak.Builder Flatpak to use the linter since it does not seem to work inside a container in GHA. So this PR relies on cloning the linter repo and using Poetry to run it.
- Linter warnings and errors are reformatted as action annotations
- Validation steps are factorized as a custom action to avoid repetition
- Some customization were added to allow validating the Flatpak repository on pushes to avoid encountering an error only while on publish validation.
- "`--exceptions`" is added since we need one and have it for `"finish-args-flatpak-spawn-access"` 
  - https://github.com/flathub/flatpak-builder-lint/pull/197

### Motivation and Context

Flathub ~~is changing~~ *has surely already changed*, and so ~should~ we *must*.

### How Has This Been Tested?

CI on PR on my fork and this PR CI.

### Types of changes

### Checklist:

  - Except if you want to squash more

### DIFF
--- a/.github/actions/flatpak-builder-lint/action.yaml
+++ b/.github/actions/flatpak-builder-lint/action.yaml
@@ -7,6 +7,10 @@ inputs:
   path:
     description: Path to flatpak-builder manifest or Flatpak build directory
     required: true
+  validateToPublish:
+    description: If false, turns some errors to non-errors for non-publish workflow
+    required: false
+    default: true
   workingDirectory:
     description: Working directory to clone flatpak-builder-lint
     required: false
@@ -65,9 +69,15 @@ runs:
         done
 
         n_errors=$(echo $ret | jq '.errors | length')
-        for ((i = 0; i < $n_errors; i++)); do
+        for ((i = 0 ; i < $(echo $ret | jq '.errors | length') ; i++)); do
           error=$(echo $ret | jq ".errors[$i]")
-          echo "::error::$error found in the Flatpak ${{ inputs.artifact }}"
+
+          if [[ "${{ inputs.validateToPublish }}" == "false" && "${error//\"}" == "appstream-screenshots-not-mirrored" ]]; then
+            echo "::notice::$error found and ignored in the Flatpak ${{ inputs.artifact }}"
+            n_errors=$(($n_errors-1))
+          else
+            echo "::error::$error found in the Flatpak ${{ inputs.artifact }}"
+          fi
         done
 
         [[ $n_errors == 0 ]] || exit 2
--- a/.github/actions/flatpak-builder-lint/action.yaml
+++ b/.github/actions/flatpak-builder-lint/action.yaml
@@ -1,0 +1,73 @@
+name: Run flatpak-builder-lint
+description: Runs flatpak-builder-lint with exceptions
+inputs:
+  artifact:
+    description: Type of artifact to lint (builddir, repo, manifest)
+    required: true
+  path:
+    description: Path to flatpak-builder manifest or Flatpak build directory
+    required: true
+  workingDirectory:
+    description: Working directory to clone flatpak-builder-lint
+    required: false
+    default: ${{ github.workspace }}
+runs:
+  using: composite
+  steps:
+    - name: Check artifact type
+      shell: bash
+      working-directory: ${{ inputs.workingDirectory }}
+      run: |
+        : Check artifact input
+        case "${{ inputs.artifact }}" in
+          builddir);;
+          repo);;
+          manifest);;
+          *)
+            echo "::error::Given artifact type is incorrect"
+            exit 2
+            ;;
+        esac
+
+    - uses: actions/checkout@v3
+      with:
+        repository: flathub/flatpak-builder-lint
+        ref: v2.0.13
+        path: flatpak-builder-lint
+        set-safe-directory: ${{ inputs.workingDirectory }}
+
+    - name: Install Dependencies 🛍️
+      shell: bash
+      working-directory: ${{ inputs.workingDirectory }}
+      run: |
+        : Install Dependencies 🛍️
+        echo ::group::Install Dependencies
+        dnf install -y -q poetry jq
+        poetry -q -C flatpak-builder-lint install
+        echo ::endgroup::
+
+    - name: Run flatpak-builder-lint
+      id: result
+      shell: bash
+      working-directory: ${{ inputs.workingDirectory }}
+      run: |
+        : Run flatpak-builder-lint
+        exit_code=0
+        ret=$(poetry -C flatpak-builder-lint run flatpak-builder-lint --exceptions ${{ inputs.artifact }} ${{ inputs.path }}) || exit_code=$?
+        if [[ $exit_code != 0 && -z "$ret" ]]; then
+          echo "::error::Error while running flatpak-builder-lint"
+          exit 2
+        fi
+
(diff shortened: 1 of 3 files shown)


tytan652:
Rebased and fixed a typo in a commit description.

.github/actions/flatpak-builder-lint/action.yaml
```diff
@@ -0,0 +1,53 @@
+name: Run flatpak-builder-lint
+description: Runs flatpak-builder-lint with exceptions
+inputs:
+  artifactToLint:
+    description: Type of artifact to lint (builddir, repo, manifest)
+    required: true
+  path:
```
GeorgesStavracas, line 7: `manifestPath`?

.github/actions/flatpak-builder-lint/action.yaml
```diff
@@ -0,0 +1,53 @@
+name: Run flatpak-builder-lint
+description: Runs flatpak-builder-lint with exceptions
+inputs:
+  artifactToLint:
```
GeorgesStavracas, line 4: Just `artifact`?

.github/actions/flatpak-builder-lint/action.yaml
```diff
@@ -7,6 +7,10 @@ inputs:
   path:
     description: Path to flatpak-builder manifest or Flatpak build directory
     required: true
+  validateToPublish:
+    description: If false, turns some errors to non-errors for non-publish workflow
+    required: false
+    default: true
```
GeorgesStavracas, line 13: What's the purpose of this input, if all consumers pass `false`?

GeorgesStavracas (changes requested):

.github/actions/flatpak-builder-lint/action.yaml
tytan652: No, the name needs cover repository and build directory cases.

.github/actions/flatpak-builder-lint/action.yaml
tytan652: Wrong, in `publish.yaml` all consumers pass implicitly `true`

.github/actions/flatpak-builder-lint/action.yaml
GeorgesStavracas: Why are you using the same variable for two different things then?

.github/actions/flatpak-builder-lint/action.yaml
tytan652: Made input name less verbose.

.github/actions/flatpak-builder-lint/action.yaml
GeorgesStavracas: Ah, I see. Understood.

.github/actions/flatpak-builder-lint/action.yaml
tytan652: If you feel like the two last commit are unnecessary, they just try to avoid encountering potential errors only while publishing.
Feel free to say it.

tytan652: It doesn't disable some test, it replace how the Github action should behave with some errors.

Screenshot from 2023-11-08 14-00-00

tytan652: > why is that undesireable?

I just realized that I misread you 🤦. You are in fact okay with what the two last commit try to implement.
Stupid me.

.github/actions/flatpak-builder-lint/action.yaml
```diff
@@ -0,0 +1,63 @@
...
+        poetry -q -C flatpak-builder-lint install
+        echo ::endgroup::
+
+    - name: Run flatpak-builder-lint
+      id: result
+      shell: bash
+      working-directory: ${{ inputs.workingDirectory }}
+      run: |
+        : Run flatpak-builder-lint
+        ret=$(poetry -C flatpak-builder-lint run flatpak-builder-lint --exceptions ${{ inputs.artifact }} ${{ inputs.path }} || true)
```
PatTheMav, line 44: From the docs it seems like `flatpak-builder-lint` would fail if `inputs.artifact` contains an unsupported positional argument, but this code always falls back to a zero return value.

So I'd probably would rather check if the input contains either "builddir", "repo", or "manifest" and fail gracefully with an error message.

This would be good practise as you're designing a repository action and it should validate the input before handing it off to the tool.

.github/actions/flatpak-builder-lint/action.yaml
```diff
@@ -0,0 +1,63 @@
...
+
+        for ((i = 0 ; i < $(echo $ret | jq '.warnings | length') ; i++)); do
```
PatTheMav, line 46: Instead of using counters (and then run `jq` on the returned data again), just loop over the entries returned by `jq` - if no values are returned, no loop iteration runs:

```suggestion
            local success=1
            poetry -C flatpak-builder-lint run flatpak-builder-lint --exceptions ${{ inputs.artifact }} ${{ inputs.path }} > linter_output.json || success=0

            if (( success )); then
              while read -r warning; do
                  echo "::warning::${warning} found in the Flatpak ${{ inputs.artifact }}"
              done <<< "$(jq -r '.warnings | .[]?' linter_output.json )"
```
The `jq` expression will only yield warning entries if the key actually exists, otherwise the output is empty (and return code is always `0`). As each warning entry will be printed to stdout on a single line, `while` will automatically iterate over each warning.

Potentially even more elegant would be to handle all that in a single loop:

```
              IFS=';' while read -r type warning; do
                  if [[ "${type}" == "warning" ]]; then
                    echo "::warning::${warning} found in the Flatpak ${{ inputs.artifact }}"
                  elif [[ "${type}" == "error" && "${{ inputs.validateToPublish }}" == "true" ]]; then
                    echo "::error::${warning} found in the Flatpak ${{ inputs.artifact }}"
                  fi
              done <<< "$(jq -r '(.warnings | try map("warning," + .)) + (.errors | try map("error," + .)) | .[]?' linter_output.json)"
```

tytan652: - When flatpak-builder-lint find an error in the manifest/builddir/repo, it returns a non-zero exit code
- The quote are kept on purpose in the annotations
- Notice are added to show "ignored" without marking them as red errors
- `validateToPublish` does not ignore everything just specific error that are clearly fixed in the publish CI
- I purposefully do not write files to not create unneeded files

.github/actions/flatpak-builder-lint/action.yaml
tytan652: I improved the management of the exit code but I kept a step to check if `artifact` is valid to exit earlier.

.github/actions/flatpak-builder-lint/action.yaml
tytan652: Using while introduce an empty line when the linter don't find any issue (empty variable/file).

```
while read -r warning; do
echo "::warning::$warning found in the Flatpak manifest"
done <<< "$(jq '.warnings[]?' test.json)"

::warning:: found in the Flatpak manifest
```

I'll keep for loops for now.

PatTheMav (approved):
Still have some code-style nits, but functionally this should be fine.

GeorgesStavracas (approved):

PR #9816, opened 2023-11-06
cmake: Force PDB generation on Windows for MSVC builds in all configs

PatTheMav:
### Description
Adds necessary compiler and linker flags to mimic MSVC's default settings which always generate PDBs regardless of optimisation and never use incremental linking.

### Motivation and Context
CMake 3.25 changed the way PDB generation is handled by only enabling it for Debug and RelWithDebInfo builds, which prohibits generation of fully optimized builds with associated symbols (which is MSVC's default).

If configuring with CMake 3.25 or above, enable this globally for builds using MSVC and fall back to embedded debug information for anything else (which would probably be clang-cl).

The necessary linker flag `/DEBUG` also needs to be added manually to the "Release" configuration, as CMake removes it by default.

With these changes, users get the benefit of optimised builds while we retain the ability to symbolicate crashes locally with the shipped PDBs.

### How Has This Been Tested?
Tested with CMake 3.27 and Visual Studio 17 2022, building in "Release" configuration" all PDBs are correctly generated and copied.

### Types of changes

### Checklist:

### DIFF
--- a/cmake/windows/compilerconfig.cmake
+++ b/cmake/windows/compilerconfig.cmake
@@ -23,6 +23,15 @@ if(CMAKE_VERSION VERSION_EQUAL 3.24.0)
   set(THREADS_HAVE_PTHREAD_ARG FALSE)
 endif()
 
+# CMake 3.25 changed the way symbol generation is handled on Windows
+if(CMAKE_VERSION VERSION_GREATER_EQUAL 3.25.0)
+  if(CMAKE_C_COMPILER_ID STREQUAL "MSVC")
+    set(CMAKE_MSVC_DEBUG_INFORMATION_FORMAT ProgramDatabase)
+  else()
+    set(CMAKE_MSVC_DEBUG_INFORMATION_FORMAT Embedded)
+  endif()
+endif()
+
 message(DEBUG "Current Windows API version: ${CMAKE_VS_WINDOWS_TARGET_PLATFORM_VERSION}")
 if(CMAKE_VS_WINDOWS_TARGET_PLATFORM_VERSION_MAXIMUM)
   message(DEBUG "Maximum Windows API version: ${CMAKE_VS_WINDOWS_TARGET_PLATFORM_VERSION_MAXIMUM}")
@@ -47,13 +56,19 @@ add_compile_options(
   "$<$<COMPILE_LANG_AND_ID:C,MSVC>:${_obs_msvc_c_options}>"
   "$<$<COMPILE_LANG_AND_ID:CXX,MSVC>:${_obs_msvc_cpp_options}>"
   "$<$<COMPILE_LANG_AND_ID:C,Clang>:${_obs_clang_c_options}>"
-  "$<$<COMPILE_LANG_AND_ID:CXX,Clang>:${_obs_clang_cxx_options}>")
+  "$<$<COMPILE_LANG_AND_ID:CXX,Clang>:${_obs_clang_cxx_options}>"
+  $<$<NOT:$<CONFIG:Debug>>:/Gy>)
 
 add_compile_definitions(UNICODE _UNICODE _CRT_SECURE_NO_WARNINGS _CRT_NONSTDC_NO_WARNINGS $<$<CONFIG:DEBUG>:DEBUG>
                         $<$<CONFIG:DEBUG>:_DEBUG>)
 
-add_link_options("$<$<NOT:$<CONFIG:Debug>>:/OPT\:REF>" "$<$<CONFIG:Debug>:/INCREMENTAL\:NO>"
-                 "$<$<CONFIG:RelWithDebInfo>:/INCREMENTAL\:NO>" "$<$<CONFIG:RelWithDebInfo>:/OPT\:ICF>" /Brepro)
+# cmake-format: off
+add_link_options($<$<NOT:$<CONFIG:Debug>>:/OPT:REF>
+                 $<$<NOT:$<CONFIG:Debug>>:/OPT:ICF>
+                 $<$<NOT:$<CONFIG:Debug>>:/INCREMENTAL:NO>
+                 /DEBUG
+                 /Brepro)
+# cmake-format: on
 
 if(CMAKE_COMPILE_WARNING_AS_ERROR)
   add_link_options(/WX)
--- a/cmake/windows/helpers.cmake
+++ b/cmake/windows/helpers.cmake
@@ -251,14 +251,14 @@ function(_target_install_obs target)
     POST_BUILD
     COMMAND "${CMAKE_COMMAND}" -E make_directory "${OBS_OUTPUT_DIR}/$<CONFIG>/${_TIO_DESTINATION}"
     COMMAND "${CMAKE_COMMAND}" -E copy ${target_file} "${OBS_OUTPUT_DIR}/$<CONFIG>/${_TIO_DESTINATION}"
-    COMMAND "${CMAKE_COMMAND}" -E $<IF:$<CONFIG:Debug,RelWithDebInfo>,copy,true> ${target_pdb_file}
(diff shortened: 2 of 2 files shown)


RytoEX:
As an aside, when installing the CMake workload/tools (I think this is selected by default, but I don't recall), Visual Studio 2022 17.7 (the current stable release) ships with CMake 3.26.4, and Visual Studio 17.8 (the current Preview release) ships with CMake 3.27.2. I think the version gating is fine for now though.

derrod (approved):

cmake/windows/compilerconfig.cmake
```diff
@@ -47,13 +56,19 @@ add_compile_options(
...
 
 add_compile_definitions(UNICODE _UNICODE _CRT_SECURE_NO_WARNINGS _CRT_NONSTDC_NO_WARNINGS $<$<CONFIG:DEBUG>:DEBUG>
                         $<$<CONFIG:DEBUG>:_DEBUG>)
 
-add_link_options("$<$<NOT:$<CONFIG:Debug>>:/OPT\:REF>" "$<$<CONFIG:Debug>:/INCREMENTAL\:NO>"
-                 "$<$<CONFIG:RelWithDebInfo>:/INCREMENTAL\:NO>" "$<$<CONFIG:RelWithDebInfo>:/OPT\:ICF>" /Brepro)
+# cmake-format: off
+add_link_options($<$<NOT:$<CONFIG:Debug>>:/OPT:REF>
+                 $<$<NOT:$<CONFIG:Debug>>:/OPT:ICF>
+                 $<$<NOT:$<CONFIG:Debug>>:/INCREMENTAL:NO>
```
RytoEX, line 68: This was previously:
```
$<$<CONFIG:Debug>:/INCREMENTAL:NO>
$<$<CONFIG:RelWithDebInfo>:/INCREMENTAL:NO>
```

Was this intentionally changed to instead exclude Debug builds?

cmake/windows/compilerconfig.cmake
```diff
@@ -47,13 +56,19 @@ add_compile_options(
...
+  $<$<NOT:$<CONFIG:Debug>>:/Gy>)
 
 add_compile_definitions(UNICODE _UNICODE _CRT_SECURE_NO_WARNINGS _CRT_NONSTDC_NO_WARNINGS $<$<CONFIG:DEBUG>:DEBUG>
                         $<$<CONFIG:DEBUG>:_DEBUG>)
 
-add_link_options("$<$<NOT:$<CONFIG:Debug>>:/OPT\:REF>" "$<$<CONFIG:Debug>:/INCREMENTAL\:NO>"
-                 "$<$<CONFIG:RelWithDebInfo>:/INCREMENTAL\:NO>" "$<$<CONFIG:RelWithDebInfo>:/OPT\:ICF>" /Brepro)
+# cmake-format: off
+add_link_options($<$<NOT:$<CONFIG:Debug>>:/OPT:REF>
+                 $<$<NOT:$<CONFIG:Debug>>:/OPT:ICF>
```
RytoEX, line 67: This was previously:
```
$<$<CONFIG:RelWithDebInfo>:/OPT:ICF>
```
Was this intentionally expanded to cover Release and MinSizeRel?

cmake/windows/compilerconfig.cmake
```diff
@@ -47,13 +56,19 @@ add_compile_options(
...
+                 $<$<NOT:$<CONFIG:Debug>>:/INCREMENTAL:NO>
+                 /DEBUG
```
RytoEX, line 69: [MS docs](https://learn.microsoft.com/en-us/cpp/build/reference/opt-optimizations?view=msvc-170) say:
> If [/DEBUG](https://learn.microsoft.com/en-us/cpp/build/reference/debug-generate-debug-info?view=msvc-170) is specified, the default for /OPT is NOREF, and all functions are preserved in the image. To override this default and optimize a debug build, specify /OPT:REF.

Here, we are unconditionally applying `/DEBUG`, but conditionally overriding `/OPT:NOREF` to `/OPT:REF` for non-Debug configurations, is that correct?

cmake/windows/compilerconfig.cmake
RytoEX, line 68: [MS docs](https://learn.microsoft.com/en-us/cpp/build/reference/incremental-link-incrementally?view=msvc-170) say:
> Most programs can be linked incrementally. However, some changes are too great, and some options are incompatible with incremental linking. LINK performs a full link if any of the following options are specified:
> * Link Incrementally isn't selected (`/INCREMENTAL:NO`)
> * `/OPT:REF` is selected
> * `/OPT:ICF` is selected

So this seems superfluous, but I am okay keeping it if we want to be explicit, or if this is required to override something CMake is otherwise doing.

cmake/windows/compilerconfig.cmake
```diff
@@ -47,13 +56,19 @@ add_compile_options(
   "$<$<COMPILE_LANG_AND_ID:C,MSVC>:${_obs_msvc_c_options}>"
   "$<$<COMPILE_LANG_AND_ID:CXX,MSVC>:${_obs_msvc_cpp_options}>"
   "$<$<COMPILE_LANG_AND_ID:C,Clang>:${_obs_clang_c_options}>"
-  "$<$<COMPILE_LANG_AND_ID:CXX,Clang>:${_obs_clang_cxx_options}>")
+  "$<$<COMPILE_LANG_AND_ID:CXX,Clang>:${_obs_clang_cxx_options}>"
+  $<$<NOT:$<CONFIG:Debug>>:/Gy>)
```
RytoEX, line 60: [MS docs](https://learn.microsoft.com/en-us/cpp/build/reference/opt-optimizations?view=msvc-170) say:
> Because /OPT:ICF can cause the same address to be assigned to different functions or read-only data members (that is, const variables when compiled by using /Gy), it can break a program that depends on unique addresses for functions or read-only data members. For more information, see [/Gy (Enable Function-Level Linking)](https://learn.microsoft.com/en-us/cpp/build/reference/gy-enable-function-level-linking?view=msvc-170).

RytoEX (commented):
Just some questions so that I understand the changes.

cmake/windows/compilerconfig.cmake
notr1ch: I thought we already has this one on for a while, it should be completely safe for our purposes.

cmake/windows/compilerconfig.cmake
PatTheMav: From what I could tell reading through CMake's open issues is that incremental linking is enabled outside of Debug and REF and ICF are also explicitly disabled when they should be enabled. Hence why we force these values for the desired configs.

cmake/windows/compilerconfig.cmake
PatTheMav: `/DEBUG` is necessary to make the linker create PDB files.

cmake/windows/compilerconfig.cmake
PatTheMav: MSVC (like Xcode) only distinguishes between Debug and Release (RelWithDebInfo and MinSizeRel are CMake-isms), this code replicates that behaviour.

cmake/windows/compilerconfig.cmake
PatTheMav: See above - this logically flattens CMake-logic to MSVC-logic.

cmake/windows/compilerconfig.cmake
RytoEX: Got it.

cmake/windows/compilerconfig.cmake
RytoEX: Got it.

cmake/windows/compilerconfig.cmake
RytoEX: Got it.

Lain-B (approved):

PR #9821, opened 2023-11-07
UI: Add crash handling options per thread

exeldro:
### Description
Add crash handling options per thread
Allows threads to not kill obs on unhandled exception
When only the thread is killed:
image
When the process is killed:
image

### Motivation and Context
Threads created in the obs process have an option to not kill obs process on unhandled exceptions.
For example for the BrowserManagerThread, that way the user can end stream on their own terms and don't end it by closing the crash popup

### How Has This Been Tested?
On windows 64

### Types of changes

### Checklist:

### DIFF
--- a/UI/obs-app.cpp
+++ b/UI/obs-app.cpp
@@ -2618,9 +2618,12 @@ static int run_program(fstream &logFile, int argc, char *argv[])
 
 #ifdef _WIN32
 
-#define CRASH_MESSAGE                                                      \
-	"Woops, OBS has crashed!\n\nWould you like to copy the crash log " \
-	"to the clipboard? The crash log will still be saved to:\n\n%s"
+#define CRASH_MESSAGE "Woops, OBS has crashed!\n\n"
+#define CRASH_THREAD_MESSAGE "Woops, a part of OBS has crashed!\n\n"
+#define CRASH_COPY_MESSAGE                      \
+	"Would you like to copy the crash log " \
+	"to the clipboard?"
+#define CRASH_MESSAGE_SAVE " The crash log will still be saved to:\n\n"
 
 static void main_crash_handler(const char *format, va_list args,
 			       void * /* param */)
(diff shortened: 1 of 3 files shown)


UI/obs-app.cpp
```diff
@@ -2618,9 +2618,12 @@ static int run_program(fstream &logFile, int argc, char *argv[])
 
 #ifdef _WIN32
 
-#define CRASH_MESSAGE                                                      \
-	"Woops, OBS has crashed!\n\nWould you like to copy the crash log " \
-	"to the clipboard? The crash log will still be saved to:\n\n%s"
+#define CRASH_MESSAGE "Woops, OBS has crashed!\n\n"
+#define CRASH_THREAD_MESSAGE "Woops, a part of OBS has crashed!\n\n"
```
WizardCM, line 2622:
```suggestion
#define CRASH_THREAD_MESSAGE "Woops, a component of OBS has crashed!\n\n"
```

WizardCM (commented):
Overall, I don't think a user will understand the difference, and they won't understand what additional things clicking "Yes" will do. Currently, you can be 100% sure that clicking "Yes" in a crash dialog will close the application, even if it still looks to be running correctly, so some users will notice that OBS itself is still running and just move the dialog out of view.

I personally think a change like this should be part of a larger crash handler overhaul that just silently restarts whatever thread died, and pops a smaller notification without interrupting the user.

notr1ch:
I don't like this idea, users don't know or care if OBS crashed in the main thread or an encoder thread or whatever. While it's true that a crashed browser thread _might_ not be immediately catastrophic, the majority of other threads crashing will end up taking down the program due to deadlocks or memory corruption or whatever. Pretending the situation is recoverable isn't wise.

exeldro:
Very often I see streamers streaming with the error dialog open, because a non essential thread has crashed, scared of touching the error dialog. It doesn't matter if they click yes, no or the close button OBS will close.
This option to not close OBS needs to be set per thread, so main and encoder thread will still close OBS.
Maybe it is better to ask the user if they want to continue unstable or not?

PatTheMav:
I think that the general idea is correct, but making thread crashes non-critical only makes sense if OBS itself can internally recover from this crash and the threads themselves are not critical to the function of OBS.

And _if_ OBS can recover from that (e.g. it detects a crash of the `media-io` thread and restarts it gracefully) then it should suppress such a warning and instead write a message to the log.

Otherwise any thread crash can lead to undefined application state because our code expects all threads to be running and makes certain assumptions about the state the threads are in.

notr1ch:
There is no safe way to recover from a crash. The thread might hold all kinds of mutexes, leak references, corrupt the heap, etc. Pretending everything is fine is not a good option as it could lead to data corruption or further bugs that become extremely difficult to diagnose.

Fenrirthviti:
What about as an alternative to a thread crash vs a full crash is that we attempt a "normal" shutdown? I.e. exiting all other threads, stopping encoders, etc. with a warning that says "OBS has experienced an error, and will try to close as safely as possible" or something similar. That way we have the end-result of a user having the opportunity to take a moment if things are sort of working before doing the shutdown, but doesn't imply this is recoverable in any way.

derrod:
> What about as an alternative to a thread crash vs a full crash is that we attempt a "normal" shutdown? I.e. exiting all other threads, stopping encoders, etc. with a warning that says "OBS has experienced an error, and will try to close as safely as possible" or something similar. That way we have the end-result of a user having the opportunity to take a moment if things are sort of working before doing the shutdown, but doesn't imply this is recoverable in any way.

The problem is data corruption. If we try to shut down normally, i.e. save data to disk, we could be writing invalid or broken data.

exeldro:
Now people switch scene collection with the crash dialog open to save the scene collection before the forced exit.
If you have done lots of changes in obs and crash occurs I would at least like to have the option to try and save, this can be with a big warning message. Not saving just gives me the same state as my backup and missing al changes I have done.

notr1ch:
That sounds like a separate issue if OBS is not saving frequently enough that significant amounts of changes are lost in a crash.

PR #9822, opened 2023-11-07
CI: Sign and package Windows builds

derrod:
### Description

This adds a new workflow that will create signed Windows builds and installer on tag pushes.

To facilitate code signing in a way that complies with the new requirements to use HSM (Hardware Security Modules) for key storage[^1] we utilise [Google Cloud's Key Management Service](https://cloud.google.com/security-key-management).

For signing Windows binaries we continue to use `signtool` with Google's CNG Provider that allows it to run the signing operation on the cloud HSM: https://cloud.google.com/kms/docs/reference/cng-signtool

#### ToDo:
- [ ] Actually acquire new code signing certificate **(In progress)**
  + Must be ECDSA, Google's CNG provider does not support RSA keys (yet): https://github.com/GoogleCloudPlatform/kms-integrations/issues/16
  + GlobalSign will work according to https://icedev.pl/posts/setting-up-ev-code-signing-google-hsm-fips-140-2/
  + DigiCert is probably also fine based on responses in issue linked above
  + If we wish to import the current keys to migrate before acquiring a new cert we will probably have to use [jsign](https://github.com/ebourg/jsign)
- [x] Figure out if we want to do delta patch generation elsewhere (takes a long time + storage on GitHub CI)
  + **Resolution:** It's not actually to bad (<20 minutes, ~600 MiB per version) as long as there's only one CEF version change to cover
- [x] Maybe find a better way for downloading old builds + PDBs for delta generation/packaging (GCS? S3? SFTP?)
  + **Resolution:** Google Cloud Storage (GCS) seems like a good option and we're pretty much going to stay within the free tier
- [x] #9816 

[^1]: https://knowledge.digicert.com/general-information/new-private-key-storage-requirement-for-standard-code-signing-certificates-november-2022

### Motivation and Context

Reduces bus factor a bit.

### How Has This Been Tested?

Tested on my fork with test cert and keys: https://github.com/derrod/obs-studio/actions/runs/6817205390/job/18540744248

### Types of changes

### Checklist:


.github/actions/bouf/action.yaml
```diff
@@ -0,0 +1,122 @@
...
+runs:
+  using: composite
+  steps:
+    - name: Prepare Artifact
+      shell: pwsh
+      run: |
+        # Extract artifact in build dir and remove zip
+        cd build
+        Expand-Archive -Path *.zip -DestinationPath .
+        del *.zip
```
PatTheMav, line 34: I'd suggest to use `Remove-Item` - our PowerShell code mostly tries to keep using first-class PowerShell idioms as much as possible, only falling back on program invocations if necessary.

.github/actions/bouf/action.yaml
```diff
@@ -0,0 +1,122 @@
...
+
+    - name: Setup bouf
+      shell: pwsh
+      env:
+        BOUF_VERSION: 'v0.6.0'
+        BOUF_HASH: '06861ae6d714b754602000bd51607b5b72e655b978b7e7497b0eb970b1613dd3'
+        BOUF_NSIS_HASH: 'bb0561a76cecaba62785642dd71052a9fe53458e45df108ca08867ee333b80bd'
+      run: |
+        # Download bouf release
+        mkdir bouf && cd bouf
```
PatTheMav, line 44: Copy the `Ensure-Location` PowerShell module from the GitHub build scripts to the GitHub action and source it from there.

.github/actions/bouf/action.yaml
```diff
@@ -0,0 +1,122 @@
...
+        
+        curl -o bouf-windows.zip -L https://github.com/obsproject/bouf/releases/download/$Env:BOUF_VERSION/bouf-windows-$Env:BOUF_VERSION.zip
```
PatTheMav, line 46: Copy the `Invoke-External` PowerShell module from the GitHub build scripts to the GitHub action and source it from there.

Also enables proper error handling whenever a CLI application fails to run.

.github/actions/bouf/action.yaml
```diff
@@ -0,0 +1,122 @@
...
+        workload_identity_provider: ${{ inputs.gcpWorkloadIdentityProvider }}
+        service_account: ${{ inputs.serviceAccountName }}
+
+    - name: Download old builds
+      shell: pwsh
+      env:
+        RCLONE_DRIVE_ENV_AUTH: 'true'
+        RCLONE_DRIVE_SHARED_WITH_ME: 'true'
+      run: |
+        rclone copy --transfers 40 :drive:old_builds "${{ github.workspace }}/old_builds"
```
PatTheMav, line 98: Shouldn't `robocopy` fulfil that need? Especially for local file operations, `rclone` as a third-party library seems unnecessary.

.github/actions/bouf/action.yaml
```diff
@@ -0,0 +1,122 @@
...
+
+    - name: Run bouf
+      shell: pwsh
+      env:
+        UPDATER_PRIVATE_KEY: ${{ inputs.updaterPrivateKey }}
+      run: |
+        &"${{ github.workspace }}\bouf\bin\bouf.exe" `
```
PatTheMav, line 105: Use splatting - put all arguments into an array, then just invoke via `@<array>`. That also avoids the use of the backtick escaping of newlines.

.github/workflows/push.yaml
```diff
@@ -194,6 +194,58 @@ jobs:
...
+          path: "repo"
+          fetch-depth: 0
+          ref: ${{ github.ref }}
+
+      - name: Set Up Environment 🔧
+        id: setup
+        shell: bash
+        run: |
+          channel='stable'
+          if [[ ${GITHUB_REF_NAME} =~ (beta|rc) ]]; then
```
PatTheMav, line 219: Aren't we always escaping the first token to the test operation in Bash? IIRC it's not required in ZSH but in Bash you should still always double-quote the first token to avoid automatic tokenisation (if the variable contains spaces the effective code would be `[[ some value =~ (beta|rc) ]]` and that's invalid shell code.

build-aux/bouf/cng-release-signing-key.pem
```diff
@@ -0,0 +1,5 @@
+-----BEGIN PUBLIC KEY-----
+MHYwEAYHKoZIzj0CAQYFK4EEACIDYgAEtfLbXkHUVc9oUPTNyaEK3hIwmuGRoTtd
+6zDhwqjJuYaMwNd1aaFQLMawTwZgR0Xn27ymVWtqJHBe0FU9BPIQ+SFmKw+9jSwu
+/FuqbJnLmTnWMJ1jRCtyHNZawvv2wbiB
+-----END PUBLIC KEY-----
```
PatTheMav, line 5: Missing newline at the end of the file.

build-aux/bouf/config.toml
PatTheMav, line 1: Is this a generic `bouf` config file that should be used locally as well? If it's only used for the GitHub action, it should be moved into the action's directory.

.github/actions/bouf/action.yaml
derrod: rclone is used to download data from a remote storage system (google cloud in this case), robocopy is just for local copies.

PatTheMav: Ah my mistake, that's fine then.

.github/actions/bouf/action.yaml
derrod: The repo is checked out in this case so I can just import it rather than having to copy it, no? I don't think they'd be imported automatically.

.github/actions/bouf/action.yaml
derrod: I don't really see an advantage to that, in fact, it would require more backticks to escape all the `"`!

.github/workflows/push.yaml
derrod: I think that is correct, though git refs shouldn't be able to have spaces. I haven't bothered rewriting this to powershell yet, so it was mostly just a stripped down version of what's used for the appcast.

.github/actions/bouf/action.yaml
PatTheMav: Yeah both work, I’d just prefer having them contained as that would make extracting them into a separate repo easier if we chose to do that.

.github/actions/bouf/action.yaml
RytoEX: I don't think there's any extra escaping? PowerShell should do that for you. Something like this @PatTheMav ?
```pwsh
$boufOptions = @(
    '--config', "${{ github.workspace }}/repo/build-aux/bouf/config.toml"
    '--version', "${{ inputs.version }}"
    '--branch', "${{ inputs.channel }}"
    '--notes-file', "${{ github.workspace }}/notes.rst"
    '-i', "${{ github.workspace }}/build"
    '-p', "${{ github.workspace }}/old_builds"
    '-o', "${{ github.workspace }}/output"
)

Invoke-External "${{ github.workspace }}\bouf\bin\bouf.exe" @boufOptions
```

derrod: I suppose that works, I was thinking of having to put flag and value into one but bouf doesn't use the `--flag=<value>` form.

PatTheMav (approved):

.github/actions/bouf/action.yaml
```diff
@@ -0,0 +1,127 @@
...
+    description: Private key for signing update manifest
+    required: true
+  gcpWorkloadIdentityProvider:
+    description: GCP Identity Provider Pool ID
+    required: true
+  serviceAccountName:
+    description: Google service account name
+    required: true
+  version:
+    description: Version string (e.g. 30.0.0-rc1)
```
RytoEX, line 14:
```suggestion
    description: Version string (e.g., 30.0.0-rc1)
```

.github/actions/bouf/action.yaml
```diff
@@ -0,0 +1,127 @@
...
+    default: ''
+
+runs:
+  using: composite
+  steps:
+    - name: Extract Artifact
+      shell: pwsh
+      run: |
+        Expand-Archive -Path build\*.zip -DestinationPath build
+        Remove-Item build\*.zip
```
RytoEX, line 32: Did the asterisks actually need to be escaped?

.github/actions/bouf/action.yaml
```diff
@@ -0,0 +1,127 @@
...
+        . $env:GITHUB_ACTION_PATH\Ensure-Location.ps1
+        . $env:GITHUB_ACTION_PATH\Invoke-External.ps1
+        Ensure-Location gcng
+        
+        Invoke-External gh release download "$env:CNG_TAG" -R "GoogleCloudPlatform/kms-integrations" -p "*amd64.zip"
+        Expand-Archive -Path *.zip
+        $sigPath = Get-ChildItem *.sig -Recurse
+        $msiPath = Get-ChildItem *.msi -Recurse
+        # Verify digital signature against Google's public key
+        Invoke-External openssl dgst -sha384 -verify "$env:GITHUB_ACTION_PATH/cng-release-signing-key.pem" -signature $sigPath $msiPath
```
RytoEX, line 74: I suppose if we're verifying the digital signature, we don't need to do a hash check?

.github/actions/bouf/action.yaml
```diff
@@ -0,0 +1,127 @@
...
+        # Finally, install the CNG provider
+        Invoke-External msiexec /i $msiPath /qn /norestart
+
+    - name: Install rclone and pandoc
```
RytoEX, line 78:
```suggestion
    - name: Install pandoc and rclone
```

Super nit. Just bothers me.

.github/actions/bouf/action.yaml
```diff
@@ -0,0 +1,127 @@
...
+      shell: pwsh
+      run: choco install pandoc rclone --no-progress -y
```
RytoEX, line 80: I'd have suggested winget for this, except pandoc is a bit outdated (3.1.1 on winget, 3.1.9 on Chocolatey) and rclone is not yet available on winget. Something to keep in mind for the future if it reaches parity maybe.

.github/actions/bouf/action.yaml
```diff
@@ -0,0 +1,127 @@
...
+
+    - name: Prepare Release Notes
+      shell: pwsh
+      run: |
+        # Release notes are just the tag body on Windows
+        cd repo
```
RytoEX, line 86:
```suggestion
        Ensure-Location repo
```
Perhaps?

.github/actions/bouf/action.yaml
```diff
@@ -0,0 +1,127 @@
...
+        git tag -l --format='%(contents:body)' ${{ inputs.version }} > "${{ github.workspace }}/notes.rst"
+    
+    - name: 'Authenticate to Google Cloud'
+      uses: 'google-github-actions/auth@v1'
+      with:
+        workload_identity_provider: ${{ inputs.gcpWorkloadIdentityProvider }}
+        service_account: ${{ inputs.serviceAccountName }}
+
+    - name: Download old builds
```
RytoEX, line 95:
```suggestion
    - name: Download Old Builds
```

.github/actions/bouf/action.yaml
```diff
@@ -0,0 +1,127 @@
...
+        $boufArgs = @(
+           "--config",     "$env:GITHUB_ACTION_PATH/config.toml",
+           "--version",    "${{ inputs.version }}"
+           "--branch",     "${{ inputs.channel }}"
+           "--notes-file", "${{ github.workspace }}/notes.rst"
+           "-i",           "${{ github.workspace }}/build"
+           "-p",           "${{ github.workspace }}/old_builds"
+           "-o",           "${{ github.workspace }}/output"
+        )
+        &"${{ github.workspace }}\bouf\bin\bouf.exe" @boufArgs
```
RytoEX, line 117:
```suggestion
        Invoke-External "${{ github.workspace }}\bouf\bin\bouf.exe" @boufArgs
```

If not, then at least:
```
& "${{ github.workspace }}\bouf\bin\bouf.exe" @boufArgs
```

.github/actions/bouf/action.yaml
```diff
@@ -0,0 +1,127 @@
...
+
+    - name: Cleanup bouf output dir
```
RytoEX, line 119:
```suggestion
    - name: Cleanup bouf Output Directory
```

.github/actions/bouf/action.yaml
```diff
@@ -0,0 +1,127 @@
...
+      shell: pwsh
+      run: |
+        # Compress updater data so github's artifact upload won't take forever
```
RytoEX, line 122:
```suggestion
        # Compress updater data so GitHub's artifact upload won't take forever
```

RytoEX (changes requested):

.github/actions/bouf/Ensure-Location.ps1
```diff
@@ -0,0 +1,29 @@
+function Ensure-Location {
```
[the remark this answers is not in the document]
PatTheMav: Correct - in theory all the utility functions could be put into a single imported PowerShell script used just by the action (instead of the separate files used by the build script), but that's more a matter of style/taste.

.github/actions/bouf/action.yaml
PatTheMav: I think it's matching files in the `build` subdirectory.

RytoEX: > I think it's matching files in the `build` subdirectory.

Oh, that makes more sense.

.github/actions/bouf/Ensure-Location.ps1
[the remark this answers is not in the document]
RytoEX: > Correct - in theory all the utility functions could be put into a single imported PowerShell script used just by the action (instead of the separate files used by the build script), but that's more a matter of style/taste.

Maybe some day it might make sense to turn the utility scripts into modules and install them as needed. Not a today problem.

If this particular method is okay to you, then I'll not block on it. We'll just need to be cognizant that there are multiple copies floating around if we have any changes to make.

.github/actions/bouf/action.yaml
derrod: Winget isn't available on runner as far as I could tell

.github/actions/bouf/action.yaml
derrod: Indeed.

.github/actions/bouf/action.yaml
derrod: If repo doesn't exist this should fail rather than creating it

.github/actions/bouf/action.yaml
derrod: Yep, remember Windows supports both `/` and `\` but `\` is technically correct (and sometimes built-in stuff won't work correctly with `/`)

.github/actions/bouf/Ensure-Location.ps1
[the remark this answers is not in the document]
derrod: I suppose the files here could be symlinks rather than direct copies, but not sure how GitHub would handle that for actions.

.github/actions/bouf/action.yaml
RytoEX: Ah, fair point.

.github/actions/bouf/action.yaml
RytoEX: Fair enough.

.github/actions/bouf/action.yaml
PatTheMav: That'd be `Set-Location` then.

derrod: `cd` is an alias for `Set-Location anyway`:
```pwsh
PS C:\Dev\obs\obs-studio> Set-Location asd
Set-Location: Cannot find path 'C:\Dev\obs\obs-studio\asd' because it does not exist.
PS C:\Dev\obs\obs-studio> cd asd
Set-Location: Cannot find path 'C:\Dev\obs\obs-studio\asd' because it does not exist.
```

PatTheMav: Indeed, but at least for use in scripts I prefer us to use the "long form" of PowerShell commands, at the very least for consistency with our other PowerShell code.

.github/actions/bouf/Ensure-Location.ps1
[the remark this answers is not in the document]
RytoEX: It's probably fine as-is for now. Just need to remember to port any fixes.

derrod:
Updated the actions a bit:
- All the files in the `actions/bouf/` are now hash-checked before running the action
- Pinned specific versions for pandoc and rclone
- Pinned specific commits for third-party actions (gcloud auth + sdk setup)
- Switched to using GCS (Google Cloud Storage) for old builds
- Signing the updater manifest is now also done via Cloud KMS using the `gcloud` CLI tool
  + This didn't seem to work via `Invoke-External` as it passes the arguments in an incompatible way, haven't investigated further but shouldn't matter, this failing isn't critical either

.github/actions/bouf/action.yaml
```diff
@@ -0,0 +1,137 @@
...
+        BOUF_NSIS_HASH: 'bb0561a76cecaba62785642dd71052a9fe53458e45df108ca08867ee333b80bd'
+        GH_TOKEN: ${{ github.token }}
+      run: |
+        # Download bouf release
+        . $env:GITHUB_ACTION_PATH\Ensure-Location.ps1
+        . $env:GITHUB_ACTION_PATH\Invoke-External.ps1
+        Ensure-Location bouf
+        Invoke-External gh release download "$env:BOUF_TAG" -R "obsproject/bouf" -p "*windows*.zip" -p "*nsis*.zip"
+        
+        if ((Get-FileHash *windows*.zip -Algorithm SHA256).Hash -ne "$env:BOUF_HASH") {
```
PatTheMav, line 41: This will work, though unexpectedly - the `gh` glob pattern would potentially download multiple files, and `Get-FileHash <pattern>).Hash` would yield multiple hashes (one per line) then. The `-ne` expression in turn yields all the hashes that do not match the comparator (so not a `$true` or `$false` boolean), but the string will be interpreted as `$true` so the intended effect is the same.

Too broad star expressions in a glob are icky, but should be fine for now.

derrod: Considering that this is one of our repos (well, mostly mine) I don't expect there to ever be more than one artifact per pattern as there's only one windows build or NSIS zip.

.github/workflows/push.yaml
```diff
@@ -194,6 +194,71 @@ jobs:
...
+          fetch-depth: 0
+          ref: ${{ github.ref }}
+
+      - name: Set Up Environment 🔧
+        id: setup
+        env:
+          BOUF_ACTION_HASH: '63933179a2cc636cb2bca2ae4ac2ae30af71bf29cc2a71f6ab01600a6f586bfe'
+        run: |
+          $shortHash = $env:GITHUB_SHA.Substring(0,9)
+          $channel = 'stable'
```
PatTheMav, line 223: Use PowerShell conditional expression:

```
$output = if ($env:GITHUB_REF_NAME -match "(beta|rc)") { "beta" } else { "stable" }
```

It's similar to pattern matching expressions.

.github/actions/bouf/action.yaml
PatTheMav: Correct, which is why I'm not concerned. I personally just prefer code to be "hardened" logically to not rely on assumptions. Doesn't require to be changed though.

derrod: I just updated it now to use the same pattern as the creation of those zips.

.github/workflows/push.yaml
```diff
@@ -194,6 +194,68 @@ jobs:
...
+          $folderHash = ''
+          $files = Get-ChildItem "${{ github.workspace }}\repo\.github\actions\bouf"
+          foreach ($file in $files) {
+            $folderHash += (Get-FileHash $file -Algorithm SHA256).Hash
+          }
+          # This is stupid but so is powershell
+          $stream = [IO.MemoryStream]::new([byte[]][char[]]$folderHash)
+          if ((Get-FileHash -InputStream $stream -Algorithm SHA256).Hash -ne "$env:BOUF_ACTION_HASH") {
+            throw "bouf action folder hash does not match."
+          }
```
RytoEX, line 237: Could you walk me through the intent here?

derrod: Ensure that all files in the `.github/actions/bouf` folder match an expected hash to verify they have not been modified. For that purpose they are hashed, those hashes are concatenated, and the hash of the concatenated hash is verified against a known one. That's a little simpler than maintaining a file or environment variables with hashes for each file individually.

RytoEX: Ah, so it's the "hash of the concatenated hash" that is why you needed a MemoryStream. Okay.

derrod: Unfortunately, despite https://github.com/PowerShell/PowerShell-RFC/issues/65 being merged over 6 years ago PowerShell still has no native means of hashing a string.

.github/workflows/push.yaml
```diff
@@ -194,6 +194,68 @@ jobs:
           name: macos-sparkle-update
           path: ${{ github.workspace }}/output
 
+  create-windows-update:
+    name: Create Windows Update 🥩
```
PatTheMav, line 198: Heh.

.github/actions/bouf/action.yaml
```diff
@@ -0,0 +1,139 @@
...
+        BOUF_TAG: 'v0.6.0-nsis-update'
+        BOUF_HASH: '892640b3322046d28f0d2ef9dd24d3ed45499dca39852c6ce70cba44269ca2d0'
+        BOUF_NSIS_HASH: '790edd61161b76b74c71fe68892bcbf4357e0d83ee67f3bbfd95af0e4d609515'
+        GH_TOKEN: ${{ github.token }}
+      run: |
+        # Download bouf release
+        . $env:GITHUB_ACTION_PATH\Ensure-Location.ps1
+        . $env:GITHUB_ACTION_PATH\Invoke-External.ps1
+        Ensure-Location bouf
+        $windows_zip = "bouf-windows-$env:BOUF_TAG.zip"
```
PatTheMav, line 39: We usually enclose variables for interpolation with braces (similar to Bash) to ensure the variable name is clearly scoped (e.g. `${env:BOUF_TAG}` or `$($Object.Method())`).

Outside of interpolation there is usually no need for enclosing (or even quoting in some cases).

PatTheMav:
Overall seems fine to me apart from PowerShell code style nits.

RytoEX (approved):
Seems fine.

derrod:
Updated to use `actions/upload-artifact@v4` instead of `v3` which massively increases the speed of uploads. Also the outputs are now separated in a logical manner (installer, signed ZIP, updater files, PDBs) rather than all in one.

RytoEX (approved):
Seems fine again to me.

.github/actions/bouf/action.yaml
```diff
@@ -0,0 +1,133 @@
...
+        BOUF_NSIS_HASH: '2f5ecff05a002913c10aafa838febc1b0ae6e779f5ca67efa545ed787ae485a0'
+        GH_TOKEN: ${{ github.token }}
+      run: |
+        # Download bouf release
+        . ${env:GITHUB_ACTION_PATH}\Ensure-Location.ps1
+        . ${env:GITHUB_ACTION_PATH}\Invoke-External.ps1
+        Ensure-Location bouf
+        $windows_zip = "bouf-windows-${env:BOUF_TAG}.zip"
+        $nsis_zip = "bouf-nsis-${env:BOUF_TAG}.zip"
+        Invoke-External gh release download "$env:BOUF_TAG" -R "obsproject/bouf" -p $windows_zip -p $nsis_zip
```
PatTheMav, line 44:
```suggestion
        Invoke-External gh release download "${env:BOUF_TAG}" -R "obsproject/bouf" -p $windows_zip -p $nsis_zip
```

.github/actions/bouf/action.yaml
```diff
@@ -0,0 +1,133 @@
...
+        
+        if ((Get-FileHash $windows_zip -Algorithm SHA256).Hash -ne "$env:BOUF_HASH") {
```
PatTheMav, line 46:
```suggestion
        if ((Get-FileHash $windows_zip -Algorithm SHA256).Hash -ne "${env:BOUF_HASH}") {
```

.github/actions/bouf/action.yaml
```diff
@@ -0,0 +1,133 @@
...
+          throw "bouf hash does not match."
+        }
+        if ((Get-FileHash $nsis_zip -Algorithm SHA256).Hash -ne "$env:BOUF_NSIS_HASH") {
```
PatTheMav, line 49:
```suggestion
        if ((Get-FileHash $nsis_zip -Algorithm SHA256).Hash -ne "${env:BOUF_NSIS_HASH}") {
```

.github/actions/bouf/action.yaml
```diff
@@ -0,0 +1,133 @@
...
+      env:
+        CNG_TAG: 'cng-v1.0'
+        GH_TOKEN: ${{ github.token }}
+      run: |
+        # Download Google CNG provider release from github
+        . ${env:GITHUB_ACTION_PATH}\Ensure-Location.ps1
+        . ${env:GITHUB_ACTION_PATH}\Invoke-External.ps1
+        Ensure-Location gcng
+        
+        Invoke-External gh release download "$env:CNG_TAG" -R "GoogleCloudPlatform/kms-integrations" -p "*amd64.zip"
```
PatTheMav, line 67:
```suggestion
        Invoke-External gh release download "${env:CNG_TAG}" -R "GoogleCloudPlatform/kms-integrations" -p "*amd64.zip"
```

PatTheMav (approved):

PR #9831, opened 2023-11-09
libobs: Add #line preprocessor directive support

exeldro:
### Description
Add #line preprocessor directive support

### Motivation and Context
Allows obs-shaderfilter to report correct line for the error message

### How Has This Been Tested?
with obs-shaderfilter on windows 11 by having `#line 1 "test"` in a shader

### Types of changes

### Checklist:


PatTheMav:
Could you update the PR title and commit message to make it clear that this is related to the shader parser and not the preprocessor used to compile OBS itself?

PatTheMav:
Do you have an example shader implementing this preprocessor directive?

exeldro:
the example shown in "how has this been tested", `#line 1 "test"` can be added to any shader in OBS

Because of the transpilation the line numbers of the errors are always wrong, this change allows for a fix for that.
more explanation about #line can be found here:
https://learn.microsoft.com/en-us/cpp/preprocessor/hash-line-directive-c-cpp?view=msvc-170

PatTheMav:
So is the expectation that the preprocessor code makes it verbatim into the actual HLSL/MSL/GLSL shader code?

exeldro:
Retested it and is not working as expected, closing this till fixed.

PR #9845, opened 2023-11-12
linux-pipewire: Ensure n_drm_formats is zero initialized

### LINKED ISSUE #9844: Pipewire Screen Capture causes OBS to crash/hang
### Operating System Info

Ubuntu 22.04

### Other OS

_No response_

### OBS Studio Version

30.0.0

_No response_

### OBS Studio Log URL

https://obsproject.com/logs/6AvQB8pOWSB7kiQY

### OBS Studio Crash Log URL

_No response_

### Expected Behavior

Does not crash on clicking the "Share" button

### Current Behavior

Crashes or hangs on clicking the "Share" button

### Steps to Reproduce

1. create screen capture or window capture pipewire and attempt to share
2. instant hang/crash on clicking "Share" in pipewire popup window

There is no crash/hang in 29.1.3 with pipewire screen capture

kkartaltepe:
### Description

We use a size_t that can be 64bit while EGL uses 32bit for the
underlying value on some platforms. Writes from our graphics function
can leave high bits uninitialized and cause iteration of the format list
to overlfow in those cases.

### Motivation and Context

fixes #9844

This short write seems uncommon,  maybe this is just due to legacy nvidia drivers on the device in issue.

### How Has This Been Tested?

OBS still works on my machine.

### Types of changes

### Checklist:

### DIFF
--- a/plugins/linux-pipewire/pipewire.c
+++ b/plugins/linux-pipewire/pipewire.c
@@ -522,7 +522,7 @@ static void init_format_info_sync(obs_pipewire_stream *obs_pw_stream)
 
 	enum gs_dmabuf_flags dmabuf_flags;
 	uint32_t *drm_formats = NULL;
-	size_t n_drm_formats;
+	size_t n_drm_formats = 0;
 
 	bool capabilities_queried = gs_query_dmabuf_capabilities(
 		&dmabuf_flags, &drm_formats, &n_drm_formats);


GeorgesStavracas (approved):

dottools:
This is just a band aid fix for an underlining problem. The EGL interface code should be the one fixing this, not this high level.

The problem lies in file `libobs-opengl/gl-egl-common.c`:
```cpp
bool gl_egl_query_dmabuf_capabilities(EGLDisplay egl_display,
				      enum gs_dmabuf_flags *dmabuf_flags,
				      uint32_t **formats, size_t *n_formats)
{
	bool ret = false;

	if (is_implicit_dmabuf_modifiers_supported()) {
		*dmabuf_flags = GS_DMABUF_FLAG_IMPLICIT_MODIFIERS_SUPPORTED;
		ret = true;
	}

	if (!glad_eglQueryDmaBufFormatsEXT) {
		blog(LOG_ERROR, "Unable to load eglQueryDmaBufFormatsEXT");
		return ret;
	}

	if (!query_dmabuf_formats(egl_display, (EGLint **)formats,
				  (EGLint *)n_formats)) {
		*n_formats = 0;
		*formats = NULL;
	}
	return ret;
}
```

The casting `size_t *` to `EGLint *` is undefined behavior.

Function should be changed to:
```cpp
bool gl_egl_query_dmabuf_capabilities(EGLDisplay egl_display,
				      enum gs_dmabuf_flags *dmabuf_flags,
				      uint32_t **formats, size_t *n_formats)
{
	EGLint num_formats;
	bool ret = false;

	if (is_implicit_dmabuf_modifiers_supported()) {
		*dmabuf_flags = GS_DMABUF_FLAG_IMPLICIT_MODIFIERS_SUPPORTED;
		ret = true;
	}

	if (!glad_eglQueryDmaBufFormatsEXT) {
		blog(LOG_ERROR, "Unable to load eglQueryDmaBufFormatsEXT");
		return ret;
	}

	if (!query_dmabuf_formats(egl_display, (EGLint **)formats,
				  &num_formats)) {
		num_formats = 0;
		*formats = NULL;
	}

	*n_formats = num_formats;

	return ret;
}
```


However still can be a problem if EGL handles are built with 64bit integers instead of 32bit.
Which means then casting `uint32_t **` to `EGLint **` is undefined behavior. Per `EGLint` declaration comment in `eglplatform.h`:
```c
/* Define EGLint. This must be a signed integral type large enough to contain
 * all legal attribute names and values passed into and out of EGL, whether
 * their type is boolean, bitmask, enumerant (symbolic constant), integer,
 * handle, or other.  While in general a 32-bit integer will suffice, if
 * handles are 64 bit types, then EGLint should be defined as a signed 64-bit
 * integer type.
 */
typedef khronos_int32_t EGLint;
```

kkartaltepe:
> This is just a band aid fix for an underlining problem. The EGL interface code should be the one fixing this, not this high level.

This is true but not something i want to deal with in a hotfix right now. So this is the fix for now.

PR #9857, opened 2023-11-14
obs-filters: Remove LTCG linker flag for SpeexDSP builds on Windows

PatTheMav:
### Description
Remove `/LTCG` linker flag for SpeexDSP implementation on Windows.

### Motivation and Context
Flag was added in CMake 2.0 update and is incompatible with incremental linking enabled by default for Debug configuration (MSVC default) in recent compiler flag update.

The original error that necessitates adding LTCG is not present with current obs-deps anymore, so remove it and disable the hard-coded default library directive instead.

### How Has This Been Tested?
Built OBS Studio on Windows 11 with Debug configuration on current master.

### Types of changes

### Checklist:

### DIFF
--- a/plugins/obs-filters/cmake/speexdsp.cmake
+++ b/plugins/obs-filters/cmake/speexdsp.cmake
@@ -6,8 +6,7 @@ if(ENABLE_SPEEXDSP)
   target_sources(obs-filters PRIVATE noise-suppress-filter.c)
   target_link_libraries(obs-filters PRIVATE SpeexDSP::Libspeexdsp)
   target_compile_definitions(obs-filters PRIVATE LIBSPEEXDSP_ENABLED)
-  target_link_options(obs-filters PRIVATE $<$<PLATFORM_ID:Windows>:/LTCG> $<$<PLATFORM_ID:Windows>:/IGNORE:4098>
-                      $<$<PLATFORM_ID:Windows>:/IGNORE:4099>)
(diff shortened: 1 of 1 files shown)


plugins/obs-filters/cmake/speexdsp.cmake
```diff
@@ -6,8 +6,7 @@ if(ENABLE_SPEEXDSP)
   target_sources(obs-filters PRIVATE noise-suppress-filter.c)
   target_link_libraries(obs-filters PRIVATE SpeexDSP::Libspeexdsp)
   target_compile_definitions(obs-filters PRIVATE LIBSPEEXDSP_ENABLED)
-  target_link_options(obs-filters PRIVATE $<$<PLATFORM_ID:Windows>:/LTCG> $<$<PLATFORM_ID:Windows>:/IGNORE:4098>
-                      $<$<PLATFORM_ID:Windows>:/IGNORE:4099>)
+  target_link_options(obs-filters PRIVATE $<$<PLATFORM_ID:Windows>:/NODEFAULTLIB:MSVCRT>)
```
derrod, line 9: This fails building right now locally and on CI, if I replace it with `msvcrtd.lib` as in `win-capture` it works: https://github.com/obsproject/obs-studio/blob/3e6797ca5bfe445381d721e407163b70b8de6d1e/plugins/win-capture/CMakeLists.txt#L80

PatTheMav: That kinda makes sense, I'm just weirded out that it didn't break in debug config for me locally.

derrod (approved):
Let's just get this done so master compiled in `Debug` again, we can think about better solutions for this whole `NODEFAULTLIB` thing later.

PR #9858, opened 2023-11-14
Enable VS default compiler optimisations

derrod:
### Description

- Enables `/GL` + `/LTCG` on non-Debug builds (enabled by default in new VS projects)
- Enables `/Oi` on non-Debug builds (enabled by default in new VS projects)
- Adds libobs change to crash if `channels` is `0`, this check is required now due to additional optimisations in `Release` configurations resulting in a divide-by-zero compiler error

Will require #9857 to compile in Debug.

### Motivation and Context

Align with default VS settings and maybe make things a tiny bit faster?

### How Has This Been Tested?

Compiled in Debug, Release, and RelWithDebInfo.

### Types of changes

### Checklist:


notr1ch (approved):
The null check removal should probably be a separate PR but I don't particularly care as long as it's a separate commit.

derrod:
I kept it in here because otherwise compiling in `Release` fails, presmably due to compiler optimisations inlining those functions.

PatTheMav (approved):

RytoEX (approved):
Seems fine. Might want to explore letting CMake handle this via INTERPROCEDURAL_OPTIMIZATION in a future revision, but this is fine for now.

derrod:
> Seems fine. Might want to explore letting CMake handle this via INTERPROCEDURAL_OPTIMIZATION in a future revision, but this is fine for now.

That used `/LTCG:INCREMENTAL` which results in lots of warnings every time you do a rebuild or fresh compile. Plus we want to specifically move away from CMake presets because they are out of date.

PatTheMav:
> > Seems fine. Might want to explore letting CMake handle this via INTERPROCEDURAL_OPTIMIZATION in a future revision, but this is fine for now.
> 
> That used `/LTCG:INCREMENTAL` which results in lots of warnings every time you do a rebuild or fresh compile. Plus we want to specifically move away from CMake presets because they are out of date.

Only for now - CMake is actively working on moving compile settings baked into defaults to MSVC-specific target property (see also the runtime library target property), so instead of passing flags to the options list, MSVC-specific options are enabled explicitly.

That might take a few versions to mature though.

PR #9875, opened 2023-11-16
Shader Cache Improvements

### LINKED ISSUE #9869: OBS Studio cannot start on version 30.0
### Operating System Info

Windows 11

### Other OS

_No response_

### OBS Studio Version

30.0.0

_No response_

### OBS Studio Log URL

https://obsproject.com/logs/c7j4ZKk3IqHHviLA

### OBS Studio Crash Log URL

_No response_

### Expected Behavior

OBS Studio 30.0 to start normally.

### Current Behavior

Shows Error "Failed to Initialize video. Your GPU May not be supported, or your graphics drivers may need to be updated".

### Steps to Reproduce

1. Updated to Version 30.0
2. Start OBS studio

- DDU and reverting the driver version back (Nvidia driver 537.58) or installing the latest driver (Nvidia driver 546.01) does not solve the issue.
- Reverting the OBS Studio version back to 29.1.3 fixes the problem, the log is as follows: https://obsproject.com/logs/OgnSR8t4AzpmlmFb or https://obsproject.com/logs/h40KBrQfYMIeoyK3

derrod:
### Description

- Adds checksum to cache files
    + New cache files are suffixed `.v2` to avoid log spam on first startup of a new version, as well as to avoid breaking older versions
- Updater now clears the cache after an update
   + NSIS script adjustments will follow if merged
- Logs if D3D compiler is outdated

Based on https://github.com/notr1ch/obs-studio/commit/4ecc62d257ac729ef3624ef6fec56a6d26537341

### Motivation and Context

- Should hopefully prevent issues like/fix #9869
- Shader cache should probably be cleared after updates to avoid having leftover files from old shader versions or changes to the shader compilation settings not being applied

### How Has This Been Tested?

- Booted up OBS and verified that invalid shader files are no longer loaded if the checksum doesn't match.
- Tested updater locally

### Types of changes

### Checklist:


libobs-d3d11/d3d11-subsystem.cpp
```diff
@@ -302,6 +302,10 @@ void gs_device::InitCompiler()
 				module, "D3DDisassemble");
 #endif
 			if (d3dCompile && d3dCreateBlob) {
+				if (ver < 47)
```
jpark37, line 305: I don't think this can be true because OBS needs Windows 10, and this is guaranteed in Windows 8.1.

jpark37: Although I think this d3dcompiler_47.dll that ships with Windows is technically for Store/UWP applications, and desktop applications are expected to include the redist DLL from the Windows SDK.

notr1ch: The idea is to catch potentially broken DLLs installed by other apps (there's a case where some broken installer overwrites the system32 D3DCompiler with an x86 version). Ideally it will never be true, but you never know what's out there.

jpark37: That seems like more of a reason to include the redist DLL.

libobs-d3d11/d3d11-subsystem.cpp
```diff
@@ -302,6 +302,10 @@ void gs_device::InitCompiler()
...
+					blog(LOG_WARNING,
+					     "InitCompiler: Using %s",
+					     d3dcompiler);
```
alinsavix: It might be worth it to log this unconditionally rather than only if the version is old, since if there's some version-specific bug in the future, having that information already logged would make it easier to identify. (assuming we don't just give up and bundle the compiler DLL)

libobs-d3d11/d3d11-subsystem.cpp
derrod: I can just remove this commit for nwo if it needs further discussion, just included it as it was part of the branch.

libobs-d3d11/d3d11-shader.cpp
```diff
@@ -224,29 +224,61 @@ void gs_shader::Compile(const char *shaderString, const char *file,
...
 	HRESULT hr;
 
+	bool is_cached = false;
 	char hashstr[20];
 
 	if (!shaderString)
 		throw "No shader string specified";
 
-	uint64_t hash = fnv1a_hash(shaderString);
+	uint64_t hash = fnv1a_hash(shaderString, strlen(shaderString));
```
notr1ch, line 233: We also use `strlen(shaderString)` in the non-cached path, so this should probably be stored in a variable.

Lain-B (approved):

Issue #9487, opened 2023-08-20
OBS fails to launch if Python partially removed with multiple installs


FireCulex:
### Operating System Info

Windows 11

### OBS Studio Version

29.1.3

### OBS Studio Log URL

https://obsproject.com/logs/TLO5Q6VOUjFXQ2fi

### Expected Behavior

For Obs to Launch 

### Current Behavior

Nothing

### Steps to Reproduce

1. Uninstall Python
2. Attempt to Launch OBS
3.
...

### Anything else we should know?

Eventually I figured out removing this line got OBS to launch again:
[Python]
Path64bit=C:/Users/Culex/AppData/Local/Programs/Python/Python310

from global.ini. I had forgotten I had removed 3.10 previously. Ideally a warning or popup should occur, or something in the log as opposed to simply nothing.

WizardCM:
So to confirm, it visually doesn't open? Or it opens and immediately closes?

Fenrirthviti:
Also, a log where OBS successfully loads would be useful as well here. It seems to fail during the load plugins stage, which is an odd place for python being missing to cause issues.

FireCulex:
> So to confirm, it visually doesn't open? Or it opens and immediately closes?

Nothing is shown on the screen. The log is generated but nothing useful. No crash. 
The easiest way to try see this for yourself is to add
```
[Python]
Path64bit=C:/Users/Culex/AppData/Local/Programs/Python/Python310
```
to %appdata%\obs-studio\global.ini

FireCulex:
I apologize it appears I was mistaken in my assessment. I added `%appdata\Local2` and it loaded perfectly fine. Then I added `C:/Users/Culex/AppData/Local/Programs/Python/Python310` and OBS locked up. Attempting to terminate it in task manager resulted in a "Access is Denied" I had to reboot. 

Upon further inspection Python310 was only partially uninstalled. python.exe, pythonw.exe,python3.dll, python310.dll were removed, however etc Lib/site-packages Scripts/jupyter related and share are still present.

I loaded up windbg and tried it again.
```
ModLoad: 00007ffa`78730000 00007ffa`78b5c000   C:\Users\Culex\AppData\Local\Programs\Python\Python38\python38.dll
Fatal Python error: failed to get the Python codec of the filesystem encoding
```
[...]

PR #9884, opened 2023-11-17
obs-scripting: Fix crash when loading Python at runtime and build errors in Debug configuration on Windows

PatTheMav:
### Description
Fixes a possible crash when loading Python at runtime on Windows and downgrades Python version support to version 3.10 on Windows.

### Motivation and Context
Existing code attempts to load the most recent Python library available in a user-specified path. Unfortunately while iterating over possible DLL file names, the string modification code also possibly changes the subdirectory name of the user-provided directory.

The PR changes this code to only modify the DLL filename and leave the provided directory intact.

Fixes https://github.com/obsproject/obs-studio/issues/9487.

This PR also lowers the upper version boundary on Windows to Python 3.10 again. Python 3.11 has replaced many macros with static inline functions which now result in unresolved link-time symbols, particularly in Debug configuration where no unused symbols are removed.

macOS is not affected by this, as it can link libraries with a dynamic runtime lookup, Windows does not have that capability.

Until this issue can be fixed in the way the Python symbols are declared for dynamic import, downgrading will re-enable ability to develop OBS Studio in Debug mode until an working fix for this is found.

### How Has This Been Tested?
Compiled OBS Studio with Python 3.11 on Windows 11 and confirmed that CMake will not generate a project. With a Python version older than Python 3.11, the project is configured and Debug configuration is correctly built.

### Types of changes

### Checklist:


Lain-B (approved):
~~Feel free to merge if you're feeling good with it, looks good to me~~ er, seems ryan self-assigned. But still looks pretty trivial

deps/obs-scripting/obs-scripting-python-import.c
```diff
@@ -116,6 +118,7 @@ bool import_python(const char *python_path, python_version_t *python_version)
 
 	IMPORT_FUNC(PyType_Ready);
 	IMPORT_FUNC(PyType_Modified);
+	IMPORT_FUNC(PyType_IsSubtype);
```
RytoEX, line 121: Is this necessary for this bugfix?

RytoEX (changes requested):

deps/obs-scripting/obs-scripting-python-import.c
PatTheMav: It could lead to crashes at runtime if we don't import the actual function, as the aliased function pointer will not point to a valid memory address.

RytoEX: Fair enough.

RytoEX (approved):
Can confirm that this fixes the crash in the linked Issue.

It is worth noting that with this change, if the Python DLL is not found in the explicit path the user has specified, OBS will no longer iterate through all Python installs if they are installed to versioned directories within the same parent directory, such as:
```
%LocalAppData%\Programs\Python\Python38
%LocalAppData%\Programs\Python\Python310
%LocalAppData%\Programs\Python\Python311
```

But I think losing that magic fallback is a fair trade-off for not crashing.
cc @WizardCM

PatTheMav:
@WizardCM any objections to this fix?

PR #9892, opened 2023-11-20
libobs/util, libobs/media-io: Fix link error including header files from C++

norihiro:
### Description

This PR adds `extern "C" {}` to the header files under `libobs/util/` that contain extern functions.

### Motivation and Context

I realized it causes a link error when I include a header file `util/crc32.h` and use the function from a C++ source code.

### How Has This Been Tested?

OS: Fedora 37

In a C++ source code for an external building plugin, included `util/crc32.h` and called `calc_crc32` from a function, and build it. Then, confirmed the linker shows no error.

CI will test this won't break existing code.

### Types of changes

### Checklist:

### DIFF
--- a/libobs/media-io/video-frame.h
+++ b/libobs/media-io/video-frame.h
@@ -20,6 +20,10 @@
 #include "../util/bmem.h"
 #include "video-io.h"
 
+#ifdef __cplusplus
+extern "C" {
+#endif
+
 struct video_frame {
 	uint8_t *data[MAX_AV_PLANES];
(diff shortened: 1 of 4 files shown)


PatTheMav (approved):
We should probably add this to all `libobs` headers, but it's fine to add them one by one until then as well.

norihiro:
There is just one more file missing `extern "C"` under libobs is `libobs/media-io/video-frame.h`. I will add it.

Lain-B (approved):

PR #9813, opened 2023-11-06
frontend-tools: Fix multiple signal-slot connections on output timer

matt818:
### Description
This will fix the output timer bug from issue #6558. If you use the output timer too many times in a run of the program, issues start happening where on every other run of the output timer, the video will stop prematurely and generate a tiny video recording file. 

Every time the start button is pressed it makes a new connection to the QTimers, increasing the amount of connections and signals being sent to the QTimers by 1 each time the start button is clicked.

This fix moves the connections to the QTimers to the function that creates them, so that the connection is only established once.

### Motivation and Context
Fixes #6558.

### How Has This Been Tested?
I used the output timer to record multiple 5 seconds videos consecutively with the issue not happening.

### Types of changes
Bug fix (non-breaking change which fixes an issue)

### Checklist:

- [x] My code has been run through [clang-format](https://github.com/obsproject/obs-studio/blob/master/.clang-format).
- [x] I have read the [**contributing** document](https://github.com/obsproject/obs-studio/blob/master/CONTRIBUTING.rst).
- [ ] My code is not on the master branch.
- [x] The code has been tested.
- [x] All commit messages are properly formatted and commits squashed where appropriate.
- [x] I have included updates to all appropriate documentation.

Issue #6558, opened 2022-05-29
Output timer stops video prematurely


weichensw:
### Operating System Info

Windows 11

### OBS Studio Version

27.2.4

### OBS Studio Log URL

https://obsproject.com/forum/threads/output-timer-stops-video-prematurely.118949/

### Expected Behavior

When using output timer, the video should be always stopped as expected

### Current Behavior

When using the output timer functionality, the recording will sometimes stop immediately, producing a 1kb file. The UI however shows that it is still recording. Normally when it is recording the file size is 0kb and will show the real size when the recording is complete. If I don't notice the 1kb file immediately, it will just keep running without throwing any errors. I for sure have the correct timer settings.

### Steps to Reproduce

When using the output timer functionality, the recording will sometimes stop immediately, producing a 1kb file. The UI however shows that it is still recording. Normally when it is recording the file size is 0kb and will show the real size when the recording is complete. If I don't notice the 1kb file immediately, it will just keep running without throwing any errors. I for sure have the correct timer settings.

### Anything else we should know?

The bug has been reproing since 25.0.4, see here: https://obsproject.com/forum/threads/output-timer-stops-video-prematurely.118949/

Fenrirthviti:
Please include a full log file instead of a link to a forum post with only a snippet.

weichensw:
I have a set of recording needs to be done asap so I reverted back to 24.0.4. I will go back to latest version after done. I might not be able to get a full log until then but I can confirm 24.0.4 doesn't have this problem.

AntoineGR:
I can reproduce and confirm this bug 100%

My Timer is set to stop recording after 4 seconds.
When opening OBS I can make a few recordings (2-4) without any issue,
but then it happens : recording stops instantly, creating a corrupt 4kb file.

**After that initial bug, every other recording stops prematurely = 50% of corrupts recordings !**
After each corrupt file the log shows an increasing number of "Stopping recording due to OutputTimer timeout" message.

Here attached is a full log showing : 
- 4 good recordings
- a first corrupt recording (@19:25:49.299)
- a series of 6 recordings alternating good/corrupt
[2022-11-16 19-24-35.txt](https://github.com/obsproject/obs-studio/files/10024791/2022-11-16.19-24-35.txt)

DevSejin:
windows 10, 28.1.2 64bit
still problem...

guidoising:
[...]

PR #9900, opened 2023-11-20
obs-ffmpeg: Initialize stopping member variable to false

RytoEX:
### Description

If an output has already stopped, but its StopRecording function was called again, then ffmpeg_mux_stop would be called and set stopping to true. On the next output start, OBS would output 1 frame, see that stopping is true, and then stop the output.

This was most easily observed using an Output Timer to record prior to 93f5b45be8c8b91199c52cf7a0ab59d8c3a08760.

Initialize stopping to false with the other state flags to ensure that the output has a clean starting state.

### Motivation and Context

Discovered while reviewing #9813 and investigating #6558.

Prior to 93f5b45be8c8b91199c52cf7a0ab59d8c3a08760 (#9813), you could run into this by following these Steps To Reproduce:
1. Set an Output Timer for recording. Use something like 5-10 seconds (short enough to make repeatedly running easy, long enough to distinguish it from a recording that immediately stopped).
2. Start recording.
3. Wait for the Output Timer to stop the recording.
4. Repeat Steps 2 and 3 several times until you get a recording that is only 1 frame.

You'll end up with a log snippet like this:
```
14:16:04.230: Starting recording due to OutputTimer
[...]
14:16:04.296: ==== Recording Start ===============================================
14:16:04.297: [ffmpeg muxer: 'adv_file_output'] Writing file 'C:/Videos/2023-11-20 14-16-04.mp4'...
14:16:05.489: [ffmpeg muxer: 'adv_file_output'] Output of file 'C:/Videos/2023-11-20 14-16-04.mp4' stopped
14:16:05.490: Output 'adv_file_output': stopping
14:16:05.490: Output 'adv_file_output': Total frames output: 1
14:16:05.490: Output 'adv_file_output': Total drawn frames: 36
14:16:05.503: ==== Recording Stop ================================================
14:16:14.282: Stopping recording due to OutputTimer timeout
14:16:14.282: Stopping recording due to OutputTimer timeout
14:16:14.283: Stopping recording due to OutputTimer timeout
14:16:14.283: Stopping recording due to OutputTimer timeout
14:16:14.283: Stopping recording due to OutputTimer timeout
```
See the key line:
```
14:16:05.490: Output 'adv_file_output': Total frames output: 1
```

If you look further up, the previous recording session should look something like this:
```
14:15:58.540: Stopping recording due to OutputTimer timeout
14:15:58.550: Stopping recording due to OutputTimer timeout
14:15:58.559: [ffmpeg muxer: 'adv_file_output'] Output of file 'C:/Videos/2023-11-20 14-15-48.mp4' stopped
14:15:58.559: Output 'adv_file_output': stopping
14:15:58.559: Output 'adv_file_output': Total frames output: 264
14:15:58.559: Output 'adv_file_output': Total drawn frames: 300
14:15:58.566: Stopping recording due to OutputTimer timeout
14:15:58.579: Stopping recording due to OutputTimer timeout
14:15:58.590: ==== Recording Stop ================================================
```
The key here is the "Stopping recording due to OutputTimer timeout" after "Output 'adv_file_output': stopping".

What seems to occur is that the output is stopped, but then the Output Timer calls [`OutputTimer::EventStopRecording()`](https://github.com/obsproject/obs-studio/blob/cfd8d4c993ba2b6d61541b66a8d08ed436de824d/UI/frontend-plugins/frontend-tools/output-timer.cpp#L245), which then calls `obs_frontend_recording_stop()`, which invokes `StopRecording`.
https://github.com/obsproject/obs-studio/blob/cfd8d4c993ba2b6d61541b66a8d08ed436de824d/UI/api-interface.cpp#L302-L305

Since the output has already actually stopped, the `stopping` member variable should be `false`. However, the calls to stop the output set `stopping` to `true`, and this is not cleared by the output's stopping process because the output has already stopped.
https://github.com/obsproject/obs-studio/blob/cfd8d4c993ba2b6d61541b66a8d08ed436de824d/plugins/obs-ffmpeg/obs-ffmpeg-mux.c#L535-L544

This may also fix a number of other "output recording does not stop" or "can not record after a while" issues that were actually caused by a bad output stop or teardown process.

### How Has This Been Tested?

Tested locally on Windows 11. Despite seeing "Stopping recording due to OutputTimer timeout" after "Output 'adv_file_output': stopping", subsequent OutputTimer invoked recording sessions do not stop early.

### Types of changes

### Checklist:

### DIFF
--- a/plugins/obs-ffmpeg/obs-ffmpeg-mux.c
+++ b/plugins/obs-ffmpeg/obs-ffmpeg-mux.c
@@ -465,6 +465,7 @@ static inline bool ffmpeg_mux_start_internal(struct ffmpeg_muxer *stream,
 	/* write headers and start capture */
 	os_atomic_set_bool(&stream->active, true);
 	os_atomic_set_bool(&stream->capturing, true);
+	os_atomic_set_bool(&stream->stopping, false);
 	stream->total_bytes = 0;
 	obs_output_begin_data_capture(stream->output, 0);
 


PatTheMav (approved):

RytoEX:
Additionally, while this is not initialized here, it is initialized in other outputs:
https://github.com/obsproject/obs-studio/blob/cfd8d4c993ba2b6d61541b66a8d08ed436de824d/plugins/obs-ffmpeg/obs-ffmpeg-output.c#L1213
https://github.com/obsproject/obs-studio/blob/cfd8d4c993ba2b6d61541b66a8d08ed436de824d/plugins/obs-ffmpeg/obs-ffmpeg-mpegts.c#L1060
https://github.com/obsproject/obs-studio/blob/cfd8d4c993ba2b6d61541b66a8d08ed436de824d/plugins/obs-outputs/flv-output.c#L165


But not in `replay_buffer_start` or `obs-ffmpeg-hls-mux`:
https://github.com/obsproject/obs-studio/blob/cfd8d4c993ba2b6d61541b66a8d08ed436de824d/plugins/obs-ffmpeg/obs-ffmpeg-mux.c#L1050-L1051
https://github.com/obsproject/obs-studio/blob/cfd8d4c993ba2b6d61541b66a8d08ed436de824d/plugins/obs-ffmpeg/obs-ffmpeg-hls-mux.c#L166-L167

notr1ch:
Shouldn't "stop" be a no-op if the output was already requested to stop? I'm not sure this fix is in the correct place - stopping should already be set to false in `deactivate`.

```14:16:14.282: Stopping recording due to OutputTimer timeout
14:16:14.282: Stopping recording due to OutputTimer timeout
14:16:14.283: Stopping recording due to OutputTimer timeout
14:16:14.283: Stopping recording due to OutputTimer timeout
14:16:14.283: Stopping recording due to OutputTimer timeout```
To me this seems more like a bug with the OutputTimer firing multiple stop events.

RytoEX:
> ```
> 14:16:14.282: Stopping recording due to OutputTimer timeout
> 14:16:14.282: Stopping recording due to OutputTimer timeout
> 14:16:14.283: Stopping recording due to OutputTimer timeout
> 14:16:14.283: Stopping recording due to OutputTimer timeout
> 14:16:14.283: Stopping recording due to OutputTimer timeout
> ```
> 
> To me this seems more like a bug with the OutputTimer firing multiple stop events.

There are two parts to this. One was that OutputTimer was firing multiple stop events. That has been fixed in #9813. However, as noted by @gxalpha on that PR:
> This change does make sense and is definitely needed, but I don't quite understand how it fixes the issue. As is it should "only" call `obs_frontend_recording_stop` multiple times which in theory shouldn't cause any issues.

The second part is that it seems that if something calls an output's `stop` _after_ the output has already been stopped, the `stopping` member variable gets set to `true` regardless, at least in the case of ffmpeg_hls_muxer and ffmpeg_muxer.

ffmpeg_hls_muxer and ffmpeg_muxer both use `ffmpeg_mux_stop`, which uses `capturing(stream) || ts == 0` as its check for if it should attempt to stop an output. ffmpeg_mpegts_muxer and ffmpeg_output use `output->active`. flv_output seems to just stop an output without any such checks, but as previously mentioned, it initalizes `stopping` to `true`, as do ffmpeg_output and ffmpeg_mpegts_muxer, so they do not (or should not) exhibit this issue.

> Shouldn't "stop" be a no-op if the output was already requested to stop? I'm not sure this fix is in the correct place - stopping should already be set to false in `deactivate`.

Perhaps. We could maybe check `active(stream)` in `ffmpeg_mux_stop`. That said, the commit history indicates that `ts == 0` was added to allow force stops, and I hadn't investigated how that factors in to this.

Lain-B (approved):
I'm going to tentatively approve this to patch this issue.

PR #3983, opened 2020-12-27
UI: Resize scene items when user resizes canvas

cg2121:
### Description
If the user resizes the canvas with the same aspect ratio as
before, resize all of the scene items in each scene.

### Motivation and Context
Before, the users would have to resize each item manually when changing the canvas size.

### How Has This Been Tested?
Started with 1920x1080, then resized to 1280x720 and made sure everything was resized correctly.

### Types of changes
- New feature (non-breaking change which adds functionality)

### Checklist:
- [x] My code has been run through [clang-format](https://github.com/obsproject/obs-studio/blob/master/.clang-format).
- [x] I have read the [**contributing** document](https://github.com/obsproject/obs-studio/blob/master/CONTRIBUTING.rst).
- [x] My code is not on the master branch.
- [x] The code has been tested.
- [x] All commit messages are properly formatted and commits squashed where appropriate.
- [x] I have included updates to all appropriate documentation.

PR #9910, opened 2023-11-23
libobs: Change scene items to using relative coordinates

derrod:
### Description

> [!IMPORTANT]
> Scenes are saved in a backwards-compatible way, stuff might not work correctly and mess up your collection.
> Make backups of your scene collections before playing around with this!

This change enables the canvas resolution to be changed while maintaining the relative positioning and sizing of items in a scene collection.

**Video demo (outdated):**

https://github.com/obsproject/obs-studio/assets/3123295/fe3ad89f-d7f7-4b44-b1b1-514a5ffbf693

Implementation notes:
- `(0.0, 0.0)` is the center of the canvas 
- Positions are stored as offsets relative to the center
- The height of the canvas has the range `[-1.0f, 1.0f]` while the width side is `[-x, x]` where `x = 1.0f * aspect ratio`
- Items without bounding boxes retain the canvas resolution used during their original creation to adjust their scaling multiplier to the current resolution
- libobs API is unchanged and converts automatically between absolute and relative positions
- Scene collections are backwards compatible, the new values are stored separately from the old ones

**Visualisation of current and proposed resizing behaviour:**
explanation_combined_v4

#### Future considerations
- adopt new APIs in UI where it makes sense
    + Perhaps we can allow users to switch between pixels and relative positions in the transform UI?
- How do we handle a situation where a user loaded a scene collection with the wrong canvas resolution and it needs to be re-migrated?

### Motivation and Context

This is similar to how other software behaves, and allows much more flexibility when it comes to adjusting a scene collection for different needs. For example, adjustments when switching switching resolutions become simpler if the aspect ratios changes, and no adjustments at all are required when maintaining aspect ratio.

In contrast to previous attempts (#3983) this does not require any changes to the scene collection or user interaction when changing canvas resolutions. The only concern (see *ToDos* above) is how to handle migration from existing collections.

### How Has This Been Tested?

Fiddled with some scene collections.

### Types of changes

### Checklist:

### DIFF
--- a/UI/data/locale/en-US.ini
+++ b/UI/data/locale/en-US.ini
@@ -594,6 +594,14 @@ Basic.Main.AddSceneCollection.Text="Please enter the name of the scene collectio
 # rename scene collection dialog
 Basic.Main.RenameSceneCollection.Title="Rename Scene Collection"
 
+# remigrate scene collection dialog
+Basic.Main.RemigrateSceneCollection.Title="Update Scene Collection Resolution"
+Basic.Main.RemigrateSceneCollection.Text="Do you want to update the scene collection resolution of \"%1\" to match the current profile's canvas resolution of %2x%3?"
+Basic.Main.RemigrateSceneCollection.CannotMigrate.Active="Cannot update scene collection resolution while outputs are active."
+Basic.Main.RemigrateSceneCollection.CannotMigrate.UnknownBaseResolution="Failed to update scene collection resolution. The original resolution is unknown."
+Basic.Main.RemigrateSceneCollection.CannotMigrate.FailedVideoReset="Reset not possible: Changing OBS resolution failed."
+Basic.Main.RemigrateSceneCollection.CannotMigrate.BaseResolutionMatches="Reset not possible: Current resolution already the base resolution of scene collection."
+
 # add profile dialog
 AddProfile.Title="Add Profile"
 AddProfile.Text="Please enter the name of the profile"
@@ -835,6 +843,8 @@ Basic.MainMenu.Profile.Import="Import Profile"
 Basic.MainMenu.Profile.Export="Export Profile"
 Basic.MainMenu.SceneCollection.Import="Import Scene Collection"
 Basic.MainMenu.SceneCollection.Export="Export Scene Collection"
+Basic.MainMenu.SceneCollection.Remigrate="Reset Base Resolution"
+Basic.MainMenu.SceneCollection.Migrate="Set Base Resolution"
 Basic.MainMenu.Profile.Exists="The profile already exists"
 
 # basic mode help menu
--- a/UI/window-basic-main.cpp
+++ b/UI/window-basic-main.cpp
@@ -953,7 +953,16 @@ void OBSBasic::Save(const char *file)
 		obs_data_set_obj(saveData, "resolution", res);
 	}
 
-	if (!obs_data_save_json_safe(saveData, file, "tmp", "bak"))
+	obs_data_set_int(saveData, "version", usingAbsoluteCoordinates ? 1 : 2);
+
+	if (migrationBaseResolution && !usingAbsoluteCoordinates) {
+		OBSDataAutoRelease res = obs_data_create();
+		obs_data_set_int(res, "x", migrationBaseResolution->first);
+		obs_data_set_int(res, "y", migrationBaseResolution->second);
+		obs_data_set_obj(saveData, "migration_resolution", res);
+	}
+
+	if (!obs_data_save_json_pretty_safe(saveData, file, "tmp", "bak"))
 		blog(LOG_ERROR, "Could not save scene data to %s", file);
 }
 
@@ -1038,6 +1047,20 @@ void OBSBasic::CreateFirstRunSources()
 				 Str("Basic.AuxDevice1"), 3);
 }
 
+void OBSBasic::DisableRelativeCoordinates(bool enable)
+{
+	/* Allow disabling relative positioning to allow loading collections
+	 * that cannot yet be migrated. */
+	OBSDataAutoRelease priv = obs_get_private_data();
+	obs_data_set_bool(priv, "AbsoluteCoordinates", enable);
+	usingAbsoluteCoordinates = enable;
+
+	ui->actionRemigrateSceneCollection->setText(
+		enable ? QTStr("Basic.MainMenu.SceneCollection.Migrate")
+		       : QTStr("Basic.MainMenu.SceneCollection.Remigrate"));
+	ui->actionRemigrateSceneCollection->setEnabled(enable);
+}
+
 void OBSBasic::CreateDefaultScene(bool firstStart)
 {
 	disableSaving++;
@@ -1048,6 +1071,7 @@ void OBSBasic::CreateDefaultScene(bool firstStart)
 	ui->transitionDuration->setValue(300);
 	SetTransition(fadeTransition);
 
+	DisableRelativeCoordinates(false);
 	OBSSceneAutoRelease scene = obs_scene_create(Str("Basic.Scene"));
 
 	if (firstStart)
@@ -1188,10 +1212,11 @@ void OBSBasic::LogScenes()
 	blog(LOG_INFO, "------------------------------------------------");
 }
 
-void OBSBasic::Load(const char *file)
+void OBSBasic::Load(const char *file, bool remigrate)
 {
 	disableSaving++;
 	lastOutputResolution.reset();
+	migrationBaseResolution.reset();
 
 	obs_data_t *data = obs_data_create_from_json_file_safe(file, "bak");
 	if (!data) {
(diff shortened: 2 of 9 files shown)


tt2468:
I assume you plan to add v2 functions that accept the new float values?

ryantheleach:
1. Are the positions exposed to the user at all, or is it merely a behind the scenes storage change?
2. Are there any circumstances where due to rounding or something, that the entered size doesn't round trip correctly? At what canvas size do these errors first occur?

derrod:
> 1. Are the positions exposed to the user at all, or is it merely a behind the scenes storage change?
> 2. Are there any circumstances where due to rounding or something, that the entered size doesn't round trip correctly? At what canvas size do these errors first occur?

1. Not in this PR, though APIs are added to access the raw values
2. Any number that is sub-pixel will be rounded to the nearest 0.5 for non-group items (to allow centering of odd-sized sources while keeping things reasonably aligned), within the limits of OBS canvas size (16k x 16k) there should be no issues

Warchamp7:
I'd like to hold off on this PR for the time being while we look into storing an "intended resolution" as part of scene collection data.

Warchamp7 (changes requested):
I'd like to hold off on this PR for the time being while we look into storing an "intended resolution" as part of scene collection data.

derrod:
Updated this PR with a behavioural change to scale to the shorter side instead of the longer one. This makes the maths less "neat" but is probably more intuitive as long as you are resizing within the aspect ratios > 1 or < 1 (swapping between those is still a  bit weird, but that's not really fixable).

explanation_v2_to_v3_combined

derrod:
> I'd like to hold off on this PR for the time being while we look into storing an "intended resolution" as part of scene collection data.

Starting with 30.1 we store the last used output resolution in the collection, with the new scene collection management we should be able to do automatic migrations with fairly good certainty, though we may still need options for manual corrections.

derrod:
Updated based on conversations off-thread:

- Migrations are now handled based on resolution saved in scene collection if an output was used
- Users can remigrate a collection if the wrong base resolution was used (Scene Collections -> Reset Base Resolution)
- Users can manually migrate a collection if it hasn't been migrated yet (same menu item)
- The legacy (absolute) mode can still be used, it requires setting a boolean value on the private OBS settings data

derrod:
Made a few improvements to the migration code, this is how users can fix an incorrectly migrated collection:

https://github.com/obsproject/obs-studio/assets/3123295/432c357b-c16b-49da-823e-b96197cdb0d9

Warchamp7 (changes requested):

derrod:
Updated strings as suggested.

nathaniasSC2:
Hey just wanted to say I used the most recent artifact build to resize a 4k canvas into 1080p and then reboot back into my standard OBS install, it worked like a charm. Tools like this are super useful! (I had base at 4k scaled to 1080p, I wanted to lower canvas to 1080p so gpu was doing less scaling since I am also running twitch enhanced broadcast and am generally streamlining things down. I'd rather use a 2nd app to record 4k feeds)

UI/data/locale/en-US.ini
```diff
@@ -591,6 +591,14 @@ Basic.Main.AddSceneCollection.Text="Please enter the name of the scene collectio
 # rename scene collection dialog
 Basic.Main.RenameSceneCollection.Title="Rename Scene Collection"
 
+# remigrate scene collection dialog
+Basic.Main.RemigrateSceneCollection.Title="Update Collection Resolution"
+Basic.Main.RemigrateSceneCollection.Text="Do you want to update the scene collection resolution of "%1" to match the current profile's canvas resolution of %2x%3?"
```
RytoEX, line 596: Do the inner quotes need to be replaced by single-quotes or escaped?
```
of "%1" to
```

UI/window-basic-main.cpp
```diff
@@ -924,6 +933,20 @@ void OBSBasic::CreateFirstRunSources()
 				 Str("Basic.AuxDevice1"), 3);
 }
 
+void OBSBasic::DisableRelativeCoordinates(bool enable)
+{
+	/* Allow disabling relative positioning to allow loading collections
+	 * that cannot yet be migrated.  */
```
RytoEX, line 939:
```suggestion
	 * that cannot yet be migrated. */
```

RytoEX (changes requested):
Quick pass. I haven't looked at `libobs/obs-scene.c` because the diff is large, and I haven't done any local builds/tests.

UI/data/locale/en-US.ini
derrod: They need to be escaped, yes. This was a mistake introduced when updating the strings based on earlier feedback.

GeorgesStavracas (approved):
This has been discussed previously from the UI / UX perspective and it was approved a while ago. This just needs code review now.

libobs/obs-scene.c
```diff
@@ -3520,6 +3896,8 @@ static void resize_group(obs_sceneitem_t *group)
 			new_pos.y = (maxv.y - minv.y) * 0.5f + minv.y;
 
 		transform_val(&new_pos, &group->draw_transform);
+		if (!group->absolute_coordinates)
+			pos_from_absolute(&new_pos, &new_pos, group);
```
RytoEX, line 3900:
```suggestion
		if (!group->absolute_coordinates)
			pos_from_absolute(&new_pos, &new_pos, group);

```

RytoEX (changes requested):
Visually this _looks_ like it makes sense. Will have to do a local build and test some migrations to see how things go, though I'm not clear on what I should expect to see. Additional review/testing would be appreciated.

derrod:
Updated the PR to address comments and squash some commits. I've also elected to make a minor change to always scale to height, rather than the shortest side, to make transitions between horizontal and vertical a bit more intuitive.

This also includes a change perhaps of interest to @exeldro in 1423687feb1fcebe898d610705cc203bd456222e and ca27351a4f0a1c8adc5b1e8a055cbcf4ab10cb54 which should get rid of the need to set a custom scene resolution in their vertical plugin since it'll use the current mix's base resolution when rendering, rather than just the main mix.

libobs/obs.h
```diff
@@ -1871,11 +1873,15 @@ EXPORT void obs_sceneitem_set_bounds_alignment(obs_sceneitem_t *item,
...
 EXPORT void obs_sceneitem_set_bounds(obs_sceneitem_t *item,
 				     const struct vec2 *bounds);
+EXPORT void obs_sceneitem_set_relative_bounds(obs_sceneitem_t *item,
+				      const struct vec2 *bounds);
 
 EXPORT int64_t obs_sceneitem_get_id(const obs_sceneitem_t *item);
 
 EXPORT void obs_sceneitem_get_pos(const obs_sceneitem_t *item,
 				  struct vec2 *pos);
+EXPORT void obs_sceneitem_get_pos2(const obs_sceneitem_t *item,
```
exeldro, line 1883: rename to obs_sceneitem_get_relative_pos

derrod: Oh somehow I missed this one.

exeldro:
Did code review and tested it, all looks good to me.

derrod:
> Reference in ne

Thanks, have you been able to verify that using a custom resolution is no longer required for vertical canvasses since it now uses the mix's base resolution?

exeldro:
With vertical I found some issues.
- when using `obs_display_add_draw_callback` to render the vertical canvas there is no way to set `current_mix`, so it does use the main canvas resolution or the custom scene resolution.
- `obs_view_render` does not set `current_mix`, so it does use the main canvas resolution or the custom scene resolution.
- In `obs_transition_start` and `obs_transition_tick` the `current_mix` is not set, so the calculated size of the transition is wrong here:
https://github.com/obsproject/obs-studio/blob/master/libobs/obs-source-transition.c#L210
https://github.com/obsproject/obs-studio/blob/master/libobs/obs-source-transition.c#L395

derrod:
I see, we will have to revisit that as it goes a bit beyond the scope of this PR. I'll  drop the first two commits then as they don't have any other function that would be useful right now. Perhaps that warrants an RFC at some point to figure these things out, as for the time being. The transitions example is particularly problematic as there's can be one `tick` but multiple `render` calls with differing mixes and resolutions, so we might need to have a fixed link between scenes and mixes or something.

docs/sphinx/reference-scenes.rst
```diff
@@ -369,6 +369,18 @@ Scene Item Functions
 
 ---------------------
 
+.. function:: void obs_sceneitem_set_relative_pos(obs_sceneitem_t *item, const struct vec2 *pos)
+              void obs_sceneitem_set_relative_pos(const obs_sceneitem_t *item, struct vec2 *pos)
```
RytoEX, line 373: Both of these function names say `set`. I imagine the second one should be get.

RytoEX (changes requested):

docs/sphinx/reference-scenes.rst
derrod: Fixed.

PR #9911, opened 2023-11-23
docs: fix missing accessor in example

Susko3:
### Description

Tested this in code, and it makes sense from the way `scenes.sources.array[i]` is accessed.

### Types of changes

### Checklist:


norihiro:
The change itself looks correct but the first word of the commit title should be capitalized. So, the commit message should be like this.
```text
docs: Fix missing accessor in example
```

Lain-B (approved):

PR #9918, opened 2023-11-26
CI: Use distributed network for service uptime check

derrod:
### Description

Updates service check job to use geographically distributed servers provided by @tt2468 to increase accuracy of uptime checks.

### Motivation and Context

Currently the check suffers from an increasing number of false-positives.

### How Has This Been Tested?

Ran on my fork: https://github.com/derrod/obs-studio/actions/runs/6992208282/job/19023560306

### Types of changes

### Checklist:

### DIFF
--- a/.github/actions/services-validator/action.yaml
+++ b/.github/actions/services-validator/action.yaml
@@ -4,6 +4,12 @@ inputs:
   repositorySecret:
     description: GitHub token for API access
     required: true
+  checkApiSecret:
+    description: Token for server check API
+    required: false
+  checkApiServers:
+    description: Servers for the check API
+    required: false
   runSchemaChecks:
     description: Enable schema checking
     required: false
@@ -46,8 +52,8 @@ runs:
           eval "$(/home/linuxbrew/.linuxbrew/bin/brew shellenv)"
           echo "/home/linuxbrew/.linuxbrew/bin:/home/linuxbrew/.linuxbrew/sbin" >> $GITHUB_PATH
         fi
-        brew install --overwrite --quiet python3
-        python3 -m pip install jsonschema json_source_map requests
+        brew install --overwrite --quiet python@3.11
+        python3 -m pip install jsonschema json_source_map requests aiohttp
         echo ::endgroup::
 
     - name: Validate Services File JSON Schema 🕵️
@@ -91,6 +97,8 @@ runs:
         GITHUB_TOKEN: ${{ inputs.repositorySecret }}
         WORKFLOW_RUN_ID: ${{ github.run_id }}
         REPOSITORY: ${{ github.repository }}
+        API_KEY: ${{ inputs.checkApiSecret }}
+        API_SERVERS: ${{ inputs.checkApiServers }}
       run: |
         : Check for defunct services 📉
         python3 -u .github/scripts/utils.py/check-services.py
--- a/.github/scripts/utils.py/check-services.py
+++ b/.github/scripts/utils.py/check-services.py
@@ -1,15 +1,15 @@
+import asyncio
 import json
-import socket
-import ssl
 import os
 import time
-import requests
 import sys
 import zipfile
 
+import aiohttp
+import requests
+
 from io import BytesIO
-from random import randbytes
-from urllib.parse import urlparse
+from typing import List, Dict
 from collections import defaultdict
 
 MINIMUM_PURGE_AGE = 9.75 * 24 * 60 * 60  # slightly less than 10 days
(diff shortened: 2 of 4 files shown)


.github/actions/services-validator/action.yaml
```diff
@@ -46,8 +52,8 @@ runs:
           eval "$(/home/linuxbrew/.linuxbrew/bin/brew shellenv)"
           echo "/home/linuxbrew/.linuxbrew/bin:/home/linuxbrew/.linuxbrew/sbin" >> $GITHUB_PATH
         fi
-        brew install --overwrite --quiet python3
-        python3 -m pip install jsonschema json_source_map requests
+        brew install --overwrite --quiet python@3.11
+        python3 -m pip install jsonschema json_source_map requests aiohttp aiodns
```
tt2468, line 56: `aiodns` doesn't seem to be used anywhere.

.github/scripts/utils.py/check-services.py
```diff
@@ -265,25 +208,47 @@ def main():
...
+                        check_servers_task(session, host, protocol, servers)
+                    )
+                )
+            results = await asyncio.gather(*tasks)
+        except Exception as e:
+            print(
+                f"❌ Querying server status for \"{service['name']}\" failed with: {e}"
+            )
+            return 1
+
```
tt2468, line 238: Could we add perhaps a 2 second wait between services in order to add a rudimentary rate limit to the API requests?

(This comment's position in the diff is not of relevance)

.github/scripts/utils.py/check-services.py
```diff
@@ -265,25 +208,47 @@ def main():
...
+        for server, result in zip(service["servers"], zip(*results)):
+            failure_count = sum(not res["status"] for res in result)
+            probe_count = len(result)
+            # only treat server as failed if all check servers reported a failure
+            is_ok = failure_count < probe_count
 
             if not is_ok:
+                failures = {res["comment"] for res in result if res["comment"]}
+                print(
+                    f"⚠️ Connecting to server failed: {server['url']} (Reason(s): {failures})"
```
tt2468, line 249: Since the URL is considered a secret key by github, this message will be censored automatically. Is there an interface where privileged members can see the actual value? If not, you could perform a `server['url'].split('.', 1)[1]` to get the region name from the URL.

.github/scripts/utils.py/check-services.py
```diff
@@ -265,25 +208,47 @@ def main():
...
+
+        # go over results
+        for server, result in zip(service["servers"], zip(*results)):
+            failure_count = sum(not res["status"] for res in result)
+            probe_count = len(result)
+            # only treat server as failed if all check servers reported a failure
+            is_ok = failure_count < probe_count
 
             if not is_ok:
+                failures = {res["comment"] for res in result if res["comment"]}
```
tt2468, line 247: I've designed the API to possibly return a string if the service checked out OK. A more accurate check would be `in result if not res['status']`

.github/actions/services-validator/action.yaml
derrod: aiohttp will use it automatically to speed up lookups, probably overkill though since we only have to do a few.

tt2468: Ah interesting, forgot about implicit library optimizations

.github/scripts/utils.py/check-services.py
derrod: The URL here is the streaming server URL, not the checker URL. If any of them fail that's currently treated as a failure of the script and it will exit.

derrod: Updated to hide the prefix and suffix and leave the region code. Tested on my repo to confirm it works as expected.

.github/scripts/utils.py/check-services.py
```diff
@@ -313,9 +278,11 @@ def main():
         if not new_service["servers"]:
             print(f'💀 Service "{service["name"]}" has no valid servers left, removing!')
             affected_services[service["name"]] = f"Service removed"
-            continue
+        else:
+            new_services["services"].append(new_service)
 
-        new_services["services"].append(new_service)
+        # wait a bit between services
+        time.sleep(2.0)
```
PatTheMav, line 285:
```suggestion
        await asyncio.sleep(2)
```

Using a sync wait in a coroutine is probably bad for the event loop.

tt2468 (approved):

RytoEX (approved):
Seemed fine at a glance.

.github/scripts/utils.py/check-services.py
PatTheMav, line 1: Did you run this file through Black?

derrod: yes

PatTheMav (approved):

PatTheMav:
@derrod feel free to merge this once the necessary repository secrets have been added.

PR #9672, opened 2023-10-06
win-wasapi: Use provided timestamp for process capture and handle flags

derrod:
### Description

Changes process audio capture to use the OS-provided timestamp instead of calculating our own. In tests this seems to resolve audio issues reported by various users (also see: https://github.com/stream-labs/obs-studio/pull/543).

Additionally adds some logging and handling for flags that may be set by `IAudioClient::GetBuffer()`:
- `AUDCLNT_BUFFERFLAGS_TIMESTAMP_ERROR` is treated as failure, buffer is free'd and capture reset
- `AUDCLNT_BUFFERFLAGS_DATA_DISCONTINUITY` is logged to debug, but ignored as it probably won't cause issues with libobs's audio data processing
- `AUDCLNT_BUFFERFLAGS_SILENT` now replaces the buffer with all zeroes to avoid processing potentially stale or garbage data (also logged to debug as it can get very spammy)

### Motivation and Context

Hopefully helps with #8064 (not closing that one quite yet).

### How Has This Been Tested?

A user with extremely reliable reproduction of the audio issue reported no problems during a test recording.

### Types of changes

- Bug fix (non-breaking change which fixes an issue)
- Tweak (non-breaking change to improve existing functionality)

### Checklist:

- [x] My code has been run through [clang-format](https://github.com/obsproject/obs-studio/blob/master/.clang-format).
- [x] I have read the [**contributing** document](https://github.com/obsproject/obs-studio/blob/master/CONTRIBUTING.rst).
- [x] My code is not on the master branch.
- [x] The code has been tested.
- [x] All commit messages are properly formatted and commits squashed where appropriate.
- [x] I have included updates to all appropriate documentation.

PR #9925, opened 2023-11-27
win-wasapi: Ignore timestamp error

prgmitchell:
### Description

Only log packets with timestamp errors rather than resetting the device.

### Motivation and Context

A user reported audio devices constantly disconnecting/reconnecting as a result of timestamp error handling added in #9672 

### How Has This Been Tested?

It has not been tested.

### Types of changes

### Checklist:


derrod:
Probably fine given that previously we simply accepted these packets anyway and users didn't report any issues with that. Per [the documentation](https://learn.microsoft.com/en-us/windows/win32/api/audioclient/ne-audioclient-_audclnt_bufferflags) this flag only indicates that the timestamp for the current packet may be uncertain, so it's probably fine to throw that one away and then continue instead of fully resetting the device.

prgmitchell:
Updated PR now that it has been tested by a user in Discord.

PR #9931, opened 2023-11-28
libobs: Reuse matching mix's render texture if possible

derrod:
### Description

Reuses the render texture for mixes that match a previously renderer mix's view/colorspace/resolution.

In combination with #9937 this allows GPU scaling to be used for encoded auxiliary views.

### Motivation and Context

Currently the encoder-only mixes created for GPU scaling (added in #9043) re-render the entire mix, which seems wasteful when we already have a perfectly good render texture!

Ideally "expensive" sources such as filters will only get rendered once per tick anyway, but some filters may not do this, e.g. `obs-composite-blur` didn't until https://github.com/FiniteSingularity/obs-composite-blur/pull/16, so this may save some GPU time for users with many filters as well.

### How Has This Been Tested?

Tested with GPU scaling enabled on recording output ([like so](https://github.com/derrod/obs-studio/commit/fadfb94749d20f7d80b6be3e1a74505a44079696)), and verified that sources no longer get rendered twice and output is still correct.

GPU usage was measurably lower in a test scene with a filter with its internal texture reuse disabled (RTX Greenscreen). See my [see test branch](https://github.com/derrod/obs-studio/commits/gpu-scaling-texture-reuse-test/).

### Types of changes

### Checklist:

### DIFF
--- a/libobs/obs-video.c
+++ b/libobs/obs-video.c
@@ -129,6 +129,44 @@ static inline void unmap_last_surface(struct obs_core_video_mix *video)
 	}
 }
 
+static inline bool can_reuse_mix_texture(const struct obs_core_video_mix *mix,
+					 size_t *idx)
+{
+	for (size_t i = 0, num = obs->video.mixes.num; i < num; i++) {
+		const struct obs_core_video_mix *other =
+			obs->video.mixes.array[i];
+		if (other == mix)
+			break;
+		if (other->view != mix->view)
+			continue;
+		if (other->render_space != mix->render_space)
+			continue;
+		if (other->ovi.base_width != mix->ovi.base_width ||
+		    other->ovi.base_height != mix->ovi.base_height)
+			continue;
+		if (!other->texture_rendered)
+			continue;
+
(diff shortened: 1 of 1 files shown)


tt2468:
Looking back on #9043 and this PR, I see there seems to be some differing behavior for main view vs aux views. I'm trying to understand what motivations there may be, but I'm having a hard time. What's the limitation?

derrod:
> Looking back on #9043 and this PR, I see there seems to be some differing behavior for main view vs aux views. I'm trying to understand what motivations there may be, but I'm having a hard time. What's the limitation?

For #9042 it probably mostly comes down to use case, aux views are generally not used for encoded outputs atm. Though arguably that could've been set up more flexibly. In this case I maintained that limitation because the main mix is the only one guaranteed to be rendered before any other mixes, so it's easy to have identical mixes reuse its texture.

derrod:
Updated the PR to be generic rather than only working on the main mix. Also PR'd #9937 to enable this for auxiliary views.

Issue #9841, opened 2023-11-11
30.0.0-rc2 (and 29.1.3) fail to build with ffmpeg 6.1


tgurr:
### Operating System Info

Other

### Other OS

Exherbo Linux

### OBS Studio Version

30.0.0-rc2

### OBS Studio Log URL

-

### Expected Behavior

No build issues.

### Current Behavior

FTBFS

```
[ 17%] Building C object libobs/CMakeFiles/libobs.dir/media-io/media-remux.c.o
[...]

PR #7462, opened 2022-09-25
libobs: Fixes for the new FFmpeg channel API

pkviet:
### Description
Also changes obs-ffmpeg.
This allows compilation with FFmpeg master.
More precisely for:
-  swresample >= 4.5.100, 
- avutil >= 57.24.100,
- avcodec >= 59.24.100.

One commit adapts the changes done in 5b6cc73 to the obs-ffmpeg-mpegts output.
The second updates obs audio resampler by swapping a deprecated function to its new version 
which uses the new channel layout API.
The third deals with the deprecation of AVFrame and AVCodecContext channels member by switching to new API.

### Motivation and Context
Enable later updates of FFmpeg library on which obs-studio relies for a variety of stuff.

### How Has This Been Tested?
test platform: win 10 pro 21H2 x64 & win 11 22h2, visual studio 2022 

- checked that audio is resampled correctly with both new & old api (both new and old FFmpeg dlls),
- checked that mono audio is upmixed correctly with both new and old api.
- checked that mpegts output works with new FFmpeg dlls.

### Types of changes
- Bug fix (non-breaking change which fixes an issue)

### Checklist:
- [x] My code has been run through [clang-format](https://github.com/obsproject/obs-studio/blob/master/.clang-format).
- [x] I have read the [**contributing** document](https://github.com/obsproject/obs-studio/blob/master/CONTRIBUTING.rst).
- [x] My code is not on the master branch.
- [x] The code has been tested.
- [x] All commit messages are properly formatted and commits squashed where appropriate.
- [x] I have included updates to all appropriate documentation.

PR #9935, opened 2023-11-29
Add support for FFmpeg 6.1

Stephen-Seo:
### Description

Deprecated FFmpeg API usage (deprecated by FFmpeg 6.1) are updated to use the [newer API as detailed here.](https://github.com/FFmpeg/FFmpeg/blob/master/doc/APIchanges).

~~Note that one of the commits removes code that uses the deprecated API without replacement code. This is because the metadata being copied by the removed code is handled by the latter usage of `avcodec_parameters_copy()` which copies over the metadata that is now included in `AVCodecParameters` since FFmpeg 6.1.~~

The code that is no longer necessary since FFmpeg 6.1 is now in a `#if` block. See the previous comment for details.

### Motivation and Context

On ArchLinux, the AUR package obs-studio-git fails to build due to usage of deprecated FFmpeg functions causing build fails (as the deprecation warnings are treated as errors). Currently, ArchLinux uses FFmpeg 6.1. This PR introduces changes that fix the usage of the FFmpeg API to follow the new changes that were made for FFmpeg 6.1.

### How Has This Been Tested?

I patched the PKGBUILD (file used to build ArchLinux packages) to use the new changes and built and installed obs-studio on my machine. I tested recording the screen and obs-studio was able to record and stop without issue, and reported no memory leaks.

### Types of changes

- Bug fix (non-breaking change which fixes an issue) (addresses https://github.com/obsproject/obs-studio/issues/9841 )

Note that these changes would probably fail to build with any version of FFmpeg less than version 6.1.

### Checklist:

Not sure if documentation is required for this change, so it is left unchecked for now.

### DIFF
--- a/deps/media-playback/media-playback/decode.c
+++ b/deps/media-playback/media-playback/decode.c
@@ -114,8 +114,14 @@ static uint16_t get_max_luminance(const AVStream *stream)
 {
 	uint32_t max_luminance = 0;
 
+#if LIBAVCODEC_VERSION_INT < AV_VERSION_INT(60, 31, 102)
 	for (int i = 0; i < stream->nb_side_data; i++) {
 		const AVPacketSideData *const sd = &stream->side_data[i];
+#else
+	for (int i = 0; i < stream->codecpar->nb_coded_side_data; i++) {
+		const AVPacketSideData *const sd =
+			&stream->codecpar->coded_side_data[i];
+#endif
 		switch (sd->type) {
 		case AV_PKT_DATA_MASTERING_DISPLAY_METADATA: {
 			const AVMasteringDisplayMetadata *mastering =
--- a/deps/media-playback/media-playback/media.c
+++ b/deps/media-playback/media-playback/media.c
@@ -504,7 +504,12 @@ void mp_media_next_video(mp_media_t *m, bool preload)
 	}
 
 	if (!m->is_local_file && !d->got_first_keyframe) {
+
+#if LIBAVUTIL_VERSION_INT < AV_VERSION_INT(58, 29, 100)
 		if (!f->key_frame)
+#else
+		if (!(f->flags & AV_FRAME_FLAG_KEY))
+#endif
 			return;
 
 		d->got_first_keyframe = true;
--- a/libobs/media-io/media-remux.c
+++ b/libobs/media-io/media-remux.c
@@ -91,6 +91,7 @@ static inline bool init_output(media_remux_job_t job, const char *out_filename)
 			return false;
 		}
 
+#if LIBAVCODEC_VERSION_INT < AV_VERSION_INT(60, 31, 102)
 #if FF_API_BUFFER_SIZE_T
 		int content_size;
 #else
@@ -125,6 +126,7 @@ static inline bool init_output(media_remux_job_t job, const char *out_filename)
 				       mastering_size);
 			}
 		}
+#endif
 
 		ret = avcodec_parameters_copy(out_stream->codecpar,
 					      in_stream->codecpar);
--- a/plugins/obs-ffmpeg/ffmpeg-mux/ffmpeg-mux.c
+++ b/plugins/obs-ffmpeg/ffmpeg-mux/ffmpeg-mux.c
@@ -498,9 +498,17 @@ static void create_video_stream(struct ffmpeg_mux *ffm)
 			av_content_light_metadata_alloc(&content_size);
 		content->MaxCLL = max_luminance;
 		content->MaxFALL = max_luminance;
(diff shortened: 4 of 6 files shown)


tt2468:
Many of the updates in this PR require ifdefs. Please update your commits to include them as other PRs in the past have, to allow for building across ffmpeg versions.

Stephen-Seo:
> Many of the updates in this PR require ifdefs. Please update your commits to include them as other PRs in the past have, to allow for building across ffmpeg versions.

Ok, I will work on this.

Stephen-Seo:
I'm going to rename the PR and edit my older comment to indicate that obs can build with the new changes even with older versions of FFmpeg since the checks have passed. Reviews are appreciated.

Stephen-Seo:
> Thanks for your PR. Could you reword the commits to state what is deprecated. Also in the commits messages, please link to the appropriate FFmpeg commits. See for instance: https://github.com/obsproject/obs-studio/pull/7462/commits

Ok I will push the requested changes tomorrow (UTC+9 here) which is when I will have more time. It might take longer, but I think I should be able to do this by the end of tomorrow.

Stephen-Seo:
Actually it seems I over-estimated how long it will take to fix the commit comments. Please let me know if there are more changes needed or if I made a mistake.

pkviet:
> Actually it seems I over-estimated how long it will take to fix the commit comments. Please let me know if there are more changes needed or if I made a mistake.

you forgot to fix the commit titles 😉

Stephen-Seo:
> you forgot to fix the commit titles 😉

I was under the impression they didn't need changing, so I wasn't sure what is more correct. Are these commit titles ok?

pkviet:
> > you forgot to fix the commit titles 😉
> 
> I was under the impression they didn't need changing, so I wasn't sure what is more correct. Are these commit titles ok?

I was thinking of something more explicit like:
- obs-ffmpeg: Use new side-data FFmpeg API
- deps/media-playback: Use new side-data FFmpeg API and new key-frame flag

Stephen-Seo:
> > I was under the impression they didn't need changing, so I wasn't sure what is more correct. Are these commit titles ok?
> 
> I was thinking of something more explicit like:
> 
>     * obs-ffmpeg: Use new side-data FFmpeg API
> 
>     * deps/media-playback: Use new side-data FFmpeg API and new key-frame flag

I split one of the commits since it was doing two somewhat different kinds of changes to make describing the change-per-commit easier. Also, the commit titles are longer than 50 characters this way, but I hope it's fine.

deps/media-playback/media-playback/decode.c
```diff
@@ -114,8 +114,15 @@ static uint16_t get_max_luminance(const AVStream *stream)
 {
 	uint32_t max_luminance = 0;
 
+// FFmpeg 6.1 puts the side data into AVCodecParameters.
+#if LIBAVCODEC_VERSION_MAJOR < 60 || LIBAVCODEC_VERSION_MINOR < 31
```
derrod, line 118: This really should be something like `#if LIBAVCODEC_VERSION_INT < AV_VERSION_INT(60, 31, 100)` to avoid false-positives if there's a version `(61, 0, 0)` or similar.

derrod (changes requested):
Many of the comments seem unnecessary, the multi-line ones should probably be in the commit message instead. We also generally use `/* ... */` for comments in C code.

Stephen-Seo:
Ok I'll make the requested changes soon. I need to sleep first. Thanks for
the review.

Stephen-Seo:
> Many of the comments seem unnecessary, the multi-line ones should probably be in the commit message instead. We also generally use `/* ... */` for comments in C code.

I fixed up the code as suggested. I did a quick test again on my end (with FFmpeg 6.1) just in case, and it looks good.

deps/media-playback/media-playback/decode.c
Stephen-Seo: This has been fixed in the latest push. Thank you for the review.

Stephen-Seo:
> Many of the comments seem unnecessary, the multi-line ones should probably be in the commit message instead. We also generally use `/* ... */` for comments in C code.

This has been fixed in the latest push. Unless there is more changes needed to be made?

Stephen-Seo:
I have no option to resolve/dismiss the "requested change" status.

screenshot of options for requested change ui

I hope the PR review process can continue regardless of this @derrod .

Stephen-Seo:
I think I misunderstood and that it was requested to remove the comment block in the code altogether in one of the commits. The same explanation exists in the commit message, so the change this time is just to remove the block comment in the `libobs` commit. Sorry for the confusion.

derrod (approved):

PR #9944, opened 2023-11-30
UI: Fall back in case of invalid scene name on load

Penwy:
### Description

This allows the UI to fall back to a correct scene when loading a scene collection if the specified `current_scene` is not a valid scene.

### Motivation and Context

When loading a scene collection where the specified `current_scene` is not the name of an existing scene (example attached below), OBS would crash.
[improper.json](https://github.com/obsproject/obs-studio/files/13517064/improper.json)

### How Has This Been Tested?

Loaded the example scene collection attached, verified there was no crash and no noticed adverse effects.

### Types of changes

### Checklist:

### DIFF
--- a/UI/window-basic-main.cpp
+++ b/UI/window-basic-main.cpp
@@ -1211,6 +1211,17 @@ retryScene:
 		goto retryScene;
 	}
 
+	if (!curScene) {
+		auto find_scene_cb = [](void *source_ptr, obs_source_t *scene) {
+			OBSSourceAutoRelease &source =
+				reinterpret_cast<OBSSourceAutoRelease &>(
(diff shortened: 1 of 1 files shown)


Fenrirthviti:
How does a scene collection get in to this state in the first place? While crash fixes are typically always a good idea, if there's another bug that's causing a collection to get in to this state in the first place, it should be corrected there.

Penwy:
As far as I know, no manipulation of the SC by obs itself can cause this state.
The case here would be due to improper editing of the file by a third party, especially given that the importer does not mark such a scene collection as invalid.

Lain-B (approved):

PR #9957, opened 2023-12-03
libobs: Pair video encoder with all audio encoders

### LINKED ISSUE #4530: Audio track 1 offset/desynced
### Operating System Info

Windows 10

### Other OS

_No response_

### OBS Studio Version

Other

OBS 27.0.0-rc2

### OBS Studio Log URL

https://obsproject.com/logs/y9XfULuwDnckI7R0

### OBS Studio Crash Log URL

_No response_

### Expected Behavior

When recording with multiple audio tracks, I would expect audio track 1 (0) to be synced with the other audio tracks.

### Current Behavior

Regardless of encoder (video or audio), regardless of container (mp4 / mkv), audio track 1 (0) is not in sync with the other tracks. It seems to be lagging behind (its later) by some degree, maybe around 10-20ms.

The PTS and DTS seem to be identical for all tracks, so I believe its earlier in the pipeline (prior to encoding and muxing).

### Steps to Reproduce

1. Enable two or more audio tracks Output>recording (advanced)
2. Ensure you have the same audio device on both (or more)
3. Hit start recording
4. Use your mic to record yourself clapping or snapping your fingers
5. Hit stop recording
6. Import the video file into an editor/NLE capable of handling multiple tracks, like Adobe Premiere or Davinci Resolve
7. Observe how audio track 1 is later than the others (which are in sync)

This occurs not only on the rc2, but also on the current release version (26.1.1). I reproduced this on rc2, and didnt want the log to differ from the reported version. This was reproduced on an nvme SSD.

I've confirmed this happens on at least one more machine. I've observed it in Davinci Resolve personally, and I've had a friend check it in premiere (same issue).

derrod:
### Description

Change encoder pairing to sync video encoder to all audio encoders and vice-versa.

Also removes unused `wait_for_video` bool.

### Motivation and Context

Fixes longstanding issue where first audio track in multi-track recordings is offset from the other audio tracks.

Fixes #4530

| **Before** | **After** |
| - | - |
| before | after |

### How Has This Been Tested?

Made some test recordings and confirmed no longer broken.

### Types of changes

### Checklist:

### DIFF
--- a/libobs/obs-encoder.c
+++ b/libobs/obs-encoder.c
@@ -678,7 +678,7 @@ void obs_encoder_shutdown(obs_encoder_t *encoder)
 	if (encoder->context.data) {
 		encoder->info.destroy(encoder->context.data);
 		encoder->context.data = NULL;
-		encoder->paired_encoder = NULL;
+		da_free(encoder->paired_encoders);
 		encoder->first_received = false;
 		encoder->offset_usec = 0;
 		encoder->start_ts = 0;
@@ -1371,13 +1371,15 @@ static void receive_video(void *param, struct video_data *frame)
 	profile_start(receive_video_name);
 
 	struct obs_encoder *encoder = param;
-	struct obs_encoder *pair = encoder->paired_encoder;
+	struct obs_encoder **paired = encoder->paired_encoders.array;
 	struct encoder_frame enc_frame;
 
-	if (!encoder->first_received && pair) {
-		if (!pair->first_received ||
-		    pair->first_raw_ts > frame->timestamp) {
-			goto wait_for_audio;
+	if (!encoder->first_received && encoder->paired_encoders.num) {
+		for (size_t i = 0; i < encoder->paired_encoders.num; i++) {
+			if (!paired[i]->first_received ||
+			    paired[i]->first_raw_ts > frame->timestamp) {
+				goto wait_for_audio;
+			}
 		}
 	}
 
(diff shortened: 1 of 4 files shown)


libobs/obs-encoder.c
```diff
@@ -678,7 +678,7 @@ void obs_encoder_shutdown(obs_encoder_t *encoder)
 	if (encoder->context.data) {
 		encoder->info.destroy(encoder->context.data);
 		encoder->context.data = NULL;
-		encoder->paired_encoder = NULL;
+		da_clear(encoder->paired_encoders);
```
PatTheMav, line 681: Is that enough? Wouldn't it be more correct to call `darray_free(encoder->paired_encoders)` (it clears implicitly) to avoid a memory leak?

libobs/obs-encoder.c
```diff
@@ -1465,9 +1467,14 @@ static bool buffer_audio(struct obs_encoder *encoder, struct audio_data *data)
 	size_t offset_size = 0;
 	bool success = true;
 
-	if (!encoder->start_ts && encoder->paired_encoder) {
+	struct obs_encoder *paired_encoder = NULL;
+	/* Audio encoders can only be paired to one video encoder */
+	if (encoder->paired_encoders.num)
```
PatTheMav, line 1472: If the comment is true, shouldn't this code also check if `num` is exactly `1` and fail otherwise?

derrod: Could be done, but `pair_encoers` will not add a video encoder here if the audio encoder already has one, so it should never be more than one. It was more meant to clarify as to why this isn't iterating over them like the video encoder counterpart is.

Lain-B (approved):

PR #9964, opened 2023-12-04
UI: Save invalid transition source data in safe mode

### LINKED ISSUE #9963: Safe mode removes third-party transitions from scene collection.
### Operating System Info

Ubuntu 22.04

### Other OS

Also windows

### OBS Studio Version

30.0.0

_No response_

### OBS Studio Log URL

https://obsproject.com/logs/mr8PiNv7t9APqOkP

### OBS Studio Crash Log URL

_No response_

### Expected Behavior

Transistions added by third-party plugins should still be present when launching normally, even after launching in safe mode once.

### Current Behavior

After having started in safe mode and closed, on restart, transitions added by third-party plugins have been removed.

### Steps to Reproduce

1. Open OBS and add a third-party transition (typically, move-transition).
2. Close and restart in safe mode.
3. Close and restart in normal mode.
4. Notice the transition isn't there anymore.
...

_No response_

derrod:
### Description

Saves transitions that failed creating the source properly in separate array if safe mode is enabled.

There may be a better way of doing this, hence a draft.

### Motivation and Context

Fixes #9963 

### How Has This Been Tested?

Verified transitions are not removed from JSON anymore.

### Types of changes

### Checklist:

### DIFF
--- a/UI/window-basic-main-transitions.cpp
+++ b/UI/window-basic-main-transitions.cpp
@@ -1779,6 +1779,10 @@ obs_data_array_t *OBSBasic::SaveTransitions()
 		obs_data_array_push_back(transitions, sourceData);
 	}
 
+	for (const OBSDataAutoRelease &transition : safeModeTransitions) {
+		obs_data_array_push_back(transitions, transition);
+	}
+
 	return transitions;
 }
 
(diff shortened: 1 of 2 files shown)


exeldro:
I like the way you made it and think the draft can be removed.

gxalpha:
Do we maybe want to always do this, not just in safe mode? That way if a user temporarily uninstalls a plugin, the transition would also not be gone (kind of how it's done for normal sources).

exeldro:
A good thing to add would be logging that it is happening. It can be logging per transition that failed to load or just a summary at the end telling how many failed to load.

Lain-B (approved):

PR #9975, opened 2023-12-08
libobs-d3d11: Add display primaries to the log

jpark37:
### Description
Can't hurt to know what Windows thinks the mastering primaries are.

### Motivation and Context
Doubt this info would be used in a support situation, but it could. Plus it's fun to have the info there.

### How Has This Been Tested?
Example output:
```
22:04:00.652: 	    space=RGB_FULL_G2084_NONE_P2020
22:04:00.652: 	    primaries=[r=(0.679688, 0.312500), g=(0.281250, 0.666016), b=(0.154297, 0.066406), wp=(0.312500, 0.329102)]
22:04:00.652: 	    relative_gamut_area=[709=1.266340, P3=0.933509, 2020=0.669730]
22:04:00.652: 	    sdr_white_nits=100
```

### Types of changes

### Checklist:

### DIFF
--- a/libobs-d3d11/d3d11-subsystem.cpp
+++ b/libobs-d3d11/d3d11-subsystem.cpp
@@ -1310,6 +1310,12 @@ static void PopulateMonitorIds(HMONITOR handle, char *id, char *alt_id,
 	}
 }
 
+static constexpr double DoubleTriangleArea(double ax, double ay, double bx,
+					   double by, double cx, double cy)
+{
+	return ax * (by - cy) + bx * (cy - ay) + cx * (ay - by);
+}
+
 static inline void LogAdapterMonitors(IDXGIAdapter1 *adapter)
 {
 	UINT i;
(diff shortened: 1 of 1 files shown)


Lain-B (approved):

WizardCM:
I'm curious, what triggered this PR? Is it an issue that has come up often in your troubleshooting? Is there documentation (even on Wikipedia) that'd explain to a support volunteer how to understand these numbers?

jpark37:
All three gamut trianlges are on this page.
https://en.wikipedia.org/wiki/DCI-P3

Examples of wide color (use a monitor with HDR enabled):
https://webkit.org/blog-files/color-gamut/

The coordinates define a triangle of color reproducibility. This information can be set in HDR metadata, which is a soft promise that the colors will not go outside of that triangle because you as the author couldn't see those colors to tune them.

One way OBS could use this information in the future is to remap colors for HDR preview so broadcasters have a gamut-remapped view that doesn't clip colors outside of their display range.

Another future use would be to copy the display primaries (manually or automaticly) into the output video metadata, which we currently just hardcode to Display P3 coordinates IIRC. A broadcaster may not want to do that though if they want to preserve colors even if they can't see them.

Also note that the reported primaries can be of dubious accuracy, and the Windows HDR calibration tool has also been reported to clamp the reported gamut to Rec. 709 although not on my machine.

Lain-B:
Thank you for the detailed explanation

PR #9978, opened 2023-12-08
libobs: Mark data path modification functions as deprecated

PatTheMav:
### Description
Marks `obs_add_data_path` and `obs_remove_data_path` as deprecated.

### Motivation and Context
Change was suggested by @Lain-B  in https://github.com/obsproject/obs-studio/pull/8179.

### How Has This Been Tested?
Successfully compiled without deprecation warnings on macOS 14.1.2.

### Types of changes

### Checklist:

### DIFF
--- a/libobs/obs.c
+++ b/libobs/obs.c
@@ -1273,13 +1273,15 @@ char *obs_find_data_file(const char *file)
 	return NULL;
 }
 
-void obs_add_data_path(const char *path)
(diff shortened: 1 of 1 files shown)


PatTheMav:
Converted to draft again because internal investigation has revealed a user of this API.

RytoEX (approved):
Discussed this again, and we should be okay to deprecate now, remove later, and the API user will add their own implementation.

PR #9979, opened 2023-12-09
Add Flatpak aarch64 builds (Part 1: build)

tytan652:
Add aarch64 support to the manifest and to the CI.

**Publish workflow will be done once the PR design for build is approved.**

#### About aarch64 on CI:

Building through QEMU emulation is really slow. So to avoid timing out
the job (6h), the original job is split in two.

The first job build dependencies (4-5h) and creates an artifact, if the cache already exist the
build is skipped.

And the second job build CEF and OBS Studio by relying on the cache if no cache on the artifact of the previous job (2-3h without cache).

aarch64 builds are not enabled by default on the main workflow on PRs.

### Motivation and Context

Publish Flatpak aarch64 builds.

### How Has This Been Tested?

CI with a PR on my fork: https://github.com/tytan652/obs-studio/pull/16 (aarch64 artifacts availlable)

### Types of changes

### Checklist:


.github/workflows/build-project.yaml
```diff
@@ -28,25 +29,30 @@ jobs:
 
           case "${GITHUB_EVENT_NAME}" in
             pull_request)
-              config_data=('codesign:false' 'notarize:false' 'package:false' 'config:RelWithDebInfo')
-              if gh pr view ${{ github.event.number }} --json labels \
+              config_data=('codesign:false' 'notarize:false' 'package:false' 'config:RelWithDebInfo' 'flatpakArch:["x86_64"]')
```
PatTheMav, line 32: I'd rather have this step create a bespoke matrix variable instead of putting this in `config_data`, as described in the GitHub action documentation:

https://docs.github.com/en/enterprise-cloud@latest/actions/learn-github-actions/expressions#example-returning-a-json-object

I'm fine with `flatpakArch` being set up as a bespoke output variable depending on GitHub event.

.github/workflows/build-project.yaml
```diff
@@ -28,25 +29,30 @@ jobs:
...
+              label_data=$(gh pr view ${{ github.event.number }} --json labels)
+              if echo "${label_data}" \
                 | jq -e -r '.labels[] | select(.name == "Seeking Testers")' > /dev/null; then
                 config_data[0]='codesign:true'
                 config_data[2]='package:true'
               fi
+              if echo "${label_data}" \
```
PatTheMav, line 39: I'm not really happy about introducing a bespoke label just for this, but code-wise it's fine.

.github/workflows/flatpak-build.yaml
```diff
@@ -0,0 +1,193 @@
+name: Flatpak Build
+on:
+  workflow_call:
+    inputs:
+      arch:
+        default: x86_64
```
PatTheMav, line 6: Add descriptions for each input, they are mandatory per the YAML schema (at least my linter automatically screams at me about them).

.github/workflows/flatpak-build.yaml
```diff
@@ -0,0 +1,193 @@
...
+        type: string
+      build-bundle:
+        default: true
+        type: boolean
+      bundle:
+        default: obs-studio-flatpak.flatpak
+        type: string
+jobs:
+  build-until-90:
+    name: Build Modules Until 90-* 🧾
```
PatTheMav, line 16: A more correct approach would be to be able to explicitly specify which modules to exclude vs using something like `stop-at-module` which depends on the order at which modules are specified. The intent here is to "build everything _but_ this one module", so why this roundabout way of achieving it?

Also doesn't this exclude nlohmann-json and websocketpp as well, as numerically they come after `asio` (another reason why making this based on order is brittle).

At the very least the job should be called something like "Build without CEF" (regardless of its module's filename number) and the filenames "-without-cef" or "-no-cef". IMO it makes more sense to use a name describing the purpose vs describing the implementation detail of achieving the purpose.

plugins/obs-qsv11/CMakeLists.txt
```diff
@@ -2,12 +2,24 @@ cmake_minimum_required(VERSION 3.22...3.25)
 
 legacy_check()
 
-option(ENABLE_QSV11 "Build Intel QSV11 Hardware Encoder." TRUE)
+if(CMAKE_SYSTEM_PROCESSOR MATCHES "^(i[3-6]86|x86|x64|x86_64|amd64|AMD64)$")
```
PatTheMav, line 5: You cannot rely on `CMAKE_SYSTEM_PROCESSOR` on Windows because the value will not correctly reflect the toolchain selected for MSVC (e.g. it will always be `x86_64` even when building for ARM64 on an Intel-based Windows machine).

As this change mainly affects Flatpak, guard this off with `if(OS_LINUX)` where CMake toolchain files are more commonly used (which set `CMAKE_SYSTEM_PROCESSOR` accordingly).

plugins/obs-qsv11/cmake/legacy.cmake
```diff
@@ -1,9 +1,21 @@
-option(ENABLE_QSV11 "Build Intel QSV11 Hardware Encoder." TRUE)
+if(CMAKE_SYSTEM_PROCESSOR MATCHES "^(i[3-6]86|x86|x64|x86_64|amd64|AMD64)$")
```
PatTheMav, line 1: See above.

PatTheMav:
Changes seem fine overall (apart from specific review comments), though I'm somewhat unsure about the added complexity.

Also how does this impact our cache usage? The Flatpak cache is the largest line item in our cache usage and I wonder how aarch64 builds will impact this.

.github/workflows/flatpak-build.yaml
tytan652: 90-* are headers-only libraries, those are excluded from the first job.

The intent is build until and stop at the first 90-* module which is `asio`

PatTheMav: As I mentioned that all seems very brittle and entirely based on conventions that are easily broken - we always need to remember not to add any further dependencies that might use any prefix above 90 or we'll break it, especially as the entire intent is to just build without CEF.

I'd rather have either:

* Two separate manifests (one with CEF and one without), though I have a hunch that Flatpak is not designed to allow that, OR
* Enhance `flatpak-builder` to be able to accept one or more "excluded modules" which are then simply skipped (and the values are based on the `name` in the module) than relying on build order and `stop-at-module`

The manifest is already a JSON file (and so are the modules), so why not use their metadata to clearly steer the build process and also express inter-module dependencies that way?

tytan652: - [ ] TODO: More bash code with `jq` to get the first 90-* module

1. Refused by Georges, he asked to improve the action to add support for `stop-at-module` since it already exist inside `flatpak-builder` and use it.
2. `flatpak-builder` has a module order and you can ask it to stop at a module, I think they try to keep the tool simple which is a good thing.

.github/workflows/flatpak-build.yaml
```diff
@@ -0,0 +1,193 @@
...
+          if [[ "${RUNNER_DEBUG}" ]]; then set -x; fi
+
+          git config --global --add safe.directory "${GITHUB_WORKSPACE}"
+
+          echo '::group::Install GitHub CLI tool'
+          dnf install -y -q gh
+          gh extension install actions/gh-actions-cache
+          echo '::endgroup::'
+
+          if [[ "${{ inputs.arch }}" != "x86_64" ]]; then
```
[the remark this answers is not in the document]
tytan652: Change some things but I prefer future-proofing with x86_64 as host than hard-coding to only aarch64 and x86_46.

.github/workflows/flatpak-build.yaml
PatTheMav: Got it - yeah I'm fine with the implementation if this is the only way to achieve it - I would like to think that being able to selectively exclude some modules is a pretty common ask and should be supported directly (I currently don't know what `stop-at-module` could be used for instead).

The main issue with it being based on file name ordering persists (a pattern I'd rather move away from), but that shouldn't hold up this PR.

tytan652: A Flatpak manifest is a recipe, `flatpak-builder` follows the recipe order with a certain strictness but the builder can be asked to stop at a module.
The `--stop-at` is mainly meant  for testing purpose, we just happen to find another usage.
This is not a "packaging system" with dependencies to solve or a tooling like buildstream.

plugins/obs-qsv11/CMakeLists.txt
```diff
@@ -2,12 +2,24 @@ cmake_minimum_required(VERSION 3.22...3.25)
 
 legacy_check()
 
-option(ENABLE_QSV11 "Build Intel QSV11 Hardware Encoder." TRUE)
+if(NOT OS_LINUX OR CMAKE_SYSTEM_PROCESSOR MATCHES "^(i[3-6]86|x86|x64|x86_64|amd64|AMD64)$")
```
PatTheMav, line 5: I actually thought this through another time and I think the better way would be to do handle this in the main plugins CMakeLists file with something similar to `check_obs_browser`:

* The plugin modules themselves should be almost agnostic as much as possible
* The decision whether to load a plugin for a specific OS or architecture should be made in the place that "selects" which plugin directories to include.

plugins/obs-qsv11/cmake/legacy.cmake
```diff
@@ -1,9 +1,21 @@
-option(ENABLE_QSV11 "Build Intel QSV11 Hardware Encoder." TRUE)
+if(NOT OS_LINUX OR CMAKE_SYSTEM_PROCESSOR MATCHES "^(i[3-6]86|x86|x64|x86_64|amd64|AMD64)$")
```
PatTheMav, line 1: This is fine, my comment above does not extend to the Legacy build path.

.github/workflows/flatpak-build.yaml
PatTheMav: Yeah I was afraid of that - we're using features for a purpose they were never meant for and we should rely on such crutches less not more, but alas. I won't block the PR over that.

plugins/obs-qsv11/CMakeLists.txt
tytan652: I wait for the aarch64 CI on my fork to finish to check the theoretical cache size. So I will not push right now.

Here is what I done for the macro in the plugins CMakeLists, what do you think ?
```cmake
macro(check_obs_qsv11)
    if(OS_WINDOWS
       OR OS_LINUX
       AND CMAKE_SYSTEM_PROCESSOR MATCHES "^(i[3-6]86|x86|x64|x86_64|amd64|AMD64)$")
      add_subdirectory(obs-qsv11)
    endif()
  endmacro()
```

tytan652:
> Also how does this impact our cache usage? The Flatpak cache is the largest line item in our cache usage and I wonder how aarch64 builds will impact this.

https://github.com/tytan652/obs-studio/actions/runs/7170036033/job/19532684838#step:9:5

Looks like it will be the almost same size as x86_64, so 2 times ~2GB.

.github/workflows/build-project.yaml
tytan652: I could work and make the aarch64 build happen on PR if a cache is there and a label to force the build without cache.
But I would prefer to know, how well/fast it is on master before enabling on every PR (if cache available).

Note that the label name can be discussed.

.github/workflows/flatpak-build.yaml
tytan652: Please open another review thread is the job's name still need to be reviewed. This one became more about the `stop-at-module` determination and the tooling.

plugins/obs-qsv11/CMakeLists.txt
tytan652: The change was pushed.

likeadragonmaid:
Are there any updates to this?

.github/workflows/build-project.yaml
tytan652: Assigning a whole matrix looks unnecessary, making the JSON object (only an array) more complex (an array in an object) where there is only one variable does not look worth it.

I was aware of this feature to set up a matrix but for one variable, it looks too much.

.github/workflows/flatpak-build.yaml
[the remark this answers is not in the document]
tytan652: Just for the sake of completeness:

flatpak-github-actions seeks to avoid dependency on docker (in docker) and the setup-qemu action.

I proposed a solution upstream https://github.com/flatpak/flatpak-github-actions/pull/173, I'm just waiting for feedback from maintainers.

tytan652:
Will be refactored after #10934 has been discussed

tytan652:
> Can the build-related changes go in first without waiting for CI / caching stuff?

No, there is no point to it if we can't use CI to test building the manifest on both architectures.

Also I prioritize getting away from the KDE Sdk with the PR I mentioned in my previous message.

valpackett:
If anyone's looking for a way to build it **now**, all the necessary changes are:

```diff
diff --git a/build-aux/com.obsproject.Studio.json b/build-aux/com.obsproject.Studio.json
index ddfe0e7729..29670b606b 100644
--- a/build-aux/com.obsproject.Studio.json
+++ b/build-aux/com.obsproject.Studio.json
@@ -83,7 +83,6 @@
                 "-DENABLE_VLC=OFF",
                 "-DENABLE_AJA=ON",
                 "-DENABLE_LIBFDK=ON",
-                "-DENABLE_QSV11=ON",
                 "-DENABLE_DECKLINK=OFF"
             ],
             "secret-opts": [
diff --git a/build-aux/modules/10-mbedtls.json b/build-aux/modules/10-mbedtls.json
index 7f63a0bb3b..3f02e9bcf1 100644
--- a/build-aux/modules/10-mbedtls.json
+++ b/build-aux/modules/10-mbedtls.json
@@ -5,6 +5,7 @@
     "config-opts": [
         "-DCMAKE_BUILD_TYPE=Release",
         "-DCMAKE_POSITION_INDEPENDENT_CODE=ON",
+        "-DMBEDTLS_FATAL_WARNINGS=OFF",
         "-DUSE_SHARED_MBEDTLS_LIBRARY=ON",
         "-DUSE_STATIC_MBEDTLS_LIBRARY=OFF",
         "-DENABLE_TESTING=OFF",
diff --git a/build-aux/modules/50-intel-media-sdk.json b/build-aux/modules/50-intel-media-sdk.json
index 8859693765..7a8befcb06 100644
--- a/build-aux/modules/50-intel-media-sdk.json
+++ b/build-aux/modules/50-intel-media-sdk.json
@@ -1,5 +1,8 @@
 {
     "name": "intel-media-sdk",
+    "only-arches": [
+        "x86_64"
+    ],
     "buildsystem": "cmake-ninja",
     "builddir": true,
     "config-opts": [
diff --git a/build-aux/modules/50-vpl-gpu-rt.json b/build-aux/modules/50-vpl-gpu-rt.json
index 8c9755a4c1..431dedaf5b 100644
--- a/build-aux/modules/50-vpl-gpu-rt.json
+++ b/build-aux/modules/50-vpl-gpu-rt.json
@@ -1,5 +1,8 @@
 {
     "name": "vpl-gpu-rt",
+    "only-arches": [
+        "x86_64"
+    ],
     "buildsystem": "cmake-ninja",
     "builddir": true,
     "config-opts": [
diff --git a/build-aux/modules/99-cef.json b/build-aux/modules/99-cef.json
index 623f423fda..91bb2b02ec 100644
--- a/build-aux/modules/99-cef.json
+++ b/build-aux/modules/99-cef.json
@@ -17,8 +17,8 @@
     "sources": [
         {
             "type": "archive",
-            "url": "https://cdn-fastly.obsproject.com/downloads/cef_binary_6533_linux_x86_64_v6.tar.xz",
-            "sha256": "7963335519a19ccdc5233f7334c5ab023026e2f3e9a0cc417007c09d86608146"
+            "url": "https://cdn-fastly.obsproject.com/downloads/cef_binary_6533_linux_aarch64_v6.tar.xz",
+            "sha256": "642514469eaa29a5c887891084d2e73f7dc2d7405f7dfa7726b2dbc24b309999"
         }
     ]
 }
```

Issue #9871, opened 2023-11-15
bug that has been occuring on all public versions of macos 14 (sonoma)~ 14.0, 14.1 & 14.1.1, where the menu bar title glitches from "OBS" -> "OBS Studio" as soon as you do something else in another app and click back on obs.


AIDaN80112:
### Operating System Info

macOS 14

### OBS Studio Version

30.0.0

### OBS Studio Log URL

https://obsproject.com/logs/XlwIpw2lVqjOoZc4

### Expected Behavior

App name to remain "OBS" at all times

### Current Behavior

App name in menu bar title glitches from "OBS" -> "OBS Studio" as soon as you do something else in another app and click back on obs. 

### Steps to Reproduce

1. launch obs
2. do something in another app (not just clicking a button, changing the actual app to display something else like searching something in a browser)
3. go back to obs

### Anything else we should know?

[VIDEO OF ISSUE OCCURING](https://youtu.be/zDU3HlJUTpw) (was too big to upload here so thats why its a yt link)

PatTheMav:
While I was able to observe this behaviour on macOS 14.1, I haven't seen it in macOS 13.5.

What is being displayed is influenced by the `CFBundleName` and `CFBundleDisplayName` in the `Info.plist` of the bundle - in version 29 those were identical, in version 30 we treated them differently, but that seems to lead to the observed behaviour.

I guess it should be fine to align both values again.

bdash:
We saw this same issue in Google Chrome and believe it has been fixed in the most recent macOS 14.2 beta.

PatTheMav:
> We saw this same issue in Google Chrome and believe it has been fixed in the most recent macOS 14.2 beta.

Good to know - "OBS Studio" is our `CFBundleDisplayName`, but iirc only `CFBundleName` is used for menu bar, dock icon, task switcher, etc.

PR #9983, opened 2023-12-10
Revert "cmake: Align Product Name and Bundle Name on macOS"

gxalpha:
### Description

This reverts commit 9f66c23825e41807c56f3a865d1c1462fa7ff73b.

This commit caused a crash in CEF on startup for many users. Let's just revert it asap for now and worry about the minor problem it was solving another time. Although according to https://github.com/obsproject/obs-studio/issues/9871#issuecomment-1815205387%3E that original issue may have been fixed in macOS anyways so I'm not worried about it.

Incidentally it also causes a problem I was having (but couldn't debug because of a lack of time) where the build folder needed cleaning before each rebuild or otherwise code signing was failing.

### Motivation and Context

People report this crash on 30.0.1 and I don't like crashes.

### How Has This Been Tested?

In my setup, which was also effected by this crash (which I only noticed a few days ago), it no longer crashes.

### Types of changes

### Checklist:

### DIFF
--- a/.github/scripts/.build.zsh
+++ b/.github/scripts/.build.zsh
@@ -179,14 +179,12 @@ build() {
       if [[ ${GITHUB_EVENT_NAME} == push && ${GITHUB_REF_NAME} =~ [0-9]+.[0-9]+.[0-9]+(-(rc|beta).+)? ]] {
         run_xcodebuild ${archive_args}
         run_xcodebuild ${export_args}
-
-        mv "OBS Studio.app" OBS.app
       } else {
         run_xcodebuild ${build_args}
 
         rm -rf OBS.app
         mkdir OBS.app
-        ditto "UI/${config}/OBS Studio.app" OBS.app
+        ditto UI/${config}/OBS.app OBS.app
       }
       popd
       ;;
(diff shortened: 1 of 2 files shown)


derrod:
If this is a temporary change, maybe we should only push it to the release branch and do a proper fix on masteR?

gxalpha:
I think pushing only to release branches is a bad habit that we shouldn’t do.
Also I don’t necessarily consider the revert to be temporary, as per https://github.com/obsproject/obs-studio/issues/9871#issuecomment-1815205387 the underlying problem may have been fixed in macOS 14.2 anyways so it should go away by itself.

PatTheMav:
What is the reason that this could fix the crash? I don't see any reason (yet) how this would influence anything regarding CEF.

Also the change was logically correct because if the application is named "OBS Studio" the bundle needs to be named the change and us renaming the bundle to just "OBS.app" is actually wrong.

PatTheMav:
Just thinking about it, I understand what the issue is - the helper apps need to be named "OBS Studio Helper" instead of "OBS Helper" obviously. Stupid oversight.. 🤣

PatTheMav (approved):
Now that I understand the root cause I'm fine with reverting this - app bundle name and file bundle name are "OBS" and we should assume that this is the name used in the menu bar.

_If_ it displays "OBS Studio" that would be the error in macOS, but the expected result is for it to be called just "OBS".

PR #9990, opened 2023-12-11
UI: Remove lock & vis checkbox classes

cg2121:
### Description
This removes the lock & visibility checkbox classes in favor of using dynamic properties.

### Motivation and Context
Less files to worry about.

### How Has This Been Tested?
Went through every theme and made sure everything looked correct.

### Types of changes

### Checklist:


cg2121:
Turns out we still need the mute checkbox, as it is tristate and uses custom code. Updated this PR to just remove the visibility and lock checkboxes.

GeorgesStavracas (approved):

Warchamp7:
I'd like to follow more in line with the phrasing used for styling widgets like QToolButtons, which is `toolButton`. As such I suggest the following for the property names:

- `visibilityCheckbox`
- `lockCheckbox`

cg2121:
Updated PR based on suggestions by @Warchamp7

Warchamp7 (approved):
Approved from my side at least. Will need to ensure no other unforeseen side effects.

Is it possible for plugins to have been using the custom widgets for example?

Lain-B:
I don't think plugins have been using those particular custom widgets. I hope.

Lain-B:
Actually even if they were, they'd probably have to duplicate the files or something anyway. So I'll just merge it.

exeldro:
Plugins have been using copies of those widgets, those will lose the icons they get from the theme, but I can do a small change in my plugins to fix that and support this change.
Custom themes will also need an update to support this change.

PR #10000, opened 2023-12-13
obs-webrtc: Add support for audio-only and video-only outputs

tt2468:
### Description
Adds the `whip_output_audio` and `whip_output_video` output kinds, which selectively advertise only video or only audio support.

To use these types, it is effectively the same as with the AV version. Just create the output, assign your video or audio encoder, and you're good.

### Motivation and Context
It's not uncommon in WebRTC systems to have video-only or audio-only streams. The WHIP protocol supports this, so this will allow plugins which do custom things with WHIP to access this functionality.

### How Has This Been Tested?
Code compiles on Ubuntu 22.04

I've been using this patch for many months in production privately, and I think it can benefit more than just myself.

### Types of changes

### Checklist:

### DIFF
--- a/plugins/obs-webrtc/whip-output.cpp
+++ b/plugins/obs-webrtc/whip-output.cpp
@@ -83,12 +83,12 @@ void WHIPOutput::Data(struct encoder_packet *packet)
 		return;
 	}
 
-	if (packet->type == OBS_ENCODER_AUDIO) {
+	if (audio_track && packet->type == OBS_ENCODER_AUDIO) {
 		int64_t duration = packet->dts_usec - last_audio_timestamp;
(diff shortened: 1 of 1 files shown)


Lain-B:
Possible we might need to think about whether we want to implement multiple outputs for audio or video only outputs or implement support like this, or change the backend to support optional video or audio on outputs. Hm.

DDRBoxman (approved):

Lain-B (approved):

Lain-B:
I forgot that we were going to update the codecs to not have the video codec on the audio output and not have the audio codec on the video output. We'll have to add a fix commit for that. My apologies for that.

PR #10006, opened 2023-12-14
UI: Add transition management APIs

Elgato-AStory:
### Description
This PR provides two new APIs for managing transitions in the frontend. This allows plugins to create and delete new instances of transitions on behalf of the user, as well as the ability to properly rename existing transitions.

### Motivation and Context
I am working on a plugin to programmatically manage OBS on behalf of the user, and I noticed that this feature I needed was absent from the frontend API. Note that this does not provide functions for managing quick transitions as that is outside of the scope for what I needed.

I did make a couple of decisions that might not sit right with everyone and I'm open to changing them in light of any substantial objections:

1. I provided the ability to rename transitions by registering a callback for the "rename" signal rather than providing an API call specifically for renaming. I felt that this was more appropriate, however this does slightly conflict with the other decision I made.
2. The rename and add-transition behaviors automatically update the name of the source to be unique, similarly to how copying and pasting sources in the UI does with a simple brute force scan of existing sources and appending a number to the name. I'm on the fence about whether this belongs in the API or if it should fall on the API consumer to not provide duplicated names. Since it's the frontend API, I'm leaning more towards having it protect itself from bad state, however the behavior might cause some surprises, especially when it comes to using `obs_source_set_name` since the name you set might be immediately overwritten. This does also open up a fun side effect, you can now successfully rename the non-configurable transitions from a plugin and have it reflect in the UI. I consider doing that to be invoking undefined behavior though.

### How Has This Been Tested?
I have only tested this manually with a debug plugin that called the functions with a series of hardcoded parameters on Windows 10 22H2

### Types of changes

### Checklist:

### DIFF
--- a/UI/api-interface.cpp
+++ b/UI/api-interface.cpp
@@ -110,6 +110,20 @@ struct OBSStudioAPI : obs_frontend_callbacks {
 		}
 	}
 
+	void obs_frontend_add_transition(obs_source_t *transition) override
+	{
+		QMetaObject::invokeMethod(main, "AddTransitionInstance",
+					  Q_ARG(OBSSource,
+						OBSSource(transition)));
+	}
+
+	void obs_frontend_remove_transition(obs_source_t *transition) override
+	{
+		QMetaObject::invokeMethod(main, "RemoveTransitionInstance",
+					  Q_ARG(OBSSource,
+						OBSSource(transition)));
+	}
+
 	void obs_frontend_get_transitions(
 		struct obs_frontend_source_list *sources) override
 	{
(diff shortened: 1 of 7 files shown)


tt2468:
Working on reviewing this, we're currently trying to consider whether accepting a raw transition pointer in the add transition method is a safe and future-proof idea, since transitions have many more quirks in order to be used correctly than other types of sources.

Elgato-AStory:
Alright, I'm open to changing it to whatever makes the most sense.

Warchamp7:
Bumping @tt2468 again on this

Warchamp7:

This gap in the API is still quite a sore spot so a resubmission of this work would be greatly appreciated.

PR #10021, opened 2023-12-19
rtmp-services: Update Switchboard Live servers

sarwin-github:
### Description
The Switchboard Live Server URLs have been recently updated in the application, so the URLs need to be updated in OBS to match.

### Motivation and Context
There were recent changes to the ingestion servers at Switchboard Live. We also removed Legacy url.

### How Has This Been Tested?
- This was tested using the latest build of OBS on Windows.
- Testing consisted of going live with every server to ensure connectivity was stable.

### Types of changes
- Tweak (non-breaking change to improve existing functionality). 
- Removed Alt names

### Checklist:

### DIFF
--- a/plugins/rtmp-services/data/services.json
+++ b/plugins/rtmp-services/data/services.json
@@ -1156,12 +1156,8 @@
             ],
             "servers": [
                 {
-                    "name": "Global - Recommended",
+                    "name": "Default",
                     "url": "rtmps://live.sb.zone:443/live"
-                },
-                {
-                    "name": "Global - Legacy",
-                    "url": "rtmp://ingest-global.switchboard.zone/live"
                 }
             ],
             "recommended": {
--- a/plugins/rtmp-services/data/package.json
+++ b/plugins/rtmp-services/data/package.json
@@ -1,11 +1,11 @@
 {
     "$schema": "schema/package-schema.json",
     "url": "https://obsproject.com/obs2_update/rtmp-services/v5",
-    "version": 242,
+    "version": 243,
     "files": [
         {
             "name": "services.json",
-            "version": 242
+            "version": 243
         }
     ]
 }


plugins/rtmp-services/data/services.json
```diff
@@ -1151,17 +1151,10 @@
         },
         {
             "name": "Switchboard Live",
-            "alt_names": [
-                "Switchboard Live (Joicaster)"
-            ],
```
gxalpha, line 1156: Is this changed intentional? This will break old setups that still have the old name set.

sarwin-github: Hello @gxalpha the `alt_names: "Switchboard Live (Joicaster)"` is no longer valid for Switchboard, so yes it's intentional.
Can you explain what are those breaking changes?  Thanks

gxalpha: The settings in OBS store the selected service by its name.
Any user that had selected "Switchboard Live (Joicaster)" before the name was changed to "Switchboard Live" in https://github.com/obsproject/obs-studio/commit/3ec053ef5df885571ba48aac6daf077412ed71b9 would still have that name stored in their OBS settings. OBS uses the `alt_names` to figure out what service that refers to today.
*As far as I know* (and I could be mistaken here), `alt_names` are not displayed anywhere in the UI.

sarwin-github: Got it, I might need to update my PR to simply remove the word (Joicaster) on the alt_names

Fenrirthviti: The `alt_name` should *not* be updated to remove the previous name, as this may potentially break existing user's setups. Only the `name` value is shown in the UI, but alt_name is used to match configurations that are from before the change.

plugins/rtmp-services/data/services.json
```diff
@@ -1152,16 +1152,12 @@
         {
             "name": "Switchboard Live",
             "alt_names": [
-                "Switchboard Live (Joicaster)"
+                "Switchboard Live"
```
gxalpha, line 1155: The previous comment still has not been addressed, see https://github.com/obsproject/obs-studio/pull/10021#discussion_r1431835434.

gxalpha (changes requested):
The commit history is getting ugly, try to only have one commit containing the change, without any merge commits. This is why it's recommended to not use the master branch for PRs.
See the comment below.

Fenrirthviti:
Commits still need to be squashed before this can be merged.

plugins/rtmp-services/data/services.json
RytoEX: @sarwin-github This has still not been addressed. Do not change existing entries in alt_names. They are there for backwards compatibility reasons. We will not merge this as-is.

RytoEX: I fixed this manually.

RytoEX (approved):
I've fixed the issues with this PR manually because the legacy ingest is gone and is causing CI failures when we try to run our services tests.

In the future:
1. Do not submit a PR from your fork's master branch.
2. Do not use merge commits to keep your branch up to date. Use `git rebase` instead.
3. Do not change or remove existing entries in the `alt_names` key. You can only add new entries. This is used so that existing user setups with the old names do not break.

If you do not follow this advice, your future PRs may be closed.

PR #10039, opened 2023-12-24
libobs/plugins: Add region of interest (ROI) encoder feature

derrod:
### Description

Adds support for specifying regions of interest (ROI) for video encoders.

### Motivation and Context

ROI is a means of prioritising certain regions of the frame for encoding, i.e. retaining higher quality, or sacrificing quality, in areas of the encoded picture.

Possible use cases are:
1. Ensuring important elements remain legible even in bitrate-starved scenarios (e.g. face cam, UI text)
2. Increasing perceptual quality by assigning priority to center of attention (e.g. crosshair in FPS games)
3. Making [fun glitch art](https://pbs.twimg.com/media/GCGxFYGWQAAUo1M?format=jpg&name=large) 😛

### Demo 

**No ROI:**

https://github.com/obsproject/obs-studio/assets/3123295/eaf54245-d9b1-4d50-9715-4b0102c8c724

**ROI on "face cam":**

https://github.com/obsproject/obs-studio/assets/3123295/8f3ed4af-d520-4db9-8b6a-55c1a24c932e

This sample was encoded using x264 slow at 4 Mbps, note that this used an earlier code revision that had an off-by-one error resulting in the ROI being misaligned from the intended element (face cam).

### How Has This Been Tested?

Tested with hardcoded regions on the advanced video recording encoder (APIs to set them via the UI TBD).

Hardware encoder changes were tested with the following combinations:
- NVENC: RTX 4090 with latest drivers
- AMF: RX 6700M with latest drivers
- QSV: Arc A750 with recent drivers

### Types of changes

### Checklist:

### DIFF
--- a/libobs/obs-encoder.c
+++ b/libobs/obs-encoder.c
@@ -52,6 +52,7 @@ static bool init_encoder(struct obs_encoder *encoder, const char *name,
 	pthread_mutex_init_value(&encoder->callbacks_mutex);
 	pthread_mutex_init_value(&encoder->outputs_mutex);
 	pthread_mutex_init_value(&encoder->pause.mutex);
+	pthread_mutex_init_value(&encoder->roi_mutex);
 
 	if (!obs_context_data_init(&encoder->context, OBS_OBJ_TYPE_ENCODER,
 				   settings, name, NULL, hotkey_data, false))
@@ -63,6 +64,8 @@ static bool init_encoder(struct obs_encoder *encoder, const char *name,
 	if (pthread_mutex_init(&encoder->outputs_mutex, NULL) != 0)
 		return false;
 	if (pthread_mutex_init(&encoder->pause.mutex, NULL) != 0)
+		return false;
+	if (pthread_mutex_init(&encoder->roi_mutex, NULL) != 0)
 		return false;
 
 	if (encoder->orig_info.get_defaults) {
@@ -377,10 +380,12 @@ static void obs_encoder_actually_destroy(obs_encoder_t *encoder)
 		if (encoder->context.data)
 			encoder->info.destroy(encoder->context.data);
 		da_free(encoder->callbacks);
+		da_free(encoder->roi);
 		pthread_mutex_destroy(&encoder->init_mutex);
 		pthread_mutex_destroy(&encoder->callbacks_mutex);
 		pthread_mutex_destroy(&encoder->outputs_mutex);
 		pthread_mutex_destroy(&encoder->pause.mutex);
+		pthread_mutex_destroy(&encoder->roi_mutex);
 		obs_context_data_free(&encoder->context);
 		if (encoder->owns_info_id)
 			bfree((void *)encoder->info.id);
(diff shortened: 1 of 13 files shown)


libobs/obs-encoder.h
```diff
@@ -102,6 +103,23 @@ struct encoder_frame {
 	int64_t pts;
 };
 
+/** Encoder region of interest */
+struct region_of_interest {
```
norihiro, line 107: In my opinion, it would be better to have a prefix of `obs_` such as `obs_encoder_roi` or `obs_region_of_interest`.

libobs/obs-encoder.c
```diff
@@ -1874,3 +1879,93 @@ uint64_t obs_encoder_get_pause_offset(const obs_encoder_t *encoder)
...
+	pthread_mutex_unlock(&encoder->roi_mutex);
+
+	return true;
+}
+
+void obs_encoder_remove_roi(obs_encoder_t *encoder,
+			    const struct region_of_interest *roi)
+{
+	pthread_mutex_lock(&encoder->roi_mutex);
+	da_erase_item(encoder->roi, roi);
```
norihiro, line 1942: We need to handle the case that the scale is enabled like `obs_encoder_add_roi` does.

norihiro: It looks the force-pushed commit resolves this.

derrod: It's not a force push yet because I'm not entirely sure if that's the appropriate way to solve this either. I'm trying to figure out if it should be expected for ROIs passed to the callback to work for the remove call, or if the caller should be responsible of keeping track of what ROIs they added. Perhaps there should be a separate means of enumerating all added ROIs and receiving properly scaled ones for use within the encoder.

libobs/obs-encoder.h
derrod: We use fairly generic names like `encoder_packet` in the encoder interface so I figured this is probably fine. We seem to mostly use `obs_` prefixes for structs that are opaque types but I don't know if this is a rule.

plugins/obs-ffmpeg/obs-nvenc.c
```diff
@@ -678,6 +683,25 @@ static bool init_encoder_base(struct nvenc_data *enc, obs_data_t *settings,
 	config->rcParams.vbvBufferSize = bitrate * 1000;
 	config->rcParams.multiPass = nv_multipass;
 
+	/* -------------------------- */
+	/* Region of interest         */
+
+	// ToDo make sure there are no side effects to setting this by default
+	config->rcParams.qpMapMode = NV_ENC_QP_MAP_DELTA;
```
derrod, line 690: There's nothing in the documentation that suggests this won't work, but I can't test it right now. Need to make sure this doesn't have any negative impact.

(I'd like to set it by default so a ROI can be added at any point, no other encoder requires configuring this feature during init so I hope NVENC is also fine if we set this but just pass a `NULL` pointer until we actually have something)

derrod: Confirmed on my NVIDIA machine that this does not have any adverse effect. In fact, the delta mode even works without excplicitly enabling it (if `qpDeltaMap` is non-NULL it always seems to be used, even if `qpMapMode` is left on `NV_ENC_QP_MAP_DISABLED`).

derrod:
With some more testing and some cleanup I now consider this PR ready for proper review. A GUI for controlling ROIs is available on my repo on the [roi-ui](https://github.com/derrod/obs-studio/tree/roi-ui) branch. A slightly outdated test build is available for Windows: https://github.com/derrod/obs-studio/releases/tag/30.0.2-roi-experiment

libobs/obs-encoder.h
Lain-B: There weren't any particular rules. When I used encoder_packet it was probably because it sounded better than obs_encoder_packet and I felt like it probably wasn't going to clash with anything. I probably should have tried to maintain some level of consistency. I'd say we should *probably* try to lean towards consistent behavior so maybe norihiro's suggestion is for the best or something.

tt2468: +1 for `obs_encoder_roi`, and also moving towards a naming syntax including a prefix of `obs_[category]_`

libobs/obs.h
```diff
@@ -2457,6 +2457,30 @@ EXPORT void obs_encoder_set_gpu_scale_type(obs_encoder_t *encoder,
...
+				   const struct region_of_interest *roi);
+/** Clear all regions */
+EXPORT void obs_encoder_clear_roi(obs_encoder_t *encoder);
+/** Enumerate regions with callback (reverse order of addition) */
+EXPORT void obs_encoder_enum_roi(obs_encoder_t *encoder,
+				 void (*enum_proc)(void *,
+						   struct region_of_interest *),
+				 void *param);
+/** Get ROI increment, encoders must rebuild their ROI map if it has changed */
+EXPORT uint32_t obs_encoder_get_roi_increment(const obs_encoder_t *encoder);
```
tt2468, line 2482: For some reason using a uint32_t for this feels risky. Maybe just ensure that nothing breaks if this value wraps and add a note to the docs?

derrod: I don't think anybody is going to update the ROIs enough to exceed 4.2 billion revisions in one OBS runtime (if you make one change every frame that would take over 2 years). Even if it rolls over, since the comparison should be `!=` not `>`, it should be fine.

libobs/obs-encoder.c
```diff
@@ -1874,3 +1879,100 @@ uint64_t obs_encoder_get_pause_offset(const obs_encoder_t *encoder)
...
+			enum_proc(param, roi);
+		}
+	}
+
+	pthread_mutex_unlock(&encoder->roi_mutex);
+}
+
+uint32_t obs_encoder_get_roi_increment(const obs_encoder_t *encoder)
+{
+	return encoder->roi_increment;
```
tt2468, line 1977: Probably a good idea to use atomics here?

derrod: Writing the value is mutexed, reading it probably doesn't need to be atomic and I don't see any situation where it not being atomic could cause an issue.

tt2468:
libobs API looks good once the struct name changes get pushed. I just have one question which is what the actual `top/bottom/left/right` correspond to. Macroblocks? Pixels?

derrod:
> I just have one question which is what the actual `top/bottom/left/right` correspond to. Macroblocks? Pixels?

Offset from top and left (i.e. column/row `0`) in pixels.

Lain-B (approved):
Looks good as far as I can tell. Very nice/interesting feature for us that will benefit all users. Question will be UI/UX down the road.

Loukious:
A question. Does this affect bitrate/max bitrate from the settings? Seems like after using this features bitrate would sometimes spike.

ma3uk:
> A question. Does this affect bitrate/max bitrate from the settings? Seems like after using this features bitrate would sometimes spike.

Yes, there is an impact on the maximum bitrate. The higher the ROI priority and the larger the ROI area, the higher the bitrate will be

Loukious:
> Yes, there is an impact on the maximum bitrate. The higher the ROI priority and the larger the ROI area, the higher the bitrate will be 
Wouldn't it make more sense if the maximum bitrate isn't touched and still respected even with this feature or do I have to change this myself (maybe making non roi lower priority)?

ma3uk:
> > Да, есть влияние на максимальный битрейт. Чем выше приоритет ROI и чем больше область ROI, тем выше будет битрейт Разве не было бы разумнее, если бы максимальный битрейт не был затронут и по-прежнему соблюдался даже с этой функцией, или я должен изменить это сам (возможно, сделать неROI более низким приоритетом)?

The ROI functionality is actually a very situational thing, it does not work wonders and to improve the quality of the ROI zone with a limited bitrate, the quality of the nonROI zone will deteriorate - these zones will be allocated less bitrate, if the bitrate selected from nonROI is not enough, the maximum bitrate will be increased.

Loukious:
> The ROI functionality is actually a very situational thing, it does not work wonders and to improve the quality of the ROI zone with a limited bitrate, the quality of the nonROI zone will deteriorate - these zones will be allocated less bitrate, if the bitrate selected from nonROI is not enough, the maximum bitrate will be increased.

I found an issue with this. Not sure if it's the UI implementation or the ROI itself. If any rescaling is enabled for the output, it breaks the coordination of the ROI zones. 
Example:
 My native resolution is 2880x1620. If I enable rescaling on the camera and the middle of the screen, the camera becomes completely distorted (even lower bitrate than non-ROI zones) and the  middle of screen zone is off it's places.

PR #10048, opened 2023-12-28
UI: Add --config-dir option to set config path

ori-sky:
### Description

This patch adds a new option `--config-dir <path>` to allow the config path used by OBS to be set on run. This new functionality can be seen as a generalization of portable mode, and the patch accordingly reimplements the portable mode option in terms of the config path option.

When the `--config-dir` option is specified, OBS now acts as if it is portable mode -- that is to say, checks against the now removed `portable_mode` flag are now done against `opt_config_path` being empty or not.

***Note:*** This change is non-breaking as `--portable` has not been removed but has instead been updated to work in terms of the `opt_config_path` string.

### Motivation and Context

The two key motivations for this patch are as follows:

1. Portable mode is unsupported on macOS. This patch does not change this behaviour, and `--portable` will still be hidden when running on macOS with `--help`, but `--config-dir` can be used on macOS without issue, enabling creators of portable OBS packages (e.g. for specific events) to target macOS with ease.
2. Running multiple instances of different portable OBS packages normally requires duplicating the program data for each portable instance. This patch allows the same instance to be run with different `--config-dir` values, avoiding duplication of identical data.

### How Has This Been Tested?

* Tested on macOS Sonoma 14.0 with various config paths being passed on the command line, as well as no config path being passed and seeing that the default non-portable-mode config path is used.
* Additionally tested on Windows 11 with various configurations of the `--portable` and `--config-dir` flags.

### Types of changes

### Checklist:

### DIFF
--- a/UI/obs-app.cpp
+++ b/UI/obs-app.cpp
@@ -86,7 +86,6 @@ static string currentLogFile;
 static string lastLogFile;
 static string lastCrashLogFile;
 
-bool portable_mode = false;
 bool steam = false;
 bool safe_mode = false;
 bool disable_3p_plugins = false;
@@ -108,6 +107,7 @@ bool opt_disable_missing_files_check = false;
 string opt_starting_collection;
 string opt_starting_profile;
 string opt_starting_scene;
+string opt_config_path;
 
 bool restart = false;
 bool restart_safe = false;
@@ -1736,7 +1736,10 @@ bool OBSApp::OBSInit()
 
 	blog(LOG_INFO, "Qt Version: %s (runtime), %s (compiled)", qVersion(),
 	     QT_VERSION_STR);
-	blog(LOG_INFO, "Portable mode: %s", portable_mode ? "true" : "false");
+	blog(LOG_INFO, "Portable mode: %s",
+	     opt_config_path.empty() ? "false" : "true");
+	blog(LOG_INFO, "Config path: %s",
+	     opt_config_path.empty() ? "default" : opt_config_path.c_str());
 
 	if (safe_mode) {
 		blog(LOG_WARNING, "Safe Mode enabled.");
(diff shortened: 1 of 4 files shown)


alinsavix:
Would `--configdir` maybe be a better name for this option? Generally when I see an option named like `--config` I'm expecting to need to specify a user-created config file, rather than a directory where all the goods will live but aren't created by the user. I'm not sure that's quite the right thing to use either, but it feels slightly better. Maybe `--datadir`?

ori-sky:
> Would `--configdir` maybe be a better name for this option? Generally when I see an option named like `--config` I'm expecting to need to specify a user-created config file, rather than a directory where all the goods will live but aren't created by the user. I'm not sure that's quite the right thing to use either, but it feels slightly better. Maybe `--datadir`?

I wondered if `--config-path` would be a good name when choosing `--config`, as I'm aware that other flags use a dash to delimit multiple words, so perhaps one of `--data-dir`, `--config-dir` or `--config-path` would be a good choice. I'm happy to change to whatever is most appropriate, I just went with config since the code internally calls it the config path.

ori-sky:
Provisionally changed the flag name from `--config` to `--config-dir`.

PatTheMav:
To add some context: "Real" portable mode (or a customisable config file directory) is unsupported on macOS on purpose because it's incompatible with macOS' application model in several ways.

Alas until we can properly decouple application settings from profile/scene collection settings (which we _should_ do), providing a custom directory hint seems to be the only way to enable users to store/backup this data in custom locations however.

Enabling OBS to load profiles and scene collections from a customisable location would require quite the effort (especially with regards to path references in them), so I'm fine with this program argument as a work-around for the foreseeable future.

(Ideally OBS would be able to create a fully self-contained "project bundle" which can be copied and moved around at will without breaking anything, but that too would require a lot of work and migrating "legacy" configurations is potentially one hell of a headache).

Warchamp7:

Issue #9864, opened 2023-11-15
Virtual Camera on OBS 30 on MacOS consuming 40% CPU


ben-gineer:
### Operating System Info

macOS 13

### OBS Studio Version

30.0.0

### OBS Studio Log URL

https://obsproject.com/logs/JNiFtFBeGtaAzKqu

### Expected Behavior

I don't expect the CPU to run at 40%, which causes the fan to run continuously

### Current Behavior

The CPU runs at 40%, and the fan continuously runs. This didn't happen with OBS 29.

Including spindump too.

[Spindump.txt](https://github.com/obsproject/obs-studio/files/13362374/Spindump.txt)

### Steps to Reproduce

1. Start OBS
2. Start Virtual Camera
3.
...

PatTheMav:
Is this 40% in OBS or 40% in activity display? Also for which process?

ben-gineer:
This is 40% for OBS Studio within Activity display:
<img width="764" alt="Screenshot 2023-11-15 at 13 32 44" src="https://github.com/obsproject/obs-studio/assets/4173590/e087710c-95bd-4463-a85f-4501464f8a34">

RytoEX:
Was fixed by #10049.

PR #10049, opened 2023-12-28
mac-virtualcam: Reduce excessive polling for new sample buffers

jcm93:
### Description

The function that [consumes sample buffers](https://developer.apple.com/documentation/coremediaio/cmioextensionstream/3915931-consumesamplebuffer) in the virtual camera's sink stream appears to be non-blocking. Since the completion handler in OBS calls the parent function directly, this causes the completion handler to be called very frequently, predominantly with a `nil` `CMSampleBuffer`. This PR is one attempt at addressing the amount of calls, modifying the callback to call `consumeBuffer` in tandem with the virtual camera's framerate, via ~~`DispatchQueue`'s `asyncAfter` method~~ a second `DispatchSourceTimer`.

### Motivation and Context

The background CPU usage from the excessive amount of calls to `consumeBuffer` adds up, with various Mac OBS users noting an increase in CPU use of the new virtual camera, and `com.apple.cmio.registerassistantservice`'s usage in particular. See https://github.com/obsproject/obs-studio/issues/9864, etc.

### How Has This Been Tested?

Tested locally on my M1 Max MacBook Pro running 14.1.2. Prior to the change, on my machine, `consumeBuffer` was being called between 20,000 and 30,000 times per second, with `com.apple.cmio.registerassistantservice` CPU use around 50%. With the change, the function is called nearer to the prescribed framerate, with `com.apple.cmio.registerassistantservice` CPU usage between ~~0-1%~~ 2-3%. Visually, I monitored the virtual camera output and did not note any significant increase in dropped frames/decreased framerate.

### Other
Documentation is thin on the ground for the best particular way to handle video streams moving from one process to another via CMIO and sink streams. Idiomatically, there may be a better way to handle this. I thought about CFNotifications when frames are prepared on the OBS end, or more precise timer shenanigans. I'm totally fine if this PR is just tracking for a better solution. Since I was tinkering around already, I merely figured I may as well submit this PR, as it does at least do the job.

### Types of changes

### Checklist:

### DIFF
--- a/plugins/mac-virtualcam/src/camera-extension/OBSCameraDeviceSource.swift
+++ b/plugins/mac-virtualcam/src/camera-extension/OBSCameraDeviceSource.swift
@@ -23,7 +23,8 @@ class OBSCameraDeviceSource: NSObject, CMIOExtensionDeviceSource {
     private var _streamingCounter: UInt32 = 0
     private var _streamingSinkCounter: UInt32 = 0
 
-    private var _timer: DispatchSourceTimer?
+    private var _placeholderTimer: DispatchSourceTimer?
+    private var _consumeBufferTimer: DispatchSourceTimer?
 
     private let _timerQueue = DispatchQueue(
         label: "timerQueue",
@@ -137,15 +138,15 @@ class OBSCameraDeviceSource: NSObject, CMIOExtensionDeviceSource {
 
         _streamingCounter += 1
 
-        _timer = DispatchSource.makeTimerSource(flags: .strict, queue: _timerQueue)
+        _placeholderTimer = DispatchSource.makeTimerSource(flags: .strict, queue: _timerQueue)
 
-        _timer!.schedule(
+        _placeholderTimer!.schedule(
             deadline: .now(),
             repeating: 1.0 / Double(OBSCameraFrameRate),
             leeway: .seconds(0)
         )
 
-        _timer!.setEventHandler {
+        _placeholderTimer!.setEventHandler {
             if self.sinkStarted {
                 return
             }
@@ -217,9 +218,9 @@ class OBSCameraDeviceSource: NSObject, CMIOExtensionDeviceSource {
                 }
             }
         }
-        _timer!.setCancelHandler {}
+        _placeholderTimer!.setCancelHandler {}
 
-        _timer!.resume()
+        _placeholderTimer!.resume()
     }
 
     func stopStreaming() {
@@ -228,9 +229,9 @@ class OBSCameraDeviceSource: NSObject, CMIOExtensionDeviceSource {
         } else {
             _streamingCounter = 0
 
-            if let timer = _timer {
+            if let timer = _placeholderTimer {
                 timer.cancel()
-                _timer = nil
+                _placeholderTimer = nil
             }
         }
     }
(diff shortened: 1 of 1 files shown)


plugins/mac-virtualcam/src/camera-extension/OBSCameraDeviceSource.swift
```diff
@@ -267,7 +268,9 @@ class OBSCameraDeviceSource: NSObject, CMIOExtensionDeviceSource {
 
                 self._streamSink.stream.notifyScheduledOutputChanged(output)
             }
-            self.consumeBuffer(client)
+            DispatchQueue.main.asyncAfter(deadline: now + (1.0 / Double(OBSCameraFrameRate))) {
```
PatTheMav, line 271: To ensure that any available sample buffer is provided in time for the output frame rate, this should use twice the frame rate.

This is obviously a trade-off between frame pacing and CPU usage, but I'd rather check if CPU usage is low enough for good pacing at first.

plugins/mac-virtualcam/src/camera-extension/OBSCameraDeviceSource.swift
PatTheMav, line 1: Using async dispatch is usually discouraged for short-term tasks (Apple's Dispatch engineers themselves suggest using Dispatch for long-running tasks).

Instead I'd suggest using a timer at twice the camera frame rate (similar to the stream output) and schedule the callback on every tick of the timer.

PatTheMav: Alternatively we could use a single timer at double the frame rate and handle _both_ (but only output a frame at every `tock` of the timer), but I'm unsure if that's more efficient than two separate timers.

plugins/mac-virtualcam/src/camera-extension/OBSCameraDeviceSource.swift
jcm93: Modified to fire every `1.0 / (Double(OBSCameraFrameRate) * 2.0)` seconds.

plugins/mac-virtualcam/src/camera-extension/OBSCameraDeviceSource.swift
jcm93: Added another DispatchSourceTimer and renamed the existing timer for readability.

PatTheMav (approved):
Change looks fine at first glance, will need to run some tests locally.

PatTheMav:
Tested this locally and the heaviest stack is indeed OBS' own render loop again (the virtual camera itself as well as QuickTime playing back the feed and everything in between is well below that).

PatTheMav:
One unfortunate side-effect in my tests is the frame-pacing however - even at twice the render frame rate there are clearly some missed/skipped frames in the output - the current implementation obviously has perfect frame-pacing (but at the cost of the immense CPU load), so maybe there is room to increase the multiplier a bit to achieve the same while still at the fraction of the current CPU usage.

jcm93:
It definitely feels like there should be a better way to do this. It seems surprising that there is no built-in messaging system for the extension to be notified when the sink has received a frame, or anything like that, and the underlying buffers certainly don't seem to be easily observable. I'll try and poke around more and see if I can think of a performant solution with better pacing.

PatTheMav:
> It definitely feels like there should be a better way to do this. It seems surprising that there is no built-in messaging system for the extension to be notified when the sink has received a frame, or anything like that, and the underlying buffers certainly don't seem to be easily observable. I'll try and poke around more and see if I can think of a performant solution with better pacing.

Yeah the missing piece is contained in their original presentation:

> Clients feed your sink stream by inserting sample buffers into a simple queue. That translates to a consumeSampleBuffer call in your extension, and once you've consumed that buffer, you notify them with notifyScheduledOutputChanged.

I haven't found a way to receive/handle the notification sent by `notifyScheduledOutputChanged` (which would suggest that the camera extension signals OBS that it's ready for another frame). The header also suggests that this is the only available notification.

On the developer forums the official response was to "avoid XPC altogether" (which would include any bespoke ways to have OBS communicate directly with the camera extension).

The name "sink" is also a bit of a hint that OBS is supposed to just dump frames into a buffer and the camera extension will handle them at its own discretion.

jcm93:
Yeah to me it seemed like that notification was traveling in the wrong direction for it to be useful, but I also just might not have my head fully wrapped around the way Apple wants people to use this system yet.

PatTheMav:
I just tested with setting the consumer frame rate to 3x the display frame rate and had no frame pacing issues - give it a whirl. CPU consumption was fine in my opinion.

jcm93:
Yeah I agree that 3.0x seems to offer an improvement in terms of pacing over 2.0x, and I did not actually register any increase in CPU use over 2.0x on my machine either, for whatever reason. I tried 4 and 5 and saw little to no further benefit, so I've pushed an update at 3x.

If I really squint, I also do see a decent amount of imperfect frame pacing even without this PR in all default receivers (QuickTime, Photo Booth, FaceTime). I don't have a 100% objective measuring device, but to my eyes at least, it really doesn't seem like some paradigm of perfect pacing is being lost here, in case that's a worry for anyone reading.

Since I'm still working on my own independent implementation of a virtual camera in my own program, I'll make a note to return to this if I end up with any kind of better solution.

PatTheMav:
I think 3x the display frame rate is good enough - perceptually animations are fluid and that's what users can expect from the feature, even if very precise frame counting might yield a limited amount of imperfect frame times.

Lain-B (approved):

PR #10052, opened 2023-12-30
libobs: Use UUIDs for all contexts

cg2121:
### Description
This adds uuids to outputs, encoders and services.

### Motivation and Context
This will help me in the output wrapper PR (#8918), so I can store/find outputs by their uuid. UUIDs are a better way of looking up contexts, as the uuid never changes and is always unique.

These are currently not using hash tables. As there are far less of these than sources, lookup times shouldn't be impacted.

### How Has This Been Tested?
Retrieved UUID from virtual camera output and also found that output by that UUID. Haven't tested with encoders or services but the code is basically the same, so they should also work.

### Types of changes

### Checklist:

### DIFF
--- a/docs/sphinx/reference-core.rst
+++ b/docs/sphinx/reference-core.rst
@@ -382,6 +382,15 @@ Libobs Objects
 
 ---------------------
 
+.. function:: obs_output_t *obs_get_output_by_uuid(const char *uuid)
+
+   Gets an output by its UUID.
+
+   Increments the output reference counter, use
+   :c:func:`obs_output_release()` to release it when complete.
+
+---------------------
+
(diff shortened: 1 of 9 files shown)


derrod (changes requested):
In documentation and commit title "UUID" should be all-uppercase as it's an acronym, otherwise seems fine though. Using a hash table for everything shouldn't be a ton of overhead but I agree it's probably overkill.

WizardCM:
Please capitalise UUID in all places where possible (commit title, documentation, comments).

cg2121:
UUID strings are now all capitalized.

cg2121:
Closing, as UUIDs are only currently useful for sources.

Issue #10026, opened 2023-12-20
there is no way to change an output timer without stopping your stream


marktheminer:
### Operating System Info

Windows 11

### OBS Studio Version

30.0.2

### OBS Studio Log URL

n/a

### Expected Behavior

If you start an output timer, you should be able to stop the timer without stopping your stream.  However, stopping the timer stops the stream (and with no warning).  In addition, you can't change the timer because to change it you have to press enter which then triggers the stop timer button which then stops the stream.

### Current Behavior

If you start an output timer, you should be able to stop the timer without stopping your stream.  However, stopping the timer stops the stream (and with no warning).  In addition, you can't change the timer because to change it you have to press enter which then triggers the stop timer button which then stops the stream.

### Steps to Reproduce

1.  open the stream timer, set it to 30 minutes and press start
2.  your stream goes live (which is also kind of odd, seems like it should be disabled until you start your stream, but that's an aside)
3.  try to stop the timer or change the value to 31 minutes
...

marktheminer:
https://obsproject.com/forum/threads/how-to-cancel-or-change-the-output-timer-once-started.137102/

gxalpha:
This is working as designed. Arguably, this isn't a great design, but this isn't a bug.

As a note, we're looking at (potentially!) completely redesigning this feature, but can't give any more details or an ETA for that at this time.

PR #10055, opened 2023-12-30
output-timer: Fixed buttons so that stopping a timer does not stop stream/recording

marktheminer:
### Description

I made it so that stopping a stream/recording timer does not cause the stream/recording to stop.

This was done by adding two member variables to the OutputTimer class and by then referencing them appropriately at the correct places.

### Motivation and Context
Multiple users including myself have had this happen while streaming when we set our timer and decide either that we want to stream longer or just set the timer incorrectly and need to change it.

https://obsproject.com/forum/threads/how-to-cancel-or-change-the-output-timer-once-started.137102/

https://github.com/obsproject/obs-studio/issues/10026

### How Has This Been Tested?
I compiled and ran it and did the following tests:

1. Start the stream timer, make sure it starts the stream as before
2. Stop the stream timer, make sure it doesn't stop the stream and still stops the timer
3. Restart the timer and make sure it actually stops the stream when the time runs out

I did the same for the recording timer.

### Types of changes

### Checklist:

### DIFF
--- a/UI/frontend-plugins/frontend-tools/output-timer.cpp
+++ b/UI/frontend-plugins/frontend-tools/output-timer.cpp
@@ -53,12 +53,14 @@ void OutputTimer::StreamingTimerButton()
 	if (!obs_frontend_streaming_active()) {
 		blog(LOG_INFO, "Starting stream due to OutputTimer");
 		obs_frontend_streaming_start();
-	} else if (streamingAlreadyActive) {
+	} else if (streamingAlreadyActive ||
+		   !this->streamingTimer->isActive()) {
 		StreamTimerStart();
 		streamingAlreadyActive = false;
 	} else if (obs_frontend_streaming_active()) {
-		blog(LOG_INFO, "Stopping stream due to OutputTimer");
-		obs_frontend_streaming_stop();
+		stoppingStreamTimer = true;
+		StreamTimerStop();
+		stoppingStreamTimer = false;
 	}
 }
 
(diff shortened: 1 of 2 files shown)


therentabrain:
Thank you, thank you, thank you.
I don't see a way to upvote this for attention, so I comment with my thanks.

marktheminer:
You are welcome!

I see any interruption of stream as a major issue.  At minimum, if this is not the right fix, it should prompt, "This will stop your stream, do you want to continue?"

GeorgesStavracas (approved):
UI / UX review: the description of the new behavior makes sense. This needs testing to validate the behaviour. The Output Timer dialog needs serious reconsideration.

marktheminer:
@GeorgesStavracas after this merges, would you be interested in coming up with a design that re-worked the issues you see with the dialog as well as adding this?

https://github.com/obsproject/obs-studio/pull/10103

I'm wondering if with a redesign if that could allow me to have that feature added, but in a different UI.

Maybe a tab for streaming and one for recording, or a https://doc.qt.io/qt-6/qgroupbox.html for each of stream and recording and then both recording timers could go on the tab or in the group box?

GeorgesStavracas:
The various OBS contributors agreed that the priority of this dialog is low, and some even raised the idea of removing it and making it an external tool. So I'm afraid the time and effort I'm willing to put on it is severely limited. Sorry.

kjetilpp:
When will this be merged? :)

marktheminer:
@kjetilpp I don't think they are going to, but you can get it at https://obsproject.com/forum/resources/advanced-output-timer.1900/

image

Fenrirthviti:
Since this is currently available as a third party plugin, and we're not particularly keen on further changes to this dialog (and would prefer a refactor/rewrite instead), I will close this.

PR #10058, opened 2024-01-01
UI: Add continue option to crash handler

exeldro:
### Description
The default crash error of OBS on windows is:
281167276-9262df8c-228e-4552-aa8b-332ff62e77f5
If OBS crashes in a thread that is not the UI, graphics or audio thread it will try to get the thread name.
With thread name:
image
Without thread name:
image

This pull request replaces #9821, this method does not require threads to set how they want errors to be handled.

### Motivation and Context
Users get crashes in not critical threads and keep the error message open to continue using obs.
https://www.youtube.com/shorts/RMiOtRFwbAg
This way the error message can be closed when it is not the main UI, graphics or audio thread.
The user can get a bit of clue what might be broken from the thread name and decide if it wants to continue;

### How Has This Been Tested?
On windows 11 by creating errors in threads like the DShowThread, UI and graphics thread.

### Types of changes

### Checklist:

### DIFF
--- a/UI/obs-app.cpp
+++ b/UI/obs-app.cpp
@@ -2618,9 +2618,19 @@ static int run_program(fstream &logFile, int argc, char *argv[])
 
 #ifdef _WIN32
 
+typedef HRESULT(WINAPI *GETTHREADDESCRIPTION)(HANDLE thread, PWSTR *desc);
+
 #define CRASH_MESSAGE                                                      \
 	"Woops, OBS has crashed!\n\nWould you like to copy the crash log " \
 	"to the clipboard? The crash log will still be saved to:\n\n%s"
+#define CRASH_THREAD_MESSAGE                                           \
+	"Woops, a non critical component of OBS has crashed! "         \
+	"Would you like to try and continue in an unstable state?\n\n" \
+	"The crash log will still be saved to:\n\n%s"
+#define CRASH_NAMED_THREAD_MESSAGE                                     \
+	"Woops, a non critical component of OBS has crashed (%s)! "    \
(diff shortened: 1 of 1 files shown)


notr1ch:
I have the same concerns as #9821, continuing from a crashed state is really not safe. Yes, it might work most of the time but I really don't like the idea of telling users a crash is recoverable. If the crash is non-fatal then why would a user choose to kill OBS? And if the choice is always to continue then why present a choice at all?

Fenrirthviti:
To play devil's advocate here, the point of providing the choice is mostly to inform the user something broke, but not enough to take down all of OBS, so they should find a convenient point to restart the program.

This feels like something that could slot in to a proper user feedback system more than presenting as a hard crash in those instances where we're confident continuing is at least an option, but we don't have that yet.

exeldro:
Added logging to make sure in the OBS log file can be seen that the user had an error and what the user selected to proceed.

WizardCM:
My concerns in #9821 also carry over to this implementation - without reading the message, pressing 'Yes' changes behaviour depending on the level of crash.

Warchamp7:
> My concerns in #9821 also carry over to this implementation - without reading the message, pressing 'Yes' changes behaviour depending on the level of crash.

I echo this concern. Do we have the ability to change the labels of the buttons in this pop up?

Warchamp7:
After further discussion, we do not think it is wise to allow continuing in an undefined state. The fact that it is currently possible in some fashion while the dialog is visible but has not been interacted with is already a concern that we would not like to make worse.

PR #10065, opened 2024-01-03
mac-capture: Update capture to include menu bar and child windows

PatTheMav:
### Description
Explicitly enable capture of the menu bar and child windows for application capture on macOS.

### Motivation and Context
macOS 14.2 changed internal defaults of SCK requiring the inclusion of the menu bar and child windows to be enabled explicitly.

This will have a slight negative impact on capture performance but is required to restore behaviour of prior macOS versions.

### How Has This Been Tested?
Tested on macOS 14.2.1.

### Types of changes

### Checklist:

### DIFF
--- a/plugins/mac-capture/mac-sck-video-capture.m
+++ b/plugins/mac-capture/mac-sck-video-capture.m
@@ -160,6 +160,10 @@ static bool init_screen_stream(struct screen_capture *sc)
 
                 [sc->stream_properties setWidth:(size_t) target_window.frame.size.width];
                 [sc->stream_properties setHeight:(size_t) target_window.frame.size.height];
+
+                if (@available(macOS 14.2, *)) {
+                    [sc->stream_properties setIncludeChildWindows:YES];
+                }
             }
         } break;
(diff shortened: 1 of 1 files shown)


plugins/mac-capture/mac-sck-video-capture.m
```diff
@@ -160,6 +160,10 @@ static bool init_screen_stream(struct screen_capture *sc)
 
                 [sc->stream_properties setWidth:(size_t) target_window.frame.size.width];
                 [sc->stream_properties setHeight:(size_t) target_window.frame.size.height];
+
+                if (@available(macOS 14.2, *)) {
+                    [sc->stream_properties setIncludeChildWindows:YES];
+                }
```
gxalpha, line 166: According to the header this should be enabled by default, is that not actually the case?

PatTheMav: True, but I'd rather have it explicitly enabled because that's what users expect right now and avoiding a similar situations where macOS' internal defaults change.

Lain-B (approved):

PR #10067, opened 2024-01-03
UI: Switch to bigger units for high bitrate/large recordings

derrod:
### Description

Switch output bitrate and file size in stats dock to larger units when appropiate.

| Before | After |
| - | - |
| before | after  |

### Motivation and Context

Suggested by a couple people and seems to make sense.

### How Has This Been Tested?

See screenshots above.

### Types of changes

### Checklist:

### DIFF
--- a/UI/window-basic-stats.cpp
+++ b/UI/window-basic-stats.cpp
@@ -546,10 +546,20 @@ void OBSBasicStats::OutputLabels::Update(obs_output_t *output, bool rec)
 	setThemeID(status, themeID);
 
 	long double num = (long double)totalBytes / (1024.0l * 1024.0l);
+	const char *unit = "MiB";
+	if (num > 1024) {
+		num /= 1024;
+		unit = "GiB";
+	}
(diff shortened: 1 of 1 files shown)


alinsavix:
I wonder if this will lead to user confusion and/or support burden, given that pretty much everything streaming-related tends to use kb/s as their only unit -- and I can guarantee (from experience) that there's a whole bunch of users that don't know that "4500kb/s" is the same as "4.5Mb/s".

This is mitigated slightly by only having the new units used on the stats dock, I suppose, but I'm not sure there still won't be a nontrivial number of people going "I set it for 4500 and I'm getting 4!"

(I'm definitely not against the idea (not that it would matter if I were!), but this concern is probably at least worth thinking about)

Fenrirthviti:
This will only convert to Mb/s after 10,000 kb/s, so it shouldn't affect most real-world use cases for streaming.

Lain-B (approved):
I approve of this, but will wait on UI/UX to make the final call.

Warchamp7 (approved):

PR #433, opened 2015-06-02
Add openSUSE build instructions to INSTALL.

boombatower:
Assumes #429 Packman instructions are present.

PR #10076, opened 2024-01-04
obs-ffmpeg: Fix issue with B-frames introducing motion blur

rhutsAMD:
### Description

If a user sets both `AdaptiveMiniGOP=true` and `EnablePreAnalysis=true` in the AMF/FFmpeg options field, AMF will adaptively insert B-pictures, and no longer uses the fixed B pattern. Since `MAX_CONSECUTIVE_BPICTURES` is hard-coded to 3, the max B-frame UI setting is not respected. Furthermore, for a fixed B-frames pattern, it is expected that increasing B-frames can cause a quality drop for certain content such as with high motion. AdaptiveMiniGOP is recommended when using B-frames to improve the quality in such cases. AdaptiveMiniGOP is dependent on PreAnalysis which means that trying to enable it without having PreAnalysis turned ON will have no negative effect (AdaptiveMiniGOP won't be enabled).

Additionally, if a UI element for enabling PreAnalysis such as a checkbox is added, it can be used to check if both B-frames > 0 and PreAnalysis is ON before trying to enable AdaptiveMiniGOP.

### Motivation and Context

This PR addresses [GPUOpen/AMF Issue \#433](https://github.com/GPUOpen-LibrariesAndSDKs/AMF/issues/433) regarding quality in OBS recording.

### How Has This Been Tested?

Tested on Windows 10 with an AMD Radeon RX 7900 XTX as well as a second machine where output bitstreams were inspected.

### Types of changes

### Checklist:

### DIFF
--- a/plugins/obs-ffmpeg/texture-amf.cpp
+++ b/plugins/obs-ffmpeg/texture-amf.cpp
@@ -1294,8 +1294,16 @@ static bool amf_avc_init(void *data, obs_data_t *settings)
 	int64_t bf = obs_data_get_int(settings, "bf");
 
 	if (enc->bframes_supported) {
-		set_avc_property(enc, MAX_CONSECUTIVE_BPICTURES, 3);
+		set_avc_property(enc, MAX_CONSECUTIVE_BPICTURES, bf);
 		set_avc_property(enc, B_PIC_PATTERN, bf);
+
+		/* AdaptiveMiniGOP is suggested for some types of content such
+		 * as those with high motion. This only takes effect if
+		 * Pre-Analysis is enabled.
+		 */
+		if (bf > 0) {
+			set_avc_property(enc, ADAPTIVE_MINIGOP, true);
+		}
 
 	} else if (bf != 0) {
 		warn("B-Frames set to %lld but b-frames are not "
--- a/plugins/obs-ffmpeg/texture-amf.cpp
+++ b/plugins/obs-ffmpeg/texture-amf.cpp
@@ -1066,6 +1066,7 @@ static void amf_defaults(obs_data_t *settings)
 	obs_data_set_default_string(settings, "rate_control", "CBR");
 	obs_data_set_default_string(settings, "preset", "quality");
 	obs_data_set_default_string(settings, "profile", "high");
+	obs_data_set_default_int(settings, "bf", 3);
 }
 
 static bool rate_control_modified(obs_properties_t *ppts, obs_property_t *p,


RytoEX:
The original commit where this was hard-coded to 3 says:
> The AMF documentation states that this value is recommended to be 3 if the bframe picture pattern property is a non-zero value.

Checking the AMF docs, this is still true as of v1.4.32:
> Maximum number of consecutive B Pictures, suggestion set to `3` if `AMF_VIDEO_ENCODER_B_PIC_PATTERN` is not `0`.

https://github.com/GPUOpen-LibrariesAndSDKs/AMF/blob/v1.4.32/amf/doc/AMF_Video_Encode_API.pdf
https://github.com/GPUOpen-LibrariesAndSDKs/AMF/blob/v1.4.32/amf/doc/AMF_Video_Encode_API.md

Also, according to the docs, `MAX_CONSECUTIVE_BPICTURES` has a range of 0..3 (maximum value is 3). If the user sets b-frames to 4 or 5, this will set `MAX_CONSECUTIVE_BPICTURES` to a value higher than 3, which should not be permitted according to the docs.

Johl7:
The issue is that even if maxbfs set to 0 in de UI, OBS still seems to hard set it to 3. Sorry if i got somethin wrong.

rhutsAMD:
> The original commit where this was hard-coded to 3 says:
> 
> > The AMF documentation states that this value is recommended to be 3 if the bframe picture pattern property is a non-zero value.
> 
> Checking the AMF docs, this is still true as of v1.4.32:
> 
> > Maximum number of consecutive B Pictures, suggestion set to `3` if `AMF_VIDEO_ENCODER_B_PIC_PATTERN` is not `0`.
> 
> https://github.com/GPUOpen-LibrariesAndSDKs/AMF/blob/v1.4.32/amf/doc/AMF_Video_Encode_API.pdf https://github.com/GPUOpen-LibrariesAndSDKs/AMF/blob/v1.4.32/amf/doc/AMF_Video_Encode_API.md
> 
> Also, according to the docs, `MAX_CONSECUTIVE_BPICTURES` has a range of 0..3 (maximum value is 3). If the user sets b-frames to 4 or 5, this will set `MAX_CONSECUTIVE_BPICTURES` to a value higher than 3, which should not be permitted according to the docs.

Thank you for noticing this in the AMF documentation. I have updated the suggestion regarding this property mentioning PA + AdaptiveMiniGOP in the doc and it will be publicly visible in the next release.

In reality, the number of B-frames can be set beyond 3 for special use cases or experiments, but for practicality we want to encourage no higher than 3 without forcefully limiting the user. I have included a change to default the number of B-frames to 3 for AVC.

RytoEX:
> I have included a change to default the number of B-frames to 3 for AVC.

The function that you modified also sets defaults for HEVC. While I don't think the b-frames setting is ever read in that path, something to keep in mind.

RytoEX (changes requested):
Incidentally, the AMF docs say that `AMF_VIDEO_ENCODER_B_PIC_PATTERN` is of type `amf_bool`, but that it support values 0..3. Other places in the docs say that `AMF_VIDEO_ENCODER_B_PIC_PATTERN` is of type `amf_int64`. If it does only support the range 0..3, then we probably should not allow the user to set b-frames higher than 3. This need not be addressed in this PR, but I wanted to bring it to your attention.

rhutsAMD:
> Incidentally, the AMF docs say that `AMF_VIDEO_ENCODER_B_PIC_PATTERN` is of type `amf_bool`, but that it support values 0..3. Other places in the docs say that `AMF_VIDEO_ENCODER_B_PIC_PATTERN` is of type `amf_int64`. If it does only support the range 0..3, then we probably should not allow the user to set b-frames higher than 3. This need not be addressed in this PR, but I wanted to bring it to your attention.

Thanks for noticing the typo in the doc. I have updated it internally and the correction will be publicly visible in the next AMF release.

plugins/obs-ffmpeg/texture-amf.cpp
```diff
@@ -1294,9 +1295,15 @@ static bool amf_avc_init(void *data, obs_data_t *settings)
 	int64_t bf = obs_data_get_int(settings, "bf");
 
 	if (enc->bframes_supported) {
-		set_avc_property(enc, MAX_CONSECUTIVE_BPICTURES, 3);
+		set_avc_property(enc, MAX_CONSECUTIVE_BPICTURES, bf);
 		set_avc_property(enc, B_PIC_PATTERN, bf);
```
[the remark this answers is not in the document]
rhutsAMD: Both are required to be set in order to use B-frames for AVC. `B_PIC_PATTERN` serves to set the number of B-frames while `MAX_CONSECUTIVE_BPICTURES` limits the maximum. `MAX_CONSECUTIVE_BPICTURES` will play an active role when used in combination with PA and AdaptiveMiniGOP where PA analyzes the input content and decides the picture type of each frame adaptively so the pattern is variable instead of a fixed number equal to `B_PIC_PATTERN` in each GOP. Yes, both the `MAX_CONSECUTIVE_BPICTURES` and `B_PIC_PATTERN` properties will be set to the same value, however, the number of B-frames will vary up to the specified maximum when PA and AdaptiveMiniGOP are enabled as well.

RytoEX (approved):
This is probably fine.

plugins/obs-ffmpeg/texture-amf.cpp
```diff
@@ -1294,9 +1295,16 @@ static bool amf_avc_init(void *data, obs_data_t *settings)
...
 
+		/* Suggested for some types of content such
+		 * as those with high motion
+		 */
```
RytoEX, line 1303:
```suggestion
		/* AdaptiveMiniGOP is suggested for some types of content such
		 * as those with high motion.
		 */
```

RytoEX (changes requested):
Please also squash the comment fix commits into the commit that introduced the comment.

plugins/obs-ffmpeg/texture-amf.cpp
```diff
@@ -1294,9 +1295,16 @@ static bool amf_avc_init(void *data, obs_data_t *settings)
 	int64_t bf = obs_data_get_int(settings, "bf");
 
 	if (enc->bframes_supported) {
-		set_avc_property(enc, MAX_CONSECUTIVE_BPICTURES, 3);
+		set_avc_property(enc, MAX_CONSECUTIVE_BPICTURES, bf);
 		set_avc_property(enc, B_PIC_PATTERN, bf);
 
+		/* AdaptiveMiniGOP is suggested for some types of content such
+		 * as those with high motion.
+		 */
```
RytoEX, line 1303: Add a note about this only taking effect if Pre-Analysis is enabled.
```suggestion
		/* AdaptiveMiniGOP is suggested for some types of content such
		 * as those with high motion. This only takes effect if
		 * Pre-Analysis is enabled.
		 */
```

RytoEX (changes requested):
Hopefully final nits. Please make the first commit message more descriptive. In particular, add a note to the commit message body about AdaptiveMiniGOP and Pre-Analysis so that git blame/log is more informative.

plugins/obs-ffmpeg/texture-amf.cpp
rhutsAMD: Added

rhutsAMD:
> Hopefully final nits. Please make the first commit message more descriptive. In particular, add a note to the commit message body about AdaptiveMiniGOP and Pre-Analysis so that git blame/log is more informative.

Added

RytoEX (changes requested):
> > Hopefully final nits. Please make the first commit message more descriptive. In particular, add a note to the commit message body about AdaptiveMiniGOP and Pre-Analysis so that git blame/log is more informative.
> 
> Added

Per our [Commit Guidelines](https://github.com/obsproject/obs-studio/blob/e284a79b48f5524a37a6e515667c8bedb0bce3f0/CONTRIBUTING.rst#commit-guidelines), you must wrap (line break) the commit message body at 72 characters.

rhutsAMD:
> > > Hopefully final nits. Please make the first commit message more descriptive. In particular, add a note to the commit message body about AdaptiveMiniGOP and Pre-Analysis so that git blame/log is more informative.
> > 
> > 
> > Added
> 
> Per our [Commit Guidelines](https://github.com/obsproject/obs-studio/blob/e284a79b48f5524a37a6e515667c8bedb0bce3f0/CONTRIBUTING.rst#commit-guidelines), you must wrap (line break) the commit message body at 72 characters.

Updated

CortexReaver:
I'll just duplicate my post from AMF github:

> Is this still not being fixed in OBS 30.1.0-beta3? Because I still have ghosting artifacts... Especially when streaming Open Red Alert:
> https://youtu.be/miAUGIbqFbE
> Look how units leaving traces behind them while moving. Also take a look at the shroud disappearing.
> VBR seems like doesn't have this bug, but the whole screen is blinking there like its refreshing - from top to bottom (this is might be a different bug though). This or rather just the bottom half of the screen does blinking.

Johl7:
> I'll just duplicate my post from AMF issues to here:
> 
> > Is this still not being fixed in OBS 30.1.0-beta3? Because I still have ghosting artifacts... This is way more noticeable in grainy static games, like Open Red Alert:
> > https://youtu.be/miAUGIbqFbE
> > Look how units leaving traces behind them while moving. Also take a look at the shroud disappearing.
> > VBR seems like doesn't have this bug, but the whole screen is blinking there like its refreshing - from top to bottom (this is might be a different bug though). This or rather just the bottom half of the screen does blinking.

what gpu is urs? feels identical to somethin i had with my Rx 470

CortexReaver:
@Johl7 
> what gpu is urs? feels identical to somethin i had with my Rx 470

RX 6600

Johl7:
> @Johl7
> 
> > what gpu is urs? feels identical to somethin i had with my Rx 470
> 
> RX 6600

Yeahh, tested a bit... those keyframe pumping aint due to bframes, just an encoder limitation on lower bitrate stuff, i guess.

CortexReaver:
VBVBufferSize sets in bps and not Kbps, right? So for the 8000 Kbps it gonna be 16000000. Hope my GPU doesn't explode or something :)

RytoEX:
@rhutsAMD 
This seems to cause issues on much older GPUs/APUs where they seem to report that bframes are supported but actually do not output any b-frames when setting b-frames to non-zero. See:
* #10368

RytoEX:
See also:
* #10435

pbaykalov:
My setup:
RX6950
OBS 31.0.2
HQVBR , CBR
Setting Max B-frames to anything higher than 0 results in frame ghosting exactly as described in https://github.com/GPUOpen-LibrariesAndSDKs/AMF/issues/433

pbaykalov:
`PASceneChangeDetectionEnable=false` mostly fixes it for me.

PR #10078, opened 2024-01-05
libobs: Log Mac model identifier

jcm93:
### Description

Adds hardware model identifier strings (e.g. `MacBookPro18,2` to logs on macOS.

### Motivation and Context

Providing support or debugging issues with OBS on Mac frequently involves querying the user about their hardware beyond just the CPU or GPU. This is because the same chip in a Mac can have different capabilities inside a different chassis, and the user may need different advice accordingly. An M1 chip inside a 13" MacBook Pro has a very different thermal envelope than inside a MacBook Air, for example.

There currently isn't any way to see a user's exact model of machine from logs for Apple Silicon machines. For Intel machines, determining a user's model year currently involves looking up either their CPU or GPU model and seeing which Macs used it, which is a relatively complex series of steps to figure out what should be fairly simple. 

With this change, the model identifier is now listed conveniently in logs to streamline debugging.

### How Has This Been Tested?

Tested locally on my Apple Silicon machine. The function added is basically a carbon copy of other string-finding functions in `obs-cocoa.m`, just with a different sysctl call, so the surface area for issues is very small.

### Other

I aimed to log a more human-readable string (e.g. `iMac (Late 2017)`), but this [seems to require a non-trivial amount of effort](https://apple.stackexchange.com/questions/98080/can-a-macs-model-year-be-determined-with-a-terminal-command). 

### Types of changes

### Checklist:

### DIFF
--- a/libobs/obs-cocoa.m
+++ b/libobs/obs-cocoa.m
@@ -72,13 +72,13 @@ static void log_processor_name(void)
     if (ret != 0)
         return;
 
-    name = malloc(size);
+    name = bmalloc(size);
 
     ret = sysctlbyname("machdep.cpu.brand_string", name, &size, NULL, 0);
     if (ret == 0)
         blog(LOG_INFO, "CPU Name: %s", name);
 
-    free(name);
+    bfree(name);
 }
 
 static void log_processor_speed(void)
--- a/libobs/obs-cocoa.m
+++ b/libobs/obs-cocoa.m
@@ -93,6 +93,25 @@ static void log_processor_speed(void)
         blog(LOG_INFO, "CPU Speed: %lldMHz", freq / 1000000);
 }
 
+static void log_model_name(void)
+{
+    char *name = NULL;
+    size_t size;
+    int ret;
+
+    ret = sysctlbyname("hw.model", NULL, &size, NULL, 0);
+    if (ret != 0)
+        return;
+
+    name = bmalloc(size);
+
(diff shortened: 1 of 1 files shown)


libobs/obs-cocoa.m
```diff
@@ -93,6 +93,25 @@ static void log_processor_speed(void)
...
+{
+    char *name = NULL;
+    size_t size;
+    int ret;
+
+    ret = sysctlbyname("hw.model", NULL, &size, NULL, 0);
+    if (ret != 0)
+        return;
+
+    name = malloc(size);
```
gxalpha, line 106: Even if memory alignment isn't needed, I'd use `bmem.h`'s  `bmalloc` and `bfree` because it's used universally across OBS.
Maybe you could even use a fixed size buffer to avoid the heap allocation altogether (say of size 128 or something), since we know the string won't be infinitely long.

Also personally I'd rather do the declaration at the same place as the initialization, but I suppose that's just style preferences.

jcm93: Changed. Added a new commit to use the aligned managers in `log_processor_name()` as well.

gxalpha (approved):

PatTheMav (commented):
I expect that this will not trip up the log parser?

jcm93:
Seems fine in my testing, tested with my own identifier and a blank identifier and neither tripped up the analyzer.

PatTheMav (approved):

PR #10089, opened 2024-01-09
libobs-d3d11: Remove new/delete replacements

jpark37:
### Description
MSVC noticed inline is not allowed, and new/delete replacments also don't belong in a header file. We're also just leaning on __STDCPP_DEFAULT_NEW_ALIGNMENT__ being 16 for MSVC x64 now.

### Motivation and Context
Don't like compiler warnings or C++ standard violations.

### How Has This Been Tested?
OBS still runs and I saw it drawing stuff in the preview window with D3D11.

### Types of changes

### Checklist:


PatTheMav (approved):

Lain-B (changes requested):
Actually changed my mind, we can probably just remove the custom new call because we don't compile in 32bit anymore. Right? Because originally the whole reason why this code existed was to force new to be aligned due to SSE issues, which should no longer be present in 64bit.

RytoEX:
> Actually changed my mind, we can probably just remove the custom new call because we don't compile in 32bit anymore. Right? Because originally the whole reason why this code existed was to force new to be aligned due to SSE issues, which should no longer be present in 64bit.

We still build a 32-bit subproject for the files that require it (win-capture for game capture hooks, win-dshow for virtual camera DLLs). I'm not sure if libobs-d3d11 is included in that build. @PatTheMav ?

jpark37:
AFAIK, the function replacements only applied to libobs-d3d11, and this doesn't change that. I don't think any file outside libobs-d3d11 ever included d3d11-subsystem.h or AlignedNew.hpp.

I thought this was really to ensure allocations get tracked in the leak count at the end.

Lain-B:
It wasn't for tracking, it was for allocation alignment because it uses vector with matrices. The default heap allocator in 32bit wouldn't be guaranteed to be aligned, so it would cause a crash, and that was the motivation for creating it for libobs-d3d11. The allocation tracking was just a side benefit, but seeing as it's C++ it's not like that has any significance anyway as long as C++ is being used correctly.

My suggestion is just to remove AlignedNew.hpp and remove the include for it and not even bother doing the new/delete operator overriding in libobs-d3d11. We don't use it anywhere else other than libobs-d3d11 for that singular purpose, so it's pretty much redundant at this point.

jpark37:
Alright, update PR with all deletes.

Lain-B (approved):

PR #10091, opened 2024-01-10
UI: Implement GPU scaling options

derrod:
### Description

Adds a combobox to set GPU scaling filter to the advanced output UI:

2024-01-21_00-49-43_iWu4oS

### Motivation and Context

- Allows people to use texture encoders such as NVENC while using rescaling
- Higher qualty scaling

### How Has This Been Tested?

NVENC and x264.

### Types of changes

### Checklist:

### DIFF
--- a/UI/data/locale/en-US.ini
+++ b/UI/data/locale/en-US.ini
@@ -1037,6 +1037,7 @@ Basic.Settings.Output.NoSpaceFileName="Generate File Name without Space"
 
 # basic mode 'output' settings - advanced section
 Basic.Settings.Output.Adv.Rescale="Rescale Output"
+Basic.Settings.Output.Adv.Rescale.Disabled="Disabled"
 Basic.Settings.Output.Adv.AudioTrack="Audio Track"
 Basic.Settings.Output.Adv.Streaming="Streaming"
 Basic.Settings.Output.Adv.Streaming.Settings="Streaming Settings"
--- a/UI/forms/OBSBasicSettings.ui
+++ b/UI/forms/OBSBasicSettings.ui
@@ -2590,7 +2590,7 @@
                             <widget class="QComboBox" name="advOutEncoder"/>
                            </item>
                            <item row="4" column="0">
-                            <widget class="QCheckBox" name="advOutUseRescale">
+                            <widget class="QLabel" name="advOutUseRescale">
                              <property name="sizePolicy">
                               <sizepolicy hsizetype="Minimum" vsizetype="Expanding">
                                <horstretch>0</horstretch>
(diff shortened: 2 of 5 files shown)


UI/data/locale/en-US.ini
```diff
@@ -1037,6 +1037,12 @@ Basic.Settings.Output.NoSpaceFileName="Generate File Name without Space"
 
 # basic mode 'output' settings - advanced section
 Basic.Settings.Output.Adv.Rescale="Rescale Output"
+Basic.Settings.Output.Adv.Rescale.Default="Fast Bilinear (CPU)"
```
WizardCM, line 1040:
```suggestion
Basic.Settings.Output.Adv.Rescale.Default="Fast Bilinear (CPU, Default)"
```

gxalpha:
Are we sure this should be a setting at all? Or at least are we sure that CPU should be an option? That seems disadvantageous if we want to eventually remove the fallback encoders (which was a long-term goal according to my understanding).

derrod:
> Are we sure this should be a setting at all? Or at least are we sure that CPU should be an option? That seems disadvantageous if we want to eventually remove the fallback encoders (which was a long-term goal according to my understanding).

The goal is to remove the FFmpeg encoders, but there will probably need to be a non-texture encoder for various scenarios where sharing doesn't work (cross-GPU, Linux, whatever). We can deprecate/remove the CPU scaling once we're confident enough that the GPU scaling works for people.

RealIndrit:
"The goal is to remove the FFmpeg encoders" and "We can deprecate/remove the CPU scaling once we're confident enough that the GPU scaling works for people".

Question: 
Does this refer to the x264 (CPU) encoders and now also the CPU based output rescaling? As from this it looks like that OBS as a project is trying to go towards ending the support CPU based video streaming as a whole? What is the main reason behind this? Is it the maintaince of the code base or something else?

gxalpha:
> Does this refer to the x264 (CPU) encoders

No. x264 does not use FFmpeg. It will not be removed.

RealIndrit:
> > Does this refer to the x264 (CPU) encoders
> 
> No. x264 does not use FFmpeg. It will not be removed.
> 
> This PR (and conversation) is only about CPU _scaling_, not CPU _encoding_.

Ah ok, the "The goal is to remove the FFmpeg encoders" did confuse me a bit,

Warchamp7:
- Remove checkbox next to Rescale Output for toggling scaling
- Move the new scaling dropdown to be first before the resolution
- Add a `No scaling` option to the dropdown, equivalent to the old checkbox being off
- Update dropdown entries to start with `CPU - ` or `GPU - ` accordingly instead of ending with it
- Put CPU at bottom of the list

Lain-B (approved):

DmitriyFrogo:
I guess, something breaks, and OBS ignores VST plugins after these changes
I limit my voice to -7 dB and game output to -15 dB with LoudMax plugin
OBS mixer scale show correct limits, but when I listen to recording - it sounds like without plugins

I tried previous artifact from master branch - no issues

Previous build https://github.com/obsproject/obs-studio/commit/fb8d228a82f6a8020166e61060a5e451303493f9:
photo_2024-01-22_15-10-26

This build https://github.com/obsproject/obs-studio/commit/0a8e00c478401119309c050de7b23111b1af3ae6:
photo_2024-01-22_15-11-13

DmitriyFrogo:
@derrod

I did some tests
This issue appears only if you using new rescaling options

Steps to reproduce issue:
1. Set smaller resolution in Settings - Output - Streaming (Advanced mode) - Rescale Output
2. In Settings - Video, set both Canvas and Scaled equals your desktop resolution
3. Add Limiter filter to any audio output source (Desktop Audio or Game/Window Capture), for example -20 dB
4. Start recording/streaming and play any sound/music louder than your limiter
5. Use any software than can show you loudness range (I using free Orban Loudness Meter)
6. Loudness >-20 dB

To see the difference:
1. Set smaller than your desktop resolution in Settings - Video - Scaled resolution
2. Disable new rescaling options in Settings - Output - Streaming (Advanced mode) - Rescale Output
Steps 3-5 same as above
6. Loudness = 20 dB

Issue #4227, opened 2021-02-13
"Overwrite if file exists" setting not respected for replay buffer


Fenrirthviti:
## Platform

Operating system and version: Windows 10
OBS Studio version: 26.1.1

## Expected Behavior
OBS should respect the "Overwrite if file exists" setting for replays as well

## Current Behavior
Local recordings will create a new file with an appended number at the end of the file if the same filename already exists, whereas replay buffer will just overwrite the existing file:

![https://pub.rachni.com/img/explorer_2021-02-12_19-09-48.png](example)

## Steps to Reproduce

1. Open OBS
2. Set local recording and replay buffer filename to something static
3. Save replay buffer multiple times
4. Observe that the file is overwritten

WizardCM:
I think I might understand why this happens.

https://github.com/obsproject/obs-studio/blob/2787e63d8ecd0f4578ab47c6ca5cf73c0bce1844/UI/window-basic-main-outputs.cpp#L1946-L1947

The filename conflict check only happens when you Start replay buffer, rather than when you save. It also gets refreshed (if you tell it to) when you start Recording. So if you were to top/start replay buffer, it would correctly add  second file rather than replacing.

RytoEX:
This also happens in Simple Output Mode, even if you stop and start the replay buffer.

PR #4327, opened 2021-03-11
UI: Add checks for overwrite setting to replay buffer

offthegrid-mike:
### Description

"Overwrite if file exists" setting not respected for replay buffer

### Motivation and Context

Refer to issue #4227
The reason behind this is there are no file overwritten checkings when saving replay buffer.
Even there is a file name checking when starting replay buffer, the variable is unused actually
> https://github.com/obsproject/obs-studio/blob/2787e63d8ecd0f4578ab47c6ca5cf73c0bce1844/UI/window-basic-main-outputs.cpp#L1946-L1947

### How Has This Been Tested?

1. Open OBS
2. Set local recording and replay buffer filename to something static
3. Save replay buffer multiple times
4. Observe that the file is overwritten

### Types of changes

- Bug fix (non-breaking change which fixes an issue)

### Checklist:

- [x] My code has been run through [clang-format](https://github.com/obsproject/obs-studio/blob/master/.clang-format).
- [x] I have read the [**contributing** document](https://github.com/obsproject/obs-studio/blob/master/CONTRIBUTING.rst).
- [x] My code is not on the master branch.
- [x] The code has been tested.
- [x] All commit messages are properly formatted and commits squashed where appropriate.
- [x] I have included updates to all appropriate documentation.

PR #10095, opened 2024-01-11
obs-ffmpeg: Allow overwrite file control for replay buffer

Penwy:
### Description

This enables the replay buffer to respect the "Overwrite if file exists" setting when creating the filename, and saving to it.

### Motivation and Context

Currently, the replay buffer systematically overwrites if existing. This is suboptimal and fails to cover some usecases, as pointed in #4227.

A solution was proposed by #4327 but it was then reverted because it caused issues. The later addition of recording file splitting (off which this piggybacks) allows a much simpler and issue-free solution.

### How Has This Been Tested?

Ubuntu 22.04.
Both in advanced and simple mode, checked that both recording and replay buffer saving correctly respected the overwrite setting, and correctly created the file.
Given that this also modifies code used by file splitting, also checked that there was no regression on it.
Given the issues of the previous attempt outlined by f295bd9, also checked there was no regression on creating subdirectories with slashes in filename.

### Types of changes

### Checklist:


RytoEX:
Use "Fixes #4227" in the PR description (not the commit message) if this is intended to fix that Issue.

UI/window-basic-main-outputs.cpp
```diff
@@ -2438,6 +2440,8 @@ bool AdvancedOutput::StartReplayBuffer()
 					     "RecRBSuffix");
 		rbTime = config_get_int(main->Config(), "AdvOut", "RecRBTime");
 		rbSize = config_get_int(main->Config(), "AdvOut", "RecRBSize");
+		bool overwriteIfExists = config_get_bool(
+			main->Config(), "Output", "OverwriteIfExists");
```
WizardCM, line 2444:
```suggestion
```

This already occurs 12 lines above.

Penwy: Thank you, completely missed that.
Dunnae why it was there, it was completely unused.

GeorgesStavracas (approved):
UI / UX review:

Short term, this seems consistent with the behavior of other properties.
Long term, it'll be better to split the controls from recording and replay buffer.

Ju8z:
Still not fixed and having same issue... It creates new folder with the "Game Capture" name and crfeate default name as "Game Capture" and overwrites it always, no matter what settings you have set. Prefix and Suffix are not being set as well!

Penwy:
> Still not fixed and having same issue

Yes indeed, until this, or another PR fixing the issue gets merged, the issue won't be fixed.

PR #10096, opened 2024-01-11
libobs: Fix texture encoder duping frames when queue full

derrod:
### Description

Fixes an oversight with `queue_frame()` where the first frame in the queue would be duplicated if there are no available free queue slots, thus keeping the encoder busy with duplicate frames rather than allowing it to work through the existing queue.

By changing the condition to only duplicate frames if there are free slots *and* the current frames needs duplication this snowballing issue no longer occurs.

Additionally, to avoid calling `circlebuf_pop_front()` on an empty buffer frames are now skipped if no items are available in the queue.

### Motivation and Context

Want to be able to push GPU encoders to the limit without it falling apart when tabbing in and out of a game.

### How Has This Been Tested?

Twitch "enhanced broadcasting" dev build with 4 NVENC encoders running slow presets/tunes.

Note that this issue is rather hard to reproduce with a single encode session, in my case I noticed it with 4 running: 1440p120 AV1 (p7), 1080p60 AV1 (p7), 720p60 HEVC (p5), and 480p30 H.264 (p3).

However, I have been able to trigger it with a single encoder in some cases, it just required tabbing in and out of a GPU intensive game running in exclusive fullscreen.

### Types of changes

### Checklist:

### DIFF
--- a/libobs/obs-internal.h
+++ b/libobs/obs-internal.h
@@ -256,6 +256,7 @@ struct obs_tex_frame {
 	uint64_t lock_key;
 	int count;
 	bool released;
+	bool skip;
 };
 
 struct obs_task_info {
--- a/libobs/obs-video-gpu-encode.c
+++ b/libobs/obs-video-gpu-encode.c
@@ -137,6 +137,18 @@ static void *gpu_encode_thread(void *data)
 			if (skip)
 				continue;
 
+			if (tf.skip) {
+				/* Skip frame but increment PTS to maintain
+				 * sync if encoder already started. */
+				if (encoder->start_ts) {
+					encoder->cur_pts +=
+						encoder->timebase_num *
+						encoder->frame_rate_divisor;
+				}
+
+				continue;
+			}
+
 			if (!encoder->start_ts)
 				encoder->start_ts = timestamp;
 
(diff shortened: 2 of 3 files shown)


derrod:
Okay this has an issue where the PTS isn't incremented when frames are skipped, need to find a way to deal with that :|

derrod:
Pushed a fix, not sure if there's a better way, but we need to keep PTS synced up so just inserting fake frames into the queue seems to do the trick.

ma3uk:
It works fine, without this commit, when recording League of Legends, each collapse and unfolding of the game window gave several missed frames, with this commit there are no frame skips.

derrod:
> this is kind of interesting in that it brings the behavior closer to what obs classic did; have you checked how this interacts with keyframe intervals? one of the issues in obs classic was that configuring a keyframe interval did not exactly work because encoders were fed a potentially variable number of frames, so there was no stability guarantee around having a keyframe interval of 2 seconds (e.g. 120 frames at 60 fps) actually be 2 seconds; I'd assume this suffers from the same issue, unless you bring back manual synchronization (which we also experimented with in obs classic)?

I'm not entirely familiar with the implementation in OBS classic but this still keeps a steady frame rate going into the encoder, with skips only happening if the encoder queue is full. I don't really think this is an issue per-se, as you already don't have guarantees for keyframe intervals being strict unless `scenecut=0` or equivalents are set.

FWIW in my testing the times this actually kicked in were fairly rare, as it requires the encoder to run close to its maximum throughput for a temporary stall to fill up the queue.

> also, thoughts on extending this to cpu based encoders?

Haven't looked into that yet, but I suppose it would be possible.

Warchamp7:
Is this still something you think we should give some attention to @derrod?

derrod:
> Is this still something you think we should give some attention to @derrod?

Maybe down the line (similar to the resync PR). There are cases where we probably want more graceful recovery, but it needs some additional work. For example, we'd need to actually store the frame's duration to ensure correct muxing.

PR #10100, opened 2024-01-13
Enable exporting CMake interface target

tytan652:
### Description

Required by:
- https://github.com/obsproject/obs-browser/pull/431
- https://github.com/obsproject/obs-websocket/pull/1196

### Motivation and Context

Needed to enable installing obs-websocket-api  and obs-browser-api headers
### How Has This Been Tested?

Build on Linux (still uses legacy path) with obs-browser-api PR and check if the header is installed, which is the case.

Ci for Windows and macOS, headers install needs to be checked locally
- Tested on Windows,  `--component Development` adds the header

### Types of changes

### Checklist:

### DIFF
--- a/cmake/windows/helpers.cmake
+++ b/cmake/windows/helpers.cmake
@@ -274,12 +274,15 @@ function(target_export target)
   set(exclude_variant EXCLUDE_FROM_ALL)
   _target_export(${target})
 
-  install(
-    FILES "$<TARGET_PDB_FILE:${target}>"
-    CONFIGURATIONS RelWithDebInfo Debug Release
-    DESTINATION "${OBS_EXECUTABLE_DESTINATION}"
-    COMPONENT Development
-    OPTIONAL)
+  get_target_property(target_type ${target} TYPE)
+  if(NOT target_type STREQUAL INTERFACE_LIBRARY)
+    install(
+      FILES "$<TARGET_PDB_FILE:${target}>"
+      CONFIGURATIONS RelWithDebInfo Debug Release
+      DESTINATION "${OBS_EXECUTABLE_DESTINATION}"
+      COMPONENT Development
+      OPTIONAL)
+  endif()
 endfunction()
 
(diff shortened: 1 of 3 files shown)


cmake/windows/helpers.cmake
```diff
@@ -274,12 +274,15 @@ function(target_export target)
...
-    OPTIONAL)
+  get_target_property(target_type ${target} TYPE)
+  if(NOT target_type STREQUAL INTERFACE_LIBRARY)
+    install(
+      FILES "$<TARGET_PDB_FILE:${target}>"
+      CONFIGURATIONS RelWithDebInfo Debug Release
+      DESTINATION "${OBS_EXECUTABLE_DESTINATION}"
+      COMPONENT Development
+      OPTIONAL)
+  endif()
```
PatTheMav, line 285: Is the condition necessary here? Because the `install` call is explicitly marked `OPTIONAL`, so if the file doesn't exist, it will skip the step anyway.

tytan652: It breaks build/install if not added.

tytan652: And modules ?

PatTheMav:
```
CMake Error:
  Error evaluating generator expression:

    $<TARGET_PDB_FILE:headers-test>

  Target "headers-test" is not an executable or library.
```

That's all it says, though I might be wrong that it could also affect static libraries (because that's still a library).

The docs state "Full path to the linker generated program database file (.pdb) where tgt is the name of a target", so anything that requires linking should be fine which seems to leave only INTERFACE libraries as the odd one out.

I guess it's fine to only check for that then.

tytan652: So I keep "not interface" then.

PatTheMav: Yep.

Lain-B (approved):

PR #10104, opened 2024-01-15
rtmp-services: Add Hakuna Live RTMP service

lenaky:
### Description
Adding Hakuna Live's RTMP Ingest server address

### Motivation and Context
Hakuna Live is one of the global live streaming service platforms.
To make it easier for streamers to access Hakuna Live, add this address.

### How Has This Been Tested?
I followed [service submission guidelines](https://github.com/obsproject/obs-studio/wiki/Service-Submission-Guidelines) and confirmed streaming's working.

### Types of changes

### Checklist:

### DIFF
--- a/plugins/rtmp-services/data/package.json
+++ b/plugins/rtmp-services/data/package.json
@@ -1,11 +1,11 @@
 {
     "$schema": "schema/package-schema.json",
     "url": "https://obsproject.com/obs2_update/rtmp-services/v5",
-    "version": 247,
+    "version": 248,
     "files": [
         {
             "name": "services.json",
-            "version": 247
+            "version": 248
         }
     ]
 }
--- a/plugins/rtmp-services/data/services.json
+++ b/plugins/rtmp-services/data/services.json
@@ -2799,6 +2799,29 @@
                 "max video bitrate": 12000,
                 "max audio bitrate": 320
             }
+        },
+        {
+            "name": "Hakuna Live - RTMP",
+            "servers": [
+                {
+                    "name": "Republic of Korea / Japan",
+                    "url": "rtmps://media-proxy.prod.an1.media.hpcnt.com/hakuna"
+                }
+            ],
+            "protocol": "RTMP",
+            "supported video codecs": [
+                "h264"
+            ],
+            "recommended": {
+                "keyint": 2,
+                "supported resolutions": [
+                    "1920x1080",
+                    "1280x720"
+                ],
+                "max video bitrate": 6000,
+                "max audio bitrate": 320,
+                "bframes": 0
+            }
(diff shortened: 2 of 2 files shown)


Fenrirthviti:
Hello, can you please confirm that you are a representative of Hakuna Live?

We are also unable to validate the service as is appears to be blocked in the US.

lenaky:
> Hello, can you please confirm that you are a representative of Hakuna Live?

Hi, I'm working at hyperconnect which is devloper of hakuna live service.
My email address (len@hpcnt.com) indicates that I am an employee of hyperconnect.
see https://www.movefast.company/company find family site.
Is it enough or should I prove myself in other way?

> We are also unable to validate the service as is appears to be blocked in the US.

We don't block any access but only registered users can stream to our server with streamKey.
Now, We just run a streaming server for south korea and japan.
What should I do in this case?

Fenrirthviti:
The service is indeed blocked in the US. Loading the page loads the front page completely, and the redirects with an error:
image

https://hakuna.live/en/unavailable-location

Can you provide any public details on current userbase? From what I was able to briefly see before the website blocked itself, there were under 50 active users.

lenaky:
You're right, I forgot that access from the U.S. is blocked. We operate a service in several regions, and when we consider all regions combined (Asia, Europe, MENA, etc.), there are around 1,000 to 1,600 users streaming simultaneously.

Our service offers both WebRTC and RTMP ingestion. The majority of our users utilize WebRTC, including WHIP, while a few continue to use RTMP.

RytoEX:
I've fixed the merge conflicts and the commit message. Is the added service ingest really identified as "Republic of Korea, Japan"? It would be helpful from our point of view if we could check/verify the site and/or ingest list from US-based IP addresses.

lenaky:
Unfortunately, I have no control over allowing access from the US. However, I can try to speak to the department which has the authorization to open U.S, although it's difficult at the moment.

RytoEX:
For my comment about the ingest name, I mostly just find "Republic of Korea, Japan" confusing, as a comma in this format usually indicates the first location is inside the second location. I would have expected "Republic of Korea / Japan" or something else. However, as I said, if your official documentation refers to it as "Republic of Korea, Japan", we aren't in the habit of mandating what you name your ingests. I just wanted to confirm that this isn't a mistake.

About opening to the US, I don't even need the ability to stream. I just wanted us to be able to easily verify the service is live/active. That said, we can probably waive this concern.

lenaky:
Thank you and I modified ingest name as you noted.

RytoEX (approved):
Seems fine. Hopefully our services checker doesn't mark it as being down.

PR #10113, opened 2024-01-18
obs-scripting: Add methods for getting scene/hotkey from calldata

Sheepolution:
### Description

For Python/Lua scripting there are already two methods provided for getting userdata. These are `calldata_source` and `calldata_sceneitem`. This PR adds two more methods: `calldata_scene` and `calldata_hotkey`. Examples of signals that would require these methods to access the data are `refresh` and `hotkey_register` respectively.

### Motivation and Context

Some signals provide scene and hotkey data. This data is currently not accessible in Python/Lua scripting. This commit is to fix this issue.

### How Has This Been Tested?

This has not been tested. Considering it's a small change where existing code was copy-pasted with types edited, I did not go through the effort of getting an OBS build ready on my PC. Sorry.

### Types of changes

### Checklist:


Sheepolution:
Resolved the conflicts. Remains untested.

Fenrirthviti:
Can you please provide a proper motivation and context for why this change is being submitted?

We're not sure what the use case actually is.

Sheepolution:
OBS scripting has signals, which we can subscribe to in Python/Lua with `signal_handler_connect`. Upon a signal triggering, it calls the callback we passed, and it gives us the relevant calldata.

There are four different types of calldata:
* Source
* Scene item
* Scene
* Hotkey

To be able to use this data in scripting, we first need to convert the data. For a signal where we get a source, we can use `calldata_source`. For a signal where we get a hotkey (e.g. `hotkey_register`) or a scene (e.g. `item_add`) there is no way for us to convert the calldata to userdata. That is what this PR solves, by adding `calldata_scene` and `calldata_hotkey`.

Is this something that I will actually use? No, that time has passed. The motivation for this now is that it fills a gap in the Python/Lua implementation of the API. I don't see why `scene` and `hotkey` would be left out, and I can only imagine that they were forgotten.

PR #10121, opened 2024-01-20
Add CheckSymbolExists module for FindMbedTLS.cmake

winterheart:
### Description

FindMbedTLS.cmake uses `check_symbol_exists` macro which requires `include(CheckSymbolExists)` before invocation (see https://cmake.org/cmake/help/latest/module/CheckSymbolExists.html).

### Motivation and Context

Without `include(CheckSymbolExists)` project fails to configure

### How Has This Been Tested?

CMake produces this error on some cases:
```
CMake Error at cmake/Modules/FindMbedTLS.cmake:97 (check_symbol_exists):
  Unknown CMake command "check_symbol_exists".
```

### Types of changes

### Checklist:

### DIFF
--- a/cmake/Modules/FindMbedTLS.cmake
+++ b/cmake/Modules/FindMbedTLS.cmake
@@ -94,6 +94,7 @@ if(MBEDTLS_LIB
    AND NOT MBEDX509_LIB)
   set(CMAKE_REQUIRED_LIBRARIES ${MBEDTLS_LIB})
   set(CMAKE_REQUIRED_INCLUDES ${MBEDTLS_INCLUDE_DIR})
+  include(CheckSymbolExists)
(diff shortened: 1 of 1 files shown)


RytoEX:
This looks correct. @PatTheMav ?

PatTheMav:
> This looks correct. @PatTheMav ?

I'd rather argue to remove the entire codepath as I'm not aware of a single instance where this occurred, but given that the legacy find module is deprecated anyway, there's not much damage in adding this.

RytoEX:
> Without include(CheckSymbolExists) project fails to configure

In what circumstance does this fail? What host system are you configuring on? How did you invoke the configuration step?

We're fine merging this, but as mentioned above, this finder is deprecated and is due to be removed in a future release.

PR #10124, opened 2024-01-22
obs-qsv11: Implement texture encoder on linux

kkartaltepe:
### Description

Implements texture encoding based on #8312 bringing performance equivalence with windows. ~~Still draft as there are a handful of TODOs and doing the right thing on multi-gpu machines might need more plumbing in libobs, but for single gpu systems this should be enough to test the texture encoding PR.~~ Multi-gpu support is in place.

### Motivation and Context

Gotta go fast. Encoding 1440p@120fps at 1% cpu is fun.

### How Has This Been Tested?

Tested on a KBL machine and a DG2 machine with h264/h265/av1 encoders. Everything seems to be working fine.

### Types of changes

### Checklist:

### DIFF
--- a/plugins/obs-qsv11/common_utils_linux.cpp
+++ b/plugins/obs-qsv11/common_utils_linux.cpp
@@ -2,9 +2,9 @@
 
 #include <time.h>
 #include <cpuid.h>
-#include <util/c99defs.h>
-#include <util/dstr.h>
+
 #include <va/va_drm.h>
+#include <va/va_drmcommon.h>
 #include <va/va_str.h>
 #include <stdio.h>
 #include <stdlib.h>
@@ -14,7 +14,12 @@
 #include <dirent.h>
 
 #include <obs.h>
+#include <obs-encoder.h>
 #include <obs-nix-platform.h>
+#include <graphics/graphics.h>
+#include <util/c99defs.h>
+#include <util/dstr.h>
+#include <util/bmem.h>
 
 // Set during check_adapters to work-around VPL dispatcher not setting a VADisplay
 // for the MSDK runtime.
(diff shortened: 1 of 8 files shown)


kkartaltepe:
There is a rare invalid first frame now, i may have introduce it during some rebasing. Will try and figure out whats up.

kkartaltepe:
Fixed an issue with uninitialized windows specific values causing some incorrect behavior. It should be correct for all platforms pending a parallel PR to work around some mesa driver bugs.

plugins/obs-qsv11/common_utils_linux.cpp
```diff
@@ -55,33 +191,93 @@ mfxStatus simple_unlock(mfxHDL pthis, mfxMemId mid, mfxFrameData *ptr)
...
-	UNUSED_PARAMETER(handle);
-	return MFX_ERR_UNSUPPORTED;
+	if (NULL == handle)
+		return MFX_ERR_INVALID_HANDLE;
+
+	// Seemingly undocumented, but Pair format defined by
+	// oneVPL-intel-gpu-intel-onevpl-23.1.0/_studio/mfx_lib/encode_hw/av1/linux/base/av1ehw_base_va_packer_lin.cpp
+	mfxHDLPair *pPair = (mfxHDLPair *)handle;
+
+	// Pointer to a VASurfaceID, will be dereferenced by the driver.
```
RytoEX, line 201: This isn't really a sentence. If it's a fragment, then it's "fine" to omit final punctuation. If this is restructured to a complete sentence, then use final punctuation.

plugins/obs-qsv11/common_utils_linux.cpp
```diff
@@ -55,33 +191,93 @@ mfxStatus simple_unlock(mfxHDL pthis, mfxMemId mid, mfxFrameData *ptr)
...
 {
 	UNUSED_PARAMETER(pthis);
-	UNUSED_PARAMETER(mid);
-	UNUSED_PARAMETER(handle);
-	return MFX_ERR_UNSUPPORTED;
+	if (NULL == handle)
+		return MFX_ERR_INVALID_HANDLE;
+
+	// Seemingly undocumented, but Pair format defined by
+	// oneVPL-intel-gpu-intel-onevpl-23.1.0/_studio/mfx_lib/encode_hw/av1/linux/base/av1ehw_base_va_packer_lin.cpp
```
RytoEX, line 198: I'm having trouble finding this repo. Could you link to it?

plugins/obs-qsv11/common_utils_linux.cpp
```diff
@@ -55,33 +191,93 @@ mfxStatus simple_unlock(mfxHDL pthis, mfxMemId mid, mfxFrameData *ptr)
...
+	mfxStatus sts =
+		MFXVideoCORE_GetHandle(*session, DEVICE_MGR_TYPE, &display);
+	MSDK_CHECK_RESULT(sts, MFX_ERR_NONE, sts);
+
+	struct encoder_texture *ptex = (struct encoder_texture *)tex;
+	struct surface_info *surf = (struct surface_info *)mid;
+
+	obs_enter_graphics();
+	// Error reporting for copy is poor, if earlier surface handling is
+	// broken this will do bad things.
```
RytoEX, line 258:
```suggestion
	/* Error reporting for copy is poor. If earlier surface handling is
	 * broken, this will do bad things. */
```

plugins/obs-qsv11/common_utils_linux.cpp
```diff
@@ -116,16 +311,27 @@ mfxStatus Initialize(mfxVersion ver, mfxSession *pSession,
...
+		obs_enter_graphics();
+		gs_enum_adapters(get_drm_device, &params);
+		obs_leave_graphics();
+	} else {
+		if (codec == QSV_CODEC_AVC && default_h264_device)
+			device_path = default_h264_device;
+		if (codec == QSV_CODEC_HEVC && default_hevc_device)
+			device_path = default_hevc_device;
+		if (codec == QSV_CODEC_AV1 && default_av1_device)
+			device_path = default_av1_device;
```
RytoEX, line 330: Aren't these mutually exclusive? Could this be an `if-else if-else if` so the other conditions are not evaluated once one is matched?

plugins/obs-qsv11/common_utils_linux.cpp
```diff
@@ -187,7 +405,7 @@ double TimeDiffMsec(mfxTime tfinish, mfxTime tstart)
 {
 	UNUSED_PARAMETER(tfinish);
 	UNUSED_PARAMETER(tstart);
-	//TODO, unused so far it seems
+	//unused so far.
```
RytoEX, line 408: If this is not a sentence, you can omit final punctuation.

RytoEX (changes requested):
Commit message nits:
* linux -> Linux
* dont -> don't
* cant -> can't
* multi-gpu -> multi-GPU
* `systems. And enables` -> `systems and enables`

The second commit message will probably need its line wrapping re-adjusted after those changes.

plugins/obs-qsv11/common_utils_linux.cpp
kkartaltepe: I dont even think intel engineers could find it honestly. this is version specific and was pulled from the actual release tarball that my distro shipped.

https://spec.oneapi.io/onevpl/2.3.1/API_ref/VPL_structs_memory.html Here is the official docs on the type, notice its a "OS-Specific handle associated with a video frame", but actually what this means its its a https://github.com/intel/vpl-gpu-rt/blob/4170dd9fa1ea319dda81b6189616ecc9b178a321/_studio/mfx_lib/ext/mpeg2/include/mfx_mpeg2_encode_interface.h#L103 "surface pair" `mfxHDLPair`, at least in practice on linux.

And this encoded into the vaapi layer responsible for doing the accelerated frame handling https://github.com/intel/vpl-gpu-rt/blob/4170dd9fa1ea319dda81b6189616ecc9b178a321/_studio/shared/src/libmfx_core_vaapi.cpp#L1251 (srcHandle is a mfxHDLPair)

What actually goes in the pair you have to find in same file https://github.com/intel/vpl-gpu-rt/blob/4170dd9fa1ea319dda81b6189616ecc9b178a321/_studio/shared/src/libmfx_core_vaapi.cpp#L1464

intel's changing their code all the time so i could change it to one of these github links, assuming they dont rename their repositories. Cause im 70% sure this repo wasnt named vpl-gpu-rt when I was writing this code a few months ago...

https://spec.oneapi.io/onevpl/2.7.1/programming_guide/VPL_prg_mem.html Also this sample code seems oddly familiar to me, and wasnt there last I looked :)

plugins/obs-qsv11/common_utils_linux.cpp
```diff
@@ -27,13 +32,144 @@ struct linux_data {
...
+	response->NumFrameActual = 0;
+
+	mfxSession *session = (mfxSession *)pthis;
+	VADisplay display;
+	mfxStatus sts =
+		MFXVideoCORE_GetHandle(*session, DEVICE_MGR_TYPE, &display);
+	MSDK_CHECK_RESULT(sts, MFX_ERR_NONE, sts);
+
+	// https://ffmpeg.org/doxygen/5.1/hwcontext__vaapi_8c_source.html#l00109
+	// though earlier comments suggest the driver ignores rt_format so we could choose whatever.
```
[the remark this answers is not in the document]
kkartaltepe: Can I die on the hill of not using block comments unless my comment is more than 3 lines?

plugins/obs-qsv11/common_utils_linux.cpp
kkartaltepe: I generally prefer to leave it if they are mutually exclusive, but I added elses.

plugins/obs-qsv11/common_utils_linux.cpp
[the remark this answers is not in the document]
RytoEX: I'll think about it. However, IIRC, @Lain-B prefers `/* */` even for single line comments. It's not the end of the world, but do please fix the other nits. I suggested the above in the hope of avoiding awkward line unwrapping when we switch from 80 to 120 columns.

plugins/obs-qsv11/common_utils_linux.cpp
kkartaltepe: Added one of the github links too.

Lain-B (approved):

PR #10125, opened 2024-01-22
libobs: Use std _Pragma

RytoEX:
### Description

Instead of using the Microsoft-specific __pragma keyword, use the standard _Pragma directive, which is now supported in C11 and C++11 /std modes.

### Motivation and Context

Use the C/C++ standard features over compiler or toolchain specific features where possible.

See:
* https://learn.microsoft.com/en-us/cpp/preprocessor/pragma-directives-and-the-pragma-keyword?view=msvc-170

### How Has This Been Tested?

Built locally on Windows 11 with VS 2022 17.8.5.

### Types of changes

### Checklist:

### DIFF
--- a/libobs/util/c99defs.h
+++ b/libobs/util/c99defs.h
@@ -52,9 +52,9 @@
 #endif
 
 #ifdef _MSC_VER
-#define PRAGMA_WARN_PUSH __pragma(warning(push))
-#define PRAGMA_WARN_POP __pragma(warning(pop))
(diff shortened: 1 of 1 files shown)


tytan652 (approved):
LGTM, my only nitpik is the commit name that could be suffixed "with MSVC".

RytoEX:
> LGTM, my only nitpik is the commit name that could be suffixed with "`with MSVC`".

Fair point. Added.

PatTheMav (approved):

PR #10122, opened 2024-01-21
obs-ffmpeg: Add native non-texture NVENC encoder

derrod:
### Description

Replaces the FFmpeg fallback in the "new" NVENC encoder with a native non-texture encoder.

Also adds support for the non-texture encoder on Linux.

Superseedes/Closes #8794 

### Motivation and Context

The FFmpeg fallback sucks I hate it.

### How Has This Been Tested?

Tested H.264 encoder on my 4090 with the latest driver on Windows and Linux. 

### Types of changes

- New feature (non-breaking change which adds functionality)

### Checklist:

- [x] My code has been run through [clang-format](https://github.com/obsproject/obs-studio/blob/master/.clang-format).
- [x] I have read the [**contributing** document](https://github.com/obsproject/obs-studio/blob/master/CONTRIBUTING.rst).
- [x] My code is not on the master branch.
- [x] The code has been tested.
- [x] All commit messages are properly formatted and commits squashed where appropriate.
- [x] I have included updates to all appropriate documentation.

PR #10133, opened 2024-01-24
obs-ffmpeg: Linux NVENC texture encoder 

derrod:
### Description

WIP NVENC texture encoder for Linux, based on software encoder introduced in #10122 and OpenGL texture encoding libobs changes in #8312.

This PR really is just the last commit.

Superseedes/closes #4974

### Motivation and Context

Gotta go fast.

### How Has This Been Tested?

OBS Flatpak on openSUSE Tumbleweed with NVIDIA 545 drivers.

### Types of changes

### Checklist:


plugins/obs-ffmpeg/obs-nvenc.c
```diff
@@ -239,18 +241,57 @@ static bool nv_texture_init(struct nvenc_data *enc, struct nv_texture *nvtex)
...
+{
+	const bool p010 = obs_p010_tex_active();
+
+	cu->cuMemAllocPitch(&nvtex->tex, &nvtex->pitch, p010 ? enc->cx * 2 : enc->cx,
+			    enc->cy + enc->cy / 2, 16);
+
+	NV_ENC_REGISTER_RESOURCE res = {NV_ENC_REGISTER_RESOURCE_VER};
+	res.resourceType = NV_ENC_INPUT_RESOURCE_TYPE_CUDADEVICEPTR;
+	res.resourceToRegister = (void *)nvtex->tex;
+	res.width = p010;
```
kkartaltepe, line 263: Not quite right?

derrod: Indeed this is wrong (might've been a fix swallowed by a GNOME Builder crash), but NVENC didn't care amusingly enough. I guess when it comes to CUDA surfaces it can get that information another way.

derrod:
Short AV1 clip recorded on openSUSE with the new texture sharing encoder:

https://github.com/obsproject/obs-studio/assets/3123295/e3939e68-ad72-4dca-b4d8-c4d22af0640c

derrod:
With #8312 merged I'll take this out of draft. It works and the PR can now be reviewed independently.

kkartaltepe:
Fallback encoders dont appear to be private and can be seen in the options, also the default encoder doesnt fallback and just fails with an error when using the wrong gpu. (couldnt actually get obs to load on the gpu atm so ill test the working path later).

derrod:
> Fallback encoders dont appear to be private and can be seen in the options, also the default encoder doesnt fallback and just fails with an error when using the wrong gpu. (couldnt actually get obs to load on the gpu atm so ill test the working path later).

Yep, until I haven't implemented an automatic fallback yet so I left the non-texture encoders visible. Gotta figure out how to determine if OBS and the CUDA context run on the same GPU.

derrod:
Updated with a commit adding a check on startup to see if mapping an OpenGL texture created by the OBS graphics context succeeds on any CUDA device. If not, only fallback encoders will be supported, otherwise the CUDA device where mapping succeeded will be used when starting a texture encoder.

This also hides the fallback encoders by default on Linux as well.

plugins/obs-ffmpeg/obs-nvenc-helpers.c
```diff
@@ -7,6 +7,8 @@
 
 #ifdef _WIN32
 #include <util/windows/device-enum.h>
+#else
+#include <GL/gl.h>
```
tytan652, line 11: This header is included without any check as a dependency in obs-ffmpeg CMake.

derrod: So it is, this was just taken from the original implementation, I thinkg the correct thing to do here would be to link against `OBS::glad` and use `#include <glad/glad.h>` instead?

All that's really used from the header is `GL_TEXTURE_2D` and `GLuint` anyway.

tytan652: Depending only on a header, is still a dependency.

tytan652: I think it's `find_package(OpenGL REQUIRED)` and `OpenGL::GL` for this case.

plugins/obs-ffmpeg/obs-nvenc.c
```diff
@@ -1155,71 +1437,104 @@ static void *nvenc_create_internal(enum codec_type codec, obs_data_t *settings,
...
 	if (obs_encoder_scaling_enabled(encoder)) {
-		if (!obs_encoder_gpu_scaling_enabled(encoder)) {
+		if (!obs_encoder_gpu_scaling_enabled(encoder) && texture) {
 			blog(LOG_INFO,
-			     "[obs-nvenc] CPU scaling enabled, falling back to ffmpeg");
+			     "[obs-nvenc] CPU scaling enabled, falling back to"
+			     " non-texture encoder");
 			goto reroute;
 		}
 		blog(LOG_INFO, "[obs-nvenc] GPU scaling enabled");
```
tt2468, line 1459: By adding the `&& texture` above, this GPU scaling message will print if CPU scaling is enabled and `texture == false`.

Bleuzen:
Hi, this is really cool, used it for streaming lately, works fine :)

Only minor things noticed just now due to the addition of AV1:
ffmpeg says the QP levels go from -1 to 51 for H264 and HEVC and from -1 to 255 for AV1.
However in OBS, the max level in CQP mode for AV1 is 63.
Looking at the code, this value is later multiplied by 4:
https://github.com/derrod/obs-studio/blob/5ced339dc7a2ce6300f720bcbb6e6d64c9c40d7c/plugins/obs-ffmpeg/obs-nvenc.c#L840
Is there a reason this multiplication is done? Why not expose the raw value like it is done for the other codecs? Seems a bit random to me.

Also, in the interface we can set a "CQ Level":
image

...while in code it looks like it does not set the CQ level, but the QP level. As far as I'm aware, OBS does not support CQ mode with NVENC currently, so unless I'm just having a dumb moment rn, the naming in the UI does not make sense.

tt2468:
These are all existing behaviors of the shared texture NVENC encoder and changing them is out of scope of the goals of this PR.

tt2468 (approved):

Bleuzen:
Did some more testing, this time with recording. There is a noticeable change in the bitrates used. I set up OBS with and without this patch the same way and used a VLC source with the same lossless video playing for reproducible input

For example when recording NVENC H264 in Simple Mode / "High Quality, Medium File Size"...

Current OBS:


With this PR:


Same with HEVC

Before:


After:


~ 75 % increase in file size

Also happens in "Advanced" Output Mode with same CQ Levels.

Didn't really expect that, not sure if this is supposed to happen or a bug

tt2468:
> Did some more testing, this time with recording. There is a noticeable change in the bitrates used. I set up OBS with and without this patch the same way and used a VLC source with the same lossless video playing for reproducible input
> 
> For example when recording NVENC H264 in Simple Mode / "High Quality, Medium File Size"...
> 
> Current OBS:
> 
> 
> With this PR:
> 
> 
> Same with HEVC
> 
> Before:
> 
> 
> After:
> 
> 
> ~ 75 % increase in file size
> 
> Also in "Advanced" Output Mode with same CQP Levels
> 
> Didn't really expect that, not sure if this is supposed to happen or a bug

The encoder implementations are different, but nothing should have been touched in relation to the bitrate regulation. That's likely just a pre-existing difference between the ffmpeg and texture encoders.

In order to get useful results, this texture encoder PR must be tested against the current texture encoder available on Windows.

ochilan:
@Bleuzen Could you provide details about each of the resulting files, for example using a tool like "mediainfo"? I wonder if there are unintended technical differences like e.g. chroma subsampling (4:2:0 vs. 4:4:4).

realnc:
Applied on top of 30.1.0 and it works fine for me. I can record AV1 with NVENC and it's fast too!

Now if we only had CQ support instead of CQP, that would make OBS the perfect capture tool for image quality.

ochilan:
> The encoder implementations are different, but nothing should have been touched in relation to the bitrate regulation. That's likely just a pre-existing difference between the ffmpeg and texture encoders.
> 
> In order to get useful results, this texture encoder PR must be tested against the current texture encoder available on Windows.

I  was curious and compared stock OBS 30.1 under Linux, stock 30.1 under Windows, and 30.1 + texture encoder under Linux (H.264, CQP 20, with as similar settings as possible). The resulting file sizes were:

Stock 30.1 Linux: 109.1 MiB
Stock 30.1 Windows: 149 MiB
Patched 30.1 Linux: 147.5 MiB

So there actually seems to be a difference before and after the patch under Linux, but the larger file size actually seems to be comparable to what the Windows version already produces.

(As far as I could tell, the metadata of all videos is identical by the way, except for the video bitrates.)

derrod:
I'm not sure if there are any differences with CQP handling in FFmpeg vs OBS, but I also don't think that's worth investigating/worrying about for now. For now we don't want to break anything on the Windows side, but this PR is a stepping stone to a larger rework that may change things.

Lain-B (approved):

PR #10138, opened 2024-01-24
libobs: Add video frame memory alignment (plus video-frame refactor)

tt2468:
### Description
Many video encoders (and raw outputs) require linesize-aligned frame buffers. In order to prevent unnecessary copies, this adds the initial functionality for creating video frames with arbitrary memory alignment.

Depends #10139

### Motivation and Context
The current OBS host-based raw video frame output pipeline makes a number of unnecessary copies, resulting in a notable penalty for encoders and raw outputs (like decklink). By allowing video frames to be created with arbitrary alignment, at least one frame copy can be completely avoided.

### How Has This Been Tested?
Tested on Ubuntu 22.04 with the video formats available in the UI, along with an unnamed hardware encoder which accepts I420 pixel format. Using amduprof to measure memory transfers, the system memory impact of a single 1080p60 encode went from 1.66GB/s to 0.93GB/s.

### Types of changes

### Checklist:

### DIFF
--- a/libobs/media-io/video-frame.c
+++ b/libobs/media-io/video-frame.c
@@ -14,404 +14,252 @@
     You should have received a copy of the GNU General Public License
     along with this program.  If not, see <http://www.gnu.org/licenses/>.
 ******************************************************************************/
-
+#include <assert.h>
 #include "video-frame.h"
 
-#define ALIGN_SIZE(size, align) size = (((size) + (align - 1)) & (~(align - 1)))
+#define HALF(size) ((size + 1) / 2)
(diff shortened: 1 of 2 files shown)


tt2468:
@jpark37 Curious your thoughts on this. Many of the texture formats had arbitrary memory alignment behavior because of a wide range of different plane size calculation methods. Seems that just many years of adding and adding resulted in some wildly different implementations per format.

Most notably I think is that the alignment currently done does not align to plane boundaries, only allocation boundaries. Correct me if I'm wrong, but on systems where alignment is something like 4 or 8 bytes, there's a reasonable chance that memcpy() won't actually be aligned during frame copies. This should fix that issue

PR #10142, opened 2024-01-26
UI: Add --no-studio-mode command line option

LyroyTheToad:
### Description

Adds `--no-studio-mode` as a new command line option.
Starting OBS with this option will run OBS with the studio mode disabled.

Logically, the `--no-studio-mode` conflicts with the `--studio-mode` option, so if both options are used, the program prints an error that says '_Can't use "--studio-mode" and "--no-studio-mode" together._' and exists with code 1.

If OBS is run without specifying `--studio-mode` and `--no-studio-mode`, the studio mode state will assume its most recent state (if it was disabled it will still be disables and vice versa).

The help message was also updated to show this new option.

### Motivation and Context

I was creating different profiles and scene collections to keep my OBS setup organized.
To simplify my use of OBS, I decided to create multiple .desktop entries to launch OBS with different profiles and scenes.
I wanted some .desktop entries to launch OBS with studio mode enabled and some others with studio mode disabled. This is when I noticed I couldn't tell OBS to disable studio mode using a command line option.
I tried searching online but I couldn't find anything, so I decided to add this feature myself.

### How Has This Been Tested?

- Tested on an Arch Linux machine, kernel version 6.7.1 .
- Compiled with `-DENABLE_BROWSER=OFF -DCMAKE_BUILD_TYPE=Debug -DENABLE_WEBRTC=OFF -DLINUX_PORTABLE=ON -DCMAKE_INSTALL_PREFIX="${HOME}/obs-studio-portable"`.
(These flags were used to facilitate the compiling process since I didn't have all the things needed to compile OBS with the defaults. I don't have any reason to believe this may not work with a _normal copy_ of OBS.)

### Types of changes

### Checklist:

### DIFF
--- a/UI/obs-app.cpp
+++ b/UI/obs-app.cpp
@@ -98,6 +98,7 @@ static bool unfiltered_log = false;
 bool opt_start_streaming = false;
 bool opt_start_recording = false;
 bool opt_studio_mode = false;
+bool opt_no_studio_mode = false;
 bool opt_start_replaybuffer = false;
 bool opt_start_virtualcam = false;
 bool opt_minimize_tray = false;
@@ -3390,6 +3391,9 @@ int main(int argc, char *argv[])
 		} else if (arg_is(argv[i], "--studio-mode", nullptr)) {
 			opt_studio_mode = true;
 
+		} else if (arg_is(argv[i], "--no-studio-mode", nullptr)) {
+			opt_no_studio_mode = true;
+
 		} else if (arg_is(argv[i], "--allow-opengl", nullptr)) {
 			opt_allow_opengl = true;
 
@@ -3415,6 +3419,7 @@ int main(int argc, char *argv[])
 				"--profile <string>: Use specific profile.\n"
 				"--scene <string>: Start with specific scene.\n\n"
 				"--studio-mode: Enable studio mode.\n"
+				"--no-studio-mode: Disable studio mode.\n"
 				"--minimize-to-tray: Minimize to system tray.\n"
 #if ALLOW_PORTABLE_MODE
 				"--portable, -p: Use portable mode.\n"
@@ -3443,6 +3448,12 @@ int main(int argc, char *argv[])
 				  << App()->GetVersionString(false) << "\n";
 			exit(0);
 		}
+	}
+
+	if (opt_studio_mode && opt_no_studio_mode) {
+		std::cout
+			<< "Can't use \"--studio-mode\" and \"--no-studio-mode\" together.\n";
+		exit(1);
 	}
 
 #if ALLOW_PORTABLE_MODE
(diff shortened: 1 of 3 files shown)


WizardCM:
Outside of "do we want this" (which is not up to me), is there a reason you went with a separate flag rather than making the existing flag a toggle? That way there'd be no possibility of conflict, and could probably simplify some of the code.

Fenrirthviti:
Logging my vote that I am against this change. Seems too niche, and would be better maintained as a local fork.

LyroyTheToad:
@WizardCM if you mean something like this: `--studio-mode [on] | off`, you are right. For some reason I discarded this option while I was coding, definitely a better solution without any ugly conflicts.
But if you mean that without specifying `--studio-mode` OBS starts with studio mode disabled, then this could be considered an unwanted behaviour by someone. In the current state, if you want to keep studio mode always enabled you just need to enable it once, a change like this will break this behaviour. 

(Another solution to my specific problem could be to memorize the studio mode state in the current profile, but mine was a much easier feature to implement.)

I definitely prefer the `--studio-mode [on] | off` approach. To update the code what is the best approach? Should I create another commit in my fork and update the pull requests, should I amend my commit to keep everything in a single commit, should I close and re-open a new pull request with just one commit?

@Fenrirthviti I do know it is a really niche feature, but it's a small change that just works and will work for years without needing to be touched. The only downside I can see is there is a little bit more text to be translated.
Also, the `--studio-mode` flag feels a little bit incomplete without its counterpart.

WizardCM:
After a bit of internal discussion, on top of my earlier feedback, we're thinking that rather than changing the behaviour of the existing flag it may be better to have a new flag purpose-built for the toggle. This would probably be best done as a "temporary runtime flag" where it doesn't persist the setting, compared to the existing flag which would remain as before.

Additionally, this will need to be built in a way that is supported by Windows (`std::cout` doesn't work on non-command-line apps).

In terms of how best to update your code, make the changes in the existing branch and force-push your changes, overriding the existing commits. Amending would be fine.

GeorgesStavracas:
> I definitely prefer the --studio-mode [on] | off approach. To update the code what is the best approach? Should I create another commit in my fork and update the pull requests, should I amend my commit to keep everything in a single commit, should I close and re-open a new pull request with just one commit?

We have discussed this today and `--studio-mode [on] | off` is the preferred approach, at least unless OBS Studio gets some sort of per-project window layout or something like that.

Please amend to this commit, and update the commit title and message to reflect the new approach. No need to recreate the PR.

jmluettgen64:
Following up - 

> We have discussed this today and `--studio-mode [on] | off` is the preferred approach, at least unless OBS Studio gets some sort of per-project window layout or something like that.

Has this been implemented or is there a plan to implement it? I think it is a useful feature especially in a multi-user environment. It adds another way to ensure that OBS opens to the same configuration every time.

Warchamp7:
Feel free to reopen if this is updated and rebased against the new file structure.

PR #10149, opened 2024-01-27
libobs: Add `obs_encoder_parent_video()` method

tt2468:
### Description
Allows parent (original) video object of an encoder with an FPS divisor to be fetched.

### Motivation and Context
When implementing a non-shared-texture encoder, I needed to be able to determine the original video_t object to use as a reference for a cross-encoder frame caching system. Since FPS division creates a new video_t object, multiple encoders cannot be associated with each other if one of them has a divisor set, as the video_t object will be different.

### How Has This Been Tested?
Ubuntu 22.04, runs

### Types of changes

### Checklist:

### DIFF
--- a/libobs/obs.h
+++ b/libobs/obs.h
@@ -2554,6 +2554,13 @@ EXPORT void obs_encoder_set_audio(obs_encoder_t *encoder, audio_t *audio);
 EXPORT video_t *obs_encoder_video(const obs_encoder_t *encoder);
 
 /**
+ * Returns the parent video output context used with this encoder, or NULL if not
+ * a video context. Used when an FPS divisor is set, where the original video
+ * context would not otherwise be gettable.
+ */
+EXPORT video_t *obs_encoder_parent_video(const obs_encoder_t *encoder);
+
+/**
  * Returns the audio output context used with this encoder, or NULL if not
  * a audio context
  */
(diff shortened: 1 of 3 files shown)


WizardCM:
Please squash your commits as the documentation change requires the code changes.

tt2468:
Commits squashed

derrod:
Seems fine, I recently had a use for this as well (GPU rescaling allows multiple video outputs with different formats, but `get_mix_for_video` fails if your encoder has an fps divisor, so if you need to get info from the mix that doesn't work right now).

libobs/obs.h
```diff
@@ -2553,6 +2553,12 @@ EXPORT void obs_encoder_set_audio(obs_encoder_t *encoder, audio_t *audio);
  */
 EXPORT video_t *obs_encoder_video(const obs_encoder_t *encoder);
 
+/**
+ * Returns the parent video output context used with this encoder, or NULL if not
+ * a video context. Used when FPS divisor is enabled
```
RytoEX, line 2558:
```suggestion
 * a video context. Used when FPS divisor is enabled.
```

Though, I confess, I'm not sure I understand the second sentence. I only see the function defined, but not used anywhere.

RytoEX (changes requested):

libobs/obs.h
tt2468: Added further context to the description. I think it should make more sense now.

RytoEX (approved):
I can't remember why this was assigned to me, but it seems fine. @derrod ?

RytoEX:
I might remove the backticks from the commit message though.

derrod (approved):

PR #10151, opened 2024-01-27
Add opengl `gs_enum_adapters` implementations

kkartaltepe:
### Description

enables using gs_enum_adapters and gs_get_adapter_count on opengl (when the platform implements them, so not for macos right now).

### Motivation and Context

We need a way to check that we are running on the right adapter for hardware encoders and texture sharing. The adapter index is used in some plugins and on windows so this allows us to avoid rewriting some of that plugin code.

This should allow the qsv/nvenc/ffmpeg texture encode PRs to safely disable/fallback to host memory if obs is not rendering on the encoder device.

### How Has This Been Tested?

Dumping the devices  when enumerating and everything looks good on Wayland and X11 on an intel+amd multi-gpu system with/without DRI_PRIME configuration.

### Types of changes

### Checklist:


kkartaltepe:
Tested with QSV linux encoding (and now a dependency of #10124)

Lain-B (approved):

PR #10153, opened 2024-01-28
libobs: Prevent encoder reconfiguration after initialization

derrod:
### Description

Prevents an encoder relying on GPU scaling from being assignment a video output that has different dimensions.

### Motivation and Context

Users reported issues about invalid output when using automatically started video recording.

The underlying issue is a race condition in the UI, but this seems like a good thing to do anyways and will prevent this issue from happening until we can fix the UI.

Race condition:
- `OBSBasic::StartStreaming()`
    + Calls `AdvancedOutput::SetupStreaming()`
        * Calls `AdvancedOutput::SetupOutputs()`
            - Assigns `obs_get_video()` to streaming/recording encoders
    + Calls `AdvancedOutput::StartStreaming()` which starts streaming output
        - The GPU scaled output is set up here during encoder init
    + Calls `OBSBasic::StartRecording()`
        * Calls `AdvancedOutput::StartRecording()`
            - Checks `Active()`, which at this point is still `false` despite the streaming output starting up and encoders being initialised
            - Calls `AdvancedOutput::SetupOutputs()` *again* which resets the scaled encoder to the unscaled output mix
                + At this point the scaled mix is leaked, as it will not be free'd until shutdown
- Outputs start and `Active()` becomes `true`
- Scaled encoder receives unscaled textures, things go wrong.

### How Has This Been Tested?

Tested with config provided by @Lordmau5 for reproduction.

### Types of changes

### Checklist:

### DIFF
--- a/libobs/obs-encoder.c
+++ b/libobs/obs-encoder.c
@@ -859,6 +859,13 @@ void obs_encoder_set_scaled_size(obs_encoder_t *encoder, uint32_t width,
 		     obs_encoder_get_name(encoder));
 		return;
 	}
+	if (encoder->initialized) {
+		blog(LOG_WARNING,
+		     "encoder '%s': Cannot set the scaled resolution "
+		     "after the encoder has been initialized",
+		     obs_encoder_get_name(encoder));
+		return;
+	}
 
 	const struct video_output_info *voi;
 	voi = video_output_get_info(encoder->media);
@@ -895,6 +902,13 @@ void obs_encoder_set_gpu_scale_type(obs_encoder_t *encoder,
 		     obs_encoder_get_name(encoder));
 		return;
 	}
+	if (encoder->initialized) {
+		blog(LOG_WARNING,
+		     "encoder '%s': Cannot enable GPU scaling "
+		     "after the encoder has been initialized",
+		     obs_encoder_get_name(encoder));
+		return;
+	}
 
 	encoder->gpu_scale_type = gpu_scale_type;
 }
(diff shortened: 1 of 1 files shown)


libobs/obs-encoder.c
```diff
@@ -1091,6 +1091,21 @@ void obs_encoder_set_video(obs_encoder_t *encoder, video_t *video)
...
+	if (obs_encoder_scaling_enabled(encoder) &&
+	    obs_encoder_gpu_scaling_enabled(encoder)) {
+		const struct video_output_info *new =
+			video_output_get_info(video);
+
+		if (encoder->scaled_width != new->width ||
+		    encoder->scaled_height != new->height) {
+			blog(LOG_WARNING,
+			     "encoder '%s': Cannot apply a video_t object with"
+			     "different dimensions to a GPU scaled encoder.",
```
RytoEX, line 1103: This looks like this logged line will be joined without a space. Technically logged lines don't need to be broken up (see style guide, "never break user-visible strings such as printk messages"), but making sure the spacing is readable is also fine.

Lordmau5:
Can confirm this works fine on my RTX 4060 and 3090 now
Had a friend test it with his AMD 6800XT and he's not getting an error anymore either

derrod:
Updated to also prevent a number of other reconfigurations after the encoder has been initialised (fps divisor, scaled size, GPU scaling, input object). In my testing it still seems to be fine. This probably should've been done sooner, but we never had the issue of UI managed encoders using different video objects.

Lordmau5:
Just tested again and still works fine on my end. Can't try on an AMD card though, I can get my buddy to try it out tomorrow again, but I'd assume it'll work fine there, too :)

RytoEX (approved):
This looks sensible. Do we need any other eyes on this?

Lordmau5:
Just had my buddy test it on AMD again, still works fine :)

derrod:
> This looks sensible. Do we need any other eyes on this?

Probably not.

PR #10159, opened 2024-01-29
obs-webrtc: Add HEVC support

murillo128:
**Description**
This adds HEVC support to obs-webrtc

**Motivation and Context**
Users wish to use HEVC

**How Has This Been Tested?**
Tested against Dolby Millicast service and Safari tech preview

**Types of changes**

**Checklist:**
 - [x]  I have read the [contributing document](https://github.com/obsproject/obs-studio/blob/master/CONTRIBUTING.rst).
 -   [x] My code is not on the master branch.
 -   [x] The code has been tested.
 -   [x] All commit messages are properly formatted and commits squashed where appropriate.
 -   [x] I have included updates to all appropriate documentation.

### DIFF
--- a/plugins/obs-webrtc/whip-output.cpp
+++ b/plugins/obs-webrtc/whip-output.cpp
@@ -23,7 +23,6 @@ const uint8_t video_payload_type = 96;
 
 WHIPOutput::WHIPOutput(obs_data_t *, obs_output_t *output)
 	: output(output),
-	  is_av1(false),
 	  endpoint_url(),
 	  bearer_token(),
 	  resource_url(),
@@ -54,13 +53,6 @@ WHIPOutput::~WHIPOutput()
 bool WHIPOutput::Start()
 {
 	std::lock_guard<std::mutex> l(start_stop_mutex);
-
-	auto encoder = obs_output_get_video_encoder2(output, 0);
-	if (encoder == nullptr) {
-		return false;
-	}
-
-	is_av1 = (strcmp("av1", obs_encoder_get_codec(encoder)) == 0);
 
 	if (!obs_output_can_begin_data_capture(output, 0))
 		return false;
@@ -148,16 +140,31 @@ void WHIPOutput::ConfigureVideoTrack(std::string media_stream_id,
 		ssrc, cname, video_payload_type,
 		rtc::H264RtpPacketizer::defaultClockRate);
 
-	if (is_av1) {
+	const obs_encoder_t *encoder = obs_output_get_video_encoder2(output, 0);
+	if (!encoder)
+		return;
+
+	const char *codec = obs_encoder_get_codec(encoder);
+	if (strcmp("h264", codec) == 0) {
+		video_description.addH264Codec(video_payload_type);
+		packetizer = std::make_shared<rtc::H264RtpPacketizer>(
+			rtc::H264RtpPacketizer::Separator::StartSequence,
+			rtp_config, MAX_VIDEO_FRAGMENT_SIZE);
+#if ENABLE_HEVC
+	} else if (strcmp("hevc", codec) == 0) {
+		video_description.addH265Codec(video_payload_type);
+		packetizer = std::make_shared<rtc::H265RtpPacketizer>(
+			rtc::H265RtpPacketizer::Separator::StartSequence,
+			rtp_config, MAX_VIDEO_FRAGMENT_SIZE);
+#endif
+	} else if (strcmp("av1", codec) == 0) {
 		video_description.addAV1Codec(video_payload_type);
 		packetizer = std::make_shared<rtc::AV1RtpPacketizer>(
 			rtc::AV1RtpPacketizer::Packetization::TemporalUnit,
 			rtp_config, MAX_VIDEO_FRAGMENT_SIZE);
 	} else {
-		video_description.addH264Codec(video_payload_type);
-		packetizer = std::make_shared<rtc::H264RtpPacketizer>(
-			rtc::H264RtpPacketizer::Separator::StartSequence,
-			rtp_config, MAX_VIDEO_FRAGMENT_SIZE);
+		do_log(LOG_ERROR, "Video codec not supported: %s", codec);
+		return;
 	}
 
 	video_sr_reporter = std::make_shared<rtc::RtcpSrReporter>(rtp_config);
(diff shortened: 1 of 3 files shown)


tt2468 (approved):

murillo128:
@RytoEX @tt2468 can I do anything to help getting this merged? is there any blocker I am not aware of?

plugins/obs-webrtc/whip-output.cpp
```diff
@@ -133,6 +125,13 @@ void WHIPOutput::ConfigureAudioTrack(std::string media_stream_id,
 void WHIPOutput::ConfigureVideoTrack(std::string media_stream_id,
 				     std::string cname)
 {
+	auto encoder = obs_output_get_video_encoder2(output, 0);
```
tt2468, line 128: - I would suggest moving this codeblock down to where the results of these getters are actually used (the codec compares)
- We generally don't like using `auto` for boundaries between c/c++ apis. `OBSEncoder` can be used here.

plugins/obs-webrtc/whip-output.cpp
```diff
@@ -133,6 +125,13 @@ void WHIPOutput::ConfigureAudioTrack(std::string media_stream_id,
...
+	if (encoder == nullptr) {
+		return;
+	}
+
+	auto codec = obs_encoder_get_codec(encoder);
```
tt2468, line 133: Please use `const char *` or `std::string` instead of `auto` here

plugins/obs-webrtc/whip-output.cpp
```diff
@@ -148,16 +147,24 @@ void WHIPOutput::ConfigureVideoTrack(std::string media_stream_id,
 		ssrc, cname, video_payload_type,
 		rtc::H264RtpPacketizer::defaultClockRate);
 
-	if (is_av1) {
+	if (strcmp("av1", codec) == 0) {
```
tt2468, line 150: Just a nitpick, I would like to see these compares in the following order: `h264`, `hevc`, `av1`

plugins/obs-webrtc/whip-service.cpp
```diff
@@ -1,7 +1,7 @@
 #include "whip-service.h"
 
 const char *audio_codecs[MAX_CODECS] = {"opus"};
-const char *video_codecs[MAX_CODECS] = {"h264", "av1"};
+const char *video_codecs[MAX_CODECS] = {"h264", "av1", "hevc"};
```
tt2468, line 4: Same nitpick from above applies here

tt2468 (changes requested):

plugins/obs-webrtc/whip-output.cpp
```diff
@@ -148,16 +147,24 @@ void WHIPOutput::ConfigureVideoTrack(std::string media_stream_id,
...
 		video_description.addH264Codec(video_payload_type);
 		packetizer = std::make_shared<rtc::H264RtpPacketizer>(
 			rtc::H264RtpPacketizer::Separator::StartSequence,
 			rtp_config, MAX_VIDEO_FRAGMENT_SIZE);
+	} else if (strcmp("hevc", codec) == 0) {
+		video_description.addH265Codec(video_payload_type);
+		packetizer = std::make_shared<rtc::H265RtpPacketizer>(
+			rtc::H265RtpPacketizer::Separator::StartSequence,
+			rtp_config, MAX_VIDEO_FRAGMENT_SIZE);
+	} else {
```
RytoEX, line 165: Guard HEVC branches behind `ENABLE_HEVC` in outputs and encoders. It's fine to not guard the codec values in enums or enum-likes.

RytoEX (changes requested):

plugins/obs-webrtc/whip-output.cpp
murillo128: done

plugins/obs-webrtc/whip-output.cpp
murillo128: done

plugins/obs-webrtc/whip-service.cpp
murillo128: done

plugins/obs-webrtc/whip-output.cpp
```diff
@@ -148,16 +140,32 @@ void WHIPOutput::ConfigureVideoTrack(std::string media_stream_id,
...
+			rtc::H264RtpPacketizer::Separator::StartSequence,
+			rtp_config, MAX_VIDEO_FRAGMENT_SIZE);
+#if ENABLE_HEVC
+	} else if (strcmp("hevc", codec) == 0) {
+		video_description.addH265Codec(video_payload_type);
+		packetizer = std::make_shared<rtc::H265RtpPacketizer>(
+			rtc::H265RtpPacketizer::Separator::StartSequence,
+			rtp_config, MAX_VIDEO_FRAGMENT_SIZE);
+#endif
+	} else if (strcmp("av1", codec) == 0) {
```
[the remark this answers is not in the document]
murillo128: me neither, strcmp was on the original code, so just followed the same pattern for adding  hevc. If there is a getter returning an enum I am happy to change it to a "switch"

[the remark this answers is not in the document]
RytoEX: I don't think this output has any available codec enum or getter for it, which is why I suggested that this is perhaps something to clean up later.

RytoEX (changes requested):
Do not use merge commits. Use `git rebase` instead.

plugins/obs-webrtc/whip-output.cpp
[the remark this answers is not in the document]
murillo128: but the `codec` is gathered from the encoder, not from the output, so even we had an enum and a codec name mapping, we would still need an `strcmp`, right?

murillo128:
I have fixed the git commits, should be ready for merging now

plugins/obs-webrtc/whip-output.cpp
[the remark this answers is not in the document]
tt2468: I believe this is currently the best method to implement this code. I also believe that we should have use a codec enum instead of string ids originally in the API. Too bad, but this works fine enough

tt2468 (approved):

RytoEX (changes requested):
The PR says:

> obs-webrtc: Add HEVC support

and the code changes do precisely that. However, the commit message says:

> obs-webrtc: Add AV1 Support

Please fix that.

murillo128:
ups... changed commit message

RytoEX (changes requested):
> obs-webrtc: Add HEVC Support

Lowercase the word "Support" please.

RytoEX (approved):
This seems fine.

PR #10161, opened 2024-01-29
CI: Update first-party GitHub Actions from v3 to v4

RytoEX:
### Description

GitHub Actions has deprecated Actions based on node16. The v4 actions are based on node20. Replace first-party v3 actions with their v4 counterparts.

See:
https://github.blog/changelog/2023-09-22-github-actions-transitioning-from-node-16-to-node-20/

### Motivation and Context

Want to resolve deprecation warnings on CI. Want to avoid CI failing when the node16 actions are disabled.

### How Has This Been Tested?

obs-plugintemplate, obs-deps, and obs-browser have both had successful CI runs. This repo's CI will also verify.

### Types of changes

### Checklist:

### DIFF
--- a/.github/workflows/push.yaml
+++ b/.github/workflows/push.yaml
@@ -33,7 +33,7 @@ jobs:
     permissions:
       checks: write
     steps:
-      - uses: actions/checkout@v3
+      - uses: actions/checkout@v4
         with:
           fetch-depth: 0
 
@@ -57,7 +57,7 @@ jobs:
     permissions:
       checks: write
     steps:
-      - uses: actions/checkout@v3
+      - uses: actions/checkout@v4
         with:
           fetch-depth: 0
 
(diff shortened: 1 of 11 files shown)


.github/workflows/push.yaml
```diff
@@ -189,7 +189,7 @@ jobs:
           customLink: 'https://obsproject.com/'
 
       - name: Upload Artifacts 📡
-        uses: actions/upload-artifact@v3
+        uses: actions/upload-artifact@v4
```
derrod, line 192: Right now the sparkle artifact relies on the fact that v3 artifacts are combined server side go produce a single output with both architectures. So just switching to v4 will not work.

RytoEX: I can revert this specific change for now, but we'll want to address that "soon", per GitHub's current plan to "transition all actions to run on Node 20 by Spring 2024".

RytoEX: Reverted this line.

derrod: Yeah that's fine, we can probably just split them up as well, having them be one single artifact was convenient but not technically necessary.

PR #10165, opened 2024-01-30
CI: Update upload-artifact from v3 to v4 for Sparkle appcast job

RytoEX:
### Description

CI: Update upload-artifact from v3 to v4 for Sparkle appcast job

Implement the new merge sub-action from upload-artifact as a separate job that requires the Sparkle appcast creation job so that we still get a single combined artifact for the Sparkle appcast. This lets us update upload-artifact from v3 to v4.

GitHub Actions has deprecated Actions based on node16. The v4 actions are based on node20.

See:
https://github.blog/changelog/2023-09-22-github-actions-transitioning-from-node-16-to-node-20/

### Motivation and Context

Follow-up to fbbfbc68585b64c84e6bd4ab21389b07c6a68755 (#10161).

### How Has This Been Tested?

This can only be tested on tag, and I don't have my fork set up to do testing of the Sparkle processes. This looks correct to @derrod and I, but only a tag can tell if we got this right.

### Types of changes

### Checklist:

### DIFF
--- a/.github/workflows/push.yaml
+++ b/.github/workflows/push.yaml
@@ -189,10 +189,21 @@ jobs:
           customLink: 'https://obsproject.com/'
 
       - name: Upload Artifacts 📡
-        uses: actions/upload-artifact@v3
+        uses: actions/upload-artifact@v4
+        with:
+          name: macos-sparkle-update-${{ matrix.target }}
+          path: ${{ github.workspace }}/output
+
+  merge-appcasts:
+    runs-on: ubuntu-22.04
+    needs: create-appcast
+    steps:
+      - name: Merge Appcasts
+        uses: actions/upload-artifact/merge@v4
         with:
           name: macos-sparkle-update
-          path: ${{ github.workspace }}/output
+          pattern: macos-sparkle-update-*
+          delete-merged: true
 
   create-windows-update:
     name: Create Windows Update 🥩


RytoEX:
I did not implement the optional `delete-merged` input to delete the individual artifacts that are being merged, though I could add that as well if we want the updated workflow to be 1-to-1 to the current workflow.

PatTheMav:
Why did the artefact need to be merged in the first place? Is there a tool that ingests the generated appcasts automatically?

derrod:
> Why did the artefact need to be merged in the first place? Is there a tool that ingests the generated appcasts automatically?

Just for convenience and to avoid possible points of failure in the manual copying process (e.g. forgetting one arch).

RytoEX:
> Why did the artefact need to be merged in the first place? Is there a tool that ingests the generated appcasts automatically?

IIRC, it's to simplify the process of getting the appcasts on the server. There are different appcasts and deltas for each architecture, but they end up in the same path on the server. Instead of having to download and extract two archives onto the server, merging them means we only need to download and extract one archives.

RytoEX:
Set the `delete-merge` input parameter to `true` to (hopefully) get the same behavior we have now of getting a single `macos-sparkle-update` artifact from this workflow.

PatTheMav:
> > Why did the artefact need to be merged in the first place? Is there a tool that ingests the generated appcasts automatically?
> 
> Just for convenience and to avoid possible points of failure in the manual copying process (e.g. forgetting one arch).

So added code complexity just in case someone forgets to download both appcasts (just as someone might forget to download both app bundles)? Seems like a prime candidate of "cute, but unnecessary" to me.

RytoEX:
While I don't disagree that this adds some complexity, I think I'm going to merge this as-is so that we can get this action updated and hopefully remove the last of our actions deprecation warnings from first-party actions (some third-party actions must still be updated). If we want to rework our workflow/processes in the future to no longer have a single artifact, we can explore that separately. If this update causes things to fail, I'll take responsibility and fix it.

PR #10168, opened 2024-01-30
UI: Remove unused 'no scene' code

cg2121:
### Description
Since we require there to be always at least one scene, this 'no scene' code is never used.

### Motivation and Context
Remove unneeded code

### How Has This Been Tested?
Compiled OBS

### Types of changes

### Checklist:


Lain-B:
Apparently it is possible to create a state where there are no scenes, so I'd say we need to keep it until that point is no longer true.

Issue #9982, opened 2023-12-10
A770 QSV AV1 encoder doesn't work when RTX4090 installed at the same PC


DP-Huanhuan:
### Operating System Info

Windows 11

### OBS Studio Version

30.0.0

### OBS Studio Log URL

https://obsproject.com/logs/VxNFhER0pOkkr7d8

### Expected Behavior

The setup:
GPU 0: UHD 770 in 13900K
GPU 1: RTX 4090
GPU 2: Intel A770

Expected Behavior: When QSV encoders selected, tasks runs on GPU 2.  Access to Intel Hyper Encoding which allows to share task between iGPU and dGPU and the option to choose which GPU to use would be even better.

### Current Behavior

QSV H264&H265 runs on GPU 0. Zero activity on GPU 2. 
QSV AV1 fails because it is only supported by GPU 2. If all QSV encoder runs on GPU 0, it will certainly fail when it comes to AV1.

### Steps to Reproduce

1. Select QSV AV1 encoder in recording tab
2. Hit record button.
3. Error window pops up. It says "Starting the output failed"

### Anything else we should know?

Driver/OS/OBS everything is the latest.
There are two GPU from different vendors installed, which might be the cause.
Tried to disable the hardware accelerated GPU scheduling. Nothing changed.

jadenquinn:
I can't even get the QSV AV1 option to show up at all. Consider yourself lucky.

tphakala:
According to logs OBS selects RTX4090 as GPU to use

    Adapter 0: NVIDIA GeForce RTX 4090
    ....
    >>> app not on intel GPU, fall back to old qsv encoder

Have you followed these steps to select ARC 770 as GPU to use with OBS?
https://obsproject.com/kb/gpu-selection-guide

What comes to hyper encoding, is does not work with OBS. I have 12th gen CPU id HD730 iGPU and ARC 750 and AV1 runs on ARC adapter but it does not share load with iGPU.

RytoEX:
I can confirm this behavior occurs even on an Intel+Intel (iGPU+dGPU; Iris Xe + Arc A770M) system if OBS is running on the iGPU. The key factor here is that the GPU OBS is running on does not have QSV AV1 capabilities, and the non-texture sharing fallback is failing. This has been reported to Intel.

Selexo:
[...]

PR #10170, opened 2024-01-31
obs-qsv11: Fix for QSV AV1 in multi-GPU system

thyintel:
### Description

Fixing a bug where OBS is selected to run on a GPU that does not support AV1 encoding, while there is an Intel dGPU in the same system. Instead of throwing an error, due to the other GPU being unable to encode AV1, the system will fall back to the dGPU that can run AV1 encoding.

### Motivation and Context

Fixing this issue: https://github.com/obsproject/obs-studio/issues/9982 

### How Has This Been Tested?

Tested on CLEVO laptop with Intel iGPU and dGPU while OBS is selected to run on the iGPU.

### Types of changes

### Checklist:

- [x ] My code has been run through [clang-format](https://github.com/obsproject/obs-studio/blob/master/.clang-format).
- [x ] All commit messages are properly formatted and commits squashed where appropriate.
- [x ] I have included updates to all appropriate documentation.

### DIFF
--- a/plugins/obs-qsv11/QSV_Encoder.cpp
+++ b/plugins/obs-qsv11/QSV_Encoder.cpp
@@ -79,30 +79,7 @@ void qsv_encoder_version(unsigned short *major, unsigned short *minor)
 
 qsv_t *qsv_encoder_open(qsv_param_t *pParams, enum qsv_codec codec)
 {
-	obs_video_info ovi;
-	obs_get_video_info(&ovi);
-	size_t adapter_idx = ovi.adapter;
-
-	// Select current adapter - will be iGPU if exists due to adapter reordering
-	if (codec == QSV_CODEC_AV1 && !adapters[adapter_idx].supports_av1) {
-		for (size_t i = 0; i < 4; i++) {
-			if (adapters[i].supports_av1) {
-				adapter_idx = i;
-				break;
-			}
-		}
-	} else if (!adapters[adapter_idx].is_intel) {
-		for (size_t i = 0; i < 4; i++) {
-			if (adapters[i].is_intel) {
-				adapter_idx = i;
-				break;
-			}
-		}
-	}
-
-	bool isDGPU = adapters[adapter_idx].is_dgpu;
-
-	QSV_Encoder_Internal *pEncoder = new QSV_Encoder_Internal(ver, isDGPU);
+	QSV_Encoder_Internal *pEncoder = new QSV_Encoder_Internal(ver);
 	mfxStatus sts = pEncoder->Open(pParams, codec);
 	if (sts != MFX_ERR_NONE) {
 
@@ -189,12 +166,6 @@ qsv_t *qsv_encoder_open(qsv_param_t *pParams, enum qsv_codec codec)
 	}
 
 	return (qsv_t *)pEncoder;
-}
-
-bool qsv_encoder_is_dgpu(qsv_t *pContext)
-{
-	QSV_Encoder_Internal *pEncoder = (QSV_Encoder_Internal *)pContext;
-	return pEncoder->IsDGPU();
 }
 
 int qsv_encoder_headers(qsv_t *pContext, uint8_t **pSPS, uint8_t **pPPS,
--- a/plugins/obs-qsv11/QSV_Encoder_Internal.cpp
+++ b/plugins/obs-qsv11/QSV_Encoder_Internal.cpp
@@ -71,7 +71,7 @@ OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
 mfxHDL QSV_Encoder_Internal::g_DX_Handle = NULL;
 mfxU16 QSV_Encoder_Internal::g_numEncodersOpen = 0;
 
-QSV_Encoder_Internal::QSV_Encoder_Internal(mfxVersion &version, bool isDGPU)
+QSV_Encoder_Internal::QSV_Encoder_Internal(mfxVersion &version)
 	: m_pmfxSurfaces(NULL),
 	  m_pmfxENC(NULL),
 	  m_nSPSBufferSize(1024),
@@ -81,8 +81,8 @@ QSV_Encoder_Internal::QSV_Encoder_Internal(mfxVersion &version, bool isDGPU)
 	  m_nTaskIdx(0),
 	  m_nFirstSyncTask(0),
 	  m_outBitstream(),
-	  m_isDGPU(isDGPU),
-	  m_sessionData(NULL)
+	  m_sessionData(NULL),
+	  m_ver(version)
 {
 	mfxVariant tempImpl;
 	mfxStatus sts;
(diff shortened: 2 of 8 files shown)


plugins/obs-qsv11/QSV_Encoder_Internal.cpp
```diff
@@ -82,8 +82,10 @@ QSV_Encoder_Internal::QSV_Encoder_Internal(mfxVersion &version, bool isDGPU)
 	  m_nFirstSyncTask(0),
 	  m_outBitstream(),
 	  m_isDGPU(isDGPU),
-	  m_sessionData(NULL)
+	  m_sessionData(NULL),
+	  m_ver(version)
 {
+#if 0
```
RytoEX, line 88: Is this intentional, or is this leftover from debugging?

plugins/obs-qsv11/QSV_Encoder_Internal.cpp
```diff
@@ -948,7 +953,7 @@ mfxStatus QSV_Encoder_Internal::Reset(qsv_param_t *pParams,
 	mfxStatus sts = ClearData();
 	MSDK_CHECK_RESULT(sts, MFX_ERR_NONE, sts);
 
-	sts = Open(pParams, codec);
+	sts = Open(pParams, codec, 0);
```
RytoEX, line 956: When is Reset called and does hard-coding the adapter to 0 here have any repercussions?

RytoEX (changes requested):
Use a present tense verb as the first word after the commit message prefix.

Technically, the bug can occur when OBS is not running on an iGPU, but on any other GPU in the system that does not have the requested capabilities (QSV AV1, for example).

plugins/obs-qsv11/QSV_Encoder.cpp
```diff
@@ -81,18 +81,18 @@ qsv_t *qsv_encoder_open(qsv_param_t *pParams, enum qsv_codec codec)
 {
 	obs_video_info ovi;
```
kkartaltepe, line 82: This whole block down to initializing the internal and opening the encoder is windows specific, and `adapters` is queried in the platform specific `check_adapters` function.

Your change is mostly trying to plumb this data down into the platform specific `Initialize` implementation.

It seems it would be easier to move this code into common_utils_windows.cpp's Initialize and storing what you need in static globals (for example a pointer `adapters` array would be fine, we know its only initialized once at module load). Then you dont need to plumb anything, and we can make some later refactors to qsv more easily.

kkartaltepe: This should help resolve the issue with not having an adapter idx during `::Reset` as well.

plugins/obs-qsv11/QSV_Encoder_Internal.cpp
thyintel: Thank you for pointing this out! This was intentional, as it is slow to create a VPL session. The only use for the session here was to query the version being used. I have moved the query version to another VPL session creation and also moved the logging texture vs sysmem to open.

plugins/obs-qsv11/QSV_Encoder.cpp
thyintel: Thank you so much for this suggestion! It should be fixed now.

plugins/obs-qsv11/QSV_Encoder_Internal.cpp
```diff
@@ -143,13 +154,13 @@ mfxStatus QSV_Encoder_Internal::Open(qsv_param_t *pParams, enum qsv_codec codec)
 #if defined(_WIN32)
 	if (m_bUseD3D11)
 		// Use D3D11 surface
-		sts = Initialize(m_ver, &m_session, &m_mfxAllocator,
+		sts = Initialize(&m_ver, &m_session, &m_mfxAllocator,
```
kkartaltepe, line 157: We need to pass through the version all the way from initialization. By taking a reference here it means the pointer didnt get propagated correctly. And when starting an encoding session I see unitialized version data in the log

```
info:       major:          1
            minor:          0
```

Which breaks this check https://github.com/obsproject/obs-studio/blob/07a66fb226700da8fa14da8c6099a30ea2353312/plugins/obs-qsv11/obs-qsv11.c#L937

plugins/obs-qsv11/QSV_Encoder_Internal.cpp
kkartaltepe: I think it looks fine to delete the code you commented out now, thanks.

plugins/obs-qsv11/QSV_Encoder_Internal.cpp
kkartaltepe: It looks like this still the case with the latest commits.

plugins/obs-qsv11/QSV_Encoder_Internal.cpp
RytoEX: Looks like this has been removed now, so this thread can be resolved.

plugins/obs-qsv11/QSV_Encoder_Internal.cpp
thyintel: Just added new commit. Added a getVersion function to QSV_Encoder_Internal to propagate the version pointer correctly.

kkartaltepe: This leaves code with misleading and unused function behind, such as passing down the `&version` out parameter to `QSV_Encoder_Internal` but instead using `QSV_Encoder_Internal::getVersion` to extract the version.

Its a bit more work to actually cleanup this usage of out parameters, and will clutter the original commit you authored. So if you want to do that its best to do it as a separate commit in this PR, splitting the changes for this cleanup and the changes related to dgpu/adapter selection.

I think its simpler to leave the out parameter usage for a later cleanup and just make the member a pointer type, and leave a larger cleanup to later.

RytoEX: @thyintel As far as I can tell, this feedback still requires action. Do you have any questions about this feedback?

thyintel: So sorry for the wait. This issue should be resolved. We added back the MFXQueryVersion() back to when QSV_Encoder_Internal is initialized. We will look to clean this up later, as we should fix the adapter reordering first.

kkartaltepe: Please also revert changes to `Initialize` so it takes a `mfxVersion ver` since we dont need it with the `MFXQueryVersion` added back.

thyintel: Changed it back. I may not need the adapter_index variable anymore now, but let me know what you think. This should be the only change left, if needed?

kkartaltepe: It should be good now, if you have a log from this commit I can take a quick look or you can wait for me to test on an intel device in a bit.

kkartaltepe (approved):

Lain-B (approved):

PR #10171, opened 2024-01-31
UI: Fix null pointer dereference

CrackThrough:
### Description

NULL check for `obs_get_encoder_codec(option)` before calling `strcmp(it, "opus")`. `strcmp()` causes segfault when NULL is passed.

### Motivation and Context

I tried to use OBS for streaming to broadcast box with WebRTC, but OBS crashed while trying to load the configuration. This PR fixes it.

### How Has This Been Tested?

It seems this code is ran only when OBS is preparing to display window. I built with this change and I confirmed the change prevents OBS from crashing. I have confirmed it passes all tests.

### Types of changes

### Checklist:

### DIFF
--- a/UI/window-basic-main.cpp
+++ b/UI/window-basic-main.cpp
@@ -1431,7 +1431,9 @@ bool OBSBasic::LoadService()
 
 		option = config_get_string(basicConfig, "AdvOut",
 					   "AudioEncoder");
-		if (strcmp(obs_get_encoder_codec(option), "opus") != 0)
+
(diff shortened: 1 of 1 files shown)


Lain-B (approved):

RytoEX:
Just to clarify a bit, this occurs on the master branch when you change the stream service to WHIP, allow OBS to change the audio codec from the prompt, and then restart OBS. It seems like the audio codec setting for Advanced Output Streaming is not actually being saved to the config file, so the setting ends up being empty/null:
```ini
[AdvOut]
AudioEncoder=
```

PR #10185, opened 2024-02-02
CI: Fix Steam workflow getting confused by PDBs

derrod:
### Description

Fix Steam workflow failing because PDBs are now part of releases.

### Motivation and Context

Do not want CI runs to fail anymore :(

### How Has This Been Tested?

Tested on my fork.

### Types of changes

### Checklist:

### DIFF
--- a/.github/actions/steam-upload/action.yaml
+++ b/.github/actions/steam-upload/action.yaml
@@ -168,7 +168,7 @@ runs:
 
         print '::group::Prepare Windows x64 assets'
         mkdir -p steam-windows && pushd steam-windows
-        unzip ${root_dir}/(#i)obs-studio-*.zip
+        unzip ${root_dir}/(#i)obs-studio-(^*-pdbs).zip
         rm ${root_dir}/(#i)obs-studio-*.zip
 
         cp -r ${root_dir}/build-aux/steam/scripts_windows scripts


PatTheMav (commented):
I'd also not be opposed to making the macOS pattern more specific as well.

derrod:
Note: We currently omit the patch version number if it is 0 in Windows artifacts, will have to update this to support both cases with some regex I guess.

Adding the Mac cases should be easy enough.

derrod:
Updated to use some regex to shorten the version name in the filename if the patch version is `0`. Seems to work in a test run on my fork.

RytoEX:
> Updated to use some regex to shorten the version name in the filename if the patch version is `0`. Seems to work in a test run on my fork.

At some point in the future, I am in favor of simply having all release numbers be full major.minor.patch so we don't have to be concerned what kind of release it is to know the version number in artifacts.

For now, while this makes some assumptions (`major` is always two digits, `minor` is always 1 digit, we only do up to 9 betas/rcs), this is probably enough to fix the Steam workflow for now.

PatTheMav:
> For now, while this makes some assumptions (major is always two digits, minor is always 1 digit, we only do up to 9 betas/rcs), this is probably enough to fix the Steam workflow for now.

Yeah I would not want to merge this, because the moment we fix our Windows artefact generation and Windows uses canonical versioning (and follows the rules) this workflow would break.. again.

Instead of doctoring around the symptom, why not use a better glob expression in the part that unzips the build, e.g. `(#i)obs-studio-(^*-PDBs).zip` which should only extract a zip that will not contain the `PDBs` suffix?

RytoEX:
> Yeah I would not want to merge this, because the moment we fix our Windows artefact generation and Windows uses canonical versioning (and follows the rules) this workflow would break.. again.

My expectation would be that when we fix the Windows artifact names across the board, we revert any changes necessary.

> Instead of doctoring around the symptom, why not use a better glob expression in the part that unzips the build, e.g. `(#i)obs-studio-(^*-PDBs).zip` which should only extract a zip that will not contain the `PDBs` suffix?

I'd be fine with this. I assume @derrod and I aren't as familiar with zsh globs as you are, so neither of us considered it.

derrod:
> > Instead of doctoring around the symptom, why not use a better glob expression in the part that unzips the build, e.g. `(#i)obs-studio-(^*-PDBs).zip` which should only extract a zip that will not contain the `PDBs` suffix?
> 
> I'd be fine with this. I assume @derrod and I aren't as familiar with zsh globs as you are, so neither of us considered it.

Yeah I'm not familiar with zsh, if that's an acceptable solution we can just do that instead.

PatTheMav:
> > > Instead of doctoring around the symptom, why not use a better glob expression in the part that unzips the build, e.g. `(#i)obs-studio-(^*-PDBs).zip` which should only extract a zip that will not contain the `PDBs` suffix?
> > 
> > 
> > I'd be fine with this. I assume @derrod and I aren't as familiar with zsh globs as you are, so neither of us considered it.
> 
> Yeah I'm not familiar with zsh, if that's an acceptable solution we can just do that instead.

I had to Google for "negate glob expression" myself I just expected it to be possible somehow. Are the PDB artefacts actually called `OBS-Studio-<VERSION>-PDBs.zip`? Then the example should work..

derrod:
> `OBS-Studio-30.1-beta1-pdbs.zip`, though I'm not sure if that was automatically or manually named that way.

Assuming Lain doesn't manually rename them it's configured here: https://github.com/derrod/bouf/blob/master/extra/config.example.toml#L132-L133

PatTheMav:
> OBS-Studio-30.1-beta1-pdbs.zip

Got it - tested the glob expression locally, had `OBS-Studio-30.0.1-PDBs.zip`, `OBS-Studio-30.0.1.zip`, and `OBS-Studio-30.1-beta1-pdbs.zip` in the directory and it yielded only `OBS-Studio-30.0.1.zip`. Also tested with `OBS-Studio-30.1-beta1.zip` available and it ignored the PDB variants.

PatTheMav:
> > `OBS-Studio-30.1-beta1-pdbs.zip`, though I'm not sure if that was automatically or manually named that way.
> 
> Assuming Lain doesn't manually rename them it's configured here: https://github.com/obsproject/bouf/blob/master/extra/config.example.toml#L132-L133

Yeah `(#i)obs-studio-(^*-pdbs).zip` should suffice, the expression is already case-insensitive, so no need to make the token itself uppercase either.

derrod:
Updated and tested, still works.

PatTheMav (approved):

RytoEX (approved):

PR #10186, opened 2024-02-02
CI: Move Windows signing to reusable workflow

derrod:
### Description

In order for the `job_workflow_sha` claim in the JWT to actually correspond to the workflow's last modification commit it has to be a reusable workflow. This is annoying, and somewhat silly given GitHub's documentation indicates otherwise, but what can you do...

Bonus points for there not being a way to get the ref of the currently executed reusable workflow, the only way I ([and others](https://github.com/actions/runner/issues/2417)) could find was to parse the JWT, which is "easy" here since this job uses it anyway.

Note: this PR needs to be merged via merge commit to avoid the hashes becoming invalid. Alternatively the second commit could be split into its own or directly pushed to master, but I'd like to not do that.

### Motivation and Context

Want to pin specific version of signing action/bouf configuration for Google Cloud Authentication.

### How Has This Been Tested?

Ran on my fork.

### Types of changes

### Checklist:

### DIFF
--- a/.github/workflows/push.yaml
+++ b/.github/workflows/push.yaml
@@ -207,8 +207,8 @@ jobs:
 
   sign-windows-build:
     name: Windows Signing ✍️
-    uses: ./.github/workflows/sign-windows.yaml
-    if: github.ref_type == 'tag'
+    uses: obsproject/obs-studio/.github/workflows/sign-windows.yaml@7dcf55274568837c37ca2d869e37b83b9ff3e7cf
+    if: github.repository_owner == 'obsproject' && github.ref_type == 'tag'
     needs: build-project
     permissions:
       contents: 'read'
--- a/.github/workflows/push.yaml
+++ b/.github/workflows/push.yaml
@@ -205,94 +205,15 @@ jobs:
           pattern: macos-sparkle-update-*
           delete-merged: true
 
-  create-windows-update:
-    name: Create Windows Update 🥩
-    if: github.repository_owner == 'obsproject' && github.ref_type == 'tag'
-    runs-on: windows-2022
+  sign-windows-build:
+    name: Windows Signing ✍️
+    uses: ./.github/workflows/sign-windows.yaml
+    if: github.ref_type == 'tag'
     needs: build-project
     permissions:
       contents: 'read'
       id-token: 'write'
-    defaults:
-      run:
-        shell: pwsh
-    environment:
-      name: bouf
-    steps:
-      - uses: actions/checkout@v4
-        with:
-          path: "repo"
-          fetch-depth: 0
-          ref: ${{ github.ref }}
-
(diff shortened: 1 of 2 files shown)


.github/workflows/push.yaml
```diff
@@ -205,94 +205,15 @@ jobs:
           pattern: macos-sparkle-update-*
           delete-merged: true
 
-  create-windows-update:
-    name: Create Windows Update 🥩
+  sign-windows-build:
+    name: Windows Signing ✍️
+    uses: obsproject/obs-studio/.github/workflows/sign-windows.yaml@744443d67fda3b40bd763206d7c072c9c1cb9981
```
PatTheMav, line 210:
```suggestion
    uses: .github/workflows/sign-windows.yaml@744443d67fda3b40bd763206d7c072c9c1cb9981
```

Shouldn't that be sufficient or does that also interfere with the hash check somehow?

derrod: Based on this it has to include the repo if you want to specify a commit hash: https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions#jobsjob_iduses

PatTheMav: Got it, so we have to use it to ensure the correct hash is used.

RytoEX: It sounds like this is concern is resolved then?

.github/workflows/push.yaml
```diff
@@ -205,94 +205,15 @@ jobs:
...
-          $folderHash = ''
-          $files = Get-ChildItem "${{ github.workspace }}\repo\.github\actions\bouf"
-          foreach ($file in $files) {
-            $folderHash += (Get-FileHash $file -Algorithm SHA256).Hash
-          }
-          # This is stupid but so is powershell
-          $stream = [IO.MemoryStream]::new([byte[]][char[]]$folderHash)
-          if ((Get-FileHash -InputStream $stream -Algorithm SHA256).Hash -ne "$env:BOUF_ACTION_HASH") {
-            throw "bouf action folder hash does not match."
-          }
```
RytoEX, line 248: Do we no longer need to check the bouf action hash?

derrod: Wince it's now pinned to a specific commit hash, no.

RytoEX: I thought `sign-windows.yaml` is pinned to a specific commit, but not the bouf action, unless I'm missing something.

RytoEX: Ah, I think I get it? The entire repo is pinned to a specific commit on checkout, and _that_ version of the action is used?

derrod: Yep.

RytoEX (approved):

PR #10211, opened 2024-02-06
docs: Add deque to the table of content and fix ROI underline

tytan652:
### Description

Fixes warnings that sphinx reported that comes from "recent" changes:
- Add deque to libobs/util table of content
- Fix title underline of obs_encoder_roi

Also fix a deprecation note for circlebuf that was set to a placeholder version (`3X.0`).

### Motivation and Context

Fix docs warnings right after I found them.

### How Has This Been Tested?

`make html` and Sphinx didn't complain of the issues fixed by the changes and deque is shown in the libobs/util TOC.

### Types of changes

### Checklist:

### DIFF
--- a/docs/sphinx/reference-libobs-util.rst
+++ b/docs/sphinx/reference-libobs-util.rst
@@ -9,6 +9,7 @@ Platform/Utility API Reference (libobs/util)
    reference-libobs-util-circlebuf
    reference-libobs-util-config-file
    reference-libobs-util-darray
+   reference-libobs-util-deque
    reference-libobs-util-dstr
    reference-libobs-util-platform
    reference-libobs-util-profiler
--- a/docs/sphinx/reference-encoders.rst
+++ b/docs/sphinx/reference-encoders.rst
@@ -267,7 +267,7 @@ Raw Frame Data Structure (encoder_frame)
 
 
 Encoder Region of Interest Structure (obs_encoder_roi)
------------------------------------------
+------------------------------------------------------
 
 .. struct:: obs_encoder_roi
 
@@ -277,7 +277,7 @@ Encoder Region of Interest Structure (obs_encoder_roi)
             uint32_t bottom
             uint32_t left
             uint32_t right
-   
+
    The rectangle edges of the region are specified as number of pixels from the input video's top and left edges (i.e. row/column 0).
 
 .. member:: float priority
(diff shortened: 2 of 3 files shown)


docs/sphinx/reference-libobs-util.rst
```diff
@@ -8,6 +8,7 @@ Platform/Utility API Reference (libobs/util)
    reference-libobs-util-bmem
    reference-libobs-util-circlebuf
```
derrod, line 9: Probably makes sense to remove this as well then.

derrod (approved):

docs/sphinx/reference-libobs-util.rst
tytan652: If it's still in available in the API, it doesn't make sense to remove it even if it's "deprecated".
But the whole page should been updated to have all the function shown as deprecated, which is out of the scope that I want to keep for this PR.

RytoEX: @derrod Seems resolved?

docs/sphinx/reference-libobs-util.rst
```diff
@@ -8,6 +8,7 @@ Platform/Utility API Reference (libobs/util)
...
    reference-libobs-util-config-file
+   reference-libobs-util-deque
    reference-libobs-util-darray
```
RytoEX, line 12:
```suggestion
   reference-libobs-util-darray
   reference-libobs-util-deque
```
Alphabetization.

RytoEX (changes requested):

docs/sphinx/reference-libobs-util.rst
tytan652: No because:
- Double-ended queue
- Dynamic Array
- Dynamic String

RytoEX: Are these _really_ alphabetized by their display name and not the name of the page?

tytan652: Realized that indeed those are alphabetized based on their identifiers and not the documentation name.

RytoEX: If the HTML rendered order ends up causing confusion, blame me, and we'll figure out something.

RytoEX (approved):

Issue #9824, opened 2023-11-08
Graphical Artifacts In Audio Mixer on Wayland


5p4r74cu5:
### Operating System Info

Other

### Other OS

Debian 12

### OBS Studio Version

29.1.3

### OBS Studio Log URL

https://obsproject.com/logs/7S6BFdLcL6LyNhn7

### Expected Behavior

No visual artifacts.

### Current Behavior

There are minor visual artifacts in the levels of the audio mixer, in this screenshot they are visible as red, and change colours, white, blue, etc. Some time the artifacting is worse, but generally the meters are clear enough to be usable.

![Screenshot_20231108_154256](https://github.com/obsproject/obs-studio/assets/26020949/be85f19b-15b1-4e12-99d5-eff447658752)

### Steps to Reproduce

Install the OBS flatpak.

### Anything else we should know?

KDE Plasma v5.27.5 in Wayland session.

kkartaltepe:
There is a minor rendering issue with the space between the meters where nothing is rendered. This means its usually transparent in modern renders and displays whatever is behind your obs.

v4u6h4n:
Ah I see, makes sense.

RytoEX:
Fixed by #10223

5p4r74cu5:
Thanks :-)

PR #5233, opened 2021-09-01
Improve scaling of audio meter (issue 3915)

OldBaldGeek:
### Description

My branch has two commits, as recommended by the fine folks on OBS Discord.

The first addresses issue 3915 by replacing a cached bitmap with selective repaint. This helps with legibility at scaled DPI, and also eliminates unnecessary repainting of the scale area.

V27.0.1 (using cached bitmap). Left 100%, right 125% scaling
Q V27

First commit. Left 100%, right 125% scaling
Q fix1

The second commit allows the appearance of the meter to be adjusted via the stylesheet. This permits the appearance to be optimized by the user to deal with scaling or other needs and preferences.

Second commit. Left 100%, right 125% scaling
Q fix2

Second commit, vertical orientation. Left 100%, right 125% scaling
Q fix2 vertical

Second commit, stylesheet specifies meter bar 9 pixels wide, silly font at 150%
Q fix2 silly at 100

Minor changes
- A horizontal meter has lots of space for the label, so the default label font is used (V27 used 1px smaller). A vertical meter has less space for a label, so the font size is reduced to 80% of the default.
- For vertical meters, the number showing the fader setting was centered on the meter/fader pair;  which made it somewhat ambiguous whether it represented fader or audio level.  Changed to be right-justified, putting the number over the fader.

### Motivation and Context

Eliminating the bitmap cache improves the appearance of the meter when scaled, fixes #3915

But in my opinion, the underlying problem is that the font used for the volume meter scale is just too small and is not adjustable. My second commit draws the scale numbers using the font settings for VolumeMeter, or its base class QWidget. It adds a qproperty to the stylesheet to adjust the size of the font by a percentage (80% default). It also adds a qproperty to specify the thickness of the meter bar in pixels, which might be used to match the width of the volume fader, etc.

[https://github.com/obsproject/obs-studio/issues/3915](url)

### How Has This Been Tested?

Tested on Windows 10 version 21H1 plus updates
- tested at 100% and 125% scaling, horizontal and vertical meters
- tested with existing stylesheets without added properties
- tested with updated stylesheets with added properties

### Types of changes

- Bug fix (non-breaking change which fixes an issue)
- New feature (non-breaking change which adds functionality) second commit adding stylesheet properties
- (minor) Performance enhancement (non-breaking change which improves efficiency) elimination of copying cached bitmap may offer a small performance improvement

### Checklist:

- [x] My code has been run through [clang-format](https://github.com/obsproject/obs-studio/blob/master/.clang-format).
[...]

PR #10223, opened 2024-02-09
UI: Fill audio meter background each update (#9842)

fzwoch:
### Description

Since the audio mixer widget is marked as being opaque it is necessary to draw all of it's pixels in the drawing region. Pixels not being drawn will leave transparent pixels (on Wayland at least) causing undesired effects.

Previously, the background was not filled when only the meters were repainted. Since the meters don't paint the complete region (spacing, padding etc.) some pixels in the region ended up not being painted.

### Motivation and Context

Addressing https://github.com/obsproject/obs-studio/issues/9824

### How Has This Been Tested?

Locally run on GNOME/Wayland which was affected by the bug. After applied fix, the audio mixer looks correctly.

Hopefully this does not regress other platforms..

### Types of changes

### Checklist:

### DIFF
--- a/UI/volume-control.cpp
+++ b/UI/volume-control.cpp
@@ -1417,6 +1417,10 @@ void VolumeMeter::paintEvent(QPaintEvent *event)
 
 	QPainter painter(this);
 
+	// Paint window background color (as widget is opaque)
+	QColor background = palette().color(QPalette::ColorRole::Window);
+	painter.fillRect(event->region().boundingRect(), background);
+
 	if (vertical)
 		height -= METER_PADDING * 2;
 
@@ -1425,11 +1429,6 @@ void VolumeMeter::paintEvent(QPaintEvent *event)
 	if (event->region().boundingRect() != getBarRect()) {
 		if (needLayoutChange())
 			doLayout();
-
-		// Paint window background color (as widget is opaque)
-		QColor background =
-			palette().color(QPalette::ColorRole::Window);
-		painter.fillRect(widgetRect, background);
 
 		if (vertical) {
 			paintVTicks(painter,


RytoEX:
I feel like the audio mixer painting was done in a very particular way due to it being very heavy on macOS. @PatTheMav ?

kkartaltepe (approved):

WizardCM:
* See #5233 for some context behind the current behaviour.

PatTheMav:
Yeah this PR needs to be checked whether it introduces a performance regression on Retina screens, especially with 120Hz ProMotion displays.

PatTheMav:
Without checking the power impact of the change (suggesting that the CPU clocks are increased) the CPU time spent on the volume meters is not significantly different with this change applied. Still the second heaviest stack in the program at idle use, but that's also the case without the change.

RytoEX:
> * See [Improve scaling of audio meter (issue 3915) #5233](https://github.com/obsproject/obs-studio/pull/5233) for some context behind the current behaviour.

In particular, I was thinking of #5233 and commit https://github.com/obsproject/obs-studio/commit/f4f2d383b1f234dc231ca0dc2804d8ba5c69635e, the latter of which specifically called out performance issues when we changed the volume meters from updating at 30Hz to 60Hz.

> Without checking the power impact of the change (suggesting that the CPU clocks are increased) the CPU time spent on the volume meters is not significantly different with this change applied. Still the second heaviest stack in the program at idle use, but that's also the case without the change.

If there are no noticeable performance regressions though, it sounds like this should be fine?

Lain-B (approved):

WizardCM (approved):

RytoEX (approved):
Seems okay. We'll give this a try and have to keep an eye out for any performance regressions (mostly on macOS).

PR #10230, opened 2024-02-10
obs-ffmpeg: Receive packets while frame queue is full

kkartaltepe:
### Description

Previously we always assumed we could push more frames before an encoded packet would be available. Actually we can get EAGAIN when the frame queue is full and before any encoded packets have been completed.

This changes attempts to read packets more often as the worst that can happen is we get EAGAIN again (i guess).

### Motivation and Context

Fixes a hang in ffmpeg vaapi encoder on intel.

### How Has This Been Tested?

The resulting recording looks correct after enabling bframes on intel's av1 drivers.

### Types of changes

### Checklist:

### DIFF
--- a/plugins/obs-ffmpeg/obs-ffmpeg-vaapi.c
+++ b/plugins/obs-ffmpeg/obs-ffmpeg-vaapi.c
@@ -543,7 +543,7 @@ static bool vaapi_encode(void *data, struct encoder_frame *frame,
 	}
 
 	ret = avcodec_send_frame(enc->context, hwframe);
-	if (ret == 0)
+	if (ret == 0 || ret == AVERROR(EAGAIN))
 		ret = avcodec_receive_packet(enc->context, enc->packet);
 
 	got_packet = (ret == 0);


plugins/obs-ffmpeg/obs-ffmpeg-vaapi.c
```diff
@@ -543,7 +543,7 @@ static bool vaapi_encode(void *data, struct encoder_frame *frame,
 	}
 
 	ret = avcodec_send_frame(enc->context, hwframe);
-	if (ret == 0)
+	if (ret == 0 || AVERROR(EAGAIN))
```
notr1ch, line 546: Shouldn't this be `ret == AVERROR(EAGAIN)`?

kkartaltepe: That is what I thought I typed, fixed and retested. Thanks.

RytoEX (changes requested):
Commit message nits:
* Previously -> Previously,
* Actually -> Actually,
* This changes -?> This change

The first two are just grammar nits. The last one I wasn't sure if it meant to say "This change" or not, hence the question mark.

As this is contained to VAAPI, I'm otherwise fine with the PR.

kkartaltepe:
Fixed up the commit message.

Lain-B (approved):

PR #10232, opened 2024-02-10
Use flathub-infra images & various other cleanups

GeorgesStavracas:
### Description

It's a ultimately a temporary workaround while the whole situation with libappstream, Flatpak, and flatpak-builder isn't figured out. But it does seem to work.

### Motivation and Context

The Flathub publishing workflow is broken due to an odd flatpak-builder / appstream impedance mismatch, and this works around that.

### How Has This Been Tested?

The publishing job goes beyond the build step: https://github.com/GeorgesStavracas/obs-studio/actions/runs/7888002165/job/21524497985 (CI fails because I don't have the OBS Studio publishing tokens in my repo, but that's not important)

### Types of changes

### Checklist:

### DIFF
--- a/.github/actions/flatpak-builder-lint/action.yaml
+++ b/.github/actions/flatpak-builder-lint/action.yaml
@@ -52,15 +52,9 @@ runs:
         done
 
         n_errors=$(echo $ret | jq '.errors | length')
-        for ((i = 0 ; i < $(echo $ret | jq '.errors | length') ; i++)); do
+        for ((i = 0 ; i < $n_errors ; i++)); do
           error=$(echo $ret | jq ".errors[$i]")
-
-          if [[ "${{ inputs.validateToPublish }}" == "false" && "${error//\"}" == "appstream-screenshots-not-mirrored" ]]; then
-            echo "::notice::$error found and ignored in the Flatpak ${{ inputs.artifact }}"
-            n_errors=$(($n_errors-1))
-          else
-            echo "::error::$error found in the Flatpak ${{ inputs.artifact }}"
-          fi
+          echo "::error::$error found in the Flatpak ${{ inputs.artifact }}"
         done
 
         [[ $n_errors == 0 ]] || exit 2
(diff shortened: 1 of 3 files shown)


tytan652 (commented):
I'm *fine* with it as long we don't switch to the pre-included `flatpak-builder-lint` installation for this image.

tytan652:
I forgot, we need Github CLI for our CIs.

.github/actions/flatpak-builder-lint/action.yaml
tytan652, line 1: v2.1.5 is a broken version to be used with a JSON parser

tytan652:
You can check #10234 for the needed error filter change.

.github/workflows/build-project.yaml
tytan652, line 1: Beside swapping images and removing dnf usage, the action as no need to be swapped to the fork.

GeorgesStavracas: Done

tytan652:
Note: `CI: Use flatpak-builder-lint from the image` can break CI if an update of the tool from upstream introduces new errors.

GeorgesStavracas:
> Note: CI: Use flatpak-builder-lint from the image can break CI if an update of the tool from upstream introduces new errors.

Yes. In which case, we adjust to the linter error.

Fenrirthviti:
This workaround seems fine, and further discussions around the best long-term solution can happen off-thread. This gets us builds working again. 

It seems we all collectively agree this isn't ideal, but it gets us from broken to not broken for now, and shouldn't have any meaningful impact on the longer term solutions. We accept the risk that this might not be as stable as other solutions, but those require more time we don't have, as we need testing ASAP for our shortened beta period.

RytoEX (changes requested):
Nit on second commit:
* setup -> set up

Other than that, so long as each commit works on its own, these are fine.

RytoEX (approved):

Issue #10218, opened 2024-02-08
Pausing and unpausing fast adds more "(PAUSE)"s to the bottom


Count-MHM:
### Operating System Info

Windows 11

### OBS Studio Version

30.0.2

### OBS Studio Log URL

https://obsproject.com/logs/DPoavmnpbs3t6UGL

### Expected Behavior

I don't think it should ever write "(PAUSED)" more than once
![Correct](https://github.com/obsproject/obs-studio/assets/71083058/c083d50e-4ea4-4ef3-b985-50cc536eecf9)

### Current Behavior

![Current](https://github.com/obsproject/obs-studio/assets/71083058/e6cfdfe3-19ae-4dec-b0ed-f9ec6f4f070b)

### Steps to Reproduce

1. Start recording
2. Pause Recording
3. Unpausing and Pausing Recording rapidly

PR #10241, opened 2024-02-13
UI: Fix paused text stacking

EmericCaramanna:
### Description
Whenever the pause button was clicked rapidly the " (PAUSED)" string next to the recording timestamp was repeated.
As shown is this screenshot:
image
I simply added a condition to prevent changing the text if the " (PAUSED)" string is already appended to the timestamp
image

### Motivation and Context
Just a simple UI contribution to fix issue #10218 

### How Has This Been Tested?
Tested on windows 10.
OBS version 30.0.2.
Simply clicked rapidly the pause button to attest that the string didn't appended more than once.

### Types of changes

### Checklist:

### DIFF
--- a/UI/window-basic-status-bar.cpp
+++ b/UI/window-basic-status-bar.cpp
@@ -550,9 +550,13 @@ void OBSBasicStatusBar::RecordingStopped()
 
 void OBSBasicStatusBar::RecordingPaused()
 {
-	QString text = statusWidget->ui->recordTime->text() +
-		       QStringLiteral(" (PAUSED)");
-	statusWidget->ui->recordTime->setText(text);
+	if (!statusWidget->ui->recordTime->text().contains(
+		    QStringLiteral(" (PAUSED)"))) {
+		QString text = statusWidget->ui->recordTime->text() +
+			       QStringLiteral(" (PAUSED)");
+
+		statusWidget->ui->recordTime->setText(text);
+	}
 
 	if (recordOutput) {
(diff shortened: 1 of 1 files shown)


gxalpha:
I think this is the wrong fix, as it addresses the symptoms instead of the cause.
The issue in my view is that whatever triggers the pause method doesn't check whether the recording is already paused, which I think should be fixed instead. I vaguely remember an old bug report (but can't find it right now) of where using the frontend API for example could also cause this.

RytoEX:
> I think this is the wrong fix, as it addresses the symptoms instead of the cause. The issue in my view is that whatever triggers the pause method doesn't check whether the recording is already paused, which I think should be fixed instead. I vaguely remember an old bug report (but can't find it right now) of where using the frontend API for example could also cause this.

You're probably thinking of the various "stopping recording or replay buffer failed" issues, which can also be caused by repeatedly clicking the button too fast.

Fenrirthviti:
Closing this since this is not an approach we are willing to accept.

As mentioned, this should be investigated to identify why the states are not being properly communicated to the UI.

Since this really shouldn't ever be triggered in normal operation (users don't typically spam the pause button), merging a band-aid fix is not necessary.

Aleksbgbg:
I think this is now correctly fixed in #10308 - although I am open to suggestions.

The issue was that the text was never undone, rather waiting for the UI timer to re-generate the timestamp - but I avoided the issue by just re-generating the text every time, rather than mutating the string.

As for the pause method, I believe it is correctly guarded against duplicate invocations - although I have not looked at any other related issues.

PR #10243, opened 2024-02-14
win-capture: Fix leak in wasapi reroute proc call

derrod:
### Description

Fixes a memory leak.

### Motivation and Context

Leaks bad.

### How Has This Been Tested?

Ran.

### Types of changes

### Checklist:

### DIFF
--- a/plugins/win-capture/audio-helpers.c
+++ b/plugins/win-capture/audio-helpers.c
@@ -22,10 +22,10 @@ static inline void reroute_wasapi_source(obs_source_t *wasapi,
 					 obs_source_t *target)
 {
 	proc_handler_t *ph = obs_source_get_proc_handler(wasapi);
-	calldata_t *cd = calldata_create();
-	calldata_set_ptr(cd, "target", target);
-	proc_handler_call(ph, "reroute_audio", cd);
-	calldata_free(cd);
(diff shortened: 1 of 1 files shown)


RytoEX:
These functions are not documented, so I had to pull up the commit that added them:
https://github.com/obsproject/obs-studio/commit/e3d19c5b16a08aa0cdcd58962bbc74cf87984ce6
>  libobs/callback: Add calldata_create and calldata_destroy
> 
> These functions aren't particularly useful in C/C++, but these functions
> are necessary to allow creating/destroying a calldata_t object via
> script.

Looking at [the one existing usage](https://github.com/obsproject/obs-studio/blob/062813eb8a352dacb97d23bddcaf776b37d37ee8/UI/frontend-plugins/frontend-tools/data/scripts/instant-replay.lua#L18-L22) in obs-studio, it is apparently that `calldata_create` must be paired with `calldata_destroy`. Apparently, that was only introduced for scripting, and the more common (expected?) form is to use `calldata_init` and `calldata_free`.

derrod:
Yeah this could just be stack allocated here instead.

RytoEX (approved):
This is a pattern used elsewhere, so this seems fine.

PR #10250, opened 2024-02-15
UI: Use signal vector for status bar

cg2121:
### Description
The stream output signals are now put inside of a vector.

### Motivation and Context
Less code

### How Has This Been Tested?
Disconnected and reconnected stream

### Types of changes

### Checklist:


PatTheMav:
I guess this change relies on the deallocators of the signals being called when the vector is cleared (which it does).

In general I'm not in favour of relying on deallocators in C++ (because C++ is a mess) versus calling cleanup code explicitly, but it's the way `OBSSignal` is designed.

PR #10251, opened 2024-02-15
UI: Don't reserve signal vector size

cg2121:
### Description
In no other place in the program do we reserve the vector size when using them for signals. This creates an extra step
if signals are added or removed.

### Motivation and Context
One less extra step to worry about

### How Has This Been Tested?
Ran OBS and everything still worked

### Types of changes

### Checklist:

### DIFF
--- a/UI/window-basic-main.cpp
+++ b/UI/window-basic-main.cpp
@@ -1856,7 +1856,6 @@ void OBSBasic::InitOBSCallbacks()
 {
 	ProfileScope("OBSBasic::InitOBSCallbacks");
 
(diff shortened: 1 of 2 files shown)


RytoEX:
This is an optimization to ensure that the vector does not need to be reallocated multiple times since we know in advance how many items we're going to add. I feel it is incumbent on people editing around these areas to increase or decrease the additional capacity reserved. If anything, this could have a clarifying comment, but it seems fairly apparent to me what this is supposed to accomplish.

cg2121:
As they are used for an optimization (I didn't know this before), it is not that big of a deal to still reserve the size.

Issue #10249, opened 2024-02-15
OBS 30.1.0 beta2 crashes while starting


viaregio:
### Operating System Info

macOS 14

### OBS Studio Version

Other

### OBS Studio Version (Other)

30.1.0 beta2

### OBS Studio Log URL

https://obsproject.com/logs/RkTXuBOskT8xLL6v

### OBS Studio Crash Log URL

https://obsproject.com/logs/RkTXuBOskT8xLL6v

### Expected Behavior

OBS 30.0.2 is starting without a crash.

Logoutput:
```
07:05:07.959: [Safe Mode] Unclean shutdown detected!
07:05:09.903: [Safe Mode] User elected to launch normally.
07:05:09.921: [macOS] Permission for audio device access granted.
07:05:09.926: [macOS] Permission for video device access granted.
07:05:09.931: [macOS] Permission for accessibility granted.
07:05:09.935: [macOS] Permission for screen capture granted.
07:05:09.935: CPU Name: Apple M1 Pro
07:05:09.935: Physical Cores: 10, Logical Cores: 10
07:05:09.935: Physical Memory: 32768MB Total
07:05:09.935: OS Name: macOS
07:05:09.936: OS Version: Version 14.3.1 (Build 23D60)
07:05:09.936: Rosetta translation used: false
07:05:09.936: Kernel Version: 23.3.0
07:05:09.937: hotkeys-cocoa: Using layout 'com.apple.keylayout.German'
07:05:09.938: Current Date/Time: 2024-02-15, 07:05:09
07:05:09.938: Browser Hardware Acceleration: true
07:05:09.938: Qt Version: 6.5.3 (runtime), 6.5.3 (compiled)
07:05:09.938: Portable mode: false
07:05:09.992: OBS 30.0.2 (mac)
07:05:09.992: ---------------------------------
07:05:09.993: ---------------------------------
07:05:09.993: audio settings reset:
07:05:09.993: 	samples per sec: 48000
07:05:09.993: 	speakers:        2
07:05:09.993: 	max buffering:   960 milliseconds
07:05:09.993: 	buffering type:  dynamically increasing
07:05:09.996: ---------------------------------
07:05:09.996: Initializing OpenGL...
07:05:10.024: Loading up OpenGL on adapter Apple Apple M1 Pro
07:05:10.024: OpenGL loaded successfully, version 4.1 Metal - 88, shading language 4.10
07:05:10.199: ---------------------------------
07:05:10.199: video settings reset:
[...]

PR #10253, opened 2024-02-15
mac-capture: Limit number of copied audio channels to libobs limits

PatTheMav:
### Description
Limit amount of audio buffers copied into libobs' audio pointer array to amount of channels supported by libobs.

### Motivation and Context
OBS handles up to 8 channels of audio, which requires the CoreAudio input callback to omit/ignore any audio data from channels above that count and also report a maximum amount of 8 channels to match the speaker setup given to swresample by libobs.

This also means that for devices with more than 8 channels the only way to get that audio in OBS is to use the channel mapping feature and manually select those channels.

### How Has This Been Tested?
Tested on macOS 14.3.1 with BlackHole 16-ch (which provides the necessary channel count above 8) and also confirmed that audio sent on channels 15 and 16 of it are properly handled when using channel mapping.

Fixes https://github.com/obsproject/obs-studio/issues/10249

### Types of changes

### Checklist:

### DIFF
--- a/plugins/mac-capture/mac-audio.c
+++ b/plugins/mac-capture/mac-audio.c
@@ -467,11 +467,16 @@ static OSStatus input_callback(void *data,
 	if (!ca_success(stat, ca, "input_callback", "audio retrieval"))
 		return noErr;
 
-	for (UInt32 i = 0; i < ca->buf_list->mNumberBuffers; i++)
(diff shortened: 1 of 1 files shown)


RytoEX (approved):
Looks fine at a glance.

PR #10254, opened 2024-02-15
mac-capture: Fix crash in device reconnect handler

PatTheMav:
### Description
Splits the memory cleanup for channel maps and channel names in macOS audio capture source.

### Motivation and Context
The uninit function prematurely released the memory allocated for the channel map setting, which exists in the scope of source life cycle (compared to channel names which are valid during the life cycle of a configured device).

Splitting up the clean-up for both (memory for channel names is released when the device is uninitialised, memory for channel map setting is released when the source is removed) ensures that the memory is released but pointers don't become unexpectedly invalid.

### How Has This Been Tested?
Tested on macOS 14.3.1, disconnecting and reconnecting a device with active channel map setting, then checking for leaked memory on shutdown.

### Types of changes

### Checklist:

### DIFF
--- a/plugins/mac-capture/mac-audio.c
+++ b/plugins/mac-capture/mac-audio.c
@@ -811,11 +811,6 @@ static void coreaudio_uninit(struct coreaudio_data *ca)
 		bfree(ca->channel_names);
 		ca->channel_names = NULL;
 	}
-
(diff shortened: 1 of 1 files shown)


RytoEX (approved):
Looks good to me.

PR #10255, opened 2024-02-15
mac-avcapture: Add color format selection for capture card source

PatTheMav:
### Description
Adds colour format selection to Capture Card source to unbreak situations where specific resolutions do not support the same colour format as the device format the source was created with.

### Motivation and Context
Some devices do not support all color formats at all resolutions, but the capture card source automatically uses an available color format and compares it against the available format for a specific resolution.

Without being able to change this format, some resolutions do not work as CMIO will not be able to find a compatible format. Thus the color format needs to be manually selected for capture card sources as well.

### How Has This Been Tested?
Tested on macOS 14.3.1 with Elgato HD 60x which supports NV12 and YUY2 up to 1080p but only supports NV12 for 2560x1440 and 3840x2160.

### Types of changes

### Checklist:

### DIFF
--- a/plugins/mac-avcapture/OBSAVCapture.m
+++ b/plugins/mac-avcapture/OBSAVCapture.m
@@ -448,8 +448,11 @@
             return NO;
         }
     } else {
+        int inputFormat;
         CMFormatDescriptionRef formatDescription = self.deviceInput.device.activeFormat.formatDescription;
-        inputFourCC = CMFormatDescriptionGetMediaSubType(formatDescription);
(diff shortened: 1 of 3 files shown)


RytoEX (approved):
Looks okay at a glance, but I would appreciate some testing/review from others.

jcm93:
I don't have a card that has the behavior from the OP, so I only tested this to make sure nothing broke with format selection for the devices I do have; looks to be fine.

MrVelidi:
Tested the change with Elgato HD60X and i was able to see NV12 and YUV2 as selectable input formats in the drop down and the NV12 option is properly working for both 1440p and 2160p.

PR #10276, opened 2024-02-22
obs-filters: Do not load NVVFX on OpenGL

exeldro:
### Description
Do not load nvvfx when started with opengl

### Motivation and Context
I want to be able to test opengl on windows without crashes as nvidia greenscreen filter uses `NvCVImage_InitFromD3D11Texture`

### How Has This Been Tested?
On windows 11 by starting with directx and with opengl

### Types of changes

### Checklist:

### DIFF
--- a/plugins/obs-filters/obs-filters.c
+++ b/plugins/obs-filters/obs-filters.c
@@ -84,7 +84,10 @@ bool obs_module_load(void)
 	obs_register_source(&luma_key_filter);
 	obs_register_source(&luma_key_filter_v2);
 #ifdef LIBNVVFX_ENABLED
-	if (load_nvvfx())
+	obs_enter_graphics();
+	const bool direct3d = gs_get_device_type() == GS_DEVICE_DIRECT3D_11;
+	obs_leave_graphics();
+	if (direct3d && load_nvvfx())
 		obs_register_source(&nvidia_greenscreen_filter_info);
 #endif
 	return true;


WizardCM:
It's worth noting we haven't officially supported OpenGL on Windows for quite a while (for good reason), so it's not expected to work and therefore this PR may be rejected outright for that reason alone.

PatTheMav:
> It's worth noting we haven't officially supported OpenGL on Windows for quite a while (for good reason), so it's not expected to work and therefore this PR may be rejected outright for that reason alone.

Should we remove `libobs-opengl` from the Windows target then? Because IMO we either ship it and then we also need to make sure that it works out of the box, or we remove it.

exeldro:
I do not want opengl for windows removed as I use it to do a simple test to see if a shader I wrote for directx is going to work in opengl

Fenrirthviti:
For posterity, enabling opengl on windows requires a) knowing it's an option and b) very explicitly and manually editing a configuration file to a specific, known, undocumented option to enable it in the first place.

I don't think removing it entirely is necessary, as it *does* still have some potential uses, especially for things like the stated development.

PatTheMav:
> For posterity, enabling opengl on windows requires a) knowing it's an option and b) very explicitly and manually editing a configuration file to a specific, known, undocumented option to enable it in the first place.
> 
> I don't think removing it entirely is necessary, as it _does_ still have some potential uses, especially for things like the stated development.

Right, but if it _can_ be enabled, then it must work. If it can be enabled by editing the configuration file, then it's a shipped feature and must be supported. 

I see no reason to keep code around (that must be maintained and kept in a compilable state) if it's not supposed to be used. "Hiding it" is just an anti-pattern.

PatTheMav (approved):
Regardless of the prior discussion the fix looks correct and follows the look-before-you-leap approach.

My only concern is if this is sufficient and if there are other places where we use Direct3D-specific stuff regardless of selected renderer.

pkviet:
If the filter is an issue in opengl, i would just disable it. There's a cmake flag for that. Surely when testing shaders, that's something that can be done ? Seems to me like a developer issue that our users won't run into.

RytoEX:
> If the filter is an issue in opengl, i would just disable it. There's a cmake flag for that. Surely when testing shaders, that's something that can be done ? Seems to me like a developer issue that our users won't run into.

That requires rebuilding OBS to do so, and I was not under the impression that @exeldro or anyone else is rebuilding OBS specifically for testing shaders or what have you. They might be rebuilding OBS incidentally for other changes, but a CMake flag requires reconfiguring the build and then rebuilding, which seems like a lot to ask to remember for this.

I'm in favor of this fix until such a time as we have a better solution.

RytoEX (approved):

Lain-B (approved):

PR #10288, opened 2024-02-26
Implement Quality of Service option for RTMP (Windows)

derrod:
### Description

Enables QoS option for RTMP(S) on Windows.

### Motivation and Context

Increase network priority of stream traffic.

Idea proposed by @notr1ch.

### How Has This Been Tested?

Verified DSCP flags are set with Wireshark.

Ran parallel test streams to ensure nothing broke.

### Types of changes

### Checklist:


alinsavix:
Do modern routers actually honor DSCP without the user having to hand-craft QoS policies for it (at which point they could just craft those policies to handle things appropriately for RTMP traffic without having to go set anything special in OBS or whatever)? Wondering if this will just be a confusing option that users think will do more than it actually does, while providing no benefit at all for 99.9999% of users.

notr1ch:
It will usually require additional setup for proper queuing behavior. The user-facing name "Increase network priority" is probably a bit confusing in this case, perhaps we could indicate it is more technical, such as "Mark RTMP traffic with DSCP AF32" (assuming MSDN docs mean decimal instead of hex).

PR #10294, opened 2024-02-27
UI: Fix a bug with Output Settings not updating & Toggle audio multi-tracks correctly

pkviet:
### Description
This does two things:
1. This removes the LoadOutputSettings() called in SaveStream1Settings. It caused a bug when trying to save Output settings when both Stream and Output settings have been changed but the changes have not been applied by hitting the 'Apply' button in Settings.
Credits: Rodney made the correct guess for the faulty line.
2. This fixes the display of audio multi-tracks which must alternate between radio buttons (single track protocols) to checkboxes
(mpegts). The LoadOutputSettings call was used to do that update but it was too broad. Instead we use (a) the signal when a service is changed to switch back from mpegts multi-track to single track; (b) the signal when a URL is input in the Server text field of Custom service; (c) when Stream settings are saved.

### Motivation and Context
Fixes a bug found in 30.1 beta.
The function was introduced by me in https://github.com/obsproject/obs-studio/pull/9028 .
It was used to toggle the display of audio multi-tracks checkboxes instead of radio buttons.

The repro steps in beta are as follows:
- change the stream service but don't hit apply,
- change the output settings from Simple to Advanced or the converse,
- hit Apply
- the output Settings are reverted.

The reason why this occurs is because when one hits apply, the stream settings are saved before the output settings.
But since they load the output settings, the changes for the outputs are reverted.

But we still want to display correctly audio multi track buttons (either radio or checkboxes). This is the goal of the second commit.

### How Has This Been Tested?
The bug is fixed.
SRT multi-track audio still works (tested from obs to wowza).
Multi-track buttons alternate correctly between radio or checkboxes as a function of the Service or as a function of the protocol when Custom service is picked.

### Types of changes

### Checklist:

### DIFF
--- a/UI/window-basic-settings-stream.cpp
+++ b/UI/window-basic-settings-stream.cpp
@@ -180,6 +180,28 @@ void OBSBasicSettings::LoadStream1Settings()
 				  Qt::QueuedConnection);
 }
 
+#define SRT_PROTOCOL "srt"
+#define RIST_PROTOCOL "rist"
+
+bool OBSBasicSettings::AllowsMultiTrack(const char *protocol)
+{
+	return astrcmpi_n(protocol, SRT_PROTOCOL, strlen(SRT_PROTOCOL)) == 0 ||
+	       astrcmpi_n(protocol, RIST_PROTOCOL, strlen(RIST_PROTOCOL)) == 0;
+}
+
+void OBSBasicSettings::SwapMultiTrack(const char *protocol)
+{
+	if (protocol) {
+		if (AllowsMultiTrack(protocol)) {
+			ui->advStreamTrackWidget->setCurrentWidget(
+				ui->streamMultiTracks);
+		} else {
+			ui->advStreamTrackWidget->setCurrentWidget(
+				ui->streamSingleTracks);
+		}
+	}
+}
+
 void OBSBasicSettings::SaveStream1Settings()
 {
 	bool customServer = IsCustomService();
@@ -264,6 +286,7 @@ void OBSBasicSettings::SaveStream1Settings()
 	}
 
 	SaveCheckBox(ui->ignoreRecommended, "Stream1", "IgnoreRecommended");
+	SwapMultiTrack(QT_TO_UTF8(protocol));
 }
 
 void OBSBasicSettings::UpdateMoreInfoLink()
@@ -509,6 +532,13 @@ void OBSBasicSettings::on_service_currentIndexChanged(int idx)
 		if (idx == 0)
 			lastCustomServer = ui->customServer->text();
 	}
+
+	if (!IsCustomService()) {
+		ui->advStreamTrackWidget->setCurrentWidget(
+			ui->streamSingleTracks);
+	} else {
+		SwapMultiTrack(QT_TO_UTF8(protocol));
+	}
 }
 
 void OBSBasicSettings::on_customServer_textChanged(const QString &)
@@ -520,6 +550,8 @@ void OBSBasicSettings::on_customServer_textChanged(const QString &)
 
 	if (ServiceSupportsCodecCheck())
 		lastCustomServer = ui->customServer->text();
+
+	SwapMultiTrack(QT_TO_UTF8(protocol));
 }
 
 void OBSBasicSettings::ServiceChanged(bool resetFields)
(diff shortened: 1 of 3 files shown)


UI/window-basic-settings.cpp
```diff
@@ -2061,7 +2061,7 @@ static inline QString makeFormatToolTip()
 #define SRT_PROTOCOL "srt"
 #define RIST_PROTOCOL "rist"
 
-inline bool allowsMultiTrack(const char *protocol)
+inline bool OBSBasicSettings::allowsMultiTrack(const char *protocol)
```
tytan652, line 2064: The inline keyword is now in the declaration in the header file.

```suggestion
bool OBSBasicSettings::allowsMultiTrack(const char *protocol)
```

UI/window-basic-settings.hpp
```diff
@@ -383,6 +383,9 @@ private slots:
 	bool ServiceAndACodecCompatible();
 	bool ServiceSupportsCodecCheck();
 
+	inline bool allowsMultiTrack(const char *protocol);
```
tytan652, line 386: Drop inline to make compilers stop complaining
```suggestion
	bool allowsMultiTrack(const char *protocol);
```

tytan652 (commented):
Maybe also rename `allowsMultiTrack` to `AllowsMultiTrack` to match our function naming.

UI/window-basic-settings.cpp
pkviet: done

UI/window-basic-settings.hpp
tytan652: Can be put as resolved

pkviet: moved the function to window-basic-settings-stream.cpp ; macos and flatpak no longer complain.

pkviet:
> Maybe also rename `allowsMultiTrack` to `AllowsMultiTrack` to match our function naming.

thanks, done.

derrod (approved):

UI/window-basic-settings.cpp
```diff
@@ -2118,20 +2108,11 @@ void OBSBasicSettings::LoadAdvOutputStreamingSettings()
...
-		is_multitrack_output = allowsMultiTrack(protocol);
-	}
-
-	if (is_multitrack_output) {
-		ui->advStreamTrackWidget->setCurrentWidget(
-			ui->streamMultiTracks);
-	} else {
-		ui->advStreamTrackWidget->setCurrentWidget(
-			ui->streamSingleTracks);
+		SwapMultiTrack(QT_UTF8(protocol));
```
RytoEX, line 2115: We check `protocol` here and also check it inside the `SwapMultiTrack` function. Do we need to check it in both places?

RytoEX, line 2115: We call `QT_UTF8` here, but then call `QT_TO_UTF8` twice inside the `SwapMultiTrack` function. Why do we need to do this?

RytoEX (commented):
Questions.

UI/window-basic-settings.cpp
pkviet: nope we don't; fixed in the next force push

pkviet: The function is called twice in window-basic-settings-steam.cpp where protocol is a QString instead of a const char*. There were also multiple iterations while multi-tasking and moved some code around.
For the sake of simplicity and readability, i'm switching the arg to const char *

UI/window-basic-settings-stream.cpp
```diff
@@ -521,6 +550,7 @@ void OBSBasicSettings::on_customServer_textChanged(const QString &)
 
 	if (ServiceSupportsCodecCheck())
 		lastCustomServer = ui->customServer->text();
+	SwapMultiTrack(QT_TO_UTF8(protocol));
```
RytoEX, line 553: Nit: line break before this line.

RytoEX (changes requested):

UI/window-basic-settings-stream.cpp
pkviet: done

RytoEX (approved):

PR #10296, opened 2024-02-27
CI: Fix broken pip install to use ephemeral virtual environment

PatTheMav:
### Description
Change validator actions to use a virtual environment to install PIP modules required for validator scripts.

### Motivation and Context
Python 3.11 and later support a system-wide configuration setting that marks the system packages as "externally managed" (e.g. to force using apt packages instead of pip packages).

This breaks installation of the modules necessary to run the validators, so use a virtual environment instead.

### How Has This Been Tested?
Needs to be tested on CI.

### Types of changes

### Checklist:


RytoEX (approved):
Seem okay.

PR #10301, opened 2024-02-27
CI: Limit format checks to changed files only

PatTheMav:
### Description
Changes CI formatting actions (and associated formatting script) to accept a list of files.

### Motivation and Context
Before the CI checks for source code and CMake files would always check the entire source code even if just a single files has been changed.

With this update, the formatting script is enhanced to accept a list of files (which is generated as a condition to run the script in the first place) which ensures that a PR will only fail validation over files it changed itself.

### How Has This Been Tested?
Tested the script locally with no list of files provided and different number of specified file paths.

### Types of changes

### Checklist:

### DIFF
--- a/.github/actions/run-clang-format/action.yaml
+++ b/.github/actions/run-clang-format/action.yaml
@@ -60,6 +60,6 @@ runs:
           print ::endgroup::
 
           print ::group::Run clang-format-16
-          ./build-aux/run-clang-format --fail-${{ inputs.failCondition }} --check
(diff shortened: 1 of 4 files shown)


RytoEX (approved):
This seems fine. If we determine that some files are slipping through the cracks, we can re-adjust this later. There do seem to be some potential time savings, though minimal. The intent here is to prevent PRs from failing on files that they did not change, and it seems to do that. We'll make sure to do a format pass over the entire codebase periodically (typically when we update formatter versions or formatter rules).

PR #10312, opened 2024-03-01
libobs/util: Add support for #if and #elif to CF parser

exeldro:
### Description
Add support for #if and #elif to CF parser
It supports without comparator and with the following comparators: ==, >=, <=, !=, <, >

### Motivation and Context
This was not supported yet and I wanted to use it in a shader. And it helps when trying to convert existing shaders to be used in OBS.

### How Has This Been Tested?
On windows 11 by having a shader with this code in it:
```
#define test 1234

#if test == 0

#elif 0

#elif test > 1234

#elif test < 1234

#elif test == 1234

#endif
```

### Types of changes

### Checklist:

### DIFF
--- a/libobs/util/cf-lexer.c
+++ b/libobs/util/cf-lexer.c
@@ -759,6 +759,28 @@ static inline struct cf_def *cf_preprocess_get_def(struct cf_preprocessor *pp, c
 	return pp->defines.array + idx;
 }
 
+bool cf_preprocess_get_def_value(struct cf_preprocessor *pp, const struct strref *def_name, int *value)
+{
+	struct cf_def *def = cf_preprocess_get_def(pp, def_name);
+	if (def == NULL || def->macro)
+		return false;
+
+	for (size_t i = 0; i < def->tokens.num; i++) {
+		if (def->tokens.array[i].type == CFTOKEN_SPACETAB)
+			continue;
+		if (def->tokens.array[i].type == CFTOKEN_NUM &&
+		    sscanf(def->tokens.array[i].str.array, "%i", value) == 1)
+			return true;
+
+		if (def->tokens.array[i].type == CFTOKEN_NAME &&
+		    cf_preprocess_get_def_value(pp, &def->tokens.array[i].str, value))
+			return true;
+
+		return false;
+	}
+	return false;
+}
+
 static char space_filler[2] = " ";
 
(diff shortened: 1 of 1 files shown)


PatTheMav (commented):
While the code seems quite robust (and was able to handle quite a few malformed preprocessor instructions), it seems to fail over one specific variation of correct code:

```HLSL
float4 PSDrawBare(VertInOut vert_in) : TARGET
{
#define test 1234

#if test == 0
    float vector = 4.0;
#elif 0
    float vector = 3.0;
#elif test > 1234
    float vector = 2.0;
#elif test < 1234
    float vector = 1.0;
#elif test == 1234
    float vector = 5.0;
#endif

	return image.Sample(def_sampler, vert_in.uv);
}
```

The parser breaks after the `#elif 0` preprocessor line, as it attempts to parse a "sub-block" (via `cf_preprocess_subblock`) but because the code already "jumped" over the newline after `0`, it finds the non-preprocessor token `float` and skips the entire line.

And indeed if the line were to say `#elif 1` instead, the parser skips the line, finds `#elif test > 1234`, somehow jumps _back_ to the `#elif 1`, throws another error, skips the current line, finds `#elif test < 1234` and then throws the `outside of "#if/#ifdef/#ifndef block"` error, so the internal state of the parser seems to get messed up.

---

Overall I think this would be fine to add, as it's clearly "missing functionality" from the parser/preprocessor (it is an open `TODO` after all).

I shouldn't forget to mention that the use of macros to create some variation of "uber-shaders" is discouraged with modern APIs which instead expect you to compose fragment and vertex shader out of smaller "function objects" that you compile and load only once, rather than creating several "fat" variants of the same shader influenced by preprocessor conditionals.

We get shader compilation caching "for free" with Metal on macOS and have implemented it for Direct3D11, but at some point we need to move to precompiled shaders (including proper shader "libraries") with the expectation that plugins also have to do the same, but that point hasn't been reached yet.

exeldro:
`cur_token` was indeed wrong for things like `#if 0`, `#if 1`, `#elif 0` and `#elif 1`
Rebased and fixed the issue

PatTheMav (approved):
Only bug I could find was fixed by most recent push, seems to work fine otherwise.

RytoEX (approved):

PR #10318, opened 2024-03-02
UI: Fix `Start Virtual Camera` Text Cut Off in Corresponding Button

### LINKED ISSUE #10210: Start Virtual Camera text is cut off in the UI
### Operating System Info

Windows 11

### Other OS

_No response_

### OBS Studio Version

30.0.2

_No response_

### OBS Studio Log URL

N/A

### OBS Studio Crash Log URL

_No response_

### Expected Behavior

The `Start Virtual Camera` text should be cut off at the right and three dots `...` should be added like in the title of the Scene Transitions dock (see screenshot).

### Current Behavior

Letters of the `Start Virtual Camera` text are cut off on both sides.

### Steps to Reproduce

1. Change the width of the Controls dock to its minimum.

_No response_

powersagitar:
> [!TIP]
> This summary has been updated to match the most recent code change

### Description

Change size policy of "Start Virtual Camera" button to ensure the minimum width of "Controls" panel is greater than or equal to the minimum width of "Start Virtual Camera" button plus the gear (settings) icon. Text cutoff is thus eliminated.

Old minimum width of "Controls" panel:
ui before fix, text gets truncated

New minimum width of "Controls" panel:
Screenshot 2024-03-03 at 5 45 24 PM

### Motivation and Context

This PR fixes #10210.

### How Has This Been Tested?

Tests are conducted by manually resizing "Controls" panel.
Solution works for both LTR and RTL languages.

Environment:  
- macOS Ventura 13.6.4 (22G513)
  Arch: x86_64
- Xcode 15.2
  Build version 15C500b

### Types of changes

### Checklist:

### DIFF
--- a/UI/window-basic-main.cpp
+++ b/UI/window-basic-main.cpp
@@ -1938,6 +1938,8 @@ void OBSBasic::AddVCamButton()
 			    QStringLiteral("configIconSmall"),
 			    &OBSBasic::VCamConfigButtonClicked);
 	vcamButton->insert(2);
+	vcamButton->first()->setSizePolicy(QSizePolicy::Minimum,
+					   QSizePolicy::Minimum);
 }
 
 void OBSBasic::ResetOutputs()


AUTHORS
```diff
@@ -651,6 +651,7 @@ Contributors:
  zhaolong
  张昆
  练亮斌
+ powersagitar
```
gxalpha, line 654: Don't add yourself to the authors list. It’s generated automatically and any manual changes will get lost.

gxalpha (changes requested):

Fenrirthviti:
I'm confused why this is creating an entire new class of button for a single button, instead of applying this attribute to the existing button. Can you please provide more context on why it was solved this way?

Was this also tested with other non-English languages (specifically, RTL languages) as you are explicitly setting `Qt::TextElideMode::ElideRight` which will break with any non-RTL language handlings.

This solution, to me, seems like the incorrect approach.

powersagitar:
> why this is creating an entire new class of button for a single button, instead of applying this attribute to the existing button.

I created a new button since 
1. With the assumption that the "attribute" you mentioned being something like `text-overflow: ellipsis`, I didn't find any, so the solution I came up with was to listen to the resize event manually and change the text displayed. 
2. The means I found to listen to resize events is to either override `QPushButton::resizeEvent` or `QPushButton::paintEvent`, but in either way a new derived class is required.
3. Right now only `Start Virtual Camera` button has overflowed text, but there could be more buttons requiring the same feature in the future, so I thought it would be better to extract this feature into its own class.

> will break with any non-RTL language handlings

As I personally cannot speak such languages so forgot accounting for them. The solution could be calling `QPushButton::locale().textDirection()` and set `Qt::TextElideMode` accordingly.

Please let me know if you think or find there's better implementations.

Warchamp7:
I'm fairly sure fixing the SizePolicy on these buttons should handle overflowing text properly. This fix is incredibly overkill and not the right approach at all.

powersagitar:
`SizePolicy` does fix this bug by not allowing `Controls` panel to shrink beyond the minimum width of the widget. It also works for both LTR and RTL languages (tested). I think this one is good to be merged if we don't want specifically ellipsis but only a fix for text clipping.

E.g., this is the new minimum width of `Controls` panel.
Screenshot 2024-03-03 at 5 45 24 PM

WizardCM:
Please squash your commits.

powersagitar:
Done.

UI/window-basic-main.cpp
```diff
@@ -1938,6 +1938,10 @@ void OBSBasic::AddVCamButton()
 			    QStringLiteral("configIconSmall"),
 			    &OBSBasic::VCamConfigButtonClicked);
 	vcamButton->insert(2);
+
+	QSizePolicy buttonSizePolicy = vcamButton->first()->sizePolicy();
+	buttonSizePolicy.setHorizontalPolicy(QSizePolicy::Minimum);
```
Warchamp7, line 1943: It should be possible to trim this PR down to simply one line

```vcamButton->first().setSizePolicy(QSizePolicy::Minimum, QSizePolicy::Minimum);```

or similar

Warchamp7 (changes requested):

Warchamp7 (approved):

PR #10320, opened 2024-03-03
UI: Fix closing OBS with floating docks

cg2121:
### Description
When OBS is closed and there is a floating dock, a Windows system sound is emitted. This fixes that by not executing the close dock dialog when OBS is closing.

### Motivation and Context
Fixes bug

### How Has This Been Tested?
Made sure the dock closing dialog is never triggered when OBS is closed

### Types of changes

### Checklist:

### DIFF
--- a/UI/window-basic-main.hpp
+++ b/UI/window-basic-main.hpp
@@ -1042,6 +1042,7 @@ public:
 	void SetDisplayAffinity(QWindow *window);
 
 	QColor GetSelectionColor() const;
+	inline bool Closing() { return closing; }
 
 protected:
 	virtual void closeEvent(QCloseEvent *event) override;
--- a/UI/window-dock.cpp
+++ b/UI/window-dock.cpp
@@ -1,5 +1,6 @@
 #include "window-dock.hpp"
 #include "obs-app.hpp"
+#include "window-basic-main.hpp"
 
 #include <QMessageBox>
 #include <QCheckBox>
@@ -27,7 +28,7 @@ void OBSDock::closeEvent(QCloseEvent *event)
 
 	bool warned = config_get_bool(App()->GlobalConfig(), "General",
 				      "WarnedAboutClosingDocks");
-	if (!warned) {
+	if (!OBSBasic::Get()->Closing() && !warned) {
 		QMetaObject::invokeMethod(App(), "Exec", Qt::QueuedConnection,
 					  Q_ARG(VoidFunc, msgBox));
 	}


PatTheMav (commented):
What we could do instead is just check if the `spontaneous` flag is set on the close event and only check for/show the warning then:

```c++
    if (((QEvent *)event)->spontaneous()) {
        bool warned = config_get_bool(App()->GlobalConfig(), "General",
                          "WarnedAboutClosingDocks");
        if (!warned) {
            QMetaObject::invokeMethod(App(), "Exec", Qt::QueuedConnection,
                          Q_ARG(VoidFunc, msgBox));
        }
    }
```

In my tests the event flag is not set when the app is closed by system shortcut, pressing the "Exit" button, or using task manager to shut down the application.

This would avoid leaking the private `closing` state variable into other parts of the code.

RytoEX:
> In my tests the event flag is not set when the app is closed by system shortcut, pressing the "Exit" button, or using task manager to shut down the application.

But it would be set in some situations:
> Returns true if the event originated outside the application (a system event); otherwise returns false.

What I'd be curious about is if there is any situation in this implementation where `closing` is not yet `true` by the time this condition is checked.

I don't think it makes sense to spawn this message box _at all_ if the application is closing for any reason. The warning about the qdockwidget itself is moot because the application will close the warning as the application closes. We'd probably have to check both proposed implementations on Windows to see if it's apparent that the message box has been opened (it will emit an audible alert on Windows) under any application close scenarios.

PatTheMav:
> But it would be set in some situations:
> 
> > Returns true if the event originated outside the application (a system event); otherwise returns false.

Which is 100% the opposite of what actually happens - it is `false` for system events and only `true` for "internal" events.

RytoEX:
> > But it would be set in some situations:
> > > Returns true if the event originated outside the application (a system event); otherwise returns false.
> 
> Which is 100% the opposite of what actually happens - it is `false` for system events and only `true` for "internal" events.

Then someone needs to file a QTBUG and tell them that either their docs are wrong or they have a bug. I would expect "spontaneous" to act as described in their docs. I'll have to check it on Windows. That's a bit off-topic for this PR though.

RytoEX:
> What we could do instead is just check if the `spontaneous` flag is set on the close event

For what it's worth, the `spontaneous()` function returned `false` for me on Windows 11 if I closed the app with:
* the Exit button
* the X button
* File -> Exit
* Task Manager -> Processes -> End task

This is consistent with the documentation. If it behaves the opposite on macOS, then that's probably a bug, possibly [QTBUG-118533](https://bugreports.qt.io/browse/QTBUG-118533).

If I closed the app with Task Manager -> Details -> End task, it doesn't even reach this point.

I never found a case where it returned `true` at this point in the code. Still, in any case where it were `true`, if the app is closing, I would still not want the warning here emitted.

We did have a UX discussion about this and considered removing this specific warning entirely. The reasons it was added is from years ago, and we think that perhaps it doesn't make sense to have this warning these days. If we remove this entire warning, we can also just remove the getter and include added here.

RytoEX (approved):
This seems to be fine, though the concerns above are noted.

PR #10324, opened 2024-03-03
libobs/graphics: Add support for param assign with type constructor

exeldro:
### Description
Add support for:
`uniform float4 test = float4(1,0,0,1);`
next to the existing:
`uniform float4 test = {1,0,0,1};`

### Motivation and Context
Support more shader stuff. This helps converting glsl shaders more easily.

### How Has This Been Tested?
On windows 11 by having a shader with the new param assignment

### Types of changes
- 
### Checklist:

### DIFF
--- a/libobs/graphics/effect-parser.c
+++ b/libobs/graphics/effect-parser.c
@@ -1082,13 +1082,27 @@ static inline int ep_parse_param_assign_intfloat_array(struct effect_parser *ep,
 	}
 
 	/* -------------------------------------------- */
-
-	code = cf_next_token_should_be(&ep->cfp, "{", ";", NULL);
-	if (code != PARSE_SUCCESS)
-		return code;
+	if (!cf_next_token(&ep->cfp)) {
+		cf_adderror_unexpected_eof(&ep->cfp);
+		return PARSE_EOF;
+	}
+	char *end_char = NULL;
+	if (strref_cmp(&ep->cfp.cur_token->str, "{") == 0) {
+		end_char = "}";
+	} else if (strref_cmp(&ep->cfp.cur_token->str, param->type) == 0) {
+		end_char = ")";
+		code = cf_next_token_should_be(&ep->cfp, "(", ";", NULL);
+		if (code != PARSE_SUCCESS)
+			return code;
+	} else if (!cf_go_to_token(&ep->cfp, ";", NULL)) {
+		return PARSE_EOF;
+	} else {
+		cf_adderror_expecting(&ep->cfp, "{");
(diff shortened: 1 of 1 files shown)


PatTheMav:
Can you post an example shader to test this with? I need to ensure this doesn't break Metal shader transpilation.

exeldro:
for example in https://github.com/obsproject/obs-studio/blob/2b613f40b472503637b251886c652eafc01238e7/libobs/data/solid.effect#L2
replace `{1.0, 1.0, 1.0, 1.0}` with `float4(1.0, 1.0, 1.0, 1.0)`

PatTheMav:
EDIT: My comment here is "wrong" insofar as the PR adds support precisely for what I'm advocating here, but I misread the code change and thought it did the opposite.

I'll keep that comment here for posterity, but it doesn't apply to this change.

---

My main issue with adding this feature is that it relies on implicit type conversions, which are not under the control of the developer and thus can lead to unexpected behaviour, particularly between the 3 shader languages we have to transpile into.

So when given the choice between expressiveness (the code spells out what is intended to happen) and correctness (using a "wrong" conversion results in an explicit error), and saving a bit of time writing shaders, we should always prefer the former.

Functionally most shader languages will just do an implicit conversion here, with all the caveats and issues that implicit conversions have to begin with:

* Which conversion actually takes place under the hood is opaque to the developer
* Because of that the code can exhibit strange, unexpected behaviour
* Implicit conversions usually compile fine if they are non-narrowing (e.g. `half` to `float`)
* Implicit conversions can compile fine even if they are narrowing if the constant itself can be fully represented in the target type
* However they will fail once the compiler cannot ensure this (either the constant is too large for the target type or a runtime value, e.g. an `int`)

And because we don't precompile shaders (even though it would be better if we did), those issues might only occur at runtime when people run the app, so our transpiler should be _more_ strict about these things (to reduce the potential runtime shader compilation errors) not less.

exeldro:
Where do you see the implicit type conversions? The change only works when the type is exactly the same, checked by `strref_cmp(&ep->cfp.cur_token->str, param->type) == 0`, which would result in no type conversions.

PatTheMav:
> > Where do you see the implicit type conversions? The change only works when the type is exactly the same, checked by `strref_cmp(&ep->cfp.cur_token->str, param->type) == 0`, which would result in no type conversions.
> 

Sorry, disregard my prior comments, I wholly misunderstood the changes of this PR.

PatTheMav (approved):

RytoEX (approved):

PR #10335, opened 2024-03-05
linux-v4l2: Give camera up to 2 seconds to start

david672orford:
### Description

Give the V4L2 devices up to 2 seconds to start sending frames after the device is opened rather than the much shorter **Frames Until Timeout** enabled by the **Autoreset on Timeout** checkbox.

### Motivation and Context
The **Autoreset on Timeout** feature is intended to close and reopen the camera if it stalls. A stall is detected when a set number of frames (by default 5) are missed. The problem is that this stall detection is active even before the first frame has been received. If it takes more than 5 frame intervals for a camera to start sending frames, it may never start because OBS will loop opening it and closing it without end.

This is an insidious problem. The user can check this box and all seems to be well (since the camera is already running), but it will fail to start if he restarts OBS and he won't know why. The discription of **Autoreset on Timeout** in no way suggests the behavior which this patch corrects. I was able to figure out what was going on only by reading the source code. 

### How Has This Been Tested?
This was tested under Ubuntu 22.04.04 with the current HEAD OBS (approaching the 3.1.0 release). I tested it with **Autoreset on Timeout** enabled and the **Frames Until Timeout** set to the default of 5. I used a Creality 3D Technology CREALITY CAM as the input devices. I tried an initial timeout of 1 second, but this still gave one spurious restart when starting OBS. Upping it to 2 seconds allowed the camera to start on the first try.

### Types of changes
Bug fix

### Checklist:
- [x] I have included updates to all appropriate documentation. (N/A)

### DIFF
--- a/plugins/linux-v4l2/v4l2-input.c
+++ b/plugins/linux-v4l2/v4l2-input.c
@@ -178,7 +178,11 @@ static void *v4l2_thread(void *vptr)
 	size_t plane_offsets[MAX_AV_PLANES];
 	int fps_num, fps_denom;
 	float ffps;
-	uint64_t timeout_usec;
+	uint64_t data_timeout_usec;
+	// v4l2_start_capture is async so we need some extra time to recieve
+	// the first frame from slow to initialize devices.
+	const uint64_t init_timeout_usec = 2000000;
+	bool initialized = false;
 
 	blog(LOG_DEBUG, "%s: new capture thread", data->device_id);
 	os_set_thread_name("v4l2: capture");
@@ -187,11 +191,11 @@ static void *v4l2_thread(void *vptr)
 	v4l2_unpack_tuple(&fps_num, &fps_denom, data->framerate);
 	ffps = (float)fps_denom / fps_num;
 	blog(LOG_DEBUG, "%s: framerate: %.2f fps", data->device_id, ffps);
-	/* Timeout set to 5 frame periods. */
-	timeout_usec = (1000000 * data->timeout_frames) / ffps;
+	/* Timeout for new frames set to 5 frame periods. */
+	data_timeout_usec = (1000000 * data->timeout_frames) / ffps;
 	blog(LOG_INFO,
 	     "%s: select timeout set to %" PRIu64 " (%dx frame periods)",
-	     data->device_id, timeout_usec, data->timeout_frames);
+	     data->device_id, data_timeout_usec, data->timeout_frames);
 
 	if (v4l2_start_capture(data->dev, &data->buffers) < 0)
 		goto exit;
@@ -210,7 +214,8 @@ static void *v4l2_thread(void *vptr)
 
 		/* Set timeout timevalue. */
 		tv.tv_sec = 0;
-		tv.tv_usec = timeout_usec;
+		tv.tv_usec = initialized ? data_timeout_usec
+					 : init_timeout_usec;
 
 		r = select(data->dev + 1, &fds, NULL, NULL, &tv);
 		if (r < 0) {
@@ -232,6 +237,7 @@ static void *v4l2_thread(void *vptr)
 			}
 
 			if (data->auto_reset) {
+				initialized = false;
 				if (v4l2_reset_capture(data->dev,
 						       &data->buffers) == 0)
 					blog(LOG_INFO,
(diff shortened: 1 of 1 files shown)


kkartaltepe:
Please include a log representative of the error and the full output for `v4l2-ctl --device=/dev/videoXXX --all` for the device in question so we can reference it in the future.

david672orford:
Output of v4l2-ctl --device=/dev/video0 --all and the OBS log are attached.
* [v4l2-ctl-output.txt](https://github.com/obsproject/obs-studio/files/14503191/v4l2-ctl-output.txt)
* [2024-03-05 19-57-56.txt](https://github.com/obsproject/obs-studio/files/14503187/2024-03-05.19-57-56.txt)

WizardCM:
If we end up doing this, please use a variable rather than a magic number. Something like `max_timeout`.

david672orford:
@WizardCM Good idea. I'll declare a const variable called first_frame_timeout.

david672orford:
@WizardCM first_frame_timeout_usec change added to PR

norihiro:
The commit message should be wrapped to 72 columns.

david672orford:
@norihiro I had wrapped to 79. Rewrapped to 72.

kkartaltepe:
> I'm sorry, there is one more nitpick on the commit message. The first word (`declare`) should be capitalized.

I think it would be better to just squash them and only keep the first commit message.

kkartaltepe:
While reviewing this, the log shows you have enabled the auto-reset feature in the video source, which also isnt the default. does the device still fail to start if you disable that feature?

david672orford:
I have merged the two commits keeping only the first message, as requested.

I have also attached a log 
[2024-03-30 23-30-02.txt](https://github.com/obsproject/obs-studio/files/14814028/2024-03-30.23-30-02.txt) showing the same camera starting with auto-reset turned off. You will note that OBS prints four error messages about the camera, but after that it starts.

In case it is not clear from my original description, the purpose of this patch is specifically to allow the use of the auto-reset feature with more cameras without compromising its intended function.

Mine is not the only camera which encounters this problem. I was telling a workmate about this PR and he immediately reproduced the problem on his laptop. He was able to work around it by raising **Frames Until Timeout** from 5 to 7 frames. (This workaround also works on my camera, but only if I set it to at least 15 frames, so I don't think changing the default is the answer.)

kkartaltepe:
I made some minor style changes, and you introduced a merge commit which are not used in OBS remember to rebase instead of pull'ing if you want to update your code to apply to master.

Unfortunately for a C920 it can take multiple frames before the device to stabilize, so the only safe option for users appears to increase the timeout to a suitable level for your device.

david672orford:
@kkartaltepe OK. I am still learning how to rebase, so I must have done something wrong when I synced the PR with head. Your force pushes are full of changes I did not make, I think this is you fixing the problem, but I am not sure. If not let me know and I will dig into it.

As I have written the patch, the timeout is 2 seconds until the first frame is received. After that it is the configured timeout (5 frame intervals by default). Once the timeout fires the device is closed, reopened, and the timeout is set back to 2 seconds until the first frame of the new session is received.

When you say that the C920 needs time to stabilize, do you mean that it stutters while starting up? If so, I think we could solve that by keeping the timeout at 2 seconds until some number of frames were received at which time we would switch to the configured timeout.

kkartaltepe:
>  If so, I think we could solve that by keeping the timeout at 2 seconds until some number of frames were received at which time we would switch to the configured timeout.

I don't think we should add more code to solve this, but that users should configure the timeout to be reasonable for their devices if they want to use the auto reset feature.

david672orford:
When you talk about not adding more code, do you mean that you don't think this patch should be merged, or do you mean that it is a good enough fix?

kkartaltepe:
> When you talk about not adding more code, do you mean that you don't think this patch should be merged, or do you mean that it is a good enough fix?

Your recommendation to add more code to handle an increasingly complicated waiting scheme, is what I mean by more code.

In the end I think we should keep the waiting code simple. Users changing the device to auto-reset can also change the timeout to something appropriate for dealing with long reset times if their device has that issue.

david672orford:
It seems to me that the change currently in this PR is worth the small increase in code size and complexity. The benefit is that it greatly reduces or entirely eliminates the need to play with the timeout. It also eliminates any need to warn to the users that turning on the timeout can make things worse by preventing their camera from starting.

I am sorry that I confused my position by offering to make furthur improvements if it turns out some cameras still required tweaking. You are right, that is a rabit hole we had best not go down if we don't have to.

I not trying to impose my opinion, just want to be clear about what I think this PR accomplishes.

PR #10343, opened 2024-03-07
frontend-tools: Fix output timer stopping recording on unpause

RytoEX:
### Description

If the "Pause timer when recording is paused" option in the Output Timer settings was enabled, even if an Output Timer was not being used, a recording may stop when attempting to unpause it. This was due to the check in the UnpauseRecordingTimer function being too loose and only checking for if the recording timer was not active. Let's initialize the recordingTimeLeft variable to -1 and check that it's greater than 0 before attempting to restart a recording timer.

### Motivation and Context

Fix a bug that appeared at least as early as OBS Studio 30.1.0-beta1. Do not want recordings stopping unexpectedly or prematurely.

When uninitialized, recordingTimeLeft could be something like -842150451. A negative value passed to QTimer's `start` function results in this:
```
13:12:39.929: QObject::startTimer: Timers cannot have negative intervals
13:12:39.929: recordingTimeLeft: -842150451
```

This negative interval warning happened at least as early as October 17, 2023 according to old notes of mine.

### How Has This Been Tested?

Tested locally on Windows 11.

### Types of changes

### Checklist:

### DIFF
--- a/UI/frontend-plugins/frontend-tools/output-timer.cpp
+++ b/UI/frontend-plugins/frontend-tools/output-timer.cpp
@@ -218,7 +218,7 @@ void OutputTimer::UnpauseRecordingTimer()
 	if (!ui->pauseRecordTimer->isChecked())
 		return;
 
-	if (!recordingTimer->isActive())
+	if (recordingTimeLeft > 0 && !recordingTimer->isActive())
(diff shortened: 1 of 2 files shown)


RytoEX:
@derrod Approved off-thread. Merging.

PR #10346, opened 2024-03-07
UI: Block Elgato Stream Deck plugin before 5.5.2.0 to avoid crashes

RytoEX:
### Description

The Elgato Stream Deck plugin older than 5.5.2.0 invokes undefined behavior that results in application crashes. Let's block older versions to prevent the crashes.

### Motivation and Context

Requested by @notr1ch . Want less crashes.

### How Has This Been Tested?

I've built this locally on Windows 11, but I have not tested this, nor am I sure how I would test it.

### Types of changes

### Checklist:

### DIFF
--- a/UI/win-dll-blocklist.c
+++ b/UI/win-dll-blocklist.c
@@ -183,6 +183,12 @@ static blocked_module_t blocked_modules[] = {
 	// Reference: https://github.com/obsproject/obs-studio/issues/8552
 	{L"\\holisticmotioncapturefilter64bit.dll", 0, 1680044549,
 	 TS_LESS_THAN},
+
+	// Elgato Stream Deck plugin < 2024-02-01
+	// Blocking all previous versions because they have undefined behavior
+	// that results in crashes.
+	// Reference: https://github.com/obsproject/obs-studio/issues/10245
+	{L"\\streamdeckplugin.dll", 0, 1706745600, TS_LESS_THAN},
 };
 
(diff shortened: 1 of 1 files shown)


notr1ch:
The only suggestion I have is to download the fixed Stream Deck version and verify the timestamp is outside of this range, otherwise LGTM.

RytoEX:
> The only suggestion I have is to download the fixed Stream Deck version and verify the timestamp is outside of this range, otherwise LGTM.

@Fenrirthviti confirmed that this at least blocks version 5.5.0.0 (dated March 3, 2023) and does not block the latest version (5.5.2.16). I'm not sure that we can specifically get the first version where this was fixed.

However, I've noticed that I blocked `streamdeckpluginqt6.dll` instead of the main `streamdeckplugin.dll`, so I might have to fix that.

derrod (approved):

RytoEX:
For what it's worth, I can confirm that Stream Deck 6.5.0 ships plugin version 5.5.0.0 dated March 3, 2023, and that plugin crashes OBS Studio 30.1.0-rc1. Stream Deck 6.5.1 and 6.5.2 ship StreamDeckPlugin.dll version 5.5.2.16 with StreamDeckPluginQt5.dll version 5.5.2.16 and StreamDeckPluginQt6.dll version 5.5.2.0 all dated February 14, 2024.

RytoEX:
Changed to block `streamdeckplugin.dll` instead of the Qt-version-specific variant. Re-tested against both known broken and working plugin versions and this seems to work as expected.

PR #10349, opened 2024-03-08
libobs: Add keyframe aligned encoder groups

palana:
### Description
This is meant to allow grouping encoders for e.g. eRTMP multitrack video (<=> Twitch Enhanced Broadcasting) such that IDR frames are aligned for seamless rendition switching when transmuxing the overall output stream into e.g. HLS or similar multi rendition formats

### Motivation and Context
Without this change, encoders aren't guaranteed to start on the same frame even if started "at the same time" due to encoder startup not being synchronized with the video thread; e.g. in practice we've observed a one frame offset that sometimes happens between encoders, though the offset can be arbitrarily high potentially due to lack of synchronization

With this there's explicit synchronization, so all encoders start on the same frame (and get the "same stream" of frames fed to them)

### How Has This Been Tested?
This code has been deployed in the Twitch Enhanced Broadcasting beta since January, and its effects can be observed when disabling the check in `receive_video` while using multitrack video outputs (and logging which input frame is first sent to encoders)

### Types of changes

### Checklist:

### DIFF
--- a/libobs/obs-output.c
+++ b/libobs/obs-output.c
@@ -290,6 +290,8 @@ void obs_output_destroy(obs_output_t *output)
 			}
 		}
 
+		da_free(output->keyframe_group_tracking);
+
 		clear_raw_audio_buffers(output);
 
 		os_event_destroy(output->stopping_event);
@@ -475,6 +477,8 @@ void obs_output_actual_stop(obs_output_t *output, bool force, uint64_t ts)
 		bfree(output->caption_head);
 		output->caption_head = output->caption_tail;
 	}
+
+	da_clear(output->keyframe_group_tracking);
 }
 
 void obs_output_stop(obs_output_t *output)
@@ -1989,6 +1993,85 @@ static void discard_unused_audio_packets(struct obs_output *output,
 		discard_to_idx(output, idx);
 }
 
+static bool purge_encoder_group_keyframe_data(obs_output_t *output, size_t idx)
+{
+	struct keyframe_group_data *data =
+		&output->keyframe_group_tracking.array[idx];
+	uint32_t modified_count = 0;
+	for (size_t i = 0; i < MAX_OUTPUT_VIDEO_ENCODERS; i++) {
+		if (data->seen_on_track[i] != KEYFRAME_TRACK_STATUS_NOT_SEEN)
+			modified_count += 1;
+	}
+
+	if (modified_count == data->required_tracks) {
+		da_erase(output->keyframe_group_tracking, idx);
+		return true;
+	}
+	return false;
+}
+
+/* Check whether keyframes are emitted from all grouped encoders, and log
+ * if keyframes haven't been emitted from all grouped encoders. */
+static void
+check_encoder_group_keyframe_alignment(obs_output_t *output,
+				       struct encoder_packet *packet)
+{
+	size_t idx = 0;
+	struct keyframe_group_data insert_data = {0};
+
+	if (!packet->keyframe || packet->type != OBS_ENCODER_VIDEO ||
+	    !packet->encoder->encoder_group)
+		return;
+
+	for (; idx < output->keyframe_group_tracking.num;) {
+		struct keyframe_group_data *data =
+			&output->keyframe_group_tracking.array[idx];
+		if (data->pts > packet->pts)
+			break;
+		if (data->group_id !=
+		    (uintptr_t)packet->encoder->encoder_group) {
+			idx += 1;
+			continue;
+		}
+
+		if (data->pts < packet->pts) {
+			if (data->seen_on_track[packet->track_idx] ==
+			    KEYFRAME_TRACK_STATUS_NOT_SEEN) {
+				blog(LOG_WARNING,
+				     "obs-output '%s': Missing keyframe with pts %" PRIi64
+				     " for encoder '%s' (track: %zu)",
+				     obs_output_get_name(output), data->pts,
+				     obs_encoder_get_name(packet->encoder),
+				     packet->track_idx);
+			}
+
+			data->seen_on_track[packet->track_idx] =
+				KEYFRAME_TRACK_STATUS_SKIPPED;
+
+			if (!purge_encoder_group_keyframe_data(output, idx))
+				idx += 1;
+			continue;
+		}
+
+		data->seen_on_track[packet->track_idx] =
+			KEYFRAME_TRACK_STATUS_SEEN;
+		purge_encoder_group_keyframe_data(output, idx);
+		return;
+	}
+
+	insert_data.group_id = (uintptr_t)packet->encoder->encoder_group;
+	insert_data.pts = packet->pts;
+	insert_data.seen_on_track[packet->track_idx] =
+		KEYFRAME_TRACK_STATUS_SEEN;
+
+	pthread_mutex_lock(&packet->encoder->encoder_group->mutex);
+	insert_data.required_tracks =
+		packet->encoder->encoder_group->encoders_started;
+	pthread_mutex_unlock(&packet->encoder->encoder_group->mutex);
+
+	da_insert(output->keyframe_group_tracking, idx, &insert_data);
+}
+
 static void interleave_packets(void *data, struct encoder_packet *packet)
 {
 	struct obs_output *output = data;
@@ -2020,6 +2103,8 @@ static void interleave_packets(void *data, struct encoder_packet *packet)
 			received_video = received_video &&
 					 output->received_video[i];
 	}
+
+	check_encoder_group_keyframe_alignment(output, packet);
 
 	was_started = output->received_audio && received_video;
 
--- a/libobs/obs-encoder.c
+++ b/libobs/obs-encoder.c
@@ -324,6 +324,17 @@ static void add_connection(struct obs_encoder *encoder)
 		}
 	}
 
+	if (encoder->encoder_group) {
+		bool ready = false;
+		pthread_mutex_lock(&encoder->encoder_group->mutex);
+		encoder->encoder_group->encoders_started += 1;
+		ready = encoder->encoder_group->encoders_started ==
+			encoder->encoder_group->encoders_added;
+		pthread_mutex_unlock(&encoder->encoder_group->mutex);
+		if (ready)
+			add_ready_encoder_group(encoder);
+	}
+
 	set_encoder_active(encoder, true);
 }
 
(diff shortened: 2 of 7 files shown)


WizardCM:
Please update the relevant documentation.

palana:
> Please update the relevant documentation.

Updated `reference-outputs.rst` in https://github.com/obsproject/obs-studio/pull/10349#event-12074277391

RytoEX (changes requested):
You can omit the `docs` prefix from these commits to shorten the commit message subjects. It will be apparent from the changed files that docs were changed. If you want to draw attention to it, mention in the commit message body that "this commit also updates/modifies docs as needed" or some such.

palana:
> You can omit the `docs` prefix from these commits to shorten the commit message subjects. It will be apparent from the changed files that docs were changed. If you want to draw attention to it, mention in the commit message body that "this commit also updates/modifies docs as needed" or some such.

Removed the `docs` prefix in https://github.com/obsproject/obs-studio/pull/10349#commits-pushed-273c2fd

tt2468:
I have one primary concern with this PR, which is that the currently supported behavior of sharing an encoder between multiple outputs, and starting those outputs at different times, becomes restricted with this addition.

Say I have encoder A, encoder B, output 1, and output 2. I apply encoder A and B to output 1, then I apply encoder A to output 2.

I can start output 1 first, then start output 2. I can also start output 2 first, and then start output 1. Say though, with this change, I "group" encoders A/B together. In this case, I **have** to start output 1 before output 2 will finish initializing encoders.

Having this be a limitation of aligning encoder keyframes, to me, is an important issue, as any future multi-destination streaming features which intend to support enhanced broadcast will need to impose this limitation on the user.

Rodney brought up that perhaps we could take control of the GOP and manually trigger keyframes. I think that this might be an adequate, and more reliable, method of synchronizing keyframes across an "encoder group." I suppose in theory, if an encoder is "grouped", it would disable any automatic keyframe generation in the implementation, and rely on signals from libobs to produce new keyframes. This could possibly be implemented as a flag in a "flags" parameter in the encode callbacks.

docs/sphinx/reference-outputs.rst
```diff
@@ -743,6 +743,42 @@ General Output Functions
...
+   on the same frame; additional properties that affect keyframe alignment
+   such as IDR/keyframe interval (e.g. *keyint*/*keyint_sec*) and
+   IDR insertion (e.g. libx264 *scenecut=0*) are not enforced.
+
+   Encoders can only be grouped/ungrouped while inactive; also, grouped
+   encoders only start processing frames when all encoders belonging to
+   a group are started, so e.g. adding a single grouped encoder out of
+   a group of encoders to a separate output and starting this output ahead
+   of the other output will cause the output to to not receive frames until
+   the other output/all other encoders are started.
```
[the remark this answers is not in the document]
palana: @RytoEX: should I remove the "the other output" bit? the output doesn't necessarily matter, all grouped encoders need to be started for any of them to actually receive frames

derrod:
> Rodney brought up that perhaps we could take control of the GOP and manually trigger keyframes. I think that this might be an adequate, and more reliable, method of synchronizing keyframes across an "encoder group." I suppose in theory, if an encoder is "grouped", it would disable any automatic keyframe generation in the implementation, and rely on signals from libobs to produce new keyframes. This could possibly be implemented as a flag in a "flags" parameter in the encode callbacks.

To clarify, what I did is simply adding a means of forcing an IDR for resynchronisation, the encoder is still managing it by itself otherwise. Based on what vendors have told us, using an open/infinite GOP and external IDR control will also have a negative impact on quality, so it's not advised. Besides that, not all encoders even support this (e.g. VideoToolbox).

> I can start output 1 first, then start output 2. I can also start output 2 first, and then start output 1. Say though, with this change, I "group" encoders A/B together. In this case, I have to start output 1 before output 2 will finish initializing encoders.

I'm not sure there's an elegant way to solve this, the resync mechanism I have build would *technically* work in this case. You just have to let the encoder B start out of sync, and then wait for the resync to happen before starting data capture on output 1, or accept out-of-sync packets for a start ("just" in this case probably means a lot of complexity to track that), However, that would result in output 2 potentially receiving a premature IDR frame.

palana:
> I have one primary concern with this PR, which is that the currently supported behavior of sharing an encoder between multiple outputs, and starting those outputs at different times, becomes restricted with this addition.
> 
> Say I have encoder A, encoder B, output 1, and output 2. I apply encoder A and B to output 1, then I apply encoder A to output 2.
> 
> I can start output 1 first, then start output 2. I can also start output 2 first, and then start output 1. Say though, with this change, I "group" encoders A/B together. In this case, I **have** to start output 1 before output 2 will finish initializing encoders.
> 
> Having this be a limitation of aligning encoder keyframes, to me, is an important issue, as any future multi-destination streaming features which intend to support enhanced broadcast will need to impose this limitation on the user.
> 
> Rodney brought up that perhaps we could take control of the GOP and manually trigger keyframes. I think that this might be an adequate, and more reliable, method of synchronizing keyframes across an "encoder group." I suppose in theory, if an encoder is "grouped", it would disable any automatic keyframe generation in the implementation, and rely on signals from libobs to produce new keyframes. This could possibly be implemented as a flag in a "flags" parameter in the encode callbacks.

At this point, we’ve tested & validated that this set of changes meet the requirements for the Twitch Enhanced Broadcasting use case. If there are additional use cases with new/different requirements, we’ll continue to iterated and bring new PRs that build on this after it gets merged

palana:
@RytoEX: squashed and rebased commits just now

Lain-B:
For the shared encoder situation (i.e. hypothetically using one of the encoders on another output and then starting that output before the output with the group), there are a few ways to handle it, but right now, these encoders will not be shared for the time being (e.g. sharing recording encoder with stream encoder) when using multi-encode. They spawn separate encoders in the enhanced broadcast implementation for the time being.

Here are some proposals to handle the keyframe alignment issue when sharing sub-encoders between multiple outputs in the future if we decide to implement it (which would be ideal at some point):

1. Drop the keyframe alignment requirement and leave it to the service to solve this in another way, likely client-side. This would be my ideal solution if it's possible to do. However, unless we do this soon, we would have to of course change the name after merge (which is annoying but fine if we have to)

2. If removing alignment requirement is not reasonable/doable, the most simple alternative is to start all encoders in a group any time one of its encoders are started up individually. This is not fantastic because if you just want to record then you have to start up many encoders no matter what, but has the benefit of being a change that should in theory be relatively trivial to implement.

3. Another proposal I've been thinking about is when starting up other encoders within an encoder group later, you could probably add somewhat complex handling to discard raw frames and make newly-started encoders wait before processing their first starting frame in order to match the precise timing of keyframes of already-started encoders. I feel like this should in theory be quite possible to do. You also might be able to discord both encoded frames *and* raw frames together as a hack in order to compensate for unknown encoder frame delay. I definitely feel this is quite possible now that I think about it.

4. There is also the proposal to manually signal keyframes ourselves rather than leave it to the encoder, which you had proposed tt. This is also a complex proposal and I am not sure all encoders among all codecs (current and upcoming) support manual keyframe placement. Plus I'm not sure how manual keyframe placement impacts encoders or how they're implemented among encocers. Probably just as complicated as solution 3 however.

Anyway, while keyframe alignment is a requirement, there are no perfect solutions. I would likely opt for my proposal of just starting all encoders in a group if one of them is started elsewhere just simply for the sake of simplicity for the time being but that's just me.

I am fairly confident however that API changes -- outside of the first proposal -- would likely not be necessary regardless of how we choose to handle it (if we choose to handle it).

Lain-B (approved):
There are more than one ways to adapt this later on without modifying the API. Right now I feel relatively good with it and would prefer not to stall this indefinitely. I feel fairly strongly that it can be adapted without necessitating API breakage if we want to add encoder sharing later (except if the alignment requirement is removed, which I'm guessing at this point in time is unlikely).

libobs/obs-output.c
```diff
@@ -1989,6 +1993,85 @@ static void discard_unused_audio_packets(struct obs_output *output,
...
+
+	if (modified_count == data->required_tracks) {
+		da_erase(output->keyframe_group_tracking, idx);
+		return true;
+	}
+	return false;
+}
+
+// Check whether keyframes are emitted from all grouped encoders, and log
+// if keyframes haven't been emitted from all grouped encoders
```
RytoEX, line 2014: Nit.
```suggestion
/* Check whether keyframes are emitted from all grouped encoders, and log
 * if keyframes haven't been emitted from all grouped encoders.
 */
```

Or
```suggestion
/* Check whether keyframes are emitted from all grouped encoders, and log
 * if keyframes haven't been emitted from all grouped encoders. */
```

RytoEX (changes requested):
Nit only, which need not block merging if it doesn't get changed. Otherwise, I trust @Lain-B 's review and assessment.

libobs/obs-output.c
palana: the second suggestion seems more common in libobs code, so going with that

libobs/obs-output.c
```diff
@@ -1989,6 +1993,85 @@ static void discard_unused_audio_packets(struct obs_output *output,
...
+		if (data->group_id !=
+		    (uintptr_t)packet->encoder->encoder_group) {
+			idx += 1;
+			continue;
+		}
+
+		if (data->pts < packet->pts) {
+			if (data->seen_on_track[packet->track_idx] ==
+			    KEYFRAME_TRACK_STATUS_NOT_SEEN) {
+				blog(LOG_WARNING,
```
RytoEX, line 2040: Just a note for the future: LOG_WARNING may be too noisy. We'll have to evaluate as this gets used in the wild if this should be LOG_DEBUG.

RytoEX (commented):
Leaving a general comment/question.

Overall this looks okay to me, and has @Lain-B 's approval. So pending an answer to this question, this should be able to be merged. If there are still pending questions/concerns after that, particularly about exported functions, we can continue to change them until we ship a stable release with those functions.

libobs/obs-output.c
Lain-B: This message doesn't get output normally, it's basically considered an error if this error message is output, so it's probably important and makes sense to output it as a general log warning in this case. Otherwise we'll never know about it from user logs.

Lain-B: (Or rather considered a warning that something is going awry and should be examined. Stream should still technically work I believe)

Lain-B: Huh, Ryan sent me a log file where the user had dynamic bitrate enabled (which I assume isn't supported) and it *does* get very verbose with multiple messages per second. Hmm.

Lain-B: Well I think that's being patched so I think it'll be fine for now. If we have any issues with logging we can adjust it later.

RytoEX: My comment was mostly, "Let's be mindful of this and keep an eye out as this code makes it into use in the wild." I wanted it noted that I raised the concern as something to keep an eye on, but not a blocker at this time.

RytoEX (approved):
If there's anything else that needs fixing here, it can be fixed in follow-ups, or we can adjust as needed on our own.

docs/sphinx/reference-outputs.rst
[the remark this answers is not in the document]
RytoEX: I do not have a strong opinion on this, and I don't want to block further. In the future, if we find that the docs lack clarity, we can clarify them then.

PR #10350, opened 2024-03-08
mac-avcapture: Fix possible crash if camera uses BGRA format

PatTheMav:
### Description
Add check for possible NULL value returned for colour primaries of a capture device' image format.

### Motivation and Context
BGRA or other formats that do not use color primaries will not yield a valid color primary value. Initializing the CFComparisonResult to a default value and replacing it only if a non-NULL color primary value was retrieved avoids a possible crash.

### How Has This Been Tested?
Tested with OBS Virtual Camera which has a BGRA output format, which has no colour primary information attached to it.

### Types of changes

### Checklist:


RytoEX (approved):
Looks good to me. Also verified this fixes the crash.

PR #10351, opened 2024-03-08
UI/obs-ffmpeg: Update NVENC internal ids

derrod:
> [!Important]
> This PR includes #10122 and #10133 as it is based on those changes. Only the last two commits are what's actually being changed compared to those.

### Description

Updates NVENC encoder ID from `jim_nvenc` to more neutral `obs_nvenc` and brings them in line with the new non-texture encoder. Maintains backwards compatibility by registering a deprecated encoder with the old ID to not break existing configs.

### Motivation and Context

Clean up naming schemes.

### How Has This Been Tested?

Verified old encoder id is still usable, but that only new ones are used in the UI.

### Types of changes

### Checklist:


ma3uk:
There is a problem with starting the encoding process on a multi-GPU system if OBS is running on a graphics card from another manufacturer
image

derrod:
> There is a problem with starting the encoding process on a multi-GPU system if OBS is running on a graphics card from another manufacturer

That's probably just due to me forgetting to update the ID in the non-texture encoder fallback, should be fixed now.

WizardCM:
I wonder if we should perform automatic migration, and provide a way to have "alternate/legacy" codes that are exposed via the API, otherwise a user manually switching from the legacy ID to the new one would clear their encoder settings.

derrod:
> I wonder if we should perform automatic migration, and provide a way to have "alternate/legacy" codes that are exposed via the API, otherwise a user manually switching from the legacy ID to the new one would clear their encoder settings.

Migrations could probably be done reasonably simple. I just didn't want to do it.

derrod:
Added a simple migration, won't be backwards compatible, i.e. loading the profile in a previous OBS version will not find the encoder, but that's the best we can do.

derrod:
Leavint this to be figured out with the NVENC refactor for 31.0.

PR #10354, opened 2024-03-09
CI: Use phased group rollouts for Sparkle stable channel updates

RytoEX:
### Description

Instead of pushing the Sparkle update to all macOS users at once, use Sparkle's "phased group rollouts" ("phased rollout interval") feature to split the notifications up into 7 groups separated by 1 day intervals. This potentially allows us to catch critical issues before all macOS users get the update notification.

### Motivation and Context

We had a critical issue with OBS Studio 30.0.1 that we had to hotfix that day with OBS Studio 30.0.2. At the time, we considered the possibility of if we could have a more gradual rollout of macOS updates. This is possible with Sparkle's phased group rollouts.

https://sparkle-project.org/documentation/publishing/#phased-group-rollouts

It's possible that we conclude that this isn't worth the added complexity, but I wanted to get the code changes done so that we didn't have to worry about _if_ it was possible, only if we wanted to actually try it.

### How Has This Been Tested?

I tested the changes locally to make sure that the CI scripts should still work and produce valid appcasts. I did not test the subsequently produced appcast files in any way other than checking them visually.

### Types of changes

### Checklist:

### DIFF
--- a/.github/actions/sparkle-appcast/action.yaml
+++ b/.github/actions/sparkle-appcast/action.yaml
@@ -160,6 +160,10 @@ runs:
           --channel '${{ inputs.channel }}'
         )
 
+        if [[ '${{ inputs.channel }}' == 'stable' ]] {
+          sparkle_options+=(--phased-rollout-interval 86400)
+        }
+
         Sparkle/bin/generate_appcast \
           ${sparkle_options} \
           builds
--- a/.github/actions/sparkle-appcast/action.yaml
+++ b/.github/actions/sparkle-appcast/action.yaml
@@ -150,15 +150,18 @@ runs:
         print -n '${{ inputs.sparklePrivateKey }}' >> eddsa_private.key
         local feed_url='${{ steps.builds.outputs.feedUrl }}'
         local arch=${${${(s:_:)feed_url:t}[2]}//x86/x86_64}
+        local sparkle_options=(
+          --verbose
+          --ed-key-file eddsa_private.key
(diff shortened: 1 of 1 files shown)


derrod (approved):
Seems fine

.github/actions/sparkle-appcast/action.yaml
```diff
@@ -150,15 +150,22 @@ runs:
         print -n '${{ inputs.sparklePrivateKey }}' >> eddsa_private.key
         local feed_url='${{ steps.builds.outputs.feedUrl }}'
         local arch=${${${(s:_:)feed_url:t}[2]}//x86/x86_64}
+        local sparkle_options=(
+          --verbose
+          --ed-key-file eddsa_private.key
+          --download-url-prefix '${{ inputs.urlPrefix }}/'
+          --full-release-notes-url "${feed_url//updates_*/notes_${{ inputs.channel }}.html}"
+          --maximum-versions 0
+          --maximum-deltas ${{ inputs.count }}
```
PatTheMav, line 159: Enclose the workflow variable in single quotes like in all other instances.

RytoEX: I thought I had issues with it locally, and it wasn't enclosed in quotes in the original code, so I didn't want to touch it. I'll try to re-test on my MacBook.

RytoEX: Changed.

PatTheMav (approved):

PR #10376, opened 2024-03-14
deps/media-playback: Fix hardware decoding of streams

### LINKED ISSUE #10369: Remote media streams not displaying video when using hardware decoding in Media Source
### Operating System Info

Windows 10

### Other OS

_No response_

### OBS Studio Version

30.1.0

_No response_

### OBS Studio Log URL

https://obsproject.com/logs/6LiWnBULTyTYyaSC

### OBS Studio Crash Log URL

_No response_

### Expected Behavior

Video output of caller PC displaying in preview of listener PC via SRT stream.

### Current Behavior

No display in preview, not even black screen.

### Steps to Reproduce

1. Start OBS as admin on both caller and listener.
2. Set stream output of caller PC to listener address.
3. Add media source on listener, untick local file, input as same address from caller stream.
4. Start streaming on caller PC.

Everything was working before the latest update, nothing un/installed on either PC since last using OBS. Updated video drivers before bug report.
Logs from caller: https://obsproject.com/logs/aT36horil2AtOrE0

pkviet:
### Description
Fixes #10369.
Since the update to FFmpeg 6.1, streams to a Media Source are broken if hardware decoding is enabled (both RTMP or SRT have been reported). The video is black although the audio is decoded fine. The manual copy of metadata introduced in commit [1] does not work any more for some unfathomable reasons.
As a fix we call instead the av_frame_copy_props function used in FFmpeg app in a similar context (hardware decoding) [2].
The metadata are copied without issues.
There is no need to guard the use of that function since it was introduced 9 years ago in avutil/frame.c.

[1] https://github.com/obsproject/obs-studio/commit/22fde5cdcd1c784cd60b8a8cd50af02e7e537182
[2] https://github.com/FFmpeg/FFmpeg/master/fftools/ffmpeg_dec.c

NB: on FFmpeg master as of today (3/15/24), the relevant line I've copied is 
https://github.com/FFmpeg/FFmpeg/blame/master/fftools/ffmpeg_dec.c#L336

### Motivation and Context
Bugfix.

### How Has This Been Tested?
HW decoding works again with Media Source when one plays an RTMP stream.
HW decoding still works for local media files.

### Types of changes

### Checklist:

### DIFF
--- a/deps/media-playback/media-playback/decode.c
+++ b/deps/media-playback/media-playback/decode.c
@@ -338,15 +338,12 @@ static int decode_packet(struct mp_decode *d, int *got_frame)
 		}
 
 		int err = av_hwframe_transfer_data(d->sw_frame, d->hw_frame, 0);
+		if (err == 0) {
+			err = av_frame_copy_props(d->sw_frame, d->hw_frame);
+		}
 		if (err) {
 			ret = 0;
 			*got_frame = false;
-		} else {
-			d->sw_frame->color_range = d->hw_frame->color_range;
-			d->sw_frame->color_primaries =
-				d->hw_frame->color_primaries;
-			d->sw_frame->color_trc = d->hw_frame->color_trc;
-			d->sw_frame->colorspace = d->hw_frame->colorspace;
 		}
 	}
 


deps/media-playback/media-playback/decode.c
```diff
@@ -338,15 +338,12 @@ static int decode_packet(struct mp_decode *d, int *got_frame)
 		}
 
 		int err = av_hwframe_transfer_data(d->sw_frame, d->hw_frame, 0);
+		if (err == 0) {
+			err = av_frame_copy_props(d->sw_frame, d->hw_frame);
+		}
 		if (err) {
```
RytoEX, line 344: These should be mutually exclusive, so I think they can be part of the same `if-else`.

| condition  | err | evaluation |
| ---------- | --- | ---------- |
| `err == 0` | `0` | `true`     |
| `err == 0` | `1` | `false`    |
| `err`      | `0` | `false`    |
| `err`      | `1` | `true`     |

RytoEX (changes requested):

deps/media-playback/media-playback/decode.c
derrod: The point is that `av_frame_copy_props()` may set `err` to a non-zero error code as well.

RytoEX: Ah, you are correct. Sorry about that. My mind was elsewhere. Disregard.

jpark37 (approved):

derrod (approved):

RytoEX (changes requested):
Nit:
Change the commit message subject to:
> deps/media-playback: Fix hardware decoding of streams

pkviet:
> Nit: Change the commit message subject to:
> 
> > deps/media-playback: Fix hardware decoding of streams

Done.

RytoEX (approved):

Issue #10377, opened 2024-03-15
replayd crashes on OBS launch if a capture source is inactive/missing on launch


MysticalOS:
### Operating System Info

macOS 14

### OBS Studio Version

30.1.0

### OBS Studio Log URL

https://obsproject.com/logs/Qr7V8oO46gI98Fid

### Expected Behavior

Not sure if even possible, as it can just be an apple bug in replayd daemon, but not to crash if a capture source just happens to be missing.

### Current Behavior

If you start OBS and any of your scenes use a capture source that is inactive, the replayd process is crashed for each instance of missing capture. (maybe application capture but app isnt open, or display capture for a display that's currently turned off, etc).

This means if two sources are missing, it actually crashes twice in a row during the launch.

### Steps to Reproduce

1. Create multiple capture sources, maybe screen capture for monitor 1 2 and 3
2. close OBS, disable any of those sources, maybe turn screens off
3. Open OBS nd watch replayd crash for each missing instance.
...

### Anything else we should know?

latest macOS 14.4
3 external monitors
Elgato Capture Card
Several scenes for various capture sources for above

if all present no crashes, if ANY missing, replayd crashes at launch for each missing capture.

PR #10378, opened 2024-03-15
mac-capture: Standardize early return behavior in init_screen_stream()

jcm93:
### Description

Makes consistent the early return behavior when creating a macOS Screen Capture, such that if the user is creating a capture with an invalid/stale window or display, we abandon initialization properly in all three capture modes.

### Motivation and Context

Per https://github.com/obsproject/obs-studio/issues/10377 as well as lingering reports about stale capture indicators, ScreenCaptureKit still does not behave well if we try to initialize streams with bad parameters, which can cause side effects for other applications relying on replayd. Additionally, the way we are currently early returning from a bad Display Capture, we seem to invalidate the source (making it unmodifiable) for the remainder of that OBS session, which is not desirable.

On 14.4, before these changes, my testing showed that:

1. If initializing a Display Capture with a no-longer-connected display, we could no longer modify that source within that OBS session. Also, replayd crashes.
2. If initializing an Application Capture with a no-longer-connected display, we can still modify the source, but replayd crashes.

With these changes:

1. Initializing a Display Capture with a no-longer-connected display fails, but the source is still modifiable and replayd does not crash.
2. Initializing an Application Capture with a no-longer-connected display, we can still modify the source, and replayd does not crash.

Initializing an Application Capture with an empty application list is supported by SCK and does not crash replayd, so we do not worry about that case.

Fixes https://github.com/obsproject/obs-studio/issues/10377

### How Has This Been Tested?
Tested locally on my 14.4 Apple Silicon machine.

### Other

I considered using a `goto`, but I didn't really want to introduce one so I opted not to.

### Types of changes

### Checklist:

### DIFF
--- a/plugins/mac-capture/mac-sck-video-capture.m
+++ b/plugins/mac-capture/mac-sck-video-capture.m
@@ -106,14 +106,14 @@ static bool init_screen_stream(struct screen_capture *sc)
     switch (sc->capture_type) {
         case ScreenCaptureDisplayStream: {
             SCDisplay *target_display = get_target_display();
-
-            if (!target_display) {
+            if (target_display == nil) {
                 MACCAP_ERR("init_screen_stream: Invalid target display ID:  %u\n", sc->display);
-
                 os_sem_post(sc->shareable_content_available);
-                return false;
+                sc->disp = NULL;
+                os_event_init(&sc->disp_finished, OS_EVENT_TYPE_MANUAL);
+                os_event_init(&sc->stream_start_completed, OS_EVENT_TYPE_MANUAL);
+                return true;
             }
-
             if (sc->hide_obs) {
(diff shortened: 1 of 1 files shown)


jcm93:
Hmm. With further testing, I've stopped seeing actual replayd *crashes* with the old behavior on 14.4. Could be due to some other factor on my machine. Regardless, replayd is definitely more crash prone on older configurations anyway so good to address.

PatTheMav (approved):
Looks right logically to me - the entire source is on my docket for a complete refactor anyway, but the change makes the source behaviour consistent across all capture types.

RytoEX (approved):
I'm not super familiar with Objective-C, but this looks like it correctly mirrors the changes in #9803.

PR #10388, opened 2024-03-15
obs-qsv11: Remove adaptive CQM option

derrod:
### Description

Removes "Subjective Video Enhancements" option from QSV encoder.

### Motivation and Context

Without setting the `ScenarioInfo` to game streaming this is a no-op[^1] and its existence is just confusing to users.

Setting the scenario was removed in #9503, see that PR for a summary of the issues encountered with this option.

### How Has This Been Tested?

Compiles, works.

### Types of changes

### Checklist:

[^1]: https://github.com/oneapi-src/oneVPL-intel-gpu/blob/main/_studio/mfx_lib/encode_hw/av1/agnostic/base/av1ehw_base_qmatrix.cpp#L110-L113


ma3uk:
@derrod I think it's better not to delete CQM, but to add ScenarioInfo SCENARIO_REMOTE_GAMING. This will restore the functionality of CQM and will not cause problems, as with SCENARIO_GAME_STREAMING. Also, with SCENARIO_REMOTE_GAMING, the Lookahead parameter is available with a maximum value of 8, but not exceeding GopRefDist.
Also, if you refer to the Intel media driver repository, you can see that SCENARIO_REMOTE_GAMING in the new drivers will include Boost frequencies for the GPU, i.e. when encoding, frequencies will not be reset to 1000 MHz on Intel Arc

derrod:
We can revisit other options at a later date, also based on feedback we get from Intel. Right now this option does nothing, so it makes sense to remove it.

Lain-B (approved):

RytoEX (approved):
Seems fine.

PR #10389, opened 2024-03-16
cmake: Fix MSVC runtime selection for win-dshow and win-capture targets

PatTheMav:
### Description
Adds the missing explicit MSVC runtime selection to the the 32-bit `win-capture` target and also moves the same selection out of the shared virtual cam interface library to the actual architecture-specific targets.

### Motivation and Context
The `MSVC_RUNTIME_LIBRARY` target property when set on an `INTERFACE` type library is not inherited by targets which consume the library, thus their property is set automatically set by CMake.

This affected the virtual camera target which uses an `INTERFACE` library for sources and settings shared between its 64-bit and 32-bit variants.

The PR sets the property directly on the "final" output targets to fix this.

### How Has This Been Tested?
Needs to be tested on an Intel-based Windows installation.

### Types of changes

### Checklist:

### DIFF
--- a/plugins/win-capture/inject-helper/cmake/32bit-build.cmake
+++ b/plugins/win-capture/inject-helper/cmake/32bit-build.cmake
@@ -2,4 +2,7 @@ project(inject-helper)
 
 add_executable(inject-helper)
 target_link_libraries(inject-helper PRIVATE _inject-helper)
-set_property(TARGET inject-helper PROPERTY OUTPUT_NAME inject-helper32)
+
+# cmake-format: off
+set_target_properties(inject-helper PROPERTIES OUTPUT_NAME inject-helper32 MSVC_RUNTIME_LIBRARY "MultiThreaded$<$<CONFIG:Debug>:Debug>")
+# cmake-format: on
--- a/plugins/win-capture/get-graphics-offsets/cmake/32bit-build.cmake
+++ b/plugins/win-capture/get-graphics-offsets/cmake/32bit-build.cmake
@@ -2,4 +2,7 @@ project(get-graphics-offsets)
 
 add_executable(get-graphics-offsets)
 target_link_libraries(get-graphics-offsets PRIVATE _get-graphics-offsets)
-set_property(TARGET get-graphics-offsets PROPERTY OUTPUT_NAME get-graphics-offsets32)
+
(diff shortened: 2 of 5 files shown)


PatTheMav:
@derrod could you check the artifact once it's done?

RytoEX (changes requested):
This also needs to update `plugins/win-capture/get-graphics-offsets/cmake/32bit-build.cmake` and `plugins/win-capture/inject-helper/cmake/32bit-build.cmake`.

PatTheMav:
> This also needs to update `plugins/win-capture/get-graphics-offsets/cmake/32bit-build.cmake` and `plugins/win-capture/inject-helper/cmake/32bit-build.cmake`.

Done.

plugins/win-capture/inject-helper/cmake/32bit-build.cmake
```diff
@@ -2,4 +2,7 @@ project(inject-helper)
 
 add_executable(inject-helper)
 target_link_libraries(inject-helper PRIVATE _inject-helper)
-set_property(TARGET inject-helper PROPERTY OUTPUT_NAME inject-helper32)
+
+# cmake-format: off
+set_target_properties_obs(inject-helper PROPERTIES OUTPUT_NAME inject-helper32 MSVC_RUNTIME_LIBRARY "MultiThreaded$<$<CONFIG:Debug>:Debug>")
```
RytoEX, line 7: Maybe `set_target_properties`?

RytoEX (changes requested):

plugins/win-capture/inject-helper/cmake/32bit-build.cmake
PatTheMav: Ugh, correct.

RytoEX:
Looks correct at a glance. Will double check later.

RytoEX (approved):
Manually checked the properties of each affected VS project, both 32-bit and 64-bit, and this seems correct now.

derrod (approved):
Confirmed using PEStudio that binaries now no longer try to load VC runtime libraries.

WizardCM (approved):
Confirmed, this fixes both issues on a system that doesn't have the 32bit redistributables installed.

PR #10391, opened 2024-03-16
CI: Update formatting actions to use git diff filter before formatting 

PatTheMav:
### Description
Update code formatting actions to use the `check-changes` action as a trigger for whether or not format checks should actually run and updates that action to pass custom `diff-filter` arguments for `git diff`.

### Motivation and Context
Using git's diff-filter when checking for changed files allows the actions to ignore any deleted files. Combined with explicitly passing glob expressions for source files to git directly allows the command to yield a list of changed files compatible with a given formatter directly.

By updating only the `check-changes` action, the code-formatting actions are able to use that actions as well, reducing code duplication.

### How Has This Been Tested?
Requires CI runs for testing.

### Types of changes

### Checklist:

### DIFF
--- a/.github/actions/check-changes/action.yaml
+++ b/.github/actions/check-changes/action.yaml
@@ -3,7 +3,7 @@ description: Checks for changed files compared to specific git reference and glo
 inputs:
   baseRef:
     description: Git reference to check against
-    required: true
+    required: false
   ref:
     description: Git reference to check with
     required: false
@@ -15,12 +15,16 @@ inputs:
     description: Use fallback compare against prior commit
     required: false
     default: 'true'
+  diffFilter:
+    description: git diff-filter string to use
+    required: false
+    default: ''
 outputs:
   hasChangedFiles:
     value: ${{ steps.checks.outputs.hasChangedFiles }}
     description: True if specified files were changed in comparison to specified git reference
   changedFiles:
-    value: ${{ toJSON(steps.checks.outputs.changedFiles) }}
+    value: ${{ steps.checks.outputs.changedFiles }}
     description: List of changed files
 runs:
   using: composite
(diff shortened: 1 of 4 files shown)


.github/actions/check-changes/action.yaml
```diff
@@ -31,26 +35,42 @@ runs:
...
+          if ! git cat-file -e "${GIT_BASE_REF}" &> /dev/null; then
+            echo "::warning::Provided base reference ${GIT_BASE_REF} is invalid"
+            if [[ "${USE_FALLBACK}" == 'true' ]]; then
+              GIT_BASE_REF='HEAD~1'
+            fi
           fi
+        else
+          if ! git cat-file -e ${GITHUB_REF_BEFORE} &> /dev/null; then
+            GITHUB_REF_BEFORE='4b825dc642cb6eb9a060e54bf8d69288fbee4904'
+          fi
```
RytoEX, line 58: I see that this was moved from the other actions, but what even is this ref?

.github/actions/check-changes/action.yaml
```diff
@@ -31,26 +35,42 @@ runs:
...
         fi
 
-        changes=($(git diff --name-only ${GIT_BASE_REF} ${GIT_REF} -- ${{ inputs.checkGlob }}))
+        changes=($(git diff --name-only --diff-filter="${DIFF_FILTER}" ${GIT_BASE_REF} ${GIT_REF} -- ${{ inputs.checkGlob }}))
 
         if (( ${#changes[@]} )); then
           file_string="${changes[*]}"
           echo "hasChangedFiles=true" >> $GITHUB_OUTPUT
-          echo "changedFiles=[${file_string// /,}]" >> GITHUB_OUTPUT
+          echo "changedFiles=[\"${file_string// /\",\"}\"]" >> $GITHUB_OUTPUT
```
RytoEX, line 73: We were originally calling toJSON on this, but now we are not. What was the intent of the curly brace element?

.github/actions/check-changes/action.yaml
```diff
@@ -15,12 +15,16 @@ inputs:
...
+    description: git diff-filter string to use
+    required: false
+    default: ''
 outputs:
   hasChangedFiles:
     value: ${{ steps.checks.outputs.hasChangedFiles }}
     description: True if specified files were changed in comparison to specified git reference
   changedFiles:
-    value: ${{ toJSON(steps.checks.outputs.changedFiles) }}
+    value: ${{ steps.checks.outputs.changedFiles }}
```
RytoEX, line 27: Intentional?

RytoEX (changes requested):
Looks fine except for my questions here.

.github/actions/check-changes/action.yaml
PatTheMav: It's a "secret" git hash of an empty tree that is guaranteed to always exist.

.github/actions/check-changes/action.yaml
PatTheMav: The original idea was to have the output be a valid JSON array with all the changed files. In tests this proved to be insufficient, as the string tokens had to enclosed in quotes, which mean that additional string operations were needed:

So `${file_string// /\",\"}` replaces every space with `\",\"`, so `a b c` becomes `a\",\"b\",\"c`.

This is then enclosed with `\"` again to generate `\"a\",\"b\",\"c\"` which is a properly list of JSON string array tokens.

PatTheMav: `toJSON` was removed because outputs for composite actions are always of type `string`, so we cannot provide a complex type as an output. Instead we provide a JSON string representation of that type.

.github/actions/check-changes/action.yaml
PatTheMav: See above, the type of outputs is `string` so converting to a complex type will not work.

.github/actions/check-changes/action.yaml
RytoEX: Oh. Neat.

.github/actions/check-changes/action.yaml
RytoEX: Ah, I see. It's ~~cursed~~ fancy string substitution in-place before it gets echoed to $GITHUB_OUTPUT. I'd have probably just done the string substitution where I didn't have to escape the quotes, but that may be personal preference.
```yaml
file_string="${file_string// /,}"
echo "changedFiles=[\"${file_string}\"]" >> $GITHUB_OUTPUT
```

.github/actions/check-changes/action.yaml
RytoEX: Got it.

.github/actions/check-changes/action.yaml
PatTheMav: That's incomplete, you cannot just use a single comma as a glue, you need to use the closing double quotes of the prior token, followed by the comma, and the opening double quotes for the next token again.

In your example you end up with `["a,b,c"]` which would be an array with a single element instead of 3 separate elements.

RytoEX: Ah, I see now what the goal is. I'll think on it.

RytoEX: To do what I was thinking (separate the string substitution from the final output), you'd have to do this:
```sh
file_string=\"${file_string// /\",\"}\"
echo "changedFiles=[${file_string}]" >> $GITHUB_OUTPUT
```

It doesn't really save any escaping, though it does provide a bit clearer separation of what is occurring at the cost of one-liner elegance. Let me know what you prefer.

PatTheMav: Again, that would not be valid syntax, because you have to wrap the first string in double quotes again (otherwise string expansion is not applied) and I don't see much benefit in splitting it up, instead we trigger two expansions where one would suffice.

RytoEX: I don't see how it's not valid syntax. I tested it repeatedly locally and ensured that it achieves the desired result.

From my perspective, the separation provides a clearer picture of why the escapes are being done. I literally could not read the original syntax in a single pass. It's not like the extra expansion costs us anything measurable here.

That said, I do not want to bike shed this to death, so it's fine as-is.

RytoEX: Discussed off-thread and @PatTheMav did say that this does seem to be valid syntax. However, it did occur to me that this action is run in bash and I tested in zsh, so I don't know for sure if it's still valid in bash, though I think it would be valid.

That said, this is still fine as-is, if not a little confusing at first glance.

RytoEX (approved):

PR #10392, opened 2024-03-16
decklink: Output BGRA only when supported 

### LINKED ISSUE #10380: Updates to the Decklink Plugin do not properly support all resolutions
### Operating System Info

macOS 14

### Other OS

_No response_

### OBS Studio Version

Git

_No response_

### OBS Studio Log URL

N/A

### OBS Studio Crash Log URL

N/A

### Expected Behavior

Decklink output should show the correct image

### Current Behavior

We send the wrong pixel format so the image is crooked

### Steps to Reproduce

Run decklink output on a ultra studio 4k mini at UHD60 

_No response_

DDRBoxman:
### Description

Currently if you select a mode that doesn't support RGBA on Decklink output the image is corrupted.

Fixes https://github.com/obsproject/obs-studio/issues/10380

### Motivation and Context

https://github.com/obsproject/obs-studio/issues/10380

### How Has This Been Tested?

Sent video to the Ultra Studio 4k mini with multiple video modes

### Types of changes

### Checklist:

### DIFF
--- a/plugins/decklink/decklink-device-instance.cpp
+++ b/plugins/decklink/decklink-device-instance.cpp
@@ -76,6 +76,21 @@ static inline enum video_format ConvertPixelFormat(BMDPixelFormat format)
 	}
 }
 
+static inline int PixelFormatByteSize(BMDPixelFormat format)
+{
+	switch (format) {
+	case bmdFormat8BitBGRA:
+	case bmdFormat8BitARGB:
+	case bmdFormat10BitRGB:
+	case bmdFormat10BitRGBXLE:
+	case bmdFormat10BitRGBX:
+		return 4;
+	default:
+	case bmdFormat8BitYUV:
+		return 2;
+	}
+}
+
 static inline int ConvertChannelFormat(speaker_layout format)
 {
 	switch (format) {
@@ -604,11 +619,19 @@ bool DeckLinkDeviceInstance::StartOutput(DeckLinkDeviceMode *mode_)
 		return false;
 	}
 
+	BMDDisplayMode *actualMode = nullptr;
+	decklink_bool_t modeSupportsBGRA = false;
+	output_->DoesSupportVideoMode(
+		bmdVideoConnectionUnspecified, mode_->GetDisplayMode(),
+		bmdFormat8BitBGRA, bmdNoVideoOutputConversion,
+		bmdSupportedVideoModeDefault, actualMode, &modeSupportsBGRA);
+
 	ComPtr<IDeckLinkKeyer> deckLinkKeyer;
 	if (device->GetKeyer(&deckLinkKeyer)) {
 		const int keyerMode = device->GetKeyerMode();
-		if (keyerMode) {
-			deckLinkKeyer->Enable(keyerMode == 1);
+		if (keyerMode && modeSupportsBGRA) {
+			const bool isExternal = keyerMode == 1;
+			deckLinkKeyer->Enable(isExternal);
 			deckLinkKeyer->SetLevel(255);
 		} else {
 			deckLinkKeyer->Disable();
@@ -618,7 +641,25 @@ bool DeckLinkDeviceInstance::StartOutput(DeckLinkDeviceMode *mode_)
 	frameQueueDecklinkToObs.reset();
 	frameQueueObsToDecklink.reset();
 
-	const int rowSize = decklinkOutput->GetWidth() * 4;
+	struct obs_video_info ovi;
+	const enum video_colorspace colorspace =
+		obs_get_video_info(&ovi) ? ovi.colorspace : VIDEO_CS_DEFAULT;
+
+	const bool source_hdr = (colorspace == VIDEO_CS_2100_PQ) ||
+				(colorspace == VIDEO_CS_2100_HLG);
+	const bool enable_hdr =
+		source_hdr &&
+		(obs_output_get_video_conversion(decklinkOutput->GetOutput())
+			 ->colorspace == VIDEO_CS_2100_PQ);
+
+	BMDPixelFormat pixelFormat = modeSupportsBGRA ? bmdFormat8BitBGRA
+						      : bmdFormat8BitYUV;
+	if (enable_hdr) {
+		pixelFormat = bmdFormat10BitRGBXLE;
+	}
+
+	const int rowSize =
+		decklinkOutput->GetWidth() * PixelFormatByteSize(pixelFormat);
 	const int frameSize = rowSize * decklinkOutput->GetHeight();
 	for (std::vector<uint8_t> &blob : frameBlobs) {
 		blob.assign(frameSize, 0);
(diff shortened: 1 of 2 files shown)


RytoEX (changes requested):
Commit message nits:
* doesnt -> doesn't

If this fixes https://github.com/obsproject/obs-studio/issues/10380, use "Fixes #10380" in the PR description to automatically link them.

plugins/decklink/decklink-device-instance.cpp
```diff
@@ -604,11 +618,21 @@ bool DeckLinkDeviceInstance::StartOutput(DeckLinkDeviceMode *mode_)
...
+				      bmdSupportedVideoModeKeying, actualMode,
+				      &keyerSupportedForMode);
+
 	ComPtr<IDeckLinkKeyer> deckLinkKeyer;
 	if (device->GetKeyer(&deckLinkKeyer)) {
 		const int keyerMode = device->GetKeyerMode();
-		if (keyerMode) {
-			deckLinkKeyer->Enable(keyerMode == 1);
+		if (keyerMode && keyerSupportedForMode) {
+			bool isExternal = keyerMode == 1;
```
RytoEX, line 634:
```suggestion
			const bool isExternal = keyerMode == 1;
```

plugins/decklink/decklink-output.cpp
```diff
@@ -32,13 +32,30 @@ static void *decklink_output_create(obs_data_t *settings, obs_output_t *output)
...
+		BMDDisplayMode *actualMode = nullptr;
+		decklink_bool_t modeSupportsBGRA = false;
+		output_->DoesSupportVideoMode(bmdVideoConnectionUnspecified,
+					      mode->GetDisplayMode(),
+					      bmdFormat8BitBGRA,
+					      bmdNoVideoOutputConversion,
+					      bmdSupportedVideoModeKeying,
+					      actualMode, &modeSupportsBGRA);
+
+		bool supportsHDR = (device->GetSupportsHDRMetadata() &&
```
RytoEX, line 47:
```suggestion
		const bool supportsHDR = (device->GetSupportsHDRMetadata() &&
```

RytoEX (changes requested):

plugins/decklink/decklink-output.cpp
```diff
@@ -32,13 +32,30 @@ static void *decklink_output_create(obs_data_t *settings, obs_output_t *output)
...
+					      actualMode, &modeSupportsBGRA);
+
+		const bool supportsHDR = (device->GetSupportsHDRMetadata() &&
+					  !decklinkOutput->force_sdr);
+
 		struct video_scale_info to = {};
-		to.format = VIDEO_FORMAT_BGRA;
+		to.format = (modeSupportsBGRA && supportsHDR)
+				    ? VIDEO_FORMAT_BGRA
+				    : VIDEO_FORMAT_UYVY;
```
jpark37, line 53: I would prefer not dropping down to UYVY unless the keyer was both unsupported and requested by user setting. That probably means reacting to KEYER setting updates in decklink_output_update among other changes.

DDRBoxman: The keyer flag is a red herring here, I'll go ahead and remove it or clarity. These video modes don't support BGRA without the keyer either which is the real issue.

jpark37: I think we should call `DoesSupportVideoMode` with `keyerMode ? bmdSupportedVideoModeKeying : bmdSupportedVideoModeDefault` or something along those lines to be sure.

DDRBoxman: Well the keyer requires the RGBA mode to work so just checking for RGBA should cover both. We may want to grey out the keyer options per video mode in the UI with that check though.

jpark37: Ok, I'm happy with Default. Thanks.

RytoEX (changes requested):
Fix the commit message so that the subject is not wrapped on GitHub (70 characters or less, I think).

If you need to elaborate, use the commit message body.

Lain-B (commented):
Is it intentional to always lock out BGRA/RGBA if it's not HDR? Or am I misunderstanding the PR?

DDRBoxman:
I tweaked the logic in decklink-output.cpp to actually match correctly on HDR and keyer. It worked on my device since it supports both but this should fix compatibility on more devices.

norihiro:
The commit message has a too long description in a line, 107 characters, limit 72.
> Not all devices support the BGRA pixel format in all video modes. Make sure we fallback to YUV when needed.

RytoEX (changes requested):
@norihiro is right. Wrap the commit message body at 72 characters.
Also, nits:
* fallback -> fall back
* Only output BGRA when supported -> Output BGRA only when supported

On "fallback", the former is a noun. The latter is a verb.

If nits are addressed and @Lain-B or @jpark37 are happy with the PR itself, this should be good afterward.

RytoEX (changes requested):
Still nit on commit message subject to place verb first:
* Only output BGRA when supported -> Output BGRA only when supported

> decklink: Output BGRA only when supported

RytoEX (approved):
Nits fixed. Will await other reviews.

jpark37 (approved):

plugins/decklink/decklink-output.cpp
```diff
@@ -32,15 +32,30 @@ static void *decklink_output_create(obs_data_t *settings, obs_output_t *output)
...
+					      bmdNoVideoOutputConversion,
+					      bmdSupportedVideoModeDefault,
+					      actualMode, &modeSupportsBGRA);
+
+		const bool supportsHDR = (device->GetSupportsHDRMetadata() &&
+					  !decklinkOutput->force_sdr);
+
 		struct video_scale_info to = {};
-		to.format = VIDEO_FORMAT_BGRA;
+		to.format = (modeSupportsBGRA || supportsHDR)
```
Lain-B, line 51: Noticed that this code is  "if supports BGR *or* HDR, use BGRA" -- wouldn't this just be "if supports BGR, use BGR"?

Lain-B: It's possible I'm misunderstanding this expression though.

DDRBoxman: yeah that probably makes more sense, I think I got confused since HDR was setup for bgra

DDRBoxman: Hmm HDR is using bmdFormat10BitRGBXLE on the device. We should also probably check if that is supported for the current format too. And then figure out like YUV HDR?

DDRBoxman: Probably set this back to always RGB and then do conversions on the GPU here for when we need YUV

https://github.com/obsproject/obs-studio/blob/004253f29633b38e7b1d8e0598e3f91a2ed790a9/UI/frontend-plugins/decklink-output-ui/decklink-ui-main.cpp#L400

RytoEX:
Setting back to draft per off-thread conversation that the open review thread requires writing new shaders to resolve.

RytoEX:
> Setting back to draft per off-thread conversation that the open review thread requires writing new shaders to resolve.

I presume this is still the case?

PR #10400, opened 2024-03-18
obs-qsv11: Check all encoders are Intel devices

kkartaltepe:
### Description

On linux we enumerate all VA-API devices, which incorrectly enables QSV AV1/HEVC encoders when an AMD devices reports these capabilities.

### Motivation and Context

AMD users shouldnt see QSV encoders.

### How Has This Been Tested?

Incorrectly displayed devices on pure amd machine are hidden. When adding an intel device the encoders are presented again.

This shouldnt affect windows as the enumeration skips non-intel devices.

### Types of changes

### Checklist:


RytoEX (approved):
Seems fine. As written in the PR, this code previously assumed that only Intel adapters were enumerated, which is the case on Windows. On Linux, all adapters are enumerated, so we must check that the adapters are actually Intel devices when checking codec support.

PR #10402, opened 2024-03-18
UI: Change advanced audio controls to use audio_active

Penwy:
### Description

This changes the active/inactive status in the advanced audio properties dialog to be based on the audio_active status of the source instead of its active status.

### Motivation and Context

With the addition of audio capture to game and window capture in 30.1, those sources would appear in the advanced audio properties regardless of whether said audio capture was enabled or not. This can quickly clutter the dialog with sources who aren't actually capturing any audio. This change will hide those by default, by marking them as inactive if the audio capture is not enabled, while still allowing them to be shown by ticking "Show inactive".

### How Has This Been Tested?

Tried every case of making sources active/inactive, verified that all was showing correctly.

### Types of changes

### Checklist:

### DIFF
--- a/UI/adv-audio-control.cpp
+++ b/UI/adv-audio-control.cpp
@@ -54,6 +54,10 @@ OBSAdvAudioCtrl::OBSAdvAudioCtrl(QGridLayout *, obs_source_t *source_)
 	activateSignal.Connect(handler, "activate", OBSSourceActivated, this);
 	deactivateSignal.Connect(handler, "deactivate", OBSSourceDeactivated,
 				 this);
+	audioActivateSignal.Connect(handler, "audio_activate",
+				    OBSSourceActivated, this);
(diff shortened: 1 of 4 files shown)


RytoEX:
Could you provide some basic steps to get into a state that illustrates why this change is necessary?

Lain-B (approved):
I approve of this but want to have it tested by someone else first to verify

Penwy:
> Could you provide some basic steps to get into a state that illustrates why this change is necessary?
The steps would be "Add a large amount of game capture/window capture sources, in a scene where you do not use them as audio capture"

derrod (approved):
Haven't tested it personally but conceptually seems fine. I think the separation of "active" and "audio active" was a more recent addition than the UI here so that's why it wasn't used back then.

Warchamp7 (approved):
Approved from the UX side and echoing the above sentiments of ensuring this gets tested

PR #10418, opened 2024-03-22
UI: Prevent double scrollbars for encoder properties

derrod:
### Description

Prevents double scrollbars in advanced output encoder properties by forcing the scroll area to always be full height.

### Motivation and Context

pictures_say_more_than_a_thousand_words

### How Has This Been Tested?

See above.

### Types of changes

### Checklist:

### DIFF
--- a/UI/properties-view.cpp
+++ b/UI/properties-view.cpp
@@ -143,6 +143,9 @@ void OBSPropertiesView::RefreshProperties()
 	adjustSize();
 	SetScrollPos(h, v, hend, vend);
 
+	if (disableScrolling)
+		setMinimumHeight(widget->minimumSizeHint().height());
+
 	lastFocused.clear();
 	if (lastWidget) {
 		lastWidget->setFocus(Qt::OtherFocusReason);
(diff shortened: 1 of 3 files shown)


Lain-B (approved):

Warchamp7:
Documenting it here for the future. The reason for this is that for some reason, Qt was taking away too much space from the properties scroll area when computing the sizes of the containers after qss gets applied. This seems to be some weirdness around 480 total height, which is the default for widgets.

This particular scrollbar was when the properties were 444 height, and Qt only gave the containing widget 432 height (Which after applying the 56 height for the combo box title and margin, totals up to 488

While the fix here does solve the problem in question, it is a workaround to some weirdness we may want to figure out eventually. I'm not entirely convinced we aren't doing something wrong with our subclassing of QScrollArea

derrod:
> While the fix here does solve the problem in question, it is a workaround to some weirdness we may want to figure out eventually. I'm not entirely convinced we aren't doing something wrong with our subclassing of QScrollArea

This fix is not really to address the scroll area height weirdness we were encountering, while it *does* also fix that, it was aimed at disabling scrolling within the properties entirely. Even if we fixed the issue where the scroll area does not expand as much as it should, you'd still end up with two scrollbars if the settings window is too small - as it is by default - to contain the encoder properties and rest of that page.

PR #3782, opened 2020-11-27
Cleanup native widgets

GeorgesStavracas:
### Description

When marking a widget as native, Qt makes part of the widget hierarchy native as well. This is for historical reasons, at least on Linux, mostly to match how X deals with frame and input windows.

However, this doesn't play well with QtWayland. As a preparation for Wayland, let's disable this behavior.

All credits go to @davidedmundson.

### Motivation and Context

This way of creating native widgets makes OBS Studio behave poorly on Wayland.

### How Has This Been Tested?

Run OBS Studio and make sure everything looks correct, specially with the previews.

### Types of changes

 - Bug fix (non-breaking change which fixes an issue)
 - Code cleanup (non-breaking change which makes code smaller or more readable) 

### Checklist:

- [x] My code has been run through [clang-format](https://github.com/obsproject/obs-studio/blob/master/.clang-format).
- [x] I have read the [**contributing** document](https://github.com/obsproject/obs-studio/blob/master/CONTRIBUTING.rst).
- [x] My code is not on the master branch.
- [x] The code has been tested.
- [x] All commit messages are properly formatted and commits squashed where appropriate.
- [x] I have included updates to all appropriate documentation.

PR #10419, opened 2024-03-22
UI: Reformat .ui files with Qt Designer

derrod:
### Description

Reformats all `.ui` files with Qt Designer by opening and re-saving them to update them to the Qt 6 XML format and fix some formatting errors.

This also fixes `multiviewProjectorMenu` being declared in the wrong section.

### Motivation and Context

Every time we edit a `.ui` file with Qt Designer and save it we have to either manually undo these changes or manually stage only the hunks that contain the changes we want. That is silly.

(Qt will randomly change the size values in the file for some reason so we'll still have to skip those if we want to keep diffs down, but those are usually only one or two per file and that's much more managable. But I propose we just ignore those going forward since they're meaningless anyway.)

Replaces #8372 which got stuck in bikeshedding limbo, only covered the main file, and is conflicted.

### How Has This Been Tested?

OBS still compiles and looks the same.

### Types of changes

### Checklist:

### DIFF
--- a/UI/forms/OBSBasic.ui
+++ b/UI/forms/OBSBasic.ui
@@ -7,7 +7,7 @@
    <rect>
     <x>0</x>
     <y>0</y>
-    <width>1079</width>
+    <width>1086</width>
     <height>729</height>
    </rect>
   </property>
@@ -176,7 +176,7 @@
            </widget>
           </item>
           <item>
-           <widget class="OBSBasicPreview" name="preview">
+           <widget class="OBSBasicPreview" name="preview" native="true">
             <property name="sizePolicy">
              <sizepolicy hsizetype="Expanding" vsizetype="Expanding">
               <horstretch>0</horstretch>
@@ -486,8 +486,8 @@
     <rect>
      <x>0</x>
      <y>0</y>
-     <width>1079</width>
-     <height>21</height>
+     <width>1086</width>
+     <height>22</height>
     </rect>
    </property>
    <widget class="QMenu" name="menu_File">
@@ -677,6 +677,11 @@
      <addaction name="actionSceneListMode"/>
      <addaction name="actionSceneGridMode"/>
     </widget>
+    <widget class="QMenu" name="multiviewProjectorMenu">
+     <property name="title">
+      <string>MultiviewProjector</string>
+     </property>
+    </widget>
     <action name="resetUI">
      <property name="text">
       <string>Basic.MainMenu.View.ResetUI</string>
@@ -706,11 +711,6 @@
     <addaction name="separator"/>
     <addaction name="actionAlwaysOnTop"/>
    </widget>
-   <widget class="QMenu" name="multiviewProjectorMenu">
-    <property name="title">
-     <string>MultiviewProjector</string>
-    </property>
-   </widget>
    <widget class="QMenu" name="menuTools">
     <property name="title">
      <string>Basic.MainMenu.Tools</string>
@@ -739,7 +739,7 @@
   <widget class="OBSBasicStatusBar" name="statusbar"/>
   <widget class="OBSDock" name="scenesDock">
    <property name="features">
-    <set>QDockWidget::DockWidgetClosable|QDockWidget::DockWidgetMovable|QDockWidget::DockWidgetFloatable</set>
+    <set>QDockWidget::DockWidgetClosable|QDockWidget::DockWidgetFloatable|QDockWidget::DockWidgetMovable</set>
    </property>
    <property name="windowTitle">
     <string>Basic.Main.Scenes</string>
@@ -808,9 +808,6 @@
             <verstretch>0</verstretch>
            </sizepolicy>
           </property>
-          <property name="spacing">
-           <number>0</number>
-          </property>
           <property name="contextMenuPolicy">
            <enum>Qt::CustomContextMenu</enum>
           </property>
@@ -831,6 +828,9 @@
           </property>
           <property name="defaultDropAction">
            <enum>Qt::TargetMoveAction</enum>
+          </property>
+          <property name="spacing">
+           <number>0</number>
           </property>
           <addaction name="actionRemoveScene"/>
          </widget>
@@ -879,7 +879,7 @@
   </widget>
   <widget class="OBSDock" name="sourcesDock">
    <property name="features">
-    <set>QDockWidget::DockWidgetClosable|QDockWidget::DockWidgetMovable|QDockWidget::DockWidgetFloatable</set>
+    <set>QDockWidget::DockWidgetClosable|QDockWidget::DockWidgetFloatable|QDockWidget::DockWidgetMovable</set>
    </property>
    <property name="windowTitle">
     <string>Basic.Main.Sources</string>
@@ -948,9 +948,6 @@
             <verstretch>0</verstretch>
            </sizepolicy>
           </property>
-          <property name="spacing">
-           <number>0</number>
-          </property>
           <property name="contextMenuPolicy">
            <enum>Qt::CustomContextMenu</enum>
           </property>
@@ -971,6 +968,9 @@
           </property>
           <property name="selectionMode">
            <enum>QAbstractItemView::ExtendedSelection</enum>
+          </property>
+          <property name="spacing">
+           <number>0</number>
           </property>
           <addaction name="actionRemoveSource"/>
          </widget>
@@ -1019,7 +1019,7 @@
   </widget>
   <widget class="OBSDock" name="mixerDock">
    <property name="features">
-    <set>QDockWidget::DockWidgetClosable|QDockWidget::DockWidgetMovable|QDockWidget::DockWidgetFloatable</set>
+    <set>QDockWidget::DockWidgetClosable|QDockWidget::DockWidgetFloatable|QDockWidget::DockWidgetMovable</set>
    </property>
    <property name="windowTitle">
     <string>Mixer</string>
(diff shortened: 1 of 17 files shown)


RytoEX:
What version of Qt Creator/Designer was used for this?

Looks like the macOS build didn't like the repeated use of "widget" as the name for some widgets.
image

I'm most of the way through checking the diff, but still have the two largest files to go.

derrod:
> What version of Qt Creator/Designer was used for this?

Qt Designer 6.7.0 actually since that's what I had installed, didn't seem to have any changes from 6.6.2 though.

Not sure why it did the double `widget` thing, might have to manually fix that.

UI/forms/OBSBasic.ui
```diff
@@ -168,15 +168,15 @@
...
+             <set>Qt::AlignmentFlag::AlignLeading|Qt::AlignmentFlag::AlignLeft|Qt::AlignmentFlag::AlignVCenter</set>
             </property>
             <property name="themeID" stdset="0">
              <string>previewProgramLabels</string>
             </property>
            </widget>
           </item>
           <item>
-           <widget class="OBSBasicPreview" name="preview">
+           <widget class="OBSBasicPreview" name="preview" native="true">
```
Warchamp7, line 179: Need someone on Mac to test if this causes issues. I remember a long time ago there was some horrific issue on Mac caused by Qt adding this attribute that was mitigated by removing it.

RytoEX: You might be thinking of Linux? I didn't find anything at a glance for macOS.

https://github.com/obsproject/obs-studio/commit/9ee7547e192fac25cf08124d57037cb124b69c38

UI/forms/OBSPermissions.ui
```diff
@@ -274,7 +274,7 @@
               <number>25</number>
              </property>
              <item>
-              <widget class="QWidget"/>
+              <widget class="QWidget" name="widget" native="true"/>
```
RytoEX, line 277: Why even are there blank nameless widgets here? Layout reasons? @gxalpha ?

I feel like there's a mechanism for "add an element to the second column in a layout" without having to spawn an empty widget. Or else a spacer item should work instead of a full widget? If so, that change would be out of scope for this PR, but we could change it in a separate PR.

If not, then yeah, they'll need unique names, which again, we could probably fix in a separate PR to keep the scope/intent of this PR constrained.

RytoEX (commented):
Musing thoughts while trying to move this forward.

UI/forms/OBSPermissions.ui
gxalpha: Would need to dig in to see what was going on there. I mostly took the layout from an earlier attempt at this dialog (by Warchamp and Matt) and then modified it, it’s quite possible that I copied this part and made bad modifications here.

UI/forms/OBSBasic.ui
RytoEX: cc @PatTheMav @gxalpha I don't recall if we've ever had issues with native widgets on macOS?
cc @GeorgesStavracas See above.

#3782 did previously remove `native="true"` from this widget.
https://github.com/obsproject/obs-studio/commit/9ee7547e192fac25cf08124d57037cb124b69c38#diff-c4723e7657a9d380fa8407cf136dc7e888e73120f82b0552722d45a709f7120fL170-R170

RytoEX (changes requested):
I've reviewed every file. The only requested changes are the ones in already opened conversations.

UI/forms/OBSBasic.ui
GeorgesStavracas: These native surfaces were causing trouble with Wayland back in the day. Not sure how that goes now.

RytoEX: Are we feeling adventurous to re-examine this? The way I see it, we have a few paths forward:
1. Test this on Wayland before merging.
2. Revert just these changes for now. Test the `native="true"` changes on Wayland later. Open a new PR for those later if testing goes well.
3. Merge it and be on the lookout for issues. Revert the `native="true"` changes if needed.

I would like this to not get blocked for a prolonged period of time.

UI/forms/OBSPermissions.ui
RytoEX: See #10473

RytoEX:
I break, I fix. As mentioned in #10473, `UI/forms/OBSPermissions.ui` will probably need a second pass from Qt Designer/Creator on your end to pick up the enum/namespace changes. For some reason, when I processed it in Qt Creator 12, it did not catch those.

UI/forms/OBSPermissions.ui
RytoEX: Should be resolved now.

UI/forms/OBSBasic.ui
GeorgesStavracas: My personal preference is (2) but it's more laborious than the others. If it's too much trouble, just merge with it.

RytoEX: Hmm. Thanks. Option 2 is probably "safest", but does also require _someone_ to go test on Wayland and check for issues and then PR/merge those changes at a later date, which is currently not in my capacity to do so. @derrod ?

RytoEX:
> I break, I fix. Merging #10473 first caused a small merge conflict.
> 
> As mentioned in #10473, `UI/forms/OBSPermissions.ui` will probably need a second pass from Qt Designer/Creator on your end to pick up the enum/namespace changes. For some reason, when I processed it in Qt Creator 12, it did not catch those.

We discussed this off-thread and discovered the differences were down to different versions of Qt Designer (`bin\designer.exe`) included with different versions of Qt. Qt 6.7.0's designer.exe produces the enum differences seen here. This is due to changes in Qt 6.7.0, but these changes _should_ be backwards compatible, and CI seems able to compile with these changes, all the way back to Qt 6.2.4 on Ubuntu 22.04.
https://bugreports.qt.io/browse/QTBUG-118473

We'll still need to decide on the native widgets issue:
https://github.com/obsproject/obs-studio/pull/10419#discussion_r1538215097

UI/forms/OBSBasic.ui
tytan652: Looks fine on GNOME Wayland after a quick test resizing arround with Studio mode on and off.

RytoEX:
Final blockers:
1. Choose which version of designer.exe or Creator to use for formatting.
2. Decide on the wayland native widgets problem.

The main challenge with Creator is that it's not delivered in-sync with Qt versions. Qt Creator 13 based on Qt 6.6.0 was released on April 3, 2024, but Qt 6.7.0 was released on April 2, 2024. There's no guarantee that the version of Creator you launch is using the version of Qt (designer) you have installed, as far as I can tell.

UI/forms/OBSBasic.ui
RytoEX: Sounds like we'll give this a scream test then.

RytoEX:
> Final blockers:
>     1. Choose which version of designer.exe or Creator to use for formatting.
>     2. Decide on the wayland native widgets problem.
> 
> The main challenge with Creator is that it's not delivered in-sync with Qt versions. Qt Creator 13 based on Qt 6.6.0 was released on April 3, 2024, but Qt 6.7.0 was released on April 2, 2024. There's no guarantee that the version of Creator you launch is using the version of Qt (designer) you have installed, as far as I can tell.

I think:
1. We should use the version of designer.exe that matches the version of Qt we ship on Windows and macOS. Right now that is 6.6.2. When we update those platforms, we can update the .ui files in tandem. Unless @Warchamp7 has a different opinion?
2. We're going to go with these changes. We'll do a scream test and be vigilant for reports of breakage.

Warchamp7:
> > Final blockers:
> > 1. Choose which version of designer.exe or Creator to use for formatting.
> > 2. Decide on the wayland native widgets problem.
> > The main challenge with Creator is that it's not delivered in-sync with Qt versions. Qt Creator 13 based on Qt 6.6.0 was released on April 3, 2024, but Qt 6.7.0 was released on April 2, 2024. There's no guarantee that the version of Creator you launch is using the version of Qt (designer) you have installed, as far as I can tell.
> 
> I think:
> 
> 1. We should use the version of designer.exe that matches the version of Qt we ship on Windows and macOS. Right now that is 6.6.2. When we update those platforms, we can update the .ui files in tandem. Unless @Warchamp7 has a different opinion?
> 2. We're going to go with these changes. We'll do a scream test and be vigilant for reports of breakage.

Stamp of approval from me.

derrod:
> 1. We should use the version of designer.exe that matches the version of Qt we ship on Windows and macOS. Right now that is 6.6.2. When we update those platforms, we can update the .ui files in tandem. Unless @Warchamp7 has a different opinion?

Could just delay until the update happens, otherwise we'd have two commits that touch many `.ui` files that make blaming even more difficult.

RytoEX:
> > 1. We should use the version of designer.exe that matches the version of Qt we ship on Windows and macOS. Right now that is 6.6.2. When we update those platforms, we can update the .ui files in tandem. Unless @Warchamp7 has a different opinion?
> 
> Could just delay until the update happens, otherwise we'd have two commits that touch many `.ui` files that make blaming even more difficult.

Sure. I'll try to spin dependencies updates soon.

RytoEX:
Per off-thread discussion with @Warchamp7 , we've re-run the .ui files through Qt 6.6.2's designer.exe for now to unblock this and fix the worst issues. Finishing up some local test builds.

RytoEX (approved):
Re-tested this locally on Windows 11 and it seems fine. This no longer includes the Qt 6.7 changes for fully-qualified enums, which is supposedly only compatible back to Qt 6.5.

PR #10426, opened 2024-03-24
UI: Fix scene rename on tab

### LINKED ISSUE #10090: Names of scenes in the Scenes list display incorrectly when hitting Tab while renaming
### Operating System Info

Windows 10

### Other OS

_No response_

### OBS Studio Version

30.0.2

_No response_

### OBS Studio Log URL

https://obsproject.com/logs/8K7EVTJOjWaigarf

### OBS Studio Crash Log URL

_No response_

### Expected Behavior

While hitting the Tab key after renaming a Scene but not yet having hit Enter or otherwise confirming the change, I expected that either:
- Nothing would happen
- The rename would be confirmed and the next scene would open
- The rename would be confirmed and the next scene would open and prompt me to rename it

### Current Behavior

Upon hitting tab, the next scene (I'll call that B) would be selected and the name would appear on both the scene I was just trying to rename (I'll call that A) as well as B. However, navigating to another scene which included the scenes as sources would show the names differently - A's name would be unchanged, and the rename that I was typing in for A would be applied to B instead. Hitting Ctrl + Z to undo my change (OR renaming B back to its original name) appeared to undo the change as expected when viewing the names of the scenes on my list of sources, but on the list of scenes it would appear to show the original name of B on both scenes.

Restarting OBS after all of this fixes the issue, and the names in the Scenes list now match the names in the Sources list.

The images below are chronologically ordered:

Original order of scenes
image

Typing in new name for desired scene
image

Scene list after hitting Tab
image

Scene list and list of scenes as sources in another scene, after hitting Tab and navigating to the scene containing all the scenes as sources **(The scenes from "[O] 1 Casters Text" to "[O] 4 Casters Text 2" are ordered the same in the sources list as they are in the scene list)**
image

Scene list and list of scenes as sources in another scene, after undoing the rename
image

### Steps to Reproduce

1. Right-click a scene in the Scenes list, and select Rename
2. Type in a new name for the scene
3. Press Tab

This is baby's first bug report so I apologize if this is over-worded, especially for how minor of an issue it is.

howellrl:
### Description
Tab key press event was triggering default Qt behavior and selecting next item in scene list before rename event could complete. This caused the incorrect scene to become the target of the rename method. Consumed the tab key event and observed that tab now completes the edit and leaves the current scene selected. This is in alignment with how tabbing away from a rename affects renaming of sources.

### Motivation and Context
Fixes #10090

### How Has This Been Tested?
Built and ran locally on Windows 11 machine before code change, observed incorrect behavior tabbing from rename. Modified code, observed tabbing from rename completes edit as expected.

### Types of changes

### Checklist:

### DIFF
--- a/UI/window-basic-main.cpp
+++ b/UI/window-basic-main.cpp
@@ -10662,10 +10662,16 @@ bool SceneRenameDelegate::eventFilter(QObject *editor, QEvent *event)
 {
 	if (event->type() == QEvent::KeyPress) {
 		QKeyEvent *keyEvent = static_cast<QKeyEvent *>(event);
-		if (keyEvent->key() == Qt::Key_Escape) {
+		switch (keyEvent->key()) {
+		case Qt::Key_Escape: {
 			QLineEdit *lineEdit = qobject_cast<QLineEdit *>(editor);
 			if (lineEdit)
 				lineEdit->undo();
+			break;
+		}
+		case Qt::Key_Tab:
+		case Qt::Key_Backtab:
+			return false;
 		}
 	}
 


Warchamp7:
This fixes hitting Tab but the same error still happens when pressing Shift + Tab

PatTheMav:
What happens on macOS without this fix is that the scene is renamed, _but also the scene being switched to_. So both scenes will have the changed name of the first scene.

With only a single scene the renaming is successful by just pressing `<TAB>`.

Pressing `<SHIFT>+<TAB>` only finishes renaming the current scene.

Warchamp7:
> What happens on macOS without this fix is that the scene is renamed, _but also the scene being switched to_. So both scenes will have the changed name of the first scene.
> 
> With only a single scene the renaming is successful by just pressing `<TAB>`.
> 
> Pressing `<SHIFT>+<TAB>` only finishes renaming the current scene.

With 3 scenes and the middle scene in the list selected, Shift + Tab doesn't cause the double rename on the current scene and the one above it?

PatTheMav:
> > What happens on macOS without this fix is that the scene is renamed, _but also the scene being switched to_. So both scenes will have the changed name of the first scene.
> > With only a single scene the renaming is successful by just pressing `<TAB>`.
> > Pressing `<SHIFT>+<TAB>` only finishes renaming the current scene.
> 
> With 3 scenes and the middle scene in the list selected, Shift + Tab doesn't cause the double rename on the current scene and the one above it?

My bad, I didn't consider that `<SHIFT>+<TAB>` would actually move the selection up - the issue occurs as you expected. Interestingly renaming the bottom source again (and naming it "Scene 3" again) triggers a "Name already in use" error.

RytoEX:
> This fixes hitting Tab but the same error still happens when pressing Shift + Tab
> 
> Will need to check for any similar issues on Mac @PatTheMav

I'm actually surprised at that. I would expect `keyEvent->key()` to always be `Qt::Key_Tab` regardless of additional modifiers. Would probably need to dump out all the keypress events during testing to see what's happening.

That said, I think it's more likely that users are hitting Tab than Shift+Tab, so I'd be inclined to accept this fix (after my own review/testing) if it fixes one known bad case, and we can return to fix other known bad cases later.

howellrl:
I was able to take a look at this and I noticed there is a separate key press event for back tab that handles the shift + tab key press combination. It appears to handle the back tab correctly now. I have updated my PR.

RytoEX (changes requested):
Could you squash the commits into a single commit and place the cases in this order:
```cpp
case Qt::Key_Escape
case Qt::Key_Tab
case Qt::Key_Backtab
```

howellrl:
Absolutely - I have squashed the commits and re-arranged the case statements.

RytoEX (approved):
Tested on Windows 11. Seems to fix the buggy behavior in the linked Issue.

RytoEX:
In the future, you can use GitHub closing keywords to automatically link a PR to an Issue:
* https://docs.github.com/en/issues/tracking-your-work-with-issues/linking-a-pull-request-to-an-issue#linking-a-pull-request-to-an-issue-using-a-keyword
* https://docs.github.com/en/get-started/writing-on-github/working-with-advanced-formatting/using-keywords-in-issues-and-pull-requests#linking-a-pull-request-to-an-issue

Waiting on retests from @Warchamp7 and @PatTheMav , or for their review requests to be removed.

Warchamp7 (approved):

Lain-B (approved):

PR #10434, opened 2024-03-26
text-freetype2: Fix bug with null characters

lourencop01:
When uploading a file containing null characters only the characters before the first null character would be rendered. Fixed by sanitizing the null characters from the bytes read from the uploaded file.

### Description

Sanitize null characters from the bytes read from an uploaded file. 

### Motivation and Context

This change fixes the issue reported on: [obs-studio/issues#7595](https://github.com/obsproject/obs-studio/issues/7595), where when a user uploaded a file containing null characters, only the characters prior to the first null character would be rendered. With this fix, the correct text from the file is rendered.

### How Has This Been Tested?

The changes were tested by uploading a file containing null characters.
The testing environment is macOS Sonoma 14.3.1

### Types of changes

### Checklist:

### DIFF
--- a/plugins/text-freetype2/text-functionality.c
+++ b/plugins/text-freetype2/text-functionality.c
@@ -424,6 +424,8 @@ void load_text_from_file(struct ft2_source *srcdata, const char *filename)
 	FILE *tmp_file = NULL;
 	uint32_t filesize = 0;
 	char *tmp_read = NULL;
+	char *sanitized = NULL;
+	size_t sanitized_length = 0;
 	uint16_t header = 0;
 	size_t bytes_read;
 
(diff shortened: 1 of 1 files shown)


kkartaltepe:
This seems like it would work, but I'm not sure we are keen on committing a workaround for this issue other than just fixing string handling throughout.

lourencop01:
Got it, no worries!

Since you're not looking for either displaying null characters as glyphs or sanitizing them, could you please clarify what the ideal solution would be? 

Thank you!

lourencop01:
Hey, just following up on this last message to bump it up on your notifications box! :)

kkartaltepe:
https://github.com/obsproject/obs-studio/issues/7595#issuecomment-1977884229 covers this. Specifically we dont need to change null bytes or use any special representation of them to fix the issue, just to manage the length of strings correctly.

PR #10436, opened 2024-03-26
obs-outputs: Fix AV1 packet priority in RTMP

### LINKED ISSUE #10383: AV1 YouTube bitrate drop until stream restart
### Operating System Info

Windows 11

### Other OS

_No response_

### OBS Studio Version

30.1.0

_No response_

### OBS Studio Log URL

https://obsproject.com/logs/VJcmhy0uy4lERRpI

### OBS Studio Crash Log URL

_No response_

### Expected Behavior

A stable AV1 stream to YouTube with a fairly beefy dedicated streaming PC.

### Current Behavior

Nvidia AV1 stream to YouTube randomly drops bitrate to ~160kbps and stays there until stream is restarted.

### Steps to Reproduce

1.
2.
3.
...

I have been trying to stream with Nvidia AV1 for about a year now (since AV1 OBS release) and tried every new update. I never have similar issues with other encoders, so I dont think it has anything to do with my ISP. Im afraid the log doesnt show anything except :
**13:31:45.549: Output 'adv_stream': Number of dropped frames due to insufficient bandwidth/connection stalls: 6283 (1.0%)**
But let me repeat, my internet runs just fine and this only happens with AV1.

**During my stream (usually 16000 kbps), the bitrate drops to ~160kbps and stays there until I restart the stream.** Nothing seems to trigger this unfortunately and it happens at random time intervals (sometimes runs fine for days, other times drops every 10 minutes)

This issue seems to be random, but I think it happens more during weekends and busy hours. (So possibly YT servers?)

I have been searching for other users with similar problems and I found a lot of them, but noone seems to have a clue about why this could be happening: https://obsproject.com/forum/threads/av1-live-streaming-on-youtube-not-working-please-help.168513/

derrod:
### Description

Adds parsing of AV1 frame type to AV1 OBU parsing to set drop priority of keyframes correctly.

Somewhat janky but matches what we did in 722ca8447e3be1ad4cd10ba1e8c8dd41e341a20f and should be good enough™ for our use case.

### Motivation and Context

Should fix #10383

### How Has This Been Tested?

Verified priority now gets set.

### Types of changes

### Checklist:

### DIFF
--- a/plugins/obs-outputs/rtmp-av1.c
+++ b/plugins/obs-outputs/rtmp-av1.c
@@ -543,12 +543,6 @@ static void serialize_av1_data(struct serializer *s, const uint8_t *data,
 	(void)priority;
 	uint8_t *buf = (uint8_t *)data;
 	uint8_t *end = (uint8_t *)data + size;
-	enum {
-		START_NOT_FOUND,
-		START_FOUND,
-		END_FOUND,
-		OFFSET_IMPOSSIBLE,
-	} state = START_NOT_FOUND;
 
 	while (buf < end) {
 		int64_t obu_size;
@@ -564,15 +558,8 @@ static void serialize_av1_data(struct serializer *s, const uint8_t *data,
 		case AV1_OBU_TEMPORAL_DELIMITER:
 		case AV1_OBU_REDUNDANT_FRAME_HEADER:
 		case AV1_OBU_TILE_LIST:
-			if (state == START_FOUND)
-				state = END_FOUND;
 			break;
 		default:
-			if (state == START_NOT_FOUND) {
-				state = START_FOUND;
-			} else if (state == END_FOUND) {
-				state = OFFSET_IMPOSSIBLE;
-			}
 			s_write(s, buf, len);
 			size += len;
 			break;
--- a/plugins/obs-outputs/rtmp-av1.c
+++ b/plugins/obs-outputs/rtmp-av1.c
@@ -33,10 +33,18 @@
 #define AV1_OBU_TILE_GROUP 4
 #define AV1_OBU_TILE_LIST 8
 #define AV1_OBU_FRAME 6
+#define AV1_OBU_FRAME_HEADER 3
 
 #define FF_PROFILE_AV1_MAIN 0
 #define FF_PROFILE_AV1_HIGH 1
 #define FF_PROFILE_AV1_PROFESSIONAL 2
+
+enum frame_type {
+	AV1_KEY_FRAME,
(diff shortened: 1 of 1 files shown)


derrod:
Adjusted the priorities based on feedback from people that know more than me about AV1. Should hopefully be good enough to keep us running for a while :)

plugins/obs-outputs/rtmp-av1.c
```diff
@@ -536,19 +544,47 @@ size_t obs_parse_av1_header(uint8_t **header, const uint8_t *data, size_t size)
 	return output.bytes.num;
 }
 
+static void compute_av1_keyframe_priority(const uint8_t *buf, bool *is_keyframe,
+					  int *priority)
+{
+	/* Skip if the packet already has a priority, e.g. assigned by the
```
RytoEX, line 550:
```suggestion
	/* Skip if the packet already has a priority, e.g., assigned by the
```

RytoEX (changes requested):
Nit only, otherwise seems fine.

norihiro:
The prefixes of the commit titles should be `obs-outputs` instead of `obs-output`.

plugins/obs-outputs/rtmp-av1.c
```diff
@@ -564,15 +600,15 @@ static void serialize_av1_data(struct serializer *s, const uint8_t *data,
...
-			if (state == START_FOUND)
-				state = END_FOUND;
 			break;
+		case AV1_OBU_FRAME:
+		case AV1_OBU_FRAME_HEADER:
+			compute_av1_keyframe_priority(buf + start_pos,
+						      is_keyframe, priority);
+#ifdef __GNUC__
+			__attribute__((fallthrough));
+#endif
```
RytoEX, line 610: It seems we normally use this instead:
```c
/* Falls through. */
```
or
```c
/* FALLTHROUGH */
```

RytoEX (changes requested):

plugins/obs-outputs/rtmp-av1.c
derrod: If we ever get C23 we can do `[[fallthrough]]` but until then it's either a comment or this. I don't think it matters.

RytoEX: It doesn't matter too much, I was just looking for consistency because I didn't want to have to try to find all the different variants we used along the way if/when we switch to C23.

RytoEX (approved):
Looks okay to me. User tests were promising.

Lain-B (approved):

PR #10439, opened 2024-03-27
rtmp-services: Update "CHZZK" platform

wly65535:
### Description
Update "recommended" settings for "CHZZK" platform

### Motivation and Context
Update "recommended" settings for "CHZZK" platform

### How Has This Been Tested?
Local Build & CI Buil

### Types of changes

### Checklist:

### DIFF
--- a/plugins/rtmp-services/data/package.json
+++ b/plugins/rtmp-services/data/package.json
@@ -1,11 +1,11 @@
 {
     "$schema": "schema/package-schema.json",
     "url": "https://obsproject.com/obs2_update/rtmp-services/v5",
-    "version": 248,
+    "version": 249,
     "files": [
         {
             "name": "services.json",
-            "version": 248
+            "version": 249
         }
     ]
 }
--- a/plugins/rtmp-services/data/services.json
+++ b/plugins/rtmp-services/data/services.json
@@ -2706,8 +2706,14 @@
             ],
             "recommended": {
                 "keyint": 1,
-                "profile": "high",
-                "bframes": 0
+                "bframes": 0,
+                "max fps": 60,
+                "max video bitrate": 8000,
+                "max audio bitrate": 192,
+                "x264opts": "tune=zerolatency scenecut=0",
+                "supported resolutions": [
+                    "1920x1080"
+                ]
             }
         },
         {


plugins/rtmp-services/data/services.json
```diff
@@ -2706,8 +2706,14 @@
             ],
             "recommended": {
                 "keyint": 1,
-                "profile": "high",
-                "bframes": 0
+                "bframes": 0,
+                "max fps": 60,
+                "max video bitrate": 8000,
+                "max audio bitrate": 192,
+                "x264opts": "tune=zerolatency scenecut=0",
```
gxalpha, line 2713: Are you *really really really* sure you want to force `tune=zerolatency` onto your users?
It significantly reduces video quality while not reducing the overall stream latency by that much (considering you're still streaming via RTMP).

wly65535: Thanks for your message.
We will consider more detailly and change the setting later.

walker-WSH: hi @gxalpha,  thanks for your advice.   We have talked about the quality risk with "zerolatency".  But for CHZZK users, low latency is much more important.  So we decide to use it, thanks a lot~

walker-WSH:
> By choosing 1080p as a supported resolution, users will explicitly have to select "Ignore recommendations" to choose anything other than 1080p. Is that OK?

Hi @WizardCM， yes, that is OK.
for chzzk,  we suggest users use 1080P instead of other resolutions.

RytoEX:
> > By choosing 1080p as a supported resolution, users will explicitly have to select "Ignore recommendations" to choose anything other than 1080p. Is that OK?
> 
> Hi @WizardCM， yes, that is OK. for chzzk, we suggest users use 1080P instead of other resolutions.

I believe the concern is that by only having 1080p in "supported resolutions", users will _not be able to stream_ with any other resolution (I do not recall if their stream resolution is automatically changed to the nearest "supported resolution" or if they get a warning) unless they check "Ignore streaming service setting recommendations", at which point none of your other recommended settings will apply anyway. Perhaps @WizardCM remembers, but you can probably test the behavior and UX yourself to confirm if it's what you actually want.

walker-WSH:
@RytoEX   We will test the limit you said and response our final decision here

thanks a lot

wly65535:
> > > By choosing 1080p as a supported resolution, users will explicitly have to select "Ignore recommendations" to choose anything other than 1080p. Is that OK?
> > 
> > 
> > Hi @WizardCM， yes, that is OK. for chzzk, we suggest users use 1080P instead of other resolutions.
> 
> I believe the concern is that by only having 1080p in "supported resolutions", users will _not be able to stream_ with any other resolution (I do not recall if their stream resolution is automatically changed to the nearest "supported resolution" or if they get a warning) unless they check "Ignore streaming service setting recommendations", at which point none of your other recommended settings will apply anyway. Perhaps @WizardCM remembers, but you can probably test the behavior and UX yourself to confirm if it's what you actually want.

Thanks for your advice.

Actually, CHZZK also supports other resolution.
For best user experience, We suggest users use 1080P.
If user checked "Ignore streaming service setting recommendations" to select other resolution. It's also no problem.

Fenrirthviti (approved):
Approved from an eligibility standpoint, seems fine.

Given above comments, this appears to be acceptable limitations to CHZZK, and they seem prepared to deal with user education so I have no objections.

plugins/rtmp-services/data/package.json
```diff
@@ -1,11 +1,11 @@
 {
     "$schema": "schema/package-schema.json",
     "url": "https://obsproject.com/obs2_update/rtmp-services/v5",
-    "version": 247,
+    "version": 248,
     "files": [
         {
             "name": "services.json",
-            "version": 247
+            "version": 248
```
RytoEX, line 8: Update to account for the latest changes in master branch.

RytoEX (changes requested):

plugins/rtmp-services/data/package.json
wly65535: Updated

plugins/rtmp-services/data/services.json
RytoEX: We explained the tradeoffs, and they have insisted. Resolving.

RytoEX (approved):

PR #10290, opened 2024-02-26
CI: Fix scheduled language file check running on same commit hash

PatTheMav:
### Description
Fix warnings in scheduled nightly workflow runs when master branch has not been updated between runs.

### Motivation and Context
When a prior scheduled nightly run is detected with the identical commit hash, then the entire language file check should be skipped.

This adds the missing condition to skip the check for changed files removing the unnecessary warning contained in the workflow results.

### How Has This Been Tested?
Needs to be checked on CI.

### Types of changes
- Bug fix (non-breaking change which fixes an issue)

### Checklist:
- [x] My code has been run through [clang-format](https://github.com/obsproject/obs-studio/blob/master/.clang-format).
- [x] I have read the [**contributing** document](https://github.com/obsproject/obs-studio/blob/master/CONTRIBUTING.rst).
- [x] My code is not on the master branch.
- [x] The code has been tested.
- [x] All commit messages are properly formatted and commits squashed where appropriate.
- [x] I have included updates to all appropriate documentation.

PR #10444, opened 2024-03-27
CI: Fix JSON runtime error for language file uploads

PatTheMav:
### Description
Removes condition from changed file check for nightly language file upload to fix JSON parsing error.

### Motivation and Context
The upload step in the job uses the output of two prior steps as its condition. One of the jobs itself is a conditional job, so if the primary condition is not met, the secondary output is not set and the fromJSON call will fail.

Making the check for changed files unconditional ensures that both outputs provide a value that can be used with fromJSON and no syntax error occurs.

Language uploads still only happen if the commit hash of the nightly run has actually changed and any language files are actually changed.

### How Has This Been Tested?
Requires nightly run to test.

### Types of changes

### Checklist:

### DIFF
--- a/.github/workflows/scheduled.yaml
+++ b/.github/workflows/scheduled.yaml
@@ -124,7 +124,7 @@ jobs:
           checkGlob: '**/en-US.ini'
 
       - name: Upload US English Language Files 🇺🇸
-        if: fromJSON(steps.checks.outputs.hasChangedFiles) && fromJSON(steps.nightly-checks.outputs.passed)
+        if: fromJSON(steps.nightly-checks.outputs.passed) && steps.checks.outputs.passed == 'true'
         uses: obsproject/obs-crowdin-sync/upload@30b5446e3b5eb19595aa68a81ddf896a857302cf
         env:
           CROWDIN_PAT: ${{ secrets.CROWDIN_SYNC_CROWDIN_PAT }}


RytoEX:
This essentially reverts #10290 , correct?

PatTheMav:
> This essentially reverts #10290 , correct?

Indeed, I might need to come up with a different solution...

RytoEX:
> > This essentially reverts #10290 , correct?
> 
> Indeed, I might need to come up with a different solution...

A different solution than this, or do we still want to merge this for now?

PatTheMav:
> > > This essentially reverts #10290 , correct?
> > 
> > 
> > Indeed, I might need to come up with a different solution...
> 
> A different solution than this, or do we still want to merge this for now?

I replaced it with a string comparison now which should be syntactically correct.

.github/workflows/scheduled.yaml
```diff
@@ -124,7 +124,7 @@ jobs:
           checkGlob: '**/en-US.ini'
 
       - name: Upload US English Language Files 🇺🇸
-        if: fromJSON(steps.checks.outputs.hasChangedFiles) && fromJSON(steps.nightly-checks.outputs.passed)
+        if: fromJSON(steps.nightly-checks.outputs.passed) && steps.checks.outputs.passed == 'true'
```
RytoEX, line 127: If `checks` step is conditional on the `nightly-checks` step returning `true`, then isn't the first conditional superfluous?

Or does this:
```yaml
if: fromJSON(steps.nightly-checks.outputs.passed)
```
actually succeed if there is _any_ output returned by the `fromJSON` call?

RytoEX (changes requested):

.github/workflows/scheduled.yaml
PatTheMav: It's not necessary as the check for changed files will not yield "true" otherwise, but it makes it more explicit that both rely on this condition.

RytoEX: Got it.

RytoEX (approved):
Seems fine.

Issue #7360, opened 2022-09-12
Letter 'a' is missing at the end


maisondasilva:
### Operating System Info

Windows 11

### OBS Studio Version

30.0.2

### OBS Studio Log URL

null

### Expected Behavior

The translation is correct, but for some reason the letter 'a' is missing at the end

![Captura de tela 2022-09-12 121111](https://user-images.githubusercontent.com/11423362/189690704-6006c9e0-4537-46cb-9747-bbaab1ba65ce.png)

### Current Behavior

![Captura de tela 2022-09-12 121111](https://user-images.githubusercontent.com/11423362/189690504-2cde8f5b-8063-419e-8350-d21e76ea933f.png)

![Captura de tela 2022-09-12 121111](https://user-images.githubusercontent.com/11423362/189692067-0eee3984-6012-4c06-b4ea-7d79a31e54c9.png)

### Steps to Reproduce

1. open obs in pt-br

### Anything else we should know?

No

maisondasilva:
Just to let you know I still see in the new version 28.1.0

Thanks

RytoEX:
Please include an OBS log to give us some information about your displays. Do your displays have a scaling factor set (Display Settings -> Scale and layout)?

For those looking at this, this is almost certainly a display scaling issue in combination with our UI padding/layout and not a code/translation issue.

On my 14" 1080p display at 125% scaling, this is what I see.
![image](https://user-images.githubusercontent.com/624931/199273830-4c2e0d0b-2461-4b64-8be9-c9e3fbf5b4fd.png)

maisondasilva:
@RytoEX I noticed that the problem is in the theme file, could you tell me which variables are used in the theme in this part, I can solve it by changing the theme file and then sending the correction here

29.0.1 have same problem in theme Yami

RytoEX:
> @RytoEX I noticed that the problem is in the theme file, could you tell me which variables are used in the theme in this part, I can solve it by changing the theme file and then sending the correction here
> 
> 29.0.1 have same problem in theme Yami

As I previously mentioned:
> For those looking at this, this is almost certainly a display scaling issue in combination with our UI padding/layout and not a code/translation issue.

We know this is an issue in the theme that is exacerbated by display scaling. The challenge is accommodating _all_ possible translation strings and display scaling combinations without taking up too much space, particularly on small displays.

maisondasilva:
Could you tell me which line in the theme so I can test it in several languages and resolutions?
[...]

PR #10456, opened 2024-03-29
UI: When no source is selected, the text indicating this appears cut off in some languages

Diog0o:
### Description
In some languages as Portuguese (PT), Portuguese-Brasil (PT-BR) and Spanish (ES) the last few words dind't appear because they were cut off. The solution was to increase the maximumSize width of the "contextSourceLabel" from 160 to 200. I tested in every single language and it's working fine.

### Motivation and Context
This change fixes the issue reported on: https://github.com/obsproject/obs-studio/issues/7360, where when so source was selected, the text suggesting that appeared cut off in some languages including: Portuguese and Spanish.

### How Has This Been Tested?
The testing environment was Windows 11, replicating the observed behavior in all languages.

### Types of changes

### Checklist:

### DIFF
--- a/UI/forms/OBSBasic.ui
+++ b/UI/forms/OBSBasic.ui
@@ -308,7 +308,7 @@
             </property>
             <property name="maximumSize">
              <size>
-              <width>160</width>
+              <width>200</width>
               <height>22</height>
              </size>
             </property>


norihiro:
The commit message should be wrapped to 72 columns.

Diog0o:
> The commit message should be wrapped to 72 columns.

Done!

Warchamp7:
We're going to discuss alternative ways to fix this, but this isn't the approach we'd like to take at this time.

maisondasilva:
I still see this bug in the new version of the theme, and in the old theme this problem does not occur, there is a faster way to resolve this, as it is a visual bug that has been around for almost 2 years!

image

https://github.com/obsproject/obs-studio/issues/7360

maisondasilva:
I still see this bug in the new version of the theme, and in the old theme this problem does not occur, there is a faster way to resolve this, as it is a visual bug that has been around for almost 2 years!

image

@Warchamp7 
https://github.com/obsproject/obs-studio/issues/7360

PR #10458, opened 2024-03-29
mac-capture: Replace pragmas with availability markers

tytan652:
### Description

Replace pragmas silencing `-Wunguarded-availability-new` by availlability markers.

Note: When placing the marker before a function declaration, clang-format can add an empty new line between the marker and the declaration. So markers were put after when it comes to function declaration.

### Motivation and Context

I tried to unvendor simde and discover that `mac-sdk-common.h` pragmas somehow contaminated other `mac-sck-*.mm` files because of our vendored simde.

Simply moving the pragmas like the following allows to fix the contamination:
```diff
diff --git a/plugins/mac-capture/mac-sck-common.h b/plugins/mac-capture/mac-sck-common.h
index 32ca8d246a8a5..e8b1911f6164d 100644
--- a/plugins/mac-capture/mac-sck-common.h
+++ b/plugins/mac-capture/mac-sck-common.h
@@ -1,9 +1,6 @@
 #include <AvailabilityMacros.h>
 #include <Cocoa/Cocoa.h>
 
-#pragma clang diagnostic push
-#pragma clang diagnostic ignored "-Wunguarded-availability-new"
-
 #include <stdlib.h>
 #include <obs-module.h>
 #include <util/threading.h>
@@ -14,6 +11,9 @@
 #include <CoreMedia/CMSampleBuffer.h>
 #include <CoreVideo/CVPixelBuffer.h>
 
+#pragma clang diagnostic push
+#pragma clang diagnostic ignored "-Wunguarded-availability-new"
+
 #define MACCAP_LOG(level, msg, ...) blog(level, "[ mac-screencapture ]: " msg, ##__VA_ARGS__)
 #define MACCAP_ERR(msg, ...)        MACCAP_LOG(LOG_ERROR, msg, ##__VA_ARGS__)
```

Rather than fixing new warnings by adding more pragmas, I tried to find a better way.
And I found `API_AVAILABLE()` and based on [Clang documentation](https://clang.llvm.org/docs/LanguageExtensions.html#objective-c-available):
> If the caller of `my_fun()` already checks that `my_fun()` is only called on 10.12, then add an availability attribute to it, which will also suppress the warning and require that calls to my_fun() are checked

And it seems to work for most of declaration not only functions.

### How Has This Been Tested?

Building yes, macOS SCK was not tested.

### Types of changes

### Checklist:

### DIFF
--- a/.clang-format
+++ b/.clang-format
@@ -181,7 +181,7 @@ ReferenceAlignment: Right
 RemoveSemicolon: false
 RequiresClausePosition: WithPreceding
 RequiresExpressionIndentation: OuterScope
-SeparateDefinitionBlocks: Always
+SeparateDefinitionBlocks: Leave
 ShortNamespaceLines: 1
 SortIncludes: false
 #SortUsingDeclarations: LexicographicNumeric
--- a/plugins/mac-capture/mac-sck-audio-capture.m
+++ b/plugins/mac-capture/mac-sck-audio-capture.m
@@ -5,7 +5,7 @@ const char *sck_audio_capture_getname(void *unused __unused)
     return obs_module_text("SCK.Audio.Name");
 }
 
-static void destroy_audio_screen_stream(struct screen_capture *sc)
+API_AVAILABLE(macos(13.0)) static void destroy_audio_screen_stream(struct screen_capture *sc)
 {
     if (sc->disp && !sc->capture_failed) {
         [sc->disp stopCaptureWithCompletionHandler:^(NSError *_Nullable error) {
@@ -32,7 +32,7 @@ static void destroy_audio_screen_stream(struct screen_capture *sc)
     os_event_destroy(sc->stream_start_completed);
 }
 
-static void sck_audio_capture_destroy(void *data)
+API_AVAILABLE(macos(13.0)) static void sck_audio_capture_destroy(void *data)
 {
     struct screen_capture *sc = data;
 
@@ -57,7 +57,7 @@ static void sck_audio_capture_destroy(void *data)
     bfree(sc);
 }
 
-static bool init_audio_screen_stream(struct screen_capture *sc)
+API_AVAILABLE(macos(13.0)) static bool init_audio_screen_stream(struct screen_capture *sc)
 {
     SCContentFilter *content_filter;
     if (sc->capture_failed) {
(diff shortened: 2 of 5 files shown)


PatTheMav (commented):
I'm not sure this is the right approach for all the instances included in this PR:

The markers are supposed to be used for API functions to mark their availability when _implementing_ a Framework or external library which provides an API (the hint is in the name `API_AVAILABLE`).

These markers mainly belong in headers that make functionality available either to other projects (exported API) or within the same project but for other compilation units (internal API).

Using them in source files to decorate static symbols does not seem like the intended use case to me: Static symbols are "private" to the compilation unit and mainly used for internal support functions or variables. A developer should always be aware of their restrictions/requirements anyway.

Typical example for markers are macOS framework headers (which is what the feature was implemented for):

```
- (BOOL)containsString:(NSString *)str API_AVAILABLE(macos(10.10), ios(8.0), watchos(2.0), tvos(9.0));
- (BOOL)localizedCaseInsensitiveContainsString:(NSString *)str API_AVAILABLE(macos(10.10), ios(8.0), watchos(2.0), tvos(9.0));

```

Static functions instead need to use `@available()` or `__builtin_available()` to avoid codepaths using unavailable APIs on older versions. The changes in the header files seem to follow convention though and seem fine.

tytan652:
The whole SCK part is behind one or two availability check.
Adding checks all over already availability-checked code would make no sense.

PatTheMav:
> The whole SCK part is behind one or two availability check. Adding checks all over already availability-checked code would make no sense.

Yeah looks like the availability attribute is necessary after all because the requirements leak across types.

Weirdly enough, when written out as `__attribute__((availability(macos, introduced=13.0)))`, clang-format keeps it in front of a function, it's only when provided as the helper macro that it does this weird indentation - maybe this can be reigned in with the correct clang-format option..

PatTheMav:
Setting `SeparateDefinitionBlocks: Leave` (instead of `Always`) disables the formatting behaviour and allows placing the macro before or just above the function declaration (which is how Xcode automatically fixes the warnings).

I'm fine with setting it to `Leave` for ObjC in clang-format for this PR.

tytan652:
Like that @PatTheMav ? I'll squash if it's good.

PatTheMav:
Yeah I think it's fine - Xcode usually puts it on its own line above the function/type, but I dunno if `clang-format` moves it onto the same line again now.

tytan652:
`clang-format` moved some of them to the line above when I formatted, those were on the same line as the function at before.

tytan652:
> but I dunno if `clang-format` moves it onto the same line again now.

No, it doesn't.

PatTheMav (approved):

PR #10459, opened 2024-03-29
linux-pipewire: Use dynamic pod builder

GeorgesStavracas:
### Description

Use the dynamic pod builder, which resizes the pod buffer as needed, instead of a static 4096 buffer which can overflow.

### Motivation and Context

The dynamic pod builder implements a dynamically growing buffer for the pods, which can indeed grow quite a lot depending on e.g. the number of modifiers and formats that the GPU advertises.

### How Has This Been Tested?

- Add any PipeWire-based source (window or monitor captures, or the PipeWire camera)
- Notice they still work

### Types of changes

### Checklist:

### DIFF
--- a/plugins/linux-pipewire/pipewire.c
+++ b/plugins/linux-pipewire/pipewire.c
@@ -36,6 +36,7 @@
 #include <spa/debug/format.h>
 #include <spa/debug/types.h>
 #include <spa/param/video/type-info.h>
+#include <spa/pod/dynamic.h>
 #include <spa/utils/result.h>
 
 //#define DEBUG_PIPEWIRE
@@ -299,10 +300,10 @@ static void swap_texture_red_blue(gs_texture_t *texture)
 }
 
 static inline struct spa_pod *build_format(obs_pipewire_stream *obs_pw_stream,
-					   struct spa_pod_builder *b,
 					   uint32_t format, uint64_t *modifiers,
 					   size_t modifier_count)
 {
+	struct spa_pod_dynamic_builder pod_builder;
 	struct spa_rectangle max_resolution = SPA_RECTANGLE(8192, 4320);
 	struct spa_rectangle min_resolution = SPA_RECTANGLE(1, 1);
 	struct spa_rectangle resolution;
@@ -310,6 +311,8 @@ static inline struct spa_pod *build_format(obs_pipewire_stream *obs_pw_stream,
 	struct spa_fraction max_framerate;
 	struct spa_fraction min_framerate;
 	struct spa_fraction framerate;
+
+	spa_pod_dynamic_builder_init(&pod_builder, NULL, 0, 4096);
 
 	if (obs_pw_stream->framerate.set) {
 		framerate = obs_pw_stream->framerate.fraction;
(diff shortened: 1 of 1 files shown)


GeorgesStavracas:
@columbarius this should have the same effect as of https://github.com/obsproject/obs-studio/pull/8293, mind reviewing this PR?

GeorgesStavracas:
For reference, I copied the approach from [PipeWire's own code in gstpipewireformat.c](https://github.com/PipeWire/pipewire/blob/a9911f68a0758c36bcd4ff0badeb440237b3b055/src/gst/gstpipewireformat.c#L429-L488).

tytan652 (approved):
Code LGTM

plugins/linux-pipewire/pipewire.c
```diff
@@ -338,63 +341,65 @@ static inline struct spa_pod *build_format(obs_pipewire_stream *obs_pw_stream,
...
 	obs_pipewire *obs_pw = obs_pw_stream->obs_pw;
-	uint32_t params_count = 0;
+	g_autoptr(GPtrArray) params = NULL;
+
+	assert(out_params != NULL);
 
-	const struct spa_pod **params;
-	params = bzalloc(2 * obs_pw_stream->format_info.num *
-			 sizeof(struct spa_pod *));
+	params = g_ptr_array_new_full(2 * obs_pw_stream->format_info.num *
```
kkartaltepe, line 400: If we want to use a dynamic array we should use the obs DARRAY to keep consistency with the rest of obs.

plugins/linux-pipewire/pipewire.c
```diff
@@ -311,6 +312,8 @@ static inline struct spa_pod *build_format(obs_pipewire_stream *obs_pw_stream,
 	struct spa_fraction min_framerate;
 	struct spa_fraction framerate;
 
+	spa_pod_dynamic_builder_init(&pod_builder, NULL, 0, 4096);
```
kkartaltepe, line 315: This should be a lot smaller since it only needs to hold one format now instead of 50 formats with 50 modifiers it just needs to hold the 50 modifiers. I assume it grows automatically, but we can just see what the size is on amd and use that I guess.

kkartaltepe, line 315: spa_pod_dynamic_builder_clean is referenced in other places. I cant find any docs on spa_pod_dynamic_builder (not sure why i hoped it would exist), but can you confirm that we dont need to clean up after init.

kkartaltepe:
Seems to work fine on my amd box, maybe we should also backport this into 30.1 or something since there are some lingering reports of crash on pipewire start which have been related to these buffers being too small before.

plugins/linux-pipewire/pipewire.c
tytan652: DARRAY in this kind of code path (not cold) should be avoided.

kkartaltepe: First of all negotiation is cold. And second we know the size so there arnt any performance concerns to begin with. Its just a matter of using the api we all know or using the gptr api.

columbarius:
> @columbarius this should have the same effect as of #8293, mind reviewing this PR?

I'm not really a fan of adding more g_ptr and autopointer magic to the pipewire code, since it complicates the code for those not familiar with them (myself included).

An alternative approach would be one builder, but recording the offsets of the different pods and then collecting them at the end: columbarius, https://gitlab.freedesktop.org/pipewire/pipewire/-/blob/master/src/modules/module-filter-chain.c#L1525

PR #10461, opened 2024-03-30
DRAFT: linux-pipewire: Allow multiple stream implementations

columbarius:
### Description

Disentangle obs_pipewire and obw_pipewire_stream

### Motivation and Context

Currently the PipeWire Context cleanup is bound to the the Stream implementation. This would disentangle those to components. 

### How Has This Been Tested?

linux-pipewire should run as before. 

### Types of changes

### Checklist:

### DIFF
--- a/plugins/linux-pipewire/pipewire.c
+++ b/plugins/linux-pipewire/pipewire.c
@@ -97,6 +97,7 @@ struct _obs_pipewire {
 
 struct _obs_pipewire_stream {
 	obs_pipewire *obs_pw;
+	struct obs_pipewire_stream_impl impl;
 	obs_source_t *source;
 
 	gs_texture_t *texture;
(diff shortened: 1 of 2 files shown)


columbarius:
@GeorgesStavracas are you fine with this approach or should each type of plugin should have its own array to register instances?

WizardCM:
Please elaborate on the motivation, as we're a bit unsure about the intent and goals behind this change.

columbarius:
I'd like to rebase #5377, which shares the obs_pipewire struct and related code, but is not abstracted by obs_pipewire_stream. obs_pipewire currently tracks all streams and closes them on shutdown. The virtual camera plugin would introduce a obs_pipewire_outputstream struct, and it would be nice if this could also be tracked by obs_pipewire.

I see two options:
1. create a common struct with pointer and funcpointer to destructor and track every plugin with it
2. create an separate array in obs_pipewire for each plugin type.

columbarius:
Found another solution. This is not longer required. Sorry

PR #10465, opened 2024-04-01
obs-ffmpeg: Add option to output audio tracks in mono

Bleuzen:
### Description

This adds the possibility to output individual audio tracks in Mono:

### Motivation and Context

Currently we cannot set individual channel layouts per audio track, all tracks are recorded with the OBS global number of audio channels.
Most people have OBS running in Stereo, which makes sense for music, game sounds and others.
However it does not make sense for other things like most (mono) microphones. For that reason I added the ability to output specific tracks in Mono

### How Has This Been Tested?

Did some recordings with this in MP4, FLV & MKV and tested different audio codecs like AAC, Opus, FLAC & PCM

Played back the recordings, ensured everything sounds as expected

Also verified with `ffprobe`

### Types of changes

### Checklist:


Bleuzen:
stupid spelling mistake, will fix in a few hours

Warchamp7:
What problem or scenario does this solve that setting a device to Mono in Advanced Audio Properties does not handle?

Bleuzen:
> What problem or scenario does this solve that setting a device to Mono in Advanced Audio Properties does not handle?

@Warchamp7 The option in the Advanced Audio Properties does make specific sources _sound like_ they are mono by mixing all channels into each other, however it still results in a multi-channel output (stereo for most people) so you end up with multiple copies of the exact same audio which does
- not help reducing bandwidth or file size while recording
- clutter up editing software by showing unnecessarily many audio channels which are just the same

The option implemented in this PR on the other hand does allow for true mono output

Warchamp7:
> > What problem or scenario does this solve that setting a device to Mono in Advanced Audio Properties does not handle?
> 
> @Warchamp7 The option in the Advanced Audio Properties does make specific sources _sound like_ they are mono by mixing all channels into each other however it still results in a multi-channel output so we end up with multiple copies of the exact same audio
> 
> The option implemented in this PR on the other hand does allow for true mono output
> 
> Also it operates per track instead of per audio source so it extends flexibility

Is there a problem with setting OBS to single channel audio output?

Do you specifically want to be able to output ex. Audio Track 1 with 2 channels and Audio Track 2 with 1 channel?

What problem does this specifically solve? If this is an ease of use in another application such as a video editor, that's not sufficient.

Warchamp7:
This PR is out of date now but more importantly introduces added complexity that I do not think outweigh the benefits.

PR #10473, opened 2024-04-03
UI: Use spacers instead of widgets in macOS Permissions UI

RytoEX:
### Description

The macOS Permissions dialog was using empty widgets to force other widgets into position. Instead, use horizontal spacers.

### Motivation and Context

Noticed when reviewing #10419 that this dialog used blank widgets to help position other elements. Seemed odd. This change should supersede the specific `native="true"` changes to the empty widgets in this file so that that and the duplicate/identical item names are no longer an issue blocking that PR.

Note that I used Qt Creator 12 on Windows 11 to make these changes to the .ui file. Some of the changes made in #10419 do not appear to have been made by Qt Creator 12 when I saved this (notably, the orientation enum namespace), so this file will still need processed again.

### How Has This Been Tested?

Compiled and checked the dialog on macOS 13.6.6. The dialog looked the same before and after.

#### Before
Screenshot 2024-04-02 at 7 57 45 PM

#### After
Screenshot 2024-04-02 at 8 11 38 PM

### Types of changes

### Checklist:

### DIFF
--- a/UI/forms/OBSPermissions.ui
+++ b/UI/forms/OBSPermissions.ui
@@ -145,7 +145,17 @@
               <number>25</number>
              </property>
              <item>
-              <widget class="QWidget"/>
+              <spacer name="horizontalSpacer">
+               <property name="orientation">
+                <enum>Qt::Horizontal</enum>
+               </property>
+               <property name="sizeHint" stdset="0">
+                <size>
+                 <width>40</width>
+                 <height>20</height>
+                </size>
+               </property>
+              </spacer>
              </item>
              <item>
               <widget class="QPushButton" name="capturePermissionButton">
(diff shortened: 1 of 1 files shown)


gxalpha (approved):
This seems fine. We might want to look into getting rid of the `QFormLayout` entirely at some later point though.

RytoEX:
> This seems fine. We might want to look into getting rid of the `QFormLayout` entirely at some later point though.

Something to keep in mind for the future. I don't want to risk disrupting #10419 more than necessary.

RytoEX:
@derrod and @Warchamp7 approved off-thread. This dialog may require further adjusts later (e.g., making the buttons not fixed size), but we agreed that's out-of-scope for this PR. Merging.

PR #10475, opened 2024-04-03
aja: Call CNTV2Card::ApplySignalRoute with a CNTV2SignalRouter instance

paulh-aja:
The version of the `CNTV2Card::ApplySignalRoute` API which accepts a const ref to `NTV2XptConnections` results in additional crosspoint connection validation before `CNTV2Card::Connect` calls are made. If the validation logic fails under this version of the API, the `Connect` calls will also fail.

This PR switches to the version of `CNTV2Card::ApplySignalRoute` which accepts a const ref to a `CNTV2SignalRouter` instance. This API still performs the crosspoint validation but will not prevent the `Connect` calls. It is perfectly safe to call Connect even if the crosspoint connections are actually invalid, so there is really no reason to use the validation step unless writing code that explicitly needs to check its result.

Crosspoint validation is an entirely optional step and is performed via a "CanConnect" ROM table in the firmware. In rare cases where the table contains a bug, the SDK will prevent the crosspoints from connecting, even though the firmware can physically connect them. In other words, a bad entry in the CanConnect table can prevent the ability to connect a perfectly valid signal route, when using the `NTV2XptConnections` version of `CNTV2Card::ApplySignalRoute`.

As an example, here is a screenshot of the NTV2 Watcher application showing a case where the CanConnect ROM reports that a particular crosspoint connection (SDI4A => TSIMux2B, highlighted in red) is invalid even though the route is physically valid in the FPGA fabric:
image

The TSIMux2 widget and subsequent Framestore2 widgets light up in blue as there is actually a valid signal between those connections.

### Description

Switch to the version of `CNTV2Card::ApplySignalRoute` which accepts a `const CNTV2SignalRouter&` instance. This change allows crosspoint connections to be made even if the optional validation step fails, due to a rare firmware bug.

### Motivation and Context

We recently found that some older Kona5 firmwares had a bug in their "CanConnect" ROM table, which is used by the ajantv2 SDK for crosspoint connection validation. The crosspoint connection validation step is entirely optional but is used by the version of the `CNTV2Card::ApplySignalRoute` API which takes a `const NTV2XptConnections&`. In this rare situation it will prevent the ability to connect otherwise physically legal connections.

### How Has This Been Tested?

Tested with a known buggy Kona5 retail firmware, which had a couple of incorrect crosspoint "CanConnect" entries.

### Types of changes

### Checklist:

### DIFF
--- a/plugins/aja/aja-routing.cpp
+++ b/plugins/aja/aja-routing.cpp
@@ -295,7 +295,9 @@ bool Routing::ConfigureSourceRoute(const SourceProps &props, NTV2Mode mode,
 	if (!ParseRouteString(route_string, cnx))
 		return false;
 
-	card->ApplySignalRoute(cnx, false);
+	CNTV2SignalRouter sr;
+	sr.ResetFrom(cnx);
+	card->ApplySignalRoute(sr, false);
 
 	// Apply SDI widget settings
 	start_channel_index = GetIndexForNTV2Channel(init_channel);
@@ -441,7 +443,9 @@ bool Routing::ConfigureOutputRoute(const OutputProps &props, NTV2Mode mode,
 	if (!ParseRouteString(route_string, cnx))
 		return false;
 
-	card->ApplySignalRoute(cnx, false);
+	CNTV2SignalRouter sr;
+	sr.ResetFrom(cnx);
+	card->ApplySignalRoute(sr, false);
 
 	// Apply SDI widget settings
 	if (props.ioSelect != IOSelection::HDMIMonitorOut ||


RytoEX:
While I appreciate the deep dive, could you provide a layman's summary for how this affects end-users and what issue this would fix from their perspective?

Lain-B:
Echoing Ryan's commentary. If I'm reading this right, could this be described as "Fix devices not working(?) on older firmware"?

paulh-aja:
Yes, that is a suitable one-liner to describe this issue.

However, one of the other AJA devs tells me that we're eventually going to deprecate the usage of the `NTV2SignalRouter` class so this will really need to be fixed in the SDK.

I'll close this PR now and fix on our side.

PR #10478, opened 2024-04-03
UI: Log available encoders on load

Penwy:
### Description

This logs all available audio and video encoders when launching

### Motivation and Context

It can often be challenging for support volunteers to remember which encoders are available on which hardware, especially with AMD.
This would greatly simplify the flow of some support interactions.

### How Has This Been Tested?

Ran obs, checked that available encoders were correctly listed and no memory leak happened.

### Types of changes

### Checklist:

### DIFF
--- a/UI/window-basic-main.cpp
+++ b/UI/window-basic-main.cpp
@@ -1982,6 +1982,34 @@ void OBSBasic::ResetOutputs()
 	"Failed to initialize video.  Your GPU may not be supported, " \
 	"or your graphics drivers may need to be updated."
 
+static inline void LogEncoders()
+{
+	constexpr uint32_t hide_flags = OBS_ENCODER_CAP_DEPRECATED |
+					OBS_ENCODER_CAP_INTERNAL;
+
+	auto list_encoders = [](obs_encoder_type type) {
+		size_t idx = 0;
+		const char *encoder_type;
+
+		while (obs_enum_encoder_types(idx++, &encoder_type)) {
+			if (obs_get_encoder_caps(encoder_type) & hide_flags ||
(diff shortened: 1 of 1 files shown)


Fenrirthviti:
Please provide a sample log with this change applied for reference.

Penwy:
[here](https://obsproject.com/logs/QVOoB3lIxRTleZAy)'s a sample.

Penwy:
I'll wait for warchamp's review on it to make changes, but yeah I agree that rod's version is more readable, especially with the right tabulation to match.

I do similarly disagree about putting it to debugging, as the whole purpose is for ease of support, and asking an user to enable it is quite the same amount of work than asking for a screenshot of the encoders dropdown, if not more.

A solution to reduce verbosity could be to only list video encoders, since those are the one with the most variability, and most actual impact on support.

WizardCM:
I vote this doesn't need to be hidden behind a verbose flag. It is useful for all known use cases for reading the log file.

derrod:
This should probably not show deprecated/internal encoders, see the flags for those: https://github.com/obsproject/obs-studio/blob/01610d8c06edb08d0cc3155cb91b3e52e9a6473e/UI/window-basic-settings.hpp#L120-L121

Warchamp7 (approved):
Fine with adding this info, I don't think this needed a UX stamp.

Agree with Fenrir on fixing up the indentation. I defer to others if we want to explicitly separate video and audio encoders as Rodney suggested.

I don't think it's necessary to hide this behind verbose.

Penwy:
Modified according to rod's suggestions.
[Here](https://obsproject.com/logs/Q5hDlrUZvYDEwJ60) is an example of a log with it.

PR #10480, opened 2024-04-04
win-capture: Unhook from hidden windows

EZ64cool:
### Description

Unhook window-capture from windows that are fully hidden but still alive.

### Motivation and Context

Currently when a window-capture hooks into a window if the window then becomes hidden it won't unhook. This means we don't recieve an unhooked signal even though the window is effectively inaccessible by the user.

The behaviour for `IsIconic` makes sense to stay hooked because it is still easily accessible by the user.

A common example of an app doing this is Steam, closing steam's main window actually just hides the window rather than destroying it. This then means we don't get an unhooked signal when closing the steam window.

### How Has This Been Tested?

I've locally tested these changes against Steam's main window, and this also doesn't affect the normal behaviour of minimised or closed windows.

### Types of changes

### Checklist:

### DIFF
--- a/plugins/win-capture/window-capture.c
+++ b/plugins/win-capture/window-capture.c
@@ -635,7 +635,8 @@ static void wc_tick(void *data, float seconds)
 	if (!obs_source_showing(wc->source))
 		return;
 
-	if (!wc->window || !IsWindow(wc->window)) {
+	if (!wc->window || !IsWindow(wc->window) ||
+	    !IsWindowVisible(wc->window)) {
 		if (wc->hooked) {
 			wc->hooked = false;
 
@@ -686,7 +687,7 @@ static void wc_tick(void *data, float seconds)
 		wc->previously_failed = false;
 		reset_capture = true;
 
-	} else if (IsIconic(wc->window) || !IsWindowVisible(wc->window)) {
+	} else if (IsIconic(wc->window)) {
 		return; /* If HWND is invisible, WGC module can't be initialized successfully */
 	}
 


WizardCM:
From both a user perspective and a technical perspective, what benefits does this provide? Or is it just "it should be done this way"?

Lain-B:
Steam goes invisible, it stops capturing, but the window still exists. This code changes it so that if steam is hidden, it considers it destroyed and thus no longer captures it and starts polling to try to find Steam again, even though Steam still technically exists and you had the window handle to it and everything.

Are you absolutely sure that's what you want to do?

RytoEX:
> The main reason I made this change was because I'd created a scene with a window-capture that looked for any steam windows, I'd then open the main steam window (which would get hooked) if I then close the main window and try to open any other steam window (I would always try Friends for example) the window-capture would never hook into any of these windows as it was still hooked into the main steam window.

I'm not sure that I as a user would want a window capture changing from my main Steam Window to my Friends List automatically because I changed window states. Is that what this does?

> I feel like I've explained it badly, but this change aims to free up a window-capture when it's hooked process has become invisible to the user.

What happens to a source's audio when this was capturing a window's application audio?

EZ64cool:
> I'm not sure that I as a user would want a window capture changing from my main Steam Window to my Friends List automatically because I changed window states. Is that what this does?
This is all relies on the window-capture having the `Window Match Priority` set to either match class or match exe. So these changes don't affect that behaviour, just that it starts to look for a window when the window becomes fully hidden from the user.
Another more useful use case would be for it to capture the seperated browser window (created from middle clicking a link)

> What happens to a source's audio when this was capturing a window's application audio?
So we'll get a proper audio capture unhooked/rehook, but all of steams audio is coming through one pipe anyway so it doesn't seem to matter which window you're actually attached to. You'll get all audio. (Tested with trailer audio on game store pages)

RytoEX:
> > What happens to a source's audio when this was capturing a window's application audio?
> 
> So we'll get a proper audio capture unhooked/rehook, but all of steams audio is coming through one pipe anyway so it doesn't seem to matter which window you're actually attached to. You'll get all audio. (Tested with trailer audio on game store pages)

So the audio stops with this change, but before it would continue? Think outside of your Steam example for a moment and just consider a user with a single window that becomes "hidden" and hits this code path.

EZ64cool:
> > > What happens to a source's audio when this was capturing a window's application audio?
> > 
> > 
> > So we'll get a proper audio capture unhooked/rehook, but all of steams audio is coming through one pipe anyway so it doesn't seem to matter which window you're actually attached to. You'll get all audio. (Tested with trailer audio on game store pages)
> 
> So the audio stops with this change, but before it would continue? Think outside of your Steam example for a moment and just consider a user with a single window that becomes "hidden" and hits this code path.

Yeah that's a good point, and after actually recording it to test it does keep the audio hooked (I must have been thinking of game-capture which acts the way I was expecting).

So to summarise, the audio doesn't stop when the window becomes hidden, so this behaviour is unchanged from the current setup. But now we recieve an unhooked signal and the window-capture starts to search for a visible window to hook into.

norihiro:
The commit 4cb151403 has a too long description in a line, containing 73 characters, where the limit is 72.
> then means we don't get an unhooked signal when closing the steam window.

PR #10482, opened 2024-04-05
Revert "UI: Change default format for release builds to fMP4/fMOV"

RytoEX:
### Description

This reverts commit 9488bb20b68ef2ca12a4f3eb040f56762c6f1447.

### Motivation and Context

This is currently a frequent support issue as Windows Media Player, Movies & TV, and File Explorer all cannot seem to handle fMP4 files correctly. Windows Media Player and Movies & TV will play the files fine, but no duration is listed and the applications cannot seek the file. File Explorer will simply not list the file's length/duration (the field is shown as having no set value). This was somehow not caught during a year's worth of beta and RC builds. The world just wasn't ready for this. =(

For the record, we have reported these issues to Microsoft.

### How Has This Been Tested?

I haven't tested this, but as this is a revert it should be fine. We've used the previous code for a year prior.

### Types of changes

### Checklist:

### DIFF
--- a/UI/window-basic-main.cpp
+++ b/UI/window-basic-main.cpp
@@ -1462,8 +1462,9 @@ static const double scaled_vals[] = {1.0,         1.25, (1.0 / 0.75), 1.5,
 				     2.5,         2.75, 3.0,          0.0};
 
 extern void CheckExistingCookieId();
-
-#ifdef __APPLE__
+#if OBS_RELEASE_CANDIDATE == 0 && OBS_BETA == 0
+#define DEFAULT_CONTAINER "mkv"
+#elif defined(__APPLE__)
 #define DEFAULT_CONTAINER "fragmented_mov"
 #else
 #define DEFAULT_CONTAINER "fragmented_mp4"


gxalpha (commented):
If the main consideration is "This doesn't work with the default OS media player and tools", imo this should be limited to Windows (and maybe Linux).
On macOS, QuickTime Player (the default video player) deals just fine with Fragmented MOV but it can't even play MKV, so on macOS we'd be going back to the situation that this PR tries to fix to Windows.

RytoEX:
> If the main consideration is "This doesn't work with the default OS media player and tools", imo this should be limited to Windows (and maybe Linux). On macOS, QuickTime Player (the default video player) deals just fine with Fragmented MOV but it can't even play MKV, so on macOS we'd be going back to the situation that this PR tries to fix to Windows.

I'm not really willing to muck around with fine-tuning this for cherry-picking to 30.1.2. If someone wants to adjust this _after_ the revert, by all means.

Fenrirthviti (approved):
Approved. Something something video editors will cry.

Fenrirthviti:
Also unless I'm missing something, this revert should still set the default to fMP4 on macOS?

RytoEX:
> Also unless I'm missing something, this revert should still set the default to fMOV on macOS?
 
Seems that way.

PR #10485, opened 2024-04-06
UI: Use selected file path when opening file dialogs

### LINKED ISSUE #6777: OBS not saving last path when browsing for LUT filter

Warchamp7:
### Description
Updates the properties view to use the existing path of a selected file when creating file dialogs, rather than relying on logic existing where `obs_properties_add_path` is called.

### Motivation and Context
Fixes #6777 at a deeper level compared to the approach in #6786

Currently, file dialogs for property views reference default_path for their working directory. However, this is set only when the property view is being created. Selecting a file or resetting to defaults does not update this value. Due to most places in the the code having a default_path of NULL, behaviour was _usually_ as expected, as re-opening the file dialog would simply fall back to the last location browsed (on Windows at least).

In the case of the LUT filter, we specify a default_path where the default LUT files are located. This meant that after selecting a LUT file outside that folder, immediately re-opening the file dialog would start in the LUT folder again, rather than the directory of the selected file. The default_path would only get updated to the selected files location when the Filters dialog is re-opened or the filters properties are reloaded via switching off of it and back.

Rather than making sure we update default_path anywhere a file selection can happen, update properties view itself to use the selected path when creating the file dialog, allowing default_path to act strictly as the fallback when there is nothing selected.

### How Has This Been Tested?
Opened the file selectors for the LUT filter, Image slideshow properties, Image source properties, and Image source context bar. Ensured that the file dialog always opened to the location of the specified file if set, the default_path if it's not NULL, and native OS fallback in the case of neither.

### Types of changes

### Checklist:

### DIFF
--- a/UI/properties-view.cpp
+++ b/UI/properties-view.cpp
@@ -1813,17 +1813,21 @@ bool WidgetInfo::PathChanged(const char *setting)
 	obs_path_type type = obs_property_path_type(property);
 	const char *filter = obs_property_path_filter(property);
 	const char *default_path = obs_property_path_default_path(property);
+
+	QLineEdit *edit = static_cast<QLineEdit *>(widget);
+
+	QString startDir = edit->text();
+	if (startDir.isEmpty())
+		startDir = default_path;
+
 	QString path;
 
 	if (type == OBS_PATH_DIRECTORY)
-		path = SelectDirectory(view, QT_UTF8(desc),
-				       QT_UTF8(default_path));
+		path = SelectDirectory(view, QT_UTF8(desc), startDir);
 	else if (type == OBS_PATH_FILE)
-		path = OpenFile(view, QT_UTF8(desc), QT_UTF8(default_path),
-				QT_UTF8(filter));
+		path = OpenFile(view, QT_UTF8(desc), startDir, QT_UTF8(filter));
 	else if (type == OBS_PATH_FILE_SAVE)
-		path = SaveFile(view, QT_UTF8(desc), QT_UTF8(default_path),
-				QT_UTF8(filter));
+		path = SaveFile(view, QT_UTF8(desc), startDir, QT_UTF8(filter));
 
 #ifdef __APPLE__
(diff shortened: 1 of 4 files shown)


RytoEX (changes requested):
The second commit should either use the prefix "plugins: " or be split into two commits, one for "image-source: " and one for "obs-filters: ".

UI/properties-view.cpp
```diff
@@ -1812,18 +1812,23 @@ bool WidgetInfo::PathChanged(const char *setting)
 	const char *desc = obs_property_description(property);
 	obs_path_type type = obs_property_path_type(property);
 	const char *filter = obs_property_path_filter(property);
+	const char *last_path = obs_data_get_string(view->settings, setting);
```
derrod, line 1815: This seems unused.

derrod (changes requested):

norihiro:
The commit 7b72331f0 should have the prefix `plugins:` instead of `Plugins:`.

derrod (approved):
Seems fine aside from the commit message complaints.

PR #10494, opened 2024-04-08
obs-outputs: Fix FLV muxer tag size logic and file info offset

palana:
### Description
This is supposed to make the FLV output emit more compliant FLV files

### Motivation and Context
To prepare eRTMP support there's an upcoming PR that adds "eRTMP" (eFLV?) support to the FLV file output; having eFLV support is generally useful so testing doesn't have to involve RTMP/an actual server, and various tools used at Twitch complained about malformed FLVs (both the off-by-one tag size and the broken header/metadata after FLV files have been finalized)

### How Has This Been Tested?
Verified functionality with https://github.com/kc5nra/ertmp-toolbox and various internal tools

### Types of changes

### Checklist:

### DIFF
--- a/plugins/obs-outputs/flv-mux.c
+++ b/plugins/obs-outputs/flv-mux.c
@@ -123,7 +123,13 @@ static inline double encoder_video_codec(obs_encoder_t *encoder)
 	return 0.0;
 }
 
-#define FLV_INFO_SIZE_OFFSET 42
+/*
+ * This is based on the position of `duration` and `fileSize` in
+ * `build_flv_meta_data` relative to the beginning of the file
+ * to allow `write_file_info` to overwrite these two fields once
+ * the file is finalized.
+ */
+#define FLV_INFO_SIZE_OFFSET 58
 
 void write_file_info(FILE *file, int64_t duration_ms, int64_t size)
 {
--- a/plugins/obs-outputs/flv-mux.c
+++ b/plugins/obs-outputs/flv-mux.c
@@ -208,6 +208,19 @@ static void build_flv_meta_data(obs_output_t *context, uint8_t **output,
 	*output = bmemdup(buf, *size);
 }
 
+static inline void write_previous_tag_size_without_header(struct serializer *s,
+							  uint32_t header_size)
+{
+	assert(serializer_get_pos(s) >= header_size);
+	/* write tag size (starting byte doesn't count) */
+	s_wb32(s, (uint32_t)serializer_get_pos(s) - header_size - 1);
+}
+
+static inline void write_previous_tag_size(struct serializer *s)
+{
+	write_previous_tag_size_without_header(s, 0);
+}
+
 void flv_meta_data(obs_output_t *context, uint8_t **output, size_t *size,
 		   bool write_header)
 {
@@ -238,7 +251,7 @@ void flv_meta_data(obs_output_t *context, uint8_t **output, size_t *size,
 
 	s_write(&s, meta_data, meta_data_size);
 
-	s_wb32(&s, (uint32_t)serializer_get_pos(&s) - start_pos - 1);
+	write_previous_tag_size_without_header(&s, start_pos);
 
 	*output = data.bytes.array;
 	*size = data.bytes.num;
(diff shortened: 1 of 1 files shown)


Lain-B (changes requested):
Could you change second commit title to:
```
obs-outputs: Change FLV tag size according to v10 spec
```

Sorry about the nitpick. Only just noticed it. That's all I've got though, otherwise looks good to me. And thank you for the fixes.

palana:
> Could you change second commit title to:
> 
> ```
> obs-outputs: Change FLV tag size according to v10 spec
> ```
> 
> Sorry about the nitpick. Only just noticed it. That's all I've got though, otherwise looks good to me. And thank you for the fixes.

done

plugins/obs-outputs/flv-mux.c
```diff
@@ -123,7 +123,13 @@ static inline double encoder_video_codec(obs_encoder_t *encoder)
 	return 0.0;
 }
 
-#define FLV_INFO_SIZE_OFFSET 42
+/*
+ * this is based on the position of `duration` and `fileSize` in
+ * `build_flv_meta_data` relative to the beginning of the file
+ * to allow `write_file_info` to overwrite these two fields once
+ * the file is finalized
```
RytoEX, line 130:
```suggestion
 * This is based on the position of `duration` and `fileSize` in
 * `build_flv_meta_data` relative to the beginning of the file
 * to allow `write_file_info` to overwrite these two fields once
 * the file is finalized.
```

RytoEX (changes requested):
Nits.

In the first commit message, end the commit message body with punctuation.

If there's a way to adjust the PR title to follow the commit message subject pattern (prefix: verb object), that would be welcome, but I understand that may be difficult when there are multiple commits present.

palana:
> If there's a way to adjust the PR title to follow the commit message subject pattern (prefix: verb object), that would be welcome, but I understand that may be difficult when there are multiple commits present.

@RytoEX: I changed the PR title, but not sure this is better; feel free to change the title however you like if you have a better suggestion

palana:
> In the first commit message, end the commit message body with punctuation.

pushed a fixup commit in https://github.com/obsproject/obs-studio/pull/10494#commits-pushed-31f855f

RytoEX:
@palana Squash the amends/fixups and I'll take a look, but these seem good.

palana:
> @palana Squash the amends/fixups and I'll take a look, but these seem good.

@RytoEX: done

plugins/obs-outputs/flv-mux.c
```diff
@@ -208,6 +214,26 @@ static void build_flv_meta_data(obs_output_t *context, uint8_t **output,
...
+	/*
+	 * From FLV file format specification version 10:
+	 * Size of previous [current] tag, including its header.
+	 * For FLV version 1 this value is 11 plus the DataSize of
+	 * the previous [current] tag.
+	 */
+	s_wb32(s, (uint32_t)serializer_get_pos(s) - header_size);
+}
+
+static void write_previous_tag_size(struct serializer *s)
```
RytoEX, line 232: Should these be marked `inline`?

RytoEX (commented):
One general question.

RytoEX:
Other than my review question, this looks good to me.

plugins/obs-outputs/flv-mux.c
palana: I don't currently see a reason why they can't be marked as `static inline`, but I'm also not sure I see a reason why they should be marked `static inline`

let me know if you want me to change these

RytoEX: I don't have a strong opinion here, so I'm fine with it as-is.

Lain-B: I'd say it makes sense here, but as to whether it matters is whether it's actually a hot path or not. This is called for each tag (packet), so it's called a fair amount, but I'd say it's ultimately not going to be hot enough to make any bit of a noticeable difference to users.

But I'd be nice to have because it's a one-liner function. In terms of assembly output this is basically unnecessarily separated code at that point. Depending on compiler settings it may automatically get inlined anyway but you'd have to check the disassembly if you're *really* curious about these cases.

So I'd say might as well do it but it's not a big deal either way :woman_shrugging:

palana: changed to `static inline` in https://github.com/obsproject/obs-studio/pull/10494#commits-pushed-8582b2e

RytoEX (approved):
LGTM

PR #10495, opened 2024-04-08
linux-pulseaudio: Implement audio output

david672orford:
### Description

The pulseaudio plugin previously supported only audio inputs and capture of outputs using their monitor sources. This PR adds output to sinks. It can also create a virtual audio cable and tear it down when the output is stopped.

* New pulse-output.c: implements pulse\_output
* New pulse-utils.c common functions extracted from pulse-input.c
* pulse-wrapper.c: new functions to get sink info, load and unload modules
* No changes to pulse input code (other than moving some to pulse-utils.c)
* No UI changes. See PR for lua scripts to test.

The pulse\_output takes two properties:

* device -- the name of the PulseAudio sink. Start with "obs-" to dynamically create a virtual audio cable
* description -- Displayname of virtual audio cable

### Motivation and Context

Existing audio outputs either stream over the Internet or require special hardware
such as Decklink. Source audio monitoring is currently used as a substitute, but setting
it up can be error-prone and it lacks certain features such as fade transitions. Setting
up virtual audio cables for feeding to audio-conferencing programs such as Zoom is
even trickier and more error prone. An actual audio output which can create its own
cables simplifies this considerably. However, more testing in required which is why this
PR is marked "draft".

### How Has This Been Tested?

Tested on Ubuntu 20.04 using both Pulseaudio and Pipewire. Output to speakers and to Zoom.

### Types of changes

### Checklist:

- [] The code has been tested.
- [] I have included updates to all appropriate documentation.


david672orford:
Repushed to get cleaner diffs.

kkartaltepe:
Thanks for the PR, FYI we generally recommend larger changes to go through RFCs (https://github.com/obsproject/rfcs) so authors can get feedback before implementing things. 

Considering pulseaudio is nearly unmaintained in OBS (and has been phased out of most distros over the last few years) is there a particular reason you implemented this against pulseaudio?

norihiro:
If you release this as a 3rd party plugin at https://obsproject.com/forum/plugins/, I think it would be welcome.

To me, the output (`obs_output_t`) type is interesting because the monitor can just monitor each source. So the user cannot listen actual streaming or recording sound. For example if the track assignment is unintentional, the audio could fails to go to the live without noticed by the monitor. On the other hand, the output-type will let user to listen to the mixed track. Probably, #4741 is another hope of the compatible feature.

However, I don't know this PR is worth to be merged for now. Especially, I think these points would be the major concerns.
- No UI is provided.
- Windows and macOS are not supported.

kkartaltepe:
> However, I don't know this PR is worth to be merged for now. Especially, I think these points would be the major concerns.
> 
>     * No UI is provided.
> 
>     * Windows and macOS are not supported.

I think these are a concern for the virtual cam audio support in general and not necessarily for this PR. Personally I don't know what folks were considering for the actual implementation whether its an output or something new.

> If you release this as a 3rd party plugin at https://obsproject.com/forum/plugins/, I think it would be welcome.

I would also like to see bugs shaken out earlier rather than later which is easier to get as a plugin than a PR historically. This is what the pipewire application capture has done for example. Of course feel free to just wait as at least I am still not sure if this a direction I would want to see merged, other folks may have other opinions.

david672orford:
I am just looking for a way to move this forward. I love OBS and am amazed by the work done on it, but am still frustrated by the lack of audio output. I sense that the maintainers are streamers who don't see a need.

But there is a considerable base of users who are using it for other things and they have been complaining about the lack of audio output. Sure, you can set up monitoring on all of your sources and connect them to a virtual cable which you create before starting OBS, but this is error prone and not very reliable. The monitor streams tend to get disconnected from the virtual cable for various reasons. I have had to write non-trivial scripts to get them reconnected quickly. And you point out, [norihiro](https://github.com/norihiro) points, this would benefit streamers too since they could monitor the final audio mix.

You, [kkartaltepe](https://github.com/kkartaltepe), refer to some issues surrounding virtual cameras. I don't know what these are, but the solution I intended to use on Linux is to install this plugin and the Lua script above to start audio output on a virtual cable whenever the virtual camera is running.

If the maintainers feel that a 3rd party plugin is the best way forward, I would respect that. I submitted this as a PR because it is non-disruptive and the recommendation is to implement major features in pieces through a series of PR's. My hope is that once there is audio output on Linux, others will be motived to implement it on Windows and MacOS.

I am not able to add audio output to the Windows and MacOS plugins myself. However, I do intent to continue working on PulseAudio support in OBS as long as it continues to use the PulseAudio API's on Linux. For example, I would like it if OBS could set metadata on audio streams so that we could see more distinctive labels in the volume control and Pipewire patchbays. I also want to look into whether the mute state of a microphone can be syncronized with the mute state in the PulseAudio/Pipewire mixer so that the mute lamp on the microphone will show the correct status. So this is not a one-off PR that I am trying to push into the project and then abandon.

tt2468:
Please update your commits to remove the submodule changes.

david672orford:
The submodule changes have been removed.

ogmkp:
> Thanks for the PR, FYI we generally recommend larger changes to go through RFCs (https://github.com/obsproject/rfcs) so authors can get feedback before implementing things.
> 
> Considering pulseaudio is nearly unmaintained in OBS (and has been phased out of most distros over the last few years) is there a particular reason you implemented this against pulseaudio?

What's more, there are still problems with pulseaudio (sound lag on VLC source or with the ndi plugin, for example), so what would be really good would be a pipewire implementation in the backend instead of pulseaudio.

I didn't quite understand this attempted PR, as pulseaudio wrapper for pipewire already creates nodes to monitor sources that can be routed.

david672orford:
> I didn't quite understand this attempted PR, as pulseaudio wrapper for pipewire already creates nodes to monitor sources that can be routed.

My motive for creating this PR is that I wanted to be able to use OBS's audio mixer at in-person events and with the virtual camera.

The source monitoring provided in OBS can be used as a workaround for the missing output, but it is clunky. You have to turn it on for each source. Audio transitions will not be heard. And you need to use a patchbay program to route the microphones separately and you can't use OBS filters on them. This makes the OBS mixer pretty useless, so I just hide it.

I will post here when I have repacked this PR as a plugin.

ipatix:
I would really love if this were added to OBS. For me it is the same case: In person events and the ability to output audio. Ideally with the ability to select which track (1-6) to output.

The last two years I've been cherry-picking and rebasing various PR commits to make this possible. Though, the maintenance work for this is really annoying every time a new OBS version releases.

My own bits about various discussed points:
- PulseAudio is a good API to choose, since compatibility is very good (legacy PulseAudio & newer PipeWire).
- I wouldn't necessarly expect good latency with PA. Perhaps JACK or native PipeWire API would be better. Though, monitoring in OBS currently works with PA only, so it wouldn't be a downgrade either way.
- Hopefully latency buildup / desync will be solved by #6351
- As for Windows and Mac OS not supported: Well, no PA available on those platform, so no point in supporting that. Though, it'd certainly be handy to have WASAPI output on Windows and CoreAudio on Mac.

PR #10501, opened 2024-04-09
UI: Adjust button order in vertical mixer

Warchamp7:
### Description
Moves the mute button before the config menu button in vertical mixer layout.

### Motivation and Context
Any action with a dedicated button has a dedicated button because we consider it important or frequent. Additional config/options should always be the last button entry. This also matches the button order in horizontal layout.

### How Has This Been Tested?
image

### Types of changes

### Checklist:

### DIFF
--- a/UI/volume-control.cpp
+++ b/UI/volume-control.cpp
@@ -306,7 +306,6 @@ VolControl::VolControl(OBSSource source_, bool showConfig, bool vertical)
 			controlLayout->setAlignment(config, Qt::AlignVCenter);
 		}
 
-		controlLayout->addItem(new QSpacerItem(3, 0));
 		// Add Headphone (audio monitoring) widget here
 		controlLayout->addWidget(mute);
 		controlLayout->setAlignment(mute, Qt::AlignVCenter);
--- a/UI/volume-control.cpp
+++ b/UI/volume-control.cpp
@@ -301,14 +301,14 @@ VolControl::VolControl(OBSSource source_, bool showConfig, bool vertical)
 		controlLayout->setContentsMargins(0, 0, 0, 0);
 		controlLayout->setSpacing(0);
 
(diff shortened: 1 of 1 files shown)


derrod:
I think the reason it is that way is so that the mute button is directly below the volume slider, which I personally think is how it should be.

Warchamp7:
> I think the reason it is that way so that the mute button is directly below the volume slider, which I personally think is how it should be.

It won't work out that way anyway if/when audio monitoring toggle finally gets added

derrod:
> It won't work out that way anyway if/when audio monitoring toggle finally gets added

But this order goes against everything that every other software - including OBS - does and looks wrong. If anything, the kebab menu shouldn't be there because it's duplicated functionality.

WizardCM:
I do have a minor concern with muscle memory, and do agree with Rodney about keeping the mute button below the slider.

PatTheMav:
At least in DAWs (e.g. Cubase, Logic Pro) it is common to place the mute button under the volume slider (because they are logically connected) and put a "solo" button next to it (which mutes all other inputs that don't have the solo button pressed):



The other variant I've seen is based off of hardware mixers which usually put the buttons atop of each other either above or below the meter/slider combo or the slider is shorter than the meter and thus the buttons live below the slider.

But the common theme is that muting/soloing is connected to "volume adjustment".

Warchamp7:
> > It won't work out that way anyway if/when audio monitoring toggle finally gets added
> 
> But this order goes against everything that every other software - including OBS - does and looks wrong. If anything, the kebab menu shouldn't be there because it's duplicated functionality.

The kebab button exists to avoid a number of actions being buried within strictly a context menu.

In any case, I'm closing this in favor of a larger pass of the audio mixer I'm working on.

PR #10511, opened 2024-04-10
UI: Reset duration when removing show/hide transition

Warchamp7:
### Description
Reset a scene items show/hide transition duration to 0 when selecting the 'None' option.

### Motivation and Context
We currently treat the globally set transition time as a default when configuring a scene item Show/Hide transition. Once a show/hide transition gets set on a scene item, it is never cleared, meaning if a show/hide transition is removed by setting it to `None`, it will retain that duration if a transition is later set again, instead of using the global duration again.

### How Has This Been Tested?
Configured show/hide transitions on multiple sources and ensured the main window global duration is only used as a default. Ensured that this value is properly reset when selecting `None` as a transition.

### Types of changes

### Checklist:

### DIFF
--- a/UI/window-basic-main-transitions.cpp
+++ b/UI/window-basic-main-transitions.cpp
@@ -1183,6 +1183,8 @@ QMenu *OBSBasic::CreateVisibilityTransitionMenu(bool visible)
 		if (id.isNull() || id.isEmpty()) {
 			obs_sceneitem_set_transition(sceneItem, visible,
 						     nullptr);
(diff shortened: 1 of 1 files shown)


WizardCM:
To confirm, this field defaults to 300ms

image

And setting the field in the Scene Transitions dock to 0 does set it to 300ms, so I think this change is fine? I approve especially of it only changing back to default when None is selected.

GeorgesStavracas (approved):

PR #10518, opened 2024-04-11
obs-webrtc: Adjust RtcpNackResponder from bitrate

Sean-Der:
### Description
Adjust RtcpNackResponder from bitrate

### Motivation and Context
At higher bitrates the cache would be undersized. Because of the undersized cache we wouldn't have enough NACK history to fix the stream.

### How Has This Been Tested?
Against Broadcast Box with a packet loss percentage of 1%

### Types of changes

### Checklist:

### DIFF
--- a/plugins/obs-webrtc/whip-output.cpp
+++ b/plugins/obs-webrtc/whip-output.cpp
@@ -21,6 +21,9 @@ const uint8_t audio_payload_type = 111;
 const char *video_mid = "1";
 const uint8_t video_payload_type = 96;
 
+// ~3 seconds of 8.5 Megabit video
+const int video_nack_buffer_size = 4000;
+
 WHIPOutput::WHIPOutput(obs_data_t *, obs_output_t *output)
 	: output(output),
 	  endpoint_url(),
@@ -181,7 +184,8 @@ void WHIPOutput::ConfigureVideoTrack(std::string media_stream_id,
 
 	video_sr_reporter = std::make_shared<rtc::RtcpSrReporter>(rtp_config);
 	packetizer->addToChain(video_sr_reporter);
-	packetizer->addToChain(std::make_shared<rtc::RtcpNackResponder>());
+	packetizer->addToChain(std::make_shared<rtc::RtcpNackResponder>(
+		video_nack_buffer_size));
 
 	video_track = peer_connection->addTrack(video_description);
 	video_track->setMediaHandler(packetizer);


plugins/obs-webrtc/whip-output.cpp
```diff
@@ -21,6 +21,9 @@ const uint8_t audio_payload_type = 111;
 const char *video_mid = "1";
 const uint8_t video_payload_type = 96;
 
+// Convert Kbps to 'How many 1500 Byte packets in 5 seconds'
+const float videoNackStorageFactor = 125.0 / 1500.0 * 5.0;
```
RytoEX, line 25: Why not `constexpr`?

RytoEX, line 25: To resolve MSVC C4305 and C22220:

```suggestion
const float videoNackStorageFactor = 125.0f / 1500.0f * 5.0f;
```

RytoEX (changes requested):

Sean-Der:
Thanks for the review @RytoEX I made the changes

plugins/obs-webrtc/whip-output.cpp
Sean-Der: Changed to `constexpr`

Sean-Der: Changes were added (and was made constexpr)

plugins/obs-webrtc/whip-output.cpp
```diff
@@ -21,6 +21,9 @@ const uint8_t audio_payload_type = 111;
 const char *video_mid = "1";
 const uint8_t video_payload_type = 96;
 
+// Convert Kbps to 'How many 1500 Byte packets in 5 seconds'
+constexpr float videoNackStorageFactor = 125.0f / 1500.0f * 5.0f;
```
tt2468, line 25: This uses an MTU of 1500 bytes, but we're using a video fragment size of 1200 bytes above. Should this be adjusted accordingly, or in some way refer to that value?

plugins/obs-webrtc/whip-output.cpp
```diff
@@ -62,6 +65,10 @@ bool WHIPOutput::Start()
 
 	is_av1 = (strcmp("av1", obs_encoder_get_codec(encoder)) == 0);
 
+	obs_data_t *video_settings = obs_encoder_get_settings(encoder);
```
tt2468, line 68: Please use OBSDataAutoRelease here.

plugins/obs-webrtc/whip-output.cpp
```diff
@@ -62,6 +65,10 @@ bool WHIPOutput::Start()
...
+	video_bitrate = (int)obs_data_get_int(video_settings, "bitrate");
```
tt2468, line 69: This only works for CBR and similar modes. Is there any consideration made for if a user uses an encoder in CQP or similar modes? Additionally, I could see it being an issue if an encoder has a "bitrate" value applied to its settings object, but the encoder is not using a mode which actually references the "bitrate" value.

Bleuzen: Also OBS does allow changing the bitrate while streaming, currently this PR does not adapt to that but only reads the bitrate once on stream start so this is not optimal as is
and I'm not sure how well this plays with VBR

Sean-Der:
@tt2468 @Bleuzen Do you think it would be better to set a 'sensible' higher default then?

In libdatachannel the default is 512. For 8mbit it  is ~1200 packets a second, I was hoping to give users ~5 seconds of cache to support people with mixed work loads. 

* People viewing in real-time get some packet loss
* The server is also saving the video and will aggressively NACK old video to make sure saved video is perfect

Bleuzen:
> Do you think it would be better to set a 'sensible' higher default then?

Not sure how this translates to resource usage (memory / bandwidth?) and if everyone would be happy with that so I don't have a strong opinion on that currently

However I wonder if it is better to make this user configurable.
Could be simple dropdown options like:
- Default (512)
- High (1024)
- Very High (2048)
- Ultra (4096)

Or a text field as used for the bitrate

Sean-Der:
The extra memory usage will be the 5 seconds of video. It shouldn't impact bandwidth, the NACKs are only requested on loss. 

I don't have a strong opinion either. I just realized when debugging lossy networks our current NACK setup isn't useful.

tt2468:
If we're just trying to determine a buffer size which has to be *available* but not always *filled*, then yeah we can probably go pretty high like 8MB without too much worry. If the behavior of the output changes depending on whether the buffer is the required size versus if it's oversized, then we might have to be more careful.

Sean-Der:
@tt2468 Yep as you described, available but not always filled.

It would be really nice to bump this especially for users with higher bitrates and packet loss, the current value is only giving them ~.5 second of buffer

derrod:
Wouldn't it make more sense to adjust the buffer size based on (configured) output bitrate? Currently OBS outputs are guaranteed to target CBR, so you could query the bitrate and set it appropiately. See the RTMP Windows send loop for example: https://github.com/obsproject/obs-studio/blob/master/plugins/obs-outputs/rtmp-stream.c#L1046-L1067

Sean-Der:
@derrod either works for me! The original worked that way, @tt2468 and @Bleuzen lefts comments around the following

```
This only works for CBR and similar modes. Is there any consideration made for if a user uses an encoder in CQP or similar modes? Additionally, I could see it being an issue if an encoder has a "bitrate" value applied to its settings object, but the encoder is not using a mode which actually references the "bitrate" value.
```

So it felt safer to have a high default value.

Bleuzen:
> Currently OBS outputs are guaranteed to target CBR

How exactly does this work?
The UI does not only show CBR

I found that for webrtc using VBR works better because most encoders do introduce big spikes in the data rate on I-Frames especially in low motion scenes
These spikes do cause freezes very often
Using VBR does lower the bitrate in low motion scenes which does also make these spikes smaller
So VBR streams freeze less often and are more stable than CBR streams
Have been using VBR for webrtc streaming over the last months for this reason

kc5nra:
I think this could be fixed in two stages. One just unblocks general usage of this feature with an appropriate sized buffer (say 4000 packets or 6 megabytes). This is a negligible amount and could be optimized later with dynamic behavior based on bitrate or configuration

Sean-Der:
To give people confidence that I have fixed this properly I wrote https://github.com/sean-der/whip-nack-test This is a WHIP server that simulates packet loss. You can start with `go run .` and then set your WHIP server to `http://localhost:8080/` and use any key.

Without packet loss OBS will stop responding to packets at `550`. This means that at most 750 kilobytes of history is available for video.
```
Sending NACK for Packet (15726) which is (550) in the past
```

With my change you now will see it go past that point now
```
Pkt with SequenceNumber(35500) correctly RTXed
Sending NACK for Packet (36358) which is (900) in the past
```

Giving us 6 Megabytes of history, giving a much better experience in broken networks.

DDRBoxman (approved):
👍 

Just making the buffer bigger seems like a good compromise for now to make the end user experience better, especially since it's only about 8mb of memory.

RytoEX (approved):
Seems fine at a glance.

PR #10524, opened 2024-04-12
obs-webrtc: Add Link Header support

Sean-Der:
### Description

WHIP/WHEP allows ICE Servers to be specified via [Link Headers](https://www.ietf.org/archive/id/draft-ietf-wish-whip-13.html#name-stun-turn-server-configurat)

### Motivation and Context
Some WebRTC deployments are behind a NAT. Companies/developers place their TURN server on the edge, and use that to access the TURN servers.

I have had 3 companies reach out to me about this, so implemented it.

### How Has This Been Tested?
Ran broadcast-box and only listened on `127.0.0.1` and confirmed that traffic could be routed through [pion/turn](https://github.com/pion/turn)

### Types of changes

### Checklist:

### DIFF
--- a/plugins/obs-webrtc/whip-output.cpp
+++ b/plugins/obs-webrtc/whip-output.cpp
@@ -220,8 +220,13 @@ bool WHIPOutput::Init()
  */
 bool WHIPOutput::Setup()
 {
+	rtc::Configuration cfg;
 
-	peer_connection = std::make_shared<rtc::PeerConnection>();
+#if RTC_VERSION_MAJOR == 0 && RTC_VERSION_MINOR > 20 || RTC_VERSION_MAJOR > 1
+	cfg.disableAutoGathering = true;
+#endif
+
+	peer_connection = std::make_shared<rtc::PeerConnection>(cfg);
 
 	peer_connection->onStateChange([this](rtc::PeerConnection::State state) {
 		switch (state) {
@@ -277,6 +282,68 @@ bool WHIPOutput::Setup()
 	peer_connection->setLocalDescription();
 
 	return true;
+}
+
+// Given a Link header extract URL/Username/Credential and create rtc::IceServer
+// <turn:turn.example.net>; username="user"; credential="myPassword";
+//
+// https://www.ietf.org/archive/id/draft-ietf-wish-whip-13.html#section-4.4
+void WHIPOutput::ParseLinkHeader(std::string val,
+				 std::vector<rtc::IceServer> &iceServers)
+{
+	std::string url, username, password;
+
+	auto extractUrl = [](std::string input) -> std::string {
+		auto head = input.find("<") + 1;
+		auto tail = input.find(">");
+
+		if (head == std::string::npos || tail == std::string::npos) {
+			return "";
+		}
+		return input.substr(head, tail - head);
+	};
+
+	auto extractValue = [](std::string input) -> std::string {
+		auto head = input.find("\"") + 1;
+		auto tail = input.find_last_of("\"");
+
+		if (head == std::string::npos || tail == std::string::npos) {
+			return "";
+		}
+		return input.substr(head, tail - head);
+	};
+
+	while (true) {
+		std::string token = val;
+		auto pos = token.find(";");
+		if (pos != std::string::npos) {
+			token = val.substr(0, pos);
+		}
+
+		if (token.find("<turn:", 0) == 0) {
+			url = extractUrl(token);
+		} else if (token.find("username=") != std::string::npos) {
+			username = extractValue(token);
+		} else if (token.find("credential=") != std::string::npos) {
+			password = extractValue(token);
+		}
+
+		if (pos == std::string::npos) {
+			break;
+		}
+		val.erase(0, pos + 1);
+	}
+
+	try {
+		auto iceServer = rtc::IceServer(url);
+		iceServer.username = username;
+		iceServer.password = password;
+		iceServers.push_back(iceServer);
+	} catch (const std::invalid_argument &err) {
+		do_log(LOG_WARNING,
+		       "Failed to construct ICE Server from %s: %s",
+		       val.c_str(), err.what());
+	}
 }
 
 bool WHIPOutput::Connect()
@@ -368,6 +435,25 @@ bool WHIPOutput::Connect()
 	CURLU *url_builder = curl_url();
 	auto last_location_header = location_headers.back();
 
+	// Parse Link headers to extract STUN/TURN server configuration URLs
+	struct curl_header *prev = nullptr;
+	struct curl_header *h = nullptr;
+	std::vector<rtc::IceServer> iceServers;
+	while ((h = curl_easy_nextheader(c, CURLH_HEADER, 0, prev))) {
+		if (astrcmpi("Link", h->name) == 0) {
+			auto value = std::string(h->value);
+			// Parse multiple links separated by ','
+			for (auto end = value.find(",");
+			     end != std::string::npos; end = value.find(",")) {
+				this->ParseLinkHeader(value.substr(0, end),
+						      iceServers);
+				value = value.substr(end + 1);
+			}
+			this->ParseLinkHeader(value, iceServers);
+		}
+		prev = h;
+	}
+
 	// If Location header doesn't start with `http` it is a relative URL.
 	// Construct a absolute URL using the host of the effective URL
 	if (last_location_header.find("http") != 0) {
@@ -442,6 +528,11 @@ bool WHIPOutput::Connect()
 		return false;
 	}
 	cleanup();
+
+#if RTC_VERSION_MAJOR == 0 && RTC_VERSION_MINOR > 20 || RTC_VERSION_MAJOR > 1
+	peer_connection->gatherLocalCandidates(iceServers);
+#endif
+
 	return true;
 }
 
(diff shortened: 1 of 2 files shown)


iameli:
I just tested this against OBS Studio and it seems to be working flawlessly.

plugins/obs-webrtc/whip-output.cpp
```diff
@@ -266,6 +271,47 @@ bool WHIPOutput::Setup()
...
+		if (head == std::string::npos || tail == std::string::npos) {
+			return "";
+		}
+		return input.substr(head, tail - head);
+	};
+
+	while ((pos = val.find(";")) != std::string::npos) {
+		auto token = val.substr(0, pos);
+		if (token.find("turn:", 0) == 0) {
+			url = token;
```
sile, line 293: Thank you for adding TURN support to WHIP!
I have confirmed that the implementation is working successfully with our SFU after making some minor modifications (please refer to the feedback below).

---

Seeing https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Link, the handling of `<` and `>` seems necessary as the doc says "The URI (absolute or relative) must be enclosed between `<` and `>`".

Besides, the doc also says "You can specify multiple links separated by commas". So it would be great if you could add comma handling to support multiple links.
(I know that the current implementation can handle multiple links if SFU returns multiple, duplicate LINK HTTP header fields. However, some HTTP server libraries (e.g., [cowboy](https://github.com/ninenines/cowboy) which is a famous library in Erlang) does not allow response headers having duplicate fields.)

Sean-Der: Thanks for testing @sile!

I will fix that tomorrow and ping you for another test :)

Sean-Der: Did you already make modifications to the C++ code or your TURN service? If you modified OBS would you be able to push to my fork?

I gave you write permission to my fork.

sile: I modified our SFU. However, if it's okay, I'm willing to fix the OBS code and push the patch to your fork 😃

sile: Pushed: https://github.com/obsproject/obs-studio/pull/10524/commits/fd771428aa538c2ddc37860e17963e2a21ba8d59

Feel free to modify (or discard) the code as I am not familiar with C++ and OBS 😅

Sean-Der: @sile Your code is fantastic! I am really impressed how quickly you were able to do that not being familiar with C++ and OBS.

I squashed the commits down into one and fixed some linter errors. Beyond that everything looks perfect to me.

build-aux/modules/50-libdatachannel.json
```diff
@@ -12,10 +12,9 @@
     "sources": [
         {
             "type": "git",
-            "url": "https://github.com/paullouisageneau/libdatachannel.git",
+            "url": "https://github.com/murillo128/libdatachannel.git",
```
mogren, line 15: I see that this fork is 24 versions behind. Is there some specific patch that's not merged back upstream that is needed?

Sean-Der: I can switch back to upstream now! Completely forgot about that  :/

Are you building something with OBS+WebRTC @mogren I had no idea!?

mogren: Sorry, not really, just reading code trying to learn. 🐰 🕳️ 🕳️ everywhere... 😁

Sean-Der:
@tt2468 Could I get a review of this one now please?

Sean-Der:
I have created a test application that confirms this is working. You can get it here https://github.com/Sean-Der/whip-turn-test

The only thing you need to change is the IP address of the host you are running the server on [here](https://github.com/Sean-Der/whip-turn-test/blob/master/main.go#L20)

OBS without this patch running against this server behaves like this. It shows WHIP connected and starting pushing media.
```
Running WHIP server at http://localhost:8085
ICE Connection State has changed: checking
ICE Connection State has changed: connected
Getting incoming media audio/opus
Getting incoming media video/H264
```

Without this change the server prints the following. OBS doesn't have TURN support so is unable to connect
```
Running WHIP server at http://localhost:8085
ICE Connection State has changed: checking
```

Some users today who are using WHIP will see OBS unable to connect and have no idea why

plugins/obs-webrtc/whip-output.cpp
```diff
@@ -368,6 +424,24 @@ bool WHIPOutput::Connect()
 	CURLU *url_builder = curl_url();
 	auto last_location_header = location_headers.back();
 
+	struct curl_header *prev = nullptr;
```
DDRBoxman, line 427: Probably want a comment here saying that the link header parsing is happening here.

plugins/obs-webrtc/whip-output.cpp
```diff
@@ -368,6 +424,24 @@ bool WHIPOutput::Connect()
...
+	struct curl_header *h = nullptr;
+	std::vector<rtc::IceServer> iceServers;
+	while ((h = curl_easy_nextheader(c, CURLH_HEADER, 0, prev))) {
+		if (strcmp("Link", h->name) == 0 ||
+		    strcmp("link", h->name) == 0) {
+			auto value = std::string(h->value);
+			for (auto end = value.find(",");
```
DDRBoxman, line 434: I'd also call out this parsing here is for having multiple values under the same header

plugins/obs-webrtc/whip-output.cpp
```diff
@@ -279,6 +284,57 @@ bool WHIPOutput::Setup()
 	return true;
 }
 
+void WHIPOutput::ParseLinkHeader(std::string val,
```
DDRBoxman, line 287: Might be nice to put some of the example strings in documentation for the method.

plugins/obs-webrtc/whip-output.cpp
```diff
@@ -279,6 +284,57 @@ bool WHIPOutput::Setup()
...
+		auto head = input.find("\"") + 1;
+		auto tail = input.find_last_of("\"");
+
+		if (head == std::string::npos || tail == std::string::npos) {
+			return "";
+		}
+		return input.substr(head, tail - head);
+	};
+
+	while ((pos = val.find(";")) != std::string::npos) {
```
DDRBoxman, line 313: I think this won't process the last item if it doesn't have a semicolon which they don't in the linked example.

DDRBoxman (changes requested):
Just a few comments, mostly around adding comments so this is easier to make sense of later.

plugins/obs-webrtc/whip-output.cpp
```diff
@@ -368,6 +424,24 @@ bool WHIPOutput::Connect()
 	CURLU *url_builder = curl_url();
 	auto last_location_header = location_headers.back();
 
+	struct curl_header *prev = nullptr;
+	struct curl_header *h = nullptr;
+	std::vector<rtc::IceServer> iceServers;
+	while ((h = curl_easy_nextheader(c, CURLH_HEADER, 0, prev))) {
+		if (strcmp("Link", h->name) == 0 ||
+		    strcmp("link", h->name) == 0) {
```
RytoEX, line 432: Are these the only two variations of "link" that are valid? If any case variation is valid, consider `astrcmpi` instead. If these are the only two valid forms, then this is fine.

RytoEX (commented):
Posed a question, but feedback is not blocking.

plugins/obs-webrtc/whip-output.cpp
Sean-Der: @DDRBoxman Fixed! Added a comment explaining what is happening

plugins/obs-webrtc/whip-output.cpp
Sean-Der: Added a note

plugins/obs-webrtc/whip-output.cpp
Sean-Der: Done!

plugins/obs-webrtc/whip-output.cpp
Sean-Der: Fixed. I now take rest of string on `npos` I tested with these values

```
< Link: <turn:192.168.1.93>; rel="ice-server"; username="username"; credential="password";
< Link: <turn:test>; rel="ice-server"; username="username"; credential="password"
```

plugins/obs-webrtc/whip-output.cpp
Sean-Der: Switched to `astrcmpi`. It makes the code simpler and more forgiving incase some does something like `LINK` and doesn't realize it.

Sean-Der:
Thank you so much @DDRBoxman @RytoEX I really appreciate the reviews :)

DDRBoxman (approved):
👍

RytoEX (approved):
LGTM

tt2468:
I'm noticing that this PR appears to be un-compilable on Ubuntu 22.04 due to the usage of `CURLH_HEADER` and Ubuntu's CURL version not having support. We have existing header parsing logic which was used to parse the Location header. I would say we will need to expand that existing logic instead of using `CURLH_HEADER`.

Sean-Der:
Fixed with https://github.com/obsproject/obs-studio/pull/10786

PR #10530, opened 2024-04-14
libobs: Remove unused variable in obs_source

norihiro:
### Description

Remove `last_audio_ts` in `struct obs_source`.

### Motivation and Context

The variable `last_audio_ts` was set but was never read.

### How Has This Been Tested?

OS: Fedora 38
Confirmed the build succeeded.

Also checked `git grep '\<last_audio_ts\>' libobs`.

### Types of changes

### Checklist:

### DIFF
--- a/libobs/obs-internal.h
+++ b/libobs/obs-internal.h
@@ -756,7 +756,6 @@ struct obs_source {
 	volatile bool timing_set;
 	volatile uint64_t timing_adjust;
 	uint64_t resample_offset;
-	uint64_t last_audio_ts;
 	uint64_t next_audio_ts_min;
 	uint64_t next_audio_sys_ts_min;
 	uint64_t last_frame_ts;
--- a/libobs/obs-source.c
+++ b/libobs/obs-source.c
@@ -1576,7 +1576,6 @@ static void source_output_audio_data(obs_source_t *source,
 		}
 	}
 
-	source->last_audio_ts = in.timestamp;
 	source->next_audio_ts_min =
 		in.timestamp + conv_frames_to_time(sample_rate, in.frames);
 


RytoEX:
It is not used for the async delay filter?
https://github.com/obsproject/obs-studio/blob/9c90c721a0b3134c00eb778d224b8ce97e1f1ec0/plugins/obs-filters/async-delay-filter.c#L202

norihiro:
> It is not used for the async delay filter?

The async delay filter has the same variable name but it is declared inside the filter.

https://github.com/obsproject/obs-studio/blob/9c90c721a0b3134c00eb778d224b8ce97e1f1ec0/plugins/obs-filters/async-delay-filter.c#L34

`struct obs_source::last_audio_ts` is defined in obs-internal.h so that the variable can be accessed only from libobs.

RytoEX:
This looks sane. @Lain-B ?

RytoEX:
> > It is not used for the async delay filter?
> 
> The async delay filter has the same variable name but it is declared inside the filter.

Yeah, I realized this after a second look.

RytoEX:
Actually, _can_ we even remove the member from the struct safely?

tt2468:
> Actually, _can_ we even remove the member from the struct safely?

Yes, since the struct definition is private.

libobs/obs-internal.h
```diff
@@ -756,7 +756,6 @@ struct obs_source {
 	volatile bool timing_set;
 	volatile uint64_t timing_adjust;
 	uint64_t resample_offset;
-	uint64_t last_audio_ts;
```
RytoEX, line 759: Can we even safely remove this from the struct definition?

Lain-B: Yep

Lain-B (approved):

PR #10536, opened 2024-04-14
NVENC Refactor

derrod:
### Description

> [!NOTE]
> This pull request is targeted for the next major version as it breaks some backwards compatibility (mainly Kepler GPUs).

- Adds refactored native NVENC implementation in separate `obs-nvenc` module
- Removes old native NVENC implementation and disables FFmpeg NVENC by default.

#### Features

1. Improved Capability Checks
    + `obs-nvenc-test[.exe]` now works on both Linux and Windows
    + Properties are now limited to what the hardware supports, making the silent retry/fallback unnecessary
    + Manual GPU blacklist is no longer required
2. New SDK (12.2) features
    + Split encoding (driver version 555 or higher) on dual-NVENC GPUs to run slower presets at higher resolutions, or achieve even higher frame rates at low resolutions
    + Ultra High Quality preset for HEVC (Turing+)
3. Additional features from old SDKs
    + B-Frames as reference to increase quality at practically no performance impact
    + Target Quality VBR mode (CQVBR) as an alternative to CQP mode
    + Maximum b-frames are now accurate (e.g. on Ada: 4 for H.264, 5 for HEVC, 7 for AV1)
4. Custom options field
    + Allows changing a number of often requested options such as AQ strength or lookahead depth without requiring more properties elements\*
    + Should help people migrate away from third-party plugins
5. Backwards compatibility
    + Old encoder IDs will stay functional until we decide to migrate users automatically

\* Current selection of options is mostly for experimentation, it will probably be cut down to only a few that are sensible but don't warrant their own UI elements.

### Motivation and Context

- Remove compatibility hack which has become a maintenance burden
- Remove NVENC code from FFmpeg module which it no longer depends on
- Support latest NVIDIA SDK and drivers
- Improved readability and maintainability

### How Has This Been Tested?

- RTX 4090 on Windows 10
- RTX 4090 on openSUSE Tumbleweed

> [!NOTE]
> This plugin is only compatible with the current cmake revision (3.0) on Linux and Windows, and will not work with legacy cmake.

### Types of changes

### Checklist:


plugins/obs-nvenc/cmake/dependencies.cmake
PatTheMav, line 1: In theory this could be rolled into the main CMakeLists file as it's not too unwieldy a block of dependencies. For `obs-ffmpeg` I split it into a separate file because it was a few more of them iirc.

derrod: Yeah makes sense, I'll combine them.

Bleuzen:
Somehow this never uses the texture encoder in my testing so far
```
info: [obs-nvenc] nv12/p010 not active, falling back to non-texture encoder
```
This always prints when I start recording. Note that I do use NV12, also tried different settings, nothing helped

Also the Lossless mode is broken
First it says
```
error: [obs-nvenc] init_encoder_h264: nv.nvEncInitializeEncoder(enc->session, &enc->params) failed: 8 (NV_ENC_ERR_INVALID_PARAM): Adaptive quantization is not supported with lossless encoding
```
which is easy to fix but would be nice if OBS could automatically disable (& hide if possible) the adaptive quantization option if lossless is selected

but even after disabling that it does error out with

```
error: [obs-nvenc] init_encoder_h264: nv.nvEncInitializeEncoder(enc->session, &enc->params) failed: 8 (NV_ENC_ERR_INVALID_PARAM): With lossless preset, RC Mode / Profile not supported with qpPrimeYZeroTransformBypassFlag.
error: Already in non_texture encoder, can't fall back further!
```

RytoEX:
> info: [obs-nvenc] nv12/p010 not active, falling back to non-texture encoder

Please see #10567.

Bleuzen:
> Please see #10567.

Can confirm it works with this

derrod:
Linux texture encoding and lossless mode for H.264 should work again now.

plugins/obs-nvenc/nvenc-opts-parser.c
```diff
@@ -0,0 +1,235 @@
+#include "nvenc-internal.h"
+
+#include <stdio.h>
+
+/* NVIDIA likes their bitfields and unfortunately the only way to do this is macros.
+ * So don't blame me for that.  */
+
+#define APPLY_BIT_OPT(opt_name, bits)                                         \
+	if (strcmp(opt->name, #opt_name) == 0) {                              \
```
PatTheMav, line 9: Why can't this be an inline function (per the comment above)? The macro is used quite often below which means that none of those steps can be debugged (except in raw assembly)?

derrod: You can't get a pointer to a specific value in a bitfield. I'm also not really concerned about the debugability of setting a value on a struct member.

sandr1x:
I checked the operation of B-frames as Reference via Elecard StramEye. For H.264 and HEVC everything works correctly. Disabled/middle/each modes

disableBadapt=1 disabled adaptive b-frames for P6,7 presets for H.264. Thank you! I think people may have difficulty where to take the arguments. Maybe add information in the manuals?

Bleuzen:
1.)
In CQVBR mode it is possible to set the maximum bitrate to 0 in the UI


My first expectation was it would probably just disable the maximum bitrate and use only the target quality to control the bitrate without a cap

Instead setting the max bitrate to 0 does for some reason cap it at some default value (depending on codec, resolution & fps), for example for 1440p 60 FPS HEVC it limits to 20 Mb/s:


... while it does go above those 20 Mb/s if the max bitrate is non-zero and higher.
A bit confusion from a users perspective, so probably either:
- disable the max bitrate when it is set to 0 (would prefer this)
- or set a lower limit for this setting so that the max bitrate could not be set to 0

---

2.)
It would be cool to also log the max bitrate when starting a stream/recording

```
info: [obs-nvenc: 'advanced_video_recording'] settings:
        codec:        AV1
        rate_control: CQVBR
        bitrate:      8000
--->    max_bitrate:      <---
        cq/cqp:       0
```

---

3.)
In the UI I think it makes more sense to swap these:


It always confuses my brain when switching rate control often during testing since the max bitrate is at another place in VBR mode

derrod:
Being able to set it to 0 is valid, and it should set it to 0 in the NVENC configuration as well. NVIDIA limits the bitrate in target quality mode even with max bitrate set to 0, this limit is based on the level. Seems for HEVC this ends up being much lower than H.264. It's one of the reasons we kept CQP around.

Bleuzen:
> Being able to set it to 0 is valid, and it should set it to 0 in the NVENC configuration as well.

I do agree from a technical perspective. However from a users perspective this is bad. You wouldn't expect a bitrate cap if max bitrate is set to 0 since there is nothing clearly indicating a fixed limit in this case. This could ruin recordings (bad quality) for new users who do not know this background info so I try to think of how this could be improved...

What do you think of extending the label


which is similar to the keyframe interval


this would at least make it clear that there is a limit so I consider this more user friendly

> NVIDIA limits the bitrate in target quality mode even with max bitrate set to 0, this limit is based on the level.

yeah just found that Handbrake and ffmpeg do the same, thanks nvidia I guess

> Seems for HEVC this ends up being much lower 

AV1 is also affected

> It's one of the reasons we kept CQP around.

I mean, CQ mode is not complete trash, as long as you manually set a high enough max bitrate

sandr1x:
This is a test build of obs, but it is unstable. In New World, I noticed that obs was unable to provide a stable framerate. If you start recording, the encoder is overloaded and stuttering occurs. With version 30 everything works fine.

derrod:
#10818 needs to be ported to here as well.

Also still need to figure out why the CUDA version dies when trying to use enhanced broadcasting with multiple of them, maybe a CUDA context management issue?

sandr1x:
The problem with unstable frame rates in OBS disappeared (my screenshot above), I assume, after updating the video drivers. I don't know if this is normal, but the test build of OBS does not reserve GPU resources for itself, so if the GPU load is around 99%, it causes the encoder to become overloaded. I used FurMark to test this. Otherwise, everything works great.

wanhongqing123:
你的邮件已收到，我会尽快的回复你

sandr1x:
The problem with the frame rate drop in OBS when the GPU is loaded has disappeared, but a new one has appeared. Now, if the GPU is loaded to its limit, the same thing happens, regardless of whether OBS is running idle or encoding video. Tested on assembly ver. 72c39ae4d. 

There is no such problem with the stable build.
image

plugins/obs-ffmpeg/obs-ffmpeg.c
```diff
@@ -59,11 +55,10 @@ extern struct obs_encoder_info hevc_vaapi_encoder_tex_info;
 #endif
 #endif
 
-#ifndef __APPLE__
+#ifdef ENABLE_FFMPEG_NVENC
```
RytoEX, line 58: Trying to understand what currently happens if you tried to compile on macOS and what would happen in this PR if you set `ENABLE_FFMPEG_NVENC` to `ON` on macOS.

RytoEX (changes requested):
Partial review.

plugins/obs-ffmpeg/obs-ffmpeg.c
derrod: I think right now this would just result in it checking if the nvenc encoder is available in FFmpeg (which it wouldn't be), all the OS-specific code is ifdef'd separately.

Lain-B (approved):

gregorytrentmartinjr:
This somehow breaks support for at least dedicated AMD encoders. Built master with this merge. AMD encoders were missing. Reverted this merge AMD encoders returned.

PR #10537, opened 2024-04-14
libobs: Mix audio of each source in a scene only once

derrod:
### Description

Only mix audio of each unique source in a scene once.

### Motivation and Context

Fixes a common pitfall as the constructive interference due to multiple scene items using the same source is not visible anywhere.

Based on discussions this should not happen.

### How Has This Been Tested?

Pure sine wave media source and ffmpeg `volumedetect` to make sure recordings after this fix have the same volume as a recording with only one scene item.

### Types of changes

### Checklist:

### DIFF
--- a/libobs/obs-scene.c
+++ b/libobs/obs-scene.c
@@ -1476,16 +1476,28 @@ static inline void mix_audio(float *p_out, float *p_in, size_t pos,
 		*(out++) += *(in++);
 }
 
-static bool scene_audio_render(void *data, uint64_t *ts_out,
-			       struct obs_source_audio_mix *audio_output,
-			       uint32_t mixers, size_t channels,
-			       size_t sample_rate)
+static inline struct scene_source_mix *get_source_mix(struct obs_scene *scene,
+						      struct obs_source *source)
+{
+	for (size_t i = 0; i < scene->mix_sources.num; i++) {
+		struct scene_source_mix *mix = &scene->mix_sources.array[i];
+		if (mix->source == source)
+			return mix;
+	}
+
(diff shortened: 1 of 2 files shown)


Warchamp7 (approved):

Warchamp7:
This solves duplicated scene items causing multiplied audio and LGTM.

Outside of this PR, we still need to solve scene items inside nested scenes causing the same issue.

Warchamp7 (approved):
I had concerns about half fixing this issue if nested scenes and groups still exhibited this issue. Rodney has now updated this PR to handle sources nested inside subscenes or groups. 

The only edge case not handled now is Show/Hide transitions on the Subscene or Group itself.

Fixing that would require a much larger overhaul of the audio subsystem and I believe it is uncommon enough to be acceptable.

libobs/obs-scene.c
```diff
@@ -1475,16 +1476,29 @@ static inline void mix_audio(float *p_out, float *p_in, size_t pos,
 		*(out++) += *(in++);
 }
 
-static bool scene_audio_render(void *data, uint64_t *ts_out,
-			       struct obs_source_audio_mix *audio_output,
-			       uint32_t mixers, size_t channels,
-			       size_t sample_rate)
+static inline struct scene_source_mix *get_source_mix(struct obs_scene *scene,
```
tt2468, line 1479: This should be able to be replaced with `da_find()` right?

derrod: No, `da_find` requires you to know the entire struct.

Lain-B (approved):

PR #10541, opened 2024-04-15
text-freetype2: Use cached glyph advances

kkartaltepe:
### Description

We had these advances laying around but didnt use them here. Loading the glyphs from the fonts is extremely slow, so lets avoid that since we take the time to cache these already.

### Motivation and Context

Makes this function ~4000x times faster. Everyone loves fast.

### How Has This Been Tested?

It didnt crash, but also this is used in the other width computation so it should be fine.

### Types of changes

### Checklist:

### DIFF
--- a/plugins/text-freetype2/text-functionality.c
+++ b/plugins/text-freetype2/text-functionality.c
@@ -570,12 +570,17 @@ uint32_t get_ft2_text_width(wchar_t *text, struct ft2_source *srcdata)
 		const FT_UInt glyph_index =
 			FT_Get_Char_Index(srcdata->font_face, text[i]);
 
-		load_glyph(srcdata, glyph_index, get_render_mode(srcdata));
-
 		if (text[i] == L'\n')
 			w = 0;
 		else {
-			w += slot->advance.x >> 6;
+			if (src_glyph) {
+				// Use the cached values.
+				w += src_glyph->xadv;
+			} else {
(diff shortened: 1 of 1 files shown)


RytoEX (approved):
This looks okay, but I am not really familiar with this patch of code. cc @norihiro ?

norihiro:
This looks good to me.

https://github.com/obsproject/obs-studio/blob/46c6efbd24cb0158286395e7765c9a3a46fb1440/plugins/text-freetype2/text-functionality.c#L579-L583

The function `get_ft2_text_width` always called after `cache_glyphs`. So, the cached width will be used for most cases. The `else` closure will be reached only if "Out of space trying to render glyphs" has happened in `cache_glyphs`. I tested both cases and the rendering looks ok.

RytoEX:
Tested this briefly on Windows 11 and it seems fine. Though, I did notice that there are apparently a few fonts that we cannot actually load (e.g., System, Courier), but that also occurs in 30.1.2 and is unrelated to the changes here.

RytoEX (changes requested):
Commit message nits:
* didnt -> didn't
* lets -> let's

This will make the first line of the commit message body too long, so the line wrapping will need to be adjusted.

RytoEX (approved):

Lain-B (approved):

PR #10546, opened 2024-04-16
libobs: Add support for multitrack, AV1, and HEVC captions

jhnbwrs:
### Description
Main components of this change

1 - We will now keep a per track list of caption data. When caption insertion is triggered we will add that to the list for each actively encoding video track. When doing interleaved sending we will pull the data from the caption data for the corresponding video. This ensures that captions get copied to all video tracks.

2 - Captions were always being inserted as AVC SEIs regardless of the codec being used for encoding. This would corrupt anything other than AVC. This type of caption insertion must be handled on a codec by codec basis, and if a codec isn't handled explicitly, no caption insertion will be attempted. This was tested by streaming to twitch with a popular caption plugin enabled, while having enhanced broadcasting configured with HEVC and AVC streams. The HEVC stream had no captions, while AVC showed all captions.

3 - Explicit handling for HEVC caption insertion. HEVC is very close to AVC, both use NALs, and a SEI NAL to add caption data. We simply need to add a 2 byte HEVC NAL header where there is a one byte AVC NAL header. I tested the same way I tested item number two, but this time was able to see captions on the HEVC stream that was sent to Twitch. I also recorded a stream that I was able to process with FFmpeg. Tracing headers showed the same number of user data registered ITUT-35 SEI NALs. Notably no errors from FFmpeg occurred processing the HEVC stream.

4 - Explicit handing for AV1 caption insertion. The data structure that is used to hold caption data is reference in the AV1 spec "5.8.2 - Metadata ITUT T35 Syntax". This defines a structure that has no differences from the HEVC/AVC structures for holding caption data, (See "D.1.6" User data registered by rec. ITU-T T.35 SEI message syntax" in the AVC spec, or D.2.6 in the HEVC spec). This structure is extracted from the SEI data and wrapped in a metadata OBU and inserted into the bitstream. Testing involved streaming AV1 to twitch and validating no video corruption. The twitch player has not yet implemented AV1 captions (as AV1 is not yet officially supported), but recording of the FLV and processing with FFmpeg shows no errors. You can find the itut t35 metadata OBUs while tracing headers with FFmpeg, showing that that data has been properly inserted into the stream.

### Motivation and Context
1 - Multitrack captioning was broken. 
2 - Captioning on any track other than H.264 would cause bitstream corruption.  There doesn't appear to be an open bug for this, but it was clearly a bug.

### How Has This Been Tested?
Being tested in the Twitch enhanced broadcasting beta currently. Details of dev testing which was done prior to releasing to the beta cohort can be found in this document:
https://www.dropbox.com/scl/fi/v5yku0a1j1m01fpk1xggo/OBS-Multitrack-Embedded-Caption-Support-Testing.paper?rlkey=yfv37vx7zs2d9m8bug9o6la2y&dl=0)

### Types of changes

### Checklist:

### DIFF
--- a/libobs/obs-output.c
+++ b/libobs/obs-output.c
@@ -1516,6 +1516,7 @@ static bool add_caption(struct obs_output *output, struct encoder_packet *out)
 	size_t size;
 	long ref = 1;
 	bool avc = false;
+	bool hevc = false;
 
 	/* Instead of exiting early for unsupported codecs, we will continue
 	 * processing to allow the freeing of caption data even if the captions
@@ -1523,6 +1524,10 @@ static bool add_caption(struct obs_output *output, struct encoder_packet *out)
 	 * the given codec. */
 	if (strcmp(out->encoder->info.codec, "h264") == 0) {
 		avc = true;
+#ifdef ENABLE_HEVC
+	} else if (strcmp(out->encoder->info.codec, "hevc") == 0) {
+		hevc = true;
+#endif
 	}
 
 	DARRAY(uint8_t) out_data;
@@ -1539,6 +1544,31 @@ static bool add_caption(struct obs_output *output, struct encoder_packet *out)
 		return false;
 	}
 
+#ifdef ENABLE_HEVC
+	uint8_t hevc_nal_header[2];
+	if (hevc) {
+		size_t nal_header_index_start = 4;
+		// Skip past the annex-b start code
+		if (memcmp(out->data, nal_start + 1, 3) == 0) {
+			nal_header_index_start = 3;
+		} else if (memcmp(out->data, nal_start, 4) == 0) {
+			nal_header_index_start = 4;
+
+		} else {
+			/* We shouldn't ever see this unless we start getting
+			 * packets without annex-b start codes. */
+			blog(LOG_DEBUG,
+			     "Annex-B start code not found. We may not "
+			     "generate a valid HEVC NAL unit header "
+			     "for our caption");
+			return false;
+		}
+		/* We will use the same 2 byte NAL unit header for the CC SEI,
+		 * but swap the NAL types out. */
+		hevc_nal_header[0] = out->data[nal_header_index_start];
+		hevc_nal_header[1] = out->data[nal_header_index_start + 1];
+	}
+#endif
 	sei_init(&sei, 0.0);
 
 	da_init(out_data);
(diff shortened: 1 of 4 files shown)


RytoEX (changes requested):
Nits while awaiting functional review.

Commit message nits...
First commit:
* interleaved send we will -> interleaved send, we will

Second commit:
* `H.264.  This` -> `H.264. This`
* The first sentence could use a bit of rewording. It currently reads a bit awkwardly.

Third commit:
* Adding -> Add

derrod (commented):
Seems fine overall, personally I would prefer that the OBU functions go into `obs-av1.c` rather than `obs-output.c`, and perhaps `extract_itut_t35_buffer_from_sei` should just be replaced with some updates to the vendored (and already modified) `libcaption`.

libobs/obs-output.c
```diff
@@ -1529,38 +1728,89 @@ static bool add_caption(struct obs_output *output, struct encoder_packet *out)
...
-	free(data);
-
-	obs_encoder_packet_release(out);
-
-	*out = backup;
-	out->data = (uint8_t *)out_data.array + sizeof(ref);
-	out->size = out_data.num - sizeof(ref);
+	if (avc || hevc || av1) {
+		if (avc || hevc) {
+			data = malloc(sei_render_size(&sei));
```
derrod, line 1745: Any reason to use `malloc` rather than `bmalloc` here?

libobs/obs-output.c
```diff
@@ -1529,38 +1728,89 @@ static bool add_caption(struct obs_output *output, struct encoder_packet *out)
...
+		out->data = (uint8_t *)out_data.array + sizeof(ref);
+		out->size = out_data.num - sizeof(ref);
+	}
 	sei_free(&sei);
-
-	return true;
+	if (avc || hevc || av1) {
+		return true;
+	}
+	return false;
```
derrod, line 1811: Could just write `return avc || hevc || av1;` instead

derrod (commented):
Seems fine overall, personally I would prefer that the OBU serialisation functions go into `obs-av1.c` rather than `obs-output.c`, and perhaps `extract_itut_t35_buffer_from_sei` should just be replaced with some updates to the vendored - and already modified - `libcaption`.

libobs/obs-output.c
jhnbwrs: only that it was previously done with malloc.

libobs/obs-output.c
```diff
@@ -1463,35 +1506,191 @@ static inline bool has_higher_opposing_ts(struct obs_output *output,
...
 	size_t size;
 	long ref = 1;
+	bool avc = false;
+	bool hevc = false;
+	bool av1 = false;
+
+	/* Instead of exiting early for unsupported codecs, we will continue
+	 * processing to allow the freeing of caption data even if the captions
+	 * will not be included in the bitstream due to being unimplemented in
+	 * the a given codec. */
```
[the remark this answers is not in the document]
jhnbwrs: done

libobs/obs-output.c
jhnbwrs: done

jhnbwrs:
> Seems fine overall, personally I would prefer that the OBU serialisation functions go into `obs-av1.c` rather than `obs-output.c`, and perhaps `extract_itut_t35_buffer_from_sei` should just be replaced with some updates to the vendored - and already modified - `libcaption`.

I have moved the av1 functions into the obs-av1.c as you suggested.  That is a bit cleaner for sure.

libobs/obs-output.c
jhnbwrs: Ok, I have changed this. As the malloc/free behavior was pre-existing, I changed it with small new commit to this PR.

libobs/obs-output.c
```diff
@@ -1529,38 +1639,86 @@ static bool add_caption(struct obs_output *output, struct encoder_packet *out)
...
+			 * forbidden_zero_bit       f(1)
+			 * nal_unit_type            u(6)
+			 * nuh_layer_id             u(6)
+			 * nuh_temporal_id_plus1    u(3)
+			 * }
+			 */
+			const uint8_t prefix_sei_nal_type = 39;
+			/* The first bit is always 0, so we just need to
+			 * save the last bit off the original header and
+			 * add the SEI NAL type */
```
RytoEX, line 1687:
```suggestion
			 * add the SEI NAL type. */
```

derrod:
I'll defer to @DDRBoxman if he has thoughts on whether or not we should just have a method in `libcaption` to get non-SEI formatted data directly rather than needing the extra function to extract it for AV1.

Still lgtm otherwise.

DDRBoxman (approved):
I'm fine with the metadata extraction from the sei staying in obs-output.c since we also so other sei manipulation there too. At this point I'm not too hot on really putting any code in the libcaption library even if it is caption related.

derrod (approved):

Lain-B (approved):

PR #10550, opened 2024-04-16
cmake: Deprecate legacy CMake for all platforms

RytoEX:
### Description

We now support the new, modern CMake path on all platforms. Encourage everyone to migrate by deprecating the legacy CMake path. The legacy CMake path will be removed in a future update.

### Motivation and Context

Want to give notice that the legacy CMake is deprecated.

### How Has This Been Tested?

Tested locally on Windows 11 to check that the deprecation warning was clearly displayed in the CMake log output.

### Types of changes

### Checklist:

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -34,21 +34,19 @@ if(CMAKE_HOST_SYSTEM_NAME MATCHES "(Darwin)" OR OBS_CMAKE_VERSION VERSION_GREATE
   return()
 endif()
 
-if(CMAKE_HOST_SYSTEM_NAME MATCHES "(Windows|Darwin)")
-  message(
-    DEPRECATION
-      "\n"
-      "============ LEGACY BUILD SYSTEM IS DEPRECATED ============"
-      "\n"
-      "You are using the legacy build system to build OBS Studio. "
-      "The legacy build system is unsupported and will be removed in the near future."
-      "\n"
-      "To migrate to the new build system, familiarize yourself with CMake presets "
-      "(https://cmake.org/cmake/help/latest/manual/cmake-presets.7.html) and create "
-      "a user preset with your customized build settings, inheriting from one of the default presets."
-      "\n"
-      "============ LEGACY BUILD SYSTEM IS DEPRECATED ============")
-endif()
+message(
+  DEPRECATION
+    "\n"
+    "============ LEGACY BUILD SYSTEM IS DEPRECATED ============"
+    "\n"
+    "You are using the legacy build system to build OBS Studio. "
+    "The legacy build system is unsupported and will be removed in the near future."
+    "\n"
+    "To migrate to the new build system, familiarize yourself with CMake presets "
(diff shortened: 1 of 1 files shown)


tytan652 (commented):
LGTM

tytan652 (approved):
LGTM

PatTheMav (approved):

Chiitoo:
Relatedly, has this been documented anywhere?

I don't see anything in these articles for example:

- https://github.com/obsproject/obs-studio/wiki/Building-OBS-Studio
- https://github.com/obsproject/obs-studio/wiki/Build-Instructions-For-Linux

RytoEX:
> Relatedly, has this been documented anywhere?
> 
> I don't see anything in these articles for example:
> 
> * https://github.com/obsproject/obs-studio/wiki/Building-OBS-Studio
> * https://github.com/obsproject/obs-studio/wiki/Build-Instructions-For-Linux

We haven't yet shipped a tag with this deprecation, so we haven't felt the need (or had the time away from other tasks) to make sweeping documentation changes. CMake prints a rather large warning if you try to use the old CMake. The new CMake invocations are documented on the Windows and macOS build pages where we have the most control and can provide presets. They are not provided on the Linux or FreeBSD pages because, if I recall correctly, we were advised that we should not provide presets for Linux/BSD aside from Ubuntu, so those pages are currently left as an exercise to the Linux/FreeBSD dev community.

If a need is found to place additional guidance on other wiki pages, we will do so. That said, the wiki pages are open to edits, so if you have edits you want to make, feel free.

Chiitoo:
I see.

I was somewhat expecting there to be /some/ mention of the change already I suppose, so people can be prepared and start transitioning (packaging wise especially).

I'll need to get more familiar with the presets, and if/when I do, and no one else has, I'll look into contributing to those notes.

One thing I'm not sure from taking a look today, is if something like this:

```
mycmakeargs=( --preset ubuntu -DENABLE_ALSA=true' )
```

is actually using the legacy style also though the warning is no longer seen.  It does seem to properly override the ubuntu presets at least to some extent, for now, but will it continue to do so?

In any case, that's a discussion for some other place I think.

Thank you!

RytoEX:
> I see.
> 
> I was somewhat expecting there to be /some/ mention of the change already I suppose, so people can be prepared and start transitioning (packaging wise especially).

We've already had a few people notice the CMake warning and switch over. I may post and pin a Discussion post later about it for added visibility.

> One thing I'm not sure from taking a look today, is if something like this:
> 
> ```
> mycmakeargs=( --preset ubuntu -DENABLE_ALSA=true' )
> ```
> 
> is actually using the legacy style also though the warning is no longer seen. It does seem to properly override the ubuntu presets at least to some extent, for now, but will it continue to do so?

The preset sets a CMake variable that forces the newer CMake path, so this would not use the legacy path.

Chiitoo:
Seems like a plan, and good to know indeed.

Thanks again!

PR #10560, opened 2024-04-17
UI: Support portable config mode on linux with new cmake version

Bleuzen:
### Description

This adds support for the `--portable` command line parameter when using the new cmake version on linux

I chose to only enable this when building with `ENABLE_RELOCATABLE` enabled because non-relocatable builds are most likely installed in a root owned directory (not writable for normal users). This mirrors how the `LINUX_PORTABLE` option behaved with the old cmake version

### Motivation and Context

The portable mode is missing since building with the new cmake version

### How Has This Been Tested?

- Built OBS with `OBS_CMAKE_VERSION=3` and `ENABLE_RELOCATABLE`
- Ran `obs --help` to check if the `--portable` option is available
- Executed OBS with the `--portable` option and checked where it places the config dir

### Types of changes

### Checklist:

### DIFF
--- a/UI/cmake/os-linux.cmake
+++ b/UI/cmake/os-linux.cmake
@@ -1,5 +1,6 @@
 target_sources(obs-studio PRIVATE platform-x11.cpp)
-target_compile_definitions(obs-studio PRIVATE OBS_INSTALL_PREFIX="${OBS_INSTALL_PREFIX}")
+target_compile_definitions(obs-studio PRIVATE OBS_INSTALL_PREFIX="${OBS_INSTALL_PREFIX}"
+                                              "$<$<BOOL:${ENABLE_RELOCATABLE}>:ENABLE_RELOCATABLE>")
 target_link_libraries(obs-studio PRIVATE Qt::GuiPrivate)
 
 if(TARGET OBS::python)
--- a/UI/obs-app.cpp
+++ b/UI/obs-app.cpp
@@ -2738,7 +2738,7 @@ static void load_debug_privilege(void)
 
 #define CONFIG_PATH BASE_PATH "/config"
 
-#if defined(LINUX_PORTABLE) || defined(_WIN32)
+#if defined(LINUX_PORTABLE) || defined(ENABLE_RELOCATABLE) || defined(_WIN32)
 #define ALLOW_PORTABLE_MODE 1
 #else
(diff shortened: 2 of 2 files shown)


Bleuzen:
This is a draft for now because there is one thing left:

With this and the new cmake, OBS places the config directory outside of the OBS portable directory currently. This is caused by the change of the directory structure.
When building with the old cmake the bin path looks like this:
`obs-studio-portable/bin/64bit/obs`
with the new cmake the bin path changed to:
`obs-studio-portable/bin/obs`

There is one less directory to go up now (`64bit` dir missing).
I'm not sure if this change was intentional or is an oversight with the new cmake version since in the latest windows version I looked at (from GH actions) the `64bit` is still there.
Thus I don't know if this should be accounted for or not in this PR

PatTheMav:
> There is one directory less to go up now (`64bit` dir missing in the path). However this directory structure change might be just an oversight. I looked at the latest windows version (from GH actions) and the `64bit` dir is still there So I guess this was not intentional and should be fixed in a separate PR

The `64bit` subdirectory is superfluous now because there are only 64bit versions of OBS Studio and no 32-bit versions exist next to 64-bit ones. It also was never valid under FHS rules, so the non-portable builds never used it and the new relocatable variant does not use a custom directory structure to allow for removal of all the additional code added for the old "portable" variant once the legacy build system is obsolete.

The only reason this still exists on Windows is because too many usecases (and OBS internals) rely on `bin/64bit/obs.exe` being a thing and the structure not changing over there.

kkartaltepe:
I think the new relocatable build covers the existing portable use-case, except we no longer specify the custom config directory and leave that to the user. Users can continue to specify their old portable config with the XDG_CONFIG_HOME when launching obs, instead of using a `--portable` flag when launching obs.

Bleuzen:
> So the change was very much intentional.
> 
> The only reason this still exists on Windows is because too many usecases (and OBS internals) rely on `bin/64bit/obs.exe` being a thing and the structure not changing over there.

Okay, thanks for clarifying



> with the XDG_CONFIG_HOME env var

Can confirm this works, cool, thanks for the hint. Although I did prefer the shorter `-p` when launching from command line it is easy enough to make a wrapper script for this to still make it short to type

So I think we can close this

PR #10565, opened 2024-04-18
linux-pipewire: Unify ScreenCast source as Screen Capture

tytan652:
### Description

Allow to select both main type of ScreenCast (monitor, window) with only one source type rather than adding a source type per ScreenCast type.

This change is made as a new Screen Capture source type which obsoletes the use of the previous Screen Capture (monitor-only) and Window Capture.

NOTE: An obsolete OBS source type is only hidden from the add list, already existing sources will still work as expected.

#### Why a new id rather than modifying display capture ?

- I can't check or predict all portals implementation, and if one them create a restore token based on the asked types rather than only selected sources. Not changing id might introduce unwanted breaks on now obsoleted type.

- Coherence in the id

#### Screenshots

GNOME monitor-only screencast selector (shown by obsoleted screen capture):
Screenshot from 2024-04-21 01-48-09

GNOME window-only screencast selector (shown by obsoleted window capture):
Screenshot from 2024-04-21 01-48-24-obfuscated

GNOME monitor+window screencast selector (shown by new screen capture):

Screenshot from 2024-04-22 23-15-35

### Motivation and Context

The way the ScreenCast portal was designed does not require a OBS source type per ScreenCast type, but the feature was introduced as 1:1 towards X11 implementations.

But unlike on X11, there is also other types like virtual type for virtual monitor that can be used by Desktop Environment (or one for applications once I finish my app to app portal).

And I think rather than adding a third OBS source type, it would better to unify them like macOS has one OBS source type for SCK.

### How Has This Been Tested?

- Added a Screen Capture with normal build
- Open OBS build with this PR, the Screen Capture (v1) was still working
- Adding a new Screen Capture allow to chose between the three ScreenCast type.

### Types of changes

### Checklist:


norihiro:
A line below contains more than 72 characters.
> Allow to select any type of ScreenCast (monitor, window, virtual) with only

Warchamp7:
Can someone get screenshots or video of the unified ScreenCaptureKit source workflow on macOS works? I'd like to try and keep the experience as similar as we reasonably can across operating systems.

tytan652:
> Can someone get screenshots or video of the unified ScreenCaptureKit source workflow on macOS works? I'd like to try and keep the experience as similar as we reasonably can across operating systems.

https://obsproject.com/kb/display-capture-sources#macos-sck

macOS Screen Capture properties window

GeorgesStavracas (approved):

tytan652:
Window capture is now 
not obsoleted which do not show in the add source menu.

kkartaltepe:
Seems like it continues to work as expected on sway as well.

GeorgesStavracas (approved):

PR #10578, opened 2024-04-21
CI: Rewrite obs-crowdin-sync in Python & add to repository

Vainock:
### Description
Rewrites the obs-crowdin-sync repository in Python and adds the action to this repository.

Changes that come along with it:
- Translates the `UI/xdg-data/com.obsproject.Studio.metainfo.xml.in` file.
- Creates a pull request instead of pushing directly: https://github.com/Vainock/obs-studio/pull/3
- Changes the name of `English` to `English (USA)` for `en-US`.
- Lowers the requirements for new languages to be added to the `UI/data/locale.ini`.
- Host the `Language` string in a separate `Language Name` file on Crowdin, remove it here.

### Motivation and Context
The obs-crowdin-sync repository was too much for what it solves.

This solution is better for integration testing, easier to extend, easier for error analysis in the log because you don't have a huge bundled script and much cleaner to look at.

Pushing directly to the repo is bad, creating a PR is safer and transparent.

### How Has This Been Tested?
Locally and https://github.com/Vainock/obs-studio/pull/3. The `update-crowdin-locales` action was tested in a obs-browser fork.

### Types of changes

### Checklist:

### DIFF
--- a/.github/scripts/utils.py/update-crowdin-locales.py
+++ b/.github/scripts/utils.py/update-crowdin-locales.py
@@ -1,0 +1,101 @@
+import os
+import subprocess
+import logging
+
+from crowdin_api import CrowdinClient
+from fnmatch import fnmatch
+
+
+CROWDIN_PROJECT_ID = 51028
+CROWDIN_API_TOKEN = os.environ["CROWDIN_PAT"]
+DEFAULT_LOCALE = "en-US"
+
+crowdin_api = CrowdinClient(token=CROWDIN_API_TOKEN, project_id=CROWDIN_PROJECT_ID, page_size=500)
+
+logging.basicConfig()
+logger = logging.getLogger()
+logger.setLevel(logging.INFO)
+
+
+def get_updated_locales(cwd=".") -> list[str]:
+    COMMITS_TO_CHECK = f"{os.environ['GITHUB_EVENT_BEFORE']}..{os.environ['GITHUB_SHA']}"
+    updated_locales = subprocess.check_output(["git", "diff", "--name-only", COMMITS_TO_CHECK], cwd=cwd)
+    updated_locales = updated_locales.decode().strip().splitlines()
+    updated_locales = [updated_locale for updated_locale in updated_locales if updated_locale.endswith(f"/{DEFAULT_LOCALE}.ini")]
+
+    return updated_locales
+
+
+def get_dir_id(name: str) -> int:
+    crowdin_api.source_files.list_directories(filter=name)["data"][0]["data"]["id"]
+
+
+def add_to_crowdin_storage(file_path: str) -> int:
+    return crowdin_api.storages.add_storage(open(file_path))["data"]["id"]
+
+
+def update_crowdin_file(file_id: int, file_path: str) -> None:
+    storage_id = add_to_crowdin_storage(file_path)
+    crowdin_api.source_files.update_file(file_id, storage_id)
+
+    logger.info(f"{file_path} updated on Crowdin.")
+
+
+def create_crowdin_file(file_path: str, name: str, directory_name: str, export_pattern: str):
+    storage_id = add_to_crowdin_storage(file_path)
+
+    crowdin_api.source_files.add_file(storage_id,
+                                      name,
+                                      directoryId=get_dir_id(directory_name),
+                                      exportOptions={"exportPattern": export_pattern})
+
+    logger.info(f"{file_path} created on Crowdin in {directory_name} as {name}.")
+
+
+def upload(updated_locales: list[str]) -> None:
+    export_paths_map = dict()
+
+    for source_file_data in crowdin_api.source_files.list_files()["data"]:
+        source_file = source_file_data["data"]
+
+        if "exportOptions" not in source_file:
+            continue
+
+        export_path: str = source_file["exportOptions"]["exportPattern"]
+        export_path = export_path[1:]
+        export_path = export_path.replace("%file_name%", os.path.basename(source_file["name"]).split(".")[0])
+        export_path = export_path.replace("%locale%", DEFAULT_LOCALE)
+
+        export_paths_map[export_path] = source_file["id"]
+
+    for file_path in updated_locales:
+        if not os.path.exists(file_path):
+            logger.warning(f"Unable to find {file_path} in working directory.")
+            continue
+
+        path_parts = file_path.split("/")
+
+        if file_path in export_paths_map:
+            crowdin_file_id = export_paths_map[file_path]
+            update_crowdin_file(crowdin_file_id, file_path)
+        elif fnmatch(file_path, f"plugins/*/data/locale/{DEFAULT_LOCALE}.ini"):
+            create_crowdin_file(file_path, f"{path_parts[1]}.ini",
+                                path_parts[0], "/plugins/%file_name%/data/locale/%locale%.ini")
+        elif fnmatch(file_path, f"UI/frontend-plugins/*/data/locale/{DEFAULT_LOCALE}.ini"):
+            create_crowdin_file(file_path, f"{path_parts[2]}.ini",
+                                path_parts[1], "/UI/frontend-plugins/%file_name%/data/locale/%locale%.ini")
+        else:
+            logger.error(f"Unable to create {file_path} on Crowdin due to its unexpected location.")
+
+
+updated_locales = [str]
+
+if "SUBMODULE_NAME" in os.environ:
+    updated_locales = get_updated_locales(f"plugins/{os.environ['SUBMODULE_NAME']}")
+else:
+    updated_locales = get_updated_locales()
+
+if len(updated_locales) != 0:
+    upload(updated_locales)
+else:
+    logger.error("Unable to find updated locales.")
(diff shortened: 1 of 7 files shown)


Vainock:
@tytan652 Should it also update `UI/cmake/linux/com.obsproject.Studio.desktop` instead of `UI/xdg-data/com.obsproject.Studio.desktop`?

tytan652:
Yes

I think the change about going away from xdg-data has completely gone unseen. We found out through metainfo regression in CI.

Vainock:
Ready for review.

.github/scripts/utils.py/update-crowdin-locales.py
```diff
@@ -0,0 +1,88 @@
...
+            create_crowdin_file(locale_path, f"{path_parts[1]}.ini",
+                                path_parts[0], "/plugins/%file_name%/data/locale/%locale%.ini")
+        elif fnmatch(locale_path, f"UI/frontend-plugins/*/data/locale/{DEFAULT_LOCALE}.ini"):
+            create_crowdin_file(locale_path, f"{path_parts[2]}.ini",
+                                path_parts[1], "/UI/frontend-plugins/%file_name%/data/locale/%locale%.ini")
+        else:
+            logger.error(f"Unable to create {locale_path} on Crowdin due to its unexpected location.")
+
+
+if "CHANGED_FILES" in os.environ:
```
PatTheMav, line 81: Don't rely on magic behaviour based on environment variables and don't do anything but variable or function declaration in the root scope - if this is supposed to be a shell command, then it should be designed as one:

* Use a `main` function
* Accept variables via command-line arguments
* Provide a usage hint when called without necessary arguments

Feel free to copy what you need for that from https://github.com/obsproject/obs-studio/blob/master/.github/scripts/utils.py/check-jsonschema.py.

.github/scripts/utils.py/update-crowdin-locales.py
```diff
@@ -0,0 +1,88 @@
...
+import json
+
+from crowdin_api import CrowdinClient
+from fnmatch import fnmatch
+
+CROWDIN_PROJECT_ID = 51028
+CROWDIN_API_TOKEN = os.environ["CROWDIN_PAT"]
+DEFAULT_LOCALE = "en-US"
+
+crowdin_api = CrowdinClient(token=CROWDIN_API_TOKEN, project_id=CROWDIN_PROJECT_ID, page_size=500)
```
PatTheMav, line 12: Does this have to be a global?

.github/scripts/utils.py/update-crowdin-locales.py
```diff
@@ -0,0 +1,88 @@
...
+
+logging.basicConfig()
```
PatTheMav, line 14: Logger should be initialized in a `main` function ,ideally with support for a log level argument.

.github/scripts/utils.py/update-crowdin-locales.py
```diff
@@ -0,0 +1,88 @@
...
+logger = logging.getLogger()
+logger.setLevel(logging.INFO)
+
+
+def get_dir_id(name: str) -> int:
+    crowdin_api.source_files.list_directories(filter=name)["data"][0]["data"]["id"]
```
PatTheMav, line 20: Is `list_directories` assured to always return a dictionary with an non-empty array with key `data`, which itself will contain objects that _always_ contain an object at key `data` which has a key `id`?

If the answer is "no" or "maybe" use a chain of `get` calls instead:

```Python
directories = crowdin_api.source_files.list_directories(filter=name).get("data", [])

if len(directories) > 0:
  id = directories[0].get("data", {}).get("id", None)
  if id is not None:
    return id
  else:
    raise Exception("No id field on directory data")
else:
  raise Exception("No crowdin source file directories found")
```

.github/scripts/utils.py/update-crowdin-locales.py
```diff
@@ -0,0 +1,88 @@
...
+
+
+def add_to_crowdin_storage(file_path: str) -> int:
+    return crowdin_api.storages.add_storage(open(file_path))["data"]["id"]
```
PatTheMav, line 24: See above re. avoidance of `ValueError`s.

.github/scripts/utils.py/update-crowdin-locales.py
```diff
@@ -0,0 +1,88 @@
...
+def upload(updated_locales: list[str]) -> None:
+    export_paths_map = dict()
+
+    for source_file_data in crowdin_api.source_files.list_files()["data"]:
+        source_file = source_file_data["data"]
+
+        if "exportOptions" not in source_file:
+            continue
+
+        export_path: str = source_file["exportOptions"]["exportPattern"]
```
PatTheMav, line 54: Multi-dimensional direct access will lead to `ValueError` if key doesn't return dictionary with the expected key. Use chained `get` calls to ensure the access does not fail and use a fallback like `None` to check for errors.

.github/scripts/utils.py/update-translations.py
PatTheMav, line 1: See the comments for `update-crowdin-locales.py` regarding globals, lack of a `main` function and command-line arguments, as well as error-prone deep access of nested dictionary keys without proper error handling (at the very least a `try.. except` block catching a `ValueError` would be necessary).

.github/scripts/utils.py/update-translations.py
```diff
@@ -0,0 +1,292 @@
...
+    for file_path, translations in build.items():
+        with open(file_path, "w") as file:
+            output = ""
+            for key, translation in translations.items():
+                output += f'{key}="{translation}"\n'
+            file.write(output)
+
+
+def write_locale_file(build: CrowdinBuild) -> None:
+    FILE_PATH = "UI/data/locale.ini"
```
PatTheMav, line 106: Don't use upper case naming for anything but actual constants.

_If_ these are actually constants, they should either be declared global in the file scope or  encapsulated (the latter is more elegant because it doesn't pollute the file scope, but not strictly necessary):

```Python
class OBSCrowdinConstants:
  FILE_PATH = ""
[..]

def write_locale_file(build: CrowdinBuild) -> None:
  [..]
  with open(OBSCrowdinConstants.FILE_PATH, encoding="utf-8") as file:
    [...]
```

.github/scripts/utils.py/update-crowdin-locales.py
Vainock: If the dictionary access fails, there is no reason to recover from there on. If there is no `exportPattern` in `exportOptions`, then the configuration on Crowdin needs to be updated. The code depends a lot on the configuration on Crowdin and assumes a specific structure. I support making the code flexible to the API results, but have trouble understanding what makes encapsulating all possible situations with error handling because of possible misconfiguration on Crowdin or broken responses beneficial.
Adding optional debug logging for all responses makes in my head more sense.

PatTheMav: Every time you access an API or process data you're not in control of, you have to ensure to either safely evaluate the state of the data before accessing it to stay in control of program execution (otherwise you openly rely on undefined state and you can make no assumptions of the states your code can end up in).

If _any_ issue while accessing that data is considered a possible error, at the very least wrap the access in a `try: [...] except (ValueError, KeyError) as e:` block because the code can potentially break on both of these errors because of an invalid access violation. You can then handle the error by logging an appropriate error function, clean up, and exit the program with an error code under your control.

.github/scripts/utils.py/update-crowdin-locales.py
Vainock: When I try...except this theses object accesses, what should I do in the except block? I caught the error, but should I raise an exception again? Just log it?

My only idea would be to try...except the upload() function in the main function (https://github.com/Vainock/obs-studio/blob/ba977d75bd6f8e1546f0d2824d6714110073af31/.github/scripts/utils.py/update-crowdin-locales.py#L101) for all `KeyError`s and `ValueError`s. This would cover all instances where these exceptions can occur in that file. My problem is that I can only return the error code from the main function. Do you have an idea for how to solve that?

PatTheMav: You don't have to catch exceptions right where they happen but where you can handle them, so it'd be fine to not handle them on lower levels and catch them on the top-most level and log the error there and return from the main function with an error code.

The exception to that would be if you want to be able to more specifically identify the error to have more specific error messages (so you want to differentiate _which_ interaction resulted in a `ValueError`). This would also enable you to `raise` errors with error messages in the inner functions instead of having to log error messages, because you're catching them in `main` and print the error there.

.github/scripts/utils.py/update-crowdin-locales.py
WizardCM: I have checked this, and the Crowdin documentation [explicitly states](https://support.crowdin.com/developer/api/v2/#tag/Source-Files/operation/api.projects.directories.getMany) that `["data"][0]["data"]` are all guaranteed ("required") in the response, **if** the response is `200`.

PatTheMav: Wouldn't be the first time a response header states `200` and the result is an empty string, or it's truncated, or it's malformed, or it's actually a piece of HTML.

There is no good reason _not_ to write code that is guaranteed to succeed instead of code that just doesn't seem to fail.

WizardCM: True enough. What I also just realised is that function is missing a `return`.

PR #10581, opened 2024-04-21
obs-ffmpeg: Fix nvidia compat ver missing definition

wanhongqing123:
### Description
Some nvidia compat definitions were missing, this change adds them

### Motivation and Context
nvidia encoder on some machines may fail to start

### How Has This Been Tested?
I have test on windows 10 19045.

### Checklist:

### DIFF
--- a/plugins/obs-ffmpeg/obs-nvenc.c
+++ b/plugins/obs-ffmpeg/obs-nvenc.c
@@ -47,6 +47,18 @@
 #define COMPATIBILITY_VERSION \
 	(NVENC_COMPAT_MAJOR_VER << 4 | NVENC_COMPAT_MINOR_VER)
 
+#define NV_ENC_CREATE_BITSTREAM_BUFFER_COMPAT_VER NVENCAPI_STRUCT_VERSION(1)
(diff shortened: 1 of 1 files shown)


derrod:
None of these define a different version than what's in the currently supported NVENC headers.

wanhongqing123:
你的邮件已收到，我会尽快的回复你

RytoEX:
As @derrod has said, none of these should be necessary. Are you actually running into a specific issue without these defined? I've double checked and none of these struct versions has changed between SDK 11.1 and 12.1.

wanhongqing123:
I found that some definitions use Compat, and some still use definitions in NVIDIA header files. I think this will make the file more confusing, so I think it is better to unify it. If the driver version is the same as the version on the actual machine, use Compat, otherwise use the definition in the NVIDIA driver @RytoEX @derrod

RytoEX:
> I found that some definitions use Compat, and some still use definitions in NVIDIA header files. I think this will make the file more confusing, so I think it is better to unify it. If the driver version is the same as the version on the actual machine, use Compat, otherwise use the definition in the NVIDIA driver @RytoEX @derrod

It really doesn't matter. The struct versions haven't changed, so this is effectively a no-op.

derrod:
Closing as per above, this is pointless.

PR #10590, opened 2024-04-24
libobs: Remove extra gs_flush calls

kkartaltepe:
### Description

These calls introduce pipeline stalls in OpenGL on low end machines which can contribute to a fairly significant 10-30% of total rendering time. On average (since not every frame stalls) this is ~5% improvement with gpu time decreasing from 6.3ms to 6.0ms when encoders are active.

The driver already needs to synchronize events within the context so these calls are not needed for correctness for OpenGL.

### Motivation and Context

Squeeze out a bit more perf on very low end devices where encoders could go faster if the rendering was faster. It matters more for intel than AMD in my tests but the intel is far weaker than my AMD device.

### How Has This Been Tested?

Rendering vaapi and qsv texture encoders on amd and intel devices and I didnt notice any differences in output.

If I'm reading the D3D11 pages correctly for flush it seems it wont even introduce the gpu-gpu sync that it does on OpenGL so these are only trying to submit commands faster in Windows? I would appreciate testing on windows to know if these were providing any significant performance improvements there.

### Types of changes

### Checklist:

### DIFF
--- a/libobs/obs-video.c
+++ b/libobs/obs-video.c
@@ -605,7 +605,6 @@ static inline void render_video(struct obs_core_video_mix *video,
 			copy_surfaces = video->copy_surfaces_encode;
 			channel_count = 1;
 #endif
-			gs_flush();
 		}
 
 		if (video->gpu_conversion) {
@@ -614,7 +613,6 @@ static inline void render_video(struct obs_core_video_mix *video,
 		}
 
 		if (gpu_active) {
-			gs_flush();
 			output_gpu_encoders(video, raw_active);
 		}
 


derrod:
Seems to work fine on Windows as well.

kkartaltepe:
There was another question about vram usage on windows without the flush, so maybe its worth checking that too.

derrod:
I don't see any difference in VRAM usage.

RytoEX (approved):
If I recall correctly, this was discussed off-thread as a bit of a mystery as to why it was originally written this way. I'm willing to merge this, but we'll probably have to keep an eye out for reports of weird behavior (either increased resource usage or performance regressions).

PR #10597, opened 2024-04-25
obs-ffmpeg: Fix deadlock on shutting down muxer

exeldro:
### Description
The things that happen to get this deadlock:
- the io thread reads `shutdown_requested` as false
- the main thread sets `shutdown_requested` to true
- the main thread gets the `data_mutex`
- the main thread signals `new_data_available_event`
- the main thread releases the `data_mutex`
- the main thread waits for the io thread to exit
- the io thread gets the `data_mutex`
- the io thread resets `new_data_available_event` because `chunk_used < 65536 && !shutting_down`
- the io thread releases the `data_mutex`
- the io thread waits for `new_data_available_event`

with the changes the `new_data_available_event` cannot be send from the main thread between the reading of `shutdown_requested` the getting of the  `data_mutex` in the io thread

### Motivation and Context
Got a memory dump Fenrir from a muxer that was stuck.
the main thread is in `ffmpeg_mux_free` waiting for the io thread to close.
the io thread is waiting for `new_data_available_event`

### How Has This Been Tested?
Muxer still works as expected 

### Types of changes

### Checklist:

### DIFF
--- a/plugins/obs-ffmpeg/ffmpeg-mux/ffmpeg-mux.c
+++ b/plugins/obs-ffmpeg/ffmpeg-mux/ffmpeg-mux.c
@@ -686,9 +686,9 @@ static void *ffmpeg_mux_io_thread(void *data)
 
 		// Loop to write in chunk_size chunks
 		for (;;) {
-			shutting_down = os_atomic_load_bool(&ffm->io.shutdown_requested);
-
 			pthread_mutex_lock(&ffm->io.data_mutex);
+
+			shutting_down = os_atomic_load_bool(&ffm->io.shutdown_requested);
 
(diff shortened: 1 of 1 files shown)


derrod:
I believe @notr1ch had some concern with this (unfortunately only noted on Discord so I have a hard time finding it right now), but I had adopted this change in the threaded writer for #10608 and it has worked flawlessly so far. So from my perspective this should be fine.

notr1ch:
My concern was why I left this outside in the first place, there was a reason for it but I didn't leave any comments and looking at it now I really don't know why. Perhaps I refactored whatever it was and this was missed.

PR #10600, opened 2024-04-25
deps/media-playback and win-dshow: Prioritize CUDA HW decoder (for 8K playback)

moocowsheep:
### Description

8K60 files would not play back at full rate in media source on Windows on Nvidia cards, due to D3D11VA and DXVA2 not being able to decode fast enough. This pull request puts CUDA decoder at highest priority, to enable high resolution / frame rate file decode at full speed.

### Motivation and Context

I have only been able to play back 8K HEVC video files on MacOS at full speed...this brings the Windows version at feature parity.

### How Has This Been Tested?

I tested the media source with HW Decode on 2 Windows machines, one with an Nvidia L4 and one with an Nvidia ADA 6000. This does not appear to affect any other code.

### Types of changes

### Checklist:

### DIFF
--- a/libobs/media-io/video-frame.c
+++ b/libobs/media-io/video-frame.c
@@ -14,362 +14,136 @@
     You should have received a copy of the GNU General Public License
     along with this program.  If not, see <http://www.gnu.org/licenses/>.
 ******************************************************************************/
-
+#include <assert.h>
 #include "video-frame.h"
 
-#define ALIGN_SIZE(size, align) size = (((size) + (align - 1)) & (~(align - 1)))
+#define HALF(size) ((size + 1) / 2)
+#define ALIGN(size, alignment) \
+	*size = (*size + alignment - 1) & (~(alignment - 1));
 
-/* messy code alarm */
-void video_frame_init(struct video_frame *frame, enum video_format format,
-		      uint32_t width, uint32_t height)
+static inline void align_size(size_t *size, size_t alignment)
 {
-	size_t size;
-	size_t offsets[MAX_AV_PLANES];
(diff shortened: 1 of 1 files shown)


tt2468:
Could you provide some profiling results of using CUDA decode versus d3d11 with your 8k60 video?

tt2468:
What about d3d11 is slower than CUDA, to where using CUDA is the difference between being able to play 8k60 or not? Have you investigated the decode and GPU download stages to determine where the actual performance degradation is occuring?

notr1ch:
CUDA tends to use parts of the GPU that are also used by by other things like games so the performance impact needs extensive testing. I think this should be scoped so that only affected videos use CUDA, or as an advanced option only.

moocowsheep:
Doing investigations and then going to resubmit properly. The CUDA decoder uses the nvdec, (I don't believe it uses the CUDA cores, it's a misnomer) but I'll get more info on the actual performance implications.

Fenrirthviti:
In the future, you can just force push your branch and the PR will update so you don't have to create a new one.

This is one reason why we request PRs not from the master branch of a fork :)

moocowsheep:
Duly noted :) I'm a bit of a GitHub novice! Fixing all for the next one

naieto:
> Doing investigations and then going to resubmit properly. The CUDA decoder uses the nvdec, (I don't believe it uses the CUDA cores, it's a misnomer) but I'll get more info on the actual performance implications.
That's why not all NVIDIA cards with CUDA cores can decode h265. Although some ffmpeg filters such as scaling, overlay, etc., use CUDA.

PR #10601, opened 2024-04-26
win-dshow: Fix compiling with FFmpeg 7.0

RytoEX:
### Description

Use "new" (FFmpeg 5.1+) channel layout API as needed.

### Motivation and Context

Want to be able to compile against FFmpeg 7.0.

Follow-up to:
* #7462

### How Has This Been Tested?

Built locally on Windows 11. I'm not clear on how to activate this code path, so further review would be appreciated.

### Types of changes

### Checklist:

### DIFF
--- a/plugins/win-dshow/ffmpeg-decode.c
+++ b/plugins/win-dshow/ffmpeg-decode.c
@@ -265,8 +265,13 @@ bool ffmpeg_decode_audio(struct ffmpeg_decode *decode, uint8_t *data,
 
 	audio->samples_per_sec = decode->frame->sample_rate;
 	audio->format = convert_sample_format(decode->frame->format);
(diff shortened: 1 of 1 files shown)


pkviet (approved):
The fix is correct.
Sidenote left here for the sake of documenting it: 
However the original code is not ideal because it assumes a given number of channel is enough to set the layout.
For instance, if the source has octogonal audio (8.0), it should be resampled first to 7.1 since obs only knows 7.1.
Another example is the various 5.1 arrangements (5.1 side, 5.1 wideback, 5.1 back ...) which will be treated as identical to our 5.1.
So the right call would be to call a swresampler instead of assuming the layout is ours.
BUT in practice this is exceedingly rare; we use the most common layouts so let's keep the code simple.

derrod (approved):

PR #9653, opened 2023-10-02
UI: Composable Themes System

derrod:
### Description

This PR introduces a system that allows "layering" of theme files, i.e. a base theme can provide styling, a variant theme the colour palette, and optional high-contrast adjustments or user customisation can adjust it to suit the user's needs.
Themes will be able to specify variables that can be overwritten and extended by child themes, to provide much more flexibility than the Qt palette system currently in use, as it allows arbitrary values to be specified.

Bonus feature: Loaded theme can be watched for changes and automatically reloaded when edited (hidden config option `[Appearance]` -> `AutoReload`).

**Also see: [Overview of syntax and design goals](https://gist.github.com/derrod/dd7711887f2e9d1efe6b51e06ebb3a48)** (will be migrated to wiki/design repo once finalised).

Additionally, theme settings are moved to a new "Appearance" tab in the settings menu, which will be expanded upon in the future:
2023-11-14_11-21-58_7gwq8h

### Motivation and Context

This PR is based around conversations with @Warchamp7 on how we can improve theming in OBS and better accomodate new UI elements in the future.

An additional benefit for third-party themes is that they can inherit stock OBS themes and adjust them to their needs without having to copy everything into their own file. This also should make them forward-compatible as styling for the UI elements introduced in PRs such as #8446 will automatically be inherited and adjusted to the themes colour palette.

### How Has This Been Tested?

Built a basic test base theme (`Yami.obt`) and variant theme (`Yami_test.ovt`) to validate parsers, UI, and composing to be functional.

### Types of changes

- New feature (non-breaking change which adds functionality)

### Checklist:

- [x] My code has been run through [clang-format](https://github.com/obsproject/obs-studio/blob/master/.clang-format).
- [x] I have read the [**contributing** document](https://github.com/obsproject/obs-studio/blob/master/CONTRIBUTING.rst).
- [x] My code is not on the master branch.
- [x] The code has been tested.
- [x] All commit messages are properly formatted and commits squashed where appropriate.
- [x] I have included updates to all appropriate documentation.

PR #10602, opened 2024-04-26
UI: Remove unnecessary ProxyStyle usage

Warchamp7:
### Description
This removes an unnecessary usage of QProxyStyle for the source context bar media QSlider and the destruction/recreation of audio mixer controls.

### Motivation and Context
It appears we abused the styleHint approach of using `SH_Slider_AbsoluteSetButtons` which causes it to treat clicks in the slider area as absolute points to jump to. While it works, this is meant to be a style implementation, such as an OS difference.

Instantiating a new custom style _just_ for this slider does not seem to be a proper usage of styleHints, and due to our heavy usage of QSS which is itself a QProxyStyle, causes some weird behaviour we've worked around in places.

Instead, this slider now handles the actual mouse events to perform the same behaviour, without having to force an update of the context bar.

Additionally, work in #9653 appears to have solved the need to completely destroy and recreate the Volume Controls on theme change.

### How Has This Been Tested?
Built OBS. Used the context bar slider for media sources. Swapped themes a bunch.

### Types of changes

### Checklist:


UI/media-slider.cpp
```diff
@@ -51,3 +63,34 @@ bool MediaSlider::eventFilter(QObject *obj, QEvent *event)
...
+{
+	QStyleOptionSlider opt;
+	initStyleOption(&opt);
+
+	const QRect groove = style()->subControlRect(
+		QStyle::CC_Slider, &opt, QStyle::SC_SliderGroove, this);
+	const QRect handle = style()->subControlRect(
+		QStyle::CC_Slider, &opt, QStyle::SC_SliderHandle, this);
+
+	int pos, sliderMin, sliderMax, handleLength;
```
RytoEX, line 77:
```suggestion
	int pos;
	int sliderMin;
	int sliderMax;
	int handleLength;
```
Also, move this to the beginning of the function with the other declarations.

RytoEX (changes requested):
I'm not sure what the Volume Control changes have to do with the rest of the PR:
> UI: Remove ProxyStyle for context bar media Slider

I feel like this should either be two commits with a rescoped PR description/title or two PRs:
1. Media Slider
2. Volume Control

UI/media-slider.cpp
Warchamp7: Fixed

Warchamp7:
Commits have been adjusted. Will need testing on mac/Linux and testing at different scaling to ensure the mouse logic is solid.

UI/media-controls.hpp
```diff
@@ -45,10 +45,10 @@ private slots:
...
 	void on_durationLabel_clicked();
 
-	void MediaSliderClicked();
-	void MediaSliderReleased();
-	void MediaSliderHovered(int val);
-	void MediaSliderMoved(int val);
+	void AbsolutesliderClicked();
+	void AbsolutesliderReleased();
+	void AbsolutesliderHovered(int val);
+	void AbsolutesliderMoved(int val);
```
gxalpha, line 51: These should probably be capatilized `AbsoluteSlider*`

sebastian-s-beckmann (commented):
Just to have it written on the PR:
The proposed alternative to the custom absolute click handling would be `SH_Slider_AbsoluteSetButtons` on `OBSProxyStyle`, personally I'd prefer that over custom handling, especially since it appears that native Windows sliders also behave that way.

RytoEX:
> Just to have it written on the PR: The proposed alternative to the custom absolute click handling would be `SH_Slider_AbsoluteSetButtons` on `OBSProxyStyle`, personally I'd prefer that over custom handling, especially since it appears that native Windows sliders also behave that way.

I've been unable to find a slider in Windows 11 that behaves otherwise. The native volume, brightness, and text size controls all move the slider exactly where I've clicked. None of them have any stepping behavior. Maybe it was this way in Windows XP/Vista? I'm more shocked that apparently no one has filed a QTBUG over this particular behavior.

UI/window-basic-main.cpp
```diff
@@ -11157,19 +11154,6 @@ float OBSBasic::GetDevicePixelRatio()
 
 void OBSBasic::ThemeChanged()
```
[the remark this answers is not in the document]
Warchamp7: I think that's out of scope for this PR. We don't know if any plugins rely on this event.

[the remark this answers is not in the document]
gxalpha: I'm not suggesting we should remove or not send the event. Either way, I can PR that cleanup after this gets merged.

RytoEX:
> I've been unable to find a slider in Windows 11 that behaves otherwise. The native volume, brightness, and text size controls all move the slider exactly where I've clicked. None of them have any stepping behavior. Maybe it was this way in Windows XP/Vista? I'm more shocked that apparently no one has filed a QTBUG over this particular behavior.

Just for visibility, I filed [QTBUG-125154](https://bugreports.qt.io/browse/QTBUG-125154) for this. All the native Windows 11 sliders and WinUI 3 slider examples I found moved to absolute position on left-click or snapped to the nearest tick (the latter can be configured).

RytoEX:
> Just to have it written on the PR: The proposed alternative to the custom absolute click handling would be `SH_Slider_AbsoluteSetButtons` on `OBSProxyStyle`, personally I'd prefer that over custom handling, especially since it appears that native Windows sliders also behave that way.

For what it's worth, I mentioned off-thread that I would also be fine with this. We actually already do OS-specific stuff in the styleHint call in OBSProxyStyle, so we could ifdef it for Windows if we were inclined (I don't know what "expected" slider behavior is on Ubuntu/Linux/etc.):
https://github.com/obsproject/obs-studio/blob/af685be40d9af51e0b2f594c94212507327b245d/UI/obs-proxy-style.cpp#L86-L89

If the primary things that we want are in the refactoring commits (the first two), we could split the third commit off and deal with that question later. Not sure how troublesome that is (if code in #10625 relies on the third commit).

In a worst case, we can merge this as-is and reconsider the implementation (OBSProxyStyle vs. events) later.

RytoEX:
> > I've been unable to find a slider in Windows 11 that behaves otherwise. The native volume, brightness, and text size controls all move the slider exactly where I've clicked. None of them have any stepping behavior. Maybe it was this way in Windows XP/Vista? I'm more shocked that apparently no one has filed a QTBUG over this particular behavior.
> 
> Just for visibility, I filed [QTBUG-125154](https://bugreports.qt.io/browse/QTBUG-125154) for this, at least for their new Windows 11 style in Qt 6.7. All the native Windows 11 sliders and WinUI 3 slider examples I found moved to absolute position on left-click or snapped to the nearest tick (the former is default behavior, but the latter can be configured).

Following up on this, I double checked the slider examples in the WinUI 3 Gallery on Windows 11 and all of them have the default behavior to set the absolute value of the slider to the StepValue nearest to the position clicked. If the StepFrequency is 1 (the default), this is indistinguishable from "left-click sets absolute value" behavior. Sliders will snap to the nearest StepValue (mostly only noticeable if you have a very small range or a very large step value) or Tick (if configured to do so). None of the examples step towards the clicked position.

I did check a few sliders on Windows 10 and they had varying behaviors. The Volume Mixer steps in the direction from the slider's current value to the cursor (this can cause it to step towards or over the cursor repeatedly). The volume widget from the system tray (not sure if it has a proper name), behaves like a Windows 11 slider. The Levels slider in the Device Properties of Sound settings behaves like Volume Mixer.

That said, the "Windows 11" / "modernwindows" style in Qt 6.7+ is following the Windows 11 and WinUI 3 guidance, which does not have the "step-towards-click" behavior that Windows 10 does. Qt has merged changes for Qt 6.7.2+ to make left-click set the absolute position of the slider. The change for 6.7.1 is still pending.
https://codereview.qt-project.org/c/qt/qtbase/+/560228
https://codereview.qt-project.org/c/qt/qtbase/+/561096
https://codereview.qt-project.org/c/qt/qtbase/+/561216
https://github.com/qt/qtbase/commit/392c5fab815cbe41df9e25e50e4d296e7e1606f6
https://github.com/qt/qtbase/commit/78bf5f7ccd63afef3e7bd6fbd7e58022b32b5b09

As far as I know, Qt's Fusion style (what is used outside of Windows and macOS), will still use left-click for "page-step" behavior.

We can either be opinionated about how we want sliders to behave in our application, or we can defer to the OS/style/framework. As far as I can tell, our issue is that we want the context bar's slider (MediaSlider/AbsoluteSlider) to behave a certain way, but do not necessarily want all sliders to behave that way.

Are all of these commits blocking #10625, or is it just the first commit? If it's just the first commit (UI: Move VolumeSlider widget to volume control), we could split that out from the other two, merge that since I think that's uncontroversial, and then discuss the remaining changes separately. The second commit (UI: Rename MediaSlider to AbsoluteSlider) might also be uncontroversial. If help is needed splitting stuff up, I can assist.

On the ProxyStyle changes, the main motivation for removing it here is because it fixes some odd behavior on theme switching, is that correct? Additionally, we always want some sliders (in this case, the context bar media seek bar) to behave a specific way on all platforms in all Qt Styles, so it must be made to do so. Do I understand that correctly?

Warchamp7:
> We can either be opinionated about how we want sliders to behave in our application, or we can defer to the OS/style/framework. As far as I can tell, our issue is that we want the context bar's slider (MediaSlider/AbsoluteSlider) to behave a certain way, but do not necessarily want all sliders to behave that way.

This is correct

> Are all of these commits blocking https://github.com/obsproject/obs-studio/pull/10625, or is it just the first commit? If it's just the first commit (UI: Move VolumeSlider widget to volume control), we could split that out from the other two, merge that since I think that's uncontroversial, and then discuss the remaining changes separately. The second commit (UI: Rename MediaSlider to AbsoluteSlider) might also be uncontroversial. If help is needed splitting stuff up, I can assist.

Only the first one really

RytoEX (approved):
Going to make a judgment call and say this is fine. If the change from QProxyStyle to mouseEvent handling causes problems, I can handle the revert of the third commit.

While looking into this, I found [QTBUG-125110](https://bugreports.qt.io/browse/QTBUG-125110), which mentions the behavior that this PR seems to describe. The response there so far is:
> Palette and font propagation follows different rules when a style sheet is active, see Qt::AA_UseStyleSheetPropagationInWidgetStyles in https://doc.qt.io/qt-6/qt.html#ApplicationAttribute-enum. Generally, mixing style sheets with other styling mechanism will result in hard-to-predict outcomes.

PR #10603, opened 2024-04-26
UI: Update media source time labels while seeking

Warchamp7:
### Description
Update the time and duration labels in the context bar when changing the seek position

https://github.com/obsproject/obs-studio/assets/1554753/a044da07-351e-43e5-9cf2-681e42236f7e

### Motivation and Context
Better feedback of user interaction

### How Has This Been Tested?
Seeked back and forth through multiple media files.

### Types of changes

### Checklist:

### DIFF
--- a/UI/media-controls.cpp
+++ b/UI/media-controls.cpp
@@ -160,6 +160,7 @@ void MediaControls::MediaSliderReleased()
 			obs_source_media_set_time(source, GetSliderTime(seek));
 		}
 
+		UpdateLabels(seek);
 		seek = lastSeek = -1;
 	}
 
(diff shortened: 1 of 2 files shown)


Lain-B (approved):

WizardCM (approved):
My only very minor concern is that `MediaSliderMoved` might run more than 60 times a second, which could cause the whole UI to lag while dragging. Wouldn't mind a side-by-side, but otherwise I'm OK with this change.

RytoEX:
A future problem/concern/refactor might be to redo this so that we store the media's duration once when it's loaded and refer back to the stored value rather than pulling a strong reference (sometimes two!) and fetching the source's duration every time we update labels.

PR #10605, opened 2024-04-26
libobs: Use floating point textures when loading floating point image files

Lain-B:
### Description
Allows the ability to open image files with floating point precision as floating point textures without unnecessarily downsampling to unsigned 8bit integer precision.

### Motivation and Context
Someone told me that they were trying to load an image file with 32bit floating point precision per channel to do some sort of advanced shader stuff, and that important precision information was being lost. They didn't know why. I noticed that they were being converted away to 8bit unsigned integer RGBA unnecessarily despite the fact that we supported the texture format natively, so I just fixed it in like 5 minutes.

### How Has This Been Tested?
Tested it with a .tiff file containing 32bit floating point precision RGBA file, opened it with an image source, checked the debugger to make sure the code path was being executed, checked the visible result, everything checked out.

### Types of changes

### Checklist:

### DIFF
--- a/libobs/graphics/graphics-ffmpeg.c
+++ b/libobs/graphics/graphics-ffmpeg.c
@@ -370,6 +370,7 @@ static void *ffmpeg_image_reformat_frame(struct ffmpeg_image *info,
 
 	if (info->format == AV_PIX_FMT_BGR0) {
 		data = ffmpeg_image_copy_data_straight(info, frame);
+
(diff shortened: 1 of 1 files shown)


tytan652:
Use `AV_PIX_FMT_RGBAF32LE` and `AV_PIX_FMT_RGBAF16LE`, need to be guarded behind libavutil version from FFmpeg 6.0

PR #10621, opened 2024-04-29
obs-webrtc: Add null terminator to codec array

kc5nra:
This fixes an issue where when the MAX_CODECS length was equal to the amount of supported codecs (3), it would leave the list without a null terminator and crash when iterating over the elements.

### Description
Addresses a bug where a non-null terminated list of strings was iterated over into invalid memory

### Motivation and Context
Crashes on some machines depending on if the memory layout had a null accidentally after the list.

### How Has This Been Tested?
Tested on Mac and Windows, no longer crashes in release mode.

### Types of changes
Bug fix

### Checklist:

### DIFF
--- a/plugins/obs-webrtc/whip-service.cpp
+++ b/plugins/obs-webrtc/whip-service.cpp
@@ -1,7 +1,7 @@
 #include "whip-service.h"
 
-const char *audio_codecs[MAX_CODECS] = {"opus"};
-const char *video_codecs[MAX_CODECS] = {"h264", "hevc", "av1"};
+const char *audio_codecs[] = {"opus", nullptr};
+const char *video_codecs[] = {"h264", "hevc", "av1", nullptr};
 
 WHIPService::WHIPService(obs_data_t *settings, obs_service_t *)
 	: server(),
--- a/plugins/obs-webrtc/whip-service.h
+++ b/plugins/obs-webrtc/whip-service.h
@@ -2,8 +2,6 @@
 #include <obs-module.h>
 #include <string>
 
-#define MAX_CODECS 3
-
 struct WHIPService {
 	std::string server;
 	std::string bearer_token;


Sean-Der (commented):
Works great for me!

RytoEX (changes requested):
Wrap the commit message body at 72 characters, not 60.

Is this a problem for all other `const char *something[]` instances across the codebase? Is that type even appropriate here? Should this just be a std vector/array or an enum (as it is elsewhere in the codebase) instead?

I'm fine getting a crash fixed quickly and merging this as-is (after fixing nits). However, I feel like this is not our first problem with this implementation, and I think a few reviewers have expressed discontent with it.

kc5nra:
> Wrap the commit message body at 72 characters, not 60.
> 
> Is this a problem for all other `const char *something[]` instances across the codebase? Is that type even appropriate here? Should this just be a std vector/array or an enum (as it is elsewhere in the codebase) instead?
> 
> I'm fine getting a crash fixed quickly and merging this as-is (after fixing nits). However, I feel like this is not our first problem with this implementation, and I think a few reviewers have expressed discontent with it.

It would be an API breaking change to address it differently. 

```
        info.get_supported_video_codecs = [](void *) -> const char ** {
		return video_codecs;
	};
	info.get_supported_audio_codecs = [](void *) -> const char ** {
		return audio_codecs;
	};
```

So the services getters would have to change to change the type

kc5nra:
But yes, it would be a problem for all the API functions that rely on (*next_item == 0) for determining last element

RytoEX:
> > Wrap the commit message body at 72 characters, not 60.
> > Is this a problem for all other `const char *something[]` instances across the codebase? Is that type even appropriate here? Should this just be a std vector/array or an enum (as it is elsewhere in the codebase) instead?
> > I'm fine getting a crash fixed quickly and merging this as-is (after fixing nits). However, I feel like this is not our first problem with this implementation, and I think a few reviewers have expressed discontent with it.
> 
> It would be an API breaking change to address it differently.

For what it's worth, I'm not saying to change the return type there. You can build the string however you like (see `rtmp-common`). I'm suggesting that relying on a raw `const char *something[]` here and doing strcmp calls is perhaps not ideal, but that is outside the scope of this PR.


> But yes, it would be a problem for all the API functions that rely on (*next_item == 0) for determining last element

If there are other places where this is problematic, we would welcome fixes for those.

kc5nra:
Makes sense! I think I just read your first message too fast, my fault! I'll ponder a better approach for this server in another PR

Thanks!

plugins/obs-webrtc/whip-service.cpp
```diff
@@ -1,7 +1,7 @@
 #include "whip-service.h"
 
-const char *audio_codecs[MAX_CODECS] = {"opus"};
-const char *video_codecs[MAX_CODECS] = {"h264", "hevc", "av1"};
+const char *audio_codecs[] = {"opus", nullptr};
+const char *video_codecs[] = {"h264", "hevc", "av1", nullptr};
```
RytoEX, line 4: Since we're returning this across a C/C++ boundary, is `nullptr` valid here?

kc5nra: https://github.com/obsproject/obs-studio/blame/dfd5798e6ca8dc72532e96239b8229f032d37d4f/UI/window-basic-main-dropfiles.cpp#L18 I just copied another example of "C stuff in C++ land". Nullptr will coerce correctly here, but happy to switch it to NULL. (I was just trying to keep nullptr's in cpp files).

derrod: `nullptr` will be fine here. And once we go to C23 we'll have it there as well!

tt2468 (approved):

PR #10625, opened 2024-04-30
UI: Update volume meter appearance

Warchamp7:
### Description
Adjusts the layout and appearance of the audio mixer volume controls. Removes minor ticks, dims the decibel indication text, widens the meters for lower channel counts, and adds 10db ticks to the slider.

**Before:**
image

image

**After:**
image

image

Depends on #10602, or at least will prefer to merge that first so that this can be rebased against it.

### Motivation and Context
Better use of space, less visual noise.

### How Has This Been Tested?
Messed with audio sources and the audio sliders

### Types of changes

### Checklist:

### DIFF
--- a/UI/data/themes/Yami.obt
+++ b/UI/data/themes/Yami.obt
@@ -129,6 +129,10 @@
 
     --spinbox_button_height: calc(var(--input_height) - 2px);
 
+    --volume_slider: calc(calc(6px + var(--font_base_value)) / 2);
+    --volume_slider_box: calc(var(--volume_slider) * 4);
+    --volume_slider_label: calc(var(--volume_slider) * 6);
+
     --scrollbar_size: 12px;
 
     /* Inputs / Controls */
@@ -295,13 +299,8 @@ SourceTree QWidget {
 
 /* Misc */
 
-QAbstractItemView,
-QStackedWidget#stackedMixerArea QWidget {
+QAbstractItemView {
     background-color: var(--bg_base);
-}
-
-QStackedWidget#stackedMixerArea QScrollBar {
-    background-color: var(--grey6);
 }
 
 QToolTip {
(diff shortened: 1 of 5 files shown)


notr1ch:
I'm not a huge fan of the larger bars and handles, it feels a bit cheap and Fisher Pricey but maybe I need to get used it. The new UI seems a bit cramped with everything so close together. I feel like the icons for properties and mute should also be the same size if they are going next to each other.

tt2468:
After hearing some responses from broadcast folks, I would suggest adding minor ticks back between -25dBFS and -5dBFS.

Warchamp7:
> I'm not a huge fan of the larger bars and handles, it feels a bit cheap and Fisher Pricey but maybe I need to get used it. The new UI seems a bit cramped with everything so close together. I feel like the icons for properties and mute should also be the same size if they are going next to each other.

I think you'd be hard pressed to find another audio application that has the meters as thin as we used to have them. 

The icons are intended to be the same size. There is an open PR to address that inconsistency



> After hearing some responses from broadcast folks, I would suggest adding minor ticks back between -25dBFS and -5dBFS.

I'll try to play around with this

UI/data/themes/Yami.qss
tytan652, line 1: I think this file should not be there, since it got removed a while ago.

lextra2:
dB font should have a higher contrast
maybe give it the same colour as the volume slider? since they're connected?

Lain-B:
I'll have to defer to everyone else for discussion about all the details but I feel like colorblind people will definitely appreciate this PR. I also like how it adds the volume level lines to the slider.

klemun:
Thoughts on aligning the 3 vertical dots on these red vertical lines instead? Otherwise, looks good!
Screenshot 2024-05-05 120225

RytoEX:
Rebased this PR. Hopefully I got the changes correct.

UI/data/themes/Yami.qss
RytoEX: Removed during rebase.

RytoEX:
After rebasing, I seem to have encountered this odd behavior:
Screenshot 2024-05-24 031650

Warchamp7:
> After rebasing, I seem to have encountered this odd behavior: Screenshot 2024-05-24 031650

I've fixed this issue

derrod:
Overall I like some of the improvements, but man do the kebab and speaker buttons stick out like a sore thumb, particularly in the horizontal one.

I'd prefer if the speaker remained without the button frame, not sure what your motivation was for adding that, but it looks "wrong" to me. We don't really do that with other symbolic buttons like the visibility or lock icons either, so it feels inconsistent with the sources dock.

notr1ch:
I agree that the mute icon should not look like a button (its behavior is more akin to a checkbox).

RytoEX (approved):
Apparently, I previously viewed these files and had no notes. LGTM

Warchamp7:
Noting the above feedback and not discounting it by merging, but going to see what the sentiment is during the beta cycle.

If it's a common point of feedback I'll revert the styling changes as a fix.

prollygon:
I use the vertical format for the Audio Mixer. Do the thicker audio meters make everything less compact? I feel like my audio sources don't fit neatly in my layout anymore. I always felt that the old design was still spacious, so I'm sad to see it get thicker. Streamers will often have many audio sources and the UI should prioritize fitting as many of them in the smallest amount of space as possible. I would love an option to make the audio mixer even more compact and space-efficient than before.

Fenrirthviti:
> ...the UI should prioritize fitting as many of them in the smallest amount of space as possible.

I disagree. Giving UI elements the space they need to breathe is very important when it comes to UI/UX design. Things in OBS have been cramped for a while. While we're aware a few select cases where people are just upset that the change has caused them to have to adjust things, we feel that "I don't like it" isn't really a valid justification for immediate change, though it's still valid feedback and we're taking all reports in to account as we continue to iterate.

In any event, this PR addresses a unintended functional change, which is a specific edge case where the changes *did* result in the meters taking up more space. In most cases, they will take up the same, or even less, despite appearances.

KAMIGENO:
I genuinely hate this change.
Please revert it---or at least please give us the option to choose between the 2 appearances.
I actively used the markers to get my audio _just_ how I want it.

TY.

KAMIGENO:
> Things in OBS have been cramped for a while.
image

KAMIGENO:
I would still like the 1 dB markers back, please.

PR #10626, opened 2024-05-01
UI: Restore MuteCheckBox indeterminate state icon in Yami

tytan652:
### Description

Restore MuteCheckBox indeterminate state icon in Yami

Release (before composable theme):
Screenshot from 2024-05-01 10-21-19

Master:
Screenshot from 2024-05-01 10-20-59

PR:
Screenshot from 2024-05-01 10-29-00

### Motivation and Context

Found weird that the button was empty rather than yellowish.

### How Has This Been Tested?

Launched OBS with a source with audio set to to none of the track.

### Types of changes

### Checklist:

### DIFF
--- a/UI/data/themes/Yami.obt
+++ b/UI/data/themes/Yami.obt
@@ -1517,6 +1517,10 @@ MuteCheckBox::indicator:checked {
     image: url(theme:Dark/mute.svg);
 }
 
+MuteCheckBox::indicator:indeterminate {
+    image: url(theme:Dark/unassigned.svg);
+}
+
 MuteCheckBox::indicator:unchecked {
     image: url(theme:Dark/settings/audio.svg);
 }


tytan652:
Added all states like `checked` and `unchecked` and the Light variant was using the same icon as base Yami so no change in it.

Light with PR:
Screenshot from 2024-05-01 10-49-51

UI/data/themes/Yami.obt
```diff
@@ -1545,6 +1549,18 @@ MuteCheckBox::indicator:unchecked:disabled {
...
+}
+
+MuteCheckBox::indicator:indeterminate:focus {
+    image: url(theme:Dark/unassigned.svg);
+}
+
+MuteCheckBox::indicator:indeterminate:disabled {
+    image: url(theme:Dark/unassigned.svg);
+}
+
```
Warchamp7, line 1563: These lines are not necessary. The explicit hover/focus states are needed for checked/unchecked states because we set overrides for those for normal checkboxes

Warchamp7 (changes requested):

UI/data/themes/Yami.obt
tytan652: Okay, I removed them.

GeorgesStavracas (approved):

PR #10633, opened 2024-05-03
UI: Add (eRTMP) multitrack video output

palana:
~~Note that this PR includes a bunch of merges from currently open PRs and a FIXME commit to help with testing, which will go away as soon as the dependency PRs are merged~~

### Description
This implements multitrack video support for Twitch Enhanced Broadcasting and Amazon IVS Multitrack Video

### Motivation and Context
Multitrack video/Enhanced Broadcasting allows sending multiple video tracks/renditions from the users machine, resulting in generally higher quality renditions at lower bitrates compared to single track streaming with server side transcoding, since e.g. losses roundtripping frames through another decode/encode step are removed.
Enhanced Broadcasting/multitrack video uses server side configuration to provide easier onboarding for users, and balanced values for the number of renditions and their bitrates.

### How Has This Been Tested?
This has been tested in the Twitch Enhanced Broadcasting beta

### Types of changes

### Checklist:


UI/window-basic-settings-stream.cpp
```diff
@@ -547,13 +630,37 @@ void OBSBasicSettings::on_customServer_textChanged(const QString &)
...
+	bool /*enabled*/)
+{
+	UpdateMultitrackVideo();
+}
+
+void OBSBasicSettings::on_multitrackVideoConfigOverrideEnable_toggled(
+	bool /*enabled*/)
+{
+	UpdateMultitrackVideo();
+}
```
tytan652, line 662: Turn `UpdateMultitrackVideo` into a slot and connect it to the widget.

UI/window-basic-settings-stream.cpp
```diff
@@ -547,13 +630,37 @@ void OBSBasicSettings::on_customServer_textChanged(const QString &)
...
+
```
tytan652, line 663: Turn `UpdateMultitrackVideo` into a slot and connect to widgets.

UI/goliveapi-postdata.cpp
```diff
@@ -0,0 +1,50 @@
...
+	using json = nlohmann::json;
+
+	json post_data;
+	post_data["service"] = "IVS";
+	post_data["schema_version"] = "2023-05-10";
+	post_data["authentication"] = streamKey.toStdString();
+
+	auto &capabilities = post_data["capabilities"];
+	capabilities["gpu"] = system_gpu_data();
+	capabilities.update(system_info(), true);
```
tytan652, line 22: The user must be able to reject having their system information being sent to a third party.

UI/obs-frontend-api/obs-frontend-api.h
```diff
@@ -62,6 +62,8 @@ enum obs_frontend_event {
 	OBS_FRONTEND_EVENT_SCENE_COLLECTION_RENAMED,
 	OBS_FRONTEND_EVENT_THEME_CHANGED,
 	OBS_FRONTEND_EVENT_SCREENSHOT_TAKEN,
+
+	OBS_FRONTEND_EVENT_STREAMING_PREPARING,
```
tytan652, line 66: With how messy service and streaming output codebase is becoming, I would be heavily against adding any Frontend API unless it's for untangling the whole broadcast code and the whole "golive" stuff and put most of it in the service code.

UI/CMakeLists.txt
```diff
@@ -84,6 +84,23 @@ target_sources(
...
 
+target_sources(
+  obs-studio
+  PRIVATE # cmake-format: sortable
+          goliveapi-censoredjson.cpp
+          goliveapi-censoredjson.hpp
+          goliveapi-network.cpp
+          goliveapi-network.hpp
+          goliveapi-postdata.cpp
+          goliveapi-postdata.hpp
```
tytan652, line 95: "goliveapi" is clearly Twitch/Amazon made thing. I would really prefer that most of this goes to where it belong: in the service plugin side.

We already have the Broadcast feature from YT to untangle to a proper service with an agnostic API to manage broadcast, this would worsen the situation.

tytan652 (changes requested):
This feature is clearly not implemented in a future-proof way, it tangles even more service behavior in the UI that should be plugin-ified (with APIs).

Continuously shoehorning service integrations was already a mistake, and this PR goes to the same wrong direction rather than trying to go to a better direction.

UI/goliveapi-postdata.cpp
koolscooby: > The user must be able to reject having their system information being sent to a third party.

The user can reject having their system information being sent by not enabling Enhanced Broadcasting or Multitrack Video in the UI.

In other words: this information is never sent to the destination service if the "Enable Enhanced Broadcasting" or "Enable Multitrack Video" are unchecked. In the UI, the user is presented with a checkbox to enable Enhanced Broadcasting or Multitrack Video.

If the feature is enabled via the UI checkbox, then `GetClientCapabilities` is required to fetch the configuration settings and further, the `GetClientCapabilities` API requires this information to vend the best possible, safe ladder for every setup.

tytan652: This is bad, the user needs be aware of this and make an explicit choice to send their data.
Hiding this detail is really bad, IIRC the data that are sent can be assigned as PII.

Fenrirthviti: The user is made aware of this with an explicit opt-in and notification when they enable the feature. Nothing is hidden, and I'm not sure why you're assuming it is.

tytan652: I didn't see the notification part when the check box is toggled in settings, maybe I'm looking at the wrong place.

Sorry then.

tytan652: Found it, it's a label. I'm sorry.

To me, it looks like a bad pattern to block a feature based on if the user wants to share data or not.
But this is out of scope.

This can be put as resolved.

UI/window-basic-settings-stream.cpp
palana: changed right after https://github.com/obsproject/obs-studio/pull/10633#event-12771455631

UI/obs-frontend-api/obs-frontend-api.h
palana: since there's currently nothing using this event it should be ok to just remove it; let me know if you want me to do that

UI/window-basic-settings-stream.cpp
tytan652: Can be put as resolved.

UI/obs-frontend-api/obs-frontend-api.h
palana: Removed in a fixup commit after https://github.com/obsproject/obs-studio/pull/10633#event-12771825219

RytoEX: Resolving.

RytoEX:
Rebased first. Now pushing cleanups:
1. The multitrack audio commits don't actually need to be in here because the later commits don't touch the same code. VOD Track just won't work without them, but removing the commits reduces some noise here.
2. We can drop "UI: Keep weak stream output ref in status bar" since it was handled in #10680.

palana:
@RytoEX, @Warchamp7: moved the remaining service settings into groupboxes after <https://github.com/obsproject/obs-studio/pull/10633#event-12870619591>
image
not super happy with the naming of either of the new groupboxes

RytoEX (changes requested):
Started a review a while back and forgot to submit it. Submitting it now so get this one feedback item raised.

UI/qt-helpers.hpp
```diff
@@ -0,0 +1,46 @@
...
+	* QTBUG-119406 and QTBUG-119103 affect Qt 6.6.0 and are fixed in Qt 6.6.2 and 6.7.0+.
+	* QTBUG-119579 and QTBUG-119810 affect Qt 6.6.1 and are fixed in Qt 6.6.2 and 6.7.0+.
+	* QTBUG-117918 is the only strange one that seems to possibly affect all Qt 6.x versions
+	* until 6.6.2, but only in Debug builds.
+	*
+	* To fix this, move relevant QFutures to another thread before resuming
+	* on main thread for affected Qt versions
+	*/
+#if (QT_VERSION >= QT_VERSION_CHECK(6, 0, 0)) && \
+	(QT_VERSION < QT_VERSION_CHECK(6, 2, 0))
```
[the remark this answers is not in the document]
palana: huh, yea, that was meant to be 6.6.2; will push a fix in a moment

palana:
> I tried to autosquash locally, but it looks like some of the autosquash commits are trying to apply changes to commits before the actual base changes the autosquash commits try to fixup. Could you squash the commits?
> 
> Per @Warchamp7 , the UI itself is okay. @Lain-B and I are still doing code review.

@RytoEX: commits are squashed, apologies for the unclean fixup commits

Lain-B (approved):
Thanks for rebasing. There are a few minor things, such as Linux/macOS just not being implemented at all and failing with "GPU not supported", but generally it has my approval. We'll probably just have to ifdef out the Linux/macOS codepath for it and disable it for the time being but otherwise again it generally has my approval.

RytoEX:
Rebased to latest master. May push trivial fixes (text nits + minor cleanups or just text nits) before merging.

koolscooby:
> Thanks for rebasing. There are a few minor things, such as Linux/macOS just not being implemented at all and failing with "GPU not supported", but generally it has my approval. We'll probably just have to ifdef out the Linux/macOS codepath for it and disable it for the time being but otherwise again it generally has my approval.

If possible, we'd like to keep the implementation as-is so we can release server-side updates to start vending ladders for Linux, MacOS, AMD AMF, and Intel QSV without requiring additional changes to OBS.

In the short term, we can improve the error message that is shown so that it makes it more apparent to users that the reason it is unsupported is not the build of OBS Studio, but instead the service that vends the configuration.

RytoEX (commented):
Leaving a brief review comment for posterity.

RytoEX (approved):
The CI failures are unrelated. The GitHub Actions Ubuntu runners are restarting due to some issue related to apt packages that indicate they require service restarts.

PR #10635, opened 2024-05-03
UI: Add support for additional canvases to multitrack video output

palana:
### Description
Adds frontend API calls to allow plugins to register additional canvases/video outputs that will participate in multitrack video configuration generation and multitrack video streams

### Motivation and Context
Currently additional views/canvases have to use separate rtmp connections or similar to be streamed, where the multitrack video output allows both multiple renditions of the same view and also different views to be streamed

### How Has This Been Tested?
Twitch Enhanced Broadcasting beta

### Types of changes

### Checklist:

### DIFF
--- a/UI/api-interface.cpp
+++ b/UI/api-interface.cpp
@@ -628,9 +628,9 @@ struct OBSStudioAPI : obs_frontend_callbacks {
 	}
 
 	void obs_frontend_multitrack_video_register(const char *name, multitrack_video_start_cb start_video,
-						    multitrack_video_stop_cb stop_video, void *param) override
+						    multitrack_video_stop_cb stop_video, void *private_data) override
 	{
-		main->MultitrackVideoRegister(name, start_video, stop_video, param);
+		main->MultitrackVideoRegister(name, start_video, stop_video, private_data);
 	}
 
 	void obs_frontend_multitrack_video_unregister(const char *name) override
--- a/UI/api-interface.cpp
+++ b/UI/api-interface.cpp
@@ -626,6 +626,17 @@ struct OBSStudioAPI : obs_frontend_callbacks {
 			cb.callback(event, cb.private_data);
 		}
 	}
+
+	void obs_frontend_multitrack_video_register(const char *name, multitrack_video_start_cb start_video,
+						    multitrack_video_stop_cb stop_video, void *param) override
+	{
+		main->MultitrackVideoRegister(name, start_video, stop_video, param);
+	}
+
+	void obs_frontend_multitrack_video_unregister(const char *name) override
+	{
+		main->MultitrackVideoUnregister(name);
+	}
 };
 
 obs_frontend_callbacks *InitializeAPIInterface(OBSBasic *main)
--- a/UI/api-interface.cpp
+++ b/UI/api-interface.cpp
@@ -627,8 +627,10 @@ struct OBSStudioAPI : obs_frontend_callbacks {
 		}
 	}
 
-	void obs_frontend_multitrack_video_register(const char *name, multitrack_video_start_cb start_video,
-						    multitrack_video_stop_cb stop_video, void *private_data) override
+	void obs_frontend_multitrack_video_register(const char *name,
+						    obs_frontend_multitrack_video_start_cb start_video,
+						    obs_frontend_multitrack_video_stop_cb stop_video,
+						    void *private_data) override
 	{
 		main->MultitrackVideoRegister(name, start_video, stop_video, private_data);
 	}
--- a/UI/goliveapi-postdata.cpp
+++ b/UI/goliveapi-postdata.cpp
@@ -7,7 +7,8 @@
 #include "models/multitrack-video.hpp"
 
 GoLiveApi::PostData constructGoLivePost(QString streamKey, const std::optional<uint64_t> &maximum_aggregate_bitrate,
-					const std::optional<uint32_t> &maximum_video_tracks, bool vod_track_enabled)
+					const std::optional<uint32_t> &maximum_video_tracks, bool vod_track_enabled,
+					const std::map<std::string, video_t *> &extra_views)
 {
 	GoLiveApi::PostData post_data{};
 	post_data.service = "IVS";
@@ -43,6 +44,22 @@ GoLiveApi::PostData constructGoLivePost(QString streamKey, const std::optional<u
 #endif
 		} else if (qstricmp(codec, "av1")) {
 			add_codec("av1");
+		}
+	}
+
+	if (!extra_views.empty()) {
+		auto &extra_views_capability = *post_data.capabilities.extra_views;
+		extra_views_capability.reserve(extra_views.size());
+		for (auto &view : extra_views) {
+			video_t *video = view.second;
+			if (!video)
+				continue;
+			const struct video_output_info *voi = video_output_get_info(video);
+			if (!voi)
+				continue;
+			extra_views_capability.push_back(
+				GoLiveApi::ExtraView{view.first, voi->width, voi->height,
+						     media_frames_per_second{voi->fps_num, voi->fps_den}});
 		}
 	}
 
--- a/UI/goliveapi-postdata.cpp
+++ b/UI/goliveapi-postdata.cpp
@@ -48,6 +48,7 @@ GoLiveApi::PostData constructGoLivePost(QString streamKey, const std::optional<u
 	}
 
 	if (!extra_views.empty()) {
+		post_data.capabilities.extra_views.emplace();
 		auto &extra_views_capability = *post_data.capabilities.extra_views;
 		extra_views_capability.reserve(extra_views.size());
 		for (auto &view : extra_views) {
--- a/UI/multitrack-video-output.cpp
+++ b/UI/multitrack-video-output.cpp
@@ -37,6 +37,7 @@
 #include "goliveapi-network.hpp"
 #include "multitrack-video-error.hpp"
 #include "models/multitrack-video.hpp"
+#include "window-basic-main.hpp"
 
 Qt::ConnectionType BlockingConnectionTypeFor(QObject *object)
 {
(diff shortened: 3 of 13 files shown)


RytoEX:
#10633 was merged some time ago. This will need to be rebased and get CI to pass.

palana:
> #10633 was merged some time ago. This will need to be rebased and get CI to pass.

@RytoEX: rebased and updated the description

RytoEX:
@palana It looks like the build failures are still present.

palana:
> @palana It looks like the build failures are still present.

@RytoEX: oh, yea, missed the bit about making CI pass; latest push should make it pass 😅

RytoEX:
@tt2468 @derrod Any feedback on this?

docs/sphinx/reference-frontend-api.rst
```diff
@@ -949,3 +957,22 @@ Functions
                       This uses the undo action from the first and the redo action from the last action.
 
    .. versionadded:: 29.1
+
+---------------------------------------
+
+.. function:: void obs_frontend_multitrack_video_register(const char *name, multitrack_video_start_cb start_video, multitrack_video_stop_cb stop_video, void *param)
```
tt2468, line 963: As much as we flip flop in the code base, the frontend API is using the `private_data` name everywhere else, so it should probably be that.

It may be a good idea to prefix the callback types with `obs_frontend_`, but that would also make them quite long. Probably good for consistency, however I am open to discussion from others.

UI/window-basic-main.cpp
```diff
@@ -11053,6 +11053,33 @@ QColor OBSBasic::GetSelectionColor() const
 	}
 }
 
+const std::vector<MultitrackVideoViewInfo> &
+OBSBasic::GetAdditionalMultitrackVideoViews()
+{
+	return multitrackVideoViews;
+}
+
+void OBSBasic::MultitrackVideoRegister(const char *name,
```
tt2468, line 11062: It seems like the register and unregister code could live in api-interface.cpp. There's no real benefit having functions here if only api-interface calls them.

UI/window-basic-main.cpp
```diff
@@ -11053,6 +11053,33 @@ QColor OBSBasic::GetSelectionColor() const
 	}
 }
 
+const std::vector<MultitrackVideoViewInfo> &
+OBSBasic::GetAdditionalMultitrackVideoViews()
```
tt2468, line 11057: This does nothing but return a copy of the value. Probably fine to fetch the value directly?

UI/multitrack-video-output.cpp
```diff
@@ -290,22 +291,36 @@ create_video_encoder(DStr &name_buffer, size_t encoder_index,
...
 		blog(LOG_WARNING,
 		     "Failed to get obs_video_info while creating encoder %zu",
 		     encoder_index);
 		throw MultitrackVideoError::warning(
 			QTStr("FailedToStartStream.FailedToGetOBSVideoInfo")
 				.arg(name_buffer->array, encoder_type));
 	}
+	obs_video_info *ovi = &ovi_storage;
+
+	const char *view_name = nullptr;
```
tt2468, line 306: This is probably safer as std::string.

docs/sphinx/reference-frontend-api.rst
```diff
@@ -239,6 +239,14 @@ Structures/Enumerations
 
    Undo redo callback
 
+.. type:: video_t *(*multitrack_video_start_cb)(void *param)
```
tt2468, line 242: It probably makes sense to include the view name in both of these callbacks.

UI/window-basic-main.cpp
palana: not sure I follow, this returns a const reference, rather than a copy (i.e. you're allowed to read entries, but not modify them)

UI/window-basic-main.cpp
palana: right, and api-interface already has access to window basic main internals

UI/multitrack-video-output.cpp
palana: safer in what way? `encoder_config` can't be modified in here (unless someone adds a const cast?), so the pointer (once assigned) stays valid for this scope at least (and it's not used outside of this scope)

docs/sphinx/reference-frontend-api.rst
palana: changed in https://github.com/obsproject/obs-studio/pull/10635#commits-pushed-f65594f

docs/sphinx/reference-frontend-api.rst
palana: changed both in https://github.com/obsproject/obs-studio/pull/10635#commits-pushed-f65594f

length of the callback type seems fine, though it is on the long side; I don't have a good alternative suggestion at this point though

RytoEX: @tt2468 Looks like this feedback was addressed.

docs/sphinx/reference-frontend-api.rst
RytoEX: It looks like the code was changed, but not the documentation.

lexano-ivs: @RytoEX I just pushed a fixup commit with documentation changes.

UI/window-basic-main.cpp
RytoEX: Looks like this feedback was addressed.

UI/window-basic-main.cpp
RytoEX: @tt2468 Could you clarify the concern here?

UI/multitrack-video-output.cpp
RytoEX: @tt2468 Could you clarify the concern here?

PatTheMav: From a cursory glance there are multiple reasons:

1. This is C++ code. By default one should use C++ types and avoid low-level C types at all costs.
2. The `view_name` is derived from `encoder_config.view`, which is a `std::string` already. So the code taps into the internal data of that string instead of using a reference to the original `std::string`, particularly as it is only used for `std::map::find` anyway.

It's precisely the kind of code style that the C++ committee and the language designers tell developers to avoid. Use C++ types and avoid pointers and C types at all costs.

palana: @PatTheMav: can you help me translate what you're suggesting? https://isocpp.github.io/CppCoreGuidelines/CppCoreGuidelines#f16-for-in-parameters-pass-cheaply-copied-types-by-value-and-others-by-reference-to-const even lists pointers as first option when passing an optional argument into a function (c++ doesn't really support rust's `Option<&str>` or similar unfortunately)

so another option would be to do special handling for empty strings, instead of requiring special handling on a different layer (c pointers, and generally everything c++ is kind of terrible for this unfortunately, since most of these won't noisily fail when you use them wrong, and the compiler also doesn't help you due to lack of borrow checking); if that's the preference here I can make that change

or are you asking for this to be translated into `std::string *view_name = nullptr;` instead?

PatTheMav: The root cause of this is that the `view` name is optional in `VideoEncoderConfiguration`, but I wonder if this has to be the case:

If OBS is supposed to support multiple views _and_ views carry a name, then the main view should also hold a name and not be identified by "the lack of any name", which is brittle.

So instead the requirement could be: OBS can have multiple views, _every_ view needs to have a unique name, _including_ the main view. This would also allow everyone working with views to clearly identify the views and gives OBS itself the ability to "reserve" view names (and there are already ternary expressions trying to use the fallback name "obs_base", so there clearly _is_ a requirement for the main/base view to have a name).

That also has the potential to solve the following issues:

* `adjust_video_encoder_scaling` doesn't have to accept a mish-mash of actual values _or_ `nullptr`s and use ternary expressions to get some video dimensions or view name from something else.
* In the same vein `adjust_video_encoder_scaling`'s argument list wouldn't need to mix mandatory and logically optional arguments instead of following the good practice of putting mandatory arguments first and optional ones last

So there should be a single list of views to which the "main view" is added and additional views can be added by the frontend API. Then the code could work with the expectation that:

* The list contains at least a single view (not having any views would be an invalid application state)
* Every view (including the main view) carries a name and the same metadata and video info

---

This is all complex stuff and I appreciate the work put into this, but that's even more reason to make it as straightforward as possible and avoid a pile of little workarounds here, there, and everywhere to squeeze in this functionality.

UI/api-interface.cpp
```diff
@@ -626,6 +626,19 @@ struct OBSStudioAPI : obs_frontend_callbacks {
 			cb.callback(event, cb.private_data);
 		}
 	}
+
+	void obs_frontend_multitrack_video_register(const char *name,
```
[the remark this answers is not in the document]
palana: its only purpose is to enable views to participate in multitrack streaming, at least I'm not aware of any other functionality that would make use of this right now

upcoming webrtc multitrack (and other/future multitrack outputs) might be interested in this at some point, but so far webrtc multitrack is stuck in review

doing this in libobs-land is somewhat very uncharted territory, since views/video outputs/general obs objects don't carry meta information about their intended purpose (and given the private data capabilities, they're currently very much in "don't touch these objects unless you know where they come from/why they're there" territory)

UI/goliveapi-postdata.cpp
```diff
@@ -46,6 +47,22 @@ GoLiveApi::PostData constructGoLivePost(QString streamKey, const std::optional<u
 		}
 	}
 
+	if (!extra_views.empty()) {
+		auto &extra_views_capability = *post_data.capabilities.extra_views;
```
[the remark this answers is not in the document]
palana: nope, seems like an accident that it currently works as is (as in, an empty `vector` looks the same as whatever contents the `optional` gets default-initialized with?)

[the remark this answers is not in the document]
palana: fixed in https://github.com/obsproject/obs-studio/pull/10635/commits/6bc5f21e7eb9a07f81cfe379095a6fd8b63f230e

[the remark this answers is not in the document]
PatTheMav: Right, I wondered the same (particularly if the declaration leads to a default initialisation of the wrapped `std::vector`) but at least this small reproducer crashes at `operator*`:

```C++
#include <iostream>
#include <optional>
#include <vector>
#include <string>

struct Test {
    std::optional<std::vector<std::string>> optionals;
};

int main(int argc, const char * argv[]) {
    Test test;
    
    auto &some = *test.optionals;
    some.reserve(10);
    
    some.push_back("Hello");
    
    for (auto &opt : some) {
        std::cout << opt.data() << "\n";
    }
    
    
    return 0;
}

RytoEX:
Closing in favor of:
* https://github.com/obsproject/obs-studio/pull/11823
* https://github.com/obsproject/obs-studio/pull/11832
* https://github.com/obsproject/obs-studio/pull/11833

PR #10640, opened 2024-05-04
build-aux: Disable DeckLink in the Flatpak

tytan652:
### Description

Disable a feature that might never be available in Flatpak since BlackMagic does not provide a stable link to download it as extra-data.

### Motivation and Context

Since the Flatpak became part of the repo, nothing has changed on BlackMagic side.

DeckLink libraries are still not redistributable and without stable download link so even extra-data is not usable.

This makes enabling DeckLink feature in the Flatpak impossible without involving customization on the end-user side which is not how Flatpak is designed for.

### How Has This Been Tested?

It builds

### Types of changes

- Flatpak cleanup

### Checklist:


RytoEX (approved):
Looks good to me.

PR #10647, opened 2024-05-06
UI: Use weak_source_expired instead of getting the source

gxalpha:
### Description

Changes an action that's getting a source from the weak source to check if the source still exists to simply use `obs_weak_source_expired`.

### Motivation and Context

Noticed this, and appears to be unnecessary.

### How Has This Been Tested?

macOS 14.4.1
Copied filters from a source.
Right clicked on the audio mixer to see that "Paste Filters" is available.
Deleted the source. Right clicked on the audio mixer again and "Paste Filters" is unavailable.

### Types of changes

### Checklist:

### DIFF
--- a/UI/window-basic-main.cpp
+++ b/UI/window-basic-main.cpp
@@ -3833,14 +3833,8 @@ void OBSBasic::VolControlContextMenu()
 
 	copyFiltersAction.setEnabled(obs_source_filter_count(vol->GetSource()) >
 				     0);
-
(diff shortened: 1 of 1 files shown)


RytoEX (approved):
Tested this locally and it seems fine. A second look would be appreciated, but I may not require it.

This code pattern is also used elsewhere.
https://github.com/obsproject/obs-studio/commit/52497d23bba73708ac952a8b12d22b73487d02cc

PR #10648, opened 2024-05-06
libobs: Add missing encoder API functions

Error-String-Expected-Got-Nil:
### Description
Adds the API functions `obs_encoder_get_signal_handler` and `obs_encoder_get_proc_handler`, both of which are mentioned in the API docs but are not actually implemented in the API. Encoders don't use signals for anything, but it was easy enough to include that I didn't see a reason not to.

### Motivation and Context
Was attempting to create a plugin, found `obs_encoder_get_proc_handler` did not actually exist, decided to make a PR to fix that since it's an easy fix.

### How Has This Been Tested?
Created an encoder in a plugin's load function, used function to get its encoder, and found it did correctly return the encoder's proc handler.

### Types of changes

### Checklist:

### DIFF
--- a/libobs/obs-encoder.c
+++ b/libobs/obs-encoder.c
@@ -591,6 +591,20 @@ obs_data_t *obs_encoder_get_settings(const obs_encoder_t *encoder)
 	return encoder->context.settings;
 }
 
+signal_handler_t *obs_encoder_get_signal_handler(const obs_encoder_t *encoder)
+{
+	return obs_encoder_valid(encoder, "obs_encoder_get_signal_handler")
+			? encoder->context.signals
+			: NULL;
+}
+
+proc_handler_t *obs_encoder_get_proc_handler(const obs_encoder_t *encoder)
+{
+	return obs_encoder_valid(encoder, "obs_encoder_get_proc_handler")
+			? encoder->context.procs
+			: NULL;
+}
+
 static inline void reset_audio_buffers(struct obs_encoder *encoder)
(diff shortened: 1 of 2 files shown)


Fenrirthviti:
If this function doesn't really serve any practical purpose, why would you add the function instead of correcting the documentation? What is the actual use case for this?

tt2468:
Encoders might have custom functionality via procs/signals that one may want to implement. I can think of some ideas in my head which would necessitate usage of these functions. The proc/signal handlers for encoders are already there, just not accessible via the frontend API like they are with most other OBS objects.

Error-String-Expected-Got-Nil:
> If this function doesn't really serve any practical purpose, why would you add the function instead of correcting the documentation? What is the actual use case for this?

I'm making this PR specifically because I wanted to use one of these functions in a plugin I'm making and it wasn't available, so yes, I think there is a use case.

Fenrirthviti:
Can you provide an example?

To be clear, I don't want to imply I'm against this, I'm mostly just curious the use case, and having a practical example will help in review for others.

Error-String-Expected-Got-Nil:
> Can you provide an example?
> 
> To be clear, I don't want to imply I'm against this, I'm mostly just curious the use case, and having a practical example will help in review for others.

Mine is a little weird, but the plugin I'm making has an encoder that has a counter for a number of frames I want it to encode before it stops. The easiest way to adjust that counter would be to use a procedure.

PR #10665, opened 2024-05-12
rnnoise: Fix compilation not working on latest gcc

cg2121:
### Description
The latest gcc spits out an error (-Werror=calloc-transposed-args) about calloc parameters that are in the wrong order.

### Motivation and Context
I'm on the latest version of Fedora which has the newest gcc and OBS doesn't compile.

### How Has This Been Tested?
Compiled OBS

### Types of changes

### Checklist:

### DIFF
--- a/plugins/obs-filters/rnnoise/src/denoise.c
+++ b/plugins/obs-filters/rnnoise/src/denoise.c
@@ -265,9 +265,9 @@ int rnnoise_init(DenoiseState *st, RNNModel *model) {
     st->rnn.model = model;
   else
     st->rnn.model = &rnnoise_model_orig;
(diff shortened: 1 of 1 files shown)


derrod:
Perhaps its time to get rid fo the vendored version as well, afaik we use system packages and obs-deps by default, this is just a fallback that existed because Ubuntu in the ancient times didn't have it.

RytoEX (approved):
This change seems to be correct. That said, this is vendored code that is sorely outdated. Once I get RNNoise updated in obs-deps, I plan to also update/replace this vendored code to the same version. However, to fix a compilation error, I'm willing to merge this for now.

PR #10674, opened 2024-05-14
libobs: Fix crash when mix is NULL

exeldro:
### Description
When `obs_reset_video` is called, it calls `obs_free_video` that sets the mix to NULL here: https://github.com/obsproject/obs-studio/blob/master/libobs/obs.c#L879

### Motivation and Context
Don't like crashes when switching resolution

### How Has This Been Tested?
On windows by changing the resolution multiple times

### Types of changes

### Checklist:

### DIFF
--- a/libobs/obs.c
+++ b/libobs/obs.c
@@ -878,13 +878,13 @@ static void obs_free_video(void)
 		obs_free_video_mix(video);
 		obs->video.mixes.array[i] = NULL;
 	}
+	da_free(obs->video.mixes);
 	if (num_views > 0)
 		blog(LOG_WARNING, "Number of remaining views: %ld", num_views);
 	pthread_mutex_unlock(&obs->video.mixes_mutex);
 
 	pthread_mutex_destroy(&obs->video.mixes_mutex);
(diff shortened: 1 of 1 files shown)


tt2468:
Just curious, in what situation did you encounter this?

exeldro:
I encountered this investigating a bug report with Aitum Vertical causing a crash after switching resolution of the main canvas

notr1ch:
In `obs_free_video `, wouldn't including the `da_free(obs->video.mixes);` in the mutex be a better solution?

exeldro:
That should work too, I have no preference

notr1ch:
I think I'd prefer that as it protects against other possible race conditions.

tt2468 (approved):
I encountered the bug referenced here yesterday. Unfortunately, I was unable to reproduce the bug today (due to it being a race condition bug), but after analyzing the changes here, I think this is a good change. I should note that this may expose further bugs related to locking an already-destroyed mutex (as `obs->video.mixes_mutex` is destroyed immediately while other threads may still be accessing it), but the da_free() here should be guarded by that mutex either way, even if the mutex does get immediately destroyed afterwards.

PR #10680, opened 2024-05-15
UI: Keep weak stream output ref in status bar

RytoEX:
### Description

The status bar doesn't currently own its references to outputs, which can be a problem if streaming outputs are released soon after stream stop. The Multitrack Video output does exactly that, so the status bar will sometimes try to access an invalid pointer for updating its stats. Keeping a weak reference around and upgrading it to collect stats, similar to how the stats window behaves.

### Motivation and Context

Picked from #10633. Want to pull some commits out of that PR to make that PR smaller.

### How Has This Been Tested?

This has been tested in the Twitch Enhanced Broadcasting beta.

### Types of changes

### Checklist:

### DIFF
--- a/UI/window-basic-status-bar.cpp
+++ b/UI/window-basic-status-bar.cpp
@@ -195,7 +195,11 @@ void OBSBasicStatusBar::UpdateBandwidth()
 	if (++seconds < bitrateUpdateSeconds)
 		return;
 
-	uint64_t bytesSent = obs_output_get_total_bytes(streamOutput);
+	OBSOutput output = OBSGetStrongRef(streamOutput);
+	if (!output)
+		return;
+
+	uint64_t bytesSent = obs_output_get_total_bytes(output);
 	uint64_t bytesSentTime = os_gettime_ns();
 
 	if (bytesSent < lastBytesSent)
@@ -338,8 +342,12 @@ void OBSBasicStatusBar::UpdateDroppedFrames()
 	if (!streamOutput)
 		return;
 
-	int totalDropped = obs_output_get_frames_dropped(streamOutput);
-	int totalFrames = obs_output_get_total_frames(streamOutput);
+	OBSOutput output = OBSGetStrongRef(streamOutput);
+	if (!output)
+		return;
+
+	int totalDropped = obs_output_get_frames_dropped(output);
+	int totalFrames = obs_output_get_total_frames(output);
 	double percent = (double)totalDropped / (double)totalFrames * 100.0;
 
 	if (!totalFrames)
@@ -356,7 +364,7 @@ void OBSBasicStatusBar::UpdateDroppedFrames()
 	/* ----------------------------------- *
 	 * calculate congestion color          */
 
-	float congestion = obs_output_get_congestion(streamOutput);
+	float congestion = obs_output_get_congestion(output);
 	float avgCongestion = (congestion + lastCongestion) * 0.5f;
 	if (avgCongestion < congestion)
 		avgCongestion = congestion;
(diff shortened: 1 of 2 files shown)


RytoEX:
If pulling commits out of #10633 isn't helpful, we can close this. I find it easier to review small PRs and had offered to start pulling commits out of that PR, so I wanted to start working towards that.

UI/window-basic-status-bar.cpp
```diff
@@ -195,7 +195,11 @@ void OBSBasicStatusBar::UpdateBandwidth()
 	if (++seconds < bitrateUpdateSeconds)
 		return;
 
-	uint64_t bytesSent = obs_output_get_total_bytes(streamOutput);
+	OBSOutputAutoRelease output = obs_weak_output_get_output(streamOutput);
+	if (!output)
+		return;
```
gxalpha, line 200: Just to confirm (can’t check atm), there are no values displayed that should potentially get reset / set to zero when the output is gone?

gxalpha (approved):
Seems sane. If we’re not holding a strong reference we should not access the pointer (and if we were holding one we should listen to the appropriate signals to release it).

That said, this same thing should probably be applied to recordOutput too (even if it’s not causing crashes right now). I’d even consider this is-scope for this PR seeing how similar it is.

UI/window-basic-status-bar.cpp
```diff
@@ -195,7 +195,11 @@ void OBSBasicStatusBar::UpdateBandwidth()
 	if (++seconds < bitrateUpdateSeconds)
 		return;
 
-	uint64_t bytesSent = obs_output_get_total_bytes(streamOutput);
+	OBSOutputAutoRelease output = obs_weak_output_get_output(streamOutput);
```
cg2121, line 198: I would suggest to use OBSGetStrongRef to be consistent with how other places in the UI get the strong reference of a weak pointer.

cg2121: And do this for all of the PR.

UI/window-basic-status-bar.hpp
```diff
@@ -27,7 +27,7 @@ class OBSBasicStatusBar : public QStatusBar {
 private:
 	StatusBarWidget *statusWidget = nullptr;
 
-	obs_output_t *streamOutput = nullptr;
+	OBSWeakOutputAutoRelease streamOutput = nullptr;
```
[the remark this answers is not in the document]
RytoEX: That does seem to be the case. Will change later.

UI/window-basic-status-bar.cpp
RytoEX: For everywhere else in the codebase that uses `OBSGetStrongRef`, the type of the receiving variable is one of:
* auto
* OBSSource
* OBSObject

If we were to use `OBSGetStrongRef` here, then it seems the type for `output` should either be `auto` or `OBSOutput`. That said, I'm not sure of the semantic difference off-hand and would have to look into it. I'm not sure if one approach is preferred over the other.

RytoEX:
Regarding the recording output, the original commit in #10633 only addresses streaming output. I don't want to expand scope too much. I just want to trim down that PR. I can make a follow-up PR for recording output later.

UI/window-basic-status-bar.cpp
RytoEX: When the stream is stopped `StreamStopped` is called which:
1. calls `ReconnectClear()`
   1. resets the various time/bytes members
2. sets `streamOutput` to `nullptr`
3. calls `Deactivate()`
   1. resets the UI elements in the status bar

Aside from that, I'm not sure what happens if the output disappears otherwise.

UI/window-basic-status-bar.cpp
RytoEX: Changed to use `OBSGetStrongRef` and `OBSOutput` accordingly.

UI/window-basic-status-bar.hpp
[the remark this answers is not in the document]
RytoEX: Changed.

UI/window-basic-status-bar.cpp
RytoEX: As far as I can tell from off-thread discussion, this concern was resolved.

PR #10693, opened 2024-05-19
libobs: Add functions to serialize JSON with default values

derrod:
### Description

Adds two new functions to serialize `obs_data` objects to JSON while also including default values.

### Motivation and Context

Alternative to #10641 that requires opt-in from the API user. Same motiviation and context.

### How Has This Been Tested?

Serialized some data.

### Types of changes

### Checklist:


Lain-B (approved):
Would recommend documentation (if documentation for this area of functions already exist) but not a big deal one way or another imho

PR #10695, opened 2024-05-19
 obs-filters: Add white balance to color filter

mar10iana:
### Description

We added three white balance sliders to the color correction filter, as shown on the screenshots below:

IMG_8447
IMG_1899

The following changes have been made to the code:
- Added new settings for red, green, and blue white balance adjustments.
- Updated the color_correction_filter_data structure to include white balance variables.
- Modified the create functions to initialize the white balance values.
- Implemented white balance matrix computation in the update functions.
- Integrated the white balance matrix into the final matrix used for color correction.
- Updated the properties functions to add sliders for white balance adjustments.
- Set default values for the new white balance settings.

### Motivation and Context

Currently, OBS offers a variety of color correction filters, but lacks a dedicated white balance feature, which is essential for many users. This feature aims to address this problem by introducing three white balance sliders (red, green and blue) that adjust the color temperature. 

Ideas website link: https://ideas.obsproject.com/posts/2582/white-balance

### How Has This Been Tested?

We tested our feature in a macOS environment. 
To test this feature you should go to the color correction filter and check the newly added sliders. If you change the value of the slider, the colours of your source will be updated.

### Types of changes

### Checklist:

### DIFF
--- a/plugins/linux-pipewire/pipewire.c
+++ b/plugins/linux-pipewire/pipewire.c
@@ -137,6 +137,14 @@ struct _obs_pipewire_stream {
 		struct spa_fraction fraction;
 		bool set;
 	} framerate;
+
+	struct {
+		int acquire_syncobj_fd;
+		int release_syncobj_fd;
+		uint64_t acquire_point;
+		uint64_t release_point;
+		bool set;
+	} sync;
 };
 
 /* auxiliary methods */
@@ -545,6 +553,20 @@ static void renegotiate_format(void *data, uint64_t expirations)
 
 /* ------------------------------------------------- */
 
+static void return_unused_pw_buffer(struct pw_stream *stream, struct pw_buffer *b)
+{
+#if PW_CHECK_VERSION(1, 2, 0)
+	struct spa_buffer *buffer = b->buffer;
+	struct spa_data *last_data = &buffer->datas[buffer->n_datas - 1];
+	struct spa_meta_sync_timeline *synctimeline =
+		spa_buffer_find_meta_data(buffer, SPA_META_SyncTimeline, sizeof(struct spa_meta_sync_timeline));
+
+	if (synctimeline && (last_data->type == SPA_DATA_SyncObj))
+		gs_sync_signal_syncobj_timeline_point(last_data->fd, synctimeline->release_point);
+#endif
+	pw_stream_queue_buffer(stream, b);
+}
+
 static inline struct pw_buffer *find_latest_buffer(struct pw_stream *stream)
 {
 	struct pw_buffer *b;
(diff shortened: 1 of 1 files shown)


WizardCM:
Is it common for white balance to be split into three separate sliders? Keeping in mind OBS is used by a variety of users, some with very intricate knowledge of video/colour and others who just want to make it look good with as little configuration as possible.

RytoEX (changes requested):
Do not use merge commits to catch up on branch/repo history. Use `git rebase`.

mar10iana:
> Is it common for white balance to be split into three separate sliders? Keeping in mind OBS is used by a variety of users, some with very intricate knowledge of video/colour and others who just want to make it look good with as little configuration as possible.

This method offers the highest level of control, allowing precise fine-tuning of color balance across the spectrum. It's ideal for professional users who need to correct complex color casts or achieve specific color grading effects.
For regular users, like myself, despite not being the simplest option, I still find this method pretty straightforward to use.
But ultimately it is up to you guys! We can change it if needed :)

Warchamp7:
I'm wondering if this makes sense as part of the color correction filter or if it should be a separate filter of it's own.

mar10iana:
> I'm wondering if this makes sense as part of the color correction filter or if it should be a separate filter of it's own.

White balance is usually under the color correction tab on photo and video editing programs such as Adobe Photoshop, After Effects, and others. Our decision was based on this and the fact that changing the white balance is color correcting the image.

Fenrirthviti:
Just putting my vote that I agree on the above. I've actually several times added an OBS color correction filter mistakenly thinking it had white balance in the past, heh. I think it's a common enough paradigm that users would expect to find it there.

Warchamp7:
Gonna stick to my original feedback here. I think it is fine to keep the 3 channels for the options but I think this should be moved into a new separate `White Balance Filter` option instead of crowding the Color Correction one even further.

This will also allow better control for whether to white balance before or after a Color Correction filter, for example.

mar10iana:
> Gonna stick to my original feedback here. I think it is fine to keep the 3 channels for the options but I think this should be moved into a new separate `White Balance Filter` option instead of crowding the Color Correction one even further.
> 
> This will also allow better control for whether to white balance before or after a Color Correction filter, for example.

Should implement this change?

Warchamp7 (changes requested):
Create a separate White Balance filter for this functionality instead

mar10iana:
> Create a separate White Balance filter for this functionality instead

I will do it this weekend, thank you.

mar10iana:
> Create a separate White Balance filter for this functionality instead

Sorry for making another PR instead of updating the last one, but this way was just easier for me.
https://github.com/obsproject/obs-studio/pull/11850

PR #10709, opened 2024-05-22
UI: Reset YT Live Chat dock on StreamingStop.

msuman-google:
### Description
This will use the same placeholder URL that is used before a stream is started.

### Motivation and Context
When a stream ends, the input bar for sending chat messages should no longer
be active. It wouldn't work even if it was active.

### How Has This Been Tested?
Built OBS locally on Linux workstation, started and stopped a live stream.

### Types of changes

### Checklist:

### DIFF
--- a/UI/auth-youtube.cpp
+++ b/UI/auth-youtube.cpp
@@ -211,6 +211,7 @@ void YoutubeAuth::ResetChat()
 {
 #ifdef BROWSER_AVAILABLE
 	if (chat && chat->cefWidget) {
+		chat->SetApiChatId("");
 		chat->cefWidget->setURL(YOUTUBE_CHAT_PLACEHOLDER_URL);
 	}
 #endif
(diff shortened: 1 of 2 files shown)


RytoEX (changes requested):
Do not use merge commits to catch up on branch/repo history. Use `git rebase`.

Do not use final punctuation in PR titles or commit message subjects (the first line of a commit message).

tytan652:
This reset the chat dock even if the broadcast has not been ended, which is wrong.
A user can still stream to the same broadcast if it has not been ended (autostop on ended manually).

msuman-google:
> This reset the chat dock even if the broadcast has not been ended, which is wrong. A user can still stream to the same broadcast if it has not been ended (autostop or ended manually).

Interesting. I was going by the prompt that is shown right after clicking `Stop streaming` (see https://ibb.co/XC477pD). I will close this pull request based on this comment.

tytan652:
This prompt is when the broadcast has autostop enabled indeed, but when it's not present there is a "end broadcast" button and the user can re-stream to this broadcast.

PR #10711, opened 2024-05-23
CI: Add PVS-Studio Analysis for Windows

derrod:
### Description

Runs PVS-Studio on Windows builds during scheduled runs.

### Motivation and Context

Static analysis is nice to have, SARIF integrates with GitHub so maintainers can easily manage and action on automatic analysis reports.

Settings this up for Linux and macOS is a bit more involved as we can't just run it on a project file as on Windows, and the command line/configuration files are different. There will be some duplication between the three platforms which may be annoying, so ideally we'd also try to cut non-Windows platforms down to platform-specific files. We'll deal with that at a later date.

### How Has This Been Tested?

Ran on my fork.

### Types of changes

### Checklist:

### DIFF
--- a/.github/actions/windows-analysis/action.yaml
+++ b/.github/actions/windows-analysis/action.yaml
@@ -1,0 +1,90 @@
+name: Run PVS-Studio Analysis
+inputs:
+  pvsUsername:
+    description: PVS-Studio License Username
+    required: true
+  pvsKey:
+    description: PVS-Studio License Key
+    required: true
+  target:
+    description: Build Target
+    required: true
+  config:
+    description: Build Configuration
+    required: true
+runs:
+  using: composite
+  steps:
+    - name: Setup PVS-Studio
+      shell: pwsh
+      run: |
+        choco install pvs-studio --version=7.30.81185.980 -y --no-progress
+
+    - name: Activate PVS-Studio
+      shell: pwsh
+      run: |
+        . ${env:GITHUB_ACTION_PATH}\Invoke-External.ps1
+        Invoke-External "C:\Program Files (x86)\PVS-Studio\PVS-Studio_Cmd.exe" credentials -u ${{ inputs.pvsUsername }} -n ${{ inputs.pvsKey }}
+
+    - name: Run PVS-Studio Analysis
+      shell: pwsh
+      run: |
+        [flags()] Enum PVSErrorCodes {
+            Success = 0
+            AnalyzerCrash = 1
+            GenericError = 2
+            InvalidCommandLine = 4
+            FileNotFound = 8
+            ConfigurationNotFound = 16
+            InvalidProject = 32
+            InvalidExtension = 64
+            LicenseInvalid = 128
+            CodeErrorsFound = 256
+            SuppressionFailed = 512
+            LicenseExpiringSoon = 1024
+        }
+        
+        $pvsParams = @(
+            "--progress"
+            "--disableLicenseExpirationCheck"
+            "-p", "${{ inputs.target }}"
+            "-c", "${{ inputs.config }}"
+            "-t", "${{ github.workspace }}\build_x64\obs-studio.sln"
+            "-o", "${{ github.workspace }}\analysis.plog"
+            "-C", "${env:GITHUB_ACTION_PATH}\obs.pvsconfig"
+        )
+        & "C:\Program Files (x86)\PVS-Studio\PVS-Studio_Cmd.exe" @pvsParams
+        
+        # Success and CodeErrorsFound are fine as error codes, we only care if it is anything but those
+        $pvs_result = $LASTEXITCODE -band (-bnot [PVSErrorCodes]::CodeErrorsFound)
+        if ($pvs_result -ne 0) {
+            Write-Output "PVS-Studio Errors: $([PVSErrorCodes]$pvs_result)"
+        }
+        exit $pvs_result
+
+    - name: Convert Analysis to SARIF
+      shell: pwsh
+      run: |
+        . ${env:GITHUB_ACTION_PATH}\Invoke-External.ps1
+        
+        $conversionParams = @(
+            "-a", "GA:1,2",
+            "-d", "V1042,Renew"
+            "-t", "Sarif"
+            "${{ github.workspace }}\analysis.plog"
+        )
+        Invoke-External "C:\Program Files (x86)\PVS-Studio\PlogConverter.exe" @conversionParams
+
+    - name: Upload PVS-Studio Logs
+      uses: actions/upload-artifact@v4
+      with:
+        name: 'pvs-analysis-log'
+        path: |
+          ${{ github.workspace }}/analysis.plog
+          ${{ github.workspace }}/analysis.plog.sarif
+
+    - name: Upload PVS-Studio Report
+      uses: github/codeql-action/upload-sarif@v3
+      with:
+        sarif_file: "${{ github.workspace }}/analysis.plog.sarif"
+        category: 'PVS-Studio (Windows)'
--- a/.github/workflows/analyze-project.yaml
+++ b/.github/workflows/analyze-project.yaml
@@ -1,0 +1,39 @@
+name: Analyze Project
+on:
+  workflow_call:
+jobs:
+  windows:
+    name: Windows 🪟 (PVS-Studio)
+    runs-on: windows-2022
+    defaults:
+      run:
+        shell: pwsh
+    steps:
+      - uses: actions/checkout@v4
+        with:
+          submodules: recursive
+          fetch-depth: 0
+
+      - name: Build OBS Studio 🧱
+        uses: ./.github/actions/build-obs
+        env:
+          TWITCH_CLIENTID: ${{ secrets.TWITCH_CLIENT_ID }}
+          TWITCH_HASH: ${{ secrets.TWITCH_HASH }}
+          RESTREAM_CLIENTID: ${{ secrets.RESTREAM_CLIENTID }}
+          RESTREAM_HASH: ${{ secrets.RESTREAM_HASH }}
+          YOUTUBE_CLIENTID: ${{ secrets.YOUTUBE_CLIENTID }}
+          YOUTUBE_CLIENTID_HASH: ${{ secrets.YOUTUBE_CLIENTID_HASH }}
+          YOUTUBE_SECRET: ${{ secrets.YOUTUBE_SECRET }}
+          YOUTUBE_SECRET_HASH: ${{ secrets.YOUTUBE_SECRET_HASH }}
+          GPU_PRIORITY_VAL: ${{ secrets.GPU_PRIORITY_VAL }}
+        with:
+          target: x64
+          config: Debug
+
+      - name: Run PVS-Studio Analysis
+        uses: ./.github/actions/windows-analysis
+        with:
+          pvsUsername: ${{ secrets.PVS_NAME }}
+          pvsKey: ${{ secrets.PVS_KEY }}
+          target: x64
+          config: Debug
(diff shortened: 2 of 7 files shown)


.github/actions/analyse-obs/action.yaml
```diff
@@ -0,0 +1,60 @@
...
+  config:
+    description: Build Configuration
+    required: true
+runs:
+  using: composite
+  steps:
+    - name: 'Setup PVS Studio'
+      shell: pwsh
+      run: |
+        Invoke-WebRequest -Uri https://cdn.pvs-studio.com/PVS-Studio_setup.exe -OutFile ..\PVS-Studio_setup.exe
```
PatTheMav, line 21: We should use hash checks here as with all other external code and binaries we download in our CI workflows/actions.

.github/actions/analyse-obs/action.yaml
```diff
@@ -0,0 +1,60 @@
...
+        $pvsParams = @(
+            "--progress"
+            "--disableLicenseExpirationCheck"
+            "-p", "${{ inputs.target }}"
+            "-c", "${{ inputs.config }}"
+            "-t", "${{ github.workspace }}\build_x64\obs-studio.sln"
+            "-o", "${{ github.workspace }}\analysis.plog"
+            "-C", "${env:GITHUB_ACTION_PATH}\obs.pvsconfig"
+        )
+        & "C:\Program Files (x86)\PVS-Studio\PVS-Studio_Cmd.exe" @pvsParams
```
PatTheMav, line 41: Should use `Invoke-External` here (which includes exit code handling already).

.github/actions/analyse-obs/action.yaml
```diff
@@ -0,0 +1,60 @@
...
+        Write-Output "PVS Exit Code (non-0 is expected): $LASTEXITCODE"
+        exit 0
+
+    - name: Convert Analysis to Sarif
+      shell: pwsh
+      run: |
+        & 'C:\Program Files (x86)\PVS-Studio\PlogConverter.exe' -a GA:1,2 -d V1042,Renew -t Sarif "${{ github.workspace }}\analysis.plog"
```
PatTheMav, line 48: See above: Use `Invoke-External`.

.github/workflows/build-project.yaml
```diff
@@ -365,6 +365,15 @@ jobs:
           target: x64
           config: ${{ needs.check-event.outputs.config }}
 
+      - name: Run PVS-Studio Analysis
```
PatTheMav, line 368: I'd rather have us use a new `analyze-project` workflow that runs in parallel to the build workflow. This one could then be extended to use the clang analyse build or ASAN and other code quality checkers.

PatTheMav (commented):
I think in general it should be fine to have it run on Windows builds at first and use clang's analysis on Linux and macOS.

.github/actions/analyse-obs/action.yaml
derrod: Since the exit code will be pretty much always non-zero (since 256 means that it found something, see the [documentation](https://pvs-studio.com/en/docs/manual/0035/)) that would always throw, so I'd have to wrap it in a `try...catch` block anyway.

The best thing to do here would probably be to write a little script that actually handles the return codes properly and only throws for those that are actual errors.

.github/actions/analyse-obs/action.yaml
derrod: Unfortunately they don't seem to have versioned download URLs, so this would break every time they change it. So a hash could be set as non-secret actions variable, or we could rely on the code signature (via `Get-AuthenticodeSignature`).

derrod:
Made the following changes:

- Switched to using chocolatey to install PVS as it allows us to pin a version, additionally it should help with @notr1ch's concerns about CDN reliability as the package will be downloaded from Chocolatey's repository
- Changed to using `Invoke-External` for activation/conversion calls
- Added flags enum with error codes to properly log errors for the main binary, ignoring the code that only signals that code issues were found (`256`)
- Renamed action to `windows-analysis` since this is Windows-specific and Linux/macOS will be separate 
- Split it into `analyze-project` workflow that runs in parallel, also uses `Debug` configuration instead of release

.github/actions/windows-analysis/action.yaml
```diff
@@ -0,0 +1,88 @@
...
+            "-t", "Sarif"
+            "${{ github.workspace }}\analysis.plog"
+        )
+        Invoke-External "C:\Program Files (x86)\PVS-Studio\PlogConverter.exe" @conversionParams
+
+    - name: 'Upload PVS Logs'
+      uses: actions/upload-artifact@v4
+      with:
+        name: 'pvs-analysis'
+        path: "${{ github.workspace }}/analysis.*"
```
PatTheMav, line 82: Are the extensions of the generated files known or are they fully arbitrary (e.g. does PVS add a randomised string to the file names)?

derrod: There's `analysis.plog` and `analysis.plog.sarif` so I simply put in a wildcard to get both.

.github/workflows/analyze-project.yaml
```diff
@@ -0,0 +1,39 @@
+name: Analyze Project
+on:
+  workflow_call:
+jobs:
+  windows-analysis:
```
PatTheMav, line 5: I think "windows" should suffice in this case as the "analysis" part is provided by the workflow name.

If it has to be part of the job name, then I'd prefer "analysis-windows", as it fits better with "analysis-linux" and "analysis-macros" later.

.github/actions/windows-analysis/action.yaml
PatTheMav: Ah suspected as much that it's additional suffixes.

PatTheMav (approved):

.github/actions/windows-analysis/action.yaml
```diff
@@ -0,0 +1,90 @@
+name: Run PVS-Studio analysis
+inputs:
+  pvsUsername:
+    description: PVS Studio Licence Username
```
RytoEX, line 4:
```suggestion
    description: PVS-Studio License Username
```

.github/actions/windows-analysis/action.yaml
```diff
@@ -0,0 +1,90 @@
...
+    required: true
+  pvsKey:
+    description: PVS Studio Licence Key
```
RytoEX, line 7:
```suggestion
    description: PVS-Studio License Key
```

.github/actions/windows-analysis/action.yaml
```diff
@@ -0,0 +1,90 @@
...
+  target:
+    description: Build Target
+    required: true
+  config:
+    description: Build Configuration
+    required: true
+runs:
+  using: composite
+  steps:
+    - name: 'Setup PVS Studio'
```
RytoEX, line 18:
```suggestion
    - name: 'Setup PVS-Studio'
```

.github/actions/windows-analysis/action.yaml
```diff
@@ -0,0 +1,90 @@
...
+      shell: pwsh
+      run: |
+        choco install pvs-studio --version=7.30.81185.980 -y --no-progress
+
+    - name: Activate PVS Studio
```
RytoEX, line 23:
```suggestion
    - name: Activate PVS-Studio
```

.github/actions/windows-analysis/action.yaml
```diff
@@ -0,0 +1,90 @@
...
+        & "C:\Program Files (x86)\PVS-Studio\PVS-Studio_Cmd.exe" @pvsParams
+        
+        # Success and CodeErrorsFound are fine as error codes, we only care if it is anything but those
+        $pvs_result = $LASTEXITCODE -band (-bnot [PVSErrorCodes]::CodeErrorsFound)
+        if ($pvs_result -ne 0) {
+            Write-Output "PVS Errors: $([PVSErrorCodes]$pvs_result)"
+        }
+        exit $pvs_result
+
+    - name: Convert Analysis to Sarif
```
RytoEX, line 65:
```suggestion
    - name: Convert Analysis to SARIF
```

.github/actions/windows-analysis/action.yaml
```diff
@@ -0,0 +1,90 @@
...
+
+    - name: 'Upload PVS Logs'
+      uses: actions/upload-artifact@v4
+      with:
+        name: 'pvs-analysis-log'
+        path: |
+          ${{ github.workspace }}/analysis.plog
+          ${{ github.workspace }}/analysis.plog.sarif
+
+    - name: 'Upload PVS report'
```
RytoEX, line 86:
```suggestion
    - name: 'Upload PVS Report'
```

.github/workflows/analyze-project.yaml
```diff
@@ -0,0 +1,39 @@
+name: Analyze Project
+on:
+  workflow_call:
+jobs:
+  windows:
+    name: Windows 🪟 (PVS Studio)
```
RytoEX, line 6:
```suggestion
    name: Windows 🪟 (PVS-Studio)
```

RytoEX (changes requested):
"PVS-Studio" is apparently the proper name of the product. SARIF is an acronym, so let's treat it that was in descriptive text to avoid confusion between that and "serif".

.github/actions/windows-analysis/action.yaml
```diff
@@ -0,0 +1,90 @@
+name: Run PVS-Studio analysis
```
RytoEX, line 1:
```suggestion
name: Run PVS-Studio Analysis
```

.github/actions/windows-analysis/action.yaml
```diff
@@ -0,0 +1,90 @@
...
+  target:
+    description: Build Target
+    required: true
+  config:
+    description: Build Configuration
+    required: true
+runs:
+  using: composite
+  steps:
+    - name: 'Setup PVS-Studio'
```
RytoEX, line 18:
```suggestion
    - name: Setup PVS-Studio
```

.github/actions/windows-analysis/action.yaml
```diff
@@ -0,0 +1,90 @@
...
+      run: |
+        choco install pvs-studio --version=7.30.81185.980 -y --no-progress
+
+    - name: Activate PVS-Studio
+      shell: pwsh
+      run: |
+        . ${env:GITHUB_ACTION_PATH}\Invoke-External.ps1
+        Invoke-External "C:\Program Files (x86)\PVS-Studio\PVS-Studio_Cmd.exe" credentials -u ${{ inputs.pvsUsername }} -n ${{ inputs.pvsKey }}
+
+    - name: Run PVS Analysis
```
RytoEX, line 29:
```suggestion
    - name: Run PVS-Studio Analysis
```

Not sure if there is a distinction here.

.github/actions/windows-analysis/action.yaml
```diff
@@ -0,0 +1,90 @@
...
+        
+        $conversionParams = @(
+            "-a", "GA:1,2",
+            "-d", "V1042,Renew"
+            "-t", "Sarif"
+            "${{ github.workspace }}\analysis.plog"
+        )
+        Invoke-External "C:\Program Files (x86)\PVS-Studio\PlogConverter.exe" @conversionParams
+
+    - name: 'Upload PVS Logs'
```
RytoEX, line 78:
```suggestion
    - name: Upload PVS Logs
```

.github/actions/windows-analysis/action.yaml
```diff
@@ -0,0 +1,90 @@
...
+      uses: actions/upload-artifact@v4
+      with:
+        name: 'pvs-analysis-log'
+        path: |
+          ${{ github.workspace }}/analysis.plog
+          ${{ github.workspace }}/analysis.plog.sarif
+
+    - name: 'Upload PVS Report'
```
RytoEX, line 86:
```suggestion
    - name: Upload PVS Report
```

RytoEX (changes requested):
Nits, but otherwise seems fine.

RytoEX (approved):

PR #10713, opened 2024-05-23
UI:Implement drag-and-drop from sources to scenes

Diog0o:
### Description

The Drag-and-Drop feature allows users to effortlessly move sources or groups of sources into scenes within the user interface. This intuitive functionality enhances user interaction by providing a visual and interactive way to organize and customize content.

The feature mimics the behaviour of a copy and paste. For example if we drag a "Input Audio Device" from Scene 1 to Scene 2, after the event, both Scene 1 and Scene 2 will have the "Input Audio Device" source.

Start:
State of Scene 1 and Scene 2
image
image

During the Drag and Drop from the "Audio Input Capture" from Scene 1 to Scene 2:
Imagem WhatsApp 2024-05-23 às 12 20 10_d1a6b6ce

After the drag and drop operation:
image

### Motivation and Context

As live streaming and digital content creation grow, efficient tools are essential. The Drag-and-Drop feature in OBS simplifies the organization of sources, mimicking copy-and-paste behavior. By allowing users to easily duplicate sources between scenes, it improves workflow and ensures consistency, making OBS even more user-friendly and effective.

We got this feature idea from the Ideas page, and here is the link to the post: https://ideas.obsproject.com/posts/225/drag-and-drop-sources-into-other-scenes

### How Has This Been Tested?

We conducted extensive manual testing to ensure the reliability and functionality of the Drag-and-Drop feature. The testing process was carried out on Windows 11 to verify compatibility and performance, ensuring a seamless user experience across different scenarios.

### Types of changes

### Checklist:

### DIFF
--- a/UI/scene-tree.cpp
+++ b/UI/scene-tree.cpp
@@ -1,16 +1,25 @@
+#include "obs-frontend-api.h"
+#include "obs-source.h"
+#include "obs.h"
 #include "scene-tree.hpp"
+#include <obs-module.h>
 
-#include <QSizePolicy>
-#include <QScrollBar>
 #include <QDropEvent>
+#include <QMimeData>
 #include <QPushButton>
+#include <QScrollBar>
+#include <QSizePolicy>
 #include <QTimer>
 #include <cmath>
 
+
+OBS_DECLARE_MODULE()
+
 SceneTree::SceneTree(QWidget *parent_) : QListWidget(parent_)
 {
+	setAcceptDrops(true);
 	installEventFilter(this);
-	setDragDropMode(InternalMove);
+	setDragDropMode(DragDrop);
 	setMovement(QListView::Snap);
 }
 
@@ -104,10 +113,32 @@ void SceneTree::startDrag(Qt::DropActions supportedActions)
 
 void SceneTree::dropEvent(QDropEvent *event)
 {
-	if (event->source() != this) {
-		QListWidget::dropEvent(event);
-		return;
-	}
+	 if (event->mimeData()->hasFormat("application/indexes")) {
+		QByteArray encodedData = event->mimeData()->data(
+			"application/indexes");
+		QDataStream stream(&encodedData, QIODevice::ReadOnly);
+
+		QStringList sourceNames;
+		stream >> sourceNames;
+
+		QPoint dropPos = event->position().toPoint();
+		QListWidgetItem *sceneItem = itemAt(dropPos);
+
+		if (sceneItem) {
+			QString sceneName = sceneItem->text();
+
+			for (const QString &sourceName : sourceNames) {
+				if (!sceneName.isEmpty() &&
+				    !sourceName.isEmpty()) {
+
+					addSourceToScene(sourceName, sceneName);
+				}
+			}
+		}
+
+		event->setDropAction(Qt::MoveAction);
+		event->accept();
+	 }
 
 	if (gridMode) {
 		int scrollWid = verticalScrollBar()->sizeHint().width();
@@ -147,6 +178,34 @@ void SceneTree::dropEvent(QDropEvent *event)
 
 	QTimer::singleShot(100, [this]() { emit scenesReordered(); });
 }
+
+void SceneTree::addSourceToScene(const QString &sourceName,
+				 const QString &sceneName)
+{
+	obs_source_t *source = obs_get_source_by_name(qPrintable(sourceName));
+	if (!source) {
+		return;
+	}
+	obs_source_t *sceneSource =
+		obs_get_source_by_name(qPrintable(sceneName));
+	if (!sceneSource) {
+		obs_source_release(source);
+		return;
+	}
+
+	obs_scene_t *scene = obs_scene_from_source(sceneSource);
+	if (!scene) {
+		obs_source_release(source);      
+		obs_source_release(sceneSource);
+		return;
+	}
+
+	obs_sceneitem_t *sceneItem = obs_scene_add(scene, source);
+
+	obs_source_release(source);
+	obs_source_release(sceneSource);
+}
+
 
 void SceneTree::RepositionGrid(QDragMoveEvent *event)
 {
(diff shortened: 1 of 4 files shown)


UI/source-tree.cpp
```diff
@@ -1,27 +1,30 @@
...
-#include <QHBoxLayout>
-#include <QMouseEvent>
-#include <QAccessible>
 
+#include <QDrag>
+#include <QMimeData>
 #include <QStylePainter>
 #include <QStyleOptionFocusRect>
 
+
```
tytan652, line 27: Unnecessary whitespace

UI/scene-tree.cpp
```diff
@@ -1,16 +1,25 @@
...
+#include <QMimeData>
 #include <QPushButton>
+#include <QScrollBar>
+#include <QSizePolicy>
 #include <QTimer>
 #include <cmath>
 
+
+OBS_DECLARE_MODULE()
+
```
tytan652, line 17:
```suggestion
```

This has nothing to do here, I guess this is why `<obs-module.h>` was added.

UI/source-tree.cpp
```diff
@@ -1075,6 +1096,119 @@ void SourceTree::UpdateIcons()
...
+	boxPainter.drawPixmap(0, 0, pixmap);
+	boxPainter.end();
+
+	drag->setPixmap(blueBox);
+	drag->setHotSpot(dragStartPosition - boundingRect.topLeft());
+
+	Qt::DropAction dropAction = drag->exec(supportedActions);
+}
+
+
```
tytan652, line 1192: Unnecessary whitespace

UI/source-tree.cpp
```diff
@@ -1075,6 +1096,119 @@ void SourceTree::UpdateIcons()
...
+	QPixmap blueBox(pixmap.size());
+	blueBox.fill(QColor(0, 0, 153, 127));
+	QPainter boxPainter(&blueBox);
+	boxPainter.drawPixmap(0, 0, pixmap);
+	boxPainter.end();
+
+	drag->setPixmap(blueBox);
+	drag->setHotSpot(dragStartPosition - boundingRect.topLeft());
+
+	Qt::DropAction dropAction = drag->exec(supportedActions);
```
tytan652, line 1189:
```suggestion
	drag->exec(supportedActions);
```
The returned object is unused

UI/source-tree.cpp
```diff
@@ -1613,16 +1747,64 @@ void SourceTree::Remove(OBSSceneItem item, OBSScene scene)
...
+		obs_scene_find_source(scene, qPrintable(sourceName));
+	if (sceneItem) {
+		obs_sceneitem_remove(sceneItem);
+	}
+
+	obs_source_release(source);
+	obs_source_release(sceneSource);
 }
 
+
```
tytan652, line 1807: Unnecessary whitespace

UI/source-tree.hpp
```diff
@@ -1,19 +1,20 @@
...
-#include <QCheckBox>
+#include <QPointer>
 #include <QStaticText>
-#include <QSvgRenderer>
-#include <QAbstractListModel>
 #include <QStyledItemDelegate>
-#include <obs.hpp>
-#include <obs-frontend-api.h>
+#include <QSvgRenderer>
+#include <QVector>
```
tytan652, line 15: Don't reorder includes

UI/source-tree.cpp
```diff
@@ -1075,6 +1096,119 @@ void SourceTree::UpdateIcons()
...
+	boxPainter.end();
+
+	drag->setPixmap(blueBox);
+	drag->setHotSpot(dragStartPosition - boundingRect.topLeft());
+
+	Qt::DropAction dropAction = drag->exec(supportedActions);
+}
+
+
+bool SourceTree::get_selected_source_names(obs_scene_t *scene,
```
tytan652, line 1193:
```suggestion
static get_selected_source_names(obs_scene_t */*scene*/,
```

This static function has no reason to be made public.

UI/source-tree.hpp
```diff
@@ -172,6 +174,13 @@ class SourceTree : public QListView {
 
 	void SelectItem(obs_sceneitem_t *sceneitem, bool select);
 
+	static bool get_selected_source_names(obs_scene_t *scene,
+					      obs_sceneitem_t *sceneitem,
+					      void *param);
```
tytan652, line 179:
```suggestion
```
I think declaring the static function as public is not needed at at all.

UI/source-tree.hpp
```diff
@@ -193,10 +202,17 @@ public slots:
...
+	virtual void startDrag(Qt::DropActions supportedActions) override;
 
 	virtual void
 	selectionChanged(const QItemSelection &selected,
 			 const QItemSelection &deselected) override;
+	virtual void dragMoveEvent(QDragMoveEvent *event) override;
+	virtual void dragEnterEvent(QDragEnterEvent *event) override;
+
+private:
+	QPoint dragStartPosition;
```
tytan652, line 215: Move it before the first `public:`, it's private by default.

UI/scene-tree.cpp
```diff
@@ -1,16 +1,25 @@
+#include "obs-frontend-api.h"
+#include "obs-source.h"
+#include "obs.h"
```
tytan652, line 3: Those includes are wrong. And before fixing them check if those are necessary.

UI/scene-tree.cpp
```diff
@@ -1,16 +1,25 @@
...
 #include "scene-tree.hpp"
+#include <obs-module.h>
```
tytan652, line 5:
```suggestion
```

Unnecessary include addition, this one is meant for mainly plugins

tytan652 (changes requested):
Don't reorder includes if you add some at the same time, it add *noise* to the readability of the PR.

Please for each added include check if the compiler requires them to be added.

gxalpha (changes requested):
> The Drag-and-Drop feature allows users to effortlessly move sources or groups of sources into scenes within the user interface. This intuitive functionality enhances user interaction by providing a visual and interactive way to organize and customize content.

I would dispute the word "intuitive" here. Personally I'm against this change, if anything I feel like a source should be moved to the other scene, not copied. But scenes aren't groups, personally I'd rather not have either.
Though to be clear, this isn't my call to make, I just wanted to voice my opinion. For now there's no need to act on the above (any action would just be closing the PR anyways).


As for the implementation:
It removes any indication of where an internal move (from source-tree to source-tree) lands - this isn't great. Additionally, there's no indication for where a copied source ends up and whether it succeeds.
Additionally, it appears to be quite buggy. Dragging a source onto the current scenes duplicates it, but then removing the second source either deletes both or straight up crashes the app. Please look into that, as well as the consequences for the undo/redo (I don't believe it's implemented at all; but it should be). For reference, I tested this on macOS.

Also, I'd recommend using the C++-Objects `OBSSource` and `OBSSourceAutoRelease` to avoid `obs_source_release` calls (this might even fix the other problem if you're lucky).

Warchamp7:
While I understand the desire here, there needs to be more structural changes to the Scenes dock before I'd be comfortable supporting this.

Right now the Scenes dock changes the active scene upon item selection when not in Studio mode. In the context of this PR it leads to one of two outcomes:

- Scene becomes active: Inadvertent scene change to the scene the source was dragged to
- Scene does not become active: Lack of feedback on user action

We may revisit this functionality in the future if/when such adjustments to the Scenes dock functionality happens.

PR #10716, opened 2024-05-24
UI: Adjust appearance of multiview borders and labels

Warchamp7:
### Description
Adjusts the size of multiview borders and the label styling to be less opaque to avoid covering actual scene content.

**Before**
obs64_fuz6xrIlx1

**After**
image

Preview/Program indicator borders are important information, so increasing the border width greatly improves their readability in some multiview layouts and display sizes/resolutions.

The size and position of the label box is adjusted to be based off the label's height plus the border width to look more consistent and fix the text not being vertically centered.

The background and label text are made less opaque to ensure they don't completely block visual content.

-----

### Motivation and Context

Many hardware multiviews either place the labels outside/around the content, or on top with no background. BlackMagic ATEM is only of the only examples that places a black background behind the label and even in that case, it is less opaque than our current labels.

image
image
image
image

### How Has This Been Tested?
Tested all multiview layouts on multiple different sized displays at different resolutions.

### Types of changes

### Checklist:

### DIFF
--- a/UI/multiview.cpp
+++ b/UI/multiview.cpp
@@ -34,9 +34,9 @@ static OBSSource CreateLabel(const char *name, size_t h)
 	OBSDataAutoRelease font = obs_data_create();
 
 	std::string text;
-	text += " ";
+	text += "  ";
 	text += name;
-	text += " ";
+	text += "  ";
 
 #if defined(_WIN32)
 	obs_data_set_string(font, "face", "Arial");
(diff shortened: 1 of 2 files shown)


Lain-B (approved):

Lain-B:
Did a quick test on freetype text source as well and still seems good.

DeeDeeG:
IMO, the labels in the example screenshot with the "cooking show" (ATEM?) are more readable. In my view, some more opacity than the "After" screenshot would help, similar to the ATEM (?) example.

FWIW, the cooking show (ATEM?) example appears to have more opacity on the Program and Preview labels as a special case. I find their non-Program, non-Preview labels on the borderline of not contrasty enough for visibility, but more readable than the "After" screenshot nonetheless.

(If it's helpful: Accessibility guidelines might recommend more contrast, though I'm not totally sure, to be honest. For example, I know of this tool to check against web standard recommended contrast levels for accessibility: https://accessibilitycheck.org/contrast/ which supports transparency.)

(Just my 2 cents. Disclaimer: I don't use multi-view, just my feedback from viewing the screenshots.)

PR #10717, opened 2024-05-24
cmake: Do not write build number file if OBS_BUILD_NUMBER set

kamalmostafa:
### Description
Do not create the build number file if OBS_BUILD_NUMBER is configured externally. ~Create the build number file under CMAKE_BINARY_DIR (never write anything to CMAKE_SOURCE_DIR).~ Allows for configuration and build from a read-only-mounted source dir.

### Motivation and Context
Docker-based builds with read-only-mounted source dir.

### How Has This Been Tested?
With just this change, I can construct a working docker-based build that mounts the source dir on the host to a read-only dir in the container -- i.e. everything else in the configuration, build, and packaging steps does respect -B /some/other/build/dir

### Types of changes

### Checklist:


kamalmostafa:
Dummy dockerfile which demonstrates the problem:
```
#!/usr/bin/env -S docker build . -f
#
# Demo of obs-studio read-only bind mounted source dir build problem
#
# Read-only mounts ./obs-studio (on the host) to /tmp/obs-studio
# (in the container)

FROM alpine
RUN apk add cmake
RUN --mount=type=bind,target=/tmp/obs-studio,source=./obs-studio \
        cmake -S /tmp/obs-studio -B /tmp/builddir

# This dummy 'cmake' fails when it tries to write the build number file into
# the source dir (/tmp/obs-studio):
#
#     0.193 CMake Error at cmake/Modules/VersionConfig.cmake:114 (file):
#     0.193   file failed to open for writing (Read-only file system):
#     0.193 
#     0.193     /tmp/obs-studio/cmake/.CMakeBuildNumber
#
# Aside from this single write into the source dir, a full build using a
# read-only-mounted source dir (and a container stuffed with dependencies)
# does work fine, writing exclusively to /tmp/builddir.
```

PatTheMav:
This defeats the purpose of the cache, which is to provide a locally unique build number regardless of location or number of build configurations and would lead to collisions between different build configurations made with the same checkout.

The repo always had the expectation that the entire checkout directory is writable (e.g. the legacy codebase copies all generated artefacts into the `additional_install_files` directory).

I would prefer that if an externally-provided build number override exists, that creation of the cache would be skipped. That way all you‘d have to do is ensure the uniqueness of the build number yourself and the checkout itself can be read-only.

kamalmostafa:
Yep, that works fine for my read-only mount build too.  @PatTheMav ?

PatTheMav (approved):
Seems fine.

RytoEX (changes requested):
The correct commit message prefix and PR prefix is "cmake: ". Otherwise, at a glance, this seems fine.

PR #10718, opened 2024-05-24
UI: Fix Qt platform override in CMake 3

tytan652:
### Description

USE_XDG definition is missing in CMake 3 code in the UI subdirectory.

This override is still needed until Qt minimum version requirement is increased to at least 6.3.0.

### Motivation and Context

We still support Qt 6.2 and Ubuntu 22.04.

The override should enforce the Wayland platform where Qt was enforcing the X11 on GNOME Wayland.
The enforce was removed in 6.3.

### How Has This Been Tested?

Not tested.

### Types of changes

### Checklist:

### DIFF
--- a/UI/cmake/os-linux.cmake
+++ b/UI/cmake/os-linux.cmake
@@ -1,5 +1,5 @@
 target_sources(obs-studio PRIVATE platform-x11.cpp)
-target_compile_definitions(obs-studio PRIVATE OBS_INSTALL_PREFIX="${OBS_INSTALL_PREFIX}")
+target_compile_definitions(obs-studio PRIVATE USE_XDG OBS_INSTALL_PREFIX="${OBS_INSTALL_PREFIX}")
 target_link_libraries(obs-studio PRIVATE Qt::GuiPrivate Qt::DBus)
 
 if(TARGET OBS::python)
--- a/UI/cmake/legacy.cmake
+++ b/UI/cmake/legacy.cmake
@@ -479,16 +479,22 @@ elseif(OS_POSIX)
     target_compile_options(obs PRIVATE -Wno-unqualified-std-cast-call)
   endif()
 
-  if(OS_LINUX AND ENABLE_WHATSNEW)
-    find_package(MbedTLS)
-    find_package(nlohmann_json REQUIRED)
-    if(NOT MBEDTLS_FOUND)
-      obs_status(FATAL_ERROR "mbedTLS not found, but required for WhatsNew support on Linux")
+  if(OS_LINUX)
+    if(USE_XDG)
+      target_compile_definitions(obs PRIVATE USE_XDG)
     endif()
 
-    target_sources(obs PRIVATE update/crypto-helpers.hpp update/crypto-helpers-mbedtls.cpp update/shared-update.cpp
-                               update/shared-update.hpp update/update-helpers.cpp update/update-helpers.hpp)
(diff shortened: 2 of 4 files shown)


UI/cmake/os-linux.cmake
```diff
@@ -8,6 +8,11 @@ if(TARGET OBS::python)
   target_link_options(obs-studio PRIVATE LINKER:-no-as-needed)
 endif()
 
+# USE_XDG is only used once for a workaround with ealier versions of Qt
+if(Qt6_VERSION VERSION_LESS 6.3)
+  target_compile_definitions(obs-studio PRIVATE USE_XDG)
+endif()
```
RytoEX, line 14: As far as I can tell, USE_XDG is already defined for all UI files that include `UI/ui-config.h` if it is defined via CMake:
https://github.com/obsproject/obs-studio/blob/52660beae2af3b5cbb792823248d52d628dcfdcb/UI/ui-config.h.in#L3
https://github.com/obsproject/obs-studio/blob/52660beae2af3b5cbb792823248d52d628dcfdcb/UI/obs-app.cpp#L79

If USE_XDG is set via CMake, the former would become:
```cpp
#define USE_XDG <value>
```

We just no longer define it as an option by default.

RytoEX (commented):
The changes in `UI/obs-app.cpp` seem fine. I'm not sure about the changes in `UI/cmake/os-linux.cmake`.

UI/cmake/os-linux.cmake
tytan652: The CMake option has been dropped in version 3, USE_XDG is only defined to libobs on Linux.

But for the UI, since it is only used for one thing, I preferred softly deprecate its use.
Once we drop support for Qt 6.2, USE_XDG in UI and the override can be removed.

RytoEX: The only thing that was dropped in CMake 3 was that we no longer expose the option by default. You can still set `USE_XDG` in CMake 3 and it will be configured and defined. See:
https://github.com/obsproject/obs-studio/blob/52660beae2af3b5cbb792823248d52d628dcfdcb/UI/CMakeLists.txt#L55

image

tytan652: I will not add back the option, this CMake 2 only thing.

I will just remove the Qt version check in CMake, and put it as a common compilation define like in libobs.

RytoEX: I wasn't suggesting to add back the option. I'm pointing out that you can, with CMake 3, simply define `USE_XDG=true`, and the `#cmakedefine USE_XDG` will correctly transform into `#define USE_XDG` such that `#if defined(USE_XDG)` is true. There is no need to add it as an expliciti CMake `option()`.

UI/cmake/os-linux.cmake
```diff
@@ -1,5 +1,5 @@
 target_sources(obs-studio PRIVATE platform-x11.cpp)
-target_compile_definitions(obs-studio PRIVATE OBS_INSTALL_PREFIX="${OBS_INSTALL_PREFIX}")
+target_compile_definitions(obs-studio PRIVATE USE_XDG OBS_INSTALL_PREFIX="${OBS_INSTALL_PREFIX}")
```
PatTheMav, line 2: This adds the `USE_XDG` definition unconditionally, is that desired?

tytan652: You did the same in libobs.

tytan652: It is desired, since the code relies on defining USE_XDG.

PR #10726, opened 2024-05-25
libobs: Fix buffer overrun in video_frame_init

derrod:
### Description

Fixes buffer overrun in `video_frame_init()`

### Motivation and Context

`offsets[]` has size `MAX_AV_PLANES - 1` but the loop would iterate over `MAX_AV_PLANES`, thus resulting in it writing to the memory past the last item.

It is worth noting that in reality this would've never overran, as `MAX_AV_PLANES` is `8`, but the highest amount of planes a video frame can have is 4 (YUV + Alpha).

Found by ~~@notr1ch~~ PVS Studio.

### How Has This Been Tested?

Hasn't.

### Types of changes

### Checklist:

### DIFF
--- a/libobs/media-io/video-frame.c
+++ b/libobs/media-io/video-frame.c
@@ -196,7 +196,7 @@ void video_frame_init(struct video_frame *frame, enum video_format format,
 	size_t size = 0;
 	uint32_t linesizes[MAX_AV_PLANES];
 	uint32_t heights[MAX_AV_PLANES];
-	size_t offsets[MAX_AV_PLANES - 1];
+	size_t offsets[MAX_AV_PLANES];
 	int alignment = base_get_alignment();
 
 	if (!frame)


libobs/media-io/video-frame.c
PatTheMav, line 1: Wouldn't it be simpler to just make the `offsets` array the same size as the others (namely `MAX_AV_PLANES`)? It's not part of an API, so there is no ABI break and makes for simpler code.

derrod: Either works, I just went this route since the last offset doesn't need to be written/stored.

derrod:
This code could use a few other improvements as well, those shouldn't be `continue`s but `break`s and instead of `MAX_AV_PLANES` we should just have a constant that is `4`, i.e. the actual max number of video planes.

notr1ch:
Found by PVS Studio, not me!

PatTheMav:
> This code could use a few other improvements as well, those shouldn't be `continue`s but `break`s and instead of `MAX_AV_PLANES` we should just have a constant that is `4`, i.e. the actual max number of video planes.

In general I'm always in favour of "when you re-visit your code, refactor while making changes" (aka "the agile way") though I know that RytoEx is not a fan of that.

RytoEX:
> In general I'm always in favour of "when you re-visit your code, refactor while making changes" (aka "the agile way") though I know that RytoEx is not a fan of that.

Depends on the scope of the changes. If refactoring makes more sense than a change that doesn't require refactoring, and the refactoring isn't huge, then it's fine. If it's so large that reviewing the refactoring takes significantly longer than a smaller, more targeted fix, then I'd prefer fixing the bug first, then refactoring, so we can merge/ship the bug fix faster.

If the refactor itself is so trivial that it's comparable to the more-targeted fix, then I don't mind doing a refactor that fixes a bug. That said, since we have the bug fix here, we can just review/merge it as-is and make the other fixes separately.

Lain-B (approved):

PR #10727, opened 2024-05-25
mac-avcapture: Improve av_capture_sync_info formatting

gxalpha:
### Description

Adds a comma to the end of the definition to improve on an... *interesting* formatting choice by clang-format.

### Motivation and Context

Noticed this when working on something else, it made my eyes hurt.
A comma at the end is established practice in the rest of the code base.

### How Has This Been Tested?

Ran clang-format to get this new formatting.
Shouldn't have

### Types of changes

### Checklist:

### DIFF
--- a/plugins/mac-avcapture/plugin-main.m
+++ b/plugins/mac-avcapture/plugin-main.m
@@ -284,23 +284,22 @@ bool obs_module_load(void)
 
     obs_register_source(&av_capture_info);
 
-    struct obs_source_info av_capture_sync_info = {.id = "macos-avcapture-fast",
-                                                   .type = OBS_SOURCE_TYPE_INPUT,
-                                                   .output_flags = OBS_SOURCE_VIDEO | OBS_SOURCE_CUSTOM_DRAW |
-                                                                   OBS_SOURCE_AUDIO | OBS_SOURCE_SRGB |
-                                                                   OBS_SOURCE_DO_NOT_DUPLICATE,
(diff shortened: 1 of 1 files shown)


PatTheMav:
Did you by any chance check which format setting influences this behaviour? Would be interesting to know if this is a bug or indeed a setting that behaves exactly as it should and instead we need to tweak the setting (otherwise we're just adding cosmetic changes instead of fixing the actual issue).

gxalpha:
I'm... confused. Seems like the column limit of 120 alone causes this? Whatever combinations I try it always becomes this exactly when I remove the column limit override for Objective-C.
This shouldn't really be the cause, but I can't figure out which setting it is, and I have no explanation of why adding a comma fixes it.

Lain-B (approved):

Lain-B:
This has been a thing forever and I'm not sure why clang-format is like that, but I usually prefer it this way so I merged it anyway (admittedly before I read the commentary). I don't know and I *doubt* it's a bug with clang-format, I just know clang-format has had this behavior since as long as I can remember. I think it's just like, a contextual thing and possibly intentional.

RytoEX:
As an aside, a quick review of some other struct initializations shows the final member definition proceeded by a comma in all cases that I checked. I did not check the entire codebase.
https://github.com/search?q=repo%3Aobsproject%2Fobs-studio%20obs_source_info&type=code

PR #10730, opened 2024-05-25
UI: Actually copy-paste filters instead of remembering the source

gxalpha:
### Description

The current behavior makes OBS remember the source that filters should
get copied from and then clones that source's filters to the source that
filters are pasted to. This means however that if the original source
has filters added or removed after the copy-operation, that will still
effect the paste-operation, so filters that were added after copying
would still get pasted.
This commit changes this to something similar to the source clipboard.
Now, we're actually copying the filters, so if the original source has
filters added to afterwards we don't copy those as well.

Depends on #10723 because code in similar locations is touched
and I would like to avoid merge conflicts later.

### Motivation and Context

Requested by @Warchamp7 on [Discord](https://discord.com/channels/348973006581923840/374636084883095554/1243673820167143514).

### How Has This Been Tested?

macOS 14.5
Copied filters via the right-click menu. Added one or more filters to the source.
Pasted filters to another source via the right-click menu. Only the copied filters were pasted, not the ones added afterwards.

### Types of changes

### Checklist:


GeorgesStavracas (approved):
Approved from the UI / UX side, needs code review

Warchamp7 (approved):

Warchamp7 (changes requested):
This appears to correctly not copy new filters added to a source after the Copy operation happened.

However if a filter is removed from the source after Copy, then it is also not included in the Paste afterwards.

gxalpha:
Marking as draft as I still have to fix the issue above, probably by just storing the filter as json or something.

PR #10732, opened 2024-05-25
UI: Fix Qt 6.7 checkbox signal deprecations

gxalpha:
### Description

qt/qtbase@3512fb1ec5ff088772170540c4e91b1886fbea45 (also cherry-picked to https://github.com/qt/qtbase/commit/02ba4b14dc1c0cee802d8d840315833fda2449b4) deprecated the stateChanged signal of QCheckBoxes in favor of a new checkStateChanged signal. The signals are the same, except that now the enum type is passed explicitly (before the enum was passed as an argument but defined as an int).

The documentation states that it's deprecated for 6.9 but I got the warning now already (dev version appears to be 6.8 currently) and it's also picked to 6.7 and the alternative exists since 6.7 so I'm not 100% what the "real" deprecation version is but 6.7 makes sense.

### Motivation and Context

Got a deprecation warning when compiling against Qt dev branch.

### How Has This Been Tested?

macOS 14.5 with Qt dev branch.

Tested all the code paths (settings, properties, transform, youtube) except the Windows `HideOBSWindowWarning` one (because I don't have a Windows environment) to see that the signal still gets sent and works.

### Types of changes

### Checklist:

### DIFF
--- a/UI/properties-view.cpp
+++ b/UI/properties-view.cpp
@@ -273,7 +273,11 @@ QWidget *OBSPropertiesView::AddCheckbox(obs_property_t *prop)
 
 	QCheckBox *checkbox = new QCheckBox(QT_UTF8(desc));
 	checkbox->setCheckState(val ? Qt::Checked : Qt::Unchecked);
+#if QT_VERSION >= QT_VERSION_CHECK(6, 7, 0)
+	return NewWidget(prop, checkbox, &QCheckBox::checkStateChanged);
+#else
 	return NewWidget(prop, checkbox, &QCheckBox::stateChanged);
+#endif
 }
 
 QWidget *OBSPropertiesView::AddText(obs_property_t *prop, QFormLayout *layout,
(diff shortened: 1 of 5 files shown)


RytoEX:
At a glance, this looks okay, though I haven't test compiled it.

RytoEX (approved):
Looks okay to me. Did a test build on Qt 6.6.3 and 6.7.1.

RytoEX:
> The documentation states that it's deprecated for 6.9 but I got the warning now already (dev version appears to be 6.8 currently) and it's also picked to 6.7 and the alternative exists since 6.7 so I'm not 100% what the "real" deprecation version is but 6.7 makes sense.

For added context:
As far as I can tell, the deprecation is version-guarded on 6.9, but maybe the dev branch just hadn't actually had its version bumped yet. I'd guess that the commit was cherry-picked so that the Qt 6.7 docs site would display the deprecation.

PR #10733, opened 2024-05-26
obs-ffmpeg: Check correct VA-API codec support

kkartaltepe:
### Description

Previously, we just checked for H264 regardless of which codec was selected. This mostly worked, but on Fedora they ship AV1 without H264 or HEVC. In that case users wont see AV1 as an option.

### Motivation and Context

We really should have been checking the right codec, and users will want to use AV1 without special drivers.

### How Has This Been Tested?

I dont see any changes on my AMD machine. I'll test a fedora machine later.

### Types of changes

### Checklist:

### DIFF
--- a/plugins/obs-ffmpeg/obs-ffmpeg-vaapi.c
+++ b/plugins/obs-ffmpeg/obs-ffmpeg-vaapi.c
@@ -1074,6 +1074,21 @@ static bool get_device_name_from_pci(struct pci_access *pacc, char *pci_slot,
 	return false;
 }
 
+static bool vaapi_device_codec_supported(const char *path,
+					 enum codec_type codec)
+{
+	switch (codec) {
+	case CODEC_H264:
+		return vaapi_device_h264_supported(path);
+	case CODEC_HEVC:
+		return vaapi_device_hevc_supported(path);
+	case CODEC_AV1:
+		return vaapi_device_av1_supported(path);
+	default:
+		return false;
+	}
+}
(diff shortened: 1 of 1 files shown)


plugins/obs-ffmpeg/obs-ffmpeg-vaapi.c
```diff
@@ -1074,6 +1074,20 @@ static bool get_device_name_from_pci(struct pci_access *pacc, char *pci_slot,
...
+{
+	switch (codec) {
+	case CODEC_H264:
+		return vaapi_device_h264_supported(path);
+	case CODEC_HEVC:
+		return vaapi_device_hevc_supported(path);
+	case CODEC_AV1:
+		return vaapi_device_av1_supported(path);
+	}
+	return false;
```
tytan652, line 1088: Nit:

```suggestion
	default:
		return false;
	}
```

kkartaltepe: Actually i want it outside the switch to keep all the branches covered. With default we wont hit the warning if someone adds a new codec (unlikely, but still).

tytan652: Normally we should remove `-Wno-error=switch`, but QSV…

tytan652 (approved):
LGTM

Lain-B (approved):

PR #10738, opened 2024-05-27
Fix Mastering Display Luminance Metadata

derrod:
### Description

Fixes mastering display metadata in various encoder plugins.

### Motivation and Context

The H.264[^1] and H.265[^2] specifications state:

> When min_display_mastering_luminance is not in the range of 1 to 50000, the nominal maximum display luminance of the mastering display is unknown or unspecified or specified by other means not specified in this Specification.

The AV1 specification does not appear to have any such limitation.[^3]

#### AMF

Set minimum to 1 for HEVC/AVC and 0 for AV1.

#### QSV

Set minimum to 1 for HEVC/AVC, and 0 for AV1. Additionally the maximum value is no longer multiplied by 10,000, as the header states `MaxDisplayMasteringLuminance is in units of 1 candela per square meter.` So the multiplication is not necessary and resulted in us writing a maximum luminance of 10 million nits by default, whoops.

#### VideoToolbox

Set minimum to 1 always since AV1 is not supported, unsure about ProRes requirements but it's probably fine.

#### Further notes

It seems some commerically released Blu-Rays have shipped with illegal minimum luminance values, so FFmpeg will likely special-case 0 as an allowed value unless strict compliance is enabled:
- https://ffmpeg.org/pipermail/ffmpeg-devel/2024-May/328170.html
- https://github.com/mpv-player/mpv/issues/14177

It also seems that the value stored in the container (MP4 et al.) is sometimes used over the one stored in the codec bitstream, so we haven't really heard of any issues reported with this, but that could also be because HDR is pretty niche and the most popular HDR-capable encoder in OBS (NVENC) does not actually support the mastering display metadata SEI/OBU.

**Edit:** The commit introducing the validation to FFmpeg is fairly recent[^4], so it's also likely that users just haven't encountered it yet.

[^1]: https://www.itu.int/rec/T-REC-H.264-202108-I/en - Section D.2.29 Mastering display colour volume SEI message semantics
[^2]: https://www.itu.int/rec/T-REC-H.265-202309-I/en - Section D.3.28 Mastering display colour volume SEI message semantics
[^3]: https://aomediacodec.github.io/av1-spec/#metadata-high-dynamic-range-mastering-display-color-volume-syntax
[^4]: https://github.com/FFmpeg/FFmpeg/commit/1c45104f74e59beb25ce7451c98527f7a3c518f6

### How Has This Been Tested?

TBD

### Types of changes

### Checklist:

### DIFF
--- a/plugins/mac-videotoolbox/encoder.c
+++ b/plugins/mac-videotoolbox/encoder.c
@@ -217,7 +217,7 @@ static CFDataRef obs_to_vt_masteringdisplay(uint32_t hdr_nominal_peak_level)
 	mdcv.white_point[1] = __builtin_bswap16(16450);
 	mdcv.max_display_mastering_luminance =
 		__builtin_bswap32(hdr_nominal_peak_level * 10000);
-	mdcv.min_display_mastering_luminance = 0;
+	mdcv.min_display_mastering_luminance = 1;
 
 	UInt8 bytes[sizeof(struct mastering_display_colour_volume)];
 	memcpy(bytes, &mdcv, sizeof(bytes));


plugins/obs-qsv11/obs-qsv11.c
```diff
@@ -635,8 +635,9 @@ static void update_params(struct obs_qsv *obsqsv, obs_data_t *settings)
 		obsqsv->params.WhitePointX = 15635;
 		obsqsv->params.WhitePointY = 16450;
 		obsqsv->params.MaxDisplayMasteringLuminance =
-			hdr_nominal_peak_level * 10000;
-		obsqsv->params.MinDisplayMasteringLuminance = 0;
+			hdr_nominal_peak_level;
+		obsqsv->params.MinDisplayMasteringLuminance =
+			obsqsv->codec == QSV_CODEC_AV1 ? 0 : 1;
```
jpark37, line 640: Did you check the metadata? Don't just trust a header comment.

jpark37 (changes requested):
0 was intentional. The value that's supposed to go in this field is the user's display minimum so we know what they are and aren't seeing in theory. It can get complicated because your display might happen to not be great, but the underlying HDR feed has a lower minimum luminance for one example. Rather than try to sort this all out, It seemed more convenient to say "I don't know" than try to plumb the value. 0 seems perfectly legal for that. 1 is a lie.
I would ask why NVENC is missing, but I'm not sure I want this change. The QSV bugfix for maximum is desirable if it's actually broken. Feels like something I would have checked, but I may have overlooked it.

jpark37:
It feels like both specs have copy/paste bugs in them because why should the minimum property affect maximum behavior but not minimum behavior?
https://www.itu.int/en/about/Pages/contact.aspx
Let me know what they say. :P

plugins/obs-qsv11/obs-qsv11.c
derrod: Hadn't been able to check yet, hence I left it as a draft. Double checked it now and the header is a lie! Reverted this change now.

Also turns out https://github.com/obsproject/obs-studio/commit/4c6e2a6722de4ed923cf29bc0bf65ddd8f45a2cf broke QSV on my NVIDIA+Intel system so that's fun...

jpark37 (changes requested):
It feels like both specs have copy/paste bugs in them because why should the min property affect max behavior but not min behavior? This also makes the same feature of the max property redundant if you accept the text at face value.
https://www.itu.int/en/about/Pages/contact.aspx
Let me know what they say. :P

PR #10739, opened 2024-05-27
cmake: Fix RPATH on Linux on non-relocatable build

### LINKED ISSUE #10737: Trying to build Flatpak from Git , `libobs-opengl.so` not found
### Operating System Info

Other

### Other OS

Fedora 40

### OBS Studio Version

Git

_No response_

### OBS Studio Log URL

https://gist.github.com/AFCMS/ce8400fcc09288c00478cf4bc2ea95c2

(not a obsproject.com uploaded log file because the interface didn't show up at all)

### OBS Studio Crash Log URL

_No response_

### Expected Behavior

I would like the software to start and work.

### Current Behavior

I wanted to build the latest Git version of OBS.

For that I cloned recursively the Git repo and executed the following:
```shell
flatpak-builder --user --install build-dir --force-clean ./build-aux/com.obsproject.Studio.json
```

The build worked fine but when I run `flatpak run --user com.obsproject.Studio` (no other versions of OBS are installed in Flatpak) in my terminal a small popup appear with `Failed to initialize video:  Graphics module not found`, while in logs there are two error lines:
```
error: os_dlopen(libobs-opengl.so.30->libobs-opengl.so.30): libobs-opengl.so.30: Ne peut ouvrir le fichier d'objet partagé: Aucun fichier ou dossier de ce type
error: Failed to initialize video:  Graphics module not found
```

image

### Steps to Reproduce

I am unsure, really. I would have expected the Flatpak to work straight since it's a cross-platform format and the build process went smoothly.

I did multiple tests on my hardware/config and all lead to the same result.

_No response_

tytan652:
### Description

Fixes #10737

RPATH is missing, so once the Flatpak is installed (not ran with gnome-builder of flatpak-builder) it doesn't search in `/app/lib` for `dlopen`.

### Motivation and Context

Fix a quite big issue.

### How Has This Been Tested?

Build and install the Flatpak, master does not found libobs-opengl and with this PR it does.

### Types of changes

### Checklist:


PatTheMav:
Not setting _any_ `RPATH` by default is by design: The default build assumes an installation into the distribution-specific system location which implies using the default search paths for libraries, so no `RPATH` is needed (which seems to be the default for such packages).

Anything else implies "relocatable" (as in: a different location than a distribution default) and uses relative `RPATH`s for maximum compatibility - wouldn't this also suffice for Flatpak?

tytan652:
Flatpak have two system location, `/usr` for the runtime and `/app` for the application.

Without it `dlopen` is unable to find libobs-opengl and I don't plan to use `LD_LIBRARY_PATH` to fix that.

tytan652:
@PatTheMav, I'll try to dig a little to find a better solution but for now this is all I have.

tytan652:
Put as draft, and undraft/close it after investigating Flatpak is done.

tytan652:
Close over #10741

PR #10740, opened 2024-05-27
obs-qsv11: Fix QSV failing on multi-vendor multi-GPU systems

derrod:
### Description

Fixes Intel QSV not working on setups where the Intel GPU is not idx 0.

### Motivation and Context

QSV started failing on my NVIDIA dGPU+Intel dGPU system with https://github.com/obsproject/obs-studio/commit/4c6e2a6722de4ed923cf29bc0bf65ddd8f45a2cf applied.

The IDX fed into `MFXCreateSession()` only counts Intel GPUs, so we need to adjust the index down if there are non-Intel GPUs earlier in the list.

This is already done in the QSV test app, but was missing here: https://github.com/obsproject/obs-studio/blob/master/plugins/obs-qsv11/obs-qsv-test/obs-qsv-test.cpp#L73-L86

### How Has This Been Tested?

Tested on system with RTX 4090 and Arc A750.

### Types of changes

### Checklist:

### DIFF
--- a/plugins/obs-qsv11/common_utils_windows.cpp
+++ b/plugins/obs-qsv11/common_utils_windows.cpp
@@ -33,26 +33,31 @@ mfxStatus Initialize(mfxVersion ver, mfxSession *pSession,
 	obs_video_info ovi;
 	obs_get_video_info(&ovi);
 	mfxU32 adapter_idx = ovi.adapter;
+	mfxU32 idx_adjustment = 0;
 
 	// Select current adapter - will be iGPU if exists due to adapter reordering
 	if (codec == QSV_CODEC_AV1 && !adapters[adapter_idx].supports_av1) {
-		for (mfxU32 i = 0; i < 4; i++) {
+		for (mfxU32 i = 0; i < MAX_ADAPTERS; i++) {
+			if (!adapters[i].is_intel) {
+				idx_adjustment++;
+				continue;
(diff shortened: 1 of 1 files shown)


kkartaltepe:
adapters are supposed to be indexed based on enumeration in plugins/obs-qsv11/obs-qsv-test/obs-qsv-test.cpp we should correct the indexes within the tester ideally.

derrod:
> adapters are supposed to be indexed based on enumeration in plugins/obs-qsv11/obs-qsv-test/obs-qsv-test.cpp we should correct the indexes within the tester ideally.

The test just enumerates all GPUs on the system, whether they're Intel or not. When I ported the NVENC tester to Linux I made it use CUDA to enumerate devices to skip all non-NVIDIA GPUs, perhaps something similar could be done here with MFX?

For now I think this is probably "good enough" as a fix, long term that stuff should probably be rewritten.

kkartaltepe:
> For now I think this is probably "good enough" as a fix, long term that stuff should probably be rewritten.

since its windows only I dont feel too strongly. I just noticed the test app also has this index adjustment...

kkartaltepe (approved):

RytoEX (approved):
Tested on my Intel+NVIDIA laptop. Seems fine. The CI failure is unrelated.

PR #10741, opened 2024-05-27
cmake,UI: Fix SOVERSION on Linux

### LINKED ISSUE #10737: Trying to build Flatpak from Git , `libobs-opengl.so` not found

tytan652:
### Description

Fixes #10737
Supersedes #10739

SOVERSION is either the major number of a semver or an arbitrary number incremented at each API/ABI break.

### Motivation and Context

CMake 3 does not apply RPATH like CMake 2 so it unraveled this issue.

We didn't released a CMake 3.0 Linux build, so no breaking change.

### How Has This Been Tested?

Build and installed the Flatpak locally and OBS finds `libobs-opengl.so.30`.

Non-Flatpak local build seems to work too.

### Types of changes

### Checklist:

### DIFF
--- a/UI/CMakeLists.txt
+++ b/UI/CMakeLists.txt
@@ -97,8 +97,11 @@ endif()
 foreach(graphics_library IN ITEMS opengl metal d3d11)
   string(TOUPPER ${graphics_library} graphics_library_U)
   if(TARGET OBS::libobs-${graphics_library})
-    target_compile_definitions(obs-studio
-                               PRIVATE DL_${graphics_library_U}="$<TARGET_FILE_NAME:OBS::libobs-${graphics_library}>")
+    target_compile_definitions(
+      obs-studio
+      PRIVATE
+        DL_${graphics_library_U}="$<$<IF:$<PLATFORM_ID:Windows>,TARGET_FILE_NAME,TARGET_SONAME_FILE_NAME>:OBS::libobs-${graphics_library}>"
+    )
   else()
     target_compile_definitions(obs-studio PRIVATE DL_${graphics_library_U}="")
   endif()
--- a/cmake/linux/helpers.cmake
+++ b/cmake/linux/helpers.cmake
@@ -55,8 +55,8 @@ function(set_target_properties_obs target)
   elseif(target_type STREQUAL SHARED_LIBRARY)
     set_target_properties(
       ${target}
-      PROPERTIES VERSION ${OBS_VERSION_MAJOR}
-                 SOVERSION ${OBS_VERSION_CANONICAL}
+      PROPERTIES VERSION ${OBS_VERSION_CANONICAL}
+                 SOVERSION ${OBS_VERSION_MAJOR}
                  BUILD_RPATH "${OBS_OUTPUT_DIR}/$<CONFIG>/${OBS_LIBRARY_DESTINATION}"
                  INSTALL_RPATH "${OBS_LIBRARY_RPATH}")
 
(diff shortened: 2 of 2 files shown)


UI/CMakeLists.txt
```diff
@@ -97,8 +97,11 @@ endif()
 foreach(graphics_library IN ITEMS opengl metal d3d11)
   string(TOUPPER ${graphics_library} graphics_library_U)
   if(TARGET OBS::libobs-${graphics_library})
-    target_compile_definitions(obs-studio
-                               PRIVATE DL_${graphics_library_U}="$<TARGET_FILE_NAME:OBS::libobs-${graphics_library}>")
+    target_compile_definitions(
+      obs-studio
+      PRIVATE
+        DL_${graphics_library_U}="$<$<IF:$<BOOL:${OS_LINUX}>,TARGET_SONAME_FILE_NAME,TARGET_FILE_NAME>:OBS::libobs-${graphics_library}>"
```
PatTheMav, line 103: Is this change necessary to get the _versioned_ name of the library set as a define?

tytan652: To get the SONAME one only on Linux yes. `TARGET_FILE_NAME` will put `libobs-opengl.so.30.1.0` where we need the SONAME (`libobs-opengl.so.30`) given by `TARGET_SONAME_FILE_NAME`.

Otherwise, we just create the same breakage we filename reversed.

AFCMS (commented):
Tested, works too

UI/CMakeLists.txt
PatTheMav: I'd check if just using `TARGET_SONAME_FILE_NAME` for all platforms shouldn't work out of the box - we set `NO_SONAME` for macOS and Windows doesn't support that anyway, so it would be interesting to see if that generator expression falls back on the target name then.

Otherwise I'd be fine using a conditional branch here instead (you can compose a string to _contain_ a generator expression, it doesn't need to be used at the immediate call site).

tytan652: > I'd check if just using `TARGET_SONAME_FILE_NAME` for all platforms shouldn't work out of the box - we set `NO_SONAME` for macOS and Windows doesn't support that anyway, so it would be interesting to see if that generator expression falls back on the target name then.

I'm only fixing the Linux path, I don't plan to test CMake behavior on other platform.

> Otherwise I'd be fine using a conditional branch here instead (you can compose a string to _contain_ a generator expression, it doesn't need to be used at the immediate call site).

I preferred to keep it simple and avoid duplicating code and/or effort in the three platform file.

PatTheMav: We cannot just fix a thing for one platform, we always have to code with _all_ supported platforms in mind, particularly for code that effects all platforms (in this case `libobs-opengl` as it's not a platform-specific target).

tytan652: Changed the logic to use `PLATFORM_ID`

AFCMS (commented):
Still works for me

PR #10751, opened 2024-05-29
libobs/util: Improve inhibit functions on Linux

### LINKED ISSUE #9968: "Failed to create xdg-screensaver: 2" error in log
### Operating System Info

Other

### Other OS

Debian 12

### OBS Studio Version

30.0.0

30.0.0 64-bit

### OBS Studio Log URL

https://obsproject.com/logs/2QYKBCszHhmN1lAF

### OBS Studio Crash Log URL

https://obsproject.com/logs/8z3SRmi03uSgFK7N

### Expected Behavior

"Failed to create xdg-screensaver: 2", and any other errors, should not occur.

### Current Behavior

Errors in the log

### Steps to Reproduce

OBS mostly works fine. However, intermittently, and without warning, I'm logged out of the user session, which closes OBS and loses the file it was recording. I haven't got to the bottom of it, and it may well not be due to OBS. However, OBS has always been running when this has happened. Thus, if it's possible to resolve any errors in OBS, it would be useful, to help eliminate it from the root cause.

It's running on a standard system, not in a docker container, or anything like that.

tytan652:
### Description

Closes #9968

Avoid creating the screensaver inhibitor thread when another method has been applied.

Make inhibition unavailable in a Flatpak if portal is not available.

### Motivation and Context

1. If the Portal or the D-Bus method is in place, adding the xdg-screensaver thread is overkill. It should be just a fallback.
2. Flatpak only support the Portal method, so spamming xdg-screensaver will do nothing

### Types of changes

### Checklist:

### DIFF
--- a/libobs/util/platform-nix.c
+++ b/libobs/util/platform-nix.c
@@ -752,10 +752,21 @@ os_inhibit_t *os_inhibit_sleep_create(const char *reason)
 
 #if defined(GIO_FOUND)
 	info->portal = portal_inhibit_info_create();
-	if (!info->portal)
-		info->dbus = dbus_sleep_info_create();
-#endif
+	if (!info->portal) {
+		/* In a Flatpak, only the portal can be used for inhibition. */
+		if (access("/.flatpak-info", F_OK) == 0) {
+			bfree(info);
+			return NULL;
+		}
 
+		info->dbus = dbus_sleep_info_create();
+	}
+
+	if (info->portal || info->dbus) {
+		info->reason = bstrdup(reason);
+		return info;
+	}
+#endif
 	os_event_init(&info->stop_event, OS_EVENT_TYPE_AUTO);
 	posix_spawnattr_init(&info->attr);
 
@@ -812,6 +823,10 @@ bool os_inhibit_sleep_set_active(os_inhibit_t *info, bool active)
 		portal_inhibit(info->portal, info->reason, active);
 	if (info->dbus)
 		dbus_inhibit_sleep(info->dbus, info->reason, active);
+	if (info->portal || info->dbus) {
+		info->active = active;
+		return true;
+	}
 #endif
 
 	if (!info->stop_event)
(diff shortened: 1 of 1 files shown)


libobs/util/platform-nix.c
```diff
@@ -752,10 +752,19 @@ os_inhibit_t *os_inhibit_sleep_create(const char *reason)
...
+			bfree(info);
+			return NULL;
+		}
+
 		info->dbus = dbus_sleep_info_create();
-#endif
+	}
 
+	if (info->portal || info->dbus)
+		goto end;
```
derrod, line 766: Rather than making this a `goto` and introducing the ifdef'd label at the end, can you not just make this an early return?

For example
```c
if (info->portal || info->dbus) {
	info->reason = bstrdup(reason);
	return info;
}
```

kkartaltepe (approved):

libobs/util/platform-nix.c
tytan652: Since we are in C, I tried to avoid duplicating code too much.

derrod: Sure but by avoiding to duplicate 2 lines you've added three for the guarded goto label, which is pretty ugly. I'd prefer a simple early return as it also exists in the other case you've added. But Since this whole thing should probably be reworked at some point it's probably fine for now if we want to merge it for 30.2. I'll leave that up to @kkartaltepe

kkartaltepe: personally I prefer just copying the code in this case, but you commented first.

tytan652: If you agree with Rodney, don't approve the PR without a comment indicating you would like the same change.

Removed all added `goto` with copied return.

libobs/util/platform-nix.c
```diff
@@ -752,10 +752,21 @@ os_inhibit_t *os_inhibit_sleep_create(const char *reason)
 
 #if defined(GIO_FOUND)
 	info->portal = portal_inhibit_info_create();
-	if (!info->portal)
+	if (!info->portal) {
+		/* In a Flatpak, only the portal can be used for inhibition */
```
RytoEX, line 756:
```suggestion
		/* In a Flatpak, only the portal can be used for inhibition. */
```

RytoEX (changes requested):
Nit.

libobs/util/platform-nix.c
tytan652: Done

RytoEX (approved):

PR #10755, opened 2024-05-31
UI: Add 'Toggle preview with window focus' option

mfro:
### Description

This pull request will implement a new preview option 'Toggle preview with window focus'. This setting will cause the preview to disable when the preview window loses focus, and re-enable when gaining focus.
image

### Motivation and Context

In some cases, the stream preview can cause reduced performance. An option to enable it only when the preview window is focused would make it easier to get the performance benefits of disabling it without manually toggling it each time it is needed.

### How Has This Been Tested?

Edition Windows 10 Pro N
Version 22H2
Installed on  ‎3/‎24/‎2021
OS build  19045.4412
Experience  Windows Feature Experience Pack 1000.19056.1000.0

Tested tabbing in and out of the preview window with the feature both enabled and disabled.

### Types of changes

### Checklist:


UI/window-basic-main.cpp
```diff
@@ -5261,6 +5261,20 @@ void OBSBasic::changeEvent(QEvent *event)
...
+					"WindowFocusTogglePreview");
+
+		if (!isActiveWindow()) {
+			previewEnabledWhileFocused = previewEnabled;
+
+			if (enablePreviewToggle) {
+				DisablePreview();
+			}
+		} else if (enablePreviewToggle && previewEnabledWhileFocused) {
+			EnablePreview();
```
notr1ch, line 5276: You probably want to use `EnablePreviewDisplay` like the code above as these functions do not work in Studio mode.

mfro: Ah yeah makes sense thanks. Updated

nabeelre:
This is great!!

matoi974:
I think the wording here should be "Disable preview when the OBS window is not active", not everyone knows what "focus" means in this context.

mfro:
Good point, I think that wording is more clear. I also moved the event handler into OBSApp so the preview isn't disabled by other windows or popups within OBS.

Warchamp7:

If this is resubmitted, I would like to see it adjusted so that the preview enabled state instead becomes a dropdown with three options:

**Enable Preview**
- Always
- When focused
- Never

PR #10760, opened 2024-06-01
docs: Add versionadded for 30.x functions

CodeYan01:
### Motivation and Context

Makes it easier to find out if a function is already in a certain release.
### How Has This Been Tested?

Built and viewed locally

### Types of changes

### Checklist:

### DIFF
--- a/docs/sphinx/reference-core.rst
+++ b/docs/sphinx/reference-core.rst
@@ -544,6 +544,8 @@ Video, Audio, and Graphics
 
    Resets all audio monitoring devices.
 
+   .. versionadded:: 30.1
+
 ---------------------
 
 .. function:: void obs_enum_audio_monitoring_devices(obs_enum_audio_device_cb cb, void *data)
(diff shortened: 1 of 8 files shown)


RytoEX (commented):
My one question is how the version-added on possible enum parameter values looks, since those are unique compared to the rest of the changes here.

RytoEX:
It's a little weird to me that we're defining an enum's values in the documentation for a function that uses it as a parameter, but that's a separate problem, I suppose. For context, here are some other places we define enums:
https://docs.obsproject.com/reference-libobs-graphics-graphics?highlight=enum#graphics-enumerations
https://docs.obsproject.com/reference-libobs-graphics-effects?highlight=enum#c.gs_effect_get_param_info
https://docs.obsproject.com/reference-libobs-media-io#c.video_format

Though, there are also other places where we do this same sort of inline definition:
https://docs.obsproject.com/reference-libobs-util-config-file?highlight=enum#c.config_open

If it's a distinct type, we should probably just document the enum itself somewhere instead of trying to shove it inline in function documentation. However, as I said, that's a separate issue from this PR's goals.

RytoEX (approved):
This looks fine. My comments above are out-of-scope for this PR.

Issue #9506, opened 2023-08-24
@ not working in youtube dock


bzzzbrasil:
### Operating System Info

Windows 10

### OBS Studio Version

30.0.0-beta2

### OBS Studio Log URL

https://obsproject.com/logs/52rMjBHupywQfn4m

### Expected Behavior

when entering the email address of the google account we should be able to enter the @

### Current Behavior

the @ key (french keyboard) does not work in the youtube dock

### Steps to Reproduce

1.just try to enter an @
2.
3.
...

### Anything else we should know?

Have a good day

Fenrirthviti:
Which dock exactly shows this issue? Can you provide a screenshot or more clear steps to reproduce? There are several places that a login prompt might be provided, we need to know which one you are trying to use.

Issue #7353, opened 2022-09-11
YouTube chat input box covers chat


icvidanova:
### Operating System Info

Windows 11

### OBS Studio Version

28.0.1

### OBS Studio Log URL

https://obsproject.com/logs/NUtDZZRXabkNJFT7

### Expected Behavior

The chat would display entirely, regardless of size

### Current Behavior

The message box covers chat if the chat height is small/minimum height
![image](https://user-images.githubusercontent.com/105178667/189552751-80e2fb8e-dab9-4349-96ba-6818d3875602.png)

### Steps to Reproduce

1. Connect to YouTube
2. Start a livestream
3. Send a message in chat
4. Make the chat dock small

### Anything else we should know?

A chat pop-out from the YouTube studio website fits in an even smaller window so I would assume my intended size should be achievable
![image](https://user-images.githubusercontent.com/105178667/189552899-b3cafbdc-f81a-4906-b72a-57731b3ab689.png)

gxalpha:
Vertically, these are the smallest dock sizes I could create without something getting cut off (making it smaller than this would cut have something cut off as you described):

<p>
<img align="top" width="170" alt="Bildschirm­foto 2022-09-12 um 01 46 59" src="https://user-images.githubusercontent.com/59806498/189554566-35943d33-7a6d-411d-a3be-4d8b9b879a2c.png">
<img align="top" width="170" alt="Bildschirm­foto 2022-09-12 um 01 45 46" src="https://user-images.githubusercontent.com/59806498/189554562-a99bf3a5-0f3a-4ed7-aa4d-6d57afa710b8.png">
<img align="top" width="170" alt="Bildschirm­foto 2022-09-12 um 01 48 54" src="https://user-images.githubusercontent.com/59806498/189554738-854f1175-d119-4330-9c37-763b8e114f78.png">
<img align="top" width="170" alt="Bildschirm­foto 2022-09-12 um 01 50 11" src="https://user-images.githubusercontent.com/59806498/189554737-fb57cfda-0f3f-43c8-b087-cc22a69e231b.png">
</p>

This is both with and without the chat box, and with and without our custom CSS. It suggests that the browser itself still has the same minimum height, but our custom chat box makes the dock larger so a dock with the same size will get cut off. Perhaps we could make the minimum size smaller as part of the script that hides the original chat.

PR #10770, opened 2024-06-03
UI: Enable first-party YouTube Chat features in OBS

msuman-google:
### Description
Unlock the full feature set of the YouTube Chat dock in OBS by removing custom scripting/CSS logic. Enable the signed-in experience for live streaming content creators while also sharing login credentials with the YouTube Control panel dock.

### Motivation and Context
This will allow OBS users to utilize features _already_ supported in the YouTube Chat plugin, such as

* creating polls
* managing Q&A sessions
* a rich emoji set in the input panel
* emoji fountains
* moderation tools

and many more. These features are available to users who are logged-in to YouTube Chat and/or the YouTube Control panel.

### How Has This Been Tested?
Built OBS locally on Linux workstation, verified existing functionality for YouTube Chat continues to work in signed-out state. Signing-in to YouTube hides the native Qt input panel in favour of the YouTube native input panel that unlocks all of the first-party functionality for live streaming content creators.

### Types of changes
  Fixes: https://github.com/obsproject/obs-studio/issues/9506
  Fixes: https://github.com/obsproject/obs-studio/issues/7353

### Checklist:

### DIFF
--- a/UI/auth-youtube.cpp
+++ b/UI/auth-youtube.cpp
@@ -138,14 +138,6 @@ bool YoutubeAuth::LoadInternal()
 	return implicit ? !token.empty() : !refresh_token.empty();
 }
 
-#ifdef BROWSER_AVAILABLE
-static const char *ytchat_script = "\
-const obsCSS = document.createElement('style');\
-obsCSS.innerHTML = \"#panel-pages.yt-live-chat-renderer {display: none;}\
-yt-live-chat-viewer-engagement-message-renderer {display: none;}\";\
-document.querySelector('head').appendChild(obsCSS);";
-#endif
-
 void YoutubeAuth::LoadUI()
 {
 	if (uiLoaded)
@@ -171,7 +163,6 @@ void YoutubeAuth::LoadUI()
 
 	browser = cef->create_widget(chat, YOUTUBE_CHAT_PLACEHOLDER_URL,
 				     panel_cookies);
-	browser->setStartupScript(ytchat_script);
 
 	chat->SetWidget(browser);
 	main->AddDockWidget(chat, Qt::RightDockWidgetArea);
@@ -219,6 +210,7 @@ void YoutubeAuth::ResetChat()
 {
 #ifdef BROWSER_AVAILABLE
 	if (chat && chat->cefWidget) {
+		chat->SetApiChatId("");
 		chat->cefWidget->setURL(YOUTUBE_CHAT_PLACEHOLDER_URL);
 	}
 #endif
@@ -399,13 +391,44 @@ void YoutubeChatDock::SetWidget(QCefWidget *widget_)
 	setWidget(widget);
 
 	cefWidget.reset(widget_);
+
+	QWidget::connect(cefWidget.get(), &QCefWidget::urlChanged, this,
+			 &YoutubeChatDock::YoutubeCookieCheck);
 }
 
 void YoutubeChatDock::SetApiChatId(const std::string &id)
 {
 	this->apiChatId = id;
-	QMetaObject::invokeMethod(this, "EnableChatInput",
-				  Qt::QueuedConnection);
+	QMetaObject::invokeMethod(this, "EnableChatInput", Qt::QueuedConnection,
+				  Q_ARG(bool, !id.empty()));
+}
+
+void YoutubeChatDock::YoutubeCookieCheck()
+{
+	QPointer<YoutubeChatDock> this_ = this;
+	auto cb = [this_](bool currentlyLoggedIn) {
+		bool previouslyLoggedIn = this_->isLoggedIn;
+		this_->isLoggedIn = currentlyLoggedIn;
+		bool loginStateChanged =
+			(currentlyLoggedIn && !previouslyLoggedIn) ||
+			(!currentlyLoggedIn && previouslyLoggedIn);
+		if (loginStateChanged) {
+			QMetaObject::invokeMethod(
+				this_, "EnableChatInput", Qt::QueuedConnection,
+				Q_ARG(bool, !currentlyLoggedIn));
+			OBSBasic *main = OBSBasic::Get();
+			if (main->GetYouTubeAppDock() != nullptr) {
+				QMetaObject::invokeMethod(
+					main->GetYouTubeAppDock(),
+					"SettingsUpdated", Qt::QueuedConnection,
+					Q_ARG(bool, !currentlyLoggedIn));
+			}
+		}
+	};
+	if (panel_cookies) {
+		panel_cookies->CheckForCookie("https://www.youtube.com", "SID",
+					      cb);
+	}
 }
 
 void YoutubeChatDock::SendChatMessage()
@@ -442,9 +465,10 @@ void YoutubeChatDock::ShowErrorMessage(const QString &error)
 			     QTStr("YouTube.Chat.Error.Text").arg(error));
 }
 
-void YoutubeChatDock::EnableChatInput()
+void YoutubeChatDock::EnableChatInput(bool visible)
 {
-	lineEdit->setVisible(true);
-	sendButton->setVisible(true);
+	bool setVisible = visible && !isLoggedIn;
+	lineEdit->setVisible(setVisible);
+	sendButton->setVisible(setVisible);
 }
 #endif
--- a/UI/auth-youtube.cpp
+++ b/UI/auth-youtube.cpp
@@ -55,6 +55,14 @@ static inline void OpenBrowser(const QString auth_uri)
 	QDesktopServices::openUrl(url);
 }
 
+static void DeleteCookies()
+{
+	if (panel_cookies) {
+		panel_cookies->DeleteCookies("youtube.com", "");
+		panel_cookies->DeleteCookies("google.com", "");
+	}
+}
+
 void RegisterYoutubeAuth()
 {
 	for (auto &service : youtubeServices) {
@@ -64,7 +72,7 @@ void RegisterYoutubeAuth()
 				return std::make_shared<YoutubeApiWrappers>(
 					service);
 			},
-			YoutubeAuth::Login, []() { return; });
+			YoutubeAuth::Login, DeleteCookies);
 	}
 }
 
@@ -212,6 +220,15 @@ void YoutubeAuth::ResetChat()
 #ifdef BROWSER_AVAILABLE
 	if (chat && chat->cefWidget) {
 		chat->cefWidget->setURL(YOUTUBE_CHAT_PLACEHOLDER_URL);
+	}
+#endif
+}
+
+void YoutubeAuth::ReloadChat()
+{
+#ifdef BROWSER_AVAILABLE
+	if (chat && chat->cefWidget) {
+		chat->cefWidget->reloadPage();
 	}
 #endif
 }
--- a/UI/auth-youtube.hpp
+++ b/UI/auth-youtube.hpp
@@ -60,6 +60,7 @@ public:
 
 	void SetChatId(const QString &chat_id, const std::string &api_chat_id);
 	void ResetChat();
+	void ReloadChat();
 
 	static std::shared_ptr<Auth> Login(QWidget *parent,
 					   const std::string &service);
--- a/UI/auth-youtube.hpp
+++ b/UI/auth-youtube.hpp
@@ -16,6 +16,7 @@ class YoutubeChatDock : public BrowserDock {
 
 private:
 	std::string apiChatId;
+	bool isLoggedIn;
 	LineEditAutoResize *lineEdit;
 	QPushButton *sendButton;
 	QHBoxLayout *chatLayout;
@@ -26,9 +27,10 @@ public:
 	void SetApiChatId(const std::string &id);
 
 private slots:
+	void YoutubeCookieCheck();
 	void SendChatMessage();
 	void ShowErrorMessage(const QString &error);
-	void EnableChatInput();
+	void EnableChatInput(bool visible);
 };
 #endif
 
--- a/UI/window-dock-youtube-app.cpp
+++ b/UI/window-dock-youtube-app.cpp
@@ -33,20 +33,11 @@ static constexpr const char *INGESTION_STOPPED = "INGESTION_STOPPED";
 
 YouTubeAppDock::YouTubeAppDock(const QString &title)
 	: BrowserDock(title),
-	  dockBrowser(nullptr),
-	  cookieManager(nullptr)
+	  dockBrowser(nullptr)
 {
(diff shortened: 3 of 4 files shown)


UI/auth-youtube.hpp
```diff
@@ -11,24 +11,35 @@
...
 
+protected:
+	void timerEvent(QTimerEvent *event);
+
 public:
-	inline YoutubeChatDock(const QString &title) : BrowserDock(title) {}
+	inline YoutubeChatDock(const QString &title) : BrowserDock(title)
+	{
+		// Check youtube.com login status periodically
+		awaitLoginTimer = startTimer(1000);
```
tytan652, line 34: I'm not sure that checking every second is a good idea.

tytan652: > * This timer is set directly on the QObject backing the YouTube Chat widget. It will only ever trigger if the widget is open.

Dock/widget closed or opened, the widget and its timer still exist. So it's every second until the chat dock (or its timer) is destroyed.

> * We want creators to have the best possible user experience _after_ logging in (either into YouTube Chat or the YouTube Control Panel docks). The 1s timer allows us to provide this experience.
> 
> * The timer probably doesn't need to run as often, post sign-in. I could change the trigger frequency when we detect the user is already signed-in to something like 10s. Would that address your concern?

I'll wait for opinions from other members.

I get that you want a better UX, I get that the browser widgets possibilities are limited (which is normal, it should not be overused), but is checking something every second all the time a good compromise ?

UI/auth-youtube.cpp
```diff
@@ -26,6 +26,7 @@
 
 #ifdef BROWSER_AVAILABLE
 #include "window-dock-browser.hpp"
+#include <QTimerEvent>
```
Ansl55555, line 29: Hab

Ansl55555 (commented):
Hz

Ansl55555 (commented):
Ja

UI/auth-youtube.hpp
```diff
@@ -11,15 +11,22 @@
...
 
 private:
 	std::string apiChatId;
+	int awaitLoginTimer;
+	bool isLoggedIn;
 	LineEditAutoResize *lineEdit;
 	QPushButton *sendButton;
 	QHBoxLayout *chatLayout;
 
+protected:
```
PatTheMav, line 27: Do not use `protected` - its use is discouraged in modern C++ and by the ISO C++ guidelines.

Use `private` instead.

UI/auth-youtube.hpp
PatTheMav: Using a timer for this seems wasteful to me. Instead I'd attempt to hook the navigation events in the browser dock and check for the cookie whenever navigation happens as that suggests a user submitted a form (ideally the login form) and thus a cookie might be set after.

So instead of constantly (possibly for the entire lifetime of the application) checking whether a cookie has been set, the code should check for the cookie at specific checkpoints of activity within the browser dock.

UI/auth-youtube.cpp
```diff
@@ -350,6 +350,36 @@ std::shared_ptr<Auth> YoutubeAuth::Login(QWidget *owner,
 }
 
 #ifdef BROWSER_AVAILABLE
+void YoutubeChatDock::timerEvent(QTimerEvent *event)
+{
+	if (event->timerId() != awaitLoginTimer)
+		return;
+
+	QPointer<YoutubeChatDock> this_ = this;
```
PatTheMav, line 358: What is this intended to accomplish? Usually capturing `this` by itself in the lambda should be enough. If there is the concern that the pointer might become invalid in between triggers of the timer, use the `*this` capture as suggested by ISO C++ and available since C++17 (which is OBS' lower version boundary).

Note that `*this` will create a copy of all class data members, but that is the only way to ensure that `this` can refer to valid data when the lambda is executed.

PatTheMav (changes requested):

UI/auth-youtube.cpp
```diff
@@ -55,6 +56,14 @@ static inline void OpenBrowser(const QString auth_uri)
 	QDesktopServices::openUrl(url);
 }
 
+static void DeleteCookies()
+{
+	if (panel_cookies) {
+		panel_cookies->DeleteCookies("youtube.com", std::string());
```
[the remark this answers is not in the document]
msuman-google: Sure. I was simply following existing precedent:
  - https://github.com/obsproject/obs-studio/blob/ce6f9a4742b40b04e68ef759e0698bd5eac4360e/UI/auth-restream.cpp#L279
  -  https://github.com/obsproject/obs-studio/blob/ce6f9a4742b40b04e68ef759e0698bd5eac4360e/UI/auth-twitch.cpp#L510

UI/auth-youtube.hpp
msuman-google: This is overriding the default QObject implementation, which is defined as a `protected` method. See https://doc.qt.io/qt-6/qobject.html#timerEvent

UI/auth-youtube.hpp
msuman-google: The complexity here is that navigation can occur in independent browser docks (i.e. the Chat and the Control Panel docks). And there is also the case of an expired login (e.g. force signout on all devices) which requires no navigation in either dock. As I mentioned earlier in the thread, we could potentially reduce the frequency of these checks in the signed-in state, if that is preferable. What do you think?

UI/auth-youtube.cpp
msuman-google: Was also following existing precedent here: 
* https://github.com/obsproject/obs-studio/blob/ce6f9a4742b40b04e68ef759e0698bd5eac4360e/UI/auth-twitch.cpp#L442

Please let me know if you still prefer capturing `this` in the lambda directly.

PatTheMav: I discussed this internally and it's fine to re-use the pattern used in the other class, as it's safer versus introducing a different way to handle data capture for the lambda here.

msuman-google: Okay, thanks. Resolving.

UI/auth-youtube.hpp
PatTheMav: Just so I don't understand the scenario right:

* There are several Youtube-related browser docks, some of them are default (non-specialised) browser docks, one of them is the specialised chat dock?
* The desired functionality is for users to log into YouTube in _any_ of these docks but _only_ the chat dock becoming aware of this and thus changing its internal state?
* This is achieved by having all these docks share the same cookie domain and running a timer to constantly check the cookie state and update the chat dock if a cookie state change occurs?
* This is also to notice a expired login (which could be triggered from independent devices)?

msuman-google: > Just so I don't understand the scenario right:
> 
> ~* There are several Youtube-related browser docks, some of them are default (non-specialised) browser docks, one of them is the specialised chat dock?~

There are only two that I know of: 
  * The default service integration dock for YouTube Live Chat, which gets setup once the user sets up their YouTube account for streaming using OAuth.
  * The custom browser dock for YouTube Control Panel, which can be used without an OAuth connection for streaming to YouTube (it uses a stream key instead). 

> * There is a specialised Youtube-specific dock type (`YouTubeAppDock`), used by all YouTube-related docks (chat, control panel, etc.)?

See above. `YouTubeAppDock` is the custom browser dock (i.e. control panel), not shared with the service integration dock. Note that the service integration dock (i.e. chat) is entirely unavailable if you don't use OAuth to connect the account.

> * The desired functionality is for users to log into YouTube in _any_ of these docks but _only_ the chat dock becoming aware of this and thus changing its internal state?

If the login happens in the control panel, the timer in the chat dock will pick up the updated state. If the login happens in chat, the timer in the chat dock will publish a signal for the control panel dock to update. So, yes, only the chat dock updates its internal state but all docks will eventually refresh.

> * This is achieved by having all these docks share the same cookie domain and running a timer to constantly check the cookie state and update the chat dock if a cookie state change occurs?

Correct.

> * This is also to notice a expired login (which could be triggered from independent devices)?

Also correct.

UI/auth-youtube.cpp
```diff
@@ -350,6 +350,36 @@ std::shared_ptr<Auth> YoutubeAuth::Login(QWidget *owner,
...
+void YoutubeChatDock::timerEvent(QTimerEvent *event)
+{
+	if (event->timerId() != awaitLoginTimer)
+		return;
+
+	QPointer<YoutubeChatDock> this_ = this;
+	auto cb = [this_](bool found) {
+		bool previouslyLoggedIn = this_->isLoggedIn;
+		this_->isLoggedIn = found;
+		if (found ^ previouslyLoggedIn) {
```
gxalpha, line 362: Please do not use binary comparators. They are terrible for code readability.

gxalpha (changes requested):

UI/auth-youtube.hpp
RytoEX: My understanding, and @PatTheMav can correct me if I'm wrong, is that this does not properly handle expired logins. If a user logs out of all devices in their browser, the cookies in our CEF instance are unaffected.

RytoEX: Is there a technical reason why CEF's `OnLoadEnd` cannot be used to check for cookies instead of the timer?

Warchamp7: Is it possible to use the existing [BrowserClient::OnLoadEnd](https://github.com/obsproject/obs-browser/blob/8e2b31f53b8a43cd816e37b4abae6d43de941207/browser-client.cpp#L669) to accomplish this instead of the timer?

msuman-google: > Is it possible to use the existing [BrowserClient::OnLoadEnd](https://github.com/obsproject/obs-browser/blob/8e2b31f53b8a43cd816e37b4abae6d43de941207/browser-client.cpp#L669) to accomplish this instead of the timer?

See explanation above:

> If the login happens in the control panel, the timer in the chat dock will pick up the updated state. If the login happens in chat, the timer in the chat dock will publish a signal for the control panel dock to update. 

Although `BrowserClient::OnLoadEnd` can help with handling expired logins, it wouldn't be useful here without some mechanism to communicate login state from the control panel to the chat dock.

UI/auth-youtube.cpp
msuman-google: Done.

UI/auth-youtube.hpp
PatTheMav: To achieve that `YouTubeAppDock` and `YouTubeChatDock` just need to connect a cookie check function to the to the appropriate event, e.g.:

```C++
connect(browser, &QCefWidget::urlChanged, this, &YouTubeCookieCheck);
```

This function would check if `SID` is set for the desired domains and then emit a signal:

```C++
emit <whichever class this is implemented in>->YouTubeSessionAvailable(hasSessionId);
```

The chat dock is also connected to this signal:
```C++
connect(<the class>, &<the class>::YouTubeSessionAvailable, chat, &YoutubeChatDock::UpdateSessionState);
```
and `UpdateSessionState` then invokes the methods to enable the chat docks and update the settings on the app dock.

I build a POC locally and had the chat controls successfully appear and disappear depending on whether I was signed-in via the app dock.

msuman-google: Thanks for the suggestion, I've updated the commit. PTAL and let me know if this new approach addresses your concerns.

UI/auth-youtube.cpp
```diff
@@ -55,6 +56,14 @@ static inline void OpenBrowser(const QString auth_uri)
 	QDesktopServices::openUrl(url);
 }
 
+static void DeleteCookies()
+{
+	if (panel_cookies) {
```
PatTheMav, line 61: Wouldn't this code also delete any existing Google or YouTube sign-in that people might have in their own custom browser docks?

As an example, I might have a custom browser dock set up with a _different_ YouTube account logged in (e.g. my personal account vs my "brand account"), or I might just be logged in to my personal Google account.

Using the global panel cookie manager to store the service-specific cookie information would conflate those two and mean that logging out of my brand account would also break my existing logins.

Vice-versa it would "magically" make the service docks use my personal account that I might be logged-in with in another dock.

IMO the service-specific docks need their own cookie manager (and share it if necessary) and not tap into the cookie manager used by user-managed docks.

UI/auth-youtube.cpp
```diff
@@ -368,6 +378,9 @@ YoutubeChatDock::YoutubeChatDock(const QString &title) : BrowserDock(title)
 			 &YoutubeChatDock::SendChatMessage);
 	QWidget::connect(sendButton, &QPushButton::pressed, this,
 			 &YoutubeChatDock::SendChatMessage);
+
+	// Ensure local state for `isLoggedIn` is updated after UI loads.
+	QTimer::singleShot(1000, this, &YoutubeChatDock::YoutubeCookieCheck);
```
PatTheMav, line 383: Should be superfluous here as `YoutubeCookieCheck` is executed automatically before the URL of the browser dock is changed, or did this not happen before the first URL change in your tests?

UI/auth-youtube.cpp
tytan652: "Custom Browser Dock" are normally not bound to the `panel_cookie` cookie manager.

tytan652: `panel_cookie` is used by service integrations (OAuth browser and docks).

msuman-google: Yes, the shared `panel_cookie` manager is only being used for non-custom (i.e. OBS native in some sense) browser docks.

PatTheMav: Got it, so users should be unaffected by this change. I'd still rather have services not being interfere with each others' browser dock state, but this particular change is harmless then.

UI/auth-youtube.cpp
msuman-google: Okay, removed this.

PatTheMav: Did it work as expected? I hadn't tested if this was still needed, hence why I formulated it as a question.

msuman-google: I was previously trying to account for the case of an existing login and subsequent restart of the OBS app, initializing the API based "send message" support for YouTube Live Chat (as opposed to using the native input panel, provided by the signed-in chat window). This was also when I previously had a bug in the cookie check logic that was reloading the page. With that bug resolved, this now works as expected, even without the initialization bit. Thanks for checking :-)

PatTheMav (approved):
Looks good to me implementation-wise, my main concern about constantly polling the cookie state has been addressed by moving that check into a pre-navigation event of the browser widget.

RytoEX (approved):

PR #10774, opened 2024-06-05
UI: Fix crash when entering settings

notr1ch:
### Description
We refer to VOD tracks as 1-6 but internally the array is 0-5 so there's an off-by-one error here. I don't really like this mix of 1-6 and 0-5 as it's prone to these kind of bugs, but refactoring this is probably best left for another time.

### Motivation and Context
Crash.

### How Has This Been Tested?
Opened settings, no longer crashed.

### Types of changes

### Checklist:


RytoEX (approved):
Reproduced the crash locally without this. Couldn't crash with this patch. LGTM

PR #10776, opened 2024-06-06
UI: Disable Multitrack Video by default

RytoEX:
### Description

Disable Multitrack Video (Twitch Enhanced Broadcasting) by default.

### Motivation and Context

Noticed in testing that this was on by default in all cases. This is a behavior change that would suddenly and without warning change streaming settings for any users configured to stream to Twitch, such as increasing the amount of bandwidth used.

### How Has This Been Tested?

Built and ran locally on Windows 11.

### Types of changes

### Checklist:

### DIFF
--- a/UI/window-basic-main.cpp
+++ b/UI/window-basic-main.cpp
@@ -1706,7 +1706,7 @@ bool OBSBasic::InitBasicConfigDefaults()
 	config_set_default_bool(basicConfig, "Stream1", "IgnoreRecommended",
 				false);
 	config_set_default_bool(basicConfig, "Stream1", "EnableMultitrackVideo",
-				true);
+				false);
 	config_set_default_bool(basicConfig, "Stream1",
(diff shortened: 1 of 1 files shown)


derrod (approved):
Migration and other stuff (data usage consent, etc.) needs to be figured out before we set this as default.

Fenrirthviti:
Agree on this change. This *has* to be an explicit opt-in from a legal standpoint anyway.

Fenrirthviti (approved):

Sean-Der:
This commit also fixes all other outputs. Currently in master if I try to stream I get

image


When I cherry-pick this commit it works again.

RytoEX:
> This commit also fixes all other outputs. Currently in master if I try to stream I get

You can just manually disable Enhanced Broadcasting in Settings -> Stream which should have the same effect in the meantime.

PR #10781, opened 2024-06-06
libobs, UI: Encoder group fixups

tt2468:
### Description
Modifies the encoder group API added previously to better follow the existing libobs API naming paradigms. This also produces much more readable code, and allows a few small additional benefits like removing encoders from an encoder group while it is active.

### Motivation and Context
The API implemented in #10349 is not yet set in stone. The design of the methods is quite obscure, and the internal implementation is difficult to maintain. This code should fix that up by a lot.

### How Has This Been Tested?
Tested on Ubuntu 22.04 - had to hardcode the GPU info for testing, but stream started to a test channel with no issues with three tracks.

### Types of changes

### Checklist:

### DIFF
--- a/libobs/obs-encoder.c
+++ b/libobs/obs-encoder.c
@@ -325,11 +325,10 @@ static void add_connection(struct obs_encoder *encoder)
 	}
 
 	if (encoder->encoder_group) {
-		bool ready = false;
 		pthread_mutex_lock(&encoder->encoder_group->mutex);
-		encoder->encoder_group->encoders_started += 1;
-		ready = encoder->encoder_group->encoders_started ==
-			encoder->encoder_group->encoders_added;
+		encoder->encoder_group->num_encoders_started += 1;
+		bool ready = encoder->encoder_group->num_encoders_started ==
+			     encoder->encoder_group->num_encoders;
 		pthread_mutex_unlock(&encoder->encoder_group->mutex);
 		if (ready)
 			add_ready_encoder_group(encoder);
@@ -353,8 +352,8 @@ static void remove_connection(struct obs_encoder *encoder, bool shutdown)
 
 	if (encoder->encoder_group) {
 		pthread_mutex_lock(&encoder->encoder_group->mutex);
-		encoder->encoder_group->encoders_started -= 1;
-		if (encoder->encoder_group->encoders_started == 0)
+		encoder->encoder_group->num_encoders_started -= 1;
+		if (encoder->encoder_group->num_encoders_started == 0)
 			encoder->encoder_group->start_timestamp = 0;
 
 		pthread_mutex_unlock(&encoder->encoder_group->mutex);
@@ -396,14 +395,15 @@ static void obs_encoder_actually_destroy(obs_encoder_t *encoder)
 		     encoder->context.name);
 
 		if (encoder->encoder_group) {
-			struct encoder_group *group = encoder->encoder_group;
+			struct obs_encoder_group *group =
+				encoder->encoder_group;
 			bool release = false;
 
 			encoder->encoder_group = NULL;
 
 			pthread_mutex_lock(&group->mutex);
-			group->encoders_added -= 1;
-			release = group->encoders_added == 0;
+			group->num_encoders -= 1;
+			release = group->num_encoders == 0;
 			pthread_mutex_unlock(&group->mutex);
 
 			if (release) {
@@ -1461,7 +1461,7 @@ static void receive_video(void *param, struct video_data *frame)
 	struct encoder_frame enc_frame;
 
 	if (encoder->encoder_group && !encoder->start_ts) {
-		struct encoder_group *group = encoder->encoder_group;
+		struct obs_encoder_group *group = encoder->encoder_group;
 		bool ready = false;
 		pthread_mutex_lock(&group->mutex);
 		ready = group->start_timestamp == frame->timestamp;
@@ -2093,7 +2093,8 @@ bool obs_encoder_group_keyframe_aligned_encoders(
 
 	bool unlock = false;
 	if (!encoder->encoder_group) {
-		encoder->encoder_group = bzalloc(sizeof(struct encoder_group));
+		encoder->encoder_group =
+			bzalloc(sizeof(struct obs_encoder_group));
 		if (pthread_mutex_init(&encoder->encoder_group->mutex, NULL) <
 		    0) {
 			bfree(encoder->encoder_group);
@@ -2101,11 +2102,11 @@ bool obs_encoder_group_keyframe_aligned_encoders(
 			return false;
 		}
 
-		encoder->encoder_group->encoders_added = 1;
+		encoder->encoder_group->num_encoders = 1;
 	} else {
 		pthread_mutex_lock(&encoder->encoder_group->mutex);
 		unlock = true;
-		if (encoder->encoder_group->encoders_started != 0) {
+		if (encoder->encoder_group->num_encoders_started != 0) {
 			blog(LOG_ERROR,
 			     "obs_encoder_group_keyframe_aligned_encoders: "
 			     "Can't add encoder '%s' to active group "
@@ -2117,7 +2118,7 @@ bool obs_encoder_group_keyframe_aligned_encoders(
 		}
 	}
 
-	encoder->encoder_group->encoders_added += 1;
+	encoder->encoder_group->num_encoders += 1;
 	encoder_to_be_grouped->encoder_group = encoder->encoder_group;
 
 	if (unlock)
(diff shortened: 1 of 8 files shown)


palana:
from what I gather, the newly exposed `obs_encoder_group` would behave like `obs_sceneitem`? not a fan of the `obs_sceneitem` API (see e.g. https://github.com/obsproject/obs-studio/commit/af03444cbe926d2d006fef84d156e0da0c3f11d3 and https://github.com/obsproject/obs-studio/commit/2fa208adfbe6861bc480eecbd6e0e569d6a5010a for an attempt to make that API "safer") but this moves this out of the twilight zone between implicit audio/video encoder pairing (that has no name) and an `obs_sceneitem` like API

I'm curious how removing encoders from a group that's being activated should work, my read of the current implementation is that you have to race `obs_output_start` internals to remove the encoder before it's active/initialized? removing an encoder from a fully active group (as in, encoders that received frames) wouldn't work because those encoders would be active/initialized?

libobs/obs-encoder.c
```diff
@@ -325,11 +325,10 @@ static void add_connection(struct obs_encoder *encoder)
...
 
 	if (encoder->encoder_group) {
-		bool ready = false;
 		pthread_mutex_lock(&encoder->encoder_group->mutex);
-		encoder->encoder_group->encoders_started += 1;
-		ready = encoder->encoder_group->encoders_started ==
-			encoder->encoder_group->encoders_added;
+		encoder->encoder_group->num_encoders_started += 1;
+		bool ready = encoder->encoder_group->num_encoders_started >=
+			     encoder->encoder_group->num_encoders;
```
palana, line 331: I don't have a test harness to check removing encoders from groups that are being activated, but if I'm not misreading things you can get encoder groups "stuck" via the following steps:
1. have multiple encoders grouped
2. start only a few of the grouped encoders, wait for all started encoders to be added to `num_encoders_started` so that their `bool ready` check is done
3. remove non-started encoders
in that scenario the `ready` check will never happen again, and thus `add_ready_encoder_group` will never be called

tt2468:
This PR should not allow encoders to be removed from groups which are in the "starting" phase. It should only allow them to be removed after they have been started. It's just a hypothetical situation but there's no reason it needs to be prevented since encoder groups are only relevant during the startup of encoders.

I acknowledge the potential dangers of having a touchable pointer to the encoder group without some kind of reference-counting, however it was a choice between "do it and explicitly document the major dangers" (what I've done) and "overengineer an API which is likely to never have much significant impact in the overall scope of things"

palana:
> This PR should not allow encoders to be removed from groups which are in the "starting" phase. It should only allow them to be removed after they have been started.

ah, ok; do you have a test harness somewhere to demonstrate this? my reading of this code (see also https://github.com/obsproject/obs-studio/pull/10781#discussion_r1631136621) is that either all encoders are started at ~"the same time" or the group gets stuck in the "starting" phase

> It's just a hypothetical situation but there's no reason it needs to be prevented since encoder groups are only relevant during the startup of encoders.

for the moment they're only relevant during the startup of encoders, however for DBR support I'm actually using encoder groups to synchronize bitrate changes across all encoders so that the IDR insertion (which generally happens when reconfiguring encoders) happens on the same frame

@derrod also has a proof of concept PR somewhere that uses encoder groups to resynchronize encoders for https://github.com/obsproject/obs-studio/pull/10096

tt2468:
I'm not super attached to the idea of allowing started encoder groups from having encoders removed from them, as I don't see much of a real-world use case for it, though I also don't quite see how the changes here would be mutually exclusive to those features.

One of my goals of fixing this up (along with a few other of my PRs) is to eventually have libobs automatically initialize all encoders in an encoder group if initialize is called on just one of the encoders. That way, we can never accidentally get into the position of having encoders which are perpetually not starting because other encoders in their group are not initialized.

palana:
> I'm not super attached to the idea of allowing started encoder groups from having encoders removed from them, as I don't see much of a real-world use case for it, though I also don't quite see how the changes here would be mutually exclusive to those features.

I don't have strong opinions on the idea of removing encoders from groups, and I agree that it's generally not mutually exclusive with the features I listed (as long as IDR alignment etc are preserved for the multitrack video output)

I am mildly curious about this implementation though, and where/how you can remove a video encoder from a started group; i.e. my current understanding of this code is that:
- either the group is started and thus all video encoders are started
- or the group is starting, so a some subset of encoders isn't started and can be removed (potentially leading to https://github.com/obsproject/obs-studio/pull/10781#pullrequestreview-2104432552)

> One of my goals of fixing this up (along with a few other of my PRs) is to eventually have libobs automatically initialize all encoders in an encoder group if initialize is called on just one of the encoders. That way, we can never accidentally get into the position of having encoders which are perpetually not starting because other encoders in their group are not initialized.

👍

tt2468:
After reading over my code again, I take back what I said about what is and isn't possible. I believe I am missing a `group->num_encoders_started && group->num_encoders != group->num_encoders_started` check in the `obs_encoder_remove_group()` function. Additionally, I am preventing a started/initialized encoder from being removed from the group. I will update the PR accordingly when I'm back at my desk.

One thing I am wondering about is what the expected behavior and limitations are for stopping an encoder group? If I stop one encoder from a started group, then start it again, what will happen?

palana:
> After reading over my code again, I take back what I said about what is and isn't possible. I believe I am missing a `group->num_encoders_started && group->num_encoders != group->num_encoders_started` check in the `obs_encoder_remove_group()` function. Additionally, I am preventing a started/initialized encoder from being removed from the group. I will update the PR accordingly when I'm back at my desk.

👍

> One thing I am wondering about is what the expected behavior and limitations are for stopping an encoder group? If I stop one encoder from a started group, then start it again, what will happen?

all encoders in an encoder group should be stopped and started at the same time; if just one encoder is stopped it won't get a new start timestamp when being restarted (this can probably be made to work with knowledge about b-frame delay/encoder internals, or by starting the encoder, throwing away the first few encoded frames and making all encoders go through a resync cycle)

derrod:
> @derrod also has a proof of concept PR somewhere that uses encoder groups to resynchronize encoders for #10096

Also updated and opened this upstream now for further discussion: https://github.com/obsproject/obs-studio/pull/10788

tt2468:
I took the day to fully brainstorm what a good API would be, and came to the conclusion that a very simple, non-ref-counted API is best. In this system, you create an empty encoder group object, assign encoders to it, and manually destroy the group when done with it. Destroying the group behaves like destroying an active encoder; it will wait for any active encoders to go inactive before fully destructing.

In the UI code, this is implemented by wrapping a newly created group with shared_ptr, to allow references to the group to be distributed. The group itself holds strong references to the encoders, meaning that holding the group will hold the encoders from being freed.

Part of how I settled on this is because normally, in order to keep an encoder object around, you need to store a strong reference to that encoder. In the case of multi-track video, you always will end up storing `n` number of encoder pointers manually, so a group object to hold all references is a handy shortcut in pretty much every case.

Finally, this design should make it relatively simple to expand the functionality to initializing/starting every encoder in the group when one member is initialized/started, fixing the hole described in a previous comment.

palana:
@tt2468 and I discussed this a bit on discord, from my last message (with some additional context added in square brackets):
> from a multitrack video output point of view, it [encoder group objects being externally visible, and keeping encoders alive] doesn't matter much; for twitch obs builds we're going to reintroduce manual tracking of all video encoders (<-> https://github.com/amazon-contributing/upstreaming-to-obs-studio/blob/74b7404648db8fdc53859d0753cb34eb700175bb/UI/multitrack-video-output.cpp#L1008-L1025) until we switch to bpm
>
> personally I don't like seeing another public api obs object with its own lifetime rules, but apart from being slightly worried about whether things will release in an orderly fashion (without some access being added [inside of libobs] at some point that clashes with that order), I'm ok with that change [i.e. making encoder group objects visible, and having them keep encoders alive through strong references]

kc5nra:
This looks fine to me - I understand your desire not to create an entirely different construct to manage the encoders (groups inside of encoders instead of groups containing encoders), it just feels a little bit awkward with the groups being intrusive to the encoder. But beyond that, seems okay!

tt2468:
@kc5nra Are you saying you would prefer to have them ref-counted rather than explicitly destroyed?

palana:
@tt2468: actually one thing that should probably be changed about the structure of this PR is that the API change commit causes the UI to not build; not sure how people want to see that squashed/whether people care that intermediate commits compile

RytoEX:
> @tt2468: actually one thing that should probably be changed about the structure of this PR is that the API change commit causes the UI to not build; not sure how people want to see that squashed/whether people care that intermediate commits compile

I care that intermediate commits compile because I don't want to be tripped up when bisecting. I may suggest that the commits are squashed, but will handle that once the changes themselves seem to have consensus.

tt2468:
I've squashed the API change commits to prevent issues during a potential bisect.

Lain-B (approved):

PR #10787, opened 2024-06-06
UI: Address logging buffer size discrepancies

jcm93:
### Description

Changes the remaining log buffer allocations in the log handler to use a buffer of size 8192 instead of 4096. Also adds improved string handling in the form of using `strncpy` rather than `strcpy` in the `too_many_repeated_entries()` check.

### Motivation and Context

OBS was crashing on startup when logging messages over 4096 characters in length with corruption of the `QPointer` object for the log viewer. It turns out that this was due to a buffer overrun in the log handling code.

https://github.com/obsproject/obs-studio/commit/43a1b309940b7bb7a92fdcd019bd011ef762621f changed the size of the log buffer in obs-app.cpp's `do_log`, but to change the size of the buffer properly we need to also modify the buffer size in the subsequent check for repeated entries, as well as the base log handler in `libobs`. Otherwise the repeated entries check will bulldoze over memory adjacent to `cmp_str` if fed a message exceeding 4096 characters in length, as well as other memory in the base log handler.

### How Has This Been Tested?

Verified that OBS no longer crashes when failing to load VLC on startup, and logging continues to work correctly.

### Types of changes

### Checklist:

### DIFF
--- a/UI/obs-app.cpp
+++ b/UI/obs-app.cpp
@@ -350,7 +350,7 @@ static inline bool too_many_repeated_entries(fstream &logFile, const char *msg,
 	static mutex log_mutex;
 	static const char *last_msg_ptr = nullptr;
 	static int last_char_sum = 0;
-	static char cmp_str[4096];
+	static char cmp_str[8192];
 	static int rep_count = 0;
 
 	int new_sum = sum_chars(output_str);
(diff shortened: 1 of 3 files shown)


RytoEX:
cc @palana who authored https://github.com/obsproject/obs-studio/commit/43a1b309940b7bb7a92fdcd019bd011ef762621f

derrod:
I've been chasing too many random crashes that seem to be all caused by this so I'm just gonna merge this to preserve what remains of my sanity while working based on current master.

RytoEX:
> looks good to me 👍 the only suggestion I'd have beyond this is to move the buffer size into a constant somewhere to avoid this situation in the future (although another change in buffer size seems unlikely)

Agree with this as a welcome future change.

PR #10788, opened 2024-06-07
Encoder Group Resynchronisation

derrod:
### Description

Adds a mechanism to resynchronise grouped encoders if keyframes are no longer aligned to the same PTS interval.

Resynchronisation works by determining a future PTS all encoders are expected to hit, and adding a flag to that frame instructing the encoder to force an IDR frame, thus starting a new GOP and bringing encoders back in-sync.

This PR adds a mechanism to adds the following on top of #10096:
- `OBS_ENCODER_CAP_FORCE_IDR` to indicate that an encoder supports forcing IDR frames
- Adds `obs_frame_flags` to `encoder_frame`/`encoder_texture`
    + Only `OBS_FRAME_FORCE_IDR` exists for now
- Repurposes the existing check for frame misalignments to determine a future resync point
- Implements forcing IDR frames for x264 and NVENC

### Motivation and Context

With the goal of #10096 (included here) being to allow encoders to skip frames to avoid the death spiral after GPU stalls while being close to the maximum throughput, we need a means of resyncing encoders if frames are skipped.

While this PR only implements this feature for x264 and NVENC, all the other relevant encoders (QSV, AMF, VideoToolbox) also support forcing/requesting IDR frames.

### How Has This Been Tested?

Has been tested in one form or another during my personal test streams going back to January.

### Types of changes

### Checklist:

### DIFF
--- a/libobs/obs-encoder.c
+++ b/libobs/obs-encoder.c
@@ -1495,6 +1495,14 @@ static void receive_video(void *param, struct video_data *frame)
 	enc_frame.frames = 1;
 	enc_frame.pts = encoder->cur_pts;
 
+	if (encoder->encoder_group &&
+	    encoder->encoder_group->force_idr_pts == encoder->cur_pts &&
+	    encoder->info.caps & OBS_ENCODER_CAP_FORCE_IDR) {
+		blog(LOG_DEBUG, "Forcing IDR on encoder \"%s\"",
+		     obs_encoder_get_name(encoder));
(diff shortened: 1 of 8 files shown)


Warchamp7:
Is this still something you think we should give some attention to @derrod?

derrod:
> Is this still something you think we should give some attention to @derrod?

Not sure, it hinges on the ability to drop frames if an encoder is overloaded. Which - historically - we didn't want to do.

Perhaps close it and then we'll see about it for whatever rework might happen w/t/t the data pipeline.

PR #10789, opened 2024-06-07
UI: Simplify multitrack encoder availability check

derrod:
### Description

Simplifies the check for whether or not an encoder is available.

### Motivation and Context

Much more readable and avoids the creation of a pointless `std::vector`.

### How Has This Been Tested?

Locally.

### Types of changes

### Checklist:

### DIFF
--- a/UI/multitrack-video-output.cpp
+++ b/UI/multitrack-video-output.cpp
@@ -259,30 +259,16 @@ static void adjust_encoder_frame_rate_divisor(
 	obs_encoder_set_frame_rate_divisor(video_encoder, divisor);
 }
 
-static const std::vector<const char *> &get_available_encoders()
-{
-	// encoders are currently only registered during startup, so keeping
-	// a static vector around shouldn't be a problem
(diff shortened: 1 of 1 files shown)


derrod:
I have no idea why it was written this, perhaps this was originally intended to use a hash set for O(1) lookups? If one wanted to do that it could be done fairly simply:

```c++
static bool encoder_available(const std::string_view &type)
{
    static const auto available_encoders = [] {
        std::unordered_set<std::string_view> available_encoders;

        const char *id = nullptr;
        for (size_t i = 0; obs_enum_encoder_types(i, &id); i++)
            available_encoders.insert(id);

        return available_encoders;
    }();

    return !!available_encoders.count(type);
}
```

RytoEX:
Perhaps @palana can shed light on the rationale behind the current implementation. This change seems fine at a glance, and we can always change it further later if we have a better understanding of the rationale/needs in the future.

RytoEX (approved):

PR #10795, opened 2024-06-07
UI: Fix multitrack stream key query parameter concatenation

derrod:
### Description

Fixes query parameters being incorrectly concatenated with the stream key if it already contains query paramters itself (e.g. `?bandwidthtest=true`).

### Motivation and Context

This should work and it doesn't.

### How Has This Been Tested?

Verified final stream key is now correctly assembled.

### Types of changes

### Checklist:

### DIFF
--- a/UI/multitrack-video-output.cpp
+++ b/UI/multitrack-video-output.cpp
@@ -114,8 +114,16 @@ create_service(const GoLiveApi::Config &go_live_config,
 				    str->len - (found - str->array));
 	}
 
+	/* The stream key itself may contain query parameters, such as
+	 * "bandwidthtest" that need to be carried over. */
+	QUrl parsed_key{stream_key};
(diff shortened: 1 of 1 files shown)


UI/multitrack-video-output.cpp
```diff
@@ -114,23 +114,29 @@ create_service(const GoLiveApi::Config &go_live_config,
 				    str->len - (found - str->array));
 	}
 
+	/* The stream key itself may contain query parameters, such as
+	 * "bandwidthtestmode" that need to be carried over. */
```
RytoEX, line 118:
```suggestion
	 * "bandwidthtest" that need to be carried over. */
```

RytoEX (changes requested):
Nit. Looks okay at a glance otherwise.

RytoEX (approved):
Did a cursory check. Seems fine.

PR #10798, opened 2024-06-08
UI: Use signal vector for advanced audio dialog

cg2121:
### Description
Simplifies code by using a vector instead of individual signals.

### Motivation and Context
Code cleanup

### How Has This Been Tested?
Made sure the advanced audio dialog still worked as expected.

### Types of changes

### Checklist:

### DIFF
--- a/UI/window-basic-adv-audio.cpp
+++ b/UI/window-basic-adv-audio.cpp
@@ -12,18 +12,16 @@ Q_DECLARE_METATYPE(OBSSource);
 OBSBasicAdvAudio::OBSBasicAdvAudio(QWidget *parent)
 	: QDialog(parent),
 	  ui(new Ui::OBSAdvAudio),
-	  sourceAddedSignal(obs_get_signal_handler(), "source_audio_activate",
-			    OBSSourceAdded, this),
-	  sourceRemovedSignal(obs_get_signal_handler(),
-			      "source_audio_deactivate", OBSSourceRemoved,
-			      this),
-	  sourceActivatedSignal(obs_get_signal_handler(), "source_activate",
-				OBSSourceActivated, this),
-	  sourceDeactivatedSignal(obs_get_signal_handler(), "source_deactivate",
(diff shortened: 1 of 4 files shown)


UI/window-basic-adv-audio.cpp
```diff
@@ -47,6 +45,8 @@ OBSBasicAdvAudio::~OBSBasicAdvAudio()
 	for (size_t i = 0; i < controls.size(); ++i)
 		delete controls[i];
 
+	sigs.clear();
```
Lain-B, line 48: Hm, is this necessary?

cg2121: It has now been removed

UI/window-basic-adv-audio.cpp
```diff
@@ -158,35 +156,25 @@ void OBSBasicAdvAudio::SetShowInactive(bool show)
 
 	showInactive = show;
 
-	sourceAddedSignal.Disconnect();
-	sourceRemovedSignal.Disconnect();
-	sourceActivatedSignal.Disconnect();
-	sourceDeactivatedSignal.Disconnect();
+	sigs.clear();
```
Lain-B, line 159: I just noticed this will clear all signals, even the ones that are only initialized in constructors. I would probably either split it off into two separate vectors and make sure to only clear relevant signals here, or maybe just close the PR.

Need to be careful for things like this. Unless I'm missing something.

Lain-B (changes requested):

UI/window-basic-adv-audio.cpp
cg2121: These are the same signals that are created in the constructor.

Lain-B: Ah I see, I misread.

Lain-B (approved):

cg2121:
There is some code cleanup that can be done here, so it is not as confusing. That will be future PR.

PR #10806, opened 2024-06-08
UI: Connect replay buffer stopping to signal instead of slot

gxalpha:
### Description

Fixes a typo that resulted in a wrong connection, where the controls dock slot would attempt to connect to the main window slot instead of signal, which of course doesn't work.
To the user, this would mean that the replay buffer button text wouldn't change to "Stopping" when pressing "Stop", instead staying on "Stop" while stopping and then going to "Start" directly.

30.2 regression.

### Motivation and Context

Saw obsproject/loganalyzer#163 that excluded an `OBSBasic`, which looked suspicious. No log should contain `OBSBasic`.
Turns out that Qt logged a warning that a signal couldn't be found. Looking into that further , this makes sense since the signal that the slot was connected to wasn't a signal.

### How Has This Been Tested?

macOS 14.5
Firstly, made sure the bug indeed exists (see observation in Description).
Tested that with this patch, the button does show "Stopping".

### Types of changes

### Checklist:

### DIFF
--- a/UI/basic-controls.cpp
+++ b/UI/basic-controls.cpp
@@ -117,7 +117,7 @@ OBSBasicControls::OBSBasicControls(OBSBasic *main)
 
 	connect(main, &OBSBasic::ReplayBufStarted, this,
 		&OBSBasicControls::ReplayBufferStarted);
-	connect(main, &OBSBasic::ReplayBufferStopping, this,
+	connect(main, &OBSBasic::ReplayBufStopping, this,
 		&OBSBasicControls::ReplayBufferStopping);
 	connect(main, &OBSBasic::ReplayBufStopped, this,
 		&OBSBasicControls::ReplayBufferStopped);


tytan652 (approved):
LGTM, my fault with autocomplete.

RytoEX (changes requested):
The change itself is fine, but the commit message could use some clarification.

WizardCM (approved):

Lain-B (approved):
Approved, but yea, I agree with Ryan, probably needs a better commit name

derrod:
I think it was meant to say "UI: Connect replay buffer stopping to signal instead of **slot**"

gxalpha:
> I think it was meant to say "UI: Connect replay buffer stopping to signal instead of slot"

Yes. Fixed.

RytoEX (approved):

Issue #10799, opened 2024-06-08
Beta version crashes on Win11 24h2 26100 release preview


Ptilopsis01:
### Operating System Info

Windows 11

### OBS Studio Version

30.2.0-beta1

### OBS Studio Log URL

cannot get it

### Expected Behavior

OBS starts correctly

### Current Behavior

[Crash 2024-06-08 16-04-59.txt](https://github.com/user-attachments/files/15747265/Crash.2024-06-08.16-04-59.txt)
It crashes just when I start the application. 

### Steps to Reproduce

1. Install the latest beta ver.
2. Start OBS
3. OBS crashes

maginhart:
same
[Crash 2024-06-08 13-33-23.txt](https://github.com/user-attachments/files/15747645/Crash.2024-06-08.13-33-23.txt)

derrod:
Ismyour Windows set to a non-English language?

maginhart:
In my case no, windows is set to English

notr1ch:
Can you try updating your VC redist? https://aka.ms/vs/17/release/vc_redist.x64.exe

knot2006:
> Can you try updating your VC redist? https://aka.ms/vs/17/release/vc_redist.x64.exe

Thank you! This fixed crashing for me.

maginhart:
Worked in my case too.

Ptilopsis01:
Works for me, thx a lot!

RytoEX:
As of OBS Studio 30.2.0 Beta 3, this should be resolved as much as we can manage now.

PR #10809, opened 2024-06-09
UI: Check VC++ Runtime version on startup

derrod:
### Description

Adds a check on startup that ensures a compatible Visual C++ runtime is being used.

The user is given the option to automatically open our existing VC runtime download page (https://obsproject.com/visual-studio-2019-runtimes),

### Motivation and Context

Due to backwards-incompatible changes in MSVC 14.40 OBS crashes on startup for some users that do not have up to date runtimes installed. See #10799.

We should also adjust the updater and installer to check for these older versions, but that will be a larger undertaking and also does not cover the use case of portable ZIPs (unlike this PR).

### How Has This Been Tested?

Tested on Windows VM with outdated vc runtime:

warning

### Types of changes

### Checklist:

### DIFF
--- a/UI/obs-app.cpp
+++ b/UI/obs-app.cpp
@@ -57,6 +57,7 @@
 #ifdef _WIN32
 #include <windows.h>
 #include <filesystem>
+#include <util/windows/win-version.h>
 #else
 #include <signal.h>
 #include <pthread.h>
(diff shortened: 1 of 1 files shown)


UI/obs-app.cpp
```diff
@@ -2901,9 +2902,41 @@ void OBSApp::commitData(QSessionManager &manager)
...
+}
+#endif
+
 int main(int argc, char *argv[])
 {
-#ifndef _WIN32
+#ifdef _WIN32
+	if (is_outdated_msvc())
+		return 1;
+#elif defined(_WIN32)
```
notr1ch, line 2939:
```suggestion
#else
```

derrod: Whoops, fixed.

UI/obs-app.cpp
```diff
@@ -2901,6 +2902,35 @@ void OBSApp::commitData(QSessionManager &manager)
 }
 #endif
 
+#ifdef _WIN32
+static constexpr char vcRunErrorTitle[] = "Outdated Visual C++ Run Version";
```
RytoEX, line 2906:
```suggestion
static constexpr char vcRunErrorTitle[] = "Outdated Visual C++ Runtime Version";
```

RytoEX (changes requested):

notr1ch:
Perhaps the wording could be changed to "You will now be directed to the download page." with OK / Cancel buttons. Currently it's unclear that pressing "No" will exit OBS.

derrod:
I updated the wording (not the screenshot though). I also wasn't super happy with it, but I didn't have any better ideas. Perhaps @Warchamp7 has a suggestion as well.

Either way I'd like to have this for beta2 (alongside the updater/installer changes) so we can mitigate the VC runtime mess as much as possible.

RytoEX (approved):
Maybe there's a better way to handle this, but if we figure that out, we can make further changes later. Let's mitigate this as much as we can.

wang-bin:
I my project I find out the std::mutex is incompatible with old msvcp140.dll, I tested with the simplest mutext code. Adding `-D_DISABLE_CONSTEXPR_MUTEX_CONSTRUCTOR` can fix the incompatibility

RytoEX:
> I my project I find out the std::mutex is incompatible with old msvcp140.dll, I tested with the simplest mutext code. Adding `-D_DISABLE_CONSTEXPR_MUTEX_CONSTRUCTOR` can fix the incompatibility

We considered this and decided against it.

PR #10817, opened 2024-06-09
updater: Check and update VC runtime version

derrod:
### Description

Updates the updater to check the VC runtime version and install the latest version if appropiate.

Detailed changes:
- Minor version of `msvcp140` is checked to be >= 40
- VC installer is downloaded from MS permalink instead of OBS mirror
- VC installer digital signature is verified before execution
- Updater is again compiled with `/MT`, otherwise it would also crash due to the same problem that OBS itself breaks (`std::mutex` changes)

### Motivation and Context

Companion to #10809.

### How Has This Been Tested?

Tested on VM and Windows Sandbox.

### Types of changes

### Checklist:


UI/win-update/updater/cmake/legacy.cmake
```diff
@@ -36,7 +36,16 @@ if(MSVC)
...
+target_link_libraries(
+  updater
+  PRIVATE OBS::blake2
+          nlohmann_json::nlohmann_json
+          zstd::libzstd_static
+          comctl32
+          shell32
+          winhttp
+          version
+          winhttp
```
RytoEX, line 48:
```suggestion
          version
          winhttp
```
`winhttp` is here twice.

RytoEX (changes requested):

RytoEX (approved):
Looks okay at a glance. Would appreciate a second review.

PR #10823, opened 2024-06-10
UI: Translate Multitrack Video error dialog buttons

RytoEX:
### Description

UI: Translate Multitrack Video error dialog buttons

The Yes and No standard buttons are not translated unless we manually set the translated text ourselves.

### Motivation and Context

A user reported to us in OBS Studio 30.2.0-beta1 that these buttons were not translated. We already use this pattern elsewhere in the codebase.
https://github.com/obsproject/obs-studio/blob/0680b642e9cb85101971e8e03bbcd9b574cff6f7/UI/ui-validation.cpp#L47-L49

### How Has This Been Tested?

I've built this locally on Windows 11, but I could not seem to get this specific dialog to pop up.

### Types of changes

### Checklist:

### DIFF
--- a/UI/multitrack-video-error.cpp
+++ b/UI/multitrack-video-error.cpp
@@ -1,6 +1,7 @@
 #include "multitrack-video-error.hpp"
 
 #include <QMessageBox>
+#include <QPushButton>
 #include "obs-app.hpp"
 
 MultitrackVideoError MultitrackVideoError::critical(QString error)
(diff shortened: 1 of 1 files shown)


RytoEX:
For further cleanup, we could probably move `QMessageBox mb` into the first `if` branch, and change the second branch to use `OBSMessageBox::critical` instead. We don't have an OBSMessageBox function that uses the Warning icon that uses Yes/No buttons, or I would suggest we use that. We could add one, but that was more work than I wanted to do this moment.

UI/multitrack-video-error.cpp
```diff
@@ -31,9 +32,14 @@ bool MultitrackVideoError::ShowDialog(
...
-				      QMessageBox::StandardButton::No);
-		return mb.exec() == QMessageBox::StandardButton::Yes;
+		QAbstractButton *yesButton =
+			mb.addButton(QTStr("Yes"), QMessageBox::YesRole);
+		mb.addButton(QTStr("No"), QMessageBox::NoRole);
+
+		if (mb.clickedButton() != yesButton)
+			return false;
+		else
+			return true;
```
derrod, line 42: Just use `return mb.clickedButton() == yesButton;`

RytoEX: Done.

derrod (changes requested):
Based on the documentation you'd still need to call `exec`, otherwise thid will always return null: https://doc.qt.io/qt-6/qmessagebox.html#clickedButton

RytoEX:
> Based on the documentation you'd still need to call `exec`, otherwise thid will always return null: https://doc.qt.io/qt-6/qmessagebox.html#clickedButton

Whoops. Fixed.

derrod (approved):
Seems fine but unable to test right now

gxalpha (approved):

PR #10827, opened 2024-06-11
obs-webrtc: Default H264 profile to main if unset

### LINKED ISSUE #10821: obs-webrtc should default to main profile
### Operating System Info

Windows 11

### Other OS

_No response_

### OBS Studio Version

30.2.0-beta1

_No response_

### OBS Studio Log URL

N/A

### OBS Studio Crash Log URL

_No response_

### Expected Behavior

Multiple users when testing OBS WebRTC reported playback errors. I was having a hard time reproducing them until I realized that the H264 profile was unset. It was defaulting to High and was causing playback issues. Browsers sometimes use OpenH264 which doesn't support H264

### Current Behavior

Video playback with OBS + H264 by default isn't reliable

### Steps to Reproduce

1. Stream using WebRTC and don't change any encoder settings
2. Video playback will be erratic/unreliable

_No response_

Sean-Der:
### Description
If user requests a specific H264 profile respect it, otherwise default to main.

### Motivation and Context
OBS defaults to H264 settings that aren't supported by browsers. This requires all WebRTC users to set advanced settings by default. 

This is a frequent debugging topic by users on discord. I had 3 different users who were inconvenienced because of this behavior. 

The new beta release of OBS was also tested by QA at Twitch. The 'WebRTC tests' were marked as failed because of the current behavior. We had to re-run the tests and ask to explicitly set the H264 profile.

### How Has This Been Tested?
I have tested locally with X264, QSV, nvenc and AOM AV1 on Windows.

On discord 3 users have been testing. Will link ticket so they can provide feedback.

### Types of changes

### Checklist:

Resolves #10821

### DIFF
--- a/plugins/obs-webrtc/whip-service.cpp
+++ b/plugins/obs-webrtc/whip-service.cpp
@@ -34,6 +34,14 @@ void WHIPService::ApplyEncoderSettings(obs_data_t *video_settings, obs_data_t *)
 	if (video_settings) {
 		obs_data_set_int(video_settings, "bf", 0);
 		obs_data_set_bool(video_settings, "repeat_headers", true);
+
+		const char *profile =
+			obs_data_get_string(video_settings, "profile");
+		if (profile == nullptr || profile[0] == '\0') {
+			obs_data_set_string(video_settings, "profile", "main");
+			blog(LOG_INFO,
+			     "[obs-webrtc] no H264 profile set, default to main");
+		}
 	}
 }
 


Sean-Der:
I ended up fixing this in Chromium itself!

OBS would crash hardware decoders. Chromium would incorrectly not fall back to software. This change would prevent the initial crash, but I don't think it makes sense anymore. The experience is better, and this will just be baggage we forget to delete.

PR #10831, opened 2024-06-11
updater: Fix defaultlib conflict

RytoEX:
### Description

Trying to build in Debug fails due a default lib conflict:
> LNK4098: defaultlib 'MSVCRTD' conflicts with use of other libs; use /NODEFAULTLIB:library

Instead, we can set the updater to always use /MT instead of /MTd.

### Motivation and Context

Want to be able to build OBS in Debug.

### How Has This Been Tested?

Tested locally on Windows 11 with VS2022 17.10.1.

### Types of changes

### Checklist:

### DIFF
--- a/UI/win-update/updater/CMakeLists.txt
+++ b/UI/win-update/updater/CMakeLists.txt
@@ -44,4 +44,4 @@ set_target_properties(
   updater
   PROPERTIES FOLDER frontend
              OUTPUT_NAME updater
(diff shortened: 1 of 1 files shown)


derrod (approved):
Let's revisit this when we're not trying to put out fires quickly.

Since nobody but myself has ever needed to debug the updater in the last few years I can live with this.

PR #10834, opened 2024-06-12
UI: Fix themeDir buffer being resized incorrectly

derrod:
### Description

Fixes `themeDir` string being too small for holding the config directory path.

### Motivation and Context

c677bac875e6b3b6e4d3358aa8b38679ff4fcb1f changed the order here, but this also resulted in the string having whatever size was necessary for the install data path, rather than being large enough to fit a userdata path. To fix this, move the resize operaetion *after* the buit-in themes are searched, and also bump it to 1024 just to be sure.

This resulted in a crash due to a bug in os_get_path_internal() which will need to be fixed separately.

### How Has This Been Tested?

Booted up OBS without a crash.

### Types of changes

### Checklist:

### DIFF
--- a/UI/obs-app-theming.cpp
+++ b/UI/obs-app-theming.cpp
@@ -417,7 +417,6 @@ static vector<OBSThemeVariable> ParseThemeVariables(const char *themeData)
 void OBSApp::FindThemes()
 {
 	string themeDir;
-	themeDir.resize(512);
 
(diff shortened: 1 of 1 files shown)


RytoEX (approved):
Can confirm this fixes the crash when running OBS normally (not in Portable Mode) on Windows. This is "good enough" for now.

PR #10835, opened 2024-06-12
libobs: Fix buffer overrun in os_get_path_internal()

derrod:
### Description

Additional fix for problem mitigated by #10834

### Motivation and Context

The function did not subtract the size already written from the size passed into `strcat_s()`, thus resulting in it assuming it can write `size` bytes, when in reality it was `size - os_wcs_to_utf8()` bytes, resulting in a buffer overflow.

### How Has This Been Tested?

Tested without #10834 applied to verify the crash no longer happens, and instead the function (correctly) returns -1 to indicate a failure.

### Types of changes

### Checklist:

### DIFF
--- a/libobs/util/platform-windows.c
+++ b/libobs/util/platform-windows.c
@@ -494,12 +494,15 @@ static int os_get_path_internal(char *dst, size_t size, const char *name,
 
 	SHGetFolderPathW(NULL, folder, NULL, SHGFP_TYPE_CURRENT, path_utf16);
 
-	if (os_wcs_to_utf8(path_utf16, 0, dst, size) != 0) {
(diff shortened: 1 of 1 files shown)


derrod:
Wait no, I can't read the documentation. Why is this overrunning?

derrod:
So turns out Microsoft's version of these functions will invoke the ["invalid parameter handler"](https://learn.microsoft.com/en-us/cpp/c-runtime-library/parameter-validation?view=msvc-170) instead of just retuning with the error code. Since we do not set one, this is the default Microsoft "watson" handler, which aborts the execution of the program. Which is just wonderful. I'm surprised we didn't run into this sooner tbh.

PR #10838, opened 2024-06-12
UI: Use unique_ptr for theme objects

derrod:
### Description

Always delete the `theme` pointer to avoid memory leaks.

### Motivation and Context

While the `std::move` will (probably) result in the object's contained members being moved the copy constructed within the hash map, the original object should be deleted as it's no longer needed.

### How Has This Been Tested?

Verified with a debugger that everything works correctly and values get moved.

### Types of changes

### Checklist:

### DIFF
--- a/UI/obs-app-theming.cpp
+++ b/UI/obs-app-theming.cpp
@@ -417,6 +417,7 @@ static vector<OBSThemeVariable> ParseThemeVariables(const char *themeData)
 void OBSApp::FindThemes()
 {
 	string themeDir;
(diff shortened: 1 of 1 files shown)


RytoEX:
If it can always be deleted, would it make more sense to use `std::unique_ptr` so we don't have to manually delete, or would that not work here?

derrod:
Using a `std::unique_ptr` would also work, yes.

RytoEX (approved):
Seems okay to me.

PR #10839, opened 2024-06-12
UI: Return std::optional from ParseThemeMeta

derrod:
### Description

Use `std::optional` as a return value instead of a pointer, thus avoiding any manual memory management shenanigans.

### Motivation and Context

Want to avoid errors like the one fixed by #10838 

This function could probably throw an exception to signal the error, but it currently doesn't and I don't want to make too many changes.

### How Has This Been Tested?

Verified themes still work.

### Types of changes

### Checklist:

### DIFF
--- a/UI/obs-app-theming.cpp
+++ b/UI/obs-app-theming.cpp
@@ -44,42 +44,42 @@ struct CFParser {
 	cf_parser *operator->() { return &cfp; }
 };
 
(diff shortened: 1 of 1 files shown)


Lain-B (approved):

RytoEX (approved):
LGTM

PR #10800, opened 2024-06-08
obs-scripting: Fix script plugin path on Linux

tytan652:
### Description

Fix the fact `obslua.so` is not found with CMake 3.

```
warning: [Lua: instant-replay.lua] Error executing startup script 1: [string "for val in pairs(package.preload) do..."]:5: module 'obslua' not found:
	no field package.preload['obslua']
	no file './obslua.lua'
	no file '/usr/share/luajit-2.1/obslua.lua'
	no file '/usr/local/share/lua/5.1/obslua.lua'
	no file '/usr/local/share/lua/5.1/obslua/init.lua'
	no file '/usr/share/lua/5.1/obslua.lua'
	no file '/usr/share/lua/5.1/obslua/init.lua'
	no file './obslua.so'
	no file '/usr/local/lib/lua/5.1/obslua.so'
	no file '/usr/lib/lua/5.1/obslua.so'
	no file '/usr/local/lib/lua/5.1/loadall.so'
	no file './/obslua.so'
	no file 'lib/obs-scripting/obslua.so'
```

### Motivation and Context

Fix a bug found by a beta-tester.

### How Has This Been Tested?

Loading any Lua script (e.g. `instant-replay.lua`).

### Types of changes

- Bug fix (non-breaking change which fixes an issue)

### Checklist:

- [x] My code has been run through [clang-format](https://github.com/obsproject/obs-studio/blob/master/.clang-format).
- [x] I have read the [**contributing** document](https://github.com/obsproject/obs-studio/blob/master/CONTRIBUTING.rst).
- [x] My code is not on the master branch.
- [x] The code has been tested.
- [x] All commit messages are properly formatted and commits squashed where appropriate.
- [x] I have included updates to all appropriate documentation.

PR #10844, opened 2024-06-13
Fix loading Lua scripts on Linux through a refactor of how C path are added

tytan652:
### Description

Try to fix the same issue as #10800 but more correctly, as the latter does not work on Linux system install.

This PR reverts #10800 and refactor how C paths are added to enable adding more than two paths.

### Motivation and Context

Fix loading Lua script on Linux.
 If you have a better approach (taking CMake 2 and 3 in mind), please feel free to supersede this PR.

### How Has This Been Tested?

Lua script loads with OBS install system-wide and relocatable.

### Types of changes

### Checklist:

### DIFF
--- a/cmake/linux/defaults.cmake
+++ b/cmake/linux/defaults.cmake
@@ -31,7 +31,7 @@ set(OBS_CMAKE_DESTINATION "${CMAKE_INSTALL_LIBDIR}/cmake")
 
 # Set additional paths used by OBS for self-discovery
 set(OBS_PLUGIN_PATH "${CMAKE_INSTALL_LIBDIR}/obs-plugins")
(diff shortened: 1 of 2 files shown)


PatTheMav (commented):
Does this affect Windows as well? If so this might need some quick testing on Win as well.

tytan652:
> Does this affect Windows as well?

If you mean the issue for loading the script plugin, I don't think so since nobody reported it on Windows and CMake 3 on Windows is a thing since a while.

Code-wise, the same paths are added to Windows and macOS with or without the PR.
Linux needed a more specific one for CMake 3.

RytoEX (approved):
I've tested this on Windows against my Lua test script in Portable and non-Portable mode and it seems to continue to work on Windows.

As far as I can tell from reviewing the code, this looks okay.

PatTheMav:
Sounds good to me then.

PR #10856, opened 2024-06-14
UI: Set Multitrack Video maximum bitrate limit default to 0/Auto

RytoEX:
### Description

UI: Set Multitrack Video maximum bitrate limit default to 0/Auto

There were some reports that the default value of 8000 was confusing because it implies that the automaticaly selected and used value is 8000 Kbps. Set it to 0 which should hopefully make it more obvious that OBS is not sending 0 Kbps.

Note that this also requires lowering the _minimum_ permissible value of the field from 500 to 0.

### Motivation and Context

Want the UI to be less confusing.

### How Has This Been Tested?

Tested locally on Windows 11.

### Types of changes

### Checklist:

### DIFF
--- a/UI/forms/OBSBasicSettings.ui
+++ b/UI/forms/OBSBasicSettings.ui
@@ -1674,13 +1674,13 @@
                         <string notr="true"> Kbps</string>
                        </property>
                        <property name="minimum">
-                        <number>500</number>
+                        <number>0</number>
                        </property>
                        <property name="maximum">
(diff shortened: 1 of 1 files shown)


derrod (approved):
Should end various discussions and is consistent with e.g. keyframe interval using 0 is "auto" or default.

tt2468 (approved):

Lain-B (approved):

PR #10863, opened 2024-06-16
UI: Add step value to Multitrack Video maximum bitrate control

RytoEX:
### Description

UI: Add step value to Multitrack Video maximum bitrate control

All other bitrate setting fields in the UI have a single-step value of 50. Without setting this, the single-step value is 1, which makes using the scroll wheel a bit tedious.

While this could arguably be higher, let's make this consistent first and then consider changing the values across the application later.

### Motivation and Context

Want consistency with other bitrate UI controls. Want scrolling the field to be less painful.

### How Has This Been Tested?

Tested on Windows 11.

### Types of changes

### Checklist:

### DIFF
--- a/UI/forms/OBSBasicSettings.ui
+++ b/UI/forms/OBSBasicSettings.ui
@@ -1679,6 +1679,9 @@
                        <property name="maximum">
                         <number>1000000</number>
                        </property>
+                       <property name="singleStep">
+                        <number>50</number>
+                       </property>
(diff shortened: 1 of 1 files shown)


derrod (approved):
This is consistent with other bitrate spinboxes.

We could probably bump all of the to 100 or 250 at some point, this kind of precision (for video) is probably unnecessary.

PR #10864, opened 2024-06-16
UI: Fix capitalization of 'OBS' and 'RTMP'

Vainock:
### Description
"obs" -> "OBS"
"rtmp" -> "RTMP"

### Motivation and Context
Consistency and correctness is good!

### How Has This Been Tested?
/

### Types of changes

### Checklist:


RytoEX:
If I recall correctly, these are both describing specific types:
* rtmp_output
* obs_video_info

I think the messages they were based on are similar, though those might have only been logged and not translated. It may be the case that we'd want to correct these for use in the UI, though I originally opted not to correct them. Will defer to @Warchamp7 .

Warchamp7:
If we're presenting this error to the user, it should be presented appropriately with correct capitalization.

RytoEX:
> If we're presenting this error to the user, it should be presented appropriately with correct capitalization.

Technically, the "correct capitalization" is:
* [rtmp_output](https://github.com/obsproject/obs-studio/blob/9d88b632ae0ad720093be22eecac37f8d1d78843/UI/multitrack-video-output.cpp#L156)
* [obs_video_info](https://github.com/obsproject/obs-studio/blob/9d88b632ae0ad720093be22eecac37f8d1d78843/UI/multitrack-video-output.cpp#L293-L296)

Because they are object/variable types. Otherwise, if we want to assert that "descriptive text should not be technical/literal", then yes, this is probably better for user-facing dialog/UI text.

RytoEX (approved):
Seems fine.

Issue #10744, opened 2024-05-28
Stats card not showing the encoded framecount


TheMasterofBlubb:
### Operating System Info

Windows 10

### OBS Studio Version

30.1.2

### OBS Studio Log URL

https://obsproject.com/logs/nRRRBOYYN0byXOkf

### Expected Behavior

When starting a stream or recording the stats card should show the correct amount of frames encoded (and skipped for that matter)

### Current Behavior

When setting a Output rescale through the Settings->Output->Rescale Output the stats card will always show 0/0 frames for the encoding stats
![grafik](https://github.com/obsproject/obs-studio/assets/12793716/9a386c9e-c250-46cf-9ead-d295000a7671)
This is reardless of the Docked or Standalone version.

### Steps to Reproduce

1. Check that Output rescale is set to off
2. Start recording and observe Stats
3. Stop recording
4. Set any Output rescale option (mine was 1664x936 with Lanc)
5. Start recording
6. observe that stats are not working properly
7. on reverse of that setting stats work again, no restart required

TheMasterofBlubb:
Possible commit that created that issue https://github.com/obsproject/obs-studio/commit/6cdfb0a8b99fd22a3f3cc01baacc9b01fa9ec2dd

WizardCM:
> Possible commit that created that issue [6cdfb0a](https://github.com/obsproject/obs-studio/commit/6cdfb0a8b99fd22a3f3cc01baacc9b01fa9ec2dd)

Is this from a bisect, or an assumption?

TheMasterofBlubb:
Assumption due to commit-history-timeline.

PR #10866, opened 2024-06-17
libobs: Fix bug with frame count in stat card

deltamaya:
Now the frame count in statcard will display correctly with rescale output  turned on

### Description

Increase both encoder media and the obs main mix video 's frame count in gpu_encode_thread

### Motivation and Context

fixes issue: https://github.com/obsproject/obs-studio/issues/10744

### How Has This Been Tested?

The encoded & lagged frame count works correctly with a simple scene

### Types of changes

Bug fix

### Checklist:

### DIFF
--- a/libobs/obs-video-gpu-encode.c
+++ b/libobs/obs-video-gpu-encode.c
@@ -63,6 +63,9 @@ static void *gpu_encode_thread(void *data)
 		next_key = tf.lock_key;
 
 		video_output_inc_texture_frames(video->video);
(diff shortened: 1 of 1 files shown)


derrod:
This really isn't fixing the issue, but creating a bunch of new ones. For example, when using the Atium Vertical plugin which uses an separate view its encoder and output stats would be counted towards the main mix's video, which is incorrect.

Really what needs fixing here is the stats window itself, as it needs to get the video object from the output's encoder, rather than relying on the main mix itself.

PR #10867, opened 2024-06-17
UI: Fix stats window

deltamaya:
Now the frame count in stats window will display correctly with rescale output turned on
I think this time it makes more sense?

### Description

Use the encoder's video frame count instead of the obs main mix video

### Motivation and Context

Fixes issue: https://github.com/obsproject/obs-studio/issues/10744

### How Has This Been Tested?

Tested with simple scene when streaming and recording

### Types of changes

Bug fix

### Checklist:


norihiro (commented):
The commit description has a too-long line. It should be wrapped at 72-column.

PR #10874, opened 2024-06-18
UI: Display dialog for multitrack video output audio channels mismatch

palana:
### Description
Since https://github.com/obsproject/obs-studio/pull/10855 is too big, this is a version that displays an error message prompting the user to switch to a supported audio channel layout

### Motivation and Context
from https://github.com/obsproject/obs-studio/pull/10855:
> The Twitch iOS app (and/or HLS playback on iOS in general?) requires audio to be stereo or mono

### How Has This Been Tested?
set audio channels to 2.1/4.1 and verified that both clicking cancel and "update settings and start streaming" worked

### Types of changes

### Checklist:


derrod:
Normally we require a restart for chaing audio settings like channels or sample rate, is that not actually required?

RytoEX:
When changing Settings -> Audio -> Channels from Stereo to 2.1, for example:
image

RytoEX:
Just acknowledging that I've seen this. This diff doesn't seem much smaller than #10855, and I'm not sure what our preference for handling this is.

derrod:
> Just acknowledging that I've seen this. This diff doesn't seem much smaller than #10855, and I'm not sure what our preference for handling this is.

I personally would've kept it simpler and simply errored out with a warning rather than handling the changing of the setting semi-automatically, but I'd rather have this in the UI than libobs changes at this point that do not strictly fix bugs.

palana:
@RytoEX, @derrod: split out the bits that change global audio settings, the first commit now only displays a warning message (and a cancel button)

RytoEX (approved):

PR #10877, opened 2024-06-18
frontend-tools: Avoid initializing the scene switcher on Wayland

tytan652:
### Description

The Linux implementation of the automatic scene switcher is X11-only and the design itself of the feature is incompatible with how Wayland works.

### Motivation and Context

If a user happen to have the scene switcher active, OBS will crash while freeing the tool under Wayland.

The saved configuration should be kept since we no longer create the scene collection JSON from scratch, so there is normally no loss.

### How Has This Been Tested?

The scene switcher is no longer loaded on Wayland.

### Types of changes

### Checklist:

### DIFF
--- a/UI/frontend-plugins/frontend-tools/auto-scene-switcher.cpp
+++ b/UI/frontend-plugins/frontend-tools/auto-scene-switcher.cpp
@@ -513,6 +513,11 @@ static void OBSEvent(enum obs_frontend_event event, void *)
 
 extern "C" void InitSceneSwitcher()
 {
+#if !defined(__APPLE__) && !defined(_WIN32)
+	if (QApplication::platformName().contains("wayland"))
+		return;
+#endif
+
 	QAction *action = (QAction *)obs_frontend_add_tools_menu_qaction(
 		obs_module_text("SceneSwitcher"));
 


Fenrirthviti:
For posterity, is this possible to fix on Wayland, or is there a fundamental missing functionality that prevents this from ever working? I understand disabling it fixes an immediate problem/crash, and a solution might take longer, but it's not clear if this is something that can be fixed or not.

If we're disabling features, I'd like to at least have an issue open (or discussion post) on what steps are required to re-enable it back to feature parity, if possible. High-level is fine, it does not need to be an exact technical roadmap.

tytan652:
> For posterity, is this possible to fix on Wayland, or is there a fundamental missing functionality that prevents this from ever working? I understand disabling it fixes an immediate problem/crash, and a solution might take longer, but it's not clear if this is something that can be fixed or not.

Under Wayland, application can't be aware of other application windows for multiple reason (e.g. security, avoid re-inventing X11).
Reminder, for Screen Capture we use portals so at no point OBS is aware of which application is selected.

This feature will never be possible on Wayland.

Fenrirthviti:
I don't quite agree with "never be possible" here, but it sounds like if Wayland offers a secure API at some point to provide window information, we can revisit this.

As it stands, Wayland doesn't provide the information that this features needs to function, so I'm fine disabling it there.

GeorgesStavracas:
This kind of interaction (listing other windows and accessing their state) isn't available in any existing Wayland protocols, and I think this is by design. Given the security requirements, this looks like something that should be provided by a portal.

Right now nothing exists in the Wayland / Portals platform that can enable this feature, so I think disabling it for the time being is all that OBS can do.

RytoEX:
Just to confirm, this happens on 30.1.x and older, correct?

GeorgesStavracas:
Yes, it happens in all current and previous releases of OBS Studio.

RytoEX:
> Yes, it happens in all current and previous releases of OBS Studio.

Okay then. Based on that information, I'll set this aside until we get into the merge window for the next release. Thanks.

Lain-B (approved):

RytoEX (approved):
LGTM

PR #10880, opened 2024-06-18
win-capture: Improve fullscreen window detection

Kira-NT:
### Description

This small change allows windows that extend slightly beyond the defined viewport to be properly detected as fullscreen ones.

While, in general, one should not assume that `bottom > top` and `right > left`, `win-capture` is tailored to Windows, where this assumption always holds true. Therefore, simply swapping `==` with `>=` when comparing `rect.right` with `mi.rcMonitor.right` and `rect.bottom` with `mi.rcMonitor.bottom` should suffice.

### Motivation and Context

While it may sound crazy, some apps on Windows use a dirty hack of extending their windows slightly beyond the monitor's declared resolution, usually by 1px, to preserve "true" windowed fullscreen and circumvent some issues related to the exclusive fullscreen mode presented by the graphics library being used, Windows itself, or both. As an example, see: https://github.com/glfw/glfw/issues/527#issuecomment-143183771.

At the moment, these windows are not recognized by OBS Game Capture because they fail the strict comparison rules defined here: https://github.com/obsproject/obs-studio/blob/48b1298fafb941f70edfa3b17bd2441e9124c3c9/plugins/win-capture/game-capture.c#L1221-L1228

And this is quite annoying because, from the user's perspective, these windows are clearly in the fullscreen mode.

Additionally, Windows itself doesn't resort to such strict comparison rules, as evidenced by the fact that the taskbar disappears, as it always does for fullscreen windows, even when a window stretches slightly beyond the defined viewport. So, if anything, this should be considered a parity feature.

### How Has This Been Tested?

There isn't much to test, so I just ran the freshly built OBS on Windows in a VM to check if it works. Thankfully, a two-symbol change didn't break the project :D

The change is very localized, shouldn't affect other areas of the code, and doesn't introduce any breaking changes. It simply fixes a minor issue/tweaks the pre-existing functionality to better match the, ahem, nuances of Windows development.

### Types of changes

### Checklist:
- [x] I have included updates to all appropriate documentation. *(I.e., none)*

### DIFF
--- a/plugins/win-capture/game-capture.c
+++ b/plugins/win-capture/game-capture.c
@@ -1131,7 +1131,15 @@ static void get_fullscreen_window(struct game_capture *gc)
 		return;
 	}
 
-	if (rect.left == mi.rcMonitor.left && rect.right == mi.rcMonitor.right && rect.bottom == mi.rcMonitor.bottom &&
+	/* some apps extend their windows ~1px beyond the monitor's
+	 * supported resolution to circumvent certain exclusive
+	 * fullscreen-related issues caused either by their graphics
+	 * library, Windows itself, or both; thus, we use a strict
+	 * comparison to check the coordinates of the top-left corner
+	 * of a window against those of a monitor, and a more relaxed
+	 * one to check the bottom-right corner, allowing it to extend
+	 * slightly past the edge of the monitor */
+	if (rect.left == mi.rcMonitor.left && rect.right >= mi.rcMonitor.right && rect.bottom >= mi.rcMonitor.bottom &&
 	    rect.top == mi.rcMonitor.top) {
 		setup_window(gc, window);
 	} else {
--- a/plugins/win-capture/game-capture.c
+++ b/plugins/win-capture/game-capture.c
@@ -1104,6 +1104,12 @@ static void setup_window(struct game_capture *gc, HWND window)
 	}
 }
 
+static int rects_nearly_equal(RECT first, RECT second, long delta)
+{
+	return abs(first.left - second.left) <= delta && abs(first.right - second.right) <= delta &&
+	       abs(first.top - second.top) <= delta && abs(first.bottom - second.bottom) <= delta;
+}
+
 static void get_fullscreen_window(struct game_capture *gc)
 {
 	HWND window = GetForegroundWindow();
@@ -1137,8 +1143,11 @@ static void get_fullscreen_window(struct game_capture *gc)
 		return;
 	}
 
-	if (rect.left == mi.rcMonitor.left && rect.right == mi.rcMonitor.right && rect.bottom == mi.rcMonitor.bottom &&
-	    rect.top == mi.rcMonitor.top) {
+	/* Some apps extend their windows 1-2 pixels beyond the monitor's supported resolution to circumvent certain
+	 * exclusive fullscreen-related issues caused either by their graphics library, Windows itself, or both. Thus,
+	 * we compare the window's corner coordinates with those of the monitor using an acceptable relative margin of
+	 * 2, allowing it to extend slightly past the edge of the monitor. */
+	if (rects_nearly_equal(rect, mi.rcMonitor, 2)) {
 		setup_window(gc, window);
 	} else {
 		gc->wait_for_target_startup = true;


Lain-B (approved):

derrod:
Please do not use merge commits, instead, rebase your PRs.

Kir-Antipov:
Apologies! It's been so long that I forgot what the contributing guidelines here look like, so I did the simplest thing possible - I resolved the conflict using GitHub's interface, since it was literally a one-line conflict, assuming the merge commit could be squashed away anyways :)

I've reordered the history, so the PR now consists of just one commit once again.

Warchamp7:
@Kira-NT Can you share any other examples of apps that exhibit this behaviour?

Kira-NT:
@Warchamp7, nope, not really. This is a Windows-specific issue, and since I don't use that OS, I can't really provide an extensive list of specific apps affected by this problem.

However, it's a widely known and commonly suggested fix for the unwanted behavior of "borderless" fullscreen windows created via OpenGL on Windows, which tend to blink every time they gain or lose focus. Here are a few examples of it being suggested in the wild, where the earliest one I found via a quick search dates back almost 12 years:

 - https://stackoverflow.com/questions/23145217/flickering-when-borderless-window-and-desktop-dimensions-are-the-same
 - https://discourse.libsdl.org/t/true-borderless-fullscreen-behaviour/24622/11
 - https://discourse.glfw.org/t/cannot-manage-to-create-borderless-windowed-form/1592
 - https://github.com/raysan5/raylib/issues/2764
 - https://github.com/pygame/pygame/issues/3619
 - https://github.com/ocornut/imgui/issues/6179

So this junky solution to a junky problem on a junky OS has been around for quite a while.

Personally, I'm interested in fixing OBS's window-detection logic, because I have a mod for Minecraft that implements borderless fullscreen for the game. On Windows, one of the approaches it uses is this "true windowed" method. Given that the mod currently has [16 million downloads on Modrinth](https://modrinth.com/mod/cubes-without-borders) and [8 million on CurseForge](https://www.curseforge.com/minecraft/mc-mods/cubes-without-borders), and considering that people often record their Minecraft gameplay, I'd say quite a few users are now exposed to this issue. Because of me, sure *(whoopsie :3)*, but still. I really don't see any downsides to adjusting a single line of code to accommodate the bogus realities of Windows development otherwise.

EZ64cool:
Helldivers 2 is currently running 1 pixel bigger than my monitor in borderless fullscreen mode, I've had to add a specific game-capture source for the exe otherwise it isn't picked up.

plugins/win-capture/game-capture.c
Warchamp7, line 1100:
```suggestion
static int rects_nearly_equal(RECT first, RECT second)
{
	return abs(first.left - second.left) <= 2 && abs(first.right - second.right) <= 2 &&
	       abs(first.top - second.top) <= 2 && abs(first.bottom - second.bottom) <= 2;
}
```

plugins/win-capture/game-capture.c
```diff
@@ -1131,7 +1131,15 @@ static void get_fullscreen_window(struct game_capture *gc)
...
+	/* some apps extend their windows ~1px beyond the monitor's
+	 * supported resolution to circumvent certain exclusive
+	 * fullscreen-related issues caused either by their graphics
+	 * library, Windows itself, or both; thus, we use a strict
+	 * comparison to check the coordinates of the top-left corner
+	 * of a window against those of a monitor, and a more relaxed
+	 * one to check the bottom-right corner, allowing it to extend
+	 * slightly past the edge of the monitor */
+	if (rect.left == mi.rcMonitor.left && rect.right >= mi.rcMonitor.right && rect.bottom >= mi.rcMonitor.bottom &&
 	    rect.top == mi.rcMonitor.top) {
```
Warchamp7, line 1143:
```suggestion
if (rects_nearly_equal(rect, mi.rcMonitor)) {
```

plugins/win-capture/game-capture.c
```diff
@@ -1131,7 +1131,15 @@ static void get_fullscreen_window(struct game_capture *gc)
...
 
-	if (rect.left == mi.rcMonitor.left && rect.right == mi.rcMonitor.right && rect.bottom == mi.rcMonitor.bottom &&
+	/* some apps extend their windows ~1px beyond the monitor's
+	 * supported resolution to circumvent certain exclusive
+	 * fullscreen-related issues caused either by their graphics
+	 * library, Windows itself, or both; thus, we use a strict
+	 * comparison to check the coordinates of the top-left corner
+	 * of a window against those of a monitor, and a more relaxed
+	 * one to check the bottom-right corner, allowing it to extend
+	 * slightly past the edge of the monitor */
```
Warchamp7, line 1141: Comment needs to be adjusted for the suggested changes.

Warchamp7 (changes requested):
Tested with Helldivers 2
_(Which was annoying since its anticheat appears to prevent even opening a debugger with the game running)_

ShareX in screenshot mode will show the window as being 2560x1440 on my system, but a Print Screen results in a screenshot that it is 2562x1440.

Logging out the `GetWindowRect` from OBS shows the window rect as being -1, 2561, -1, 1441, which means it is reported by the Windows API as being 2562x1442.

This means that the solution in this PR of checking the top left corner explicitly would still fail, so I've proposed a +/- 2 px check on each corner instead.

I've suggested a 2px tolerance instead of 1px because testing with Slay the Spire 2 showed it also being 2562 wide, but the left of both rects is 0, and it is the right that is off by 2.

Warchamp7:
@Kira-NT This is currently on our board for 33.0 but awaiting changes.

Kira-NT:
Thanks for the ping! Not sure how I managed to miss the previous one.

I've rebased the PR against the current master and adjusted the code as per your suggestions. I did add the acceptable delta as a parameter, though, instead of hardcoding it, so it's a bit easier to change in case in the future someone stumbles upon a game that does the same trick but with a 4px offset, for example, because after all my time researching this topic, I've never come across anyone suggesting a 2px offset instead of the 1px one *(although it does kinda make sense if you want to keep the window size a multiple of 2)*, so who knows what else might be out there in the wild.

RytoEX (approved):

Warchamp7 (approved):

PR #10887, opened 2024-06-19
UI: Only cache multitrack config URL startup argument

palana:
### Description
Allow usage of different multitrack config URLs between different stream attempts without restarting the program

### Motivation and Context
This is an artifact from when development on multitrack video started, before services could supply their config URLs through services.json

This currently works when switching between Twitch and Amazon IVS due to some infrastructure being shared, but won't work if switching between Twitch/Amazon IVS and a completely different service

### How Has This Been Tested?
Switched between Twitch and Amazon IVS and verified that they use their own service URLs:
```
13:27:18.508: Go live URL: https://ingest.contribute.live-video.net/api/v3/GetClientConfiguration
[…]
13:27:38.260: Go live URL: https://ingest.twitch.tv/api/v3/GetClientConfiguration
```

### Types of changes

### Checklist:

### DIFF
--- a/UI/goliveapi-network.cpp
+++ b/UI/goliveapi-network.cpp
@@ -128,7 +128,8 @@ GoLiveApi::Config DownloadGoLiveConfig(QWidget *parent, QString url,
 
 QString MultitrackVideoAutoConfigURL(obs_service_t *service)
 {
-	static const QString url = [service]() -> QString {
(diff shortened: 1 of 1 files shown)


derrod (approved):
I don't think you need to explicitly call `.has_value()` in an if expression but it's fine.

RytoEX (approved):

PR #10888, opened 2024-06-19
UI: Disable multitrack video settings on non-win32 platforms

palana:
### Description
Disables multitrack video output settings for platforms other than windows

### Motivation and Context
System info is not implemented for platforms other than windows, so the overall sentiment is/was to disable the multitrack video output for other platforms until system info is implemented (and ideally encoders for those other platforms are validated and/or encoder ladders/configs are available)

### How Has This Been Tested?
Verified that these settings are available on windows builds, but aren't available on MacOS builds

### Types of changes

### Checklist:

### DIFF
--- a/UI/window-basic-settings.cpp
+++ b/UI/window-basic-settings.cpp
@@ -6325,6 +6325,10 @@ void OBSBasicSettings::UpdateMultitrackVideo()
 			ui->enableMultitrackVideo->setChecked(false);
 	}
 
+#ifndef _WIN32
+	available = available && MultitrackVideoDeveloperModeEnabled();
+#endif
+
 	if (IsCustomService())
 		available = available && MultitrackVideoDeveloperModeEnabled();
 


UI/window-basic-settings.cpp
```diff
@@ -6325,6 +6325,10 @@ void OBSBasicSettings::UpdateMultitrackVideo()
 			ui->enableMultitrackVideo->setChecked(false);
 	}
 
+#ifndef _WIN32
+	available = available && MultitrackVideoDeveloperModeEnabled();
```
palana, line 6329: there's an argument to be made about changing this to
```suggestion
	available = IsCustomService() && MultitrackVideoDeveloperModeEnabled();
```
instead, but I think it's still potentially useful as is in some cases (e.g. someone developing their own configuration service and testing hitting that endpoint from linux?)

derrod (approved):
Seems fine as-is.

UI/window-basic-settings.cpp
RytoEX: As @derrod said, this seems fine as-is.

PR #10890, opened 2024-06-19
UI: Fix crash when exiting with a multitrack video stream active

palana:
Not entirely sure if this is a good fix/the right fix, commit title etc is somewhat not final cc @tt2468, @RytoEX 

### Description
changes order that multitrack video output encoders and outputs are released in, to work around a crash on shutdown while the multitrack video output is active

### Motivation and Context
exiting obs while the multitrack video output is active will currently cause obs to crash:
code around the crash https://github.com/obsproject/obs-studio/blob/608d3bfc26b5ff88ea6467d0183826c344bb1bd5/libobs/obs-encoder.c#L358-L360
-> `remove_connection` calls `obs_encoder_group_actually_destroy` which in turn releases the last strong ref to the current encoder so encoder is an invalid pointer in `remove_connection` after that point, and it currently crashes in `pthread_mutex_unlock(&encoder->encoder_group->mutex)` for me

### How Has This Been Tested?
exited obs with multitrack video output active and it no longer crashes

### Types of changes

### Checklist:

### DIFF
--- a/UI/multitrack-video-output.hpp
+++ b/UI/multitrack-video-output.hpp
@@ -52,9 +52,13 @@ public:
 
 private:
 	struct OBSOutputObjects {
-		OBSOutputAutoRelease output_;
 		std::shared_ptr<obs_encoder_group_t> video_encoder_group_;
 		std::vector<OBSEncoderAutoRelease> audio_encoders_;
+		/* Outputs do not hold on to their encoders,
+		 * so the output needs to be stopped/released
+		 * before the encoders are released
+		 */
+		OBSOutputAutoRelease output_;
 		OBSServiceAutoRelease multitrack_video_service_;
 		OBSSignal start_signal, stop_signal, deactivate_signal;
 	};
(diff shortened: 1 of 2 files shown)


UI/multitrack-video-output.hpp
```diff
@@ -52,9 +52,13 @@ struct MultitrackVideoOutput {
...
 private:
 	struct OBSOutputObjects {
-		OBSOutputAutoRelease output_;
 		std::shared_ptr<obs_encoder_group_t> video_encoder_group_;
 		std::vector<OBSEncoderAutoRelease> audio_encoders_;
+		/* Outputs do not hold on to their encoders,
+		 * so the output needs to be stopped/released
+		 * before the encoders are released
+		 */
+		OBSOutputAutoRelease output_;
```
tt2468, line 61: I'm not really a fan of how this relies on implicit c++ delete behavior to keep libobs from crashing. I'm looking into if it's possible to solve the issue from the libobs side.

PR #10892, opened 2024-06-19
UI: Adjust Yami (Classic) styling

Warchamp7:
### Description
Many of the changes to Yami had some unintended side effects to the Classic variant that were not properly accounted for before.

### Motivation and Context
Trying to match the old Dark theme more closely

### How Has This Been Tested?
Before:
image

After:
image

Before:
image

After:
image

### Types of changes

### Checklist:

### DIFF
--- a/UI/data/themes/Yami.obt
+++ b/UI/data/themes/Yami.obt
@@ -477,7 +477,7 @@ OBSDock > QWidget {
 }
 
 #transitionsFrame {
-    padding: 4px 8px;
+    padding: var(--padding_large);
 }
 
 OBSDock QLabel {
@@ -827,7 +827,7 @@ QComboBox::drop-down,
 QDateTimeEdit::drop-down {
     border:none;
     border-left: 1px solid var(--grey6);
-    width: 32px;
+    width: var(--input_height);
 }
 
 QComboBox::down-arrow,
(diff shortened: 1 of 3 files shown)


derrod (changes requested):
The "active" colour for buttons still is wrong

controls dock

Warchamp7:
> The "active" colour for buttons still is wrong
> 
> controls dock

That was more or less a deliberate change

derrod:
It does look pretty out of place

Warchamp7:
> It does look pretty out of place

It's using the 'Primary' color of the theme, same as the Scene and Source lists (Which also had a minor visual behaviour change from stable). It's a more consistent use of 'Selected' or Active

derrod:
None of the other buttons use that colour when pressed and it looks out of place imo.

Warchamp7:
> None of the other buttons use that colour when pressed and it looks out of place imo.

It's the checked state versus pressed, it's meant to be different

derrod:
The blue still looks weird though, I'd rather just have the old dark grey.

RytoEX:
> The blue still looks weird though, I'd rather just have the old dark grey.

It looks like the blue is part of beta4, so this is not an issue with this PR per se, but rather something else that you would also like to see changed. Is that correct? If so, I don't want to block this PR on additional changes we'd like to see, and I'd like to get this merged to fix the issues that it _does_ fix.

RytoEX (approved):

PR #10895, opened 2024-06-19
mac-avcapture: Use list-based format selector

jcm93:
### Description

Updates the properties window for macOS `mac-avcapture` sources to use a unified list-based format selector when in the non-preset mode. Also adds a contextual warning in the properties window to show when a device has macOS system effects active.

###
Screenshot 2024-07-10 at 9 39 50 AMScreenshot 2025-03-17 at 1 58 31 PM
Bildschirmfoto_2025-03-17_um_19 35 50

Bildschirmfoto_2025-03-17_um_19 36 27

###

The frame rate selector is retained, so that we can continue to have explicit control over the device frame rate. The list selectors for resolution, color space, pixel format and video range, however, have all been consolidated inside the single format selector.

### Motivation and Context

https://github.com/obsproject/obs-studio/pull/9344 provided a much-needed update to the Video Capture Device source on macOS and added a lower-latency Capture Card Device source. A number of factors, however, have conspired to make the new version of the source harder to use, for both novice and adept users:

- The new version does not reimplement the 'auto-populate parameters' feature of the source on other platforms. When a property is updated, the user must check all other properties to ensure that the combination of selected parameters is still a valid configuration, which is not always immediately apparent.
- Because of "quirks" with OBS's frame rate selector and the default size for the source properties window, it is not apparent to many users that the list selectors for pixel format, video range and color space exist 'below the fold' in the source properties window.

This PR aims to simplify and improve the UI for the source while also reducing the complexity of the property selection code.
- All parameters are now visible in the source properties window at its default size. The only parameter that can be invalid after changing the selected format is the frame rate, and its status should be more apparent within the context of the window.
- It is easier to understand at a glance what combinations of parameters are valid for a given device. Making an appropriate selection should now be a quicker process.

Additionally, the new effects warning in the properties window should help to alleviate some user confusion. On macOS, enabling an effect such as Studio Light or Portrait mode for one video input will often enable them on all inputs (filed as FB13302833). Users are often confused to see webcam effects on e.g. capture card inputs. With this PR, the source properties window displays a warning for active system effects, which should make this issue more apparent.

### How Has This Been Tested?

Tested locally on an M1 Max machine running the macOS Sequoia beta with all video capture devices on hand, including the built-in webcam, Continuity Camera from an iPhone, a virtual camera, and a single capture card from a popular brand, to ensure both OBS source types remained in good working order.

### Types of changes

### Checklist:


plugins/mac-avcapture/OBSAVCapture.m
```diff
@@ -700,21 +628,67 @@ + (NSString *)stringFromSettings:(void *)settings withSetting:(NSString *)settin
     return result;
 }
 
++ (const char *)effectsWarningForDevice:(AVCaptureDevice *)device
```
jcm93, line 631: I was wondering if this should return an `NSString *` instead, and we wait to convert it to a C string until after we call this from `plugin-properties.m`. That would allow us to remove the `<obs-module.h>` import above.

Also curious if there is a good way to check compile-time SDK availability, since the background replacement code that I've commented out here will only compile with the macOS 15 SDK.

PatTheMav: I was about to suggest that - try to stay within Foundation and Cocoa as much as possible and only convert to pure C types at the boundary of the `libobs` API functions.

plugins/mac-avcapture/plugin-main.m
```diff
@@ -79,42 +79,44 @@ static void av_fast_capture_set_defaults(obs_data_t *settings)
...
-                                                          OBS_COMBO_TYPE_LIST, OBS_COMBO_FORMAT_INT);
-    obs_property_t *video_range = obs_properties_add_list(properties, "video_range", obs_module_text("VideoRange"),
-                                                          OBS_COMBO_TYPE_LIST, OBS_COMBO_FORMAT_INT);
 
     if (capture_info) {
         bool isFastPath = capture_info->isFastPath;
 
         // Add Property Visibility and Callbacks
         configure_property(device_list, true, true, properties_changed, capture);
+        //todo fix
```
PatTheMav, line 108: What's the TODO here?

plugins/mac-avcapture/OBSAVCapture.h
```diff
@@ -336,6 +350,50 @@ typedef struct av_capture_info {
 
 #pragma mark - Static helper functions
 
+/// Compare two `AVCaptureDeviceFormat`s for purposes of sorting in the properties window.
+static NSComparator const OBSAVCaptureDeviceFormatCompare =
+    ^NSComparisonResult(AVCaptureDeviceFormat *a, AVCaptureDeviceFormat *b) {
+        CMVideoDimensions aDimensions = CMVideoFormatDescriptionGetDimensions(a.formatDescription);
```
PatTheMav, line 356: Have you tried calculating a "bandwidth" for each format (which would necessarily include resolution, number of planes, size per plane, colour depth, etc.) so you get a nice single numerical value representing a format and then you can just compare the value of `A` and `B`.

Also the default names for custom comparators in Swift is `lhs` and `rhs`, which I'd prefer here as well.

PatTheMav (commented):
Left some code-specific comments, overall I'd like to see more discussion about:

1. Presentation of the formats in the dropdown (@Warchamp7 got some ideas there?)
2. How to "upgrade" existing source configurations

plugins/mac-avcapture/OBSAVCapture.m
```diff
@@ -700,21 +628,67 @@ + (NSString *)stringFromSettings:(void *)settings withSetting:(NSString *)settin
...
+    }
+    */
+    /* Not available until we are building on Xcode 16. Is there a way to check SDK availability?
+    if (@available(macOS 15.0, *)) {
+        if (device.backgroundReplacementActive) {
+            effectWarning = obs_module_text("BackgroundReplacementWarning");
+            effectsCount++;
+        }
+    }
+     */
```
gxalpha, line 668: > Also curious if there is a good way to check compile-time SDK availability, since the background replacement code that I've commented out here will only compile with the macOS 15 SDK.

@jcm93 like this:
```c
#if __MAC_OS_X_VERSION_MAX_ALLOWED >= 150000
...
#endif
```

At a later point we could raise the minimum version but since Xcode 16 is still in Beta 1 I feel like that's to early, the current minimum is this:
https://github.com/obsproject/obs-studio/blob/5f98d34e2c4d81a43eafd3d87ae65e67a3d04e06/cmake/macos/compilerconfig.cmake#L26-L27


On GHA we're on `macos-14` with macOS 14.5 and Xcode 15.2 (see the [docs](https://github.com/actions/runner-images/blob/main/images/macos/macos-14-arm64-Readme.md), default is 15.0.1 but we [change](https://github.com/obsproject/obs-studio/blob/5f98d34e2c4d81a43eafd3d87ae65e67a3d04e06/.github/workflows/build-project.yaml#L84) that), so it wouldn't be compiled anyways. GHA does provide Xcode 16 though if we're happy to compile on a Beta versions on CI, historically we've done that a couple of times.

jcm93: Thanks; added this check to the background replacement block.

plugins/mac-avcapture/plugin-main.m
jcm93: Just left over from something fixed and forgot to remove the todo. This part has been rewritten now also with the new effects warning return type.

plugins/mac-avcapture/OBSAVCapture.m
jcm93: Changed, import removed.

plugins/mac-avcapture/data/locale/en-US.ini
```diff
@@ -15,3 +15,9 @@ Auto="Auto"
 Unknown="Unknown (%1)"
 EnableAudio="Enable audio if supported by device"
 Resolution="Resolution"
+PortraitEffectWarning="Portrait system effect is active on the selected device"
+CenterStageWarning="Center Stage system effect is active on the selected device"
+StudioLightWarning="Studio Light system effect is active on the selected device"
+ReactionsWarning="Reaction effects are active for the selected device"
+BackgroundReplacementWarning="Background replacement is active for the selected device"
+MultipleEffectsWarning="Multiple system effects are active on the selected device"
```
gxalpha, line 23: Personally I'd prefer keys like `Warning.PortraitEffect` etc for better readability. Additionally, some contain "Effect" and others don't, is that intentional?

plugins/mac-avcapture/OBSAVCapture.h
jcm93: > Have you tried calculating a "bandwidth" for each format (which would necessarily include resolution, number of planes, size per plane, colour depth, etc.) so you get a nice single numerical value representing a format and then you can just compare the value of A and B.

This seems like a nice idea, I'll just have to look into the specifics. Adjusted argument names for now.

plugins/mac-avcapture/data/locale/en-US.ini
jcm93: Agree; adjusted the naming and the descriptions to be more uniform.

plugins/mac-avcapture/OBSAVCapture.h
jcm93: Just a thought here so I don't forget; the sort order may also want to cleanly delineate aspect ratio in some way at the top level; i.e. all 16:9 formats separated from all 4:3 formats, etc. for any portrait or square formats that might be in there.

jcm93: Pushed an update to the sort comparator; it now sorts by aspect ratio, then computed bandwidth, then color space, which seems like a better ordering.

Bandwidth is computed a bit lazily, relying partially on a simple bitsPerPixel block with me providing the values. Since this will only used inside this sort comparator and it mostly just wants to be fast, it seemed okay. Not sure what other formats I may want to include values for, though.

As I'm thinking about this more, I'm wondering if, after aspect ratio, we may actually want to sort by (resolution * fps) first, and then bits per pixel second, possibly in reverse order. Reason being, If we're listing all the formats available at a particular FPS and resolution, from there the "best" option (i.e., the one that most users should pick) may be the one with the lowest bandwidth rather than the highest.

PatTheMav: So the way you'd normally do this is to create a custom category for `AVCaptureDeviceFormat` (say `AVCaptureDeviceFormat+Comparable.m`) and extend the original class with a lazily computed property and add `- (NSComparisonResult)compare:(MyClass *)object` as the comparator.

That way you'd encapsulate the calculation of the comparison value within the format class and also provide the canonical `compare` function implicitly used by Foundation.

jcm93: Switched it around to compute/store a bandwidth scalar and sort using a selector instead of the NSComparator.

Thoughts now:

1. We could just store a generic "sort order" value instead of a 'bandwidth scalar' value. Since the sort order doesn't seem like it should be strictly/only bandwidth-based, and that way we could avoid computing aspect ratio in the selector. If we did that we could maybe even make a sort descriptor and sort solely on that computed property.
2. still same reverse-bandwidth thoughts as above

Also let me know if you see any other issues.

jcm93:
Sidebar that I just thought of; now that the scaffolding is in place to store computed properties in an AVCaptureDeviceFormat+Comparable category, The format's list description (given by `[OBSAVCapture descriptionFromFormat:]` presently) could also be a lazily computed property in the same way, which would definitely be more efficient...

plugins/mac-avcapture/OBSAVCapture.h
jcm93: Got on a bit of a roll so went ahead and revamped sorting with my idea so it could be tested. Refactored to use sort descriptors on top of the computed properties.

Currently sorts by aspect ratio (rounded to two places) descending, then 'pixels per second' (i.e. resolution * fps) descending, then 'bits per pixel' *ascending*, with the idea that you would present less bandwidth-intense formats first for the same pixel throughput. Couple screenshots beneath; certainly needs more testing.

Screenshot 2024-07-10 at 1 04 07 AM

Screenshot 2024-07-10 at 1 14 52 AM

plugins/mac-avcapture/OBSAVCapture.m
```diff
@@ -475,34 +419,64 @@ - (BOOL)configureSession:(NSError *__autoreleasing *)error
 
     CMTime time = {.value = fps.denominator, .timescale = fps.numerator, .flags = 1};
 
-    AVCaptureDeviceFormat *format = nil;
-
-    for (AVCaptureDeviceFormat *formatCandidate in [self.deviceInput.device.formats reverseObjectEnumerator]) {
-        CMVideoDimensions formatDimensions = CMVideoFormatDescriptionGetDimensions(formatCandidate.formatDescription);
+    const char *selectedFormat = obs_data_get_string(self.captureInfo->settings, "supported_format");
+    NSString *selectedFormatNSString = [NSString stringWithCString:selectedFormat encoding:NSUTF8StringEncoding];
```
PatTheMav, line 423: In theory you should be able to simply box the character array pointer here via `@(selectedFormat)`, but you need to explicitly check against a possible NULL pointer before boxing.

plugins/mac-avcapture/OBSAVCapture.m
```diff
@@ -475,34 +419,64 @@ - (BOOL)configureSession:(NSError *__autoreleasing *)error
...
-            continue;
-        }
+    AVCaptureDeviceFormat *format = nil;
+    FourCharCode subtype;
+    OBSAVCaptureColorSpace colorSpace;
+    bool fpsSupported = false;
 
-        for (AVFrameRateRange *range in formatCandidate.videoSupportedFrameRateRanges) {
-            if (CMTimeCompare(range.maxFrameDuration, time) >= 0 && CMTimeCompare(range.minFrameDuration, time) <= 0) {
+    if (![selectedFormatNSString isEqual:@""]) {
```
PatTheMav, line 430: If you know that both objects are `NSString`, it is at least measurably faster to call `isEqualToString`.

plugins/mac-avcapture/plugin-properties.m
```diff
@@ -135,7 +136,7 @@ bool properties_changed_use_preset(OBSAVCapture *capture, obs_properties_t *prop
         properties_changed_preset(capture, properties, preset_list, settings);
     }
 
-    const char *update_properties[5] = {"resolution", "frame_rate", "color_space", "video_range", "input_format"};
+    const char *update_properties[2] = {"frame_rate", "supported_format"};
```
PatTheMav, line 139: This is fine, but given that it's only 2 properties whose state needs to be changed it might be almost simpler to directly call `obs_property_set_<thing>` directly without the loop.

plugins/mac-avcapture/OBSAVCapture.m
jcm93: Fixed.

plugins/mac-avcapture/OBSAVCapture.m
jcm93: Changed.

plugins/mac-avcapture/OBSAVCapture.m
```diff
@@ -700,21 +688,66 @@ + (NSString *)stringFromSettings:(void *)settings withSetting:(NSString *)settin
...
+        }
+    }
+    if (@available(macOS 13.0, *)) {
+        if (device.studioLightActive) {
+            effectWarning = @"Warning.Effect.StudioLight";
+            effectsCount++;
+        }
+    }
+    if (@available(macOS 14.0, *)) {
+        /// Reaction effects do not follow the same paradigm as other effects in terms of checking whether they are active. According to Apple, this is because a device instance property `reactionEffectsActive` would have been ambiguous (conflicting with whether a reaction is currently rendering). Instead, Apple exposes the `AVCaptureDevice.reactionEffectGesturesEnabled` class property (which apparently exists, but is hidden/private for all other effects) to tell us whether the effect is enabled application-wide, while `device.canPerformReactionEffects` tells us whether the device's active format currently supports the effect. The logical conjunction of these two properties tells us whether putting our thumbs inside the video frame will make fireworks appear. The device instance properties for other effects are a convenience 'shorthand' for this private class/instance property combination.
```
PatTheMav, line 714: Please add some paragraph breaks (you should be able to add an empty line via `///`) to make this more readable.

plugins/mac-avcapture/plugin-main.m
```diff
@@ -79,42 +79,52 @@ static void av_fast_capture_set_defaults(obs_data_t *settings)
...
 
     obs_properties_t *properties = obs_properties_create();
 
     // Create Properties
     obs_property_t *device_list = obs_properties_add_list(properties, "device", obs_module_text("Device"),
                                                           OBS_COMBO_TYPE_LIST, OBS_COMBO_FORMAT_STRING);
+
+    obs_property_t *effects_warning;
+    if (has_effects) {
+        effects_warning = obs_properties_add_text(properties, "effects_warning", effects_warning_string, OBS_TEXT_INFO);
```
PatTheMav, line 98: Why not use the return of `obs_module_text(effectsWarningKey.UTF8String)` here directly and also use `if (effectsWarningKey)` to check for a valid key (`effectsWarningForDevice` returns nil if no warnings are available).

This avoids an additional check for `has_effects` and the declaration of a boolean and a `char` array. `effectsWarningKey` can be used for the later check as well.

plugins/mac-avcapture/OBSAVCapture.m
jcm93: Changed. Used one line of code per paragraph still; let me know if you want different formatting (such as picking a column width and justifying manually with more lines, etc.)

plugins/mac-avcapture/plugin-main.m
jcm93: Thanks, fixed.

plugins/mac-avcapture/AVCaptureDeviceFormat+OBSListable.m
```diff
@@ -77,13 +77,15 @@ - (NSNumber *)aspectRatioComparisonValue
         CMFormatDescriptionRef formatDescription = self.formatDescription;
         CMVideoDimensions dimensions = CMVideoFormatDescriptionGetDimensions(formatDescription);
 
-        //for sorting purposes, treat aspect ratios sufficiently close together as equivalent
-        double ratio = (double) dimensions.width / (double) dimensions.height;
-        double roundedRatio = round(ratio * 100) / 100;
-        NSNumber *compareRatio = @(roundedRatio);
+        if (dimensions.height != 0) {
```
PatTheMav, line 80: This would evaluate to true if `dimensions.height` is `-500` - is that desired?

jcm93: I'm not sure. My priority was to avoid a crash if a card advertises some nonsense format with zero in some dimension and that squeaks past AVFoundation.

It's may be more 'correct'/desired to just not assign an aspect ratio sort value to those hypothetical formats with negative dimensions, I suppose? I would have to double check that an NSComparison between an object and nil doesn't raise an exception in that case too.

jcm93: Changed to just not set a value for anything zero or negative, and also changed the NSString representation to return `@""` if either dimension is <= 0. Verified that nil values for the aspect ratio do not crash if used in a sort descriptor.

jcm93:
Some submodules went weird on the latest commit, and in the process of fixing it I went ahead and rebased and squashed the commits.

Warchamp7:
> Screenshot 2024-07-10 at 1 04 07 AM Screenshot 2024-07-10 at 1 14 52 AM

Assuming this is the current state of the lists, this looks great to me!

jcm93:
The only difference from there to the current state is the decimal formatting. Here is the current state of the list format for the devices I have available:
Screenshot 2024-07-31 at 12 26 35 PMScreenshot 2024-07-31 at 12 26 48 PMScreenshot 2024-07-31 at 12 26 48 PM

PatTheMav:
> The only difference from there to the current state is the decimal formatting. Here is the current state of the list format for the devices I have available: Screenshot 2024-07-31 at 12 26 35 PMScreenshot 2024-07-31 at 12 26 48 PMScreenshot 2024-07-31 at 12 26 48 PM

The third screenshot still show separate framerates, is that because each is its own framerate range or is that just an old screenshot?

Also I don't know what "Default" is supposed to mean in the second screenshot? Is it undefined? OBS will need to interpret it as _some_ value, so which one is it and why not show the user?

jcm93:
In the third screenshot, yes, each of those is its own frame rate range on the format where the range consists of one discrete value (min and max are equal). I do not have a good sense of how common of a pattern this is. If it is very common, with long lists of ranges on each format, I think I might need to reconsider the string representation to be more simple for large numbers of frame rate ranges.

In the second screenshot, this is just the weird way Continuity Camera advertises itself. Its formats have no YUV matrix attached (or primaries or transfer function), so OBS concludes `VIDEO_CS_DEFAULT`. This is existing behavior. When OBS eventually receives sample buffers from Continuity Camera, they seem to be in the 709 color space. I agree that this isn't super desirable, but I don't think there's a great way around it without very special handling.

PatTheMav:
> In the third screenshot, yes, each of those is its own frame rate range on the format where the range represents one discrete value (min and max are equal). I do not have a good sense of how common of a pattern this is. If it is very common, with long lists of ranges on each format, I think I might need to reconsider the string representation to be more simple for large numbers of frame rate ranges.
> 
> In the second screenshot, this is just the weird way Continuity Camera advertises itself. Its formats have no YUV matrix attached (or primaries or transfer function), so OBS concludes `VIDEO_CS_DEFAULT`. This is existing behavior. When OBS eventually receives sample buffers from Continuity Camera, they seem to be in the 709 color space. I agree that this isn't super desirable, but I don't think there's a great way around it without very special handling.

Right, yeah I dunno if the color space selection even makes sense, though I reported the discrepancy between selection and what is actually used in the sample buffers to Apple and I haven't checked if this has been changed in macOS 15 yet.

Warchamp7:
> The only difference from there to the current state is the decimal formatting. Here is the current state of the list format for the devices I have available: 

In that case this looks good on my end from the UI side

PatTheMav:
@jcm93 Could you rebase this please?

jcm93:
Rebased, and also changed `AVCaptureDeviceFormat+OBSListable.m`'s  `- (NSString *)obsPropertyListDescription` to more gracefully handle bad values.

I chose to just not include string values if they are invalid, unspecified, or unknown. So if a device provides resolution, color space, pixel format, but not FPS, the string description will just be "\<resolution> - \<color space> - \<pixel format>", instead of showing `" fps"` for an unspecified fps, or `"0x0 ()"` for an invalid resolution. This can sometimes happen for color spaces, i.e. Continuity Camera does not specify a color space in its format advertisements.

If the resultant string is completely empty, that means that all parameters are invalid, unspecified, or unknown. For this case, we show "Default" for the format instead of an empty string. This occurs with iPhone screen mirroring.

Hopefully (!) this PR should now be in a good state!

RytoEX:
I'd like to see an approval from someone who runs macOS daily and I'd also like to see a UI/UX approval from @Warchamp7 .

Warchamp7 (approved):

plugins/mac-avcapture/OBSAVCapture.m
```diff
@@ -649,16 +618,14 @@ - (BOOL)updateSessionwithError:(NSError *__autoreleasing *)error
 
 #pragma mark - OBS Settings Helpers
 
-+ (CMVideoDimensions)dimensionsFromSettings:(void *)settings
++ (CMVideoDimensions)legacyDimensionsFromSettings:(void *)settings
```
PatTheMav, line 621: I don't think this needs to be prefixed with "legacy" as it's a getter function to convert the string from settings into an actual `CMVideoDimensions` object. It might only be _used_ by legacy code, but it's still a correct getter method.

PatTheMav (changes requested):
Looks fine apart from my other comment(s), also worked as expected in local tests.

plugins/mac-avcapture/OBSAVCapture.m
jcm93: Hmm, I had it marked as `legacy` because we won't even store the resolution directly as its own key/value any longer, since that value is now encapsulated in the stored format. So it seems like an incorrect getter if it was called `dimensionsFromSettings` unless it also tried to parse the dimensions from the new `supported_format` key, no? Since it won't work unless you're calling it for a source that has the old set of properties.

PatTheMav: Ah forgive me, you're right of course - the setting itself is not some global, universal, OBS dimension setting, but a setting specifically used for legacy configurations. Ignore my comment then.

jcm93:
Per discussion out-of-band, we needed to add an internal representation for the value of device formats separate from their human-readable descriptions. This is so that we can freely change the format of device format descriptions if we want to without requiring any special handling to migrate sources. I did this in a very simplistic way with [obsPropertyListInternalRepresentation](https://github.com/obsproject/obs-studio/pull/10895/files#diff-b5dcba3f140b061d005256627203d5581bf038c1875a335cb0f294b3e643c30dR54), serializing string values with the dimensions, frame rate ranges, and raw values for the OBS color space enum and the CMFormatDescription subtype. I would welcome any scrutiny on whether there is a superior way to represent these internally than what I picked here.

plugins/mac-avcapture/AVCaptureDeviceFormat+OBSListable.h
```diff
@@ -0,0 +1,26 @@
...
+@property (nonatomic, strong) NSNumber *aspectRatioComparisonValue;
+
+/// Lazily computed property containing the bits per pixel for the format. For planar formats, accounts for the number of bytes of chrominance and luminance per pixel block.
+@property (nonatomic, strong) NSNumber *bitsPerPixel;
+
+/// Lazily computed property containing a general localized description of the format, suitable for use in picking a format in the OBS source properties window.
+@property (nonatomic, strong) NSString *obsPropertyListDescription;
+
+/// Lazily ocmputed property containing the string value used by OBS to represent and uniquely identify the format by its dimensions, supported frame rate ranges, color space and [four character code](https://developer.apple.com/documentation/coremedia/1489255-cmformatdescriptiongetmediasubty?language=objc).
```
RytoEX, line 23:
```suggestion
/// Lazily computed property containing the string value used by OBS to represent and uniquely identify the format by its dimensions, supported frame rate ranges, color space and [four character code](https://developer.apple.com/documentation/coremedia/1489255-cmformatdescriptiongetmediasubty?language=objc).
```

RytoEX (changes requested):

plugins/mac-avcapture/AVCaptureDeviceFormat+OBSListable.h
jcm93: Thanks, fixed.

jcm93:
Revised to add a trivial base64 mask / hash for the internal value used by OBS to identify device formats. Leaving this value human-readable risked the impression that this value could or should be modified by users.

Masking this value also more or less precludes OBS from being aware of (hypothetical) future material breakages in macOS in how device formats are represented. Elegantly responding to such a breakage would be a large maintenance burden anyway, however, so this seems an acceptable tradeoff. Even in the far off future, a device format should still be encapsulated by 1. its resolution, 2. its supported frame rate ranges, 3. its four character code, and 4. its native color space, all four of which are encapsulated in our mask. So our chosen mask should still be relatively "future proof."

RytoEX (approved):

plugins/mac-avcapture/AVCaptureDeviceFormat+OBSListable.m
```diff
@@ -0,0 +1,153 @@
...
+        for (AVFrameRateRange *range in sortedRanges) {
+            frameRatesInternalDescription = [frameRatesInternalDescription
+                stringByAppendingFormat:@"%.3f-%.3f", range.minFrameRate, range.maxFrameRate];
+        }
+        NSString *internalRepresentation =
+            [NSString stringWithFormat:@"%dx%d %@ %i %i", formatDimensions.width, formatDimensions.height,
+                                       frameRatesInternalDescription, deviceColorSpace, formatSubType];
+        NSData *internalRepresentationBytes = [internalRepresentation dataUsingEncoding:NSUTF8StringEncoding];
+        NSString *base64String = [internalRepresentationBytes base64EncodedStringWithOptions:0];
+        objc_setAssociatedObject(self, @selector(obsPropertyListInternalRepresentation), base64String,
```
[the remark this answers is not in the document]
jcm93: According to the [runtime source](https://github.com/apple-oss-distributions/objc4/blob/89543e2c0f67d38ca5211cea33f42c51500287d5/runtime/objc-runtime-new.mm#L9284) associated object references are indeed taken care of whenever the object itself is being deallocated. The only actual verification I did on my end however was creating/destroying sources, adding/removing devices and checking that no zombies/leaks were reported.

jcm93:
Made one small fix; I wanted to make sure this PR logs the format a device is configuring with, and I noticed that we were logging the internal representation of the format (which is base64 scrambled) rather than the human readable version; now fixed. Also rebased; will squash commits whenever we get around to looking at this again post-31.

PatTheMav:
@RytoEX @jcm93 What's the status here? Can we make this land at the next best possible release?

jcm93:
As far as I am concerned it is in good shape, and I remain able to address any other feedback. I will rebase and squash the commits.

RytoEX:
> @RytoEX @jcm93 What's the status here? Can we make this land at the next best possible release?

This didn't land for 31.0. The intent at the time was to land it for the next release.

jcm93:
Per discussions out of band, I am adding one last feature to this in a separate commit; if the user selects a format that does not support their previously configured frame rate, we will fall back to an appropriate FPS supported by the format based on OBS's configured output frame rate. This fallback frame rate is determined by the new function `fallbackFrameRateForFormat:`.

After sitting on it, I think that the benefits are large enough that the addition of this functionality is merited. Namely, it seems like selecting a navigating to a format in the properties window should try to offer immediate feedback to help the user select a format, rather than requiring fiddling with the frame rate selector on each selection just to get a visual preview. Hitting cancel in the window is always a reliable "bail out" to return to the previous committed configuration.

While it is a relatively small addition, it's nontrivial enough that I am adding it in a separate commit so it can be looked at, and I can squash the commits later.

Warchamp7 (approved):

PatTheMav (commented):
Works as advertised apart from the code-style issues I commented on separately.

I'm not really a fan of the aspect ratio in the format drop-down and exposing all the frame rate ranges in it: With some devices I tested with this list becomes a big bunch of textual noise, so the usefulness of the information is drowned out by that.

Personally I'd rather have something that is simple and low-effort to parse for the user with a straight-forward and consistent behaviour. The existence of a bespoke frame rate selector makes exposing the ranges in the first dropdown redundant.

@Warchamp7 what do you think? I've posted some screenshots off-thread that display the "noise" I mention.

plugins/mac-avcapture/AVCaptureDeviceFormat+OBSListable.m
```diff
@@ -0,0 +1,153 @@
...
+    return objc_getAssociatedObject(self, @selector(bitsPerPixel));
+}
+
+- (NSNumber *)pixelBandwidthComparisonValue
+{
+    if (!objc_getAssociatedObject(self, @selector(pixelBandwidthComparisonValue))) {
+        CMFormatDescriptionRef formatDescription = self.formatDescription;
+        CMVideoDimensions dimensions = CMVideoFormatDescriptionGetDimensions(formatDescription);
+        NSNumber *bandwidth =
+            @(dimensions.width * dimensions.height * self.videoSupportedFrameRateRanges.firstObject.maxFrameRate);
```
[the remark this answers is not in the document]
jcm93: Fixed to sort the ranges before computing this value.

plugins/mac-avcapture/OBSAVCapture.m
```diff
@@ -946,6 +1032,39 @@ + (FourCharCode)fourCharCodeFromFormat:(OBSAVCaptureVideoFormat)format
...
+                    [frameRateDescription stringByAppendingFormat:@"%.0f-%.0f", minFrameRate, maxFrameRate];
+            } else {
+                frameRateDescription =
+                    [frameRateDescription stringByAppendingFormat:@"%.2f-%.2f", minFrameRate, maxFrameRate];
+            }
+        }
+        index++;
+        if (index < (ranges.count)) {
+            frameRateDescription = [frameRateDescription stringByAppendingString:@", "];
+        }
```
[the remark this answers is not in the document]
jcm93: Addressed.

plugins/mac-avcapture/AVCaptureDeviceFormat+OBSListable.m
```diff
@@ -0,0 +1,153 @@
...
+        NSString *pixelFormatDescription = [OBSAVCapture stringFromSubType:formatSubType];
+        OBSAVCaptureColorSpace deviceColorSpace = [OBSAVCapture colorspaceFromDescription:self.formatDescription];
+        NSString *colorspaceDescription = [OBSAVCapture stringFromColorspace:deviceColorSpace];
+        NSString *fpsRangesDescription = [OBSAVCapture frameRateDescription:self.videoSupportedFrameRateRanges];
+        NSString *aspectRatioString = [OBSAVCapture aspectRatioStringFromDimensions:formatDimensions];
+
+        NSString *resolutionDescriptionComponent;
+        if (formatDimensions.width > 0 && formatDimensions.height > 0) {
+            resolutionDescriptionComponent = [NSString
+                stringWithFormat:@"%dx%d (%@) - ", formatDimensions.width, formatDimensions.height, aspectRatioString];
```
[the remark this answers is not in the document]
jcm93: Addressed mostly; we still pass some empty strings around which I will look at, though I verified everything works correctly (including showing "Default" for the case where a device format contains no specific information whatsoever).

jcm93:
Added a second commit to address some visual noise when there are a great many FPS ranges present on a format.

The approach taken in the second commit simplifies the FPS ranges description if there are more than ten ranges, showing the minimum and maximum FPS in the array of ranges and the number of ranges. This can happen if a card advertises many discrete values in the form of ranges, i.e. 30-30, 45-45, 60-60. 

This scenario is currently disambiguated from a "genuine range" with a parenthetical indicating the number of ranges. So for example a format advertising 60 ranges between 1 and 120 fps would show "1 - 120 FPS (60 values)". While a genuine FPS range between 1-120 (i.e. any frame rate between 1-120 is supported) would show just "1 - 120 FPS".

This is not perfect, but it seems necessary to handle this edge case if a format has, for example, 60 FPS ranges, each representing discrete integers between 1 and 60.

plugins/mac-avcapture/AVCaptureDeviceFormat+OBSListable.m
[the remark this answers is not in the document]
jcm93: Changed to use a null NSString for the default FPS ranges description which is a bit neater.

PatTheMav (approved):
Works fine with all devices available to me.

Warchamp7 (approved):

Warchamp7:
My final request for this is gonna be to change the discrete FPS listings to be from low to high, so that it's consistent with FPS ranges being low to high.

plugins/mac-avcapture/OBSAVCapture.h
```diff
@@ -58,10 +62,13 @@ typedef struct av_capture {
 /// C struct for sample buffer validity checks in capture callback
 typedef struct av_capture_info {
     OBSAVCaptureColorSpace colorSpace;
-    OBSAVCaptureVideoRange videoRange;
+    FourCharCode fourCC;
     bool isValid;
 } OBSAVCaptureVideoInfo;
 
+/// Defines the maximum number of frame rate ranges to show complete information for inside of a device format description before providing a more generic description of the supported frame rates.
+static const UInt32 kMaxFrameRateRangesInDescription = 10;
```
PatTheMav, line 70: Declaring a `static` variable in a header creates a copy of this value for _each_ compilation unit that includes the header (regardless of whether it's actually used). Is this the desired result?

jcm93: After considering it I think it might be more appropriate to move this declaration to the top of the implementation file (after imports), since only the implementation file needs to care about it. That seems to be one conventional way to declare these types of constants.

jcm93: Changed.

PatTheMav (approved):

PR #10896, opened 2024-06-19
libobs: Fix destroy_on_stop functionality of encoder groups

tt2468:
### Description
Fixes a crash when the following steps occur:
- Encoder group created and then started with encoders, only the group owning encoder refs
- Encoder group destroy called: group has `destroy_on_stop` set without actual destroy
- Outputs holding encoders from stopping are stopped
- `remove_connection()` function destroys encoder group, releasing encoders and causing the currently processed encoder to be destroyed early
- parent scopes try to access destroyed encoder pointer and crash

This change moves some logic around to improve the release/destruct order of `obs_encoder_stop()` to fix the race above.

Note: Cases of encoder errors will not destruct the group if it has `destroy_on_stop` set, as the encoder is also not destroyed if it also is set to destroy on stop. This part of the code should be revisited at a later date and fixed up to prevent memory leaks.

Closes #10890 

### Motivation and Context
Shutting down OBS with an EB stream running was resulting in a crash when the output object was destroyed after "destroy" was called on the group.

### How Has This Been Tested?
Todo testing

### Types of changes

### Checklist:

### DIFF
--- a/libobs/obs-encoder.c
+++ b/libobs/obs-encoder.c
@@ -786,13 +786,18 @@ void obs_encoder_start(obs_encoder_t *encoder,
 	pthread_mutex_unlock(&encoder->init_mutex);
 }
 
-static inline void obs_encoder_stop_internal(
-	obs_encoder_t *encoder,
-	void (*new_packet)(void *param, struct encoder_packet *packet),
-	void *param)
+void obs_encoder_stop(obs_encoder_t *encoder,
+		      void (*new_packet)(void *param,
+					 struct encoder_packet *packet),
+		      void *param)
 {
 	bool last = false;
 	size_t idx;
+
+	if (!obs_encoder_valid(encoder, "obs_encoder_stop"))
+		return;
+	if (!obs_ptr_valid(new_packet, "obs_encoder_stop"))
+		return;
 
 	pthread_mutex_lock(&encoder->init_mutex);
 	pthread_mutex_lock(&encoder->callbacks_mutex);
@@ -833,21 +838,6 @@ static inline void obs_encoder_stop_internal(
 	}
 
 	pthread_mutex_unlock(&encoder->init_mutex);
-}
-
-void obs_encoder_stop(obs_encoder_t *encoder,
-		      void (*new_packet)(void *param,
-					 struct encoder_packet *packet),
-		      void *param)
-{
-	bool destroyed;
-
-	if (!obs_encoder_valid(encoder, "obs_encoder_stop"))
-		return;
-	if (!obs_ptr_valid(new_packet, "obs_encoder_stop"))
-		return;
-
-	obs_encoder_stop_internal(encoder, new_packet, param);
 }
 
 const char *obs_encoder_get_codec(const obs_encoder_t *encoder)
--- a/libobs/obs-encoder.c
+++ b/libobs/obs-encoder.c
@@ -371,6 +371,8 @@ static void remove_connection(struct obs_encoder *encoder, bool shutdown)
 	 * up again */
 	if (shutdown)
 		obs_encoder_shutdown(encoder);
+	encoder->initialized = false;
+
 	set_encoder_active(encoder, false);
 }
 
@@ -810,7 +812,6 @@ static inline bool obs_encoder_stop_internal(
 
 	if (last) {
 		remove_connection(encoder, true);
-		encoder->initialized = false;
 
 		if (encoder->destroy_on_stop) {
 			pthread_mutex_unlock(&encoder->init_mutex);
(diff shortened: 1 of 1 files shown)


tt2468:
@palana could you give this a test when you're around and let me know if this solves the issue? I wasn't able to reproduce a crash on Windows or Linux, but I was able to definitely observe some out of bounds accesses in `master` along with some log warnings which this fixes.

libobs/obs-encoder.c
```diff
@@ -810,34 +812,26 @@ static inline bool obs_encoder_stop_internal(
...
-		      void *param)
-{
-	bool destroyed;
+		/* destroying the group all the way back here prevents a race where destruction
+		 * of the group can prematurely destroy the encoder within internal functions.
+		 * this is the point where it is safe to destroy the group, even if the encoder
+		 * is then also destroyed */
+		if (group && group->destroy_on_stop &&
+		    group->num_encoders_started == 0)
+			obs_encoder_group_actually_destroy(group);
```
palana, line 828:
```suggestion
		if (!group)
			return;
		
		pthread_mutex_lock(&group->mutex);
		if (group->destroy_on_stop &&
		    group->num_encoders_started == 0) {
			obs_encoder_group_actually_destroy(group);
		} else {
			pthread_mutex_unlock(&group->mutex);
		}
```

`obs_encoder_group_actually_destroy` currently unlocks the mutex before destroying it, so the mutex needs to be locked before calling `obs_encoder_group_actually_destroy`; additionally, accesses to group members are protected by the mutex

palana (changes requested):
this fixes the crash for me, however I haven't been able to convince myself that the changes to e.g. `encoder->initialized` and `obs_encoder_stop` behavior are sound

also, see comment about the group mutex

tt2468:
Issue should be resolved now.

tt2468:
> I haven't been able to convince myself that the changes to e.g. encoder->initialized and obs_encoder_stop behavior are sound

Both of these changes are simply moving code. `encoder->initialized = false` was always called in the two places it was removed from, which was always right after the call to remove_connection(). `obs_encoder_stop_internal()` only existed AFAIK because of the bool response which determined whether to unlock the mutex. It was only ever invoked from `obs_encoder_stop()`.

palana (commented):
> > I haven't been able to convince myself that the changes to e.g. encoder->initialized and obs_encoder_stop behavior are sound
> 
> Both of these changes are simply moving code. `encoder->initialized = false` was always called in the two places it was removed from, which was always right after the call to remove_connection(). `obs_encoder_stop_internal()` only existed AFAIK because of the bool response which determined whether to unlock the mutex. It was only ever invoked from `obs_encoder_stop()`.

@tt2468: I don't currently see `encoder->initialized = false;` in the code path that calls `obs_encoder_group_actually_destroy`; but I also haven't (yet) verified what code paths end up in that place, i.e. is there a path where the encoder is not destroyed after reaching that code? I'm thinking there might be, but not sure yet; will take another look early next week

tt2468:
I can assure you that moving `encoder->initialized` in the first commit changes no behavior. It was just worth doing to keep the code concise.

libobs/obs-encoder.c
```diff
@@ -371,6 +365,8 @@ static void remove_connection(struct obs_encoder *encoder, bool shutdown)
 	 * up again */
 	if (shutdown)
 		obs_encoder_shutdown(encoder);
+	encoder->initialized = false;
+
 	set_encoder_active(encoder, false);
```
palana, line 370: @tt2468:
> I can assure you that moving `encoder->initialized` in the first commit changes no behavior. It was just worth doing to keep the code concise.

the order of `encoder->initialized = false` and `set_encoder_active(encoder, false)` is swapped with this change; I don't currently see if this change is significant, however

arguably the old behavior was probably wrong, since e.g. the call chain around `full_stop` didn't include any memory barriers after `encoder->initialized = false;`?

this is probably fine in the grand scheme of things, and I also mostly prefer that this is fixed in libobs, rather than requiring [libobs-user changes](https://github.com/obsproject/obs-studio/pull/10890) (this* is already part of Twitch beta `30.2.0-beta4-enhanced-broadcasting-v27` see https://github.com/amazon-contributing/upstreaming-to-obs-studio/commits/release/v27/, and we haven't gottan any issue reports related to this)

RytoEX:
Based on off-thread discussion, it seems this approach has enough consensus to merge. I'll fix up some nits, and then take care of this.

libobs/obs-encoder.c
RytoEX: Resolving per off-thread discussion and because merging requires conversations to be closed.

PR #10898, opened 2024-06-20
win-wasapi: Remove noisy and useless debug logging

derrod:
### Description

Removes some leftover debug logging messages that have outlived their usefulness.

### Motivation and Context

These were originally added to help troubleshoot the application audio capture issues. Since it has been established that those are a Windows bug we can't detect or work around, we don't need these any longer.

They also keep annoying me because I run OBS with verbose logging for testing and especially the silent flag gets quite spammy.

### How Has This Been Tested?

N/A

### Types of changes

### Checklist:

### DIFF
--- a/plugins/win-wasapi/win-wasapi.cpp
+++ b/plugins/win-wasapi/win-wasapi.cpp
@@ -1109,16 +1109,7 @@ bool WASAPISource::ProcessCaptureData()
 			sawBadTimestamp = true;
 		}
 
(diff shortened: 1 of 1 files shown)


Lain-B (approved):

Lain-B:
Just going to merge this for 30.1 because it's trivial and it just cleans up the debug output. I'd rather have unnecessary noise even if it is in verbose logging. Additionally I don't think these logging message have ever proven be actually be useful in any way/shape/form.

PR #10899, opened 2024-06-20
linux-pipewire: Remove extra loop teardown

kkartaltepe:
### Description

pw_thread_loop_wait must be called with the lock held, so this code was throwing warnings on newer pipewire libraries. With it removed there seems to be no reason to perform a stop before destroying the loop which will internally call stop since at least 7 years ago.

https://gitlab.freedesktop.org/pipewire/pipewire/-/blame/d0a2e6316bb652b064db8f2ff63b86afd52bddfb/src/pipewire/thread-loop.c#L247

### Motivation and Context

No more pipewire warnings on shutdown.

### How Has This Been Tested?

Test scene shutsdown fine still.

### Types of changes

### Checklist:

### DIFF
--- a/plugins/linux-pipewire/pipewire.c
+++ b/plugins/linux-pipewire/pipewire.c
@@ -173,11 +173,6 @@ static void update_pw_versions(obs_pipewire *obs_pw, const char *version)
 
 static void teardown_pipewire(obs_pipewire *obs_pw)
 {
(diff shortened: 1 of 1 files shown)


GeorgesStavracas (approved):

pobrn:
The loop should be stopped or at least its lock held when the context is destroyed.

PR #10901, opened 2024-06-20
UI: Try to opt out of reaction effects by default on macOS

jcm93:
> [!IMPORTANT]
> This PR most needs testing by a user on macOS 14+ who has not yet modified system camera effects with OBS open, to see if "reaction" effects trigger for camera inputs.

### Description

Adds the `NSCameraReactionEffectsEnabled = false` key-value pair to the default Info.plist file on macOS, in an attempt to make reaction effects not appear by default for all camera inputs in OBS.

### Motivation and Context

Much has been written complaining about macOS system "reaction" effects inappropriately triggering in different scenarios on macOS, due to Apple's decision to opt all applications into these effects by default for all camera inputs. This PR is an attempt to make OBS opt out of these effects by default.

### How Has This Been Tested?

Tested locally on my machine (on the macOS Sequoia beta) to verify that if the user intentionally enables reaction effects with OBS open, the setting will still persist between restarts of OBS. Thus, if users want reaction effects, this PR will not prevent them from enabling them and keeping them enabled.

Was *unable* to test if this PR successfully prevents reaction effects from appearing by default. This is because there is no mechanism in macOS (that I know of) to reset the status of effects preferences for programs across the system. Attempted to test in a VM, but no existing VM software (that I know of) supports camera input for macOS VMs.

This could also use testing from users still on macOS Sonoma to verify that effects settings continue to persist there as they do on macOS Sequoia.

Even if this PR is ineffective at preventing reaction effects from appearing by default, it is unlikely to hurt either.

### Types of changes

### Checklist:

### DIFF
--- a/UI/cmake/macos/Info.plist.in
+++ b/UI/cmake/macos/Info.plist.in
@@ -4,6 +4,8 @@
 <dict>
 	<key>NSHighResolutionCapable</key>
 	<true/>
+	<key>NSCameraReactionEffectsEnabled</key>
+	<false/>
 	<key>LSAppNapIsDisabled</key>
 	<true/>
 	<key>SUFeedURL</key>


PatTheMav:
I don't think this change has any value because per the official documentation the setting is merely for iOS apps to be included in the default activation of these effects (as only apps with voice over IP background activity get this enabled by default otherwise).

AliceMousie:
I tested this commit out on MacOS 15.1.1, I had not touched the reactions settings previously since I never use a camera on OBS. Reactions were still enabled (lit up green in the camera dropdown menu) with this build, although I'm not sure if this config change would not be picked up due to having OBS installed already?

[The documentation does now list MacOS](https://developer.apple.com/documentation/avfoundation/avcapturedevice/reactioneffectsenabled), and anecdotally I have seen a podcast struggle with reaction effects appearing when they record with OBS. The setting is hard to find, and it's not obvious or intuitive that it applies per-app. These are Apple's shortcomings but it would still be good to find a workaround to this if possible.

PatTheMav:
> I tested this commit out on MacOS 15.1.1, I had not touched the reactions settings previously since I never use a camera on OBS. Reactions were still enabled (lit up green in the camera dropdown menu) with this build, although I'm not sure if this config change would not be picked up due to having OBS installed already?
> 
> [The documentation does now list MacOS](https://developer.apple.com/documentation/avfoundation/avcapturedevice/reactioneffectsenabled), and anecdotally I have seen a podcast struggle with reaction effects appearing when they record with OBS. The setting is hard to find, and it's not obvious or intuitive that it applies per-app. These are Apple's shortcomings but it would still be good to find a workaround to this if possible.

The documentation has always listed macOS here, because the page refers to the availability of the `reactionEffectsEnabled` property on an `AVDevice` instance. It has no effect on the availability of the `NSCameraReactionEffectsEnabled` property list key.

Further the page states (emphasis mine):

> In macOS, the system enables reaction effects **for all apps by default**. In iOS, it enables them by default **only for video conferencing apps** (apps that enable the Voice over IP option in their [UIBackgroundModes](https://developer.apple.com/documentation/BundleResources/Information-Property-List/UIBackgroundModes) configuration). Apps that don’t use this background mode may **opt in** to this feature by addiing the following key

My reading of this is that the property key is only supposed to allow iOS apps that are _not_ video conferencing apps to also get this feature enabled for them. As it is enabled by default for all apps on macOS, the key has no effect.

jcm93:
Closing; superseded by https://github.com/obsproject/obs-studio/pull/11652.

PR #10902, opened 2024-06-20
obs-qsv11: Fix CBR Spike

thyintel:
### Description

CBR improvement for new Intel platform, the peak bitrate is now down to 120% on average from 200% while keeping same bitrate and quality. 

This PR will not build until VPL version has been updated: https://github.com/obsproject/obs-deps/pull/254.

### Motivation and Context

This change has been made to decrease the CBR spike on new Intel platforms.

### How Has This Been Tested?

new Intel platforms on various driver versions

### Types of changes

### Checklist:

- [ x] I have read the [**contributing** document](https://github.com/obsproject/obs-studio/blob/master/CONTRIBUTING.rst).
- [ x] My code is not on the master branch.
- [ x] The code has been tested.
- [ x] All commit messages are properly formatted and commits squashed where appropriate.
- [ x] I have included updates to all appropriate documentation.

### DIFF
--- a/plugins/obs-qsv11/QSV_Encoder_Internal.cpp
+++ b/plugins/obs-qsv11/QSV_Encoder_Internal.cpp
@@ -177,6 +177,26 @@ mfxStatus QSV_Encoder_Internal::Open(qsv_param_t *pParams, enum qsv_codec codec)
 	return sts;
 }
 
+PRAGMA_WARN_PUSH
+PRAGMA_WARN_DEPRECATION
+static inline bool HasOptimizedBRCSupport(const mfxPlatform &platform,
+					  const mfxVersion &version,
+					  mfxU16 rateControl)
+{
+#if (MFX_VERSION_MAJOR >= 2 && MFX_VERSION_MINOR >= 12) || MFX_VERSION_MAJOR > 2
+	if ((version.Major >= 2 && version.Minor >= 12) || version.Major > 2)
+		if (rateControl == MFX_RATECONTROL_CBR &&
+		    (platform.CodeName >= MFX_PLATFORM_LUNARLAKE &&
+		     platform.CodeName != MFX_PLATFORM_ALDERLAKE_N))
+			return true;
+#endif
+	UNUSED_PARAMETER(platform);
+	UNUSED_PARAMETER(version);
+	UNUSED_PARAMETER(rateControl);
+	return false;
+}
+PRAGMA_WARN_POP
+
 mfxStatus QSV_Encoder_Internal::InitParams(qsv_param_t *pParams,
 					   enum qsv_codec codec)
 {
@@ -233,10 +253,18 @@ mfxStatus QSV_Encoder_Internal::InitParams(qsv_param_t *pParams,
 	switch (pParams->nRateControl) {
 	case MFX_RATECONTROL_CBR:
 		m_mfxEncParams.mfx.TargetKbps = pParams->nTargetBitRate;
-		m_mfxEncParams.mfx.BufferSizeInKB =
-			(pParams->nTargetBitRate / 8) * 2;
+
+		if (HasOptimizedBRCSupport(platform, m_ver,
+					   pParams->nRateControl)) {
+			m_mfxEncParams.mfx.BufferSizeInKB =
+				(pParams->nTargetBitRate / 8) * 1;
+		} else {
+			m_mfxEncParams.mfx.BufferSizeInKB =
+				(pParams->nTargetBitRate / 8) * 2;
+		}
+
 		m_mfxEncParams.mfx.InitialDelayInKB =
-			(pParams->nTargetBitRate / 8) * 1;
+			m_mfxEncParams.mfx.BufferSizeInKB / 2;
 		break;
 	case MFX_RATECONTROL_VBR:
 		m_mfxEncParams.mfx.TargetKbps = pParams->nTargetBitRate;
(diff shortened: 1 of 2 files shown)


RytoEX:
> This PR will not build until VPL update (predicted by end of this month).

Even when the new update is available, this code will probably need ifdef'd on the libvpl version, because obs-qsv11 is available on Linux where we do not control the version of libvpl available. On Ubuntu 22.04, which we still support, the version of libvpl available is 2022.1.0 (API 2.6). On Ubuntu 24.04, the version of libvpl available is 2023.3.0 (API 2.9).

thyintel:
> > This PR will not build until VPL update (predicted by end of this month).
> 
> Even when the new update is available, this code will probably need ifdef'd on the libvpl version, because obs-qsv11 is available on Linux where we do not control the version of libvpl available. On Ubuntu 22.04, which we still support, the version of libvpl available is 2022.1.0 (API 2.6). On Ubuntu 24.04, the version of libvpl available is 2023.3.0 (API 2.9).

Sounds good. I will add the ifdef when the new version of libvpl is released.

thyintel:
> Now that #11097 has landed, please rebase this.

Rebased!

plugins/obs-qsv11/QSV_Encoder_Internal.cpp
```diff
@@ -177,6 +177,23 @@ mfxStatus QSV_Encoder_Internal::Open(qsv_param_t *pParams, enum qsv_codec codec)
 	return sts;
 }
 
+PRAGMA_WARN_PUSH
+PRAGMA_WARN_DEPRECATION
+static inline bool IsOptimizedBRCSupport(const mfxPlatform &platform,
+					 mfxU16 rateControl)
+{
+#if defined(_WIN32)
```
RytoEX, line 185: This seems a bit like a sledgehammer and will prevent any non-Windows platform from being able to utilize this even if they are building against libvpl 2.12 or newer. Would it not be more robust to use the `MFX_VERSION` defines?
```suggestion
#if MFX_VERSION_MAJOR >= 2 && MFX_VERSION_MINOR >= 12
```
or:
```suggestion
#if MFX_VERSION >= 2012
```

Unless there is a driver/runtime requirement as well.

plugins/obs-qsv11/QSV_Encoder_Internal.cpp
```diff
@@ -177,6 +177,23 @@ mfxStatus QSV_Encoder_Internal::Open(qsv_param_t *pParams, enum qsv_codec codec)
...
+	if (rateControl == MFX_RATECONTROL_CBR &&
+	    (platform.CodeName >= MFX_PLATFORM_LUNARLAKE &&
+	     platform.CodeName != MFX_PLATFORM_ALDERLAKE_N))
+		return true;
+#endif
+	(void)platform;
+	(void)rateControl;
```
RytoEX, line 192:
```suggestion
	UNUSED_PARAMETER(platform);
	UNUSED_PARAMETER(rateControl);
```

RytoEX (changes requested):

plugins/obs-qsv11/QSV_Encoder_Internal.cpp
```diff
@@ -177,6 +177,23 @@ mfxStatus QSV_Encoder_Internal::Open(qsv_param_t *pParams, enum qsv_codec codec)
 	return sts;
 }
 
+PRAGMA_WARN_PUSH
+PRAGMA_WARN_DEPRECATION
+static inline bool IsOptimizedBRCSupport(const mfxPlatform &platform,
```
RytoEX, line 182: Might be better to name this `HasOptimizedBRCSupport`.

plugins/obs-qsv11/QSV_Encoder_Internal.cpp
```diff
@@ -177,6 +177,23 @@ mfxStatus QSV_Encoder_Internal::Open(qsv_param_t *pParams, enum qsv_codec codec)
...
+					 mfxU16 rateControl)
+{
+#if MFX_VERSION_MAJOR >= 2 && MFX_VERSION_MINOR >= 12
+	if (rateControl == MFX_RATECONTROL_CBR &&
```
RytoEX, line 186: Does this also require a runtime version check like #11060 or is this purely a build dependency requirement?

RytoEX (changes requested):

plugins/obs-qsv11/QSV_Encoder_Internal.cpp
```diff
@@ -177,6 +177,23 @@ mfxStatus QSV_Encoder_Internal::Open(qsv_param_t *pParams, enum qsv_codec codec)
 	return sts;
 }
 
+PRAGMA_WARN_PUSH
+PRAGMA_WARN_DEPRECATION
+static inline bool IsOptimizedBRCSupport(const mfxPlatform &platform,
+					 mfxU16 rateControl)
+{
+#if MFX_VERSION_MAJOR >= 2 && MFX_VERSION_MINOR >= 12
```
RytoEX, line 185: It has occurred to me that should there ever be a libvpl 3.0, this will fail. I'm not sure if that's a scenario we're concerned with, but if it is, this would need to be amended to something like:
```suggestion
#if (MFX_VERSION_MAJOR >= 2 && MFX_VERSION_MINOR >= 12) || MFX_VERSION_MAJOR > 2
```
or, if the construction of `MFX_VERSION` is not expected to change:
```suggestion
#if MFX_VERSION >= 2012
```

RytoEX (changes requested):

plugins/obs-qsv11/QSV_Encoder_Internal.cpp
thyintel: VPL version 2.12 adds the MFX_PLATFORM_LUNARLAKE to mfxcommon.h, which is the reason for this check.

RytoEX: I suppose my question is: does OptimizedBRCSupport required VPL 2.12 on the running system as well, or is it _only_ a build-time requirement?

thyintel: It is a running requirement, but also build because the updated header needs to be on the system.

RytoEX: If it's both, then this needs to also check the runtime version, like in the other PR.

plugins/obs-qsv11/QSV_Encoder_Internal.cpp
```diff
@@ -177,6 +177,23 @@ mfxStatus QSV_Encoder_Internal::Open(qsv_param_t *pParams, enum qsv_codec codec)
 	return sts;
 }
 
+PRAGMA_WARN_PUSH
+PRAGMA_WARN_DEPRECATION
+static inline bool HasOptimizedBRCSupport(const mfxPlatform &platform,
+					  mfxU16 rateControl)
+{
+#if MFX_VERSION_MAJOR >= 2 && MFX_VERSION_MINOR >= 12 || MFX_VERSION_MAJOR > 2
```
RytoEX, line 185:
```suggestion
#if (MFX_VERSION_MAJOR >= 2 && MFX_VERSION_MINOR >= 12) || MFX_VERSION_MAJOR > 2
```

RytoEX (changes requested):

RytoEX (approved):
This seems fine, though I haven't tested it on hardware that would use the new path (Lunar Lake or newer, except Alder Lake N).

PR #10911, opened 2024-06-22
obs-ffmpeg: Check if current NVENC configuration supports 4:4:4 encode

derrod:
### Description

Adds an error and fails encoder initialisation if 4:4:4 is unsupported.

### Motivation and Context

The new fallback encoder supports 444, but not all encoders do (AV1 does not on 40-series cards), instead of erroring out *after* the encoder has started we should abort during initialisation.

Reported on the forums: https://obsproject.com/forum/threads/an-issue-where-an-error-occurs-with-certain-color-formats.176240/

### How Has This Been Tested?

Verified 4:4:4 HEVC still works but AV1 will now show an error.

### Types of changes

### Checklist:


RytoEX (approved):
Looks okay to me.

PR #10912, opened 2024-06-22
obs-outputs: Skip trak box if track has no data

derrod:
### Description

Skips writing a `trak` box for tracks that do not have any data (chunks) when writing the full moov.

### Motivation and Context

This would crash if the encoder failed after starting, but before producing any packets, due to an assumption that there would be at least one chunk. Since there's no point in writing the box at all if there are no chunks (i.e. data) we can just skip the entire box.

Reported on the forums: https://obsproject.com/forum/threads/an-issue-where-an-error-occurs-with-certain-color-formats.176240/

### How Has This Been Tested?

Verified with setup outlined in the forum thread.

### Types of changes

### Checklist:


RytoEX (approved):
Looks okay at a glance.

PR #10919, opened 2024-06-25
cmake: Fix FFmpeg version regex

Penwy:
### Description

This adds `~` to the list of characters accepted as part of a correct version name for ffmpeg in `/cmake/finders/FindFFmpeg.cmake`.

### Motivation and Context

Ffmpeg version `6.1-1build2~22.04` (from https://launchpad.net/~ubuntuhandbook1/+archive/ubuntu/ffmpeg6) failed to be found while setting up cmake, due to the `~` character in the version name. Tildes are frequent enough in linux version names that I feel their inclusion here is warranted.
Legacy cmake hasn't been changed as it correctly finds that ffmpeg version.

### How Has This Been Tested?

Verified that it was still correctly compiling, both with that ffmpeg and the distro one (which doesn't have a tilde), both in legacy and modern cmake.

### Types of changes

### Checklist:

### DIFF
--- a/cmake/finders/FindFFmpeg.cmake
+++ b/cmake/finders/FindFFmpeg.cmake
@@ -277,7 +277,7 @@ endif()
 
 if(EXISTS "${FFmpeg_avutil_INCLUDE_DIR}/libavutil/ffversion.h")
   file(STRINGS "${FFmpeg_avutil_INCLUDE_DIR}/libavutil/ffversion.h" _version_string
-       REGEX "^.*FFMPEG_VERSION[ \t]+\"n?[0-9a-z\\.-]+\"[ \t]*$")
+       REGEX "^.*FFMPEG_VERSION[ \t]+\"n?[0-9a-z\\~.-]+\"[ \t]*$")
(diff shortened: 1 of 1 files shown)


PatTheMav (approved):
Should be fine.

RytoEX:
Could you provide the full line in the header file that this targets?

RytoEX (changes requested):
Nit:
* ffmpeg -> FFmpeg

Penwy:
> Could you provide the full line in the header file that this targets?

This targets line 4 of `ffversion.h`.
On the PPA version I linked : `#define FFMPEG_VERSION "6.1-1build2~22.04"`
On the base distro version : `#define FFMPEG_VERSION "4.4.2-0ubuntu0.22.04.1"`

> ffmpeg -> FFmpeg

Done

RytoEX:
Thanks. I see the issue now. The regex is quite strict and requires a matching pair of double-quotes (`"`) around the target version number string. This is why it won't simply match `6.1-1build2`, which would otherwise be valid. I do not remember why exactly we did this, but I have a vague recollection that we did it based on specific findings in the wild.

PR #10951, opened 2024-07-06
libobs: Add `OBS_ENCODER_CAP_SCALING` cap

tt2468:
### Description
Adds a new encoder cap which tells libobs that rather than scaling video frames in software, the encoder is capable of scaling them via its own (presumably more efficient) means.

An encoder may implement this cap by comparing the VOI of its assigned `video_t` and the results of `obs_encoder_get_width/height()`. If the width/height values differ, then the encoder is being asked by libobs to self-scale, and the resolution in VOI will be the raw frame size, with the `...get_width/height()` being the intended output resolution of the encoder.

It is important to note that GPU rescaling mode will take priority over self-scaling. If GPU rescaling is enabled, the encoder will never be asked to self-scale.

### Motivation and Context
This is useful for discrete hardware encoders, where they might have fixed-function video scaling logic that is highly efficient and fast. Additionally, this feature allows a hardware device which is encoding a full ABR ladder of tracks to be smart and only copy a video frame from GPU -> Host -> Device once for the entire ladder, rather than once for every track.

### How Has This Been Tested?
Tested with a plugin of mine for the MA35D card. Frames received by the encoder are the normal output resolution of OBS, regardless of what the "scaled" size is (this change does not make the UI aware of self-scaling, so this requires custom encoder management).

Ubuntu 22.04

### Types of changes

### Checklist:

### DIFF
--- a/libobs/obs-encoder.h
+++ b/libobs/obs-encoder.h
@@ -37,6 +37,7 @@ typedef struct obs_encoder obs_encoder_t;
 #define OBS_ENCODER_CAP_DYN_BITRATE (1 << 2)
 #define OBS_ENCODER_CAP_INTERNAL (1 << 3)
 #define OBS_ENCODER_CAP_ROI (1 << 4)
+#define OBS_ENCODER_CAP_SCALING (1 << 5)
 
 /** Specifies the encoder type */
 enum obs_encoder_type {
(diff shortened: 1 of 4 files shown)


libobs/obs-encoder.h
```diff
@@ -37,6 +37,7 @@ typedef struct obs_encoder obs_encoder_t;
 #define OBS_ENCODER_CAP_DYN_BITRATE (1 << 2)
 #define OBS_ENCODER_CAP_INTERNAL (1 << 3)
 #define OBS_ENCODER_CAP_ROI (1 << 4)
+#define OBS_ENCODER_CAP_SELF_SCALE (1 << 5)
```
WizardCM, line 40: As this is an encoder cap, we can probably just name it `OBS_ENCODER_CAP_SCALING`. Documentation can cover what it does exactly.

WizardCM (approved):

derrod (approved):
Seems fine to me now.

PR #10952, opened 2024-07-06
libobs: Add `obs_encoder_get_group()`

tt2468:
### Description
Allows getting of the current encoder group of an encoder. Not useful for too much on its own, but allows encoder implementations to know if they are in a group or not.

### Motivation and Context
Needed in my case for enabling some performance optimizations in an encoder plugin. I can rely on encoder groups in my plugin to create a shared scaler object for encoders within the group, rather than always one scaler per encoder always. This gives me a >30% efficiency boost in my scaler utilization.

### How Has This Been Tested?
Compiles on Ubuntu 22.04

### Types of changes

### Checklist:

### DIFF
--- a/libobs/obs-encoder.c
+++ b/libobs/obs-encoder.c
@@ -2046,6 +2046,13 @@ uint32_t obs_encoder_get_roi_increment(const obs_encoder_t *encoder)
 	return encoder->roi_increment;
 }
 
+obs_encoder_group_t *obs_encoder_get_group(obs_encoder_t *encoder)
+{
+	return obs_encoder_valid(encoder, "obs_encoder_get_group")
+		       ? encoder->encoder_group
+		       : NULL;
+}
+
 bool obs_encoder_set_group(obs_encoder_t *encoder, obs_encoder_group_t *group)
 {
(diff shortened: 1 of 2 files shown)


libobs/obs-encoder.c
```diff
@@ -2046,6 +2046,13 @@ uint32_t obs_encoder_get_roi_increment(const obs_encoder_t *encoder)
 	return encoder->roi_increment;
 }
 
+obs_encoder_group_t *obs_encoder_get_group(obs_encoder_t *encoder)
+{
+	return obs_encoder_valid(encoder, "obs_encoder_get_group")
+		       ? encoder->encoder_group
+		       : NULL;
+}
```
palana, line 2054: given the intricacies around the lifetime of groups I'd suggest going with a design that doesn't expose the pointer to code that doesn't own the group, e.g.
```c++
bool obs_encoder_is_grouped_with(obs_encoder_t *encoder, obs_encoder_t *other_encoder)
```

or maybe add a comment to the function along the lines of
> this pointer is only safe to use before encoders have been started or while encoders are active; the pointer might get invalidated as soon as encoders become inactive

PR #10954, opened 2024-07-07
CI: Fix flatpak-builder-lint action

tytan652:
### Description

`jq` does not return zero if the array does not exist but an empty output

### Motivation and Context

Stop the action tab becoming more red

### How Has This Been Tested?

Here (I will temporarily push a commit to trigger errors/warnings )

### Types of changes

### Checklist:

### DIFF
--- a/.github/actions/flatpak-builder-lint/action.yaml
+++ b/.github/actions/flatpak-builder-lint/action.yaml
@@ -51,15 +51,16 @@ runs:
           exit $exit_code
         fi
 
-        for ((i = 0 ; i < $(echo $ret | jq '.warnings | length') ; i++)); do
+        n_warnings=$(echo $ret | jq '.warnings | length')
+        for ((i = 0 ; i < n_warnings ; i++)); do
           warning=$(echo $ret | jq ".warnings[$i]")
           echo "::warning::$warning found in the Flatpak ${{ inputs.artifact }}"
         done
 
         n_errors=$(echo $ret | jq '.errors | length')
-        for ((i = 0 ; i < $n_errors ; i++)); do
+        for ((i = 0 ; i < n_errors ; i++)); do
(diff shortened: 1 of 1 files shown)


tytan652:
Flatpak CI still fails because of a likely upstream issue…

GeorgesStavracas (approved):

PatTheMav:
This PR doesn't make sense to me - if you pass an empty string to `jq` it will also return just nothing with a `0` return code.

`echo $test | jq -r '.some | length'` with `test` being an empty string yields return code `0`, the same happens if `test` is not even defined. Does this require a particular version of `jq`? 

Also if the field is truly optional, then we should use `jq`'s optional syntax:

`jq -r ' .someOptionalArray // ["default", "values"] | length'` though even with a non-existing key `jq` will usually return just the actual value `0` then.

PatTheMav:
Also a lot of the shell script code can be simplified by using `jq`'s built-in functionality:

`echo "<JSON_STRING"> | jq -r '(.warnings // [] | try map("::warning:: " + . + " found in the Flatpak ARTIFACT")) + (.errors // [] | try map("::error:: " + . + " found in the Flatpak ARTIFACT "))' | jq -r '.[]'` will print any warning or error found already pre-composed as a GitHub actions output.

You can pipe this output to `while read -r log_line; do echo "${log_line}"; done <<< "$(<THE JQ CALL)"` to have it automatically printed then.

tytan652:
*Ignores the two previous message as the actual is minimal*

tytan652:
@PatTheMav, this PR is a fix not a re-write. It is out of question that it becomes one.

Also your proposition add complexity, and this is wrong for maintainability.
This action is a simplish bash script, not a `jq` one.

RytoEX (approved):
Seems fine to fix the immediate issue. We'll consider further adjustments in the future as needed.

Issue #10929, opened 2024-06-27
Start streaming freezes on beta.


Penwy:
### Operating System Info

Ubuntu 22.04

### OBS Studio Version

30.2.0-beta4

### OBS Studio Log URL

https://obsproject.com/logs/YuJnDwiIyZlg7um6

### Expected Behavior

Pressing "Start Stream" should start the stream.

### Current Behavior

Pressing "Start Stream" fully freezes obs.

### Steps to Reproduce

1. Open non-flatpak obs on ubuntu 22.04
2. Press "Start Stream"
...

### Anything else we should know?

Flatpak works correctly, only deb and self-compile have the issue.
Bisect pointed to c8950900c36fd6d0bd014eb6fd24da829a63033d as the culprit.
[Backtrace](https://github.com/user-attachments/files/16014293/message.txt) pointed to [this](https://github.com/obsproject/obs-studio/blob/c8950900c36fd6d0bd014eb6fd24da829a63033d/UI/window-basic-main-outputs.cpp#L1143) as the very likely issue.

RytoEX:
Given that it doesn't happen in Flatpak (Qt 6.6.x) but does happen in presumably the PPA version (Qt 6.2.4), this may just be an unfixed Qt Futures bug in older Qt versions. It would be good to determine if there's a version of Qt where this does _not_ happen so we can scope out what a possible fix might be.

RytoEX:
Off-thread testing indicates that this is not a problem with builds on Ubuntu 23.10 with Qt 6.4.2.

RytoEX:
This is possibly one of these:
* https://bugreports.qt.io/browse/QTBUG-105182
* https://bugreports.qt.io/browse/QTBUG-106083

which are the same core bug, and only fixed in 6.2.7, 6.4.1, and 6.5+. Since Ubuntu 22.04 is frozen on 6.2.4, we can't expect it to be fixed on Ubuntu 22.04 with a Qt update. We probably need to implement a code path to Start Streaming that either:
a) avoids this specific behavior
b) does not use QFuture at all

palana:
> This is possibly one of these:
> 
> * https://bugreports.qt.io/browse/QTBUG-105182
> * https://bugreports.qt.io/browse/QTBUG-106083
> 
[...]

PR #10957, opened 2024-07-08
UI: Remove `QFuture` usage

palana:
### Description
Unbreak OBS builds using older Qt versions, since `QFuture`s are broken on older Qt versions, even with the deadlock workaround (see <https://github.com/obsproject/obs-studio/issues/10929>)

### Motivation and Context
This is meant to fix https://github.com/obsproject/obs-studio/issues/10929

Originally the workaround for bouncing continuations between threads seemed sufficient, but there are Qt versions where even creating a ready future and trying to observe it breaks/causes deadlocks

### How Has This Been Tested?
Tested "Start Streaming" works and error conditions (e.g. invalid configs) are still handled appropriately

### Types of changes

### Checklist:


PatTheMav:
In general I'm in favour of using more C++ standard functionality instead of Qt variants if possible (and only rely on Qt variants when interacting with UI code) as this allows us to more easily decouple core functionality from the presentation layer.

palana:
Split out the first commit to make code motion more visible vs actual changes

UI/window-basic-main-outputs.hpp
```diff
@@ -1,15 +1,17 @@
 #pragma once
 
+#include <future>
 #include <memory>
 #include <string>
 
 #include <QFuture>
```
DDRBoxman, line 7: Can this QFuture be removed?

RytoEX (commented):
The parts that I understand look okay. This is mostly shuffling code and changing types. The user who reported #10929 said that this PR fixes that Issue. I would still welcome additional review.

UI/window-basic-main-outputs.hpp
RytoEX: Given that the blame points to https://github.com/obsproject/obs-studio/commit/c8950900c36fd6d0bd014eb6fd24da829a63033d (#10633), it should be safe to remove.

UI/window-basic-main.cpp
```diff
@@ -5170,18 +5172,11 @@ void OBSBasic::ClearSceneData()
...
-		startStreamingFuture.cancelAll();
-		event->ignore();
-		return;
-	} else if (startStreamingFuture.future.isCanceled() &&
-		   !startStreamingFuture.future.isFinished()) {
+	/* Wait for multitrack video stream to start/finish processing in the background */
+	if (setupStreamingGuard.valid() &&
+	    setupStreamingGuard.wait_for(std::chrono::seconds{0}) !=
+		    std::future_status::ready) {
+		QTimer::singleShot(1000, this, &OBSBasic::close);
```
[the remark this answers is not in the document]
palana: `OBSBasic::close` causes another call to `closeEvent`, so this check gets repeated until `setupStreamingGuard` is actually resolved, so the behavior should be fairly similar to the previous code (except the previous code scheduled a continuation that fired immediately after the future was resolved, as opposed to this)

an alternative would be to spawn a "task" on another thread that waits for `setupStreamingGuard` to resolve and then immediately `QMetaObject::invoke`s `OBSBasic::close`; keeping track of that task is somewhat annoying though in that it requires yet another `std::future` or some other synchronization primitive

this could then change to something like
```c++
if (setupStreamingGuard.valid() && setupStreamingGuard.wait_for(std::chrono::seconds{0}) != std::future_status::ready) {
   if (!setupStreamingShutdownMonitor.valid()) {
      auto monitor = std::make_shared<StartMultitrackVideoStreamingGuard>();
      setupStreamingShutdownMonitor = monitor.GetFuture();
      QThreadPool::globalInstance()->start([self = this, setupStreamingGuard, monitor] {
        setupStreamingGuard.wait();
        monitor.reset();
        QMetaObject::invokeMethod(self, &OBSBasic::close);
      });
    }
}
if (setupStreamingShutdownMonitor.valid()) {
  event->ignore();
  return;
}
```

ultimately it all depends on `setupStreamingGuard` only being resolved when the background task is finished (same as `startStreamingFuture` previously, except `setupStreamingGuard` requires more manual work to keep alive long enough unfortunately

UI/window-basic-main-outputs.hpp
palana: removed this and a `QPromise` include elsewhere too

UI/window-basic-main.cpp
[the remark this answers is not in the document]
RytoEX: It's probably fine as-is. I just wanted to note it.

RytoEX (approved):

Chiitoo:
This seems to break legacy CMake configuration:

```
CMake Error at UI/cmake/legacy.cmake:285 (target_sources):
  Cannot find source file:

    qt-helpers.cpp
Call Stack (most recent call first):
  cmake/Modules/ObsHelpers.cmake:473 (include)
  UI/CMakeLists.txt:3 (legacy_check)
```

Is that to be expected and intentional?

Thank you!

palana:
> This seems to break legacy CMake configuration:
> 
> ```
> CMake Error at UI/cmake/legacy.cmake:285 (target_sources):
>   Cannot find source file:
> 
>     qt-helpers.cpp
> Call Stack (most recent call first):
>   cmake/Modules/ObsHelpers.cmake:473 (include)
>   UI/CMakeLists.txt:3 (legacy_check)
> ```
> 
> Is that to be expected and intentional?
> 
> Thank you!

@Chiitoo: not intentional, and not expected necessarily 😅

filed a follow-up PR to fix this at https://github.com/obsproject/obs-studio/pull/10967

RytoEX:
> This seems to break legacy CMake configuration:
> 
> ```
> CMake Error at UI/cmake/legacy.cmake:285 (target_sources):
>   Cannot find source file:
> 
>     qt-helpers.cpp
> Call Stack (most recent call first):
>   cmake/Modules/ObsHelpers.cmake:473 (include)
>   UI/CMakeLists.txt:3 (legacy_check)
> ```
> 
> Is that to be expected and intentional?
> 
> Thank you!

While we'll fix this, please note that legacy CMake _will_ be removed in the near future.

Chiitoo:
> While we'll fix this, please note that legacy CMake _will_ be removed in the near future.

Yep, already getting into the new way of things.

Thanks!

PR #10961, opened 2024-07-09
Replace vendored SIMD Everywhere by prefix/system install

tytan652:
### Description

> [!IMPORTANT]  
> Depends on:
> - https://github.com/obsproject/obs-deps/pull/255

Replace vendored SIMD Everywhere by prefix/system install.

SIMD Everywhere finder needs to be installed alongside libobs CMake
package since its headers depends on it.

macOS AVX intrinsics is included to make sure that AVX defines are
present.

### Motivation and Context

Reduce in-tree vendored deps in OBS Studio repo.

### How Has This Been Tested?

Build in CI on my fork.

### Types of changes

### Checklist:


.github/actions/flatpak-builder-lint/action.yaml
PatTheMav, line 1: I guess this will be removed after a rebase?

tytan652: Yes, I forgot to do it sooner.

cmake/common/helpers_common.cmake
```diff
@@ -426,6 +426,14 @@ function(target_export target)
     DESTINATION "${package_destination}"
     COMPONENT Development
     ${exclude_variant})
+
+  if(target STREQUAL libobs)
```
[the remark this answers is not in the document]
tytan652: The headers include a SIMDE header which make it a dependency of the CMake package and so the finder needs to be installed alongside the package.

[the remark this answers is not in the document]
PatTheMav: Yeah the question was more about which ones? Because SIMDE is an implementation detail of `libobs` internals and should not be exposed publicly. It doesn't make sense to expose it from an API point of view and if we do that seems like a mistake to me.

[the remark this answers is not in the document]
tytan652: `util/sse-intrin.h` used in some provided header.

```
grep -r "sse-intrin.h" /usr/include/obs/. 
/usr/include/obs/./graphics/quat.h:#include "../util/sse-intrin.h"
/usr/include/obs/./graphics/vec3.h:#include "../util/sse-intrin.h"
/usr/include/obs/./graphics/vec4.h:#include "../util/sse-intrin.h"
```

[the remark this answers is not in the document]
PatTheMav: Right and of course we inlined all those functions so we created a tight coupling between API users and the internal data types..

That should have never happened, but I guess that's water under the bridge and can be cleaned up once the current API is not used externally anymore in the future.

cmake/finders/FindSIMDE.cmake
```diff
@@ -0,0 +1,105 @@
...
+
+``SIMDE_INCLUDE_DIR``
+  Directory containing ``simde/simde-common.h``.
+
+#]=======================================================================]
+
+# cmake-format: off
+# cmake-lint: disable=C0103
+# cmake-lint: disable=C0301
+# cmake-format: on
```
RytoEX, line 37: Should no longer need these.

cmake/finders/FindSIMDE.cmake
```diff
@@ -0,0 +1,105 @@
...
+    set_target_properties(SIMDE::SIMDE PROPERTIES INTERFACE_INCLUDE_DIRECTORIES "${SIMDE_INCLUDE_DIR}")
+  endif()
+endif()
+
+include(FeatureSummary)
+set_package_properties(
+  SIMDE
+  PROPERTIES
+    URL "https://github.com/simd-everywhere/simde"
+    DESCRIPTION "Implementations of SIMD instruction sets for systems which don't natively support them. "
```
RytoEX, line 104:
```suggestion
    DESCRIPTION "Implementations of SIMD instruction sets for systems which don't natively support them."
```

RytoEX (changes requested):

cmake/common/helpers_common.cmake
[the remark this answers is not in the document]
RytoEX: Seems like this thread can be resolved as "todo in a future effort".

cmake/finders/FindSIMDE.cmake
tytan652: Removed, since indeed cmake-format was replaced with gersemi.

cmake/finders/FindSIMDE.cmake
tytan652: Fixed

cmake/common/helpers_common.cmake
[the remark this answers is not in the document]
PatTheMav: Ideally it should be fixed before making it an external dependency, because by fixing _that_, we exposed another open wound (namely that it now becomes an explicit public 3rd party dependency) and we just trade one problem (vendored dependency) for another (internal dependency becomes a hard public external dependency) even though there is no architectural reason for that to be the case.

We're doctoring around symptoms again and make a change that is mostly cosmetic instead of fixing the _actual_ issue. But it's a fight I've never won, so I gave up fighting it.

PatTheMav:
I still don't like that we have to ship a finder for the SIMDE package because associated types are used for types like `vec4` and thus exposed globally for the entire `libobs` library but I don't see an easy way to fix this.

@tytan652 if you update/rebase the PR I think this is good to go because even if it's inelegant to ship a finder with the package, you implemented it elegantly.

tytan652:
Seems like I depends a little on #12460 since gersemi add me out of scope formatting which are included in the mentioned PR.

tytan652:
~~Looks like I need to find how to fix including sse2 in Windows ARM64…~~

Add math header.

PatTheMav:
Just a minor nit while we're trying to fix Windows ARM64 - according to their README the correct spelling seems to be "SIMDe" not "SIMDE". So where full capitalisation is not needed we should use that spelling, similar to how we try to use "FFmpeg" as much as possible.

libobs/util/sse-intrin.h
```diff
@@ -25,8 +25,17 @@
...
 #define WIN32_LEAN_AND_MEAN
 #endif
+
+#if defined(_MSC_VER) && defined(__cplusplus)
+#include <cmath>
+#endif
+
+#if defined(__APPLE__) && !defined(__arm64__)
+#include <immintrin.h>
+#endif
```
PatTheMav: We didn't need to include this before for Intel Macs - did it fail to compile without this header? IIRC `immintrin` represents the `avx` class of vector functions.

tytan652:
Yes it did fail to compile without it because of redefinition errors

Le 1 août 2025 21:03:51 GMT+02:00, Patrick Heyer ***@***.***> a écrit :

libobs/util/sse-intrin.h
PatTheMav:
```suggestion
#if defined(__APPLE__)
#include <simd/base.h>
#endif
```
This will include either `immintrin.h` or `arm_neon.h` based on architecture and availability on the platform, which should ensure that SIMDe "sees" the native types once its own header is included.

tytan652: Answered by mail so the comment is out of the conversation.
https://github.com/obsproject/obs-studio/pull/10961#issuecomment-3145557980

tytan652: Applied the change now x86_64 CI has to go green.

PR #10779, opened 2024-06-06
UI: Clean up Settings widget structure and styling

Warchamp7:
### Description
Many of the tabs in Settings right now have margins or sizings that do not match, or have widget heirarchies that vary between them. This PR makes adjustments to many containers and widgets in Settings to make their margins and appearance more consistent.

### Motivation and Context
Want settings to look better and cleaner.

obs64_2m99ot6HaD
obs64_wZwgFFfo9a

obs64_uCkJL8gLDs
obs64_ehPbopyScX

obs64_Gf89PbiYwI
obs64_0dmM4u68qt

obs64_3agSmlNVrO
obs64_GvXvXe9HAx

obs64_xdYiS6qTow
obs64_eppZFD163Y

obs64_5KlKbcoeHP
obs64_LEdgfnUpQb

obs64_BYSkTinwQj
obs64_Sp50el0sfv

### How Has This Been Tested?
Checked through all tabs of Settings, including nested stacked pages.

### Types of changes

- Tweak (non-breaking change to improve existing functionality)

### Checklist:
- [x] My code has been run through [clang-format](https://github.com/obsproject/obs-studio/blob/master/.clang-format).
- [x] I have read the [**contributing** document](https://github.com/obsproject/obs-studio/blob/master/CONTRIBUTING.rst).
- [x] My code is not on the master branch.
- [x] The code has been tested.
- [x] All commit messages are properly formatted and commits squashed where appropriate.
- [x] I have included updates to all appropriate documentation.

PR #10963, opened 2024-07-10
UI: Add field growth policy to form layouts

Warchamp7:
### Description
This adds a defined fieldGrowthPolicy back to the Settings form.

### Motivation and Context
Part of the cleanup in #10779 involved removing unnecessary properties as they were being set to their defaults.

Unfortunately, some of these defaults are not the same across operating systems. The default fieldGrowthPolicy on Windows is AllNonFixedFieldsGrow, so this removal was fine there. However on mac the default policy is FieldsStayAtSizeHint which causes the below issue:

image
image

Re-adding this property should fix this.

### How Has This Been Tested?
Built on Windows and observed everything is still fine. This will need to be checked on mac.

### Types of changes

### Checklist:

### DIFF
--- a/UI/forms/OBSBasicSettings.ui
+++ b/UI/forms/OBSBasicSettings.ui
@@ -219,6 +219,9 @@
                    <string>Basic.Settings.General</string>
                   </property>
                   <layout class="QFormLayout" name="formLayout_32">
+                   <property name="fieldGrowthPolicy">
+                    <enum>QFormLayout::AllNonFixedFieldsGrow</enum>
+                   </property>
                    <property name="labelAlignment">
                     <set>Qt::AlignRight|Qt::AlignTrailing|Qt::AlignVCenter</set>
                    </property>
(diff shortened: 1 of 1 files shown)


UI/forms/OBSBasicSettings.ui
```diff
@@ -2910,7 +2964,7 @@
                               </sizepolicy>
                              </property>
                              <property name="currentIndex">
-                              <number>1</number>
+                              <number>0</number>
```
RytoEX, line 2967: Was this a Qt Designer change or something else?

jcm93:
Can confirm this resolves layout issues on macOS present in 30.2 beta 2 through RC 1.

UI/forms/OBSBasicSettings.ui
Warchamp7: This was me trying to reset all stacked widgets to their default tab.

As far as I can tell that widget _should_ be on index 0 (The radio group) by default, so this is also a regression fix.

RytoEX (approved):

PR #10966, opened 2024-07-11
obs-webrtc: Enable PacingHandler

Sean-Der:
### Description
Adds packet pacing to WebRTC output.

### Motivation and Context
Pacing reduces the amount of packet loss for users. Spreading bursty content over a larger window.

### How Has This Been Tested?
This isn't a problem that I have personally. I also haven't set up network simulations that exhibit the behavior either.

WHIP output users have reported they have a better experience with this. I will have them report their results in the thread.

### Types of changes

### Checklist:

### DIFF
--- a/plugins/obs-webrtc/whip-output.cpp
+++ b/plugins/obs-webrtc/whip-output.cpp
@@ -1,6 +1,8 @@
 #include "whip-output.h"
 #include "whip-utils.h"
 
+#include <obs.hpp>
+
 /*
  * Sets the maximum size for a video fragment. Effective range is
  * 576-1470, with a lower value equating to more packets created,
@@ -147,6 +149,9 @@ void WHIPOutput::ConfigureVideoTrack(std::string media_stream_id, std::string cn
 	if (!encoder)
 		return;
 
+	OBSDataAutoRelease settings = obs_encoder_get_settings(encoder);
+	auto video_bitrate = (int)obs_data_get_int(settings, "bitrate");
+
 	const char *codec = obs_encoder_get_codec(encoder);
 	if (strcmp("h264", codec) == 0) {
 		video_description.addH264Codec(video_payload_type);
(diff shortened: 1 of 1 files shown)


DDRBoxman (approved):

ProudlyTM:
Me and a friend of mine both had unexplainable artifacts and/or freezes (depending on the Encoder settings) when streaming to WHIP, but that patch seems to have gotten rid of the issues for both of us!

I can only say a big thanks @Sean-Der for the fix!

RytoEX:
How severe/urgent/widespread is this?

Sean-Der:
I don't believe it is very wide spread. I implemented it [3 months ago](https://github.com/paullouisageneau/libdatachannel/commit/f61fe88e698fdcf091464131e27b8a11b3a2e4c6) because I hoped it would help users. In those 3 months I have gotten two reports that it helped.

I think it would be good to wait for 31.

plugins/obs-webrtc/whip-output.h
```diff
@@ -1,6 +1,7 @@
 #pragma once
 
 #include <obs-module.h>
+#include <obs.hpp>
```
norihiro, line 4: Just a nitpick.
I think it is better to add this line to `whip-output.cpp` instead of `whip-output.h` because the header file does not rely on `obs.hpp`.

Sean-Der: Fixed! Thank you for the review @norihiro

leodog256:
It fixed long freeze. But there is still skip of a frame every like x seconds when there is not that much movement on the screen (Only in some games that i tested).

tt2468:
Would it be possible for libdatachannel to rolling average the incoming packet data bitrate and pace off of that? That's what libsrt does, and it seems to work quite well. It means that different forms of rate control can be used.

Sean-Der:
@tt2468 Can do! I will make the window size user adjustable

RytoEX:
clang-format is failing here for some reason.

Sean-Der:
@RytoEX Fixed, sorry about that :(

Should be good now

Sean-Der:
This patch has quite a following in the realtime broadcasting discord now :) anytime someone has an issue people recommend it. Seems to help a lot!

plugins/obs-webrtc/whip-output.cpp
```diff
@@ -1,3 +1,4 @@
+#include <obs.hpp>
 #include "whip-output.h"
 #include "whip-utils.h"
```
RytoEX, line 3: I think our preferred header order is:
```cpp
#include "whip-output.h"
#include "whip-utils.h"

#include <obs.hpp>
```

Unless there is a specific reason they must be in this order.

RytoEX (changes requested):

plugins/obs-webrtc/whip-output.cpp
Sean-Der: @RytoEX fixed! Could I get another review, thank you :)

RytoEX: Thanks.

RytoEX (approved):

PR #10974, opened 2024-07-12
UI: Create OBS Studio folder for default recording location

Warchamp7:
### Description
This updates the default directory for recordings to be a newly created `OBS Studio` folder inside the default directory
`%UserProfile%/Videos/OBS Studio` on Windows
`/home/<User>/OBS Studio` on Linux (?)
`/Users/<User>/Movies/OBS Studio` on macOS (?)

### Motivation and Context
Better organization for users

### How Has This Been Tested?
Tested on Windows.

Ensured the directory is created during initial startup and properly set for new Profiles.

### Types of changes

### Checklist:


UI/obs-app.cpp
```diff
@@ -1180,6 +1189,8 @@ void OBSApp::AppInit()
 
 	if (!MakeUserDirs())
 		throw "Failed to create required user directories";
+	if (!MakeRecordingDir())
+		throw "Failed to create recording directory";
```
notr1ch, line 1193: This should be unnecessary as the directory is created on demand when a recording is started. OBS should not create an empty folder in the user's video folder just from being run.

Warchamp7: 1. It does not create the folder on demand in my testing
image

2. I disagree and believe the folder should be initialized when OBS first runs like any app other storage location

UI/platform-osx.mm
jcm93, line 1: Can't comment on the directory creation logic, but I would suggest the following for this function:

```objc
string GetDefaultVideoSavePath()
{
    NSFileManager *fm = [NSFileManager defaultManager];
    NSURL *url = [fm URLForDirectory:NSMoviesDirectory inDomain:NSUserDomainMask appropriateForURL:nil create:true
                               error:nil];
    if (url)
        url = [url URLByAppendingPathComponent:@"OBS Studio" isDirectory:true];

    if (!url)
        return getenv("HOME");

    return url.path.fileSystemRepresentation;
}
```

Cleaner to keep the fallback as is, and I don't see a compelling reason to change anything else in the function besides the URL we're trying for.

UI/obs-app.cpp
notr1ch: This isn't an app storage location though, this is the user's data folders. Well-designed apps should respect a user's folder layout and not create empty files and directories.

UI/platform-osx.mm
PatTheMav: No need to use two separate conditional branches though:

```ObjC++
string GetDefaultVideoSavePath() const
{
    NSFileManager *defaultManager = [NSFileManager defaultManager];
    NSURL *moviesDirectory = [defaultManager URLForDirectory:NSMoviesDirectory inDomain:NSUserDomainMask
                                           appropriateForURL:nil
                                                      create:YES
                                                       error:nil];

    if (moviesDirectory) {
        const string videoSavePath =
            [moviesDirectory URLByAppendingPathComponent:@"OBS Studio" isDirectory:YES].path.fileSystemRepresentation;

        return videoSavePath;
    } else {
        return getenv("HOME");
    }
}
```

As the string will not be modified, the return type as well as the function could be marked as `const` (I don't know if copy elision will be used, otherwise the copy constructor will be used to copy the returned string into the target variable).

UI/platform-osx.mm
```diff
@@ -62,10 +62,13 @@ string GetDefaultVideoSavePath()
     NSURL *url = [fm URLForDirectory:NSMoviesDirectory inDomain:NSUserDomainMask appropriateForURL:nil create:true
                                error:nil];
 
-    if (!url)
-        return getenv("HOME");
+    if (!url) {
+        NSString *path = [NSString pathWithComponents:@[getenv("HOME"), @"/OBS Studio"]];
```
PatTheMav, line 66:
```suggestion
        NSString *path = [NSString pathWithComponents:@[@(getenv("HOME")), @"OBS-Studio"]];
```

The error is "collection element of type 'char *' is not an Objective-C object" - `pathWithComponents` takes an `NSArray`, but that class requires that the type of values in the array itself are Foundation classes. `getenv("HOME")` returns a `char *`, which is not a Foundation type, so it needs to be boxed with another `@()` to convert it into an `NSString`. The danger about this is that the program will crash if `getenv` returns a NULL pointer.

Also `pathWithComponents` already concatenates with a path separator between each component, so the forward slash in the second string is superfluous.

UI/platform-osx.mm
```diff
@@ -62,10 +62,13 @@ string GetDefaultVideoSavePath()
...
+        return path;
```
PatTheMav, line 67: You can't simply return an `NSString` pointer (which is reference counted by the ObjC runtime) as a `std::string` object (which either uses C++ semantics and memory management).

Instead you need to take a `char` pointer reference of the `NSString` and use that to initialise a `std::string` - this is usually done by using the `UTF8String` property of the `NSString` object.

The compiler will insert the required code to ensure this all works (the `char` pointer will become invalid once the `NSString` object is automatically deallocated by the ObjC runtime once the function goes out of scope) so it will implicitly create a `std::string` with a copy of that `char` array before returning the string and then _that_ string will be used in the copy constructor of the target string (so you will have the entire string being copied around in memory at least twice).

UI/obs-app.cpp
Warchamp7: Do you have some sort of guidance or documentation from Microsoft about that?

I have multiple applications on my system that have created folders for themselves in the Videos library location

UI/platform-osx.mm
Warchamp7: Is there a reason for the else branch here or could it just be
```
string GetDefaultVideoSavePath() const
{
    NSFileManager *defaultManager = [NSFileManager defaultManager];
    NSURL *moviesDirectory = [defaultManager URLForDirectory:NSMoviesDirectory inDomain:NSUserDomainMask
                                           appropriateForURL:nil
                                                      create:YES
                                                       error:nil];

    if (moviesDirectory) {
        const string videoSavePath =
            [moviesDirectory URLByAppendingPathComponent:@"OBS Studio" isDirectory:YES].path.fileSystemRepresentation;

        return videoSavePath;
    }
    
    return getenv("HOME");
}```

Comment (PatTheMav) [reply UI/platform-osx.mm:1]: That is indeed also fine, comes down to personal preference.

File: UI/obs-app.cpp (hunk 1 above)
Comment (notr1ch) [reply UI/obs-app.cpp:1193]: It's just good behavior, just because other apps are shitty doesn't mean we should also be shitty. There's [plenty of evidence of of users who desire such behavior](https://www.google.com/search?q=prevent+apps+from+creating+folders+in+my+documents) from their apps.

Comment (PatTheMav) [reply UI/obs-app.cpp:1193]: One's shitty behaviour is another's expected behaviour though. And by that token I'd rather have an application create files in its own managed subdirectory than mess up the root of a data storage location.

I do agree that the directory should be created upon first use though. I don't remember Shadowplay creating directories for recordings or highlights until I actually triggered either.

Comment (Fenrirthviti) [reply UI/obs-app.cpp:1193]: I concur with first use, as a user who does not enjoy all the random empty folders I have no intention of using.

File: UI/platform-osx.mm [hunk 4]
```diff
@@ -58,14 +58,20 @@ void CheckIfAlreadyRunning(bool &already_running)
...
+    NSURL *moviesDirectory = [defaultManager URLForDirectory:NSMoviesDirectory inDomain:NSUserDomainMask
+                                           appropriateForURL:nil
+                                                      create:YES
+                                                       error:nil];
 
-    if (!url)
-        return getenv("HOME");
+    if (moviesDirectory) {
+        const string videoSavePath =
+            [moviesDirectory URLByAppendingPathComponent:@"OBS-Studio" isDirectory:YES].path.fileSystemRepresentation;
```
jcm93, line 69:
```suggestion
            [moviesDirectory URLByAppendingPathComponent:@"OBS Studio" isDirectory:YES].path.fileSystemRepresentation;
```
Shouldn't this just be "OBS Studio"? Spaces are fine to pass as a component to this function, and I'm not aware of any issues with spaces in folder names for OBS.

Warchamp7:
Better approach to the underlying problem in #12395

PR #10979, opened 2024-07-13
UI,cmake: Get default video path on Linux with GLib

tytan652:
### Description

Qt does not rely on xdg-user-dirs behavior to get the video directory. So GLib API that relies on this behavior is used.

Note that GLib is already used by Qt itself for its event loop on Linux, so adding GLib as a dependency of the UI is not an issue.

Closes #3746

### Motivation and Context

Get the XDG video directory through a widespread implementation since Qt does not try to rely on xdg-user-dirs.

### How Has This Been Tested?

- Removed `RecFilePath` and `FFFilePath` from my profile config
- Ran OBS and change a random profile setting and save it
- Check the profile config and see that the value are the XDG video directory.

### Types of changes

### Checklist:


kkartaltepe:
If we were going to do this, not adding 224 lines of cmake is a strong benefit of #3746

tytan652:
> If we were going to do this, not adding 224 lines of cmake is a strong benefit of #3746

Half of it is because of the legacy path (duplication), also a finder does not usually need changes post merge (low maintenance burden).

PatTheMav (approved):
Can't say much about the correctness of using `glib` but changes look fine to me.

tytan652:
Custom Glib2 was dropped for the one provided by ECM.

frontend/utility/platform-x11.cpp
```diff
@@ -220,7 +222,7 @@ bool GetDataFilePath(const char *data, string &output)
 
 string GetDefaultVideoSavePath()
 {
-	return string(getenv("HOME"));
+	return string(g_get_user_special_dir(G_USER_DIRECTORY_VIDEOS));
```
PatTheMav: Are there scenarios where this might return a `nullptr`? Because the `std::string` will throw an exception if passed such a pointer.

tytan652: On Unix the function (with a a valid `G_USER_` enum) is unlikely to return NULL since it has a fallback value in its implementation if XDG directories are not correctly set.

But if really desired I can add a fallback to $HOME if null.

PatTheMav: I'll leave that to you. Personally I'd rather have this code be robust to handle the possibility of the returned `char *` to be `NULL` and avoid a crash, particularly because the old code already has that issue and this is a good chance to fix that with a single commit.

Annoyingly the code calling this function looks like this:

```C++
config_set_default_string(activeConfiguration, "SimpleOutput", "FilePath", GetDefaultVideoSavePath().c_str());
```
So this function _needs_ to return some `std::string` instance to ensure that `c_str()` will not fail as well.

tytan652: Added `getenv("HOME")` if Glib somehow fails to return a path.

Issue #10676, opened 2024-05-15
Media source always played if close file when inactive is checked


Xarkam:
### Operating System Info

Windows 11

### OBS Studio Version

30.1.2

### OBS Studio Log URL

https://obsproject.com/logs/rOSyWMH7fRWelcVT

### Expected Behavior

The media source should not be played if the "close file when inactive" option is checked

### Current Behavior

The media source is played even though the option has been checked so that obs does not do so.

I noticed that the progress bars for playing a media source below the preview moved even though I checked the option not to open the file if it is not visible. (the countdown also turns indicating the time already read on the media)

See screenshot below.

### Steps to Reproduce

1. Create group in Sources
2. Add media source into group
3. Hide group & media source
4. Close Obs
5. Open Obs and see media is read as muted
6.
...

### Anything else we should know?

![image](https://github.com/obsproject/obs-studio/assets/331816/6704fea9-8537-4010-a8aa-1d9b265ea03c)

Sugam-Chand-M:
Hi @Xarkam 
Can please provide a video recording of the issue.

Warchamp7:
I am not able to reproduce this issue on OBS 32. Please open a new issue if this is still happening.

PR #10980, opened 2024-07-13
UI: Media source always played if close file when inactive is checked

Sugam-Chand-M:
### Description

Even when the "Close when Inactive" property is checked  and the source is inactive, the media was being played. 
Issue - [ #10676 ]

### Motivation and Context

This PR aims at resolving the issue by pausing the media play when the source is inactive and the "Close when Inactive" property is checked.
Issue - [ #10676 ]

### How Has This Been Tested?

Tested manually in Windows 11.
Added a video drive link - [Video Link](https://drive.google.com/file/d/1vTx85_dKfeiJ30B25Tx9OsGrbcPHDwO6/view?usp=sharing) of the fix working

### Types of changes

### Checklist:

### DIFF
--- a/UI/media-controls.cpp
+++ b/UI/media-controls.cpp
@@ -267,9 +267,19 @@ void MediaControls::RefreshControls()
 	uint32_t flags = 0;
 	const char *id = nullptr;
 
+	obs_data_t *settings = obs_source_get_settings(source);
+	bool sourceIsActive = obs_source_active(source);
+	bool closeWhenInactive =
+		obs_data_get_bool(settings, "close_when_inactive");
+
 	if (source) {
 		flags = obs_source_get_output_flags(source);
 		id = obs_source_get_unversioned_id(source);
+	}
+
+	if (!sourceIsActive && closeWhenInactive) {
+		SetPausedState();
+		return;
 	}
 
 	if (!source || !(flags & OBS_SOURCE_CONTROLLABLE_MEDIA)) {


SuslikV:
I think, the fix is incorrect. The option meaning changed in this PR. This may lead to confusion in future when new programmers will read the code.

The option to "close file" designed to free the file handler. Thus, the file can be in use and not exclusively by OBS.
Maybe the tip to the option is not clear enough and could be reworded (don't know is it really worth so much attention or not):
https://github.com/obsproject/obs-studio/blob/7912d06b65042501837b5fc669e28cfa0412e020/plugins/obs-ffmpeg/data/locale/en-US.ini#L81-L82
here the phrase "This allows the file to be changed..." has wider meaning: the file can be renamed, old file can be deleted and new file can be created with the same name, whole content of the media file can be modified (edited) when this option is checked.

In my opinion, the "Pause of playback" and "File close" is completely different actions. Current OBS logic states that media continues its playback even when the media source in not in the current scene ("not active" in terms of OBS - not shown in the stream/recording).

Yes, logically the "close of the file" is exclusive to the "not active" state of the source, but defaults of OBS is to "always playback" the source (and never stop until media ends or loop is forced).

The proposed changes can be considered as a fix, but previously such huge changes in logic used kind of versioned variants of the sources (to keep behavior unchanged for scenes that already created by user). Maybe additional checkbox to force the pause when media is not active - is a better option here? For example, such option exist for VLC Source of OBS.

Sugam-Chand-M:
Hi @SuslikV 

Sorry for the very delayed response

"Close the file when inactive" option is for not displaying the sources when it is inactive and not to be deleted from the source.
When the product was designed and coded, I believe they have considered the scenario where if the user wants to temporarily disable the source which shouldn't be visible so they have included this option.

For the questions,
1) Ideally when the source is inactive, the media shouldn't not be played or should be paused and played from the position when it becomes active again
2) It can be either, active or inactive, its like inactive => paused but vice versa is not true
3) Same as the answer for 2nd

Warchamp7:

In my testing I was also no longer able to reproduce the linked issue, so this may have been fixed already.

PR #10982, opened 2024-07-13
cmake: Set CFBundleDisplayName for camera extension

gxalpha:
### Description

macOS 15 shows this in the settings (or at least I think it's only from 15, haven't seen it earlier).
This is in System Settings -> Common [or whatever it's named in English] -> Login Objects and Extensions [or similar] -> Extensions.

Before/After (German, but the important part should be obvious):
<p>
Bildschirmfoto 2024-07-13 um 23 55 24

Bildschirmfoto 2024-07-14 um 00 20 54
</p>

### Motivation and Context

Noticed this randomly when browsing through the login items settings.
If we don't set the display name macOS will fall back on the identifier which isn't great for users.

### How Has This Been Tested?

macOS 15, installed vcam and saw the new name in system settings.

### Types of changes

### Checklist:


PatTheMav (approved):
Should be fine.

RytoEX (approved):
LGTM

PR #10992, opened 2024-07-16
UI: Don't create default desktop audio source on macOS 13+

jcm93:
### Description

Removes the default device-based "Desktop Audio" source created on first run if the user is on macOS 13+.

### Motivation and Context

Since adding macOS Screen Capture in macOS 13, the recommended way to capture audio on macOS is via that source (or alternately, the macOS Audio Capture source for strictly audio-only sources). The device-based audio source idiom is only relevant for older versions of macOS, or for some specialized workflows that require advanced audio routing via programs like Blackhole, VB-Cable, et. al.

The creation of this device-based source by default on new macOS 13+ systems very often leads to new user confusion. It doesn't provide any audio, and users often think that they need to configure it to get desktop audio, which can actually point them *away* from the newer, easier-to-use audio capture methods.

On top of being nonfunctional, the source also just takes up space in the audio mixer until the user disables it.

### How Has This Been Tested?

Tested on macOS 15. On a fresh startup of OBS, the Desktop Audio device in Settings is now set to "Disabled," with only the Mic/Aux source configured in the audio mixer.

### Types of changes

### Checklist:

### DIFF
--- a/UI/platform-osx.mm
+++ b/UI/platform-osx.mm
@@ -133,6 +133,15 @@ void SetAlwaysOnTop(QWidget *window, bool enable)
     window->show();
 }
 
+bool shouldCreateDefaultAudioSource(void)
+{
+    if (@available(macOS 13, *)) {
+        return false;
+    } else {
+        return true;
+    }
+}
+
 bool SetDisplayAffinitySupported(void)
 {
     // Not implemented yet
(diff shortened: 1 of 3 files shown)


Lain-B (approved):
I suppose this is fine. You okay with it @Warchamp7?

Might be good to change the commit email though.

GeorgesStavracas (approved):
UI / UX approved

PR #5455, opened 2021-10-27
UI: Relatively center multiple scene items

cg2121:
### Description
Before the scene items would be centered at the same spot
when centering them, now they are moved relatively to each
other.

Before:

https://user-images.githubusercontent.com/19962531/139073933-2be27287-04b2-4b65-a220-b11e72f53f27.mp4

After:

https://user-images.githubusercontent.com/19962531/139073975-0b132e82-f585-45dd-a595-ed9d38f0d5b5.mp4

### Motivation and Context
Easier to align multiple scene items.

### How Has This Been Tested?
Selected multiple scene items and centered them vertically, horizontally and to scene.

### Types of changes
- New feature (non-breaking change which adds functionality)

### Checklist:
- [x] My code has been run through [clang-format](https://github.com/obsproject/obs-studio/blob/master/.clang-format).
- [x] I have read the [**contributing** document](https://github.com/obsproject/obs-studio/blob/master/CONTRIBUTING.rst).
- [x] My code is not on the master branch.
- [x] The code has been tested.
- [x] All commit messages are properly formatted and commits squashed where appropriate.
- [x] I have included updates to all appropriate documentation.

PR #10997, opened 2024-07-16
Add support for compiling for Windows on ARM

thirumalai-qcom:
Introduces Windows on ARM support to the OBS-studio application. This includes modifications that are guarded with ARM-specific macros to ensure compatibility without affecting the existing codebase.

### Description
This pull request introduces several code refactoring changes to enhance WOA support and resolve conflicts in the OBS-Studio application. The key changes include:

- **w32-pthreads:** Added additional definitions of ARM macros `_M_ARM` and `_M_ARM64`.
- **libobs:** updated `obs-win-crash-handler.c` to handle instruction pointers and stack traces, improving crash reporting on ARM64 systems .
- **obs-filters:** Renaming `S_THRESHOLD` to `S_FILTER_THRESHOLD` to avoid a conflict with existing macro with a numeric value.
- **obs-outputs:** Replaced `__tzcnt_u32()` with `_CountTrailingZeros()` for ARM.
- **Frontend:** ensured `WIN32_LEAN_AND_MEAN` is defined conditionally if not defined.

### Motivation and Context

The primary motivation for these changes is to expand compatibility with ARM architecture (windows) by addressing conflicts with existing macros and functions. By introducing ARM-specific macros and guarding ARM-specific code, we ensure that the application runs smoothly without affecting the existing codebase.

- **Additional ARM definitions in deps/w32-pthreads/context.h:**
    - Update w32-pthreads/context.h to include definitions for _M_ARM and _M_ARM64 in addition to existing ARM definitions.

- **Libobs changes:**
  - Adjust obs-win-crash-handler.c to handle instruction pointers & stack traces for crash reporting.

- **Obs-filters changes:**
   - Renaming `S_THRESHOLD` macro to `S_FILTER_THRESHOLD` to avoid conflicts with existing definition

- **Obs-outputs: Use _CountTrailingZeros() in ctz32**
    - Replaced `tzcnt_u32` with `_CountTrailingZeros` to ensure compatibility with WOA machines.

- **Frontend Changes:**
  - Ensured WIN32_LEAN_AND_MEAN is included conditionally in OBSBasicFilters.cpp, OBSBasicInteraction.cpp, and OBSBasicProperties.cpp.

### How Has This Been Tested?
The changes have been tested by building OBS Studio on an ARM machine (X Elite machine, Win11). All builds were successful, Basic functionality tests confirmed that the OBS Studio application with the modifications is functional.

### Types of changes

### Checklist:

### DIFF
--- a/deps/w32-pthreads/context.h
+++ b/deps/w32-pthreads/context.h
@@ -63,7 +63,7 @@
 #define PTW32_PROGCTR(Context)  ((Context).Rip)
 #endif
 
-#if defined(_ARM_) || defined(ARM)
+#if defined(_ARM_) || defined(ARM) || defined(_M_ARM) || defined(_M_ARM64)
 #define PTW32_PROGCTR(Context)  ((Context).Pc)
 #endif
 
--- a/plugins/obs-filters/compressor-filter.c
+++ b/plugins/obs-filters/compressor-filter.c
@@ -27,7 +27,7 @@
 /* clang-format off */
 
 #define S_RATIO                         "ratio"
-#define S_THRESHOLD                     "threshold"
+#define S_FILTER_THRESHOLD              "threshold"
 #define S_ATTACK_TIME                   "attack_time"
 #define S_RELEASE_TIME                  "release_time"
 #define S_OUTPUT_GAIN                   "output_gain"
@@ -190,7 +190,7 @@ static void compressor_update(void *data, obs_data_t *s)
 	const char *sidechain_name = obs_data_get_string(s, S_SIDECHAIN_SOURCE);
 
 	cd->ratio = (float)obs_data_get_double(s, S_RATIO);
-	cd->threshold = (float)obs_data_get_double(s, S_THRESHOLD);
+	cd->threshold = (float)obs_data_get_double(s, S_FILTER_THRESHOLD);
 	cd->attack_gain = gain_coefficient(sample_rate, attack_time_ms / MS_IN_S_F);
 	cd->release_gain = gain_coefficient(sample_rate, release_time_ms / MS_IN_S_F);
 	cd->output_gain = db_to_mul(output_gain_db);
@@ -438,7 +438,7 @@ static struct obs_audio_data *compressor_filter_audio(void *data, struct obs_aud
 static void compressor_defaults(obs_data_t *s)
 {
 	obs_data_set_default_double(s, S_RATIO, 10.0f);
-	obs_data_set_default_double(s, S_THRESHOLD, -18.0f);
+	obs_data_set_default_double(s, S_FILTER_THRESHOLD, -18.0f);
 	obs_data_set_default_int(s, S_ATTACK_TIME, 6);
 	obs_data_set_default_int(s, S_RELEASE_TIME, 60);
 	obs_data_set_default_double(s, S_OUTPUT_GAIN, 0.0f);
@@ -477,8 +477,8 @@ static obs_properties_t *compressor_properties(void *data)
 
 	p = obs_properties_add_float_slider(props, S_RATIO, TEXT_RATIO, MIN_RATIO, MAX_RATIO, 0.5);
 	obs_property_float_set_suffix(p, ":1");
-	p = obs_properties_add_float_slider(props, S_THRESHOLD, TEXT_THRESHOLD, MIN_THRESHOLD_DB, MAX_THRESHOLD_DB,
-					    0.1);
+	p = obs_properties_add_float_slider(props, S_FILTER_THRESHOLD, TEXT_THRESHOLD, MIN_THRESHOLD_DB,
+					    MAX_THRESHOLD_DB, 0.1);
 	obs_property_float_set_suffix(p, " dB");
 	p = obs_properties_add_int_slider(props, S_ATTACK_TIME, TEXT_ATTACK_TIME, MIN_ATK_RLS_MS, MAX_ATK_MS, 1);
 	obs_property_int_set_suffix(p, " ms");
--- a/frontend/dialogs/OBSBasicFilters.cpp
+++ b/frontend/dialogs/OBSBasicFilters.cpp
@@ -30,7 +30,9 @@
 #include <QLineEdit>
 
 #ifdef _WIN32
+#ifndef WIN32_LEAN_AND_MEAN
 #define WIN32_LEAN_AND_MEAN 1
+#endif
 #include <Windows.h>
 #endif
 
--- a/frontend/dialogs/OBSBasicInteraction.cpp
+++ b/frontend/dialogs/OBSBasicInteraction.cpp
@@ -27,7 +27,9 @@
 #include <QInputEvent>
 
 #ifdef _WIN32
+#ifndef WIN32_LEAN_AND_MEAN
 #define WIN32_LEAN_AND_MEAN 1
+#endif
 #include <Windows.h>
 #endif
 
--- a/frontend/dialogs/OBSBasicProperties.cpp
+++ b/frontend/dialogs/OBSBasicProperties.cpp
@@ -27,7 +27,9 @@
 #include <QPushButton>
 
 #ifdef _WIN32
+#ifndef WIN32_LEAN_AND_MEAN
 #define WIN32_LEAN_AND_MEAN 1
+#endif
 #include <Windows.h>
 #endif
 
--- a/libobs/obs-win-crash-handler.c
+++ b/libobs/obs-win-crash-handler.c
@@ -138,7 +138,13 @@ static inline bool get_dbghelp_imports(struct exception_handler_data *data)
 
 static inline void init_instruction_data(struct stack_trace *trace)
 {
-#ifdef _WIN64
+#if defined(_M_ARM64)
+	trace->instruction_ptr = trace->context.Pc;
+	trace->frame.AddrPC.Offset = trace->instruction_ptr;
+	trace->frame.AddrFrame.Offset = trace->context.Fp;
+	trace->frame.AddrStack.Offset = trace->context.Sp;
+	trace->image_type = IMAGE_FILE_MACHINE_ARM64;
+#elif defined(_WIN64)
 	trace->instruction_ptr = trace->context.Rip;
 	trace->frame.AddrPC.Offset = trace->instruction_ptr;
 	trace->frame.AddrFrame.Offset = trace->context.Rbp;
--- a/plugins/obs-filters/expander-filter.c
+++ b/plugins/obs-filters/expander-filter.c
@@ -27,7 +27,7 @@
 /* clang-format off */
 
 #define S_RATIO                         "ratio"
-#define S_THRESHOLD                     "threshold"
+#define S_FILTER_THRESHOLD              "threshold"
 #define S_ATTACK_TIME                   "attack_time"
 #define S_RELEASE_TIME                  "release_time"
 #define S_OUTPUT_GAIN                   "output_gain"
@@ -158,7 +158,7 @@ static void expander_defaults(obs_data_t *s)
 		is_expander_preset = false;
 	obs_data_set_default_string(s, S_PRESETS, is_expander_preset ? "expander" : "gate");
 	obs_data_set_default_double(s, S_RATIO, is_expander_preset ? 2.0 : 10.0);
-	obs_data_set_default_double(s, S_THRESHOLD, -40.0f);
+	obs_data_set_default_double(s, S_FILTER_THRESHOLD, -40.0f);
 	obs_data_set_default_int(s, S_ATTACK_TIME, 10);
 	obs_data_set_default_int(s, S_RELEASE_TIME, is_expander_preset ? 50 : 125);
 	obs_data_set_default_double(s, S_OUTPUT_GAIN, 0.0);
@@ -168,7 +168,7 @@ static void expander_defaults(obs_data_t *s)
 static void upward_compressor_defaults(obs_data_t *s)
 {
 	obs_data_set_default_double(s, S_RATIO, 0.5);
-	obs_data_set_default_double(s, S_THRESHOLD, -20.0f);
+	obs_data_set_default_double(s, S_FILTER_THRESHOLD, -20.0f);
 	obs_data_set_default_int(s, S_ATTACK_TIME, 10);
 	obs_data_set_default_int(s, S_RELEASE_TIME, 50);
 	obs_data_set_default_double(s, S_OUTPUT_GAIN, 0.0);
@@ -204,7 +204,7 @@ static void expander_update(void *data, obs_data_t *s)
 
 	cd->ratio = (float)obs_data_get_double(s, S_RATIO);
 
-	cd->threshold = (float)obs_data_get_double(s, S_THRESHOLD);
+	cd->threshold = (float)obs_data_get_double(s, S_FILTER_THRESHOLD);
 	cd->attack_gain = gain_coefficient(sample_rate, attack_time_ms / MS_IN_S_F);
 	cd->release_gain = gain_coefficient(sample_rate, release_time_ms / MS_IN_S_F);
 	cd->output_gain = db_to_mul(output_gain_db);
@@ -445,8 +445,8 @@ static obs_properties_t *expander_properties(void *data)
 	p = obs_properties_add_float_slider(props, S_RATIO, TEXT_RATIO, !cd->is_upwcomp ? MIN_RATIO : MIN_RATIO_UPW,
 					    !cd->is_upwcomp ? MAX_RATIO : MAX_RATIO_UPW, 0.1);
 	obs_property_float_set_suffix(p, ":1");
-	p = obs_properties_add_float_slider(props, S_THRESHOLD, TEXT_THRESHOLD, MIN_THRESHOLD_DB, MAX_THRESHOLD_DB,
-					    0.1);
+	p = obs_properties_add_float_slider(props, S_FILTER_THRESHOLD, TEXT_THRESHOLD, MIN_THRESHOLD_DB,
+					    MAX_THRESHOLD_DB, 0.1);
 	obs_property_float_set_suffix(p, " dB");
 	p = obs_properties_add_int_slider(props, S_ATTACK_TIME, TEXT_ATTACK_TIME, MIN_ATK_RLS_MS, MAX_ATK_MS, 1);
 	obs_property_int_set_suffix(p, " ms");
--- a/plugins/obs-filters/limiter-filter.c
+++ b/plugins/obs-filters/limiter-filter.c
@@ -24,7 +24,7 @@
 
 /* clang-format off */
 
-#define S_THRESHOLD                     "threshold"
+#define S_FILTER_THRESHOLD              "threshold"
 #define S_RELEASE_TIME                  "release_time"
 
 #define MT_ obs_module_text
@@ -90,7 +90,7 @@ static void limiter_update(void *data, obs_data_t *s)
 	const float release_time_ms = (float)obs_data_get_int(s, S_RELEASE_TIME);
 	const float output_gain_db = 0;
 
-	cd->threshold = (float)obs_data_get_double(s, S_THRESHOLD);
+	cd->threshold = (float)obs_data_get_double(s, S_FILTER_THRESHOLD);
 
 	cd->attack_gain = gain_coefficient(sample_rate, attack_time_ms / MS_IN_S_F);
 	cd->release_gain = gain_coefficient(sample_rate, release_time_ms / MS_IN_S_F);
@@ -181,7 +181,7 @@ static struct obs_audio_data *limiter_filter_audio(void *data, struct obs_audio_
 
 static void limiter_defaults(obs_data_t *s)
 {
-	obs_data_set_default_double(s, S_THRESHOLD, -6.0f);
+	obs_data_set_default_double(s, S_FILTER_THRESHOLD, -6.0f);
 	obs_data_set_default_int(s, S_RELEASE_TIME, 60);
 }
 
@@ -190,8 +190,8 @@ static obs_properties_t *limiter_properties(void *data)
 	obs_properties_t *props = obs_properties_create();
 	obs_property_t *p;
 
-	p = obs_properties_add_float_slider(props, S_THRESHOLD, TEXT_THRESHOLD, MIN_THRESHOLD_DB, MAX_THRESHOLD_DB,
-					    0.1);
+	p = obs_properties_add_float_slider(props, S_FILTER_THRESHOLD, TEXT_THRESHOLD, MIN_THRESHOLD_DB,
+					    MAX_THRESHOLD_DB, 0.1);
 	obs_property_float_set_suffix(p, " dB");
 	p = obs_properties_add_int_slider(props, S_RELEASE_TIME, TEXT_RELEASE_TIME, MIN_ATK_RLS_MS, MAX_RLS_MS, 1);
 	obs_property_int_set_suffix(p, " ms");
--- a/plugins/obs-outputs/utils.h
+++ b/plugins/obs-outputs/utils.h
@@ -22,7 +22,11 @@ static inline uint32_t clz32(unsigned long val)
 
 static inline uint32_t ctz32(unsigned long val)
 {
+#if defined(_M_ARM64)
+	return _CountTrailingZeros(val);
+#else
 	return _tzcnt_u32(val);
+#endif
 }
 #else
 static uint32_t popcnt(uint32_t x)


kkartaltepe:
Unless you have tested QSV with your arm devices you should drop the proposed changes there and ensure the plugin is not built for arm.

tytan652:
Please reword the PR as "adding Windows on ARM support", we already support ARM64 on macOS and OBS can be built on Linux aarch64.

Also there is no CI build check which does not help to avoid/find build issue on WoA, which is not good at all since not all contributors will have WoA machines to check if it builds.

UI/volume-control.cpp
```diff
@@ -1153,8 +1153,13 @@ void VolumeMeter::paintVTicks(QPainter &painter, int x, int y, int height)
 
 inline int VolumeMeter::convertToInt(float number)
 {
+#if defined(_M_ARM64)
+	constexpr int min = (std::numeric_limits<int>::min)();
+	constexpr int max = (std::numeric_limits<int>::max)();
+#else
```
derrod, line 1159: What's the reason for this?

UI/obs-app.cpp
```diff
@@ -56,6 +56,9 @@
 
 #ifdef _WIN32
 #include <windows.h>
+#if defined(_M_ARM64)
+	#include <shellapi.h>
+#endif
```
derrod, line 61: Is this even used anywhere? It should also be included through `windows.h` anyway unless `WIN32_LEAN_AND_MEAN` is defined.

UI/window-basic-filters.cpp
```diff
@@ -36,9 +36,18 @@
...
-#ifdef _WIN32
-#define WIN32_LEAN_AND_MEAN 1
-#include <Windows.h>
+#if defined(_M_ARM64)
+	#ifdef _WIN32
+		#ifndef WIN32_LEAN_AND_MEAN
+		#define WIN32_LEAN_AND_MEAN 1
+		#endif
+	#endif
+	#include <Windows.h>
```
derrod, line 45: This should be inside the previous definition, but in general this whole thing seems pointless

UI/window-basic-preview.cpp
```diff
@@ -1861,6 +1861,18 @@ static void DrawStripedLine(float x1, float y1, float x2, float y2,
...
+			dx = (std::min)(xx1 + 7.5f * offX, x2);
+		} else {
+			dx = (std::max)(xx1 + 7.5f * offX, x2);
+		}
+		if (y1 < y2) {
+			dy = (std::min)(yy1 + 7.5f * offY, y2);
+		} else {
+			dy = (std::max)(yy1 + 7.5f * offY, y2);
+		}
+#else
```
derrod, line 1875: See previous comment about changes related to `std::max`/`std::min`

libobs/graphics/graphics.c
```diff
@@ -947,8 +947,13 @@ static inline void assign_sprite_uv(float *start, float *end, bool flip)
 		*start = 0.0f;
 		*end = 1.0f;
 	} else {
+#if defined(_M_ARM64)
+		*start = 0.0f;
+		*end = 1.0f;
+#else
```
derrod, line 953: Please elaborate why this change was made.

plugins/obs-filters/compressor-filter.c
```diff
@@ -28,6 +28,9 @@
 /* clang-format off */
 
 #define S_RATIO                         "ratio"
+#if defined(_M_ARM64)
+	#undef S_THRESHOLD
```
derrod, line 32: Why is this being undefined here?

plugins/obs-qsv11/common_utils_windows.cpp
```diff
@@ -172,7 +172,9 @@ double TimeDiffMsec(mfxTime tfinish, mfxTime tstart)
 
 void util_cpuid(int cpuinfo[4], int flags)
 {
+#if !defined(_M_ARM64)
 	return __cpuid(cpuinfo, flags);
+#endif
```
derrod, line 177: Is there an alternative for ARM? Probably not important since QSV would only work a with a dGPU anyway.

plugins/vlc-video/vlc-video-plugin.c
derrod, line 1: The VLC changes seem to target VLC4, which we do not support. And that should be split out into a separate PR.

derrod (changes requested):
Please actually read through the [contributing document](https://github.com/obsproject/obs-studio/blob/master/CONTRIBUTING.rst). Changes for different modules/components should be in separate commits. Also the motivation leaves a lot to be desired, there isn't really any explanation for why some of those changes were made/deemed necessary.

plugins/vlc-video/vlc-video-plugin.c
thirumalai-qcom: Thank you for the feedback. I will go ahead and separate the changes related to the VLC into a new PR.

plugins/obs-qsv11/common_utils_windows.cpp
thirumalai-qcom: Right now, there is no direct equivalent for this on ARM.
Reverting these changes as QSV11 is not supported on WOA yet.

plugins/obs-filters/compressor-filter.c
thirumalai-qcom: The `#undef S_THRESHOLD` is necessary to avoid a macro redefinition warning.
This is because `windows.h`, which includes `winbase.h`, defines `S_THRESHOLD` with a value of 1. Our source file requires `S_THRESHOLD` to be defined as `"threshold"`.

UI/window-basic-filters.cpp
thirumalai-qcom: Thanks for pointing that out. It was a mistake, I've made the necessary corrections now.

UI/volume-control.cpp
thirumalai-qcom: This change was made to address compilation errors related to the use of `std::numeric_limits<int>::min()` and `std::numeric_limits<int>::max()`. The `()` are included to ensure that we are explicitly invoking the functions from the `<limits>` header rather than the macros that are defined.

This approach is derived from a previous [PR #5455](https://github.com/obsproject/obs-studio/pull/5455#discussion_r779210328) that addressed a similar Issue.

UI/window-basic-preview.cpp
thirumalai-qcom: I've provided a brief explanation about the `std::max`/`std::min` modifications in the previous comment. You can review it [here](https://github.com/obsproject/obs-studio/pull/10997#discussion_r1688385006)

UI/obs-app.cpp
thirumalai-qcom: This header is used by the `ShellExecuteA()` function in `obs-app.cpp`.

https://github.com/obsproject/obs-studio/blob/5854f3b9e5861246ea57dd4a26d3d847a8552c4b/UI/obs-app.cpp#L2926

As `WIN32_LEAN_AND_MEAN` is defined, `shellapi.h` is not included by `windows.h`. By explicitly including `shellapi.h`, we ensure that sources get compiled, preventing compilation errors.

libobs/graphics/graphics.c
thirumalai-qcom: During Functional testing, we noticed that the browser page was displayed upside down when compiled on ARM64. To address this, we adjusted the `start` and `end` values to ensure proper browser-page orientation in OBS Studio.

**Before Modification:**
Before

**After Modifications:**
After

UI/volume-control.cpp
RytoEX: It looks like our options are to `#define NOMINMAX` whenever doing `#include <Windows.h>` in C++ files (and making sure there are no issues), using this syntax all the time, or this (only using this syntax within an appropriate ifdef).

UI/obs-app.cpp
RytoEX: I'm not aware of `WIN32_LEAN_AND_MEAN` being defined at this point or this currently causing compilation errors. A cursory check inside Visual Studio shows that `WIN32_LEAN_AND_MEAN` is not defined here. It is defined in other places in the codebase, but not here.

image

Have I misunderstood something?

libobs/graphics/graphics.c
WizardCM: If that is the only purpose of this change, it can be removed. It's due to the newer shared texture API within CEF, and therefore is handled in https://github.com/obsproject/obs-browser/pull/434/files#diff-22b4408287e08499d62e82ed1c8b4e586d4dfe80c4ecdf18fb561c1027be627aL661-R661.

UI/obs-app.cpp
thirumalai-qcom: Hi Ryan, The cursory check on WOA device (X-Elite) points that `WIN32_LEAN_AND_MEAN` is defined, so i have added `Shellapi.h` explicitly to resolve compilation issues.

shellapi

UI/volume-control.cpp
thirumalai-qcom: Using `()` explicitly the code ensures clarity and avoids macro conflicts. This inline change only affects the particular line if broken, making it a localized fix.

Defining `NOMINMAX` affects the entire file, meaning any legitimate use of the `min` and `max` macros would be disabled throughout the file. This could lead to potential issues elsewhere in the code where the `min` and `max` macros might be needed.

Introducing conditional statements for this purpose adds redundancy and complexity to the code. It again adds `()` only in specific conditions, which can make the code harder to maintain and understand.

plugins/vlc-video/vlc-video-plugin.c
thirumalai-qcom: Hi @derrod 
While expanding compatibility of OBS for Windows on ARM, we have noted that VLC for ARM is compatible only from VLC 4.0 onwards. Should the new PR changes focus solely on upgrading VLC to v4.0 for all architectures, or is the OBS team planning to support both VLC v3.x and v4.0?

UI/volume-control.cpp
RytoEX: I cannot think of legitimate uses of the `min` and `max` macros unless you're working with specific Windows headers that require them (such as perhaps GdiPlus headers, which we only use in two places).

The change shown in this review had already introduced conditional preprocessor statements here. The example you linked to unconditionally adds `()`. Unfortunately, this code preview was outdated, and I had not seen that the code was changed to unconditionally add `()`, which is what I was trying to hint at.

norihiro:
The commit messages need to follow the guideline. https://github.com/obsproject/obs-studio/blob/master/CONTRIBUTING.rst#commit-guidelines

UI/volume-control.cpp
tytan652: @thirumalai-qcom, OBS Studio is a multi-platform application which relies on common API in C library API and C++ library API.

As it is, the `min` and `max` macro that are not part of the C/C++ APIs. And moreover it interfere with the C++ STL that is multi-platform.

Defining `NOMINMAX` is almost mandatory to make sure that the same C/C++ APIs is present on the 3 main platforms.

Those macros are not a thing on macOS or Linux, so disabling is simpler more sensible than making workaround for macros that OBS multi-platform code will never use.

plugins/obs-filters/compressor-filter.c
tytan652: Looks like a namespace conflict (here `S_`), the reasonable change would be to change all `compressor-filter.c` `S_` macros to something like `FILTER_S_` in a **separate** PR.

UI/volume-control.cpp
thirumalai-qcom: @RytoEX Could you share your thoughts on which approach to implement here? Should we continue using the `()` explicitly, or would it be more efficient to define `NOMINMAX` given the context of OBS Studio's multi-platform nature?

plugins/obs-filters/compressor-filter.c
norihiro: The majority of the plugins use `S_` as the prefix to define setting names. Changing the macro name makes the history dirty. IMHO, `#undef` is reasonable.

There are various ways in which the setting names are written among the 1st-party plugins. Unfortunately, the majority write string literals directly without macros. The 2nd place is `S_`.

Besides this PR, I want to suggest the style should be unified among the plugins. In my opinion, `SETTING_`, `OPT_`, or `PROP_` (no one uses this though) would be descriptive rather than `S_`.

Plugins directly writing string literals
- https://github.com/obsproject/obs-studio/blob/ff2fa24dc25519475687041ba6af829e9ba41335/plugins/coreaudio-encoder/encoder.cpp#L1375-L1386
- https://github.com/obsproject/obs-studio/blob/ff2fa24dc25519475687041ba6af829e9ba41335/plugins/image-source/color-source.c#L22-L24
- https://github.com/obsproject/obs-studio/blob/ff2fa24dc25519475687041ba6af829e9ba41335/plugins/image-source/image-source.c#L104-L107
- https://github.com/obsproject/obs-studio/blob/ff2fa24dc25519475687041ba6af829e9ba41335/plugins/linux-alsa/alsa-input.c#L190-L201
- https://github.com/obsproject/obs-studio/blob/ff2fa24dc25519475687041ba6af829e9ba41335/plugins/linux-capture/xcomposite-input.c#L855-L863
- https://github.com/obsproject/obs-studio/blob/ff2fa24dc25519475687041ba6af829e9ba41335/plugins/linux-capture/xshm-input.c#L275-L283
- https://github.com/obsproject/obs-studio/blob/ff2fa24dc25519475687041ba6af829e9ba41335/plugins/linux-jack/jack-input.c#L61-L62
- https://github.com/obsproject/obs-studio/blob/ff2fa24dc25519475687041ba6af829e9ba41335/plugins/linux-pipewire/camera-portal.c#L1307-L1330
- https://github.com/obsproject/obs-studio/blob/ff2fa24dc25519475687041ba6af829e9ba41335/plugins/linux-pipewire/screencast-portal.c#L609-L611
- https://github.com/obsproject/obs-studio/blob/ff2fa24dc25519475687041ba6af829e9ba41335/plugins/linux-pulseaudio/pulse-input.c#L552
- https://github.com/obsproject/obs-studio/blob/ff2fa24dc25519475687041ba6af829e9ba41335/plugins/linux-v4l2/v4l2-input.c#L1154-L1163
- https://github.com/obsproject/obs-studio/blob/ff2fa24dc25519475687041ba6af829e9ba41335/plugins/mac-avcapture/plugin-properties.m#L321-L326
- https://github.com/obsproject/obs-studio/blob/ff2fa24dc25519475687041ba6af829e9ba41335/plugins/mac-videotoolbox/encoder.c#L823-L833
- https://github.com/obsproject/obs-studio/blob/ff2fa24dc25519475687041ba6af829e9ba41335/plugins/obs-ffmpeg/obs-ffmpeg-av1.c#L69-73
- https://github.com/obsproject/obs-studio/blob/ff2fa24dc25519475687041ba6af829e9ba41335/plugins/obs-ffmpeg/obs-ffmpeg-hls-mux.c#L143
- https://github.com/obsproject/obs-studio/blob/ff2fa24dc25519475687041ba6af829e9ba41335/plugins/obs-ffmpeg/obs-ffmpeg-mpegts.c#L976-L977
- https://github.com/obsproject/obs-studio/blob/ff2fa24dc25519475687041ba6af829e9ba41335/plugins/obs-ffmpeg/obs-nvenc.c#L647-L657
- https://github.com/obsproject/obs-studio/blob/ff2fa24dc25519475687041ba6af829e9ba41335/plugins/obs-filters/crop-filter.c#L76-L82
- https://github.com/obsproject/obs-studio/blob/ff2fa24dc25519475687041ba6af829e9ba41335/plugins/obs-filters/eq-filter.c#L47-L49
- https://github.com/obsproject/obs-studio/blob/ff2fa24dc25519475687041ba6af829e9ba41335/plugins/obs-filters/hdr-tonemap-filter.c#L76-L86
- https://github.com/obsproject/obs-studio/blob/ff2fa24dc25519475687041ba6af829e9ba41335/plugins/obs-filters/scroll-filter.c#L79-L89
- https://github.com/obsproject/obs-studio/blob/ff2fa24dc25519475687041ba6af829e9ba41335/plugins/obs-filters/sharpness-filter.c#L26
- https://github.com/obsproject/obs-studio/blob/ff2fa24dc25519475687041ba6af829e9ba41335/plugins/obs-libfdk/obs-libfdk.c#L96-L99
- https://github.com/obsproject/obs-studio/blob/ff2fa24dc25519475687041ba6af829e9ba41335/plugins/obs-outputs/flv-mux.c#L139
- https://github.com/obsproject/obs-studio/blob/ff2fa24dc25519475687041ba6af829e9ba41335/plugins/obs-outputs/flv-output.c#L509
- https://github.com/obsproject/obs-studio/blob/ff2fa24dc25519475687041ba6af829e9ba41335/plugins/obs-outputs/mp4-output.c#L263-L274
- https://github.com/obsproject/obs-studio/blob/ff2fa24dc25519475687041ba6af829e9ba41335/plugins/obs-qsv11/obs-qsv11.c#L262-344
- https://github.com/obsproject/obs-studio/blob/ff2fa24dc25519475687041ba6af829e9ba41335/plugins/obs-transitions/transition-stinger.c:#L70-L72
- https://github.com/obsproject/obs-studio/blob/ff2fa24dc25519475687041ba6af829e9ba41335/plugins/obs-x264/obs-x264.c#L402-L416
- https://github.com/obsproject/obs-studio/blob/ff2fa24dc25519475687041ba6af829e9ba41335/plugins/text-freetype2/text-freetype2.c#L341-358
- https://github.com/obsproject/obs-studio/blob/ff2fa24dc25519475687041ba6af829e9ba41335/plugins/win-capture/duplicator-monitor-capture.c
- https://github.com/obsproject/obs-studio/blob/ff2fa24dc25519475687041ba6af829e9ba41335/plugins/win-capture/monitor-capture.c
- https://github.com/obsproject/obs-studio/blob/ff2fa24dc25519475687041ba6af829e9ba41335/plugins/win-capture/window-capture.c
- https://github.com/obsproject/obs-studio/blob/ff2fa24dc25519475687041ba6af829e9ba41335/plugins/win-dshow/win-dshow-encoder.cpp

Image slideshow uses `static const char *` with the prefix `S_`. (Probably, this is Lain's preference, isn't this?)
- https://github.com/obsproject/obs-studio/blob/ff2fa24dc25519475687041ba6af829e9ba41335/plugins/image-source/obs-slideshow-mk2.c#L19-L33

Plugins using macros with the prefix `S_`
- https://github.com/obsproject/obs-studio/blob/ff2fa24dc25519475687041ba6af829e9ba41335/plugins/obs-filters/compressor-filter.c#L30-L35
- https://github.com/obsproject/obs-studio/blob/ff2fa24dc25519475687041ba6af829e9ba41335/plugins/obs-filters/expander-filter.c#L30-L37
- https://github.com/obsproject/obs-studio/blob/ff2fa24dc25519475687041ba6af829e9ba41335/plugins/obs-filters/gain-filter.c#L12
- https://github.com/obsproject/obs-studio/blob/ff2fa24dc25519475687041ba6af829e9ba41335/plugins/obs-filters/gpu-delay.c#L5
- https://github.com/obsproject/obs-studio/blob/ff2fa24dc25519475687041ba6af829e9ba41335/plugins/obs-filters/limiter-filter.c#L28-L29
- https://github.com/obsproject/obs-studio/blob/ff2fa24dc25519475687041ba6af829e9ba41335/plugins/obs-filters/noise-gate-filter.c#L14-L18
- https://github.com/obsproject/obs-studio/blob/ff2fa24dc25519475687041ba6af829e9ba41335/plugins/obs-filters/noise-suppress-filter.c#L43-L50
- https://github.com/obsproject/obs-studio/blob/ff2fa24dc25519475687041ba6af829e9ba41335/plugins/obs-filters/nvidia-greenscreen-filter.c#L25-L30
- https://github.com/obsproject/obs-studio/blob/ff2fa24dc25519475687041ba6af829e9ba41335/plugins/obs-filters/scale-filter.c#L11-L30
- https://github.com/obsproject/obs-studio/blob/ff2fa24dc25519475687041ba6af829e9ba41335/plugins/obs-text/gdiplus/obs-text.cpp#L46-L80
- https://github.com/obsproject/obs-studio/blob/ff2fa24dc25519475687041ba6af829e9ba41335/plugins/obs-transitions/transition-fade-to-color.c#L3-L7
- https://github.com/obsproject/obs-studio/blob/ff2fa24dc25519475687041ba6af829e9ba41335/plugins/obs-transitions/transition-luma-wipe.c#L7-L9
- https://github.com/obsproject/obs-studio/blob/ff2fa24dc25519475687041ba6af829e9ba41335/plugins/obs-transitions/transition-slide.c#L5
- https://github.com/obsproject/obs-studio/blob/ff2fa24dc25519475687041ba6af829e9ba41335/plugins/obs-transitions/transition-swipe.c#L17-L18
- https://github.com/obsproject/obs-studio/blob/ff2fa24dc25519475687041ba6af829e9ba41335/plugins/vlc-video/vlc-video-source.c#L15-L25

Plugins using macros with a prefix `SETTING_`
- https://github.com/obsproject/obs-studio/blob/ff2fa24dc25519475687041ba6af829e9ba41335/plugins/obs-filters/async-delay-filter.c#L17
- https://github.com/obsproject/obs-studio/blob/ff2fa24dc25519475687041ba6af829e9ba41335/plugins/obs-filters/chroma-key-filter.c#L8-L17
- https://github.com/obsproject/obs-studio/blob/ff2fa24dc25519475687041ba6af829e9ba41335/plugins/obs-filters/color-correction-filter.c#L23-L32
- https://github.com/obsproject/obs-studio/blob/ff2fa24dc25519475687041ba6af829e9ba41335/plugins/obs-filters/color-grade-filter.c#L9-L12
- https://github.com/obsproject/obs-studio/blob/ff2fa24dc25519475687041ba6af829e9ba41335/plugins/obs-filters/color-key-filter.c#L9-12
- https://github.com/obsproject/obs-studio/blob/ff2fa24dc25519475687041ba6af829e9ba41335/plugins/obs-filters/color-key-filter.c
- https://github.com/obsproject/obs-studio/blob/ff2fa24dc25519475687041ba6af829e9ba41335/plugins/obs-filters/luma-key-filter.c
- https://github.com/obsproject/obs-studio/blob/ff2fa24dc25519475687041ba6af829e9ba41335/plugins/obs-filters/mask-filter.c
- https://github.com/obsproject/obs-studio/blob/ff2fa24dc25519475687041ba6af829e9ba41335/plugins/win-capture/game-capture.c

Plugins using macros without prefix
- https://github.com/obsproject/obs-studio/blob/ff2fa24dc25519475687041ba6af829e9ba41335/plugins/decklink/const.h#L1-L17
- https://github.com/obsproject/obs-studio/blob/ff2fa24dc25519475687041ba6af829e9ba41335/plugins/win-dshow/win-dshow.cpp#L34-L50
- https://github.com/obsproject/obs-studio/blob/ff2fa24dc25519475687041ba6af829e9ba41335/plugins/win-wasapi/win-wasapi.cpp#L27-L30

Plugins using macros with a prefix `OPT_`
- https://github.com/obsproject/obs-studio/blob/ff2fa24dc25519475687041ba6af829e9ba41335/plugins/obs-outputs/rtmp-stream.c#L26-L34
- https://github.com/obsproject/obs-studio/blob/ff2fa24dc25519475687041ba6af829e9ba41335/plugins/win-wasapi/win-wasapi.cpp#L27-L30

Only AJA plugin has a structure.
- https://github.com/obsproject/obs-studio/blob/ff2fa24dc25519475687041ba6af829e9ba41335/plugins/aja/aja-ui-props.hpp

thirumalai-qcom:
@RytoEX I would like to know if AJA functionality is planned to be included in the Phase-1 release of the arm64 version of OBS. Currently, in AJA the [`__cpuid`](https://github.com/aja-video/libajantv2/blob/7a53c590243868b76e5401fdcb23ec743bc07f7a/ajabase/system/windows/infoimpl.cpp#L375) method is not supported on ARM devices, Could you provide some insights into what functionality we will miss in OBS on ARM due to this limitation ?

plugins/obs-filters/compressor-filter.c
tytan652: > Changing the macro name makes the history dirty.

Undefining macros that are not ours is not a good idea either.

plugins/vlc-video/vlc-video-plugin.c
RytoEX: There should be no work required to support VLC 3.0.21, unless their headers have had changes that we would require to build the vlc-video plugin/source against them. The VLC source in OBS relies on host-system binaries.

thirumalai-qcom: Hi @RytoEX,
I just want to clarify that it's not that straight-forward.
Whatever you've stated is true incase of Windows on x86_64 because as 

> The VLC source in OBS relies on host-system binaries 

these binaries are detected and their paths are extracted using the [registry](https://github.com/obsproject/obs-studio/blob/95a753b9d8f16ba1cc73ba83936368ff2c74ff5e/plugins/vlc-video/vlc-video-plugin.c#L164).  However, on Windows on ARM (WOA), the official binary installer is not available and VLC(v3.0.21) is only available through MSYS2 packages. Installation via MSYS2 with pacman (the package manager of MSYS2) doesn't create a registry key, so VLC is not detected by the system.

To solve this, one solution is to check the default MSYS2 installation path `(c:/msys64/clangarm64/bin)` to detect VLC,For example, here's a quick code fix to handle this in [plugins/vlc-video/vlc-video-plugin.c](https://github.com/obsproject/obs-studio/blob/95a753b9d8f16ba1cc73ba83936368ff2c74ff5e/plugins/vlc-video/vlc-video-plugin.c#L155) :
```
#ifdef _WIN32
    char *path_utf8 = NULL;
    wchar_t path[1024];
    #ifdef _M_ARM64
        memset(path, 0, 1024 * sizeof(wchar_t));
        //checking if VLC is present in Clangarm64/include.
        if (_waccess(L"c:/msys64/clangarm64/include/vlc/", 0) == 0)
        {
           wcscpy(path, L"c:/msys64/clangarm64/bin"); //Default msys2 folder
           wcscat(path, L"[\\libvlc.dll](file://libvlc.dll/)");
           os_wcs_to_utf8_ptr(path, 0, &path_utf8);
           libvlc_module = os_dlopen(path_utf8);
           bfree(path_utf8);
        }
    #else
        LSTATUS status;
	DWORD size;
	HKEY key;

	memset(path, 0, 1024 * sizeof(wchar_t));

	status = RegOpenKeyW(HKEY_LOCAL_MACHINE, L"SOFTWARE\\VideoLAN\\VLC",
			     &key);
	if (status != ERROR_SUCCESS)
		return false;

	size = 1024;
	status = RegQueryValueExW(key, L"InstallDir", NULL, NULL, (LPBYTE)path,
				  &size);
	if (status == ERROR_SUCCESS) {
		wcscat(path, L"\\libvlc.dll");
		os_wcs_to_utf8_ptr(path, 0, &path_utf8);
		libvlc_module = os_dlopen(path_utf8);
		bfree(path_utf8);
	}

	RegCloseKey(key);
   #endif
```

With the above fix, it should be seamless on WOA also & I've verified it too, what do you think of this?

plugins/obs-filters/compressor-filter.c
Fenrirthviti: Poking @paulh-aja for insight in to AJA's efforts/status on ARM support.

plugins/vlc-video/vlc-video-plugin.c
RytoEX: Off-thread discussion concluded that VLC support will be set aside for now.

Fenrirthviti:
> @RytoEX I would like to know if AJA functionality is planned to be included in the Phase-1 release of the arm64 version of OBS. Currently, in AJA the [`__cpuid`](https://github.com/aja-video/libajantv2/blob/7a53c590243868b76e5401fdcb23ec743bc07f7a/ajabase/system/windows/infoimpl.cpp#L375) method is not supported on ARM devices, Could you provide some insights into what functionality we will miss in OBS on ARM due to this limitation ?
> 
> Edit: Identified an alternative way to retrieve the CPU string from the register (as done in `libobs/obs-windows.c` [log_processor_info()](https://github.com/obsproject/obs-studio/blob/3b266fec22462719b1dfe0b56380811ec791437a/libobs/obs-windows.c#L63C1-L86C3)). Can this be a replacement for the `__cpuid` method on ARM devices?

Poking @paulh-aja for AJA's current stats on ARM support. (sorry for double ping here, accidentally replied to the wrong comment!)

RytoEX:
> > @RytoEX I would like to know if AJA functionality is planned to be included in the Phase-1 release of the arm64 version of OBS. Currently, in AJA the [`__cpuid`](https://github.com/aja-video/libajantv2/blob/7a53c590243868b76e5401fdcb23ec743bc07f7a/ajabase/system/windows/infoimpl.cpp#L375) method is not supported on ARM devices, Could you provide some insights into what functionality we will miss in OBS on ARM due to this limitation ?
> > Edit: Identified an alternative way to retrieve the CPU string from the register (as done in `libobs/obs-windows.c` [log_processor_info()](https://github.com/obsproject/obs-studio/blob/3b266fec22462719b1dfe0b56380811ec791437a/libobs/obs-windows.c#L63C1-L86C3)). Can this be a replacement for the `__cpuid` method on ARM devices?
> 
> Poking @paulh-aja for AJA's current stats on ARM support. (sorry for double ping here, accidentally replied to the wrong comment!)

My thinking on this would be that if there's an ARM64 equivalent to `__cpuid`, then it should (probably) be trivial to ifdef appropriately, which should be fine. That said, the linked code is in AJA's SDK, so the changes will have to be made there, not in OBS. It is worth noting that you cannot simply lift existing code from the obs-studio repo to drop into AJA's repo as the licenses, while compatible, are different. The correct way to address this is likely to file an Issue (or PR) on their repo.

paulh-aja:
> Currently, in AJA the [__cpuid](https://github.com/aja-video/libajantv2/blob/7a53c590243868b76e5401fdcb23ec743bc07f7a/ajabase/system/windows/infoimpl.cpp#L375) method is not supported on ARM devices.

> Poking @paulh-aja for AJA's current stats on ARM support. 

Please open a PR or at least raise an issue in the [libajantv2](https://github.com/aja-video/libajantv2) repo as RytoEX mentioned.

We do not have Windows ARM support on our current release roadmap. However, if the only thing blocking it is replacing the use of `__cpuid` in the ajabase code, then it may be trivial to add it in. Our Mac and Linux drivers and SDK do support ARM64 today.

thirumalai-qcom:
> Please open a PR or at least raise an issue in the [libajantv2](https://github.com/aja-video/libajantv2) repo as RytoEX mentioned.

@paulh-aja I've raised a PR in the libajantv2 repo to replace the `__cpuid` in the ajabase code.

thirumalai-qcom:
> The correct way to address this is likely to file an Issue (or PR) on their repo.

@RytoEX 
As the AJA PR has been successfully merged into the mainline, I was curious to know if there are any updates on this PR merging for the initial release on Windows on ARM (WOA)? Could you also provide your inputs on which fix would be appropriate to resolve the compilation errors related to `std::numeric_limits<int>::min()` error and `S_THRESHOLD`macro redefinition warning?

UI/obs-app.cpp
RytoEX: That's strange, but I suppose it's some platform difference that makes this required.

plugins/obs-filters/compressor-filter.c
RytoEX: To clarify, the larger conversation here is whether or not we should explore a more comprehensive change that renames our macros to avoid the conflicts with generic defines such as `S_THRESHOLD`. The smaller scoped conversation is "are we okay with undefining this macro to get this to build?" The secondary question is "does `WIN32_LEAN_AND_MEAN` solve this?"

UI/volume-control.cpp
thirumalai-qcom: - I have modified the PR by adding `NOMINMAX` flag to the 3 specific files where the macro conflicts arose.
- Adding `NOMINMAX` globally as suggested by @RytoEX breaks the compilation as some files already have it defined.

https://github.com/thirumalai-qcom/obs-studio/blob/ec5297b549aa2884287a1b7e986e0f2bf08ad143/UI/multitrack-video-output.hpp#L13

plugins/obs-filters/compressor-filter.c
thirumalai-qcom: The `winbase.h` header file defines `S_THRESHOLD` with a value of `1`, and this definition cannot be excluded by the `WIN32_LEAN_AND_MEAN` directive.
So the available options are to `#undef S_THRESHOLD` or Renaming the OBS macro which would affect the project's history, as mentioned by @norihiro

UI/volume-control.cpp
thirumalai-qcom: - Refactored PR for initial ARM64 enablement; added #NOMINMAX macro to relevant headers to fix compilation issues in arm64.

RytoEX (changes requested):
With https://github.com/obsproject/obs-studio/pull/11385 merged, please drop the NOMINMAX changes from this PR. Additionally, the CMake changes seem to affect more lines than they should. Please drop those and revert the scope of this PR back to just the remaining code changes necessary to make compiling possible

deps/w32-pthreads/context.h
```diff
@@ -9,25 +9,25 @@
...
  *      License as published by the Free Software Foundation; either
  *      version 2 of the License, or (at your option) any later version.
- * 
+ *
  *      This library is distributed in the hope that it will be useful,
  *      but WITHOUT ANY WARRANTY; without even the implied warranty of
  *      MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the GNU
  *      Lesser General Public License for more details.
- * 
+ *
```
RytoEX, line 30: Revert these cosmetic changes.

plugins/obs-vst/headers/VSTPlugin.h
```diff
@@ -21,14 +21,14 @@ along with this program.  If not, see <http://www.gnu.org/licenses/>.
...
 
+#include "EditorWidget.h"
 #include <mutex>
 #include <atomic>
 #include <string>
 #include <QDirIterator>
 #include <obs-module.h>
 #include "aeffectx.h"
 #include "vst-plugin-callbacks.hpp"
-#include "EditorWidget.h"
```
RytoEX, line 31: This seems unnecessary. Revert this unless it is necessary for successful compilation.

RytoEX (changes requested):

deps/w32-pthreads/context.h
thirumalai-qcom: Done, thanks for your feedback

plugins/obs-vst/headers/VSTPlugin.h
thirumalai-qcom: Thanks for pointing out that, this can be reverted as #11385 is merged.

thirumalai-qcom:
> With #11385 merged, please drop the NOMINMAX changes from this PR. Additionally, the CMake changes seem to affect more lines than they should. Please drop those and revert the scope of this PR back to just the remaining code changes necessary to make compiling possible

@RytoEX 
Thanks for the information regarding `NOMINMAX`, I have refactored the PR with your suggestions.

plugins/obs-filters/compressor-filter.c
RytoEX: For now, let us change `S_THRESHOLD` in the affected files to `S_FILTER_THRESHOLD`. At a later time, we will pursue a more comprehensive solution.

thirumalai-qcom: Thanks for your inputs @RytoEX. The PR is updated as per the suggestion.

RytoEX (changes requested):
Please update the commit messages to reflect the current changes.

thirumalai-qcom:
> Please update the commit messages to reflect the current changes.

Thanks for the feedback @RytoEX, I've updated the commit messages and descriptions to reflect the current changes. 
Let me know if any further adjustments are needed.

RytoEX (approved):
Tentatively approved, though I may force push a change before merging to fix nitpicks on commit messages.

PR #11000, opened 2024-07-17
Add architecture support to add_obs_plugin and allow QSV only on x86_64

tytan652:
### Description

Add architecture support to add_obs_plugin and allow QSV only on x86_64

### Motivation and Context

Intel stack is only completely buildable on x86_64 architectures.

### How Has This Been Tested?

I need CI to check Windows.

### Types of changes

### Checklist:


cmake/common/helpers_common.cmake
```diff
@@ -492,7 +495,17 @@ function(add_obs_plugin target)
     endforeach()
   endif()
 
-  if(found_platform)
+  if(_AOP_NUM_ARCHITECTURES EQUAL 0)
+    set(found_architecture TRUE)
+  else()
+    foreach(architecture IN LISTS _AOP_ARCHITECTURES)
+      if("${architecture}" STREQUAL CMAKE_SYSTEM_PROCESSOR)
```
PatTheMav, line 502: You cannot just check for `CMAKE_SYSTEM_PROCESSOR` because it will not match the build configuration on Windows or macOS (as neither uses this value for target architecture selection).

On macOS you'd need to check for `CMAKE_OSX_ARCHITECTURES` (which can be a list, i.e. `x86_64;arm64` because you can build universal binaries) and on Windows you need to check for `CMAKE_GENERATOR_PLATFORM` which can be either `Win32`, `x64`, or `ARM64`.

cmake/common/helpers_common.cmake
```diff
@@ -499,8 +499,18 @@ function(add_obs_plugin target)
     set(found_architecture TRUE)
   else()
     foreach(architecture IN LISTS _AOP_ARCHITECTURES)
-      if("${architecture}" STREQUAL CMAKE_SYSTEM_PROCESSOR)
-        set(found_architecture TRUE)
+      if(OS_WINDOWS AND CMAKE_GENERATOR MATCHES "Visual Studio .+")
```
PatTheMav, line 502: We usually don't explicitly check for Visual Studio (because we don't support any other way to build OBS on Windows). It's fine as-is, but it's implicitly expected and stuff is expected to break if you don't use MSVC.

tytan652: I remove the check, if only MSVC is supported as generator.

cmake/common/helpers_common.cmake
```diff
@@ -492,7 +495,27 @@ function(add_obs_plugin target)
...
-  if(found_platform)
+  if(_AOP_NUM_ARCHITECTURES EQUAL 0)
+    set(found_architecture TRUE)
+  else()
+    foreach(architecture IN LISTS _AOP_ARCHITECTURES)
+      if(OS_WINDOWS)
+        if(CMAKE_GENERATOR_PLATFORM STREQUAL "${architecture}")
+          set(found_architecture TRUE)
+        endif()
+      elseif(OS_MACOS)
```
PatTheMav, line 506: You can use `elseif(OS_MACOS AND ${architecture} IN_LIST CMAKE_OSX_ARCHITECTURES)` here - `IN_LIST` will yield true even if `CMAKE_OSX_ARCHITECTURES` is just a single value.

cmake/common/helpers_common.cmake
tytan652: Tell me if you are fine with the change/direction so I can squash.

tytan652: Please mark as resolved if you are okay with the change.

PatTheMav: See my other review regarding `CMAKE_OSX_ARCHITECTURES`.

cmake/common/helpers_common.cmake
tytan652: I keep a `elseif` and a `if` to not fallback to `CMAKE_SYSTEM_PROCESSOR`.

PatTheMav: Yeah that's fine, using `IN_LIST` should still be simpler than the regex match.

cmake/common/helpers_common.cmake
tytan652: Done

PatTheMav (approved):
Looks fine to me, if CI shows that QSV was correctly included for Windows and Linux (and other plugins without specified architectures were added in all cases), should be good to go.

RytoEX (approved):
This looks good at a glance. Will target for 31.

PR #11004, opened 2024-07-18
libobs: Add Broadcast Performance Metrics (BPM)

lexano-ivs:
### Description
This PR implements initial support for Broadcast Performance Metrics (BPM). BPM currently sends the following metrics inband with the video bitstream using SEI (for AVC/HEVC) and OBU (for AV1) with each IDR-frame, which is typically at 2 second intervals:

- UTC-based wall-clock timestamps using RFC3339 format
- Session frame counters: rendered, lagged, dropped, output
- Encoded rendition frame counters: input, skipped, output

The metrics are intended to be used by live-streaming services such as Twitch and Amazon IVS, for example in plugins such as multitrack/eRTMP. BPM generation and delivery is enabled for the output by registering the `bpm_injection()` callback with `obs_output_add_packet_callback()`. This packet callback mechanism is introduced with this PR after the initial implementation (within libobs) was removed in favour of a callback API at the packet processing level.

Detailed BPM documentation is available in the [Multitrack Video Integration Guide](https://d50yg09cghihd.cloudfront.net/other/20240718-MultitrackVideoIntegrationGuide.pdf).

### Motivation and Context
To enable ongoing improvement of automatic stream configuration, to deliver the best possible stream settings, broadcast performance metrics (BPM) must be measured and sent. The metrics are collected and sent in-band via either SEI (for AVC/HEVC) or OBU (AV1) messages. Two classes of data are collected:

- Timestamps are collected to measure end-to-end latency between the broadcaster and the viewer. They are useful for:
  - Providing the broadcaster or audience with an estimate of end-to-end latency
  - Analyzing timestamp jitter that may indicate system stress or poor first-mile network connectivity
  - Referencing real-world event time for aligning and aggregating time-series counter data

- Frame counters are collected to measure the performance of the broadcast software and video encoders at the frame level. They are useful for:
  - Providing broadcasters with a performance dashboard that includes additional signals, to help
them improve their streaming setup
  - Providing a proactive signal that may correlate with environmental changes like newly released GPU drivers or OS versions/patches
  - Providing feedback to enable video services to safely iterate and release improvements to GetClientConfiguration, including support for new hardware vendors, new GPU models, new codecs, new driver features, additional video-encoder setting tuning, and new user-controlled presets (e.g., “Dual PC Setup” vs. “Gaming+Streaming Setup”).

The timestamp sent from the broadcaster is based on a global common reference clock, typically an NTP-synchronized clock using the UTC+0 timezone. RFC3339 is commonly used for this scenario of "Internet time." This provides an absolute reference, making temporal difference calculations trivial. The timestamps have millisecond resolution which is necessary to provide resolution suitable for measuring end-to-end latency, and to provide enough resolution for frame-level timestamps. 

The events being timestamped for video frames are:

1. CTS (Composition timestamp): When the composited frame was rendered by OBS
2. FER (Frame encode request timestamp): When the frame was requested to be encoded
3. FERC (Frame encode request complete timestamp): When the frame encode request was completed
4. PIR (Packet interleave request): When the compressed frame (packet) was sent to the output plugin.

The time difference between PIR and CTS represents the majority of the client latency.

### How Has This Been Tested?
BPM was tested in the Twitch Enhanced Broadcasting beta, streaming multi-rendition and single-track sessions into the Twitch ingest service, parsing the metrics, and sending the results to the control plane back-end. Additionally, it was tested with AVC, HEVC, and AV1, with and without closed captions injections. BPM and closed captions use a very similar mechanism of injecting data using SEI/OBU hence the need to test both alone and simultaneously. Multiple bitstreams were also captured and verified with a stream analyzer for correctness.

### Types of changes

### Checklist:

### DIFF
--- a/libobs/obs-av1.h
+++ b/libobs/obs-av1.h
@@ -22,6 +22,15 @@ enum {
 	OBS_OBU_PADDING = 15,
 };
 
+enum av1_obu_metadata_type {
+	METADATA_TYPE_HDR_CLL = 1,
+	METADATA_TYPE_HDR_MDCV,
+	METADATA_TYPE_SCALABILITY,
+	METADATA_TYPE_ITUT_T35,
+	METADATA_TYPE_TIMECODE,
+	METADATA_TYPE_USER_PRIVATE_6
+};
+
 /* Helpers for parsing AV1 OB units.  */
 
 EXPORT bool obs_av1_keyframe(const uint8_t *data, size_t size);
@@ -33,6 +42,9 @@ EXPORT void obs_extract_av1_headers(const uint8_t *packet, size_t size,
 EXPORT void metadata_obu_itu_t35(const uint8_t *itut_t35_buffer,
 				 size_t itut_bufsize, uint8_t **out_buffer,
 				 size_t *outbuf_size);
+EXPORT void metadata_obu(const uint8_t *source_buffer, size_t source_bufsize,
+			 uint8_t **out_buffer, size_t *outbuf_size,
+			 uint8_t metadata_type);
 
 #ifdef __cplusplus
 }
--- a/libobs/obs-internal.h
+++ b/libobs/obs-internal.h
@@ -1319,6 +1319,9 @@ struct obs_encoder {
 	uint32_t frame_rate_divisor_counter; // only used for GPU encoders
 	video_t *fps_override;
 
+	// Number of frames successfully encoded
+	uint32_t encoded_frames;
+
 	/* Regions of interest to prioritize during encoding */
 	pthread_mutex_t roi_mutex;
 	DARRAY(struct obs_encoder_roi) roi;
--- a/libobs/obs-internal.h
+++ b/libobs/obs-internal.h
@@ -26,6 +26,7 @@
 #include "util/profiler.h"
 #include "util/task.h"
 #include "util/uthash.h"
+#include "util/array-serializer.h"
 #include "callback/signal.h"
 #include "callback/proc.h"
 
@@ -1065,9 +1066,12 @@ struct delay_data {
 	enum delay_msg msg;
 	uint64_t ts;
 	struct encoder_packet packet;
+	bool packet_time_valid;
+	struct encoder_packet_time packet_time;
 };
 
-typedef void (*encoded_callback_t)(void *data, struct encoder_packet *packet);
+typedef void (*encoded_callback_t)(void *data, struct encoder_packet *packet,
+				   struct encoder_packet_time *frame_time);
 
 struct obs_weak_output {
 	struct obs_weak_ref ref;
@@ -1185,6 +1189,8 @@ struct obs_output {
 	// captions are output per track
 	struct caption_track_data *caption_tracks[MAX_OUTPUT_VIDEO_ENCODERS];
 
+	DARRAY(struct encoder_packet_time) encoder_packet_times[MAX_OUTPUT_VIDEO_ENCODERS];
+
 	bool valid;
 
 	uint64_t active_delay_ns;
@@ -1212,7 +1218,8 @@ static inline void do_output_signal(struct obs_output *output,
 	calldata_free(&params);
 }
 
-extern void process_delay(void *data, struct encoder_packet *packet);
+extern void process_delay(void *data, struct encoder_packet *packet,
+			  struct encoder_packet_time *packet_time);
 extern void obs_output_cleanup_delay(obs_output_t *output);
 extern bool obs_output_delay_start(obs_output_t *output);
 extern void obs_output_delay_stop(obs_output_t *output);
@@ -1240,7 +1247,7 @@ struct obs_weak_encoder {
 
 struct encoder_callback {
 	bool sent_first_packet;
-	void (*new_packet)(void *param, struct encoder_packet *packet);
+	encoded_callback_t new_packet;
 	void *param;
 };
 
@@ -1332,6 +1339,8 @@ struct obs_encoder {
 	pthread_mutex_t callbacks_mutex;
 	DARRAY(struct encoder_callback) callbacks;
 
+	DARRAY(struct encoder_packet_time) encoder_packet_times;
+
 	struct pause_data pause;
 
 	const char *profile_encoder_encode_name;
@@ -1347,13 +1356,9 @@ extern bool obs_encoder_initialize(obs_encoder_t *encoder);
 extern void obs_encoder_shutdown(obs_encoder_t *encoder);
 
 extern void obs_encoder_start(obs_encoder_t *encoder,
-			      void (*new_packet)(void *param,
-						 struct encoder_packet *packet),
-			      void *param);
+			      encoded_callback_t new_packet, void *param);
 extern void obs_encoder_stop(obs_encoder_t *encoder,
-			     void (*new_packet)(void *param,
-						struct encoder_packet *packet),
-			     void *param);
+			     encoded_callback_t new_packet, void *param);
 
 extern void obs_encoder_add_output(struct obs_encoder *encoder,
 				   struct obs_output *output);
@@ -1363,7 +1368,8 @@ extern void obs_encoder_remove_output(struct obs_encoder *encoder,
 extern bool start_gpu_encode(obs_encoder_t *encoder);
 extern void stop_gpu_encode(obs_encoder_t *encoder);
 
-extern bool do_encode(struct obs_encoder *encoder, struct encoder_frame *frame);
+extern bool do_encode(struct obs_encoder *encoder, struct encoder_frame *frame,
+		      const uint64_t *frame_cts);
 extern void send_off_encoder_packet(obs_encoder_t *encoder, bool success,
 				    bool received, struct encoder_packet *pkt);
 
--- a/libobs/obs-internal.h
+++ b/libobs/obs-internal.h
@@ -74,6 +74,12 @@ struct rendered_callback {
 	void *param;
 };
 
+struct packet_callback {
+	void (*packet_cb)(obs_output_t *output, struct encoder_packet *pkt,
+			  struct encoder_packet_time *pkt_time, void *param);
+	void *param;
+};
+
 /* ------------------------------------------------------------------------- */
 /* validity checks */
 
@@ -1189,7 +1195,12 @@ struct obs_output {
 	// captions are output per track
 	struct caption_track_data *caption_tracks[MAX_OUTPUT_VIDEO_ENCODERS];
 
-	DARRAY(struct encoder_packet_time) encoder_packet_times[MAX_OUTPUT_VIDEO_ENCODERS];
+	DARRAY(struct encoder_packet_time)
+	encoder_packet_times[MAX_OUTPUT_VIDEO_ENCODERS];
+
+	/* Packet callbacks */
+	pthread_mutex_t pkt_callbacks_mutex;
+	DARRAY(struct packet_callback) pkt_callbacks;
 
 	bool valid;
 
--- a/libobs/obs-output.c
+++ b/libobs/obs-output.c
@@ -18,6 +18,7 @@
 #include <inttypes.h>
 #include "util/platform.h"
 #include "util/util_uint64.h"
+#include "util/array-serializer.h"
 #include "graphics/math-extra.h"
 #include "obs.h"
 #include "obs-internal.h"
@@ -308,6 +309,9 @@ void obs_output_destroy(obs_output_t *output)
 
 		da_free(output->keyframe_group_tracking);
 
+		for (size_t i = 0; i < MAX_OUTPUT_VIDEO_ENCODERS; i++)
+			da_free(output->encoder_packet_times[i]);
+
 		clear_raw_audio_buffers(output);
 
 		os_event_destroy(output->stopping_event);
@@ -365,6 +369,7 @@ bool obs_output_actual_start(obs_output_t *output)
 		deque_init(&ctrack->caption_data);
 		pthread_mutex_unlock(&ctrack->caption_mutex);
 	}
+
 	return success;
 }
 
@@ -1476,8 +1481,10 @@ static inline void check_received(struct obs_output *output,
 	}
 }
 
-static inline void apply_interleaved_packet_offset(struct obs_output *output,
-						   struct encoder_packet *out)
+static inline void
+apply_interleaved_packet_offset(struct obs_output *output,
+				struct encoder_packet *out,
+				struct encoder_packet_time *packet_time)
 {
 	int64_t offset;
 
@@ -1491,6 +1498,8 @@ static inline void apply_interleaved_packet_offset(struct obs_output *output,
 
 	out->dts -= offset;
 	out->pts -= offset;
+	if (packet_time)
+		packet_time->pts -= offset;
 
 	/* convert the newly adjusted dts to relative dts time to ensure proper
 	 * interleaving.  if we're using an audio encoder that's already been
@@ -1526,7 +1535,7 @@ static size_t extract_itut_t35_buffer_from_sei(sei_t *sei, uint8_t **data_out)
 	if (!sei || !sei->head) {
 		return 0;
 	}
-	/* We should only need to get one payload, because the SEI that was 
+	/* We should only need to get one payload, because the SEI that was
 	 * generated should only have one message, so no need to iterate. If
 	 * we did iterate, we would need to generate multiple OBUs. */
 	sei_message_t *msg = sei_message_head(sei);
@@ -1712,8 +1721,8 @@ static bool add_caption(struct obs_output *output, struct encoder_packet *out)
 			uint8_t *obu_buffer = NULL;
 			size_t obu_buffer_size = 0;
 			size = extract_itut_t35_buffer_from_sei(&sei, &data);
-			metadata_obu_itu_t35(data, size, &obu_buffer,
-					     &obu_buffer_size);
+			metadata_obu_itut_t35(data, size, &obu_buffer,
+					      &obu_buffer_size);
 			if (obu_buffer) {
 				da_push_back_array(out_data, obu_buffer,
 						   obu_buffer_size);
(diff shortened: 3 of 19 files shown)


alinsavix:
I'm curious, does/will this information get passed through to the client side (of twitch, at least)? Because that would be really useful.

(For bonus points, make it so the timestamps represent the time the associated frame was created, rather than when it was output from the encoder, which would not only make them even more useful, but also measure encoder delay as part of the end-to-end measurement. At least, I don't *think* that's what it's doing right now)

libobs/obs-av1.h
```diff
@@ -30,9 +39,9 @@ EXPORT void obs_extract_av1_headers(const uint8_t *packet, size_t size,
 				    size_t *new_packet_size,
 				    uint8_t **header_data, size_t *header_size);
 
-EXPORT void metadata_obu_itu_t35(const uint8_t *itut_t35_buffer,
```
kkartaltepe, line 33: This symbol is public, please preserve the old function it can be implemented with the new function trivially.

libobs/obs-output.h
```diff
@@ -44,6 +45,9 @@ struct obs_output_info {
 
 	uint32_t flags;
 
+	/* Broadcast Performance Metrics control */
+	bool enable_bpm;
+
```
tytan652, line 50: Info structure should be considered non-mutable once registered.

```suggestion
```

libobs/obs-output.c
```diff
@@ -1340,6 +1397,24 @@ void obs_output_set_audio_conversion(
...
+void obs_output_enable_bpm(obs_output_t *output, bool bpm_enable)
+{
+	if (!obs_output_valid(output, "obs_output_enable_bpm"))
+		return;
+	if (!log_flag_service(output, __FUNCTION__))
+		return;
+	if (active(output))
+		return;
+
+	os_atomic_set_bool(&output->info.enable_bpm, bpm_enable);
```
tytan652, line 1410: You should not modify an info structure.

tytan652 (changes requested):

koolscooby:
> I'm curious, does/will this information get passed through to the client side (of twitch, at least)? Because that would be really useful.

Short answer: yes. Long answer, a few points!

* Yes, the embedded SEIs do get passed through transmuxing to the delivered video segments on Twitch. 
* However, we haven't decided exactly if/how we will intentionally expose it to either broadcasters or viewers; we are considering ideas like putting it in Twitch Inspector or in "Video Stats for Nerds" in the player. First, we have to validate that the data pipeline works at scale and that we trust the data to expose it to end users. 
* In the interim, this data is absolutely crucial for us to continue iterating on Automatic Stream Configuration settings (e.g. deploying HEVC or 1440p/4k ladders.) Without it, we can't safely iterate and change client settings (because we don't want to break users' performance.)

libobs/obs-output.h
lexano-ivs: Thanks for suggestion tytan652. If the `bool enable_bpm` field is removed, should I use the `flags` field in `struct obs_output_info` instead to store the BPM enable bit?

libobs/obs-av1.h
lexano-ivs: Thank you @kkartaltepe. I'll fix this up. Appreciate your review.

norihiro:
The title of the commit fea442ebd should start from `obs-webrtc: ` instead of `WHIP: `.

tt2468:
A few things:
- I would highly prefer that we include a millisecond-accurate signed 64 bit integer UNIX epoch timestamp, instead of an RFC3339 string.
- Would it be possible to split out BPM from libobs, and instead add functionality to add private SEI payloads for encoders? Something along the lines of `obs_encoder_add_private_sei(obs_encoder_t *, void *buf, size_t size)`. The location of the BPM code could probably be in the UI, running off of a QTimer to trigger.

libobs/obs-av1.h
lexano-ivs: I've update this with the latest git push. I put the original export back and wrapped the call to the new `metadata_obu()` function.

libobs/obs-output.h
lexano-ivs: @tytan652 I moved the `enable_bpm` boolean to the `obs_output` struct instead of the `obs_output_info` struct and adjusted references accordingly. Please let me know if this make sense.

libobs/obs-output.c
lexano-ivs: Thank you, I understand now what the info structure is meant to do now. I've reworked this, hopefully it's fine now.

lexano-ivs:
> The title of the commit [fea442e](https://github.com/obsproject/obs-studio/commit/fea442ebdfb63d7dbf787fa5a58ee8059f284131) should start from `obs-webrtc: ` instead of `WHIP: `.

Thank you @norihiro. I've updated the commit with the correct name.

lexano-ivs:
> A few things:
> 
>     * I would highly prefer that we include a millisecond-accurate signed 64 bit integer UNIX epoch timestamp, instead of an RFC3339 string.

In the [integration guide](https://d50yg09cghihd.cloudfront.net/other/20240718-MultitrackVideoIntegrationGuide.pdf), on page 28, we defined a type system for timestamp data. `timestamp_type==2` is the `duration_since_epoch_ts`, as well as a `delta_ts` format to signal differences. At the moment we're using type 1, which is the RFC3339 format, mainly because it was the best way for us to get started in the beta testing because the ingest servers were already speaking RFC3339. We can add type 2 and type 3 formats in the future, and any parser would have to respect them. Adding type 2 and/or type 3 (or any other types we come up with) is on the backlog, just simply not scoped/planned right at the moment.

> 
>     * Would it be possible to split out BPM from libobs, and instead add functionality to add private SEI payloads for encoders? Something along the lines of `obs_encoder_add_private_sei(obs_encoder_t *, void *buf, size_t size)`. The location of the BPM code could probably be in the UI, running off of a QTimer to trigger.

I need to think about this a bit more. Initial thought revolves around trigger. The current implementation is triggered from the `send_interleaved()` function because it was the best spot I could find to identify IDR frames to align with. We need the data aligned to IDR frames, so moving the trigger to an async QTimer seems like it might introduce jitter between the IDR time and the QTimer firing, with the assumption the we want tight (preferably synchronous) alignment.

libobs/obs-output.h
tytan652: > Please let me know if this make sense.

Implementing this inside of libobs does not make sense at all for me since it's using SEI metadata to send custom information, but I needed second opinions which @tt2468 gave.

tt2468 (changes requested):
I would like to suggest that we implement a packet callback for outputs. This callback would be triggered for every video packet, closely before being passed to the output to be transmitted. The BPM SEI rendering code can be moved to the `common` folder in the base of this repository, similar to happy eyeballs and media-playback.

Here some suggested APIs:
- Typedefs
  - `struct array_output_data *(*obs_output_packet_callback_t)(obs_output_t *, struct encoder_packet *pkt, struct encoder_packet_timing *timing, void *)`
- Functions
  - `obs_output_add_packet_callback(obs_output_t *output, obs_packet_callback_t callback, void *data);
  - `obs_output_add_packet_remove(obs_output_t *output, obs_packet_callback_t callback, void *data);`

It is important to note that this callback is implementer per-output, so the primary buffer in the packet is going to likely need to be duplicated before libobs inserts the SEI payload, to prevent SEI data from being sent to other outputs subscribed to the same encoders.

lexano-ivs:
> I would like to suggest that we implement a packet callback for outputs. This callback would be triggered for every video packet, closely before being passed to the output to be transmitted. The BPM SEI rendering code can be moved to the `common` folder in the base of this repository, similar to happy eyeballs and media-playback.
> 
> Here some suggested APIs:
> 
>     * Typedefs
>       
>       * `struct array_output_data *(*obs_output_packet_callback_t)(obs_output_t *, struct encoder_packet *pkt, struct encoder_packet_timing *timing, void *)`
> 
>     * Functions
>       
>       * `obs_output_add_packet_callback(obs_output_t *output, obs_packet_callback_t callback, void *data);
>       * `obs_output_add_packet_remove(obs_output_t *output, obs_packet_callback_t callback, void *data);`
> 
> 
> It is important to note that this callback is implementer per-output, so the primary buffer in the packet is going to likely need to be duplicated before libobs inserts the SEI payload, to prevent SEI data from being sent to other outputs subscribed to the same encoders.

Thanks for the suggestions. The "BPM SEI rendering code" can be thought of as 2 parts: 1. The calculations of the frame counters and timing deltas (once we implement delta time); and 2. The rendering of this data into SEI payload syntax. Are you envisioning moving both parts 1 and 2 into the callback, or just part 2? I'm leaning towards both parts should be handled in the callback function, but wanted to understand what you were thinking as well.

Also, if we employ this approach, the closed caption SEI could be supported with the same callback technique. I'd like to keep the migration of the current closed caption code to the potentially new callback mechanism as a separate workstream, and focus purely on the BPM support for this PR. Is this OK with you? 

I'd also like to understand if this packet callback should be available for use by any plugin, or would this be only for "internal" plugins (for lack of a better term) compiled natively with OBS? I believe it's the former (any plugin could use it), but would like to know for sure.

libobs/obs-internal.h
```diff
@@ -1089,6 +1090,66 @@ struct caption_track_data {
...
 };
 
+struct counter_data {
+	uint32_t diff;
+	uint32_t ref;
+	uint32_t curr;
+};
+
+// Broadcast Performance Metrics frame timing
+struct bpm_frame_time {
```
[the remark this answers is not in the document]
lexano-ivs: The idea of not calling this `bpm_frame_time` seems like a good idea. When I implemented it, it was part of the overall BPM work and hadn't realized it might be useful for more generic uses, so this makes sense. `encoder_packet_timing` is a good name, however some of the timing info is related to the uncompressed frames, so it's not totally accurate. The frame times straddle the uncompressed frames and compressed packets, so it's a bit tricky to name. I'll think about this a bit more, and we may end up using `encoder_packet_timing`.

libobs/obs-internal.h
```diff
@@ -1089,6 +1090,66 @@ struct caption_track_data {
...
+	/* PTS used to associate uncompressed frames with encoded packets. */
+	int64_t pts;
+	/* Composition timestamp is when the frame was rendered. */
+	uint64_t cts;
```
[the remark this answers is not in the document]
lexano-ivs: Will do, thanks.

libobs/obs-internal.h
```diff
@@ -1183,6 +1244,14 @@ struct obs_output {
 	// captions are output per track
 	struct caption_track_data *caption_tracks[MAX_OUTPUT_VIDEO_ENCODERS];
 
+	/* Broadcast Performance Metrics control */
+	bool enable_bpm;
```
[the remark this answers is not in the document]
lexano-ivs: Overall I like the idea of removing the `enable_bpm` flag. For the frame counters in BPM, this makes sense to make the logic implicit based on whether or not callbacks are registered. For the frame timing, specifically managing the entries in the BPM frame timing DARRAY, there might need to be a bit more intelligence to ensure we don't keep filling it up, and also we keep the required entries in the DARRAY long enough to consume them. 

I'll work on this with the goal towards eliminating the `enable_bpm` flag.

lexano-ivs:
With the latest updates, frame timing is now propagated from the encoder array to each output array. This allows the frame timing to work in with multiple output services in tandem. The initial implementation would drain the encoder frame timing array after the first consumption of the data.

I'm moving the PR to "Ready for review", and will work on renaming `bpm_frame_time` to `encoder_packet_timing` (or something along these lines).

tt2468:
> Thanks for the suggestions. The "BPM SEI rendering code" can be thought of as 2 parts: 1. The calculations of the frame counters and timing deltas (once we implement delta time); and 2. The rendering of this data into SEI payload syntax. Are you envisioning moving both parts 1 and 2 into the callback, or just part 2? I'm leaning towards both parts should be handled in the callback function, but wanted to understand what you were thinking as well.

Uhh I think both parts would be handled in the callback? I don't know if the frame counters and timing deltas are before or after the `encoder_packet_timing` struct finalization.

> Also, if we employ this approach, the closed caption SEI could be supported with the same callback technique. I'd like to keep the migration of the current closed caption code to the potentially new callback mechanism as a separate workstream, and focus purely on the BPM support for this PR. Is this OK with you?

It is definitely not necessary to do any immediate migration of the captions code to this callback, just as long as it's kept in mind.

> I'd also like to understand if this packet callback should be available for use by any plugin, or would this be only for "internal" plugins (for lack of a better term) compiled natively with OBS? I believe it's the former (any plugin could use it), but would like to know for sure.

As with other libobs APIs of similar design, this would be available to any code which has a copy of the encoder's pointer. That is usually the same code that created the encoder, but there are places where a third party plugin could fetch an encoder it is not managing, like through the frontend API. Realistically, if you register a callback of your own on this API, you would be expected to disconnect the callback as a part of your own cleanup procedure, or it would be disconnected by libobs if the encoder is destroyed.

lexano-ivs:
@tt2468 @RytoEX I've pushed a series of commits earlier today that should satisfy the requested changes:

1. Renamed `bpm_frame_time` to `encoder_packet_time`
2. `encoder_packet_time` handling remains within libobs and flows from encoding functions (stored in the encoder_t packet_times array) to the output_t packet_times arrays. 
3. BPM code has been completely removed from libobs, and moved to `deps/bpm`
4. The `enable_bpm` flag has been removed, along with the public API.
5. A packet callback mechanism has been created. Registered packet callbacks are invoked from a loop in `send_interleaved()`.
6. BPM is the first user of the new packet callback mechanism. `bpm_injection()` is introduced at the callback function as well as `bpm_destroy()` to deallocate. `bpm.h` is the public interface to BPM and simply contains the 2 function signatures.
7. BPM is enabled in multitrack video by using the new packet callback add/remove functions.
8. I've disabled BPM in WebRTC for the moment as it needs more testing with the new mechanism.

We can squash the new commits into old ones, and/or rework the branch as needed. I'd like to get an initial review first as I don't want to rewrite the branch history until the new code seems OK. I'll definitely keep a local copy of the original branch though for safekeeping before squashing anything.

tt2468 (commented):
Just a reminder that we'll want to also update the docs in this PR with the new methods. You may also want to include a note that any reallocation of the packet's buffer MUST be done with a util that allocates with bmem, otherwise a memory leak may occur.

We will want to verify that the packet struct is copied appropriately and that no memory leaks are occuring. I assume you've already done that, though.

plugins/obs-outputs/CMakeLists.txt
```diff
@@ -9,6 +9,10 @@ if(NOT TARGET happy-eyeballs)
   add_subdirectory("${CMAKE_SOURCE_DIR}/deps/happy-eyeballs" "${CMAKE_BINARY_DIR}/deps/happy-eyeballs")
 endif()
 
+if(NOT TARGET bpm)
+  add_subdirectory("${CMAKE_SOURCE_DIR}/deps/bpm" "${CMAKE_BINARY_DIR}/deps/bpm")
+endif()
+
```
tytan652, line 15:
```suggestion
```

plugins/obs-outputs/CMakeLists.txt
```diff
@@ -65,6 +69,7 @@ target_link_libraries(
   PRIVATE OBS::libobs
           OBS::happy-eyeballs
           OBS::opts-parser
+          OBS::bpm
```
tytan652, line 72:
```suggestion
```

tytan652 (changes requested):
Unused

plugins/obs-outputs/CMakeLists.txt
lexano-ivs: Thanks @tytan652. I just force pushed an update that moves this to the UI/CMakeLists.txt file.

deps/CMakeLists.txt
```diff
@@ -4,6 +4,7 @@ if(OS_WINDOWS)
 endif()
 
 add_subdirectory(blake2)
+add_subdirectory(bpm)
```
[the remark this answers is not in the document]
lexano-ivs: @tt2468 The current branch point for this PR doesn't have a "shared" directory. I'm going to rebase shortly with master, fix conflicts (if any), then move the BPM stuff to "shared" instead of "deps".

lexano-ivs:
@tt2468 I've rebased against master as of this morning, fixed conflicts, moved from deps/bpm to shared/bpm, and added a commit with documentation for the add/remove packet callback functions. This should be close to complete now, and if it looks OK, I can squash commits to make it prettier. 

Regarding memory leaks, I have been running with debug mode and checking for the num of memory leaks, which is 0, so I'm confident there's no issue there. 

I don't understand why the "gersemi" validation is complaining about the format of the CMakeLists.txt files and could use some guidance on what might be wrong.

kkartaltepe:
> I don't understand why the "gersemi" validation is complaining about the format of the CMakeLists.txt files and could use some guidance on what might be wrong.

Gersemi si the new cmake formatter being used in CI, you should change from using cmake-format to https://github.com/BlankSpruce/gersemi to keep your files formatted.

shared/bpm/bpm.h
```diff
@@ -0,0 +1,26 @@
...
+ * with the first invokation of the callback associated with the output.
+ * Deallocation must be done explicitly with bpm_destroy(), after the
+ * callback is removed.
+ *
+ * BPM is designed to operate at the packet level. The bpm_injection()
+ * callback function must be registered and unregistered with
+ * obs_output_add_packet_callback() and obs_output_remove_packet_callback(),
+ * respectively.
+ */
+void bpm_injection(obs_output_t *output, struct encoder_packet *pkt,
```
tt2468, line 17: Microscopic nitpick: Should this perhaps be named `bpm_inject` to match the tense of `bpm_destroy`?

lexano-ivs: I've renamed it to `bpm_inject()` just now. Also added a note about memset() to the documentation on `obs_output_add_packet_callback()`. 

Next step is squashing commits, which I'll start shortly.

lexano-ivs:
@tt2468 I've squashed the commits now. Much cleaner. Please review and let me know next steps, if any.

shared/bpm/bpm.h
RytoEX: This feedback seems to have been addressed.

RytoEX (approved):
The parts I reviewed seem fine. Builds and runs on Windows, and confirmed that the functionality is working. This has gone through multiple rounds of review and design revision, and it's as good as it's likely to get before merging.

PR #11006, opened 2024-07-19
mac-virtualcam: Update NotInstalled error text for macOS 15

jcm93:
### Description

Updates the error message for an uninstalled/not-yet-approved camera system extension on macOS 15.

### Motivation and Context

Apple moved where the user approves system extensions away from "Privacy & Security" to "General -> Login Items & Extensions". Don't want to provide outdated/incorrect directions. Beta users have already arrived confused in support about the old message.

### How Has This Been Tested?

Tested in an offline branch with a custom signing configuration to allow for the installation of system extensions, then relevant changes manually copied to this branch but not technically tested. Should be double checked.

### Types of changes

### Checklist:

### DIFF
--- a/plugins/mac-virtualcam/src/obs-plugin/plugin-main.mm
+++ b/plugins/mac-virtualcam/src/obs-plugin/plugin-main.mm
@@ -306,7 +306,12 @@ static bool virtualcam_output_start(void *data)
                                                delegate.lastErrorMessage]
                         .UTF8String);
             } else {
-                obs_output_set_last_error(vcam->output, obs_module_text("Error.SystemExtension.NotInstalled"));
+                if (@available(macOS 15.0, *)) {
+                    obs_output_set_last_error(vcam->output,
+                                              obs_module_text("Error.SystemExtension.NotInstalled.MacOS15"));
+                } else {
+                    obs_output_set_last_error(vcam->output, obs_module_text("Error.SystemExtension.NotInstalled"));
+                }
             }
 
             return false;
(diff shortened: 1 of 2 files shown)


plugins/mac-virtualcam/src/obs-plugin/plugin-main.mm
```diff
@@ -306,7 +306,12 @@ static bool virtualcam_output_start(void *data)
                                                delegate.lastErrorMessage]
                         .UTF8String);
             } else {
-                obs_output_set_last_error(vcam->output, obs_module_text("Error.SystemExtension.NotInstalled"));
+                if (@available(macOS 15.0, *)) {
```
PatTheMav, line 309: I could be wrong but I think this will only ever be true once we compile against the macOS 15 platform SDK.

We'll probably upgrade the used Xcode version on CI as well as the platform SDK once they become available on GitHub Actions after macOS 15 is about to release.

PatTheMav (approved):
Should be fine.

On the whole I'd rather present a dialog box with a specific button that will take you to the corresponding settings dialog instead of using a breadcrumb string which is not particularly easy to parse.

plugins/mac-virtualcam/src/obs-plugin/plugin-main.mm
RytoEX: This feedback still needs to be resolved.

jcm93: I just checked a CI build in another branch with an `@available(macOS 15.0, *)` check built on Xcode 15.2, and it evaluated to true on my macOS 15 system. So this check does not actually require the macOS 15 SDK to function as intended. I'm comfortable resolving this given that info.

PR #10999, opened 2024-07-17
obs-outputs: Set FLV DTS offset based on first audio or video packet

derrod:
### Description

Ensures that PTS/DTS for audio start at 0 if there by ensuring the DTS offset is based on the first packet that comes in, regardless of whether or not it's video (b-frame delay), or audio (priming sample delay).

### Motivation and Context

Fixes #10998

This became necessary due to #10689 and #10690 which fix handling of audio priming samples with negative timestamps. Note that if users have a sufficiently large number of b-frames configured (1 is enough) this currently doesn't actually cause any issues.

### How Has This Been Tested?

Logged output packets and ensured the resulting FLV files are playable.

### Types of changes

- Bug fix (non-breaking change which fixes an issue)

### Checklist:

- [x] My code has been run through [clang-format](https://github.com/obsproject/obs-studio/blob/master/.clang-format).
- [x] I have read the [**contributing** document](https://github.com/obsproject/obs-studio/blob/master/CONTRIBUTING.rst).
- [x] My code is not on the master branch.
- [x] The code has been tested.
- [x] All commit messages are properly formatted and commits squashed where appropriate.
- [x] I have included updates to all appropriate documentation.

PR #11013, opened 2024-07-20
obs-outputs: Preserve signedness of FLV timestamp

jeongki-kim:
### Description
obs-outputs: Preserve signedness of FLV timestamp

### Motivation and Context
[Video File Format Specification Version 10](https://rtmp.veriskope.com/pdf/video_file_format_spec_v10.pdf) states that a timestamp in FLV format is a signed 32-bit integer, which are composed of two integers.

> |Field|Type|Comment|
> |-|-|-|
> |...|...|...|
> |Timestamp|UI24|Time in milliseconds at which the data in this tag applies. This value is relative to the first tag in the FLV file, which always has a timestamp of 0.|
> |TimestampExtended|UI8|Extension of the Timestamp fields to form a **SI32** value. This field represents the upper 8 bits, while the previous Timestamp field represents the lower 24 bits of the time in milliseconds.|
> |...|...|...|

### How Has This Been Tested?
Inspected RTMP packets with Wireshark.

### Types of changes

### Checklist:

### DIFF
--- a/plugins/obs-outputs/flv-mux.c
+++ b/plugins/obs-outputs/flv-mux.c
@@ -130,7 +130,7 @@ static void s_wstring(struct serializer *s, const char *str)
 static inline void s_wtimestamp(struct serializer *s, int32_t i32)
 {
 	s_wb24(s, (uint32_t)(i32 & 0xFFFFFF));
-	s_w8(s, (uint32_t)(i32 >> 24) & 0x7F);
+	s_w8(s, (uint32_t)(i32 >> 24) & 0xFF);
 }
 
 static inline double encoder_bitrate(obs_encoder_t *encoder)
@@ -340,7 +340,7 @@ static void flv_video(struct serializer *s, int32_t dts_offset,
 
 	s_wb24(s, (uint32_t)packet->size + 5);
 	s_wb24(s, (uint32_t)time_ms);
-	s_w8(s, (time_ms >> 24) & 0x7F);
+	s_w8(s, (time_ms >> 24) & 0xFF);
 	s_wb24(s, 0);
 
 	/* these are the 5 extra bytes mentioned above */
@@ -373,7 +373,7 @@ static void flv_audio(struct serializer *s, int32_t dts_offset,
 
 	s_wb24(s, (uint32_t)packet->size + 2);
 	s_wb24(s, (uint32_t)time_ms);
-	s_w8(s, (time_ms >> 24) & 0x7F);
+	s_w8(s, (time_ms >> 24) & 0xFF);
 	s_wb24(s, 0);
 
 	/* these are the two extra bytes mentioned above */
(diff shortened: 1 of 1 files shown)


derrod:
The timestamp in RTMP is unsigned, so this would result in invalid or out-of-order timestamps being sent. Based on advice from one of the authors of the Enhanced FLV/RTMP spec, we should never sent negative  or non-monotonic timestamps (he also believes the `SI32` in the legacy FLV spec might be a mistake).

Since we already start the presentation at a non-0 time to compensate for b-frame delay, it follows that doing the same for audio encoder delay is probably the right approach, hence #10999.

jeongki-kim:
As RTMP timestamps are unsigned like you said, and they are wrapped around when they overflows, the most significant bit in the timestamp value no longer represents signedness. Also discarding the MSB results the timestamp to roll over every 2,147,483,647 miliseconds, while the RTMP specification says it rolls over every 4,294,967,295 milliseconds.

jeongki-kim:
Even if OBS is going to count RTMP timestamp from zero when it starts streaming, there still might be a slight possibility that one might keep streaming longer than about 50 days, which the timestamp will overflow eventually.

Adobe RTMP Specification([HTML ver](https://rtmp.veriskope.com/docs/spec), [PDF ver](https://rtmp.veriskope.com/pdf/rtmp_specification_1.0.pdf)) covers such situation:
> Because timestamps are 32 bits long, they roll over every 49 days, 17 hours, 2 minutes and 47.296 seconds. Because streams are allowed to run continuously, potentially for years on end, an RTMP application SHOULD use serial number arithmetic [RFC1982] when processing timestamps, and SHOULD be capable of handling wraparound. For example, an application assumes that all adjacent timestamps are within 2^31 - 1 milliseconds of each other, so 10000 comes after 4000000000, and 3000000000 comes before 4000000000.

Also I suspect the following code of librtmp will not work as expected, if the MSB is dropped.
https://github.com/obsproject/obs-studio/blob/2fa77c402149c0bb797b5838e5fc6a7391e6a86d/plugins/obs-outputs/librtmp/rtmp.c#L4134

RytoEX:
With #10999 merged, is this still needed?

jeongki-kim:
I think so because this is not only related to the initial timestamp, which was already handled by #10999, but also covers RTMP timestamp overflow.

Discarding the MSB, which is the identical bit that represents signedness in case of signed integers, limits its maximum value up to (2^31)-1 (==0x7FFF FFFF), and therefore results the timestamp resets to 0 after 0x7FFF FFFF. This behavior does not comply RTMP specification as it specifies that 0 comes after 0xFFFF FFFF.

Only few users would be affected because it matters when they streams over 0x7FFF FFFF miliseconds or about 24 days approximately. If you don't think it counts, this PR can be closed without merging.

PR #11030, opened 2024-07-26
win-update: Use correct winhttp proxy type

eebssk1:
### Description

The legacy type may not work on some setups, use new one on 8.1+  instead.

### Motivation and Context

Updater does not go through my system proxy.
The legacy type seems do not check system proxy settings throughly.

### How Has This Been Tested?

I haven't test compile the changes,but should not break since it's a rather simple change.

### Types of changes

Tweak

### Checklist:

### DIFF
--- a/UI/win-update/updater/http.cpp
+++ b/UI/win-update/updater/http.cpp
@@ -76,7 +76,7 @@ bool HTTPPostData(const wchar_t *url, const BYTE *data, int dataLen,
 	 * connect to server                      */
 
 	hSession = WinHttpOpen(L"OBS Studio Updater/3.0",
-			       WINHTTP_ACCESS_TYPE_DEFAULT_PROXY,
+			       WINHTTP_ACCESS_TYPE_AUTOMATIC_PROXY,
 			       WINHTTP_NO_PROXY_NAME, WINHTTP_NO_PROXY_BYPASS,
 			       0);
 	if (!hSession) {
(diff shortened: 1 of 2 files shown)


derrod (changes requested):
We don't support anything below Windows 10 so you can just replace the constant, no need to overcomplicate things.

derrod (approved):

RytoEX:
> We don't support anything below Windows 10 so you can just replace the constant, no need to overcomplicate things.

The updater does run on Windows 7/8/8.1, but bails out pretty early.
https://github.com/obsproject/obs-studio/blob/a9082de880adaca7d10706b3d750c9ebfddf1e2d/UI/win-update/updater/updater.cpp#L1925-L1935

I guess in this case, since these paths are never reached on those older Windows versions, the change is safe?

derrod:
> > We don't support anything below Windows 10 so you can just replace the constant, no need to overcomplicate things.
> 
> The updater does run on Windows 7/8/8.1, but bails out pretty early.

Yeah but we don't support it, hence the bailout.

> I guess in this case, since these paths are never reached on those older Windows versions, the change is safe?

Yes. Even if it wasn't, I couldn't care less if it crashes on unsupported Windows versions.

norihiro:
Just a nitpick,
The commit a9082de88 has a 2nd line which is not empty.

RytoEX (changes requested):
Do not use merge commits to catch up on git history. Use `git rebase`.

eebssk1:
> Do not use merge commits to catch up on git history. Use `git rebase`.

done

PR #11031, opened 2024-07-26
UI: Don't calculate physical screen size for Wayland

jrelvas-ipc:
### Description
The UI code assumes QScreen.size is in the logical screen resolution space. This appears to be true with X11 (and probably Win32/MacOS as well), but not with Wayland sessions. Qscreen.size is the physical screen resolution on Wayland.

This PR skips physical resolution conversion when on a Wayland backend.

### Motivation and Context
Currently, the screen resolution found in the settings UI will **not** match the actual resolution on Wayland if DPI scaling is used.

imagem

This PR fixes this issue, making the Canvas resolution match the user's actual screen resolution in the settings UI and auto-configuration wizard:

imagem

### How Has This Been Tested?

Environment:
- OS: Fedora Workstation 41 (Rawhide)
- Machine: Thinkpad P1 Gen 6
  - CPU: i7-13800h
  - Graphics: Intel Gen12 Graphics (Intel Xe igpu) + Nvidia RTX 4000 Ada (Laptop) [Nvidia 560 open]
  - Desktop: GNOME 47 (Alpha)
  - Display: 3840x2400 (16:10)

I've tested the Wayland and X11 backends. On X11, there's no changes with the screen resolution option (which was already correct). On Wayland, the screen resolution now matches the real resolution.

Steps: 
1. Set the screen's DPI to any value other than 1.
2. Launch OBS studio (in Wayland)
3. Check the "Output" tab in the Settings UI - Ensure the first Canvas resolution in the list is equal to the screen's resolution.
4. Same step as above, but in the Auto-configuration wizard.

Since this only adds a single conditional check for Wayland, this change should be isolated to Linux Desktop Wayland targets. X11, Win32 and MacOS should not be affected. 

### Types of changes

### Checklist:


tytan652:
Please, guard platform name check behind `#if !defined(__APPLE__) && !defined(_WIN32)`.

jrelvas-ipc:
Like this @tytan652?

jrelvas-ipc:
Better not merge this for now - this doesn't work at all with fractional scaling :upside_down_face:

UI/window-basic-auto-config.cpp
```diff
@@ -174,8 +174,16 @@ AutoConfigVideoPage::AutoConfigVideoPage(QWidget *parent)
 
 		// Calculate physical screen resolution based on the virtual screen resolution
 		// They might differ if scaling is enabled, e.g. for HiDPI screens
-		as_width = round(as_width * screen->devicePixelRatio());
-		as_height = round(as_height * screen->devicePixelRatio());
+		// NOTE: On Wayland, QScreen.size is physical rather than logical, so skip it
+
+#if !defined(__APPLE__) && !defined(_WIN32)
+		if (!QApplication::platformName().contains("wayland"))
```
redstrate, line 180: In order to properly support fractional scales under Wayland (for the purposes of figuring out the true screen resolution) do this:

1. Create a new protected function (stick it above `private:`) for these two classes. You want to override `QWidget::event` like this: `bool event(QEvent *event) override;`
2. Create a definition for that function, and write it something like this:

```cpp
bool OBSBasicSettings::event(QEvent *event)
{
	// When the DPR changes, we must recalculate the list since we depend on it's value
	if (event->type() == QEvent::DevicePixelRatioChange) {
		LoadResolutionLists();
	}
	return QWidget::event(event);
}
```
3. Ensure that you only access `devicePixelRatio` through the QWidget surface, like `window()->devicePixelRatio()`. Qt currently doesn't provide the true DPR on Wayland to any `QScreen()` AFAIK.

Note that this is inherently troublesome because it seems OBS will add every resolution for every screen? But you can only get the DPR of the screen this window is on. This is something you'll have to discuss with the maintainers though.

tytan652: > 3\. …Qt currently doesn't provide the true DPR on Wayland to any `QScreen()` AFAIK.

@redstrate, can we consider it a Qt bug ?

redstrate: > > 3. …Qt currently doesn't provide the true DPR on Wayland to any `QScreen()` AFAIK.
> 
> @redstrate, can we consider it a Qt bug ?

I would say yes, consider it a Qt bug (more specifically, in the Wayland QPA.) I personally don't know the feasibility of pushing the correct DPR up to QScreen (there are abandoned attempts like https://codereview.qt-project.org/c/qt/qtwayland/+/563344) but someone on the Qt bugtracker might.

Until such a thing happens though, we would still need the workaround described in my comment. Qt just did it to fix broken WebEngine fractional scaling see https://codereview.qt-project.org/c/qt/qtwebengine/+/569910

tytan652: Unfortunately, `LoadResolutionLists` is meant to be ran only once while creating the settings windows.
I don't think making it in an event is a good idea.

tytan652: Before I forgot, thanks for the feedback.

davidedmundson:
QScreen::size *should* be in logical co-ordinates. There is a Gnome bug where that sometimes sends us incorrect logical sizes. Fractional scaling messes this up because when that's enabled Gnome does send the right value.

GTK ended up hacking round it, I made a patch in Qt to do the same, but it hasn't landed yet:  https://codereview.qt-project.org/c/qt/qtwayland/+/549372 

The other discussion about window::scale vs screen::scale is a red-herring here, as we already have API for the logical size.

norihiro:
The commit 138b70183 has a too long description in a line, 96 characters. It should be wrapped at 72 characters.
> When Wayland is the windowing backend, QScreen.size already represents the physical screen size,

Warchamp7:

PR #11043, opened 2024-07-29
Remove unused functions

norihiro:
### Description

This PR removes unused functions that were not exported, ie. functions declared with `static` qualifier.

This PR also adds `inline` qualifier to `static` functions defined in header files.
This will affect 3rd party plugins but it should not make something wrong.

text-freetype2: Some functions were unnecessarily declared in a header file. They were moved to a proper location.

libobs/graphics/effect-parser.c: Exceptionally,
I didn't remove `ep_getparam`
as it is a variant of `ep_get*` functions so that it might be used in future.

libobs/obs-source.c: The removed function `set_eparami` was introduced at e7f754df97 but later removed by bdd8d6405.

### Motivation and Context

Unused functions should not be left.

I want to enable `-Wunused-function` (but I gave up for now since qt automoc generates an unused function without `static inline` qualifier.)

For public header files, it's better to avoid `static` functions without `inline`
so that 3rd party plugins can enable `-Wunused-function`.

### How Has This Been Tested?

OS: Fedora 39

Added `-Wunused-function` to the compile option.
Ran a command below to temporarily remove `inline` qualifier.
```sh
for f in $(git grep -l 'inline' ':*.c*'); do
  ed $f <<EOF
\$
?#include
a
#define inline
.
wq
EOF
done
```
Ran build flow and checked errors.

### Types of changes

### Checklist:

### DIFF
--- a/UI/window-basic-settings-a11y.cpp
+++ b/UI/window-basic-settings-a11y.cpp
@@ -11,11 +11,6 @@ enum ColorPreset {
 	COLOR_PRESET_CUSTOM = 99,
 };
 
-static inline bool WidgetChanged(QWidget *widget)
-{
-	return widget->property("changed").toBool();
-}
-
 static inline QColor color_from_int(long long val)
 {
 	return QColor(val & 0xff, (val >> 8) & 0xff, (val >> 16) & 0xff,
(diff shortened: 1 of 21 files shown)


kkartaltepe (approved):

RytoEX (approved):
Looks okay at a glance. Would not mind more eyes on this.

Lain-B (approved):

PR #11052, opened 2024-07-31
shared/media-playback: Unref sw_frame before reuse if using hardware decoding

tt2468:
### Description
By not performing an unref on sw_frame before using it again, a memory "leak" was being created if the frame had side data. This removes a previously-added check that optionally unrefs sw_frame, and just does it every tick.

### Motivation and Context
Noticed a memory leak the other day when playing back an HDR media file. It was quite extreme (ticking up by 1MB every few seconds).

### How Has This Been Tested?
Memory usage no longer goes up over time, other behaviors that I could test still work as expected.

### Types of changes

### Checklist:

### DIFF
--- a/shared/media-playback/media-playback/decode.c
+++ b/shared/media-playback/media-playback/decode.c
@@ -337,16 +337,7 @@ static int decode_packet(struct mp_decode *d, int *got_frame)
 			return ret;
 		}
 
-		/* does not check for color format or other parameter changes which would require frame buffer realloc */
-		if (d->sw_frame->data[0] &&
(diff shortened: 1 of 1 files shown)


derrod (approved):
Probably less efficient since we have to allocate a new frame every time but arguably much easier than trying work around the issues that come with re-using frames.

tt2468:
For the record, I did verify that ffmpeg itself is doing a lot of frame unreffing every tick, so I felt confident that it's safe to use for every decoded frame here.

pkviet (approved):
Tested the change as well as an alternative. Works fine.

Lain-B (approved):

PR #9183, opened 2023-06-30
UI: Fix utf-8 paths in shared updater components

derrod:
### Description

Fixes reading/writing files with non-ASCII characters in their path in the shared code of the update framework.

### Motivation and Context

AAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAA

### How Has This Been Tested?

Compiled and ran on my machine. Still needs to be tested on an affected system.

### Types of changes

- Bug fix (non-breaking change which fixes an issue)

### Checklist:

- [x] My code has been run through [clang-format](https://github.com/obsproject/obs-studio/blob/master/.clang-format).
- [x] I have read the [**contributing** document](https://github.com/obsproject/obs-studio/blob/master/CONTRIBUTING.rst).
- [x] My code is not on the master branch.
- [x] The code has been tested.
- [x] All commit messages are properly formatted and commits squashed where appropriate.
- [x] I have included updates to all appropriate documentation.

PR #11055, opened 2024-07-31
UI: Implement backend work for configurable storage locations for settings, profiles, and scene collections

PatTheMav:
### Description
This PR implements backend changes required to make the storage locations of settings, profiles, and scene collections configurable and decoupled from OBS' current default storage location for its settings, log files, crash logs, and other generated files.

This is achieved in 3 major parts:

1. The configuration is split up into an "app" configuration (mainly for information that pertains to the work of the app itself, e.g. actual location of user configuration, last time the app has checked for updates, etc.) and a "user" configuration (mainly for settings that are exposed to the user and can be changed according to their preference).
2. Profile storage location is decoupled from any default path for settings and as such all code that uses the settings' path to access profile information has been changed
3. Scene collection storage location is decoupled from any default path for settings, like for profiles, code that uses the settings' path to access them has been changed

The implementation also introduces additional changes:

* Changes rely on C++17 functionality, such as availability of `std::string_view`, `std::filesystem` and others.
* Reliance of `using namespace std` was avoided as much as possible as it's a bad design pattern, when functionality from `std` was used, it's made explicit
* Use of pointers was avoided in favour of either `std::optional` (where it's use of copies were deemed good enough) or reference types to keep using value semantics as much as possible
* Some names were changed to make their function more explicit, e.g. `basicConfig` was renamed to `activeConfiguration` because it represents the currently active configuration object in use by the application (it's a state-dependent variable).
* In some cases (e.g. deletion of profiles or scene collections) copies are used on purpose as the underlying references are removed from the owning collection container
* Exception handling is favoured over returning single error values or checking for validity of modified variables as favoured by modern C++ practices.

Architecturally an attempt was made to introduce some encapsulation to the management of profiles and scene collections. Available scene collections or profiles are discovered on disk and cached in memory (with refreshes from disk happening only when explicitly requested). 

Handling of encoder configuration (which is stored in separate JSON files next to the profile settings file) has not been changed, but should be moved out of the encoder configuration and into the profile management, but was out of scope for this PR.

### Motivation and Context
The long-term goal of this work is to enable placing configuration files as well as profiles and scene collections in arbitrary locations provided by the user, such as:

* A network drive
* A vendor-specific directory that is automatically synced with cloud services (e.g. iCloud Drive)
* (Future) Non-file based storage (e.g. cloud storage APIs)

### How Has This Been Tested?
Tested on macOS 14 with the following scenarios:

* Launched OBS with existing configuration, settings were migrated to `user.ini`
* Launched OBS with missing scene collection, new collection was created in-place
* Launched OBS with missing profile, profile was created in-place
* Created/duplicated/removed/renamed profiles with and without "invalid" characters (i.e. emoji)
* Created/duplicated/removed/renamed scene collections with and without "invalid" characters
* Checked that UUIDs of sources in duplicated scene collections were correctly changed
* Checked that name of renamed/duplicated scene collections was correctly changed
* Checked that currently active profiles or scene collections were saved before any change took place

### Types of changes

### Checklist:

### DIFF
--- a/UI/obs-app-theming.cpp
+++ b/UI/obs-app-theming.cpp
@@ -409,7 +409,6 @@ static vector<OBSThemeVariable> ParseThemeVariables(const char *themeData)
 
 void OBSApp::FindThemes()
 {
-	string themeDir;
 
 	QStringList filters;
 	filters << "*.obt" // OBS Base Theme
@@ -417,18 +416,24 @@ void OBSApp::FindThemes()
 		<< "*.oha" // OBS High-contrast Adjustment layer
 		;
 
-	GetDataFilePath("themes/", themeDir);
-	QDirIterator it(QString::fromStdString(themeDir), filters, QDir::Files);
-	while (it.hasNext()) {
-		auto theme = ParseThemeMeta(it.next());
-		if (theme && !themes.contains(theme->id))
-			themes[theme->id] = std::move(*theme);
+	{
+		string themeDir;
+		GetDataFilePath("themes/", themeDir);
+		QDirIterator it(QString::fromStdString(themeDir), filters,
+				QDir::Files);
+		while (it.hasNext()) {
+			auto theme = ParseThemeMeta(it.next());
+			if (theme && !themes.contains(theme->id))
+				themes[theme->id] = std::move(*theme);
+		}
 	}
 
-	themeDir.resize(1024);
-	if (GetConfigPath(themeDir.data(), themeDir.capacity(),
-			  "obs-studio/themes/") > 0) {
-		QDirIterator it(QT_UTF8(themeDir.c_str()), filters,
+	{
+		const std::string themeDir =
+			App()->userConfigLocation.u8string() +
+			"/obs-studio/themes";
+
+		QDirIterator it(QString::fromStdString(themeDir), filters,
 				QDir::Files);
 
 		while (it.hasNext()) {
@@ -876,7 +881,8 @@ bool OBSApp::SetTheme(const QString &name)
 
 	filesystem::path debugOut;
 	char configPath[512];
-	if (GetConfigPath(configPath, sizeof(configPath), filename.c_str())) {
+	if (GetAppConfigPath(configPath, sizeof(configPath),
+			     filename.c_str())) {
 		debugOut = absolute(filesystem::u8path(configPath));
 		filesystem::create_directories(debugOut.parent_path());
 	}
@@ -940,7 +946,7 @@ bool OBSApp::InitTheme()
 	}
 
 	char userDir[512];
-	if (GetConfigPath(userDir, sizeof(userDir), "obs-studio/themes")) {
+	if (GetAppConfigPath(userDir, sizeof(userDir), "obs-studio/themes")) {
 		auto configSearchDir = filesystem::u8path(userDir);
 		QDir::addSearchPath("theme", absolute(configSearchDir));
 	}
@@ -948,7 +954,7 @@ bool OBSApp::InitTheme()
 	/* Load list of themes and read their metadata */
 	FindThemes();
 
-	if (config_get_bool(globalConfig, "Appearance", "AutoReload")) {
+	if (config_get_bool(userConfig, "Appearance", "AutoReload")) {
 		/* Set up Qt file watcher to automatically reload themes */
 		themeWatcher = new QFileSystemWatcher(this);
 		connect(themeWatcher.get(), &QFileSystemWatcher::fileChanged,
@@ -956,19 +962,19 @@ bool OBSApp::InitTheme()
 	}
 
 	/* Migrate old theme config key */
-	if (config_has_user_value(globalConfig, "General", "CurrentTheme3") &&
-	    !config_has_user_value(globalConfig, "Appearance", "Theme")) {
-		const char *old = config_get_string(globalConfig, "General",
+	if (config_has_user_value(userConfig, "General", "CurrentTheme3") &&
+	    !config_has_user_value(userConfig, "Appearance", "Theme")) {
+		const char *old = config_get_string(userConfig, "General",
 						    "CurrentTheme3");
 
 		if (themeMigrations.count(old)) {
-			config_set_string(globalConfig, "Appearance", "Theme",
+			config_set_string(userConfig, "Appearance", "Theme",
 					  themeMigrations[old].c_str());
 		}
 	}
 
 	QString themeName =
-		config_get_string(globalConfig, "Appearance", "Theme");
+		config_get_string(userConfig, "Appearance", "Theme");
 
 	if (themeName.isEmpty() || !GetTheme(themeName)) {
 		if (!themeName.isEmpty()) {
--- a/UI/obs-app.cpp
+++ b/UI/obs-app.cpp
@@ -23,6 +23,7 @@
 #include <string>
 #include <sstream>
 #include <mutex>
+#include <filesystem>
 #include <util/bmem.h>
 #include <util/dstr.hpp>
 #include <util/platform.h>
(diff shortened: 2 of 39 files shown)


UI/obs-frontend-api/obs-frontend-api.h
```diff
@@ -208,7 +208,9 @@ EXPORT obs_output_t *obs_frontend_get_recording_output(void);
 EXPORT obs_output_t *obs_frontend_get_replay_buffer_output(void);
 
 EXPORT config_t *obs_frontend_get_profile_config(void);
-EXPORT config_t *obs_frontend_get_global_config(void);
+EXPORT config_t *obs_frontend_get_global_config(void) OBS_DEPRECATED;
+EXPORT config_t *obs_frontend_get_app_config(void);
+EXPORT config_t *obs_frontend_get_user_config(void);
```
tytan652, line 213: This is something I wish we stop doing.
I think we should create another way for add-ons to put their settings and never have to touch a `config_t *`.

PatTheMav: I agree, but the PR is massive enough as it is and I tried to focus on one thing without rocking the boat too much.

E.g. with these changes it should be easier to encapsulate even further and implement something like `OBSConfig` or `OBSProfile` classes which are interacted with (so instead of `GetCurrentProfile` returning a struct, it returns a reference to a class instance) so instead of taking the profile's path and adding `recordEncoder.json` to it, you would do `OBSProfile::GetRecordEncoder()` or something.

But - again - would have been out of scope, but is definitely where we should want to end up.

tytan652: In my opinion adding new API is not a middle-ground, it just "worsen" to at least 2-3 major version.

PatTheMav:
Will require https://github.com/obsproject/obs-websocket/pull/1241 to be merged first to unbreak Linux compilation.

UI/obs-frontend-api/obs-frontend-api.h
PatTheMav: Just removing access to the configuration objects without offering an alternative is not an option either.

And - again - the defined scope for this change was to just make the locations of these things to be adjustable, not to change the way plugins interact with the settings system.

Such a change can be done, but would constitute a rewrite of large chunks of the application and would be stuck in review-hell for eternity.

tytan652: Feel free to mark it as resolve, I have no motivation.

UI/obs-app-theming.cpp
```diff
@@ -409,26 +409,29 @@ static vector<OBSThemeVariable> ParseThemeVariables(const char *themeData)
...
+			auto theme = ParseThemeMeta(it.next());
+			if (theme && !themes.contains(theme->id))
+				themes[theme->id] = std::move(*theme);
+		}
 	}
 
-	themeDir.resize(1024);
-	if (GetConfigPath(themeDir.data(), themeDir.capacity(),
-			  "obs-studio/themes/") > 0) {
-		QDirIterator it(QT_UTF8(themeDir.c_str()), filters,
```
RytoEX, line 431: IIRC, this was done this way to fix a specific bug with theme paths. We should confirm that the bug isn't reintroduced.

UI/window-basic-main.hpp
```diff
@@ -1331,6 +1330,106 @@ public slots:
...
+	void on_actionNewSceneCollection_triggered();
+	void on_actionDupSceneCollection_triggered();
+	void on_actionRenameSceneCollection_triggered();
+	void
+	on_actionRemoveSceneCollection_triggered(bool skipConfirmation = false);
+	void on_actionImportSceneCollection_triggered();
+	void on_actionExportSceneCollection_triggered();
+
+public slots:
+	bool NewSceneCollection(const QString &name);
```
RytoEX, line 1432: It's a little odd to me to start a slot/function name with `New`. I'd expect a verb, such as `Create` or `Add` or even `CreateNew` (redundant though it may be). That said, I see there are private functions that use `Create`, so I'm not sure what to do to both use a verb and avoid confusion.

RytoEX (commented):
Some quick general comments/questions.

UI/obs-app-theming.cpp
RytoEX: @derrod I think you submitted the fix here with `.resize(1024)`. Do you remember what the test case was?

derrod: That's just to ensure the buffer is big enough. Since the string is constructed directly here rather than being written by an external function to the underlying buffer that's fine.

UI/obs-app.hpp
```diff
@@ -154,7 +164,11 @@ private slots:
 
 	inline QMainWindow *GetMainWindow() const { return mainWindow.data(); }
 
-	inline config_t *GlobalConfig() const { return globalConfig; }
+	inline config_t *GetAppConfig() const { return appConfig; }
+	inline config_t *GetUserConfig() const { return userConfig; }
+	std::string userConfigLocation;
+	std::string userScenesLocation;
+	std::string userProfilesLocation;
```
derrod, line 171: Is there any particular reason not to use `std::filesystem::path` for these?

PatTheMav: Mainly because they are treated as strings throughout most of the codebase. In theory one could keep them as filesystem paths and always convert into `u8string` and into `c_str` from those, but that would create unnecessary overhead as those variables are more often used for string composition than actual filesystem operations (in contrast to the path objects used in the `OBSSceneCollection` and `OBSProfile` objects).

derrod (changes requested):
There are a number of naïve conversion from `std::string` to `std::filesystem::path`, those should all use `std::filesystem::u8path` to ensure proper conversion from UTF-8. Otherwise a native encoding is assumed that - mainly on Windows - may result in invalid paths if non-ASCII characters are used.

Also see #9183 and https://github.com/obsproject/obs-websocket/pull/1231

UI/obs-app.hpp
derrod: Just from a brief look through the diff it seems that those aren't used in a ton of places, and in the majority of them they are converted to `std::filesystem::path` anyway. So it seems to me that storing them as such directly would be simpler. This may also need adjustments of some helper functions like `GetClosestUnusedFileName()` to support it, but that should be trivial (in that specific example the use of libobs `os_` APIs could also be substituted with more `std::filesystem` features). Maybe also add an overload to the `ConfigFile` wrapper's `Open()` method that takes a reference to an `std::filesystem::path` object.

PatTheMav:
> There are a number of naïve conversion from `std::string` to `std::filesystem::path`, those should all use `std::filesystem::u8path` to ensure proper conversion from UTF-8, otherwise a native encoding is used that - especially on Windows - may result in invalid paths if non-ASCII characters are used.
> 
> Also see #9183 and [obsproject/obs-websocket#1231](https://github.com/obsproject/obs-websocket/pull/1231)

Oh right, I didn't consider the conversion from string _to_ 

> There are a number of naïve conversion from `std::string` to `std::filesystem::path`, those should all use `std::filesystem::u8path` to ensure proper conversion from UTF-8. Otherwise a native encoding is assumed that - mainly on Windows - may result in invalid paths if non-ASCII characters are used.
> 
> Also see #9183 and [obsproject/obs-websocket#1231](https://github.com/obsproject/obs-websocket/pull/1231)

Good point, if the `std::string` storage is UTF-8-based, the value needs to be converted to the correct encoding of the path `value_type`. 

_Unfortunately_ `u8path` has been deprecated in C++20 already and the approach suggested by the author of the deprecation proposals suggests the following:

```C++
std::string utf8_encoded_filename = ...;
auto char8_view =
    std::ranges::views::transform(utf8_encoded_filename,
                                  [](char c) { return (char8_t)c; });
std::filesystem::path p(char8_view.begin(), char8_view.end());
```

The C++ STL discourages use of UTF-8 encoded text when using `char` types and as such the language wants us to jump through these hoops if we do.

UI/obs-app.hpp
PatTheMav: Those APIs are precisely the reason why they are strings - all these APIs are involved:

* `ConfigFile.open()` (`std::string`) 
* `GetClosestUnusedFileName()` (`std::string`)
* `GetFileSafeName()` (`char *` and `std::string`)
* `os_quick_write_utf8_file` (`char *`)

So in general I opted to keep the strings as such for as long as possible to minimise the impact on existing API functions and convert to paths only when actual file system interaction takes place.

The necessity to either explicitly convert them into a UTF-8 type (or use deprecated UTF-8-specific functions) seems to makes it even more prudent to me to keep using strings for as long as possible and not use `std::filesystem::path` as storage.

derrod:
> Good point, if the `std::string` storage is UTF-8-based, the value needs to be converted to the correct encoding of the path `value_type`.
> 
> _Unfortunately_ `u8path` has been deprecated in C++20 already and the approach suggested by the author of the deprecation proposals suggests the following:
> 
> The C++ STL discourages use of UTF-8 encoded text when using `char` types and as such the language wants us to jump through these hoops if we do.

We disable `char8_t` as of #9194 because it's half-assed and the C++ standards committee should feel bad about introducing it. It's not supported in various STL functions and features (including `std::format`/`std::cin`/`std::cout`), there's no conversion features provided, and it doesn't solve anything while also breaking backwards compatibility, e.g. by changing the `u8` prefix to be `char8_t` instead of `char`. And of course, many third-party libraries such as `nlohmann_json` simply do not support it (or `u8string`s) because nobody actually uses it in their codebases.

PatTheMav:
Pushed an update with the following changes:

* Functions `NewProfile`, `DuplicateProfile` and their variants for scene collections have been renamed to `CreateNewProfile` and `CreateDuplicateProfile` respectively
* File name composition and handling has been revised:
  * If a `std::filesystem::path` exists already, it is used directly
  * If a `std::filesystem::path` needs to be composed of two existing paths, they are both used directly
  * If a `std::filesystem::path` needs to be created of a `std::string` or `char *`, the `u8path` constructor is used
  * If a `std::filesystem::path` needs to be composed of an existing path and a `std::string` or `char *`, the latter are first converted to a path via `u8path` before being used in the composition
  * If a `std::string` or `char *` is required, the existing `std::filesystem::path` is converted into a UTF-8 character array via `u8string` and then either its `c_str()` value is used or it's composed into a new string with other tokens directly (it's faster to convert _just_ the path back to UTF-8 than to convert all string tokens first from UTF-8 to the local file system encoding and then appending all parts)

This probably makes the code less efficient because of more temporary objects being created and (mostly on Windows) more character conversion taking place, but in the greater context of file system operations those losses are probably negligible.

UI/window-basic-main.hpp
RytoEX: I think this was missed in the latest round of revisions.

RytoEX:
Building this on Windows 11 and running with a fresh installation in Portable Mode results in a crash on startup:
image

PatTheMav:
> Building this on Windows 11 and running with a fresh installation in Portable Mode results in a crash on startup: image
> 
> ```
>  	KernelBase.dll!00007ffd125bfabc()	Unknown
>  	vcruntime140d.dll!00007ffc2d8c9362()	Unknown
> >	obs64.exe!std::filesystem::_Throw_fs_error(const char * _Op, const std::error_code & _Error, const std::filesystem::path & _Path1, const std::filesystem::path & _Path2) Line 1875	C++
>  	obs64.exe!std::filesystem::copy(const std::filesystem::path & _From, const std::filesystem::path & _To, const std::filesystem::copy_options _Options) Line 4431	C++
>  	obs64.exe!std::filesystem::copy(const std::filesystem::path & _From, const std::filesystem::path & _To) Line 4440	C++
>  	obs64.exe!OBSApp::MigrateGlobalSettings() Line 885	C++
>  	obs64.exe!OBSApp::InitGlobalConfig() Line 712	C++
>  	obs64.exe!OBSApp::AppInit() Line 1272	C++
>  	obs64.exe!run_program(std::basic_fstream<char,std::char_traits<char>> & logFile, int argc, char * * argv) Line 2107	C++
>  	obs64.exe!main(int argc, char * * argv) Line 2854	C++
>  	obs64.exe!qtEntryPoint() Line 50	C++
>  	obs64.exe!WinMain(HINSTANCE__ * __formal, HINSTANCE__ * __formal, char * __formal, int __formal) Line 60	C++
> ```
> 
> Even copying in some test profiles and scene collections resulted in the same crash on startup.

In `obs-app.cpp` on line 872, change `resize` into `reserve` - does that fix the crash?

RytoEX:
> > Building this on Windows 11 and running with a fresh installation in Portable Mode results in a crash on startup:
> > Even copying in some test profiles and scene collections resulted in the same crash on startup.
> 
> In `obs-app.cpp` on line 872, change `resize` into `reserve` - does that fix the crash?

Yes, that fixes the crash.

Warchamp7 (approved):
Approval from my side pending code review from others.

UI/obs-app.cpp
```diff
@@ -734,120 +689,211 @@ bool OBSApp::UpdatePre22MultiviewLayout(const char *layout)
...
+
+	bool migratedUserSettings =
+		config_get_bool(appConfig, "General", "UserSettingsMigrated");
+
+	if (!migratedUserSettings &&
+	    lastVersion < MAKE_SEMANTIC_VERSION(31, 0, 0)) {
+		bool migrated = MigrateGlobalSettings();
+
+		config_set_bool(appConfig, "General", "UserSettingsMigrated",
+				migrated);
```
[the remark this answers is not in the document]
PatTheMav: The idea is that the "global" settings have been migrated _to_ the "user" settings, which should happen if the last version OBS ran as is lower than whichever version we release this change with (which I'd also need to adapt accordingly).

I might need to give this another pass anyway, because I don't think the migration setting is set on a "fresh" OBS install (as there is nothing to migrate there anyway), but the key idea here is to mark the case of "legacy global config exists and was already successfully migrated".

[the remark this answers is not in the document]
PatTheMav: I've changed this code now a bit - I call the setting `Pre31Migrated` to keep it in line with the other similar settings.

I also check for the older version first (I'd still need to change this to the actual version we target for this change) and for the migration flag second.

This mainly serves as a guard against unnecessary duplicate migrations which will trigger an error message as the copied `user.ini` already exists.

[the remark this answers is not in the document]
RytoEX: Changes look fine at a glance.

RytoEX:
It looks like the `resize`->`reserve` change got reverted and this again crashes on Windows.

PatTheMav:
> It looks like the `resize`->`reserve` change got reverted and this again crashes on Windows.

Pushed an update to fix that.

RytoEX:
> > It looks like the `resize`->`reserve` change got reverted and this again crashes on Windows.
> 
> Pushed an update to fix that.

That fixed the crash again.

RytoEX (approved):
Seems okay.

PR #11056, opened 2024-07-31
cmake: Update use of MbedTLS to support versions 3.6.0 and beyond

PatTheMav:
### Description
Update MbedTLS find module and usage in CMake files to prepare codebase for later update to CMake version 3.6.0 and beyond.

### Motivation and Context
MbedTLS changed a lot of internals with their LTS version 3.6.0, which are incompatible with the find module currently shipped with OBS Studio.

The solution requires several changes to be applied at once:

* Rename the generated target name to MbedTLS::mbedtls to match the name used by MbedTLS' own CMake package
* Update find module to use the updated target name(s)
* Set CMAKE_FIND_PACKAGE_PREFER_CONFIG to TRUE before trying to find MbedTLS to ensure that CMake package files are used with priority (Those are shipped only with MbedTLS 3.6.0 in obs-deps).
* A deprecation warning is emitted if the find module is used with MbedTLS 3.6.0 available

### How Has This Been Tested?
Tested on macOS with either current MbedTLS version provided by `obs-deps` as well as MbedTLS 3.6.0 provided via its own CMake package.

### Types of changes

### Checklist:

### DIFF
--- a/UI/cmake/feature-whatsnew.cmake
+++ b/UI/cmake/feature-whatsnew.cmake
@@ -6,14 +6,16 @@ if(ENABLE_WHATSNEW AND TARGET OBS::browser-panels)
   if(OS_MACOS)
     include(cmake/feature-macos-update.cmake)
   elseif(OS_LINUX)
+    set(CMAKE_FIND_PACKAGE_PREFER_CONFIG TRUE)
     find_package(MbedTLS REQUIRED)
+    set(CMAKE_FIND_PACKAGE_PREFER_CONFIG FALSE)
     find_package(nlohmann_json REQUIRED)
 
     if(NOT TARGET OBS::blake2)
       add_subdirectory("${CMAKE_SOURCE_DIR}/deps/blake2" "${CMAKE_BINARY_DIR}/deps/blake2")
     endif()
 
-    target_link_libraries(obs-studio PRIVATE MbedTLS::MbedTLS nlohmann_json::nlohmann_json OBS::blake2)
+    target_link_libraries(obs-studio PRIVATE MbedTLS::mbedtls nlohmann_json::nlohmann_json OBS::blake2)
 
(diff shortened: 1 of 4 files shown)


tytan652:
3.6.0 caused a breakage (which break RTMPS), until MbedTLS fix their mess with a bugfix release.
I don't think this PR should be "ready for review"

PatTheMav:
> 3.6.0 caused a breakage (which break RTMPS), until MbedTLS fix their mess with a bugfix release. I don't think this PR should be "ready for review"

It is, and also it needs to be merged _before_ we update MbedTLS in `obs-deps` as that will immediately break the current codebase.

Thus we need to merge a forward-compatible change that will work with the current release of `obs-deps` as well as the future release which will be based of 3.6.0 or beyond and which creates CMake packages. There is currently no suggestion that the updated version of MbedTLS will not continue using its updated build system.

tytan652:
I, at least, don't think we should advertise any support of versions `==3.6.0` in any capacity.

RytoEX:
This needs to be rebased and get CI to pass. IIRC, building SRT 1.5.x against mbedTLS 3.5.0+ fails because of changes in mbedTLS that needed addressed in libsrt, but there is no new libsrt tag with the needed changes. That being the case, we currently cannot update to mbedTLS 3.5+, but we can attempt to lay the foundation to allow building against newer versions so we don't have to juggle those changes in the future.

PatTheMav:
Requires a new `obs-deps` release with https://github.com/obsproject/obs-deps/pull/261 merged to fix macOS.

RytoEX (approved):
This seems fine. We can merge this to unblock ourselves for future updates when it's possible for us to update libsrt and MbedTLS without declaring explicit support for MbedTLS 3.6.0+.

PR #11060, opened 2024-08-01
obs-qsv11: Add AV1 Screen Content Coding

thyintel:
### Description

This change adds AV1 SCC for new Intel platforms to improve quality. 
When tested, target usage target usage 4 had an average of a 20% quality gain (up to 40%) and target usage 6 had an average of a 25% quality gain (up to 40%).

### Motivation and Context

This change added quality improvements for future Intel platforms. 

### How Has This Been Tested?

new Intel platforms on various driver versions

### Types of changes

### Checklist:

- [x ] My code has been run through [clang-format](https://github.com/obsproject/obs-studio/blob/master/.clang-format).
- [x ] All commit messages are properly formatted and commits squashed where appropriate.
- [x ] I have included updates to all appropriate documentation.

### DIFF
--- a/plugins/obs-qsv11/QSV_Encoder_Internal.cpp
+++ b/plugins/obs-qsv11/QSV_Encoder_Internal.cpp
@@ -195,6 +195,22 @@ static inline bool HasOptimizedBRCSupport(const mfxPlatform &platform,
 	UNUSED_PARAMETER(rateControl);
 	return false;
 }
+
+static inline bool HasAV1ScreenContentSupport(const mfxPlatform &platform,
+					      const mfxVersion &version)
+{
+#if (MFX_VERSION_MAJOR >= 2 && MFX_VERSION_MINOR >= 12) || MFX_VERSION_MAJOR > 2
+	// Platform enums needed are introduced in VPL version 2.12
+	if ((version.Major >= 2 && version.Minor >= 12) || version.Major > 2)
+		if (platform.CodeName >= MFX_PLATFORM_LUNARLAKE &&
+		    platform.CodeName != MFX_PLATFORM_ALDERLAKE_N &&
+		    platform.CodeName != MFX_PLATFORM_ARROWLAKE)
+			return true;
+#endif
+	UNUSED_PARAMETER(platform);
+	UNUSED_PARAMETER(version);
+	return false;
+}
 PRAGMA_WARN_POP
 
 mfxStatus QSV_Encoder_Internal::InitParams(qsv_param_t *pParams,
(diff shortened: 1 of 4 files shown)


plugins/obs-qsv11/QSV_Encoder_Internal.cpp
```diff
@@ -328,6 +347,24 @@ mfxStatus QSV_Encoder_Internal::InitParams(qsv_param_t *pParams,
 		extendedBuffers.push_back((mfxExtBuffer *)&m_ExtAv1TileParam);
 	}
 
+#if MFX_VERSION_MAJOR >= 2 && MFX_VERSION_MINOR >= 11
+	PRAGMA_WARN_PUSH
+	PRAGMA_WARN_DEPRECATION
```
RytoEX, line 352: Is there a reason this is also checked with preprocessor checks?

plugins/obs-qsv11/QSV_Encoder_Internal.h
```diff
@@ -54,6 +54,7 @@ THEORY OF LIABILITY, WHETHER IN CONTRACT, STRICT LIABILITY, OR TORT
 OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
 */
 #pragma once
+#define ONEVPL_EXPERIMENTAL
```
RytoEX, line 57: Right now in obs-deps, `BUILD_DISPATCHER_ONEVPL_EXPERIMENTAL` is set to `OFF`. Does that matter here with obs-studio as a consumer of VPL via obs-deps?

RytoEX (changes requested):
Had some questions about a few things.

plugins/obs-qsv11/QSV_Encoder_Internal.cpp
thyintel: The preprocessor check is for the m_ExtAV1ScreenContentTools variable. This check will be consistent with the check in the QSV_Encoder_Internal.h where the variable is defined.

thyintel: As well as the MFX_EXT_BUFF_AV1_SCREEN_CONTENT_TOOLS flag

plugins/obs-qsv11/QSV_Encoder_Internal.h
thyintel: It shouldn't need to be turned on. The flag can be kept off for now.

RytoEX: Understood.

plugins/obs-qsv11/QSV_Encoder_Internal.cpp
RytoEX: Got it.

RytoEX:
This will need the merge conflict due to merging #10902 resolved.

plugins/obs-qsv11/QSV_Encoder_Internal.cpp
```diff
@@ -195,6 +195,22 @@ static inline bool HasOptimizedBRCSupport(const mfxPlatform &platform,
 	UNUSED_PARAMETER(rateControl);
 	return false;
 }
+
+static inline bool HasAV1ScreenContentSupport(const mfxPlatform &platform,
+					      const mfxVersion &version)
+{
+#if (MFX_VERSION_MAJOR >= 2 && MFX_VERSION_MINOR >= 12) || MFX_VERSION_MAJOR > 2
+	// AV1_SCREEN_CONTENT_TOOLS API is introduced in VPL version 2.11
+	if ((version.Major >= 2 && version.Minor >= 12) || version.Major > 2)
```
RytoEX, line 204: If this is introduced in 2.11, then shouldn't the version checks check for >= 2.11 instead of 2.12?

RytoEX (changes requested):

plugins/obs-qsv11/QSV_Encoder_Internal.cpp
thyintel: The checks for the AV1_SCEEN_CONTENT_TOOLS are all in version 2.11, but the checks for the platforms (Lunarlake, Alderlake, and Arrowlake) are in 2.12

RytoEX: So `ExtAV1ScreenContentTools` is available in VPL 2.11, but the platform enums are only defined in 2.12 or the runtime requirement is 2.12+?

thyintel: yes, the platform enums are only defined in 2.12, so the runtime requirement is also 2.12

RytoEX: The comment may need to be updated to make that clearer.

thyintel: I moved this specific comment to the AV1 SCC code in the InitParams function and changed the comment here to be more clear.

RytoEX (approved):

PR #10438, opened 2024-03-26
cmake: Replace cmake-format with gersemi for file formatting

PatTheMav:
### Description
Updates the CMake formatter used by the project to replace `cmake-format` with `gersemi`.

### Motivation and Context
`cmake-format` has been abandoned for a few years now and hasn't seen any further development, making it incompatible with recent CMake changes to function calls and syntax.

All legacy files are excluded from formatting checks to avoid unnecessary CI failures.

### How Has This Been Tested?
All CMake files have been successfully formatted using `gersemi`.

### Types of changes
- Code cleanup (non-breaking change which makes code smaller or more readable)

### Checklist:
- [x] My code has been run through [clang-format](https://github.com/obsproject/obs-studio/blob/master/.clang-format).
- [x] I have read the [**contributing** document](https://github.com/obsproject/obs-studio/blob/master/CONTRIBUTING.rst).
- [x] My code is not on the master branch.
- [x] The code has been tested.
- [x] All commit messages are properly formatted and commits squashed where appropriate.
- [x] I have included updates to all appropriate documentation.

PR #11066, opened 2024-08-02
cmake: Update 32-bit project generation on Windows

PatTheMav:
### Description
Updates the way the 32-bit subproject is generated on Windows to create the capture and virtual camera modules for 32-bit host applications.

The update removes the dependency on `legacy_check` as an early entry point and is technically able to handle additional architectures on Windows.

### Motivation and Context
To enable removal of the legacy CMake codepath on Windows, the code to generate the 32-bit subproject cannot rely on `legacy_check` as an always-present entry point to overload the CMake script path with 32-bit specific code.

### How Has This Been Tested?
Tested on Windows 11.

### Types of changes

### Checklist:


RytoEX:
This will need the files reformatted with gersemi now that #10438 has been merged.

shared/obs-shared-memory-queue/shared-memory-queue.c
```diff
@@ -0,0 +1,225 @@
+#include <windows.h>
+#include "shared-memory-queue.h"
+#include "tiny-nv12-scale.h"
+
+#define VIDEO_NAME L"OBSVirtualCamVideo"
```
RytoEX, line 5: Conceptually, this feels a bit weird because `shared-memory-queue` was obviously written to only be used by the Virtual Camera on Windows, but now it's in `shared` and not in a "Windows-only" project/folder. Though, having said that, it looks like win-dshow didn't early return when not targeting Windows either, so maybe that concern is invalid.

RytoEX (commented):
Visually, this seems to make sense. I'll have to do some local test builds. I don't know if anyone has opinions about moving the various sub-projects (d3d8-api, hook-config, inject-library, libdshowcapture, shared-memory-queue, tiny-nv12-scale) to the shared or deps folder, though I believe that's sort of required to create and manage the various CMake targets to make this work, and to enable removing legacy CMake entirely.

PatTheMav:
> At a glance, the main thing that stands out to me as a bit weird is moving libdshowcapture out of win-dshow. I suppose in practice, one isn't building win-dshow without building libobs, and probably one is just building the whole obs-studio repo. I'll probably want @Lain-B 's opinion, but I'll carry on with review/testing.

The main issue is that `libdshowcapture` is for all intents and purposes a 3rd party dependency used by `win-dshow` and the virtual camera module, but for historical reasons one has been architected to be a pseudo-dependency of the other.

It is this pseudo-dependency that has to be broken up to be able to build the virtual camera as its own independent target, because otherwise one would _always_ need to build `win-dshow` (and `libobs` by extension) just to get the virtual camera even though factually it has no code dependencies on either.

There is no real reason why the virtual camera is contained within `win-dshow`, in fact it should be its own "module" or independent target (but it cannot be due to historical reasons and the way it's shipped/managed by OBS code).

PatTheMav:
> Visually, this seems to make sense. I'll have to do some local test builds. I don't know if anyone has opinions about moving the various sub-projects (d3d8-api, hook-config, inject-library, libdshowcapture, shared-memory-queue, tiny-nv12-scale) to the shared or deps folder, though I believe that's sort of required to create and manage the various CMake targets to make this work, and to enable removing legacy CMake entirely.

Correct, it is necessary to fix those pseudo-dependencies (see also the virtual camera) because they are in fact entirely separate binaries/library (they don't have code dependencies on the  `win-capture` module) but have unfortunately been architected in such a way to share bespoke source code and header with the main module.

As such all those sub-projects are very much code shared between modules (exactly what the `shared` directory was created for).

I agree that it's inelegant, but the root cause for that is the way that the modules have been written originally. 

The proper fix for that would be an obs-utilities library that could be built and linked to statically and that contains all common utility code used by `libobs`, runtime modules, and standalone binaries (lots of stuff contained in `libobs`' `utils` directory should end up in there as well).

shared/obs-shared-memory-queue/shared-memory-queue.c
PatTheMav: Yeah the main issue that `shared-memory-queue` is a set of utility functions used by `win-dshow` to communicate with the virtual camera. It is shared by both targets, but the targets have no direct dependency on each other.

So once again that code fits the very definition of "shared" code. See also my comment regarding the "utility library" which would be the correct fix, but lots of Windows code had been written with this implicit sharing of source code and headers across target/project boundaries.

RytoEX:
I did a test build of this on Windows 11 and it seemed to work as expected. The 32-bit/x86 versions of get-graphics-offsets, graphics-hook, inject-helper, obs-vulkan, and obs-virtualcam-module were all built and copied to their correct locations. Would still appreciate knowing if anyone had any strong opinions about this.

Lain-B:
> At a glance, the main thing that stands out to me as a bit weird is moving libdshowcapture out of win-dshow. I suppose in practice, one isn't building win-dshow without building libobs, and probably one is just building the whole obs-studio repo. I'll probably want @Lain-B 's opinion, but I'll carry on with review/testing.

I don't mind moving `libdshowcapture` out of `win-dshow`. It doesn't really make a difference for our sake.

RytoEX:
Do these commits function independently, or does this require a merge commit?

PatTheMav:
> Do these commits function independently, or does this require a merge commit?

Squashed them in a way that should make that possible - the new shared sources are added before, the project changes now include the move of `libdshowcapture`, the shared sources are then removed separately.

RytoEX (approved):
LGTM

PR #11075, opened 2024-08-04
UI: Fix mem leak if --disable-missing-files-check

### LINKED ISSUE #11074: Memory leak when --disable-missing-files-check is used
### Operating System Info

Ubuntu 22.04

### Other OS

_No response_

### OBS Studio Version

30.2.2

_No response_

### OBS Studio Log URL

https://obsproject.com/logs/BvNHbI5BxygZL2lw

### OBS Studio Crash Log URL

_No response_

### Expected Behavior

No memory leak should occur when using the `--disable-missing-files-check` flag.

### Current Behavior

Use of `--disable-missing-files-check` (when there are any missing files) actually does allocate and build the list of missing `files` - the flag just inhibits the display of the dialog - but also accidentally inhibits the release of the `files` list ...
```
Number of memory leaks: 1
```

### Steps to Reproduce

1. Construct a scene with a file reference.
2. Exit OBS.
3. Remove the file.
4. Start OBS with the `--disable-missing-files-check` flag.
5. Exit OBS.  Observe `Number of memory leaks: 1` in the output log.

_No response_

kamalmostafa:
### Description

Make the --disable-missing-files-check flag actually inhibit the collection of a `files` list, not just the display of the missing files dialog.

Fixes a memory leak of the `files` list itself when the flag is used (and there are some missing files).

Closes: obsproject/obs-studio#11074

### Motivation and Context

Fix memory leak: https://github.com/obsproject/obs-studio/issues/11074

### How Has This Been Tested?

Manual testing of consistent behavior.

### Types of changes

### Checklist:

### DIFF
--- a/UI/window-basic-main.cpp
+++ b/UI/window-basic-main.cpp
@@ -1176,6 +1176,9 @@ void OBSBasic::Load(const char *file)
 
 static inline void AddMissingFiles(void *data, obs_source_t *source)
 {
+	if (!data)
+		return;
+
 	obs_missing_files_t *f = (obs_missing_files_t *)data;
 	obs_missing_files_t *sf = obs_source_get_missing_files(source);
 
@@ -1264,7 +1267,10 @@ void OBSBasic::LoadData(obs_data_t *data, const char *file)
 		obs_data_array_push_back_array(sources, groups);
 	}
 
-	obs_missing_files_t *files = obs_missing_files_create();
+	obs_missing_files_t *files = NULL;
+	if (!App()->IsMissingFilesCheckDisabled())
+		files = obs_missing_files_create();
+
 	obs_load_sources(sources, AddMissingFiles, files);
 
 	if (transitions)
@@ -1435,7 +1441,7 @@ retryScene:
 
 	LogScenes();
 
-	if (!App()->IsMissingFilesCheckDisabled())
+	if (files)
 		ShowMissingFilesDialog(files);
 
 	disableSaving--;


PatTheMav (commented):
The bugfix is a bit messy, but that's because of the original implementation is messy to begin with:

* `AddMissingFiles` will automagically call `obs_missing_files_create` when `obs_source_get_missing_files` is called (because that function has two fallback paths that create a new struct)
* `obs_load_sources` is only used in this single instance in the entire application, so its _only_ purpose is to provide the `AddMissingFiles` callback when loading sources
* This suggests that the check for missing files is an architectural fixture, not an "optional" feature (as the startup argument suggests)
* The memory allocated by `obs_missing_files_create` (which is _always_ allocated somehow) is only explicitly released when `ShowMissingFilesDialog` is called:
  * Either no files are missing, so `obs_missing_files_destroy` is called directly
  * The `OBSMissingFiles` instance initialised is deallocated at some point (because it's destructor calls `obs_missing_files_destroy`.

This change only fixes the symptoms of the underlying architectural issue (e.g. the memory is allocated independently from UI functionality and thus should also be released independently from showing any dialogs), but it does avoid the memory leak.

UI/window-basic-main.cpp
```diff
@@ -1176,6 +1176,8 @@ void OBSBasic::Load(const char *file)
 
 static inline void AddMissingFiles(void *data, obs_source_t *source)
 {
+	if (!data)
+		return;
```
WizardCM, line 1180: Code consistency nitpick: please include a newline after the return.

PatTheMav:
Please rebase this PR and give it another pass to make it ready for review/merging again. Cursory check makes this approach preferable to the one in #12628.

PR #11077, opened 2024-08-04
text-freetype: Use the pixel format for reading pixels

kkartaltepe:
### Description

Fixes rendering for terrible bitmap fonts.

### Motivation and Context

Resolves a long standing issue where people replace their normal monospace font with fonts from the 60s. Our code previously only tried this if we asked for 1bit formats, but freetype will do this automatically for 1bit fonts when the glyph doesnt need to be resized. So we need to use the format of the actual resulting rendered bitmap.

### How Has This Been Tested?

Loaded up a sample font from a user and it renders at exact resolutions now.

I have no idea who setup the original 1bit rendering stuff but maybe we should test whatever fonts that was originally setup for.

### Types of changes

### Checklist:

### DIFF
--- a/plugins/text-freetype2/text-functionality.c
+++ b/plugins/text-freetype2/text-functionality.c
@@ -251,10 +251,10 @@ struct glyph_info *init_glyph(FT_GlyphSlot slot, const uint32_t dx,
 	return glyph;
 }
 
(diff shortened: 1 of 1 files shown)


norihiro:
The commit prefix should be `text-freetype2` (`2` is missing).

kkartaltepe:
Thanks, fixed.

